PR #90068, opened 2023-08-05
Fix mono's Wsingle-bit-bitfield-constant-conversion

am11:
Warnings are of the form:

> warning: implicit truncation from 'int' to a one-bit wide bit-field changes value from 1 to -1 [-Wsingle-bit-bitfield-constant-conversion]
 
reported on linux-x64 build (with cc: clang 17 rc1).

PR #91314, opened 2023-08-30
[mono] Fix warning in class-internals.h

akoeplinger:
It was supposed to be fixed in https://github.com/dotnet/runtime/pull/90068 but we should've used `guint` for the one-bit bitfield.


am11 (approved):
LGTM!

lambdageek (approved):

PR #91315, opened 2023-08-30
[browser] Print message arguments in forwarded console

maraf:
- Print arguments for forwarded console messages from browser to apphost console
- It matches the behavior with log to actual browser console

Previous
```
MONO_WASM [0xf390c-main]: Error in bindings_init
MONO_WASM [0xf390c-main]: onRuntimeInitializedAsync() failed
MONO_WASM [0x5f4244-main]: afterLoadWasmModuleToWorker added message event handler
```
New
```
MONO_WASM [0x5f4244-main]: Error in bindings_init Can't find method System.Runtime.InteropServices.JavaScript.JavaScriptExports.InstallSynchronizationContext
MONO_WASM [0x5f4244-main]: onRuntimeInitializedAsync() failed Can't find method System.Runtime.InteropServices.JavaScript.JavaScriptExports.InstallSynchronizationContext
MONO_WASM [0x5f4244-main]: afterLoadWasmModuleToWorker added message event handler {"workerID":1}
```

### DIFF
--- a/src/mono/wasm/host/WasmTestMessagesProcessor.cs
+++ b/src/mono/wasm/host/WasmTestMessagesProcessor.cs
@@ -31,7 +31,7 @@ internal sealed class WasmTestMessagesProcessor
                 logMessage = JsonSerializer.Deserialize<WasmLogMessage>(message);
                 if (logMessage != null)
                 {
-                    line = logMessage.payload + " " + string.Join(" ", logMessage.arguments ?? Enumerable.Empty<string>());
+                    line = logMessage.payload + " " + string.Join(" ", logMessage.arguments ?? Enumerable.Empty<object>());
                 }
                 else
                 {
--- a/src/mono/wasm/host/WasmTestMessagesProcessor.cs
+++ b/src/mono/wasm/host/WasmTestMessagesProcessor.cs
@@ -40,14 +40,15 @@ internal sealed class WasmTestMessagesProcessor
             }
             catch (JsonException)
             {
-                line = message.TrimEnd();
+                line = message;
             }
         }
         else
         {
-            line = message.TrimEnd();
+            line = message;
         }
 
+        line = line.TrimEnd();
         switch (logMessage?.method?.ToLowerInvariant())
         {
             case "console.debug": _logger.LogDebug(line); break;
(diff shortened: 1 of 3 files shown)


pavelsavara (approved):
do we have same problem in xharness ?

maraf:
> do we have same problem in xharness ?

Yes, we do

src/mono/wasm/host/WasmTestMessagesProcessor.cs
```diff
@@ -27,7 +29,14 @@ public void Invoke(string message)
             try
             {
                 logMessage = JsonSerializer.Deserialize<WasmLogMessage>(message);
-                line = logMessage?.payload ?? message.TrimEnd();
```
campersau, line 30: No `TrimEnd` anymore? Maybe it could be done for all after this `if` like:
```js
line = line.TrimEnd()
```

src/mono/wasm/host/WasmTestMessagesProcessor.cs
```diff
@@ -27,18 +29,26 @@ public void Invoke(string message)
...
+                if (logMessage != null)
+                {
+                    line = logMessage.payload + " " + string.Join(" ", logMessage.arguments ?? Enumerable.Empty<object>());
```
radical, line 34: Why `Enumerable.Empty<object>` and not `Enumerable.Empty<string>`? Just curious, shouldn't matter though.

radical (approved):

src/mono/wasm/host/WasmTestMessagesProcessor.cs
radical: I think you can also do `logMessage.arguments ?? []`

kg: `[]` will allocate every time, I think `Array.Empty<>` is what you want

maraf: The arguments can contain any object. At least one call from worker init sends `{worker: 1}` as JS object. So I changed the arguments on apphost side to array of objects

maraf: `Enumerable.Empty` has the same "caching frequently used values" as `Array.Empty` and enumerable is all I need. Is there any other reason to prefere array over enumerable?

kg: Enumerable.Empty is fine too then

PR #91286, opened 2023-08-29
Fix nullable annotation for Validator.TryValidateValue and ValidateValue

jeffhandley:
Fixes #91162

There was an erroneous comment in the `Validator` `TryValidateValue` and `ValidateValue` methods indicating that the `value` could not be null. This comment led to incorrect nullable annotations being applied. Propagation of that surfaced in the new Options source generator.

This removes the erroneous comment, updates the annotation to be more relaxed (non-breaking), and removes the `!` from all call sites in the repo. The intent is to backport this fix to .NET 8 RC2.

Issue #91162, opened 2023-08-26
Incorrect nullable annotation on Validator.TryValidateValue's `value`?


stephentoub:
It's defined like:
```C#
public static bool TryValidateValue(object value, ValidationContext validationContext, ICollection<ValidationResult>? validationResults, IEnumerable<ValidationAttribute> validationAttributes)
```
and the XML doc comment states:
```C#
/// <param name="value">The value to test.  It cannot be null.</param>
```
However, the implementation appears to permit null (the only place it's used is being passed to an `object?` argument in a helper method), the purpose of an attribute like `[Required]` is to validate one isn't null (which suggests null could otherwise be valid, and the new .NET 8 source generator for options is having to emit code like:
```C#
if (!global::System.ComponentModel.DataAnnotations.Validator.TryValidateValue(options.Path!, context, validationResults, validationAttributes))
{
    builder.AddResults(validationResults);
}
```
(note the `!` on the first argument) because the input might be null but the method is defined to not accept null.

cc: @jeffhandley, @eiriktsarpalis

jeffhandley:
I tried doing some archaeology to figure out where the mismatch between the comment and the behavior came from, but I came up empty-handed.

PR #91351, opened 2023-08-30
Fix nullable annotation for Validator.ValidateValue ref source

jeffhandley:
Contributes to #91162. This ref source change was missed in #91286 and caught during the release/8.0 backport review in https://github.com/dotnet/runtime/pull/91293#discussion_r1310601858.


tarekgh (approved):
Sorry I missed that in my original review. Thanks to @stephentoub for catching it!

tarekgh:
@jeffhandley should we mark it with 8.0 milestone instead? just to indicate this change went to 8.0 too.

PR #91355, opened 2023-08-30
Add DebuggerTypeProxy to ListDictionaryInternal

### LINKED ISSUE #91254: Add DebuggerTypeProxy for ListDictionaryInternal
[This PR](https://github.com/dotnet/runtime/pull/91224) added debugger display to `ListDictionaryInternal`. A DebuggerTypeProxy should also be added to improve displaying the collection's contents.

The current display in VS:
image

For example: https://github.com/dotnet/runtime/blob/86252f7b723eeae2a827144ee9c15bc675028ad8/src/libraries/System.Private.CoreLib/src/System/Collections/ArrayList.cs#L2640-L2653

arturek:
The ListDictionaryInternalDebugView class improves the debug view of ListDictionaryInternal instances which are used to implement the Data property of the Exception class.

fixes #91254

The old debugger view:
Screenshot Current

The new debugger view:
Screenshot Updated

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
@@ -10,6 +10,7 @@ namespace System.Collections
     /// Recommended for collections that typically include fewer than 10 items.
     /// </summary>
     [DebuggerDisplay("Count = {count}")]
+    [DebuggerTypeProxy(typeof(ListDictionaryInternalDebugView))]
     [Serializable]
     [TypeForwardedFrom("mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089")]
     // Needs to be public to support binary serialization compatibility
@@ -200,6 +201,17 @@ namespace System.Collections
                 last!.next = node.next;
             }
             count--;
+        }
+
+        internal KeyValuePairs[] ToKeyValuePairsArray()
+        {
+            var array = new KeyValuePairs[count];
+            int index = 0;
+            for (DictionaryNode? node = head; node != null; node = node.next)
+            {
+                array[index++] = new KeyValuePairs(node.key, node.value);
+            }
+            return array;
         }
 
         private sealed class NodeEnumerator : IDictionaryEnumerator
@@ -404,6 +416,20 @@ namespace System.Collections
             public object key = null!;
             public object? value;
             public DictionaryNode? next;
+        }
+
+        internal sealed class ListDictionaryInternalDebugView
+        {
+            private readonly ListDictionaryInternal list;
+
+            public ListDictionaryInternalDebugView(ListDictionaryInternal list)
+            {
+                ArgumentNullException.ThrowIfNull(list);
+                this.list = list;
+            }
+
+            [DebuggerBrowsable(DebuggerBrowsableState.RootHidden)]
+            public KeyValuePairs[] Items => list.ToKeyValuePairsArray();
         }
     }
 }
--- a/src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
@@ -203,7 +203,7 @@ namespace System.Collections
             count--;
         }
 
-        internal KeyValuePairs[] ToKeyValuePairsArray()
+        private KeyValuePairs[] ToKeyValuePairsArray()
         {
             var array = new KeyValuePairs[count];
             int index = 0;
@@ -418,18 +418,18 @@ namespace System.Collections
             public DictionaryNode? next;
         }
 
-        internal sealed class ListDictionaryInternalDebugView
+        private sealed class ListDictionaryInternalDebugView
         {
-            private readonly ListDictionaryInternal list;
+            private readonly ListDictionaryInternal _list;
 
             public ListDictionaryInternalDebugView(ListDictionaryInternal list)
             {
                 ArgumentNullException.ThrowIfNull(list);
-                this.list = list;
+                _list = list;
             }
 
             [DebuggerBrowsable(DebuggerBrowsableState.RootHidden)]
-            public KeyValuePairs[] Items => list.ToKeyValuePairsArray();
+            public KeyValuePairs[] Items => _list.ToKeyValuePairsArray();
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
```diff
@@ -202,6 +203,17 @@ public void Remove(object key)
             count--;
         }
 
+        internal KeyValuePairs[] ToKeyValuePairsArray()
+        {
+            var array = new KeyValuePairs[count];
```
tarekgh, line 208: Is it possible the list can change while executing this? I am asking if we can instead use `List<KVP>`?

tarekgh:
CC @JamesNK if want to have a quick look?

src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
arturek: This code is executed only when the program is suspended by a debugger. Hashtable has similar code.

tarekgh: ok, thanks!

tarekgh (approved):

src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
JamesNK, line 208: `KeyValuePairs` is an unusual type, and the debugger output differs from other dictionaries. It looks like only `Hashtable` uses it, and every other `IDictionary` uses the generic `KeyValuePair<object, object>`.

I think this should use `KeyValuePair<object, object>` to be consistent with the rest of the ecosystem.

JamesNK: This method is only used by the debug view. Can it be moved into the debug view?

stephentoub: This can just be the implementation of the Items property in the debug view; doesn't need its own method.

src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
```diff
@@ -405,5 +417,19 @@ private sealed class DictionaryNode
             public object? value;
             public DictionaryNode? next;
         }
+
+        internal sealed class ListDictionaryInternalDebugView
```
stephentoub, line 421: private?

src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
```diff
@@ -405,5 +417,19 @@ private sealed class DictionaryNode
...
+        {
+            private readonly ListDictionaryInternal list;
```
stephentoub, line 423: _list

tarekgh:
@arturek could you please address the feedback so we can proceed merging it?

src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
arturek: I based my changes on two other non-generic collection classes, `Hashtable` and `SortedList`. To keep code and behaviour consistent with existing examples I used a similar code structure and item type.

Personally, I prefer the way how `KeyValuePairs` presents items (it does it using an additional parameter of `DebuggerDisplay`) but I agree that it would be nice to have greater consistency between generic and non-generic collections.

Maybe instead of modifying this PR, it would be better to make suggested refactorings on all mentioned classes in a separate PR? Otherwise, inconsistencies in behaviour and code between collection classes will remain. But if you think that it is important to modify this PR please let me know and I'll update it.

src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
arturek: done

src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
arturek: done

arturek:
Sorry, I was a little busy. I made some updates and I'll do the rest here or in a separate PR..

src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs
arturek: @JamesNK, @stephentoub I inlined the method as suggested. Earlier I incorrectly interpreted why that method was needed in other places.

@JamesNK I created #91506 to discuss how to display dictionaries.

tarekgh:
@arturek could you please agree to the license agreement CLA according to https://github.com/microsoft/contributorlicenseagreement?

stephentoub (approved):

arturek:
@microsoft-github-policy-service agree

tarekgh:
@arturek looks one more step to sign the CLA. Thanks for your patience. Somehow the following post gets removed after it shows up for a few seconds. CC @dotnet/dnceng 

image

garath:
> Somehow the following post gets removed after it shows up for a few seconds. CC dotnet/dnceng

Alas, dnceng does not manage the CLI bot. For support, check with the [OSPO](https://docs.opensource.microsoft.com/tools/cla/). 

It looks like dotnet is running its own instance of Microsoft CLA tool (notice the different bot name in the screen shot). Be sure to follow those instructions for dotnet. 

(Perhaps there is some conflict between the Microsoft and Dotnet bots that is causing the post to disappear?)

tarekgh:
@arturek could you please finish the CLA agreement to proceed merging it?

arturek:
@tarekgh I don't know how I can do it. I don't see the comment from a license bot anymore, and adding just my agree comment doesn't seem to work.

tarekgh:
@arturek did you see my comment? need to use @dotnet-policy-service instead @microsoft-github-policy-service to agree to the CLA.

arturek:
@dotnet-policy-service agree

arturek:
Thank you, @tarekgh . I haven't noticed that it has a different version.

tarekgh:
Thanks @arturek for providing the change and be patient :-)

PR #91284, opened 2023-08-29
Start sharing code between ILLink/ILCompiler test infra

sbomer:
This creates a new directory `Mono.Linker.Tests.Shared` with sources that are shared between the test infra of ILLink/ILCompiler. This starts by sharing most of the code in `TestRunner.cs`, renaming some types to make the sharing easier.

PR #91358, opened 2023-08-30
Share more code between ILLink/ILCompiler test infra

sbomer:
Uses the code sharing infra added in https://github.com/dotnet/runtime/pull/91284. This shares most of the logic that was easily shareable in the `TestCasesRunner` directory.


src/tools/illink/test/Mono.Linker.Tests.Shared/AssemblyChecker.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using Mono.Cecil;
+
+namespace Mono.Linker.Tests.TestCasesRunner
+{
+	public partial class AssemblyChecker
+	{
+		private readonly AssemblyDefinition originalAssembly;
+
+		static bool IsCompilerGeneratedMemberName (string memberName)
```
vitek-karas, line 12: Maybe we should consider moving these into a separate utils extension method class - they're not really related to the assembly checker as such, they're more general use methods (even if they're not currently used as such).

vitek-karas (approved):

src/tools/illink/test/Mono.Linker.Tests.Shared/AssemblyChecker.cs
vitek-karas: OK - makes sense

PR #91180, opened 2023-08-27
Update binder gen emitted-interceptor nullability to match framework impl

layomia:
Addresses https://github.com/dotnet/runtime/issues/90987. RC 2 candidate.

PR #91359, opened 2023-08-30
[release/8.0] Update binder gen emitted-interceptor nullability to match framework impl

### LINKED ISSUE #90987: Config generator doesn't match nullability semantics of built-in Bind
The built-in Bind method annotates the object as nullable:
```C#
public static void Bind(this IConfiguration configuration, object? instance)
```
and then explicitly nops if it's null:
```C#
if (instance != null)
{
    ...
}
```
In contrast, the code generated by the generator annotates it as non-nullable and throws an exception if it's null, changing the semantics when using the source generator.
```C#
if (obj is null)
{
    throw new ArgumentNullException(nameof(obj));
}
```
(It also names the parameter `obj` instead of `instance`.)

layomia:
Backport of #91180 to release/8.0
Fixes #90987 for 8.0.

## Customer Impact

Updates null input handling by the generated replacement binding methods to align with the reflection-based implementation. This follows the general user expectation of consistency between the two implementations.

## Testing

Automated null handling tests have been added for all parameters of all binder methods. They assert the same behavior for source-gen and reflection. Some null refs in the reflection binder were fixed as well. This also provides complete build coverage for all binding inputs.

## Risk

Low. It's a contained fix for an off-by-default component.

cc @carlossanlop.

### DIFF
--- a/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Helpers/Emitter/OptionsConfigurationServiceCollectionExtensions.cs
+++ b/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Helpers/Emitter/OptionsConfigurationServiceCollectionExtensions.cs
@@ -24,12 +24,12 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
             private void EmitConfigureMethods()
             {
                 const string defaultNameExpr = "string.Empty";
-                string configParam = $"{Identifier.IConfiguration} {Identifier.configuration}";
+                string configParam = $"{Identifier.IConfiguration} {Identifier.config}";
 
                 if (ShouldEmitMethods(MethodsToGen_Extensions_ServiceCollection.Configure_T))
                 {
                     EmitStartMethod(MethodsToGen_Extensions_ServiceCollection.Configure_T, configParam);
-                    _writer.WriteLine($"return {Identifier.Configure}<{Identifier.TOptions}>({Identifier.services}, {defaultNameExpr}, {Identifier.configuration}, {Identifier.configureOptions}: null);");
+                    _writer.WriteLine($"return {Identifier.Configure}<{Identifier.TOptions}>({Identifier.services}, {defaultNameExpr}, {Identifier.config}, {Identifier.configureOptions}: null);");
                     EmitEndBlock();
                 }
 
@@ -38,7 +38,7 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
                     EmitStartMethod(
                         MethodsToGen_Extensions_ServiceCollection.Configure_T_name,
                         paramList: $"string? {Identifier.name}, " + configParam);
-                    _writer.WriteLine($"return {Identifier.Configure}<{Identifier.TOptions}>({Identifier.services}, {Identifier.name}, {Identifier.configuration}, {Identifier.configureOptions}: null);");
+                    _writer.WriteLine($"return {Identifier.Configure}<{Identifier.TOptions}>({Identifier.services}, {Identifier.name}, {Identifier.config}, {Identifier.configureOptions}: null);");
                     EmitEndBlock();
                 }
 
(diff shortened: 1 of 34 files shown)


tarekgh (approved):

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Helpers/Emitter/OptionsConfigurationServiceCollectionExtensions.cs
```diff
@@ -47,22 +47,20 @@ private void EmitConfigureMethods()
...
-                    return {{Identifier.services}}.{{Identifier.AddSingleton}}<{{optionsNamespaceName}}.IConfigureOptions<{{Identifier.TOptions}}>>(new {{optionsNamespaceName}}.ConfigureNamedOptions<{{Identifier.TOptions}}>({{Identifier.name}}, {{Identifier.obj}} => {{nameof(MethodsToGen_CoreBindingHelper.BindCoreMain)}}({{Identifier.configuration}}, {{Identifier.obj}}, typeof({{Identifier.TOptions}}), {{Identifier.configureOptions}})));
+                    {{Identifier.services}}.{{Identifier.AddSingleton}}<{{Identifier.IOptionsChangeTokenSource}}<{{Identifier.TOptions}}>>(new {{Identifier.ConfigurationChangeTokenSource}}<{{Identifier.TOptions}}>({{Identifier.name}}, {{Identifier.config}}));
+                    return {{Identifier.services}}.{{Identifier.AddSingleton}}<IConfigureOptions<{{Identifier.TOptions}}>>(new ConfigureNamedOptions<{{Identifier.TOptions}}>({{Identifier.name}}, {{Identifier.instance}} => {{nameof(MethodsToGen_CoreBindingHelper.BindCoreMain)}}({{Identifier.config}}, {{Identifier.instance}}, typeof({{Identifier.TOptions}}), {{Identifier.configureOptions}})));
```
jeffhandley, line 63: Why isn't the `optionsNamespace` needed anymore?

src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
```diff
@@ -305,125 +312,122 @@ private static void BindProperty(PropertyInfo property, object instance, IConfig
...
+                // -----------------------------------------------------------------------------------------------------------------------------
+                //                  |  bindingPoint |  bindingPoint |
+                //     Interface    |     Value     |   IsReadOnly  |  Behavior
+                // -----------------------------------------------------------------------------------------------------------------------------
+                //  ISet<T>         |   not null    |  true/false   | Use the Value instance to populate the configuration
+                //  ISet<T>         |     null      |     false     | Create HashSet<T> instance to populate the configuration
+                //  ISet<T>         |     null      |     true      | nothing
+                //  IReadOnlySet<T> | null/not null |     false     | Create HashSet<T> instance, copy over existing values, and populate the configuration
+                //  IReadOnlySet<T> | null/not null |     true      | nothing
+                // -----------------------------------------------------------------------------------------------------------------------------
```
jeffhandley, line 338: These comments with the table of scenarios is really helpful

src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
```diff
@@ -305,125 +312,122 @@ private static void BindProperty(PropertyInfo property, object instance, IConfig
...
+                //  IReadOnlyDictionary<T> | null/not null |     true      | nothing
+                // -----------------------------------------------------------------------------------------------------------------------------
+                if (TypeIsADictionaryInterface(type))
+                {
+                    if (!bindingPoint.IsReadOnly || bindingPoint.Value is not null)
                     {
-                        // if the binding point doesn't let us set a new instance, there's nothing more we can do
-                        if (bindingPoint.IsReadOnly)
+                        object? newValue = BindDictionaryInterface(bindingPoint.Value, type, config, options);
+                        if (!bindingPoint.IsReadOnly && newValue != null)
```
jeffhandley, line 368: Nit/question (not blocking): Would `is not null` be preferred here? `is (not) null` is used through most of this code, but there are some `!= null`s mixed in.

jeffhandley (approved):
This is approved for RC2 as it improves the quality of a new .NET 8 feature. I had a couple non-blocking questions in the review.

src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
layomia: Yup!
Don't mean to take credit here... the diff to this file is purely indentation due to removing the unneeded `if (config != null)` check above.

src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
layomia: Per https://github.com/dotnet/runtime/pull/91359#discussion_r1312214668, we could def clean up aspects of the reflection code later.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Helpers/Emitter/OptionsConfigurationServiceCollectionExtensions.cs
layomia: This was missed in the interceptors change, where we switched away from using fully qualified names for the top level replacement binders (emitted in global namespace). With the Roslyn feature, we now emit all the code within a namespace (same one needed for the granular interceptors opt-in btw), and emit the `usings` within the namespace.

src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
layomia: Yeah the source gen impl followed this closely.

Issue #91073, opened 2023-08-24
Options generator perf improvements


stephentoub:
The options generator is emitting code like this:
```C#
            var baseName = (string.IsNullOrEmpty(name) ? "FirstModel" : name) + ".";
            var builder = new global::Microsoft.Extensions.Options.ValidateOptionsResultBuilder();
            var context = new global::System.ComponentModel.DataAnnotations.ValidationContext(options);
            var validationResults = new global::System.Collections.Generic.List<global::System.ComponentModel.DataAnnotations.ValidationResult>();
            var validationAttributes = new global::System.Collections.Generic.List<global::System.ComponentModel.DataAnnotations.ValidationAttribute>(1);

            context.MemberName = "P1";
            context.DisplayName = baseName + "P1";
            validationAttributes.Add(global::__OptionValidationStaticInstances.__Attributes.A3);
            if (!global::System.ComponentModel.DataAnnotations.Validator.TryValidateValue(options.P1!, context, validationResults, validationAttributes))
            {
                builder.AddResults(validationResults);
            }

            context.MemberName = "P2";
            context.DisplayName = baseName + "P2";
            validationResults.Clear();
            validationAttributes.Clear();
            validationAttributes.Add(global::__OptionValidationStaticInstances.__Attributes.A4);
            if (!global::System.ComponentModel.DataAnnotations.Validator.TryValidateValue(options.P2!, context, validationResults, validationAttributes))
            {
                builder.AddResults(validationResults);
            }

            return builder.Build();
```

1. There's a concatenation string allocation that results as part of creating `baseName`, even in the common case where name is null and both things being concatenated are literals.  There's then another concatenation allocation per member, as the implementation tacks on a literal suffix.  Instead of `var baseName = (string.IsNullOrEmpty(name) ? "FirstModel" : name) + ".";`, that line should be deleted, the each display name creation should instead be like: `context.DisplayName = string.IsNullOrEmpty(name) ? "FirstModel.P2" : name + ".P2";`. That way, when the name is null or empty, there are zero string allocations, and when the name is non-null and non-empty, it saves the one intermediate allocation.
2. `ValidateOptionsResultBuilder` is a class and is being allocated even when there aren't any validation errors.  We should instead lazily allocate it only when one of the TryValidateValue calls returns false.

layomia:
cc @tarekgh.

tarekgh:
This is fixed by https://github.com/dotnet/runtime/pull/91432

Issue #90990, opened 2023-08-23
Options source generator shouldn't use random names


eerhardt:
https://github.com/dotnet/runtime/blob/3ef2c804b4fea8c94c9bd5cc39f79525361a03d2/src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs#L40-L42

Generated code should be deterministic. Many companies and orgs have requirements around deterministic builds.

              Is this the code that prompted https://github.com/dotnet/roslyn-analyzers/pull/6882/? Should we change this code to not use Random, and instead be deterministic?

cc @jaredpar

_Originally posted by @eerhardt in https://github.com/dotnet/runtime/pull/89148#discussion_r1303095388_

tarekgh:
@jaredpar what is the suggestion here? We need to have a class name that is global internal and not conflict when two independent compilations using the generator and one of them is internal visible to the other one. The source gen supports Lang versions 8+. We already detect if using Lang Version 11+ we use `file` keyword and not using random at all.

stephentoub:
> what is the suggestion here?

Incorporate a hash of more data, like the assembly's name.

Or use C#11 and/or don't use InternalsVisibleTo.

tarekgh:
> Or use C#11 and/or don't use InternalsVisibleTo.

I want to avoid restricting the language version only because of that. I am not sure what do you mean by `don't use InternalsVisibleTo.`? It is the user who is using it and not the generator. Do you mean we shouldn't support this scenario? This is a real scenario we got from R9.

> Incorporate a hash of more data, like the assembly's name.

This looks like a reasonable idea to try. Thanks!

jaredpar:
> I want to avoid restricting the language version only because of that.

I would start with requiring C# 11 and wait to see if users push back. Other generators like the regex generator have set  [that standard](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.RegularExpressions/gen/RegexGenerator.cs#L306) and, to my understanding at least, that has not been a problem thus far. I would not invest in making my generator more complex to support this scenario unless I had data saying that would be valuable.

tarekgh:
@geeknoid @stephentoub @eerhardt would you be ok if we require Language version 11+ for the options source gen? 

CC @ericstj

stephentoub:
> . I am not sure what do you mean by don't use InternalsVisibleTo.? It is the user who is using it and not the generator. Do you mean we shouldn't support this scenario?

[...]

PR #91363, opened 2023-08-30
Options Source Gen Fixes

tarekgh:
Planning to port these fixes to rc2

The changes in this PR fix the following issues:

- Remove using `Random` in the source generator and instead use the module name hash. https://github.com/dotnet/runtime/issues/90990
- Address the optimization requested in the issue: https://github.com/dotnet/runtime/issues/91073.
- Detection when need to emit the nullable `!` when calling `Validator.TryValidateValue`
- Allow generating code for properties of type `IEnumerable<T>` and tagged with the attribute `ValidateEnumeratedItems`

### DIFF
--- a/src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
+++ b/src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Extensions.Options.Generators
             else
             {
                 _modifier = "internal";
-                string suffix = $"_{GetNonRandomizedHashCode($"{compilation.SourceModule.Name}"):X8}";
+                string suffix = $"_{GetNonRandomizedHashCode(compilation.SourceModule.Name):X8}";
                 _staticValidationAttributeHolderClassName += suffix;
                 _staticValidatorHolderClassName += suffix;
             }
--- a/src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
+++ b/src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Extensions.Options.Generators
         }
 
         /// <summary>
-        /// Returns the nullable annotation string to use in the code generation according to the first first parameter of
+        /// Returns the nullable annotation string to use in the code generation according to the first parameter of
         /// <see cref="System.ComponentModel.DataAnnotations.Validator.TryValidateValue(object, ValidationContext, ICollection{ValidationResult}, IEnumerable{ValidationAttribute})"/> is nullable annotated.
         /// </summary>
         /// <param name="compilation">The <see cref="Compilation"/> to consider for analysis.</param>
--- a/src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
+++ b/src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
@@ -3,10 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Linq;
 using System.Threading;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.DotnetRuntime.Extensions;
 
 namespace Microsoft.Extensions.Options.Generators
 {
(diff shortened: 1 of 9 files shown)


tarekgh:
CC @jeffhandley @geeknoid

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
```diff
@@ -161,6 +190,7 @@ private void GenModelSelfValidationIfNecessary(ValidatedModel modelToValidate)
         {
             if (modelToValidate.SelfValidates)
             {
+                OutLn($"builder ??= new global::Microsoft.Extensions.Options.ValidateOptionsResultBuilder();");
                 OutLn($"builder.AddResults(((global::System.ComponentModel.DataAnnotations.IValidatableObject)options).Validate(context));");
```
stephentoub, line 194: What you have is fine. It could also be:
```C#
OutLn($"(builder ??= new()).AddResults(((global::System.ComponentModel.DataAnnotations.IValidatableObject)options).Validate(context));");
```

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
```diff
@@ -239,8 +268,9 @@ private void GenMemberValidation(ValidatedMember vm, ref Dictionary<string, Stat
                 OutLn($"validationAttributes.Add({_staticValidationAttributeHolderClassFQN}.{staticValidationAttributeInstance.FieldName});");
             }
 
-            OutLn($"if (!global::System.ComponentModel.DataAnnotations.Validator.TryValidateValue(options.{vm.Name}!, context, validationResults, validationAttributes))");
+            OutLn($"if (!global::System.ComponentModel.DataAnnotations.Validator.TryValidateValue(options.{vm.Name}{_TryGetValueNullableAnnotation}, context, validationResults, validationAttributes))");
             OutOpenBrace();
+            OutLn($"builder ??= new global::Microsoft.Extensions.Options.ValidateOptionsResultBuilder();");
             OutLn($"builder.AddResults(validationResults);");
```
stephentoub, line 274: Same for all of these.  It could (but doesn't have to be) simplified to:
```C#
OutLn($"(builder ??= new()).AddResults(validationResults);");
```

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
```diff
@@ -405,5 +445,19 @@ private StaticFieldInfo GetOrAddStaticValidator(ref Dictionary<string, StaticFie
...
+        /// </summary>
+        internal static int GetNonRandomizedHashCode(string s)
+        {
+            uint result = 2166136261u;
+            foreach (char c in s)
+            {
+                result = (c ^ result) * 16777619;
+            }
+            return Math.Abs((int)result);
+        }
```
stephentoub, line 461: We might want to consider using SHA256 here instead of FNV-1a. I _think_ that's what the compiler itself uses when it needs uniqueness.

stephentoub (approved):
Nice. Thanks.

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
tarekgh: After offline discussion, we decided to stick with `FNV-1a`.

tarekgh:
Failing tests tracked by https://github.com/dotnet/runtime/issues/91410

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
```diff
@@ -65,6 +69,31 @@ public Emitter(Compilation compilation, bool emitPreamble = true) : base(emitPre
             return Capture();
         }
 
+        /// <summary>
+        /// Returns the nullable annotation string to use in the code generation according to the first first parameter of
```
jeffhandley, line 73: Nit: "...the first first parameter..."

jeffhandley (approved):

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
tarekgh: I am duplicating it to ensure users will read it :-) just kidding. I'll fix it.

PR #91371, opened 2023-08-31
Allow superpmi.exe to find a default JIT

BruceForstall:
This makes it slightly easier to invoke superpmi.exe manually: most of the time you want to use the clrjit.dll (or similar) in the same directory where superpmi.exe lives (in a Core_Root directory), so use that if not JIT is specified.

### DIFF
--- a/src/coreclr/tools/superpmi/superpmi/commandline.cpp
+++ b/src/coreclr/tools/superpmi/superpmi/commandline.cpp
@@ -12,6 +12,7 @@
 #include "mclist.h"
 #include "methodcontext.h"
 #include "logging.h"
+#include "spmiutil.h"
 
 // NOTE: this is parsed by parallelsuperpmi.cpp::ProcessChildStdOut() to determine if an incorrect
 // argument usage error has occurred.
@@ -21,8 +22,8 @@ void CommandLine::DumpHelp(const char* program)
 {
     printf("%s\n", g_SuperPMIUsageFirstLine);
     printf("\n");
-    printf("Usage: %s [options] jitname [jitname2] filename.mc\n", program);
-    printf(" jitname" PLATFORM_SHARED_LIB_SUFFIX_A " - path of jit to be tested\n");
+    printf("Usage: %s [options] [jitname] [jitname2] filename.mc\n", program);
+    printf(" jitname" PLATFORM_SHARED_LIB_SUFFIX_A " - optional path of jit to be tested (default is JIT in same directory as %s)\n", program);
     printf(" jitname2" PLATFORM_SHARED_LIB_SUFFIX_A " - optional path of second jit to be tested\n");
     printf(" filename.mc - load method contexts from filename.mc\n");
     printf(" -j[it] Name - optionally -jit can be used to specify jits\n");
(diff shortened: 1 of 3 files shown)


BruceForstall:
@jakobbotsch PTAL
cc @dotnet/jit-contrib

src/coreclr/tools/superpmi/superpmi/commandline.cpp
```diff
@@ -669,12 +659,164 @@ bool CommandLine::Parse(int argc, char* argv[], /* OUT */ Options* o)
             return false;
         }
     }
+
+    SPMI_TARGET_ARCHITECTURE DefaultSpmiTargetArchitecture = GetSpmiTargetArchitecture();
```
jakobbotsch, line 663: Wrong naming conventions (here and below)

src/coreclr/tools/superpmi/superpmi/commandline.cpp
```diff
@@ -669,12 +659,164 @@ bool CommandLine::Parse(int argc, char* argv[], /* OUT */ Options* o)
...
+#if defined(HOST_UNIX)
+        host_os_tag = "unix";
+        jit_host_os_prefix = "lib";
+        jit_host_os_extension = ".so";
+#elif defined(HOST_WINDOWS)
+        host_os_tag = "win";
+        jit_host_os_extension = ".dll";
+#else
+        // TODO: handle Mac
+        allowDefaultJIT = false;
```
jakobbotsch, line 692: Needs to be addressed? Presumably it's just .dylib?

jakobbotsch (commented):
Seems fine to me, but looks like there are build failures.

src/coreclr/tools/superpmi/superpmi/commandline.cpp
```diff
@@ -669,12 +659,170 @@ bool CommandLine::Parse(int argc, char* argv[], /* OUT */ Options* o)
...
+
+    if (o->nameOfJit == nullptr)
+    {
+        // Try to find a JIT in the same directory as superpmi.exe. If found (based on targetArchitecture), use it.
+
+        bool allowDefaultJIT = true;
+
+        const char* host_os_tag = "";
+        const char* jit_host_os_prefix = "";
+        const char* jit_host_os_extension = "";
```
jakobbotsch, line 681: Wrong naming convention

src/coreclr/tools/superpmi/superpmi/commandline.cpp
```diff
@@ -669,12 +659,170 @@ bool CommandLine::Parse(int argc, char* argv[], /* OUT */ Options* o)
...
+                sprintf_s(tempStr, len, "%s%c%s%s_%s_%s_%s%s",
+                    programPath,
+                    DIRECTORY_SEPARATOR_CHAR_A,
+                    jitHostOSPrefix,
+                    jitBaseName,
+                    jitOSName,
+                    targetArch,
+                    hostArch,
+                    jitHostOSExtension);
+            }
```
jakobbotsch, line 798: Is there any reason to not just use the "default" JIT (clrjit.dll/libclrjit.so/libclrjit.dylib)?

jakobbotsch (approved):

src/coreclr/tools/superpmi/superpmi/commandline.cpp
BruceForstall: We actually do use the default. Only if you use the `-target` option do the non-default ones get used.

PR #91374, opened 2023-08-31
Fix runtime dispatch to static virtuals on interface types

### LINKED ISSUE #90333: NativeAOT is crashing on static interface method
### Description

On .NET 8 Preview 7.

Repro:

```csharp
using System.Reflection;

IUnknown obj = new Impl();
if (obj.QueryInterface<IBar>(out var bar)) bar.Bar();
if (obj.QueryInterface<IFoo>(out var foo)) foo.Foo();

interface IComInterface
{
    abstract static Guid IID { get; }
}

interface IUnknown : IComInterface
{
    static Guid IComInterface.IID => new("00000000-0000-0000-C000-000000000046");
    static Guid GetIID<TVtbl>() where TVtbl : IComInterface => TVtbl.IID;

    [UnconditionalSuppressMessage("ILLink", "IL2111")]
    [UnconditionalSuppressMessage("ILLink", "IL2075")]
    bool QueryInterface<
        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.Interfaces | DynamicallyAccessedMemberTypes.NonPublicProperties)] T
        >([NotNullWhen(true)] out T? result) where T : IComInterface
    {
        foreach (var p in GetType()
            .GetInterfaces()
            .SelectMany(([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.NonPublicProperties)] i)
                => i.GetProperty("IComInterface.IID", BindingFlags.NonPublic | BindingFlags.Static)
                    is PropertyInfo propertyInfo ? [propertyInfo] : Array.Empty<PropertyInfo>()))
        {
            if (p.GetValue(this) is Guid guid && guid == GetIID<T>())
            {
                result = (T)this;
                return true;
            }
        }
        result = default;
        return false;
    }
}

interface IFoo : IComInterface, IUnknown
{
    static Guid IComInterface.IID => new("12345678-1234-1234-1234-1234567890AB");
    void Foo();
}

interface IBar : IComInterface, IUnknown
{
    static Guid IComInterface.IID => new("12345678-1234-1234-1234-1234567890CD");
    void Bar();
}

interface IBaz : IComInterface, IFoo, IBar
{
    static Guid IComInterface.IID => new("12345678-1234-1234-1234-1234567890EF");
}

partial struct Impl : IBaz
{
    public void Foo() => Console.WriteLine("foo");
    public void Bar() => Console.WriteLine("bar");
}
```

### Reproduction Steps

Build the source with NativeAOT and run it.

### Expected behavior

No crashing.

### Actual behavior

Crashing.

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

MichalStrehovsky:
Fixes #90333.

We were not generating information about static virtuals on interface types. Information about default interface methods normally goes to the class, but if the T we're dispatching on is an interface, this information wasn't generated. The fix is to put this information into dispatch maps and sealed vtables, same way we do for classes.

The test shows what the problem is - if we change `IBar` to be a class, things would work even before this PR.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/tools/Common/TypeSystem/Common/MetadataVirtualMethodAlgorithm.cs
+++ b/src/coreclr/tools/Common/TypeSystem/Common/MetadataVirtualMethodAlgorithm.cs
@@ -614,6 +614,8 @@ namespace Internal.TypeSystem
         {
             Debug.Assert(!interfaceMethod.Signature.IsStatic);
 
+            // This would be a default interface method resolution. The algorithm below would sort of work, but doesn't handle
+            // things like diamond cases and it's better not to let it resolve as such.
             if (currentType.IsInterface)
                 return null;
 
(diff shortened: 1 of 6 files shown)


jkotas (approved):
Thanks

Issue #91381, opened 2023-08-31
Emit custom modifiers into native layout metadata


MichalStrehovsky:
When we started generating custom modifiers into metadata format to support new function pointer APIs in #85504, we should have also added it to native layout format. We currently have a mismatch. It's likely generic virtual dispatch and LDTOKEN is going to be wrong if we're in a situation where two method signatures only differ by modifiers.

PR #85504, opened 2023-04-28
Emission and reading of custom modifiers

MichalStrehovsky:
Fixes #71883.

* Emit custom modifiers to the native metadata format
* Read custom modifiers at runtime
* Bugfix in the type loader

PR #91382, opened 2023-08-31
Fix LDTOKEN of methods that have modifiers

### LINKED ISSUE #91065: native aot regression in .net 8.0
### Description

i am using this [library](https://github.com/Handlebars-Net/Handlebars.Net) to generate strings with templates. it can work fine using JsonDocument as the model under native aot with .net 7. however, somehow when i upgraded my dotnet version to 8.0 it breaks with `System.ArgumentException: The handle is invalid.`. i know that this library does not support native aot directly but im sure this is a regression bug in .net

### Reproduction Steps

[nativeaot.zip](https://github.com/dotnet/runtime/files/12431092/nativeaot.zip)

- first, compile with `dotnet publish -c Release` and run the generated binary.
- change the `TargetFramework` to `net7.0` to make it work.

### Expected behavior

works on net 8.0

### Actual behavior

crashed under native aot

### Regression?

net 7.0

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

MichalStrehovsky:
Fixes #91065.

When we started generating custom modifiers into metadata format to support new function pointer APIs in #85504, we should have also added it to native layout format. We currently have a mismatch. This is a low risk bugfix to ignore modifiers on the metadata side that we can take to 8.0. We'll want to do a full fix to actually emit and compare this. Tracked in a .NET 9 bug at #91381.

No regression test because I spent too much time being puzzled at why https://github.com/Handlebars-Net/Handlebars.Net/blob/50614fd844e5360eb10e76154aa74da4d7bf12ce/source/Handlebars/Helpers/IHelperDescriptor.cs#L13 is generated as a custom modifier (`[in] !TOptions& modreq([netstandard]System.Runtime.InteropServices.InAttribute) options`) whereas if I do it, I get `[in] !T& 'value'` with a custom attribute. We'll want to write a proper set of tests with ambiguities for the bug I opened anyway.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/TypeLoaderEnvironment.MetadataSignatureParsing.cs
+++ b/src/coreclr/nativeaot/System.Private.TypeLoader/src/Internal/Runtime/TypeLoader/TypeLoaderEnvironment.MetadataSignatureParsing.cs
@@ -173,12 +173,23 @@ namespace Internal.Runtime.TypeLoader
 
         private bool CompareTypeSigWithType(ref NativeParser parser, TypeManagerHandle moduleHandle, Handle typeHandle)
         {
-            while (typeHandle.HandleType == HandleType.TypeSpecification)
+            while (typeHandle.HandleType == HandleType.TypeSpecification
+                || typeHandle.HandleType == HandleType.ModifiedType)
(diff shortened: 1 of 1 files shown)


jkotas (approved):
Thanks

PR #91383, opened 2023-08-31
Remove public provider from rundown session

### LINKED ISSUE #90575: EventPipe logs unintended events during rundown
When capturing an EventPipe trace configured to only capture informational level events, unintended verbose events are incorrectly captured as well. 

## Repro

1. Write a C# app that does this and run it:
```C#
{
    Task.Run(() =>
    {
        List<object> list = new List<object>();
        while (true)
        {
            list.Add(new object());
            list.Clear();
        }
    });
    Console.WriteLine("Hit enter to exit");
    Console.ReadLine();
}
```

2. Run dotnet-trace (replace `ConsoleApp33` with the name of your app):
```
dotnet-trace collect -n ConsoleApp33 --clreventlevel informational --clrevents gc
```

3. Once the trace has run for a few seconds, hit Enter to stop the trace and let dotnet-trace gracefully complete.
4. Open the trace in PerfView and look for the presence of the GC/AllocationTick event in the Events or EventStats windows.

Expected behavior: No such events should be found because they are verbose events
Actual behavior: Some number of these events will be present. In my attempt it was 100 of them but I'd expect the number to vary based on timing.

## Affected versions

I reproduced this with on a x64 Windows machine and the .NET 8 Preview 7 build of the runtime:
> dotnet --version
8.0.100-preview.7.23376.3
> dotnet-trace --version
7.0.430602+c9c987585e1fa9a8a8a5e1ffb60a70f85b0da4ca

However investigating the code below I expect the problem is likely to reproduce on any OS and architecture and probably several past major versions of .NET Core, perhaps all of them.

## Likely root cause

Neither [write_event_2](https://github.com/dotnet/runtime/blob/main/src/native/eventpipe/ep.c#L772) nor [ep_session_write_event](https://github.com/dotnet/runtime/blob/main/src/native/eventpipe/ep-session.c#L523) filters out events which are written by a non-rundown thread when a session is in rundown mode. During rundown it is not sufficient to rely on keyword and level filtering alone because the keywords and levels used for rundown events are also (unfortunately) shared with non-rundown events. In particular the GC keyword is enabled during rundown at the verbose level, but other non-rundown threads may be concurrently emitting AllocTick or other GC verbose events. The events coming from these non-rundown threads need to be filtered out.

## Problem Investigation

I set up the repro app as above and attached VS debugger for native code before running dotnet-trace. Then I set a conditional breakpoint at https://github.com/dotnet/runtime/blob/main/src/native/eventpipe/ep-buffer-manager.c#L946 with condition 
`ep_event->level == 5`.

Once the debugger is ready, run steps (2) and (3) of the repro which caused the breakpoint to be hit. The first time I saw a normal rundown event that happened to have verbose level which is by design behavior. I observed the thread id for the rundown thread was 264, then updated the breakpoint conditional to exclude all further hits on this thread:
`ep_event->level == 5 && $TID != 264`

When the breakpoint hits again we see a callstack like this:
```
>	coreclr.dll!ep_buffer_manager_write_event(_EventPipeBufferManager * buffer_manager, Thread * thread, _EventPipeSession * session, _EventPipeEvent * ep_event, _EventPipeEventPayload * payload, const unsigned char * activity_id, const unsigned char * related_activity_id, Thread * event_thread, _EventPipeStackContents * stack) Line 945	C++
 	coreclr.dll!ep_session_write_event(_EventPipeSession * session, Thread * thread, _EventPipeEvent * ep_event, _EventPipeEventPayload * payload, const unsigned char * activity_id, const unsigned char * related_activity_id, Thread * event_thread, _EventPipeStackContents * stack) Line 494	C++
 	coreclr.dll!write_event_2(Thread * thread, _EventPipeEvent * ep_event, _EventPipeEventPayload * payload, const unsigned char * activity_id, const unsigned char * related_activity_id, Thread * event_thread, _EventPipeStackContents * stack) Line 786	C++
 	coreclr.dll!write_event(_EventPipeEvent * ep_event, _EventPipeEventPayload * payload, const unsigned char * activity_id, const unsigned char * related_activity_id) Line 715	C++
 	[Inline Frame] coreclr.dll!ep_write_event(_EventPipeEvent *) Line 1500	C++
 	[Inline Frame] coreclr.dll!EventPipeAdapter::WriteEvent(_EventPipeEvent * ep_event, unsigned char *) Line 499	C++
 	coreclr.dll!EventPipeWriteEventGCAllocationTick_V4(const unsigned int AllocationAmount, const unsigned int AllocationKind, const unsigned short ClrInstanceID, const unsigned __int64 AllocationAmount64, const void * TypeID, const wchar_t * TypeName, const unsigned int HeapIndex, const void * Address, const unsigned __int64 ObjectSize, const _GUID *) Line 1288	C++
 	coreclr.dll!FireEtwGCAllocationTick_V4(const unsigned int AllocationAmount, const unsigned int AllocationKind, const unsigned short ClrInstanceID, const unsigned __int64 AllocationAmount64, const void * TypeID, const wchar_t * TypeName, const unsigned int HeapIndex, const void * Address, const unsigned __int64 ObjectSize, const _GUID *) Line 437	C++
 	coreclr.dll!GCToCLREventSink::FireGCAllocationTick_V4(unsigned __int64 allocationAmount, unsigned int allocationKind, unsigned int heapIndex, void * objectAddress, unsigned __int64 objectSize) Line 184	C++
 	coreclr.dll!WKS::gc_heap::adjust_limit_clr(unsigned char * start, unsigned __int64 limit_size, unsigned __int64 size, alloc_context * acontext, unsigned int flags, WKS::heap_segment * seg, int align_const, int gen_number) Line 16686	C++
 	coreclr.dll!WKS::gc_heap::a_fit_segment_end_p(int gen_number, WKS::heap_segment * seg, unsigned __int64 size, alloc_context * acontext, unsigned int flags, int align_const, int * commit_failed_p) Line 17510	C++
 	[Inline Frame] coreclr.dll!WKS::gc_heap::soh_try_fit(int) Line 17665	C++
 	coreclr.dll!WKS::gc_heap::allocate_soh(int gen_number, unsigned __int64 size, alloc_context * acontext, unsigned int flags, int align_const) Line 17784	C++
 	coreclr.dll!WKS::gc_heap::try_allocate_more_space(alloc_context * acontext, unsigned __int64 size, unsigned int flags, int gen_number) Line 18813	C++
 	coreclr.dll!WKS::gc_heap::allocate_more_space(alloc_context * acontext, unsigned __int64 size, unsigned int flags, int alloc_generation_number) Line 19305	C++
 	[Inline Frame] coreclr.dll!WKS::gc_heap::allocate(unsigned __int64) Line 19336	C++
 	coreclr.dll!WKS::GCHeap::Alloc(gc_alloc_context * context, unsigned __int64 size, unsigned int flags) Line 49030	C++
 	[Inline Frame] coreclr.dll!Alloc(unsigned __int64 size, GC_ALLOC_FLAGS flags) Line 227	C++
 	[Inline Frame] coreclr.dll!AllocateObject(MethodTable *) Line 1098	C++
 	coreclr.dll!JIT_New(CORINFO_CLASS_STRUCT_ * typeHnd_) Line 2465	C++
 	[External Code]	
 	coreclr.dll!CallDescrWorkerInternal() Line 100	Unknown
 	coreclr.dll!DispatchCallSimple(unsigned __int64 * pSrc, unsigned long numStackSlotsToCopy, unsigned __int64 pTargetAddress, unsigned long dwDispatchCallSimpleFlags) Line 221	C++
 	coreclr.dll!ThreadNative::KickOffThread_Worker(void * ptr) Line 158	C++
 	[Inline Frame] coreclr.dll!ManagedThreadBase_DispatchInner(ManagedThreadCallState *) Line 7222	C++
 	coreclr.dll!ManagedThreadBase_DispatchMiddle(ManagedThreadCallState * pCallState) Line 7266	C++
 	coreclr.dll!ManagedThreadBase_DispatchOuter(ManagedThreadCallState * pCallState) Line 7425	C++
 	[Inline Frame] coreclr.dll!ManagedThreadBase_FullTransition(void(*)(void *)) Line 7470	C++
 	[Inline Frame] coreclr.dll!ManagedThreadBase::KickOff(void(*)(void *)) Line 7505	C++
 	coreclr.dll!ThreadNative::KickOffThread(void * pass) Line 230	C++
 	[External Code]	
```

Presumably one of the EventPipe code frames prior to buffer manager should have filtered this event out, but none did so. We just need to pick a place to add the filtering logic for this case.

davmason:
Fixes #90575

When we start rundown we set the level/keywords on the public and the rundown provider: https://github.com/dotnet/runtime/blob/79c021d65c280020246d1035b0e87ae36f2d36a9/src/native/eventpipe/ep-session.c#L335-L349

If the user has a different set of events enabled for the public provider this can introduce unwanted events in the trace - i.e. GC events in a trace that specifically excludes them.

I tested that a CPU trace still symbolicates code properly in perfview and VS, if there are other scenarios people think of please let me know.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/eventpipe/ep-rt-types-aot.h
+++ b/src/coreclr/nativeaot/Runtime/eventpipe/ep-rt-types-aot.h
@@ -27,6 +27,8 @@
 #define EP_ASSERT(expr)
 #endif
 
+#define EP_FAILFAST(msg) 
+
 #undef EP_UNREACHABLE
 #define EP_UNREACHABLE(msg) do { UNREACHABLE_MSG(msg); } while (0)
 
--- a/src/coreclr/nativeaot/Runtime/eventpipe/ep-rt-types-aot.h
+++ b/src/coreclr/nativeaot/Runtime/eventpipe/ep-rt-types-aot.h
@@ -27,8 +27,6 @@
 #define EP_ASSERT(expr)
 #endif
 
-#define EP_FAILFAST(msg) 
-
 #undef EP_UNREACHABLE
 #define EP_UNREACHABLE(msg) do { UNREACHABLE_MSG(msg); } while (0)
 
--- a/src/mono/mono/eventpipe/ep-rt-types-mono.h
+++ b/src/mono/mono/eventpipe/ep-rt-types-mono.h
@@ -29,7 +29,7 @@
 #undef EP_UNLIKELY
 #define EP_UNLIKELY(expr) G_UNLIKELY(expr)
 
-#define EP_FAILFAST(msg) 
+#define EP_FAILFAST(msg) g_error(msg)
 
 struct _rt_mono_event_internal_t {
 	gpointer event;
--- a/src/mono/mono/eventpipe/ep-rt-types-mono.h
+++ b/src/mono/mono/eventpipe/ep-rt-types-mono.h
@@ -29,6 +29,8 @@
 #undef EP_UNLIKELY
 #define EP_UNLIKELY(expr) G_UNLIKELY(expr)
 
+#define EP_FAILFAST(msg) 
+
 struct _rt_mono_event_internal_t {
 	gpointer event;
 };
--- a/src/mono/mono/eventpipe/ep-rt-types-mono.h
+++ b/src/mono/mono/eventpipe/ep-rt-types-mono.h
@@ -29,8 +29,6 @@
 #undef EP_UNLIKELY
 #define EP_UNLIKELY(expr) G_UNLIKELY(expr)
 
-#define EP_FAILFAST(msg) g_error(msg)
-
 struct _rt_mono_event_internal_t {
 	gpointer event;
 };
--- a/src/native/eventpipe/ep-session.c
+++ b/src/native/eventpipe/ep-session.c
@@ -329,23 +329,16 @@ ep_session_enable_rundown (EventPipeSession *session)
 	const uint64_t keywords = 0x80020139;
 	const EventPipeEventLevel verbose_logging_level = EP_EVENT_LEVEL_VERBOSE;
 
-	EventPipeProviderConfiguration rundown_providers [2];
-	uint32_t rundown_providers_len = (uint32_t)ARRAY_SIZE (rundown_providers);
+	EventPipeProviderConfiguration rundown_provider;
+	ep_provider_config_init (&rundown_provider, ep_config_get_rundown_provider_name_utf8 (), keywords, verbose_logging_level, NULL); // Rundown provider.
 
-	ep_provider_config_init (&rundown_providers [1], ep_config_get_rundown_provider_name_utf8 (), keywords, verbose_logging_level, NULL); // Rundown provider.
+	EventPipeSessionProvider *session_provider = ep_session_provider_alloc (
+		ep_provider_config_get_provider_name (&rundown_provider),
+		ep_provider_config_get_keywords (&rundown_provider),
+		ep_provider_config_get_logging_level (&rundown_provider),
+		ep_provider_config_get_filter_data (&rundown_provider));
 
-	// Update provider list with rundown configuration.
-	for (uint32_t i = 0; i < rundown_providers_len; ++i) {
-		const EventPipeProviderConfiguration *config = &rundown_providers [i];
-
-		EventPipeSessionProvider *session_provider = ep_session_provider_alloc (
-			ep_provider_config_get_provider_name (config),
-			ep_provider_config_get_keywords (config),
-			ep_provider_config_get_logging_level (config),
-			ep_provider_config_get_filter_data (config));
-
-		ep_raise_error_if_nok (ep_session_add_session_provider (session, session_provider));
-	}
+	ep_raise_error_if_nok (ep_session_add_session_provider (session, session_provider));
 
 	ep_session_set_rundown_enabled (session, true);
 	result = true;
--- a/src/native/eventpipe/ep-session.c
+++ b/src/native/eventpipe/ep-session.c
@@ -164,6 +164,7 @@ ep_session_alloc (
 	instance->session_type = session_type;
 	instance->format = format;
 	instance->rundown_requested = rundown_requested;
+	instance->rundown_thread_id = 0;
 	instance->synchronous_callback = sync_callback;
 	instance->callback_additional_data = callback_additional_data;
 
--- a/src/native/eventpipe/ep-session.c
+++ b/src/native/eventpipe/ep-session.c
@@ -348,6 +348,7 @@ ep_session_enable_rundown (EventPipeSession *session)
 		ep_raise_error_if_nok (ep_session_add_session_provider (session, session_provider));
 	}
 
+	ep_session_set_rundown_thread_id (session, ep_thread_get_os_thread_id (ep_thread_get ()));
 	ep_session_set_rundown_enabled (session, true);
 	result = true;
 
@@ -521,6 +522,11 @@ ep_session_write_event (
 
 	// Filter events specific to "this" session based on precomputed flag on provider/events.
 	if (ep_event_is_enabled_by_mask (ep_event, ep_session_get_mask (session))) {
+		if (ep_session_get_rundown_enabled (session) && (ep_session_get_rundown_thread_id (session) != ep_thread_get_os_thread_id (ep_thread_get ()))) {
+			EP_ASSERT (ep_session_get_rundown_thread_id (session) != 0);
+			return false;
+		}
+
 		if (session->synchronous_callback) {
 			session->synchronous_callback (
 				ep_event_get_provider (ep_event),
--- a/src/native/eventpipe/ep-session.c
+++ b/src/native/eventpipe/ep-session.c
@@ -332,7 +332,6 @@ ep_session_enable_rundown (EventPipeSession *session)
 	EventPipeProviderConfiguration rundown_providers [2];
 	uint32_t rundown_providers_len = (uint32_t)ARRAY_SIZE (rundown_providers);
 
-	ep_provider_config_init (&rundown_providers [0], ep_config_get_public_provider_name_utf8 (), keywords, verbose_logging_level, NULL); // Public provider.
 	ep_provider_config_init (&rundown_providers [1], ep_config_get_rundown_provider_name_utf8 (), keywords, verbose_logging_level, NULL); // Rundown provider.
 
 	// Update provider list with rundown configuration.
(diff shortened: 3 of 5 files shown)


lateralusX (commented):
Couldn't we just remove adding the public provider to rundown? I seen other strange artifacts due to this in the provider callback that would be eliminated if we didn't add the public provider during rundown. I looked through Mono and nothing in its rundown implementation uses events outside of the rundown provider, I assume the same applies to CoreCLR/NativeAOT.

brianrob:
> Couldn't we just remove adding the public provider to rundown? I seen other strange artifacts due to this in the provider callback that would be eliminated if we didn't add the public provider during rundown. I looked through Mono and nothing in its rundown implementation uses events outside of the rundown provider, I assume the same applies to CoreCLR/NativeAOT.

I am wondering about this as well.  Is the reason that these events show up because the hardcoded rundown configuration enables the public provider at verbose level?  If so, then I'm thinking that removing the public provider is probably the right answer.

davmason:
When Noah and I were chatting about this we couldn't be sure if any scenario relied on the public provider during rundown. How confident are we that we won't break any existing scenario by removing the public provider?

brianrob:
> When Noah and I were chatting about this we couldn't be sure if any scenario relied on the public provider during rundown. How confident are we that we won't break any existing scenario by removing the public provider?

Reasonably confident, but perhaps this is a good opportunity to consider making rundown configurable?  I'm not sure how much work that is though.

lateralusX:
> When Noah and I were chatting about this we couldn't be sure if any scenario relied on the public provider during rundown. How confident are we that we won't break any existing scenario by removing the public provider?

I looked through Mono's rundown implementation and the events we emit there and all are from the rundown provider. Maybe we could do similar check on CoreCLR, ETW::EnumerationLog::EndRundown (). A brief look indicates that is mainly using the rundown provider to decide what different events to emit. The events that it emits all seems to be DC kind of events, there is one exception checking the private provider:

BOOL bIsRichDebugInfoEnabled =
    ETW_EVENT_ENABLED(MICROSOFT_WINDOWS_DOTNETRUNTIME_PRIVATE_PROVIDER_DOTNET_Context, JittedMethodRichDebugInfo);

but since the private provider has not been part of rundown, this is either used in some different scenario (maybe ETW) or not working.

Maybe we could use a variation of current fix and do some validation checks on CI, checking that written events into a session that is in rundown mode and validate that events written from rundown thread only comes from the rundown provider, if not log and abort the process so we can track it on CI?

davmason:
> Maybe we could use a variation of current fix and do some validation checks on CI, checking that written events into a session that is in rundown mode and validate that events written from rundown thread only comes from the rundown provider, if not, log and abort the process so we can track it on CI?

Great idea! I'll give it a shot

src/mono/mono/eventpipe/ep-rt-types-mono.h
```diff
@@ -29,6 +29,8 @@
 #undef EP_UNLIKELY
 #define EP_UNLIKELY(expr) G_UNLIKELY(expr)
 
+#define EP_FAILFAST(msg)
```
lambdageek, line 32:
```suggestion
#define EP_FAILFAST(msg) g_error(msg)
```

src/native/eventpipe/ep-session.c
```diff
@@ -521,6 +523,19 @@ ep_session_write_event (
 
 	// Filter events specific to "this" session based on precomputed flag on provider/events.
 	if (ep_event_is_enabled_by_mask (ep_event, ep_session_get_mask (session))) {
+		if (ep_session_get_rundown_enabled (session) && (ep_session_get_rundown_thread_id (session) != ep_thread_get_os_thread_id (ep_thread_get ()))) {
+			EP_ASSERT (ep_session_get_rundown_thread_id (session) != 0);
+			return false;
+		}
+
+		if (ep_session_get_rundown_enabled (session) && (ep_session_get_rundown_thread_id (session) == ep_thread_get_os_thread_id (ep_thread_get ()))) {
```
noahfalk, line 531: Is 531-537 just for debugging? If so can we put an `#ifdef` around it? If not failfast seems like an overly aggresive response to logging something we didn't expect to log.

src/coreclr/nativeaot/Runtime/eventpipe/ep-rt-types-aot.h
```diff
@@ -27,6 +27,8 @@
 #define EP_ASSERT(expr)
 #endif
 
+#define EP_FAILFAST(msg)
```
noahfalk, line 30: It wasn't clear that we really needed a failfast based on how it was being used, but if we do then this needs a real implementation right?

noahfalk (commented):
The filtering seemed fine, the error checking I'm skeptical on.

src/native/eventpipe/ep-session.c
noahfalk: I see you guys were discussing having this fail in CI. That seems fine to me, but I assume the code comes back out before the PR is merged?

davmason: Right, this code will be backed out. It's just to verify that we aren't getting rundown events from the public provider so we can be confident in removing the public provider rather than doing the thread filtering.

My plan is to run a diagnostc repo test pass with these changes and then call it good and change this PR to remove the public provider rather than filter by thread

noahfalk: Sounds good, that satisfies all my concerns around the failfast stuff :)

davmason:
After running the test CI build and local testing I flipped this to removing the public provider from rundown and changed the title to represent that.

src/native/eventpipe/ep-session.c
```diff
@@ -332,7 +332,6 @@ ep_session_enable_rundown (EventPipeSession *session)
 	EventPipeProviderConfiguration rundown_providers [2];
```
lateralusX, line 332: Should be 1 rundown_providers.

src/native/eventpipe/ep-session.c
```diff
@@ -332,7 +332,6 @@ ep_session_enable_rundown (EventPipeSession *session)
...
 	uint32_t rundown_providers_len = (uint32_t)ARRAY_SIZE (rundown_providers);
 
-	ep_provider_config_init (&rundown_providers [0], ep_config_get_public_provider_name_utf8 (), keywords, verbose_logging_level, NULL); // Public provider.
 	ep_provider_config_init (&rundown_providers [1], ep_config_get_rundown_provider_name_utf8 (), keywords, verbose_logging_level, NULL); // Rundown provider.
```
lateralusX, line 335: Should use index 0.

src/native/eventpipe/ep-session.c
davmason: Thanks, fixed this to get rid of the loop since there is just one provider now

lateralusX (approved):
LGTM!

PR #69236, opened 2022-05-12
[Mono] Fix 4244 warnings.

lateralusX:
Fixes all 4244 warnings on Mono runtime x86/x64 Windows build, aligning with SDL requirements (https://github.com/dotnet/runtime/issues/66154). There will be additional warnings in the cross compilers that needs to be fixed separately, but this should fix the major bulk of 4244 warnings.

All changes done in this PR are mitigating compiler issued data type truncation 4244 warnings:

`'conversion' conversion from 'type1' to 'type2', possible loss of data`

and PR will align to the same type as the compiler issued in the warning.

PR introduce a new set of macros to cast between types following patter of existing GPOINTER_TO_INT, GPOINTER_TO_UINT, GTYPE1_TO_TYPE2. The idea of use these macros is that it will be clear where casts are done, and we will have ability to intercept and add validation logic in specialized builds to catch truncation errors. The PR also introduce the needed inline functions under `ENABLE_CHECKED_CASTS` that could be used for extended validation when using the macros. Failure actions in these inline functions are not currently fully implemented, since that will be different from case to case, but can be added when needed since all infrastructure is prepared.

If it is possible to change types used in code to mitigate the warnings, that will be the initial strategy used by this PR, if that however would cause ripple effects through the source code (unfortunately quite common), we will fallback using the GTYPE1_TO_TYPE2 macros to clearly show the intent, keep trackability of the change as well as adding ability to do additional validation of casts in the future.

PR #91395, opened 2023-08-31
[mono] Reenable C4244 MSVC warning and fix occurrences

### LINKED ISSUE #91249: Mono build disables required warning C4244 under MSVC
In cmakelists.txt, we have:

https://github.com/dotnet/runtime/blob/41878765a826b5bbf54c00fd9b934bae337de650/src/mono/CMakeLists.txt#L500-L511

And in zlib.cmake, we have:

https://github.com/dotnet/runtime/blob/41878765a826b5bbf54c00fd9b934bae337de650/src/native/external/zlib.cmake#L1-L5

The end result is that _all_ of the mono native libs - not just the zlib components - are being built with C4244 disabled, which is causing binskim violations in the official build.

The mono sources should re-enable C4244 _as error_ per SDL and binskim requirements.

See also https://github.com/dotnet/runtime/pull/91245.

akoeplinger:
Fixes https://github.com/dotnet/runtime/issues/91249

We originally fixed the warnings with https://github.com/dotnet/runtime/pull/69236 but due to `add_compile_options()` calls in zlib.cmake _all_ of the mono native libs - not just the zlib components - were being built with C4244 disabled and a few couple more leaked in.

This fixes the occurrences and reworks the .cmake files a bit to hopefully prevent this in the future.


src/mono/mono/mini/aot-runtime.c
```diff
@@ -2275,7 +2275,7 @@ load_aot_module (MonoAssemblyLoadContext *alc, MonoAssembly *assembly, gpointer
 
 		addr = amodule->mem_begin;
 		g_assert (addr);
-		len = amodule->mem_end - amodule->mem_begin;
+		len = (int)(amodule->mem_end - amodule->mem_begin);
```
akoeplinger, line 2278: this one is a bit weird to me in general, I'm not sure why we need `addr` and `len` since for calculating `page_end` we're adding them back together so couldn't we just use `amodule->mem_end` instead?

akoeplinger: removed the intermediate variables

src/mono/mono/mini/graph.c
```diff
@@ -228,6 +227,7 @@ mono_draw_graph (MonoCompile *cfg, MonoGraphOptions draw_options)
 #ifdef HAVE_SYSTEM
 	//com = g_strdup_printf ("dot %s -Tpng -o %s.png; eog %s.png", fn, fn, fn);
 	com = g_strdup_printf ("dot %s -Tps -o %s.ps;gv %s.ps", fn, fn, fn);
+	int _i G_GNUC_UNUSED;
```
fanyang-mono, line 230: Do you still need `G_GNUC_UNUSED` here?

src/mono/mono/mini/helpers.c
```diff
@@ -255,6 +254,7 @@ MONO_RESTORE_WARNING
 #endif
 
 #ifdef HAVE_SYSTEM
+	int unused G_GNUC_UNUSED;
```
fanyang-mono, line 257: Same question here.

src/mono/mono/mini/mini-llvm-cpp.cpp
```diff
@@ -779,7 +779,7 @@ mono_llvm_register_overloaded_intrinsic (LLVMModuleRef module, IntrinsicId id, L
 unsigned int
 mono_llvm_get_prim_size_bits (LLVMTypeRef type)
 {
-	return unwrap (type)->getPrimitiveSizeInBits ();
+	return (unsigned int)(unwrap (type)->getPrimitiveSizeInBits ());
```
fanyang-mono, line 782: Why not using the macros here?

fanyang-mono (approved):
Overall, the change looks good to me.

src/mono/mono/mini/graph.c
akoeplinger: Yes because otherwise we'd get a warning about the variable being assigned but never used.

src/mono/mono/mini/mini-llvm-cpp.cpp
akoeplinger: Not intentional, will change, thanks!

src/mono/mono/mini/mini-llvm.c
```diff
@@ -14387,7 +14387,7 @@ emit_dbg_loc (EmitContext *ctx, LLVMBuilderRef builder, const unsigned char *cil
 		MonoDebugSourceLocation *loc;
 		LLVMValueRef loc_md;
 
-		loc = mono_debug_method_lookup_location (ctx->minfo, cil_code - cfg->header->code);
+		loc = mono_debug_method_lookup_location (ctx->minfo, GINT64_TO_INT (cil_code - cfg->header->code));
```
lateralusX, line 14390: GPTRDIFF_TO_INT

src/mono/mono/mini/mini-x86.c
```diff
@@ -3388,7 +3388,7 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
 			break;
 
 		case OP_LABEL:
-			ins->inst_c0 = code - cfg->native_code;
+			ins->inst_c0 = GINT64_TO_TMREG (code - cfg->native_code);
```
lateralusX, line 3391: GPTRDIFF_TO_TMREG

src/mono/mono/sgen/sgen-scan-object.h
```diff
@@ -14,7 +14,7 @@
  * "start" will point to the start of the next object, if the scanned
  * object contained references.  If not, the value of "start" should
  * be considered undefined after executing this code.  The object's
- * GC descriptor must be in the variable "mword desc".
+ * GC descriptor must be in the variable "SgenDescriptor desc".
```
lateralusX, line 17: These are different type depending on host/target, mword is host machine word while SgenDescriptor is target machine word, have you validated that calls only uses SgenDescriptor and we don't end up casting mword to SgenDescriptor or SgenDescriptor to mword since that can lead to data loss depending on host/target combinations.

src/mono/mono/metadata/sgen-mono.c
```diff
@@ -1297,7 +1297,7 @@ two_args_report_root (void *address, MonoObject *obj, void *gc_data)
 }
 
 static void
-precisely_report_roots_from (GCRootReport *report, void** start_root, void** end_root, mword desc)
+precisely_report_roots_from (GCRootReport *report, void** start_root, void** end_root, SgenDescriptor desc)
```
lateralusX, line 1300: Same comment as above, only called with SgenDescriptor, no casting happening on callsite?

lateralusX (commented):
Overall looks good, great work. Mainly a couple of additional places to use different conversion macros or opportunity to add a new more precise conversion macro.

The only thing to look out for is the mword SGenDescriptor related changes, mword is based on host size while SGenDescriptor is based on target size, so changing the type we will need to make sure that it doesn't introduce changed behavior or potentially loss of precision in case you have different host/target pointer sizes.

src/mono/mono/metadata/sgen-mono.c
```diff
@@ -98,7 +98,7 @@ ptr_on_stack (void *ptr)
 	} while (0)
 
 static void
-scan_object_for_binary_protocol_copy_wbarrier (gpointer dest, char *start, mword desc)
+scan_object_for_binary_protocol_copy_wbarrier (gpointer dest, char *start, SgenDescriptor desc)
```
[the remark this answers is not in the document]
akoeplinger: Yes.

src/mono/mono/metadata/sgen-client-mono.h
```diff
@@ -90,7 +90,7 @@ enum {
 };
 
 static inline mword
-sgen_mono_array_size (GCVTable vtable, MonoArray *array, mword *bounds_size, mword descr)
+sgen_mono_array_size (GCVTable vtable, MonoArray *array, mword *bounds_size, SgenDescriptor descr)
```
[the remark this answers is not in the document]
akoeplinger: It is only called with SgenDescriptor.

src/mono/mono/sgen/sgen-scan-object.h
akoeplinger: Yep I checked all the callsites and they are all called with SgenDescriptor type without casting.

lateralusX (approved):

PR #91399, opened 2023-08-31
Add late binding lookup in VB.NET for COM objects

### LINKED ISSUE #90541: Visual Basic late binding failure on COM objects
### Description

When a COM object is used (through an interop assembly) in Visual Basic, late binding can fail in two cases:

1.  If the object has been cast to a plain Object type.
2.  If the object implements more than one interface (e.g. InterfaceA and InterfaceB). If the object is declared as InterfaceA, late binding will fail on a property that is defined on InterfaceB.

This problem shows up in .NET 7 and .NET 8 preview 7. It is a regression from .NET Framework behavior.
The problem does not show up in C# when doing late binding on the dynamic type.

### Reproduction Steps

Problem 1 can be shown with Microsoft Excel COM Interop.

The attached Visual Studio solution demonstrates problems 1 and 2. It contains a COM server and .NET test projects. Details are in the README.md file.
[ATLTest563.zip](https://github.com/dotnet/runtime/files/12338706/ATLTest563.zip)

### Expected behavior

On the Windows OS, late binding on COM objects works the same as in the .NET Framework.

### Actual behavior

Late binding fails under the conditions decribed above (more details in the README.md file).

### Regression?

It is a regression from .NET Framework 4.8 (and previous versions of .NET Framework).

It is probably not a regression from previous releases of .NET Core.
Problem 1 has been [reported previously](https://github.com/dotnet/runtime/issues/66847) in .NET 5.

### Known Workarounds

The workaround is to use early (compile-time) binding. 
That workaround is fine for new code, but there is a lot of existing late-binding code in .NET Framework written against the Autodesk Inventor API by Inventor customers. We want to provide a seemless upgrade path for the authors of that code.

### Configuration

.NET 7, and .NET 8 preview 7.
Windows 10
x64 or x86

### Other information

_No response_

AaronRobinsonMSFT:
Add back support for COM late-binding in VisualBasic

Fixes #90541

### DIFF
--- a/src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
+++ b/src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
@@ -269,21 +269,21 @@ Namespace Microsoft.VisualBasic.CompilerServices
 
         <RequiresUnreferencedCode(LateBindingTrimMessage)>
         Private Shared Function InternalLateIndexGet(
-                ByVal Instance As Object,
-                ByVal Arguments() As Object,
-                ByVal ArgumentNames() As String,
-                ByVal ReportErrors As Boolean,
-                ByRef Failure As ResolutionFailure,
-                ByVal CopyBack As Boolean()) As Object
+                ByVal instance As Object,
+                ByVal arguments() As Object,
+                ByVal argumentNames() As String,
+                ByVal reportErrors As Boolean,
+                ByRef failure As ResolutionFailure,
+                ByVal copyBack As Boolean()) As Object
 
-            Failure = ResolutionFailure.None
+            failure = ResolutionFailure.None
 
-            If Arguments Is Nothing Then Arguments = NoArguments
-            If ArgumentNames Is Nothing Then ArgumentNames = NoArgumentNames
+            If arguments Is Nothing Then arguments = NoArguments
+            If argumentNames Is Nothing Then argumentNames = NoArgumentNames
 
-            Dim baseReference As Container = New Container(Instance)
+            Dim baseReference As Container = New Container(instance)
             If baseReference.IsCOMObject And Not baseReference.IsWindowsRuntimeObject Then
-                Return LateBinding.LateIndexGet(Instance, Arguments, ArgumentNames)
+                Return LateBinding.LateIndexGet(instance, arguments, argumentNames)
             End If
 
             'An r-value expression o(a) has two possible forms:
@@ -293,10 +293,10 @@ Namespace Microsoft.VisualBasic.CompilerServices
             If baseReference.IsArray Then
                 'This is an array lookup o(a).
 
-                If ArgumentNames.Length > 0 Then
-                    Failure = ResolutionFailure.InvalidArgument
+                If argumentNames.Length > 0 Then
+                    failure = ResolutionFailure.InvalidArgument
 
-                    If ReportErrors Then
+                    If reportErrors Then
                         Throw New ArgumentException(SR.Argument_InvalidNamedArgs)
                     End If
 
@@ -304,21 +304,21 @@ Namespace Microsoft.VisualBasic.CompilerServices
                 End If
 
                 ' Initialize the copy back array to all ByVal
-                ResetCopyback(CopyBack)
-                Return baseReference.GetArrayValue(Arguments)
+                ResetCopyback(copyBack)
+                Return baseReference.GetArrayValue(arguments)
             End If
 
             'This is a default member access o.d(a), which is a call to method "".
             Return CallMethod(
                        baseReference,
                        "",
-                       Arguments,
-                       ArgumentNames,
+                       arguments,
+                       argumentNames,
                        NoTypeArguments,
-                       CopyBack,
+                       copyBack,
                        BindingFlagsInvokeMethod Or BindingFlagsGetProperty,
-                       ReportErrors,
-                       Failure)
+                       reportErrors,
+                       failure)
         End Function
 
         <RequiresUnreferencedCode(LateBindingTrimMessage)>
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
```diff
@@ -265,19 +269,22 @@ Namespace Microsoft.VisualBasic.CompilerServices
 
         <RequiresUnreferencedCode(LateBindingTrimMessage)>
         Private Shared Function InternalLateIndexGet(
-                ByVal instance As Object,
```
agocke, line 268: Shouldn't these be lower case?

AaronRobinsonMSFT: A majority of the VB code is upper case for locals. I was just trying to follow the most recent style. Happy to update them if that is appropriate.

src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
```diff
@@ -47,6 +47,10 @@ Namespace Microsoft.VisualBasic.CompilerServices
                 baseReference = New Container(Instance)
             End If
 
+            If baseReference.IsCOMObject And Not baseReference.IsWindowsRuntimeObject Then
```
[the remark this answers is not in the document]
AaronRobinsonMSFT: I don't think so. Since the API was there, I wanted to call it and let the more knowledgable people in this area speak up. Happy to remove it.

src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
cston: Consider reverting the uppercase changes. It looks like a significant fraction of the locals in this file still use lowercase, and it would be good to limit the diffs to just essential changes.

src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
```diff
@@ -927,6 +942,18 @@ Namespace Microsoft.VisualBasic.CompilerServices
...
+                Try
+                    LateBinding.InternalLateSet(Instance, Type, MemberName, Arguments, ArgumentNames, OptimisticSet, CallType)
+                    If (RValueBase And Type.IsValueType) Then
```
cston, line 948: >Type.IsValueType [](http://example.com/codeflow?start=39&length=16)

Why use `Type.IsValueType` here rather than `baseReference.IsValueType` as used elsewhere?

cston, line 948: >( [](http://example.com/codeflow?start=23&length=1)

Minor: `If` condition does not require parentheses.

src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
```diff
@@ -927,6 +942,18 @@ Namespace Microsoft.VisualBasic.CompilerServices
...
+                        Throw New Exception(Utils.GetResourceString(SR.RValueBaseForValueType, Utils.VBFriendlyName(Type, Instance), Utils.VBFriendlyName(Type, Instance)))
```
cston, line 949: >Utils.VBFriendlyName(Type, Instance), Utils.VBFriendlyName(Type, Instance) [](http://example.com/codeflow?start=95&length=74)

`baseReference.VBFriendlyName, baseReference.VBFriendlyName`?

src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
cston, line 948: >If (RValueBase And Type.IsValueType) Then [](http://example.com/codeflow?start=20&length=41)

Should the `If` block be before `LateBinding.InternalLateSet()`? (Is there a reason to call the `InternalLateSet()` if the we'll throw an exception afterwards?) And consider moving the `If` block outside of the `Try / Catch`.

src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
```diff
@@ -47,6 +47,10 @@ Namespace Microsoft.VisualBasic.CompilerServices
                 baseReference = New Container(Instance)
             End If
 
+            If baseReference.IsCOMObject And Not baseReference.IsWindowsRuntimeObject Then
+                Return LateBinding.InternalLateCall(Instance, Type, MemberName, Arguments, ArgumentNames, CopyBack, IgnoreReturn)
```
cston, line 51: Are we able to test the COM cases?

src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
AaronRobinsonMSFT: The `Type` is a byref to `InternalLateSet` and can be updated.

AaronRobinsonMSFT: Do I need to mark something to properly pass a value byref to a method in VisualBasic?

AaronRobinsonMSFT: This is needed because `Type` is byref and can be updated.

src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
AaronRobinsonMSFT: I can, but it is going to get complicated. I'm dubious it is worth the effort.

src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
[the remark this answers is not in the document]
AaronRobinsonMSFT: Actually this API is used in a few places so I'm going to leave this as such.

src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
cston: > The `Type` is a byref to `InternalLateSet` and can be updated.

Thanks. Consider adding a comment to indicate that, perhaps:
`' Using Type.IsValueType rather than baseReference.IsValueType since Type may be modified in InternalLateSet.`

src/libraries/Microsoft.VisualBasic.Core/src/Microsoft/VisualBasic/CompilerServices/NewLateBinding.vb
cston, line 50: >And Not [](http://example.com/codeflow?start=41&length=7)

The new `If` conditions should use `AndAlso Not ...` rather than `And Not ...`.

cston (approved):

AaronRobinsonMSFT:
Failure is https://github.com/dotnet/runtime/issues/91410

PR #85558, opened 2023-04-29
Share UTF8 converters between coreclr and mono

am11:
Fix https://github.com/dotnet/runtime/issues/78490

Issue #72796, opened 2022-07-25
[mono] Reduce loader lock contention in class loading


lambdageek:
The global loader lock has two high-level problems:

1. It inhibits concurrency
2. We sometimes hold it while calling managed code (class cctors, assembly load callbacks) which can lead to deadlocks.

We should continue to look for opportunities to move code out of the global loader lock into specific locks for individual data structures.

Some ideas / Tracking:

- [ ] Implement "class readiness levels" https://github.com/mono/mono/issues/7856
- [ ] Implement "please retry" mode for class initialization https://github.com/dotnet/runtime/issues/51864 (comment: https://github.com/dotnet/runtime/issues/51864#issuecomment-953977929)

lambdageek:
Came up again in https://github.com/dotnet/runtime/issues/93686

Issue #51881, opened 2021-04-26
[mono] AppDomain.CurrentDomain.AssemblyLoad Console.WriteLine assertion


lambdageek:
Running the following program with Mono will crash

```csharp
using System;
using System.Threading;
using System.Reflection;

namespace X
{
    class Program
    {
        static void Main(string[] args)
        {

	    Thread.CurrentThread.Name = "Main-Thread";
	    
	    AppDomain.CurrentDomain.AssemblyLoad += (_sender, evt) => {
		var assm = evt.LoadedAssembly;
		var name = assm.GetName();
		// removing this line fixes the crash
#if true
		Console.WriteLine ($" : Loaded {name.Name} from '{assm.Location}' on {Thread.CurrentThread.Name}");
#endif
	    };
	    
	    var allDone = new ManualResetEventSlim ();
	    var proceed = new ManualResetEventSlim ();

	    var conflictThread = new Thread(() => {
		proceed.Set();
		Console.WriteLine ($" - {Thread.CurrentThread.Name} waiting for all done");
		allDone.Wait();
	    });
	    conflictThread.Name = "Conflict-Thread";
	    conflictThread.Start();

	    proceed.Wait ();
	    allDone.Set ();
	    Thread.Sleep (1000);
	    Console.WriteLine ($" {Thread.CurrentThread.Name} exiting");
        }
    }
}
```

Expected Output: no crash

Actual Output:
```
* Assertion at /Users/alklig/work/dotnet-runtime/runtime-bugs/src/mono/mono/metadata/assembly.c:1374, condition `is_ok (hook_error)' not met, function:mono_assembly_invoke_load_hook_internal, (null) assembly:/Users/alklig/work/dotnet-runtime/runtime-bugs/artifacts/bin/HelloWorld/x64/Debug/osx-x64/publish/System.Private.CoreLib.dll type:TypeLoadException member:(null)


=================================================================
	Native Crash Reporting
=================================================================
Got a SIGABRT while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries
used by your application.
=================================================================

=================================================================
	Native stacktrace:
=================================================================
	0x100c748a4 - /Users/alklig/work/dotnet-runtime/runtime-bugs/artifacts/bin/HelloWorld/x64/Debug/osx-x64/publish/libcoreclr.dylib : dump_native_stacktrace
[...]

PR #91400, opened 2023-08-31
Create a managed implementation of assembly binder

huoyaoyuan:
https://github.com/dotnet/runtime/pull/85558#issuecomment-1602642549

> It should be possible to move the whole assembly loader to C#. We would need a special unmanaged path to load CoreLib to bootstrap, but the rest can be managed. It is too big to do it all at once. I think a good place to start with this refactoring would be `BINDER_SPACE::*`.

I've ported the types under `BINDER_SPACE::`, and a large portion of binding logic in assemblybindercommon.cpp. I also created a disabled feature switch to guard the code.

The logic is ported closely from C++ and aims line-to-line match, to keep the behavior. Initially I want to convert all the HRESULTs to exceptions, but it's not practical to use exception in control logic of binder, so I kept HRESULTs in AssemblyBinderCommon.

/cc @jkotas

Is this in the desired direction? I haven't do anything around managed/unmanaged boundary. I think code should be ported to managed, until something we don't want.


jkotas:
This is good raw material. Creating a clean managed/unmanaged boundary is probably going to be the harder part. Some thoughts:

- The unmanaged binder calls out to managed to log tracing messages and to fire resolve events. This back and forth between managed and unmanaged should go away, the unmanaged code should just do a single call to managed to do everything.
- The binding using TPA list should be folded into DefaultAssemblyLoadContext, no need for it to live in separate type.
- The managed equivalent of `TextualIdentityParser` exists as `System.Reflection.AssemblyNameFormatter`. We should just use the managed `AssemblyNameFormatter` instead of duplicating it as `TextualIdentifyParser`.
- `TextualIdentifyParser` may be used by debugger or in places that are not able to run managed code. If you run into situations like this, leave it alone in C++ for now.
- We avoid C# `record` in core libraries. C# `record` generates too much hidden code bloat that the trimmer is not able to delete.
- I am not sure what is the purpose of the binding failure HRESULT error cache. The VM caches the binding failures with the exception details (look for `AppDomain::AddExceptionToCache`). I would think that the binding failure HRESULT error cache should be unnecessary.

src/coreclr/System.Private.CoreLib/src/Internal/Runtime/Binder/AssemblyName.cs
```diff
@@ -0,0 +1,389 @@
...
+            }
+
+            // Set public key and/or public key token (if we have it)
+            if (dwPublicKeyToken != 0 && pvPublicKeyToken != null)
+            {
+                if ((dwRefOrDefFlags & CorAssemblyFlags.afPublicKey) != 0)
+                {
+                    byte* pByteToken;
+                    uint dwTokenLen;
+                    StrongNameTokenFromPublicKey(pvPublicKeyToken, dwPublicKeyToken, &pByteToken, &dwTokenLen);
```
jkotas, line 163: We have `AssemblyNameHelpers.ComputePublicKeyToken` for this. No need to call unmanaged code.

Also, we ignore the strong name key for binding purposes, so we may be able to skip eagerly computing the public key token in the first place.

src/coreclr/System.Private.CoreLib/src/Internal/Runtime/Binder/ApplicationContext.cs
```diff
@@ -0,0 +1,309 @@
...
+            }
+
+            Debug.Assert(startPos < iEnd);
+
+            outPath = paths[startPos..iEnd];
+            startPos = iNext;
+            return true;
+        }
+
+        private static bool IsRelativePath(ReadOnlySpan<char> path)
```
jkotas, line 114: It should be fine to replace this by `Path.IsPathFullyQualified`.

src/coreclr/System.Private.CoreLib/src/Internal/Runtime/Binder/AssemblyName.cs
```diff
@@ -0,0 +1,389 @@
...
+
+        public AssemblyName(IntPtr pPEImage)
+        {
+            IdentityFlags |=
+                AssemblyIdentityFlags.IDENTITY_FLAG_CULTURE | AssemblyIdentityFlags.IDENTITY_FLAG_PUBLIC_KEY_TOKEN_NULL;
+
+            int* dwPAFlags = stackalloc int[2];
+            using IMdInternalImport pIMetaDataAssemblyImport = BinderAcquireImport(pPEImage, dwPAFlags);
+
+            Architecture = AssemblyBinderCommon.TranslatePEToArchitectureType(dwPAFlags);
```
jkotas, line 77: The managed binder should not need to deal with image architectures. The unmanaged runtime should validate very early whether the image is good for current architecture (and reject it if it is not). By the time we get to the managed binder, we should know that the image architecture is good.

src/coreclr/System.Private.CoreLib/src/Internal/Runtime/Binder/ApplicationContext.cs
huoyaoyuan: I noticed the behavior difference with `IsPathRooted`, but didn't mention this.

vitek-karas:
@elinor-fung FYI

src/coreclr/System.Private.CoreLib/src/Internal/Runtime/Binder/ApplicationContext.cs
jkotas: The corner-case behavior differences should not be a problem.

huoyaoyuan:
Good points for the behaviors that we don't need in managed binder. I'm aiming for a working POC first, and not tweaking yet. Records are just simplification since I don't want to write GetHashCode etc. first.

Rough thought of steps:

- Explore the scope of managed binder, create shape for things to port, and use placeholder for boundary and complex functions.
- Fulfill implementation and boundary incrementally, to get the POC working.
- Create feature switch guard for unmanaged code replaced, and filter out things we can't remove, for example which used by debugger.
- Tweak boundary and optimise managed code.
- Turn on the feature switch eventually.

Questions:
- Should I create a tracking issue for this? This is likely not a single-PR work.
- Should we get some code merged with feature switch disabled first? How complete should it be?
- Will this affect bootstrapping on new platform/architecture?

src/coreclr/System.Private.CoreLib/src/Internal/Runtime/Binder/AssemblyName.cs
huoyaoyuan: I started with focusing the data structures, and did not look at the caller. Leaving the cleanup later should be fine, since cleaning managed code is easy.

jkotas:
> Should I create a tracking issue for this?

Sounds good to me.

> Should we get some code merged with feature switch disabled first? How complete should it be?

I would like to avoid merging untested prototype code. The code being merged should be tested and it should not have too many loose ends.

I think this work can be done without a feature switch by carving our and merging one part at a time. If you have a strong preference for a temporary feature switch, it is ok to it that way too. The problem with feature switches is that they hamper refactoring since the code has to work with both feature switch on and off.

> Will this affect bootstrapping on new platform/architecture?

I do not think so. We need to execute hundreds of managed methods to get to Main today. This may add some more to it, but it won't change the fact that the JIT has to be nearly functionally complete in order to reach Main.

Thank you!

huoyaoyuan:
> I think this work can be done without a feature switch by carving our and merging one part at a time.

The concern is about final product size, since the code should be non-trivial amount in CoreLib. But it may be fine since we are in the beginning of a new version.

> the code has to work with both feature switch on and off.

If the goal is to replace the binder/loader instead of just creating an alternative, not switching should be fine.

Goal of this PR:

Get POC working at a reasonable boundary, for example the binder first. The unmanaged binder code will not be deleted until the whole feature is complete.
There may be temporary unmanaged/managed callback setup, but will be in limited amount and clear feature boundary.

src/coreclr/System.Private.CoreLib/src/Internal/Runtime/Binder/AssemblyBinderCommon.cs
```diff
@@ -0,0 +1,862 @@
...
+        public static int BindAssembly(AssemblyBinder binder, AssemblyName assemblyName, bool excludeAppPaths, out Assembly? result)
+        {
+            int kContextVersion = 0;
+            BindResult bindResult = default;
+            int hr;
+            result = null;
+            ApplicationContext applicationContext = binder.AppContext;
+
+        // Tracing happens outside the binder lock to avoid calling into managed code within the lock
+        //BinderTracing::ResolutionAttemptedOperation tracer{ pAssemblyName, pBinder, 0 /*managedALC*/, hr};
```
huoyaoyuan, line 197: A question for how eventing works here:
It looks like that NativeRuntimeEventSource is the corresponding place for native events. It requires defining runtime-specific P/Invokes at NativeRuntimeEventSource.Threading.NativeSinks.CoreCLR.cs, and modifying the python script to avoid generating dummy methods. Is this the correct way?

jkotas: > Is this the correct way?

Yes, I think it is the correct way to do it.

If it works better, it should be fine to leave the fully managed binder tracing for a separate PR and call the unmanaged tracing using QCalls for now.

src/coreclr/binder/bindertracing.cpp
```diff
@@ -95,7 +95,7 @@ namespace
             _ASSERTE(pPEAssembly != nullptr);
             pPEAssembly->GetDisplayName(request.RequestingAssembly);
 
-            AssemblyBinder *binder = pPEAssembly->GetAssemblyBinder();
+            AssemblyBinder *binder = /*pPEAssembly->GetAssemblyBinder()*/ NULL;
```
huoyaoyuan, line 98: This should be unused now and the change is to make compiler happy

src/coreclr/vm/appdomain.cpp
```diff
@@ -5065,7 +5142,8 @@ HRESULT RuntimeInvokeHostAssemblyResolver(INT_PTR pManagedAssemblyLoadContextToB
                 pParentLoaderAllocator->EnsureReference(pResultAssemblyLoaderAllocator);
             }
 
-            pResolvedAssembly = pLoadedPEAssembly->GetHostAssembly();
+            // unused with managed binder
+            // pResolvedAssembly = pLoadedPEAssembly->GetHostAssembly();
```
huoyaoyuan, line 5146: Make compiler happy

src/coreclr/vm/assemblybinder.cpp
```diff
@@ -32,11 +32,11 @@ NativeImage* AssemblyBinder::LoadNativeImage(Module* componentModule, LPCUTF8 na
     STANDARD_VM_CONTRACT;
 
     BaseDomain::LoadLockHolder lock(AppDomain::GetCurrentDomain());
-    AssemblyBinder* binder = componentModule->GetPEAssembly()->GetAssemblyBinder();
+    AssemblyBinder* binder = /* componentModule->GetPEAssembly()->GetAssemblyBinder() */ NULL;
```
huoyaoyuan, line 35: Unused function now.

src/coreclr/vm/baseassemblyspec.inl
```diff
@@ -14,6 +14,11 @@
 #ifndef __BASE_ASSEMBLY_SPEC_INL__
 #define __BASE_ASSEMBLY_SPEC_INL__
 
+#include <clrex.h>
+#include <util.hpp>
+#include <gchandleutilities.h>
+#include <threads.h>
```
huoyaoyuan, line 20: Some components don't include necessary headers for object handle and gc.

src/coreclr/vm/ceeload.cpp
```diff
@@ -3023,20 +3023,22 @@ DomainAssembly * Module::LoadAssemblyImpl(mdAssemblyRef kAssemblyRef)
...
     }
 
     if (pDomainAssembly != NULL)
     {
+        GCX_COOP();
         _ASSERTE(
             pDomainAssembly->IsSystem() ||                  // GetAssemblyIfLoaded will not find CoreLib (see AppDomain::FindCachedFile)
             !pDomainAssembly->IsLoaded() ||                 // GetAssemblyIfLoaded will not find not-yet-loaded assemblies
-            GetAssemblyIfLoaded(kAssemblyRef, NULL, FALSE, pDomainAssembly->GetPEAssembly()->GetHostAssembly()->GetBinder()) != NULL);     // GetAssemblyIfLoaded should find all remaining cases
+            GetAssemblyIfLoaded(kAssemblyRef, NULL, FALSE, GetDomain()->CreateHandle(((BINDERASSEMBLYREF)ObjectFromHandle(pDomainAssembly->GetPEAssembly()->GetHostAssembly()))->m_binder)) != NULL);     // GetAssemblyIfLoaded should find all remaining cases
```
huoyaoyuan, line 3041: Handle leak here

src/coreclr/vm/coreassemblyspec.cpp
```diff
@@ -28,12 +28,14 @@
 #include "../binder/inc/assemblybindercommon.hpp"
 #include "../binder/inc/applicationcontext.hpp"
 
-HRESULT  AssemblySpec::Bind(AppDomain *pAppDomain, BINDER_SPACE::Assembly** ppAssembly)
+HRESULT  AssemblySpec::Bind(AppDomain *pAppDomain, BINDERASSEMBLYREF* ppAssembly)
```
huoyaoyuan, line 31: Good candidate to convert to managed in follow up.

huoyaoyuan (commented):
@jkotas Hello world is now working. It's really a lot of work comparing to my initial estimation. There are still unimplemented paths and handle leaks etc.

The next step is to bring the quality to be mergable for main. To not make code review too hard, I don't want to include more refactors in this PR. I'm having some questions now:

- How to debug managed code during bootstrap? I used printf tracing to indentify several managed code bug, but it's really not convinient.
- I have no knowledge how DAC and debugger works, and not confident to convert code used by them. How will it affected by additional managed code and objects?
- How to cover and debug other loading patterns, such as unloadability?

jkotas:
> How to debug managed code during bootstrap? 

We use native debuggers - windbg on Windows or lldb on Unix.

> I have no knowledge how DAC and debugger works, and not confident to convert code used by them.

The managed code cannot run in DAC, so the DAC should be unaffected by this change. We can do some ad-hoc testing once it is ready. 

> How to cover and debug other loading patterns, such as unloadability?

We have unloadability tests in the default test matrix. These tests should be good enough to catch most of the bugs.

huoyaoyuan:
I should have implemented all the feature gaps, so I'm requiring a round of review now.
I need some help here for lifetime of OBJECTHANDLE. There are many native types holding references of AssemblyBinder. There are definitely leaks and over-releases now.

src/coreclr/System.Private.CoreLib/System.Private.CoreLib.csproj
```diff
@@ -119,6 +119,18 @@
 
   <!-- Sources -->
   <ItemGroup>
+    <Compile Include="$(BclSourcesRoot)\Internal\Runtime\Binder\ApplicationContext.cs" />
+    <Compile Include="$(BclSourcesRoot)\Internal\Runtime\Binder\Assembly.cs" />
+    <Compile Include="$(BclSourcesRoot)\Internal\Runtime\Binder\AssemblyBinder.cs" />
```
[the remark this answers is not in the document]
huoyaoyuan: Yes, that's what I'm looking forward. I'm seeing places referencing it as "native ALC". I'd like to put it into separated PR, to help reviewing of this one.

The lifetime should follow current ALC/LoaderAllocator management, I think?

src/coreclr/vm/multicorejitimpl.h
```diff
@@ -266,7 +266,7 @@ class MulticoreJitProfilePlayer
 friend class MulticoreJitRecorder;
 
 private:
-    AssemblyBinder                   * m_pBinder;
+    OBJECTHANDLE                       m_pBinder;
```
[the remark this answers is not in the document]
huoyaoyuan: > I think we should keep an unmanaged AssemblyBinder that has the OBJECTHANDLE as an internal implementation detail.

That's really desired. I need to figure out the lifetime of collectible AssemlyBinder though. Currently I'm checking https://github.com/dotnet/runtime/blob/main/docs/design/features/unloadability.md

src/coreclr/System.Private.CoreLib/src/Internal/Runtime/Binder/TextualIdentityParser.cs
```diff
@@ -0,0 +1,163 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Diagnostics;
+using System.Text;
+
+namespace Internal.Runtime.Binder
+{
+    internal static class TextualIdentityParser
```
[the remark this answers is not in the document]
huoyaoyuan: Should refactoring be separated since this PR is really big?

It's also worth tracking the usage of this formatter. If it's only used for diagnostics, we should not be concerned to change it.

src/coreclr/vm/peassembly.inl
```diff
@@ -92,7 +92,11 @@ inline ULONG PEAssembly::HashIdentity()
         GC_TRIGGERS;
     }
     CONTRACTL_END;
-    return m_pHostAssembly->GetAssemblyName()->Hash(BINDER_SPACE::AssemblyName::INCLUDE_VERSION);
+
+    GCX_COOP();
```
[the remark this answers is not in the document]
huoyaoyuan: Yeah it's probably cleaner to delete dead code from main first.

src/coreclr/System.Private.CoreLib/System.Private.CoreLib.csproj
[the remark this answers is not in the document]
jkotas: There is a managed/unmanaged pair for AssemblyLoadContext today with the right lifetime management. I think you should keep the lifetime management as is, and just attach more things to managed or unmanaged sides of this pair.

> I'd like to put it into separated PR, to help reviewing of this one.

I think having a simple code without unnecessary or duplicate abstractions is more helpful for reviewing than similarity with the current unmanaged code.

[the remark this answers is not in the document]
huoyaoyuan: I'm reverting some changes and rewriting in a new structure. The native CustomAssemblyBinder/DefaultAssemblyBinder types will be preserved and referenced in native code with lifetime management only, and the binding logic will still be managed. The managed code will be attached into ALC type, but in separated files.
This significantly reduces the change at native side, and also some boilerplate to invoke ALC resolving events.

src/coreclr/System.Private.CoreLib/src/System/Runtime/Loader/Assembly.cs
```diff
@@ -0,0 +1,57 @@
...
+    // We also store whether it was bound using TPA list
+    internal sealed class BinderAssembly
+    {
+        // fields used by VM
+#pragma warning disable CA1823, 414, 169
+        private BinderAssemblyName m_assemblyName;
+
+        // The assembly object is rooted by native object.
+        // Reference native ALC to allow managed ALC object to be collected.
+        private IntPtr m_binder;
```
huoyaoyuan, line 17: This is a viable solution for now, but I don't know how it affects future design if the BINDER_SPACE::Assembly type is somehow attached with RuntimeAssembly.

src/coreclr/System.Private.CoreLib/src/System/Runtime/Loader/AssemblyLoadContext.CoreCLR.cs
```diff
@@ -38,20 +38,21 @@ public partial class AssemblyLoadContext
         internal static extern bool IsTracingEnabled();
 
         [LibraryImport(RuntimeHelpers.QCall, EntryPoint = "AssemblyNative_TraceResolvingHandlerInvoked", StringMarshalling = StringMarshalling.Utf16)]
-        [return: MarshalAs(UnmanagedType.Bool)]
-        internal static partial bool TraceResolvingHandlerInvoked(string assemblyName, string handlerName, string? alcName, string? resultAssemblyName, string? resultAssemblyPath);
```
huoyaoyuan, line 42: This looks like a bug that native QCall implementation returns void.

src/coreclr/System.Private.CoreLib/src/System/Runtime/Loader/AssemblyName.cs
```diff
@@ -0,0 +1,393 @@
...
+        INCLUDE_PUBLIC_KEY_TOKEN = 0x10,
+        EXCLUDE_CULTURE = 0x20,
+        INCLUDE_ALL = INCLUDE_VERSION
+                    | INCLUDE_ARCHITECTURE
+                    | INCLUDE_RETARGETABLE
+                    | INCLUDE_CONTENT_TYPE
+                    | INCLUDE_PUBLIC_KEY_TOKEN,
+    }
+
+    internal sealed unsafe class BinderAssemblyName : AssemblyIdentity, IEquatable<BinderAssemblyName>
```
huoyaoyuan, line 65: This should be replaced with `System.Reflection.AssemblyName` in following PR. It may require more native side changes so I'm not willing to include in this PR.

The `BINDER_SPACE::AssemblyName` type is still used in native. The conversion should cover all its usage.

src/coreclr/System.Private.CoreLib/src/System/Runtime/Loader/AssemblyVersion.cs
```diff
@@ -0,0 +1,42 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Runtime.Loader
+{
+    internal struct AssemblyVersion : IEquatable<AssemblyVersion>
```
huoyaoyuan, line 6: Behavioral difference with `System.Version`: allows Unspecified value on each element.
Can be replaced together with AssemblyName replacement.

src/coreclr/inc/CrstTypes.def
```diff
@@ -87,6 +87,7 @@ End
 
 Crst AssemblyLoader
     AcquiredBefore DeadlockDetection UniqueStack DebuggerMutex
+                   ThreadStore GlobalStrLiteralMap LoaderAllocatorReferences
 End
```
huoyaoyuan, line 91: **Design question:**

Managed code are starting to be called under AssemblyLoader lock. Currently only DeclareDependencyOnMvid for R2R path, but I assume that after more components converted to managed, there will be more chance to invoke managed code.

Currently there is no dependency cycle on this. But I want to ensure this is correct in architecture and won't block future changes.

src/coreclr/vm/ceemain.cpp
```diff
@@ -962,6 +962,8 @@ void EEStartupHelper()
         }
 #endif // _DEBUG
 
+        // Initialize managed binder. Ensure g_fEEStarted and COM initialized before executing managed code.
+        SystemDomain::System()->PostStartInit();
```
huoyaoyuan, line 966: Is it correct time to initialize managed binder here? COM components assert for g_fEEStarted if the initialization is invoked earlier.

src/coreclr/vm/object.h
```diff
@@ -1466,6 +1469,9 @@ class AssemblyLoadContextBaseObject : public Object
     CLR_BOOL      _isCollectible;
 #else // TARGET_64BIT
     int64_t       _id; // On 32-bit platforms this 64-bit value type is larger than a pointer so JIT places it first
+    OBJECTREF     _appContext;
+    OBJECTREF     _assemblySimpleNameMvidCheckHash;
+    OBJECTREF     _loadedAssemblies;
```
huoyaoyuan, line 1474: There's layout mismatch asserted only on ARM32. Need some help about field layout since I don't have ARM environment.

huoyaoyuan (commented):
Most CI legs are passing now. I'm calling for code review and asking help for remaining failures that I really don't have idea.

There are also some questions commented in PR. I still prefer to not do big refactor in this PR and keep native-managed mapping, since I've found several bugs when comparing the converted code.

src/libraries/System.Private.CoreLib/src/System/Reflection/AssemblyNameHelpers.StrongName.cs
```diff
@@ -8,9 +8,9 @@ namespace System.Reflection
 {
     internal static partial class AssemblyNameHelpers
     {
-        public static byte[]? ComputePublicKeyToken(byte[]? publicKey)
+        public static byte[]? ComputePublicKeyToken(ReadOnlySpan<byte> publicKey)
         {
-            if (publicKey == null)
+            if (publicKey == default)
```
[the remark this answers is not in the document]
huoyaoyuan: The original implementation has behavioral difference on empty and `null`. There is a `Length == 0` check below.

I need better understanding about when will null/empty be passed in.

src/coreclr/System.Private.CoreLib/src/System/Runtime/Loader/AssemblyName.cs
huoyaoyuan: An interesting thing is that `S.R.AssemblyName` can be modified by malicious event handler on ALC. I'm not sure about the scope it's affecting. Anyway, cloning shouldn't more complex than creating a new type.

src/coreclr/vm/object.h
am11: ([wild guess](https://github.com/dotnet/coreclr/pull/23906/files)) Does it help if we enable `pshpack4` and `poppack` (from line 1447 and 1493) for unix arm? There is currently a four byte difference between managed and unmanaged object sizes.

jkotas:
I have been looking at this on and off. The main thing that I am worried about with the current change is impact on startup.

On Windows x64 release, I see about 20ms startup time regression (70ms -> 90ms) and about 3MB workingset regression (20MB -> 23MB) for simple console `Hello world`. The workingset regression is probably the easier one to drill into. To reproduce it, dotnet publish `Console.WriteLine(Environment.WorkingSet)` and overwrite the runtime in the output with your local build.

lambdageek:
@jkotas More for my own education, rather than as a comment on the details of this PR:  how does a managed assembly binder interact with class loading/initialization in coreclr?  In Mono we have scenarios where constructing a class triggers assembly loading while we're holding the global loader lock - which can run managed loader events which can lead to deadlocks (#72796) .  My understanding is that CoreCLR class loading is less atomic - relying on load levels and individual class locks.  But I was under the impression that there are still scenarios where managed callbacks fire while a native lock is held.  Does moving the assembly binder to managed code affect those scenarios?

Also what happens if you trigger class initialization from the managed assembly binder?  One way I could confuse mono was by doing a `Console.WriteLine` in the `AppDomain.AssemblyLoad` event handler before calling `WriteLine` (#51881) - which triggers loading of `System.Console.dll`  which triggers a recursive class initialization of `Console`

It's possible that the interaction between the assembly binder and class initialization is more well behaved in CoreCLR, but I'm curious if porting the binder to managed has any impact.

jkotas:
Right, assembly loader needs to be able to call AssemblyLoadContext callbacks and AppDomain events that are managed code. Rewriting more of the assembly loader in managed code means that there is going to be more managed code running, but it does not change any fundamental invariants.

> I was under the impression that there are still scenarios where managed callbacks fire while a native lock is held. Does moving the assembly binder to managed code affect those scenarios?

CoreCLR class loading/initialization has lock per type. (The implementation of this lock is being rewritten in #96653 to be more scalable and performant.) Triggering assembly loading or more type loading from type loading works fine, as long as there is no recursive loading of the same type.

> class initialization

CoreCLR does not trigger class initializers during type or assembly loader. The class initializers are triggered only once the code runs. We are out of the assembly and type loader by that time.

agocke:
Moved to draft while we figure out if we can either fix the perf regressions, or accept them.

huoyaoyuan:
I've done some measurement about working set:

It's about 3.5MB of total regression comparing to main for a given run. Break down by vmmap:
Heap (unmanaged?): +0.8MB
Image: +1.5MB. The working set of CoreLib, coreclr, clrjit and icu.dll has all increased.
Managed: vmmap can't show this category of PR build, but successes for main branch build. It's about 1.5MB for main.
Mapped file: +0.6MB for icudtl.dat
Page table: -0.3MB
Private data: +276KB, a block of R/W data
Sharable: +2MB. Not sure what it represents.

So one obvious conclusion is that some globalization related code are touched unintentionally.

src/coreclr/System.Private.CoreLib/src/System/Runtime/Loader/AssemblyName.cs
```diff
@@ -0,0 +1,392 @@
...
+            }
+
+            static uint HashCaseInsensitive(string str)
+            {
+                // ported from SString::HashCaseInsensitive
+                uint hash = 5381;
+
+                foreach (char ch in str)
+                {
+                    hash = ((hash << 5) + hash) ^ char.ToUpperInvariant(ch);
```
ANahr, line 248: Might call into unwanted code. Maybe str.GetHashCode(StringComparison.OrdinalIgnoreCase)

ANahr (commented):
There are multiple cases in the code that might trigger globalization code

src/coreclr/System.Private.CoreLib/src/System/Runtime/Loader/ApplicationContext.cs
```diff
@@ -0,0 +1,281 @@
...
+                if (TrustedPlatformAssemblyMap != null)
+                {
+                    return;
+                }
+
+                //
+                // Parse TrustedPlatformAssemblies
+                //
+
+                TrustedPlatformAssemblyMap = new Dictionary<string, TPAEntry>(StringComparer.InvariantCultureIgnoreCase);
```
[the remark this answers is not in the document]
huoyaoyuan: Oh how did I write this. I'm always aware that `OrdinalIgnoreCase` should be preferred than `InvariantCultureIgnoreCase`.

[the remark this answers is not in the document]
huoyaoyuan: Nice catch @ANahr . Changing these call sites to use `OrdinalIgnoreCase` reduces the working set regression to ~1.1MB.

huoyaoyuan:
Keep open since it's functionally complete.

huoyaoyuan:
### Status report:
##### Working Set:
About 15.2 MB -> 16.3 MB under default configuration.
The difference is mainly +0.6MB in mapped image of CoreLib, and +0.3MB of managed heap, +0.2MB of sharable page.
There is reduction of mapped image of coreclr.dll, but the difference is much smaller.
##### Start time:
Measured with `Measure-Command` in PowerShell as an external source. The startup time range shifts from 23.6 ~ 24.8ms to 25.3 ~ 26.7ms. The regression is clearly measurable.
##### File size:
-39KB in coreclr.dll, +92KB in CoreLib (R2R).

There are still some space of dead code for optimization, but I kept this PR as close to native code as possible to help reviewing.
Cleanup around managed->native calls may also improve performance in the future.

src/coreclr/System.Private.CoreLib/src/System/Runtime/Loader/ApplicationContext.cs
```diff
@@ -0,0 +1,301 @@
...
+                    // Skip exe files when the caller requested only dlls
+                    continue;
+                }
+
+                if (outPath.EndsWith(".ni.dll", StringComparison.OrdinalIgnoreCase)
+                    || outPath.EndsWith(".ni.exe", StringComparison.OrdinalIgnoreCase))
+                {
+                    simpleName = outPath[iSimpleNameStart..^7];
+                    isNativeImage = true;
+                }
```
huoyaoyuan, line 176: Is `ni.dll` still a thing? I remember it's output of NGen.

huoyaoyuan:
I'm going to abandon this for now.

Per offline discussion with @jkotas , there's no net improvement for this PR. The newly created managed-native boundaries are quite a lot. It's not easy to eliminate them since many of the native components are used by diagnostics interfaces like DAC, or bootstrap path for executing managed application. There's also duplications between managed and native code, and potential discrepancies with CoreLib loading.

If anyone want to revisit this in the future, consider to create a branch in runtimelab. There are also improvable small points discovered during this.

Although not a successful experiment, still great thanks to @jkotas for assisting on this!

agocke:
@huoyaoyuan  thanks for your effort! I’m sure it will be useful for the next changes in this area.

Issue #90965, opened 2023-08-22
Tier0 castclass pattern allocates


EgorBo:
The following minimal repro is the reason why we regressed TE benchmarks in https://github.com/dotnet/runtime/issues/90899
```cs
using System.Runtime.CompilerServices;

Test<MyStruct>(default);

[MethodImpl(MethodImplOptions.NoInlining)] 
static void Test<T>(T t) => ((IMyInterface)t).DoWork();

public interface IMyInterface
{
    void DoWork();
}

public struct MyStruct : IMyInterface
{
    int a;

    [MethodImpl(MethodImplOptions.NoInlining)]
    public void DoWork() {}
}

```

Tier0 codegen for `Test`:
```asm
; Assembly listing for method Program:<<Main>$>g__Test|0_0[MyStruct](MyStruct) (Tier0)
       push     rbp
       sub      rsp, 48
       lea      rbp, [rsp+0x30]
       xor      eax, eax
       mov      qword ptr [rbp-0x08], rax
       mov      qword ptr [rbp-0x10], rax
       mov      qword ptr [rbp+0x10], rcx
       lea      rdx, [rbp+0x10]
       mov      rcx, 0x7FFAFBFEE288      ; MyStruct
       call     CORINFO_HELP_BOX  ;;;; <------------------------- allocation!
       mov      gword ptr [rbp-0x08], rax
       mov      rdx, gword ptr [rbp-0x08]
       mov      rcx, 0x7FFAFBFEE0B8      ; IMyInterface
       call     [CORINFO_HELP_CHKCASTINTERFACE]
       mov      gword ptr [rbp-0x10], rax
       mov      rcx, gword ptr [rbp-0x10]
       mov      r11, 0x7FFAFB4A0248
       call     [r11]IMyInterface:DoWork():this
       nop      
       add      rsp, 48
       pop      rbp
       ret
```
Tier1 codegen for it:
```asm
; Assembly listing for method Program:<<Main>$>g__Test|0_0[MyStruct](MyStruct) (FullOpts)
       sub      rsp, 40
       mov      dword ptr [rsp+0x20], ecx
       lea      rcx, [rsp+0x20]
       call     [MyStruct:DoWork():this]
       nop      
       add      rsp, 40
       ret
```

Presumably, it needs:
1) Don't share temps for boxed values in `impImportAndPushBox` (so locals can have exact class info)
2) `impOptimizeCastClassOrIsInst` should not quit if opts are disabled
[...]

PR #91403, opened 2023-08-31
Expand CreateSpan in T0

MichalPetryka:
I've noticed my app allocate 100MB of `RuntimeFieldInfoStub` instances in the VM implementation while running the T0 code in my app, expanding in T0 should avoid such allocations.


MichalPetryka:
@jkotas Could this get a backport to 8.0?

jkotas:
It is up to the JIT team to make the call on a backport and bring it through the servicing process.

I do not think that this meets the bar for backport to .NET 8. It is by design that T0 compilation consumes extra memory, including extra GC allocations that would be eliminated by full JIT optimizations.

> I've noticed my app allocate 100GB of RuntimeFieldInfoStub instances in the VM implementation

Why is not T1 compilation kicking fast enough for your app? It takes a while to allocate 100GB of RuntimeFieldInfoStub, I would expect T1 compilation to kick in a long before this much of RuntimeFieldInfoStub instances get allocated.

MichalPetryka:
> It is by design that T0 compilation consumes extra memory, including extra GC allocations that would be eliminated by full JIT optimizations.

I understand that but with this optimalization pessimizing T0 code so much using it without AggressiveOptimization turns out to be a bad idea, making it effectively useless and a serious performance issue.

> It takes a while to allocate 100GB of RuntimeFieldInfoStub

I've actually misread the data, it's able to allocate 100MB, not 100GB. The code in question consists of many hot loops which probably take a lot of time to OSR.

EgorBo:
I agree with Jan that we already have tons of similar cases in tier0, e.g. `Enum.HasFlag` (already fixed), also everything related to devirtualizations of interfaces calls on structs when jit can see the exact type (see https://github.com/dotnet/runtime/issues/90965) and other patterns where only optimized code can get rid of allocations.

We also know that sometimes a method may stay in Tier0 for too long because of the "promotion queue" problem (We had an idea to make it priority-based and e.g. all methods with allocators inside could get a higher priority).

src/coreclr/jit/importercalls.cpp
```diff
@@ -2551,6 +2551,9 @@ GenTree* Compiler::impIntrinsic(GenTree*                newobjThis,
             // This one is just `return true/false`
             case NI_System_Runtime_CompilerServices_RuntimeHelpers_IsKnownConstant:
 
+            // Not expanding this leads to huge allocation in T0
```
EgorBo, line 2554: Can you make the comment a bit less drammatic? 🙂 I'm referring to 'huge" word

EgorBo (approved):

JulieLeeMSFT:
@MichalPetryka, there are build errors. Please resolve it before merging.

MichalPetryka:
@MihuBot -tier0

EgorBo:
@MichalPetryka @MihaZupan is MihuBot failing because of this PR or it's a general issue? Does it fail on other changes?

MihaZupan:
It's failing somewhere in `jit-diff`. I'll try it on different PRs now.

MihaZupan:
Seems like the issue is specifically with this PR and tier 0.
It's working fine on other PRs, or on this PR without the `--tier0` flag.

EgorBo:
Let's run some pipelines which don't disable TieredCompilation because green inner-loop CI likely means it was never executed in Tier0.

EgorBo:
@MihuBot -tier0

EgorBo:
I've ran jitdiff locally with --tier0 and this PR didn't seem to cause any issues for me so I am going to merge it.

PR #91413, opened 2023-08-31
Fix x64 crossbuild on macOS arm64

janvorli:
The cross build is failing due to the cpufeatures.c being compiled in. This file tries to extract cpu features using cpuid / getauxval that don't make sense to execute in the cross tools.

This change disables compiling the cpufeatures.c for cross build and changes the `JitGetProcessorFeatures` to return zero in this case.

### DIFF
--- a/src/coreclr/tools/aot/jitinterface/CMakeLists.txt
+++ b/src/coreclr/tools/aot/jitinterface/CMakeLists.txt
@@ -5,9 +5,13 @@ set(NATIVE_SOURCES
     jitinterface.cpp
     jitwrapper.cpp
     corinfoexception.cpp
-
-    ${CLR_SRC_NATIVE_DIR}/minipal/cpufeatures.c
 )
+
+if(NOT CLR_CROSS_COMPONENTS_BUILD)
+    list(APPEND NATIVE_SOURCES
+        ${CLR_SRC_NATIVE_DIR}/minipal/cpufeatures.c
+    )
+endif()
 
 if(CLR_CMAKE_TARGET_WIN32)
   set(JITINTERFACE_RESOURCES Native.rc)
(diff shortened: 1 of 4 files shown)


jkotas:
jitinterface should not have any target specific code. Why are we cross-compiling it with TARGET != HOST in the first place?

janvorli:
> Why are we cross-compiling it with TARGET != HOST in the first place?

It compiles that when building libjitinterface_arm64.dylib during cross components build.

MichalStrehovsky:
Isn't the problem more that cpufeatures.h/.c uses `TARGET_` ifdefs, but it should use `HOST_`?

ILC/crossgen2 (that use this) are both crosscompilers. They already know not to call into this logic when TARGET != HOST. I would expect x64 version of crossgen2 that we built on arm64 to be able to detect CPU features on x64.

janvorli:
> Isn't the problem more that cpufeatures.h/.c uses `TARGET_` ifdefs, but it should use `HOST_`?

The host doesn't help here. That was actually my first thought too. Extracting the features when TARGET != HOST doesn't seem to make sense. Say we have JIT targeting arm64 running on x64. Extracting x64 features is useless for the JIT isn't it? And vice versa.

MichalStrehovsky:
> > Isn't the problem more that cpufeatures.h/.c uses `TARGET_` ifdefs, but it should use `HOST_`?
> 
> The host doesn't help here. That was actually my first thought too. Extracting the features when TARGET != HOST doesn't seem to make sense. Say we have JIT targeting arm64 running on x64. Extracting x64 features is useless for the JIT isn't it? And vice versa.

ILC/crossgen is a crosscompiler and what we're building is still a crosscompiler - one can just pass `--targetarch:XYZ` to the built ILC/crossgen and it will do the right thing (error out if `--instruction-set:native` was specified but hostarch!=targetarch, or call into this API to find the right flags if hostarch==targetarch).

jkotas:
`jitinterface*.dll` build should be only parametrized by the host architecture. If we are building it with HOST!=TARGET for some reason, the binary should not be affected by the TARGET in any way.

This is different for the clrjit itself. clrjit build is parametrized by both the host architecture and the target architecture.

jkotas:
Conflicts...

janvorli:
Resolved

jkotas (approved):
Thanks

jkotas:
Backport candidate?

PR #91415, opened 2023-08-31
Fix NativeAOT unhandled exception stack trace

### LINKED ISSUE #91298: [native aot] Unhandled exceptions are not printed to console
Compile and run with nativeaot on current main:

```
throw new Exception("My unhandled exception");
```

## Expected result

Unhandled exception message is printed to console.

## Actual result

No message printed to console

janvorli:
The recent change to add a new exception handling mechanism to coreclr has broken stack traces on unhandled exceptions in nativeaot.

This change fixes it by reverting the part of the change that caused the problem and that turned out to not to be needed for the new exception handling either.

Close #91298

### DIFF
--- a/src/tests/baseservices/exceptions/unhandled/unhandled.cs
+++ b/src/tests/baseservices/exceptions/unhandled/unhandled.cs
@@ -24,7 +24,6 @@ namespace TestUnhandledException
 
             testProcess.StartInfo.FileName = Environment.ProcessPath;
             testProcess.StartInfo.Arguments = Environment.CommandLine + " throw";
-            testProcess.StartInfo.UseShellExecute = false;
             testProcess.StartInfo.RedirectStandardError = true;
             testProcess.ErrorDataReceived += (sender, line) => 
             {
@@ -41,7 +40,7 @@ namespace TestUnhandledException
             testProcess.CancelErrorRead();
 
             int expectedExitCode;
-            if ((Environment.OSVersion.Platform == PlatformID.Unix) || (Environment.OSVersion.Platform == PlatformID.MacOSX))
+            if (!OperatingSystem.IsWindows())
             {
                 expectedExitCode = 128 + 6;
             }
@@ -60,7 +59,7 @@ namespace TestUnhandledException
                 return 101;
             }
 
-            if (Regex.Match(lines[0], @"Unhandled exception[.:] System\.Exception\: Test", RegexOptions.IgnoreCase) == Match.Empty)
+            if (lines[0] != "Unhandled exception. System.Exception: Test")
             {
                 Console.WriteLine("Missing Unhandled exception header");
                 return 102;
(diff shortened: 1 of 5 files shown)


agocke:
Should we add a smoke test for this?

jkotas (approved):
Thanks!

jkotas:
> Should we add a smoke test for this?

I think a regular test - that runs on all runtime flavors - would be enough.

janvorli:
Ok, let me add a test here.

src/tests/baseservices/exceptions/unhandled/unhandled.cs
```diff
@@ -0,0 +1,78 @@
...
+                throw new Exception("Test");
+            }
+
+            List<string> lines = new List<string>();
+
+            Process testProcess = new Process();
+
+            testProcess.StartInfo.FileName = Environment.ProcessPath;
+            testProcess.StartInfo.Arguments = Environment.CommandLine + " throw";
+            testProcess.StartInfo.UseShellExecute = false;
```
jkotas, line 27:
```suggestion
```
Unnecessary - it is the default in .NET Core.

src/tests/baseservices/exceptions/unhandled/unhandled.cs
```diff
@@ -0,0 +1,78 @@
...
+                }
+            };
+
+            testProcess.Start();
+            testProcess.BeginErrorReadLine();
+            testProcess.WaitForExit();
+            testProcess.CancelErrorRead();
+
+            int expectedExitCode;
+            if ((Environment.OSVersion.Platform == PlatformID.Unix) || (Environment.OSVersion.Platform == PlatformID.MacOSX))
```
jkotas, line 44:
```suggestion
            if (!OperatingSystem.IsWindows())
```
`Environment.OSVersion.Platform` is a soft-obsolete API. It behaves in non-sensical way for historic reasons (for example, it returnns `PlatformID.Unix` even on macOS).

src/tests/baseservices/exceptions/unhandled/unhandled.cs
```diff
@@ -0,0 +1,78 @@
...
+                expectedExitCode = unchecked((int)0xE0434352);
+            }
+
+            if (expectedExitCode != testProcess.ExitCode)
+            {
+                Console.WriteLine($"Wrong exit code 0x{testProcess.ExitCode:X8}");
+                return 101;
+            }
+
+            if (Regex.Match(lines[0], @"Unhandled exception[.:] System\.Exception\: Test", RegexOptions.IgnoreCase) == Match.Empty)
```
jkotas, line 63: Feel free to change the native AOT string to `Unhandled exception` and use a simple exact match here instead.

jkotas:
Thank you!

Issue #90404, opened 2023-08-11
[Mono] Consider switching to C11


lambdageek:
Currently Mono is supposed to use C99. As a result, we're relying on platform or compiler intrinsics for some functionality that has been added to the standard since then such as `stdatomic.h`.

We should consider upgrading Mono to a more recent language standard.

This came up as part of #88279, for example.

Some concrete things we could take advantage of:

* `stdatomic.h` for atomic reads/writes, CAS, etc
* `_Alignas`
* static assertions
* `_Noreturn`
* possibly thread locals (although the JIT likes to take advantage of understanding TLS layout, so we need to ensure we don't take a perf hit)

Some reasons we might not want to do it:
* Anecdotally, I've heard that MSVC C11 support has some issues - we will need to understand the specific pitfalls to avoid so we don't break `mono-aot-cross` running on Win32 and also Mono on non-Windows Win32 platforms

lambdageek:
cc @vargaz @AaronRobinsonMSFT @lateralusX @am11

MichalPetryka:
> * I've heard that MSVC C11 support has some issues

MSVC doesn't support C11, only C17.

AaronRobinsonMSFT:
> > * I've heard that MSVC C11 support has some issues
> 
> MSVC doesn't support C11, only C17.

Yes it does - https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/. There is no support for optional features and there are bugs around atomics, but it has support.

am11:
In libraries, we are using C11 for couple of years:

https://github.com/dotnet/runtime/blob/376e2353e4a06bd79639cad4ea83fd9001e62883/eng/native/configurecompiler.cmake#L10-L11
and e.g. its 'Generic Selection' feature.: https://github.com/dotnet/runtime/blob/376e2353e4a06bd79639cad4ea83fd9001e62883/src/native/libs/System.Native/pal_process.c#L559

VCR handles it just fine. Only these features are not supported: https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/#whats-not which we don't use in libs.

AaronRobinsonMSFT:
The atomics in C11 is recent and experimental, announced at https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/.

barcharcraz:
[...]

PR #91421, opened 2023-08-31
Bump mono to -std=gnu11

lambdageek:
Change the mono build to require C11 (with gnu extensions on gcc/clang platforms).
* Change `g_static_assert` to be `_Static_assert` or `static_assert` as apropriate.
* Change `_DN_STATIC_ASSERT` to be `static_assert`
* Add static asserts in `jiterp.c` when it casts between `T*` and `atomic_T*`
* Add C11 guidance to the mono coding guide doc

Contributes to #90404

### DIFF
--- a/docs/coding-guidelines/mono-code-guide.md
+++ b/docs/coding-guidelines/mono-code-guide.md
@@ -80,6 +80,11 @@ If you need them, use `mono_mutex_t`, `mono_cond_t`, etc for cases where GC tran
 prohibitively costly and you can guarantee that the lock will never be taken by a GC initiator or by
 a mix of threads in GC cooperative and GC preemptive mode.
 
+Using standard C threads and `call_once` in native library PInvokes outside the runtime is okay.
+
+Using standard C mutexes and condition variables in native library PInvokes outside the runtime is
+okay if the locks are not shared with the runtime internals.
+
 ### Thread locals
 
 FIXME: no guidance yet
@@ -91,6 +96,9 @@ The C standard atomics are not guaranteed to be lock-free.  Use the mono `mono_a
 general want locking because it would not be GC aware and may deadlock the cooperative GC.
 
 Uses of `_Atomic` are a code smell in Mono.
+
+Using standard C atomics in native library PInvokes outside the runtime is okay provided the atomics
+aren't also accessed inside the runtime internals.
 
 ### Generic operations
 
--- a/docs/coding-guidelines/mono-code-guide.md
+++ b/docs/coding-guidelines/mono-code-guide.md
@@ -61,6 +61,41 @@ There are actually three boolean types to keep in mind:
 * `MonoBoolean` used as an interop type with C# bool in internal calls
 * `mono_bool` used by the public C API - generally new code shouldn't use it except when adding a new public API function.
 
+## Newer features of C11 and later
+
+Mono is currently (2023) written in C11.
+
+### Static asserts
+
+Use `static_assert` liberally.  Include `<assert.h>`.  If you cannot include `<assert.h>` (for
+example because it introduces conflicting symbols or macros), use `g_static_assert`.  Do not use
+`_Static_assert` directly (see how `g_static_assert` is defined). Rationale: C23 deprecates
+`_Static_assert`.
+
+### Threads, locks, `call_once`
+
+Due to mono's threading model and cooperative GC, using C threading and locking primitives directly
+is not ok.  Prefer `MonoCoopMutex`, `MonoCoopCond` that have GC-aware locking/waiting operations.
+If you need them, use `mono_mutex_t`, `mono_cond_t`, etc for cases where GC transitions are
+prohibitively costly and you can guarantee that the lock will never be taken by a GC initiator or by
+a mix of threads in GC cooperative and GC preemptive mode.
+
+### Thread locals
+
+FIXME: no guidance yet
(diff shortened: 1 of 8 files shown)


src/mono/CMakeLists.txt
```diff
@@ -525,7 +525,7 @@ if(GCC)
     set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu11")
   else()
     # We require C99 with some GNU extensions, e.g. `linux` macro
-    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99")
+    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu11")
```
jkoritzinsky, line 528: Instead of passing the standard flag directly, can we change this to set the CMake properties that emit these versions (`CMAKE_C_STANDARD_REQUIRED`, `CMAKE_C_STANDARD`, and `CMAKE_C_EXTENSIONS`)?

lambdageek: yes, CMAKE_C_EXTENSIONS is what I didn't know about, thanks!

And that means we can actually just inherit `CMAKE_C_STANDARD_REQUIRED` and `CMAKE_C_STANDARD` from `eng/native/configurecompiler.cmake`.  Nice.

am11: > ah nevermind we don't include that in Mono yet.

Yup, once it's done, it will cleanup some workarounds such as: https://github.com/dotnet/runtime/blob/c540d01de6a2656722cbe85b9fa3eac400510279/src/native/libs/System.IO.Compression.Native/extra_libs.cmake#L3 and cleanup duplicate code from cmake files under src/mono.

src/mono/mono/mini/interp/jiterpreter.c
```diff
@@ -925,6 +928,7 @@ mono_interp_tier_prepare_jiterpreter_fast (
 #ifdef DISABLE_THREADS
 	gint64 count = trace_info->hit_count++;
 #else
+	static_assert (sizeof(atomic_long) == sizeof(gint64) && ATOMIC_LONG_LOCK_FREE == 2);
 	gint64 count = atomic_fetch_add ((atomic_long *)&trace_info->hit_count, 1);
```
lambdageek, line 932: @kg PTAL at this file.

This particular use of `atomic _long` is actually kind of suspect.  on some platforms it's actually `long long` (and `atomic_llong`) that is 64-bits.  I assume that's not the case on wasm?

docs/coding-guidelines/mono-code-guide.md
```diff
@@ -61,6 +61,38 @@ There are actually three boolean types to keep in mind:
 * `MonoBoolean` used as an interop type with C# bool in internal calls
 * `mono_bool` used by the public C API - generally new code shouldn't use it except when adding a new public API function.
 
+## Newer features of C11 and later
```
lambdageek, line 64: I added some coding guidelines based on what I know right now. Feedback welcome

src/mono/mono/mini/interp/jiterpreter.c
kg: Yeah this might need to be atomic_llong. I did test it originally but it's possible it's truncating it.

src/mono/mono/mini/interp/jiterpreter.c
```diff
@@ -751,11 +753,12 @@ trace_info_get (gint32 index) {
 static gint32
 trace_info_alloc () {
 #ifdef DISABLE_THREADS
-	gint32 index = trace_count++,
+	gint32 index = trace_count++;
 #else
-	gint32 index = atomic_fetch_add ((atomic_int *)&trace_count, 1),
+	static_assert (sizeof(atomic_int) == sizeof(gint32) && ATOMIC_INT_LOCK_FREE == 2);
+	gint32 index = atomic_fetch_add ((atomic_int *)&trace_count, 1);
```
AaronRobinsonMSFT, line 759: Would it be more desirable to use the target instead of the target's type for the assert?

For example:

```c
static_assert (sizeof(atomic_int) == sizeof(trace_count) && ATOMIC_INT_LOCK_FREE == 2);
```

src/mono/mono/mini/interp/jiterpreter.c
lambdageek:
```
/__w/1/s/src/mono/mono/mini/interp/jiterpreter.c:931:2: error: static assertion failed due to requirement 'sizeof(_Atomic(long)) == sizeof (trace_info->hit_count)': 
          static_assert (sizeof(atomic_long) == sizeof(trace_info->hit_count) && ATOMIC_LONG_LOCK_FREE == 2, "");
          ^              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /usr/local/emscripten/emsdk/upstream/emscripten/cache/sysroot/include/assert.h:12:23: note: expanded from macro 'static_assert'
  #define static_assert _Static_assert
                        ^
  /__w/1/s/src/mono/mono/mini/interp/jiterpreter.c:931:37: note: expression evaluates to '4 == 8'
```

yep. looks like it needs to be `atomic_llong`

AaronRobinsonMSFT (approved):
LGTM

PR #91441, opened 2023-09-01
Update CustomValidationAttribute ref to include missing properties

jeffhandley:
After applying [Fix nullable annotation for Validator.ValidateValue ref source · Pull Request #91351](https://github.com/dotnet/runtime/pull/91351), I reran `dotnet msbuild /t:GenerateReferenceAssemblySource`, and it picked up on 2 properties missing from the ref source for `CustomValidationAttribute`.

Since this is a pretty esoteric scenario and not reported by any customers, I figured we'd just fix this in net9.

### DIFF
--- a/src/libraries/System.ComponentModel.Annotations/ref/System.ComponentModel.Annotations.cs
+++ b/src/libraries/System.ComponentModel.Annotations/ref/System.ComponentModel.Annotations.cs
@@ -65,6 +65,8 @@ namespace System.ComponentModel.DataAnnotations
     {
         public CustomValidationAttribute([System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembersAttribute(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicMethods)] System.Type validatorType, string method) { }
         public string Method { get { throw null; } }
+        public override bool RequiresValidationContext { get { throw null; } }
+        public override object TypeId { get { throw null; } }
         [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembersAttribute(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicMethods)]
         public System.Type ValidatorType { get { throw null; } }
         public override string FormatErrorMessage(string name) { throw null; }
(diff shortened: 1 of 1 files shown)


tarekgh (approved):
Interesting, it looks like the appcompat tool doesn't detect the overridden members? 

CC @ericstj @ViktorHofer

src/libraries/System.ComponentModel.Annotations/ref/System.ComponentModel.Annotations.cs
```diff
@@ -65,6 +65,8 @@ public sealed partial class CustomValidationAttribute : System.ComponentModel.Da
     {
         public CustomValidationAttribute([System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembersAttribute(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicMethods)] System.Type validatorType, string method) { }
         public string Method { get { throw null; } }
+        public override bool RequiresValidationContext { get { throw null; } }
+        public override object TypeId { get { throw null; } }
```
stephentoub, line 69: CustomValidationAttribute is sealed, so these technically aren't necessary. Overrides need to be in the ref so that a derived type doing `base.Member` will call to the override rather than skipping it, but that's not an issue on sealed types.

ViktorHofer:
> Interesting, it looks like the appcompat tool doesn't detect the overridden members?

That's probably https://github.com/dotnet/sdk/issues/24390.

ViktorHofer (approved):

stephentoub:
> > Interesting, it looks like the appcompat tool doesn't detect the overridden members?
> 
> That's probably [dotnet/sdk#24390](https://github.com/dotnet/sdk/issues/24390).

As noted in https://github.com/dotnet/runtime/pull/91441#discussion_r1312477674, there isn't actually compat risk to adding or removing these, nor is there a technical benefit to doing so, because the type is sealed. If there's a special mode for flagging all discrepancies, these could be included, but we've frequently removed (or not added) overrides on sealed types.

jeffhandley:
Closing without merging since there's no benefit in adding these in.

PR #91445, opened 2023-09-01
Cleanup only changes

Maoni0:
+ fixed compiler errors for when BACKGROUND_GC is not defined

+ made a few things only defind when USE_REGIONS is not defined as they are only used for segments case

+ fixed an AV in a dprintf in descr_generations that @cshung noticed

+ reverted the dprintf change I accidently left in a previous PR that @mrsharm noticed

+ got rid of some unused methods

+ got rid of some useless, obsolete or wrong comments and move the finalize lock comment to an appropriate place

+ fixed some inappropriate line width, inconsistent formatting, unused args/fields and typos. 
there are other inconsistent formating like spacing but there are many so they should be in their own PR to not completely dwarf changes in this one.

### DIFF
--- a/src/coreclr/gc/gc.cpp
+++ b/src/coreclr/gc/gc.cpp
@@ -41,7 +41,8 @@ public:
     }
 
     // obtain random number in the range 0 .. r-1
-    static uint64_t get_rand(uint64_t r) {
+    static uint64_t get_rand(uint64_t r)
+    {
         // require r >= 0
         uint64_t x = (uint64_t)((get_rand() * r) >> 31);
         return x;
@@ -1359,11 +1360,9 @@ inline bool can_use_write_watch_for_card_table()
 #endif
 }
 
-#else
+#else //WRITE_WATCH
 #define mem_reserve (MEM_RESERVE)
 #endif //WRITE_WATCH
-
-//check if the low memory notification is supported
 
 void WaitLongerNoInstru (int i)
 {
@@ -1479,7 +1478,8 @@ enter_msl_status gc_heap::enter_spin_lock_msl_helper (GCSpinLock* msl)
                     {
                         if (VolatileLoad (&msl->lock) == lock_free || IsGCInProgress ())
                             break;
-                        YieldProcessor ();           // indicate to the processor that we are spinning
+                        // give the HT neighbor a chance to run
+                        YieldProcessor ();
                     }
                     if (VolatileLoad (&msl->lock) != lock_free && !IsGCInProgress ())
                     {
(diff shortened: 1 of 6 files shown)


src/coreclr/gc/gcpriv.h
```diff
@@ -294,7 +294,8 @@ void GCLogConfig (const char *fmt, ... );
 #define MAX_NUM_BUCKETS (MAX_INDEX_POWER2 - MIN_INDEX_POWER2 + 1)
 
 #ifdef USE_REGIONS
-#define MAX_REGION_SIZE 0x80000000
+// We constrain this as many bookkeeping fields use an int
+#define MAX_REGION_SIZE (1 << (sizeof (int32_t) * 8 - 1))
```
mangod9, line 298:
```suggestion
#define MAX_REGION_SIZE (1 << ((sizeof (int32_t) * 8) - 1))
```

src/coreclr/gc/gc.cpp
```diff
@@ -2961,8 +2960,8 @@ BOOL        gc_heap::heap_analyze_enabled = FALSE;
 
 #ifndef MULTIPLE_HEAPS
 
+alloc_list gc_heap::gen2_alloc_list[NUM_GEN2_ALIST - 1];
 alloc_list gc_heap::loh_alloc_list [NUM_LOH_ALIST-1];
```
mangod9, line 2964:
```suggestion
alloc_list gc_heap::loh_alloc_list [NUM_LOH_ALIST - 1];
```

src/coreclr/gc/gc.cpp
```diff
@@ -2961,8 +2960,8 @@ BOOL        gc_heap::heap_analyze_enabled = FALSE;
...
-alloc_list gc_heap::gen2_alloc_list[NUM_GEN2_ALIST-1];
 alloc_list gc_heap::poh_alloc_list [NUM_POH_ALIST-1];
```
mangod9, line 2965:
```suggestion
alloc_list gc_heap::poh_alloc_list [NUM_POH_ALIST - 1];
```

src/coreclr/gc/gc.cpp
```diff
@@ -29192,18 +29194,14 @@ void gc_heap::mark_phase (int condemned_gen_number, BOOL mark_only_p)
             }
         }
 
-#ifdef FEATURE_BASICFREEZE
-#ifdef USE_REGIONS
-        assert (!ro_segments_in_range);
```
mangod9, line 29197: assume removal of this assert is intentional?

Maoni0: yes `ro_segments_in_range` is not defined unless FEATURE_BASICFREEZE is defined.

src/coreclr/gc/gc.cpp
Maoni0: done

src/coreclr/gc/gc.cpp
Maoni0: done

mangod9:
There appear to be test failures related to this change.

Maoni0:
yep, my bad for missing a line. I've pushed a fix. I already observed this in my own testing when I tried things with segments.

mangod9 (approved):
LGTM, but wondering if this change will conflict with any backports to 8?

Maoni0:
> LGTM, but wondering if this change will conflict with any backports to 8?

it could since it does touch the DATAS code path a little; but I'm happy to help resolve any conflicts. I wouldn't want to not merge just because we need to resolve a few conflicts.

mangod9:
ok, thats fine then.

PR #91233, opened 2023-08-28
Use live illink to trim framework

sbomer:
https://github.com/dotnet/runtime/pull/90517 added a workaround to trim the framework using the SDK's version of ILLink.Tasks (see https://github.com/dotnet/runtime/pull/90517#issuecomment-1677995252 for context). This removes the workaround and replaces it with logic to use the live build of ILLink.Tasks.

The dependency on ILLink.Tasks is expressed via a magical `ProjectReference` that @ViktorHofer kindly helped me construct.

Unfortunately I had to add a small extension point to `Microsoft.NET.ILLink.Tasks.props`, to allow overriding the analyzer props path. This lets it be imported right out of the source tree, instead of from the build output of ILLink.Tasks.

PR #91454, opened 2023-09-01
Fix official builds for NativeAOT

ViktorHofer:
Fixes regression introduced in https://github.com/dotnet/runtime/pull/91233

The ILLink ProjectReference output shouldn't be copied to consuming projects.

From the docs:
> Private	Optional boolean. Specifies whether the reference should be copied to the output folder. This attribute matches the Copy Local property of the reference that's in the Visual Studio IDE.

From what I can see, this is only happening on NativeAOT legs because of this glob: https://github.com/dotnet/runtime/blob/91ac6b3544d53fcf9e6962575813c1cd54aa27bd/eng/liveBuilds.targets#L82-L87

### DIFF
--- a/eng/illink.targets
+++ b/eng/illink.targets
@@ -6,12 +6,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <!-- SkipGetTargetFrameworkProperties builds ILLink.Tasks without any TargetFramework settings
-         based on the referencing project, preventing errors when illink.targets is imported by a
-         project with TargetFrameworks that are incompatible with those of ILLink.Tasks. -->
+    <!-- THe assembly shouldn't be referenced, nor promoted to a package dependency, nor copied to the output directory. -->
     <ProjectReference Include="$(_ILLinkTasksSourceDir)ILLink.Tasks.csproj"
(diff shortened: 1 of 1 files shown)


eng/illink.targets
```diff
@@ -6,12 +6,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <!-- SkipGetTargetFrameworkProperties builds ILLink.Tasks without any TargetFramework settings
-         based on the referencing project, preventing errors when illink.targets is imported by a
-         project with TargetFrameworks that are incompatible with those of ILLink.Tasks. -->
+    <!-- THe assembly shouldn't be referenced, nor promoted to a package dependency, nor copied to the output directory. -->
```
akoeplinger, line 9:
```suggestion
    <!-- The assembly shouldn't be referenced, nor promoted to a package dependency, nor copied to the output directory. -->
```

akoeplinger (approved):

agocke:
Thanks! I have to say, "Private=false" means don't copy makes absolutely no sense to me, but at least it works!

ViktorHofer:
At least it's documented ;)

Issue #91394, opened 2023-08-31
[Mono] Add missing SIMD support for some methods from System.Numerics.Vector2/3/4 and Quaternion on Arm64


matouskozak:
### Vector2
- [x] SN_CopyTo - CoreClr doesn't intrinsify this API
- [x] SN_Distance
- [x] SN_DistanceSquared
- [x] SN_Length
- [x] SN_LengthSquared
- [x] SN_Lerp
- [x] SN_Normalize

### Vector3
- [x] SN_CopyTo - CoreClr doesn't intrinsify this API
- [x] SN_Distance
- [x] SN_DistanceSquared
- [x] SN_Length
- [x] SN_LengthSquared
- [x] SN_Lerp
- [x] SN_Normalize

### Vector4
- [x] SN_CopyTo - CoreClr doesn't intrinsify this API
- [x] SN_Distance
- [x] SN_DistanceSquared
- [x] SN_Length
- [x] SN_LengthSquared
- [x] SN_Lerp
- [x] SN_Normalize

### Quaternion
- [x] Conjugate

fanyang-mono:
This issue was completed via https://github.com/dotnet/runtime/pull/98761

Issue #73462, opened 2022-08-05
[Mono] Add SIMD support for System.Numerics.Vector2 on Arm64


fanyang-mono:
Methods:

- [x] ctor
- [x] Abs
- [x] Dot
- [x] GetElement
- [x] Max
- [x] Min
- [x] SquareRoot
- [x] WithElement
- [x] get_Item
- [x] get_One
- [x] get_Zero
- [x] set_Item

Operators:

- [x] Additon
- [x] Division
- [x] Equality
- [x] Inequality
- [x] Multiply
- [x] Subtraction

fanyang-mono:
Couldn't get to this within .NET8, due to other higher priority projects.

PR #91455, opened 2023-09-01
[WIP][mono] Basic SIMD support for System.Numerics.Vector2 on arm64

matouskozak:
This PR adds basic SIMD support for `System.Numerics.Vector2` on arm64. Equaling the current support present for `System.Numerics.Vector4`.

Current SIMD support for `Vector2` with mini/llvm:
- SN_ctor
- SN_Abs
- SN_Add
- SN_Clamp
- SN_Divide (currently disabled `Vector2` / `float` scenario, will enable in the next PR)
- SN_Dot
- SN_Max
- SN_Min
- SN_Multiply
- SN_Negate
- SN_SquareRoot
- SN_Subtract
- SN_get_Item
- SN_get_One
- SN_get_UnitX
- SN_get_UnitY
- SN_get_Zero
- SN_op_Addition
- SN_op_Division
- SN_op_Equality
- SN_op_Inequality
- SN_op_Multiply
- SN_op_Subtraction
- SN_op_UnaryNegation
- SN_set_Item

Future work on the missing intrinsic is tracked here https://github.com/dotnet/runtime/issues/91394.
Contributes to: https://github.com/dotnet/runtime/issues/73462

---
p.s. These getters currently use 128-bit code paths for emitting const values (`emit_xconst_v128`) even for Vector2 (64-bit vector):
- SN_get_UnitX
- SN_get_UnitY
- SN_get_One

While it looks like the generated code is okay locally, it could potentially cause problems in `mono_arch_emit_exceptions` or in subsequent exception handling https://github.com/dotnet/runtime/blob/c12429abaa1762dc971d6e4ef2f34e7140c1b364/src/mono/mono/mini/mini-arm64.c#L6092-L6094

@fanyang-mono @jandupej Any thoughts on that? Is dedicated 64-bit const value emit worth implementing?


matouskozak:
More paths where 128-bit vector paths are used detected (e.g. `OP_XZERO`). Changing to WIP and I'll investigate further.

jandupej:
> These getters currently use 128-bit code paths for emitting const values...

You can use a `fmov` to flood the lower two floats with `1.0f`. This gives you the fastest `SN_get_One` possible (there is a 64-bit variant of this, with `q=0`). To make `SN_get_UnitX/Y` you can shift the vector left or right as doubles by 32. Zeros are shifted in, so this will give you a `(0.0f, 1.0f)` or reverse. This will destroy the upper 64 bits of the register, but it shouldn't be a problem as only the lower 64 bits are of importance.

PR #91463, opened 2023-09-01
[browser] [wasm] remove special StringContent handling

### LINKED ISSUE #91379: [browser] [wasm] Remove StringContent special case
While working on #91295 I noticed that there is a special handling for `StringContent`:

https://github.com/dotnet/runtime/blob/6d3be9e69dc385722851b0c165255cc4828aec7c/src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs#L214-L227

But because `StringContent` inherits from `ByteArrayContent` the string is actually stored as a byte array, which means that `ReadAsStringAsync` actually allocates a new string an thus needs more memory then a call to `ReadAsByteArrayAsync` would need and is also slower.
Since the browser supports sending byte array anyway I would propose to remove this special case. E.g. like this: https://github.com/campersau/runtime/commit/c887f0caf6b0329ef10272f1b8b0d4f41c7c5ec8

Currently:
string > StringContent > new byte array (encoding)
byte array > ReadAsStringAsync > new string (encoding)
string > Fetch > "complex" string to js Marshal (encoding?)

Proposed:
string > StringContent > new byte array (encoding)
byte array > ReadAsByteArrayAsync > new byte array
byte array > Fetch > easy byte array to js

Or was there any other reason why this was special cased in the first place?
Edit: Looks like it was there from the beginning: https://github.com/mono/mono/pull/10464

|               Method |     Mean |    Error |   StdDev |   Gen0 | Allocated |
|--------------------- |---------:|---------:|---------:|-------:|----------:|
|    ReadAsStringAsync | 570.8 ns | 11.39 ns | 13.99 ns | 0.0973 |    1280 B |
| ReadAsByteArrayAsync | 187.0 ns |  3.73 ns |  6.43 ns | 0.0525 |     688 B |

<details>
<summary>Benchmark</summary>

```c#
[MemoryDiagnoser]
public class Program
{
    {
        BenchmarkRunner.Run<Program>(args: args);
    }

    private StringContent content;

    [GlobalSetup]
    public void Setup()
    {
        content = new StringContent("Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.");
    }

    [Benchmark]
    public async Task ReadAsStringAsync()
    {
        await content.ReadAsStringAsync();
    }

    [Benchmark]
    public async Task ReadAsByteArrayAsync()
    {
        await content.ReadAsByteArrayAsync();
    }

}
```

</details>

---

If we really want to remove allocations I would propose adding an `internal virtual bool TryGetMemory(out ReadOnlyMemory<byte> memory);` method to `HttpContent` and implement it for `ByteArrayContent` / `StringContent` / `EmptyContent` / `FormUrlEncodedContent` / `ReadOnlyMemoryContent` and special case this instead.

campersau:
Fixes #91379

### DIFF
--- a/src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
+++ b/src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
@@ -212,37 +212,27 @@ namespace System.Net.Http
                 cancellationToken.ThrowIfCancellationRequested();
                 if (request.Content != null)
                 {
-                    if (request.Content is StringContent)
+                    bool streamingEnabled = false;
+                    if (BrowserHttpInterop.SupportsStreamingRequest())
                     {
-                        string body = await request.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(true);
+                        request.Options.TryGetValue(EnableStreamingRequest, out streamingEnabled);
+                    }
+
+                    if (streamingEnabled)
+                    {
+                        Stream stream = await request.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(true);
                         cancellationToken.ThrowIfCancellationRequested();
 
-                        promise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, body);
+                        ReadableStreamPullState pullState = new ReadableStreamPullState(stream, cancellationToken);
+
+                        promise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, ReadableStreamPull, pullState);
                     }
                     else
                     {
-                        bool streamingEnabled = false;
-                        if (BrowserHttpInterop.SupportsStreamingRequest())
-                        {
-                            request.Options.TryGetValue(EnableStreamingRequest, out streamingEnabled);
-                        }
+                        byte[] buffer = await request.Content.ReadAsByteArrayAsync(cancellationToken).ConfigureAwait(true);
+                        cancellationToken.ThrowIfCancellationRequested();
 
-                        if (streamingEnabled)
-                        {
-                            Stream stream = await request.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(true);
-                            cancellationToken.ThrowIfCancellationRequested();
-
-                            ReadableStreamPullState pullState = new ReadableStreamPullState(stream, cancellationToken);
-
-                            promise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, ReadableStreamPull, pullState);
-                        }
-                        else
-                        {
-                            byte[] buffer = await request.Content.ReadAsByteArrayAsync(cancellationToken).ConfigureAwait(true);
-                            cancellationToken.ThrowIfCancellationRequested();
-
-                            promise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, buffer);
-                        }
+                        promise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, buffer);
                     }
                 }
                 else
--- a/src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpInterop.cs
+++ b/src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpInterop.cs
@@ -69,8 +69,7 @@ namespace System.Net.Http
             string[] headerValues,
             string[] optionNames,
             [JSMarshalAs<JSType.Array<JSType.Any>>] object?[] optionValues,
-            JSObject abortControler,
-            string? body = null);
+            JSObject abortControler);
 
         [JSImport("INTERNAL.http_wasm_fetch_stream")]
         public static partial Task<JSObject> Fetch(
--- a/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
+++ b/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
@@ -285,10 +285,8 @@ namespace System.Net.Http.Functional.Tests
         }
 
         [OuterLoop]
-        [ConditionalTheory(typeof(PlatformDetection), nameof(PlatformDetection.IsBrowser))]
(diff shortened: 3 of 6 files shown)


src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpInterop.cs
```diff
@@ -69,8 +69,7 @@ public static void GetResponseHeaders(JSObject fetchResponse, HttpHeaders respos
             string[] headerValues,
             string[] optionNames,
             [JSMarshalAs<JSType.Array<JSType.Any>>] object?[] optionValues,
-            JSObject abortControler,
-            string? body = null);
+            JSObject abortControler);
```
pavelsavara, line 72: I think this is now un-used, no ?

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
```diff
@@ -212,37 +212,27 @@ private static async Task<WasmFetchResponse> CallFetch(HttpRequestMessage reques
...
-                        else
-                        {
-                            byte[] buffer = await request.Content.ReadAsByteArrayAsync(cancellationToken).ConfigureAwait(true);
-                            cancellationToken.ThrowIfCancellationRequested();
+                        byte[] buffer = await request.Content.ReadAsByteArrayAsync(cancellationToken).ConfigureAwait(true);
+                        cancellationToken.ThrowIfCancellationRequested();
 
-                            promise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, buffer);
-                        }
+                        promise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, buffer);
```
pavelsavara, line 235: What would happen to encoding UTF8/UTF16/WTF16? 
Right now it's JS interop doing the conversion to JS string and the browser is doing the rest.
https://github.com/dotnet/runtime/blob/f6b6415119b4d6d4421b3186a6a8aaa4803af969/src/mono/wasm/runtime/strings.ts#L71-L78

What would happen to `Content-Encoding` header ?

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpInterop.cs
campersau: It is needed for requests without content like `GET`.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
campersau: The `Content-Encoding` isn't set automatically and is usually for `gzip` / `br` etc. Or am I wrong?

The string is getting encoded with the encoding specified by the constructor of `StringContent` and stored as bytes because `StringContent` actually inherits from `ByteArrayContent`.

https://github.com/dotnet/runtime/blob/f6b6415119b4d6d4421b3186a6a8aaa4803af969/src/libraries/System.Net.Http/src/System/Net/Http/StringContent.cs#L56-L60

https://github.com/dotnet/runtime/blob/f6b6415119b4d6d4421b3186a6a8aaa4803af969/src/libraries/System.Net.Http/src/System/Net/Http/HttpContent.cs#L25

campersau:
There is one behavior change when someone is using this `StringContent` constructor 
https://github.com/dotnet/runtime/blob/f6b6415119b4d6d4421b3186a6a8aaa4803af969/src/libraries/System.Net.Http/src/System/Net/Http/StringContent.cs#L56-L60
and passes `null` for `mediaType`, then the browser wouldn't send a `Content-Type` anymore.

With the old logic the browser would send a `Content-Type` because of this browser behavior:
> When a string is used as a body, it conveniently sets the Content-Type header to text/plain;charset=UTF-8.

So with the new logic it would actually do what the developer wants and can also be seen as a bugfix.

---

Example:

```c#
new StringContent("Hello World", null, (System.Net.Http.Headers.MediaTypeHeaderValue)null);
```

Before (browser sets `Content-Type: text/plain;charset=UTF-8` because we send the value as `string`):
```
Accept: */*
Host: 127.0.0.1:36165
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/116.0.5845.0 Safari/537.36
Accept-Encoding: gzip, deflate, br
Content-Type: text/plain;charset=UTF-8
Origin: http://127.0.0.1:39987
Referer: http://127.0.0.1:39987/
Content-Length: 11
sec-ch-ua: \"Chromium\";v=\"116\ \"Not)A;Brand\";v=\"24\ \"HeadlessChrome\";v=\"116\"
sec-ch-ua-platform: \"Linux\"
sec-ch-ua-mobile: ?0
sec-fetch-site: cross-site
sec-fetch-mode: cors
sec-fetch-dest: empty
```

After (there is no `Content-Type` header):
```
Accept: */*
Host: 127.0.0.1:46361
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/116.0.5845.0 Safari/537.36
Accept-Encoding: gzip, deflate, br
Origin: http://127.0.0.1:43851
Referer: http://127.0.0.1:43851/
Content-Length: 11
sec-ch-ua: \"Chromium\";v=\"116\ \"Not)A;Brand\";v=\"24\ \"HeadlessChrome\";v=\"116\"
sec-ch-ua-mobile: ?0
sec-ch-ua-platform: \"Linux\"
sec-fetch-site: cross-site
sec-fetch-mode: cors
sec-fetch-dest: empty
```

pavelsavara:
> So with the new logic it would actually do what the developer wants and can also be seen as a bugfix.

Sounds good.

pavelsavara (approved):
LGTM, @maraf any comments ?

maraf (approved):
Looks good to me 👍

PR #91465, opened 2023-09-01
Improve DevContainer configuration for Codespaces

akoeplinger:
Adds a default configuration that comes with all the dependencies preinstalled, but not prebuilt. This reduces confusion if you don't use the "New with options..." dropdown when creating the Codespace since you'd just get a standard Ubuntu container then.

Also install a specific version of dotnet-serve.


.devcontainer/wasm/devcontainer.json
```diff
@@ -1,6 +1,6 @@
 // For format details, see https://aka.ms/devcontainer.json.
 {
-	"name": "C# (.NET)",
+	"name": "WASM development (prebuilt)",
 	"build": {
 		"dockerfile": "Dockerfile",
```
ilonatommy, line 5:
```suggestion
		"dockerfile": "../Dockerfile",
```

ilonatommy (approved):

PR #91468, opened 2023-09-01
Remove dependencies on non-live illink

sbomer:


jkoritzinsky (commented):
Can you take a look in targetingpacks.targets and update the KnownILLink item changes to point to the live ILLink?

ViktorHofer (commented):
First of all: Amazing change!!

Slightly unrelated to this PR but still important from my point of view: Is there a feature flag that we can condition the illink.targets import on so that only project that actually use the linker, import the file and with it build the P2P to ILLink.Tasks.csproj?

eng/liveILLink.targets
```diff
@@ -0,0 +1,29 @@
...
+    <_ILLinkTasksSourceDir>$(ToolsProjectRoot)illink\src\ILLink.Tasks\</_ILLinkTasksSourceDir>
+    <ILLinkAnalyzersPropsPath>$(ToolsProjectRoot)illink\src\ILLink.RoslynAnalyzer\build\Microsoft.NET.ILLink.Analyzers.props</ILLinkAnalyzersPropsPath>
+  </PropertyGroup>
+
+  <Import Project="$(_ILLinkTasksSourceDir)build\Microsoft.NET.ILLink.Tasks.props" />
+
+  <PropertyGroup>
+    <!-- Don't use SDK's trimming functionality. -->
+    <_RequiresILLinkPack>false</_RequiresILLinkPack>
+    <ILLinkTasksAssembly>$(ToolsILLinkDir)$(NetCoreAppToolCurrent)/ILLink.Tasks.dll</ILLinkTasksAssembly>
```
ViktorHofer, line 13: Please use a windows separator char. We standardized on using that in anything under src/libraries and that file is imported by our projects.

```suggestion
    <ILLinkTasksAssembly>$(ToolsILLinkDir)$(NetCoreAppToolCurrent)\ILLink.Tasks.dll</ILLinkTasksAssembly>     
```

eng/Analyzers.targets
```diff
@@ -14,7 +14,8 @@
   <PropertyGroup Condition="'$(RunAnalyzers)' != 'false'">
     <EnableSingleFileAnalyzer Condition="
       '$(EnableSingleFileAnalyzer)' == '' And
-      '$(TargetFrameworkIdentifier)' == '.NETCoreApp'">true</EnableSingleFileAnalyzer>
+      '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
+      '$(IsGeneratorProject)' != 'true'">true</EnableSingleFileAnalyzer>
```
ViktorHofer, line 18: What about IsReferenceSourceProject?

eng/Analyzers.targets
```diff
@@ -14,7 +14,8 @@
   <PropertyGroup Condition="'$(RunAnalyzers)' != 'false'">
     <EnableSingleFileAnalyzer Condition="
       '$(EnableSingleFileAnalyzer)' == '' And
-      '$(TargetFrameworkIdentifier)' == '.NETCoreApp'">true</EnableSingleFileAnalyzer>
+      '$(TargetFrameworkIdentifier)' == '.NETCoreApp' And
```
ViktorHofer, line 17: nit: I know this was already there before but please use lowercase conditional operators. Just style. I would really like msbuild to add a style analyzer so that we don't have the inconsistencies in our repo.

```suggestion
      '$(TargetFrameworkIdentifier)' == '.NETCoreApp' and
```

src/libraries/Directory.Build.targets
```diff
@@ -128,6 +128,8 @@
   <Import Project="$(RepositoryEngineeringDir)slngen.targets" Condition="'$(IsSlnGen)' == 'true'" />
 
   <Import Project="$(RepositoryEngineeringDir)illink.targets" Condition="'$(IsSourceProject)' == 'true' or '$(ExplicitlyImportCustomILLinkTargets)' == 'true'" />
+  <!-- liveILLink.targets should be conditionally imported after illink.targets (which also imports liveILLink.targets) to prevent a double-import. -->
+  <Import Project="$(RepositoryEngineeringDir)liveILLink.targets" Condition="'$(_ImportedLiveILLinkTargets)' != 'true'" />
```
ViktorHofer, line 132: Changes look really good except for these two lines (which I don't understand). illink.targets already imports liveILLink.targets. Why do we need to import it here explicitly?

ViktorHofer (approved):

eng/Analyzers.targets
ViktorHofer: Yes, I don't think we should have this analyzer enabled for test projects at the moment. So, I would prefer if you keep your change.

ViktorHofer:
> Assets file '/Users/runner/work/1/s/artifacts/obj/ILLink.Tasks/project.assets.json' doesn't have a target for 'net8.0/osx-x64'

src/coreclr/Directory.Build.targets
```diff
@@ -26,7 +27,8 @@
           UseHardlinksIfPossible="false" />
   </Target>
 
-  <!-- Import targets here to have TargetPath and other macros defined. Limit to CoreLib. -->
-  <Import Condition="'$(MSBuildProjectName)' == 'System.Private.CoreLib' or '$(IsNativeAotProject)' == 'true'" Project="$(RepositoryEngineeringDir)illink.targets" />
+    <!-- Import targets here to have TargetPath and other macros defined. Limit to CoreLib. -->
+    <Import Condition="'$(MSBuildProjectName)' == 'System.Private.CoreLib' or '$(IsNativeAotProject)' == 'true'" Project="$(RepositoryEngineeringDir)illink.targets" />
+    <Import Project="$(RepositoryEngineeringDir)liveILLink.targets" />
```
ViktorHofer, line 32: Apparently this causes crossgen2_publish.proj to reference the linker projects now. These projects flow some global properties between each other which causes binclashes, race conditions and in this case, restore errors because the project.assets.json file doesn't list a required RID.

see https://github.com/dotnet/runtime/pull/92826/files#r1341932080 for more details

Can we further fine tune this import to only apply to the projects that really need the live linker?

ViktorHofer: Please look at a binlog and see if all the projects that now reference the linker project actually need it. I'm not sure about the crossgen2_publish case though.

jkoritzinsky: Crossgen2_publish needs the linker for platforms that don’t support NativeAOT.

ViktorHofer, line 32: nit: indentation

```suggestion
  <!-- Import targets here to have TargetPath and other macros defined. Limit to CoreLib. -->
  <Import Condition="'$(MSBuildProjectName)' == 'System.Private.CoreLib' or '$(IsNativeAotProject)' == 'true'" Project="$(RepositoryEngineeringDir)illink.targets" />
  <Import Project="$(RepositoryEngineeringDir)liveILLink.targets" />
```

src/libraries/Directory.Build.targets
```diff
@@ -127,6 +127,7 @@
   <Import Project="$(RepositoryEngineeringDir)testing\coverage.targets" Condition="'$(EnableRunSettingsSupport)' == 'true' or '$(EnableCoverageSupport)' == 'true'" />
 
   <Import Project="$(RepositoryEngineeringDir)illink.targets" Condition="'$(IsSourceProject)' == 'true' or '$(ExplicitlyImportCustomILLinkTargets)' == 'true'" />
+  <Import Project="$(RepositoryEngineeringDir)liveILLink.targets" />
```
ViktorHofer, line 130: Would it make sense to only import this file when the illink.targets file is imported?

```suggestion
  <ImportGroup Condition="'$(IsSourceProject)' == 'true' or '$(ExplicitlyImportCustomILLinkTargets)' == 'true'">
    <Import Project="$(RepositoryEngineeringDir)illink.targets" />
    <Import Project="$(RepositoryEngineeringDir)liveILLink.targets" />
  </ImportGroup>
```

ViktorHofer (approved):

eng/liveILLink.targets
```diff
@@ -0,0 +1,42 @@
...
+    <_ILLinkTasksSourceDir>$(ToolsProjectRoot)illink\src\ILLink.Tasks\</_ILLinkTasksSourceDir>
+    <ILLinkAnalyzersPropsPath>$(ToolsProjectRoot)illink\src\ILLink.RoslynAnalyzer\build\Microsoft.NET.ILLink.Analyzers.props</ILLinkAnalyzersPropsPath>
+  </PropertyGroup>
+
+  <Import Project="$(_ILLinkTasksSourceDir)build\Microsoft.NET.ILLink.Tasks.props" Condition="'$(_RequiresLiveILLink)' == 'true'" />
+
+  <PropertyGroup Condition="'$(_RequiresLiveILLink)' == 'true'">
+    <!-- Don't use SDK's trimming functionality. -->
+    <_RequiresILLinkPack>false</_RequiresILLinkPack>
+    <ILLinkTasksAssembly>$(ToolsILLinkDir)$(NetCoreAppToolCurrent)\ILLink.Tasks.dll</ILLinkTasksAssembly>
```
ViktorHofer, line 26: This needs to respect the MSBuildRuntimeType and pick the .NET Framework one, same as https://github.com/dotnet/runtime/blob/d3569b99ead8f909556f5b416b6f57d94f30e668/eng/illink.targets#L15-L17

ViktorHofer:
```suggestion
    <ILLinkTasksAssembly Condition="'$(MSBuildRuntimeType)' == 'Core'">$(ToolsILLinkDir)$(NetCoreAppToolCurrent)\ILLink.Tasks.dll</ILLinkTasksAssembly>
    <ILLinkTasksAssembly Condition="'$(MSBuildRuntimeType)' != 'Core'">$(ToolsILLinkDir)$(NetFrameworkToolCurrent)\ILLink.Tasks.dll</ILLinkTasksAssembly>
```

eng/targetingpacks.targets
jkoritzinsky, line 1: We should update the logic in this file that removes the PackageReference to ILCompiler to also remove the reference to the ILLink packages that can be pulled in.

ViktorHofer:
Looks like there is a remaining failure in the NativeAOT leg .

jkoritzinsky (approved):

ViktorHofer (approved):

PR #84379, opened 2023-04-05
Simplify target to enable config binding src generator from NuGet package, ensure it runs for all TFMs, and rename enabling property

layomia:
- Renders https://github.com/dotnet/runtime/issues/83532 obsolete.
- Renames enabling property from `EnableMicrosoftExtensionsConfigurationBinderGenerator` to `EnableConfigurationBindingGenerator` (fyi @eerhardt).

PR #91469, opened 2023-09-01
Enable interceptors implicitly in binder gen nupkg when generator is enabled

### LINKED ISSUE #91419: Update config binding gen nupkg transitive build target to use granular interceptor opt-in
- Implicitly enable interceptors for projects with `<EnableConfigurationBindingGenerator>true<... />`
- Corresponding nupkg change for https://github.com/dotnet/sdk/pull/35024

@dotnet/area-extensions-configuration @captainsafia @eerhardt.

layomia:
- Fixes https://github.com/dotnet/runtime/issues/91419. RC-2 candidate.
- Interceptors feature is required for source-gen to work; enable it implicitly using granular opt-in (i.e only for the namespace that we emit binding logic into)
- Corresponding web SDK change https://github.com/dotnet/sdk/pull/35024/
- Not blocked on the new Roslyn version that contains the granular feature flowing into the SDK (https://github.com/dotnet/sdk/pull/35018).

@dotnet/area-extensions-configuration @dotnet/area-infrastructure-libraries  @captainsafia

### DIFF
--- a/src/libraries/Microsoft.Extensions.Configuration.Binder/src/buildTransitive/Microsoft.Extensions.Configuration.Binder.targets
+++ b/src/libraries/Microsoft.Extensions.Configuration.Binder/src/buildTransitive/Microsoft.Extensions.Configuration.Binder.targets
@@ -1,4 +1,9 @@
 <Project InitialTargets="NETStandardCompatError_Microsoft_Extensions_Configuration_Binder">
+    <PropertyGroup Condition="'$(EnableConfigurationBindingGenerator)' == 'true'">
+        <!-- The configuration binding source generator uses a preview version of the compiler interceptors feature. Enable it implicitly when the generator is enabled. -->
+        <InterceptorsPreviewNamespaces>$(InterceptorsPreviewNamespaces);Microsoft.Extensions.Configuration.Binder.SourceGeneration</InterceptorsPreviewNamespaces>
+    </PropertyGroup>
+
     <Target Name="_Microsoft_Extensions_Configuration_Binder_RemoveAnalyzer"
             Condition="'$(EnableConfigurationBindingGenerator)' != 'true'"
             AfterTargets="ResolvePackageDependenciesForBuild;ResolveNuGetPackageAssets">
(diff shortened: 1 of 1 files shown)


tarekgh (approved):
LGTM. 

Did we test this manually? Is there a way we can add a test for this? This shouldn't block this PR though.

layomia:
Yes this was manually tested; like the previous change (https://github.com/dotnet/runtime/pull/84379).

Unlike the SDK (e.g. in https://github.com/dotnet/sdk/pull/35024), we don't seem to have a pattern for testing inbox/nupkg target changes.

@ViktorHofer do you have thoughts on possibilities here (for future/follow-up) ?

src/libraries/Microsoft.Extensions.Configuration.Binder/src/buildTransitive/Microsoft.Extensions.Configuration.Binder.targets
```diff
@@ -8,6 +8,14 @@
         </ItemGroup>
     </Target>
 
+    <Target Name="_Microsoft_Extensions_Configuration_Binder_EnableInterceptors"
```
captainsafia, line 11: It looks like we're not removing the old feature flag in this change set. From looking at [this commit](https://github.com/dotnet/runtime/blob/release/8.0-rc1/src/libraries/Microsoft.Extensions.Configuration.Binder/src/buildTransitive/Microsoft.Extensions.Configuration.Binder.targets) it doesn't look like it was ever included.

I assume this means that behavior for RC1 + NuPkg will mean that users have to manually set the flag themselves?

layomia: That's right. For RC-1 the question of enabling interceptors implicitly (along with the other dimension of behavior diff between web SDK vs not) was still very open. I decided not to pursue adding the implicit opt-in here given the concerns around the global opt-in.

I think the breaking change with new build errors might be net positive. IMO opinion it might be interesting to see the customer reaction to the warnings. We should probably add breaking change notes to the RC-1 blog though.

captainsafia: > We should probably add breaking change notes to the RC-1 blog though.

Yes, agreed.

src/libraries/Microsoft.Extensions.Configuration.Binder/src/buildTransitive/Microsoft.Extensions.Configuration.Binder.targets
```diff
@@ -8,6 +8,14 @@
...
+            Condition="'$(EnableConfigurationBindingGenerator)' = 'true'"
+            AfterTargets="ResolvePackageDependenciesForBuild;ResolveNuGetPackageAssets">
+            
+        <!-- The configuration binding source generator uses a preview version of the compiler interceptors feature. Enable it implicitly when the generator is enabled. -->
+        <InterceptorsPreviewNamespaces>$(InterceptorsPreviewNamespaces);Microsoft.Extensions.Configuration.Binder.SourceGeneration</InterceptorsPreviewNamespaces>
```
ViktorHofer, line 16: I'm surprised this is valid msbuild syntax. Is this a property or an item? I would expect these this element to be in a property group.

layomia: Pushed stale commit.

src/libraries/Microsoft.Extensions.Configuration.Binder/src/buildTransitive/Microsoft.Extensions.Configuration.Binder.targets
layomia: https://github.com/dotnet/core/issues/8439.

ViktorHofer (approved):

src/libraries/Microsoft.Extensions.Configuration.Binder/src/buildTransitive/Microsoft.Extensions.Configuration.Binder.targets
```diff
@@ -8,6 +8,16 @@
         </ItemGroup>
     </Target>
 
+    <Target Name="_Microsoft_Extensions_Configuration_Binder_EnableInterceptors"
+            Condition="'$(EnableConfigurationBindingGenerator)' == 'true'"
+            AfterTargets="ResolvePackageDependenciesForBuild;ResolveNuGetPackageAssets">
```
ViktorHofer, line 13: Why do you need an AfterTargets here to just a property? That doesn't make sense.

ViktorHofer: Do you even need a target here at all?

ViktorHofer (approved):

PR #48652, opened 2021-02-23
Import iOS from MacCatalyst in libraries OSGroups.json

akoeplinger:
This allows us to reuse the existing iOS build configs since we don't have MacCatalyst-specific logic yet.

PR #47823, opened 2021-02-03
Add Mac Catalyst (iOS API, Mac ABI) runtime

directhex:
This ties into several concurrent discussions:

https://github.com/dotnet/runtime/pull/47645
https://github.com/dotnet/runtime/issues/47517
https://github.com/dotnet/runtime/issues/47518
https://github.com/dotnet/runtime/issues/44882
https://github.com/dotnet/designs/pull/174
https://github.com/dotnet/runtime/issues/47768
https://github.com/dotnet/xamarin/issues/29

I've made a few assumptions for now:

* RIDs will be `maccatalyst-x64` and `maccatalyst-arm64`
* TFM is `net6.0-maccatalyst`
* `TargetsMacCatalyst` gets defined in the build
* `IsMacCatalyst()` and `IsMacCatalystVersionAtLeast()` get added to System.Runtime (an API change in need of review)
* `--os maccatalyst` gets passed to build.sh
* xharness support for `maccatalyst` https://github.com/dotnet/xharness/issues/435

It should work well enough right now to happily build a `maccatalyst-x64` runtime pack. `maccatalyst-arm64` requires an LLVM build, which we don't have as of today, plus a newer macOS/Xcode than my current dev machine. I know I've made some incorrect assumptions regarding ARM64 (e.g. the target iOS version for Catalyst needs to map to macOS 11.0+, but the number I'm using maps to 10.15.1)

PR #91473, opened 2023-09-01
[maccatalyst] Make sure MacProxy is included in System.Net.Http

### LINKED ISSUE #90258: ClientWebSocket.ConnectAsync() does not work on Mac
### Description

in a simple MAUI test app, call to `ClientWebSocket.ConnectAsync()` on Mac does not return. 
some experimental work suggests that MAUI is defaulting to `HttpNoProxy`:
when we don’t set proxy to `ClientWebSocket` Option the MAUI `ClientWebSocket` is defaulting to `NoProxy` and the connect fails. I checked the values of `HttpClient.DefaultProxy`:
In MAUI app:
```c#
  HttpClient.DefaultProxy    {System.Net.Http.HttpNoProxy}
```

comparing that to .Net7 App, it's set to `MacProxy`:
```c#
  HttpClient.DefaultProxy    {System.Net.Http.MacProxy}  
```
for reference:
https://github.com/dotnet/dotnet/blob/main/src/runtime/src/libraries/System.Net.WebSockets.Client/src/System/Net/WebSockets/WebSocketHandle.Managed.cs#L35

in customer situation, WPAD (Web Proxy Auto-Discovery) is used to configure internal laptops. 
while testing on MS VPN, the code works. It's worth considering whether the MS VPN proxy setup differs from WPAD, 

we need to get Proxy Url/object into MAUI app for all different VPN configuration and create `WebProxy` object and set it to `ClientWebSocket`.

### Steps to Reproduce

download the repro to Mac connected via VPN with WPAD configured.
in the repro app, click on button to start test. the call to `ClientWebSocket.ConnectAsync()` will be stuck 

### Link to public reproduction project repository

https://github.com/linkaiyu/ClientWebSocket

### Version with bug

7.0.49

### Last version that worked well

Unknown/Other

### Affected platforms

macOS

### Affected platform versions

_No response_

### Did you find any workaround?

short term workaround is to create a webproxy for the connection, but for long term when the app is going to be released to the public, the short term solution won't meet the needs

### Relevant log output

_No response_

steveisok:
Fixes #90258

#90258 identified that `HttpClient.DefaultProxy` returned `System.Net.Http.HttpNoProxy` when using MacCatalyst. This is due to System.Net.Http.csproj not having maccatalyst as a condition when including MacProxy sources.

Way back when, #47823 had the condition, but this appears to have changed in #48652 when we thought MacCatalyst would inherit settings from iOS. That turned out to not be the case and this check was never added back.

### DIFF
--- a/src/libraries/System.Net.Http/src/System.Net.Http.csproj
+++ b/src/libraries/System.Net.Http/src/System.Net.Http.csproj
@@ -316,11 +316,11 @@
     <Compile Include="System\Net\Http\SocketsHttpHandler\HttpEnvironmentProxy.Unix.cs" />
     <Compile Include="System\Net\Http\SocketsHttpHandler\CurrentUserIdentityProvider.Unix.cs" />
   </ItemGroup>
-  <ItemGroup Condition="'$(TargetPlatformIdentifier)' != '' and '$(TargetPlatformIdentifier)' != 'windows' and '$(TargetPlatformIdentifier)' != 'browser' and '$(TargetPlatformIdentifier)' != 'osx' and '$(TargetPlatformIdentifier)' != 'ios' and '$(TargetPlatformIdentifier)' != 'tvos'">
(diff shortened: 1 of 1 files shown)


simonrozsival (approved):

MihaZupan:
https://github.com/dotnet/runtime/blob/e3925e345acf8b9522c204b188def0741d45454b/src/libraries/Common/tests/System/Net/Http/HttpClientHandlerTest.Proxy.cs#L120 looks like it could be the relevant test.

PR #91474, opened 2023-09-01
Condition EnableSingleFileAnalyzer

agocke:
With the latest SDK, a warning is generated if the analyzer is enabled on an unsupported SDK.

### DIFF
--- a/eng/Analyzers.targets
+++ b/eng/Analyzers.targets
@@ -14,7 +14,7 @@
   <PropertyGroup Condition="'$(RunAnalyzers)' != 'false'">
     <EnableSingleFileAnalyzer Condition="
       '$(EnableSingleFileAnalyzer)' == '' And
-      '$(TargetFrameworkIdentifier)' == '.NETCoreApp'">true</EnableSingleFileAnalyzer>
+      $([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">true</EnableSingleFileAnalyzer>
   </PropertyGroup>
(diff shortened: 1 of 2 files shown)


hoyosjs (approved):

eng/Analyzers.targets
```diff
@@ -14,7 +14,7 @@
   <PropertyGroup Condition="'$(RunAnalyzers)' != 'false'">
     <EnableSingleFileAnalyzer Condition="
       '$(EnableSingleFileAnalyzer)' == '' And
-      '$(TargetFrameworkIdentifier)' == '.NETCoreApp'">true</EnableSingleFileAnalyzer>
+      $([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">true</EnableSingleFileAnalyzer>
```
ViktorHofer, line 17: I don't understand this change. dotnet/runtime main only builds net8.0 and net9.0. Why do we need this check?

ViktorHofer (changes requested):
Please respond to my question before merging as I don't think this is right.

eng/Analyzers.targets
agocke: Are you sure? I see a lot in Microsoft.Extensions that has netstandard2.0

ViktorHofer: Exactly. The `TFI == .NETCoreApp` check should already take care of only applying to .NETCoreApp TFMs.

agocke:
Hmm, can't repro this anymore... I wonder if it was a stale interaction.

PR #91483, opened 2023-09-01
Fix usage of load library

AaronRobinsonMSFT:
Use `LoadLibraryExW` instead of `LoadLibraryW` on Windows.

### DIFF
--- a/src/coreclr/debug/di/process.cpp
+++ b/src/coreclr/debug/di/process.cpp
@@ -181,7 +181,7 @@ STDAPI DLLEXPORT OpenVirtualProcessImpl2(
     CLR_DEBUGGING_PROCESS_FLAGS* pFlagsOut)
 {
 #ifdef TARGET_WINDOWS
-    HMODULE hDac = LoadLibraryExW(pDacModulePath, NULL, LOAD_LIBRARY_SEARCH_SYSTEM32);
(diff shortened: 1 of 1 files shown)


mikem8361 (approved):

elinor-fung (approved):

tommcdon (approved):

jkoritzinsky (approved):

AaronRobinsonMSFT:
Failure is unrelated

src/coreclr/debug/di/process.cpp
```diff
@@ -180,7 +180,11 @@ STDAPI DLLEXPORT OpenVirtualProcessImpl2(
     IUnknown ** ppInstance,
     CLR_DEBUGGING_PROCESS_FLAGS* pFlagsOut)
 {
+#ifdef TARGET_WINDOWS
```
am11, line 183: Maybe we can use LoadLibraryExW in Unix as well and delete LoadLibraryW from PAL; since LoadLibraryW calls LoadLibraryExW in PAL, and other few usages don't seem special.

AaronRobinsonMSFT: Yep. That is the plan. This function in the PAL is about to get an overhaul. We needed to port this to .NET 8 so this PR was the narrow fix. The broader .NET 9 fix will be coming soon.

AaronRobinsonMSFT: See https://github.com/dotnet/runtime/pull/91693

PR #91487, opened 2023-09-02
DNMD in Mono

jkoritzinsky:
Rewrite most of Mono's metadata usage to use DNMD. Let's see what CI says.

Expected things to be broken to at least some extent:

 - Hot reload
 - System.Reflection.Emit


lambdageek:
Interesting...

src/mono/mono/metadata/metadata.c
```diff
@@ -7077,23 +7036,24 @@ mono_type_to_unmanaged (MonoType *type, MonoMarshalSpec *mspec, gboolean as_fiel
...
+	if (!md_create_cursor (meta->metadata_handle, mdtid_FieldMarshal, &marshal_info, &count))
+		return NULL;
+	
+	if (!md_find_row_from_cursor (marshal_info, mdtFieldMarshal_Parent, tok, &marshal_info))
+		return NULL;
 
-	return mono_metadata_blob_heap (meta, mono_metadata_decode_row_col (tdef, loc.result, MONO_FIELD_MARSHAL_NATIVE_TYPE));
+	// TODO-DNMD: Hot Reload.
+	// If we go with DNMD's metadata update mechanism, then we don't need to do anything special here.
+	// If we keep Mono's layered mechanism and don't use DNMD's md_apply_delta API, then we'll need to do something here.
```
lambdageek, line 7050: For the most part, mono's layered approach didn't work out:  I was hoping to avoid having to fully materialize the merged metadata tables after an update (ie: make row decoding slower but avoid having to place mutated image data into writable memory).  But eventually I ended up giving up on that so now Mono ends up creating mutated copies of every modified table when applying a delta anyway.

There is one artifact of the original design that's kind of subtle: we never have to stop the world while we're applying the update because we still never write over old table data - we always create copies. So when we use the managed hot reload API, other threads keep running and using the old tables until the update is done.

I think if there was in-place mutation (I assume that's what DNMD does - AFAICT that's what CoreCLR's EnC does), we would have to be careful to stop the world before applying the update (easy, the GC's STW api is available).  And ideally we'd have some way to ensure that no thread is inside the interpreter's transform stage when the world stops - we don't want it to see a mix of old and new metadata.

src/mono/CMakeLists.txt
```diff
@@ -952,6 +952,19 @@ endif()
 
 include_directories("${CLR_SRC_NATIVE_DIR}")
 
+include(FetchContent)
+
+FetchContent_Declare(
+  dnmd
+  GIT_REPOSITORY https://github.com/AaronRobinsonMSFT/DNMD
```
am11, line 959: This will introduce C++ runtime dependency in mono and we are ok with this, neither we can lower the dependency in DNMD to just use the plan C, right?

jkoritzinsky: DNMD core libraries (dnmd and dnmd::pdb) only use C. The dnmd::interfaces library uses C++, but that is intended for use in CoreCLR, not here.

src/mono/mono/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 project(mono)
 
+find_package(dncp REQUIRED)
```
filipnavara, line 3: Is `dncp` required? If yes, should it get git-linked like `dnmd`?

jkoritzinsky: I think I can avoid the dncp dependency. I did some work in dnmd to remove the requirement. I'll try it out locally.

src/mono/CMakeLists.txt
huoyaoyuan: I wonder if this would block or be a pain of offline build.

jkoritzinsky: This PR is just for testing. If we were to actually ship this, we would likely integrate DNMD into the repo or use dependency flow to pass the dependencies.

PR #91489, opened 2023-09-02
[c11] define mono atomics in terms of standard atomics

lambdageek:
Contributes to #90404

### DIFF
--- a/src/mono/mono/utils/atomic.h
+++ b/src/mono/mono/utils/atomic.h
@@ -25,19 +25,7 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
 Apple targets have historically being problematic, xcode 4.6 would miscompile the intrinsic.
 */
 
-#if defined(_MSC_VER)
-
-#define MONO_IGNORE_STDATOMIC 1
-#endif
-
-#if defined(HOST_ANDROID) && (defined(HOST_X86) || defined(HOST_AMD64) || defined(HOST_ARM))
-/* on Android-x86 ATOMIC_LONG_LONG_LOCK_FREE == 1, not 2 like we want. */
-/* on Andriod-x64 ATOMIC_LONG_LOCK_FREE == 1, not 2 */
-/* on Android-armv7 ATOMIC_INT_LOCK_FREE == 1, not 2 */
-#define MONO_IGNORE_STDATOMIC 1
-#endif
-
-#if !defined(__STDC_NO_ATOMICS__) && !defined(MONO_IGNORE_STDATOMIC)
+#ifdef ENABLE_STDATOMIC
 
 #include<stdatomic.h>
 
--- a/src/mono/mono/utils/atomic.h
+++ b/src/mono/mono/utils/atomic.h
@@ -25,6 +25,11 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
 Apple targets have historically being problematic, xcode 4.6 would miscompile the intrinsic.
 */
 
+#if defined(_MSC_VER)
+
+#define MONO_IGNORE_STDATOMIC 1
+#endif
+
 #if defined(HOST_ANDROID) && (defined(HOST_X86) || defined(HOST_AMD64))
 /* on Android-x86 ATOMIC_LONG_LONG_LOCK_FREE == 1, not 2 like we want. */
 /* on Andriod-x64 ATOMIC_LONG_LOCK_FREE == 1, not 2 */
--- a/src/mono/mono/utils/atomic.h
+++ b/src/mono/mono/utils/atomic.h
@@ -25,6 +25,11 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
 Apple targets have historically being problematic, xcode 4.6 would miscompile the intrinsic.
 */
 
+#if defined(HOST_ANDROID) && defined(HOST_X86)
+/* on Android-x86 ATOMIC_LONG_LONG_LOCK_FREE == 1, not 2 like we want. */
+#define MONO_IGNORE_STDATOMIC 1
+#endif
+
 #if !defined(__STDC_NO_ATOMICS__) && !defined(MONO_IGNORE_STDATOMIC)
 
 #include<stdatomic.h>
--- a/src/mono/mono/utils/atomic.h
+++ b/src/mono/mono/utils/atomic.h
@@ -92,7 +92,7 @@ mono_atomic_cas_i64 (volatile gint64 *dest, gint64 exch, gint64 comp)
 	(void)atomic_compare_exchange_strong ((volatile atomic_llong *)dest, (long long*)&comp, exch);
 	return comp;
 #else
-#error gint64 not same size atomic_llong or atomic_long, define MONO_IGNORE_STDATOMIC
+#error gint64 not same size atomic_llong or atomic_long, don't define MONO_USE_STDATOMIC
 #endif
 }
 
@@ -165,7 +165,7 @@ mono_atomic_xchg_i64 (volatile gint64 *dest, gint64 exch)
 	g_static_assert (sizeof (atomic_llong) == sizeof (*dest) && ATOMIC_LLONG_LOCK_FREE == 2);
 	return atomic_exchange ((volatile atomic_llong *)dest, exch);
 #else
-#error gint64 not same size atomic_llong or atomic_long, define MONO_IGNORE_STDATOMIC
+#error gint64 not same size atomic_llong or atomic_long, don't define MONO_USE_STDATOMIC
 #endif
 }
 
@@ -193,7 +193,7 @@ mono_atomic_fetch_add_i64 (volatile gint64 *dest, gint64 add)
 	g_static_assert (sizeof (atomic_llong) == sizeof (*dest) && ATOMIC_LLONG_LOCK_FREE == 2);
 	return atomic_fetch_add ((volatile atomic_llong *)dest, add);
 #else
-#error gint64 not same size atomic_llong or atomic_long, define MONO_IGNORE_STDATOMIC
+#error gint64 not same size atomic_llong or atomic_long, don't define MONO_USE_STDATOMIC
 #endif
 }
 
@@ -227,7 +227,7 @@ mono_atomic_load_i64 (volatile gint64 *src)
 	g_static_assert (sizeof (atomic_llong) == sizeof (*src) && ATOMIC_LLONG_LOCK_FREE == 2);
 	return atomic_load ((volatile atomic_llong *)src);
 #else
-#error gint64 not same size atomic_llong or atomic_long, define MONO_IGNORE_STDATOMIC
+#error gint64 not same size atomic_llong or atomic_long, don't define MONO_USE_STDATOMIC
 #endif
 }
 
@@ -269,7 +269,7 @@ mono_atomic_store_i64 (volatile gint64 *dst, gint64 val)
 	g_static_assert (sizeof (atomic_llong) == sizeof (*dst) && ATOMIC_LLONG_LOCK_FREE == 2);
 	atomic_store ((volatile atomic_llong *)dst, val);
 #else
-#error gint64 not same size atomic_llong or atomic_long, define MONO_IGNORE_STDATOMIC
+#error gint64 not same size atomic_llong or atomic_long, don't define MONO_USE_STDATOMIC
 #endif
 }
 
--- a/src/mono/mono/utils/atomic.h
+++ b/src/mono/mono/utils/atomic.h
@@ -25,7 +25,19 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
 Apple targets have historically being problematic, xcode 4.6 would miscompile the intrinsic.
 */
 
-#ifdef ENABLE_STDATOMIC
+#if defined(_MSC_VER)
+
+#define MONO_IGNORE_STDATOMIC 1
+#endif
+
+#if defined(HOST_ANDROID) && (defined(HOST_X86) || defined(HOST_AMD64) || defined(HOST_ARM))
+/* on Android-x86 ATOMIC_LONG_LONG_LOCK_FREE == 1, not 2 like we want. */
+/* on Andriod-x64 ATOMIC_LONG_LOCK_FREE == 1, not 2 */
+/* on Android-armv7 ATOMIC_INT_LOCK_FREE == 1, not 2 */
+#define MONO_IGNORE_STDATOMIC 1
+#endif
+
+#if !defined(__STDC_NO_ATOMICS__) && !defined(MONO_IGNORE_STDATOMIC)
 
 #include<stdatomic.h>
 
--- a/src/mono/mono/utils/atomic.h
+++ b/src/mono/mono/utils/atomic.h
@@ -25,19 +25,50 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
 Apple targets have historically being problematic, xcode 4.6 would miscompile the intrinsic.
 */
 
+/* Decide if we will use stdatomic.h */
+/*
+ * Generally, we can enable C11 atomics if the header is available and if all the primitive types we
+ * care about (int, long, void*, long long) are lock-free.
+ *
+ * Note that we genrally don't want the compiler's locking implementation because it may take a
+ * global lock, in which case if the atomic is used by both the GC implementation and runtime
+ * internals we may have deadlocks during GC suspend.
+ *
+ * It might be possible to use some Mono specific implementation for specific types (e.g. long long)
+ * on some platforms if the standard atomics for some type are not lock-free (for example: long
+ * long).  We might be able to use a GC-aware lock, for example.
+ *
+ */
 #if defined(_MSC_VER)
-
-#define MONO_IGNORE_STDATOMIC 1
+  /*
+   * we need two things:
+   *
+   * 1. MSVC atomics support is not experimental, or we pass /experimental:c11atomics
+   *
+   * 2. We build our C++ code with C++23 or later (otherwise MSVC will complain about including
+   * stdatomic.h)
+   *
+   */
+#  undef MONO_USE_STDATOMIC
+#elif defined(HOST_IOS) || defined(HOST_OSX) || defined(HOST_WATCHOS) || defined(HOST_TVOS)
+#  define MONO_USE_STDATOMIC 1
+#elif defined(HOST_ANDROID)
+  /* on Android-x86 ATOMIC_LONG_LONG_LOCK_FREE == 1, not 2 like we want. */
+  /* on Andriod-x64 ATOMIC_LONG_LOCK_FREE == 1, not 2 */
+  /* on Android-armv7 ATOMIC_INT_LOCK_FREE == 1, not 2 */
+#  if defined(HOST_ARM64)
+#    define MONO_USE_STDATOMIC 1
+#  endif
+#elif defined(HOST_LINUX)
+  /* FIXME: probably need arch checks */
+#  define MONO_USE_STDATOMIC 1
+#elif defined(HOST_WASI) || defined(HOST_BROWSER)
+#  define MONO_USE_STDATOMIC 1
+#else
+#  undef MONO_USE_STDATOMIC
 #endif
 
-#if defined(HOST_ANDROID) && (defined(HOST_X86) || defined(HOST_AMD64) || defined(HOST_ARM))
-/* on Android-x86 ATOMIC_LONG_LONG_LOCK_FREE == 1, not 2 like we want. */
-/* on Andriod-x64 ATOMIC_LONG_LOCK_FREE == 1, not 2 */
-/* on Android-armv7 ATOMIC_INT_LOCK_FREE == 1, not 2 */
-#define MONO_IGNORE_STDATOMIC 1
-#endif
-
-#if !defined(__STDC_NO_ATOMICS__) && !defined(MONO_IGNORE_STDATOMIC)
+#ifdef MONO_USE_STDATOMIC
 
 #include<stdatomic.h>
 
--- a/src/mono/mono/utils/atomic.h
+++ b/src/mono/mono/utils/atomic.h
@@ -25,8 +25,9 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
 Apple targets have historically being problematic, xcode 4.6 would miscompile the intrinsic.
 */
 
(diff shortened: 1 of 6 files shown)


src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,236 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+static inline gint64
+mono_atomic_fetch_add_i64 (volatile gint64 *dest, gint64 add);
+
+static inline gint32
+mono_atomic_add_i32 (volatile gint32 *dest, gint32 add)
+{
+	// mono_atomic_add_ is supposed to return the value that is stored.
+	// the atomic_add intrinsic returns the previous value instead.
+	// so we return prev+add which should be the new value
+	// FIXME: is this okay?
```
alexrp, line 91: yes

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,236 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+#include<stdatomic.h>
+
+/* we can't have nice things.  InterlockedCompareExchange and __sync_val_compare_and_swap both
+ * return the previous value of dest.  On the other hand C11 atomics return a boolean success/failure value.
+ *
+ * Emulate the value-returning atomic in terms of the bool one.
+ *
+ * 
+ */
+#define mono_atomic_cas_kernel(VALTYPE,dest,exch,comp) do {	\
```
alexrp, line 39: `atomic_compare_exchange` updates `comp` (a pointer!) with the actual value at `dest`, so you don't actually need this hack.

lambdageek: You know... I shouldn't write code late at night. I read the cppreference page a dozen times tonight and still somehow missed that one sentence

lambdageek:
```
FAILED: mono/mini/CMakeFiles/monosgen-objects.dir/llvm-jit.cpp.obj 
  C:\PROGRA~1\MICROS~1\2022\ENTERP~1\VC\Tools\MSVC\1436~1.325\bin\Hostx86\x64\cl.exe  /nologo /TP -DHAVE_CONFIG_H -DLLVM_API_VERSION=1600 -DMY_ZCALLOC -DNO_GLOBALIZATION_SHIM -ID:\a\_work\1\s\artifacts\obj -ID:\a\_work\1\s\src\native -ID:\a\_work\1\s\artifacts\obj\mono\browser.wasm.Release\cross\mono\mini -ID:\a\_work\1\s\artifacts\obj\mono\browser.wasm.Release\cross\mono\mini\..\.. -ID:\a\_work\1\s\src\mono\mono\mini\..\.. -ID:\a\_work\1\s\src\mono\mono\mini\.. -ID:\a\_work\1\s\src\mono\mono\mini\..\eventpipe -ID:\a\_work\1\s\artifacts\obj\mono\browser.wasm.Release\cross\mono\mini\eventpipe -ID:\a\_work\1\s\src\native\public\. -ID:\a\_work\1\s\artifacts\obj\mono\browser.wasm.Release\cross\mono\eglib -ID:\a\_work\1\s\src\mono\mono\eglib -external:ID:\a\_work\1\s\artifacts\obj\mono\browser.wasm.Release\llvm\x64\include -external:W0 -DHOST_WIN32 -D__WIN32__ -DWIN32 -DWIN32_LEAN_AND_MEAN -DNDEBUG -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_DEPRECATE -DWIN32_THREADS -DWINVER=0x0601 -D_WIN32_WINNT=0x0601 -D_WIN32_IE=0x0501 -D_UNICODE -DUNICODE -DFD_SETSIZE=1024 -DNVALGRIND -ID:\a\_work\1\s\artifacts\obj\mono\browser.wasm.Release\llvm\\x64/include  /std:c++17  -fno-exceptions -fno-rtti -D__STDC_CONSTANT_MACROS -D__STD_FORMAT_MACROS -D__STDC_LIMIT_MACROS /O2 /Ob2 /DNDEBUG -MT /source-charset:utf-8 /wd4244 /W4 /WX /EHsc /FC /Oi /GF /Zi /GL -Os /showIncludes /Fomono\mini\CMakeFiles\monosgen-objects.dir\llvm-jit.cpp.obj /Fdmono\mini\CMakeFiles\monosgen-objects.dir\ /FS -c D:\a\_work\1\s\src\mono\mono\mini\llvm-jit.cpp
  cl : Command line warning D9002 : ignoring unknown option '-fno-exceptions'
  cl : Command line warning D9002 : ignoring unknown option '-fno-rtti'
  C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.36.32532\include\__msvc_cxx_stdatomic.hpp(25): warning STL4038: The contents of <stdatomic.h> are available only with C++23 or later.
```

So this is actually pointing out a problem on other platforms, too.  because we have a handful of C++ files (like `llvm-jit.cpp`) that may include `mono/utils/atomic.h` we need a new enough C++ version so that `stdatomic.h` is available.

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,231 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
 Apple targets have historically being problematic, xcode 4.6 would miscompile the intrinsic.
 */
 
-#if defined(HOST_WIN32)
+#if defined(_MSC_VER)
+
+#define MONO_IGNORE_STDATOMIC 1
+#endif
+
+#if defined(HOST_ANDROID) && (defined(HOST_X86) || defined(HOST_AMD64) || defined(HOST_ARM))
```
marek-safar, line 33: I think the opt-in setup would make more sense here

lambdageek: I'm not so sure. I would expect the number of platforms that don't support stdatomic to decrease over time.

I can try it both ways.  (And I want to move this decision to cmake to make it more visible)

lambdageek: Off by default seems ok

lambdageek:
Oh right, moving the logic for `ENABLE_STDATOMIC` to cmake isn't great because we include the atomic.h header is consumed by offset-tool.py

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
 Apple targets have historically being problematic, xcode 4.6 would miscompile the intrinsic.
 */
 
-#if defined(HOST_WIN32)
+/* Decide if we will use stdatomic.h */
+/*
+ * Generally, we can enable C11 atomics if the header is available and if all the primitive types we
+ * care about (int, long, void*, long long) are lock-free.
+ *
+ * Note that we genrally don't want the compiler's locking implementation because it may take a
```
AaronRobinsonMSFT, line 33:
```suggestion
 * Note that we generally don't want the compiler's locking implementation because it may take a
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+#else
+#error gint64 not same size atomic_llong or atomic_long, define MONO_IGNORE_STDATOMIC
+#endif
+}
+
+static inline gpointer
+mono_atomic_xchg_ptr (volatile gpointer *dest, gpointer exch)
+{
+	g_static_assert (ATOMIC_POINTER_LOCK_FREE == 2);
+	return atomic_exchange ((_Atomic gpointer*)dest, exch);
```
AaronRobinsonMSFT, line 176:
```suggestion
	return atomic_exchange ((_Atomic(gpointer)*)dest, exch);
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+#else
+#error gint64 not same size atomic_llong or atomic_long, define MONO_IGNORE_STDATOMIC
+#endif
+}
+
+static inline gpointer
+mono_atomic_load_ptr (volatile gpointer *src)
+{
+	g_static_assert (ATOMIC_POINTER_LOCK_FREE == 2);
+	return atomic_load ((_Atomic gpointer*)src);
```
AaronRobinsonMSFT, line 238:
```suggestion
	return atomic_load ((_Atomic(gpointer)*)src);
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+#else
+#error gint64 not same size atomic_llong or atomic_long, define MONO_IGNORE_STDATOMIC
+#endif
+}
+
+static inline void
+mono_atomic_store_ptr (volatile gpointer *dst, gpointer val)
+{
+	g_static_assert (ATOMIC_POINTER_LOCK_FREE == 2);
+	atomic_store ((_Atomic gpointer*)dst, val);
```
AaronRobinsonMSFT, line 280:
```suggestion
	atomic_store ((_Atomic(gpointer)*)dst, val);
```

AaronRobinsonMSFT (approved):
Some nits. I changed to using the `_Atomic()` pattern since it is the preferred style. Otherwise LGTM.

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+#  define MONO_USE_STDATOMIC 1
+#elif defined(HOST_WASI) || defined(HOST_BROWSER)
+#  define MONO_USE_STDATOMIC 1
+#else
+#  undef MONO_USE_STDATOMIC
+#endif
+
+#ifdef MONO_USE_STDATOMIC
+
+#include<stdatomic.h>
```
lateralusX, line 73:
```suggestion
#include <stdatomic.h>
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+#if SIZEOF_LONG == 8
+	g_static_assert (sizeof (atomic_long) == sizeof (*dest) && ATOMIC_LONG_LOCK_FREE == 2);
+	(void)atomic_compare_exchange_strong ((atomic_long*)dest, (long*)&comp, exch);
+	return comp;
+#elif SIZEOF_LONG_LONG == 8
+	g_static_assert (sizeof (atomic_llong) == sizeof (*dest) && ATOMIC_LLONG_LOCK_FREE == 2);
+	(void)atomic_compare_exchange_strong ((atomic_llong*)dest, (long long*)&comp, exch);
+	return comp;
+#else
+#error gint64 not same size atomic_llong or atomic_long, define MONO_IGNORE_STDATOMIC
```
lateralusX, line 95: should it say undefine MONO_USE_STDATOMIC?

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+mono_atomic_xchg_i64 (volatile gint64 *dest, gint64 exch)
+{
+#if SIZEOF_LONG == 8
+	g_static_assert (sizeof (atomic_long) == sizeof (*dest) && ATOMIC_LONG_LOCK_FREE == 2);
+	return atomic_exchange ((atomic_long*)dest, exch);
+#elif SIZEOF_LONG_LONG == 8
+	g_static_assert (sizeof (atomic_llong) == sizeof (*dest) && ATOMIC_LLONG_LOCK_FREE == 2);
+	return atomic_exchange ((atomic_llong*)dest, exch);
+#else
+#error gint64 not same size atomic_llong or atomic_long, define MONO_IGNORE_STDATOMIC
```
lateralusX, line 168: should it say undefine MONO_USE_STDATOMIC?

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+#endif
+
+#ifdef MONO_USE_STDATOMIC
+
+#include<stdatomic.h>
+
+static inline gint32
+mono_atomic_cas_i32 (volatile gint32 *dest, gint32 exch, gint32 comp)
+{
+	g_static_assert (sizeof(atomic_int) == sizeof(*dest) && ATOMIC_INT_LOCK_FREE == 2);
```
lateralusX, line 78:
```suggestion
	g_static_assert (sizeof (atomic_int) == sizeof (*dest) && ATOMIC_INT_LOCK_FREE == 2);
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+static inline gint64
+mono_atomic_dec_i64 (volatile gint64 *dest)
+{
+	return mono_atomic_add_i64 (dest, -1);
+}
+
+static inline gint32
+mono_atomic_xchg_i32 (volatile gint32 *dest, gint32 exch)
+{
+	g_static_assert (sizeof(atomic_int) == sizeof(*dest) && ATOMIC_INT_LOCK_FREE == 2);
```
lateralusX, line 154:
```suggestion
	g_static_assert (sizeof (atomic_int) == sizeof (*dest) && ATOMIC_INT_LOCK_FREE == 2);
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+mono_atomic_xchg_ptr (volatile gpointer *dest, gpointer exch)
+{
+	g_static_assert (ATOMIC_POINTER_LOCK_FREE == 2);
+	return atomic_exchange ((_Atomic(gpointer)*)dest, exch);
+}
+
+static inline gint32
+mono_atomic_fetch_add_i32 (volatile gint32 *dest, gint32 add)
+{
+	g_static_assert (sizeof(atomic_int) == sizeof(*dest) && ATOMIC_INT_LOCK_FREE == 2);
```
lateralusX, line 182:
```suggestion
	g_static_assert (sizeof (atomic_int) == sizeof (*dest) && ATOMIC_INT_LOCK_FREE == 2);
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+	return atomic_fetch_add ((atomic_llong*)dest, add);
+#else
+#error gint64 not same size atomic_llong or atomic_long, define MONO_IGNORE_STDATOMIC
+#endif
+}
+
+static inline gint8
+mono_atomic_load_i8 (volatile gint8 *src)
+{
+	g_static_assert (sizeof(atomic_char) == sizeof(*src) && ATOMIC_CHAR_LOCK_FREE == 2);
```
lateralusX, line 203:
```suggestion
	g_static_assert (sizeof (atomic_char) == sizeof (*src) && ATOMIC_CHAR_LOCK_FREE == 2);
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+	return atomic_load ((atomic_char *)src);
+}
+
+static inline gint16
+mono_atomic_load_i16 (volatile gint16 *src)
+{
+	g_static_assert (sizeof(atomic_short) == sizeof(*src) && ATOMIC_SHORT_LOCK_FREE == 2);
```
lateralusX, line 210:
```suggestion
	g_static_assert (sizeof (atomic_short) == sizeof (*src) && ATOMIC_SHORT_LOCK_FREE == 2);
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+	return atomic_load ((atomic_short*)src);
+}
+
+static inline gint32 mono_atomic_load_i32 (volatile gint32 *src)
+{
+	g_static_assert (sizeof(atomic_int) == sizeof(*src) && ATOMIC_INT_LOCK_FREE == 2);
```
lateralusX, line 216:
```suggestion
	g_static_assert (sizeof (atomic_int) == sizeof (*src) && ATOMIC_INT_LOCK_FREE == 2);
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+mono_atomic_load_ptr (volatile gpointer *src)
+{
+	g_static_assert (ATOMIC_POINTER_LOCK_FREE == 2);
+	return atomic_load ((_Atomic(gpointer)*)src);
+}
+
+static inline void
+mono_atomic_store_i8 (volatile gint8 *dst, gint8 val)
+{
+	g_static_assert (sizeof(atomic_char) == sizeof(*dst) && ATOMIC_CHAR_LOCK_FREE == 2);
```
lateralusX, line 244:
```suggestion
	g_static_assert (sizeof (atomic_char) == sizeof (*dst) && ATOMIC_CHAR_LOCK_FREE == 2);
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+	atomic_store ((atomic_char*)dst, val);
+}
+
+static inline void
+mono_atomic_store_i16 (volatile gint16 *dst, gint16 val)
+{
+	g_static_assert (sizeof(atomic_short) == sizeof(*dst) && ATOMIC_SHORT_LOCK_FREE == 2);
```
lateralusX, line 251:
```suggestion
	g_static_assert (sizeof (atomic_short) == sizeof (*dst) && ATOMIC_SHORT_LOCK_FREE == 2);
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+	atomic_store ((atomic_short*)dst, val);
+}
+
+static inline void
+mono_atomic_store_i32 (volatile gint32 *dst, gint32 val)
+{
+	g_static_assert (sizeof(atomic_int) == sizeof(*dst) && ATOMIC_INT_LOCK_FREE == 2);
```
lateralusX, line 258:
```suggestion
	g_static_assert (sizeof (atomic_int) == sizeof (*dst) && ATOMIC_INT_LOCK_FREE == 2);
```

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+
+#ifdef MONO_USE_STDATOMIC
+
+#include<stdatomic.h>
+
+static inline gint32
+mono_atomic_cas_i32 (volatile gint32 *dest, gint32 exch, gint32 comp)
+{
+	g_static_assert (sizeof(atomic_int) == sizeof(*dest) && ATOMIC_INT_LOCK_FREE == 2);
+	(void)atomic_compare_exchange_strong ((atomic_int*)dest, &comp, exch);
```
lateralusX, line 79: Do we have strong guarantees for mono_atomic_cas_* functions? I believe we normally call these in loops to handle potential spuriously fails on platforms that could trigger them. If we would like to make the default implementation using strong guarantees, then maybe we should also offer the option to use weak when we already do cas in loops to potentially be more performant on platforms that takes advantage of spuriously fails.

src/mono/mono/utils/atomic.h
```diff
@@ -25,7 +25,262 @@ F/MonoDroid( 1568): shared runtime initialization error: Cannot load library: re
...
+	return comp;
+}
+
+static inline gint64
+mono_atomic_cas_i64 (volatile gint64 *dest, gint64 exch, gint64 comp)
+{
+#if SIZEOF_LONG == 8
+	g_static_assert (sizeof (atomic_long) == sizeof (*dest) && ATOMIC_LONG_LOCK_FREE == 2);
+	(void)atomic_compare_exchange_strong ((atomic_long*)dest, (long*)&comp, exch);
```
lateralusX, line 88: when we do an explicit cast like this, wouldn't it make sense to cast to the full type, like `volatile atomic_long *` in this case. This will be done implicit, but since we do the cast it might make sense to make it explicit. We do similar on callsites to `mono_atomic_cas_*` in runtime where we have both implicit and explicit.

Another nit that we are inconsistent with in the runtime, when we cast to pointers do we have a preference if we should write (type*) or (type *)?

src/mono/mono/utils/atomic.h
lambdageek: yea good point. I need to look at the callers and check if they're all in loops.  My feeling is that we should make weak the default and add strong as a second option.  I didn't want to change the mono functions as part of this PR (and strong seemed like a safer default).  But I can audit and report back and then we can add the non-default versions

lambdageek: Found 173 occurrences of `mono_atomic_cas` in src/mono/mono.  Looked through about 2/3rds of them, chasing down callers if it's in a static helper method or a macro.   About half are weak uses and about half are strong.  So `strong` is definitely the right default.  In a few places we could switch over to weak (particularly inside sgen there are some loops that could use weak).  But it should be follow-up work, IMO.

src/mono/mono/utils/atomic.h
lambdageek: done.

I prefer `(type *)` (and in declarations put the `*` next to the variable, not the type name in C code, and the other way in C++). Although I'm not sure we have that written down in the style guide.

lateralusX: Jupp, me too and I see that you have updated to follow that style in this PR now.

src/mono/mono/utils/atomic.h
lateralusX: That is what I expected, we have made different assumption around the strong/weak guarantees around the runtime over years. Moving over to strong will potential correct some false assumptions previously made in runtime (at least it won't break anything), so it should be safe, and I agree that we could identify and use weak versions where it might benefit.

lambdageek: Created https://github.com/dotnet/runtime/issues/91747 for follow-up work

PR #91507, opened 2023-09-02
Implement Linq CountBy method for IEnumerable and IQueryable

### LINKED ISSUE #77716: [API Proposal]: Linq CountBy method
[Draft implementation: alternative design 2](https://github.com/dotnet/runtime/commit/793c97a294fb4a6093694988bbebffceba1729d0)

### Background and motivation

Currently, it is not so easy to count occurrences of `IEnumerable<T>` elements with existing LINQ methods in C#. 

```csharp
var arr = new[]{1, 2, 2, 3, 3, 3};

Dictionary<int, int> counts = arr
  .GroupBy(x => x)
  .ToDictionary(x => x.Key, x => x.Count());
```

### Motivation
 - It is not efficient to use `GroupBy` for counting because it materializes `Grouping` objects with a list of values related to each key.
 - It is not intuitive that `.Count()` will not enumerate the sequence again.
 - Other popular languages provide simpler API for this

### Other language examples
F#:
```fsharp
let arr = [|1; 2; 2; 3; 3; 3|]
let counts: seq<int * int> = arr |> Seq.countBy id
```

Python:
```py

from collections import Counter

arr = [1, 2, 2, 3, 3, 3]
counts: Counter = Counter(arr)
```

Kotlin:
```kt
val arr = arrayOf(1, 2, 2, 3, 3, 3)
val counts: Map<Int, Int> = arr.groupingBy { it }.eachCount()
```

### Benchmark
 [Benchmark](https://gist.github.com/epeshk/4afa1ecab272c76e25e2b0e6f29da232) results:
```
BenchmarkDotNet=v0.13.2, OS=Windows 10 (10.0.19044.2130/21H2/November2021Update)
AMD Ryzen 9 5950X, 1 CPU, 32 logical and 16 physical cores
.NET SDK=7.0.100-rc.2.22477.23
  [Host]     : .NET 7.0.0 (7.0.22.47203), X64 RyuJIT AVX2
  DefaultJob : .NET 7.0.0 (7.0.22.47203), X64 RyuJIT AVX2
```

|       Method |       Mean |    Error |   StdDev |    Gen0 |    Gen1 |  Allocated |
|------------- |-----------:|---------:|---------:|--------:|--------:|-----------:|
|     Grouping | 1,031.7 us | 18.61 us | 17.41 us | 60.5469 | 39.0625 | 1017.17 KB |
|   LookupDict | 18.919 ms | 0.3766 ms | 0.6790 ms | 1343.7500 | 1312.5000 | 937.5000 |  10.22 MB |
|       Lookup | 16.927 ms | 0.3320 ms | 0.4078 ms |  625.0000 |  593.7500 | 218.7500 |   7.44 MB |
|     Counting |   673.5 us |  5.99 us |  5.31 us |       - |       - |    9.52 KB |
| CountingCast |   635.3 us |  2.42 us |  2.26 us |       - |       - |    7.25 KB |

Benchmark for case without repetitions:
`Seq = Enumerable.Range(0, SequenceLength).ToArray();`

|       Method |      Mean |     Error |    StdDev |      Gen0 |      Gen1 |     Gen2 | Allocated |
|------------- |----------:|----------:|----------:|----------:|----------:|---------:|----------:|
|     Grouping | 18.902 ms | 0.3639 ms | 0.4469 ms | 1343.7500 | 1312.5000 | 937.5000 |  10.22 MB |
|   LookupDict | 22.138 ms | 0.4110 ms | 0.4037 ms | 1468.7500 | 1437.5000 | 937.5000 |  16.15 MB |
|       Lookup | 14.548 ms | 0.0455 ms | 0.0380 ms |  968.7500 |  953.1250 | 468.7500 |  10.39 MB |
|     Counting |  3.844 ms | 0.0307 ms | 0.0287 ms | 1007.8125 |  992.1875 | 992.1875 |   4.21 MB |
| CountingCast |  2.211 ms | 0.0271 ms | 0.0240 ms |  652.3438 |  636.7188 | 636.7188 |   2.77 MB |

### API Proposal

```csharp
namespace System.Linq;

public static partial class Enumerable
{
    public static IEnumerable<KeyValuePair<TKey, int>> CountBy<TSource, TKey>(
        this IEnumerable<TSource> source,
        Func<TSource, TKey> keySelector,
        IEqualityComparer<TKey>? keyComparer = null) where TKey : notnull;


    public static IEnumerable<KeyValuePair<TKey, long>> LongCountBy<TSource, TKey>(
        this IEnumerable<TSource> source,
        Func<TSource, TKey> keySelector,
        IEqualityComparer<TKey>? keyComparer = null) where TKey : notnull;
}
```

### API Usage

```csharp
var array = new []{1, 2, 2, 3, 3, 3};

var counts = array.CountBy(x => x);

foreach (var (key, count) in counts)
    Console.WriteLine($"{key}: {count}");
```

#### Find any most frequent element:
```csharp
enumerable
    .CountBy(x => x)
    .MaxBy(x => x.Value)
    .Key;
```

#### Find duplicates:
```csharp
enumerable
    .CountBy(x => x)
    .Where(x => x.Value > 1)
    .Select(x => x.Key);
```

#### 1. With separated result type, inspired by `ILookup`
```csharp
namespace System.Linq;
public interface ICounts<TKey, TCount> : IEnumerable<(TKey Key, TCount Count)> where TKey : notnull
{
    int KeyCount { get; }

    TCount this[TKey key] { get; }

    bool Contains(TKey key);
}

public static partial class Enumerable
{
    public static ICounts<TKey, int> CountBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector) where TKey : notnull;
    public static ICounts<TKey, int> CountBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? keyComparer) where TKey : notnull;
    public static ICounts<TKey, long> LongCountBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector) where TKey : notnull;
    public static ICounts<TKey, long> LongCountBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? keyComparer) where TKey : notnull;
}
```

#### 2. Same as previous, but without `TKey : notnull` constraint
```csharp
namespace System.Linq;
public interface ICounts<TKey, TCount> : IEnumerable<(TKey Key, TCount Count)>
{
    int KeyCount { get; }

    TCount this[TKey key] { get; }

    bool Contains(TKey key);
}

public static partial class Enumerable
{
    public static ICounts<TKey, int> CountBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector);
    public static ICounts<TKey, int> CountBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? keyComparer);
    public static ICounts<TKey, long> LongCountBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector);
    public static ICounts<TKey, long> LongCountBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? keyComparer);
}
```

##### Find duplicates:
```csharp
ICounts<TKey, int> counts = enumerable.CountBy(x => x);
IEnumerable<TKey> duplicates = counts
    .Where(x => x.Count > 1)
    .Select(x => x.Key);
```

#### 3.
The extension method for `IEnumerable<IGrouping<TKey, TValue>>` with internal knowledge about `GroupingEnumerator` allowing to Count values without materializing lists in memory.
```csharp
array.GroupBy(x => x).Counts();
```

#### Questions to decide:
- Should we use `IEnumerable<KeyValuePair<TKey, TValue>>` as a return type or specific type instead for better naming, like `IEnumerable<(TKey Key, TCount Count)>`:
##### Find duplicates:
```csharp
IEnumerable<KeyCountPair<TKey, int>> counts = enumerable.CountBy(x => x);
IEnumerable<TKey> duplicates = counts
    .Where(x => x.Count > 1)
    .Select(x => x.Key);
```

- Should we return `IEnumerable<>` for simplicity or go with `IDictioinary<>/IReadOnlyDictionary<>/ICounts<>` to avoid copying? Specific return type could improve performance for ~x1.5 for cases when user needs a lookup.
F# returns sequence, Kotlin — map, Python — special map-like type

- Should `TKey` allow `null`? For example, `ToLookup` allow `null` and `Dictionary` not.

### Risks

Looks like there is no C# method with the same name in `dotnet` organization projects.
https://github.com/search?q=org%3Adotnet+CountBy&type=code

manandre:
Closes #77716

@epeshk I started from your initial commit https://github.com/dotnet/runtime/commit/793c97a294fb4a6093694988bbebffceba1729d0 and I made required changes to respect [approved API](https://github.com/dotnet/runtime/issues/77716#issuecomment-1664448577)
Your comments are welcome indeed.

### DIFF
--- a/src/libraries/System.Linq/src/System/Linq/CountBy.cs
+++ b/src/libraries/System.Linq/src/System/Linq/CountBy.cs
@@ -8,7 +8,7 @@ namespace System.Linq
 {
     public static partial class Enumerable
     {
-        public static IEnumerable<KeyValuePair<TKey, int>> CountBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? comparer = null) where TKey : notnull
+        public static IEnumerable<KeyValuePair<TKey, int>> CountBy<TSource, TKey>(this IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? keyComparer = null) where TKey : notnull
         {
             if (source is null)
             {
@@ -19,32 +19,39 @@ namespace System.Linq
                 ThrowHelper.ThrowArgumentNullException(ExceptionArgument.keySelector);
             }
 
-            return Core(source, keySelector, comparer);
+            return Core(source, keySelector, keyComparer);
 
-            static IEnumerable<KeyValuePair<TKey, int>> Core(IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? comparer)
+            static IEnumerable<KeyValuePair<TKey, int>> Core(IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? keyComparer)
             {
-                Dictionary<TKey, int> countsBy = BuildCountDictionary(source, keySelector, comparer);
+                using IEnumerator<TSource> enumerator = source.GetEnumerator();
 
-                foreach (KeyValuePair<TKey, int> countBy in countsBy)
+                if (!enumerator.MoveNext())
+                {
+                    yield break;
+                }
+
+                foreach (KeyValuePair<TKey, int> countBy in BuildCountDictionary(enumerator, keySelector, keyComparer))
                 {
                     yield return countBy;
                 }
             }
 
-            static Dictionary<TKey, int> BuildCountDictionary(IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? comparer)
+            static Dictionary<TKey, int> BuildCountDictionary(IEnumerator<TSource> enumerator, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? keyComparer)
             {
-                Dictionary<TKey, int> countsBy = new(comparer);
+                Dictionary<TKey, int> countsBy = new(keyComparer);
 
-                foreach (TSource element in source)
+                do
                 {
-                    TKey currentKey = keySelector(element);
+                    TSource value = enumerator.Current;
+                    TKey key = keySelector(value);
 
-                    ref int currentCount = ref CollectionsMarshal.GetValueRefOrAddDefault(countsBy, currentKey, out _);
+                    ref int currentCount = ref CollectionsMarshal.GetValueRefOrAddDefault(countsBy, key, out _);
                     checked
                     {
                         currentCount++;
                     }
                 }
+                while (enumerator.MoveNext());
 
                 return countsBy;
             }
--- a/src/libraries/System.Linq/src/System/Linq/CountBy.cs
+++ b/src/libraries/System.Linq/src/System/Linq/CountBy.cs
@@ -19,42 +19,42 @@ namespace System.Linq
                 ThrowHelper.ThrowArgumentNullException(ExceptionArgument.keySelector);
             }
 
-            return Core(source, keySelector, keyComparer);
+            return CountByIterator(source, keySelector, keyComparer);
+        }
 
-            static IEnumerable<KeyValuePair<TKey, int>> Core(IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? keyComparer)
+        private static IEnumerable<KeyValuePair<TKey, int>> CountByIterator<TSource, TKey>(IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? keyComparer) where TKey : notnull
+        {
+            using IEnumerator<TSource> enumerator = source.GetEnumerator();
+
+            if (!enumerator.MoveNext())
             {
-                using IEnumerator<TSource> enumerator = source.GetEnumerator();
-
-                if (!enumerator.MoveNext())
-                {
-                    yield break;
-                }
-
-                foreach (KeyValuePair<TKey, int> countBy in BuildCountDictionary(enumerator, keySelector, keyComparer))
-                {
-                    yield return countBy;
-                }
+                yield break;
             }
 
-            static Dictionary<TKey, int> BuildCountDictionary(IEnumerator<TSource> enumerator, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? keyComparer)
+            foreach (KeyValuePair<TKey, int> countBy in BuildCountDictionary(enumerator, keySelector, keyComparer))
             {
-                Dictionary<TKey, int> countsBy = new(keyComparer);
-
-                do
-                {
-                    TSource value = enumerator.Current;
-                    TKey key = keySelector(value);
-
-                    ref int currentCount = ref CollectionsMarshal.GetValueRefOrAddDefault(countsBy, key, out _);
-                    checked
-                    {
-                        currentCount++;
-                    }
-                }
-                while (enumerator.MoveNext());
-
-                return countsBy;
+                yield return countBy;
             }
+        }
+
+        private static Dictionary<TKey, int> BuildCountDictionary<TSource, TKey>(IEnumerator<TSource> enumerator, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? keyComparer) where TKey : notnull
+        {
+            Dictionary<TKey, int> countsBy = new(keyComparer);
+
+            do
+            {
+                TSource value = enumerator.Current;
+                TKey key = keySelector(value);
+
+                ref int currentCount = ref CollectionsMarshal.GetValueRefOrAddDefault(countsBy, key, out _);
+                checked
+                {
+                    currentCount++;
+                }
+            }
+            while (enumerator.MoveNext());
+
+            return countsBy;
         }
     }
 }
--- a/src/libraries/System.Linq/src/System/Linq/CountBy.cs
+++ b/src/libraries/System.Linq/src/System/Linq/CountBy.cs
@@ -19,22 +19,35 @@ namespace System.Linq
                 ThrowHelper.ThrowArgumentNullException(ExceptionArgument.keySelector);
             }
 
-            Dictionary<TKey, int> countsBy = new(comparer);
+            return Core(source, keySelector, comparer);
 
-            using IEnumerator<TSource> e = source.GetEnumerator();
-
-            while (e.MoveNext())
+            static IEnumerable<KeyValuePair<TKey, int>> Core(IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? comparer)
             {
-                TKey currentKey = keySelector(e.Current);
+                Dictionary<TKey, int> countsBy = BuildCountDictionary(source, keySelector, comparer);
 
-                ref int currentCount = ref CollectionsMarshal.GetValueRefOrAddDefault(countsBy, currentKey, out _);
-                checked
+                foreach (KeyValuePair<TKey, int> countBy in countsBy)
                 {
-                    currentCount++;
+                    yield return countBy;
                 }
             }
 
-            return countsBy;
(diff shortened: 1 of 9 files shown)


src/libraries/System.Linq/src/System/Linq/CountBy.cs
```diff
@@ -0,0 +1,40 @@
...
+            if (keySelector is null)
+            {
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.keySelector);
+            }
+
+            Dictionary<TKey, int> countsBy = new(comparer);
+
+            using IEnumerator<TSource> e = source.GetEnumerator();
+
+            while (e.MoveNext())
```
huoyaoyuan, line 26: Why not using a normal `foreach` here?

src/libraries/System.Linq/src/System/Linq/CountBy.cs
```diff
@@ -0,0 +1,40 @@
...
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.source);
+            }
+            if (keySelector is null)
+            {
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.keySelector);
+            }
+
+            Dictionary<TKey, int> countsBy = new(comparer);
+
+            using IEnumerator<TSource> e = source.GetEnumerator();
```
eiriktsarpalis, line 24: This method is eagerly evaluating the source enumerable, returning a materialized dictionary with frequency counts. This violates the invariant of most LINQ methods returning IEnumerable where deferred evaluation is being used (this is how the related `GroupBy` works).

I wouldn't be entirely opposed to this being an eager computation, however in that case we should change the return type to be a concrete `Dictionary<TSource, int>`.

@stephentoub thoughts?

src/libraries/System.Linq/src/System/Linq/CountBy.cs
manandre: you're right, no special reason here. I have changed it for a `foreach` loop.

src/libraries/System.Linq/src/System/Linq/CountBy.cs
manandre: I have added a commit to propose a lazy loading approach. Does it fit the LINQ requirements?

stephentoub: > @stephentoub thoughts?

In general, I think we've kept the prompt vs lazy aspect consistent between the Xx and XxBy methods, e.g. Max and MaxBy are both prompt, Order and OrderBy are both lazy, etc. So given that `Count` is prompt, I think I'd expect `CountBy` to be prompt as well, and then as you suggest _not_ return `IEnumerable<>` but instead return an `IReadOnlyDictionary<TSource, int>` (or maybe a `long` instead of `int` if we want to avoid a future where we need `LongCountBy`).  I suggest `IReadOnlyDictionary<>` purely so that we have some flexibility, e.g. returning a singleton if the input was empty.

We should run it through API review again to discuss.

eiriktsarpalis: We re-reviewed this and ended up favouring the delayed approach.

src/libraries/System.Linq/src/System/Linq/CountBy.cs
```diff
@@ -0,0 +1,60 @@
...
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.source);
+            }
+            if (keySelector is null)
+            {
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.keySelector);
+            }
+
+            return Core(source, keySelector, keyComparer);
+
+            static IEnumerable<KeyValuePair<TKey, int>> Core(IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? keyComparer)
```
eiriktsarpalis, line 24: Following naming convention elsewhere this should be called `CountByIterator`, see for example:

https://github.com/dotnet/runtime/blob/ba10d55c99968ab577a9d96f19fbb2dbddbc83a4/src/libraries/System.Linq/src/System/Linq/Zip.cs#L87-L111

I would also recommend not making it a local method so that any exception stacktraces show up cleaner.

src/libraries/System.Linq/tests/CountByTests.cs
```diff
@@ -0,0 +1,116 @@
...
+                source: Enumerable.Repeat(5, 20),
+                keySelector: x => x,
+                comparer: null,
+                expected: Enumerable.Repeat(5, 1).ToDictionary(x => x, x => 20));
+
+            yield return WrapArgs(
+                source: new string[] { "Bob", "bob", "tim", "Bob", "Tim" },
+                keySelector: x => x,
+                null,
+                expected: new string[] { "Bob", "bob", "tim", "Tim" }.ToDictionary(x => x, x => x == "Bob" ? 2 : 1));
```
eiriktsarpalis, line 80: Nit: consider using a fully evaluated result here, there current expression seems a bit complicated.

src/libraries/System.Linq/tests/CountByTests.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            Assert.Equal(expected, source.CountBy(keySelector, comparer));
+        }
+
+        [Theory]
+        [MemberData(nameof(CountBy_TestData))]
+        public static void CountBy_RunOnce_HasExpectedOutput<TSource, TKey>(IEnumerable<TSource> source, Func<TSource, TKey> keySelector, IEqualityComparer<TKey>? comparer, IEnumerable<KeyValuePair<TKey, int>> expected)
+        {
+            Assert.Equal(expected, source.RunOnce().CountBy(keySelector, comparer));
+        }
```
eiriktsarpalis, line 42: Could you also add a couple of tests validating the deferred semantics of the method?

src/libraries/System.Linq/src/System/Linq/CountBy.cs
manandre: Method renamed and local methods removed.

src/libraries/System.Linq/tests/CountByTests.cs
manandre: Complex results are now more explicit.

src/libraries/System.Linq/tests/CountByTests.cs
manandre: Yes, tell me if the new test methods fit the need.

src/libraries/System.Linq/tests/CountByTests.cs
```diff
@@ -27,6 +27,36 @@ public void CountBy_KeySelectorNull_ThrowsArgumentNullException()
             AssertExtensions.Throws<ArgumentNullException>("keySelector", () => source.CountBy(keySelector, new AnagramEqualityComparer()));
         }
 
+        [Fact]
+        public void CountBy_SourceThrowsOnGetEnumerator()
+        {
+            IEnumerable<int> source = new ThrowsOnGetEnumerator();
+
+            var enumerator = source.CountBy(x => x).GetEnumerator();
```
eiriktsarpalis, line 35: Is this test correct? Isn't this meant to throw when calling `GetEnumerator()`?

manandre: Yes, the test is correct. The input/source `GetEnumerator()` method is (lazily) called on the `MoveNext()` call on the output enumerator.
The `GetEnumerator()` method of the output enumerator only returns the compiler-generated "YieldIterator" instance which is the same as the returned enumerable (as it implements both IEnumerator* and IEnumerable*)

eiriktsarpalis: Ah, right. Thanks.

eiriktsarpalis (approved):
Thanks! @manandre would you be interested in contributing the `AggregateBy` implementation as well? It should be pretty similar to this one.

Issue #90081, opened 2023-08-06
[API Proposal]: `UnsafeAccessorTypeAttribute` for static or private type access


AaronRobinsonMSFT:
### Background and motivation

The [`UnsafeAccessorAttribute`](https://github.com/dotnet/runtime/issues/81741) mechanism was designed to provide access to a non-visible static or instance member (that is, method or field). There are however limitations with this design when involving `static` types or accessing members on non-visible types. This attribute helps bridge that gap.

Consider the following two scenarios involving methods. Note that fields suffer from the same issue.

## Scenario 1 - Private type

```csharp
// Assembly A
private class C
{
    private static int Method(int a) { ... }
}
```

```csharp
// Assembly B
[UnsafeAccessor(UnsafeAccessorKind.StaticMethod, Name="Method")]
static extern int CallMethod(??? c, int a); // One cannot write type C due to visibility.
```

## Scenario 2 - Static type

```csharp
// Assembly A
public static class C
{
    private static int Method(int a) { ... }
}
```

```csharp
// Assembly B
[UnsafeAccessor(UnsafeAccessorKind.StaticMethod, Name="Method")]
static extern int CallMethod(??? c, int a); // A static class cannot be used as a parameter in a signature.
```

## Scenario 3 - Private class parameters

```csharp
// Assembly A
public class C
{
    private class D { }
    private static int Method(D d) { ... }
}
```

```csharp
// Assembly B
[UnsafeAccessor(UnsafeAccessorKind.StaticMethod, Name="Method")]
static extern int CallMethod(??? d); // Unable express D type as a parameter.
```

### API Proposal

The following attribute would accept a [fully qualified](https://learn.microsoft.com/dotnet/framework/reflection-and-codedom/specifying-fully-qualified-type-names) or partially qualified type name to use for member look-up. The string supplied via the `typeName` would be in a format that is accepted by [`Type.GetType(string typeName)`](https://learn.microsoft.com/dotnet/api/system.type.gettype#system-type-gettype(system-string)). This API will be specifcally limited by what that API is capable of. There may be additional restrictions for the initial implementation of this. For example, it only works on [reference types](https://learn.microsoft.com/dotnet/csharp/language-reference/keywords/reference-types) for now.

```csharp
namespace System.Runtime.CompilerServices;

[AttributeUsage(AttributeTargets.Parameter | AttributeTargets.ReturnValue, AllowMultiple = false, Inherited = false)]
[...]

PR #91520, opened 2023-09-03
[iOS] Fix trimming warnings in HttpClientHandler.AnyMobile

simonrozsival:
While testing NativeAOT on iOS, I started noticed this warning from ILC:
```
ILC : Trim analysis warning IL2075: System.Net.Http.HttpClientHandler.InvokeNativeHandlerMethod(String,Object[]): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'System.Type.GetMethod(String)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. 
```

This PR fixes the warning.


MichalStrehovsky:
We typically write targeted trimming tests if there is a suppression. Suppressions are not great because someone could add a new callsite and forget to add DynamicDependency (or DynamicDependency the wrong thing).

I think this can be rewritten to be correct by construction. All the text strings are there, they're just too far apart for analysis to figure them out.

Basically, I'd do this:

* Delete all the DynamicDependency stuff.
* Replace `_nativeHandler!.GetType()!.GetMethod(name);` in `InvokeNativeHandlerMethod` with `Type.GetType(NativeHandlerType + ", " + AssemblyName)!.GetMethod(name);`. The C# compiler will concatenate these strings because they're all consts.
* This is already enough to make everything work. Trimming will keep all public methods on the type and there will be no warning. This is because the name of the type is visible within the `InvokeNativeHandlerMethod` method.
* If there is a size concern (we don't want to keep all the public methods on this type), we could hoist the `Type.GetType(NativeHandlerType + ", " + AssemblyName)!.GetMethod(name);` part into a delegate (add an `Action<MethodInfo>` delegate parameter to `InvokeNativeHandlerMethod` and provide the right implementation at each callsite as a lambda). This should result in comparable size to what was achieved with DynamicDependency (the attributes are not cheap, size-wise - it's a lot of characters and none of them can be shared even though they're identical because custom attributes are encoded that way).

simonrozsival:
@MichalStrehovsky thanks for the suggestion!

src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.AnyMobile.cs
```diff
@@ -796,6 +797,8 @@ private void ThrowForModifiedManagedSslOptionsIfStarted()
             _socketHandler!.SslOptions = _socketHandler!.SslOptions;
         }
 
+        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2075:GetMethod",
+            Justification = "The DynamicDependency declarations will ensure the methods will be preserved.")]
```
eerhardt, line 801: Where are these DynamicDependency declaration? I don't see them in this file. It might be good to point to them in the Justification.

Or even better, annotate this method as `RequiresUnreferencedCode` and then suppress the warnings at the callsites that have the DynamicDependency attributes on them.

MichalStrehovsky (commented):
Looks great! (Reviewing from the trimmability perspective only.)

Cc @AaronRobinsonMSFT - I think this is a nice candidate for #90081. We'll be able to simplify all of this and get rid of the reflection.

src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.AnyMobile.cs
simonrozsival: Thanks for the comment. All the call sites were in a separate file - `HttpClientHandler.AnyMobile.InvokeNativeHandler.cs`. I moved the method into that file so that they are right next to each other.

I'm not sure if the `RequiresUnreferencedCode` attribute would still make sense after the last change based on Michal's suggestions.

MichalStrehovsky: The comment was applicable to the previous iteration only. The code is trim safe by construction now.

akoeplinger (approved):

stephentoub:
Was release/8.0-rc1 correct or should it have been release/8.0 instead?

simonrozsival:
@stephentoub you're right, it shouldn't be RC1. I'll open a new backport.

src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.AnyMobile.InvokeNativeHandler.cs
```diff
@@ -156,5 +153,36 @@ private static HttpMessageHandler CreateNativeHandler()
...
+        {
+            return InvokeNativeHandlerMethod(getMethod, parameters: null, cachingKey!);
+        }
+
+        private void InvokeNativeHandlerSetter(Func<MethodInfo> getMethod, object? value, [CallerMemberName] string? cachingKey = null)
+        {
+            InvokeNativeHandlerMethod(getMethod, parameters: new object?[] { value }, cachingKey!);
+        }
+
+        private object InvokeNativeHandlerMethod(Func<MethodInfo> getMethod, object?[]? parameters, string cachingKey)
```
marek-safar, line 167: Why is it not static?

src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.AnyMobile.InvokeNativeHandler.cs
```diff
@@ -156,5 +153,36 @@ private static HttpMessageHandler CreateNativeHandler()
 
             return (HttpMessageHandler)_nativeHandlerMethod!.Invoke(null, null)!;
         }
+
+        private object InvokeNativeHandlerGetter(Func<MethodInfo> getMethod, [CallerMemberName] string? cachingKey = null)
+        {
+            return InvokeNativeHandlerMethod(getMethod, parameters: null, cachingKey!);
```
marek-safar, line 159: Please add Debug.Assert instead of only `!`

src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.AnyMobile.InvokeNativeHandler.cs
```diff
@@ -16,135 +18,130 @@ public partial class HttpClientHandler : HttpMessageHandler
...
 
-        [DynamicDependency("get_Credentials", NativeHandlerType, AssemblyName)]
-        private ICredentials GetCredentials() => (ICredentials)InvokeNativeHandlerMethod("get_Credentials");
+        private ICredentials GetCredentials()
+            => (ICredentials)InvokeNativeHandlerGetter(() => Type.GetType(NativeHandlerType)!.GetMethod("get_Credentials")!);
 
-        [DynamicDependency("set_Credentials", NativeHandlerType, AssemblyName)]
-        private void SetCredentials(ICredentials? value) => InvokeNativeHandlerMethod("set_Credentials", value);
+        private void SetCredentials(ICredentials? value)
+            => InvokeNativeHandlerSetter(() => Type.GetType(NativeHandlerType)!.GetMethod("set_Credentials")!, value);
```
marek-safar, line 144: Perhaps `Type.GetType(NativeHandlerType)` could come as an argument to the lambda and you could cache it too.

marek-safar, line 144: All the lambdas should be static (not sure why the analyzer does not complain).

src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.AnyMobile.InvokeNativeHandler.cs
```diff
@@ -156,5 +153,36 @@ private static HttpMessageHandler CreateNativeHandler()
 
             return (HttpMessageHandler)_nativeHandlerMethod!.Invoke(null, null)!;
         }
+
+        private object InvokeNativeHandlerGetter(Func<MethodInfo> getMethod, [CallerMemberName] string? cachingKey = null)
```
marek-safar, line 157: Perhaps we could use CallerLineNumberAttribute or some other key which is not a next string in the assembly

src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.AnyMobile.InvokeNativeHandler.cs
simonrozsival: As far as I understand it, by keeping the GetType and GetMethod calls together with constant arguments will give the trimmer enough information to keep only the methods that are actually used. If we cached the type, I think it wouldn't be trimmable anymore.

vitek-karas: That's true. We don't have a way to do this with caching the Type (without suppressing some warnings).
Once we have `UnsafeAccessorType` then this can be rewritten to use that and it will not use any reflection and be faster.

PR #91522, opened 2023-09-03
Unsigned check of index inside ElementAt

YohDeadfall:
# Description

This pull request brings an unsigned index check trick into `ElementAt` which should reduce the resulting instruction count.

# Customer Impact

A tiny boost.

# Regression

No regression.

# Testing

The existing tests cover the change too.

# Risk

No risk.

### DIFF
--- a/src/libraries/System.Linq/src/System/Linq/ElementAt.cs
+++ b/src/libraries/System.Linq/src/System/Linq/ElementAt.cs
@@ -87,12 +87,11 @@ namespace System.Linq
 
             if (source is IList<TSource> list)
             {
-                return index >= 0 && index < list.Count ? list[index] : default;
+                return (uint)index < (uint)list.Count ? list[index] : default;
             }
 
             TryGetElement(source, index, out TSource? element);
             return element;
-
         }
 
         /// <summary>Returns the element at a specified index in a sequence or a default value if the index is out of range.</summary>


danmoseley (approved):
I wonder whether codegen could assume count is non negative and do this itself.

stephentoub:
> I wonder whether codegen could assume count is non negative and do this itself.

Not for an arbitrary `IList<T>.Count` implementation.

YohDeadfall:
So, is that `index >= 0` was intentionally made to return a default value instead of validating the index by the actual `IList<T>` implementation because it can be broken and return a negative count and/or accept negative indices? Is it a real case?

danmoseley:
> Not for an arbitrary `IList<T>.Count` implementation.

Isn't that an assumption we're making here? If Count is negative, this changes behavior.

eiriktsarpalis:
Lot of our code assumes `ICollection/IList.Count` values to always be non-negative and can break in unexpected ways if we pass it pathological implementations like that. One example that jumps to mind is the `TryGetNonEnumeratedCount` implementation which is used in a lot of places internally:

https://github.com/dotnet/runtime/blob/90042669ef7e2d829aa605cb3751760b6161d12a/src/libraries/System.Linq/src/System/Linq/Count.cs#L102-L106

danmoseley:
I almost wonder whether an attribute might be helpful, like `[NonNegative]` -- ie a hint for codegen -- given it's super common to store non negative values in signed types like int as they're the natural .NET types. Does codegen make this assumption today for concrete `List<T>` ?

@AndyAyersMS

stephentoub:
> Does codegen make this assumption today for concrete List<T> ?

No

EgorBo:
> I almost wonder whether an attribute might be helpful, like `[NonNegative]` -- ie a hint for codegen that negative values may have undefined behavior -- given it's super common to store non negative values in signed types like int as they're the natural .NET types. Does codegen make this assumption today for concrete `List<T>` ?
> 
> @AndyAyersMS

We try to avoid custom attributes (they're not cheap to check) + afair, the length can be in fact negative in a non-safe multithread environment. Overall, List<> has many perf issues compared to plain array like e.g. its indexer has two bound checks for the same reason

AndyAyersMS:
Yeah, I don't see an attribute being helpful here.

danmoseley:
OK -- what criteria should we use in the C# when using this uint trick for arbitrary types (eg IList<T>`). Are we always comfortable doing it -- as we did here? In what circumstances would we pause -- maybe when there's unsafe code nearby?

danmoseley:
> Overall, List<> has many perf issues compared to plain array like e.g. its indexer has two bound checks for the same reason

I assume preventing negative count (interlocked operation) would be more expensive than keeping those two bound checks?

stephentoub:
> OK -- what criteria should we use in the C# when using this uint trick for arbitrary types

We generally shouldn't; security/correctness concerns aside, it makes the intent harder to understand/read.  It's most valuable on really hot paths and when subsequent optimizations are possible because of it, e.g. when the JIT can actually eliminate a resulting bounds check. This typically only applies to spans/arrays/strings.

> maybe when there's unsafe code nearby?

We can basically _never_ trust such a `Count` member from a security perspective. Even with `List<T>`, even if we validate some `i` against its `Count`, that 'i' could be out-of-bounds at the very next moment if that `List<T>` might have come from other code that could still be manipulating it, or could have resulted in the instance being corrupted, such that we should not be using unsafe code to bypass the JIT's defenses here.

> I assume preventing negative count (interlocked operation) would be more expensive than keeping those two bound checks?

Yes. Interlockeds would be prohibitively expensive.  They'd also be insufficient; you'd very likely need a full lock to ensure that the `List<T>`'s `_size` and `_items` remained consistent with each other.

danmoseley:
> We generally shouldn't; security/correctness concerns aside, it makes the intent harder to understand/read. It's most valuable on really hot paths

OK so we are OK with this PR because it's a really hot path? fair enough

Issue #69705, opened 2022-05-24
[LoongArch64] A plan for amending the LoongArch64‘s port


shushanhf:
### Description

After the supporting for LoongArch64 finished by #59561 , 
there are still some TODOs for LA that should be implemented.

Here creating a new issue to track these.

(1) Fixing all the bugs anytime;
(2) Amend the LoongArch64's ABI;
(3) Enable the promote feature;
(4) Pushing the crossgen2 for LA;
(5) Supporting the SIMD/intrinsic for LA;
(6) ...

The CoreRoot for LoongArch64 will be updated  [here](https://github.com/shushanhf/LoongArch64-Blog-for-.NET-Mono/tree/main/rumtime_main_CoreRoot)

### Reproduction Steps

Testing all the test-cases of the runtime on LoongArch64's Linux.

### Expected behavior

The test cases should pass.

### Actual behavior

Failed.

### Known Workarounds

Using the SDK6-LoongArch64 which published by  Loongson based on the offical-SDK6.

### Configuration

LoongArch64’s Linux

shushanhf:
@jkotas @BruceForstall 

we will push some PRs to amend the LA's  port after we had fixed some test cases on LA-linux.

huoyaoyuan:
> Supporting the SIMD/intrinsic for LA

I've read the instruction spec for LA, and found some instructions interesting. As a bread new architecture, we may be concerned about exposing intrinsics in public API. However, we can implement an internal intrinsic class first and use in BitOperations etc.

shushanhf:
> > Supporting the SIMD/intrinsic for LA
> 
> I've read the instruction spec for LA, and found some instructions interesting. As a bread new architecture, we may be concerned about exposing intrinsics in public API. However, we can implement an internal intrinsic class first and use in BitOperations etc.

Thanks.

am11:
@shushanhf, should we have `native->uc_mcontext.__fcsr = lpContext->Fcsr;` before this line: https://github.com/dotnet/runtime/blob/48a3dbe5509f53f89db0a48c8285a2eec0758fea/src/coreclr/pal/src/thread/context.cpp#L626 as we do the opposite here: https://github.com/dotnet/runtime/blob/48a3dbe5509f53f89db0a48c8285a2eec0758fea/src/coreclr/pal/src/thread/context.cpp#L776

shushanhf:
[...]

PR #89781, opened 2023-08-01
Fix handling ThreadAbortException at the end of catch

janvorli:
After the catch handler for ThreadAbortException exits, the
exception needs to be rethrown immediately. This was not
implemented on Unix and this change makes that work on 
Unix amd64, arm64 and arm.

Close #72703

PR #91531, opened 2023-09-04
[LoongArch64, RISC-V] Fix handling ThreadAbortException at the end of catch for LoongArch64 and RISC-V.

LuckyXu-HF:

Fix handling ThreadAbortException at the end of catch for LoongArch64 and RISC-V.

### DIFF
--- a/src/coreclr/vm/exceptionhandling.cpp
+++ b/src/coreclr/vm/exceptionhandling.cpp
@@ -808,7 +808,7 @@ UINT_PTR ExceptionTracker::FinishSecondPass(
     {
         CopyOSContext(pThread->m_OSContext, pContextRecord);
         SetIP(pThread->m_OSContext, (PCODE)uResumePC);
-#if defined(TARGET_UNIX) && !(defined(TARGET_AMD64) || defined(TARGET_ARM64) || defined(TARGET_ARM) || defined(TARGET_LOONGARCH64))
+#if defined(TARGET_UNIX) && !(defined(TARGET_AMD64) || defined(TARGET_ARM64) || defined(TARGET_ARM) || defined(TARGET_LOONGARCH64) || defined(TARGET_RISCV64))
         uAbortAddr = NULL;
 #else
         uAbortAddr = (UINT_PTR)COMPlusCheckForAbort(uResumePC);
--- a/src/coreclr/vm/exceptionhandling.cpp
+++ b/src/coreclr/vm/exceptionhandling.cpp
@@ -808,7 +808,7 @@ UINT_PTR ExceptionTracker::FinishSecondPass(
     {
         CopyOSContext(pThread->m_OSContext, pContextRecord);
         SetIP(pThread->m_OSContext, (PCODE)uResumePC);
-#if defined(TARGET_UNIX) && !(defined(TARGET_AMD64) || defined(TARGET_ARM64) || defined(TARGET_ARM) || defined(TARGET_LOONGARCH64) || defined(TARGET_RISCV64))
+#if defined(TARGET_UNIX) && defined(TARGET_X86)
         uAbortAddr = NULL;
 #else
         uAbortAddr = (UINT_PTR)COMPlusCheckForAbort(uResumePC);
--- a/src/coreclr/vm/exceptionhandling.cpp
+++ b/src/coreclr/vm/exceptionhandling.cpp
@@ -808,7 +808,7 @@ UINT_PTR ExceptionTracker::FinishSecondPass(
     {
         CopyOSContext(pThread->m_OSContext, pContextRecord);
         SetIP(pThread->m_OSContext, (PCODE)uResumePC);
-#if defined(TARGET_UNIX) && !(defined(TARGET_AMD64) || defined(TARGET_ARM64) || defined(TARGET_ARM))
+#if defined(TARGET_UNIX) && !(defined(TARGET_AMD64) || defined(TARGET_ARM64) || defined(TARGET_ARM) || defined(TARGET_LOONGARCH64))
         uAbortAddr = NULL;
 #else
         uAbortAddr = (UINT_PTR)COMPlusCheckForAbort(uResumePC);
--- a/src/coreclr/vm/loongarch64/asmhelpers.S
+++ b/src/coreclr/vm/loongarch64/asmhelpers.S
@@ -669,29 +669,7 @@ LEAF_END JIT_GetSharedGCStaticBase_SingleAppDomain, _TEXT
 
 .endm
 
-// ------------------------------------------------------------------
-//
-// Helpers for ThreadAbort exceptions
-//
-        NESTED_ENTRY RedirectForThreadAbort2, _TEXT, NoHandler
-        PROLOG_SAVE_REG_PAIR_INDEXED  22, 1, 16
-
-        // stack must be 16 byte aligned
-        CHECK_STACK_ALIGNMENT
-
-        // On entry:
-        //
-        // a0 = address of FaultingExceptionFrame
-        //
-        // Invoke the helper to setup the FaultingExceptionFrame and raise the exception
-        bl    C_FUNC(ThrowControlForThread)
-
-        // ThrowControlForThread doesn't return.
-        EMIT_BREAKPOINT
-
-        NESTED_END RedirectForThreadAbort2, _TEXT
-
-GenerateRedirectedStubWithFrame RedirectForThreadAbort, RedirectForThreadAbort2
+GenerateRedirectedStubWithFrame RedirectForThreadAbort, ThrowControlForThread
 
 // ------------------------------------------------------------------
 // ResolveWorkerChainLookupAsmStub
--- a/src/coreclr/vm/loongarch64/asmhelpers.S
+++ b/src/coreclr/vm/loongarch64/asmhelpers.S
@@ -631,6 +631,68 @@ LEAF_ENTRY JIT_GetSharedGCStaticBase_SingleAppDomain, _TEXT
     bl  JIT_GetSharedGCStaticBase_Helper
 LEAF_END JIT_GetSharedGCStaticBase_SingleAppDomain, _TEXT
 
+// Make sure the `FaultingExceptionFrame_StackAlloc` is 16-byte aligned.
+#define FaultingExceptionFrame_StackAlloc (SIZEOF__GSCookie + SIZEOF__FaultingExceptionFrame + 0x8)
+#define FaultingExceptionFrame_FrameOffset SIZEOF__GSCookie
+
+.macro GenerateRedirectedStubWithFrame stub, target
+
+    //
+    // This is the primary function to which execution will be redirected to.
+    //
+    NESTED_ENTRY \stub, _TEXT, NoHandler
+
+        //
+        // IN: ra: original IP before redirect
+        //
+
+        PROLOG_SAVE_REG_PAIR_INDEXED  22, 1, 16
+
+        // alloc stack for FaultingExceptionFrame.
+        addi.d  $sp, $sp, -FaultingExceptionFrame_StackAlloc
+
+        // stack must be 16 bytes aligned
+        CHECK_STACK_ALIGNMENT
+
+        // Save pointer to FEF for GetFrameFromRedirectedStubStackFrame
+        addi.d  $a0, $sp, FaultingExceptionFrame_FrameOffset
+
+        // Prepare to initialize to NULL
+        st.d    $r0, $a0, 0  // Initialize vtbl (it is not strictly necessary)
+        st.d    $r0, $a0, FaultingExceptionFrame__m_fFilterExecuted  // Initialize BOOL for personality routine
+
+        bl   C_FUNC(\target)
+        // Target should not return.
+        EMIT_BREAKPOINT
+
+    NESTED_END \stub, _TEXT
+
+.endm
+
+// ------------------------------------------------------------------
+//
+// Helpers for ThreadAbort exceptions
+//
+        NESTED_ENTRY RedirectForThreadAbort2, _TEXT, NoHandler
+        PROLOG_SAVE_REG_PAIR_INDEXED  22, 1, 16
+
+        // stack must be 16 byte aligned
+        CHECK_STACK_ALIGNMENT
+
+        // On entry:
+        //
+        // a0 = address of FaultingExceptionFrame
+        //
+        // Invoke the helper to setup the FaultingExceptionFrame and raise the exception
+        bl    C_FUNC(ThrowControlForThread)
+
+        // ThrowControlForThread doesn't return.
+        EMIT_BREAKPOINT
+
+        NESTED_END RedirectForThreadAbort2, _TEXT
+
+GenerateRedirectedStubWithFrame RedirectForThreadAbort, RedirectForThreadAbort2
+
 // ------------------------------------------------------------------
 // ResolveWorkerChainLookupAsmStub
 //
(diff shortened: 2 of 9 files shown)


LuckyXu-HF:
This PR mainly fixes `System.Runtime.Tests.ControlledExecutionTests.CancelInTryAndExitCatchNormally` in `System.Runtime.Tests` after #89781.
Based on this PR, The `System.Runtime.Tests` has all passed on LoongArch64:
<pre>
Testing: @@@@@ home/xuliangyu/libs-8-tests/System.Runtime.Tests/Release/net8.0-unix @@@@@
  Discovering: System.Runtime.Tests (method display = ClassAndMethod, method display options = None)
  Discovered:  System.Runtime.Tests (found 9232 of 9279 test cases)
  Starting:    System.Runtime.Tests (parallel test collections = on, max threads = 4)
    System.Tests.ArgIteratorTests.ArgIterator_GetNextArgType [SKIP]
      Condition(s) not met: "IsArgIteratorSupported"
    System.Tests.ArgIteratorTests.ArgIterator_GetRemainingCount_GetNextArg [SKIP]
      Condition(s) not met: "IsArgIteratorSupported"
    System.Tests.StringTests.IndexOf_SingleLetter(s: "Hello", target: '\0', startIndex: 0, count: 5, expected: -1) [SKIP]
      Target \0 is not supported in ICU
    System.Tests.StringTests.LastIndexOf_NullInStrings [SKIP]
      Condition(s) not met: "IsNlsGlobalization"
    System.Tests.StringTests.IndexOf_NullInStrings [SKIP]
      Condition(s) not met: "IsNlsGlobalization"
    System.Tests.StringTests.EndsWith_NullInStrings_NonOrdinal [SKIP]
      Condition(s) not met: "IsNlsGlobalization"
    System.Tests.StringTests.StartsWith_NullInStrings_NonOrdinal [SKIP]
      Condition(s) not met: "IsNlsGlobalization"
    System.Tests.StringTests.EndsWith_NullInStrings [SKIP]
      Condition(s) not met: "IsNlsGlobalization"
    System.Tests.StringTests.StartsWith_NullInStrings [SKIP]
      Condition(s) not met: "IsNlsGlobalization"
    System.Runtime.Tests.JitInfoTests.JitInfoIsNotPopulated [SKIP]
      Condition(s) not met: "IsMonoAOT"
    System.Tests.DateTimeOffsetTests.ToLocalTime_MinValue [SKIP]
      Condition(s) not met: "IsMinValueNegativeLocalOffset"
    System.Tests.DateTimeOffsetTests.ToLocalTime_Ambiguous [SKIP]
      Condition(s) not met: "IsPacificTime"
    System.Tests.TimeZoneInfoTests.UnsupportedImplicitConversionTest [SKIP]
      Condition(s) not met: "DoesNotSupportIanaNamesConversion"
  Finished:    System.Runtime.Tests
=== TEST EXECUTION SUMMARY ===
   System.Runtime.Tests  Total: 65391, Errors: 0, Failed: 0, Skipped: 13, Time: 42.992s
</pre>

LuckyXu-HF:
cc @shushanhf ,Could you please review this PR?
Thanks

shushanhf (approved):
LGTM
Thanks

shushanhf:
cc  @clamp03 
Also add RISC-V64.

clamp03 (approved):
Thank you so much!!!

cc @alpencolt @t-mustafin @gbalykov @tomeksowi

src/coreclr/vm/riscv64/asmhelpers.S
```diff
@@ -539,6 +539,68 @@ LOCAL_LABEL(JIT_GetSharedGCStaticBase_SingleAppDomain_CallHelper):
...
+
+        // stack must be 16 bytes aligned
+        CHECK_STACK_ALIGNMENT
+
+        // Save pointer to FEF for GetFrameFromRedirectedStubStackFrame
+        addi  a0, sp, FaultingExceptionFrame_FrameOffset
+
+        // Prepare to initialize to NULL
+        sd    zero, a0, 0  // Initialize vtbl (it is not strictly necessary)
+        sd    zero, a0, FaultingExceptionFrame__m_fFilterExecuted  // Initialize BOOL for personality routine
```
clamp03, line 570: Please update to
`sd zero, 0(a0)`
`sd zero, FaultingExceptionFrame__m_fFilterExecuted(a0)`

Thank you!

LuckyXu-HF: Already updated, thank you.

shushanhf:
@jkotas  Could you please review this PR ?

src/coreclr/vm/exceptionhandling.cpp
```diff
@@ -808,7 +808,7 @@ UINT_PTR ExceptionTracker::FinishSecondPass(
     {
         CopyOSContext(pThread->m_OSContext, pContextRecord);
         SetIP(pThread->m_OSContext, (PCODE)uResumePC);
-#if defined(TARGET_UNIX) && !(defined(TARGET_AMD64) || defined(TARGET_ARM64) || defined(TARGET_ARM))
+#if defined(TARGET_UNIX) && !(defined(TARGET_AMD64) || defined(TARGET_ARM64) || defined(TARGET_ARM) || defined(TARGET_LOONGARCH64) || defined(TARGET_RISCV64))
         uAbortAddr = NULL;
 #else
```
jkotas, line 813:
```suggestion
```
This should not be needed.

The only platform where the #if condition true is Linux x86 that looks like a bug.

jkotas: @alpencolt @clamp03 Is this still needed for Linux x86?

gbalykov: > Is this still needed for Linux/Tizen x86?

I suppose yes, but not sure, because main branch has been crashing on linux x86 for a long time (all coreclr tests fail). We did not have time to investigate that yet.

LuckyXu-HF: > > Is this still needed for Linux/Tizen x86?
> 
> I suppose yes, but not sure, because main branch has been crashing on linux x86 for a long time (all coreclr tests fail). We did not have time to investigate that yet.

@jkotas  OK, thank you.
If this still needed for Linux/Tizen x86 , perhaps we can modify it this way?
```suggestion
-#if defined(TARGET_UNIX) && !(defined(TARGET_AMD64) || defined(TARGET_ARM64) || defined(TARGET_ARM) || defined(TARGET_LOONGARCH64) || defined(TARGET_RISCV64))
+#if defined(TARGET_UNIX) && defined(TARGET_X86)
         uAbortAddr = NULL;
 #else
         uAbortAddr = (UINT_PTR)COMPlusCheckForAbort(uResumePC);
```

src/coreclr/vm/riscv64/asmhelpers.S
```diff
@@ -539,6 +539,68 @@ LOCAL_LABEL(JIT_GetSharedGCStaticBase_SingleAppDomain_CallHelper):
...
+        // IN: ra: original IP before redirect
+        //
+
+        PROLOG_SAVE_REG_PAIR_INDEXED  fp, ra, 16
+
+        // alloc stack for FaultingExceptionFrame.
+        addi  sp, sp, -FaultingExceptionFrame_StackAlloc
+
+        // stack must be 16 bytes aligned
+        CHECK_STACK_ALIGNMENT
```
alpencolt, line 563: I think it's not correct. Original ARM64 code always align stack, but `CHECK_STACK_ALIGNMENT` works only on debug mode.

alpencolt (changes requested):

src/coreclr/vm/riscv64/asmhelpers.S
shushanhf: The line:542:543,
```
// Make sure the `FaultingExceptionFrame_StackAlloc` is 16-byte aligned.
#define FaultingExceptionFrame_StackAlloc (SIZEOF__GSCookie + SIZEOF__FaultingExceptionFrame + 0x8)
```
we define the `FaultingExceptionFrame_StackAlloc`  is 16-byte aligned while the ARM64 is 8-byte aligned.

LuckyXu-HF: Yes, as @shushanhf said, this is a small optimization.

alpencolt: May be I don't understand your idea clearly, but comment in ARM64 says:
```
; At this point, the stack maybe misaligned if the thread abort was asynchronously
; triggered in the prolog or epilog of the managed method. For such a case, we must
; align the stack before calling into the VM.
;
; Runtime check for 16-byte alignment.
```
That means that stack might be already misaligned before `addi sp, sp, -FaultingExceptionFrame_StackAlloc`.
@jkotas am I right?

jkotas: Yes, you are right. The thread abort stops the thread, checks whether the thread is inside managed code. If the thread is inside managed code, it patches the thread IP to point to this method. If the thread happened to be stopped in the middle of epilog or prolog, the stack is going to be misaligned.

shushanhf: First, for LoongArch64, we always keep the sp is 16-byte aligned, even in the manged code.
Second, during the processes of the abort by vm, I think the vm can get the sp whether is 16-byte aligned by the the FaultingExceptionFrame and should throw the error for unaligned sp.

shushanhf: > May be I don't understand your idea clearly, but comment in ARM64 says:
> 
> ```
> ; At this point, the stack maybe misaligned if the thread abort was asynchronously
> ; triggered in the prolog or epilog of the managed method. For such a case, we must
> ; align the stack before calling into the VM.
> ;
> ; Runtime check for 16-byte alignment.
> ```

Where is the sp is not 16 byte aligned within the prolog or epilog ?
For LoongArch64, even the stack is too large while the sp should be adjust by two or more times, every adjusting the sp is 16 byte aligned. I think the ARM64 should be also this.

I think the prolog and epilog for ARM64, RISCV64 and LoongArch64, they should keep the sp always 16-byte aligned during the adjusting the sp.

jkotas: Ok, I can see that the sp should be always aligned given the codegen used for prologs/epilogs.

src/coreclr/vm/loongarch64/asmhelpers.S
```diff
@@ -631,6 +631,68 @@ LEAF_ENTRY JIT_GetSharedGCStaticBase_SingleAppDomain, _TEXT
...
+
+    NESTED_END \stub, _TEXT
+
+.endm
+
+// ------------------------------------------------------------------
+//
+// Helpers for ThreadAbort exceptions
+//
+        NESTED_ENTRY RedirectForThreadAbort2, _TEXT, NoHandler
```
jkotas, line 676: Why do we need `RedirectForThreadAbort2`? Can `RedirectForThreadAbort` call `ThrowControlForThread`directly?

jkotas: (It is needed on Windows in some cases, but I do not see why it should be needed on Linux.)

shushanhf: The `.macro GenerateRedirectedStubWithFrame` existed very long.
If we don't add new function liking `GenerateRedirectedStubWithFrame RedirectForThreadAbort, RedirectForThreadAbort2`, we can replace the `.macro GenerateRedirectedStubWithFrame`  by `RedirectForThreadAbort`.

LuckyXu-HF: > Why do we need `RedirectForThreadAbort2`? Can `RedirectForThreadAbort` call `ThrowControlForThread`directly?

Thank you for your advice, I will investigate this.

LuckyXu-HF: > Why do we need `RedirectForThreadAbort2`? Can `RedirectForThreadAbort` call `ThrowControlForThread`directly?

I have tested this, it's ok to use `RedirectForThreadAbort` call `ThrowControlForThreaddirectly` on LoongArch64,  if necessary, I will update this function implementation.
<pre>
Testing: @@@@@ home/xuliangyu/runtime/artifacts/bin/System.Runtime.Tests/Release/net8.0-unix @@@@@
  Discovering: System.Runtime.Tests (method display = ClassAndMethod, method display options = None)
  Discovered:  System.Runtime.Tests (found 1 of 9279 test case)
  Starting:    System.Runtime.Tests (parallel test collections = on, max threads = 1)
  Finished:    System.Runtime.Tests
=== TEST EXECUTION SUMMARY ===
   System.Runtime.Tests  Total: 1, Errors: 0, Failed: 0, Skipped: 0, Time: 8.085s
</pre>

src/coreclr/vm/riscv64/asmhelpers.S
alpencolt: > For LoongArch64, even the stack is too large while the sp should be adjusted by two or more times, every time adjusting the sp is 16 byte aligned. I think the ARM64 should be also this.

It doesn't work for ARM64, e.g. there is instruction like `str x1, [sp], 8` which change stack to `8`. Situation with RISC-V and LA64 is better since we change `sp` manualy, but can we grantee that it will work that way now and with future changes?

src/coreclr/vm/loongarch64/asmhelpers.S
jkotas: Yes, I think simpler is better.

src/coreclr/vm/riscv64/asmhelpers.S
shushanhf: > > For LoongArch64, even the stack is too large while the sp should be adjusted by two or more times, every time adjusting the sp is 16 byte aligned. I think the ARM64 should be also this.
> 
> It doesn't work for ARM64, e.g. there is instruction like `str x1, [sp], 8` which change stack to `8`. Situation with RISC-V and LA64 is better since we change `sp` manualy, but can we grantee that it will work that way now and with future changes?

This instruction is not used within the manged code and I think we should not use this instruction to change the sp.
The increases should be only used when store paried and keep 16-byte aligned when changing the sp.

The proper way is first allock the enough stack space and then store, rather than store one by one while changing the sp 8-byte each time.

This instruction should only be used by pointer except the sp pointer.

Now the prolog and epilog don't change the sp by this way.

src/coreclr/vm/loongarch64/asmhelpers.S
LuckyXu-HF: Ok, I have updated it.

src/coreclr/vm/riscv64/asmhelpers.S
LuckyXu-HF: According to RISC-V psABI, it required `$sp` to be 16-byte aligned. I think if the `$sp` here is not 16-byte aligned, then the `$sp` is likely to be incorrect already.

LuckyXu-HF: > It doesn't work for ARM64, e.g. there is instruction like `str x1, [sp], 8` which change stack to `8`. Situation with RISC-V and LA64 is better since we change `sp` manualy, but can we grantee that it will work that way now and with future changes?

I didn't modify the code for ARM64 in this commit, maybe we just need to focus on how to make LoongArch64 and RISC-V work properly.

jkotas (approved):
LGTM!

PR #89681, opened 2023-07-29
Emit armv8.4 ldapur* for volatile loads with contained offsets

EgorBo:
Contributes to https://github.com/dotnet/runtime/issues/64457

```cs
volatile int a;
volatile int b;
volatile int c;
volatile int d;

[Benchmark]
public int Test()
{
    return a + b + c + d;
}
```
Current codegen for `Test`:
```asm
; Assembly listing for method P:Test():int:this (Tier1)		
            stp     fp, lr, [sp, #-0x10]!		
            mov     fp, sp		
            add     x1, x0, #8		
            ldapr   w1, [x1]		
            add     x2, x0, #12		
            ldapr   w2, [x2]		
            add     w1, w1, w2		
            add     x2, x0, #16		
            ldapr   w2, [x2]		
            add     w1, w1, w2		
            add     x0, x0, #20		
            ldapr   w0, [x0]		
            add     w0, w1, w0		
            ldp     fp, lr, [sp], #0x10		
            ret     lr		
; Total bytes of code 60
```
New codegen for `Test`:
```asm
; Assembly listing for method P:Test():int:this (Tier1)
            stp     fp, lr, [sp, #-0x10]!
            mov     fp, sp
            ldapur  w1, [x0, #0x08]
            ldapur  w2, [x0, #0x0C]
            add     w1, w1, w2
            ldapur  w2, [x0, #0x10]
            add     w1, w1, w2
            ldapur  w0, [x0, #0x14]
            add     w0, w1, w0
            ldp     fp, lr, [sp], #0x10
            ret     lr
; Total bytes of code 44
```

Apple M2 Max:

| Method |        Job |               Toolchain |      Mean |     Error |    StdDev | Ratio |
|------- |----------- |------------------------ |----------:|----------:|----------:|------:|
|   Test | Job-LOXHRO |      /Core_Root_PR/corerun | 0.4724 ns | 0.0050 ns | 0.0045 ns |  0.89 |
|   Test | Job-MPJUHL | /Core_Root_main/corerun | 0.5327 ns | 0.0063 ns | 0.0059 ns |  1.00 |

Perf difference is not super exciting but it is what it is 🙂

PR #91553, opened 2023-09-04
ARM64: Use stlur for volatile stores

EgorBo:
```cs
volatile int x;

void Test() => x = 42;
```
Codegen diff on arm64 v8.4+ (e.g. Apple M2):
```diff
; Assembly listing for method Program:Test()
            stp     fp, lr, [sp, #-0x10]!
            mov     fp, sp
            mov     w1, #42
-           dmb     ish
-           str     w1, [x0, #0x08]
+           stlur   w1, [x0, #0x08]
            ldp     fp, lr, [sp], #0x10
            ret     lr
```


EgorBo:
@TIHan @dotnet/jit-contrib PTAL, it's a follow-up to https://github.com/dotnet/runtime/pull/89681 but for stores.

TLDR: on arm64-v8.4 we can use stlur instead of strl or str+dmb because the former now supports IMM offsets. stlur, just like ldapur, has acquire-release semantics which matches our memory model expectations.

EgorBo:
also, cc @VSadov

src/coreclr/jit/codegenarm64.cpp
```diff
@@ -4209,26 +4209,55 @@ void CodeGen::genCodeForStoreInd(GenTreeStoreInd* tree)
         {
             bool addrIsInReg   = addr->isUsedFromReg();
             bool addrIsAligned = ((tree->gtFlags & GTF_IND_UNALIGNED) == 0);
+            bool handledByRcpc = false;
 
-            if ((ins == INS_strb) && addrIsInReg)
+            // On arm64-v8.4+ we can use stlur* instructions with acquire/release semantics
+            // if the address is LEA with just imm offset (unscaled)
+            if (compiler->compOpportunisticallyDependsOn(InstructionSet_Rcpc2) && addrIsAligned &&
```
BruceForstall, line 4216: Maybe you should make this symmetric with the ldapur implementation, and only require `addrIsAligned` for the non-byte cases (INS_strh, INS_str)?

src/coreclr/jit/codegenarm64.cpp
```diff
@@ -4209,26 +4209,55 @@ void CodeGen::genCodeForStoreInd(GenTreeStoreInd* tree)
...
+                (tree->Addr()->OperIs(GT_LEA) && !tree->HasIndex() && (tree->Scale() == 1) &&
+                 emitter::emitIns_valid_imm_for_unscaled_ldst_offset(tree->Offset())))
```
BruceForstall, line 4218: Looks like an unnecessary extra level of parens here.
```suggestion
                tree->Addr()->OperIs(GT_LEA) && !tree->HasIndex() && (tree->Scale() == 1) &&
                emitter::emitIns_valid_imm_for_unscaled_ldst_offset(tree->Offset()))
```

src/coreclr/jit/codegenarm64.cpp
EgorBo: thanks, unified code with loads to have a single method to decide which load/store instruction to choose.

jakobbotsch:
Will this need a fix similar to #91668?

EgorBo:
> Will this need a fix similar to #91668?

I'd guess - yes? since it's the same ISA. 

```
//  stlur  1X011001000iiiii iiii00nnnnnttttt  9900 0000   [Xn imm(-256..+255)] ARMv8.4 RCPC2
//  stlurb 00011001000iiiii iiii00nnnnnttttt  1900 0000   [Xn imm(-256..+255)] ARMv8.4 RCPC2
//  stlurh 01011001000iiiii iiii00nnnnnttttt  5900 0000   [Xn imm(-256..+255)] ARMv8.4 RCPC2
```
doc: https://developer.arm.com/documentation/ddi0602/2023-06/Base-Instructions/STLUR--Store-Release-Register--unscaled--?lang=en

can you add them as well?

jakobbotsch:
> doc: https://developer.arm.com/documentation/ddi0602/2023-06/Base-Instructions/STLUR--Store-Release-Register--unscaled--?lang=en
> 
> can you add them as well?

Done (but untested)

EgorBo:
@BruceForstall PTAL refactored changes

src/coreclr/jit/codegenarmarch.cpp
```diff
@@ -1728,6 +1728,122 @@ void CodeGen::genCodeForIndexAddr(GenTreeIndexAddr* node)
...
+//
+// Arguments:
+//    currentIns   - the current instruction to perform load/store
+//    indir        - the indirection node representing the volatile load/store
+//    needsBarrier - OUT parameter. Set to true if an explicit memory barrier is required.
+//
+// Return Value:
+//    instruction to perform the volatile load/store with.
+//
+instruction CodeGen::genGetVolatileLdStIns(instruction currentIns, GenTreeIndir* indir, bool* needsBarrier)
```
BruceForstall, line 1743: Couldn't this function also do the `genIsValidIntReg(targetReg)` check instead of the callers?

BruceForstall (approved):
LGTM. One question/suggestion.

PR #91524, opened 2023-09-03
Add FreeBSD case to gcenv.

Thefrank:
Closes https://github.com/dotnet/runtime/issues/64935

PR #91554, opened 2023-09-04
Delete unused methods from CoreCLR PAL

jkotas:


jkotas:
https://github.com/dotnet/runtime/pull/91524#discussion_r1314767022

src/coreclr/pal/src/misc/sysinfo.cpp
```diff
@@ -249,426 +237,3 @@ GetSystemInfo(
...
-    lpBuffer->ullAvailVirtual = lpBuffer->ullAvailPhys;
-
-    LOGEXIT("GlobalMemoryStatusEx returns %d\n", fRetVal);
-    PERF_EXIT(GlobalMemoryStatusEx);
-
-    return fRetVal;
-}
-
-bool
-PAL_ReadMemoryValueFromFile(const char* filename, uint64_t* val)
```
AaronRobinsonMSFT, line 493: I think this is used in a few other places.

am11: It is used in cgroup, which also is now-unused code in PAL (GC has a separate copy). I think the overarching idea was to move all memory related PAL code to GC, which happened a couple of months ago, but these leftovers weren't cleaned up for some reason at the same time.

AaronRobinsonMSFT (approved):

src/coreclr/pal/src/misc/sysinfo.cpp
AaronRobinsonMSFT: > cgroup is used to read processor quota limits outside the GC.

Confirmed. I just tried to remove it also and hit that :(

am11: Ah, I see. Maybe in the future, we could rewire it so those values are read from single place (GC PAL).

PR #91556, opened 2023-09-04
[mono] Use minipal/random.c for random number generation in the runtime.

vargaz:


vargaz:
Looks like random.c needs a windows implementation.

PR #91557, opened 2023-09-04
Fix Interlocked.Or/And for arm64

EgorBo:
I assume it's a typo made during some code clean up - we stopped recognizing `Interlocked.Or` and `Interlocked.And` as jit intrinsics on arm64 (they're not accelerated on x86/64)
Example:
```cs
void Test(ref int x) => Interlocked.Or(ref x, 1);
```
Current codegen in Main:
```asm
; Method Program:Test(byref)
            stp     fp, lr, [sp, #-0x10]!
            mov     fp, sp
            ldr     w0, [x1]
            align   [0 bytes for IG03]
            align   [0 bytes]
            align   [0 bytes]
            align   [0 bytes]
G_M18773_IG03:
            orr     w2, w0, #1
            mov     w3, w0
            casal   w3, w2, [x1]
            cmp     w3, w0
            beq     G_M18773_IG05
            sxtw    w0, w3
            b       G_M18773_IG03
G_M18773_IG05:
            ldp     fp, lr, [sp], #0x10
            ret     lr
; Total bytes of code: 48
```
New codegen with this PR (arm64 v8.1+):
```asm
; Method Program:Test(byref)
            stp     fp, lr, [sp, #-0x10]!
            mov     fp, sp
            mov     w0, #1
            ldsetal w0, w0, [x1]
            ldp     fp, lr, [sp], #0x10
            ret     lr
; Total bytes of code: 24
```

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=395159&view=ms.vss-build-web.run-extensions-tab)


jakobbotsch (approved):

stephentoub:
Do we know when this regressed?

PR #91560, opened 2023-09-04
Convert all tests under baseservices to the merged test infrastructure

trylek:
This change converts all remaining tests under baseservices to use the merged test model.
Apart from a few tiny tweaks I found a somewhat funny case in the test

baseservices/RuntimeConfiguration/TestConfig.csproj

This test had two problematic characteristics w.r.t. test merging - it was launching itself as
a child process and it (ab)used the [Fact] attributes for "its own mini-harness" that naturally
didn't go well with the merged test infra. I have refactored the test to use a separate "Tester"
app (akin to similar cases like ParallelCrash) and I removed the [Fact] attributes as they were
actually superfluous because the test uses two additional attributes (ConfigProperty and
EnvVar) to mark the "interesting" methods.

Thanks

Tomas


markples:
How have you checked whether all tests have been preserved across the conversion?  It seems like the changes to how the internal test harness work could impact the number of reported tests and make this difficult.  (If so, do you have recommendations on how to review this with that in mind?)

markples:
For the tests that define a `Main`, I believe that `BuildAsStandalone` will be broken since it will try to generate an entry point.  I think setting `ReferenceXUnitWrapperGenerator` to `false` will fix this.

trylek:
Well, in the local testing I have so far just monitored the number of tests; in fact even that has changed slightly as I split one or two tests to multiple ```[Fact]```s (I tried to clearly separate those individual changes from the bulk conversions as individual commits on the PR). I'm still working on fixing the BasicTestWithMcj test, once I'm done with that, I think I'll compare the output xml files, I think the last time I wrote a one-off managed app for the purpose, I need to take a look if I still have it around but it's about half an hour's work to write.

trylek:
Thanks for the suggestion regarding ```ReferenceXUnitWrapperGenerator```, that should make it nicely visible which tests are problematic for some future cleanup wave.

src/tests/Directory.Build.targets
```diff
@@ -153,6 +153,7 @@
   <Target Name="CheckForTestOutsideOfGroup" BeforeTargets="Build">
     <Error Condition="'$(InMergedTestDirectory)' == 'true'
       and '$(OutputType)' == 'Exe'
+      and '$(CLRTestKind)' == 'BuildAndRun'
```
markples, line 156: It looks like I never had to deal with the kind of case that led to this (in the JIT tests).  I'm a bit concerned because this is the check that we don't exclude a test (say, it was `BuildOnly` with multiple `RunOnly` projects that specify different parameters - which likely should just become `[Theory]s`).  This covers both conversion of tests and authoring new ones (which happened, though this is less likely since test authoring is generally copy-and-paste and we've gotten past conversion of most tests).

On the other hand, `BuildOnly` is exactly how one should specify "this is a weird test - I will take care of running it".

Perhaps auditing is sufficient for these (easy here - I only see ParallelCrash* and stackoverflow* - it will be something to keep in mind, especially in GC)

src/tests/baseservices/RuntimeConfiguration/TestConfig.csproj
```diff
@@ -1,15 +1,12 @@
...
   <PropertyGroup>
-    <OutputType>Exe</OutputType>
-    <!-- This test provides no interesting scenarios for GCStress -->
-    <GCStressIncompatible>true</GCStressIncompatible>
-    <UnloadabilityIncompatible>true</UnloadabilityIncompatible>
-    <DisableProjectBuild Condition="'$(RuntimeFlavor)' == 'Mono'">true</DisableProjectBuild>
+    <!-- This is a separate app launched by the actual test in TestConfigTester.csproj -->
+    <RequiresProcessIsolation>true</RequiresProcessIsolation>
+    <CLRTestKind>BuildOnly</CLRTestKind>
+    <NoWarn>$(NoWarn);XUW1001</NoWarn>
```
markples, line 6: Is setting `ReferenceXUnitWrapperGenerator` to `false` appropriate here?

It would remove the need for the `NoWarn`. The other warnings might not be relevant anyway, and I think we don't want BuildAsStandalone trying to build a simple wrapper for this. So I somewhat think the answer is "yes", but I don't feel certain about it.

src/tests/baseservices/RuntimeConfiguration/TestConfigTester.csproj
```diff
@@ -0,0 +1,21 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="$(TestSourceDir)Common/CoreCLRTestLibrary/CoreCLRTestLibrary.csproj" />
+    <ProjectReference Include="$(MSBuildThisFileDirectory)TestConfig.csproj">
+      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+      <OutputItemType>Content</OutputItemType>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </ProjectReference>
+  </ItemGroup>
+</Project>
```
markples, line 21: I feel like this could all be part of wrapper generation but also that would take a bunch of work that belongs in the future - something like recognizing EnvVar and then executing that test (or a batch of similar tests) in a new process (sort of like RequiresProcessIsolation but at the individual test level so it would take some more wrapper work).

So I think I'm checking my understanding and saying you have the right thing for now.

src/tests/baseservices/ilasm_ildasm/regression/vswhidbey305155/305155.cs
```diff
@@ -1,15 +1,17 @@
...
+    public Type[] Types;
 }
 
 public class Test
 {
-	[MyAttribute(Types = new Type[]{typeof(string), typeof(void)})]
-	public static int Main() { return 0; }
+    [MyAttribute(Types = new Type[]{typeof(string), typeof(void)})]
+    [Fact]
+    public static void TestEntryPoint() { }
```
markples, line 16: This test is marked as a library (was it running before?), which I think will now break under BuildAsStandalone and can just be removed from the project.  I don't know if other Library annotations under baseservices are accurate.

trylek: I'll try to double-check, I think that for the merged tests we default to Library OutputMode.

markples: Oh, I think you're right.  We override it (questionably, but beyond the scope there of course) to Exe when BuildAsStandalone is set.  So it's ok but just redundant.

trylek:
@markples - I have stood up the xml differ and I have managed to convince myself that the results are fine:

<pre>
Left-only tests (5 total):
--------------------------
baseservices\exceptions\exceptioninterop\ExceptionInterop_ro\ExceptionInterop_ro
baseservices\exceptions\exceptioninterop\ExceptionInterop\ExceptionInterop
baseservices\exceptions\simple\ParallelCrashMainThread\ParallelCrashMainThread
baseservices\exceptions\simple\ParallelCrashWorkerThreads\ParallelCrashWorkerThreads
baseservices\TieredCompilation\BasicTestWithMcj\RunBasicTestWithMcj

Right-only tests (24 total):
----------------------------
_ExceptionInterop_ro::ExceptionInterop.ThrowManagedExceptionThroughNativeAndCatchInFrame()
_ExceptionInterop_ro::ExceptionInterop.ThrowNativeExceptionAndCatchInFrame()
_ExceptionInterop_ro::ExceptionInterop.ThrowNativeExceptionAndCatchInFrameWithFilter()
_ExceptionInterop_ro::ExceptionInterop.ThrowNativeExceptionAndCatchInFrameWithFinally()
_ExceptionInterop_ro::ExceptionInterop.ThrowNativeExceptionInFrameWithFinallyCatchInOuterFrame()
_ExceptionInterop::ExceptionInterop.ThrowManagedExceptionThroughNativeAndCatchInFrame()
_ExceptionInterop::ExceptionInterop.ThrowNativeExceptionAndCatchInFrame()
_ExceptionInterop::ExceptionInterop.ThrowNativeExceptionAndCatchInFrameWithFilter()
_ExceptionInterop::ExceptionInterop.ThrowNativeExceptionAndCatchInFrameWithFinally()
_ExceptionInterop::ExceptionInterop.ThrowNativeExceptionInFrameWithFinallyCatchInOuterFrame()
_ParallelCrashTester::ParallelCrashTester.ParallelCrashMainThread()
_ParallelCrashTester::ParallelCrashTester.ParallelCrashMainThreadAndWorkerThreads()
_ParallelCrashTester::ParallelCrashTester.ParallelCrashWorkerThreads()
baseservices\exceptions\stackoverflow\stackoverflow\stackoverflow
baseservices\exceptions\stackoverflow\stackoverflow3\stackoverflow3
baseservices\exceptions\StackTracePreserve\StackTracePreserveTests\StackTracePreserveTests
baseservices\exceptions\unhandled\unhandledTester\unhandledTester
baseservices\finalization\CriticalFinalizer\CriticalFinalizer
baseservices\ilasm_ildasm\regression\vswhidbey305155\305155\305155
baseservices\mono\runningmono\runningmono
baseservices\multidimmarray\enum\enum
baseservices\RuntimeConfiguration\TestConfigTester\TestConfigTester
baseservices\varargs\varargsupport_r\varargsupport_r
baseservices\varargs\varargsupport\varargsupport
</pre>

There are several interesting observations based on this diff.

* We have an ugly discontinuity regarding test naming based on whether a given test project contains one or multiple ```[Fact]```` clauses - when there are multiple Fact clauses, we drop the test path information and just name the test based on the method. I think we should probably concatenate these two or concatenate the test dll / script path with the method name or something, the completely different representation of tests with multiple test cases is confusing and makes it harder for developers to correlate their test failures to the source / binary code.

* We have a subtle difference with regard to reporting tests blocked in ```issues.targets```. The legacy infrastructure filtered them out at the msbuild level so that they never made it into the XUnit wrappers and simply aren't reported in any manner in the results. With the merged infrastructure we filter them out in a different manner and in the report we end up reporting them as "skipped". I think that's actually more appropriate, the only downside is that today there's no indication in the xml results file that the test in question was skipped due to an issues.targets clause; I have filed https://github.com/dotnet/runtime/issues/91562 to fix that. In the above list, that pertains to the "Right-only tests" ```305155```, ```runningmono```, ```enum```, ```varargsupport_r``` and ```varargsupport```.

* For ```stackoverflow```, ```stackoverflow3``` or ```unhandled```, I believe these need fixing in the merged wrapper generator. These projects are marked as ```BuildOnly``` and shouldn't be actually run, the xml result file pretends it has run the corresponding cmd / sh script but that is nonsense, there's no cmd / sh script getting generated for these. I'll follow up with Jeremy on fixing this, I'm not sure to what extent it's blocking for this particular PR. (In practice the projects actually getting run are ```stackoverflowtester``` and ```unhandledTester``` introduced as part of this PR.)

Thanks

Tomas

markples:
Thanks Tomas!

- re: discontinuity in naming: I believe that we ended up in this position because we had a legacy naming system (path/project) which couldn't differentiate between multiple tests in a project, and the "new" way is closer to (or the same as?) standard xunit naming.  The discontinuity was kept to minimize the number of tests that were renamed as part of adding merged test groups.  However, with the eventual hope of merging the builds too, perhaps moving more towards xunit-style (always use it) would have less eventual churn that going back to incorporating paths?  It is important for now that the assembly name is there (though it appears to be?) because we don't dedup class/test names.
- agreed on your analysis and new work item
- I believe that these are being caused by the RequiresProcessIsolation attribute, for which the processing doesn't check for BuildOnly, etc.  Is this needed in the final iteration?  If not, then instead of excluding BuildOnly, perhaps Dir.build.targets could consider this an error.
  - It seems odd that the wrapper is reporting it as run when it doesn't exist.  It looks like it is [this code](https://github.com/dotnet/runtime/blob/9ae1a7cb4e024fce01bee35f533c268858e3320b/src/tests/Common/CoreCLRTestLibrary/OutOfProcessTest.cs#L75-L79), which appears to be necessary but could be a point where tests are discarded incorrectly.

src/tests/baseservices/ilasm_ildasm/regression/vswhidbey305155/305155.cs
trylek: I have removed the superfluous Library OutputType annotation, thanks for spotting.

src/tests/baseservices/RuntimeConfiguration/TestConfig.csproj
trylek: Removed. good point.

trylek: (Removed NoWarn and added ReferenceXUnitWrapperGenerator=false to be precise.)

src/tests/Directory.Build.targets
trylek: I just retried and the check seems to be no longer needed so I removed it. I guess I was hitting the problem in some of the internal test dependencies that should specify ReferenceXUnitWrapperGenerator=false instead. Thanks for pointing that out!

trylek:
@markples - I have rebased this change against the latest main and I believe I have addressed all your pending PR feedback, can you please take another look when you have a chance?

Thanks Tomas

src/tests/baseservices/exceptions/stackoverflow/stackoverflowtester.cs
```diff
@@ -71,12 +69,13 @@ static bool TestStackOverflow(string testName, string testArgs, out List<string>
             return true;
         }
 
-        static bool TestStackOverflowSmallFrameMainThread()
+        [Fact]
+        public static bool TestStackOverflowSmallFrameMainThread()
```
markples, line 73: `[Fact]` returning `bool` probably should be rejected by the wrapper generation.  (I'm guessing that the value will be ignored today.)

trylek: Thanks for pointing that out, I have updated the ```[Fact]``` methods in the ```stackoverflowtester.cs``` source file, according to your previous comments I think you might like the additional delta.

markples (approved):

PR #91566, opened 2023-09-05
Improve scalability of monitored Counter / UpDownCounter

stephentoub:
Avoid locking around every Update. Instead, use interlocked operations, and for CounterAggregator, partition the value being updated into one per core rather than one all-up. Without the partitioning, the compare-exchange loop can actually be measurably slower than the use of locking under heavy contention. With the partitioning, given we're dealing with floating-point, this can conceivably end up changing the reported values slightly, since we end up changing the order of operations (but with concurrency, such order of operations was already non-deterministic).

Contributes to https://github.com/dotnet/aspnetcore/issues/50412

|                 Method |         Toolchain |     Mean | Ratio |
|----------------------- |------------------ |---------:|------:|
|   UpDownCounter_Serial | \main\corerun.exe | 36.65 ms |  1.00 |
|   UpDownCounter_Serial |   \pr\corerun.exe | 24.77 ms |  0.68 |
|                        |                   |          |       |
| UpDownCounter_Parallel | \main\corerun.exe | 55.55 ms |  1.00 |
| UpDownCounter_Parallel |   \pr\corerun.exe | 14.02 ms |  0.25 |
|                        |                   |          |       |
|         Counter_Serial | \main\corerun.exe | 36.27 ms |  1.00 |
|         Counter_Serial |   \pr\corerun.exe | 25.15 ms |  0.69 |
|                        |                   |          |       |
|       Counter_Parallel | \main\corerun.exe | 58.72 ms |  1.00 |
|       Counter_Parallel |   \pr\corerun.exe | 13.97 ms |  0.22 |

```C#
using System.Diagnostics.Metrics;
using System.Diagnostics.Tracing;

public class Tests
{
    const int Iters = 1_000_000;

    private Meter _meter;
    private UpDownCounter<int> _upDownCounter;
    private Counter<int> _counter;
    private MetricsEventListener _listener;

    [GlobalSetup]
    public void Setup()
    {
        _meter = new Meter("Example");
        _upDownCounter = _meter.CreateUpDownCounter<int>("upDownCounter");
        _counter = _meter.CreateCounter<int>("counter");
        _listener = new MetricsEventListener();
    }

    [GlobalCleanup]
    public void Cleanup()
    {
        _listener.Dispose();
        _meter.Dispose();
    }

    [Benchmark]
    public void UpDownCounter_Serial()
    {
        for (int i = 0; i < Iters; i++)
        {
            _upDownCounter.Add(1);
            _upDownCounter.Add(-1);
        }
    }

    [Benchmark]
    public void UpDownCounter_Parallel()
    {
        Parallel.For(0, Iters, i =>
        {
            _upDownCounter.Add(1);
            _upDownCounter.Add(-1);
        });
    }

    [Benchmark]
    public void Counter_Serial()
    {
        for (int i = 0; i < Iters; i++)
        {
            _counter.Add(1);
            _counter.Add(1);
        }
    }

    [Benchmark]
    public void Counter_Parallel()
    {
        Parallel.For(0, Iters, i =>
        {
            _counter.Add(1);
            _counter.Add(1);
        });
    }

    private sealed class MetricsEventListener : EventListener
    {
        protected override void OnEventSourceCreated(EventSource eventSource)
        {
            if (eventSource.Name == "System.Diagnostics.Metrics")
            {
                EnableEvents(eventSource, EventLevel.LogAlways, EventKeywords.All, new Dictionary<string, string>() { { "Metrics", "Example\\upDownCounter;Example\\counter" } });
            }
        }
    }
}
```


src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/CounterAggregator.cs
```diff
@@ -16,21 +24,56 @@ public CounterAggregator(bool isMonotonic)
...
+            // We're not guaranteed uncontented access, so we still need to add the value
+            // to the delta with synchronization. Contention could come from other threads
+            // assigned to the same slot or from Collect zero'ing out the delta.
+            double currentValue;
+            do
             {
-                _delta += value;
+                currentValue = delta.Value;
             }
+            while (Interlocked.CompareExchange(ref delta.Value, currentValue + value, currentValue) != currentValue);
```
EgorBo, line 48: Perhaps, worth filing an api proposal for `Interlocked.Add(double)` ?

stephentoub: What would it do?

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/CounterAggregator.cs
```diff
@@ -16,21 +24,56 @@ public CounterAggregator(bool isMonotonic)
...
+
+            // Get the delta best associated with the current thread, preferring to use core ID rather than
+            // thread ID to reduce contention.
+            ref PaddedDouble delta = ref deltas[
+#if NETCOREAPP2_1_OR_GREATER
+                Thread.GetCurrentProcessorId()
+#else
+                Environment.CurrentManagedThreadId
+#endif
+                % deltas.Length];
```
EgorBo, line 38: Should we use `% Environment.ProcessorCount` here instead? Thus it will be converted to a jit-time constant and the whole mod operation will be cheaper

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/CounterAggregator.cs
EgorBo: The same as what integer overloads do? E.g.:
```cs
public float Add(ref float location, float value)
{
    float initialValue, computedValue;
    do
    {
        initialValue = totalValue;
        computedValue = initialValue + value;
    }
    while (initialValue != Interlocked.CompareExchange(ref totalValue, 
        computedValue, initialValue));
    return computedValue;
}
```

stephentoub: The integer Add methods end up in intrinsics that result in a single lock add instruction on some platforms, no?

EgorBo: Yes, this won't be accelerated (although, e.g. Add on arm64 8.0 is also not accelerated), I was just thinking of hiding this a bit obscured impl under a simple API if users might needed - no idea if they do 🙂

stephentoub: > although, e.g. Add on arm64 8.0 is also not accelerated

Yes, I just only think we should be adding such apis if they're accelerated on at least one platform, especially when they're overloads of methods that are accelerated. Otherwise overloads end up having very different perf characteristics.

EgorBo (approved):

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/CounterAggregator.cs
```diff
@@ -16,21 +25,57 @@ public CounterAggregator(bool isMonotonic)
...
+            PaddedDouble[] deltas = _deltas;
+            Debug.Assert(deltas.Length == Environment.ProcessorCount);
+
+            // Get the delta best associated with the current thread, preferring to use core ID rather than
+            // thread ID to reduce contention.
+            ref PaddedDouble delta = ref deltas[
+#if NETCOREAPP2_1_OR_GREATER
+                Thread.GetCurrentProcessorId()
+#else
+                Environment.CurrentManagedThreadId
```
tarekgh, line 38: > Environment.CurrentManagedThreadId [](http://example.com/codeflow?start=15&length=35)

does this mean on netfx it is possible to be slower as we are not using the core Id?

tarekgh (approved):
LGTM. Will be good to have @noahfalk have a look too.

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/CounterAggregator.cs
stephentoub: > > Environment.CurrentManagedThreadId
> 
> does this mean on netfx it is possible to be slower as we are not using the core Id?

Slower than with GetCurrentProcessorId, yes, as it's more likely for there to be contention.

EgorBo: `Thread.CurrentThread.ManagedThreadId` should work fine on .NET Framework btw, it should use a jit opt that folds it to fast a helper call: https://sharplab.io/#v2:EYLgHgbALANALiAhgZwLYB8ACAmAjAWAChMAGAAk1wDoAVACwCcBTRAEwEsA7AcwG4jSFXAFZ+hIgIDMFbGQDCZAN5EyqslzhkAsgAoN6gJRkAvAD4y9ZmypyArg2ac4llqypbEnRNyasXbAElWMTUyIgBfIA===

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/CounterAggregator.cs
```diff
@@ -16,21 +24,56 @@ public CounterAggregator(bool isMonotonic)
...
+                // Add the delta to the aggregated value.
+                _aggregatedValue += delta;
+                aggregatedValue = _aggregatedValue;
             }
+
+            return new CounterStatistics(delta, _isMonotonic, aggregatedValue);
+        }
+
+        [StructLayout(LayoutKind.Explicit, Size = 128)] // common cache-line size
+        private struct PaddedDouble
```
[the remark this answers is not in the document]
huoyaoyuan: Out of scope for this PR, but we may re-evaluate the cache line size for new architectures. The padding size was originally introduced in 2018.

noahfalk:
Sorry I was OOF for last week and getting caught up.

I like this as something people could opt into but I worry about the memory usage if we make this the default implementation. In the case of the 80 core machines mentioned in the issue, we are trading in an 8 byte double for a ref to an array with 80*cache_line_size bytes, 80*64=~5KB per time series I think? MetricsEventSource has a default tracking limit of 1000 time series. A time series = 1 combination of counter name + all dimension values so its not that hard to hit the time series limit. For people running this in production it could be a +5MB VM cost. In some cases thats probably no big deal, but in a container with tight limits that might be a significant regression or an OOM.

In terms of fixes I think we could:
- cap parallelism at 8 as default behavior
- make the cap configurable using a new filter argument on MetricsEventSource and pass it down MetricsEventSource -> AggregationManager -> AggregatorStore -> Aggregator.
- Rather than put 1 value per cache line, use the entire cache line for different aggregated values belonging to same core. The data structure gets more complicated but we get much better memory usage density with hopefully minimal increase in contention.

I'd like to find out if they are hitting similar contention issues with the lock on histograms. Histograms are a much bigger data structure so defaulting them to be one copy-per-core probably isn't viable unless it is opt-in or we change the data structure to make them grow more dynamically than they already do.

EgorBo:
> 80*cache_line_size bytes, 80*64=~5KB per time series I think? MetricsEventSource has a default tracking limit of 1000 time series. A time series = 1 combination of counter name + all dimension values so its not that hard to hit the time series limit. For people running this in production it could be a +5MB VM cost

> cap parallelism at 8 as default behavior

I wonder if it's better to just remove padding then to cap it to 8 as the main beneficiar of this opt are many-cores systems.

so then 80*8=640B, *1000 = 642kb for the worst case.

noahfalk:
> I wonder if it's better to just remove padding

If that delivers a better perf result I am fine with it, though I still think we need to cap total memory used for these counters around 512KB (a somewhat arbitrarily selected value). If we had no padding presumably that raises the cap to 64 doubles per time series.

EgorBo:
A microbenchmark to play with (on a system with many cores): https://gist.github.com/EgorBo/fb944b46bf9b1ffd35b01bb3b5726e7f

JamesNK:
What about an adaptive solution? If the number of dimensions increases beyond a limit, then switch back to the old behavior to save memory?

And if a counter has a lot of dimensions, then there is less contention when updating counts because increments are spread out across the dimensions.

noahfalk:
> What about an adaptive solution? If the number of dimensions increases beyond a limit, then switch back to the old behavior to save memory?

I worry that the assignment of which time series are optimized for memory and which are optimized for latency will appear largely random based on creation order of the different time series. Under high contention the perf cliff seems severe enough I think I'd rather get an error asking me to raise the memory limits vs. having unexplained significant changes in perf. From I can see in the perf numbers, I think the contention impacts in @EgorBo's ARM multi-core machine are much larger than what I assume @stephentoub measured on x64.

> is there a lock per dimension?

Today there is one lock per time-series. (Time series being the a unique combination of instrument + all the dimension values, which I think is what you meant)

stephentoub:
@noahfalk, what would you like me to do with this? Should I do "cap parallelism at 8"?

noahfalk:
Yeah, I think capping at 8 is a fine place to start. I'd still be interested in making further improvements but its likely to get more involved. I assume you'd prefer to bank your incremental improvements and move on vs. working on more complicated options?

stephentoub:
> I assume you'd prefer to bank your incremental improvements and move on vs. working on more complicated options?

Yes, as that's acceptable to you, that's my preference.

noahfalk (approved):
LGTM, thanks @stephentoub!

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/CounterAggregator.cs
```diff
@@ -1,13 +1,24 @@
...
-        private double _delta;
         private double _aggregatedValue;
+        /// <summary>Per-core deltas.</summary>
+        /// <remarks>
+        /// Stored as an array of deltas rather than a single delta to reduce contention from
+        /// highly-parallel updates. The elements are padded to reduce false sharing.
+        /// The array is limited to a semi-arbitrary limit of 8 in order to avoid excessive memory
+        /// consumption when many counters are being used.
+        /// </remarks>
+        private readonly PaddedDouble[] _deltas = new PaddedDouble[Math.Min(Environment.ProcessorCount, 8)];
```
bencyoung-Fignum, line 21: @stephentoub shouldn't this be a max if the comment is accurate?

stephentoub: I believe they're both correct. We want to increase the number up to the processor count but not go above 8, so we want the processor count or 8, whichever is lower.

OwnageIsMagic: @bencyoung-Fignum min is ensuring limit, here it means 8 or less.

bencyoung-Fignum: Doh you are right, apologies!

PR #91572, opened 2023-09-05
Respect StackTraceHiddenAttribute

### LINKED ISSUE #91309: StackTraceHiddenAttribute not respected

MichalStrehovsky:
Fixes #91309.

NativeAOT sources stack trace data from two places: reflection metadata (if the method was reflection-visible), or specialized stack trace metadata. Finding out if a method should be hidden from stack traces in the former case is easy: just look for the attribute. The latter case requires compiler work.

In this PR, I'm extending the specialized stack trace metadata format to contain a bit that says if the frame should be hidden or not. I'm doing it in a way that we can also do this for compiler-generated methods (that otherwise don't show up in stack trace metadata).

The runtime behavior is similar to CoreCLR - the methods show up in individual stack frames, but if we stringify the stack trace, they'll be skipped. I'm adding a specialized test that tests the two sources of metadata.

I'm also marking the methods involved in the startup path as stack trace hidden.

There is one thing I skipped - CoreCLR has logic to force generate the stack trace metadata for the first frame. If we do that, the compiler-generated startup code starts showing up in stack traces again. Marking `Main` as stack trace hidden might lead to an empty stack trace.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/nativeaot/System.Private.StackTraceMetadata/src/Internal/StackTraceMetadata/StackTraceMetadata.cs
+++ b/src/coreclr/nativeaot/System.Private.StackTraceMetadata/src/Internal/StackTraceMetadata/StackTraceMetadata.cs
@@ -329,7 +329,7 @@ namespace Internal.StackTraceMetadata
 
             private struct StackTraceData : IComparable<StackTraceData>
             {
-                private const int IsHiddenFlag = 0x1;
+                private const int IsHiddenFlag = 0x2;
 
                 private readonly int _rvaAndIsHiddenBit;
 
--- a/src/coreclr/nativeaot/System.Private.StackTraceMetadata/src/Internal/StackTraceMetadata/StackTraceMetadata.cs
+++ b/src/coreclr/nativeaot/System.Private.StackTraceMetadata/src/Internal/StackTraceMetadata/StackTraceMetadata.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Reflection.Runtime.General;
 
 using Internal.Metadata.NativeFormat;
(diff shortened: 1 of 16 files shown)


src/coreclr/nativeaot/System.Private.StackTraceMetadata/src/Internal/StackTraceMetadata/StackTraceMetadata.cs
```diff
@@ -290,12 +314,43 @@ public string GetMethodNameFromRvaIfAvailable(int rva)
...
+                    return null;
+                }
+
                 return MethodNameFormatter.FormatMethodName(_metadataReader, data.OwningType, data.Name, data.Signature, data.GenericArguments);
             }
 
             private struct StackTraceData : IComparable<StackTraceData>
             {
-                public int Rva { get; init; }
+                private const int IsHiddenFlag = 0x1;
```
jkotas, line 332: Stealing the lowest bit like this tends to be a problem on Arm32, but we can deal with it if we ever get there.

jkotas (approved):
Thank you!

src/coreclr/nativeaot/System.Private.StackTraceMetadata/src/Internal/StackTraceMetadata/StackTraceMetadata.cs
MichalStrehovsky: We can just switch this to second highest bit - we already align everything at 4 bytes minimum because of ARM and fat function pointers. The assert should help guarding against this ever breaking the invariant.

```suggestion
                private const int IsHiddenFlag = 0x2;
```

PR #91578, opened 2023-09-05
Support for easy lazy with implicit operator

Ali-YousefiTelori:
Hi,
These changes let you use Lazy<T> in an easy way.

In the past we have to use Lazy<T> this way:

```CSharp
var lazy = new Lazy<string>(() => "data");
var value = lazy.Value;
```

with this feature, you can use it more easily:

```CSharp
Lazy<string> lazy = () => "data";
string value = lazy;
```

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Lazy.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Lazy.cs
@@ -505,6 +505,17 @@ namespace System
         /// </remarks>
         [DebuggerBrowsable(DebuggerBrowsableState.Never)]
         public T Value => _state == null ? _value! : CreateValue();
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


hez2010:
Please go through an API approval process before submitting PRs involving API changes.

stephentoub:
Thank you for the interest in contributing. As noted in a previous reply, https://github.com/dotnet/runtime/blob/main/docs/project/api-review-process.md outlines the process we follow for adding new APIs. It starts with an issue, so if you'd like to open one proposing these additions, please feel free; in the meantime, I'm closing this PR. That said, I think it's unlikely we'd choose to add such implicit operators; we avoid doing so when the operation allocates, blocks, potentially takes a long time, etc., and this has the potential to do all of those.

PR #91580, opened 2023-09-05
JIT: Fix invalid zero-init suppression for untracked variables

### LINKED ISSUE #91576: JIT: Bad codegen with try/finally
```csharp
// Generated by Fuzzlyn v1.6 on 2023-09-03 15:59:01
// Run on X64 Windows
// Seed: 11520325105937570553
// Reduced from 294.5 KiB to 0.7 KiB in 00:04:32
// Debug: Outputs False
// Release: Outputs True
public class Program
{
    public static IRuntime s_rt;
    public static void Main()
    {
        s_rt = new Runtime();
        bool vr1 = default(bool);
        try
        {
            long[] vr2 = default(long[]);
            var vr3 = vr2[0];
            bool vr4 = default(bool);
            vr1 = !vr4;
        }
        finally
        {
            ref bool vr5 = ref vr1;
            s_rt.WriteLine(vr5);
        }
    }
}

public interface IRuntime
{
    void WriteLine<T>(T value);
}

public class Runtime : IRuntime
{
    public void WriteLine<T>(T value) => System.Console.WriteLine(value);
}
```

jakobbotsch:
optRemoveRedundantZeroInits has logic to remove unnecessary zero inits if we can determine that the local will be zeroed in the prolog. In addition, it also has logic to suppress the prolog zero init if there is a dominating initialization already. The latter logic was trying to reason about liveness for untracked locals, which does not make sense.

Fix #91576

Also makes the handling for potential implicit control flow more precise, to counteract regressions.


jakobbotsch:
This sadly has some quite significant regressions, will see if I can address some of these.

jakobbotsch:

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=395646&view=ms.vss-build-web.run-extensions-tab) where we now need to zero init in some cases again.

AndyAyersMS (approved):
Are the handful cases where we now do less zero init ones where there were no eh successors?

jakobbotsch:
> Are the handful cases where we now do less zero init ones where there were no eh successors?

Yep, for example in https://github.com/dotnet/runtime/blob/cf65ea21bfcb4f53a416506d43d79900413e5b1b/src/tests/JIT/Regression/JitBlue/Runtime_85892/Runtime_85892.cs `vr4` no longer needs zero initing because we now realize that even though it is handler live, the handler is not reachable before the location where it is initialized.

It seems like we could be even more precise by tracking the set of locals that are actually live to handlers we have seen and only disqualifying those in this optimization, though I doubt it would be significantly better than what we have with this PR.

jakobbotsch:
Opened #91839 for the Fuzzlyn failures. The jitstress failure is #91705. The superpmi-diffs failure was the "No space left on device" due to some machines in a bad state.

PR #91596, opened 2023-09-05
Start vectorizing TensorPrimitives

stephentoub:
Note I added an ifdef for `NET8_0_OR_GREATER` because my assumption is all of this is going to be backported to the release/8.0 branch, at which point we'll also be building for net7.0.


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -0,0 +1,163 @@
...
+            if (x.Length > destination.Length)
+            {
+                ThrowHelper.ThrowArgument_DestinationTooShort();
+            }
+
+            ref float xRef = ref MemoryMarshal.GetReference(x);
+            ref float dRef = ref MemoryMarshal.GetReference(destination);
+            int remaining = x.Length;
+
+#if NET8_0_OR_GREATER
```
tannergooding, line 109: Are we compiling for net 7 or something, why would this ifdef be needed for the `*.netcore.cs` file?

stephentoub: See my comment in the PR description :)

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <exception cref="ArgumentException">Destination is too short.</exception>
+        /// <remarks>This method effectively does <c><paramref name="destination" />[i] = <paramref name="x" />[i] + <paramref name="y" /></c>.</remarks>
+        public static void Add(ReadOnlySpan<float> x, float y, Span<float> destination)
+        {
+            if (x.Length > destination.Length)
+            {
+                ThrowHelper.ThrowArgument_DestinationTooShort();
+            }
+
+            fixed (float* xPtr = &MemoryMarshal.GetReference(x), destPtr = &MemoryMarshal.GetReference(destination))
```
tannergooding, line 72: Can we avoid pinning by using `Unsafe.As` or `Unsafe.ReadUnaligned`?

I believe that generates good code on .NET Framework still and is also available downlevel.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -0,0 +1,161 @@
...
+                {
+                    Vector512<float> sum = Vector512.LoadUnsafe(ref xRef, (uint)i) + Vector512.LoadUnsafe(ref yRef, (uint)i);
+                    Vector512.StoreUnsafe(sum, ref dRef, (uint)i);
+
+                    i += Vector512<float>.Count;
+                }
+            }
+#endif
+
+            if (Vector256.IsHardwareAccelerated)
```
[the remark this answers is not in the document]
stephentoub: > Ideally we'd get a trailing mask and merge, then store the final result instead

If you can share an example of what that looks like, I can change it.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
stephentoub: Ok

stephentoub:
(I'm revamping as well, to share more code between the various operations. Will post an update shortly...)

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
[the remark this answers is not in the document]
tannergooding: Noting that if we're not supporting overlapping src/dst buffers, we can just backtrack by `remainder` and do 1x full Vector128/256/512 operation, since `Add` is idempotent

otherwise, we want something like: https://github.com/dotnet/machinelearning/blob/d6927515d8014442d35ee22817ec7c40b9e7eec7/src/Microsoft.ML.CpuMath/AvxIntrinsics.cs#L459-L564

which is an example in ML.NET that handles alignment and remaining data. It was written back around .NET Core 3.1, but the general premise should apply equally to the xplat APIs.

Basically we need a small upfront table covering the 4/8/16 different possible leading/trailing masks: https://github.com/dotnet/machinelearning/blob/d6927515d8014442d35ee22817ec7c40b9e7eec7/src/Microsoft.ML.CpuMath/AvxIntrinsics.cs#L24-L46

and we backtrack by `remainder` so that we do 1 last operation that exactly fills a Vector128/256/512, simply doing a conditional select of the already computed results with the new results.

----------------------------

Particularly when processing large data (as may be expected for tensors), handling alignment is important to avoid known perf penalties. We don't strictly have to pin the data, we can "opportunistically align" under the presumption the GC won't move the data by checking the base alignment of the underlying pointer. This will perform well if the GC doesn't move the data and remain correct even if it does (just no longer with the benefit of alignment).

For small data, handling "remaining" elements (that doesn't fit into a vector) efficiently is important and its pretty easy to do. Ideally we add some official APIs to help with this, which is part of the broader consideration around efficient masking support that has already been getting discussed alongside AVX-512, SVE, and general existing ISAs.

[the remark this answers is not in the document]
stephentoub: > Noting that if we're not supporting overlapping src/dst buffers

Are we?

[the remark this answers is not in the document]
tannergooding: I think most of the other APIs we've added to the BCL don't support them.

We should probably double check with @luisquintanilla / @michaelgsharp on whether ML.NET or SK need it.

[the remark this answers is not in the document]
michaelgsharp: I think for the case of ML.NET that @tannergooding is right about just having a 1x `Span<T>` API when we want it that way. At least thats what it looks like to me.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -0,0 +1,1062 @@
...
+    public static partial class TensorPrimitives
+    {
+        /// <summary>Computes the element-wise result of: <c><paramref name="x" /> + <paramref name="y" /></c>.</summary>
+        /// <param name="x">The first tensor, represented as a span.</param>
+        /// <param name="y">The second tensor, represented as a span.</param>
+        /// <param name="destination">The destination tensor, represented as a span.</param>
+        /// <exception cref="ArgumentException">Length of '<paramref name="x" />' must be same as length of '<paramref name="y" />'.</exception>
+        /// <exception cref="ArgumentException">Destination is too short.</exception>
+        /// <remarks>This method effectively does <c><paramref name="destination" />[i] = <paramref name="x" />[i] + <paramref name="y" />[i]</c>.</remarks>
+        public static unsafe void Add(ReadOnlySpan<float> x, ReadOnlySpan<float> y, Span<float> destination) =>
```
tannergooding, line 20: Can this public surface area stay shared and just the generic helper be abstracted?

Should be possible if you make the signature:
```csharp
private static unsafe void InvokeSpanSpanIntoSpan<TBinaryOperator>(
            ReadOnlySpan<float> x, ReadOnlySpan<float> y, Span<float> destination)
            where TBinaryOperator : struct, IBinaryOperator
```

and
```csharp
private static unsafe void InvokeSpanSpanIntoSpan<TBinaryOperator>(
            ReadOnlySpan<float> x, ReadOnlySpan<float> y, Span<float> destination, TBinaryOperator op = default)
            where TBinaryOperator : struct, IBinaryOperator
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -0,0 +1,1062 @@
...
+                oneVectorFromEnd = x.Length - Vector512<float>.Count;
+                if (i <= oneVectorFromEnd)
+                {
+                    // Loop handling one vector at a time.
+                    do
+                    {
+                        Vector512.StoreUnsafe(
+                            TUnaryOperator.Invoke(
+                                Vector512.LoadUnsafe(ref xRef, (uint)i)),
+                            ref dRef, (uint)i);
```
tannergooding, line 196: Just wondering, why this way rather than something like:
```csharp
Vector512<float> result = TUnaryOperator.Invoke(Vector512.LoadUnsafe(ref xRef, (uint)i));
result.StoreUnsafe(ref dRef, (uint)i);
```

or
```csharp
TUnaryOperator.Invoke(Vector512.LoadUnsafe(ref xRef, (uint)i))
              .StoreUnsafe(ref dRef, (uint)i);
```

I personally find them a little bit more readable than the nesting/indention here (but not a big deal by any means)

tannergooding (approved):

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: Ah, using `default` is a good idea. I'll do that.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: > Just wondering, why this way

I was mainly trying to be consistent with the other methods that had more arguments, e.g. three LoadUnsafe calls as arguments to Invoke. I also didn't realize there was an instance StoreUnsafe method.

I'll try to find a happy middle ground.

PR #91600, opened 2023-09-05
Add windows stylearguments format to coreclr test scripts

janvorli:
For some reason lost in the past, the generated Windows and Unix coreclr test scripts use different style of options passing. On windows, the option and its value are passed separated by space, e.g. `-coreroot c:\runtime\coreclr` while Unix separate them by `=`, e.g. `-coreroot=~/runtime/coreclr`. The Unix way has a disadvantage when I try to use command line completion to find the coreroot path I want to pass in. Most shells try to treat the `-coreroot=` as part of the path and so the completion doesn't work. So I keep doing an awkward thing - putting space after the `=` and deleting it after the completion.

This change adds the windows style option passing as an optional way to these scripts.


jkoritzinsky (approved):

trylek (approved):
LGTM, thank you!

Issue #88688, opened 2023-07-11
Simplify CMake build system now that our minimum is 3.20


jkoritzinsky:
There's a few simplifications we can do now that our minimum CMake version is 3.20 on all platforms:

- [ ] Remove the `_obj` suffix on our CMake targets and remove the "interface" targets now that `target_link_libraries` can link to object libraries.
- [ ] Remove the polyfills used with older CMake versions in `functions.cmake`
- [ ] Remove the `add_linker_flag` function and instead use `target_link_options`
- [ ] Remove our conditional setting of CMake policies for policies that are now always available.
- [ ] Enable CMake policy [CMP0117](https://cmake.org/cmake/help/latest/policy/CMP0117.html) and update the corresponding compile options throughout the repo.
- [ ] Enable CMake policy [CMP0092](https://cmake.org/cmake/help/latest/policy/CMP0092.html) and update the corresponding compile options throughout the repo.
- [ ] Use CMake's support to set up IPO optimizations: https://cmake.org/cmake/help/latest/policy/CMP0069.html, https://cmake.org/cmake/help/latest/module/CheckIPOSupported.html#module:CheckIPOSupported
- [ ] Use CMake's support to set up PIE options: https://cmake.org/cmake/help/latest/policy/CMP0083.html, https://cmake.org/cmake/help/latest/module/CheckPIESupported.html#module:CheckPIESupported
- [ ] Use the `FindPython3` CMake module instead of the deprecated `FindPythonInterp` module. The deprecated module is being removed in newer versions of CMake.
- [ ] Remove any CMake version checks that check for versions less than 3.20
- [ ] Use improvements in CMake generator expressions to simplify our CMake generator expressions (there are many cases where we can simplify expressions like `$<OR:$<CONFIG:X>,$<CONFIG:Y>>:-flag>` to `$<CONFIG:X,Y>`)
- [ ] Investigate if we can simplify the build configuration for Apple Silicon using CMake's support that was added in CMake 3.19
- [ ] Set [CMP0112](https://cmake.org/cmake/help/latest/policy/CMP0112.html) to the `NEW` behavior and fix any target dependency issues.
- [ ] Add [CMake presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) for common build command line settings (like our standard Debug/Checked/Release options) to help enable better IDE integration and simplify our build process a little. (We can only use features available in version 1 or 2 of the presets format).
[...]

PR #91632, opened 2023-09-05
Remove CMake version checks

jkoritzinsky:
Now that our minimum CMake version is 3.20, we can remove these version checks.

Contributes to https://github.com/dotnet/runtime/issues/88688


janvorli (approved):
LGTM, thank you!

src/tests/CMakeLists.txt
```diff
@@ -1,9 +1,7 @@
 cmake_minimum_required(VERSION 3.20)
 
 cmake_policy(SET CMP0042 NEW)
-if (CMAKE_VERSION VERSION_GREATER 3.15 OR CMAKE_VERSION VERSION_EQUAL 3.15)
-  cmake_policy(SET CMP0091 NEW)
-endif()
+cmake_policy(SET CMP0091 NEW)
```
jkotas, line 4: Do we need to set this explicitly? 

`cmake_mimum_required(VERSION 3.20)` should set these policies to NEW by default.

jkoritzinsky: I can check. I know that some of these policies need to be explicitly set to the `NEW` behavior. I believe this policy is one of them based on the docs.

jkotas: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html says that all policies are set to `NEW` behavior based on min version, it does not mention any exceptions.

> The cmake_minimum_required(VERSION) command implicitly invokes the [cmake_policy(VERSION)](https://cmake.org/cmake/help/latest/command/cmake_policy.html#command:cmake_policy) command to specify that the current project code is written for the given range of CMake versions. All policies known to the running version of CMake and introduced in the `<min>` (or `<max>`, if specified) version or earlier will be set to use NEW behavior.

jkoritzinsky: The documentation for this policy states the following:

> This policy was introduced in CMake version 3.15. Use the cmake_policy() command to set it to OLD or NEW explicitly. Unlike many policies, CMake version 3.27.4 does not warn when this policy is not set and simply uses OLD behavior.

jkoritzinsky: I'll try removing it and see what happens.

jkotas (approved):

PR #91633, opened 2023-09-05
[wasm] Use specific version of v8 for tests

radical:
On CI, provision `v8` using version from `eng/testing/ChromeVersions.props` and use that for running library tests, and Wasm.Build.Tests .

The only remaining build which will be using system `v8` after this would be the runtime tests.

### DIFF
--- a/eng/testing/tests.browser.targets
+++ b/eng/testing/tests.browser.targets
@@ -24,6 +24,7 @@
     <WasmIgnoreNet6WorkloadInstallErrors Condition="'$(WasmIgnoreNet6WorkloadInstallErrors)' != 'true'">false</WasmIgnoreNet6WorkloadInstallErrors>
     <!--<InstallWorkloadUsingArtifactsDependsOn>_GetWorkloadsToInstall;$(InstallWorkloadUsingArtifactsDependsOn)</InstallWorkloadUsingArtifactsDependsOn>-->
     <GetWorkloadInputsDependsOn>_GetWorkloadsToInstall;$(GetWorkloadInputsDependsOn)</GetWorkloadInputsDependsOn>
+    <!-- '/.dockerenv' - is to check if this is running in a codespace -->
     <InstallChromeForTests Condition="'$(InstallChromeForTests)' == '' and
                                         ('$(ContinuousIntegrationBuild)' != 'true' or Exists('/.dockerenv')) and
                                         '$(Scenario)' == 'WasmTestOnBrowser'">true</InstallChromeForTests>
--- a/eng/testing/tests.browser.targets
+++ b/eng/testing/tests.browser.targets
@@ -27,6 +27,9 @@
     <InstallChromeForTests Condition="'$(InstallChromeForTests)' == '' and
                                         ('$(ContinuousIntegrationBuild)' != 'true' or Exists('/.dockerenv')) and
                                         '$(Scenario)' == 'WasmTestOnBrowser'">true</InstallChromeForTests>
+    <InstallV8ForTests Condition="'$(InstallV8ForTests)' == '' and
+                                        ('$(ContinuousIntegrationBuild)' != 'true' or Exists('/.dockerenv')) and
+                                        ('$(Scenario)' == 'normal' or '$(Scenario)' == '')">true</InstallV8ForTests>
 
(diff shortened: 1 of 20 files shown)


eng/testing/tests.browser.targets
```diff
@@ -27,6 +27,9 @@
     <InstallChromeForTests Condition="'$(InstallChromeForTests)' == '' and
                                         ('$(ContinuousIntegrationBuild)' != 'true' or Exists('/.dockerenv')) and
                                         '$(Scenario)' == 'WasmTestOnBrowser'">true</InstallChromeForTests>
+    <InstallV8ForTests Condition="'$(InstallV8ForTests)' == '' and
+                                        ('$(ContinuousIntegrationBuild)' != 'true' or Exists('/.dockerenv')) and
```
kg, line 31: this dockerenv check bothers me, but i guess it's fine since we were already doing it? it feels too magical

radical: That's the common way I found of detecting that you are running in a codespace.

kg: Maybe add a comment explaining that

src/libraries/sendtohelix-wasm.targets
```diff
@@ -101,6 +102,8 @@
 
     <HelixPreCommand Condition="'$(NeedsToRunOnBrowser)' == 'true' and '$(DebuggerHost)' == 'chrome'" Include="export PATH=$HELIX_CORRELATION_PAYLOAD/$(ChromeDriverDirName):$PATH" />
     <HelixPreCommand Condition="'$(NeedsToRunOnBrowser)' == 'true' and '$(DebuggerHost)' == 'chrome'" Include="export PATH=$HELIX_CORRELATION_PAYLOAD/$(ChromeDirName):$PATH" />
+
+    <HelixPreCommand Condition="'$(NeedsToRunOnV8)' == 'true'" Include="export V8_PATH_FOR_TESTS=$HELIX_CORRELATION_PAYLOAD/$(V8DirName)/$(V8BinaryName)" />
```
kg, line 106: do we need quotes for this or the set= further down?

radical: Not really. Helix paths are all "space" free.

src/mono/wasm/build/WasmApp.targets
```diff
@@ -475,7 +475,7 @@
     </PropertyGroup>
 
     <Error Condition="'$(WasmMainAssemblyFileName)' == ''" Text="%24(WasmMainAssemblyFileName) property needs to be set for generating $(WasmRunV8ScriptPath)." />
-    
+
```
kg, line 478: whitespace damage (not important)

src/mono/wasm/Wasm.Build.Tests/Wasm.Build.Tests.csproj
```diff
@@ -16,6 +16,7 @@
     <SkipWorkloadsTestingTargetsImport>false</SkipWorkloadsTestingTargetsImport>
     <InstallWorkloadForTesting>true</InstallWorkloadForTesting>
     <InstallChromeForTests Condition="'$(InstallChromeForTests)' == '' and ('$(ContinuousIntegrationBuild)' == 'true' or Exists('/.dockerenv'))">true</InstallChromeForTests>
+    <InstallV8ForTests Condition="'$(InstallV8ForTests)' == '' and ('$(ContinuousIntegrationBuild)' == 'true' or Exists('/.dockerenv'))">true</InstallV8ForTests>
```
kg, line 19: could add a comment explaining dockerenv here too, not important though

kg (approved):

src/mono/wasm/Wasm.Build.Tests/Wasm.Build.Tests.csproj
radical: I was trying to be lazy lol. It is used in a bunch of places, i'll add there too :D

src/mono/wasm/build/WasmApp.targets
radical: Removed trailing spaces.

PR #91637, opened 2023-09-05
SPMI: Optimize collection download, increase diffs timeout

jakobbotsch:
With the latest collections we are frequently hitting timeouts during
linux tpdiff runs. This PR applies two changes to improve the
situation:
- Optimize the download/extraction of collections by extracting directly
  to the target location. The Helix machines actually seem to spend many
  times longer to extract and copy the collections than they do
  downloading them.
- Increase the timeout of superpmi-diffs by one hour.


jakobbotsch:

This passed CI in https://github.com/dotnet/runtime/pull/91637/checks?check_run_id=16522220574

BruceForstall (approved):
LGTM

PR #91176, opened 2023-08-27
BigInteger constructor perf

Rob-Hague:
Use BinaryPrimitives/MemoryMarshal in BigInteger constructor

| Method |        Job | ByteCount | IsBigEndian |      Mean |    Error |   StdDev | Ratio | Code Size |
|------- |----------- |---------- |------------ |----------:|---------:|---------:|------:|----------:|
|   **Ctor** | **pr** |         **5** |       **False** |  **13.28 ns** | **0.073 ns** | **0.065 ns** |  **1.05** |
|   Ctor | main |         5 |       False |  12.60 ns | 0.067 ns | 0.063 ns |  1.00 |
|        |            |           |             |           |          |          |       |
|   **Ctor** | **pr** |         **5** |        **True** |  **16.38 ns** | **0.036 ns** | **0.032 ns** |  **1.32** |
|   Ctor | main |         5 |        True |  12.37 ns | 0.016 ns | 0.013 ns |  1.00 |
|        |            |           |             |           |          |          |       |
|   **Ctor** | **pr** |        **16** |       **False** |  **11.78 ns** | **0.019 ns** | **0.018 ns** |  **0.66** |
|   Ctor | main |        16 |       False |  17.72 ns | 0.113 ns | 0.094 ns |  1.00 |
|        |            |           |             |           |          |          |       |
|   **Ctor** | **pr** |        **16** |        **True** |  **15.08 ns** | **0.059 ns** | **0.055 ns** |  **0.84** |
|   Ctor | main |        16 |        True |  17.90 ns | 0.015 ns | 0.012 ns |  1.00 |
|        |            |           |             |           |          |          |       |
|   **Ctor** | **pr** |        **64** |       **False** |  **13.78 ns** | **0.022 ns** | **0.019 ns** |  **0.31** |
|   Ctor | main |        64 |       False |  44.43 ns | 0.040 ns | 0.036 ns |  1.00 |
|        |            |           |             |           |          |          |       |
|   **Ctor** | **pr** |        **64** |        **True** |  **17.45 ns** | **0.027 ns** | **0.021 ns** |  **0.39** |
|   Ctor | main |        64 |        True |  44.82 ns | 0.040 ns | 0.033 ns |  1.00 |
|        |            |           |             |           |          |          |       |
|   **Ctor** | **pr** |       **256** |       **False** |  **22.16 ns** | **0.020 ns** | **0.017 ns** |  **0.15** |
[...]

PR #91639, opened 2023-09-05
Use `Unsafe.WriteUnaligned` in `BitConverter.GetBytes`

xtqqczze:
[sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAmARgFgAoQgBgAJDSA6AJQFcA7DbAWxiYDCEHgAc8MKAGUJAN2xgYAZwDcVWg2bsuvfgEkuEiCOlQ5ClVTUBmBsToC6Abyp1XDG4yR1gATwwwAbQBdOgBxGAwAIT8lAH0AEWwAMySACiTcCABDDDoZLNw2GABKFzdnSjcq7xjgmv9FOgBeOg4YAHd6wMVsAC8YCDSM7IxioNVK6tcAVQ5FLKT+AHUobH9ZguwAczaAE1TYJLoAWRgeaB9jrKhFAAsCpnCMAEEoKCyfeJyslhhF2A4ClSvgaxTQeQKRWKEymrkIAHYuhZJq4AL5lOEeUheEGBEJPaINWJPV7vT7fX7/GCAmDpTI5CGFEoYpwsqq4uq4xotNqdDk9fqDOkjMYw2GzeaLJjPRQAHlx4OGOQAfAc/iczhcrjd7rhHhFSR8vhgfn8JNSgVzisVmoyimKpgikQ66KjqizCFicbV8RFCXFIllFLSlbl8kzSijWVH2T6kbbeV0AgKBkN6aNxmy3BKFvwZfKYor06rDkiAjQgjaWuH7Vm4YiuS70ZRUUA==)

Added `AggressiveInlining` to prevent regressions from from failure to inline.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/BitConverter.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/BitConverter.cs
@@ -29,6 +29,7 @@ namespace System
         /// </summary>
         /// <param name="value">A Boolean value.</param>
         /// <returns>A byte array with length 1.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static byte[] GetBytes(bool value)
         {
             byte[] r = new byte[1];
--- a/src/libraries/System.Private.CoreLib/src/System/BitConverter.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/BitConverter.cs
@@ -56,6 +56,7 @@ namespace System
         /// </summary>
         /// <param name="value">A Char value.</param>
         /// <returns>An array of bytes with length 2.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static byte[] GetBytes(char value)
         {
             byte[] bytes = new byte[sizeof(char)];
@@ -83,6 +84,7 @@ namespace System
         /// </summary>
         /// <param name="value">The number to convert.</param>
         /// <returns>An array of bytes with length 2.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static byte[] GetBytes(short value)
         {
             byte[] bytes = new byte[sizeof(short)];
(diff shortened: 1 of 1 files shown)


xtqqczze:
Similar changes should also be made in #91299.

xtqqczze:
@MihuBot

xtqqczze:
`jit-diffs` show regressions due to failure to inline and no improvements.

stephentoub:
I appreciate the effort, but what is the concrete benefit of this change? A total of 20 bytes of IL and 0 impact on the resulting asm?  If that's all this buys, I don't think it's worth the churn.

xtqqczze:
> I appreciate the effort, but what is the concrete benefit of this change? A total of 20 bytes of IL and 0 impact on the resulting asm? If that's all this buys, I don't think it's worth the churn.

The motivation was to avoid depending an implementation alignment assumption, i.e. the array data is 8-byte aligned.

I was hoping the reduced IL size might bring a benefit from improved inlining, but `jit-diffs` show the opposite, see the description for an example.

xtqqczze:
@EgorBo I was surprised that the misses inlining a 19-byte method consisting of an allocation and two intrinsics.

huoyaoyuan:
> i.e. the array data is 8-byte aligned.

https://github.com/dotnet/runtime/pull/91176#discussion_r1307186945
There can be some places depending on the fact. Is this clarified in any spec?

MichalPetryka:
> Is this clarified in any spec?

AFAIR it's just an implementation detail of CoreCLR.

jkotas:
> the array data is 8-byte aligned.

The array data is not 8-byte aligned on 32-bit platforms, and there is a risk of alignment fault on arm. We seem to be lucky that the JIT does not generate code that would hit the alignment fault - https://github.com/dotnet/runtime/issues/20729 has some discussion about it. It is a fragile assumption - it may be possible to trick the JIT into generating code that would hit the alignment fault on Arm32.

MichalPetryka:
> I was hoping the reduced IL size

Does adding a call reduce the size here? `ref bytes[0]` should be smaller and should perform the same here.

jkotas:
> Does adding a call reduce the size here? 

It does reduce IL size by one byte. It is not something we care about.

xtqqczze:
@MihuBot

xtqqczze:
@MihuBot

xtqqczze:
@MihuBot

xtqqczze:
The `System.Data.Common.ObjectStorage:Set` diffs may improve as a result of  #91945.

PR #91643, opened 2023-09-05
Separate running PAL tests out to their own jobs/steps.

jkoritzinsky:
This helps us separate the different test legs in our product that are different. Instead of including the PAL tests in the cross-job copying and the significant logic around generating work items for the src/tests test tree in Helix, create a simple dedicated job for building and running the PAL tests on each platform.


eng/pipelines/runtime.yml
```diff
@@ -62,1443 +62,21 @@ extends:
...
-      #
-      - template: /eng/pipelines/common/platform-matrix.yml
-        parameters:
-          jobTemplate: /eng/pipelines/common/global-build-job.yml
-          buildConfig: checked
-          platforms:
-          - gcc_linux_x64
-          jobParameters:
-            testGroup: innerloop
-            nameSuffix: Native_GCC
```
janvorli, line 106: Why are so many legs  including the gcc one being removed?

jkoritzinsky: This is temporary to make testing easier. I'll revert this before I make the PR non-draft

jkoritzinsky: I've reverted this portion of the change and this PR is now ready for review.

janvorli (approved):
It looks good, but I'd feel more confident if someone more knowledgeble in the CI yml scripts took a look too.

trylek (approved):
Looks great to me, thanks Jeremy!

trylek:
Having said that, I see that helixpublishwitharcade.proj still refers to CreatePALTestArchive at line 135, I would guess that's why multiple legs of your PR failed, that will need addressing.

jkoritzinsky:
The new jobs all passed, so merging this in.

PR #91646, opened 2023-09-05
Move CoreCLR community platforms to build with the global build template

jkoritzinsky:
This makes our community-supported platform testing story consistent (we do all testing of community-supported targets in the runtime-community pipeline and none in the runtime pipeline) and allows us to remove more custom code from the CoreCLR build-job.yml file.

Alternatively, we could keep running these legs in the runtime.yml file and just do the move to the global-build-job template.


jkotas:
runtime-community is optional targets for targets that are too expensive and/or too flaky to run unconditionally. It includes tests.

We have build for linux-x86 and linux-armel in the unconditional pipeline to catch the trivial linux-x86 and linux-armel build breaks. The idea is that the build-only leg for linux-x86 and linux-armel is cheap enough and reliable enough to make it worth run unconditionally - it saves us from having to having to deal with the fix build break PRs.

jkoritzinsky:
Sounds good 👍
I'll move these jobs back into the runtime pipeline.

jkoritzinsky:
Jobs have been moved back into the runtime pipeline and fixed to pass.

trylek (approved):
LGTM, thanks Jeremy!

BruceForstall (approved):

clamp03 (approved):
Thank you

wscho77 (approved):
Thank you 👍

jkoritzinsky:
Failures are all known.

Issue #91648, opened 2023-09-05
Test failure: JIT/Directed/PREFIX/volatile/1/Box_Unbox_volatile/Box_Unbox_volatile.dll


BruceForstall:
pipeline: runtime-coreclr outerloop
coreclr osx arm64 Checked no_tiered_compilation @ OSX.1200.ARM64.Open
Directed_2 Work Item

https://dev.azure.com/dnceng-public/public/_build/results?buildId=392066&view=ms.vss-test-web.build-test-results-tab&runId=8427280&paneView=dotnet-dnceng.dnceng-build-release-tasks.helix-test-information-tab&resultId=120605

```
12:57:36.545 Running test: JIT/Directed/PREFIX/volatile/1/Box_Unbox_volatile/Box_Unbox_volatile.dll
Fatal error. System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception.
   at Box_Unbox.main()
   at Program.<<Main>$>g__TestExecutor31|0_32(System.IO.StreamWriter, System.IO.StreamWriter, <>c__DisplayClass0_0 ByRef)
   at Program.<Main>$(System.String[])
[createdump] Invalid process id: task_for_pid(80065) FAILED (os/kern) failure (5)
[createdump] This failure may be because createdump or the application is not properly signed and entitled.
[createdump] Target process is alive
[createdump] Failure took 1ms
waitpid() returned successfully (wstatus 0000ff00) WEXITSTATUS ff WTERMSIG 0
App Exit Code: 22
Expected: 100
Actual: 22
END EXECUTION - FAILED
```

Same failure in `coreclr osx arm64 Checked @ OSX.1200.ARM64.Open.`

Related? https://github.com/dotnet/runtime/issues/91647 @EgorBo

Issue #91647, opened 2023-09-05
Test failure: JIT/Directed/coverage/oldtests/stfldstatic2_r/stfldstatic2_r.dll


BruceForstall:
pipeline: runtime-coreclr outerloop
coreclr osx arm64 Checked no_tiered_compilation @ OSX.1200.ARM64.Open
Directed_1 Work Item

https://dev.azure.com/dnceng-public/public/_build/results?buildId=392066&view=ms.vss-test-web.build-test-results-tab&runId=8427280&paneView=debug&resultId=120604

(failure started with runtime-coreclr outerloop 20230831.3)

```
12:57:21.890 Running test: JIT/Directed/coverage/oldtests/stfldstatic2_r/stfldstatic2_r.dll
Fatal error. System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception.
   at stfldstatic2.Main()
   at Program.<<Main>$>g__TestExecutor171|0_172(System.IO.StreamWriter, System.IO.StreamWriter, <>c__DisplayClass0_0 ByRef)
   at Program.<Main>$(System.String[])
[createdump] Invalid process id: task_for_pid(58591) FAILED (os/kern) failure (5)
[createdump] This failure may be because createdump or the application is not properly signed and entitled.
[createdump] Target process is alive
[createdump] Failure took 0ms
waitpid() returned successfully (wstatus 0000ff00) WEXITSTATUS ff WTERMSIG 0
App Exit Code: 22
Expected: 100
Actual: 22
END EXECUTION - FAILED
```

Maybe related to https://github.com/dotnet/runtime/pull/89681 @EgorBo ?

EgorBo:
Yes, seems like ldapur is the culprit, here is the asm diff: 
![image](https://github.com/dotnet/runtime/assets/523221/7249454c-6664-4884-a07e-d88c194fe1f6)

it looks correct to me so most likely the problem is on the encoding side, I'll check and file a fix shortly

EgorBo:
```
1D404011          ldapur  s17, [x0, #0x04]
```

PR #91650, opened 2023-09-05
Disable stfldstatic2_r, Box_Unbox_volatile on osx-arm64

BruceForstall:
Tracking: #91647, #91648

### DIFF
--- a/src/tests/issues.targets
+++ b/src/tests/issues.targets
@@ -443,6 +443,12 @@
 
     <!-- OSX arm64 on CoreCLR Runtime -->
     <ItemGroup Condition="'$(XunitTestBinBase)' != '' and '$(TargetsOSX)' == 'true' and '$(TargetArchitecture)' == 'arm64' and '$(RuntimeFlavor)' == 'coreclr' ">
+        <ExcludeList Include="$(XunitTestBinBase)/JIT/Directed/PREFIX/volatile/1/Box_Unbox_volatile/*">
+            <Issue>https://github.com/dotnet/runtime/issues/91648</Issue>
+        </ExcludeList>
+        <ExcludeList Include="$(XunitTestBinBase)/JIT/Directed/coverage/oldtests/stfldstatic2_r/*">
+            <Issue>https://github.com/dotnet/runtime/issues/91647</Issue>
+        </ExcludeList>
     </ItemGroup>
 
     <!-- Tests that need to be triaged for vararg usage as that is not supported -->


BruceForstall:
@EgorBo @dotnet/jit-contrib PTAL

EgorBo (approved):

trylek (approved):
LGTM, thanks Bruce!

BruceForstall:
@trylek @markples @jkoritzinsky @dotnet/runtime-infrastructure One of the big problems with our merged test system is that if one test fails and brings down the run (e.g., crashes, or hits a JIT/VM assert), all the tests after it don't run at all. This PR is an example of how problematic that is. There were failing tests. I submitted a PR to disable them and make outerloop (and many other pipelines) clean. But then I get a failure on one of the tests that previously didn't run.

I notice in previous runs the text "30/211 tests run". With this PR, it's "80/211 tests run". There's no telling how many times I would need to (1) add new test to exclude, (2) rerun outerloop testing, (3) see new failure, (4) repeat; before the pipeline is clean again. I don't have access to an osx-arm64 box to do this more quickly (nor should I need to have such access).

Even if these tests were marked as "Known Issues", it would mean losing coverage on many un-run tests.

Is there a plan to fix this problem?

BruceForstall:
https://github.com/dotnet/runtime/pull/91652 should fix the bug, so closing.

BruceForstall:
> @BruceForstall I believe we have a command or attribute that can run a specific test out of process. Can we modify such tests to run out-of-proc temporarily?

That doesn't solve the problem, since I don't know which tests to mark (it's the same problem as knowing which tests to disable).

One option would be to (1) disable merged tests for all merged groups that had such a failure, (2) re-run the pipeline, (3) disable all tests that failed and re-enable merged groups.

This is not a viable option; I don't think we have the ability right now to "un-merge".

The real solution is certainly to ensure that all tests that should run, do run (and have their results reported).

agocke:
Isn't that equivalent to saying that every test needs to run in its own process?

BruceForstall:
> Isn't that equivalent to saying that every test needs to run in its own process?

No.

One option that has been discussed: a "watcher process". If a merged process fails, the "watcher" notices where it failed, and restarts it, starting after the failing test.

fwiw, our "PMI" JIT testing tool has done this for many years.

trylek:
The "watcher process" is what I also remember from the last time we discussed this. At some point it seemed to be part of the "merge on red" plan but IIRC I was later told it just didn't fit in its time schedule. I still believe it's the most straightforward way to fix this issue.

Issue #90976, opened 2023-08-23
Config generator emits uncompilable code for certain type conflicts


stephentoub:
The implementation does name mangling, but those mangled names can conflict with other names, e.g.
```C#
using Microsoft.Extensions.Configuration;

IConfiguration configuration = new ConfigurationBuilder().Build();

var c1 = new Cint();
var c2 = new C<int>();
configuration.Bind(c1);
configuration.Bind(c2);

internal class C<T>
{
    public int Value { get; set; }
}

internal class Cint
{
    public int Value { get; set; }
}
```

layomia:
I'll resolve with a shared abstraction doing whatever JSON does here.

layomia:
Inlining this scenario from https://github.com/dotnet/runtime/issues/90969:

---

e.g.
```C#
namespace S1
{
    class C1
    {
        public void M(IConfiguration configuration)
        {
            var c = new C1();
            configuration.Bind(c);
        }
    }
}
namespace S2
{
    class C1
    {
        public void M(IConfiguration configuration)
        {
            var c = new C1();
            configuration.Bind(c);
        }
    }
}
```
results in lots of errors like:
```
error CS0111: Type 'BindingExtensions' already defines a member called 'Bind_C1' with the same parameter types
```

ericstj:
Fixed in #91657

PR #91657, opened 2023-09-06
Fix binder gen compile issues due to inaccessible members and identifier name clashes

### LINKED ISSUE #90909: Config binder generator generates uncompilable code when a private nested type is used
This compiles fine:
```C#

internal class Program
{
    private static void Main() { }

    private static C Foo(IConfiguration configuration)
    {
        C c = new();
        configuration.Bind(c);
        return c;
    }

    internal sealed class C
    {
        public int Value { get; set; }
    }
}
```
but changing:
```C#
internal sealed class C
```
to
```C#
private sealed class C
```
produces an error about `C` being inaccessible, as the generator emits code into another class that tries to access that property.

layomia:
Fixes https://github.com/dotnet/runtime/issues/90909 and https://github.com/dotnet/runtime/issues/90976. RC-2 candidate.

cc @ericstj.

### DIFF
--- a/src/libraries/Common/src/SourceGenerators/TypeModelHelper.cs
+++ b/src/libraries/Common/src/SourceGenerators/TypeModelHelper.cs
@@ -1,7 +1,6 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System;
 using System.Text;
 using Microsoft.CodeAnalysis;
 using System.Collections.Generic;
@@ -74,7 +73,7 @@ namespace SourceGenerators
         /// </summary>
         public static string ToMinimalDisplayString(this ITypeSymbol type) => type.ToDisplayString(s_minimalDisplayFormat);
 
-        private static List<ITypeSymbol>? GetAllTypeArgumentsInScope(this INamedTypeSymbol type)
+        public static List<ITypeSymbol>? GetAllTypeArgumentsInScope(this INamedTypeSymbol type)
         {
             if (!type.IsGenericType)
             {
--- a/src/libraries/Common/src/SourceGenerators/TypeModelHelper.cs
+++ b/src/libraries/Common/src/SourceGenerators/TypeModelHelper.cs
@@ -9,7 +9,7 @@ using System.Diagnostics;
 
 namespace SourceGenerators
 {
-    internal static class TypeModelHelpers
+    internal static class TypeModelHelper
     {
         private static readonly SymbolDisplayFormat s_minimalDisplayFormat = new SymbolDisplayFormat(
             globalNamespaceStyle: SymbolDisplayGlobalNamespaceStyle.Omitted,
--- a/src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
+++ b/src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
@@ -5,6 +5,7 @@ using System;
 using System.Text;
 using Microsoft.CodeAnalysis;
 using System.Collections.Generic;
+using System.Diagnostics;
 
 namespace SourceGenerators
 {
@@ -25,16 +26,37 @@ namespace SourceGenerators
                 return ToIdentifierCompatibleSubstring(arrayType.ElementType) + suffix;
             }
 
-            string name = GetName(type);
+            StringBuilder? sb = null;
+            string? symbolName = null;
+
+            if (useUniqueName)
+            {
+                string uniqueDisplayString = type.ToMinimalDisplayString();
+                PopulateIdentifierCompatibleSubstring(sb = new(), uniqueDisplayString);
+            }
+            else
+            {
+                symbolName = type.Name;
+            }
+
+#if DEBUG
+            bool usingUniqueName = (symbolName is null || sb is not null) && useUniqueName;
+            bool usingSymbolName = (symbolName is not null && sb is null) && !useUniqueName;
+            bool configuredNameCorrectly = (usingUniqueName && !usingSymbolName) || (!usingUniqueName && usingSymbolName);
+            Debug.Assert(configuredNameCorrectly);
+#endif
 
             if (type is not INamedTypeSymbol namedType || !namedType.IsGenericType)
             {
-                return name;
+                return symbolName ?? sb!.ToString();
             }
 
-            StringBuilder sb = new();
+            if (sb is null)
+            {
+                (sb = new()).Append(symbolName!);
+            }
 
-            sb.Append(name);
+            Debug.Assert(sb.Length > 0);
 
             foreach (ITypeSymbol genericArg in namedType.GetAllTypeArgumentsInScope())
             {
@@ -42,10 +64,6 @@ namespace SourceGenerators
             }
 
             return sb.ToString();
-
-            string GetName(ITypeSymbol type) => useUniqueName
-                ? ToIdentifierCompatibleSubstring(type.ToMinimalDisplayString())
-                : type.Name;
         }
 
         /// <summary>
@@ -76,27 +94,23 @@ namespace SourceGenerators
             }
         }
 
-        private static string ToIdentifierCompatibleSubstring(string input)
+        private static void PopulateIdentifierCompatibleSubstring(StringBuilder sb, string input)
         {
-            StringBuilder sb = new();
-
             foreach (char c in input)
             {
                 if (c is '[')
                 {
-                    sb.Append("Arr");
+                    sb.Append("Array");
                 }
-                else if (c is ']')
+                else if (c is ',')
                 {
-                    sb.Append("ay");
+                    sb.Append("Comma");
                 }
-                else if (c is not (',' or ' ' or '.' or '<' or '>' or '_'))
+                else if (char.IsLetterOrDigit(c))
                 {
                     sb.Append(c);
                 }
             }
-
-            return sb.ToString();
         }
     }
 }
--- a/src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
+++ b/src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
@@ -17,13 +17,13 @@ namespace SourceGenerators
             genericsOptions: SymbolDisplayGenericsOptions.IncludeTypeParameters,
             miscellaneousOptions: SymbolDisplayMiscellaneousOptions.UseSpecialTypes);
 
-        public static string ToIdentifierCompatibleSubstring(this ITypeSymbol type, bool useUniqueName = false)
+        public static string ToIdentifierCompatibleSubstring(this ITypeSymbol type, bool useUniqueName)
         {
             if (type is IArrayTypeSymbol arrayType)
             {
                 int rank = arrayType.Rank;
                 string suffix = rank == 1 ? "Array" : $"Array{rank}D"; // Array, Array2D, Array3D, ...
-                return ToIdentifierCompatibleSubstring(arrayType.ElementType) + suffix;
+                return ToIdentifierCompatibleSubstring(arrayType.ElementType, useUniqueName) + suffix;
             }
 
             StringBuilder? sb = null;
@@ -62,7 +62,7 @@ namespace SourceGenerators
             {
                 foreach (ITypeSymbol genericArg in typeArgsInScope)
                 {
-                    sb.Append(ToIdentifierCompatibleSubstring(genericArg));
+                    sb.Append(ToIdentifierCompatibleSubstring(genericArg, useUniqueName));
                 }
             }
 
@@ -94,7 +94,7 @@ namespace SourceGenerators
 
                 if (!current.TypeArguments.IsEmpty)
                 {
-                    (args ?? new()).AddRange(current.TypeArguments);
+                    (args ??= new()).AddRange(current.TypeArguments);
                 }
             }
         }
--- a/src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
+++ b/src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
@@ -1,0 +1,102 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Text;
+using Microsoft.CodeAnalysis;
+using System.Collections.Generic;
+
+namespace SourceGenerators
+{
+    internal static class TypeModelHelpers
+    {
+        private static readonly SymbolDisplayFormat s_minimalDisplayFormat = new SymbolDisplayFormat(
+            globalNamespaceStyle: SymbolDisplayGlobalNamespaceStyle.Omitted,
+            typeQualificationStyle: SymbolDisplayTypeQualificationStyle.NameAndContainingTypes,
+            genericsOptions: SymbolDisplayGenericsOptions.IncludeTypeParameters,
+            miscellaneousOptions: SymbolDisplayMiscellaneousOptions.UseSpecialTypes);
+
+        public static string ToIdentifierCompatibleSubstring(this ITypeSymbol type, bool useUniqueName = false)
+        {
+            if (type is IArrayTypeSymbol arrayType)
+            {
+                int rank = arrayType.Rank;
+                string suffix = rank == 1 ? "Array" : $"Array{rank}D"; // Array, Array2D, Array3D, ...
+                return ToIdentifierCompatibleSubstring(arrayType.ElementType) + suffix;
+            }
+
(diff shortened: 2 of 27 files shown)


src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
```diff
@@ -0,0 +1,102 @@
...
+            foreach (ITypeSymbol genericArg in namedType.GetAllTypeArgumentsInScope())
+            {
+                sb.Append(ToIdentifierCompatibleSubstring(genericArg));
+            }
+
+            return sb.ToString();
+
+            string GetName(ITypeSymbol type) => useUniqueName
+                ? ToIdentifierCompatibleSubstring(type.ToMinimalDisplayString())
+                : type.Name;
```
tarekgh, line 48: Why we prefer this style? this is called once in the code and can be in lined in the code . 

string name = useUniqueName 
                ? ToIdentifierCompatibleSubstring(type.ToMinimalDisplayString())
                : type.Name;

just wondering :-)

src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        /// <summary>
+        /// Type name, prefixed with containing type names if it is nested (e.g. ContainingType.NestedType).
+        /// </summary>
+        /// <returns></returns>
```
tarekgh, line 54: >  /// <returns></returns> [](http://example.com/codeflow?start=6&length=25)

do you want to fill this?

src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
```diff
@@ -0,0 +1,102 @@
...
+        public static string ToMinimalDisplayString(this ITypeSymbol type) => type.ToDisplayString(s_minimalDisplayFormat);
+
+        private static ITypeSymbol[] GetAllTypeArgumentsInScope(this INamedTypeSymbol type)
```
tarekgh, line 57: >ITypeSymbol[] [](http://example.com/codeflow?start=23&length=13)

what can't returning List<ITypeSymbol> instead of array here? avoid the array allocation

src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            foreach (ITypeSymbol genericArg in namedType.GetAllTypeArgumentsInScope())
+            {
+                sb.Append(ToIdentifierCompatibleSubstring(genericArg));
+            }
+
+            return sb.ToString();
+
+            string GetName(ITypeSymbol type) => useUniqueName
+                ? ToIdentifierCompatibleSubstring(type.ToMinimalDisplayString())
```
tarekgh, line 47: >ToIdentifierCompatibleSubstring [](http://example.com/codeflow?start=18&length=31)

You can pass the created `sb` here instead of creating another StringBuilder object inside this method.

src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+        }
+
+        private static string ToIdentifierCompatibleSubstring(string input)
+        {
+            StringBuilder sb = new();
+
+            foreach (char c in input)
+            {
+                if (c is '[')
```
tarekgh, line 85: >if (c is '[') [](http://example.com/codeflow?start=16&length=13)

Is it possible you can get characters between `[` and `]`?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -70,6 +70,7 @@ private static bool IsValidRootConfigType(ITypeSymbol? type)
             {
                 if (type is null ||
                     type.SpecialType is SpecialType.System_Object or SpecialType.System_Void ||
+                    !IsAccessibleFromGenBinders(type) ||
```
tarekgh, line 73: >!IsAccessibleFromGenBinders(type) || [](http://example.com/codeflow?start=20&length=36)

Is this mean we'll have a difference between the source gen and the runtime in such cases?

layomia: In a sense... we won't generate a replacement interceptor for the invocation, meaning that it will always fallback to reflection. AOT users will get a warning and the fix would be to adjust the visibility.

src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
layomia: Perhaps multi-dim arrays. I think the current code handles it okay; I'll add a test. We also have existing tests that show we skip multidim arrays as unsupported, consistent with reflection. So we shouldn't hit this codepath for multi-dim arrays.

I'll adjust the code as necessary.

src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
layomia: Will switch thanks. There was a reason I didn't do that in an earlier iteration but it no longer holds.

src/libraries/Common/src/SourceGenerators/TypeModelHelpers.cs
layomia: Was just porting the code from JSON. Makes sense to optimize it here.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: Will the users get anything to help them tell them what to do? I mean how the users know the source gen didn't handle it because of the class visibility? Would it make sense to produce info diagnostics for this one? It doesn't have to be part of this PR.

layomia: Yup, should have mentioned this -- we produce a diagnostic for this. Hence, we need these suppressions:

https://github.com/dotnet/runtime/pull/91657/files#diff-81e6cde105e57e3f3b3a0107c173264b96aee943a0faa6b8f1bd72d7a57a43a3R333-R343

tarekgh (approved):

tarekgh (approved):

layomia:
CI failures unrelated; referenced above.

src/libraries/System.Text.Json/gen/JsonSourceGenerator.Parser.cs
```diff
@@ -589,17 +590,19 @@ private TypeGenerationSpec ParseTypeGenerationSpec(in TypeToGenerate typeToGener
...
-                string typeInfoPropertyName = typeToGenerate.TypeInfoPropertyName ?? GetTypeInfoPropertyName(type);
+                string typeInfoPropertyName = typeToGenerate.TypeInfoPropertyName ??
+                    type.ToIdentifierCompatibleSubstring(useUniqueName: false /* We leave identifier name conflicts to users, as codified below. */);
 
                 if (classType is ClassType.TypeUnsupportedBySourceGen)
                 {
                     ReportDiagnostic(DiagnosticDescriptors.TypeNotSupported, typeToGenerate.AttributeLocation ?? typeToGenerate.Location, type.ToDisplayString());
                 }
 
+                // Workaround for https://github.com/dotnet/roslyn/issues/54185 by keeping track of the file names we've used.
```
eiriktsarpalis, line 601: The issue as described in https://github.com/dotnet/roslyn/issues/54185 hasn't been fixed in Roslyn 3.11 (and won't be fixed). We should keep this comment (and possibly with a TODO explaining that we should remove it once Roslyn 3.11 support is removed).

src/libraries/System.Text.Json/gen/JsonSourceGenerator.Parser.cs
```diff
@@ -1589,34 +1592,6 @@ private static string DeterminePropertyNameFieldName(string effectiveJsonPropert
                 return null;
             }
 
-            private static string GetTypeInfoPropertyName(ITypeSymbol type)
```
eiriktsarpalis, line 1592: Changing the identifier generation algorithm can be a potential source breaking change, since user code referencing source generated members from earlier SG versions could fail to compile.

Even if the new logic is equivalent to the old one, this will get backported to RC2 and it's not clear to me why we should take the risk this late.

src/libraries/Common/src/SourceGenerators/TypeModelHelper.cs
```diff
@@ -0,0 +1,120 @@
...
+{
+    internal static class TypeModelHelper
+    {
+        private static readonly SymbolDisplayFormat s_minimalDisplayFormat = new SymbolDisplayFormat(
+            globalNamespaceStyle: SymbolDisplayGlobalNamespaceStyle.Omitted,
+            typeQualificationStyle: SymbolDisplayTypeQualificationStyle.NameAndContainingTypes,
+            genericsOptions: SymbolDisplayGenericsOptions.IncludeTypeParameters,
+            miscellaneousOptions: SymbolDisplayMiscellaneousOptions.UseSpecialTypes);
+
+        public static string ToIdentifierCompatibleSubstring(this ITypeSymbol type, bool useUniqueName)
```
eiriktsarpalis, line 19: I don't think this algorithm should live in shared code. It should be specific to each application since changing the algorithm in the future is a potential breaking change, and what we have here isn't perfect.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -78,6 +80,27 @@ private static bool IsValidRootConfigType(ITypeSymbol? type)
                 }
 
                 return true;
+
+                static bool IsAccessibleFromGenBinders(ITypeSymbol type)
```
eiriktsarpalis, line 84: Couldn't you use `IsSymbolAccessibleWithin` instead?

https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.compilation.issymbolaccessiblewithin?view=roslyn-dotnet-4.6.0

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -78,6 +80,27 @@ private static bool IsValidRootConfigType(ITypeSymbol? type)
...
+                        foreach (ITypeSymbol genericArg in typeArgsInScope)
+                        {
+                            if (!IsAccessibleFromGenBinders(genericArg))
+                            {
+                                return false;
+                            }
+                        }
+                    }
+
+                    return type.DeclaredAccessibility is Accessibility.Public or Accessibility.Internal or Accessibility.Friend;
```
eiriktsarpalis, line 102: What about `internal` members that are in referenced assemblies?

src/libraries/System.Text.Json/gen/JsonSourceGenerator.Parser.cs
```diff
@@ -589,17 +590,19 @@ private TypeGenerationSpec ParseTypeGenerationSpec(in TypeToGenerate typeToGener
                 }
 
                 var typeRef = new TypeRef(type);
-                string typeInfoPropertyName = typeToGenerate.TypeInfoPropertyName ?? GetTypeInfoPropertyName(type);
+                string typeInfoPropertyName = typeToGenerate.TypeInfoPropertyName ??
```
eiriktsarpalis, line 593: I'm not comfortable making cross-cutting changes in RC2 backports. To minimize risk, we should be making targeted changes even if it results in temporary code duplication.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
layomia: I don't have a "within" symbol. The symbol would be the generated binder extension class which doesn't exist yet.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
layomia: We shouldn't be seeing those in this routine. This method is for root binding types only. `InternalsVisibleTo` usage is interesting but the current logic handles the scenario.

For transitive types - the generator only inspects public properties. The compiler would have complained if a type had a public member whose declared type is internal. So we don't have to guard against that explicitly

src/libraries/Common/src/SourceGenerators/TypeModelHelper.cs
layomia: Okay I'll revert in this PR. 

---

For 9.0, I still think we should at least have the separate implementations live next to each other so they can be kept in sync. If one generator makes an important improvement, then we would address the other one if needed.

IMO this would also increase discoverability of common concerns when handling type graphs in source-gen (formatting identifiers, visibility checks, formatting names with special syntax e.g. `@int` etc.) Authors would know to check the shared code dir for helper APIs that already exist and they might not need to rewrite.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
eiriktsarpalis: Couldn't you use another type symbol in the same assembly as a proxy for the one that will be generated?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
eiriktsarpalis: My point is we should try to use IsSymbolAccessibleWithin rather than attempt to reimplement accessibility resolution.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
eiriktsarpalis: It should be possible to use the current IAsse blySymbol as the 'within' parameter although I haven't tested that myself.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
layomia: Yeah I fully intended to use it; will do if possible per discussion here https://github.com/dotnet/runtime/pull/91657#discussion_r1318574410.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
layomia: Yes I considered the proxy approach. It didn't feel deterministic on first thought, but yes it would be a correct/better check. I'll also try using the assembly symbol.

layomia: Using a proxy would work in the common case, but we'd need to account for the edge case of the assembly having just one (e.g. a singular, simple target-config POCO with primitive fields). Any handwritten fallback would have a vastly different impl.

Comparing with assembly doesn't work - nested privates aren't visible.

eiriktsarpalis: > Comparing with assembly doesn't work - nested privates aren't visible.

Are any of the types being generated nested within user-defined types? If not, I would expect its visibility to be equivalent to that of the assembly?

layomia: > Comparing with assembly doesn't work - nested privates aren't visible.

I tried it again; the assembly check works. Must have not used the `!` operator when I tried. Thanks.

PR #91658, opened 2023-09-06
Append job attempt number to log artifact names

akoeplinger:
Avoids errors like "Artifact Installer-Logs-coreclr--windows_arm64-Debug already exists for build 1636423." when retrying jobs.

### DIFF
--- a/eng/pipelines/common/global-build-job.yml
+++ b/eng/pipelines/common/global-build-job.yml
@@ -235,8 +235,8 @@ jobs:
         PathtoPublish: '$(Build.SourcesDirectory)/artifacts/log/'
         PublishLocation: Container
         ${{ if notin(parameters.osGroup, 'browser', 'wasi') }}:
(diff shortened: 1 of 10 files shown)


ViktorHofer (approved):
I think we tried that once in the past but for some reason had to revert. But if it works, great.

akoeplinger:
@ViktorHofer yes it seems to work. looking at the artifacts UI I wonder thought whether moving the attempt as a prefix looks better, i.e. this

image

instead of this

image

what do you think?

ViktorHofer:
Can we skip the attempt part for the first run and only add it for attempt number >= 2?

> wonder though whether moving the attempt as a prefix looks better, i.e. this

AttemptX as a prefix looks better IMO.

akoeplinger:
> Can we skip the attempt part for the first run and only add it for attempt number >= 2?

I thought about that but the only way I can see it working is defining a separate variable in `eng/pipelines/common/variables.yml` but I think we hit issues in the past with not every pipeline/template importing that one. Besides that it'd probably be confusing because it'd sort the second-attempt-logs below the first-attempt ones.

> AttemptX as a prefix looks better IMO.

Changed to prefix.

ViktorHofer (approved):

PR #85810, opened 2023-05-05
Get rid of reflection blocked types

MichalStrehovsky:
Fixes #72570.

Still need to delete workarounds that make things public in corelib but maybe this diff is large enough already?

We were gradually getting less and less from reflection blocking:

* We stopped blocking things outside corelib (.NET Native blocked all of BCL; we don't).
* We trim reflection metadata and that allows us to have method bodies without metadata.

With this, we should be able to get type definition-level reflection metadata for any MethodTable there is in the system.

This is a ~30 kB size regression for BasicMinimalApis. It is pretty much a wash for Hello World, because all the BCL cruft to handle reflection blocked types costs as much as the benefit of blocking.

Cc @dotnet/ilc-contrib

PR #91660, opened 2023-09-06
Emit less metadata for not-reflection-visible types

MichalStrehovsky:
In .NET 8 we massively regressed the size of an empty WinForms app. A WinForms app now brings in a big chunk of WPF with it. I traced it down to the `ICommand` interface having a WPF `TypeConverter` and `ValueSerializer` attribute on it: https://github.com/dotnet/runtime/blob/04bd438844482c907062583153a43a9e3b37dbb8/src/libraries/System.ObjectModel/src/System/Windows/Input/ICommand.cs#L13-L16.

An empty app will have a call to a method on `ICommand`, but nothing actually implements `ICommand`. Previously this would mean we generate an unconstructed `MethodTable` for `ICommand`, the unconstructed `MethodTable`s get no reflection metadata, and that's the end of the story.

After #85810 however, the reflection stack can no longer reason about `MethodTable`s that don't have reflection metadata, so we need to generate it. This means we end up with the custom attribute and all the reflection dataflow that comes out of it.

But this metadata is not actually visible in trim safe apps (the only place where reflection could see these method tables in trim safe code is if they're used in a type comparison `x == typeof(Foo)` and we were able to optimize the method table to the unconstructed version because of that). So we can generate less of it and still get away with it.

In this PR I'm adding support for skipping generation of custom attribute metadata for such types. The size of an empty WinForms app goes from 50-something MB to 20-something MB. I think we'll be able to further reduce this number to ~7 MB or less because 12 MB of this are embedded resources that look designer related.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TypeMetadataNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TypeMetadataNode.cs
@@ -24,13 +24,13 @@ namespace ILCompiler.DependencyAnalysis
     internal sealed class TypeMetadataNode : DependencyNodeCore<NodeFactory>
     {
         private readonly MetadataType _type;
-        private readonly bool _isMinimal;
+        private readonly bool _includeCustomAttributes;
 
-        public TypeMetadataNode(MetadataType type, bool isMinimal)
+        public TypeMetadataNode(MetadataType type, bool includeCustomAttributes)
         {
             Debug.Assert(type.IsTypeDefinition);
             _type = type;
-            _isMinimal = isMinimal;
+            _includeCustomAttributes = includeCustomAttributes;
         }
 
         public MetadataType Type => _type;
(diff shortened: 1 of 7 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TypeMetadataNode.cs
```diff
@@ -24,11 +24,13 @@ namespace ILCompiler.DependencyAnalysis
     internal sealed class TypeMetadataNode : DependencyNodeCore<NodeFactory>
     {
         private readonly MetadataType _type;
+        private readonly bool _isMinimal;
```
vitek-karas, line 27: Could you please add a comment describing what it means for type metadata to be "minimal"?
And why is it OK to not include custom attributes in this case for example.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TypeMetadataNode.cs
```diff
@@ -100,7 +110,7 @@ public override IEnumerable<DependencyListEntry> GetStaticDependencies(NodeFacto
         /// Decomposes a constructed type into individual <see cref="TypeMetadataNode"/> units that will be needed to
         /// express the constructed type in metadata.
         /// </summary>
-        public static void GetMetadataDependencies(ref DependencyList dependencies, NodeFactory nodeFactory, TypeDesc type, string reason)
+        public static void GetMetadataDependencies(ref DependencyList dependencies, NodeFactory nodeFactory, TypeDesc type, string reason, bool isFullType = true)
```
vitek-karas, line 113: Personally, I find it confusing the "minimal"/"full" usage. If they're complementary (type is either minimal or full, but nothing else), then I would prefer to use only one of the two terms and negation.
If they're not complementary then I think there should be a comment somewhere describing the differences and where/how it's used.

MichalStrehovsky: Tried to address both.

vitek-karas (approved):
Thanks!

Issue #49904, opened 2021-03-19
[Mono] implement static methods in interfaces support


lambdageek:
Implement support for [static interface methods](https://github.com/dotnet/runtime/pull/49558) (csharp design: https://github.com/dotnet/csharplang/issues/4436)

Work items:
- [ ] unconditionally enable JIT support (can borrow from partydonk prototype)
- [ ] unconditionally enable interpreter support
- [ ] support for the `RuntimeFeature` flag to check for [experimental features](https://github.com/dotnet/designs/pull/185) - will be used to tag BCL additions.

lambdageek:
@vargaz I think we're probably going to need static vtables.  The ecma spec addendum says the static interface methods are going to be in the methodimpl table.  The partydonk prototype completely ignores that and does matching by name.  But we should just add a `PROP_STATIC_VTABLE` to `InfrequentDataKind` and just write down the impls from the MethodImpl in there when setting up the class.  It can be completely separate from instance vtable layout.  What do you think?

davidwrighton:
There is now a working implmentation of all of this for CoreCLR checked in. Tests are disabled for mono, but otherwise it should be ready for implementation work to begin in Mono.

lambdageek:
Zoltan implemented the initial version in #53465 

Notably generic sharing does not work correctly, so the AOT support isn't quite there yet.

lambdageek:
Remaining work pushed back to .net 7

lambdageek:
@tannergooding @stephentoub I think in general Mono is in good shape to use static virtual methods in libraries.  There may be bugs, but the only remaining completely unimplemented functionality that I'm aware of is the `constrained . ldftn` opcode - ie `&T.StaticMethod`  and creation of delegates from static virtual methods from inside another generic method.

SamMonoRT:
@lambdageek @BrzVlad - There are a couple unchecked ones. Are these required for 7.0 or can this be moved to 8.0.0 ?

BrzVlad:
Moving to 8.0 since only corner cases from runtime tests are left implementing

PR #91664, opened 2023-09-06
[mono][metadata] Fixes for static virtual methods with interface variance

### LINKED ISSUE #88689: Variant static interface dispatch behaves incorrectly on Mono
### Description

New test added in #88639 fails on Mono

### Reproduction Steps

Run the src/tests/Loader/classloader/StaticVirtualMethods/InterfaceVariance/ComplexHierarchyPositive.csproj test on a Mono platform.

### Expected behavior

Test passes

### Actual behavior

Test fails

### Regression?

No

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

BrzVlad:
Fixes https://github.com/dotnet/runtime/issues/88689
Contributes to https://github.com/dotnet/runtime/issues/49904

### DIFF
--- a/src/mono/mono/metadata/class-setup-vtable.c
+++ b/src/mono/mono/metadata/class-setup-vtable.c
@@ -1035,6 +1035,37 @@ apply_override (MonoClass *klass, MonoClass *override_class, MonoMethod **vtable
 	return TRUE;
 }
 
+/*
+ * Interface method `decl` is overridden in `klass` with `override`. In addition to overriding the method slot
+ * for this interface method, if parent klass implements a different variant compatible interface, then we also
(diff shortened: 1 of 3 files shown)


BrzVlad:
@lambdageek Could you take a look at this ?

kg:
The actual algorithm and everything look right here, but I'm concerned that applying the variance at the vtable level is wrong as a general approach, since the spec seems to require the check to be done at call sites. Or does every specific generic instantiation of a given interface have its own slot in every class's vtable? I was under the impression that wasn't true, and that was part of the motivation for the IMT in minijit.

On the other hand, for the specific scenario of static virtual methods, the vtable approach seems like it's probably safe.

kg (approved):

PR #91674, opened 2023-09-06
JIT: DNER multiregs with SIMD12s

### LINKED ISSUE #91214: Assertion failed 'isValidVectorElemsizeFloat(size)' during 'Generate code'
```c#
// Found by Antigen

using System;
using System.Numerics;
public class TestClass
{
    public struct S1
    {
    }
    public struct S2
    {
    }
    public struct S3
    {
        public struct S3_D1_F1
        {
        }
        public struct S3_D1_F4
        {
        }
        public Vector3 v3_5;
        public bool bool_6;
    }
    public struct S4
    {
    }
    public struct S5
    {
    }
    static Vector64<ulong> s_v64_ulong_31 = Vector64.Create((ulong)1);
    static S3.S3_D1_F1 s_s3_s3_d1_f1_61 = new S3.S3_D1_F1();
    static S3 s_s3_63 = new S3();
    Vector64<ulong> v64_ulong_87 = Vector64.Create((ulong)5);
    static int s_loopInvariant = 6;
    public S3 Method2(ref S3.S3_D1_F1 p_s3_s3_d1_f1_162)
    {
        unchecked
        {
            S1 s1_179 = new S1();
            S2 s2_180 = new S2();
            S3.S3_D1_F4 s3_s3_d1_f4_183 = new S3.S3_D1_F4();
            S4 s4_185 = new S4();
            S5 s5_186 = new S5();
            int __loopvar17 = s_loopInvariant + 9;
            Log("s1_179", s1_179);
            Log("s2_180", s2_180);
            Log("s3_s3_d1_f", s3_s3_d1_f4_183);
            Log("s4_185", s4_185);
            Log("s5_186", s5_186);
            return s_s3_63;
        }
    }
    public void Method0()
    {
        unchecked
        {
            S3 s3_221 = new S3();
            S3 s3_222 = s3_221;
            s_v64_ulong_31 = Vector64.ConditionalSelect(v64_ulong_87, v64_ulong_87 *= 15|4, s_v64_ulong_31 | (v64_ulong_87 *= Vector64<ulong>.Zero)^ (v64_ulong_87 += s_v64_ulong_31));
            s3_222 = Method2(ref s_s3_s3_d1_f1_61);
            Log("s3_222.v3_", s3_222.v3_5);
            return;
        }
    }
    {
        new TestClass().Method0();
    }
    public static void Log(string varName, object varValue)
{}    }
/*
Environment:

set COMPlus_TieredCompilation=0

Assert failure(PID 3508 [0x00000db4], Thread: 11984 [0x2ed0]): Assertion failed 'isValidVectorElemsizeFloat(size)' in 'TestClass:Method0():this' during 'Generate code' (IL size 148; hash 0x46e9aa75; FullOpts)
    File: D:\a\_work\1\s\src\coreclr\jit\emitarm64.cpp Line: 4225
    Image: e:\kpathak\CORE_ROOT\corerun.exe
*/
```

jakobbotsch:
Locals with SIMD12 fields will span multiple registers when they end up as multireg returns, so these should always be DNER'd.

Fix #91214

### DIFF
--- a/src/coreclr/jit/lower.cpp
+++ b/src/coreclr/jit/lower.cpp
@@ -7499,6 +7499,28 @@ bool Lowering::CheckMultiRegLclVar(GenTreeLclVar* lclNode, int registerCount)
             if (registerCount == varDsc->lvFieldCnt)
             {
                 canEnregisterAsMultiReg = true;
+
+#ifdef FEATURE_SIMD
+                // TYP_SIMD12 breaks the above invariant that "we won't have
+                // matching reg and field counts"; for example, consider
+                //
+                // * STORE_LCL_VAR<struct{Vector3, int}>(CALL)
+                // * RETURN(LCL_VAR<struct{Vector3, int}>)
+                //
+                // These return in two GPR registers, while the fields of the
+                // local are stored in SIMD and GPR register, so registerCount
+                // == varDsc->lvFieldCnt == 2. But the backend cannot handle
+                // this.
+
+                for (int i = 0; i < varDsc->lvFieldCnt; i++)
+                {
+                    if (comp->lvaGetDesc(varDsc->lvFieldLclStart + i)->TypeGet() == TYP_SIMD12)
+                    {
+                        canEnregisterAsMultiReg = false;
+                        break;
+                    }
+                }
+#endif
             }
         }
     }
--- a/src/tests/JIT/Regression/JitBlue/Runtime_91062/Runtime_91062.cs
+++ b/src/tests/JIT/Regression/JitBlue/Runtime_91062/Runtime_91062.cs
@@ -1,5 +1,5 @@
 // Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.aa
+// The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Runtime.CompilerServices;
 using System.Runtime.Intrinsics;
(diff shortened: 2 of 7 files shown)


EgorBo:
hm... I thought that the fix is more complicated per our discussion in discord: 
image

that's why I closed mine that touches the same path

EgorBo (approved):

jakobbotsch:
> hm... I thought that the fix is more complicated per our discussion in discord with you and @SingleAccretion : image
> 
> that's why I closed mine that touches the same path

After thinking about it some more (and checking diffs) I don't think we need anything more complicated here. I believe that all cases where we call `CheckMultiRegLclVar` that involves SIMD12 in the source or destination is going to end up with some kind of mismatch that the backend cannot handle; this includes the call and return cases (where the return/call has it spread out over two GPR registers, and the LCL always has it in a SIMD register) and the HWI case (where it will be split in multiple SIMD registers). I do not think that `CheckMultiRegLclVar` is used for `LCL_VAR <-> LCL_VAR` compatibility checks (these should already have been DNER'd earlier, or block morphing should have rewritten them). Maybe @SingleAccretion disagrees?

It might still be nice to do the `ReturnTypeDesc` thing, which is more general and obviously correct logic, but it's also a much larger refactoring so probably not something we'd want to do for 8.0.

SingleAccretion:
> Maybe @SingleAccretion disagrees?

The change looks good to me. Though I it would be nice to "formally prove" mismatches can only arise from SIMD12 on local side, I agree it is convincing that it is indeed the case.

jakobbotsch:
Failures are #91705 and #91723

PR #91684, opened 2023-09-06
Move Native AOT outerloop tests to their own pipeline

agocke:
Native AOT tests are regularly queued on their own and are very reliable. By putting the tests in their own queue we can both queue the Native AOT tests individually and ensure that the pipeline stays green.


eng/pipelines/coreclr/runtime-nativeaot-outerloop.yml
```diff
@@ -0,0 +1,196 @@
...
+    - eng/Version.Details.xml
+    - .devcontainer/*
+    - .github/*
+    - docs/*
+    - LICENSE.TXT
+    - PATENTS.TXT
+    - THIRD-PARTY-NOTICES.TXT
+
+schedules:
+  - cron: "0 9,21 * * *" # run at 9:00 and 21:00 (UTC) which is 1:00 and 13:00 (PST).
```
MichalStrehovsky, line 31: When you ask the first responders to create a pipeline definition for this in AzDO, could you also ask if there's a guidance on how to spread  out scheduled pipeline runs? I see this one would happen at the same time as extra-platforms. Maybe we should spread them out?

MichalStrehovsky (approved):
Looks good to me but obviously we shouldn't merge before we validate it works.

eng/pipelines/coreclr/runtime-nativeaot-outerloop.yml
```diff
@@ -0,0 +1,196 @@
+
+# This pipeline includes the Native AOT tests that we don't have resources to
+# run on every PR but that we want to still have test coverage, we run this
+# pipeline on a schedule and also developers can run it via /azp run command on
```
ivanpovazan, line 4: Just a clarification, to initiate a build manually, one would need to do: `/azp run runtime-nativeaot-outerloop` ?

ivanpovazan (approved):

eng/pipelines/coreclr/runtime-nativeaot-outerloop.yml
```diff
@@ -0,0 +1,196 @@
...
+    platforms:
+    - windows_x64
+    - windows_arm64
+    - osx_x64
+    - osx_arm64
+    - linux_x64
+    - linux_arm64
+    - linux_musl_x64
+    jobParameters:
+      testGroup: innerloop
```
kotlarmilos, line 57: Is the test group still `innerloop`?

eng/pipelines/coreclr/runtime-nativeaot-outerloop.yml
agocke: Yup that's the intent

eng/pipelines/coreclr/runtime-nativeaot-outerloop.yml
agocke: I think so. This is a terminology issue, but even though this pipeline is "outer-loop", the test group is "inner loop." I could be wrong about that, though.

PR #91622, opened 2023-09-05
Delete the RequiresPreviewFeaturesAttribute from RefreshMemoryLimit

cshung:

PR #91688, opened 2023-09-06
[release/8.0] Delete the RequiresPreviewFeaturesAttribute from RefreshMemoryLimit

cshung:
Backport of #91622 to release/8.0

### Customer Impact
This is to enable the RefreshMemoryLimit feature so that it won't require the `RequiresPreviewFeaturesAttribute`. This is going to be used by Amazon Lambda [here](https://github.com/aws/aws-lambda-dotnet/pull/1578).

### Testing
The internal working is tested by enabling the `COMMITTED_BYTES_SHADOW` feature switch - the switch will trigger a call to `RefreshMemoryLimit` for each GC, and it was run as part of the GC exit criteria run.

The managed part of the method is not currently tested, work is in progress to provide some coverage through https://github.com/dotnet/runtime/pull/91833.

### Risk
The API is already public, this change does not add any more risk for customers who will call it regardless of the existence of a warning or not.

### DIFF
--- a/src/coreclr/System.Private.CoreLib/src/System/GC.CoreCLR.cs
+++ b/src/coreclr/System.Private.CoreLib/src/System/GC.CoreCLR.cs
@@ -899,13 +899,10 @@ namespace System
         ///
         /// This API will only handle configs that could be handled when the runtime is loaded, for example, for configs that don't have any effects on 32-bit systems (like the GCHeapHardLimit* ones), this API will not handle it.
         ///
-        /// As of now, this API is feature preview only and subject to changes as necessary.
-        ///
         /// <exception cref="InvalidOperationException">If the hard limit is too low. This can happen if the heap hard limit that the refresh will set, either because of new AppData settings or implied by the container memory limit changes, is lower than what is already committed.</exception>
         /// <exception cref="InvalidOperationException">If the hard limit is invalid. This can happen, for example, with negative heap hard limit percentages.</exception>
         ///
         /// </summary>
-        [RequiresPreviewFeatures("RefreshMemoryLimit is in preview.")]
         public static void RefreshMemoryLimit()
         {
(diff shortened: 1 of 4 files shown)


stephentoub:
Are there tests somewhere for GC.RefreshMemoryLimit()?  I'm not seeing any.

Maoni0 (approved):

cshung:
> Are there tests somewhere for GC.RefreshMemoryLimit()? I'm not seeing any.

If we turn on `#define COMMITTED_BYTES_SHADOW`, this will stress test the `RefreshMemoryLimit()` before every GC. We tested it that way.

stephentoub:
> If we turn on #define COMMITTED_BYTES_SHADOW, this will stress test the RefreshMemoryLimit() before every GC. We tested it that way.

It calls the managed method, or the underlying functionality. If I were to change the managed method to nop accidentally, what test would fail?

mangod9 (approved):

cshung:
> > If we turn on #define COMMITTED_BYTES_SHADOW, this will stress test the RefreshMemoryLimit() before every GC. We tested it that way.
> 
> It calls the managed method, or the underlying functionality? If I were to change the managed method to nop or throw accidentally, what test would fail? And if the answer is "none", is that ok?

You are right, if someone nop the method, no test will detect that failure.

From a risk perspective, it is so much easier for a GC dev to forget hard limit could be changed and therefore write incompatible code than someone nop the managed method, that's why testing has been focused there. There are some glue code at the qcall layer that won't be exercised without a test calling from the managed side as well.

I think there is value to add a managed test, let me add one.

carlossanlop:
@cshung can you please fill out the servicing template from the PR description and get an M2 approval when this is ready to merge?

jeffschwMSFT (approved):
approved.  this can be merged when ready

Issue #85592, opened 2023-05-01
[browser][MT] Multithreading and JavaScript async interop in .NET 9


lambdageek:
Tracking issue for further work on JS interop and multithreading.

Constituent part of https://github.com/dotnet/runtime/issues/76956

## Goals
- CPU intensive workloads on dotnet thread pool
- Allow user to start new managed threads using `new Thread` and join it.
- Add new C# API for creating dedicated web workers with JS interop. Allow JS async/promises via external event loop.
- enable blocking `Task.Wait` and `lock()` like APIs from C# user code on all threads or non-JS/UI threads
    - Current public API throws PNSE for it
    - This is core part on MT value proposition.
    - If people want to use existing MT code-bases, most of the time, the code is full of locks.
    - People want to use existing desktop/server multi-threaded code as is.
- allow HTTP and WS C# APIs to be used from any thread despite underlying JS object affinity
- JSImport/JSExport interop in maximum possible extent.
- don't change/break single threaded build. †

## Lower priority goals
- try to make it debugging friendly
- sync C# to async JS
    - dynamic creation of new pthread
    - implement crypto via `subtle` browser API
    - allow MonoVM to lazily download DLLs from the server, instead of during startup.
    - implement synchronous APIs of the HTTP and WS clients. At the moment they throw PNSE.
- sync JS to async JS to sync C#
    - allow calls to synchronous JSExport from UI thread (callback)
- don't prevent future marshaling of JS [transferable objects](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects), like streams and canvas.
- offload CPU intensive part of WASM startup to WebWorker, so that the pre-rendered (blazor) UI could stay responsive during Mono VM startup.

## Non-goals
- interact with JS state on WebWorker of managed threads other than UI thread or dedicated `JSWebWorker`

## Design discussion and experiment
* [x] Design discussion https://github.com/dotnet/designs/pull/301
* [x] test deputy/sidecar design with Blazor (in some limited way)
* https://github.com/dotnet/runtime/pull/92226
* https://github.com/dotnet/runtime/pull/93920
* https://github.com/dotnet/runtime/pull/94525

## Depending on selected design
* [x] call dispatch to/from UI thread
* [x] call dispatch to/from JSWebWorker thread
* [x] Make Promise -> Task marshaling ahead of call
* [x] Make string marshaling non GC boundary breaching
* [x] Make Array marshaling non GC boundary breaching
[...]

Issue #69409, opened 2022-05-16
[wasm-mt] Add a System.Runtime.InteropServices.JavaScript.BrowserSynchronizationContext


lambdageek:
**Update** this is no longer a public API proposal.  instead, we should make an `internal` SynchronizationContext subclass that is installed by the runtime on the main thread.

---

### Background and motivation

In order to enable multi-threaded WebAssembly programming, we want to allow users to continue to program using `async`/`await`.  In the current single-threaded `browser-wasm` runtime, all asynchronous tasks run on the main browser thread implicitly as there is nowhere else to run them.  Our multi-threaded WebAssembly programming model is built on WebWorkers - dedicated threads that do not share system memory with the main browser thread but may pass messages to it back and forth including WebAssembly main memory using a SharedArrayBuffer.

In multi-threaded WebAssembly, certain operations (for example code that manipulates the DOM of the current page) must run on the browser's main event loop, not in a WebWorker.  As a result, it is necessary to add a `SynchronizationContext` subclass that allows asynchronous work to be posted back to the main browser thread.

### API Proposal

```csharp
using System.Threading;

namespace System.Runtime.InteropServices.JavaScript;

/// A synchronization context that queues work to run on the browser's main thread.
/// In a multi-threaded web application certain operations (such as manipulating the DOM) must be done on the
/// main thread and not in a WebWorker.  The BrowserSynchronizationContext allows asynchronous callbacks to
/// return to run their continuations on the main thread.
[SupportedOSPlatform("browser")]
internal sealed class BrowserSynchronizationContext : SynchronizationContext
{
    internal BrowserSynchronizationContext();
    public override SynchronizatonContext CreateCopy();
    public override void Post(SendOrPostCallback d, object state);
    //// Always throws NotSupportedException
    public override void Send(SendOrPostCallback d, object state);
}
```

### API Usage

In a .razor page in Blazor:
```razor
<button @onclick="DoSomeWork">Start Work</button>
<text>Result is: @output</text>

@code {
    string output = "";
    public async Task DoSomeWork() {
        output = "pending";
        Debug.Assert (SynchronizationContext.Current instanceof BrowserSynchronizationContext);
        var result = await Task.Run(Model.SomeAsyncWork); // start some work on the threadpool
        output = result.ToString();  // return to the UI thread and update the model
    }
}
```

### Alternative Designs

[...]

PR #91696, opened 2023-09-06
[browser] WASM deputy worker - multi-threading proposal

pavelsavara:
This is proposal on how to enable threads on WASM, while allowing for blocking `.Wait`, `lock() { ... }` and similar from user C# code.

Feedback and questions are welcome.

Contributes to https://github.com/dotnet/runtime/issues/85592

There is alternative proposal https://github.com/dotnet/runtime/pull/91731

### DIFF
--- a/docs/design/mono/wasm-threads.md
+++ b/docs/design/mono/wasm-threads.md
@@ -96,11 +96,13 @@
 - has SynchronizationContext installed on it
     - So that C# calls could be dispatched to it by runtime
 - throw PNSE on attempt to marshal sync C# delegate to UI JavaScript
+    - or throw later only when you try to call the function from JS side.
 - can run C# finalizers
 - will run GC
 - this cross-threading dispatch will have performance impact for the JS interop.
     - TODO: measure how much
     - this should not impact Blazor `renderBatch` perf.
+- VS debugger would connect to mono as usual. But chrome dev tools experience may be different, because it's will be async with the C# part.
 
 ## JSWebWorker with JS interop
 - is C# thread created and disposed by new API for it
@@ -194,7 +196,7 @@
 - Blazor renderBatch will continue working even with legacy interop in place.
     - Because it only reads memory and it doesn't call back to Mono VM.
 - Blazor's [`IJSInProcessRuntime.Invoke`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.jsinterop.ijsinprocessruntime.invoke) should still work, because it's C#->JS direction
-- Blazor's [`IJSUnmarshalledRuntime `](https://learn.microsoft.com/en-us/dotnet/api/microsoft.jsinterop.ijsunmarshalledruntime) should still work, because it's C#->JS direction
+- Blazor's [`IJSUnmarshalledRuntime`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.jsinterop.ijsunmarshalledruntime) should still work, because it's C#->JS direction
 - TODO: Review Blazor's JavaScript APIs!
 
 # Current state 2023 Sep
--- a/docs/design/mono/wasm-threads.md
+++ b/docs/design/mono/wasm-threads.md
@@ -176,7 +176,7 @@
 - on the deputy thread
     - to dispatch async calls from UI thread to it
 
-### Blazor - what breaks when MT build
+## Blazor - what breaks when MT build
 - as compared to single threaded runtime, the major difference would be no synchronous callbacks.
     - for example from DOM `onClick`. This is one of the reasons people prefer ST WASM over Blazor Server.
     - but there is really [no way around it](#problem), because you can't have both MT and sync calls from UI.
@@ -187,7 +187,7 @@
 - throw PNSE from Blazor's [`IJSInProcessRuntime.Invoke`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.jsinterop.ijsinprocessruntime.invoke)
 - throw PNSE from Blazor's any call to [`IJSUnmarshalledRuntime `](https://learn.microsoft.com/en-us/dotnet/api/microsoft.jsinterop.ijsunmarshalledruntime)
 
-## Current state 2023 Sep
+# Current state 2023 Sep
  - we already ship MT version of the runtime in the wasm-tools workload.
  - It's enabled by `<WasmEnableThreads>true</WasmEnableThreads>` and it requires COOP HTTP headers.
  - It will serve extra file `dotnet.native.worker.js`.
--- a/docs/design/mono/wasm-threads.md
+++ b/docs/design/mono/wasm-threads.md
@@ -176,9 +176,9 @@
 - on the deputy thread
     - to dispatch async calls from UI thread to it
 
-### Blazor - what breaks
+### Blazor - what breaks when MT build
 - as compared to single threaded runtime, the major difference would be no synchronous callbacks.
-    - for example from DOM `onClick`. This is one of the reasons people prefer WASM over Blazor Server.
+    - for example from DOM `onClick`. This is one of the reasons people prefer ST WASM over Blazor Server.
     - but there is really [no way around it](#problem), because you can't have both MT and sync calls from UI.
 - implement Blazor's `WebAssemblyDispatcher` to dispatch [`Component.InvokeAsync`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.componentbase.invokeasync) to deputy thread.
     - process feedback from https://github.com/dotnet/aspnetcore/pull/48991 and make more async
--- a/docs/design/mono/wasm-threads.md
+++ b/docs/design/mono/wasm-threads.md
@@ -150,7 +150,7 @@
     - how to do that during GC/finalizer ?
     - should we run finalizer per worker ?
 - is it ok for `SynchronizationContext` to be public API
-    - because it could contain UI thread SynchronizationContext, which user code should not be disposed on.
+    - because it could contain UI thread SynchronizationContext, which user code should not be dispatched on.
 
 ## JSHost.GlobalThis, JSHost.DotnetInstance, JSHost.ImportAsync
 - calls will be dispatched from deputy thread to UI JavaScript
@@ -165,8 +165,9 @@
     - is solution for deputy's SynchronizationContext same as for JSWebWorker's SynchronizationContext, from the code-gen perspective ?
     - how could "HTTP from any C# thread" redirect this to the thread of fetch JS object affinity ?
     - should generated code or the SynchronizationContext detect it from passed arguments ?
-    - TODO: explain why not make user responsible for doing it, instaed of changing generator
     - TODO: figure out backward compatibility of already generated code. Must work on single threaded
+- why not make user responsible for doing it, instead of changing generator ?
+    - I implemented MT version of HTTP and WS by calling `SynchronizationContext.Send` and it's less than perfect. It's difficult to do it right: Error handling, asynchrony.
 - on a JSWebWorker
     - to dispatch any calls of JSObject proxy members
     - to dispatch `Dispose()` of JSObject proxy
--- a/docs/design/mono/wasm-threads.md
+++ b/docs/design/mono/wasm-threads.md
@@ -23,6 +23,7 @@
      - in single-threaded build of the runtime, all of this is NOOP. That's why it works on UI thread.
 2. UI thread in the browser can't synchronously block
      - you can spin-lock but it's bad idea.
+         - Deadlock: when you spin-block, the JS timer loop and any messages are not pumping. But code in other threads may be waiting for some such event to resolve.
          - It eats your battery
          - Browser will kill your tab at random point (Aw, snap).
          - It's not deterministic and you can't really test your app to prove it harmless.
@@ -129,7 +130,7 @@
     - fast on worker -> worker
 
 ## Promise
-- passing promise should work everywhere.
+- passing Promise should work everywhere.
 - from UI javaScript it would be passed as Task to deputy worker
 - open question: passing JS promise to deputy should be fine. But does the `resolve()` need to block UI thread ?
 
@@ -157,6 +158,7 @@
 
 ## SynchronizationContext
 - we will need public C# API for it, `JSHost.xxxSynchronizationContext`
+- maybe `JSHost.Post(direction, lambda)` without exposing the `SynchronizationContext` would be better.
     - we could avoid it by generating late bound ICall. Very ugly.
 - hide `SynchronizationContext.Send` and `SynchronizationContext.Post` inside of the generated code.
     - needs to be also inside generated nested marshalers
@@ -175,6 +177,10 @@
         - it will need some public API any way, to be called from generated code.
 - on the deputy thread
     - to dispatch async calls from UI thread to it
+
+### dispatch alternatives
+- we could use emscripten's `emscripten_dispatch_to_thread_async` or JS `postMessage`
+- the details on how to interleave that with calls to `ToManaged` and `ToJS` for each argument may be tricky.
 
 ## Blazor - what breaks when MT build
 - as compared to single threaded runtime, the major difference would be no synchronous callbacks.
@@ -223,7 +229,7 @@
     - [ ] on UI synchronous JSImport
     - [ ] on UI synchronous C# delegate callback
     - [ ] throw fatal if somehow C# code was blocking on UI thread.
-- [ ] optinal: make underlying emscripten WebWorker pool allocation dynamic
+- [ ] optinal: make underlying emscripten WebWorker pool allocation dynamic, or provide C# API for that.
 - [ ] optinal: implement async function/delegate marshaling in JSImport/JSExport parameters.
 - [ ] optinal: enable blocking HTTP/WS APIs
 - [ ] optinal: enable lazy DLL download by blocking the caller
--- a/docs/design/mono/wasm-threads.md
+++ b/docs/design/mono/wasm-threads.md
@@ -138,7 +138,7 @@
 - from UI javaScript it would be passed as Task to deputy worker
 - open question: passing JS promise to deputy should be fine. But does the `resolve()` need to block UI thread ?
 
-## Task, Task<T>
+## Task, Task\<T>
 - passing Task should work everywhere.
 - when marshaled to JS they bind to specific Promise and have affinity
     - the `SetResult` need to be marshaled on thread of the Promise.
--- a/docs/design/mono/wasm-threads.md
+++ b/docs/design/mono/wasm-threads.md
@@ -1,0 +1,125 @@
+# Multi-threading on browser
+
+## Goals
+ - don't change/break single threaded build. †
+ - JS interop
+     - both sync and async calls and callbacks
+        - sync calls from JS to C# are part of the problem, see below
+     - from UI thread to "some C# main thread" and back
+     - from dedicated worker to it's JavaScript state and back
+ - CPU intensive workloads on dotnet thread pool
+ - enable blocking .Wait APIs from C# user code
+     - Current public API throws PNSE for it
+ - allow HTTP and WS C# APIs to be used from any thread
+     - Underlying JS object have thread affinity
+
+<sub><sup>† Note: all the text below discusses MT build only, unless explicit about ST build.</sup></sub>
+
+# Problem
+1. If you have multithreading, any thread might need to block while waiting for any other to release a lock.
+     - in user code, in nuget packages, in Mono VM itself, there are managed and un-managed locks
+     - in single-threaded build of the runtime, all of this is NOOP. That's why it works.
+2. UI thread in the browser can't synchronously block
+     - you can spin-lock but it's bad idea. 
+         - It eats your battery 
+         - Browser will kill your tab at random point (Aw, snap).
+         - It's not deterministic and you can't really test your app to prove it harmless.
+     - all the other threads/workers could synchronously block
+3. JavaScript engine APIs and objects have thread affinity. The DOM and few other browser APIs are only available on the main UI "thread"
+     - and so, you need to have C# interop with UI, but you can't block there.
+
+## Design proposal
+
+### TL;DR
+4. execute C# code on "deputy worker", executing all user C# on behalf of the UI JavaScript
+5. throw PNSE when UI JavaScript would call in any synchronous JSExport or callback to C#. This will prevent Mono from trying to randomly block the call.
+
+### Alternatives
+10. create emscripten engine on worker
+- this is similar to 4. but not feasible
+- it would break lot of existing JavaScript APIs
+- it would make startup callbacks on wrong thread (blazor JS integration)
+- we would have to re-write Blazor renderBatch
+11. throw PNSE any time C# code needs to block
+- It's not deterministic and you can't really test your app to prove it harmless.
+12. throw PNSE any time C# code or VM code needs needs to block
+- Mono VM needs to hold lock while allocating memory, even on the UI thread.
+
+## Design proposal details
+
+### UI thread
+- this is the main browser "thread", the one with DOM on it
+- will start emscripten as usual
+- will create Deputy worker as C# thread
+- dispatch execution of C# `Task Main()` to the deputy worker.
+- dispatch all async JSImport calls to the deputy worker.
+- dispatch all async callbacks to the deputy worker.
+- throw PNSE on any JSImport call from JS
+- it will be valid C# thread, but not used directly by user code.
+    - we will try to prevent user code from running on it and from needing to do so
+- we will spin lock only for Mono VM code
+    - we assume that Mono VM will block only shortly for operations like:
+        - alloc/free memory
+        - transform IL -> IR and update Mono VM shared state
+    - we will spin lock before Blazor `renderBatch`
+        - to wait for "pause GC"
+    - we will spin lock during GC, if we hit barrier
+        - TODO: is that short enough ?
+    - we should never block for file operations or for network operations
+        - TODO: how to prove it ?
+
+### Deputy worker
+- executing all **user C# code** on behalf of the UI JavaScript "thread"
+    - that is also C# entry point `Task Main()` or `void Main()`
+    - `void Main()` would return promise that never resolves to UI JavaScript
+- doesn't expose JavaScript state to user code.
+    - as optimization we could consider running HTTP and WS client here, instead of UI thread.
+- has SynchronizationContext installed on it
+    - So that C# calls could be dispatched to it by runtime
+- throw PNSE on attempt to marshal sync C# delegate to UI JavaScript
+- can run C# finalizers
+- will run GC
+
+### WebWorker with JS interop
+- is C# thread created and disposed by new API for it
+- there is JSSynchronizationContext installed on it
+    - so that user code could dispatch back to it, in case that it needs to call JSObject proxy (with thread affinity)
+
+### C# Thread
+- without JS interop. calling JSImport will PNSE.
+
+### C# Threadpool Thread
+- without JS interop. calling JSImport will PNSE.
+
+### JSImport
+
+### JSExport
+
+### Promise
+
+### Task, Task<T>
+
+### Blazor
+ - provide SynchronizationContext which would dispatch [`Component.InvokeAsync`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.components.componentbase.invokeasync) to 
+ - Blazor renderBatch will continue working even with legacy interop in place. 
+   - Because it only reads memory and it doesn't call back to Mono VM.
+ - throw PNSE from Blazor's [`IJSInProcessRuntime.Invoke`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.jsinterop.ijsinprocessruntime.invoke)
+ - throw PNSE from Blazor's any call to [`IJSUnmarshalledRuntime `](https://learn.microsoft.com/en-us/dotnet/api/microsoft.jsinterop.ijsunmarshalledruntime)
+
+## Current state 2023 Sep
+ - we have pre-allocated pool of webWorker which are mapped to pthread dynamically.
+ - we can configure pthread to keep running after synchronous thread_main finished. That's necessary to run any async tasks involving JavaScript interop.
+ - GC is running on UI thread/worker.
+ - legacy interop has problems with GC boundaries.
+ - JSImport & JSExport work
+ - There is private JSSynchronizationContext implementation which is too synchronous
+ - There is draft of public C# API for creating WebWorker with JS interop. It must be dedicated un-managed resource, because we could not cleanup JS state created by user code.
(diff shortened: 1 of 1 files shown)


docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,231 @@
...
+- is it ok for `SynchronizationContext` to be public API
+    - because it could contain UI thread SynchronizationContext, which user code should not be disposed on.
+
+## JSHost.GlobalThis, JSHost.DotnetInstance, JSHost.ImportAsync
+- calls will be dispatched from deputy thread to UI JavaScript
+- on JSWebWorker call will stay on the same thread.
+
+## SynchronizationContext
+- we will need public C# API for it, `JSHost.xxxSynchronizationContext`
+    - we could avoid it by generating late bound ICall. Very ugly.
```
lambdageek, line 160: this is for the benefit of the JSImport/Export source generator, right?

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,231 @@
...
+- hide `SynchronizationContext.Send` and `SynchronizationContext.Post` inside of the generated code.
+    - needs to be also inside generated nested marshalers
+    - is solution for deputy's SynchronizationContext same as for JSWebWorker's SynchronizationContext, from the code-gen perspective ?
```
lambdageek, line 163: I want the answer to be "yes"

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,231 @@
...
+    - how could "HTTP from any C# thread" redirect this to the thread of fetch JS object affinity ?
+    - should generated code or the SynchronizationContext detect it from passed arguments ?
+    - TODO: explain why not make user responsible for doing it, instaed of changing generator
+    - TODO: figure out backward compatibility of already generated code. Must work on single threaded
```
lambdageek, line 167: This seems hard

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,231 @@
...
+    - same as above
+    - as alternative we could only have there emscripten C dispatcher
+        - it will need some public API any way, to be called from generated code.
+- on the deputy thread
+    - to dispatch async calls from UI thread to it
+
+## Blazor - what breaks when MT build
+- as compared to single threaded runtime, the major difference would be no synchronous callbacks.
+    - for example from DOM `onClick`. This is one of the reasons people prefer ST WASM over Blazor Server.
+    - but there is really [no way around it](#problem), because you can't have both MT and sync calls from UI.
```
lambdageek, line 182: We could allow it but tell users not to use it.  This is similar ot guidance in other UI framworks not to block UI threads.  They don't prohibit you from running code but then you're on the hook for your UI freezing up (or in this case draining the battery and getting killed by the mobile browser)

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,231 @@
...
+- [ ] reimplement HTTP and WS with the new code gen without direct SynchronizationContext use
+    - [ ] there is synchronous callback from JS event to C# in HTTP code.
+- [ ] make C# finalizers work
+- [ ] throw PNSE - fail fast, so that users discover limits in the dev loop
+    - [ ] on any MT use of `mono_bind_static_method` from legacy interop.
+        - Because it's synchronous. It throws on JSWebWorker already.
+    - [ ] on UI synchronous JSImport
+    - [ ] on UI synchronous C# delegate callback
+    - [ ] throw fatal if somehow C# code was blocking on UI thread.
+- [ ] optinal: make underlying emscripten WebWorker pool allocation dynamic
```
lambdageek, line 225: we could also expose an async managed API to spin up additional workers (there's a GH Issue for that somewhere...)

lambdageek (commented):
nice writeup. I'm a little bit worried the model with the deputy thread will be difficult to explain.

docs/design/mono/wasm-threads.md
pavelsavara: Generators uses public APIs because it generates user code.

docs/design/mono/wasm-threads.md
pavelsavara: I hope so too. Over night I realized that shape of this API could be 2 methods on the `JSHost` rather than exposing the `SynchronizationContext`, because there could more complex logic on which of the should be used.

docs/design/mono/wasm-threads.md
pavelsavara: We could have MSBuild property for it, which would be opt-in. Right now, I think if we are doing all of this complex stuff to prevent it, we should gain the benefits of it. Otherwise we will have to support crazy things.

I though that we could get away with spin block. But MT unit tests on CI are hanging in too many cases, and probable for this reason. We just need to get over it, me included.

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,231 @@
...
+
+## Problem
+1. If you have multithreading, any thread might need to block while waiting for any other to release a lock.
+     - locks are in the user code, in nuget packages, in Mono VM itself
+     - there are managed and un-managed locks
+     - in single-threaded build of the runtime, all of this is NOOP. That's why it works on UI thread.
+2. UI thread in the browser can't synchronously block
+     - you can spin-lock but it's bad idea.
+         - It eats your battery
+         - Browser will kill your tab at random point (Aw, snap).
```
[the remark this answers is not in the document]
pavelsavara: If we could get away with not preventing it, this whole design doesn't need to be implemented. I would be happy to save myself lot of work. But right now, I think this is necessary.

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,231 @@
...
+    - we will spin lock before Blazor `renderBatch`
+        - to wait for "pause GC"
+    - we will spin lock during GC, if we hit barrier
+        - TODO: is that short enough ?
+    - we should never block for file operations or for network operations
+        - TODO: how to prove it ?
+        - Could we unregister the thread from Mono VM ? No, because we need the C# to dispatch calls in both directions in
+- it will actually execute small chunks of C# code
+    - the pieces which are in the generated code of the JSExport
+    - containing dispatch to deputy worker's synchronization context
```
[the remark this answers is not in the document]
pavelsavara: JS interop is creating JS objects and that needs to happen on the right JS worker, because of affinity.
The JS side of the marshaller is calling static C# methods, to create C# instances of `Task`.

To the second question, that's alternative 10. The main reason is that JS interop and also the JS embedding APIs actually needs to make those short-lived synchronous calls to Mono/C#. Otherwise even JS code of marshaling individual arguments would have to be async, because it would send messages to a "server". 

The Blazor `renderBatch` is also touching memory and converting `MonoString*` etc. That could not become async, perf would just die.

It could be done, but it would not be less trouble than having C# on the UI thread internally.

pavelsavara:
> I'm a little bit worried the model with the deputy thread will be difficult to explain.

Yes, but hopefully the users don't need to know that the C# is running elsewhere and the only visible difference would be sync calls from JS and bit of interop perf. This "deputy thread" concept is mostly for this small audience.

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,231 @@
...
+- could sync void C# methods be dispatched as fire and forget ?
+    - no, because that would break the contract that they are blocking until finished.
+    - also the errors would not propagate
+- TODO: is there anything special about error propagation over the interop/thread boundary ?
+
+## Deputy worker
```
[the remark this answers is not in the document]
pavelsavara: Yes, deputy is similar to `JSWebWorker` except it talks to UI thread JS space, not it's own (on logical level).
The implementation of how it works is not JS event loop, but rather C# `SynchronizationContext`.

Passing those messages could be also done via emscripten's `emscripten_dispatch_to_thread_async` or by JS `postMessage`.

Doing it in C# allows us to interleave it with the code generated by Roslyn codegen.
Let's do more thinking about this, how to actually do it.

One of the open questions for me is selection of target thread based on affinity of the `JSObject` proxy passed as argument. Note we are talking about static methods which could have more or also none just object. 
This question is relevant for "call HTTP from any C# thread" scenario.

docs/design/mono/wasm-threads.md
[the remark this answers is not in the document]
pavelsavara: I added one more reason: `Deadlock: when you spin-block, the JS timer loop and any messages are not pumping. But code in other threads may be waiting for some such event to resolve.`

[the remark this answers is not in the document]
pavelsavara: In discussion with @kg I realized that that we could implement just first part of the "deputy" as experiment and see if it solves some of the CI failures we have. After that we would know more about if this is valuable direction.

kg:
I think you've done a fantastic job coming up with a solution for this problem, but I think it might be best to not solve this problem. While reading over the design I kept thinking "this has a lot in common with blazor server, maybe we can reuse all its remoting and etc for this scenario".

The way I see it this proposes a sort of hybrid model between existing blazor WASM and existing blazor server:

1. Everything runs on the client
2. (new) The client is running a client thread and a server thread (the latter is the well-named deputy worker)
3. There is a client thread on the client and a server thread on the server

I think there's a lot of value in having this hybrid option - it reduces latency vs blazor server, and reduces the resources spent on the server and reduces bandwidth spent communicating between client and server. That's great.

But do we actually want to have three models? Will users comprehend all the nuances of the 3 models and know which one to pick for their scenario? Can we communicate that clearly to them? Will users feel betrayed if they commit to model 2 and then discover they actually needed model 1, or vice versa, and it's hard to swap after the fact? Adding a third distinct model with its own approach to interop feels like it adds a lot of education and maintenance burden since we can't say 'everything that applies to Blazor Server applies to this' or 'everything that applies to Blazor WASM applies to this'. Users may have selected Blazor WASM specifically for the ability to seamlessly communicate between C# and JS, which goes away in this new model.

I think at the end of the day a lot of code makes assumptions that don't hold for the browser, and the existence of the deputy worker will protect that code from some of its assumptions but the code is still fundamentally broken for this environment. It assumes things that aren't true. It would be ideal if we could make all existing C# concurrency constructs and code work in the browser as-is but I'm not sure if the costs we pay for doing so are justified - the existence of the deputy worker and all the inter-thread coordination will have a measurable cost (hopefully small), startup and shutdown get more complex, debugging gets WAY more complex, etc. This could be especially bad on mobile, where increased resource usage = increased odds of a "tab crash"

For the specific scenario of "some code needs to block while waiting for a lock to be released":
* We successfully move the stall off of the UI thread onto the deputy thread. That is good, it means the browser thread isn't stalled. But the app isn't actually responsive, you will click buttons or interact with widgets and nothing will happen because the deputy thread is hung waiting for a lock. Is that improvement tangible enough to justify the undertaking of moving to the hybrid model? I think this new failure mode is potentially more confusing for the user. Imagine you press a button, the deputy worker starts crunching data, then you type some text into a textarea *and the deputy worker wakes up and immediately obliterates the textarea*. A blocked UI thread prevents this, which is *good*!
* In general if an app is waiting on locks long enough to perceptibly hang the browser thread, the app's locking model is wrong. The deputy worker becomes a heroic effort to compensate for this root problem that will not be able to fully alleviate it.
* I'm especially concerned about the debugging scenario. If I hit the pause button in the browser debugger right now for a hung blazor app, nothing happens because browser debuggers are bad... okay, but if I open the debugger *before* launching the app and then hit pause while it's hung, I have a call stack that shows what's going on, and I can at least tell I have a locking problem. If my app is AOTd the stack may even suggest what code is waiting on a lock, which is fantastic. In the deputy worker model *nothing will happen when I hit pause* because the unresponsiveness is now due to another thread being stuck. I now need to figure out which thread is actually preventing the browser thread from doing its job. The debugging story for this sort of exists on native but on the web it doesn't. Debugging concurrency issues in the browser is already awful and we would be exposing more people to that. (Maybe we just have to build a proper concurrency debugger for the web.)
* Ultimately any user-exposed code should not be able to block for a long period of time. If it's waiting it should be waiting with a short timeout, if it even waits at all. It should probably be using queues instead, and the UI should be designed to avoid situations where the user performs a bunch of inputs that are all blocked on some invisible processing in the background (i.e. while the app is Busy the UI is locked out to communicate that clicks won't do anything). We can make this bad code work better via the heroic introduction of the deputy worker but it will still be broken and end users will probably still demand better than the experience it's possible to offer them.
* This code might be in the BCL. Purging the BCL of locks seems like a tall order (maybe I'm overestimating this) and since the main thread will still be running C#, it's still possible it will take locks, it just won't be directly the user's fault anymore. The deputy worker or any other thread will be able to take a lock and then cause BCL code to block the main thread when it tries to acquire the lock, right? So we've improved the situation but not completely prevented the "UI thread is waiting on a lock" scenario, I think.

You've clearly already thought through the problem scenarios here and all the nuances we have to tackle, like remoting errors across thread boundaries, the thread affinity of JS objects, etc. I'm confident that if we go forward with this design we'll eventually end up with something really good. My concern is ultimately that I don't know how many end users would actually be happy after adopting it, because the code they're trying to run is ill-suited for the web environment. I think our time and resources might be better spent on continued efforts to try and move users away from synchronous blocking models for their web apps, and perhaps improved tooling to ease that transition. 

On the other hand, maybe the deputy worker model gets users 90% of the way there, and they can fix the other 10% of cases and ship their app and not have to think about it ever again, and their users will be satisfied even if they're unhappy with how the app misbehaves while the deputy worker is hung. That's a real possibility and maybe it's good enough.

At the end of the day though I'm not sure I can convince myself that `Monitor.Enter` or `Task.Wait` should work in the browser unless a very low timeout is specified. We want to surface:
* it's not guaranteed you will acquire your lock / your work will complete
* so since it's not guaranteed, you need to handle the case where what you expected to happen hasn't happened
* this means you need to think about how your application should behave in both scenarios

I'm also mildly concerned that any illusion we manage to maintain in this regard will eventually break. Browser vendors already made the arbitrary decision to ban timers, what's stopping them from deciding to terminate any web worker that runs for too long without yielding to the event loop? or to flip it around: browser vendors got away with this, so why can't we tell people infinite-timeout locks are banned? 😊

In conclusion (this comment has gotten away from me) I think maybe the right decision is instead to gently force developers to take their medicine, and make the medicine taste as good as possible with a set of really good debugging and development tools that build on what we've already got.

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,239 @@
...
+         - It eats your battery
+         - Browser will kill your tab at random point (Aw, snap).
+         - It's not deterministic and you can't really test your app to prove it harmless.
+     - all the other threads/workers could synchronously block
+3. JavaScript engine APIs and objects have thread affinity. The DOM and few other browser APIs are only available on the main UI "thread"
+     - and so, you need to have C# interop with UI, but you can't block there.
+
+## Design proposal TL;DR
+4. execute C# code on "deputy worker", executing all user C# on behalf of the UI JavaScript
+5. throw PNSE when UI JavaScript would call in any synchronous JSExport or callback to C#. ††
```
kg, line 36: this seems to contradict the "both sync and async calls and callbacks" part of the goals section (i'm not opposed to it though, just confused)

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,239 @@
...
+<sub><sup>†† This will prevent user C# code from trying to randomly synchronously block the caller on UI thread.</sup></sub>
+
+## Alternatives
+10. create emscripten engine on worker
+- this is similar to 4. but not feasible
+- it would break lot of existing JavaScript APIs
+- it would make startup callbacks on wrong thread (blazor JS integration)
+- we would have to re-write Blazor's `renderBatch` to bytes streaming.
+11. throw PNSE any time C# code needs to block
+- It's not deterministic and you can't really test your app to prove it harmless.
```
kg, line 47: technically, we could deterministically PNSE on any attempt to acquire a lock, even successful ones, since those are all 'possible blocks'. But you're 100% right here.

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,239 @@
...
+- dispatch all async callbacks to the deputy worker.
+- throw PNSE on any JSImport call from JS
+- it will be valid C# thread, but not used directly by user code.
+    - we will try to prevent user code from running on it and from needing to do so
+- we will spin lock only for Mono VM code
+    - we assume that Mono VM will block only shortly for operations like:
+        - alloc/free memory
+        - transform IL -> IR and update Mono VM shared state
+    - we will spin lock before Blazor `renderBatch`
+        - to wait for "pause GC"
```
kg, line 71: what happens if we can't pause the GC because the deputy worker is stuck? I guess the browser thread hangs, right?

docs/design/mono/wasm-threads.md
pavelsavara: I should re-word it. throwing PNSE is easy, but it doesn't solve the "test my app and prove it valid" problem

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,239 @@
...
+    - we will spin lock during GC, if we hit barrier
+        - TODO: is that short enough ?
```
kg, line 73: GCs can take hundreds of MS in the real world (I'm not sure why, but our GC under WASM seems to be much slower than it would be on native) so we should assume that users on lower spec devices will see hangs if we have to do this. It might still be the right choice

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,239 @@
...
+    - that is also C# entry point `Task Main()` or `void Main()`
+    - `void Main()` would return promise that never resolves to UI JavaScript
+- this thread **could block** on synchronization primitives just fine!
+- doesn't expose JavaScript state to user code.
+    - because JSHandle has thread affinity and it's unique per JS thread.
+    - as optimization we could consider running HTTP and WS client here, instead of UI thread. But JSHandle problem.
+- has SynchronizationContext installed on it
+    - So that C# calls could be dispatched to it by runtime
+- throw PNSE on attempt to marshal sync C# delegate to UI JavaScript
+- can run C# finalizers
```
kg, line 97: finalizers already run on a separate thread, not the main thread, AFAIK. the finalizer thread is owned by the GC, I think?

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,239 @@
...
+- will run GC
```
kg, line 98: Can we somehow guarantee that GC will never happen on the browser thread? The browser thread will potentially have to stall and wait for GC at the very least, if we ever allow C# to run on it.

docs/design/mono/wasm-threads.md
pavelsavara: Updated

docs/design/mono/wasm-threads.md
pavelsavara: The other alternative is to rewrite `renderBatch` to streaming bytes.

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,239 @@
...
+- this cross-threading dispatch will have performance impact for the JS interop.
+    - TODO: measure how much
+    - this should not impact Blazor `renderBatch` perf.
+
+## JSWebWorker with JS interop
+- is C# thread created and disposed by new API for it
+- could block on synchronization primitives
+- there is JSSynchronizationContext installed on it
```
kg, line 108: FYI since you mentioned performance, be aware that (at present) having any kind of custom synchronization context installed causes some of the Task/Await machinery to go through a slow-path. see https://github.com/dotnet/runtime/issues/69409#issuecomment-1219733545 (though I hope this won't actually matter)

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,239 @@
...
+- executing all **user C# code** on behalf of the UI JavaScript "thread"
+    - that is also C# entry point `Task Main()` or `void Main()`
+    - `void Main()` would return promise that never resolves to UI JavaScript
+- this thread **could block** on synchronization primitives just fine!
+- doesn't expose JavaScript state to user code.
+    - because JSHandle has thread affinity and it's unique per JS thread.
+    - as optimization we could consider running HTTP and WS client here, instead of UI thread. But JSHandle problem.
+- has SynchronizationContext installed on it
+    - So that C# calls could be dispatched to it by runtime
+- throw PNSE on attempt to marshal sync C# delegate to UI JavaScript
```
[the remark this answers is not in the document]
pavelsavara: We could fail bit later, when you actually try to call it. But you have to call it on the right thread, because of JS affinity.
I will update the doc about it.

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,239 @@
...
+- from UI javaScript it would be passed as Task to deputy worker
+- open question: passing JS promise to deputy should be fine. But does the `resolve()` need to block UI thread ?
+
+## Task, Task<T>
+- passing Task should work everywhere.
+- when marshaled to JS they bind to specific Promise and have affinity
+    - the `SetResult` need to be marshaled on thread of the Promise.
+    - The proxy of the Promise knows which `SynchronizationContext` to dispatch to.
+    - on UI thread it's the UI thread's SynchronizationContext, not deputy's
+    - TODO: could same task be marshaled to multiple JS workers ?
```
kg, line 145: we should assume it can and will happen

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,239 @@
...
+    - in deputy worker, it will be always UI thread Id
+    - the JSHandle always belongs to UI thread
+- `Dispose` need to be called on the right thread.
+    - how to do that during GC/finalizer ?
+    - should we run finalizer per worker ?
+- is it ok for `SynchronizationContext` to be public API
+    - because it could contain UI thread SynchronizationContext, which user code should not be dispatched on.
+
+## JSHost.GlobalThis, JSHost.DotnetInstance, JSHost.ImportAsync
+- calls will be dispatched from deputy thread to UI JavaScript
```
kg, line 158: this is a tough pill to swallow. i can see how it's necessary to maintain compatibility for existing code, but we would now need 'actually local' versions of this for the deputy worker, right? and a way to distinguish between them. i'll think about it more

docs/design/mono/wasm-threads.md
pavelsavara: I think we don't run finalizers in WASM right now.

docs/design/mono/wasm-threads.md
pavelsavara: There is no "actually local" JS state on the deputy worker. We could say (on logical level), that JS interop doesn't even have to be initialized there.

pavelsavara:
> I think there's a lot of value in having this hybrid option - it reduces latency vs blazor server, and reduces the resources spent on the server and reduces bandwidth spent communicating between client and server. That's great.

This is interesting angle. 

Could we do "blazor server" in the MT (and emscripten started on the worker as in alternative 10), while the UI would just run the same light weight JS protocol ? It would make the user mental model easier. There would be no non-blazor JS interop and no `IJSInProcessRuntime.Invoke` and `IJSUnmarshalledRuntime`.

That could solve blazor, but not non-blazor MT is that acceptable outcome ?

pavelsavara:
> debugging gets WAY more complex

Yes, chrome dev tools experience may be different, because it's will be async with the C# part.
VS debugger would connect to mono as usual. 

> For the specific scenario of "some code needs to block while waiting for a lock to be released":
>     * Ultimately any user-exposed code should not be able to block for a long period of time. 
>     * This code might be in the BCL. Purging the BCL of locks seems like a tall order 
> At the end of the day though I'm not sure I can convince myself that `Monitor.Enter` or `Task.Wait` should work in the browser unless a very low timeout is specified. 

I disagree, I think C# desktop ecosystem is full of locks and MT.  
And one of the reasons people want MT is that this blocking code would work. 
If we say, don't block, they could run the code on single-threaded runtime already. 
But that kills 90% of MT value proposition.

SteveSandersonMS:
> Imagine you press a button, the deputy worker starts crunching data, then you type some text into a textarea and the deputy worker wakes up and immediately obliterates the textarea. A blocked UI thread prevents this, which is good!

You're right that this is a difficult problem, but it's one we've already solved in an even more challenging environment (Blazor Server doing these calls over the internet). We already track old, incoming, and new values separately so we can pick the right final state based on what actually changed, and we have an event dispatch system that takes account of the fact that the UI shape can change in between event handlers being exposed and event calls arriving at the server.

There are some edge cases where developers run into issues with overlapping state changes, but it's rare in practice because we've learned how common UX patterns interact with network latency and what developers usually want to happen.

docs/design/mono/wasm-threads.md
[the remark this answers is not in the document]
pavelsavara: There is more developed version of this "why not aternative 10." [lower in the conversation](https://github.com/dotnet/runtime/pull/91696#issuecomment-1709774964). I will create separate PR/proposal for that.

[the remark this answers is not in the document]
pavelsavara: See https://github.com/dotnet/runtime/pull/91731

docs/design/mono/wasm-threads.md
lambdageek: in single-threaded wasm finalizers run in `mono_runtime_do_background_work` which is scheduled as a background job when `mono_wasm_gc_finalize_notify` is called.  In other words after a GC is finished we queue up an idle task to run finalizers.

docs/design/mono/wasm-threads.md
lambdageek: There's [`GC.TryStartNoGCRegion()`](https://learn.microsoft.com/en-us/dotnet/api/system.gc.trystartnogcregion?view=net-7.0) (no-op on Mono, I think).  It's a weird global mode (so if you have background threads that are quickly allocating, your critical region can still run out of memory), but it might be good enough for some scenarios.

lambdageek:
> Could we do "blazor server" in the MT (and emscripten started on the worker as in alternative 10), while the UI would just run the same light weight "blazor server"" signalR like protocol via JS only? It would make the user mental model easier. 

FWIW this is basically the design that [VS Code wasm-wasi](https://code.visualstudio.com/blogs/2023/06/05/vscode-wasm-wasi) settled on, too, for different reasons (plugin isolation).

docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,243 @@
...
+- `getAssemblyExports` need to bind JS on UI thread, but register on deputy thread
+- hide `SynchronizationContext.Send` and `SynchronizationContext.Post` inside of the generated code.
+    - fast on worker -> worker
+
+## Promise
+- passing Promise should work everywhere.
+- from UI javaScript it would be passed as Task to deputy worker
+- open question: passing JS promise to deputy should be fine. But does the `resolve()` need to block UI thread ?
+
+## Task, Task<T>
```
AustinWise, line 141:
```suggestion
## Task, Task\<T>
```

The \<T> is being treated like it is a HTML tag and is not rendering. This escaping causes it to render correctly.

Elringus:
Excuse me if I'll be off the point here, but just wanted to share my experience in a similar (as I see it) matter.

I'm building a web (PWA) app with C# "backend" compiled to WASM and react frontend communicating in sync/blocking fashion. I've started with .NET 6 where I had to modify the .NET runtime a bit to make it work with the bundlers. I've also added some quality of life stuff on top of the interop, like auto-generating bindings based on C# interfaces, TS types generation, events, etc.

It all worked great in the end, UX was basically like the C# part was another JS library consumed on the frontend. Testing and debuging were straightforward.

At some point I had an idea of moving the C# "backend" to a web worker. I thought this would ultimately dedicate browser's UI thread to the UI-only stuff, like rendering and animation and improve responsiveness (even though there were no real issues with that at the moment).

To minimize breaking changes in the APIs (I've already had a chunky codebase at that point), I've used [comlink](https://github.com/GoogleChromeLabs/comlink) (which uses [JS proxies](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) under the hood) to re-route the interop calls to the webworker. Basically, all the existing APIs were intact, except that they all become async, and blocking calls from JS to C# were no longer possible.

In the end, however, I've rolled all that back to blocking. Main reason was debug experience. And it's not about connecting a debugger or anything, but just understanding the flow control. Complex UI frameworks (like React) have lots going on under the hood and when coupled with an async behaviour on the other "side" it just becomes a real pain to work with. The sweet UX of consuming C# API as a JS library goes away and you have to think about it as a remote server.

Another point is performance/responsiveness. While the UI was not blocking anymore (though it was a hardly noticable improvement in itself), total time spend on interop became noticeably larger, especially in tight back-and-forth loops.

Imho, running the entire C# module on webwoker is not worth it, at least in the case of apps with common "point-click" interfaces (it may be different for something realtime, like games, though in this case input latency will suffer).

However, in case of computation-heavy tasks, it sure make sense to off-load them on worker threads. I hoped that maybe it could be possible to add C# API to spawn worker threads just for that. No interop or any sort of communication with the JS side, just pure C# code running on the webworker. Is this at all possible w/o moving the whole runtime to worker as well?

pavelsavara:
> Excuse me if I'll be off the point here, but just wanted to share my experience in a similar (as I see it) matter.

Thanks for chiming in, much appreciated!

I move the response to side-car design, because your feedback is very relevant [there](https://github.com/dotnet/runtime/pull/91731#issuecomment-1727167722).

docs/design/mono/wasm-threads.md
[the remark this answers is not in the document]
pavelsavara: > More generally I don't understand why running emscripten off the main thread is a non a workable solution. 

@lambdageek here is more detailed answer https://github.com/dotnet/runtime/pull/91731#discussion_r1334249216

pavelsavara:
There could be variation on this design, when 
- emscripten still starts on the UI thread
- dotnet starts on deputy-thread

Benefits
- UI thread could call simple synchronous C functions without dispatch
    - like memory access and allocations
    - probably not any Mono functions, as that would require Mono registered thread and GC barrier
- the UI thread would be mostly doing nothing, and so it could serve others (managing thread pool) without being blocked

Especially compared to [sidecar option D)](https://github.com/dotnet/runtime/pull/91731#discussion_r1334264015)

marek-safar:
Should this rather go to dotnet/design repo?

docs/design/mono/wasm-threads.md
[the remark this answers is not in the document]
pavelsavara: Short answer: because marshaling JSImport/JSExport parameters requires lot on Mono infrastructure. Like allocating `MonoString`, `TaskCompletionSource` and many others. Those are mostly synchronous calls and need to be fast. Trying to dispatch those calls from UI to sidecar thread would lead to terrible latency. Alternatively we could have double proxy for some data types, which is not great either, mostly for GC.

pavelsavara:
Closing in favor of https://github.com/dotnet/designs/pull/301

PR #91295, opened 2023-08-29
[browser] [wasm] Request Streaming upload

campersau:
Uses a `ReadableStream` as fetch request body and reads chunks from .NET `HttpContent` stream.
The feature needs to be enabled via `HttpRequestOptionsKey<bool>("WebAssemblyEnableStreamingRequest")`. A helper could be added to [`WebAssemblyHttpRequestMessageExtensions`](https://github.com/dotnet/aspnetcore/blob/9402bfac90a695bb732dff17ba624801076df77f/src/Components/WebAssembly/WebAssembly/src/Http/WebAssemblyHttpRequestMessageExtensions.cs#L134-L151).
It is only supported using HTTP/2 or higher and is currently only implemented in Chromium (Chrome / Edge) (see https://github.com/dotnet/runtime/issues/36634#issuecomment-1418701286).

~It currently tries to get the buffer size from the browser via [`desiredSize`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableByteStreamController/desiredSize) which in my testing was always `0`. We can also remove this detection and just use the fixed default buffer size.
I also enabled [`autoAllocateChunkSize`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/ReadableStream#autoallocatechunksize) which provides us a [`byobRequest`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableByteStreamController/byobRequest) so we can copy directly into the browsers request buffer. Unfortunately https://github.com/WebAssembly/design/issues/1162 is not yet implemented which means we can't do a zero-byte transfer by passing the `ArrayBuffer` to .NET and avoid any buffers in  .NET.~
The default buffer size is currently set to `65536`. Maybe it should also be made configurable via a `HttpRequestOptionsKey`?

Exceptions from reading the `HttpContent` stream will fail the request with the thrown exception.
The special case for `StringContent` is still preferred and bypasses streaming.

I added some `BrowserHttpHandler_*` tests to `ResponseStreamTest.cs` analog to the `WebAssemblyEnableStreamingResponse` one. Maybe these `BrowserHttpHandler_*` test should be moved in their own file?
I duplicated the `PostAsync_ThrowFromContentCopy_RequsetFails` tests to use remote server, as the `Http2LoopbackServer` does not work in the browser.

Do I need to adjust something regarding `FEATURE_WASM_THREADS`?

Fixes https://github.com/dotnet/runtime/issues/36634

PR #91699, opened 2023-09-06
[browser] [wasm] Refactor Request Streaming to use HttpContent.CopyToAsync

campersau:
Uses a [`TransformStream`](https://developer.mozilla.org/en-US/docs/Web/API/TransformStream) into which the `HttpContent` gets [copied into](https://developer.mozilla.org/en-US/docs/Web/API/WritableStreamDefaultWriter). Sync writes are not allowed like sync reads in response streaming.

Ref: https://github.com/dotnet/runtime/pull/91295#issuecomment-1706919710

### DIFF
--- a/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
+++ b/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
@@ -302,6 +302,7 @@ namespace System.Net.Http.Functional.Tests
             var content = new MultipartFormDataContent();
             content.Add(new StreamContent(new DelegateStream(
                 canReadFunc: () => true,
+                readFunc: (buffer, offset, count) => throw new FormatException(),
                 readAsyncFunc: (buffer, offset, count, cancellationToken) =>
                 {
                     if (remaining > 0)
--- a/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
+++ b/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
@@ -356,18 +356,20 @@ namespace System.Net.Http.Functional.Tests
         }
 
         public static TheoryData CancelRequestReadFunctions
-            => new TheoryData<Func<Task<int>>>
+            => new TheoryData<bool, Func<Task<int>>>
             {
-                () => Task.FromResult(0),
-                () => Task.FromResult(1),
-                () => new TaskCompletionSource<int>().Task,
-                () => throw new FormatException(),
+                { false, () => Task.FromResult(0) },
+                { true, () => Task.FromResult(0) },
+                { false, () => Task.FromResult(1) },
+                { true, () => Task.FromResult(1) },
+                { false, () => throw new FormatException() },
+                { true, () => throw new FormatException() },
             };
 
         [OuterLoop]
         [ConditionalTheory(typeof(PlatformDetection), nameof(PlatformDetection.IsBrowser))]
         [MemberData(nameof(CancelRequestReadFunctions))]
-        public async Task BrowserHttpHandler_StreamingRequest_CancelRequest(Func<Task<int>> readFunc)
+        public async Task BrowserHttpHandler_StreamingRequest_CancelRequest(bool cancelAsync, Func<Task<int>> readFunc)
         {
             var WebAssemblyEnableStreamingRequestKey = new HttpRequestOptionsKey<bool>("WebAssemblyEnableStreamingRequest");
 
@@ -381,8 +383,9 @@ namespace System.Net.Http.Functional.Tests
             req.Content = new StreamContent(new DelegateStream(
                 canReadFunc: () => true,
                 readFunc: (buffer, offset, count) => throw new FormatException(),
-                readAsyncFunc: (buffer, offset, count, cancellationToken) =>
+                readAsyncFunc: async (buffer, offset, count, cancellationToken) =>
                 {
+                    if (cancelAsync) await Task.Delay(1);
                     Assert.Equal(token.IsCancellationRequested, cancellationToken.IsCancellationRequested);
                     if (!token.IsCancellationRequested)
                     {
@@ -393,7 +396,7 @@ namespace System.Net.Http.Functional.Tests
                     {
                         readCancelledCount++;
                     }
-                    return readFunc();
+                    return await readFunc();
                 }));
 
             using (HttpClient client = CreateHttpClientForRemoteServer(Configuration.Http.RemoteHttp2Server))
--- a/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
+++ b/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
@@ -247,6 +247,7 @@ namespace System.Net.Http.Functional.Tests
             int readOffset = 0;
             req.Content = new StreamContent(new DelegateStream(
                 canReadFunc: () => true,
+                readFunc: (buffer, offset, count) => throw new FormatException(),
                 readAsyncFunc: async (buffer, offset, count, cancellationToken) =>
                 {
                     await Task.Delay(1);
@@ -340,19 +341,85 @@ namespace System.Net.Http.Functional.Tests
             req.Options.Set(WebAssemblyEnableStreamingRequestKey, true);
 
             Exception error = new FormatException();
-            var content = new StreamContent(new DelegateStream(
+            req.Content = new StreamContent(new DelegateStream(
                 canSeekFunc: () => true,
                 lengthFunc: () => 12345678,
                 positionGetFunc: () => 0,
                 canReadFunc: () => true,
-                readFunc: (buffer, offset, count) => throw error,
+                readFunc: (buffer, offset, count) => throw new FormatException(),
                 readAsyncFunc: (buffer, offset, count, cancellationToken) => syncFailure ? throw error : Task.Delay(1).ContinueWith<int>(_ => throw error)));
-
-            req.Content = content;
 
             using (HttpClient client = CreateHttpClientForRemoteServer(Configuration.Http.RemoteHttp2Server))
             {
                 Assert.Same(error, await Assert.ThrowsAsync<FormatException>(() => client.SendAsync(req)));
+            }
+        }
+
+        [OuterLoop]
+        [ConditionalFact(typeof(PlatformDetection), nameof(PlatformDetection.IsBrowser))]
+        public async Task BrowserHttpHandler_StreamingRequest_CancelRequest()
+        {
+            var WebAssemblyEnableStreamingRequestKey = new HttpRequestOptionsKey<bool>("WebAssemblyEnableStreamingRequest");
+
+            var req = new HttpRequestMessage(HttpMethod.Post, Configuration.Http.Http2RemoteEchoServer);
+
+            req.Options.Set(WebAssemblyEnableStreamingRequestKey, true);
+
+            using var cts = new CancellationTokenSource();
+            var token = cts.Token;
+            int readNotCancelledCount = 0, readCancelledCount = 0;
+            req.Content = new StreamContent(new DelegateStream(
+                canReadFunc: () => true,
+                readFunc: (buffer, offset, count) => throw new FormatException(),
+                readAsyncFunc: (buffer, offset, count, cancellationToken) =>
+                {
+                    Assert.Equal(token.IsCancellationRequested, cancellationToken.IsCancellationRequested);
+                    if (!token.IsCancellationRequested)
+                    {
+                        readNotCancelledCount++;
+                        cts.Cancel();
+                    }
+                    else
+                    {
+                        readCancelledCount++;
+                    }
+                    return Task.FromResult(1);
+                }));
+
+            using (HttpClient client = CreateHttpClientForRemoteServer(Configuration.Http.RemoteHttp2Server))
+            {
+                TaskCanceledException ex = await Assert.ThrowsAsync<TaskCanceledException>(() => client.SendAsync(req, token));
+                Assert.Equal(token, ex.CancellationToken);
+                Assert.Equal(1, readNotCancelledCount);
+                Assert.Equal(0, readCancelledCount);
+            }
+        }
+
+        [OuterLoop]
+        [ConditionalFact(typeof(PlatformDetection), nameof(PlatformDetection.IsBrowser))]
+        public async Task BrowserHttpHandler_StreamingRequest_Http1Fails()
+        {
+            var WebAssemblyEnableStreamingRequestKey = new HttpRequestOptionsKey<bool>("WebAssemblyEnableStreamingRequest");
+
+            var req = new HttpRequestMessage(HttpMethod.Post, Configuration.Http.RemoteHttp11Server.BaseUri);
+
+            req.Options.Set(WebAssemblyEnableStreamingRequestKey, true);
+
+            int readCount = 0;
+            req.Content = new StreamContent(new DelegateStream(
+                canReadFunc: () => true,
+                readFunc: (buffer, offset, count) => throw new FormatException(),
+                readAsyncFunc: (buffer, offset, count, cancellationToken) =>
+                {
+                    readCount++;
+                    return Task.FromResult(1);
+                }));
+
+            using (HttpClient client = CreateHttpClientForRemoteServer(Configuration.Http.RemoteHttp11Server))
+            {
+                HttpRequestException ex = await Assert.ThrowsAsync<HttpRequestException>(() => client.SendAsync(req));
+                Assert.Equal("TypeError: Failed to fetch", ex.Message);
+                Assert.Equal(1, readCount);
             }
         }
 
--- a/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
+++ b/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
@@ -296,6 +296,7 @@ namespace System.Net.Http.Functional.Tests
             req.Options.Set(WebAssemblyEnableStreamingRequestKey, true);
 
             int size = 1500 * 1024 * 1024;
+            int multipartOverhead = 125 + 4 /* "test" */;
             int remaining = size;
             var content = new MultipartFormDataContent();
             content.Add(new StreamContent(new DelegateStream(
@@ -319,7 +320,7 @@ namespace System.Net.Http.Functional.Tests
             using (HttpResponseMessage response = await client.SendAsync(req))
             {
                 Assert.Equal(HttpStatusCode.OK, response.StatusCode);
-                Assert.Equal((size + 129).ToString(), Assert.Single(response.Headers.GetValues("X-HttpRequest-Body-Length")));
+                Assert.Equal((size + multipartOverhead).ToString(), Assert.Single(response.Headers.GetValues("X-HttpRequest-Body-Length")));
                 // Streaming requests can't set Content-Length
                 Assert.False(response.Headers.Contains("X-HttpRequest-Headers-ContentLength"));
             }
--- a/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
+++ b/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
@@ -246,6 +246,7 @@ namespace System.Net.Http.Functional.Tests
 
             int readOffset = 0;
             req.Content = new StreamContent(new DelegateStream(
+                canReadFunc: () => true,
                 readAsyncFunc: async (buffer, offset, count, cancellationToken) =>
                 {
                     await Task.Delay(1);
@@ -296,7 +297,9 @@ namespace System.Net.Http.Functional.Tests
 
             int size = 1500 * 1024 * 1024;
             int remaining = size;
-            req.Content = new StreamContent(new DelegateStream(
+            var content = new MultipartFormDataContent();
+            content.Add(new StreamContent(new DelegateStream(
+                canReadFunc: () => true,
                 readAsyncFunc: (buffer, offset, count, cancellationToken) =>
                 {
                     if (remaining > 0)
@@ -307,7 +310,8 @@ namespace System.Net.Http.Functional.Tests
                         return Task.FromResult(send);
                     }
                     return Task.FromResult(0);
-                }));
+                })), "test");
+            req.Content = content;
 
             req.Content.Headers.Add("Content-MD5-Skip", "browser");
 
@@ -315,7 +319,7 @@ namespace System.Net.Http.Functional.Tests
             using (HttpResponseMessage response = await client.SendAsync(req))
             {
                 Assert.Equal(HttpStatusCode.OK, response.StatusCode);
-                Assert.Equal(size.ToString(), Assert.Single(response.Headers.GetValues("X-HttpRequest-Body-Length")));
+                Assert.Equal((size + 129).ToString(), Assert.Single(response.Headers.GetValues("X-HttpRequest-Body-Length")));
                 // Streaming requests can't set Content-Length
                 Assert.False(response.Headers.Contains("X-HttpRequest-Headers-ContentLength"));
             }
--- a/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
+++ b/src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
@@ -355,9 +355,19 @@ namespace System.Net.Http.Functional.Tests
             }
         }
 
+        public static TheoryData CancelRequestReadFunctions
+            => new TheoryData<Func<Task<int>>>
+            {
+                () => Task.FromResult(0),
+                () => Task.FromResult(1),
+                () => new TaskCompletionSource<int>().Task,
+                () => throw new FormatException(),
+            };
+
         [OuterLoop]
-        [ConditionalFact(typeof(PlatformDetection), nameof(PlatformDetection.IsBrowser))]
-        public async Task BrowserHttpHandler_StreamingRequest_CancelRequest()
+        [ConditionalTheory(typeof(PlatformDetection), nameof(PlatformDetection.IsBrowser))]
+        [MemberData(nameof(CancelRequestReadFunctions))]
+        public async Task BrowserHttpHandler_StreamingRequest_CancelRequest(Func<Task<int>> readFunc)
         {
             var WebAssemblyEnableStreamingRequestKey = new HttpRequestOptionsKey<bool>("WebAssemblyEnableStreamingRequest");
 
@@ -383,7 +393,7 @@ namespace System.Net.Http.Functional.Tests
                     {
                         readCancelledCount++;
                     }
-                    return Task.FromResult(1);
+                    return readFunc();
                 }));
 
             using (HttpClient client = CreateHttpClientForRemoteServer(Configuration.Http.RemoteHttp2Server))
--- a/src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
+++ b/src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
@@ -223,8 +223,8 @@ namespace System.Net.Http
                         JSObject transformStream = BrowserHttpInterop.CreateTransformStream();
                         try
                         {
-                            promise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, transformStream);
-                            promise = WasmHttpWriteStream.CopyToAsync(promise, transformStream, request.Content, cancellationToken);
+                            Task<JSObject> fetchPromise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, transformStream);
+                            promise = WasmHttpWriteStream.CopyToAsync(fetchPromise, transformStream, request.Content, cancellationToken);
                         }
                         catch
                         {
--- a/src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
+++ b/src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
@@ -133,9 +133,9 @@ namespace System.Net.Http
             List<string> headerNames = new List<string>(headerCount);
             List<string> headerValues = new List<string>(headerCount);
             JSObject abortController = BrowserHttpInterop.CreateAbortController();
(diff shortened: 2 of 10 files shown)


src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
```diff
@@ -307,15 +310,16 @@ public async Task BrowserHttpHandler_StreamingRequest()
...
+            req.Content = content;
 
             req.Content.Headers.Add("Content-MD5-Skip", "browser");
 
             using (HttpClient client = CreateHttpClientForRemoteServer(Configuration.Http.RemoteHttp2Server))
             using (HttpResponseMessage response = await client.SendAsync(req))
             {
                 Assert.Equal(HttpStatusCode.OK, response.StatusCode);
-                Assert.Equal(size.ToString(), Assert.Single(response.Headers.GetValues("X-HttpRequest-Body-Length")));
+                Assert.Equal((size + 129).ToString(), Assert.Single(response.Headers.GetValues("X-HttpRequest-Body-Length")));
```
pavelsavara, line 322: Could you please explain +129 ? Maybe made it a named constant.

src/mono/wasm/runtime/http.ts
```diff
@@ -65,80 +65,32 @@ export function http_wasm_abort_response(res: ResponseExtension): void {
...
-    pull_state.dispose();
-    controller.error(error);
-    pull_state.__pull_promise_control = null;
+export async function http_wasm_request_stream_write(ts: TransformStreamExtension, bufferPtr: VoidPtr, bufferLength: number): Promise<void> {
+    mono_assert(bufferLength > 0, "expected bufferLength > 0");
+    await ts.__writer.ready;
+    // the bufferPtr is pinned by the caller
+    const view = new Span(bufferPtr, bufferLength, MemoryViewType.Byte);
+    const copy = view.slice() as Uint8Array;
+    await ts.__writer.write(copy);
```
pavelsavara, line 74: I would love this to be cancellable.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpInterop.cs
```diff
@@ -28,16 +28,16 @@ internal static partial class BrowserHttpInterop
...
+            JSObject requestStream,
             IntPtr bufferPtr,
             int bufferLength);
 
-        [JSImport("INTERNAL.http_wasm_readable_stream_controller_error")]
-        public static partial void ReadableStreamControllerError(
-            [JSMarshalAs<JSType.Any>] object pullState,
-            Exception error);
+        [JSImport("INTERNAL.http_wasm_request_stream_close")]
+        public static partial Task<JSObject> RequestStreamClose(
```
pavelsavara, line 38: This confuses me, could you please explain why we are returning promise here ?

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
```diff
@@ -329,41 +318,98 @@ static async Task<HttpResponseMessage> Impl(HttpRequestMessage request, Cancella
...
         {
-            fixed (byte* ptr = buffer)
+            cancellationToken.ThrowIfCancellationRequested();
+            using (Buffers.MemoryHandle handle = buffer.Pin())
             {
-                BrowserHttpInterop.ReadableStreamControllerEnqueue(pullState, (nint)ptr, length);
+                await RequestStreamWriteUnsafe(_requestStream, buffer, handle).ConfigureAwait(true);
             }
+
+            static unsafe Task RequestStreamWriteUnsafe(JSObject requestStream, ReadOnlyMemory<byte> buffer, Buffers.MemoryHandle handle)
```
pavelsavara, line 357: Please make it cancellable promise. The pattern is already in the previous code. Or I can try to help :)

src/mono/wasm/runtime/http.ts
pavelsavara: I assume this may throw when
- TPC breaks or server kills the connection mid flight
- we triggered AbortController

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
pavelsavara: When this propagates cancelation exception, or if this was aborted,
how do we propagate that error to the caller of the top level `.SendAsync`, does that work fine ?

pavelsavara (commented):
Nice work thanks! 
I wonder if the transformer could also improve response streaming :)

src/libraries/Common/tests/System/Net/Http/ResponseStreamTest.cs
campersau: This test now sends `MultipartFormDataContent` which adds some more things like boundaries, field name ("test" in this case) and multi part headers. Will add a constant for it.

src/mono/wasm/runtime/http.ts
campersau: Unfortunately not, I need to race all of these with the `fetch_promise_control` e.g. 
```js
await Promise.race([ts.__writer.write(copy), ts.__fetch_promise_control.promise]);
```
or do you have another idea?

```js
var t = new TransformStream();
var w = t.writable.getWriter();
var a = new AbortController();

var f = fetch("", {
    method: "POST",
    body: t.readable,
    duplex: "half",
    signal: a.signal
});

a.abort();

var wp = w.write(new Uint8Array([1, 2, 3]));

var done = Promise.race([wp, f]);

// wp is pending
// done / f is rejected
```

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpInterop.cs
campersau: This returns the actual `fetchResponse` promise.

So it works like this:

1. The `FetchStream` method returns the `TransformStream`.
2. `CopyToAsync` to the `TransformStream`.
3. Close the `TransformStream` and return the actual `fetchResponse` to read the response from.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
campersau: I am planning to do it like this https://github.com/dotnet/runtime/pull/91699#discussion_r1319082854 so it should be automatically cancellable since fetch is cancellable.

src/mono/wasm/runtime/http.ts
```diff
@@ -65,79 +65,36 @@ export function http_wasm_abort_response(res: ResponseExtension): void {
...
-            pull_state.__controller = null;
-            pull_state.__pull_promise_control = null;
-            pull_state.__fetch_promise_control?.reject(error);
-            return Promise.reject(error);
-        }
-    }
-
-    function cancel(error: any) {
-        pull_state.__fetch_promise_control?.reject(error);
+export async function http_wasm_transform_stream_close(ts: TransformStreamExtension, error: Error | undefined): Promise<void> {
```
campersau, line 84: This could be split into two methods, but it currently saves one export/import.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
```diff
@@ -329,41 +326,103 @@ static async Task<HttpResponseMessage> Impl(HttpRequestMessage request, Cancella
...
-                ReadableStreamControllerEnqueueUnsafe(this, _buffer, length);
-            }
-            catch (Exception ex)
+            using (WasmHttpWriteStream stream = new WasmHttpWriteStream(transformStream))
             {
-                BrowserHttpInterop.ReadableStreamControllerError(this, ex);
+                try
+                {
+                    Task copyTask = content.CopyToAsync(stream, cancellationToken);
+                    if (await Task.WhenAny(copyTask, fetchResponsePromise).ConfigureAwait(true) == copyTask)
```
campersau, line 347: Race with fetch to get errors and cancellation because TransformStream does not propagate these.

src/mono/wasm/runtime/http.ts
```diff
@@ -65,79 +65,36 @@ export function http_wasm_abort_response(res: ResponseExtension): void {
...
-    mono_assert(controller, "expected controller");
-    pull_state.__pull_promise_control?.reject(error);
-    pull_state.__fetch_promise_control?.reject(error);
-    pull_state.dispose();
-    controller.error(error);
-    pull_state.__pull_promise_control = null;
+export async function http_wasm_transform_stream_write(ts: TransformStreamExtension, bufferPtr: VoidPtr, bufferLength: number): Promise<void> {
+    mono_assert(bufferLength > 0, "expected bufferLength > 0");
+    mono_assert(ts.__fetch_promise_control, "expected fetch promise control");
+    await Promise.race([ts.__writer.ready, ts.__fetch_promise_control.promise]);
```
campersau, line 77: Race with fetch everywhere to get errors and cancellation because TransformStream does not propagate these.

src/mono/wasm/runtime/http.ts
campersau: I opened a bug report in chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1480250
Because I would expect that any fetch error would properly cancel the `ReadableStream` (which the browser locks) and thus any writer promises would be rejected.

src/mono/wasm/runtime/http.ts
pavelsavara: Let's split it, for clarity sake

src/mono/wasm/runtime/http.ts
pavelsavara: If fetch failed/aborted/whatever it would throw and we could catch it here ? And abort the other promise via it's controller too ?

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
pavelsavara: I wonder if race on TS side could propagate the failure to copyTask, so that we don't have to race here.

src/mono/wasm/runtime/http.ts
pavelsavara: transformer ends have `cancel()` and `abort()`

pavelsavara: We could still wrap those promises with cancelable and have our own controller

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
```diff
@@ -220,12 +220,17 @@ private static async Task<WasmFetchResponse> CallFetch(HttpRequestMessage reques
...
-                        Stream stream = await request.Content.ReadAsStreamAsync(cancellationToken).ConfigureAwait(true);
-                        cancellationToken.ThrowIfCancellationRequested();
-
-                        ReadableStreamPullState pullState = new ReadableStreamPullState(stream, cancellationToken);
-
-                        promise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, ReadableStreamPull, pullState);
+                        JSObject transformStream = BrowserHttpInterop.CreateTransformStream();
+                        try
+                        {
+                            promise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, transformStream);
```
pavelsavara, line 226: let's rename this one to fetchPromise or something.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
```diff
@@ -220,12 +220,17 @@ private static async Task<WasmFetchResponse> CallFetch(HttpRequestMessage reques
...
+                            promise = WasmHttpWriteStream.CopyToAsync(promise, transformStream, request.Content, cancellationToken);
```
pavelsavara, line 227: this second promise is not a proxy of the JS promise and so, it would not work with `BrowserHttpInterop.CancelationHelper`, it will assert/throw I think.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
```diff
@@ -329,41 +326,99 @@ static async Task<HttpResponseMessage> Impl(HttpRequestMessage request, Cancella
...
-            }
-            catch (Exception ex)
+            using (WasmHttpWriteStream stream = new WasmHttpWriteStream(transformStream))
             {
-                BrowserHttpInterop.ReadableStreamControllerError(this, ex);
+                try
+                {
+                    await content.CopyToAsync(stream, cancellationToken).ConfigureAwait(true);
+                    await BrowserHttpInterop.TransformStreamClose(transformStream).ConfigureAwait(true);
+                    return await fetchResponsePromise.ConfigureAwait(true);
```
pavelsavara, line 348: could this await move back up to CallFetch ?

src/mono/wasm/runtime/http.ts
```diff
@@ -65,79 +65,37 @@ export function http_wasm_abort_response(res: ResponseExtension): void {
...
+    // the bufferPtr is pinned by the caller
+    const view = new Span(bufferPtr, bufferLength, MemoryViewType.Byte);
+    const copy = view.slice() as Uint8Array;
+    await Promise.race([ts.__writer.write(copy), ts.__fetch_promise_control.promise]);
```
pavelsavara, line 81: wrap this promise with 
```
return wrap_as_cancelable_promise(async () => {
        return Promise.race([ts.__writer.write(copy), ts.__fetch_promise_control.promise])
    });
```

and make the method synchronous (on JS side only)

pavelsavara: So that `async` doesn't create new promise (without the controller on it)

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
campersau: It depends on what the behavior should be when a request content just waits indefinitely in `SerializeToStreamAsync`. E.g. should it throw a cancellation exception or not.

I checked the normal `SocketsHttpHandler` and it waits indefinitely so this can be removed.

```c#
var cts = new CancellationTokenSource();
cts.CancelAfter(TimeSpan.FromSeconds(5));

using var client = new HttpClient();
client.Timeout = TimeSpan.FromSeconds(15);
await client.PostAsync("https://github.com", new MyHttpContent(), cts.Token); // this will wait indefinitely

class MyHttpContent : HttpContent
{
    protected override Task SerializeToStreamAsync(Stream stream, TransportContext? context)
        => new TaskCompletionSource().Task;

    protected override bool TryComputeLength(out long length)
    {
        length = 0;
        return false;
    }
}
```

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
campersau: Yes, I have rearranged the code.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
```diff
@@ -329,41 +336,81 @@ static async Task<HttpResponseMessage> Impl(HttpRequestMessage request, Cancella
...
-            {
-                int length = await _stream.ReadAsync(_buffer, _cancellationToken).ConfigureAwait(true);
-                ReadableStreamControllerEnqueueUnsafe(this, _buffer, length);
-            }
-            catch (Exception ex)
+            cancellationToken.ThrowIfCancellationRequested();
+            using (Buffers.MemoryHandle handle = buffer.Pin())
             {
-                BrowserHttpInterop.ReadableStreamControllerError(this, ex);
+                await TransformStreamWriteUnsafe(_transformStream, buffer, handle).ConfigureAwait(true);
```
pavelsavara, line 355: add `BrowserHttpInterop.CancelationHelper`

campersau: Yes, currently working on it :)

pavelsavara: Because also C# async creates new Task instances

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
```diff
@@ -220,28 +220,43 @@ private static async Task<WasmFetchResponse> CallFetch(HttpRequestMessage reques
...
+                        {
+                            Task<JSObject> fetchPromise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, transformStream);
+                            ValueTask<JSObject> fetchTask = BrowserHttpInterop.CancelationHelper(fetchPromise, cancellationToken, abortController, null); // initialize fetch cancellation
+
+                            using (WasmHttpWriteStream stream = new WasmHttpWriteStream(transformStream))
+                            {
+                                try
+                                {
+                                    await request.Content.CopyToAsync(stream, cancellationToken).ConfigureAwait(true);
+                                    await BrowserHttpInterop.TransformStreamClose(transformStream).ConfigureAwait(true);
```
pavelsavara, line 233: does close need to be cancelable too ? could it take long time to finish pusing bytes to network in the stage ?

src/mono/wasm/runtime/http.ts
pavelsavara: `ts.__writer.abort(error);` is not locked after all ?

Anyway, that solves the case when we are aborting it. I think the case when TCP fails is still hanging ?

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
campersau: Yes, it does.

src/mono/wasm/runtime/http.ts
campersau: > ts.__writer.abort(error); is not locked after all ?

Everything on the writer is not locked because we created it.
Only the `ReadableStream` is locked because the browser is using it.

I tried forwarding fetch errors to `writer.abort(error)` but it doesn't work in all cases. E.g.

```js
var t = new TransformStream();
var w = t.writable.getWriter();
var a = new AbortController();
var f = fetch("", {
    method: "POST",
    body: t.readable,
    duplex: "half",
    signal: a.signal
});
f.catch(error => w.abort(error));

var wp = w.ready.then(() => { a.abort(); return w.write(new Uint8Array([1, 2, 3])); });

// wp is pending and never resolves
// f will be rejected
```

The `BrowserHttpHandler_StreamingRequest_Http1Fails` would fail because the `write()` promise never resolves as fetch gets rejected during that write...

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
```diff
@@ -133,25 +133,26 @@ private static async Task<WasmFetchResponse> CallFetch(HttpRequestMessage reques
             List<string> headerNames = new List<string>(headerCount);
             List<string> headerValues = new List<string>(headerCount);
             JSObject abortController = BrowserHttpInterop.CreateAbortController();
-            CancellationTokenRegistration? abortRegistration = cancellationToken.Register(() =>
+            CancellationTokenRegistration? abortRegistration = cancellationToken.Register(static s =>
```
campersau, line 136: I noticed that most of the `cancellationToken.Register` usages didn't use the `state`, so I have updated them to use it.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpInterop.cs
```diff
@@ -117,30 +123,27 @@ public static unsafe Task<JSObject> Fetch(string uri, string[] headerNames, stri
             [JSMarshalAs<JSType.MemoryView>] Span<byte> buffer);
 
 
-        public static async ValueTask<T> CancelationHelper<T>(Task<T> promise, CancellationToken cancellationToken, JSObject? abortController, JSObject? fetchResponse)
+        public static async ValueTask CancelationHelper(Task promise, CancellationToken cancellationToken, JSObject? fetchResponse = null)
```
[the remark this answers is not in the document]
pavelsavara: I don't know if that's important. Probably not.

pavelsavara:
I guess the ball is on my side of the court now, I will do full review early next week. Sorry for delay.

src/mono/wasm/runtime/http.ts
```diff
@@ -148,7 +109,7 @@ export function http_wasm_fetch_bytes(url: string, header_names: string[], heade
     return http_wasm_fetch(url, header_names, header_values, option_names, option_values, abort_controller, copy);
 }
 
-export function http_wasm_fetch(url: string, header_names: string[], header_values: string[], option_names: string[], option_values: any[], abort_controller: AbortController, body: Uint8Array | ReadableStream | null): ControllablePromise<ResponseExtension> {
+export function http_wasm_fetch(url: string, header_names: string[], header_values: string[], option_names: string[], option_values: any[], abort_controller: AbortController, body: Uint8Array | ReadableStream | null): Promise<ResponseExtension> {
```
pavelsavara, line 112: this should stay `ControllablePromise`

src/mono/wasm/runtime/http.ts
```diff
@@ -65,80 +64,42 @@ export function http_wasm_abort_response(res: ResponseExtension): void {
...
+}
+
+export function http_wasm_transform_stream_abort(ts: TransformStreamExtension): void {
+    ts.__writer.abort();
+}
 
-    const cancelable_promise = http_wasm_fetch(url, header_names, header_values, option_names, option_values, abort_controller, body);
-    pull_state.__fetch_promise_control = loaderHelpers.getPromiseController(cancelable_promise);
-    return cancelable_promise;
+export function http_wasm_fetch_stream(url: string, header_names: string[], header_values: string[], option_names: string[], option_values: any[], abort_controller: AbortController, body: TransformStreamExtension): Promise<ResponseExtension> {
```
pavelsavara, line 99: return `ControllablePromise`

src/mono/wasm/runtime/http.ts
```diff
@@ -65,80 +64,42 @@ export function http_wasm_abort_response(res: ResponseExtension): void {
...
 
-export function http_wasm_readable_stream_controller_error(pull_state: PullStateExtension, error: Error): void {
-    const controller = pull_state.__controller;
-    mono_assert(controller, "expected controller");
-    pull_state.__pull_promise_control?.reject(error);
-    pull_state.__fetch_promise_control?.reject(error);
-    pull_state.dispose();
-    controller.error(error);
-    pull_state.__pull_promise_control = null;
+export function http_wasm_transform_stream_write(ts: TransformStreamExtension, bufferPtr: VoidPtr, bufferLength: number): Promise<void> {
```
pavelsavara, line 73: `ControllablePromise`

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
```diff
@@ -220,51 +221,71 @@ private static async Task<WasmFetchResponse> CallFetch(HttpRequestMessage reques
...
+                            fetchResponse = await fetchTask.ConfigureAwait(true);
+                        }
                     }
                     else
                     {
                         byte[] buffer = await request.Content.ReadAsByteArrayAsync(cancellationToken).ConfigureAwait(true);
                         cancellationToken.ThrowIfCancellationRequested();
 
-                        promise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, buffer);
+                        Task<JSObject> fetchPromise = BrowserHttpInterop.Fetch(uri, headerNames.ToArray(), headerValues.ToArray(), optionNames, optionValues, abortController, buffer);
```
pavelsavara, line 265: did we lose ability to abort request here ? I mean, not just abandon the promise, but to stop wasting bandwidth.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
```diff
@@ -329,41 +350,82 @@ static async Task<HttpResponseMessage> Impl(HttpRequestMessage request, Cancella
...
-        public async Task PullAsync()
+        private async Task WriteAsyncCore(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken)
         {
-            try
-            {
-                int length = await _stream.ReadAsync(_buffer, _cancellationToken).ConfigureAwait(true);
-                ReadableStreamControllerEnqueueUnsafe(this, _buffer, length);
-            }
-            catch (Exception ex)
+            cancellationToken.ThrowIfCancellationRequested();
```
pavelsavara, line 366: since we switched from pull to push model, this call could come from any thread. And so it will need to be dispatch to the thread/worker which has the JS objects on it.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
```diff
@@ -220,51 +221,71 @@ private static async Task<WasmFetchResponse> CallFetch(HttpRequestMessage reques
...
+                                catch (Exception)
+                                {
+                                    BrowserHttpInterop.TransformStreamAbort(transformStream);
+                                    if (!abortController.IsDisposed)
+                                    {
+                                        BrowserHttpInterop.AbortRequest(abortController);
+                                    }
+                                    try
+                                    {
+                                        using (fetchResponse = await fetchTask.ConfigureAwait(true)) // observe exception
```
pavelsavara, line 246: Could you please explain, I don't follow fully.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
campersau: No, we still do abort the fetch with the `abortController`.
See https://github.com/dotnet/runtime/pull/91699/files#diff-66ce677ded58c3a6a0639d321f4460bb0b2c4bcaeb3216ff3db4381937365bd8R136

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
campersau: This will observe the abort / fetch exception from fetch otherwise this would throw an `UnobservedTaskException`.
The `using` and `BrowserHttpInterop.AbortResponse(fetchResponse)` part isn't really needed because this will always throw.

pavelsavara: Thanks. Is that strategy valid for all types of exceptions here ? Or just cancellation ?

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
campersau: At this point we are only doing two `abortController.Dispose` and `transformStream.Dispose` calls which already dispatch to their `SynchronisationContext`. Also we do use `ConfigureAwait(true)`. Or do will still need to dispatch? Is it only required for `FEATURE_WASM_THREADS`?

E.g.
```c#
#if FEATURE_WASM_THREADS
 transformStream.SynchronizationContext.Send(...);
```

pavelsavara: I mean `TransformStreamWriteUnsafe` is pushing bytes, so wrapping it with `FEATURE_WASM_THREADS` and dispatch should be necessary I think. 

The `FEATURE_WASM_THREADS` is the multi-threaded build of runtime. 
You can build it with `/p:MonoWasmBuildVariant=multithread` 
Make sure to `rm -rf artifacts` before doing so.

pavelsavara: Or you can keep streaming not implemented in multi-threaded build and return false for `SupportsStreamingRequest` under `FEATURE_WASM_THREADS`

pavelsavara: and throw here, just to be sure.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
campersau: Yes, this is valid for all types of exceptions.

src/libraries/System.Net.Http/src/System/Net/Http/BrowserHttpHandler/BrowserHttpHandler.cs
campersau: > I mean TransformStreamWriteUnsafe is pushing bytes, so wrapping it with FEATURE_WASM_THREADS and dispatch should be necessary I think.

Sorry, I was looking at the wrong code.
Yes, it is indeed needed and with https://github.com/dotnet/runtime/pull/91699/commits/0f2ee8f47beb7d92a015c9e3b15489c6d45670d5 the tests passed when building / running with `/p:MonoWasmBuildVariant=multithread` otherwise they would fail.

pavelsavara (approved):

PR #91701, opened 2023-09-06
Remove unnecessary imports

sbomer:
I think these are unnecessary. I noticed the eventpipe test build produces a warning about a duplicate import of liveBuilds.targets, which is already imported by src/tests/Directory.Build.targets.


ViktorHofer:
I don't think I'm the right person to review this change. Don't know enough about this source tree.

LakshanF (approved):
Thanks!

PR #91702, opened 2023-09-06
Try to disable VSIX update service and collect VS update logs

hoyosjs:


jkoritzinsky (approved):

hoyosjs:
Hmm this already happens too late

PR #91712, opened 2023-09-07
fixed problems with how sampling is done and how we suspend to change heap count in DATAS

Maoni0:
+ Moved the sample recording into when we are suspended. The way we were calculating the throughput cost was in `check_heap_count` (which is called right after we restart EE on heap0), we record the `msl_wait_time `(and reset it to 0 for soh/uoh). This is not synchronized with the allocating threads (which are already running at this point). So what can happen is the allocating threads are already accumulated more wait time which is attributed to this GC but it's not within the period we are counting for this GC (and we lose this part for the next GC). For BGC this is incorrect. If an ephemeral GC did happen before the BGC starts, we'd be adding a sample for that GC which is _basically_ correct for that eph GC. But if an eph GC did not happen, we are just adding a random sample which is calculating the tcp as (msl wait + whatever GC that was finished before this BGC) so obviously incorrect.

+ Added gen2 sampling - this was adapted from Peter's gen2 sampling changes. This serves as a backstop in case the existing sampling doesn't ever pick gen2 GC costs. I made the following fixes -
1) changed the way we calculated the median
2) moved where this is calculated to again avoid timing issues
3) made the gen2 samples actually count instead of losing that info if we happen to sample when a gen2 didn't just occur.

+ Changed when `check_heap_count` is called - the previous place is right after a suspension which does not help with spacing the suspension time out (it was "suspend for GC" then "immediately suspend to change heap count"). And it caused a problem with BGC which was it always tried to change heap count when it couldn't because BGC was in progress. I changed this to be on a timeout to intentionally space the suspensions out. Now most of the time, heap count changes happen due to this time out. If we are really in a situation where GCs are happening too quickly and we return from waiting on the `ee_suspend_event` due to a GC started, we change the heap count right before we do a GC. So this also helps with the BGC problem.

### DIFF
--- a/src/coreclr/gc/gc.cpp
+++ b/src/coreclr/gc/gc.cpp
@@ -2940,6 +2940,7 @@ BOOL gc_heap::should_expand_in_full_gc = FALSE;
 int gc_heap::dynamic_adaptation_mode = dynamic_adaptation_default;
 gc_heap::dynamic_heap_count_data_t SVR::gc_heap::dynamic_heap_count_data;
 uint64_t gc_heap::last_suspended_end_time = 0;
+GCEvent gc_heap::bgc_heap_count_change_event;
 
 #ifdef STRESS_DYNAMIC_HEAP_COUNT
 int gc_heap::heaps_in_this_gc = 0;
@@ -7002,6 +7003,12 @@ BOOL gc_heap::create_thread_support (int number_of_heaps)
     {
         goto cleanup;
     }
+#ifdef DYNAMIC_HEAP_COUNT
+    if (!bgc_heap_count_change_event.CreateOSAutoEventNoThrow (FALSE))
+    {
+        goto cleanup;
+    }
+#endif //DYNAMIC_HEAP_COUNT
     if (!gc_t_join.init (number_of_heaps, join_flavor_server_gc))
     {
         goto cleanup;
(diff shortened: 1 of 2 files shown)


src/coreclr/gc/gc.cpp
```diff
@@ -7067,16 +7064,63 @@ void gc_heap::gc_thread_function ()
...
+                        dprintf (6666, ("changing heap count due to timeout"));
+                        check_heap_count();
+                    }
+                }
+#endif //DYNAMIC_HEAP_COUNT
+
+                if (gradual_decommit_in_progress_p)
+                {
+                    decommit_lock.Enter ();
+                    gradual_decommit_in_progress_p = decommit_step (DECOMMIT_TIME_STEP_MILLISECONDS);
```
mangod9, line 7104: is there a ceiling for `DECOMMIT_TIME_STEP_MILLISECONDS`? I worry it might be in the lock for at most that much time?

src/coreclr/gc/gc.cpp
```diff
@@ -25152,266 +25276,262 @@ void gc_heap::recommission_heap()
...
+    step_up = min (step_up, n_max_heaps - extra_heaps - n_heaps);
 
-    dprintf (9999, ("sample %d: soh_msl_wait_time: %zd, uoh_msl_wait_time: %zd, elapsed_between_gcs: %zd, gc_elapsed_time: %d, heap_size: %zd MB",
-        dynamic_heap_count_data.sample_index,
-        sample.soh_msl_wait_time,
-        sample.uoh_msl_wait_time,
-        sample.elapsed_between_gcs,
-        sample.gc_elapsed_time,
-        sample.heap_size/(1024*1024)));
+    // on the way down, we essentially divide the heap count by 1.5
```
mangod9, line 25394: on the way down it steps down by a third?

src/coreclr/gc/gc.cpp
```diff
@@ -25152,266 +25276,262 @@ void gc_heap::recommission_heap()
...
+        GCToEEInterface::SuspendEE(SUSPEND_FOR_GC_PREP);
+        dprintf (9999, ("h0 suspended EE in check"));
 
-        // if we are adjusting down, make sure we adjust lower than the lowest uoh msl heap
-        if ((new_n_heaps < n_heaps) && (dynamic_heap_count_data.lowest_heap_with_msl_uoh != -1))
+#ifdef BACKGROUND_GC
+        if (gc_heap::background_running_p())
         {
-            new_n_heaps = min (dynamic_heap_count_data.lowest_heap_with_msl_uoh, new_n_heaps);
+            // background GC is running - reset the new heap count
```
mangod9, line 25523: not sure I follow this, so if BGC is in progress there is no change in heap count and it will be recalculated on next iteration?

src/coreclr/gc/gc.cpp
Maoni0: this is unchanged from before.

src/coreclr/gc/gc.cpp
Maoni0: yes, also unchanged from before.

src/coreclr/gc/gc.cpp
Maoni0: exactly. also unchanged from before.

mangod9 (approved):

PR #91717, opened 2023-09-07
Improve binder gen binding logic (init, member binding, polymorphism)

### LINKED ISSUE #90974: Config generator wrongly emits init-related exception in bind operations
This code without the generator:
```C#

Environment.SetEnvironmentVariable("Value", "42");
IConfiguration b = new ConfigurationBuilder().AddEnvironmentVariables().Build();

Base d = new Derived();
b.Bind(d);
Console.WriteLine(d.Value);

internal abstract class Base
{
    public int Value { get; set; }
}

internal sealed class Derived : Base { }
```
prints `42`. 

With the config generator, that same code throws an exception, because the generator emits this:
```C#
public static void Bind_Base(this IConfiguration configuration, object? obj)
{
    throw new InvalidOperationException("Cannot create instance of type '<global namespace>.Base' because it is missing a public instance constructor.");
}
```

layomia:
Fixes https://github.com/dotnet/runtime/issues/90974. RC-2 candidate.


src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -360,59 +370,26 @@ private bool IsParsableFromString(ITypeSymbol type, out StringParsableTypeKind t
...
-            }
-
-            private EnumerableSpec? CreateArraySpec(IArrayTypeSymbol arrayType)
+            private EnumerableSpec? CreateArraySpec(IArrayTypeSymbol arrayTypeSymbol)
             {
-                if (!TryGetTypeSpec(arrayType.ElementType, Diagnostics.ElementTypeNotSupported, out TypeSpec elementSpec))
+                ITypeSymbol elementTypeSymbol = arrayTypeSymbol.ElementType;
+                if (!MemberTypeIsBindable(arrayTypeSymbol, elementTypeSymbol, Diagnostics.ElementTypeNotSupported, out TypeSpec elementTypeSpec) ||
+                    // We want a BindCore method for List<TElement> as a temp holder for the array values
+                    GetOrCreateTypeSpec(_typeSymbols.List.Construct(elementTypeSymbol)) is not EnumerableSpec listTypeSpec)
```
tarekgh, line 378: >GetOrCreateTypeSpec(_typeSymbols.List.Construct(elementTypeSymbol)) is not EnumerableSpec listTypeSpec [](http://example.com/codeflow?start=20&length=102)

When this is possible to be true?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -360,59 +370,26 @@ private bool IsParsableFromString(ITypeSymbol type, out StringParsableTypeKind t
...
-                }
-
-                if (arrayType.Rank > 1)
-                {
-                    RegisterUnsupportedType(arrayType, Diagnostics.MultiDimArraysNotSupported);
-                    return false;
-                }
-
-                return true;
+                return TryRegisterTypeForBindCoreGen(listTypeSpec) && TryRegisterTypeForBindCoreGen(spec) ? spec : null;
```
tarekgh, line 392: >TryRegisterTypeForBindCoreGen(listTypeSpec) [](http://example.com/codeflow?start=23&length=43)

should you try this check before creating `spec`?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -724,6 +691,7 @@ private DictionarySpec CreateDictionarySpec(INamedTypeSymbol type, ITypeSymbol k
                                 ConfigurationKeyName = propertySpec.ConfigurationKeyName,
                             };
 
+                            canInitAllCtorParams &= propertySpec.Type.CanInstantiate;
```
tarekgh, line 694: > canInitAllCtorParams &= propertySpec.Type.CanInstantiate; [](http://example.com/codeflow?start=27&length=58)

I am not seeing you are using `canInitAllCtorParams` anywhere else.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
```diff
@@ -82,36 +81,62 @@ private void EmitGetCoreMethod()
...
+                                    EmitBindingLogic(complexType, Identifier.instance, Identifier.configuration, InitializationKind.Declaration);
+                                    _writer.WriteLine($"return {Identifier.instance};");
+                                }
+                                else if (type is ObjectSpec { InitExceptionMessage: string exMsg })
+                                {
+                                    _writer.WriteLine($@"throw new {Identifier.InvalidOperationException}(""{exMsg}"");");
+                                }
+                                else
+                                {
+                                    _writer.WriteLine($"// Cannot create an instance of {type.DisplayString}.");
```
tarekgh, line 117: >_writer.WriteLine($"// Cannot create an instance of {type.DisplayString}."); [](http://example.com/codeflow?start=36&length=76)

how this is useful to emit in the code?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
```diff
@@ -627,22 +650,34 @@ private void EmitPopulationImplForEnumerableWithAdd(EnumerableSpec type)
...
+                        }
+                        break;
+                    case ComplexTypeSpec { CanInstantiate: true } complexType:
+                        {
+                            EmitBindingLogic(complexType, Identifier.value, Identifier.section, InitializationKind.Declaration);
+                            _writer.WriteLine($"{addExpr}({Identifier.value});");
+                        }
+                        break;
+                    default:
+                        break;
```
tarekgh, line 680: is this needed at all?

tarekgh (approved):
Added a few comments, LGTM otherwise.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -186,42 +186,44 @@ static bool IsAccessibleFromGenBinders(ITypeSymbol type)
...
             {
-                if (typeSpec.NeedsMemberBinding)
+                if (type.HasBindableMembers)
                 {
-                    RegisterTypeForMethodGen(MethodsToGen_CoreBindingHelper.BindCore, typeSpec);
+                    RegisterTypeForMethodGen(MethodsToGen_CoreBindingHelper.BindCore, type);
+                    return true;
                 }
+
+                return true;
```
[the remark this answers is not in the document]
layomia: Yes thanks. The difference after I correct this is

- if the type appears as a member in the graph, we'd skip binding for that property
- if it's a root config type, we'd still generate an interceptor, but it will be a no op (e.g. `public static void Bind_EmptyType(this IConfig, object) => { }`)

I'll add a test.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
layomia: Given the new changes, I expect these methods to always return true. Replacing with debug asserts.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
layomia: Never; will use debug assert.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
layomia: Added it for UX but the "not supported" state should be clear from looking at the empty code block. I'll remove it.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Emitter.cs
```diff
@@ -64,18 +64,27 @@ public void Emit()
...
                 InitializationKind initKind,
                 Action<string>? writeOnSuccess = null)
             {
-                Debug.Assert(type.CanInitialize);
-
-                if (!type.NeedsMemberBinding)
+                if (!type.HasBindableMembers)
                 {
-                    EmitObjectInit(memberAccessExpr, initKind);
+                    if (initKind is not InitializationKind.None)
```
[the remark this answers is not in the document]
layomia: I'm finding that this method & related logic for "complex" type binding are too cluttered. They handle multiple edge cases which makes it hard to make a concise description.

I'll follow-up with a small refactoring PR to separate/silo the concerns. This will help me tell a better story.
- do we need to init?
- how are we assigning to the target object? (declaration for new instances returned by `Get`, or to bind to object members?)
- is this a value type (copy assignment)?
- etc.

layomia:
Queuing after https://github.com/dotnet/runtime/pull/91657 & needed follow-up PR are considered for RC-2 port.

layomia:
@ericstj thanks, I'll backport manually.

PR #91595, opened 2023-09-05
Validate hardcoded offsets to PROFILE_PLATFORM_SPECIFIC_DATA struct

tomeksowi:
These offsets are now in asmconstants.h validated against the C struct with static asserts.

This refactoring is a follow-up to [discussion](https://github.com/dotnet/runtime/pull/91313#discussion_r1312507740) under PR #91313.

Note: I didn't define offset constants for i386 because its asm profile stubs are based on a sequence of `push`es rather than field offsets so they wouldn't use them.

Part of #84834
cc @wscho77 @HJLeee @JongHeonChoi @t-mustafin @alpencolt @gbalykov @clamp03

PR #91722, opened 2023-09-07
[LoongArch64-RISCV64] Refactor the profiler for LoongArch64 and also fix some errors for RISCV64.

shushanhf:

Refactor the profiler for LoongArch64 and also fix some errors for RISCV64  after #91595


tomeksowi (approved):
RISC-V changes look good to me, thanks for the stack alignment fix.

src/coreclr/vm/riscv64/asmconstants.h
```diff
@@ -255,7 +255,10 @@ ASMCONSTANTS_C_ASSERT(CallCountingStubData__TargetForThresholdReached == offseto
 #define PROFILE_LEAVE    2
 #define PROFILE_TAILCALL 4
 
-#define SIZEOF__PROFILE_PLATFORM_SPECIFIC_DATA 312
+// NOTE: this should be 16-byte aligned as stack size.
+#define SIZEOF__PROFILE_PLATFORM_SPECIFIC_DATA  0x140
```
tomeksowi, line 259: Suggestion:
```C
ASMCONSTANTS_C_ASSERT(SIZEOF__PROFILE_PLATFORM_SPECIFIC_DATA % 16 == 0)
```

shushanhf: Thanks very much.
Very good.
I will add it.

clamp03 (approved):
Thank you.

src/coreclr/jit/codegenloongarch64.cpp
```diff
@@ -8677,10 +8649,78 @@ void CodeGen::genProfilingEnterCallback(regNumber initReg, bool* pInitRegZeroed)
...
+    assert((helper == CORINFO_HELP_PROF_FCN_LEAVE) || (helper == CORINFO_HELP_PROF_FCN_TAILCALL));
+
+    if (!compiler->compIsProfilerHookNeeded())
+    {
+        return;
+    }
+
+    compiler->info.compProfilerCallback = true;
+
+    if (compiler->compProfilerMethHndIndirected)
```
shushanhf, line 8702: @jakobbotsch 
Can you give me some help?
`if (compiler->compProfilerMethHndIndirected)` , is this used under AOT's relocation ? which will be recomputed before running liking AOT ?

shushanhf:
If don‘t add the `update the args registers within the GenerateProfileHelper`， the origin patch of this PR is OK while you run the `RunTest` within the `src/tests/profiler/unittest/inlining.cs`.

The `update the args registers within the GenerateProfileHelper` is used to fix the error about the `args` within the `Main(string[] args)`.

Now the related profiler tests is OK.

@jkotas 
Could you please review this PR?
Thanks very much!

src/coreclr/jit/codegenloongarch64.cpp
jakobbotsch: It looks like dead code nowadays, I think you can just assert that it is false.

shushanhf: Thanks very for your help!

jkotas (approved):
LGTM

davmason (approved):

PR #90392, opened 2023-08-11
[browser] Allow JS modules to be injected as pre-loaded assets

pavelsavara:
- new `asset.moduleExports` field which could be pre-populated with ES6 exports of runtime modules
- demo in `browser-minimal-config\main.js`
- allow promise in `asset.buffer`

To address user feedback 
Fixes https://github.com/dotnet/runtime/issues/89510

PR #91731, opened 2023-09-07
[browser] WASM sidecar - multi-threading proposal

pavelsavara:
This is (anternative 10. like) "sidecar" proposal on how to enable threads on WASM, while allowing for blocking `.Wait`, `lock() { ... }` and similar from user C# code.

Feedback and questions are welcome.

Contributes to https://github.com/dotnet/runtime/issues/85592

Alternative to https://github.com/dotnet/runtime/pull/91696

### DIFF
--- a/docs/design/mono/wasm-threads.md
+++ b/docs/design/mono/wasm-threads.md
@@ -1,0 +1,158 @@
+# Multi-threading on browser
+
+## Goals
+ - CPU intensive workloads on dotnet thread pool
+ - enable blocking .Wait APIs from C# user code
+     - Current public API throws PNSE for it
+     - This is core part on MT value proposition.
+     - If people want to use existing MT code-bases, most of the time, the code is full of locks. People want to use existing code as is.
+ - allow HTTP and WS C# APIs to be used from any thread
+     - Underlying JS object have thread affinity
+ - don't change/break single threaded build. †
+ - don't try to block on UI thread.
+
+<sub><sup>† Note: all the text below discusses MT build only, unless explicit about ST build.</sup></sub>
+
+## Design proposal TL;DR - Alternative 10
+20. MAUI/BlazorWebView ... is the same thing
+21. execute whole WASM runtime on a worker. Blocking is fine there.
+22. The UI thread will only have `blazor.server.js` with small modifications.
+23. This needs new marketing name! -> `WASM server` ???
+
+# Detailed design
+
+## Blazor - what changes vs MAUI server
+- it still needs to compile for WASM target
+- it will have threadpool and blocking C# `.Wait` and `lock()`
+- it would not have Socket or DB connection
+
+## Blazor - what changes vs "Blazor WASM"
+- "onClick" would be asynchronous, same way as in Blazor server
+- Blazor's [`IJSInProcessRuntime.Invoke`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.jsinterop.ijsinprocessruntime.invoke) would not be available
+- Blazor's [`IJSUnmarshalledRuntime`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.jsinterop.ijsunmarshalledruntime) would not be available
+- no JavaScript APIs on `globalThis.Blazor` or `globalThis.MONO`
+- no runtime JS interop
+- no runtime on UI thread
+- Blazor united: progressive switch from remote server to WASM server, after WASM binaries arrived
+
+## Blazor startup
+- we will need to start emscripten on a worker and `postMessage` there all the configuration.
+    - this is possibly the biggest effort
+- probably no `Blazor.start`, `start.withResourceLoader` with returning `Response` object
+- we may keep `dotnet.js` loader to run in the UI
+    - TODO: prototype
+
+### Blazor `renderBatch`
+- streaming bytes - as Blazor server does
+    - we use [RenderBatchWriter](https://github.com/dotnet/aspnetcore/blob/045afcd68e6cab65502fa307e306d967a4d28df6/src/Components/Shared/src/RenderBatchWriter.cs) in the WASM
+    - we use `blazor.server.js` to render it
+    - preferred if fast enough
+    - could be ok to render batch by another thread while UI is rendering
+- plan b) use existing `SharedArrayBuffer`
+    - WASM would:
+        - stop the GC
+        - send the "read-memory-now"
+        - await for "done" message
+        - enable GC
+    - UI could duplicate `conv_string` and `unbox_mono_obj` dark magic and use it
+
+## Blazor JS interop
+- will work as usual on "Blazor server"
+
+## runtime JS Interop
+- would be available only to workers, but it would not have access to UI thread and DOM
+- could call sync and async calls both directions, promises and callbacks
+
+## HTTP & WS
+- would be running on a worker, will be accessible from all C# threads
+    - we will overcome thread affinity internally
+
+## C# Thread
+- could block on synchronization primitives
+- without JS interop. calling JSImport will PNSE.
+
+## C# Threadpool Thread
+- could block on synchronization primitives
+- without JS interop. calling JSImport will PNSE.
+
+## Blocking problem
+1. If you have multithreading, any thread might need to block while waiting for any other to release a lock.
+     - locks are in the user code, in nuget packages, in Mono VM itself
+     - there are managed and un-managed locks
+     - in single-threaded build of the runtime, all of this is NOOP. That's why it works on UI thread.
+2. UI thread in the browser can't synchronously block
+     - you can spin-lock but it's bad idea.
+         - Deadlock: when you spin-block, the JS timer loop and any messages are not pumping. But code in other threads may be waiting for some such event to resolve.
+         - It eats your battery
+         - Browser will kill your tab at random point (Aw, snap).
+         - It's not deterministic and you can't really test your app to prove it harmless.
+     - all the other threads/workers could synchronously block
+3. JavaScript engine APIs and objects have thread affinity. The DOM and few other browser APIs are only available on the main UI "thread"
+     - and so, you need to have some way how to talk to UI
+
+This is the main reason why we can't run MT dotnet also on UI thread.
+
+## Alternatives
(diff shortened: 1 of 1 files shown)


docs/design/mono/wasm-threads.md
```diff
@@ -0,0 +1,158 @@
...
+- would be available only to workers, but it would not have access to UI thread and DOM
+- could call sync and async calls both directions, promises and callbacks
+
+## HTTP & WS
+- would be running on a worker, will be accessible from all C# threads
+    - we will overcome thread affinity internally
+
+## C# Thread
+- could block on synchronization primitives
+- without JS interop. calling JSImport will PNSE.
```
kg, line 72: I think leaving out JSImport makes sense as a 1.0 version of this, but it's meaningful to support JSImport so that the "server" here can utilize browser APIs, to compensate for the lack of native APIs. Like for example if the "server" is able to use an offscreen canvas it would be able to access hardware accelerated rendering, which would enable developers to use stuff like browser machine learning APIs.

kg:
This is a robust proposal, I could see either this or the original one working really well. I think it ends up depending on the user scenarios each proposal is best at solving.

pavelsavara:
@Elringus thanks for chiming in, appreciated. I moved [our conversation](https://github.com/dotnet/runtime/pull/91696#issuecomment-1726700990) here.

> To minimize breaking changes in the APIs, I've used [comlink](https://github.com/GoogleChromeLabs/comlink)
> except that they all become async, and blocking calls from JS to C# were no longer possible.

Something like that is what I'm considering in this "side-car" design.
This is approach is dispatch to UI thread via JS. 

[The other approach](https://github.com/dotnet/runtime/pull/91696) describes how to do the call dispatch via C#/managed/emscripten.

During this research, I realized that out of 4 combinatios [JSImport/JSExport] x [Sync/Async] the only thing which need to be spin-waiting is sync JSExport (and internal variations). The other sync scenario, sync JSImport (and internal methods) could be truly blocking via Atomics. 

There are hairy details of managing C# stack. Also `JSObject`/`ManagedObject` proxies which I'm thinking about now, because it would be better to have proxy directly created from managed call in the UI thread, rather than have comlink-like proxy of a proxy.

BTW: how does the comlink deal with nested structures ? Like

```
    setModuleImports("main.js", {
        Sample: {
            Test: {
                displayMeaning: () => { return something }
            }
        }
    });
```

> In the end, however, I've rolled all that back to blocking. Main reason was debug experience. 

This is valuable feedback. 

With this design, I think that

 - All the legacy interop and memory functions will just throw.
 - There are few sync APIs of the `dotnet.create()` which need to stay working. I consider making them spin-blocking.
 - The sync JSExport would do spin-wait, and complain loudly into browser console, like emscripten does.
Same for callback passed on JSImport .There should be MSBuild switch that makes it throw instead. I'm not sure if that will be opt-in or opt-out.

> Another point is performance/responsiveness. While the UI was not blocking anymore (though it was a hardly noticable improvement in itself), total time spend on interop became noticeably larger, especially in tight back-and-forth loops.

I have similar concerns about dispatch between C# threads. We need to measure it.
I hope that Atomic wait would be faster than `self.postMesssage`
I'm not sure if that's reason enough to skip `postMesssage` completely and go full `SharedArrayBuffer` for all messages.

> Imho, running the entire C# module on webwoker is not worth it

The main motivation of moving out off the UI thread, is to allow C# to have consistent and blocking .Wait , `lock()` etc on all threads.

We tried to do threads in Net8 and one of the reasons we didn't deliver that, are the the implications treating one of the threads differently. That's confusing to the developer and difficult in existing code-bases.

Allowing blocking everywhere would also allow us to finally implement C# crypto via browser's `subtle`.

> However, in case of computation-heavy tasks, it sure make sense to off-load them on worker threads.

That's already possible with Net8 experimental workload. It just has some rough edges, like running out of emscripten's thread pool and maybe leaking threads.

pavelsavara:
[partytown](https://partytown.builder.io/how-does-partytown-work) have neat trick, they use synchronous [XHR](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests#example_http_synchronous_request) instead of spin-wait. 
That's deprecated, but better because it doesn't eat battery.

Elringus:
> BTW: how does the comlink deal with nested structures ?

Iirc, it uses deep clone, so yes. Alternatively, they also use [transferable](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects) for types that support it and JS Proxies for callbacks (as it's not possible to clone or transfer functions to another context/thread in JS).

One more thing I've forgot to mention is the frontend behavior with this kind of setup, which was another reason I've switched back to full blocking. Imagine a button with hover and active animations. With blocking, when user clicks the button, the UI thread is blocked until the underlying code is fully executed and the button remains in active (pressed) visual state during this time. While in theory it may be even desirable to not block the UI here (to get rid of slight stutter on button click), in reality UX becomes weird when user clicks the button, sees the active state, which then immediately returns to normal and then nothing happens (while the code is executed async). This also open doors to all kind of unspecified behavior, as user may accidentally click twice or interact with something else, while the call is not finished. The solution here would be blocking the interaction and authoring special CSS styles for the time when the async calls is executing, but that's additional layers of complexity.

> That's already possible with Net8 experimental workload. It just has some rough edges, like running out of emscripten's thread pool and maybe leaking threads.
Oh, so that's possible without all the limitations with sync/blocking calls from JS to C# (`JSExport`)? I wasn't able to test it, as it doesn't seem to support module injection I'm using when creating the runtime (https://github.com/dotnet/runtime/pull/90392), but it'll be possible in the future, this would completely cover all the needs I have for threading.

Elringus:
> I think that he is right and that this is not dotnet runtime issue to solve.

Sure, but if there are no useful cases for this mode, no one will bother to solve it. Even worse, users will spend time and resources adapting to this mode only to later discard all the work. Maybe some kind of warning explaining all the inherent pitfalls and additional complexity would help here.

> You are on your own, we would be happy to hear about trouble when you try that, but we probably would not try to fix it.

I mean, does it run the main/entry .NET runtime on webworker (in which case all the blocking interop limitations apply, which is not useful in my case) or it runs in the same way as ST, but has an API to dispatch tasks to worker threads?

pavelsavara:
> additional complexity would help here.

This is for advanced user like authors of higher level UI frameworks, like Blazor and Uno.

> I mean, does it run the main/entry .NET runtime on webworker 

Net 8, no emscripten doesn't work on UI worker with MT build in Net8. I'm fixing it for Net9.

> but has an API to dispatch tasks to worker threads?

You can use thread pool. 

JSImport needs thread affinity of the caller and we have not exposed API for that. JSExport has the blocking problem.

docs/design/mono/wasm-threads.md
pavelsavara: To be able to do that we need JSImport/JSExport interop on the UI thread.

There are 3 choices how to do that
A) dispatch via managed as described in [deput-worker proposal](https://github.com/dotnet/runtime/pull/91696)
B) dispatch via just JS, which means we will have JS proxy in UI thread of C# proxy in side-car worker. [Comlink ](https://github.com/GoogleChromeLabs/comlink) style. We could write out own and make in also spin-blocking the UI. But this is double dispatch on each call. Hopping over 2 threads and their main loop. 
C) dispatch just the `mono_wasm_bind_js_function` and `mono_wasm_bind_cs_function` and then do the JS side of the marshaling in the UI thread. But the implementation is heavily dependent on 
- memory (that's easy one)
- shared code `invoke-cs.ts`, `invoke-js.ts` etc
- but that is dependent on Mono C methods and emscripten methods. Most of them are synchronous. 
   - emscripten: stack alllocation, memory views (growing)
   - GC and JS handles (I guess those should be UI thread local)
   - various JS helpers (logging, exception handling, asserts)
   - mono: string marshaling, gc roots
   - mono: call dispatch to managed code: instantiate `TaskCompletionSource` etc

pavelsavara: D) we could further move the boundary and have it between full `dotnet.runtime.js` on the UI thread and emscripten on the side-car. That would narrow it down to
- memory view events
- stackalloc/alloc/free
- mono_wasm_assembly_find_class
- mono_wasm_assembly_find_method
- mono_wasm_deregister_root
- mono_wasm_invoke_method_bound
- mono_wasm_invoke_method_ref
- mono_wasm_register_root
- mono_wasm_runtime_run_module_cctor

kg: I was proposing that the sidecar worker be able to access JS inside the sidecar context, to be clear. Not JS objects from the main thread. No remoting, so it can still be synchronous.

pavelsavara: We need some JS interop anyway for Blazor, the current surface is

- INTERNAL.loadLazyAssembly
- INTERNAL.loadSatelliteAssemblies
- Blazor._internal.dotNetCriticalError
- Blazor._internal.renderBatch
- globalThis.console.debug
- globalThis.console.error
- globalThis.console.info
- globalThis.console.warn
- Blazor._internal.endInvokeDotNetFromJS
- Blazor._internal.getApplicationEnvironment
- Blazor._internal.getPersistedState
- Blazor._internal.invokeJSJson
- Blazor._internal.receiveByteArray
- receiveHotReloadAsync
- MONO.getI16
- MONO.getI32
- MONO.getF32
- BINDING.js_string_to_mono_string
- BINDING.unbox_mono_obj
- BINDING.conv_string
- ICall InvokeJS

pavelsavara:
> [partytown](https://partytown.builder.io/how-does-partytown-work) have neat trick, they use synchronous [XHR](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests#example_http_synchronous_request) + service worker instead of spin-wait. That's deprecated, but better because it doesn't eat battery.

But it only works on Firefox https://wpt.fyi/results/service-workers/service-worker/fetch-request-xhr-sync.https.html?label=experimental&label=master&aligned

pavelsavara:
closing in favor of https://github.com/dotnet/designs/pull/301

PR #91735, opened 2023-09-07
Set DOTNET_PreferredVectorBitWidth for x86/x64

kunalspathak:
Set the `DOTNET_PreferredVectorBitWidth=512` for Antigen process so it can generate Avx512 intrinsic methods.

### DIFF
--- a/src/coreclr/scripts/antigen_run.py
+++ b/src/coreclr/scripts/antigen_run.py
@@ -213,6 +213,12 @@ def main(main_args):
         # Run tool such that issues are placed in a temp folder
         with TempDir() as temp_location:
             antigen_log = path.join(temp_location, get_antigen_filename(tag_name))
+            env_variable = {}
+            if 'arm' not in run_duration:
+                # For x86 and x64, set this environment variable so Antigen process would
+                # get the hardware acceleration of Avx512 and through that, Antigen will
+                # generate code containing Avx512 specific intrinsic methods.
+                env_variable["DOTNET_PreferredVectorBitWidth"] = "512"
             run_command([path_to_tool, "-c", path_to_corerun, "-o", temp_location, "-d", str(run_duration)], _exit_on_fail=True, _output_file= antigen_log)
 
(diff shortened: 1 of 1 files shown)


kunalspathak:
@BruceForstall @jakobbotsch

src/coreclr/scripts/antigen_run.py
```diff
@@ -213,6 +213,12 @@ def main(main_args):
         # Run tool such that issues are placed in a temp folder
         with TempDir() as temp_location:
             antigen_log = path.join(temp_location, get_antigen_filename(tag_name))
+            env_variable = {}
+            if 'arm' not in run_duration:
```
jakobbotsch, line 217:
```suggestion
            if 'x86' in coreclr_args.run_configuration or 'x64' in coreclr_args.run_configuration:
```

kunalspathak: i thought about this initially, but `arm` checked looked reasonable to me. I will change it back to x86/x64.

src/coreclr/scripts/antigen_run.py
```diff
@@ -213,6 +213,12 @@ def main(main_args):
...
+                # For x86 and x64, set this environment variable so Antigen process would
+                # get the hardware acceleration of Avx512 and through that, Antigen will
+                # generate code containing Avx512 specific intrinsic methods.
+                env_variable["DOTNET_PreferredVectorBitWidth"] = "512"
```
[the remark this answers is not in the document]
kunalspathak: ah yes.

src/coreclr/scripts/antigen_run.py
jakobbotsch: It was mainly the `run_duration` => `run_configuration` change that is important, but I think it's better to be inclusive rather than exclusive for something like this.

jakobbotsch (approved):

BruceForstall (approved):

Issue #83658, opened 2023-03-20
Test failure baseservices/threading/regressions/2164/foreground-shutdown/foreground-shutdown.cmd


v-wenyuxu:
Failed in Run: [runtime-coreclr outerloop 20230318.3](https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=209759)
 
 **Failed tests:**
 ```
 R2R-CG2 windows x64 Checked no_tiered_compilation @ Windows.10.Amd64.Open
     - baseservices/threading/regressions/2164/foreground-shutdown/foreground-shutdown.cmd
 ```
 
 **Error message:**
 ```
 
Return code:      1
Raw output file:      C:hwAC570944wB68409ADuploads
egressions2164oreground-shutdownoutput.txt
Raw output:
BEGIN EXECUTION
foreground-shutdown.dll
        1 file(s) copied.
20:26:15.68
Response file: C:hwAC570944wB68409ADeaseservices	hreading
egressions2164oreground-shutdownoreground-shutdown.dll.rsp
C:hwAC570944wB68409ADeaseservices	hreading
egressions2164oreground-shutdownIL-CG2oreground-shutdown.dll
-o:C:hwAC570944wB68409ADeaseservices	hreading
egressions2164oreground-shutdownoreground-shutdown.dll
--targetarch:x64
--targetos:windows
--verify-type-and-field-layout
--method-layout:random
-r:C:hwAC570944pSystem.*.dll
-r:C:hwAC570944pMicrosoft.*.dll
-r:C:hwAC570944p\xunit.*.dll
-r:C:hwAC570944pmscorlib.dll
-r:C:hwAC570944p
etstandard.dll
-O
" "dotnet" "C:hwAC570944pcrossgen2crossgen2.dll" @"C:hwAC570944wB68409ADeaseservices	hreading
egressions2164oreground-shutdownoreground-shutdown.dll.rsp"   -r:C:hwAC570944wB68409ADeaseservices	hreading	hreading_group1IL-CG2*.dll  -r:C:hwAC570944wB68409ADeaseservices	hreading
egressions2164oreground-shutdownIL-CG2*.dll"
Emitting R2R PE file: C:hwAC570944wB68409ADeaseservices	hreading
egressions2164oreground-shutdownoreground-shutdown.dll
" "dotnet" "C:hwAC570944p
2rdump
[...]

PR #43423, opened 2020-10-14
Reducing DllMain responsibilities, so that singlefile would not need it

VSadov:
- Removed CoreDllMain.
It just initializes CRT and GS cookie in the same way it would happen by itself.  
Comments imply that this was needed because of certain static constructors. Looks like nothing requires this any more.

- Switched thread detachment handling in EE to be off a `thread_local` object destructor 
This is a cross-platform way that works regardless of .dll or .exe scenario and thus preferable.

- Refactor DLL_PROCESS_ATTACH so it would do only Dll-specific work (saving module handles)
Other initialization was moved to `EEStartupHelper` where it runs regardless of platform or whether it is .dll/.exe scenario

contributes towards:#43072

PR #91739, opened 2023-09-07
Fix a potential shutdown hang with `Environment.Exit()` on Windows depending on timing with GCs

### LINKED ISSUE #84006: Test failure baseservices/threading/regressions/2164/foreground-shutdown/foreground-shutdown.cmd
Run: [runtime-coreclr gcstress-extra 20230326.1](https://dev.azure.com/dnceng-public/public/_build/results?buildId=217853&view=ms.vss-test-web.build-test-results-tab&runId=4073620&resultId=347720&paneView=debug)

Failed test:
```
coreclr windows arm64 Checked gcstress0xc_jitstress1 @ Windows.11.Arm64.Open
- baseservices/threading/regressions/2164/foreground-shutdown/foreground-shutdown.cmd

coreclr windows x64 Checked gcstress0xc_zapdisable @ Windows.10.Amd64.Open
- baseservices/threading/regressions/2164/foreground-shutdown/foreground-shutdown.cmd

coreclr windows x64 Checked gcstress0xc_jitstress1 @ Windows.10.Amd64.Open
- baseservices/threading/regressions/2164/foreground-shutdown/foreground-shutdown.cmd

coreclr windows x86 Checked gcstress0xc_tailcallstress @ Windows.10.Amd64.Open
- baseservices/threading/regressions/2164/foreground-shutdown/foreground-shutdown.cmd

coreclr windows arm64 Checked gcstress0xc_jitstress2 @ Windows.11.Arm64.Open
- baseservices/threading/regressions/2164/foreground-shutdown/foreground-shutdown.cmd

coreclr windows x86 Checked gcstress0xc_jitstress1 @ Windows.10.Amd64.Open
- baseservices/threading/regressions/2164/foreground-shutdown/foreground-shutdown.cmd

coreclr windows x86 Checked gcstress0xc_zapdisable @ Windows.10.Amd64.Open
- baseservices/threading/regressions/2164/foreground-shutdown/foreground-shutdown.cmd

coreclr windows x86 Checked gcstress0xc_jitminopts_heapverify1 @ Windows.10.Amd64.Open
- baseservices/threading/regressions/2164/foreground-shutdown/foreground-shutdown.cmd
```
**Error message:**
```
cmdLine:C:hwA05C0952wACDA0907easeservices	hreading	hreading_group1../regressions/2164/foreground-shutdown/foreground-shutdown.cmd Timed Out (timeout in milliseconds: 5400000 from variable __TestTimeout, start: 3/26/2023 11:30:07 PM, end: 3/27/2023 1:00:07 AM)

Return code:      -100
Raw output file:      C:hwA05C0952wACDA0907uploads
egressions2164oreground-shutdownoutput.txt
Raw output:
BEGIN EXECUTION
"C:hwA05C0952pcorerun.exe" -p "System.Reflection.Metadata.MetadataUpdater.IsSupported=false"  foreground-shutdown.dll

cmdLine:C:hwA05C0952wACDA0907easeservices	hreading	hreading_group1../regressions/2164/foreground-shutdown/foreground-shutdown.cmd Timed Out (timeout in milliseconds: 5400000 from variable __TestTimeout, start: 3/26/2023 11:30:07 PM, end: 3/27/2023 1:00:07 AM)
Test Harness Exitcode is : -100
To run the test:
Set up CORE_ROOT and run.

C:hwA05C0952wACDA0907easeservices	hreading	hreading_group1../regressions/2164/foreground-shutdown/foreground-shutdown.cmd
Expected: True
Actual:   False


Stack trace
   at Xunit.Assert.True(Nullable`1 condition, String userMessage) in /_/src/xunit.assert/Asserts/BooleanAsserts.cs:line 132
   at TestLibrary.OutOfProcessTest.RunOutOfProcessTest(String basePath, String assemblyPath)
   at Program.<<Main>$>g__TestExecutor2|0_1(StreamWriter tempLogSw, StreamWriter statsCsvSw, <>c__DisplayClass0_0&)
```

kouvel:
- On Windows when a thread calls `ExitProcess`, the `TlsDestructionMonitor` for the thread appears to be destructed after all other threads in the process are torn down. It's possible for a GC to be in progress during that time, and the thread cleanup code in `TlsDestructionMonitor` tries to enter cooperative GC mode to fix the frame pointer, leading to a hang. Fixed by deactivating the `TlsDestructionMonitor` for the thread before calling `ExitProcess`.
  - I believe it wouldn't have been an issue prior to https://github.com/dotnet/runtime/pull/43423 and https://github.com/dotnet/runtime/pull/65358, since the thread cleanup code was done on `DLL_THREAD_DETACH`, which is not raised by `ExitProcess`.
  - On Linux, upon a call to `exit` the `TlsDestructionMonitor` for the thread appears to be destructed while other threads are still running, so it is eventually able to enter cooperative GC mode. It seems the cleanup is unnecessary in this case anyway, and the behavior is similar to before, so I didn't special-case the fix for Windows.
- Also disabled the relevant test due to a different issue (https://github.com/dotnet/runtime/issues/83658) occurring in the same test on multiple platforms/architectures, which is not understood yet.

Fixes https://github.com/dotnet/runtime/issues/84006

### DIFF
--- a/src/coreclr/vm/ceemain.cpp
+++ b/src/coreclr/vm/ceemain.cpp
@@ -1727,6 +1727,11 @@ struct TlsDestructionMonitor
         m_activated = true;
     }
 
+    void Deactivate()
+    {
+        m_activated = false;
+    }
+
     ~TlsDestructionMonitor()
     {
         if (m_activated)
@@ -1766,6 +1771,11 @@ thread_local TlsDestructionMonitor tls_destructionMonitor;
 void EnsureTlsDestructionMonitor()
 {
     tls_destructionMonitor.Activate();
+}
+
+void DeactivateTlsDestructionMonitor()
+{
+    tls_destructionMonitor.Deactivate();
 }
 
 #ifdef DEBUGGING_SUPPORTED
--- a/src/coreclr/vm/ceemain.cpp
+++ b/src/coreclr/vm/ceemain.cpp
@@ -1720,47 +1720,62 @@ BOOL STDMETHODCALLTYPE EEDllMain( // TRUE on success, FALSE on error.
 
 struct TlsDestructionMonitor
 {
-    bool m_activated = false;
+    SIZE_T m_activatedOrThreadId = 0;
 
-    void Activate()
+    void Activate(Thread *thread)
     {
-        m_activated = true;
+#ifdef TARGET_UNIX
+        // The thread ID is checked in the destructor
+        m_activatedOrThreadId = thread->GetOSThreadId64();
+#else // !TARGET_UNIX
+        m_activatedOrThreadId = 1;
+#endif // TARGET_UNIX
     }
 
     void Deactivate()
     {
-        m_activated = false;
+        m_activatedOrThreadId = 0;
     }
 
     ~TlsDestructionMonitor()
     {
-        if (m_activated)
+        if (m_activatedOrThreadId == 0)
         {
-            Thread* thread = GetThreadNULLOk();
-            if (thread)
-            {
-#ifdef FEATURE_COMINTEROP
-                // reset the CoInitialize state
-                // so we don't call CoUninitialize during thread detach
-                thread->ResetCoInitialized();
-#endif // FEATURE_COMINTEROP
-                // For case where thread calls ExitThread directly, we need to reset the
-                // frame pointer. Otherwise stackwalk would AV. We need to do it in cooperative mode.
-                // We need to set m_GCOnTransitionsOK so this thread won't trigger GC when toggle GC mode
-                if (thread->m_pFrame != FRAME_TOP)
-                {
-#ifdef _DEBUG
-                    thread->m_GCOnTransitionsOK = FALSE;
-#endif
-                    GCX_COOP_NO_DTOR();
-                    thread->m_pFrame = FRAME_TOP;
-                    GCX_COOP_NO_DTOR_END();
-                }
-                thread->DetachThread(TRUE);
-            }
-
-            ThreadDetaching();
+            return;
         }
+
+#ifdef TARGET_UNIX
+        // Some Linux toolset versions call thread-local destructors during shutdown on a wrong thread.
+        if (PAL_GetCurrentOSThreadId() != m_activatedOrThreadId)
+        {
+            return;
+        }
+#endif // TARGET_UNIX
+
+        Thread* thread = GetThreadNULLOk();
+        if (thread)
+        {
+#ifdef FEATURE_COMINTEROP
+            // reset the CoInitialize state
+            // so we don't call CoUninitialize during thread detach
+            thread->ResetCoInitialized();
+#endif // FEATURE_COMINTEROP
+            // For case where thread calls ExitThread directly, we need to reset the
+            // frame pointer. Otherwise stackwalk would AV. We need to do it in cooperative mode.
+            // We need to set m_GCOnTransitionsOK so this thread won't trigger GC when toggle GC mode
+            if (thread->m_pFrame != FRAME_TOP)
+            {
+#ifdef _DEBUG
+                thread->m_GCOnTransitionsOK = FALSE;
+#endif
+                GCX_COOP_NO_DTOR();
+                thread->m_pFrame = FRAME_TOP;
+                GCX_COOP_NO_DTOR_END();
+            }
+            thread->DetachThread(TRUE);
+        }
+
+        ThreadDetaching();
     }
 };
 
@@ -1768,9 +1783,9 @@ struct TlsDestructionMonitor
 // is called when a thread is being shut down.
 thread_local TlsDestructionMonitor tls_destructionMonitor;
 
-void EnsureTlsDestructionMonitor()
+void EnsureTlsDestructionMonitor(Thread *thread)
 {
-    tls_destructionMonitor.Activate();
+    tls_destructionMonitor.Activate(thread);
 }
 
 void DeactivateTlsDestructionMonitor()
--- a/src/coreclr/vm/ceemain.h
+++ b/src/coreclr/vm/ceemain.h
@@ -46,6 +46,7 @@ void ForceEEShutdown(ShutdownCompleteAction sca = SCA_ExitProcessWhenShutdownCom
 void ThreadDetaching();
 
 void EnsureTlsDestructionMonitor();
+void DeactivateTlsDestructionMonitor();
 
 void SetLatchedExitCode (INT32 code);
 INT32 GetLatchedExitCode (void);
--- a/src/coreclr/vm/ceemain.h
+++ b/src/coreclr/vm/ceemain.h
@@ -17,6 +17,7 @@
 #include <windef.h> // for HFILE, HANDLE, HMODULE
 
 class EEDbgInterfaceImpl;
+class Thread;
 
 // Ensure the EE is started up.
 HRESULT EnsureEEStarted();
@@ -45,7 +46,7 @@ void ForceEEShutdown(ShutdownCompleteAction sca = SCA_ExitProcessWhenShutdownCom
 // Notification of a DLL_THREAD_DETACH or a Thread Terminate.
 void ThreadDetaching();
 
-void EnsureTlsDestructionMonitor();
+void EnsureTlsDestructionMonitor(Thread *thread);
 void DeactivateTlsDestructionMonitor();
 
 void SetLatchedExitCode (INT32 code);
--- a/src/coreclr/vm/eepolicy.cpp
+++ b/src/coreclr/vm/eepolicy.cpp
@@ -78,6 +78,13 @@ void SafeExitProcess(UINT exitCode, ShutdownCompleteAction sca = SCA_ExitProcess
         // disabled because if we fault in this code path we will trigger our Watson code
         CONTRACT_VIOLATION(ThrowsViolation);
 
+        // The TlsDestructionMonitor for this thread would likely be destructed at some point after ExitProcess is called. On
+        // Windows, this happens after all other threads in the process are torn down, and may occur while a GC is in progress.
+        // The thread cleanup code in TlsDestructionMonitor may try to enter cooperative GC mode to fix the frame pointer and
+        // wait for the pending GC to complete, leading to a hang. Since the process is being exited, deactivate the
+        // TlsDestructionMonitor for this thread before calling ExitProcess.
+        DeactivateTlsDestructionMonitor();
+
         ExitProcess(exitCode);
     }
 }
--- a/src/coreclr/vm/threads.cpp
+++ b/src/coreclr/vm/threads.cpp
@@ -369,7 +369,7 @@ void SetThread(Thread* t)
     gCurrentThreadInfo.m_pThread = t;
     if (t != NULL)
     {
-        EnsureTlsDestructionMonitor();
+        EnsureTlsDestructionMonitor(t);
     }
 }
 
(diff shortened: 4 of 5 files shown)


mangod9 (approved):
Assume this isnt a regression and the issue started showing up now due to timing changes?

janvorli (approved):
LGTM, thank you!

jkotas:
Can this change introduce deadlocks or data corruptions? Are there places in the runtime where suppressing unregistration of the exiting thread is going to lead to orphaned pointer?

jkotas:
For example, what is going to happen if a thread exits right after `DeactivateTlsDestructionMonitor` and then another thread triggers GC. The suspend for GC is going to hang or crash - it will try to wait the thread that exited to get to the safe spot, but that is not ever going to happen since the thread is gone.

kouvel:
> For example, what is going to happen if a thread exits right after `DeactivateTlsDestructionMonitor` and then another thread triggers GC. The suspend for GC is going to hang or crash - it will try to wait the thread that exited to get to the safe spot, but that is not ever going to happen since the thread is gone.

Is it possible for a thread to exit right after `DeactivateTlsDestructionMonitor` but before `ExitProcess` starts tearing down the threads?

My rationale was that `DeactivateTlsDestructionMonitor` is called just before the thread (T1) calls `ExitProcess`, both in preemptive mode. If a different thread T2 triggers GC in-between T1 deactivating and the call to ExitProcess, T2 shouldn't have to wait for T1 since T1 is in preemptive mode.

Also I was thinking the pointers relevant to T1 would continue to be valid (like for walking the stack) until the OS starts tearing down all threads. If it's possible that the memory behind those pointers would be deallocated while another thread is still running, then it would be problematic, but the same issue would have existed back when thread cleanup was done upon `DLL_THREAD_DETACH` since that's not raised for `ExitProcess` and the thread calling `ExitProcess` would not have done any cleanup. On Windows anyway, the thread that called `ExitProcess` appears to be the last thread that gets torn down.

jkotas:
A different example: `ExitProcess` can be called by 3rd party component. This is going to the common situation when CoreCLR is hosted and does not own "Main". Is it going to hit the same problem?

danmoseley:
Just curious, why does any code at all run on Environment.Exit? What is expected to happen other than rapid process termination?

jkotas:
>Just curious, why does any code at all run on Environment.Exit?

Environment.Exit is graceful process termination. It runs both managed (`AppDomain.ExitProcess`) and unmanaged (`atexit`, `DLL_PROCESS_DETATCH`) exit handlers.

kouvel:
> A different example: ExitProcess can be called by 3rd party component. This is going to the common situation when CoreCLR is hosted and does not own "Main". Is it going to hit the same problem?

I think returning from the main function would be ok, even if a 3rd party component owns main, because there wouldn't be any managed frames on the stack, and the frame pointer of that thread would be the top frame.

If a thread that has managed frames on the stack calls `ExitProcess` directly, bypassing `Environment.Exit`, then it could run into the same type of hang. I'm not sure at the moment what would be a good way of fixing that. Fixing the frame pointer only in `DllMain` I expect would resolve the issue when coreclr is loaded as a DLL, but I'm not sure about statically-linked cases.

> Just curious, why does any code at all run on Environment.Exit? What is expected to happen other than rapid process termination?

I'm not sure about why C++ `thread_local` objects are destructed for a thread that calls `ExitProcess` or `exit`. The `thread_local` objects for other threads are not destructed, so in that sense it's not a completely graceful exit anyway.

jkotas:
Note that this comment https://github.com/dotnet/runtime/blob/773c1807b73abfa1204b8e260ffd20518d34c328/src/coreclr/vm/ceemain.cpp#L1742-L1744 talks about `ExitThread`. I do not see how this can help to make `ExitThread` work better. And even if it did make `ExitThread` work better in some cases, I would not have a problem with breaking it. `ExitThread` on a managed thread is one of those cases that are always going to work poorly.

kouvel:
> I do not see how this can help to make `ExitThread` work better. And even if it did make `ExitThread` work better in some cases, I would not have a problem with breaking it. `ExitThread` on a managed thread is one of those cases that are always going to work poorly.

I had done some brief testing with `ExitThread` also based on the comment, it looks like while `ExitThread` would obviously not run destructors on the stack, it does run thread-local destructors both on Windows and Linux. That I believe is similar in behavior to before, where `DLL_THREAD_DETACH` is raised for a thread that calls `ExitThread`.

That said, I don't think this would make `ExitThread` scenarios any better, I'm supposing that it wouldn't make them worse either.

jkotas:
> If a thread that has managed frames on the stack calls ExitProcess directly, bypassing Environment.Exit, then it could run into the same type of hang. I'm not sure at the moment what would be a good way of fixing that. Fixing the frame pointer only in DllMain I expect would resolve the issue when coreclr is loaded as a DLL, but I'm not sure about statically-linked cases.

Would switching to `FlsAlloc` callback on Windows, like we do in native AOT https://github.com/dotnet/runtime/blob/9ae1a7cb4e024fce01bee35f533c268858e3320b/src/coreclr/nativeaot/Runtime/windows/PalRedhawkMinWin.cpp#L149 help with addressing this?

We do not set any flags before exiting the process in native AOT. It would be best for CoreCLR to be on the same plan.

kouvel:
> Would switching to `FlsAlloc` callback on Windows, like we do in native AOT help with addressing this?

I did some testing, it looks like while the `FlsAlloc` callbacks are raised a bit earlier, they can still run after all other threads have been torn down. It looks like other threads can be torn down asynchronously with running the `FlsAlloc` or thread-local callbacks for the thread that called `ExitProcess`.

I also checked out `atexit` on Windows to see if it would detect a direct call to `ExitProcess` after noticing that NativeAOT is using it to avoid performing cleanup on the same thread:

https://github.com/dotnet/runtime/blob/d34ed289d34b2f9588e10673800d269d7b1a6376/src/coreclr/nativeaot/Runtime/startup.cpp#L324-L330

It looks like the `atexit` callback is not called when a thread calls `ExitProcess` unfortunately, it is called though when exiting from `main`.

Also noticed the following:

https://github.com/dotnet/runtime/blob/d34ed289d34b2f9588e10673800d269d7b1a6376/src/coreclr/nativeaot/Runtime/startup.cpp#L332-L336

Which should probably also be checked in CoreCLR, I'll update to check the thread ID on Unixes.

kouvel:
Updated to check the thread ID on Unixes in the `TlsDestructionMonitor`, please take a look again. The diff is a bit simpler when ignoring whitespace.

jkotas:
I still think that setting the flag that suppresses thread exit notifications before exiting is going in a wrong direction.

I think we should rather go in the direction of deleting the problematic pieces from the thread shutdown notification. It is the plan that native AOT is on. The thread shutdown notification does very little work on native AOT.

If it is not possible for some reason, it likely means that there is a problem in native AOT too.

kouvel:
> I think we should rather go in the direction of deleting or rearranging the problematic pieces in the thread shutdown notification.

If it weren't for `ExitThread` cases, the thread shutdown notification probably wouldn't need to fix the frame pointer at all, as I suppose in all other cases there wouldn't be any managed frames on the stack when the thread is exiting. Maybe we can break `ExitThread` cases and remove the frame pointer fixup?

jkotas:
I would not mind regressing `ExitThread`. Do you know whether `ExitThread` will be actually broken if you delete the code the resets m_pFrame?

kouvel:
> Do you know whether ExitThread will be actually broken if you delete the code the resets m_pFrame?

I believe it would be broken, my understanding is that the main purpose of entering cooperative GC mode is to ensure that no other thread is waking that thread's stack before the thread's stack is deallocated, which would make those frame pointers invalid.

jkotas:
The thread store lock is taken when the runtime is suspended and the stackwalks are happening. The thread detach takes the thread store lock to delete the thread from the thread store. The thread detach will wait on the thread store lock until the runtime is resumed. It should prevent stack walk from ever encountering deallocated thread stack.

kouvel:
> The thread detach takes the thread store lock to delete the thread from the thread store. The thread detach will wait on the thread store lock until the runtime is resumed.

It's quite possible that I missed that, I had checked but I didn't see the thread store lock being taken, could you please point me to where that happens?

kouvel:
From a very brief rescan it looks like the thread detach adjusts the state of the thread object to "detached" but the thread object wouldn't actually be removed from the thread store until some other point in time, as there could still be references to the managed thread object.

kouvel:
> The thread detach will wait on the thread store lock until the runtime is resumed.

Could this also lead to a hang? The resume may never happen in `ExitProcess` cases on Windows.

kouvel:
It seems like this:
https://github.com/dotnet/runtime/blob/49a0633fc77ff548740d4e8b3e7b8e210f0f7294/src/coreclr/vm/threads.cpp#L995
Should be done inside the thread store lock but I don't see it being taken.

There is another lock taken here:
https://github.com/dotnet/runtime/blob/49a0633fc77ff548740d4e8b3e7b8e210f0f7294/src/coreclr/vm/threads.cpp#L989

If the detach code would run when the same thread called `ExitProcess`, I don't think it's safe for it to wait for anything. Otherwise, the detach code shouldn't run in that case.

I suspect there's an issue in NativeAOT too:
https://github.com/dotnet/runtime/blob/49a0633fc77ff548740d4e8b3e7b8e210f0f7294/src/coreclr/nativeaot/Runtime/threadstore.cpp#L183-L185

I believe the comment above would not hold when it's the same thread that called `ExitProcess` and the detach code is running as a result of that. A different thread may have been torn down while holding the thread store lock, and this thread would hang.

jkotas:
> From a very brief rescan it looks like the thread detach adjusts the state of the thread object to "detached" 

Yes, you are right. (I have described how the detach works in native AOT. I forgot that CoreCLR uses different scheme.)

jkotas:
> > The thread detach will wait on the thread store lock until the runtime is resumed.

> Could this also lead to a hang? The resume may never happen in ExitProcess cases on Windows.

It won't hang, assuming thread store lock is CRITICAL_SECTION. Instead, it is going to exit the process immediately, skipping the remaining cleanup. During shutdown, EnterCriticalSection terminates the process immediately when it is called on a critical section taken by some other thread.

jkotas:
> It seems like this:
https://github.com/dotnet/runtime/blob/49a0633fc77ff548740d4e8b3e7b8e210f0f7294/src/coreclr/vm/threads.cpp#L995
> Should be done inside the thread store lock but I don't see it being taken.

Yes, it is a tricky lock-free scheme. It is likely that it is has subtle race conditions.

I would avoid coupling it with thread store lock. Instead, introduce a single counter that counts the number of foreground threads and check that counter atomically to detect when it is safe to exit. It is what native does here: https://github.com/dotnet/runtime/blob/4b75f938a6a76a7dea177ca70de7b81878f14d33/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Thread.NativeAot.cs#L489-L502. (If you would like to clean it up, it should be a separate PR.)

kouvel:
Should I close this PR and open a separate one to remove the frame pointer fixup that breaks `ExitThread` cases?

kouvel:
> remove the frame pointer fixup

Actually nevermind, that may not work either, and there may be more things to investigate. Closing for now.

kouvel:
> It won't hang, assuming thread store lock is CRITICAL_SECTION. Instead, it is going to exit the process immediately, skipping the remaining cleanup. During shutdown, EnterCriticalSection terminates the process immediately when it is called on a critical section taken by some other thread.

I imagine this would also prevent thread-local destructors and `DLL_PROCESS_DETACH` from being raised. Is that ok?

jkotas:
Correct, the rest of the cleanup code won't run.

kouvel:
I guess it's a necessity.

PR #91745, opened 2023-09-07
Disable CodeQL for jitrollingbuild pipeline

BruceForstall:
It appears to cause osx builds to fail due to timeout or crash.

I followed the pattern of #90329

Verified this worked on internal pipeline.


BruceForstall:
@jakobbotsch PTAL
cc @akoeplinger @dotnet/jit-contrib

hoyosjs (approved):

jakobbotsch (approved):

PR #91749, opened 2023-09-07
[wasm] Add blazor First UI measurement

radekdoulik:
This measures the time of `OnAfterRender` event on index page, to mimic the blazor benchmark's `First UI` measurement.

Also

 * fix blazor frame url base

 * let simple-server handle non-root urls with trailing '/'

### DIFF
--- a/src/mono/sample/wasm/blazor-frame/wwwroot/index.html
+++ b/src/mono/sample/wasm/blazor-frame/wwwroot/index.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
     <title>blazor</title>
-    <!-- <base href="/blazor-template/" /> -->
+    <base href="/blazor-template/" />
     <link rel="stylesheet" href="css/bootstrap/bootstrap.min.css" />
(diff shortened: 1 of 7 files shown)


src/mono/sample/wasm/blazor-frame/wwwroot/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
     <title>blazor</title>
-    <!-- <base href="/blazor-template/" /> -->
```
maraf, line 7: Just curious, what's the reason for this change?

maraf (approved):
Looks good to me 👍

src/mono/sample/wasm/blazor-frame/wwwroot/index.html
radekdoulik: The `Index.razor` page wasn't loading properly, so I changed it according to https://learn.microsoft.com/en-us/aspnet/core/blazor/host-and-deploy/?view=aspnetcore-8.0&tabs=visual-studio#app-base-path

Interestingly I had to also change the `blazor.webassembly.js` path to make it work.

PR #91751, opened 2023-09-07
[NativeAOT] Switch macOS to mmap thunks like iOS platforms

### LINKED ISSUE #91738: [NativeAOT] Thunk mapping doesn't work on macOS/x64 with hardened runtime
*Filing to keep track of an issue that was experienced by the Avalonia developers*

References:
[ThunksMapping 1](https://github.com/dotnet/runtime/blob/1e5ee2dc46508223f1bb04fc21a8003cc7605442/src/coreclr/nativeaot/Runtime/ThunksMapping.cpp#L125-L134)
[ThunksMapping 2](https://github.com/dotnet/runtime/blob/1e5ee2dc46508223f1bb04fc21a8003cc7605442/src/coreclr/nativeaot/Runtime/ThunksMapping.cpp#L236-L250)
[PalVirtualAlloc](https://github.com/dotnet/runtime/blob/1e5ee2dc46508223f1bb04fc21a8003cc7605442/src/coreclr/nativeaot/Runtime/unix/PalRedhawkUnix.cpp#L819-L824)

These are essentially the three places that need to be adjusted. Firstly, `PalVirtualAlloc` needs a fallback to use `MAP_JIT` on macOS/x64. Secondly, the call to `pthread_jit_write_protect_np` in ThunksMapping need to be executed on x64 if `MAP_JIT` was used.

The code is not very cleanly commented but it does work on macOS/arm64. On macOS/x64 we traditionally used the same code path as on Linux, but that's not valid when running under hardened runtime. So essentially macOS/x64 needs to have both versions of the code and switch at runtime depending on whether the executable uses hardened runtime or not. And we can detect the hardened runtime upfront by doing an allocation that would not succeed. Essentially it would be a copy of [this CoreCLR code]( https://github.com/dotnet/runtime/blob/1e5ee2dc46508223f1bb04fc21a8003cc7605442/src/coreclr/pal/src/misc/utils.cpp#L333-L366).

Additionally, it needs to be verified that we link weakly to the `pthread_jit_write_protect_np` symbol if targeting old enough macOS version that doesn't have it yet. This should be done by the compiler/linker transparently, but we should double check. 

cc @kekekeks @jmacato If you have a small repro, please attach it.
cc @agocke

filipnavara:
Fixes #91738

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
@@ -121,7 +121,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     </ItemGroup>
 
     <ItemGroup>
-      <ExtraLinkerArg Include="-segprot,__THUNKS,rx,rx" Condition="'$(_IsiOSLikePlatform)' == 'true'" />
+      <ExtraLinkerArg Include="-segprot,__THUNKS,rx,rx" Condition="'$(_IsApplePlatform)' == 'true'" />
     </ItemGroup>
 
     <ItemGroup>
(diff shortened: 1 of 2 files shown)


jkotas (approved):

jkotas:
Thank you!

MichalPetryka:
Does backporting this to 8.0 make sense?

jkotas:
> Does backporting this to 8.0 make sense?

@MichalPetryka Thank you for pointing this out. The fix has been backported to .NET 8 RC2.

PR #91755, opened 2023-09-07
[mono] add nullable annotation to InternalInvoke

lambdageek:
We had two of these icalls with different signatures in RuntimeMethodInfo and RuntimeConstructorInfo.

They both end up calling `ves_icall_InternalInvoke` which sometimes returns null


lambdageek:
This is wrong actually.  they should both be nullable

jkotas (approved):

vargaz (approved):

PR #91758, opened 2023-09-07
Split the informational version on both SemVer2 separators when calculating TFM

jkoritzinsky:
This fixes a build break found when we stabilize the package version (and the only separator in the informational version is '+' before the commit id)


AaronRobinsonMSFT (approved):

jkoritzinsky:
Failures are all known issues.

PR #91759, opened 2023-09-07
Fix whole program optimization option for Windows

janvorli:
The recent change to clean up the way some options in Windows builds of the runtime are overridden missed the fact that the `CMAKE_INTERPROCEDURAL_OPTIMIZATION` cannot contain generator expressions and so it was always set to "off".

The fix is to set` CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE` and `CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELWITHDEBUGINFO` instead.


janvorli:
cc: @elinor-fung

hoyosjs (approved):

jkoritzinsky (approved):

jakobbotsch:
image

:+1:

PR #91765, opened 2023-09-07
Remove .NET6/7 SafeHandle marshalling logic

jkoritzinsky:
We don't build for .NET 6 or 7 in dotnet/runtime any more, so we don't need to include this marshalling logic.


jkoritzinsky:
I'm going to merge this into another PR instead of doing this here. It'll end up being a cleaner change.

PR #91766, opened 2023-09-07
Fix fgDebugCheckFlags checks for FTN_ADDR

### LINKED ISSUE #91505: Assertion failed '!"Missing flags on tree"'
```cs
static unsafe void PrintFPtr()
{
    delegate*<void> ptr = &Foo;
    Console.WriteLine(*(long*)ptr);
}

static void Foo()
{
}
```
```
Missing flags on tree [000001]: #------------
               [000001] n---G+-----                         *  IND       long  
               [000000] H----+-----                         \--*  CNS_INT(h) long   0x7ffb48ff03c0 ftn

Assert failure(PID 2888 [0x00000b48], Thread: 50468 [0xc524]): Assertion failed '!"Missing flags on tree"' in 'Tests:PrintFPtr()' during 'Morph - Global' (IL size 13; hash 0x403dcc3a; Tier0)

    File: C:\prj\runtime-base\src\coreclr\jit\fgdiagnostic.cpp Line: 3319
    Image: C:\prj\runtime-base\artifacts\bin\coreclr\windows.x64.Checked\CoreRun.exe
```

EgorBo:
Closes https://github.com/dotnet/runtime/issues/91505

the check used to complain that `IND(ICON)` should have `GTF_IND_INVARIANT` flag, but it's not true - `IND(FTNPTR)` is never invariant because it points to a potentially mutable data (e.g. tiering)

### DIFF
--- a/src/coreclr/jit/fgdiagnostic.cpp
+++ b/src/coreclr/jit/fgdiagnostic.cpp
@@ -3152,6 +3152,7 @@ void Compiler::fgDebugCheckFlags(GenTree* tree)
                 // Some of these aren't handles to invariant data...
                 if ((handleKind == GTF_ICON_STATIC_HDL) || // Pointer to a mutable class Static variable
                     (handleKind == GTF_ICON_BBC_PTR) ||    // Pointer to a mutable basic block count value
+                    (handleKind == GTF_ICON_FTN_ADDR) ||   // Pointer to a potentially mutable VM slot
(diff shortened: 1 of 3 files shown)


EgorBo:
@AndyAyersMS PTAL

AndyAyersMS (approved):

EgorBo:
Failures are https://github.com/dotnet/runtime/issues/91838

Also, filed an issue for a case that still sets `IND_INVARIANT` for `IND(FTN_PTR)` in the existing code: https://github.com/dotnet/runtime/issues/91846

PR #91767, opened 2023-09-07
Metrics Feature Switch

tarekgh:
https://github.com/dotnet/runtime/issues/89880

### DIFF
--- a/docs/workflow/trimming/feature-switches.md
+++ b/docs/workflow/trimming/feature-switches.md
@@ -13,6 +13,7 @@ configurations but their defaults might vary as any SDK can set the defaults dif
 | EnableUnsafeBinaryFormatterSerialization | System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization | BinaryFormatter serialization support is trimmed when set to false |
 | EventSourceSupport | System.Diagnostics.Tracing.EventSource.IsSupported | Any EventSource related code or logic is trimmed when set to false |
 | InvariantGlobalization | System.Globalization.Invariant | All globalization specific code and data is trimmed when set to true |
+| MetricsSupport | System.Diagnostics.Metrics.IsSupported | Any Metrics related code or logic is trimmed when set to false |
 | PredefinedCulturesOnly | System.Globalization.PredefinedCulturesOnly |  Don't allow creating a culture for which the platform does not have data |
 | HybridGlobalization | System.Globalization.Hybrid |  Properties connected with the mixed: platform-specific + icu-based globalization will be trimmed  |
 | UseSystemResourceKeys | System.Resources.UseSystemResourceKeys |  Any localizable resources for system assemblies is trimmed when set to true |
--- a/docs/workflow/trimming/feature-switches.md
+++ b/docs/workflow/trimming/feature-switches.md
@@ -13,7 +13,7 @@ configurations but their defaults might vary as any SDK can set the defaults dif
 | EnableUnsafeBinaryFormatterSerialization | System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization | BinaryFormatter serialization support is trimmed when set to false |
 | EventSourceSupport | System.Diagnostics.Tracing.EventSource.IsSupported | Any EventSource related code or logic is trimmed when set to false |
 | InvariantGlobalization | System.Globalization.Invariant | All globalization specific code and data is trimmed when set to true |
-| MetricsSupport | System.Diagnostics.Metrics.IsSupported | Any Metrics related code or logic is trimmed when set to false |
+| MetricsSupport | System.Diagnostics.Metrics.Meter.IsSupported | Any Metrics related code or logic is trimmed when set to false |
 | PredefinedCulturesOnly | System.Globalization.PredefinedCulturesOnly |  Don't allow creating a culture for which the platform does not have data |
 | HybridGlobalization | System.Globalization.Hybrid |  Properties connected with the mixed: platform-specific + icu-based globalization will be trimmed  |
 | UseSystemResourceKeys | System.Resources.UseSystemResourceKeys |  Any localizable resources for system assemblies is trimmed when set to true |
--- a/src/libraries/System.Diagnostics.DiagnosticSource/src/ILLink/ILLink.Substitutions.Shared.xml
+++ b/src/libraries/System.Diagnostics.DiagnosticSource/src/ILLink/ILLink.Substitutions.Shared.xml
@@ -1,0 +1,10 @@
+<linker>
+  <assembly fullname="System.Diagnostics.DiagnosticSource">
+    <type fullname="System.Diagnostics.Metrics.Meter" feature="System.Diagnostics.Metrics.IsSupported" featurevalue="false">
+      <method signature="System.Boolean get_IsSupported()" body="stub" value="false" />
+    </type>
+    <type fullname="System.Diagnostics.Metrics.Instrument" feature="System.Diagnostics.Metrics.IsSupported" featurevalue="false">
+      <method signature="System.Boolean get_Enabled()" body="stub" value="false" />
+    </type>
+  </assembly>
+</linker>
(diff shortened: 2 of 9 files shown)


src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/Instrument.common.cs
```diff
@@ -52,6 +52,11 @@ protected Instrument(Meter meter, string name, string? unit, string? description
         /// <param name="tags">A span of key-value pair tags associated with the measurement.</param>
         protected void RecordMeasurement(T measurement, ReadOnlySpan<KeyValuePair<string, object?>> tags)
         {
+            if (!Meter.IsSupported)
```
noahfalk, line 55: Can we avoid adding this check and count on _subscription.First to be null? It looks like you already blocked all the code paths that could have registered a subscription.

tarekgh:
@eerhardt @noahfalk

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/Meter.cs
```diff
@@ -17,11 +17,22 @@ public class Meter : IDisposable
...
+            AppContext.TryGetSwitch("System.Diagnostics.Metrics.IsSupported", out bool isSupported) ? isSupported : true;
+
         /// <summary>
         /// Initialize a new instance of the Meter using the <see cref="MeterOptions" />.
         /// </summary>
         public Meter(MeterOptions options)
         {
+            if (!IsSupported)
+            {
+                Name = options is not null && options.Name is not null ? options.Name : throw new ArgumentNullException(nameof(options));
```
noahfalk, line 32: It seems odd to me that we would initialize the Name property but not the other properties. Is there a reason to split it this way?

noahfalk (approved):
LGTM with a couple questions inline

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/Instrument.common.cs
tarekgh: Yes, we can do that. According to my understanding (per @eerhardt explanation) jit is going to optimize that in tier two and will fully remove this check. I was hoping when `IsSupported` is false, the linker can empty the whole method body including accessing the static fields. Let me know if you still think we should remove this check.

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/Meter.cs
tarekgh: I only added Name because the compiler was complaining not initializing non-nullable member. I am ok either way, either suppress the compiler error and not initialize the name or initialize all other properties. I don't have strong opinion which way we should go with.

noahfalk: How about we initialize name to empty string? Keeps the compiler happy and preserves consistency that when metrics aren't supported none of constructor inputs are preserved.

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/Instrument.common.cs
noahfalk: Can you confirm the JIT does optimize it as expected. As long as that is confirmed then I have no qualms leaving it there :)

stephentoub: Adding this check will allow the trimmer to remove the rest of the method body, right?

docs/workflow/trimming/feature-switches.md
```diff
@@ -13,6 +13,7 @@ configurations but their defaults might vary as any SDK can set the defaults dif
 | EnableUnsafeBinaryFormatterSerialization | System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization | BinaryFormatter serialization support is trimmed when set to false |
 | EventSourceSupport | System.Diagnostics.Tracing.EventSource.IsSupported | Any EventSource related code or logic is trimmed when set to false |
 | InvariantGlobalization | System.Globalization.Invariant | All globalization specific code and data is trimmed when set to true |
+| MetricsSupport | System.Diagnostics.Metrics.IsSupported | Any Metrics related code or logic is trimmed when set to false |
```
eerhardt, line 16: One naming suggestion I've been making is to make the appcontext switch a possible namespace + Type + property name. That way if we ever needed to expose this in a public API, they wouldn't be inconsistent.

`System.Diagnostics.Metrics` is a namespace, which can't have properties. So I'd suggest something like:

System.Diagnostics.Metrics.Meter.IsSupported

Which makes it consistent with `System.Diagnostics.Tracing.EventSource.IsSupported`.

src/libraries/System.Diagnostics.DiagnosticSource/src/System.Diagnostics.DiagnosticSource.csproj
```diff
@@ -135,4 +135,8 @@ System.Diagnostics.DiagnosticSource</PackageDescription>
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="$(SystemRuntimeCompilerServicesUnsafeVersion)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ILLinkSubstitutionsXmls Include="ILLink/ILLink.Substitutions.Shared.xml" />
```
eerhardt, line 139: (nit) typically these go above the `Compile` items.

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/Meter.cs
eerhardt: It might make sense to just move this check into the `Initialize` method - then we don't need to do this mess with `Name` and in 2 places.

eerhardt (commented):
Should we also have a trimming test that ensures certain things are trimmed?

Can you show a before and after size comparison (maybe using ILSpy or https://github.com/MichalStrehovsky/sizoscope) that shows what is all trimmed when this switch is set?

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Metrics/Instrument.common.cs
eerhardt: > Adding this check will allow the trimmer to remove the rest of the method body, right?

Correct - and all the methods it calls that aren't called elsewhere.

tarekgh: I created a simple app using my private built. The app and the runtime were built as a release. I have `IsSupported = true` to test the normal case. I run it under the debugger, and I am seeing the call `Meter.IsSupported` is not removed inside `RecordMeasurement` method. I'll remove this check from there as this is a hot path, we want to avoid regressions there. 

image

tarekgh:
@eerhardt confirmed by Sizoscope the trimming is occurred as desired. I'll try to start do Xamarin measurements with this version and looks if we need to do more here.  

image

src/libraries/System.Diagnostics.DiagnosticSource/src/ILLink/ILLink.Substitutions.Shared.xml
```diff
@@ -0,0 +1,10 @@
+<linker>
+  <assembly fullname="System.Diagnostics.DiagnosticSource">
+    <type fullname="System.Diagnostics.Metrics.Meter" feature="System.Diagnostics.Metrics.Meter.IsSupported" featurevalue="false">
+      <method signature="System.Boolean get_IsSupported()" body="stub" value="false" />
+    </type>
+    <type fullname="System.Diagnostics.Metrics.Instrument" feature="System.Diagnostics.Metrics.Meter.IsSupported" featurevalue="false">
+      <method signature="System.Boolean get_Enabled()" body="stub" value="false" />
+    </type>
+  </assembly>
```
eerhardt, line 9: (nit) you can refactor this to avoid the duplication.

```suggestion
  <assembly fullname="System.Diagnostics.DiagnosticSource" feature="System.Diagnostics.Metrics.Meter.IsSupported" featurevalue="false">
    <type fullname="System.Diagnostics.Metrics.Meter">
      <method signature="System.Boolean get_IsSupported()" body="stub" value="false" />
    </type>
    <type fullname="System.Diagnostics.Metrics.Instrument">
      <method signature="System.Boolean get_Enabled()" body="stub" value="false" />
    </type>
  </assembly>
```

eerhardt (approved):
LGTM. Thanks, @tarekgh.

PR #91768, opened 2023-09-07
Remove "Is supported on this TFM" logic from marshalling generators and instead handle it during factory construction

jkoritzinsky:
The only generator that supports down-level is the LibraryImportGenerator, so remove this infrastructure from the base APIs and instead have LibraryImportGenerator create a different generator factory depending on the TFM.

Also, include #91765 as we would have had to separate out the legacy SafeHandle marshaller instantiation to a separate marshalling generator factory, but we don't use it anyway any more.


src/libraries/System.Runtime.InteropServices/gen/LibraryImportGenerator/LibraryImportGenerator.cs
```diff
@@ -313,6 +313,22 @@ private static SyntaxTokenList StripTriviaFromModifiers(SyntaxTokenList tokenLis
                 return (PrintForwarderStub(pinvokeStub.StubMethodSyntaxTemplate, explicitForwarding: true, pinvokeStub, diagnostics), pinvokeStub.Diagnostics.Array.AddRange(diagnostics.Diagnostics));
             }
 
+            bool supportsTargetFramework = true;
+            if (pinvokeStub.LibraryImportData.SetLastError && !options.GenerateForwarders)
+            {
+                supportsTargetFramework = pinvokeStub.GeneratorFactoryKey.Key.TargetFramework == TargetFramework.Net
+                    && pinvokeStub.GeneratorFactoryKey.Key.Version.Major >= 6;
+            }
```
jtschuster, line 321: Should we just add the conditions for the if block to the `supportsTargetFramework` assignment?

```suggestion
            bool supportsTargetFramework = pinvokeStub.LibraryImportData.SetLastError
                && !options.GenerateForwarders
                && pinvokeStub.GeneratorFactoryKey.Key.TargetFramework == TargetFramework.Net
                && pinvokeStub.GeneratorFactoryKey.Key.Version.Major >= 6;
```

jtschuster (approved):

PR #91774, opened 2023-09-08
Remove uses of IlcTrimMetadata from the test tree

### LINKED ISSUE #71506: [NativeAOT] Replace IlcTrimMetadata on test projects
The project property fix, IlcTrimMetadata, is a quick fix for the regression introduced by PR, #70201. The right fix is to either make the test analyzable or add a RD.XML to each test project. The impacted tests are as below,

```
System.Collections.NonGeneric/tests/System.Collections.NonGeneric.Tests.csproj
System.Net.Security/tests/UnitTests/System.Net.Security.Unit.Tests.csproj
System.Net.WebSockets.Client/tests/System.Net.WebSockets.Client.Tests.csproj
System.Numerics.Vectors/tests/System.Numerics.Vectors.Tests.csproj
System.Private.Xml/tests/XmlSchema/XmlSchemaSet/System.Xml.XmlSchemaSet.Tests.csproj
System.Security.Cryptography.Csp/tests/System.Security.Cryptography.Csp.Tests.csproj
System.Threading.Tasks.Extensions/tests/System.Threading.Tasks.Extensions.Tests.csproj
System.Threading.Tasks/tests/System.Threading.Tasks.Tests.csproj

```

MichalStrehovsky:
This is basically just making tests trimmable. We want to be testing the shipping configuration without weird switches.

I left IlcTrimMetadata in two test projects because the uses there were too annoying to fix up. For System.Security, the problem is in non-trimmability of the APIs and would need RD.XML/whatever. For Vectors, the problem is in the use of `dynamic` in the tests due to laziness and I'm equally lazy to rewrite all of it.

Fixes #71506.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/libraries/System.Numerics.Vectors/tests/System.Numerics.Vectors.Tests.csproj
+++ b/src/libraries/System.Numerics.Vectors/tests/System.Numerics.Vectors.Tests.csproj
@@ -3,7 +3,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <TargetFramework>$(NetCoreAppCurrent)</TargetFramework>
     <TestRuntime>true</TestRuntime>
-    <!-- Uses a lot of `dynamic` that IlcTrimMetadata=false happens for fix -->
+    <!-- Uses a lot of `dynamic` that IlcTrimMetadata=false happens to fix -->
     <IlcTrimMetadata>false</IlcTrimMetadata>
   </PropertyGroup>
   <ItemGroup>
--- a/src/libraries/System.Numerics.Vectors/tests/System.Numerics.Vectors.Tests.csproj
+++ b/src/libraries/System.Numerics.Vectors/tests/System.Numerics.Vectors.Tests.csproj
@@ -3,7 +3,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <TargetFramework>$(NetCoreAppCurrent)</TargetFramework>
     <TestRuntime>true</TestRuntime>
-    <!--Remove once this is fixed, https://github.com/dotnet/runtime/issues/71506 -->
+    <!-- Uses a lot of `dynamic` that IlcTrimMetadata=false happens for fix -->
     <IlcTrimMetadata>false</IlcTrimMetadata>
   </PropertyGroup>
   <ItemGroup>
--- a/src/libraries/System.Private.Xml/tests/AllowDefaultResolverContext.cs
+++ b/src/libraries/System.Private.Xml/tests/AllowDefaultResolverContext.cs
@@ -32,7 +32,7 @@ namespace System.Xml.Tests
             // the default, or alternatively separate all such tests into a separate test project
             // where that project contains only those tests that require the switch set.
 
-            Type t = typeof(XmlConvert).Assembly.GetType("System.Xml.LocalAppContextSwitches");
+            Type t = Type.GetType("System.Xml.LocalAppContextSwitches, System.Private.Xml");
             Assert.NotNull(t);
 
             FieldInfo fi = t.GetField("s_allowDefaultResolver", BindingFlags.NonPublic | BindingFlags.Static);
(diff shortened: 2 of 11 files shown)


src/libraries/System.Numerics.Vectors/tests/System.Numerics.Vectors.Tests.csproj
```diff
@@ -3,7 +3,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <TargetFramework>$(NetCoreAppCurrent)</TargetFramework>
     <TestRuntime>true</TestRuntime>
-    <!--Remove once this is fixed, https://github.com/dotnet/runtime/issues/71506 -->
+    <!-- Uses a lot of `dynamic` that IlcTrimMetadata=false happens for fix -->
```
vitek-karas, line 6:
```suggestion
    <!-- Uses a lot of `dynamic` that IlcTrimMetadata=false happens to fix -->
```

src/libraries/System.Private.Xml/tests/AllowDefaultResolverContext.cs
```diff
@@ -32,7 +32,7 @@ private static void ClearCachedSwitch()
             // the default, or alternatively separate all such tests into a separate test project
             // where that project contains only those tests that require the switch set.
 
-            Type t = typeof(XmlConvert).Assembly.GetType("System.Xml.LocalAppContextSwitches");
+            Type t = Type.GetType("System.Xml.LocalAppContextSwitches, System.Private.Xml");
```
vitek-karas, line 35: I've been seeing so much of this lately that I'm wondering if we should just implement the necessary data flow for this.

src/libraries/System.Threading.Tasks.Extensions/tests/ValueTaskTests.cs
```diff
@@ -1278,7 +1278,7 @@ public void NonGeneric_TornRead_DoesNotCrashOrHang()
 
             // _obj is a Task but other fields are from result construction
             vtBoxed = new ValueTask();
-            vtBoxed.GetType().GetField("_obj", BindingFlags.NonPublic | BindingFlags.Instance).SetValue(vtBoxed, Task.CompletedTask);
+            typeof(ValueTask).GetField("_obj", BindingFlags.NonPublic | BindingFlags.Instance).SetValue(vtBoxed, Task.CompletedTask);
```
vitek-karas, line 1281: Another potential candidate for better data flow.
We know the static type of the variable here - maybe we should use that to try to resolve the field, and only if that fails produce a warning. But it's a bit weird.

vitek-karas (approved):

src/libraries/System.Threading.Tasks.Extensions/tests/ValueTaskTests.cs
MichalStrehovsky: The static type of the variable is object, which is why it doesn't work (it would work if it was typed as ValueTask, because it's struct and therefore sealed). We'd need to track the types of things stored here. Analysis in .NET Native did it. I don't know if we really want to go there again.

vitek-karas: Sorry - I missed that. You're right, I don't think we should do this.

src/libraries/System.Private.Xml/tests/AllowDefaultResolverContext.cs
Suchiman: :+1: this is my favorite go to way of doing things

vitek-karas: The sad part is that even if we implement it (which I think we should) it will only work when the app targets net9+. It's not much use for almost every library out there right now. It will take something like 4 years to really be usable by libraries.
And backporting it to .NET 6 is... difficult to say the least.

LakshanF (approved):

PR #89061, opened 2023-07-18
Add support for default implementation of static virtuals with method constraints

davidwrighton:
- The major problem was the logic which incorrectly would instantiate the methods when it wasn't necessary
- As the number of flags to the implementation functions has grown very large, this change also includes logic converting them all to a single flags variable when passing them around

Fixes #73658
Fixes #78865

PR #91780, opened 2023-09-08
Start running test added in #89061

MichalStrehovsky:
GitHub_73658.cs added in #89061 was lacking a csproj so we were not actually running it.
GitHub_78865.cs added in the same PR is an exact copy of StaticVirtualsAndMethodConstraintsAndDefaultImplementation.cs that was added in the same PR. It also lacked a csproj. I'm deleting it instead of adding csproj.


davidwrighton (approved):

MichalStrehovsky:
Turns out GitHub_73658.cs is also a dupe, of GetInterfaceMapWithStaticVirtualsAndConstraints.cs this time. So deleting that one too.

PR #91782, opened 2023-09-08
Don't log warnings for certain RemoveAttributeInstances

### LINKED ISSUE #88994: Multiple IL2049 warnings when publishing with NativeAOT and _AggressiveAttributeTrimming
### Description

I tried using `_AggressiveAttributeTrimming` to check the size diffs, and noticed a whole bunch of warnings:

> resource ILLink.LinkAttributes.xml in System.Private.CoreLib(377,8): warning IL2049: System.Private.CoreLib: The internal attribute name 'RemoveAttributeInstances' being used in the xml is not supported by ILLink, check the spelling and the supported internal attributes. [C:\Users\sergiopedri\git\ComputeSharp\samples\ComputeSharp.SwapChain.Cli\ComputeSharp.SwapChain.Cli.csproj]

### Reproduction Steps

- Checkout to https://github.com/Sergio0694/ComputeSharp/tree/dev/_AggressiveAttributeTrimming-warn
- Build the CLI sample with these steps:
```powershell
$env:COMPUTESHARP_SWAPCHAIN_CLI_PUBLISH_AOT='true';
dotnet publish samples\ComputeSharp.SwapChain.Cli\ComputeSharp.SwapChain.Cli.csproj -c Release -f net8.0 -r win-x64
```

### Expected behavior

Should build with no warnings.

### Actual behavior

When `_AggressiveAttributeTrimming` is set, you get a whole bunch of warnings.

### Configuration

- 8.0.100-preview.7.23364.32

MichalStrehovsky:
Fixes #88994.

ILLinker attribute XML format has a way to express "only remove these attributes when parameter X has value of Y". We currently generate a warning and don't trim the attribute at all.

The savings from trimming these attributes are going to be miniscule. Even in IL, this is scraping the bottom of the barrel. In Native AOT these attributes are pretty effectively deduplicated across assemblies. We'll likely never need this.

This keeps the existing behavior (don't trim the attribute), but removes the warning.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/UsageBasedMetadataManager.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/UsageBasedMetadataManager.cs
@@ -1093,7 +1093,7 @@ namespace ILCompiler
                 string internalValue = GetAttribute(nav, "internal");
                 if (!string.IsNullOrEmpty(internalValue))
                 {
-                    if (!IsRemoveAttributeInstances(internalValue) || !nav.IsEmptyElement)
+                    if (!IsRemoveAttributeInstances(internalValue))
                     {
                         LogWarning(nav, DiagnosticId.UnrecognizedInternalAttribute, internalValue);
                     }


rolfbjarne (approved):

vitek-karas (approved):
We're likely missing a test here - this should have been caught by a test.

I agree that if NativeAOT doesn't support that feature, it should not warn because of it being used.

MichalStrehovsky:
> We're likely missing a test here - this should have been caught by a test.

We have validation for the things that work (unconditional removal) in Native AOT smoke test suite. We just don't run the shared ILLinker tests for these. We probably don't have any sort of Kept validation for attributes. One could extract that information from the dependency graph in theory.

PR #89654, opened 2023-07-28
SPMI: Gracefully handle managed exceptions during replay

jakobbotsch:
Various JIT-EE APIs are expected to be able to throw exceptions, which SPMI will record and rethrow during replay. This change makes us handled managed exception that escapes from `CILJit::compileMethod` gracefully by not considering them as replay failures.
The expectation is that an exception with this exception code will never be thrown due to a JIT bug, and is always thrown as a result of a recorded exception that the JIT itself did not want to handle.

Also do a small amount of cleanup for the EH in SPMI.

Contributes to #47540 and #47546.

PR #91783, opened 2023-09-08
SPMI: Handle "successful error cases" in near differ on arm64

### LINKED ISSUE #91257: SPMI asmdiff errors on Arm64 Linux
### Description

On Arm64 Linux, running SPMI asmdiffs causes errors:

```
[15:39:34] ERROR: Couldn't load base metrics summary created by child process
```

And the script fails with no diffs.

### Reproduction Steps

```
./build.sh -rc checked -lc release -s clr+libs
./src/tests/build.sh generatelayoutonly Checked
python3 ./src/coreclr/scripts/superpmi.py asmdiffs -build_type Checked -log_level debug
```

### Expected behavior

SPMI will run to completion without any errors

### Actual behavior

Full log file:  [superpmi.4.log](https://github.com/dotnet/runtime/files/12431134/superpmi.4.log)

### Regression?

Used to work.

### Known Workarounds

None.

### Configuration

runtime HEAD:
```
3bda6e0013d 2023-08-21.. Vladimir Sadov  [NativeAOT] Missing memory fence before bulk move of objects (#90890) HEAD -> main, origin/main, origin/HEAD
```

Linux Ubuntu 22.04, Arm64 Altra.

### Other information

Works on: Linux Ubuntu 22.04, X64
Works on: Windows, Arm64

SPMI replay works on Linux, Arm64

jakobbotsch:
After #89654 SPMI replay will succeed instead of result in replay errors in expected error cases (such as BADCODE or EE exception). To support diffing such contexts, we record zero-sized assembly that the near differ uses. However, on arm64 there is some additional code that calls repCompileMethod to make some additional adjustments to the code blob, and in the "EE exception" cases we cannot replay this function, resulting in crash during asmdiff. This fixes the problem by only making the adjustments when we know there is any code.

An alternative solution could be to avoid invoking the neardiffer at all in the succeeding error cases, but this seemed like an ok pragmatic solution. However, if there are differing opinions then I would be fine with also pursuing this.

Fix #91257

### DIFF
--- a/src/coreclr/tools/superpmi/superpmi/neardiffer.cpp
+++ b/src/coreclr/tools/superpmi/superpmi/neardiffer.cpp
@@ -1247,28 +1247,29 @@ bool NearDiffer::compare(MethodContext* mc, CompileResult* cr1, CompileResult* c
     // is a sum of their sizes. The following is to adjust their sizes and the roDataBlock_{1,2} pointers.
     if (GetSpmiTargetArchitecture() == SPMI_TARGET_ARCHITECTURE_ARM64)
     {
-        BYTE*        nativeEntry_1;
-        ULONG        nativeSizeOfCode_1;
-        CorJitResult jitResult_1;
+        if (hotCodeSize_1 > 0)
+        {
+            BYTE* nativeEntry_1;
+            ULONG        nativeSizeOfCode_1;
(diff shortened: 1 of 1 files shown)


jakobbotsch:

BruceForstall (approved):
LGTM

PR #91788, opened 2023-09-08
Fix handling of variant casting in the managed type system

### LINKED ISSUE #85146: Managed type system casting logic doesn't handle all array variance
Compile following with optimizations on:

```csharp
class Program
{
    static void Main()
    {
        object source = new Enum0[] { Enum0.First, Enum0.Second, Enum0.Third };
        Console.WriteLine(source is IEnumerable<Enum1>);
    }

    enum Enum0 { First, Second, Third }
    enum Enum1 { }
}
```

Expected: prints true
Observed: prints false

This is because the CanCastTo logic within the managed type system doesn't think these are castable. RyuJIT hardcodes this to false in codegen.

MichalStrehovsky:
Fixes #85146.

Cc @dotnet/ilc-contrib


jkotas (approved):
Thanks

Issue #84834, opened 2023-04-14
CoreCLR RISC-V architecture port


gbalykov:
## Current status

Our initial goal was to get to 100% pass rate for CodeGen BringUp tests in Debug mode. We've achieved this both on qemu and on StarFive VisionFive2 board.

For this, all changes are currenlty in PRs created by my colleagues @clamp03 @t-mustafin:
- already merged (https://github.com/dotnet/runtime/pull/82379, https://github.com/dotnet/runtime/pull/82382, https://github.com/dotnet/runtime/pull/84584, https://github.com/dotnet/runtime/pull/84608, https://github.com/dotnet/runtime/pull/82380)
- not yet merged at the moment of writing (https://github.com/dotnet/runtime/pull/84748, https://github.com/dotnet/runtime/pull/82381, https://github.com/dotnet/runtime/pull/84797).

So, when all of the above are merged, CodeGen BringUp tests launched on main branch should pass both on VisionFive2 and qemu (*for StarFive VisionFive2 board `GC_REGIONS` also need to be disabled currently, we are already investigating this*).

Current test results for full JIT directory on VisionFive2 in Debug with `DOTNET_GCgen0size=99999999`:
```
# Tests Discovered : 4737
# Passed           : 4304
# Failed           : 413
# Skipped          : 20
```

Current test results for all tests on VisionFive2 in Debug with `DOTNET_GCgen0size=99999999`:
```
# Tests Discovered : 6565
# Passed           : 5663
# Failed           : 861
# Skipped          : 41
```

#### Our further plans for near future:
- focus on full JIT tests directory to make tests pass in Debug mode and with `DOTNET_GCgen0size=99999999`
- fix TODOs, which are not related to jit optimizations
- add CI in dotnet/runtime repo for RISC-V with some tests run on qemu
- enable JitDisasm and JitDump
- fix or disable `GC_REGIONS` on VisionFive2
- focus on more tests dirs with Debug build of runtime after all of the above

#### What's currently not in our plans for near future:
- full jit optimizations with all optimization passes (i.e. disabled MinOpt, tiered compilation, etc.)
- optimizations of codegen itself
- crossgen2 support
- SIMD and intrinsics support
- jitstress, jitstressregs, gcstress testing
- GC (not sure about this yet, need to check current status without `DOTNET_GCgen0size=99999999` and with/without `FEATURE_CONSERVATIVE_GC`)
- EH, stackwalking, and interop portions (also not sure, need to check current status)
[...]

PR #91797, opened 2023-09-08
[RISC-V] ELT profiler: fix reconstruction of struct args passed partially on the stack

tomeksowi:
The fix is similar to what [ShuffleIterator is doing](https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/comdelegate.cpp#L193-L201), i.e. if `ArgIterator` returns a general register offset that can't fit `argSize`, this means the other part of the struct was spilled on the stack.

Part of #84834


tomeksowi:
@davmason PR ready for review
\+ @jkotas

davmason (approved):
Profiler test changes LGTM

PR #91802, opened 2023-09-08
[mono][aot] Avoid compiling the same method multiple times during dedup.

vargaz:


kotlarmilos (approved):
LGTM, thanks!

Question -- Hash key is `MonoMethod *method`. Would the same method from two assemblies share the pointer? Probably not?

src/mono/mono/mini/aot-compiler.c
```diff
@@ -4334,14 +4335,17 @@ collect_dedup_method (MonoAotCompile *acfg, MonoMethod *method)
 		if (acfg->dedup_phase == DEDUP_SKIP)
 			return TRUE;
 		// Remember for later
-		if (acfg->dedup_phase == DEDUP_COLLECT && !g_hash_table_lookup (dedup_methods, method))
-			g_hash_table_insert (dedup_methods, method, method);
+		if (acfg->dedup_phase == DEDUP_COLLECT) {
```
tannergooding, line 4338: Is this second check needed given the first?

That is, the outermost `if` on L4334 is `(acfg->dedup_phase == DEDUP_SKIP || acfg->dedup_phase == DEDUP_COLLECT)`

and the check on L4335 unconditionally returns, so this line can only evaluate to true already.

You could seemingly just assert and simplify the nesting by inverting the `if (!g_hash_table_lookup (dedup_methods, method))`

so you end up with:
```cpp
    if ((acfg->dedup_phase == DEDUP_SKIP || acfg->dedup_phase == DEDUP_COLLECT) && mono_aot_can_dedup (method)) {
		if (acfg->dedup_phase == DEDUP_SKIP)
			return TRUE;

        g_assert(acfg->dedup_phase == DEDUP_COLLECT);
        
        if (g_hash_table_lookup (dedup_methods, method))
				return TRUE;
				
        g_hash_table_insert (dedup_methods, method, method);
    }
    return FALSE;
```

vargaz:
> LGTM, thanks!
> 
> Question -- Hash key is `MonoMethod *method`. Would the same method from two assemblies share the pointer? Probably not?

These methods are generic instances, there is only one MonoMethod for each.

PR #91811, opened 2023-09-08
ARM64: Enable jumptable to BT optimization

EgorBo:
"jump-table to BT" optimization is currently limited to XARCH only, this PR enables it for ARM64
```cs
int Test(int x)
{
    switch (x)
    {
        case 11:
        case 13:
        case 15:
        case 17:
        case 18:
        case 20:
            return 1;
    }
    return 2;
}
```
```diff
; Method Prog:Test(int):int:this (FullOpts)	
            stp     fp, lr, [sp, #-0x10]!	
            mov     fp, sp	
            sub     w0, w1, #11	
            cmp     w0, #9	
            bhi     G_M9307_IG05	
-           mov     w0, w0	
-           adr     x1, [@RWD00]	
-           ldr     w1, [x1, x0, LSL #2]	
-           adr     x2, [G_M9307_IG02]	
-           add     x1, x1, x2	
-           br      x1	
-G_M9307_IG03:	
+           mov     w1, #725
+           lsr     w0, w1, w0
+           tbz     w0, #0, G_M9307_IG05
            mov     w0, #1	
            ldp     fp, lr, [sp], #0x10	
            ret     lr	
G_M9307_IG05:	
            mov     w0, #2	
            ldp     fp, lr, [sp], #0x10	
            ret     lr	
-RWD00  dd 00000024h ; case G_M9307_IG03	
-       dd 00000030h ; case G_M9307_IG05	
-       dd 00000024h ; case G_M9307_IG03	
-       dd 00000030h ; case G_M9307_IG05	
-       dd 00000024h ; case G_M9307_IG03	
-       dd 00000030h ; case G_M9307_IG05	
-       dd 00000024h ; case G_M9307_IG03	
-       dd 00000024h ; case G_M9307_IG03	
-       dd 00000030h ; case G_M9307_IG05	
-       dd 00000024h ; case G_M9307_IG03	
-; Total bytes of code: 68
+; Total bytes of code: 56
```

### DIFF
--- a/src/coreclr/jit/lower.cpp
+++ b/src/coreclr/jit/lower.cpp
@@ -1150,10 +1150,6 @@ GenTree* Lowering::LowerSwitch(GenTree* node)
 bool Lowering::TryLowerSwitchToBitTest(
     BasicBlock* jumpTable[], unsigned jumpCount, unsigned targetCount, BasicBlock* bbSwitch, GenTree* switchValue)
 {
-#ifndef TARGET_XARCH
-    // Other architectures may use this if they substitute GT_BT with equivalent code.
-    return false;
-#else
     assert(jumpCount >= 2);
     assert(targetCount >= 2);
     assert(bbSwitch->bbJumpKind == BBJ_SWITCH);
@@ -1237,6 +1233,10 @@ bool Lowering::TryLowerSwitchToBitTest(
         bitTable = ~bitTable;
         std::swap(bbCase0, bbCase1);
     }
+
+    const bool useBt = true;
+#else
+    const bool useBt = false;
 #endif
 
     //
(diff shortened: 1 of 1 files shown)


src/coreclr/jit/lower.cpp
```diff
@@ -1270,20 +1270,32 @@ bool Lowering::TryLowerSwitchToBitTest(
...
+    //
+    // Fallback to AND(RSZ(bitTable, switchValue), 1)
+    //
+    GenTree* tstCns  = comp->gtNewIconNode(bbSwitch->bbNext == bbCase0 ? 0 : 1, bitTableType);
+    GenTree* shift   = comp->gtNewOperNode(GT_RSZ, bitTableType, bitTableIcon, switchValue);
+    GenTree* bitTest = comp->gtNewOperNode(GT_TEST, TYP_VOID, shift, tstCns);
+    bitTest->gtFlags |= GTF_SET_FLAGS;
+    GenTreeCC* jcc = comp->gtNewCC(GT_JCC, TYP_VOID, GenCondition::EQ);
+    LIR::AsRange(bbSwitch).InsertAfter(switchValue, bitTableIcon, shift, tstCns, bitTest);
+    LIR::AsRange(bbSwitch).InsertAfter(bitTest, jcc);
```
jakobbotsch, line 1294: I would create `JTRUE` and continue lowering from the `JTRUE`, agnostically on all platforms. I believe it will produce more optimal codegen for ARM64 (`tbnz`) and the same codegen on x64.

jakobbotsch: Also this logic (with `GT_TEST`) likely does not work on RISC-V and LA64. So it's best to just create normal IR (regular ANDs) and leave it up to regular lowering to optimize it.

EgorBo: Good point, that's why I enabled the whole thing only for XARCH and ARM64 (arm32 doesn't support TEST), will try your suggestion

src/coreclr/jit/lower.cpp
```diff
@@ -1270,20 +1266,31 @@ bool Lowering::TryLowerSwitchToBitTest(
...
+    //
+    GenTree* tstCns = comp->gtNewIconNode(bbSwitch->bbNext != bbCase0 ? 0 : 1, bitTableType);
+    GenTree* shift  = comp->gtNewOperNode(GT_RSZ, bitTableType, bitTableIcon, switchValue);
+    GenTree* one    = comp->gtNewIconNode(1, bitTableType);
+    GenTree* andOp  = comp->gtNewOperNode(GT_AND, bitTableType, shift, one);
+    GenTree* cmp    = comp->gtNewOperNode(GT_EQ, TYP_INT, andOp, tstCns);
+    GenTree* jcc    = comp->gtNewOperNode(GT_JTRUE, TYP_VOID, cmp);
+    LIR::AsRange(bbSwitch).InsertAfter(switchValue, bitTableIcon, shift, tstCns, one);
+    LIR::AsRange(bbSwitch).InsertAfter(one, andOp, cmp, jcc);
+#endif // !TARGET_XARCH
```
jakobbotsch, line 1292: Why do we still need the xarch special case? Does `OptimzeConstCompare` not handle the case? Ideally we would teach it about the missing opportunity so that everyone benefits instead of special casing it here

EgorBo: I thought about it, but I'd like to leave it for a future follow up since it involves more work to add that peephole and I was mainly interesting in improving arm64

EgorBo: Ah, you mean that the peephole already exists?

jakobbotsch: Ok, sounds fine to me. I would personally expect `x & (1 << y)` to be quite common, and it looks like we are missing this opportunity for arm64 (`TEST_EQ/TEST_NE(x, LSH(1, y)) => TEST_EQ/TEST_NE(RSZ(x, y), 1)`). Then this transform could produce `EQ/NE(AND(x, LSH(1, y)))`

jakobbotsch: I don't think it exists, but it seems like adding it here would be straightforward:
https://github.com/dotnet/runtime/blob/e04c75e7c65814f7f4813262effbd56d2c90e5e3/src/coreclr/jit/lower.cpp#L3596-L3621

jakobbotsch: Of course we could also teach it about `TEST_EQ/TEST_NE(RSZ(x, y), 1) => BT(x, y)` on x64/x86, but I guess this is a less common pattern

jakobbotsch: Anyway I'm ok with keeping this PR as is.

EgorBo: @jakobbotsch ok let's then keep as is and I'll file a "good-first-issue" to recognize BT pattern (or will work myself when I have time)

EgorBo:
Diffs: https://dev.azure.com/dnceng-public/public/_build/results?buildId=400689&view=ms.vss-build-web.run-extensions-tab

(bigger in fact because we don't take data section into account)

jakobbotsch (approved):

EgorBo:
Failures are https://github.com/dotnet/runtime/issues/91838

kunalspathak:
Why are the regressions on arm?

EgorBo:
> Why are the regressions on arm?

arm32 doesn't have `tbz` short path, but overall it should still be sensitive size improvements for arm -- all the regressions come with jump-table deleted.

kunalspathak:
> > Why are the regressions on arm?
> 
> arm32 doesn't have `tbz` short path, but overall it should still be sensitive size improvements for arm -- all the regressions come with jump-tables deleted (those aren't counted in `Total bytes of code`)

I am asking because the PR title says that this only impacts Arm64.

EgorBo:
> > > Why are the regressions on arm?
> > 
> > 
> > arm32 doesn't have `tbz` short path, but overall it should still be sensitive size improvements for arm -- all the regressions come with jump-tables deleted (those aren't counted in `Total bytes of code`)
> 
> I am asking because the PR title says that this only impacts Arm64.

it used to, Jakob suggested a fix that enabled it for all archs including arm32, risc-v and la64

Issue #91779, opened 2023-09-08
[Android][arm64] Use of undeclared identifier 'ICALL_SIG_TYPE__Bool'


simonrozsival:
## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=399447
Build error leg or test failing: Build / android-arm64 Release AllSubsets_Mono / Build product
Pull request: https://github.com/dotnet/runtime/pull/91520

Android arm64 builds are failing (x64 and arm builds don't seem to be affected) with the following error:
```
 /__w/1/s/src/mono/mono/metadata/icall.c:7252:1: error: use of undeclared identifier 'ICALL_SIG_TYPE__Bool'; did you mean 'ICALL_SIG_TYPE_bool'?
```

### Known Issue Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json
{
  "ErrorMessage": "(NETCORE_ENGINEERING_TELEMETRY=Build) The command \" cmake --build . --target install --config Release --parallel 4\" exited with code 2.",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=399447
**Error message validated:** `(NETCORE_ENGINEERING_TELEMETRY=Build) The command " cmake --build . --target install --config Release --parallel 4" exited with code 2.`
**Result validation: :white_check_mark:** Known issue matched with the provided build.
**Validation performed at:** 9/8/2023 8:20:10 AM UTC

### Report
|Build|Definition|Step Name|Console log|Pull Request|
|---|---|---|---|---|
|[2263085](https://dev.azure.com/dnceng/internal/_build/results?buildId=2263085)|dotnet-runtime|Build product|[Log](https://dev.azure.com/dnceng/7ea9116e-9fac-403d-b258-b31fcf1bb293/_apis/build/builds/2263085/logs/863)||
[...]

PR #91812, opened 2023-09-08
Revert "[c11] define mono atomics in terms of standard atomics"

lewing:
Reverts dotnet/runtime#91489


mdh1418 (approved):

alexrp:
What's the reason for the revert?

lewing:
> What's the reason for the revert?

https://github.com/dotnet/runtime/issues/91779

PR #89364, opened 2023-07-23
Fix potential BitCast bug for platforms without unaligned reads

MichalPetryka:
BitCast could potentially make an invalid read when reinterpreting a type to another that requires higher alignment, which could possibly cause a fault on Mono ARMv6.

PR #91813, opened 2023-09-08
[mono] Unaligned read/write only when NO_UNALIGNED_ACCESS is set

matouskozak:
Using unaligned read code path only when `NO_UNALIGNED_ACCESS` is set. Otherwise emit regular memory load (`EMIT_NEW_LOAD_MEMBASE_TYPE`) without using `memcpy`.

Edit: Increased the scope of the change to both load/store operations. When CPU supports unaligned memory access (`NO_UNALIGNED_ACCESS` is not set) we let CPU handle the unaligned load/store operation rather than handling it ourselves.

---

Locally confirmed that it fixes `Perf_Matrix3x2.CreateRotation*` benchmarks (https://github.com/dotnet/runtime/issues/91361)

### DIFF
--- a/src/mono/mono/mini/intrinsics.c
+++ b/src/mono/mono/mini/intrinsics.c
@@ -556,7 +556,10 @@ MONO_RESTORE_WARNING
 		t = mini_get_underlying_type (t);
 		if (cfg->gshared && t != ctx->method_inst->type_argv [0] && MONO_TYPE_ISSTRUCT (t) && mono_class_check_context_used (mono_class_from_mono_type_internal (t)))
 			cfg->prefer_instances = TRUE;
-		return mini_emit_memory_load (cfg, t, args [0], 0, MONO_INST_UNALIGNED);
+			
+		int align;
+		int esize = mono_type_size (t, &align);
+		return mini_emit_memory_load (cfg, t, args [0], 0, ((guintptr)&(args [0]->dreg) % align == 0) ? 0 : MONO_INST_UNALIGNED);
 	} else if (!strcmp (cmethod->name, "WriteUnaligned")) {
 		g_assert (ctx);
 		g_assert (ctx->method_inst);
--- a/src/mono/mono/mini/intrinsics.c
+++ b/src/mono/mono/mini/intrinsics.c
@@ -556,10 +556,7 @@ MONO_RESTORE_WARNING
 		t = mini_get_underlying_type (t);
 		if (cfg->gshared && t != ctx->method_inst->type_argv [0] && MONO_TYPE_ISSTRUCT (t) && mono_class_check_context_used (mono_class_from_mono_type_internal (t)))
 			cfg->prefer_instances = TRUE;
-			
-		int align;
-		int esize = mono_type_size (t, &align);
-		return mini_emit_memory_load (cfg, t, args [0], 0, ((guintptr)&(args [0]->dreg) % align == 0) ? 0 : MONO_INST_UNALIGNED);
+		return mini_emit_memory_load (cfg, t, args [0], 0, MONO_INST_UNALIGNED);
 	} else if (!strcmp (cmethod->name, "WriteUnaligned")) {
 		g_assert (ctx);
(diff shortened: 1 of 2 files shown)


src/mono/mono/mini/intrinsics.c
```diff
@@ -556,7 +556,10 @@ MONO_RESTORE_WARNING
 		t = mini_get_underlying_type (t);
 		if (cfg->gshared && t != ctx->method_inst->type_argv [0] && MONO_TYPE_ISSTRUCT (t) && mono_class_check_context_used (mono_class_from_mono_type_internal (t)))
 			cfg->prefer_instances = TRUE;
-		return mini_emit_memory_load (cfg, t, args [0], 0, MONO_INST_UNALIGNED);
+			
+		int align;
+		mono_type_size (t, &align);
+		return mini_emit_memory_load (cfg, t, args [0], 0, ((guintptr)&(args [0]->dreg) % align == 0) ? 0 : MONO_INST_UNALIGNED);
```
filipnavara, line 562: I don't think `((guintptr)&(args [0]->dreg) % align == 0)` does the right thing. That tests address of a member in a structure, so it's essentially a constant expression at type of emit.

You can always emit normal load on x86/x64 because the processors do support unaligned reads internally (with a performance penalty). For everything else it would need to emit a runtime check which is possibly not any better than unrolled memcpy.

matouskozak: Ah, I see. I thought that the `agrs [0]->dreg` points to the actual address we are trying to read from.

src/mono/mono/mini/intrinsics.c
```diff
@@ -556,7 +556,12 @@ MONO_RESTORE_WARNING
 		t = mini_get_underlying_type (t);
 		if (cfg->gshared && t != ctx->method_inst->type_argv [0] && MONO_TYPE_ISSTRUCT (t) && mono_class_check_context_used (mono_class_from_mono_type_internal (t)))
 			cfg->prefer_instances = TRUE;
-		return mini_emit_memory_load (cfg, t, args [0], 0, MONO_INST_UNALIGNED);
+
+		#ifdef NO_UNALIGNED_ACCESS
+			return mini_emit_memory_load (cfg, t, args [0], 0, MONO_INST_UNALIGNED);
+		#else
+			return mini_emit_memory_load (cfg, t, args [0], 0, 0);
+		#endif
```
MichalPetryka, line 564: Doesn't it make more sense to have the check inside of `mini_emit_memory_load`?

matouskozak: I wasn't 100% confident that this solution will be the correct one so for now I put it only for `ReadUnaligned` as it was the immediate solution for the regression. I agree that it would be better to have it inside `mini_emit_memory_load` to optimize all memory load emits. I could move the `#ifdef` to `mini_emit_memory_load` method and zero-out the `MONO_INST_UNALIGNED` flag when `NO_UNALIGNED_ACCESS` is not set. What do you think?

MichalPetryka: I think it should be enough to just if-out the if there. Also, `mini_emit_memory_store` should probably be changed too.

matouskozak:
Microbenchmarks results from x64-Ubuntu22.04
| Microbenchmark                    | ecef854c27861d9d1e25a3d7de76295cd8a92d3b | 65536eb2f01ebb85a54e792a2b974fe76af7e230 | this PR |
|-----------------------------------|------------------------------------------|------------------------------------------|---------|
| CreateRotationBenchmark           | 95.55                                    | 201.7                                    | 95.04   |
| CreateRotationWithCenterBenchmark | 99.09                                    | 209.7                                    | 98.62   |

vargaz (approved):

vargaz:
What change caused the regression ?

matouskozak:
> What change caused the regression ?

It was introduced in https://github.com/dotnet/runtime/pull/89364/. by changing `BitCast` to use `ReadUnaligned`.

PR #91815, opened 2023-09-08
JIT: Optimize JitStressOnly

jakobbotsch:
Calling `MethodSet::contains` on every invocation to `compStressCompile` is very expensive since it requires multiple JIT-EE calls to print the full method name. This merges the logic with the existing cached bool.

I was seeing 5 second tests running in 50+ seconds with a moderately long `JitStressOnly` method set before this change.

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -1789,7 +1789,7 @@ void Compiler::compInit(ArenaAllocator*       pAlloc,
     info.compClassHnd   = compHnd->getMethodClass(methodHnd);
 
 #ifdef DEBUG
-    bRangeAllowStress = false;
+    compAllowStress = true;
 
     // set this early so we can use it without relying on random memory values
     verbose = compIsForInlining() ? impInlineInfo->InlinerCompiler->verbose : false;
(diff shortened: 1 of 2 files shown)


Wraith2:
The name `compAllowStress` feels strange. The usage looks more like `compStressEnabled` or `compStressIncluded`

jakobbotsch:
> The name `compAllowStress` feels strange. The usage looks more like `compStressEnabled` or `compStressIncluded`

How do you figure? Setting this bool to true does not mean the function has stress enabled, that requires a stress mode (`DOTNET_JitStress=XYZ`, `DOTNET_JitStressModeNames=...`). However, setting it to false means that a function is disallowed from having stress modes enabled.

Wraith2:
With that context it makes more sense. Seeing only the code usage it wasn't clear.

jakobbotsch:
cc @dotnet/jit-contrib

kunalspathak (approved):
`superpmi-replay` failures should be unrelated, but do you recall why they were happening initially? The logs of those runs are deleted now, so can't check.

BruceForstall (approved):

jakobbotsch:
> `superpmi-replay` failures should be unrelated, but do you recall why they were happening initially? The logs of those runs are deleted now, so can't check.

I'm pretty sure it was the "out of space" issue, but let me rerun it for good measure.

PR #91818, opened 2023-09-08
Update Double.cs

meokullu:
InNan(double d) method checks if double d is equal to itself. To avoid unnecessary computation, returning false directly will give same result.

Deleting unintentionally added empty space.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Double.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Double.cs
@@ -167,19 +167,11 @@ namespace System
             return (bits & 0x7FFFFFFFFFFFFFFF) == 0x7FF0000000000000;
         }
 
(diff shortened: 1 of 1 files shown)


src/libraries/System.Private.CoreLib/src/System/Double.cs
```diff
@@ -167,19 +167,11 @@ public static unsafe bool IsInfinity(double d)
...
-        {
-            // A NaN will never equal itself so this is an
-            // easy and efficient way to check for NaN.
-
-            #pragma warning disable CS1718
-            return d != d;
-            #pragma warning restore CS1718
-        }
-
+        public static unsafe bool IsNaN(double d) => false;
```
tannergooding, line 173: NaN won't equal itself and so `d != d` returns `true` for `NaN` and `false` otherwise.

Always returning `false` would violate the check and be an invalid change.

PR #91038, opened 2023-08-23
Implement standalone GC loading in NativeAOT

cshung:
This is currently a prototype - the silly naming reflects that. It probably doesn't work in the integrated settings yet, creating this PR to test only.

Highlights:
- [x] Known to work on Windows/Linux targetting 8.0 preview 7 to power ASP.Net BasicMinimalApi benchmark
- [x] Module loading code is confirmed to be not in the final binary using dead code elimination in the linker.
- [x] MSBuild integration

Lowlights: 
- [x] We have to build a NativeAOT-specific variant of the clrgc. (See https://github.com/dotnet/runtime/pull/91821 for the progress towards fixing this)
- [ ] All diagnostics events from CLRGC are not fired 
- [x] Naming of various classes
- [x] Reading module name from configuration
- [x] Fallback to default GC in the standalone mode if the configuration is not set.
- [x] Proper error reporting

@dotnet/gc

PR #91821, opened 2023-09-08
Converge Representations between NativeAOT and CoreCLR

cshung:
To facilitate the work to produce a standalone GC for NativeAOT as in https://github.com/dotnet/runtime/pull/91038, I am making this PR to make the GC view of the runtime look more closely to each other. The eventual goal is to avoid special casing in the GC for NativeAOT as much as possible.

This is a work in progress - checked in the list below means either there is no difference, or the difference is understood and handled so that it is okay.

> With this change, I am able to use a standard release mode clrgc on NativeAOT and it appears to work for BasicMinimalApi

Here is an analysis of how NativeAOT and CoreCLR differ based on analyzing `gcenv.object.h` [here](https://github.com/dotnet/runtime/blob/main/src/coreclr/gc/env/gcenv.object.h):

> Caveat: This analysis is based on reading code only, it maybe not be complete and extra differences might be found by more testing.

- [x] Both NativeAOT and CoreCLR use the `ObjHeader` to represent `BIT_SBLK_GC_RESERVE` and `BIT_SBLK_FINALIZER_RUN` in the same way.

The method table bits:
- [x] NativeAOT never use the `MTFlag_RequireAlign8`, `MTFlag_Category_ValueType_Mask`, `MTFlag_Category_ValueType` flag in the GC (*)
- [x] NativeAOT and CoreCLR use a different bit for `MTFlag_ContainsPointers`, and this change fixed it by moving the NativeAOT bit so that it is the same as the CoreCLR one. 
- [x] NativeAOT and CoreCLR use a different bit for `MTFlag_HasCriticalFinalizer`, and this change fixed it by moving the CoreCLR bit so that it is the same as the NativeAOT one. Care is taken to make sure `clrgc` is still backward compatible with older `coreclr` before the move.  
- [x] NativeAOT and CoreCLR use the same bit for `MTFlag_HasFinalizer` and `MTFlag_HasComponentSize`
- [ ] NativeAOT does not use a bit to determine if it is an array or value type.
- [x] NativeAOT doesn't have collectible types, and this change fixed it by having a reserved bit that is always zero for NativeAOT, also changed CoreCLR so they use the same bit for Collectible.

The method table methods:
- [x] InitializeFreeObject, incorrectly used the `MTFlag_IsArray` flag for NativeAOT, only used for GCSample, okay.
- [x] GetBaseSize, okay
- [x] RawGetComponentSize, okay
- [x] Collectible, fixed by changing bits, okay.
- [x] ContainsPointers, okay
- [x] ContainsPointersOrCollectible, okay
- [x] RequiresAlign8, not called for NativeAOT, okay
- [x] IsValueType, not called for NativeAOT, okay
- [x] HasComponentSize, okay
- [x] HasFinalizer, okay
- [x] HasCriticalFinalizer, okay
- [x] IsArray, deleted, okay
- [x] GetParent, deleted, okay
- [x] SanityCheck, okay

(*) As of the current code, `pMT->RequiresAlign8()` and `pMT->IsValueType()` is used only in validation, and it is ifdef out by NativeAOT.

### DIFF
--- a/src/coreclr/gc/env/gcenv.object.h
+++ b/src/coreclr/gc/env/gcenv.object.h
@@ -46,15 +46,15 @@ public:
 
 static_assert(sizeof(ObjHeader) == sizeof(uintptr_t), "this assumption is made by the VM!");
 
-#define MTFlag_RequireAlign8            0x00001000
-#define MTFlag_Category_ValueType       0x00040000
-#define MTFlag_Category_ValueType_Mask  0x000C0000
-#define MTFlag_ContainsPointers         0x01000000
-#define MTFlag_HasCriticalFinalizer     0x00000002
-#define MTFlag_HasFinalizer             0x00100000
-#define MTFlag_IsArray                  0x00080000
-#define MTFlag_Collectible              0x00200000
-#define MTFlag_HasComponentSize         0x80000000
+#define MTFlag_RequiresAlign8           0x00001000 // enum_flag_RequiresAlign8
+#define MTFlag_Category_ValueType       0x00040000 // enum_flag_Category_ValueType
+#define MTFlag_Category_ValueType_Mask  0x000C0000 // enum_flag_Category_ValueType_Mask
+#define MTFlag_ContainsPointers         0x01000000 // enum_flag_ContainsPointers
+#define MTFlag_HasCriticalFinalizer     0x00000002 // enum_flag_HasCriticalFinalizer
+#define MTFlag_HasFinalizer             0x00100000 // enum_flag_HasFinalizer
+#define MTFlag_IsArray                  0x00080000 // enum_flag_Category_Array
+#define MTFlag_Collectible              0x00200000 // enum_flag_Collectible
+#define MTFlag_HasComponentSize         0x80000000 // enum_flag_HasComponentSize
 
 class MethodTable
 {
@@ -112,7 +112,7 @@ public:
 
     bool RequiresAlign8()
     {
-        return (m_flags & MTFlag_RequireAlign8) != 0;
+        return (m_flags & MTFlag_RequiresAlign8) != 0;
     }
 
     bool IsValueType()
@@ -147,7 +147,7 @@ public:
             return (m_flags & Old_MTFlag_HasCriticalFinalizer) != 0;
         }
 #endif
-        return (m_flags & MTFlag_HasCriticalFinalizer) && !HasComponentSize();
+        return !HasComponentSize() && (m_flags & MTFlag_HasCriticalFinalizer);
     }
 
     bool SanityCheck()
--- a/src/coreclr/gc/env/gcenv.object.h
+++ b/src/coreclr/gc/env/gcenv.object.h
@@ -4,6 +4,10 @@
 #ifndef __GCENV_OBJECT_H__
 #define __GCENV_OBJECT_H__
 
+#ifdef BUILD_AS_STANDALONE
+extern bool g_oldMethodTableFlags;
+#endif
+
 // ARM requires that 64-bit primitive types are aligned at 64-bit boundaries for interlocked-like operations.
 // Additionally the platform ABI requires these types and composite type containing them to be similarly
 // aligned when passed as arguments.
@@ -46,7 +50,7 @@ static_assert(sizeof(ObjHeader) == sizeof(uintptr_t), "this assumption is made b
 #define MTFlag_Category_ValueType       0x00040000
 #define MTFlag_Category_ValueType_Mask  0x000C0000
 #define MTFlag_ContainsPointers         0x01000000
-#define MTFlag_HasCriticalFinalizer     0x08000000
+#define MTFlag_HasCriticalFinalizer     0x00000002
 #define MTFlag_HasFinalizer             0x00100000
 #define MTFlag_IsArray                  0x00080000
 #define MTFlag_Collectible              0x10000000
@@ -127,12 +131,15 @@ public:
 
     bool HasCriticalFinalizer()
     {
-#ifdef FEATURE_NATIVEAOT
-        const int HasCriticalFinalizerFlag = 0x0002;
-        return (m_flags & HasCriticalFinalizerFlag) && !HasComponentSize();
-#else
-        return (m_flags & MTFlag_HasCriticalFinalizer) != 0;
+#ifdef BUILD_AS_STANDALONE
+        if (g_oldMethodTableFlags)
+        {
+            // This flag is used for .NET 8 or below
+            const int Old_MTFlag_HasCriticalFinalizer = 0x08000000;
+            return (m_flags & Old_MTFlag_HasCriticalFinalizer) != 0;
+        }
 #endif
+        return (m_flags & MTFlag_HasCriticalFinalizer) && !HasComponentSize();
     }
 
     bool IsArray()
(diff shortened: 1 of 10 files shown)


src/coreclr/gc/objecthandle.cpp
```diff
@@ -1641,7 +1641,7 @@ void CALLBACK ScanPointer(_UNCHECKED_OBJECTREF *pObjRef, uintptr_t *pExtraInfo,
     promote_func* callback = (promote_func*)lp2;
     callback(pRef, (ScanContext *)lp1, 0);
 }
-
+#ifdef FEATURE_NATIVEAOT
```
jkotas, line 1644: Is this ifdef really needed?

src/coreclr/gc/env/gcenv.object.h
```diff
@@ -127,7 +127,12 @@ class MethodTable
 
     bool HasCriticalFinalizer()
     {
+#ifdef FEATURE_NATIVEAOT
+        const int HasCriticalFinalizerFlag = 0x0002;
```
jkotas, line 131: Can we unify the critical finalizer flag too?

src/coreclr/gc/objecthandle.cpp
cshung: Not really, the linker will probably make sure this method doesn't get to the `coreclr.dll`. But I figure it would be easier for the reader to know this method is not needed in the standard CoreCLR world and that `HNDTYPE_VARIABLE` is a NativeAOT-only thing. That was done as part of the exercise to determine what differences are there between CoreCLR and NativeAOT.

src/coreclr/gc/env/gcenv.object.h
cshung: I wish, but I don't know how though. The ideal location for that bit is currently being occupied by `ElementType`, which needs to be there regardless of `HasComponentSize`.

src/coreclr/gc/objecthandle.cpp
jkotas: > `HNDTYPE_VARIABLE` is a NativeAOT-only thing

`HNDTYPE_VARIABLE` is not used by NativeAOT (see https://github.com/dotnet/runtime/discussions/86840).

If it helps, I would be fine with deleting it.

src/coreclr/gc/env/gcenv.object.h
jkotas: If you want to keep the CoreCLR intact, you should be able to reshuffle the NativeAOT Flags bits to place the critical finalizer bit into the same spot as where it is in CoreCLR,

It will work, but it is not ideal. The spot that CoreCLR uses for the critical finalizer bit is not the best - it is rarely checked bit, set only on a few types. It would be better to store the bit in the "low bits" area. I guess that you are not keen on changing the bits for CoreCLR because of the constraints on the GC/EE interface evolution that you are trying to fit into.

cshung: Right now, clrgc can work with .NET 6+. I would hope to keep it that way as long as .NET 6 is still supported. If there is a way that we can eventually make it such that we can move the bit when .NET 6 is no longer supported, that would be good.

jkotas: You can make it work by adding version checks in standalone GC impl like `if (version > something) check this bit; else check other bit;`. The version checks like this are the only how to evolve the interface while having one clrgc binary that works on multiple version.

src/coreclr/gc/objecthandle.cpp
cshung: I can't actually delete `Ref_ScanPointers` because it is actually called by NativeAOT [here](https://github.com/dotnet/runtime/blob/671d42c39cc85f3a4c33f4feff2c556415b19395/src/coreclr/nativeaot/Runtime/gcrhenv.cpp#L502).

jkotas: This is all dead code. I have submitted PR to clean this up: #91847

src/coreclr/gc/env/gcenv.object.h
```diff
@@ -127,7 +131,15 @@ class MethodTable
...
-        return (m_flags & MTFlag_HasCriticalFinalizer) != 0;
+#ifdef BUILD_AS_STANDALONE
+        if (g_oldMethodTableFlags)
+        {
+            // This flag is used for .NET 8 or below
+            const int Old_MTFlag_HasCriticalFinalizer = 0x08000000;
+            return (m_flags & Old_MTFlag_HasCriticalFinalizer) != 0;
+        }
+#endif
+        return (m_flags & MTFlag_HasCriticalFinalizer) && !HasComponentSize();
```
Maoni0, line 142: >return (m_flags & MTFlag_HasCriticalFinalizer) && !HasComponentSize(); [](http://example.com/codeflow?start=8&length=70)

nit - it'd be a bit better to check for the array type first just to make it clear that if it's an array it cannot have a finalizer

src/coreclr/vm/methodtable.h
```diff
@@ -3401,8 +3403,11 @@ public :
 
         enum_flag_IsTrackedReferenceWithFinalizer   = 0x04000000,
 
-        enum_flag_HasCriticalFinalizer        = 0x08000000, // finalizer must be run on Appdomain Unload
-        enum_flag_Collectible                 = 0x10000000,
+        // TODO, andrewau, move fields to reflect value changes
```
davidwrighton, line 3406: What is this TODO for?

src/coreclr/vm/methodtable.h
```diff
@@ -3342,6 +3342,7 @@ public :
         // As you change the flags in WFLAGS_LOW_ENUM you also need to change this
         // to be up to date to reflect the default values of those flags for the
         // case where this MethodTable is for a String or Array
+        // TODO, AndrewAu, this looks important, do it!
```
davidwrighton, line 3345: Since you are not adjusting the set of flags, you should not need to change anything here.

src/coreclr/vm/methodtable.h
cshung: I wonder if I should make `enum_flag_HasCriticalFinalizer` part of `WFLAGS_LOW_ENUM` and therefore need change for the code above.

src/coreclr/nativeaot/Runtime/DebugHeader.cpp
```diff
@@ -269,7 +269,8 @@ extern "C" void PopulateDebugHeaders()
 
     static_assert(MethodTable::Flags::EETypeKindMask         == 0x00030000, "The debugging data contract has a hard coded dependency on this value of MethodTable::Flags. If you change this value you must bump major_version_number.");
     static_assert(MethodTable::Flags::HasFinalizerFlag       == 0x00100000, "The debugging data contract has a hard coded dependency on this value of MethodTable::Flags. If you change this value you must bump major_version_number.");
-    static_assert(MethodTable::Flags::HasPointersFlag        == 0x00200000, "The debugging data contract has a hard coded dependency on this value of MethodTable::Flags. If you change this value you must bump major_version_number.");
+    // TODO, andrewau, bump version number, testing
```
mikem8361, line 272: You need to bump the version number [here](https://github.com/dotnet/runtime/blob/8a5283fcf1526e80255a35acbb8f9bc362c9f050/src/coreclr/nativeaot/Runtime/DebugHeader.cpp#L80).

davidwrighton (approved):

PR #91832, opened 2023-09-09
The IConfiguration.GetSection method will not return null, call the IConfigurationSection.Exists extension method to determine.

AlexKuai:
The implementation type of IConfiguration always returns a ConfigurationSection instance in the implementation of the GetSection method, regardless of whether the configuration section exists or not.

Should the IConfigurationSection.Exists extension method be called here to determine whether the configuration section exists？


tarekgh:
@AlexKuai what exactly you are trying to fix here? is it just optimization? if so, could you please provide some benchmark numbers to see if it is worth changing it?

Issue #67405, opened 2022-03-31
SegFault while trying to Copy native test components in monointerpreter test run build


joperezr:
Latest dotnet\runtime Rolling Build failed due to a problem when trying to copy native test components on the `Mono monointerpreter Pri0 Runtime Tests Run OSX x64 release` job. Bellow are more details to help with diagnosing.

Builds affected

[20220331.4](https://dev.azure.com/dnceng/public/_build/results?buildId=1692244) => [Log](https://dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_apis/build/builds/1692244/logs/1965)

```
2022-03-31T09:09:29.3477940Z     Copying native test binaries...
2022-03-31T09:09:46.6888520Z   /Users/runner/work/1/s/dotnet.sh: line 27:  2070 Segmentation fault: 11  ${dotnetPath} "$@"
2022-03-31T09:09:46.6991160Z /Users/runner/work/1/s/src/tests/build.proj(494,5): error MSB3073: The command ""/Users/runner/work/1/s/dotnet.sh" msbuild /Users/runner/work/1/s/src/tests/build.proj /t:CopyAllNativeTestProjectBinaries "/p:TargetArchitecture=x64" "/p:Configuration=Release" "/p:LibrariesConfiguration=Release" "/p:TargetOS=OSX" "/p:RuntimeOS=" "/p:RuntimeFlavor=mono" "/p:RuntimeVariant=" "/p:CLRTestBuildAllTargets=" "/p:__TestGroupToBuild=1" "/p:__SkipRestorePackages=1" /nodeReuse:false /maxcpucount /bl:/Users/runner/work/1/s/artifacts//log/Release/InnerManagedTestBuild.1.binlog" exited with code 139.
2022-03-31T09:09:46.7194120Z 
2022-03-31T09:09:46.7256460Z Build FAILED.
```

cc: @radical  can you please take a look and tag relevant folks?

radical:
cc @steveisok @trylek

trylek:
Hmm, unless I'm mistaken this looks like a (previously known but exceedingly rare) crash of the dotnet.exe app; I doubt we can shed more light on this without being able to capture a crash dump so I guess that is the primary goal this issue should track.

PR #91844, opened 2023-09-09
Remove unnecessary rid argument to dotnet-restore

am11:
We build managed tests with AnyCPU configuration. This `-r PackageRID` causes errors during the restore when cross building for different OS (`-os $other` and not different arch).


am11:
mono interpreter segmentation fault seems unrelated (possibly https://github.com/dotnet/runtime/issues/67405).

ViktorHofer (approved):

agocke (approved):
LGTM

PR #91845, opened 2023-09-09
Fix old gcc build

am11:
gcc 4.9 builds coreclr with this patch.

### DIFF
--- a/src/coreclr/jit/hwintrinsicxarch.cpp
+++ b/src/coreclr/jit/hwintrinsicxarch.cpp
@@ -3617,7 +3617,8 @@ GenTree* Compiler::impSpecialIntrinsic(NamedIntrinsic        intrinsic,
                         impAppendTree(gtUnusedValNode(*val3), CHECK_SPILL_ALL, impCurStmtDI);
                     }
 
-                    switch (info.oper1)
+                    // cast in switch clause is needed for old gcc
+                    switch ((TernaryLogicOperKind)info.oper1)
                     {
                         case TernaryLogicOperKind::Select:
                         {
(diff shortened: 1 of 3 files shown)


src/coreclr/jit/hwintrinsicxarch.cpp
```diff
@@ -3617,7 +3617,8 @@ GenTree* Compiler::impSpecialIntrinsic(NamedIntrinsic        intrinsic,
                         impAppendTree(gtUnusedValNode(*val3), CHECK_SPILL_ALL, impCurStmtDI);
                     }
 
-                    switch (info.oper1)
+                    // cast in switch clause is needed for old gcc
```
AaronRobinsonMSFT, line 3620:
```suggestion
```

This is just good code. No one is going to remove that cast.

AaronRobinsonMSFT (approved):

janvorli (approved):
LGTM, thank you!

src/coreclr/jit/hwintrinsicxarch.cpp
am11: We added a similar comment in past:
```
src/coreclr/debug/createdump/crashinfounix.cpp:                    // explicit initialization for old gcc support; instead of just runtimeInfo { }
src/coreclr/jit/gentree.cpp:    // static_cast to "enum class" for old gcc support 
```

basically if someone removes these casts, the CI will be green; hence the reminder comment.

jakobbotsch: Given that `oper1` is already of type `TernaryLogicOperKind` the cast does seem strange to me on first glance, and I would probably remove it if I saw it.

By the way, what is the benefit of keeping this compatibility with the old GCC version?

am11: > By the way, what is the benefit of keeping this compatibility with the old GCC version?

gcc 4.9 and clang 3.5 are listed as the lowest versions to build runtime in init-compiler script: https://github.com/dotnet/runtime/blob/02feed083350b66c6b7a956343612c1ebfbadba4/eng/common/native/init-compiler.sh#L66-L67

So far, we have maintained the build with older toolchains. Going by the code changes, it does not seem too hard to continue supporting them. IMO, until we have a reason to give up the older toolchains support (e.g. for C++ 17 support), we should keep the option.

agocke:
Thanks!

PR #91847, opened 2023-09-09
Cleanup defines for specialized handle types

jkotas:
- Introduce FEATURE_VARIABLE_HANDLES, FEATURE_REFCOUNTED_HANDLES and FEATURE_WEAK_NATIVE_COM_HANDLES defines to control compilation of support for these specialized handle types.
- Stop compiling FEATURE_VARIABLE_HANDLES and FEATURE_WEAK_NATIVE_COM_HANDLES support into the live runtime since these handle types are not currently used. (The support is still compiled into standalone GC for backward compatibility.)
- Delete some dead code


src/coreclr/gc/objecthandle.cpp
```diff
@@ -66,7 +66,7 @@ void CALLBACK VariableTraceDispatcher(_UNCHECKED_OBJECTREF *pObjRef, uintptr_t *
     }
 }
 
-#if defined(FEATURE_COMINTEROP) || defined(FEATURE_COMWRAPPERS) || defined(FEATURE_OBJCMARSHAL) || defined(FEATURE_NATIVEAOT)
```
jkotas, line 69: This deletes number of `FEATURE_NATIVEAOT` ifdefs in the GC. `FEATURE_NATIVEAOT` ifdefs in the GC are anti-patern.

src/coreclr/gc/gcinterface.h
```diff
@@ -474,9 +472,8 @@ typedef enum
      * useful when passing a pointer to object innards out of the runtime while GC
      * may be enabled.
```
teo-tsirpanis, line 473: The documentation is identical to `HNDTYPE_PINNED`. Would it be worth it to add some text to differentiate them?

jkotas: I have updated the comment. (I expect that we are going to delete the support for async pinned handles at some point - it is not worth it to put much effort into it.)

src/coreclr/gc/objecthandle.cpp
```diff
@@ -424,9 +427,9 @@ void CALLBACK ScanPointerForProfilerAndETW(_UNCHECKED_OBJECTREF *pObjRef, uintpt
         break;
     case    HNDTYPE_WEAK_SHORT:
     case    HNDTYPE_WEAK_LONG:
-#ifdef FEATURE_COMINTEROP
+#ifdef FEATURE_WEAK_NATIVE_COM_HANDLES
     case    HNDTYPE_WEAK_NATIVE_COM:
-#endif // FEATURE_COMINTEROP
+#endif
```
mangod9, line 432:
```suggestion
#endif // FEATURE_WEAK_NATIVE_COM_HANDLES
```

src/coreclr/gc/objecthandle.cpp
```diff
@@ -1015,6 +993,7 @@ void TraceVariableHandlesBySingleThread(HANDLESCANPROC pfnTrace, uintptr_t lp1,
         walk = walk->pNext;
     }
 }
+#endif
```
mangod9, line 996:
```suggestion
#endif // FEATURE_VARIABLE_HANDLES
```

src/coreclr/gc/objecthandle.cpp
```diff
@@ -439,29 +442,27 @@ void CALLBACK ScanPointerForProfilerAndETW(_UNCHECKED_OBJECTREF *pObjRef, uintpt
...
+            }
+            if ((nVarHandleType & VHT_PINNED) != 0)
+            {
+                rootFlags |= kEtwGCRootFlagsPinning;
+            }
 
-        // No special ETW flag for strong handles (VHT_STRONG)
-    }
-#else
-        _ASSERTE(!"Variable handle encountered");
```
mangod9, line 460: assume this assert is no longer necessary?

jkotas: It can be moved to the `default:` case.

cshung:
I have found some more opportunities for cleanup here https://github.com/cshung/runtime/commit/590eae65a5e79df0ab17fa68a4f4e7327c6e1731, can you please take a look?

jkotas:
> I have found some more opportunities for cleanup here https://github.com/cshung/runtime/commit/590eae65a5e79df0ab17fa68a4f4e7327c6e1731, can you please take a look?

Thank you. I have pushed a commit inspired by your change:
- Deleted runtime support for handles that the runtime no longer uses instead of ifdefing it.
- I have omitted the change that added the `FEATURE_...HANDLE` ifdef into the contract itself. The `FEATURE_...HANDLE` ifdefs have applied to the implementation only and ifdefs do not work well in contracts.

mangod9 (approved):

PR #91865, opened 2023-09-11
dumpwriterelf: increase buffer to fit a RHEL8 arm64 page.

### LINKED ISSUE #91864: createdump segfaults on RHEL8 arm64 with .NET 8
With .NET 8, `createdump` gives a segfault on RHEL8 arm64.

Attaching the debugger shows that `DumpWriter` is trying to write a 64kB block here:

https://github.com/dotnet/runtime/blob/de0ab156194eb64deae0e1018db9a58f7b02f4a3/src/coreclr/debug/createdump/dumpwriterelf.cpp#L181

using the 16kB `m_tempBuffer` allocated here:

https://github.com/dotnet/runtime/blob/de0ab156194eb64deae0e1018db9a58f7b02f4a3/src/coreclr/debug/createdump/dumpwriterelf.h#L45

This leads to stack corruption.

arm64 on RHEL8 has an atypical page size of  64kB pages.

When I update `m_tempBuffer` to have a 64kB size, `createdump` works.

cc @mikem8361 @omajid

tmds:
Fixes https://github.com/dotnet/runtime/issues/91864.

@mikem8361 ptal.

### DIFF
--- a/src/coreclr/debug/createdump/crashinfo.cpp
+++ b/src/coreclr/debug/createdump/crashinfo.cpp
@@ -195,7 +195,7 @@ CrashInfo::GatherCrashInfo(DumpType dumpType)
         return false;
     }
     // Add the special (fake) memory region for the special diagnostics info
-    MemoryRegion special(PF_R, SpecialDiagInfoAddress, SpecialDiagInfoAddress + PAGE_SIZE);
+    MemoryRegion special(PF_R, SpecialDiagInfoAddress, SpecialDiagInfoAddress + SpecialDiagInfoSize);
     m_memoryRegions.insert(special);
 #ifdef __APPLE__
     InitializeOtherMappings();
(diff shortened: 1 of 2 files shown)


hoyosjs (approved):

mikem8361:
We need to change the MacOS dump writer's m_tempBuffer in dumpwritermacho.h to 64K also for arm64/M1.  I think M1 has a 64K PAGE_SIZE.

mikem8361:
Another way to fix this is to change the size of the SpecialDiagInfo block to 4K always instead of PAGE_SIZE [here]( https://github.com/dotnet/runtime/blob/02feed083350b66c6b7a956343612c1ebfbadba4/src/coreclr/debug/createdump/crashinfo.cpp#L198).  Adding a SPECIAL_DIAGINFO_SIZE define to specialdiaginfo.h is probably a good idea if you make this change.

tmds:
> Another way to fix this is to change the size of the SpecialDiagInfo block to 4K always instead of PAGE_SIZE [here](https://github.com/dotnet/runtime/blob/02feed083350b66c6b7a956343612c1ebfbadba4/src/coreclr/debug/createdump/crashinfo.cpp#L198). Adding a SPECIAL_DIAGINFO_SIZE define to specialdiaginfo.h is probably a good idea if you make this change.

I've adopted this suggestion, and validated it works on RHEL 8 arm64.

mikem8361 (approved):

PR #91876, opened 2023-09-11
Improve the trim incompatibility message for Assembly.GetType

vitek-karas:
Currently calling `Assembly.GetType` is always a warning when trimming. But typically the code can be changed to use `Type.GetType` with assembly qualified type name, which is a trim compatible pattern.

This modifies the warning message on the `RequiresUnreferencedCode` attribute for `Assembly.GetType` to suggest using `Type.GetType` instead.

### DIFF
--- a/src/coreclr/System.Private.CoreLib/src/System/Reflection/RuntimeAssembly.cs
+++ b/src/coreclr/System.Private.CoreLib/src/System/Reflection/RuntimeAssembly.cs
@@ -242,7 +242,7 @@ namespace System.Reflection
             return type;
         }
 
-        [RequiresUnreferencedCode("Types might be removed by trimming. Consider using Type.GetType instead.")]
+        [RequiresUnreferencedCode("Types might be removed by trimming. If the type name is a string literal, consider using Type.GetType instead.")]
         public override Type? GetType(
             string name, // throw on null strings regardless of the value of "throwOnError"
             bool throwOnError, bool ignoreCase)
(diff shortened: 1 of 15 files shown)


eerhardt:
Don't you need to update the `ref` assembly code as well?

vitek-karas:
> Don't you need to update the ref assembly code as well?

Of course - I was... dumb. Should be fixed now.

src/coreclr/System.Private.CoreLib/src/System/Reflection/RuntimeAssembly.cs
```diff
@@ -242,7 +242,7 @@ public override AssemblyName GetName(bool copiedName)
             return type;
         }
 
-        [RequiresUnreferencedCode("Types might be removed")]
+        [RequiresUnreferencedCode("Types might be removed by trimming. Consider using Type.GetType instead.")]
```
MichalStrehovsky, line 245: If we're updating CoreCLR's RuntimeAssembly, should we also update Mono's and NativeAOT's?

MichalStrehovsky (commented):
If we want to be prescriptive in the message, should we change this to `Types might be removed by trimming. If the type name is a string literal, consider using Type.GetType instead.`.

Basically the advice to use `Type.GetType` is not useful if this is not a string constant that we can analyze or is properly annotated.

src/coreclr/System.Private.CoreLib/src/System/Reflection/RuntimeAssembly.cs
vitek-karas: Done - thanks for reminding me

vitek-karas:
Thanks for the suggestion @MichalStrehovsky  - I changed the message everywhere.

MichalStrehovsky (approved):
Thank you!

PR #91884, opened 2023-09-11
Remove Latin1CharSearchValues

MihaZupan:
After #89981, this implementation would only end up being chosen if neither X86 nor Arm HW intrinsics were available.
Throughput-wise, the scalar loops perform almost identically (assuming we don't have too many bloom filter false positives to deal with in the prob map).


stephentoub (approved):

stephentoub:
Thanks for following up on our discussion.

Might be worth backporting this to release/8.0 as well, to avoid shipping code we don't actually want anymore.

jeffhandley (approved):
This looks like a valuable reliability improvement to me. I will go ahead and merge this into `main` and check with @artl93 on porting it to release/8.0.

lewing:
Just want to add a note here that this causes a pretty large perf regression on wasm if simd has to be disabled so I'd prefer to keep it around in net8 if possible

MihaZupan:
We were using this loop before: [Latin1](https://github.com/dotnet/runtime/blob/d34a059b42a333956dfab085c8011b8d05ba5066/src/libraries/System.Private.CoreLib/src/System/SearchValues/Latin1CharSearchValues.cs#L55-L64)
Now we'd be using this: [ProbabilisticMap](https://github.com/dotnet/runtime/blob/d34a059b42a333956dfab085c8011b8d05ba5066/src/libraries/System.Private.CoreLib/src/System/SearchValues/ProbabilisticMap.cs#L311-L320)

I'm guessing this doesn't have so much to do with the throughput of the two methods then, but rather that the ProbabilisticMap must do a `values.Contains(char)` for every match. If there's no vectorization and there's a bunch of values, that could show up.

stephentoub:
> Just want to add a note here that this causes a pretty large perf regression on wasm AOT if simd has to be disabled

Which perf test in particular?

lewing:
to be specific there was a regression in this range
https://github.com/dotnet/runtime/compare/6504cdb355b8bfbd0f1ae316ffe87ecddedc5dbc...4a09c82215399c27f52277a8db7178270410c693
of this benchmark
https://github.com/dotnet/runtime/blob/7f8da5754e382fcc5d0992a3f72b17d4ee1e9b05/src/mono/sample/wasm/browser-bench/Span.cs#L87-L101
when wasm SIMD is disabled on AOT
image
I'm assuming this is the commit that caused it

stephentoub:
I don't see how this PR would impact that test.

lewing:
it was likely https://github.com/dotnet/runtime/pull/91887

lewing:
feel free to backport

stephentoub:
But the cited test doesn't involve SearchValues at all.

lewing:
Those are the commits, it regressed in firefox and chrome without simd+eh I'll look more closely when I have time

MihaZupan:
> it was likely https://github.com/dotnet/runtime/pull/91887

It definitely wasn't that given that the `SearchValues<string>` API is not yet used anywhere.

lewing:
I'm sorry for the noise, It is quite confusing to me too.

MihaZupan:
@lewing could [this benchmark](https://github.com/dotnet/runtime/pull/91884#issuecomment-1718392962) be running with `Icu = true` and `TARGET_BROWSER = false`?

I think that's the only possible configuration where you'd hit `SearchValues` with an Invariant `IndexOf` in [`CompareInfo.Icu.cs`](https://github.com/dotnet/runtime/blob/d34ed289d34b2f9588e10673800d269d7b1a6376/src/libraries/System.Private.CoreLib/src/System/Globalization/CompareInfo.Icu.cs#L230-L240)

lewing:
It would be Icu = true and TARGET_BROWSE=1 I believe

MihaZupan:
In that case it could be this PR with the usages here
https://github.com/dotnet/runtime/blob/d34ed289d34b2f9588e10673800d269d7b1a6376/src/libraries/System.Private.CoreLib/src/System/Globalization/CompareInfo.Icu.cs#L230-L237


Given this logic is there only as a "fast path", it could make sense to guard it behind `Vector128.IsHardwareAccelerated` to avoid it becoming a deoptimization.

stephentoub:
> In that case it could be this PR with the usages here
> 
> https://github.com/dotnet/runtime/blob/d34ed289d34b2f9588e10673800d269d7b1a6376/src/libraries/System.Private.CoreLib/src/System/Globalization/CompareInfo.Icu.cs#L230-L237

Nice find. I could not have told you we have that fast-path there to avoid the interop call.

MihaZupan:
> on wasm AOT if simd has to be disabled

Is this a temporary limitation of some kind, or are we expecting that we'll have a target without any HW acceleration support going forward?

Issue #68028, opened 2022-04-14
Review the multi-op instruction usage for Arm64


tannergooding:
After seeing the `msub` PR (https://github.com/dotnet/runtime/pull/66621) which folds `mul, sub` into a single `msub` I went and looked in the Arm64 manual for other interesting "combined operation" instructions.

There is a set of `(shifted register)` instructions which can combine a `shift, op` (the variants ending with `s` also set flags):
- [x] `add`, `adds` - Add
- [x] `sub`, `subs` - Subtract
- [x] `cmp` - Compare [#84605](https://github.com/dotnet/runtime/pull/84605)
- [x] `cmn` - Compare Negative [#84667](https://github.com/dotnet/runtime/pull/84667)
- [x] `neg`, `negs` - Negate [#84667](https://github.com/dotnet/runtime/pull/84667)
- [x] `and`, `ands` - Bitwise AND
- [x] `bic`, `bics` - Bitwise bit clear
- [x] `eon` - Bitwise exclusive OR NOT 
- [x] `eor` - Bitwise exclusive OR
- [x] `orr` - Bitwise inclusive OR
- [x] `mvn` - Bitwise NOT  
- [x] `orn` - Bitwise inclusive OR NOT 
- [x] `tst` - Test Bits  
- [x] `cbz` instead of `cmp #0` #42514

There is a set of `(extended register)` instructions which can combine a `zero-extend, op` or `sign-extend, op`:
- [x] `add`, `adds` - Add
- [x] `sub`, `subs` - Subtract
- [x] `cmp` - Compare
- [x] `cmn` - Compare Negative

There is a set of `(carry)` instructions which can utilize the carry from a previous operation:
- [ ] `adc`, `adcs` - Add with carry
- [ ] `sbc`, `sbcs` - Subtract with carry
- [ ] `ngc`, `ngcs` - Negate with carry

There are the `multiply integer` instructions which can combine an `op, mul`:
- [x] `madd` - Multiply-add
- [x] `msub` - Multiply-subtract
- [x] `mneg` - Multiply-negate

There are then some `long multiply` instructions which can return a product twice the size of the inputs (`i32 * i32 = i64` or similar; effectively good for covering `zero or sign extend, multiply`):
- [ ] `smull`, `umull` - Multiply long
- [x] `smaddl`, `umaddl` - Multiply-add long https://github.com/dotnet/runtime/pull/91886
[...]

PR #91886, opened 2023-09-11
Optimise long multiply + add/sub/neg on arm64.

c272:
This PR optimises an extending multiply of two integers out to a long followed by an addition, subtraction or negation operation down to a single `[s/u]maddl`, `[s/u]subl` or `smnegl` instruction. It also adds a generic ternary operator which can be used in the future to support other three argument nodes.

The `spmidiff` results for this patch can be found at the below gist (performed on `win-arm64` since `linux-arm64` is currently broken as per #91257):
https://gist.github.com/c272/944721d2d083660e0ba7ec9fddb24de4

Minor regressions seen here are due to additional `mov` instructions changing the register allocation in cases where the add value is containable, preventing some previously performed `ldp` optimisations. Eliminating all cases of containable adds also removes many valid optimisations, so I have currently left it as is.

Contributes to https://github.com/dotnet/runtime/issues/68028


src/coreclr/jit/codegenarm64.cpp
```diff
@@ -2521,6 +2521,84 @@ void CodeGen::genCodeForMulHi(GenTreeOp* treeNode)
     genProduceReg(treeNode);
 }
 
+//------------------------------------------------------------------------
+// genCodeForMulAddSubLong: Generates code for int*int + long -> long.
```
kunalspathak, line 2525: also include an example for `SUB`.

kunalspathak (commented):
It seems adding `GenTreeTernOp` can be avoided. In the past, we directly recognized the pattern in codegen e.g. `madd`, `msub`. Did you try that approach and faced any problem?

a74nh:
> It seems adding `GenTreeTernOp` can be avoided. In the past, we directly recognized the pattern in codegen e.g. `madd`, `msub`. Did you try that approach and faced any problem?

I guess that would be possible, but it feels like it should be done during lowering (I have no overly strong opinion though).

My thought for `GenTreeTernOp` is that a follow on patch could move `SELECT`, `CMPXCHG` over to use it, which should simplify things a little.

tannergooding:
Rather than `GenTreeTernOp`, we already have `GenTreeHWIntrinsic` which supports multi-op and which is really oriented around these types of specialized instructions for both SIMD and Scalar code.

We can just introduce an `ArmBase_LongMultiplyAdd` or similar for internal use only and have lowering generate that node. The existing metadata table will make this very simple to support and likely require less overall code changes.

c272:
@tannergooding Thank you for the comment, I have reworked the patch to instead utilise `GT_HWINTRINSIC` instead of creating a new ternary node. The updated patch should also resolve the testing failures.

src/coreclr/jit/hwintrinsiccodegenarm64.cpp
```diff
@@ -486,6 +486,14 @@ void CodeGen::genHWIntrinsic(GenTreeHWIntrinsic* node)
                 break;
             }
 
+            case NI_ArmBase_Arm64_MultiplyLongAdd:
+                ins = node->TypeIs(TYP_ULONG) ? INS_umaddl : INS_smaddl;
```
kunalspathak, line 490: should this be `varTypeIsUnsigned(intrin.baseType)`?

src/coreclr/jit/hwintrinsiccodegenarm64.cpp
```diff
@@ -1112,6 +1120,13 @@ void CodeGen::genHWIntrinsic(GenTreeHWIntrinsic* node)
                 GetEmitter()->emitIns_R_R_R(ins, emitSize, targetReg, op2Reg, op3Reg, opt);
                 break;
             }
+
+            case NI_ArmBase_Arm64_MultiplyLongAdd:
+            case NI_ArmBase_Arm64_MultiplyLongSub:
+                assert(opt == INS_OPTS_NONE);
+                GetEmitter()->emitIns_R_R_R_R(ins, EA_8BYTE, targetReg, op1Reg, op2Reg, op3Reg);
```
kunalspathak, line 1127:
```suggestion
                GetEmitter()->emitIns_R_R_R_R(ins, emitSize, targetReg, op1Reg, op2Reg, op3Reg);
```

src/coreclr/jit/lowerarmarch.cpp
```diff
@@ -532,6 +532,16 @@ GenTree* Lowering::LowerBinaryArithmetic(GenTreeOp* binOp)
                 return next;
             }
         }
+
+        if (binOp->OperIs(GT_SUB))
+        {
+            // Attempt to optimise for smsubl.
```
kunalspathak, line 538:
```suggestion
            // Attempt to optimize for umsubl/smsubl.
```

src/coreclr/jit/lowerarmarch.cpp
```diff
@@ -2701,6 +2711,155 @@ void Lowering::TryLowerCselToCinc(GenTreeOp* select, GenTree* cond)
         }
     }
 }
+
+//----------------------------------------------------------------------------------------------
+// TryLowerAddSubToCombinedMulOp: Attempt to convert ADD and SUB nodes to a combined multiply
+// and add/sub operation, extending. Conversion can only happen if the operands to the
```
kunalspathak, line 2717:
```suggestion
// and add/sub operation. Conversion can only happen if the operands to the
```

src/coreclr/jit/lowerarmarch.cpp
```diff
@@ -2701,6 +2711,155 @@ void Lowering::TryLowerCselToCinc(GenTreeOp* select, GenTree* cond)
...
+        addVal = op1;
+    }
+    else
+    {
+        // Exit if neither operation are GT_MUL_LONG.
+        return nullptr;
+    }
+
+    // Additional value must be of long size.
+    // The compiler will check the relevant signage for us, so we don't need to worry about that.
```
kunalspathak, line 2776: Not sure if I understood this comment. Can you please elaborate?

kunalspathak (changes requested):
Need to resolve merge conflict and also fix the test failure:

```
Assert failure(PID 275 [0x00000113], Thread: 275 [0x0113]): Assertion failed 'compiler->compIsaSupportedDebugOnly(HWIntrinsicInfo::lookupIsa(intrin.id))' in 'System.Collections.Generic.Dictionary`2[System.__Canon,System.__Canon]:TryInsert(System.__Canon,System.__Canon,ubyte):ubyte:this' during 'Generate code' (IL size 668; hash 0x26ab038f; FullOpts)

    File: /__w/1/s/src/coreclr/jit/hwintrinsiccodegenarm64.cpp Line: 210
    Image: /root/helix/work/correlation/corerun
````

File: src/coreclr/jit/lowerarmarch.cpp [hunk 7]
```diff
@@ -2701,6 +2711,155 @@ void Lowering::TryLowerCselToCinc(GenTreeOp* select, GenTree* cond)
...
+
+    // Mul values must both be integers.
+    if (!genActualTypeIsInt(mul->gtOp1) || !genActualTypeIsInt(mul->gtOp2))
+        return nullptr;
+
+    // Create the new node and replace the original.
+    {
+        NamedIntrinsic intrinsicId =
+            op->OperIs(GT_ADD) ? NI_ArmBase_Arm64_MultiplyLongAdd : NI_ArmBase_Arm64_MultiplyLongSub;
+        GenTreeHWIntrinsic* outOp = comp->gtNewScalarHWIntrinsicNode(mul->IsUnsigned() ? TYP_ULONG : TYP_LONG,
```
SingleAccretion, line 2788: > mul->IsUnsigned() ? TYP_ULONG : TYP_LONG

It is illegal to type nodes as `TYP_ULONG`.

SingleAccretion: > add an assert to make it self-explanatory?

Definitely - there is a long-term desire to implement an "IR verifier" component (we already have a lot of the pieces, but they're a bit ad-hoc) that would check for things like that. It just hasn't been done yet.

src/coreclr/jit/lowerarmarch.cpp
c272: Honestly, I'm also not sure what I was trying to get at here. It was probably to denote that we didn't need to check for `TYP_ULONG`, but I have removed the comment line.

src/coreclr/jit/lowerarmarch.cpp
c272: Thank you, wasn't aware of this. I've fixed it as `TYP_LONG` and used the intrinsic's base type to determine unsigned/signed.

tannergooding (approved):
This looks correct to me. It should get sign off from someone from @dotnet/jit-contrib before being merged

kunalspathak (approved):
LGTM

PR #90847, opened 2023-08-19
Fix a deadlock in NonGC + Profiler API

EgorBo:
Fixes https://github.com/dotnet/runtime/issues/90830

Quick explanation how's the dead-lock happening:
**Thread1:**
Someone (typically, JIT) tries to allocate an object on NonGC heap. `FrozenObjectHeapManager` (FOHM) acquires its lock and calls GC's API `RegisterNewSegment`. That API internally **can** hit a case when a GC is happening so it has to wait for GC to complete.

**Thread2 (GC's):**
GC is executing a callback (e.g. `GarbageCollectionFinished` or `*Started`) and Profiler uses that callback to enumerate objects on NonGC heap via `ICorProfilerInfo14::GetNonGCHeapBounds` thus, it also tries to acquire FOHM's lock (to be able to safely enumerate the objects). Thus, GC's thread (Thread2) is wating for FOHM's lock to release (it's taken by Thread1) while Thread1 is waiting for GC to finish.

The fix is https://github.com/dotnet/runtime/issues/90830#issuecomment-1684841908

PR #91895, opened 2023-09-11
#ifdef out code that requires FEATURE_EVENT_TRACE

yowl:
This PR follows #90847 and removes some more code when FEATURE_EVENT_TRACE is not defined.  Surfaced as part of https://github.com/dotnet/runtimelab/pull/2396

Not sure about the `gcrhenv.cpp` change.  Maybe I should be removing all the `DiagGC*` methods and `#ifdef`ing their usages ?

cc @EgorBo 

Thanks

### DIFF
--- a/src/coreclr/gc/gc.cpp
+++ b/src/coreclr/gc/gc.cpp
@@ -10056,7 +10056,9 @@ void gc_heap::update_ro_segment (heap_segment* seg, uint8_t* allocated, uint8_t*
 {
     enter_spin_lock (&gc_heap::gc_lock);
 
+#ifdef FEATURE_EVENT_TRACE
     assert (use_frozen_segments_p);
+#endif // FEATURE_EVENT_TRACE
     assert (heap_segment_read_only_p (seg));
     assert (allocated <= committed);
     assert (committed <= heap_segment_reserved (seg));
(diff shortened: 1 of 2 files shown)


EgorBo:
LGTM, but leaving for @dotnet/gc to approve/merge

yowl:
> We can also just remove the assert - it doesn't add a lot of value

I have no preference here.

yowl:
Sorry I closed this by mistake.  

> for the changed related to the `FEATURE_EVENT_TRACE ifdef` checks in gc.cpp, why are only some removed but not all?

I've gone ahead and just removed the assert so `gc.cpp` only has that as a change.  @Maoni0 is there anything else I can answer or change here?  Thanks.

EgorBo (approved):
LGTM, but leaving for @dotnet/gc to review/merge

mangod9:
@cshung, can you please review and merge?

TheSpydog:
Is this fix going to be merged into a .NET 8.x release? For our game console ports of NativeAOT, we do not enable FEATURE_EVENT_TRACE, so we’re running into compile errors when upgrading our runtime support to .NET 8.

jkotas:
We will take it for the consideration. We do not typically backport fixes for custom configuration like this one, but I think we can make an exception for this one given that it is a regression from .NET 7, it is shortly after release and the fix is very low risk.

Issue #91380, opened 2023-08-31
Config generator overwrites existing properties when provided with empty config section


adamsitnik:
In a project that I am working on, we have hierarchical configuration. It means that the parent and child config sections can provide same properties.

Following repro shows the issue:

```cs
using Microsoft.Extensions.Configuration;
using Microsoft.Extensions.Hosting;

namespace ConfigBindRepro
{
    internal class Program
    {
        static void Main()
        {
            HostApplicationBuilder builder = Host.CreateEmptyApplicationBuilder(settings: null);
            builder.Configuration.AddInMemoryCollection(new KeyValuePair<string, string?>[]
            {
                 new("A:B:ConnectionString", "localhost"),
            });

            AClass instance = new();
            builder.Configuration.GetSection("A:B").Bind(instance); // sets ConnectionString to "localhost"
            builder.Configuration.GetSection("A").Bind(instance); // source gen: sets ConnectionString to null, reflection: does not change the value

            Console.WriteLine($"The value is '{instance.ConnectionString}'");
        }
    }

    public class AClass
    {
        public string? ConnectionString { get; set; }
    }
}
```

When using reflection based binder:

```log
The value is 'localhost'
```

When using source gen:

```log
The value is ''
```

cc @layomia @eerhardt

layomia:
Need to resolve alongside https://github.com/dotnet/runtime/issues/86333; same underlying logic.

Issue #86333, opened 2023-05-16
[Binder gen] To align with reflection impl, update property binding logic that handles missing config values


layomia:
Background: https://github.com/dotnet/runtime/pull/86285#discussion_r1195339024. This is for consistency with the reflection binder.

layomia:
Per feedback from @tarekgh all binding differences from the reflection implementation need to be addressed before we ship; this includes differences in exception handling. Marking as blocking release.

Divergences are reflected in `ConditionalFact` applications in the source gen tests.

ericstj:
Triage: users may notice this difference.  We have had multiple reports of this when we regressed in the runtime implementation.

layomia:
**Scenarios**

- [Resolve this diff](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.cs#L1715-L1721)
- https://github.com/dotnet/runtime/issues/91380 (@adamsitnik)

ericstj:
Fixed in RC2 with  #92167

PR #91899, opened 2023-09-11
Make src gen for property setters consistent with reflection

steveharter:
Address https://github.com/dotnet/runtime/issues/86333 (RC2 candidate as well)
- Call the setter for value type properties (with the default value for the property type) if the value is not in the config, unless it is during a Bind (not "Get"). This requires a adding a `bool` arg to the `BindCore()` method.
- Always check for a `null` value or missing section for properties and if null\missing (this was not done in all cases previously) then don't call the setter with the `null`. This addresses https://github.com/dotnet/runtime/issues/91380.

Note that this PR matches known semantics regarding the above with the goal of maintaining compat with reflection; this is the case even for non-intuitive cases including:
- Only value type properties get their setter called; not reference types or `Nullable<T>`.
- Also, properties on objects that are nested objects or those added to a collection are also not defaulted through the setter (only properties on root objects are).

We could create an issue for these non-intuitive cases, but would need to consider the benefit:risk including breaking changes.

### DIFF
--- a/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
+++ b/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
@@ -258,7 +258,7 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
                 }
                 else
                 {
-                    EmitBindCoreImplForObject((ObjectSpec)effectiveType, ValueDefaulting.None);
+                    EmitBindCoreImplForObject((ObjectSpec)effectiveType);
                 }
 
                 EmitEndBlock();
@@ -384,8 +384,7 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
                         parsedMemberAssignmentLhsExpr,
                         sectionPathExpr: GetSectionPathFromConfigurationExpression(configKeyName),
                         canSet: true,
-                        firstTimeInitialization: true,
-                        ValueDefaulting.None);
+                        firstTimeInitialization: true);
 
                     if (canBindToMember)
                     {
@@ -767,7 +766,7 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
                 EmitEndBlock();
             }
 
-            private void EmitBindCoreImplForObject(ObjectSpec type, ValueDefaulting valueDefaulting)
+            private void EmitBindCoreImplForObject(ObjectSpec type)
             {
                 Debug.Assert(type.HasBindableMembers);
 
@@ -786,8 +785,7 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
                             memberAccessExpr: $"{containingTypeRef}.{property.Name}",
                             GetSectionPathFromConfigurationExpression(property.ConfigurationKeyName),
                             canSet: property.CanSet,
-                            firstTimeInitialization: false,
-                            valueDefaulting);
+                            firstTimeInitialization: false);
                     }
                 }
             }
@@ -797,8 +795,7 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
                 string memberAccessExpr,
                 string sectionPathExpr,
                 bool canSet,
-                bool firstTimeInitialization,
-                ValueDefaulting valueDefaulting)
+                bool firstTimeInitialization)
             {
                 TypeSpec effectiveMemberType = member.Type.EffectiveType;
 
@@ -845,7 +842,7 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
 
                             EmitBlankLineIfRequired();
                             EmitStartBlock($"if ({sectionValidationCall} is {Identifier.IConfigurationSection} {sectionIdentifier})");
-                            EmitBindingLogicForComplexMember(member, memberAccessExpr, sectionIdentifier, canSet, valueDefaulting);
+                            EmitBindingLogicForComplexMember(member, memberAccessExpr, sectionIdentifier, canSet);
                             EmitEndBlock();
 
                             return complexType.CanInstantiate;
@@ -859,8 +856,7 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
                 MemberSpec member,
                 string memberAccessExpr,
                 string configArgExpr,
-                bool canSet,
-                ValueDefaulting valueDefaulting)
+                bool canSet)
             {
 
                 TypeSpec memberType = member.Type;
@@ -923,7 +919,7 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
                     targetObjAccessExpr,
                     configArgExpr,
                     initKind,
-                    valueDefaulting,
+                    ValueDefaulting.None,
                     writeOnSuccess
                     );
             }
--- a/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
+++ b/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
@@ -807,7 +807,6 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
                         {
                             if (canSet)
                             {
-                                string nullBangExpr = member.Type.IsValueType ? string.Empty : "!";
                                 bool useDefaultValueIfSectionValueIsNull = !firstTimeInitialization &&
                                     member is PropertySpec &&
                                     member.Type.IsValueType &&
@@ -818,7 +817,7 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
                                     stringParsableType,
                                     $@"{Identifier.configuration}[""{member.ConfigurationKeyName}""]",
                                     sectionPathExpr,
-                                    writeOnSuccess: parsedValueExpr => _writer.WriteLine($"{memberAccessExpr} = {parsedValueExpr}{nullBangExpr};"),
+                                    writeOnSuccess: parsedValueExpr => _writer.WriteLine($"{memberAccessExpr} = {parsedValueExpr};"),
                                     checkForNullSectionValue: true,
                                     useDefaultValueIfSectionValueIsNull,
                                     useIncrementalStringValueIdentifier: true);
--- a/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
+++ b/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
@@ -237,7 +237,7 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
             private void EmitBindCoreMethod(ComplexTypeSpec type)
             {
                 string objParameterExpression = $"ref {type.DisplayString} {Identifier.instance}";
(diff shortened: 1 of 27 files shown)


tarekgh:
@steveharter there are some failures that seem related to the change. Could you please have a look?

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/Baselines/OptionsBuilder/Bind_T_BinderOptions.generated.txt
```diff
@@ -108,12 +108,19 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
         {
             ValidateConfigurationKeys(typeof(Program.MyClass), s_configKeys_ProgramMyClass, configuration, binderOptions);
 
-            instance.MyString = configuration["MyString"]!;
+            if (configuration["MyString"] is string value1)
+            {
+                instance.MyString = value1!;
```
layomia, line 113:
```suggestion
                instance.MyString = value1;
```

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/Baselines/OptionsBuilder/Bind_T_BinderOptions.generated.txt
```diff
@@ -108,12 +108,19 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
...
+            }
 
             if (configuration["MyInt"] is string value2)
             {
                 instance.MyInt = ParseInt(value2, () => configuration.GetSection("MyInt").Path);
             }
+            else
+            {
+                instance.MyInt = default;
```
layomia, line 122: Perhaps make this a conditional assignment to reduce LOC?

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/Baselines/OptionsBuilder/Bind_T.generated.txt
```diff
@@ -114,12 +114,19 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
         {
             ValidateConfigurationKeys(typeof(Program.MyClass), s_configKeys_ProgramMyClass, configuration, binderOptions);
 
-            instance.MyString = configuration["MyString"]!;
+            if (configuration["MyString"] is string value1)
+            {
+                instance.MyString = value1!;
+            }
```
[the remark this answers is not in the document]
steveharter: The reflection appears to only call the setter for value types and not `Nullable<T>` -- basically something that can never be `null`.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/Baselines/OptionsBuilder/Bind_T_BinderOptions.generated.txt
steveharter: Thanks

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/Baselines/OptionsBuilder/Bind_T_BinderOptions.generated.txt
steveharter: A newer commit changed the `else` to `else if` so I don't think a conditional assignment works well anymore.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/Baselines/Collections.generated.txt
```diff
@@ -161,39 +161,39 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
...
+        public static void BindCore(IConfiguration configuration, ref Program.MyClassWithCustomCollections instance, bool defaultValueIfNotFound, BinderOptions? binderOptions)
         {
             ValidateConfigurationKeys(typeof(Program.MyClassWithCustomCollections), s_configKeys_ProgramMyClassWithCustomCollections, configuration, binderOptions);
 
             if (AsConfigWithChildren(configuration.GetSection("CustomDictionary")) is IConfigurationSection section1)
             {
                 Program.CustomDictionary<string, int>? temp3 = instance.CustomDictionary;
                 temp3 ??= new Program.CustomDictionary<string, int>();
-                BindCore(section1, ref temp3, binderOptions);
+                BindCore(section1, ref temp3, true, binderOptions);
```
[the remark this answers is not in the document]
steveharter: A later commit addresses this; it should be `false` here to match reflection semantics.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/Baselines/ConfigurationBinder/Bind.generated.txt
```diff
@@ -45,7 +45,7 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
             }
 
             var typedObj = (Program.MyClass)instance;
-            BindCore(configuration, ref typedObj, binderOptions: null);
+            BindCore(configuration, ref typedObj, false, binderOptions: null);
```
layomia, line 48:
```suggestion
            BindCore(configuration, ref typedObj, defaultValueIfNotFound: false, binderOptions: null);
```

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
```diff
@@ -307,6 +309,10 @@ private void EmitInitializeMethod(ObjectSpec type)
                 {
                     if (property.ShouldBindTo && property.MatchingCtorParam is null)
                     {
+                        // Currently properties don't have an 'else' failure case. Doing that may collide with the
+                        // feature to set properties to their default values if they don't exist in the config section.
+                        Debug.Assert(property.ErrorOnFailedBinding == false);
```
layomia, line 314: > // Currently properties don't have an 'else' failure case.

Seems like we should remove the abstraction to avoid these two features from clashing. Not sure that the property binding expectation here will change.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
```diff
@@ -633,7 +640,7 @@ private void EmitPopulationImplForArray(EnumerableSpec type)
 
                 // Create list and bind elements.
                 string tempIdentifier = GetIncrementalIdentifier(Identifier.temp);
```
layomia, line 642: Seems to me that `InitializationKind.Declaration` and `ValueDefaulting.None` indicate the same scenario, when a new object is being declared. Maybe we could remove `ValueDefaulting`?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
```diff
@@ -377,7 +383,8 @@ void EmitBindImplForMember(MemberSpec member)
                         member,
                         parsedMemberAssignmentLhsExpr,
                         sectionPathExpr: GetSectionPathFromConfigurationExpression(configKeyName),
-                        canSet: true);
+                        canSet: true,
+                        firstTimeInitialization: true);
```
layomia, line 387: Would `InitializationKind.Declaration` express the same intent?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
steveharter: The only place with `ValueDefaulting.CallSetter` is for complex types (objects) which also has  `InitializationKind.Declaration`. However, arrays and enumerables also have  `InitializationKind.Declaration` but should not have `ValueDefaulting.CallSetter`.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
steveharter: Sure.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
layomia: Seems like something we can simplify by adding intermediary logic/methods before we get to the core binding methods. I wouldn't block on this though.

Generally I'm trying to tease out the decisions made in the core binding methods so that it's to understand (https://github.com/dotnet/runtime/pull/91717#discussion_r1320459650).

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/Baselines/ConfigurationBinder/Bind.generated.txt
steveharter: OK; originally I was assuming we wanted terse generated code, but I see that we use `binderOptions: null`

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
steveharter: The reason for the difference is because of the non-intuitive reflection semantics around not calling setter for objects added to arrays.

I could look at adding support for collections\enumerable; it would be a small breaking change for reflection, but that would mean modifying the reflection code.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
steveharter: I moved the ErrorOnFailedBindingProperty from the base class to the parameter class, and removed this assert.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/Baselines/ConfigurationBinder/Bind.generated.txt
layomia: We do want terse code, but generally prioritize readability.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
layomia: > I could look at adding support for collections\enumerable; it would be a small breaking change for reflection, but that would mean modifying the reflection code.

We should definitely avoid this, especially not for 8. The binder / user expectations are very sensitive to any breaking changes, as @tarekgh would probably echo.

layomia: The approach you have now looks good to me.

tarekgh: Yes, please try avoiding introducing any breaking change in for the reflection. Any slight change gets noticed and causes us servicing it. At least try to avoid it in the current stage of the project.

layomia (approved):
Thanks!

steveharter:
Need to get in at least this before backport merge can be successful: https://github.com/dotnet/runtime/pull/91717

PR #91901, opened 2023-09-11
Merge CoreMangLib, Exceptions and Interop folders

trylek:
This change converts tests in the CoreMangLib, Exceptions and Interop to the merged model. As next step I plan to work on review feedback to my previous PR merging baseservices and on merging the remaining tests.

Thanks

Tomas

### DIFF
--- a/src/tests/CoreMangLib/system/delegate/VSD/OpenDelegate.csproj
+++ b/src/tests/CoreMangLib/system/delegate/VSD/OpenDelegate.csproj
@@ -1,5 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
+    <!-- Needed for mechanical merging of all remaining tests, this particular project may not actually need process isolation -->
+    <RequiresProcessIsolation>true</RequiresProcessIsolation>
     <OutputType>exe</OutputType>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <CLRTestPriority>1</CLRTestPriority>
--- a/src/tests/CoreMangLib/system/delegate/VSD/OpenDelegate.csproj
+++ b/src/tests/CoreMangLib/system/delegate/VSD/OpenDelegate.csproj
@@ -2,7 +2,6 @@
   <PropertyGroup>
     <!-- Needed for mechanical merging of all remaining tests, this particular project may not actually need process isolation -->
     <RequiresProcessIsolation>true</RequiresProcessIsolation>
-    <OutputType>exe</OutputType>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <CLRTestPriority>1</CLRTestPriority>
   </PropertyGroup>
--- a/src/tests/Interop/MonoAPI/MonoMono/InstallEHCallback.cs
+++ b/src/tests/Interop/MonoAPI/MonoMono/InstallEHCallback.cs
@@ -5,183 +5,183 @@ using System;
 using System.Runtime.InteropServices;
 using System.Runtime.CompilerServices;
 
+using Xunit;
+
 namespace MonoAPI.Tests.MonoMono.InstallEHCallback;
 
 public class MonoPInvokeCallbackAttribute : Attribute {
-	public MonoPInvokeCallbackAttribute (Type delegateType) { }
+    public MonoPInvokeCallbackAttribute (Type delegateType) { }
 }
 
 public class InstallEHCallback {
 
-	[DllImport (MonoAPISupport.TestLibName)]
-	public static extern void mono_test_setjmp_and_call (VoidVoidDelegate del, out IntPtr handle);
+    [DllImport (MonoAPISupport.TestLibName)]
+    public static extern void mono_test_setjmp_and_call (VoidVoidDelegate del, out IntPtr handle);
 
-	[DllImport (MonoAPISupport.TestLibName)]
-	public static extern void mono_test_setup_ftnptr_eh_callback (VoidVoidDelegate del, VoidHandleHandleOutDelegate inside_eh_callback);
+    [DllImport (MonoAPISupport.TestLibName)]
+    public static extern void mono_test_setup_ftnptr_eh_callback (VoidVoidDelegate del, VoidHandleHandleOutDelegate inside_eh_callback);
 
-	[DllImport (MonoAPISupport.TestLibName)]
-	public static extern void mono_test_cleanup_ftptr_eh_callback ();
-	
-	public delegate void VoidVoidDelegate ();
-	public delegate void VoidHandleHandleOutDelegate (uint handle, out int exception_handle);
+    [DllImport (MonoAPISupport.TestLibName)]
+    public static extern void mono_test_cleanup_ftptr_eh_callback ();
+    
+    public delegate void VoidVoidDelegate ();
+    public delegate void VoidHandleHandleOutDelegate (uint handle, out int exception_handle);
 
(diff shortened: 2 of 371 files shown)


src/tests/CoreMangLib/system/delegate/VSD/OpenDelegate.csproj
```diff
@@ -1,5 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
+    <!-- Needed for mechanical merging of all remaining tests, this particular project may not actually need process isolation -->
+    <RequiresProcessIsolation>true</RequiresProcessIsolation>
     <OutputType>exe</OutputType>
```
markples, line 5: It looks like the lowercase 'exe' slipped through the scan.

markples: (though, now that I think about it, I guess that's where ReqProcIso ends up anyway, though I think we automatically set it (kind of hacky in Dir.B.targets since it is after the SDK has had a look at it))

src/tests/Interop/MonoAPI/MonoMono/InstallEHCallback.cs
```diff
@@ -5,183 +5,183 @@
...
+        return 0;
+    }
+
+    [Fact]
+    public static void TestEntryPoint ()
+    {
+        MonoAPI.Tests.MonoAPISupport.Setup();
+        int result;
+        result = test_0_setjmp_exn_handler ();
+        Assert.Equal(0, result);
```
markples, line 182: I suspect that this (and the others) end up being fine (mono test, int equality), but I don't know for sure that these don't end up being run in some configuration that breaks.  But hopefully outerloop (and extra platforms?) is good enough to check that?

src/tests/CoreMangLib/system/delegate/VSD/OpenDelegate.csproj
trylek: I think you're right that we shouldn't keep obsolete ```OutputType=exe``` properties on most tests except the tiny subset opting out of the merged wrappers for special reasons like the half a dozen tests under baseservices that do weird things. I have experimentally removed all these annotations from the interop tests and I'm retesting the change locally, I'll be happy to update the PR by removing those if it turns out to be working.

src/tests/Interop/PInvoke/Array/MarshalArrayAsParam/AsDefault/AsDefaultTest.cs
```diff
@@ -395,7 +395,7 @@ private static void TestMarshalInOut_ByVal()
         Console.WriteLine("CStyle_Array_Int_InOut_ZeroLength");
         int[] iArrLength0 = InitArray<int>(0);
         Assert.True(CStyle_Array_Int_InOut_ZeroLength(iArrLength0));
-        Assert.Equal(0, iArrLength0.Length);
+        Assert.Empty(iArrLength0);
```
markples, line 398: similarly, I have no idea whether int equality or `Empty` is more dangerous in the various configurations

src/tests/Interop/PInvoke/BestFitMapping/Program.cs
```diff
@@ -6,37 +6,33 @@
...
-            LPStr.PInvoke_Default.RunTest();
-            LPStr.PInvoke_False_False.RunTest();
-            LPStr.PInvoke_False_True.RunTest();
-            LPStr.PInvoke_True_False.RunTest();
-            LPStr.PInvoke_True_True.RunTest();
-            return 100;
-        } catch (Exception e){
-            Console.WriteLine($"Test Failure: {e}");
-            return 101;
-        }
```
markples, line 40: I like this cleanup that we get!

src/tests/Interop/MonoAPI/MonoMono/InstallEHCallback.cs
trylek: All I can say to this now is that in general Mono tests are supported by the merged test infrastructure, @jkoritzinsky went to great lengths to make that fully functional and IIRC Mono people were happy about his implementation in terms of efficiency and cleanness.

src/tests/Interop/PInvoke/Array/MarshalArrayAsParam/AsDefault/AsDefaultTest.cs
trylek: I just fixed those as the analyzer slapped me because of them and it seemed to me that somehow exempting them from the scanning would be more complicated than just going with the flow.

src/tests/Interop/UnmanagedCallersOnly/UnmanagedCallersOnlyTest.cs
```diff
@@ -45,11 +46,6 @@ public static int Main(string[] args)
                 TestUnmanagedCallersOnlyValid_ThrowException();
                 TestUnmanagedCallersOnlyViaUnmanagedCalli_ThrowException();
             }
-
-            if (args.Length != 0 && args[0].Equals("calli"))
```
markples, line 49: Was this being tested before?  (if not, you'll get a +1 in your counts)

jkoritzinsky: No, this was only for manual testing. In the interop space, we do very little automated testing of "kills the process" behaviors around invalid code patterns.

trylek: The test count is a complete mess, under the merged scheme the test count is more than twice the original number. I think this is because multiple tests under this subtree used the ```[Fact]``` annotations even before test merging and in the legacy system those reported as a single test within a particular project whereas today they smoothly blend with the overall infra and report as multiple tests. There were two tests exercising CoreCLR runtime crashes under special circumstances, the bit you have pointed out and a similar bit in the SuppressGCTransition test. Apparently they haven't been running recently - we may have broken them at some point in the past by removing the RunOnly projects (tests running other tests with specific sets of command-line parameters) or perhaps they were always broken. So now they are broken no more and I expect follow-up work once they start throwing different errors on other platforms.

trylek: OK, thanks @jkoritzinsky, so please let me know whether I should keep these two as negative tests or throw them away.

jkoritzinsky: I think it's worth keeping them as Pri1 tests, at least for now.

src/tests/Interop/MonoAPI/MonoMono/InstallEHCallback.cs
markples: Sorry, I wasn't clear.  I just don't know if this test will end up running in some other configuration (from a quick look, I don't see it being excluded from other runtimes).  However, I meant it more as a "heads up" in case something breaks, not an objection to the PR.

markples (approved):

src/tests/Interop/MonoAPI/MonoMono/InstallEHCallback.cs
jkoritzinsky: If we have concerns, we can use the SkipOnCoreClrAttribute on the test and it will automatically be skipped on non-Mono platforms. Tbh I think that's a better solution than suppressing the test in issues.targets.

PR #91904, opened 2023-09-11
[RISC-V] regArg dependcies unrolling in genFnPrologCalleeRegArgs()

t-mustafin:
Fixes `NYI_RISCV64("-----------CodeGen::genFnPrologCalleeRegArgs() error!--");` on register dependencies with length > 2 and potentially contain a loop.

cc @alpencolt @gbalykov @ashaurtaev @clamp03 @tomeksowi

### DIFF
--- a/src/coreclr/jit/codegenriscv64.cpp
+++ b/src/coreclr/jit/codegenriscv64.cpp
@@ -7578,11 +7578,22 @@ void CodeGen::genFnPrologCalleeRegArgs()
     noway_assert(regArgMaskLive != 0);
 
     unsigned varNum;
-    unsigned regArgMaskIsInt = 0;
-    unsigned regArgNum       = 0;
-    // Process any circular dependencies
-    unsigned regArg[MAX_REG_ARG * 2]     = {0};
-    unsigned regArgInit[MAX_REG_ARG * 2] = {0};
+    unsigned regArgNum = 0;
+    // Process any rearrangements including circular dependencies.
+    regNumber regArg[MAX_REG_ARG + MAX_FLOAT_REG_ARG];
+    regNumber regArgInit[MAX_REG_ARG + MAX_FLOAT_REG_ARG];
+    emitAttr  regArgAttr[MAX_REG_ARG + MAX_FLOAT_REG_ARG];
+
+    for (int i = 0; i < MAX_REG_ARG + MAX_FLOAT_REG_ARG; i++)
+    {
+        regArg[i] = REG_NA;
+
+#ifdef DEBUG
+        regArgInit[i] = REG_NA;
+        regArgAttr[i] = EA_UNKNOWN;
+#endif
+    }
+
     for (varNum = 0; varNum < compiler->lvaCount; ++varNum)
     {
         LclVarDsc* varDsc = compiler->lvaTable + varNum;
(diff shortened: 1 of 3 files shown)


tomeksowi:
Which test does this PR fix?

t-mustafin:
> Which test does this PR fix?

System.Text.RegularExpressions.Tests catches NYI assert in Debug mode.

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -7578,11 +7578,12 @@ void CodeGen::genFnPrologCalleeRegArgs()
     noway_assert(regArgMaskLive != 0);
 
     unsigned varNum;
-    unsigned regArgMaskIsInt = 0;
-    unsigned regArgNum       = 0;
+    unsigned regArgNum = 0;
     // Process any circular dependencies
-    unsigned regArg[MAX_REG_ARG * 2]     = {0};
-    unsigned regArgInit[MAX_REG_ARG * 2] = {0};
+    regNumber regArg[MAX_REG_ARG + MAX_FLOAT_REG_ARG]     = {REG_R0};
```
clamp03, line 7583: How about using `REG_NA` instead of `REG_R0` for initial value?

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -7638,20 +7639,19 @@ void CodeGen::genFnPrologCalleeRegArgs()
...
-                            assert(varDsc->GetArgReg() >= REG_ARG_FIRST && varDsc->GetArgReg() <= REG_ARG_LAST);
+                            assert(isValidIntArgReg(varDsc->GetArgReg()));
                             regArg[varDsc->GetArgReg() - REG_ARG_FIRST]     = varDsc->GetArgReg();
                             regArgInit[varDsc->GetArgReg() - REG_ARG_FIRST] = varDsc->GetArgInitReg();
-                            if (varDsc->TypeGet() == TYP_INT)
-                            {
-                                regArgMaskIsInt = 1 << (unsigned)varDsc->GetArgReg();
-                            }
+                            regArgAttr[varDsc->GetArgReg() - REG_ARG_FIRST] =
+                                varDsc->TypeGet() == TYP_INT ? EA_4BYTE : EA_PTRSIZE;
```
clamp03, line 7646: How about  changing like `EA_SIZE(emitActualTypeSize(varDsc->TypeGet()))`?

gbalykov:
cc @jakobbotsch

jakobbotsch (approved):
Looks fine to me. But I wonder why RISCV and LA64 do not use the same `genFnPrologCalleeRegArgs` as all other architectures? What special treatment do they need?

t-mustafin:
> Looks fine to me. But I wonder why RISCV and LA64 do not use the same `genFnPrologCalleeRegArgs` as all other architectures? What special treatment do they need?

I don't have answer for this question yet. @clamp03 @shushanhf do you know the answer?

jakobbotsch:
I will merge this, but you may consider seeing whether RISC-V can just use the same function as the officially supported targets.

shushanhf:
Hi, @jakobbotsch 
Sorry for late reponse.

(1) The LA64 and RISCV64 can use the same `genFnPrologCalleeRegArgs`.

> Looks fine to me. But I wonder why RISCV and LA64 do not use the same `genFnPrologCalleeRegArgs` as all other architectures? What special treatment do they need?
(2) The ABI of the RISCV64 and LA64 is quite different with the other architectures. liking the float-arg might be passed by integer register.  including the field of struct passed by registers.

PR #89809, opened 2023-08-01
Adjust System.Net.Http metrics

antonfirsov:
Adjust System.Net.Http metrics naming and semantics according to the outcome of the discussion in https://github.com/lmolkova/semantic-conventions/pull/1:

https://github.com/lmolkova/semantic-conventions/blob/dotnet8-metrics/docs/dotnet/dotnet-http-metrics.md

Contributes to #89093.
Fixes #89451.

Issue #90639, opened 2023-08-15
French abbreviated month names not showing up as expected


carlossanlop:
### Reproduction Steps

Hitting an 8.0 RC1 PR.
- Queue: `browser-wasm linux Release LibraryTests`
- Build: https://dev.azure.com/dnceng/internal/_build/results?buildId=2244190&view=logs&j=d4e38924-13a0-58bd-9074-6a4810543e7c&t=7a581d54-fbdb-5b9d-4377-cc346f489112
- Log file: https://helixri107v0xd1eu3ibi6ka.blob.core.windows.net/dotnet-runtime-refs-pull-32920-merge-ea7d46fe093c40c687/WasmTestOnBrowser-System.Globalization.Hybrid.WASM.Tests/1/console.7f0e9ff7.log?sv=2021-08-06&se=2023-11-13T20%3A20%3A15Z&sr=c&sp=rl&sig=W%2F2gF08hpz1loT%2FCaXrJ1MIlITlV%2BvXkL7uTMw67Pzw%3D
- Output:
```
info: [2023-08-15T20:36:42.955Z] [FAIL] System.Globalization.Tests.DateTimeFormatInfoAbbreviatedMonthGenitiveNames.AbbreviatedMonthGenitiveNames_Get_ReturnsExpected_HybridGlobalization(format: DateTimeFormatInfo { AbbreviatedDayNames = ["søn.", "man.", "tir.", "ons.", "tor.", ...], AbbreviatedMonthGenitiveNames = ["jan.", "feb.", "mars", "apr.", "mai", ...], AbbreviatedMonthNames = ["jan", "feb", "mar", "apr", "mai", ...], AMDesignator = "a.m.", Calendar = GregorianCalendar { AlgorithmType = SolarCalendar, CalendarType = Localized, Eras = [...], IsReadOnly = False, MaxSupportedDateTime = 9999-12-31T23:59:59.9999999, ... }, ... }, expected: ["jan.", "feb.", "mar.", "apr.", "mai", ...])
info: Assert.Equal() Failure
info:                            ↓ (pos 2)
info: Expected: ["jan.", "feb.", "mar.", "apr.", "mai", ...]
info: Actual:   ["jan.", "feb.", "mars", "apr.", "mai", ...]
info:                            ↑ (pos 2)
[...]

PR #91910, opened 2023-09-12
Replace http.error.reason with OTel standard error.type

### LINKED ISSUE #91909: [Http Metrics] Replace http.error.reason with OTel standard error.type
Align attribute name and semantics with the OpenTelemetry spec (which was finalized yesterday 2023/9/11 - see https://github.com/open-telemetry/semantic-conventions/pull/205) -- `http.error.reason` attribute has been standardized under name `error.type`. We need to:
- Rename `http.error.reason` to `error.type`.
- Report `error.type` also for valid responses where the HTTP status code indicates an error (4xx and 5xx) -- in such case, the value should be the string representation of the HTTP status code.
    - Note: Currently we report `http.error.reason` only when the underlying handler fails to fetch a response.

See `error.type` in [the spec](https://github.com/open-telemetry/semantic-conventions/blob/2bad9afad58fbd6b33cc683d1ad1f006e35e4a5d/docs/http/http-spans.md#common-attributes) for more details.

For context: We introduced the attribute `http.error.reason` in PR #89809 (on 2023/8/3) based on the in-progress draft of the OTel spec (https://github.com/open-telemetry/semantic-conventions/pull/205).

Given that this is new 8.0 feature, we should adapt to the spec in 8.0 to avoid breaking changes with 9.0.

antonfirsov:
Fixes #91909

Align attribute name and semantics with the OpenTelemetry spec (which was finalized yesterday 2023/9/11 - see https://github.com/open-telemetry/semantic-conventions/pull/205) -- `http.error.reason` attribute has been standardized under name `error.type`. We need to:
- Rename `http.error.reason` to `error.type`.
- Report `error.type` also for valid responses where the HTTP status code indicates an error (4xx or 5xx) -- in such case, the value should be the string representation of the HTTP status code.
    - Note: Currently we report `http.error.reason` only when the underlying handler fails to fetch a response.

See `error.type` in [the spec](https://github.com/open-telemetry/semantic-conventions/blob/2bad9afad58fbd6b33cc683d1ad1f006e35e4a5d/docs/http/http-spans.md#common-attributes) for more details.

For context: We introduced the attribute `http.error.reason` in PR #89809 (on 2023/8/3) based on the in-progress draft of the OTel spec (https://github.com/open-telemetry/semantic-conventions/pull/205).

Given that this is new 8.0 feature, we should adapt to the spec in 8.0 to avoid breaking changes with 9.0.

### DIFF
--- a/src/libraries/System.Net.Http/src/System/Net/Http/Metrics/MetricsHandler.cs
+++ b/src/libraries/System.Net.Http/src/System/Net/Http/Metrics/MetricsHandler.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.Metrics;
+using System.Net.Http.HPack;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -112,11 +113,12 @@ namespace System.Net.Http.Metrics
                 tags.Add("http.response.status_code", GetBoxedStatusCode((int)response.StatusCode));
                 tags.Add("network.protocol.version", GetProtocolVersionString(response.Version));
             }
-            else
+
+            if (TryGetErrorType(response, exception, out string? errorType))
             {
-                Debug.Assert(exception is not null);
-                tags.Add("http.error.reason", GetErrorReason(exception));
+                tags.Add("error.type", errorType);
             }
+
             TimeSpan durationTime = Stopwatch.GetElapsedTime(startTimestamp, Stopwatch.GetTimestamp());
 
             HttpMetricsEnrichmentContext? enrichmentContext = HttpMetricsEnrichmentContext.GetEnrichmentContextForRequest(request);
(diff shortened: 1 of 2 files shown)


danmoseley:
Is there any similar change needed in Kestrel @JamesNK

JamesNK:
No aspnetcore changes are required. This attribute is for HTTP clients only.

danmoseley:
Ok. Yes sounds like we should definitely get into 8.0 as suggested

karelz:
Let's get this into 9.0 first, then we can immediately create backport to 8.0. IMO it should not have a problem to get in.

@artl93 heads up. If you have concerns, please let us know ASAP.

src/libraries/System.Net.Http/src/System/Net/Http/Metrics/MetricsHandler.cs
```diff
@@ -209,6 +221,14 @@ private static object GetBoxedStatusCode(int statusCode)
                 : statusCode;
         }
 
+        private static string GetStatusCodeString(int statusCode)
```
MihaZupan, line 224: @stephentoub should we consider bumping `SmallNumberCacheLength` to 512 instead and we can delete this code?

MihaZupan (approved):
Looks good, thank you

src/libraries/System.Net.Http/src/System/Net/Http/Metrics/MetricsHandler.cs
```diff
@@ -209,6 +221,14 @@ private static object GetBoxedStatusCode(int statusCode)
...
+        {
+            string[] strings = LazyInitializer.EnsureInitialized(ref s_statusCodeStrings, static () => new string[512]);
```
karelz, line 226: Would it make sense to save size by offsetting indexes by 400?
Likely not needed, but perhaps in future? (we would have to change the method to `GetStatusCode4xx5xxString`)

karelz (approved):

src/libraries/System.Net.Http/src/System/Net/Http/Metrics/MetricsHandler.cs
antonfirsov: Would we do it in 8.0 as well? If not I would suggest to do that in a separate PR.

stephentoub: Or delete it regardless; the cache size is currently 300, so all success status code are covered.

I'm not against also raising the size, but we do need to be careful with it creeping up too high.

antonfirsov: This method will be never called with success status codes. (I will address https://github.com/dotnet/runtime/pull/91910/files#r1323214809.)

lmolkova (approved):
:shipit:

MihaZupan (approved):

karelz (approved):

noahfalk (approved):
Looks great 👍

antonfirsov:
CI failures are #90639, #91910 + many unrelated failures in `runtime-extra-platforms`.

PR #91912, opened 2023-09-12
[mono] Implement X86Base.Pause in JIT mode as well.

vargaz:


jandupej (approved):
I believe we also have `OP_NOP_RELAXED` that emits just that. LGTM otherwise.

PR #91915, opened 2023-09-12
Make Options source gen support Validation attributes having constructor with params Parameter

tarekgh:
This change is to support the validation attributes having constructor with `params` keyword parameter like [`DeniedValuesAttribute`](https://github.com/dotnet/runtime/blob/4a09c82215399c27f52277a8db7178270410c693/src/libraries/System.ComponentModel.Annotations/src/System/ComponentModel/DataAnnotations/DeniedValuesAttribute.cs#L20) and [`AllowedValuesAttribute`](https://github.com/dotnet/runtime/blob/4a09c82215399c27f52277a8db7178270410c693/src/libraries/System.ComponentModel.Annotations/src/System/ComponentModel/DataAnnotations/AllowedValuesAttribute.cs#L20).

This fix need to be ported to rc2.


tarekgh:
CC @geeknoid @ericstj

src/libraries/Microsoft.Extensions.Options/gen/Parser.cs
```diff
@@ -469,14 +470,36 @@ private List<ValidatedMember> GetMembersToValidate(ITypeSymbol modelType, bool s
...
+                    ImmutableArray<IParameterSymbol> parameters = attribute.AttributeConstructor?.Parameters ?? ImmutableArray<IParameterSymbol>.Empty;
+                    bool lastParameterDeclaredWithParamsKeyword =  parameters.Length > 0 && parameters[parameters.Length - 1].IsParams;
+
+                    ImmutableArray<TypedConstant> arguments = attribute.ConstructorArguments;
+
+                    for (int i = 0; i < arguments.Length; i++)
                     {
-                        validationAttr.ConstructorArguments.Add(GetArgumentExpression(constructorArgument.Type!, constructorArgument.Value));
+                        TypedConstant argument = arguments[i];
+                        if (argument.Kind == TypedConstantKind.Array)
```
eiriktsarpalis, line 481: Probably too much of a niche scenario to consider here, but could this logic be tripped up by something like the following?
```C#
Test(new int[] { 1, 2, 3 });

static void Test(params int[][] values) { }
```

eiriktsarpalis (approved):

geeknoid (approved):

src/libraries/Microsoft.Extensions.Options/gen/Parser.cs
ericstj: That's actually not a valid attribute parameter type.
https://sharplab.io/#v2:D4AQTAjAsAUCDMACciCyBPAagQwDYEsATbAF3wHsA7AQRJICd8AjAVxIFNEAuRWh5tpwDesRGORIMOAsTJU+jVhwAUAB2z1sAWwDOifJRIBtALqnEANzwt2OgJSIhiAL6xXMWEal4ipCjTpFQWVKdgB3fUNzJwgAGkQwePhnOxNYBGQwNHQAFXRVdlgRGHcgA===

> error CS0181: Attribute constructor parameter 'values' has type 'int[][]', which is not a valid attribute parameter type

Attribute's parameters need to be represented in metadata and can only take a very limited set of types.

eiriktsarpalis: Good point. I did recall that it's possible to nest arrays in attribute declarations, but it looks like it's only possible for `object[]` parameters:

https://sharplab.io/#v2:D4AQTAjAsAUCDMACciCyBPAagQwDYEsATbAF3wHsA7AQRJICd8AjAVxIFNEAuRWh5tpwDesRGORIMOAsTJU+jVhwAUAB2z1sAWwDOickwBW7AMYkA2gF1EANzwt2OgJSIhiAL6xPMWOal4iUgpKZUp2AHd9I1MLazcwyINjMytXRAgPDydLWARkMDR0ABV0VXZYERhvIA===

I _think_ that case would work fine with the logic as-is, but maybe a test is in order?

eiriktsarpalis: I see this was brought up here as well: https://github.com/dotnet/runtime/pull/91934#discussion_r1323326105

PR #91919, opened 2023-09-12
Fix InPlaceRuntime layering issues

MichalStrehovsky:
For the hackathon I'm looking at layering in the runtime and I started by getting `InPlaceRuntime=false` building again.

In this build configuration, we use Runtime.Base and System.Private.CoreLib as separate things, with separate System.Object definitions (instead of Runtime.Base being folded into CoreLib).

This is the setup .NET Native used.

Cc @dotnet/ilc-contrib


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/ThunkPool.cs
```diff
@@ -226,11 +226,11 @@ public unsafe void FreeThunk(IntPtr thunkAddress)
 
             IntPtr dataAddress = TryGetThunkDataAddress(thunkAddress);
             if (dataAddress == IntPtr.Zero)
-                EH.FallbackFailFast(RhFailFastReason.InternalError, null);
+                Environment.FailFast(null);
```
jkotas, line 229: I think all these `Environment.FailFasts` can be just debug-only asserts.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/RuntimeExceptionHelpers.cs
```diff
@@ -118,6 +118,16 @@ public class RuntimeExceptionHelpers
             }
         }
 
+        internal enum RhFailFastReason
```
jkotas, line 121: Move this under native\Common to avoid multiple copies?

jkotas (approved):
Thanks

PR #91924, opened 2023-09-12
[mono] Optimize Vector128.Create if there are some zero elements

jandupej:
When using `Vector128<T>.Create(...)` with some arguments set to zero, these do not generate their `ins` as their respective lanes are already correctly initialized.


vargaz (approved):

jandupej:
CI green, one lane was cancelled.

PR #87241, opened 2023-06-08
Import isinst/castclass Nullable as underlying type

huoyaoyuan:
According to ECMA-335 III.4.6:

>  If *typeTok* is a non-nullable value type or a generic parameter type it is interpreted 
as “boxed” *typeTok*. If *typeTok* is a nullable type, `Nullable<T>`, it is interpreted as 
“boxed” `T`. 

In my understanding, it means that `isinst V?` is the same as `isinst V`.

Code gen:
```csharp
        bool TestNullable(object o) => o is int?;

        int TestCoalesce(object o) => (o as int?) ?? 1234;
```

Before:
```asm
; Method CSPlayground.Program:TestNullable(System.Object):bool:this
G_M50267_IG01:  ;; offset=0000H
       sub      rsp, 40
						;; size=4 bbWeight=1 PerfScore 0.25

G_M50267_IG02:  ;; offset=0004H
       mov      rcx, 0x7FF8A0D4A9B8      ; System.Nullable`1[int]
       call     [CORINFO_HELP_ISINSTANCEOFANY]
       test     rax, rax
       setne    al
       movzx    rax, al
						;; size=25 bbWeight=1 PerfScore 4.75

G_M50267_IG03:  ;; offset=001DH
       add      rsp, 40
       ret      
						;; size=5 bbWeight=1 PerfScore 1.25
; Total bytes of code: 34

; Method CSPlayground.Program:TestCoalesce(System.Object):int:this
G_M24774_IG01:  ;; offset=0000H
       sub      rsp, 40
       xor      eax, eax
       mov      qword ptr [rsp+20H], rax
						;; size=11 bbWeight=1 PerfScore 1.50

G_M24774_IG02:  ;; offset=000BH
       mov      rcx, 0x7FF8A0D1A9B8      ; System.Nullable`1[int]
       call     [CORINFO_HELP_ISINSTANCEOFANY]
       mov      r8, rax
       lea      rcx, [rsp+20H]
       mov      rdx, 0x7FF8A0D1A9B8      ; System.Nullable`1[int]
       call     CORINFO_HELP_UNBOX_NULLABLE
       mov      eax, 0x4D2
       cmp      byte  ptr [rsp+20H], 0
       cmovne   eax, dword ptr [rsp+24H]
						;; size=54 bbWeight=1 PerfScore 9.50

G_M24774_IG03:  ;; offset=0041H
       add      rsp, 40
       ret      
						;; size=5 bbWeight=1 PerfScore 1.25
[...]

Issue #47920, opened 2021-02-05
Ineffective codegen for or pattern type checking


huoyaoyuan:
### Description

Examine this code:

```csharp
using System;
public class C
{
    public bool M0(object o)
    {
        return o is int or uint or long;
    }
    public bool M1(object o)
    {
        return o != null && (o is int or uint or long);
    }
    public bool M2(object o)
    {
        return o != null && (o.GetType() == typeof(int) || o.GetType() == typeof(uint) || o.GetType() == typeof(long));
    }
}
```

IL Compiled:
```msil
.method public hidebysig 
        instance bool M0 (
            object o
        ) cil managed 
    {
        // Method begins at RVA 0x2050
        // Code size 28 (0x1c)
        .maxstack 8

        IL_0000: ldarg.1
        IL_0001: isinst [System.Private.CoreLib]System.Int32
        IL_0006: brtrue.s IL_001a

        IL_0008: ldarg.1
        IL_0009: isinst [System.Private.CoreLib]System.UInt32
        IL_000e: brtrue.s IL_001a

        IL_0010: ldarg.1
        IL_0011: isinst [System.Private.CoreLib]System.Int64
        IL_0016: ldnull
        IL_0017: cgt.un
        IL_0019: ret

        IL_001a: ldc.i4.1
        IL_001b: ret
    } // end of method C::M0

    .method public hidebysig 
        instance bool M1 (
            object o
        ) cil managed 
    {
        // Method begins at RVA 0x206d
        // Code size 33 (0x21)
        .maxstack 8

        IL_0000: ldarg.1
        IL_0001: brfalse.s IL_001f

        IL_0003: ldarg.1
        IL_0004: isinst [System.Private.CoreLib]System.Int32
        IL_0009: brtrue.s IL_001d

        IL_000b: ldarg.1
        IL_000c: isinst [System.Private.CoreLib]System.UInt32
        IL_0011: brtrue.s IL_001d

        IL_0013: ldarg.1
        IL_0014: isinst [System.Private.CoreLib]System.Int64
        IL_0019: ldnull
        IL_001a: cgt.un
        IL_001c: ret

        IL_001d: ldc.i4.1
        IL_001e: ret

        IL_001f: ldc.i4.0
        IL_0020: ret
    } // end of method C::M1

    .method public hidebysig 
        instance bool M2 (
            object o
        ) cil managed 
    {
        // Method begins at RVA 0x2090
        // Code size 75 (0x4b)
        .maxstack 2

[...]

PR #91945, opened 2023-09-12
Optimize unboxing in `System.Data.Common.ObjectStorage:Set`

xtqqczze:
Inline `value.GetType()` instead of using a local, to enable a JIT optimization.

Diffs: https://www.diffchecker.com/QgpKJ9UD/

```
Top method improvements (bytes):
        -456 (-32.50 % of base) : System.Data.Common.dasm - System.Data.Common.ObjectStorage:Set(int,System.Object):this (FullOpts)
```

### DIFF
--- a/src/libraries/System.Data.Common/src/System/Data/Common/ObjectStorage.cs
+++ b/src/libraries/System.Data.Common/src/System/Data/Common/ObjectStorage.cs
@@ -233,7 +233,7 @@ namespace System.Data.Common
                 return;
             }
 
-            if (_dataType == typeof(object) || value.GetType() == typeof(object))
+            if (_dataType == typeof(object) || _dataType.IsInstanceOfType(value))
             {
                 _values[recordNo] = value;
                 return;
--- a/src/libraries/System.Data.Common/src/System/Data/Common/ObjectStorage.cs
+++ b/src/libraries/System.Data.Common/src/System/Data/Common/ObjectStorage.cs
@@ -236,50 +236,49 @@ namespace System.Data.Common
             }
             else
             {
-                Type valType = value.GetType();
-                if (_dataType == typeof(Guid) && valType == typeof(string))
+                if (_dataType == typeof(Guid) && value is string)
                 {
                     _values[recordNo] = new Guid((string)value);
                 }
                 else if (_dataType == typeof(byte[]))
                 {
-                    if (valType == typeof(bool))
+                    if (value is bool)
                     {
                         _values[recordNo] = BitConverter.GetBytes((bool)value);
                     }
-                    else if (valType == typeof(char))
+                    else if (value is char)
                     {
                         _values[recordNo] = BitConverter.GetBytes((char)value);
                     }
-                    else if (valType == typeof(short))
+                    else if (value is short)
                     {
                         _values[recordNo] = BitConverter.GetBytes((short)value);
                     }
-                    else if (valType == typeof(int))
+                    else if (value is int)
                     {
                         _values[recordNo] = BitConverter.GetBytes((int)value);
                     }
-                    else if (valType == typeof(long))
+                    else if (value is long)
                     {
                         _values[recordNo] = BitConverter.GetBytes((long)value);
                     }
-                    else if (valType == typeof(ushort))
+                    else if (value is ushort)
                     {
                         _values[recordNo] = BitConverter.GetBytes((ushort)value);
                     }
-                    else if (valType == typeof(uint))
+                    else if (value is uint)
                     {
                         _values[recordNo] = BitConverter.GetBytes((uint)value);
                     }
-                    else if (valType == typeof(ulong))
+                    else if (value is ulong)
                     {
                         _values[recordNo] = BitConverter.GetBytes((ulong)value);
                     }
-                    else if (valType == typeof(float))
+                    else if (value is float)
                     {
                         _values[recordNo] = BitConverter.GetBytes((float)value);
                     }
-                    else if (valType == typeof(double))
+                    else if (value is double)
                     {
                         _values[recordNo] = BitConverter.GetBytes((double)value);
                     }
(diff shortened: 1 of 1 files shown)


src/libraries/System.Data.Common/src/System/Data/Common/ObjectStorage.cs
```diff
@@ -236,50 +236,49 @@ public override void Set(int recordNo, object value)
...
-                Type valType = value.GetType();
-                if (_dataType == typeof(Guid) && valType == typeof(string))
+                if (_dataType == typeof(Guid) && value is string)
                 {
                     _values[recordNo] = new Guid((string)value);
                 }
                 else if (_dataType == typeof(byte[]))
                 {
-                    if (valType == typeof(bool))
+                    if (value is bool)
```
roji, line 245: nit: can use a switch expression

EgorBo:
there should be no perf difference between `valType == typeof(..)` and `val is PrimitiveType` I belived, at least for JIT.

xtqqczze:
> there should be no perf difference between `valType == typeof(..)` and `val is PrimitiveType` I believe, at least for JIT.

This is part of the current codegen, from the build https://github.com/MihuBot/runtime-utils/issues/177, showing a call to `CORINFO_HELP_UNBOX` that could be eliminated.

```asm
G_M32004_IG28:
       mov      rsi, 0xD1FFAB1E      ; 'System.UInt64'

       cmp      rax, rsi
       jne      SHORT G_M32004_IG31
       mov      r15, gword ptr [r15+0x40]
       mov      rsi, 0xD1FFAB1E      ; System.UInt64
       cmp      qword ptr [rbx], rsi
       je       SHORT G_M32004_IG30
						;; size=34 bbWeight=0.50 PerfScore 3.88
G_M32004_IG29:
       mov      rsi, rbx
       mov      rdi, 0xD1FFAB1E      ; System.UInt64
       mov      rax, 0xD1FFAB1E      ; code for CORINFO_HELP_UNBOX
       call     [rax]CORINFO_HELP_UNBOX
						;; size=25 bbWeight=0.25 PerfScore 0.94
G_M32004_IG30:
       mov      rdi, qword ptr [rbx+0x08]
       mov      rax, 0xD1FFAB1E      ; code for System.BitConverter:GetBytes(ulong):ubyte[]
       call     [rax]System.BitConverter:GetBytes(ulong):ubyte[]
       mov      rdx, rax
       movsxd   rsi, r14d
       mov      rdi, r15
       call     CORINFO_HELP_ARRADDR_ST
       jmp      G_M32004_IG37
						;; size=35 bbWeight=0.50 PerfScore 4.50
```

The issue may be that the JIT is unable to prove `value` is not null.

xtqqczze:
@EgorBo This change is also a little more efficient in terms of IL.

```
  isinst [System.Runtime]System.Single
```

vs:
```
  callvirt instance class [System.Runtime]System.Type [System.Runtime]System.Object::GetType()
  ldtoken [System.Runtime]System.Single
  call class [System.Runtime]System.Type [System.Runtime]System.Type::GetTypeFromHandle(valuetype [System.Runtime]System.RuntimeTypeHandle)
  call bool [System.Runtime]System.Type::op_Equality(class [System.Runtime]System.Type, class [System.Runtime]System.Type)
```

src/libraries/System.Data.Common/src/System/Data/Common/ObjectStorage.cs
huoyaoyuan: This may hit the performance issue logged in #47920. I actually have a branch to improve this, but it's not expandable-proof so I haven't push it.

It's also interesting to see if a following `unbox.any` changes the codegen of `isinst`.

xtqqczze:
@MihuBot

src/libraries/System.Data.Common/src/System/Data/Common/ObjectStorage.cs
xtqqczze: In the case of a switch expression, the JIT is not able to eliminate the calls to `CORINFO_HELP_UNBOX`.

xtqqczze:
@MihuBot

src/libraries/System.Data.Common/src/System/Data/Common/ObjectStorage.cs
huoyaoyuan: I'm surprised that `CORINFO_HELP_UNBOX` is ever evolved after #87241.

huoyaoyuan:
I think we would be able to use `switch` when #47920 is fully resolved.

src/libraries/System.Data.Common/src/System/Data/Common/ObjectStorage.cs
xtqqczze: > I'm surprised that `CORINFO_HELP_UNBOX` is ever evolved after #87241.

The `isinst` + `unbox.any` optimization appears in .NET 7.

xtqqczze:
@MihuBot

src/libraries/System.Data.Common/src/System/Data/Common/ObjectStorage.cs
```diff
@@ -226,73 +226,89 @@ public override bool IsNull(int record)
         public override void Set(int recordNo, object value)
         {
             Debug.Assert(null != value, "null value");
+
             if (_nullValue == value)
             {
                 _values[recordNo] = null;
+                return;
```
MihaZupan, line 233: What is the aim behind replacing `else` blocks with many duplicated `return`s?

xtqqczze: To use a single throw at the end of the method whilst keeping the control flow simple. The JIT will create a single exit.

If this is less readable I can revert to `else` blocks.

xtqqczze: @MihaZupan What is your view on this?

xtqqczze:
> there should be no perf difference between `valType == typeof(..)` and `val is PrimitiveType` I believe, at least for JIT.

This is correct, as long as `valType` isn't stored to a local.

I've refactored to use `value.GetType()` inline.

PR #91960, opened 2023-09-12
Delete .NET Native left overs

jkotas:


src/coreclr/nativeaot/Common/src/Internal/Runtime/MethodTable.cs
```diff
@@ -506,20 +506,19 @@ internal bool IsGenericTypeDefinition
             get
             {
                 Debug.Assert(IsGeneric);
+
+                uint offset = GetFieldOffset(EETypeField.ETF_GenericDefinition);
```
jkotas, line 510: This avoids duplication of the code that computes the offset between the relative and absolute paths. We have used this pattern in a few places already.

MichalStrehovsky (approved):
Thanks!

Issue #91258, opened 2023-08-29
Config binding gen omits a needed comma in an emitted Configure overload replacement


martincostello:
### Description

Updating to .NET SDK `8.0.100-rc.2.23428.11` from `8.0.100-rc.2.23427.4` introduces [compilation errors](https://github.com/martincostello/costellobot/actions/runs/6010809962/job/16303508696?pr=821#step:9:87) in the configuration binding source generator:

```
Error: /home/runner/work/costellobot/costellobot/src/Costellobot/Microsoft.Extensions.Configuration.Binder.SourceGeneration/Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator/BindingExtensions.g.cs(59,227): error CS1003: Syntax error, ',' expected [/home/runner/work/costellobot/costellobot/src/Costellobot/Costellobot.csproj]
Error: /home/runner/work/costellobot/costellobot/src/Costellobot/Microsoft.Extensions.Configuration.Binder.SourceGeneration/Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator/BindingExtensions.g.cs(58,35): error CS0246: The type or namespace name 'IOptionsChangeTokenSource<>' could not be found (are you missing a using directive or an assembly reference?) [/home/runner/work/costellobot/costellobot/src/Costellobot/Costellobot.csproj]
Error: /home/runner/work/costellobot/costellobot/src/Costellobot/Microsoft.Extensions.Configuration.Binder.SourceGeneration/Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator/BindingExtensions.g.cs(58,76): error CS0246: The type or namespace name 'ConfigurationChangeTokenSource<>' could not be found (are you missing a using directive or an assembly reference?) [/home/runner/work/costellobot/costellobot/src/Costellobot/Costellobot.csproj]
```

/cc @layomia

### Reproduction Steps

1. Clone https://github.com/martincostello/costellobot/commit/09ae9eab3d6780af4710b93051e6442f065b3c18.
2. Run `build.ps1` in the root of the repository.

### Expected behavior

The application compiles.

### Actual behavior

The application fails to compile with `CS1003` and `CS0246` errors.

### Regression?

Yes compared to recent preview builds.

### Known Workarounds

None.

### Configuration

.NET SDK `8.0.100-rc.2.23428.11`

ShreyasJejurkar:
[...]

PR #91961, opened 2023-09-12
Fix: Config binder generator doesn't generate code when named arguments are out of order

### LINKED ISSUE #90908: Config binder generator doesn't generate code when named arguments are out of order
This will successfully generate code:
```C#
internal void Configure(IConfiguration configuration) =>
    ConfigurationBinder.Bind(configuration: configuration, instance: this);
```
but this will not:
```C#
internal void Configure(IConfiguration configuration) =>
    ConfigurationBinder.Bind(instance: this, configuration: configuration);
```

buyaa-n:
Config binder generator doesn't generate code when named arguments are out of order like in:

```C#
internal void Configure(IConfiguration configuration) =>
    ConfigurationBinder.Bind(instance: this, configuration: configuration);
```

The issue also happens for ConfigurationBinder.Get(...) overloads. 

Fixes https://github.com/dotnet/runtime/issues/90908


src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Parser/ConfigurationBinder.cs
```diff
@@ -218,7 +231,7 @@ private void ParseGetValueInvocation(BinderInvocation invocation)
                         return;
                     }
 
-                    ITypeOfOperation? typeOfOperation = operation.Arguments[1].ChildOperations.FirstOrDefault() as ITypeOfOperation;
+                    ITypeOfOperation? typeOfOperation = GetArgumentForParameterAtIndex(operation.Arguments, 1).ChildOperations.FirstOrDefault() as ITypeOfOperation;
```
buyaa-n, line 234: The same issue found in `GetValue(...)` invocations, but could not test. Not sure if it is known issue or it is by design, but the generator is not generating the source for any test case I tried. CC @layomia

buyaa-n: Apparently `GetValue` only works for primitive types

layomia:
> The issue also happens for ConfigurationBinder.Get(...) overloads.

I believe it applies to all the extensions methods -- so in addition to `ConfigurationBinder.cs`, we'd need to change the parsing in `OptionsBuilder...cs` and `OptionsConfigurationServiceCollection...cs`.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/Baselines/ConfigurationBinder/Bind_NamedParameters.generated.txt
```diff
@@ -0,0 +1,170 @@
+// <auto-generated/>
```
layomia, line 1: I think these would be better as regular runtime tests. I'd put them here instead https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/ConfigurationBinderTests.Generator.cs.

For the `OptionsBuilder` and `IServiceCollection` extension methods, I'd add tests to https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Options.ConfigurationExtensions/tests/SourceGenerationTests/ConfigurationExtensionsTest.Generator.cs.

Like others in the files I shared, specific syntax patterns are crucial for these new tests. The comments need to be clear on what scenario is being tested so that future devs won't change them.

buyaa-n:
> I believe it applies to all the extensions methods -- so in addition to `ConfigurationBinder.cs`, we'd need to change the parsing in `OptionsBuilder...cs` and `OptionsConfigurationServiceCollection...cs`.

Did not found any indexed access of arguments in those files, only `TypeArguments` which is safe and no need to change.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/Baselines/ConfigurationBinder/Bind_NamedParameters.generated.txt
buyaa-n: > I think these would be better as regular runtime tests. I'd put them here instead https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/ConfigurationBinderTests.Generator.cs.

I did added test there, for repro removed the fix and run tests but test did not fail, for example I added this::

```cs
[Fact]
public void NamedParameterTests_Bind()
{

    IConfiguration configuration = TestHelpers.GetConfigurationFromJsonString("""{"MyString":"hi", "MyList":[1], "MyInt":1}""");

    var m1 = new MyClass();

    ConfigurationBinder.Bind(instance: m1, configuration: configuration);

    Assert.Equal(1, m1.MyInt);
    Assert.Equal("hi", m1.MyString);
    Assert.Equal(1, m1.MyList[0]);
}

public class MyClass
{
    public string MyString { get; set; }
    public int MyInt { get; set; }
    public List<int> MyList { get; set; }
    public Dictionary<string, string> MyDictionary { get; set; }
}
```

I could not debug because my VS showing error, so could not check which Bind method it is calling (generated or not), I guess the test doesn't throw/fail when the generator did not produce source and it silently calls the original Bind method.

buyaa-n: Or am I missing something here @layomia ?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Parser/ConfigurationBinder.cs
buyaa-n: Added corresponding test

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/Baselines/ConfigurationBinder/Bind_NamedParameters.generated.txt
buyaa-n: Confirmed offline that the test falling back to use reflection version. As discussed with @layomia removed the text files added, and just testing that the file is generated and row count, no need to test row by row as other tests already doing that

layomia:
> > I believe it applies to all the extensions methods -- so in addition to `ConfigurationBinder.cs`, we'd need to change the parsing in `OptionsBuilder...cs` and `OptionsConfigurationServiceCollection...cs`.
> 
> Did not found any indexed access of arguments in those files, only `TypeArguments` which is safe and no need to change.

Per offline discussion, we do need a fix/tests for these as well, e.g. 

```cs
using System;
using Microsoft.Extensions.DependencyInjection;
					
public class Program
{
	public static void Main()
	{
		ServiceCollection services = new();
		IConfiguration configuration = new ConfigurationBuilder().Build();
		services.Configure<string>(name: "string", configureBinder: null, config: configuration);
	}
}
```

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.cs
```diff
@@ -15,6 +15,107 @@ public partial class ConfigurationBindingGeneratorTests
...
+                        using Microsoft.Extensions.Configuration;
+
+                        public class Program
+                        {
+                            public static void Main()
+                            {
+                                ConfigurationBuilder configurationBuilder = new();
+                                IConfigurationRoot config = configurationBuilder.Build();
+
+                                var str = ConfigurationBinder.GetValue(key: "key", configuration: config, type: typeof(string));
```
layomia, line 100: We also need variants for regular extension method call syntax - `config.GetValue(key: "key", type: typeof(string))`. Here & wherever applicable.

buyaa-n:
> Per offline discussion, we do need a fix/tests for these as well, e.g.
> 
> ```cs
> using System;
> using Microsoft.Extensions.Configuration;
> using Microsoft.Extensions.DependencyInjection;
> 					
> public class Program
> {
> 	public static void Main()
> 	{
> 		ServiceCollection services = new();
> 		IConfiguration configuration = new ConfigurationBuilder().Build();
> 		services.Configure<string>(name: "string", configureBinder: null, config: configuration);
> 	}
> }
> ```

@layomia I have been testing similar code to see if I can repro it, but encountering errors similar to https://github.com/dotnet/runtime/issues/91258 even though I am using latest `8.0.100-rc.2.23464.22` SDK:

```log
\ConfigTestWebApp\Microsoft.Extensions.Configuration.Binder.SourceGeneration\Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator\BindingExtensions.g.cs(70,35): 
  error CS0246: The type or namespace name 'IOptionsChangeTokenSource<>' could not be found (are you missing a using directive or an assembly reference?)
\ConfigTestWebApp\Microsoft.Extensions.Configuration.Binder.SourceGeneration\Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator\BindingExtensions.g.cs(70,76): 
  error CS0246: The type or namespace name 'ConfigurationChangeTokenSource<>' could not be found (are you missing a using directive or an assembly reference?) 
\ConfigTestWebApp\Microsoft.Extensions.Configuration.Binder.SourceGeneration\Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator\BindingExtensions.g.cs(71,178): 
  error CS0103: The name 'BindCoreMain' does not exist in the current context
```

layomia:
That's odd. Checking that out.

In the meantime, you can verify your fix by adding unit tests like the ones you add for `ConfigurationBinder`

buyaa-n:
> In the meantime, you can verify your fix by adding unit tests like the ones you add for `ConfigurationBinder`

Yes, of course that is the plan, just wanted to let you know about these errors as it might be important.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.cs
buyaa-n: Done

buyaa-n:
> > > I believe it applies to all the extensions methods -- so in addition to `ConfigurationBinder.cs`, we'd need to change the parsing in `OptionsBuilder...cs` and `OptionsConfigurationServiceCollection...cs`.
> > 
> > 
> > Did not found any indexed access of arguments in those files, only `TypeArguments` which is safe and no need to change.
> 
> Per offline discussion, we do need a fix/tests for these as well, e.g.
> 
> ```cs
> using System;
> using Microsoft.Extensions.Configuration;
> using Microsoft.Extensions.DependencyInjection;
> 					
> public class Program
> {
> 	public static void Main()
> 	{
> 		ServiceCollection services = new();
> 		IConfiguration configuration = new ConfigurationBuilder().Build();
> 		services.Configure<string>(name: "string", configureBinder: null, config: configuration);
> 	}
> }
> ```

Added unit tests for those, there were no need a code fix for those extension methods

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.Options.cs
```diff
@@ -60,6 +60,85 @@ public class MyClass2
...
+                            public List<int> MyList { get; set; }
+                            public Dictionary<string, string> MyDictionary { get; set; }
+                        }
+                    }
+                    """;
+
+            var (d, r) = await RunGenerator(source);
+            Assert.Equal(1, r.Length);
+            Assert.Empty(d);
+            Assert.Equal(234, r[0].SourceText.Lines.Count);
```
buyaa-n, line 100: The line count produced changes with the generator updates, we might want to remove these count asserts or maybe just assert that the line count is greater than `0` or some other specific number.

layomia: I think `Assert.Single` and `Assert.Empty` are fine. We just need to know that the generator picked up those invocations and produced something.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.Options.cs
```diff
@@ -60,6 +60,79 @@ public class MyClass2
...
+                    public class Program
+                    {
+                        public static void Main()
+                        {
+                            ConfigurationBuilder configurationBuilder = new();
+                            IConfiguration config = configurationBuilder.Build();
+                            IConfigurationSection section = config.GetSection("MySection");
+                            ServiceCollection services = new();
+
+                            OptionsConfigurationServiceCollectionExtensions.Configure<MyClass>(config: section, services: services);
```
layomia, line 80: We need to (also) have separate tests for each overload, to validate that every overload is handled correctly. For these & for `ConfigurationBinder`.

layomia: I'm surprised that no changes were needed for the `Options.ConfigurationExtensions` methods. Testing each one in isolation might reveal a gap.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.cs
```diff
@@ -15,6 +15,110 @@ public partial class ConfigurationBindingGeneratorTests
...
+                            {
+                                public string MyString { get; set; }
+                                public int MyInt { get; set; }
+                                public List<int> MyList { get; set; }
+                                public Dictionary<string, string> MyDictionary { get; set; }
+                            }
+                        }
+                    """;
+
+            await VerifyThatSourceGenerated(source);
```
layomia, line 119:
```suggestion
            await VerifyThatSourceIsGenerated(source);
```

eiriktsarpalis (approved):

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.Options.cs
buyaa-n: > I'm surprised that no changes were needed for the `Options.ConfigurationExtensions` methods. Testing each one in isolation might reveal a gap.

Not surprising to me, because those implementations did not use/reference arguments by index at all. Anyway, separated the test for each overload.

layomia (approved):
Thanks!

buyaa-n:
One CI build not finishes even running for more than 3.5hrs, going to merge

PR #91968, opened 2023-09-12
crashinfo.cpp/crashinfounix.cpp: use off_t instead of off64_t

ayakael:
Build on latest Alpine Linux fails, as musl does not define `off64_t` as `off_t`. It's also my understanding that `off64_t` is a product of a bygone era. This patches replaces `off64_t` with `off_t` (with other 64 derivatives)


mikem8361 (approved):

tommcdon:
All checks have passed, merging

PR #91974, opened 2023-09-12
Minimize net diff of binder-gen visibility/identifier-clash fix

layomia:
Follow up to https://github.com/dotnet/runtime/pull/91657. Primarily reverts JSON changes made in that PR.

https://github.com/dotnet/runtime/pull/91967 shows the overall diff of the bug fix and addresses feedback to switch to use the Roslyn API for accessibility checks.

### DIFF
--- a/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Specs/KnownTypeSymbols.cs
+++ b/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Specs/KnownTypeSymbols.cs
@@ -13,6 +13,8 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
 {
     internal sealed record KnownTypeSymbols
     {
+        public CSharpCompilation Compilation { get; }
+
         public INamedTypeSymbol String { get; }
         public INamedTypeSymbol? CultureInfo { get; }
         public INamedTypeSymbol? DateOnly { get; }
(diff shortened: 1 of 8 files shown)


src/libraries/System.Text.Json/gen/JsonSourceGenerator.Parser.cs
```diff
@@ -1592,6 +1590,37 @@ private static string DeterminePropertyNameFieldName(string effectiveJsonPropert
...
+                }
+
+                if (type is not INamedTypeSymbol namedType || !namedType.IsGenericType)
+                {
+                    return type.Name;
+                }
+
+                StringBuilder sb = new();
+
+                string name = namedType.Name;
```
tarekgh, line 1609: >string name = namedType.Name; [](http://example.com/codeflow?start=16&length=29)

why you need to declare this? can't you just append `namedType.Name` to the string builder.

tarekgh (approved):
LGTM. will be good to wait @eiriktsarpalis to look at the change in the json source gen.

layomia:
Following offline discussion with @eiriktsarpalis, I've addressed his last concern regarding using the built-in API for accessibility checks. https://github.com/dotnet/runtime/pull/91974/commits/3b828344ffbb1147538ba3d178cbeea16e45c405

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Specs/KnownTypeSymbols.cs
```diff
@@ -13,6 +13,8 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
 {
     internal sealed record KnownTypeSymbols
     {
+        public CSharpCompilation Compilation { get; }
```
layomia, line 16: (This diff has no bearing on generator incrementality.)

eiriktsarpalis (approved):
Thanks!

PR #89736, opened 2023-07-31
Utilize the new `ref readonly` language feature

tannergooding:
This resolves #85911. The dotnet/sdk currently has the same version of Roslyn inserted (`v4.8.0-1.23378.8`).

This is broken into 3 main commits:
* `in` to `ref readonly`
* `ref` to `in`
* `ref` to `ref readonly`

As per the API review, none of these are binary breaking changes and none of these introduce new errors. However, users may see new warnings for two of the scenarios:
* An API was changed from `in` to `ref readonly`, users will now be asked to use in explicitly
* An API was changed from `ref` to `in`, users will now be asked to use in or no longer specify ref

PR #91983, opened 2023-09-13
Update `Marshal.QueryInterface()` argument modifier

### LINKED ISSUE #91981: .NET 8 RC1 can not compile with Windows app sdk and CsWinRT
### Description

.NET 8 RC1 can not compile with Windows app sdk and CsWinRT

In.net8 RC1, the Marshal.QueryInterface() method is defined from
public static int QueryInterface (IntPtr pUnk, ref Guid iid, out IntPtr ppv);
Become a
public static int QueryInterface (IntPtr pUnk, in Guid iid, out IntPtr ppv);
Please update the content in WinRT.cs immediately, otherwise the application will not compile

image

### Reproduction Steps

In describe bug

### Expected behavior

In describe bug

### Actual behavior

The Marshal.QueryInterface() method should be compatible with both ref and in arguments

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

.NET 8 RC1
Windows App SDK 1.4.0
CSWINRT 2.0.3

### Other information

None

AaronRobinsonMSFT:
Fixes #91981

Change `in` modifier to `ref readonly` to avoid warnings in existing interop code.

### DIFF
--- a/src/libraries/System.Runtime.InteropServices/src/System/Runtime/InteropServices/Marshalling/IIUnknownStrategy.cs
+++ b/src/libraries/System.Runtime.InteropServices/src/System/Runtime/InteropServices/Marshalling/IIUnknownStrategy.cs
@@ -32,7 +32,7 @@ namespace System.Runtime.InteropServices.Marshalling
         /// <param name="iid">The IID (Interface ID) to query for.</param>
         /// <param name="ppObj">The resulting interface.</param>
         /// <returns>Returns an HRESULT represents the success of the operation.</returns>
-        /// <seealso cref="Marshal.QueryInterface(nint, in Guid, out nint)"/>
+        /// <seealso cref="Marshal.QueryInterface(nint, ref readonly Guid, out nint)"/>
         public int QueryInterface(void* instancePtr, in Guid iid, out void* ppObj);
 
         /// <summary>
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs
@@ -130,7 +130,7 @@ namespace System.Runtime.InteropServices
             return SizeOfHelper(t, throwIfNotMarshalable: true);
         }
 
-        public static unsafe int QueryInterface(IntPtr pUnk, in Guid iid, out IntPtr ppv)
+        public static unsafe int QueryInterface(IntPtr pUnk, ref readonly Guid iid, out IntPtr ppv)
         {
             ArgumentNullException.ThrowIfNull(pUnk);
 
--- a/src/libraries/System.Runtime.InteropServices/ref/System.Runtime.InteropServices.cs
+++ b/src/libraries/System.Runtime.InteropServices/ref/System.Runtime.InteropServices.cs
@@ -1104,7 +1104,7 @@ namespace System.Runtime.InteropServices
         public static object? PtrToStructure(System.IntPtr ptr, [System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembersAttribute(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.NonPublicConstructors| System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicConstructors)] System.Type structureType) { throw null; }
         public static T? PtrToStructure<[System.Diagnostics.CodeAnalysis.DynamicallyAccessedMembersAttribute(System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.NonPublicConstructors | System.Diagnostics.CodeAnalysis.DynamicallyAccessedMemberTypes.PublicConstructors)]T>(System.IntPtr ptr) { throw null; }
         public static void PtrToStructure<T>(System.IntPtr ptr, [System.Diagnostics.CodeAnalysis.DisallowNullAttribute] T structure) { }
-        public static int QueryInterface(System.IntPtr pUnk, in System.Guid iid, out System.IntPtr ppv) { throw null; }
+        public static int QueryInterface(System.IntPtr pUnk, ref readonly System.Guid iid, out System.IntPtr ppv) { throw null; }
         public static byte ReadByte(System.IntPtr ptr) { throw null; }
         public static byte ReadByte(System.IntPtr ptr, int ofs) { throw null; }
         [System.Diagnostics.CodeAnalysis.RequiresDynamicCode("Marshalling code for the object might not be available")]


src/libraries/System.Runtime.InteropServices/src/System/Runtime/InteropServices/Marshalling/IIUnknownStrategy.cs
```diff
@@ -32,7 +32,7 @@ public unsafe interface IIUnknownStrategy
         /// <param name="iid">The IID (Interface ID) to query for.</param>
         /// <param name="ppObj">The resulting interface.</param>
         /// <returns>Returns an HRESULT represents the success of the operation.</returns>
-        /// <seealso cref="Marshal.QueryInterface(nint, in Guid, out nint)"/>
+        /// <seealso cref="Marshal.QueryInterface(nint, ref readonly Guid, out nint)"/>
         public int QueryInterface(void* instancePtr, in Guid iid, out void* ppObj);
```
stephentoub, line 36: `in` is correct here?

stephentoub (approved):
cc: @tannergooding

src/libraries/System.Runtime.InteropServices/src/System/Runtime/InteropServices/Marshalling/IIUnknownStrategy.cs
AaronRobinsonMSFT: That one is. It is a new API and we designed it to be `in`.

jkotas:
> Change in modifier to ref readonly to avoid warnings in existing interop code.

I thought that it is "by design" that the change will produce warnings in existing code. There are other APIs with the same problem. For example, this compiled fine in .NET 7 and it will produce the exact same warning in .NET 8 as the Marshal.QueryInterface case.

```csharp
var span = new byte[10];
int value = 42;
MemoryMarshal.Write<int>(span, ref value);
```

If we are not happy with this change, should be roll it back in all APIs that were affected?

AaronRobinsonMSFT:
> If we are not happy with this change, should be roll it back in all APIs that were affected?

I think in general I would agree that introducing new warnings should be avoided, but I do see the benefit in most cases. My specific concern here is the legacy nature of the `Marshal.QueryInterface()` API. I would be surprised if we see many users complain about the `MemoryMarshal.Write<int>()` APIs, but the `Marshal.QueryInterface()` API is old enough and in code that people rarely want to touch at all.

My calculus is on the nature of the `Marshal.QueryInterface()` API, it is very old and used in code that is rarely touched, and it adds no benefit to the interop scenario. Therefore any impact should be weight more.

jkotas:
These two changes from https://github.com/dotnet/runtime/pull/89736#issue-1829841309 can introduce warnings in existing code:
- An API was changed from in to ref readonly, users will now be asked to use in explicitly
- An API was changed from ref to in, users will now be asked to use in or no longer specify ref

jkotas:
> the legacy nature of the Marshal.QueryInterface() API

Ok, makes sense. 

It is unfortunate that we will end up with no consistency in how Guids are treated by various QI APIs:
- `ICustomQueryInterface.GetInterface` - ref Guid
- `Marshal.QueryInterface` - ref readonly Guid
- `IIUnknownStrategy.QueryInterface` - in Guid

jkotas:
cc @Sergio0694

Sergio0694:
Oh, well this is unfortunate 🥲

I was hoping starting with .NET 8 we'd be able to just pass `Type.GUID` without having to manually copy to a local first, which would've been nice.

Sergio0694:
If people are _actually_ failing to build, isn't that simply because they're on an older version of Roslyn? The changes needed to properly support `ref readonly` didn't make it to VS 2022 Preview just yet I think. Like, the build error in #91981 will not happen if you're using an up to date version of Roslyn, it will only generate a warning (see language spec for `ref readonly`). Considering that, are we sure we actually do need to revert this signature change? The fact some new warnings would be generated was discussed and approved during API review, but there's no actual breaking changes if you have updated tooling 🤔

hamarb123:
This should be fine to have as `in`. There should only be an issue if they've converted the warning into an error. We should ask them if they've turned that on (if they've got this on, then it's no different to other new warnings suddenly becoming errors on new TFMs) AND check if they're on an up-to-date compiler, as an older compiler would obviously have issues. And if that's not the case, ping the roslyn team to see if they know what's gone wrong, instead of merging this since it is supposed to be fine.

To be clear: this should not be merged on any version, because the issue shouldn't exist to start with. There is something else that has gone wrong. Either it's that they've enabled WarningsAsErrors (in which case they're being told about some unideal usage of an API as a warning, which they've opted to convert into an error), they've got an out-of-date compiler (either their fault, or the latest .NET 8 build has an old compiler which doesn't support the feature properly), or there's a bug in roslyn (which would obviously need to be fixed).

On a side note, this is why I suggested to the roslyn team on the issue (iirc I'm pretty sure I did anyway) that there should be no warning when using `ref` on an `in` parameter. If the user's issue is caused by WarningsAsErrors or similar, then would it be too late to fix the problem properly by changing this (i.e., by requesting the roslyn team to allow `ref` to be used for `in` parameters without a warning)?

stephentoub:
> The fact some new warnings would be generated was discussed and approved during API review

In general, yes. We didn't go API by API and discuss every single one being changed. And for things in interop, the folks responsible for interop (aka Aaron) get to weigh in (and apparently hadn't been aware of this).

> there's no actual breaking changes if you have updated tooling

Whether or not it's something we consider to be "breaking", it's still friction. And the person responsible for our interop system is saying that friction isn't warranted.

> WarningsAsErrors / they've got an out-of-date compiler / or there's a bug in roslyn

Or `LangVersion` is set to something 11 or lower. It's a supported scenario to upgrade your TFM but not your LangVersion, and in that case, this will end up being an error, not a warning, regardless of WarningsAsErrors.

Sergio0694:
> "Whether or not it's something we consider to be "breaking", it's still friction. And the person responsible for our interop system is saying that friction isn't warranted."

Sorry, let me rephrase that. Of course I do respect what Aaron is saying here, all I'm trying to say is just, given this PR is marked as a fix for an issue which is just a consequence of that user trying to compile with outdated tooling, and not a real break caused by the signature change when using C# 12, is it possible that the apparent need to revert this might've been influenced by that? As in, that considering the additional context (and the fact that user would not in fact see a build error as soon as they update Roslyn to latest), is there a chance undoing this might not actually be needed? 😅

> "It's a supported scenario to upgrade your TFM but not your LangVersion, and in that case, this will end up being an error, not a warning, regardless of WarningsAsErrors."

This is more of a side note, but I find this to be a bit confusing. One of the main arguments for including those relaxed combinations of ref modifiers as part of the spec was specifically to allow changing the signature of APIs like this without it being a source breaking change. If se say that upgrading TFM without keeping the language version in sync is a perfectly supported scenario, doesn't that kinda just invalidate all of that? Because you'd still end up in a situation where you'd have users on an "officially supported" scenario getting (unwanted) breaking changes when migrating to .NET 8 🤔

stephentoub:
> doesn't that kinda just invalidate all of that?

No, it's just one more thing to factor in. It needs to be worth it. And Aaron is saying it's not worth it. There's a scale here and with all changes that introduce friction; this isn't something where everything is definitively 100% right or 100% wrong.

stephentoub:
cc: @jaredpar

tannergooding (approved):
Seems good to me overall.

I'm not particularly a fan of avoiding new warnings, but I understand why there is a desire to do it here in particular.

It being `ref readonly` notably also doesn't block `typeof(T).GUID`, it just transfers the warning there instead which the people who want to explicit take advantage of this can suppress instead.

This means by changing from `ref` to `ref readonly` (rather than `in`) here on this very old API means we maintain the status quo for existing code, we improve the experience for new code by removing the need to `Unsafe.As` to strip the readonlyness, and we allow power users to still pass rvalues if they are willing to suppress a warning.

PR #91985, opened 2023-09-13
Ensure the adapter name 100% matches when parsing proc/net/dev

### LINKED ISSUE #91982: Network Traffic Statistics on Linux
Recently I'm working on getting network traffic (bytes received/sent) on Linux. I learned that it can be done by reading `/proc/net/dev`. But later I found this has already been implemented in `NetworkInterface` and it just does the same thing: parsing `/proc/net/dev`.

https://github.com/dotnet/runtime/blob/3aeefbdd5e975e287effaa8670422837dc661d68/src/libraries/System.Net.NetworkInformation/src/System/Net/NetworkInformation/StringParsingHelpers.Statistics.cs#L404-L446

I notice that it uses `line.Contains(name)` to locate the specific adapter. I'm curious if it's safe to do so. For example, is it possible that both `eth1` and `eth10` exist?

skyoxZ:
Fix #91982

* Add checking the adapter name.

* Remove unused variable `index`.


skyoxZ:
Sorry for a PR without verification. I'm closing it and I may not have enough time to fix it recently.

Issue #91965, opened 2023-09-12
AOT compilation fails when NoWarn spans separate lines


heaths:
### Description

The `.rsp` file that is generated may contain `NoWarn` which, as isn't uncommon, may span lines based on project configuration. For example, in a `Directory.Build.props`, someone may have specified:

```xml
<NoWarn>
  $(NoWarn);
  CS1234;
</NoWarn>
```

In this case, you end up with an error like:

```text
Generating native code
EXEC : error : No files matching ;NU5105;CA1812; [C:\src\myproject.csproj::TargetFramework=net7.0]
  Internal.CommandLine.CommandLineException: No files matching ;NU5105;CA1812;
     at Internal.CommandLine.Helpers.AppendExpandedPaths(Dictionary`2, String, Boolean) in /_/src/coreclr/tools/Common/CommandLine/CommandLineHelpers.cs:line 57
     at ILCompiler.Program.ParseCommandLine(String[]) in /_/src/coreclr/tools/aot/ILCompiler/Program.cs:line 347
     at ILCompiler.Program.Run(String[]) in /_/src/coreclr/tools/aot/ILCompiler/Program.cs:line 427
     at ILCompiler.Program.Main(String[]) in /_/src/coreclr/tools/aot/ILCompiler/Program.cs:line 1150
```

### Reproduction Steps

1. Create a simple console application.
2. Add `<PublishAot>true</PublishAot>` to a `PropertyGroup`.
3. Add the following to a `PropertyGroup` exactly as formatted:
   ```xml
   <NoWarn>
     $(NoWarn>;
     NU5105;
   </NoWarn>
4. Run `dotnet publish`.

### Expected behavior

Publishes native image.

### Actual behavior

Fails because `.rsp` file has line breaks that make the CLI think the `NoWarn` values are files e.g. partial response file:

```
--feature:System.Threading.Thread.EnableAutoreleasePool=false
--feature:System.Text.Encoding.EnableUnsafeUTF7Encoding=false
--stacktracedata
--scanreflection
--nowarn:"
      
      ;NU5105;CA1812;
      CA1716; 
      CA1308; 
      CA1819; 
      CA1710; 
      CA1028; 
      CA1032; 
      CA1063; 
      CA1066; 
      CA1815; 
      CA2007; 
      CA2231; 
      CA2225; 
      CA1714; 
      CA1062; 
      CA1031; 
      CA2000; 
      CA2012; ;
      MSB3245; ;IL2121;1701;1702"
```

[...]

PR #92010, opened 2023-09-13
Replace newline characters in NoWarn

am11:
When writing the rsp file, we should not emit the newline characters due to $(NoWarn) evaluation because for one, it breaks response file parsing. Alternative (or additional) fix is to make parsing handle multi-line values among other things, which is a bit involved.

This is a simplest fix for https://github.com/dotnet/runtime/issues/91965.

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -263,7 +263,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <IlcArg Condition="$(IlcSystemModule) != ''" Include="--systemmodule:$(IlcSystemModule)" />
       <IlcArg Condition="'$(_targetOS)' == 'win' and $(IlcMultiModule) != 'true' and '$(IlcGenerateWin32Resources)' != 'false'" Include="--win32resourcemodule:%(ManagedBinary.Filename)" />
       <IlcArg Condition="$(IlcDumpIL) == 'true'" Include="--ildump:$(NativeIntermediateOutputPath)%(ManagedBinary.Filename).il" />
-      <IlcArg Condition="$(NoWarn) != ''" Include='--nowarn:"$([MSBuild]::Escape($(NoWarn)))"' />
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -263,7 +263,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <IlcArg Condition="$(IlcSystemModule) != ''" Include="--systemmodule:$(IlcSystemModule)" />
       <IlcArg Condition="'$(_targetOS)' == 'win' and $(IlcMultiModule) != 'true' and '$(IlcGenerateWin32Resources)' != 'false'" Include="--win32resourcemodule:%(ManagedBinary.Filename)" />
       <IlcArg Condition="$(IlcDumpIL) == 'true'" Include="--ildump:$(NativeIntermediateOutputPath)%(ManagedBinary.Filename).il" />
-      <IlcArg Condition="$(NoWarn) != ''" Include='--nowarn:"$([MSBuild]::Escape($(NoWarn)))"' />
+      <IlcArg Condition="$(NoWarn) != ''" Include='--nowarn:"$([MSBuild]::Escape($(NoWarn)).Replace(`%0A`, ``).Replace(`%0D`, ``))"' />
```
jkotas, line 266: @adamsitnik Is it expected that System.CommandLine response file parser terminates the argument at EOL even when the EOL is enclosed in quotes?

am11: It's our own helper which parses the response file: https://github.com/dotnet/runtime/blob/519d681ab2e5f9ed870d55128088f12d355e32cf/src/coreclr/tools/Common/CommandLineHelpers.cs#L386

jkotas: Ok, I agree that this is a simplest fix.

jkotas (approved):
LGTM. 

@dotnet/ilc-contrib Any concerns?

MichalStrehovsky (approved):
Thank you!

PR #92023, opened 2023-09-13
Fix si.ibAllocated in FrozenObjectHeapManager

EgorBo:
`ibAllocated` is expected to be a real size of a segment, not a pointer to the end.

Regressed in https://github.com/dotnet/runtime/pull/90847

### DIFF
--- a/src/coreclr/vm/frozenobjectheap.cpp
+++ b/src/coreclr/vm/frozenobjectheap.cpp
@@ -184,6 +184,8 @@ void FrozenObjectSegment::RegisterOrUpdate(uint8_t* current, size_t sizeCommited
         si.ibCommit = sizeCommited;
         si.ibReserved = m_Size;
 
+        assert((size_t)current >= (size_t)si.pvMem);
+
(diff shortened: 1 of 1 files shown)


EgorBo:
PTAL @cshung @tommcdon, thanks for help!

src/coreclr/vm/frozenobjectheap.cpp
```diff
@@ -180,7 +180,7 @@ void FrozenObjectSegment::RegisterOrUpdate(uint8_t* current, size_t sizeCommited
         segment_info si;
         si.pvMem = m_pStart;
         si.ibFirstObject = sizeof(ObjHeader);
-        si.ibAllocated = (size_t)current;
+        si.ibAllocated = (size_t)current - (size_t)si.pvMem;
```
cshung, line 183: You might want to `assert ((size_t)current >= (size_t)si.pvMem)`.

cshung (approved):
LGTM, thanks for the quick fix!

tommcdon (approved):
Thanks!

Issue #91139, opened 2023-08-25
JIT: antigen test case fails when unwind fragment size is set very low


AndyAyersMS:
Compile: [TestClass156-test-assertion.cs.txt](https://github.com/dotnet/runtime/files/12443699/TestClass156-test-assertion.cs.txt)

Run on osx arm64 with
```
export COMPlus_JitAggressiveInlining=1
export COMPlus_JitSplitFunctionSize=100
export COMPlus_GCStress=0x4

Assert failure(PID 94626 [0x000171a2], Thread: 23328476 [0x163f6dc]): !"About to FailFast. set DOTNET_AssertOnFailFast=0 if this is expected"
    File: /Users/andya/repos/runtime0/src/coreclr/vm/jithelpers.cpp Line: 4616
    Image: /Users/andya/repos/runtime0/artifacts/tests/coreclr/osx.arm64.Checked/Tests/Core_Root/corerun
```
(may also repro on other arm64 hosts, did not try)

By increasing the split function size you can make the test "pass":
```
export COMPlus_JitAggressiveInlining=1
export COMPlus_JitSplitFunctionSize=138
export COMPlus_GCStress=0x4

Unhandled exception. System.OverflowException: Value was either too large or too small for a Decimal.
   at System.Decimal.op_Multiply(Decimal d1, Decimal d2)
   at TestClass.LeafMethod0() in /Users/andya/bugs/antigen/156.cs:line 66
   at TestClass.Method21(S2 p_s2_593, UInt16& p_ushort_594, Double& p_double_595, S2 p_s2_596, S1 p_s1_597, UInt32 p_uint_598, Decimal& p_decimal_599, S1& p_s1_600) in /Users/andya/bugs/antigen/156.cs:line 1309
   at TestClass.Method0() in /Users/andya/bugs/antigen/156.cs:line 5701
   at TestClass.Main(String[] args) in /Users/andya/bugs/antigen/156.cs:line 5798
```

AndyAyersMS:
@BruceForstall FYI

BruceForstall:
Looks like we're failing fast due to a GS cookie check failure.

I've seen it also fail with more "normal" GC hole asserts (e.g., CREATE_CHECK_STRING).

BruceForstall:
I've been repro'ing on win-arm64 with:

```
set COMPlus_JitAggressiveInlining=1
set COMPlus_JitSplitFunctionSize=100
set COMPlus_GCStress=F
```

[...]

PR #92027, opened 2023-09-13
Fix assert about maximum fragment split size

### LINKED ISSUE #91251: Assertion failed 'curSize < maxSplitSize' during 'Emit code'
```c#
// Found by Antigen


using System;
public class TestClass
{
    {
    }
}
/*
Environment:

set COMPlus_JitSplitFunctionSize=4

Assert failure(PID 5728 [0x00001660], Thread: 10212 [0x27e4]): Assertion failed 'curSize < maxSplitSize' in '(dynamicClass):InvokeStub_EventAttribute.set_Level(System.Object,System.Object,ulong):System.Object' during 'Emit code' (IL size 25; hash 0x433e37b5; FullOpts)
    File: D:\a\_work\1\s\src\coreclr\jit\emit.cpp Line: 3146
    Image: e:\kpathak\CORE_ROOT\corerun.exe
*/

BruceForstall:
We want to support arbitrary `DOTNET_JitSplitFunctionSize` values for stressing the unwind fragment splitting logic, but there are limitations to the smallest size that can be split. Thus, we can exceed a specified minimum. Loosen an assert that the last fragment be less than the specified split size, since it might exceed that minimum, e.g., by being a large epilog. Leave the assert specifying the hard requirement of maximum fragment size.

Fixes #91251

### DIFF
--- a/src/coreclr/jit/emit.cpp
+++ b/src/coreclr/jit/emit.cpp
@@ -3143,7 +3143,7 @@ void emitter::emitSplit(emitLocation*         startLoc,
     } // end for loop
 
     splitIfNecessary();
-    assert(curSize < maxSplitSize);
+    assert(curSize < UW_MAX_FRAGMENT_SIZE_BYTES);
 }
 
 /*****************************************************************************


BruceForstall:
@jakobbotsch PTAL
cc @dotnet/jit-contrib

jakobbotsch (approved):

AndyAyersMS:
Is this for #91139 or for #91251 or for both of them?

BruceForstall:
> Is this for #91139 or for #91251 or for both of them?

Just #91251 (I updated the comment above).

PR #92029, opened 2023-09-13
Merge tests under tracing / Reflection / ReadyToRun / Regressions

trylek:


src/tests/Regressions/coreclr/0416/hello.cs
```diff
@@ -1,15 +1,16 @@
...
-		return 100;
-	}
+    [Fact]
+    public static void TestEntryPoint()
+    {
+        Console.WriteLine("Make sure to set the following env vars:");
+        Console.WriteLine("  DOTNET_LogEnable=1");
+        Console.WriteLine("  DOTNET_LogToConsole=1");
+        Console.WriteLine("This test ensures that logging is working");
+    }
```
markples, line 15: This test doesn't seem to be doing that.

This isn't related to your change except that I was going to comment that the comment for RPI should indicate the environment variables, except...

(I wouldn't hold this change up - just file an item or even just delete this test)

trylek: Thanks Mark for spotting, I do admit I haven't scrutinized every single test but this one does look weird. I'm going to push a commit deleting it and, if someone complains, we can bring it back and possibly try to make sure it does something reasonable.

src/tests/Regressions/coreclr/22021/provider.ilproj
```diff
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.IL">
   <PropertyGroup>
     <OutputType>Library</OutputType>
+    <CLRTestKind>BuildOnly</CLRTestKind>
```
markples, line 4: I think D.B.targets will now automatically set this to `SharedLibrary` - did you hit a problem with that?

src/tests/readytorun/coreroot_determinism/coreroot_determinism.csproj
```diff
@@ -1,5 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk" ToolsVersion="Current">
   <PropertyGroup>
+    <!-- Needed for CLRTestTargetUnsupported, GCStressIncompatible, CrossGenTest -->
+    <RequiresProcessIsolation>true</RequiresProcessIsolation>
     <OutputType>exe</OutputType>
```
markples, line 5: There are several lowercase exes in this batch that got left behind like before.

markples (approved):
Thank you for continuing to push on this!

src/tests/readytorun/coreroot_determinism/coreroot_determinism.csproj
trylek: Fixed in 12th commit.

src/tests/Regressions/coreclr/22021/provider.ilproj
trylek: Thanks for pointing that out, I tried to rebuild the tests without the property and it works fine, in fact I could remove both properties as I believe OutputType is Library by default. Fixed in 15th commit.

PR #92034, opened 2023-09-13
Remove code that is only applicable to emitting mscorlib.dll for .NET Framework

jkoritzinsky:
These code blocks only apply to compiling/emitting the core library with a GUID (which we only do in .NET Framework) and a custom module name (which we only do in .NET Framework). Additionally, we build SPCL with Roslyn, so even then these code-paths are dead.


jkotas:
Also delete `LIBID_ComPlusRuntime` from cor.h (and other GUIDs next to it that appear to be unused)?

jkotas (approved):
Thanks!

PR #92040, opened 2023-09-14
[mono][wasm] Only catch the exception thrown by mono_llvm_cpp_throw_e…

vargaz:
…xception() in AOT-ed code.

Part of the fix for https://github.com/dotnet/runtime/issues/90692.

### DIFF
--- a/src/mono/mono/mini/mini-llvm.c
+++ b/src/mono/mono/mini/mini-llvm.c
@@ -5001,8 +5001,47 @@ emit_llvmonly_landing_pad (EmitContext *ctx, int group_index, int group_size)
 		LLVMPositionBuilderAtEnd (builder, catch_bb);
 
 		LLVMValueRef catchpad_args [1];
-		catchpad_args [0] = LLVMConstNull (pointer_type (LLVMInt8Type ()));
+		catchpad_args [0] = ctx->module->sentinel_exception;
 		catchpad = LLVMBuildCatchPad (builder, lpad, catchpad_args, 1, "");
+
+		// The ifdef is needed because of INTRINS_WASM
+#ifdef TARGET_WASM
+		// LLVM seems to require at least the following code sequence
+		// %7 = catchpad within %5 [ptr @typeinfo for int*]
+		// %8 = tail call ptr @llvm.wasm.get.exception(token %7)
+		// %9 = tail call i32 @llvm.wasm.get.ehselector(token %7)
+
+		LLVMValueRef args [] = { catchpad };
+		LLVMValueRef call = call_intrins (ctx, INTRINS_WASM_GET_EXCEPTION, args, "");
+		LLVMSetTailCall (call, TRUE);
+		LLVMValueRef sel_call = call_intrins (ctx, INTRINS_WASM_GET_EHSELECTOR, args, "");
+		LLVMSetTailCall (sel_call, TRUE);
+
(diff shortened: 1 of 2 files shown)


src/mono/mono/mini/mini-llvm.c
```diff
@@ -5001,8 +5001,48 @@ emit_llvmonly_landing_pad (EmitContext *ctx, int group_index, int group_size)
 		LLVMPositionBuilderAtEnd (builder, catch_bb);
 
 		LLVMValueRef catchpad_args [1];
-		catchpad_args [0] = LLVMConstNull (pointer_type (LLVMInt8Type ()));
+		catchpad_args [0] = ctx->module->sentinel_exception;
 		catchpad = LLVMBuildCatchPad (builder, lpad, catchpad_args, 1, "");
+
+#ifdef TARGET_WASM
+		LLVMValueRef args [16];
```
kg, line 5008: is it safe for slots 1-15 of this to be uninitialized?

kg:
When combined with https://github.com/dotnet/runtime/pull/91364, this PR fixes https://github.com/dotnet/runtime/issues/90692 for me.

kg (approved):
LGTM but I'm not an LLVM expert

src/mono/mono/mini/mini-llvm.c
```diff
@@ -5001,8 +5001,47 @@ emit_llvmonly_landing_pad (EmitContext *ctx, int group_index, int group_size)
...
+		// The ifdef is needed because of INTRINS_WASM
+#ifdef TARGET_WASM
+		// LLVM seems to require at least the following code sequence
+		// %7 = catchpad within %5 [ptr @typeinfo for int*]
+		// %8 = tail call ptr @llvm.wasm.get.exception(token %7)
+		// %9 = tail call i32 @llvm.wasm.get.ehselector(token %7)
+
+		LLVMValueRef args [] = { catchpad };
+		LLVMValueRef call = call_intrins (ctx, INTRINS_WASM_GET_EXCEPTION, args, "");
+		LLVMSetTailCall (call, TRUE);
```
SingleAccretion, line 5016: I am pretty sure there is no need for the `tail` annotation and Clang just adds it because that's what it does for most calls.

src/mono/mono/mini/mini-llvm.c
```diff
@@ -5001,8 +5001,47 @@ emit_llvmonly_landing_pad (EmitContext *ctx, int group_index, int group_size)
...
+		LLVMValueRef sel_call = call_intrins (ctx, INTRINS_WASM_GET_EHSELECTOR, args, "");
+		LLVMSetTailCall (sel_call, TRUE);
+
+		//%10 = tail call i32 @llvm.eh.typeid.for(ptr nonnull @typeinfo for int*) #5, !dbg !31
+		//%11 = icmp eq i32 %9, %10, !dbg !31
+		//br i1 %11, label %12, label %14, !dbg !31
+
+		LLVMValueRef typeid_args [] = { ctx->module->sentinel_exception };
+		LLVMValueRef typeid = call_intrins (ctx, INTRINS_EH_TYPEID_FOR, typeid_args, "");
+		LLVMSetTailCall (typeid, TRUE);
```
SingleAccretion, line 5026: What is the scenario served by the `int*` check? There is a size cost to using the C++ selector machinery.

It would support propagating (most) foreign C++ exceptions through managed code; is that the intention?

kg:
We may want to backport this to 8.0 staging. (I thought this fix was already in 8, but it looks like it's not?)

Issue #92022, opened 2023-09-13
Error in `ImmutableArray` initialization using C#12 collection literals


eduherminio:
### Description

`ImmutableArray`s are throwing an exception when being initialized at runtime in some circumstances:
- When debugging the project from Visual Studio
- When publishing the project (self-contained, trimmed, single-file in my case).

However, they work from CLI

This can be especially painful because tests pass CI but then everything collapses at runtime initialization. Real life CI example:
- This CI passes: https://github.com/lynx-chess/Lynx/actions/runs/6165990734
- The windows executable can be downloaded from there fails.
- Checking out the branch ([`immutable-arrays-cleaned`](https://github.com/lynx-chess/Lynx/tree/immutable-arrays-cleaned)) and running it locally has the same behavior: works OK from command line, fails from VS

Simplified example below:

### Reproduction Steps

```
dotnet new console -n immutable
```

Program.cs
```csharp
using System.Collections.Immutable;

Console.WriteLine($"Hello, World from {Constants.Coordinates[0]}");


public static class Constants
{
    public static readonly ImmutableArray<string> Coordinates =
    [
        "a8", "b8", "c8", "d8", "e8", "f8", "g8", "h8",
        "a7", "b7", "c7", "d7", "e7", "f7", "g7", "h7",
        "a6", "b6", "c6", "d6", "e6", "f6", "g6", "h6",
        "a5", "b5", "c5", "d5", "e5", "f5", "g5", "h5",
        "a4", "b4", "c4", "d4", "e4", "f4", "g4", "h4",
        "a3", "b3", "c3", "d3", "e3", "f3", "g3", "h3",
        "a2", "b2", "c2", "d2", "e2", "f2", "g2", "h2",
        "a1", "b1", "c1", "d1", "e1", "f1", "g1", "h1"
    ];
}

```

immutable.csproj
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <LangVersion>preview</LangVersion>
    <Nullable>enable</Nullable>
  </PropertyGroup>

</Project>

```

### Expected behavior

Runs normally, prints this in all the cases
```bash
Hello, World from a8
```

### Actual behavior

Failes from Visual Studio or when the project is published:

```
$> dotnet run -c Release
Hello, World from a8
$> dotnet run -c Debug
[...]

Issue #88684, opened 2023-07-11
`DynamicallyAccessedMembersAnalyzer` throws for `InlineArrayAccessOperation`


tannergooding:
As per the title, the `DynamicallyAccessedMembersAnalyzer` currently throws when encountering an `InlineArrayAccessOperation`:
```
Severity	Code	Description	Project	File	Line	Suppression State
Error	AD0001	Analyzer 'ILLink.RoslynAnalyzer.DynamicallyAccessedMembersAnalyzer' threw an exception of type 'System.NotImplementedException' with message 'Microsoft.CodeAnalysis.Operations.InlineArrayAccessOperation: D:\Users\tagoo\source\repos\terrafx.interop.windows\samples\DirectX\D3D12\HelloConstBuffer12.cs: (177,8)-(177,29)'.
Exception occurred with following context:
Compilation: TerraFX.Samples.DirectX
ISymbol: CreatePipelineState (Method)

System.NotImplementedException: Microsoft.CodeAnalysis.Operations.InlineArrayAccessOperation: D:\Users\tagoo\source\repos\terrafx.interop.windows\samples\DirectX\D3D12\HelloConstBuffer12.cs: (177,8)-(177,29)
   at ILLink.RoslynAnalyzer.DataFlow.LocalDataFlowVisitor`2.VisitSimpleAssignment(ISimpleAssignmentOperation operation, LocalDataFlowState`2 state)
   at ILLink.RoslynAnalyzer.DataFlow.OperationWalker`2.Visit(IOperation operation, TArgument argument)
   at ILLink.RoslynAnalyzer.TrimAnalysis.TrimAnalysisVisitor.Visit(IOperation operation, LocalDataFlowState`2 argument)
   at ILLink.RoslynAnalyzer.DataFlow.LocalDataFlowVisitor`2.VisitExpressionStatement(IExpressionStatementOperation operation, LocalDataFlowState`2 state)
   at ILLink.RoslynAnalyzer.DataFlow.OperationWalker`2.Visit(IOperation operation, TArgument argument)
   at ILLink.RoslynAnalyzer.TrimAnalysis.TrimAnalysisVisitor.Visit(IOperation operation, LocalDataFlowState`2 argument)
   at ILLink.RoslynAnalyzer.DataFlow.LocalDataFlowVisitor`2.Transfer(BlockProxy block, LocalDataFlowState`2 state)
[...]

PR #92060, opened 2023-09-14
Implement support for `InlineArray` in the trimmer

vitek-karas:
Types annotated with `InlineArray` need to preserve all of their fields, even if otherwise this would not be the case. It's possible to have a struct with `LayoutKind.Auto` and with `InlineArray` - trimmer would normally trim fields on such struct. This leads to corruption since the field is never accessed directly.

This changes the marking to preserve all fields on a type with `InlineArray` attribute just like we would for explicit layout struct and similar other types.

Adds tests to cover both the explicit usage of `InlineArray` attribute as well as the compiler generate usage via collection literals.

This should resolve the trimming part of https://github.com/dotnet/runtime/issues/92022
And it also adds tests for https://github.com/dotnet/runtime/issues/88684


agocke (approved):
LGTM, thanks!

sbomer (approved):
LGTM, thanks!

PR #92086, opened 2023-09-14
fix ReceiveFrom with dual mode socket

### LINKED ISSUE #92006: Incorrect port from Socket.ReceiveFromAsync(SocketAsyncEventArgs)
Regression from .net 7.

As part of managed implementation of QUIC, I have QuicSocketContext, which uses Socket.ReceiveFromAsync (for servers) and SocketReceiveAsync (for clients) to receive UDP datagrams. After rebasing the project onto latest main, server started receiving wrong port and thus ends up sending datagrams to itself, instead of to the clients.

Minimized repro code:

```cs
using System.Net.Http;
using System.Security.Authentication;
using System.Buffers;
using System.Diagnostics;
using System.Net.Sockets;
using System.Net;
using System.Threading;

var serverListeningAddress = new IPEndPoint(IPAddress.Loopback, 5000);

System.Console.WriteLine($"Starting server on {serverListeningAddress}");
var server = new QuicSocketContext(serverListeningAddress, null, true);
server.Start();
System.Console.WriteLine($"Server started on {server.LocalEndPoint}");

var client = new QuicSocketContext(null, server.LocalEndPoint, false);
client.Start();

System.Console.WriteLine($"===== Sending first datagram =====");
client.SendDatagram(new DatagramInfo(new byte[] { 1, 2, 3 }, 3, server.LocalEndPoint));

Thread.Sleep(-1);

internal readonly record struct DatagramInfo(byte[] Buffer, int Length, EndPoint RemoteEndpoint);

internal class QuicSocketContext
{
    private readonly EndPoint? _localEndPoint;
    private readonly EndPoint? _remoteEndPoint;
    private readonly bool _isServer;

    private bool _started;

    private readonly Socket _socket = new Socket(SocketType.Dgram, ProtocolType.Udp);
    private readonly SocketAsyncEventArgs _socketReceiveEventArgs = new SocketAsyncEventArgs();

    public QuicSocketContext(EndPoint? localEndPoint, EndPoint? remoteEndPoint, bool isServer)
    {
        _localEndPoint = localEndPoint;
        _remoteEndPoint = remoteEndPoint;

        _isServer = isServer;

        SetupSocket(localEndPoint, remoteEndPoint);

        _socketReceiveEventArgs.Completed += (sender, args) => OnReceiveFinished(args);
    }

    private void SetupSocket(EndPoint? localEndPoint, EndPoint? remoteEndPoint)
    {
        if (localEndPoint != null)
        {
            _socket.Bind(localEndPoint);
        }

        if (remoteEndPoint != null)
        {
            _socket.Connect(remoteEndPoint);
        }
    }

    public IPEndPoint LocalEndPoint => (IPEndPoint)_socket.LocalEndPoint!;

    public void Start()
    {
        if (_started)
        {
            return;
        }

        _started = true;

        var args = _socketReceiveEventArgs;
        while (!ReceiveFromAsync(args))
        {
            // this should not really happen, as the socket should be just opened, but we want to be sure and don't
            // miss any incoming datagrams
            OnDatagramReceived(ExtractDatagram(args));
        }
    }

    private DatagramInfo ExtractDatagram(SocketAsyncEventArgs args)
    {
        return new DatagramInfo(args.Buffer!, args.SocketError == SocketError.Success ? args.BytesTransferred : 0,
            args.RemoteEndPoint ?? _remoteEndPoint!);
    }

    private void OnReceiveFinished(SocketAsyncEventArgs args)
    {
        System.Console.WriteLine($"{(_isServer ? "Server" : "Client")}: OnReceiveFinished, RemoteEndPoint = {args.RemoteEndPoint}");
        bool pending;
        do
        {
            DatagramInfo datagram = ExtractDatagram(args);

            // immediately issue another async receive
            System.Console.WriteLine($"{(_isServer ? "Server" : "Client")}: Starting next receive");
            pending = !ReceiveFromAsync(args);

            OnDatagramReceived(datagram);
        } while (pending);
    }

    protected void OnDatagramReceived(in DatagramInfo datagram)
    {
        System.Console.WriteLine($"{(_isServer ? "Server" : "Client")}: processing datagram from {datagram.RemoteEndpoint}");
        Thread.Sleep(1000);

        SendDatagram(datagram);

        ArrayPool.Return(datagram.Buffer);
    }

    internal ArrayPool<byte> ArrayPool { get; } = ArrayPool<byte>.Shared;

    private bool ReceiveFromAsync(SocketAsyncEventArgs args)
    {
        // use fresh buffer for each receive, since the previous one is still being processed
        var buffer = ArrayPool.Rent(1200);
        args.SetBuffer(buffer, 0, buffer.Length);

        if (_remoteEndPoint != null)
        {
            System.Console.WriteLine($"Client: Socket.ReceiveAsync");
            return _socket.ReceiveAsync(args);
        }

        args.RemoteEndPoint = _localEndPoint!;
        System.Console.WriteLine($"Server: Socket.ReceiveFromAsync with RemotEndPoint = {_localEndPoint}");
        return _socket.ReceiveFromAsync(args);
    }

    internal void SendDatagram(in DatagramInfo datagram)
    {
        if (_remoteEndPoint != null)
        {
            // connected socket - > Socket.Send
            System.Console.WriteLine($"Client: Send over connected socket to {_remoteEndPoint}");
            _socket.Send(datagram.Buffer.AsSpan(0, datagram.Length), SocketFlags.None);
        }
        else
        {
            System.Console.WriteLine($"Server: SendTo {datagram.RemoteEndpoint}");
            _socket.SendTo(datagram.Buffer, 0, datagram.Length, SocketFlags.None, datagram.RemoteEndpoint);
        }
    }
}
```

.NET 7 output:
```
Starting server on 127.0.0.1:5000
Server: Socket.ReceiveFromAsync with RemotEndPoint = 127.0.0.1:5000
Server started on [::ffff:127.0.0.1]:5000
Client: Socket.ReceiveAsync
===== Sending first datagram =====
Client: Send over connected socket to [::ffff:127.0.0.1]:5000
Server: OnReceiveFinished, RemoteEndPoint = [::ffff:127.0.0.1]:57206
Server: Starting next receive
Server: Socket.ReceiveFromAsync with RemotEndPoint = 127.0.0.1:5000
Server: processing datagram from [::ffff:127.0.0.1]:57206
Server: SendTo [::ffff:127.0.0.1]:57206
Client: OnReceiveFinished, RemoteEndPoint =
Client: Starting next receive
Client: Socket.ReceiveAsync
Client: processing datagram from [::ffff:127.0.0.1]:5000
Client: Send over connected socket to [::ffff:127.0.0.1]:5000
Server: OnReceiveFinished, RemoteEndPoint = [::ffff:127.0.0.1]:57206
Server: Starting next receive
Server: Socket.ReceiveFromAsync with RemotEndPoint = 127.0.0.1:5000
Server: processing datagram from [::ffff:127.0.0.1]:57206
Server: SendTo [::ffff:127.0.0.1]:57206
Client: OnReceiveFinished, RemoteEndPoint =
Client: Starting next receive
Client: Socket.ReceiveAsync
Client: processing datagram from [::ffff:127.0.0.1]:5000
Client: Send over connected socket to [::ffff:127.0.0.1]:5000
Server: OnReceiveFinished, RemoteEndPoint = [::ffff:127.0.0.1]:57206
Server: Starting next receive
Server: Socket.ReceiveFromAsync with RemotEndPoint = 127.0.0.1:5000
Server: processing datagram from [::ffff:127.0.0.1]:57206
....
````

.NET 8 output, manual emphasis on first wrong line

```
Starting server on 127.0.0.1:5000
Server: Socket.ReceiveFromAsync with RemotEndPoint = 127.0.0.1:5000
Server started on [::ffff:127.0.0.1]:5000
Client: Socket.ReceiveAsync
===== Sending first datagram =====
Client: Send over connected socket to [::ffff:127.0.0.1]:5000
Server: OnReceiveFinished, RemoteEndPoint = [::ffff:127.0.0.1]:65505
Server: Starting next receive
Server: Socket.ReceiveFromAsync with RemotEndPoint = 127.0.0.1:5000
Server: processing datagram from [::ffff:127.0.0.1]:65505
Server: SendTo [::ffff:127.0.0.1]:65505
Client: OnReceiveFinished, RemoteEndPoint =
Client: Starting next receive
Client: Socket.ReceiveAsync
Client: processing datagram from [::ffff:127.0.0.1]:5000
Client: Send over connected socket to [::ffff:127.0.0.1]:5000
Server: OnReceiveFinished, RemoteEndPoint = 127.0.0.1:5000                // <---- here, server starts talking to itself because provided RemoteEndPoint is incorrect.
Server: Starting next receive
Server: Socket.ReceiveFromAsync with RemotEndPoint = 127.0.0.1:5000
Server: processing datagram from 127.0.0.1:5000
Server: SendTo 127.0.0.1:5000
Server: OnReceiveFinished, RemoteEndPoint = 127.0.0.1:5000
Server: Starting next receive
Server: Socket.ReceiveFromAsync with RemotEndPoint = 127.0.0.1:5000
Server: processing datagram from 127.0.0.1:5000
Server: SendTo 127.0.0.1:5000
Server: OnReceiveFinished, RemoteEndPoint = 127.0.0.1:5000
Server: Starting next receive
Server: Socket.ReceiveFromAsync with RemotEndPoint = 127.0.0.1:5000
Server: processing datagram from 127.0.0.1:5000
...
```

### PR DESCRIPTION
**Pull request by wfurt**
Fixes #92006

This is regression against 7.0
It seems like the old internal code handled cases when IPv4 `EndPoint` was passed to `ReceiveFrom` on dual mode socket and the received address is actually IPv4 mapped into IPv6. But that case fails now and we fail to set `RemoteEndPoint` leaving whatever was there - in case of the provided repro creating loop or possibly sending message to wrong peer. 

After thinking about it I see no reason why `IPEndPoint.Create` needs to really care about `AddressFamily` as long as it is IP. The documentation claims that it needs to be correct type but it does not talk about IPv4 vs IPv6. And we allocate new `EndPoint` so the actual value of the old one does not matter. It will also make it easier to consume new API from #88970 as we exposed `SocketAddresses` to consumers and they may need to eat with it. 

The regression was introduced by PR #89841

### DIFF
--- a/src/libraries/System.Net.Primitives/src/System/Net/IPEndPoint.cs
+++ b/src/libraries/System.Net.Primitives/src/System/Net/IPEndPoint.cs
@@ -155,8 +155,8 @@ namespace System.Net
         {
             ArgumentNullException.ThrowIfNull(socketAddress);
 
-            if (socketAddress.Family != AddressFamily.InterNetwork && socketAddress.Family != AddressFamily.InterNetworkV6)
-            {
+            if (socketAddress.Family is not (AddressFamily.InterNetwork or AddressFamily.InterNetworkV6))
+                {
                 throw new ArgumentException(SR.Format(SR.net_InvalidAddressFamily, socketAddress.Family.ToString(), GetType().FullName), nameof(socketAddress));
             }
 
--- a/src/libraries/System.Net.Primitives/src/System/Net/IPEndPoint.cs
+++ b/src/libraries/System.Net.Primitives/src/System/Net/IPEndPoint.cs
@@ -155,9 +155,9 @@ namespace System.Net
         {
             ArgumentNullException.ThrowIfNull(socketAddress);
 
-            if (socketAddress.Family != AddressFamily)
+            if (socketAddress.Family != AddressFamily.InterNetwork && socketAddress.Family != AddressFamily.InterNetworkV6)
             {
-                throw new ArgumentException(SR.Format(SR.net_InvalidAddressFamily, socketAddress.Family.ToString(), GetType().FullName, AddressFamily.ToString()), nameof(socketAddress));
+                throw new ArgumentException(SR.Format(SR.net_InvalidAddressFamily, socketAddress.Family.ToString(), GetType().FullName), nameof(socketAddress));
             }
 
             int minSize = AddressFamily == AddressFamily.InterNetworkV6 ? SocketAddress.IPv6AddressSize : SocketAddress.IPv4AddressSize;
--- a/src/libraries/System.Net.Primitives/tests/FunctionalTests/IPEndPointTest.cs
+++ b/src/libraries/System.Net.Primitives/tests/FunctionalTests/IPEndPointTest.cs
@@ -195,8 +195,7 @@ namespace System.Net.Primitives.Functional.Tests
 
         public static IEnumerable<object[]> Create_InvalidAddressFamily_TestData()
         {
-            yield return new object[] { new IPEndPoint(2, 500), new SocketAddress(Sockets.AddressFamily.InterNetworkV6) };
-            yield return new object[] { new IPEndPoint(IPAddress.Parse("192.169.0.9"), 500), new SocketAddress(Sockets.AddressFamily.InterNetworkV6) };
+            yield return new object[] { new IPEndPoint(2, 500), new SocketAddress(Sockets.AddressFamily.Unknown) };
             yield return new object[] { new IPEndPoint(IPAddress.Parse("0:0:0:0:0:0:0:1"), 500), new SocketAddress(Sockets.AddressFamily.InterNetwork) };
         }
 
--- a/src/libraries/System.Net.Primitives/tests/FunctionalTests/IPEndPointTest.cs
+++ b/src/libraries/System.Net.Primitives/tests/FunctionalTests/IPEndPointTest.cs
@@ -143,6 +143,19 @@ namespace System.Net.Primitives.Functional.Tests
             Assert.Equal(expected, endPoint.ToString());
         }
 
+        [Fact]
+        public static void Create_DifferentAF_Success()
+        {
+            SocketAddress sa = new SocketAddress(AddressFamily.InterNetwork, SocketAddress.GetMaximumAddressSize(AddressFamily.InterNetworkV6));
+            var ep = new IPEndPoint(IPAddress.IPv6Any, 0);
+            Assert.NotNull(ep.Create(sa));
+
+            sa = new SocketAddress(AddressFamily.InterNetworkV6);
+            ep = new IPEndPoint(IPAddress.Any, 0);
+
+            Assert.NotNull(ep.Create(sa));
+        }
+
         public static IEnumerable<object[]> Serialize_TestData()
         {
             yield return new object[] { new IPAddress(2), 16 };
--- a/src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketAsyncEventArgs.cs
+++ b/src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketAsyncEventArgs.cs
@@ -927,13 +927,13 @@ namespace System.Net.Sockets
                     {
                         try
                         {
-                            if (_remoteEndPoint!.AddressFamily == _socketAddress!.Family)
-                            {
-                                _remoteEndPoint = _remoteEndPoint!.Create(_socketAddress);
-                            }
-                            else if (_remoteEndPoint!.AddressFamily == AddressFamily.InterNetworkV6 && _socketAddress.Family == AddressFamily.InterNetwork)
+                            if (_remoteEndPoint!.AddressFamily == AddressFamily.InterNetworkV6 && _socketAddress!.Family == AddressFamily.InterNetwork)
                             {
                                 _remoteEndPoint = new IPEndPoint(_socketAddress.GetIPAddress().MapToIPv6(), _socketAddress.GetPort());
+                            }
+                            else
+                            {
+                                _remoteEndPoint = _remoteEndPoint!.Create(_socketAddress!);
                             }
                         }
                         catch
@@ -949,7 +949,14 @@ namespace System.Net.Sockets
                     {
                         try
                         {
-                            _remoteEndPoint = _remoteEndPoint!.Create(_socketAddress!);
+                            if (_remoteEndPoint!.AddressFamily == AddressFamily.InterNetworkV6 && _socketAddress!.Family == AddressFamily.InterNetwork)
+                            {
+                                _remoteEndPoint = new IPEndPoint(_socketAddress.GetIPAddress().MapToIPv6(), _socketAddress.GetPort());
+                            }
+                            else
+                            {
+                                _remoteEndPoint = _remoteEndPoint!.Create(_socketAddress!);
+                            }
                         }
                         catch
                         {
--- a/src/libraries/System.Net.Primitives/src/Resources/Strings.resx
+++ b/src/libraries/System.Net.Primitives/src/Resources/Strings.resx
@@ -64,7 +64,7 @@
     <value>This property is not implemented by this class.</value>
   </data>
   <data name="net_InvalidAddressFamily" xml:space="preserve">
-    <value>The AddressFamily {0} is not valid for the {1} end point, use {2} instead.</value>
+    <value>The AddressFamily {0} is not valid for the {1} end point.</value>
   </data>
   <data name="net_InvalidSocketAddressSize" xml:space="preserve">
     <value>The supplied {0} is an invalid size for the {1} end point.</value>
--- a/src/libraries/System.Net.Sockets/tests/FunctionalTests/ReceiveFrom.cs
+++ b/src/libraries/System.Net.Sockets/tests/FunctionalTests/ReceiveFrom.cs
@@ -171,6 +171,52 @@ namespace System.Net.Sockets.Tests
         [Theory]
         [InlineData(false)]
         [InlineData(true)]
+        public async Task ReceiveSent_DualMode_Success(bool ipv4)
+        {
+            const int Offset = 10;
+            const int DatagramSize = 256;
+            const int DatagramsToSend = 16;
+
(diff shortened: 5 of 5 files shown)

### CODE REVIEW

File: src/libraries/System.Net.Primitives/src/System/Net/IPEndPoint.cs [hunk 1]
```diff
@@ -155,9 +155,9 @@ public override EndPoint Create(SocketAddress socketAddress)
         {
             ArgumentNullException.ThrowIfNull(socketAddress);
 
-            if (socketAddress.Family != AddressFamily)
+            if (socketAddress.Family != AddressFamily.InterNetwork && socketAddress.Family != AddressFamily.InterNetworkV6)
```
stephentoub, line 158: Nit: can be simplified to:
```C#
if (socketAddress.Family is not AddressFamily.InterNetwork and not AddressFamily.InterNetworkV6)
```
or
```C#
if (socketAddress.Family is not (AddressFamily.InterNetwork or AddressFamily.InterNetworkV6)
```

src/libraries/System.Net.Primitives/tests/FunctionalTests/IPEndPointTest.cs
```diff
@@ -143,6 +143,19 @@ public static void ToString_Invoke_ReturnsExpected(IPEndPoint endPoint, string e
...
+        {
+            SocketAddress sa = new SocketAddress(AddressFamily.InterNetwork, SocketAddress.GetMaximumAddressSize(AddressFamily.InterNetworkV6));
+            var ep = new IPEndPoint(IPAddress.IPv6Any, 0);
+            Assert.NotNull(ep.Create(sa));
+
+            sa = new SocketAddress(AddressFamily.InterNetworkV6);
+            ep = new IPEndPoint(IPAddress.Any, 0);
+
+            Assert.NotNull(ep.Create(sa));
+        }
```
stephentoub, line 157: Do we have an exception test for when the SocketAddress is for something other than IPv4/6?

src/libraries/System.Net.Primitives/src/System/Net/IPEndPoint.cs
```diff
@@ -155,9 +155,9 @@ public override EndPoint Create(SocketAddress socketAddress)
...
             {
-                throw new ArgumentException(SR.Format(SR.net_InvalidAddressFamily, socketAddress.Family.ToString(), GetType().FullName, AddressFamily.ToString()), nameof(socketAddress));
+                throw new ArgumentException(SR.Format(SR.net_InvalidAddressFamily, socketAddress.Family.ToString(), GetType().FullName), nameof(socketAddress));
             }
 
             int minSize = AddressFamily == AddressFamily.InterNetworkV6 ? SocketAddress.IPv6AddressSize : SocketAddress.IPv4AddressSize;
```
stephentoub, line 163: If we're no longer requiring that socketAddress.Family == AddressFamily, does this size check still matter?

src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketAsyncEventArgs.cs
```diff
@@ -927,13 +927,13 @@ internal void FinishOperationSyncSuccess(int bytesTransferred, SocketFlags flags
                     {
                         try
                         {
-                            if (_remoteEndPoint!.AddressFamily == _socketAddress!.Family)
+                            if (_remoteEndPoint!.AddressFamily == AddressFamily.InterNetworkV6 && _socketAddress!.Family == AddressFamily.InterNetwork)
                             {
-                                _remoteEndPoint = _remoteEndPoint!.Create(_socketAddress);
+                                _remoteEndPoint = new IPEndPoint(_socketAddress.GetIPAddress().MapToIPv6(), _socketAddress.GetPort());
                             }
```
stephentoub, line 933: Is this something IPEndPoint.Create(SocketAddress) should be doing instead of doing it at this one call site?

src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketAsyncEventArgs.cs
```diff
@@ -949,7 +949,14 @@ internal void FinishOperationSyncSuccess(int bytesTransferred, SocketFlags flags
                     {
                         try
                         {
-                            _remoteEndPoint = _remoteEndPoint!.Create(_socketAddress!);
+                            if (_remoteEndPoint!.AddressFamily == AddressFamily.InterNetworkV6 && _socketAddress!.Family == AddressFamily.InterNetwork)
+                            {
+                                _remoteEndPoint = new IPEndPoint(_socketAddress.GetIPAddress().MapToIPv6(), _socketAddress.GetPort());
+                            }
+                            else
```
stephentoub, line 956: Similarly here: if this logic were moved into IPEndPoint.Create, would that a) address the duplication between these two call sites and b) potentially avoid bugs elsewhere?  Or would that introduce problems?

src/libraries/System.Net.Primitives/src/System/Net/IPEndPoint.cs
wfurt: We don't need to but than we can possibly get exception when we access the underlying buffer and it is too small. So I feel the extra check provides better feedback.

src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketAsyncEventArgs.cs
wfurt: this one is tricky - we have cases when the address is IPv4 and `Create` would just work - but we want IPv6 IPv4Mapped addresses for DualMode sockets.  I guess mostly for compatibility reasons so this is handled as special case.

src/libraries/System.Net.Primitives/tests/FunctionalTests/IPEndPointTest.cs
wfurt: yes, that should be covered by `Create_InvalidAddressFamily_ThrowsArgumentException`

stephentoub (approved):

src/libraries/System.Net.Primitives/src/System/Net/IPEndPoint.cs
wfurt: updated.

PR #84620, opened 2023-04-11
Avoid warnings due to RUC on type when marking assembly as library

vitek-karas:
Marking an assembly as "library" will mark all public APIs. If some of the marked members have RUC on them we decided we don't want to produce warnings, since there's nothing especially wrong with the library. The callers of those members would still get the right warnings.

Currently linker implements this correctly for methods and properties, but it fails for fields and events and it still produces some warnings, especially if the RUC is on the parent type.

This change fixes those cases and adds a special test for the library marked assembly in combination with several RUCs.

Fixes https://github.com/dotnet/runtime/issues/81864

Issue #83581, opened 2023-03-17
ILLink reports duplicate warnings from properties in some cases


vitek-karas:
```C#
RequireAllProperties(typeof(TestType));

void RequireAllProperties([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicProperties)] Type type) { }

[RequiresUnreferencedCode("attr")]
class AttributeWhichRequires : Attribute
{
}

class TestType
{
    [AttributeWhichRequires]
    public int Property // IL2026
    {
        get;
        set;
    }
}
```

The above code will produce two identical warnings `IL2026` from the analyzer, and three identical warnings from illink.

Ideally it should only produce one warning from each tool.
The reason this is not the case in the illink is that we don't avoid duplicate property marking (via the `IsProcessed` hashtables), unlike we do for methods and types.
The analyzer is a similar reason, but there we don't go through hashtables anywhere, so we might need to come up with some better mechanism.

In the case of illink the three warnings come from:
* DAM marking of the property via the `RequireAllProperties` call - this will mark the property for reflection, which will in turn mark all of its accessors
* Marking of the getter will also mark the property -> warning
* Marking of the setter will also mark the property -> warning

PR #92094, opened 2023-09-14
Avoid marking property/event attributes multiple times

### LINKED ISSUE #92064: dotnet publish command fails with error NETSDK1144: Optimizing assemblies for size failed
I'm running the following command to build maui ios project

`dotnet publish .ProjectName.csproj -c Staging -f net7.0-ios -p:ApplicationDisplayVersion=3.0.0 -p:ApplicationVersion=3.0.0.123 -p:ArchiveOnBuild=true -p:RuntimeIdentifier=ios-arm64 -p:CodesignKey="Apple Distribution: <title is here>" -p:MtouchLink=SdkOnly`

It fails with 

` /usr/local/share/dotnet/sdk/7.0.306/Sdks/Microsoft.NET.ILLink.Tasks/build/Microsoft.NET.ILLink.targets(86,5): error NETSDK1144: Optimizing assemblies for size failed. Optimization can be disabled by setting the PublishTrimmed property to false`

I'm not able to provide a detailed log because the project is under NDA, but can give any other details regarding the libraries usage, etc.

sbomer:
This tracks whether properties have been marked yet, and avoids marking them multiple times. Fixes a recursion issue where marking a property can cause marking of attributes which may cause marking of the same property.

I removed the `AlreadyMarked` assert because https://github.com/dotnet/runtime/pull/84620 propagates the `Kind` from methods to properties - so it can look as if a property is being marked from a descriptor the first time it is encountered.

Fixes https://github.com/dotnet/runtime/issues/92064

Fixes the illink part of https://github.com/dotnet/runtime/issues/83581. That issue mentions a problem with the analyzer behavior, but the testcase I'm fixing in this PR doesn't seem to show duplicate warnings from the analyzer. Can we consider https://github.com/dotnet/runtime/issues/83581 fixed by this change?

### DIFF
--- a/src/tools/illink/src/linker/Linker.Steps/MarkStep.cs
+++ b/src/tools/illink/src/linker/Linker.Steps/MarkStep.cs
@@ -3537,7 +3537,8 @@ namespace Mono.Linker.Steps
 
 		protected internal void MarkProperty (PropertyDefinition prop, in DependencyInfo reason)
 		{
-			Tracer.AddDirectDependency (prop, reason, marked: false);
+			if (!Annotations.MarkProcessed (prop, reason))
+				return;
 
 			using var propertyScope = ScopeStack.PushScope (new MessageOrigin (prop));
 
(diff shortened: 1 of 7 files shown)


vitek-karas (approved):

vitek-karas (approved):
Nice

PR #92095, opened 2023-09-14
[wasi] Add AOT compiler build to the official build 

steveisok:


vargaz (approved):

lewing (commented):
...

lewing (commented):
...

lewing:
@directhex I think the remaining issue here might be that the system python we're picking up on windows is 32bit and the libclang is 64bit? Which python is the windows build without emsdk using?

directhex:
> @directhex I think the remaining issue here might be that the system python we're picking up on windows is 32bit and the libclang is 64bit? Which python is the windows build without emsdk using?

This analysis seems valid. It's p/invoking (or whatever it's called on Python) from D:\a\_work\1\s\artifacts\obj\mono\wasi.wasm.Release\llvm\x64\bin\libclang.dll and https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#language-and-runtime says it comes with Python 3.9 (which has no x64 Windows builds)

lewing:
Could we use the emsdk python package directly for this step?

radical (approved):

lewing:
@microsoft-github-policy-service rerun

PR #92099, opened 2023-09-15
[wasi] Generate scripts to run with node.

vargaz:


vargaz:
The advantage of this is that node is running on v8 allowing the testing of AOT etc.

radical (approved):

PR #92102, opened 2023-09-15
[RISC-V] Atomics

tomeksowi:
Add atomic instructions (RV64A) to JIT.
Intrinsify calls to Interlocked.ExchangeAdd|Exchange|And|Or|CompareExchange.

Part of #84834

### DIFF
--- a/src/coreclr/jit/codegenriscv64.cpp
+++ b/src/coreclr/jit/codegenriscv64.cpp
@@ -2609,14 +2609,40 @@ void CodeGen::genJumpTable(GenTree* treeNode)
 }
 
 //------------------------------------------------------------------------
-// genLockedInstructions: Generate code for a GT_XADD or GT_XCHG node.
+// genLockedInstructions: Generate code for a GT_XADD, GT_XAND, GT_XORR or GT_XCHG node.
 //
 // Arguments:
-//    treeNode - the GT_XADD/XCHG node
+//    treeNode - the GT_XADD/XAND/XORR/XCHG node
 //
 void CodeGen::genLockedInstructions(GenTreeOp* treeNode)
 {
-    NYI_RISCV64("genLockedInstructions-----unimplemented/unused on RISCV64 yet----");
+    GenTree* data = treeNode->AsOp()->gtOp2;
+    GenTree* addr = treeNode->AsOp()->gtOp1;
+    regNumber dataReg = data->GetRegNum();
+    regNumber addrReg = addr->GetRegNum();
+    regNumber targetReg = treeNode->GetRegNum();
+    if (targetReg == REG_NA)
+    {
+        targetReg = REG_R0;
+    }
+
+    genConsumeAddress(addr);
+    genConsumeRegs(data);
+
+    emitAttr dataSize = emitActualTypeSize(data);
+    bool is4 = (dataSize == EA_4BYTE);
+
+    assert(!data->isContainedIntOrIImmed());
+
+    genTreeOps op = treeNode->gtOper;
+    instruction ins =
+        op == GT_XORR ? (is4 ? INS_amoor_w   : INS_amoor_d) :
+        op == GT_XAND ? (is4 ? INS_amoand_w  : INS_amoand_d) :
+        op == GT_XCHG ? (is4 ? INS_amoswap_w : INS_amoswap_d) :
+        op == GT_XADD ? (is4 ? INS_amoadd_w  : INS_amoadd_d) :
+        (assert(!"Unexpected treeNode->gtOper"), INS_none);
+
+    GetEmitter()->emitIns_R_R_R(ins, dataSize, targetReg, addrReg, dataReg);
 }
 
 //------------------------------------------------------------------------
@@ -4643,6 +4669,8 @@ void CodeGen::genCodeForTreeNode(GenTree* treeNode)
 
         case GT_XCHG:
         case GT_XADD:
+        case GT_XORR:
+        case GT_XAND:
             genLockedInstructions(treeNode->AsOp());
             break;
 
--- a/src/coreclr/jit/codegenriscv64.cpp
+++ b/src/coreclr/jit/codegenriscv64.cpp
@@ -2616,10 +2616,10 @@ void CodeGen::genJumpTable(GenTree* treeNode)
 //
 void CodeGen::genLockedInstructions(GenTreeOp* treeNode)
 {
-    GenTree* data = treeNode->AsOp()->gtOp2;
-    GenTree* addr = treeNode->AsOp()->gtOp1;
-    regNumber dataReg = data->GetRegNum();
-    regNumber addrReg = addr->GetRegNum();
+    GenTree*  data      = treeNode->AsOp()->gtOp2;
+    GenTree*  addr      = treeNode->AsOp()->gtOp1;
+    regNumber dataReg   = data->GetRegNum();
+    regNumber addrReg   = addr->GetRegNum();
     regNumber targetReg = treeNode->GetRegNum();
     if (targetReg == REG_NA)
     {
@@ -2630,18 +2630,19 @@ void CodeGen::genLockedInstructions(GenTreeOp* treeNode)
     genConsumeRegs(data);
 
     emitAttr dataSize = emitActualTypeSize(data);
-    bool is4 = (dataSize == EA_4BYTE);
+    bool     is4      = (dataSize == EA_4BYTE);
 
     assert(!data->isContainedIntOrIImmed());
 
     genTreeOps op = treeNode->gtOper;
+    // clang-format off
     instruction ins =
         op == GT_XORR ? (is4 ? INS_amoor_w   : INS_amoor_d) :
         op == GT_XAND ? (is4 ? INS_amoand_w  : INS_amoand_d) :
         op == GT_XCHG ? (is4 ? INS_amoswap_w : INS_amoswap_d) :
         op == GT_XADD ? (is4 ? INS_amoadd_w  : INS_amoadd_d) :
         (assert(!"Unexpected treeNode->gtOper"), INS_none);
-
+    // clang-format on
     GetEmitter()->emitIns_R_R_R(ins, dataSize, targetReg, addrReg, dataReg);
 }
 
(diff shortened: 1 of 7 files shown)


tomeksowi:
@jakobbotsch @jkotas PR ready for review.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -700,7 +700,7 @@ void emitter::emitIns_R_R_R(
     if ((INS_add <= ins && ins <= INS_and) || (INS_mul <= ins && ins <= INS_remuw) ||
         (INS_addw <= ins && ins <= INS_sraw) || (INS_fadd_s <= ins && ins <= INS_fmax_s) ||
         (INS_fadd_d <= ins && ins <= INS_fmax_d) || (INS_feq_s <= ins && ins <= INS_fle_s) ||
-        (INS_feq_d <= ins && ins <= INS_fle_d))
+        (INS_feq_d <= ins && ins <= INS_fle_d) || (INS_lr_w <= ins && ins <= INS_amomaxu_d))
```
t-mustafin, line 703: Seems lesser atomic instruction is `INS_amoadd_w`, not `INS_lr_w`.

tomeksowi: `INS_lr_w` comes first. `INS_*` is an enum, not the instruction itself.

t-mustafin (approved):

jkotas:
cc @dotnet/jit-contrib

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -2609,14 +2609,41 @@ void CodeGen::genJumpTable(GenTree* treeNode)
...
-    NYI_RISCV64("genLockedInstructions-----unimplemented/unused on RISCV64 yet----");
+    GenTree*  data      = treeNode->AsOp()->gtOp2;
+    GenTree*  addr      = treeNode->AsOp()->gtOp1;
+    regNumber dataReg   = data->GetRegNum();
+    regNumber addrReg   = addr->GetRegNum();
+    regNumber targetReg = treeNode->GetRegNum();
+    if (targetReg == REG_NA)
+    {
+        targetReg = REG_R0;
+    }
```
jakobbotsch, line 2627: When is `targetReg == REG_NA`?

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -2609,14 +2609,41 @@ void CodeGen::genJumpTable(GenTree* treeNode)
...
+    assert(!data->isContainedIntOrIImmed());
+
+    genTreeOps op = treeNode->gtOper;
+    // clang-format off
+    instruction ins =
+        op == GT_XORR ? (is4 ? INS_amoor_w   : INS_amoor_d) :
+        op == GT_XAND ? (is4 ? INS_amoand_w  : INS_amoand_d) :
+        op == GT_XCHG ? (is4 ? INS_amoswap_w : INS_amoswap_d) :
+        op == GT_XADD ? (is4 ? INS_amoadd_w  : INS_amoadd_d) :
+        (assert(!"Unexpected treeNode->gtOper"), INS_none);
```
jakobbotsch, line 2644: Write this as a switch? We generally avoid comma expressions in the JIT sources.

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -2609,14 +2609,41 @@ void CodeGen::genJumpTable(GenTree* treeNode)
...
+    // clang-format on
+    GetEmitter()->emitIns_R_R_R(ins, dataSize, targetReg, addrReg, dataReg);
 }
```
jakobbotsch, line 2647: Looks like this function forgets to call `genProduceReg` anywhere.

src/coreclr/jit/lsrariscv64.cpp
```diff
@@ -368,17 +368,48 @@ int LinearScan::BuildNode(GenTree* tree)
...
+            assert(!"-----unimplemented on RISCV64----");
+            break;
+
         case GT_XORR:
         case GT_XAND:
         case GT_XADD:
         case GT_XCHG:
         {
-            NYI_RISCV64("-----unimplemented on RISCV64 yet----");
+            assert(dstCount == (tree->TypeGet() == TYP_VOID) ? 0 : 1);
```
jakobbotsch, line 398:
```suggestion
            assert(dstCount == (tree->TypeIs(TYP_VOID)) ? 0 : 1);
```

src/coreclr/jit/codegenriscv64.cpp
jakobbotsch: Nevermind, I can see that LSRA build does not build a def for the `TYP_VOID` case.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3659,6 +3684,50 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
+                (assert(!"Illegal funct5 within atomic memory operation, emitDisInsName"), "?");
+            // clang-format on
+
+            unsigned int funct3 = (code >> 12) & 0x7;
+            // clang-format off
+            char width =
+                funct3 == 0b010 ? 'w' :
+                funct3 == 0b011 ? 'd' :
+                (assert(!"Illegal width tag within atomic memory operation, emitDisInsName"), '?');
+            // clang-format on
```
jakobbotsch, line 3712: Ditto here -- write a switch instead.

src/coreclr/jit/instrsriscv64.h
```diff
@@ -241,6 +241,31 @@ INST(fcvt_d_l,      "fcvt.d.l",       0,   0xd2200053)
 INST(fcvt_d_lu,     "fcvt.d.lu",      0,   0xd2300053)
 INST(fmv_d_x,       "fmv.d.x",        0,   0xf2000053)
 
+// RV32A + RV64A (R-type, R_R_R)
+// For now all of them are seqentially consistent because the Interlocked.* APIs don't expose acquire/release ordering
+INST(lr_w,          "lr.w",           0,   (0b00010 << 27) | (0b11 << 25) | (0b010 << 12) | 0b0101111)
```
clamp03, line 246: Could you write hex instruction code like all other instructions and write detail as comment ?
`0b11` is variable part, please write in `emitIns_R_R_R`.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3659,6 +3684,50 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
+            {
+                return;
+            }
+            assert(len <= 12);
+
+            const char* dest = RegNames[(code >> 7) & 0x1f];
+            const char* addr = RegNames[(code >> 15) & 0x1f];
+            const char* data = RegNames[(code >> 20) & 0x1f];
+
+            printf("%*s %s, %s, %s\n", 12 - len, "", dest, addr, data);
```
clamp03, line 3728: Please print risc-v instructions not yours. (*. https://github.com/llvm/llvm-project/blob/main/llvm/test/MC/RISCV/rv64a-valid.s)
If then, instruction size will exceed 12 length. Please update spaces in other instructions as well as atomic instructions.

src/coreclr/jit/lsrariscv64.cpp
```diff
@@ -368,17 +368,48 @@ int LinearScan::BuildNode(GenTree* tree)
...
+            GenTree* addr = tree->gtGetOp1();
+            GenTree* data = tree->gtGetOp2();
+            assert(!addr->isContained() && !data->isContained());
+            srcCount = 2;
+
+            buildInternalIntRegisterDefForNode(tree);
```
clamp03, line 404: Does it need internal register?

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3659,6 +3684,50 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
             }
             return;
         }
+        case 0b0101111: // AMO - atomic memory operation
```
clamp03, line 3687: Could you update to hex?

clamp03 (changes requested):
Could you share tests to fix?
Please check the style and naming of existing sources? Please~~~

src/coreclr/jit/codegenriscv64.cpp
tomeksowi: I wrote it with `?:` to save space, 3x as few LOCs vs. switch-case while retaining readability. But I don't have a strong opinion, it can be a switch if that's the style around here.

src/coreclr/jit/emitriscv64.cpp
tomeksowi: ok

src/coreclr/jit/instrsriscv64.h
tomeksowi: I did it to be consistent with the RISC-V ISA spec, they use binary for instructions everywhere and no hex. But I don't have a strong opinion, can be hex + comment.

As for the ordering flags, sure, I can OR them in emitIns_R_R_R.

src/coreclr/jit/codegenriscv64.cpp
jakobbotsch: Yes, please do. Nested ternaries are discouraged in the JIT coding conventions (https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/clr-jit-coding-conventions.md#1519-ternary-operators).

src/coreclr/jit/codegenriscv64.cpp
jakobbotsch: Right.

src/coreclr/jit/emitriscv64.cpp
tomeksowi: OK, I wanted to avoid changing padding everywhere, but I can do that and print `.aqrl` like LLVM.

src/coreclr/jit/lsrariscv64.cpp
tomeksowi: I don't think so, good catch.

tomeksowi:
> Could you share tests to fix? Please check the style and naming of existing sources? Please~~~

I stumbled on lack of atomics in JIT looking for `NYI_RISCV64` assertions. I constructed a testing playlist for it by grepping for `Interlocked.*` in C# tests.

Yes, I'll adjust style as per remarks in comments.

jakobbotsch:
> I stumbled on lack of atomics in JIT looking for `NYI_RISCV64` assertions. I constructed a testing playlist for it by grepping for `Interlocked.*` in C# tests:

To be clear, this is just an optimization, right? Not recognizing these as intrinsics should usually just result in the managed fallback, so these tests aren't broken on RISC-V.

tomeksowi:
> > I stumbled on lack of atomics in JIT looking for `NYI_RISCV64` assertions. I constructed a testing playlist for it by grepping for `Interlocked.*` in C# tests:
> 
> To be clear, this is just an optimization, right? Not recognizing these as intrinsics should usually just result in the managed fallback, so these tests aren't broken on RISC-V.

Exactly.

src/coreclr/jit/emitriscv64.cpp
tomeksowi: Sure, I'll leave the funct5 binary though because it doesn't fall on the nibble boundary and would be difficult to eye-match with hex codes in instrsriscv64.h.

clamp03 (approved):
Thank you so much.
I know my initial codes are not good enough. I am so sorry for that. But that is the reason why we need you. Thank you for all your hard work.

jakobbotsch (approved):

src/coreclr/jit/lsrariscv64.cpp
```diff
@@ -368,17 +368,45 @@ int LinearScan::BuildNode(GenTree* tree)
...
+            assert(!"-----unimplemented on RISCV64----");
+            break;
+
         case GT_XORR:
         case GT_XAND:
         case GT_XADD:
         case GT_XCHG:
         {
-            NYI_RISCV64("-----unimplemented on RISCV64 yet----");
+            assert(dstCount == tree->TypeIs(TYP_VOID) ? 0 : 1);
```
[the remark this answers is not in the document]
tomeksowi: No, no, it was always passing:) I fixed also in arm64 where I copied it from.

JulieLeeMSFT:
Please check CI build failures before merging.

jakobbotsch:
SPMI failures are the out-of-space ones we have been hitting recently.

Issue #91862, opened 2023-09-11
Assertion failed '!node->OperIs(GT_CALL) || (node->AsCall()->gtCallType == CT_HELPER)' during 'Optimize index checks'


kunalspathak:
```c#

using System;
using System.Collections.Generic;
using System.Runtime.CompilerServices;
using System.Runtime.Intrinsics;
using System.Runtime.Intrinsics.X86;
using System.Numerics;
public class TestClass
{
    public struct S1
    {
        public struct S1_D1_F3
        {
        }
    }
    public struct S2
    {
    }
    static byte s_byte_5 = 127;
    static Vector128<ushort> s_v128_ushort_21 = Vector128.Create((ushort)2, 2, 5, 32767, 2, 5, 54, 5);
    static Vector128<int> s_v128_int_22 = Vector128<int>.Zero;
    static S1.S1_D1_F3 s_s1_s1_d1_f3_51 = new S1.S1_D1_F3();
    static S1 s_s1_52 = new S1();
    static S2 s_s2_53 = new S2();
    S1.S1_D1_F3 s1_s1_d1_f3_101 = new S1.S1_D1_F3();
    S1 s1_102 = new S1();
    S2 s2_103 = new S2();
    public S1 Method54(out S1 p_s1_1625, out S1.S1_D1_F3 p_s1_s1_d1_f3_1626, ref S1.S1_D1_F3 p_s1_s1_d1_f3_1627, S1 p_s1_1628, S1 p_s1_1629, ref S1 p_s1_1630, ref S1.S1_D1_F3 p_s1_s1_d1_f3_1631, out S2 p_s2_1632, out S2 p_s2_1633)
    {
        unchecked
        {
            byte byte_1635 = 2;
            int int_1640 = 2;
            ushort ushort_1645 = 54;
            if (15<=4)
            {
            }
            else
            {
                ushort_1645 += Vector128.GetElement(Sse2.ShiftRightLogical128BitLane(s_v128_ushort_21 = s_v128_ushort_21, s_byte_5 = s_byte_5), Sse41.Extract(s_v128_int_22, byte_1635) + (int_1640 *= 15-4)& 5);
            }
            return s_s1_52;
        }
    }
    public void Method0()
    {
        unchecked
        {
            S1.S1_D1_F3 s1_s1_d1_f3_2935 = new S1.S1_D1_F3();
            S1 s1_2936 = new S1();
[...]

PR #92112, opened 2023-09-15
JIT: Add some convenience classes for node count JIT metrics, and automatic dumping from compShutdown

jakobbotsch:
- Add a NodeCounts class that can be used to easily record the count of node types seen
- Add a DumpOnShutdown class that can be used to easily register either a Histogram class or NodeCounts class to be dumped as part of Compiler::compShutdown

For example, I was investigating #91862 and wanted to know what types of nodes we usually see in the index/length positions of `BOUNDS_CHECK` inside `optRemoveRangeCheck`. I made the following change:
```diff
--- a/src/coreclr/jit/optimizer.cpp
+++ b/src/coreclr/jit/optimizer.cpp
@@ -8849,6 +8849,11 @@ void Compiler::AddModifiedElemTypeAllContainingLoops(unsigned lnum, CORINFO_CLAS
     }
 }
 
+NodeCounts     s_rangeCheckLengthNodes;
+NodeCounts     s_rangeCheckIndexNodes;
+DumpOnShutdown d1("BOUNDS_CHECK length nodes", &s_rangeCheckLengthNodes);
+DumpOnShutdown d2("BOUNDS_CHECK index nodes", &s_rangeCheckIndexNodes);
+
 //------------------------------------------------------------------------------
 // optRemoveRangeCheck : Given an indexing node, mark it as not needing a range check.
 //
@@ -8878,6 +8883,9 @@ GenTree* Compiler::optRemoveRangeCheck(GenTreeBoundsChk* check, GenTree* comma,
                  (check != nullptr && check->OperIs(GT_BOUNDS_CHECK) && comma == nullptr));
     noway_assert(check->OperIs(GT_BOUNDS_CHECK));
 
+    s_rangeCheckIndexNodes.record(check->GetIndex()->OperGet());
+    s_rangeCheckLengthNodes.record(check->GetArrayLength()->OperGet());
+
     GenTree* tree = comma != nullptr ? comma : check;
 
 #ifdef DEBUG
```

After that, a superpmi.exe replay over libraries_tests.run outputs the following when it finishes:
```scala
Loaded 674679  Jitted 674679  FailedCompile 0 Excluded 0 Missing 0
Total time: 565472.612700ms
BOUNDS_CHECK length nodes
ARR_LENGTH           :   51990
LCL_VAR              :   47159
CNS_INT              :    6751
COMMA                :       9
LCL_FLD              :       5
CAST                 :       4

BOUNDS_CHECK index nodes
LCL_VAR              :   62221
CNS_INT              :   42139
COMMA                :     623
CAST                 :     460
ADD                  :     397
RSH                  :      72
NEG                  :       5
UDIV                 :       1
```


jakobbotsch:

src/coreclr/jit/utils.cpp
```diff
@@ -999,6 +1000,90 @@ void Histogram::record(unsigned size)
...
+{
+    for (DumpOnShutdownEntry& entry : s_dumpOnShutdown)
+    {
+        if ((entry.Name == nullptr) && (entry.Dumpable == nullptr))
+        {
+            entry.Name     = name;
+            entry.Dumpable = dumpable;
+            break;
+        }
+    }
```
BruceForstall, line 1067: Maybe add an assert that the DumpOnShutdownEntry table isn't all full?

BruceForstall (approved):
Looks nice. Would be useful to have some comments for the classes showing how to use them.

src/coreclr/jit/utils.cpp
```diff
@@ -999,6 +1000,90 @@ void Histogram::record(unsigned size)
...
+        }
+
+        fprintf(output, "%-20s : %7u\n", GenTree::OpName(entry.oper), entry.count);
+    }
+}
+
+void NodeCounts::record(genTreeOps oper)
+{
+    assert(oper < GT_COUNT);
+    m_counts[oper]++;
```
AndyAyersMS, line 1046: Should this do an interlocked add?

BruceForstall: Maybe that would depend on whether this class is instantiated as global, or as a member of Compiler?

src/coreclr/jit/utils.cpp
jakobbotsch: Will do.

src/coreclr/jit/utils.cpp
jakobbotsch: Yeah, for good measure it should be interlocked add. `NodeCounts` is meant to be used as a static field (like `Histogram`) so you can collect stats for multiple compilations, so it can be racing with other threads, although I expect we mainly use it through SPMI where it won't due to the way parallelization works there.

Looks like `Histogram` should also be changed to use an interlocked add.

PR #90430, opened 2023-08-11
[NativeAOT] Do not use private APIs on iOS/macOS

filipnavara:
Fixes #90408.

Tested with upload of empty iOS app through TestFlight process.

Issue #88737, opened 2023-07-12
NativeAOT library mode on iOS-like platforms


filipnavara:
As documented in https://github.com/royalapplications/beyondnet/issues/34#issuecomment-1632510996 there's a demand to get NativeAOT library mode working on iOS-like platforms. This is specifically for `net8.0` TFM and NOT the Xamarin `net8.0-ios` TFM which may run into similar issues but requires fixes in the Xamarin SDK. It currently requires a couple of workaround and we should look into fixing the underlying issues:

- Linking step doesn't use correct `-sysroot`/`-target`. We currently bundle the apps through AppleAppBuilder in dotnet/runtime, and a custom bundler in Xamarin. However, for dylib builds the NativeAOT linker step on its own is mostly sufficient. We just need to specify proper `-target` (eg. `apple-arm64-ios11.0` or similar) and sysroot. The sysroot can be obtained through `xcrun`, or we can simply execute the whole linker using `xcrun -sdk iphoneos clang <options>` command to achieve the same result.
- There's a conflict with the new `mobile-librarybuilder-experimental` workload which currently needs to be disabled with `_IsAppleMobileLibraryMode=false`. We should disable it automatically for NativeAOT publishing.

shana:
+10 for not having to hardcode the sysroot path to get this to work. Everything else in the build system seems to be working great (as seen in https://github.com/godotengine/godot/pull/82729), this is the only stumbling block that's making it slightly awkward.

shana:
@MichalStrehovsky BTW, looks like the workaround mentioned in #91997 is only for the `-lc++` linker argument, and not for the `-isysroot` path, so there's no current (and easy) workarounds for this.

PR #92114, opened 2023-09-15
[NativeAOT] Fix iOS library build by linking standard C++ library by default

### LINKED ISSUE #91997: .NET 8 RC1 fails to NativeAOT compile when targeting ios-arm64
### Description

Up until .NET RC1, NativeAOT compilation worked out of the box when targeting `ios-arm64`.
This requires adding `<PublishAotUsingRuntimePack>true</PublishAotUsingRuntimePack>` and `<LinkerArg Include="-isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk" />` to the csproj.
Everything else is pretty much standard fare.

Now, since RC1, `dotnet publish` fails because of linking errors in `libicuuc.a` related to C++.

As a workaround, one can add `<LinkerArg Include="-lc++" />` to the csproj like in the example below.

### Reproduction Steps

Running `dotnet publish` against the following example csproj works fine because `-lc++` was re-added using a `LinkerArg` flag. Removing this line makes the build fail.

```xml
  <PropertyGroup>
    <RuntimeIdentifier>ios-arm64</RuntimeIdentifier>
    <PublishAOT>true</PublishAOT>

    <!-- TODO: Temporary workaround for iOS/iOS Simulator support -->
    <PublishAotUsingRuntimePack>true</PublishAotUsingRuntimePack>
  </PropertyGroup>

  <ItemGroup>
    <!-- TODO: Temporary workaround for iOS support -->
    <LinkerArg Include="-isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk" />

    <!-- TODO: Temporary workaround for .NET 8 RC1 which seems to have removed -lc++ from its linker flags -->
    <LinkerArg Include="-lc++" />
  </ItemGroup>
</Project>
```

### Expected behavior

Compiles out of the box.

### Actual behavior

Does not compile without adding `<LinkerArg Include="-lc++" />`.

### Regression?

It did work in all .NET 8 previews until RC1.

### Known Workarounds

Adding `<LinkerArg Include="-lc++" />` to the csproj.

### Configuration

.NET 8.0.100-rc.1.23455.8

### Other information

_No response_

ivanpovazan:
This PR fixes building an iOS-like NativeAOT libraries by linking standard C++ library by default.

Recently we started statically linking ICU libraries when targeting iOS-like platforms in: https://github.com/dotnet/runtime/pull/90430
However globalization and ICU support has a dependency on the standard C++ lib which is not included by NativeAOT build integration targets and produces a build error when trying to build NativeAOT iOS library:
```
Undefined symbols for architecture arm64:
    "std::__1::__call_once(unsigned long volatile&, void*, void (*)(void*))", referenced from:
        void std::__1::call_once<void (&)()>(std::__1::once_flag&, void (&)()) in libicuuc.a(umutex.ao)
        ...
```

This is not visible when build iOS apps, as we use `AppleAppBuilder` and `Xamarin` application bundlers that are taking care of native linking. 

NativeAOT with iOS in library mode is not officially supported (this is tracked in: https://github.com/dotnet/runtime/issues/88737), but this fix is a prerequisite of going in the right direction and unblocking customers who are using the experimental support.  The official support for iOS library mode will come in a separate PR.
 
---
Fixes: https://github.com/dotnet/runtime/issues/91997

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
@@ -50,6 +50,9 @@ The .NET Foundation licenses this file to you under the MIT license.
 
       <EventPipeName>libeventpipe-disabled</EventPipeName>
       <EventPipeName Condition="'$(EventSourceSupport)' == 'true'">libeventpipe-enabled</EventPipeName>
+
+      <LinkStandardCPlusPlusLibrary Condition="'$(LinkStandardCPlusPlusLibrary)' == '' and '$(NativeLib)' != '' and '$(_IsiOSLikePlatform)' == 'true' and '$(InvariantGlobalization)' != 'true'">true</LinkStandardCPlusPlusLibrary>
+      <LinkStandardCPlusPlusLibrary Condition="'$(LinkStandardCPlusPlusLibrary)' == ''">false</LinkStandardCPlusPlusLibrary>
     </PropertyGroup>
 
     <ItemGroup>
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
@@ -51,8 +51,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <EventPipeName>libeventpipe-disabled</EventPipeName>
       <EventPipeName Condition="'$(EventSourceSupport)' == 'true'">libeventpipe-enabled</EventPipeName>
 
-      <LinkStandardCPlusPlusLibrary Condition="'$(LinkStandardCPlusPlusLibrary)' == '' and '$(NativeLib)' != '' and '$(_IsiOSLikePlatform)' == 'true' and '$(InvariantGlobalization)' != 'true'">true</LinkStandardCPlusPlusLibrary>
-      <LinkStandardCPlusPlusLibrary Condition="'$(LinkStandardCPlusPlusLibrary)' == ''">false</LinkStandardCPlusPlusLibrary>
+      <LinkStandardCPlusPlusLibrary Condition="'$(LinkStandardCPlusPlusLibrary)' == '' and '$(_IsiOSLikePlatform)' == 'true' and '$(InvariantGlobalization)' != 'true'">true</LinkStandardCPlusPlusLibrary>
     </PropertyGroup>
 
     <ItemGroup>


rolfbjarne (approved):

steveisok (approved):
@ivanpovazan @rolfbjarne this doesn't clash with anything in the iOS sdk, right?

ivanpovazan:
> @ivanpovazan @rolfbjarne this doesn't clash with anything in the iOS sdk, right?

I do not think it should. I manually tested building/running a MAUI iOS sample app by explicitly passing `-p:LinkStandardCPlusPlusLibrary=true` with the RC1 release and did not have any issues.

rolfbjarne:
> @ivanpovazan @rolfbjarne this doesn't clash with anything in the iOS sdk, right?

It shouldn't, at most we'd pass the same linker argument twice to the native linker.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
```diff
@@ -50,6 +50,9 @@ The .NET Foundation licenses this file to you under the MIT license.
 
       <EventPipeName>libeventpipe-disabled</EventPipeName>
       <EventPipeName Condition="'$(EventSourceSupport)' == 'true'">libeventpipe-enabled</EventPipeName>
+
+      <LinkStandardCPlusPlusLibrary Condition="'$(LinkStandardCPlusPlusLibrary)' == '' and '$(NativeLib)' != '' and '$(_IsiOSLikePlatform)' == 'true' and '$(InvariantGlobalization)' != 'true'">true</LinkStandardCPlusPlusLibrary>
+      <LinkStandardCPlusPlusLibrary Condition="'$(LinkStandardCPlusPlusLibrary)' == ''">false</LinkStandardCPlusPlusLibrary>
```
jkotas, line 55:
```suggestion
```
Nit: Unnecessary.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
```diff
@@ -50,6 +50,9 @@ The .NET Foundation licenses this file to you under the MIT license.
 
       <EventPipeName>libeventpipe-disabled</EventPipeName>
       <EventPipeName Condition="'$(EventSourceSupport)' == 'true'">libeventpipe-enabled</EventPipeName>
+
+      <LinkStandardCPlusPlusLibrary Condition="'$(LinkStandardCPlusPlusLibrary)' == '' and '$(NativeLib)' != '' and '$(_IsiOSLikePlatform)' == 'true' and '$(InvariantGlobalization)' != 'true'">true</LinkStandardCPlusPlusLibrary>
```
jkotas, line 54: Is `'$(NativeLib)' != '' and` necessary and logically correct here? 

Xamarin link targets may be adding the c++ library reference today, but it is not right for runtime to depend on it. The runtime build targets should be coherent.

ivanpovazan: > Is '$(NativeLib)' != '' and necessary and logically correct here?

It is not need and I will remove it, thank you.

> Xamarin link targets may be adding the c++ library reference today, but it is not right for runtime to depend on it. The runtime build targets should be coherent.

From my understanding the Xamarin build uses `clang++` as the native compiler/linker and does not explicitly include the c++ library.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
ivanpovazan: Fixed.

jkotas (approved):
Thanks

ivanpovazan:
The failures are unrelated and are being tracked.

Issue #84248, opened 2023-04-03
Redundant nullcheck in ARR_LEN/IND on top of non-null array


EgorBo:
```csharp
void Foo()
{
    var arr = new int[10];
    var len = arr.Length;
}
```
Codegen:
```asm
; Method Program:Foo():this
G_M52879_IG01:              
       4883EC28             sub      rsp, 40
G_M52879_IG02:              
       48B928E4DFFCF87F0000 mov      rcx, 0x7FF8FCDFE428      ; int[]
       BA0A000000           mov      edx, 10
       E8384A865F           call     CORINFO_HELP_NEWARR_1_VC
       8B4008               mov      eax, dword ptr [rax+08H]  ;;  <----
G_M52879_IG03:              
       4883C428             add      rsp, 40
       C3                   ret      
; Total bytes of code: 32
```
and other patterns where JIT should be able to detect that ARR_LEN/IND is invoked on something known to be non-null, e.g.:
```csharp
void Foo(int[] arr)
{
    if (arr != null)
    {
        var _ = arr.Length;
    }
}
```

JulieLeeMSFT:
Pushing to Future.

PR #92116, opened 2023-09-15
JIT: Extract all side effects of the index in optRemoveRangeCheck

### LINKED ISSUE #91862: Assertion failed '!node->OperIs(GT_CALL) || (node->AsCall()->gtCallType == CT_HELPER)' during 'Optimize index checks'

jakobbotsch:
optRemoveRangeCheck extracts only GTF_ASG from the bounds check. If the BOUNDS_CHECK is complex, that results in silently dropping side effects on the floor (see the example case).

The ideal fix is that we should always extract all side effects from the index and length operands, however this has large regressions because the length typically has an ARR_LENGTH that we then extract. This PR instead has a surgical fix for the problem case that can be backported to .NET 8. It extracts all side effects from the index, but keeps extracting only GTF_ASG from the length to get around the issue mentioned above.

Fix #91862


EgorBo:
> The ideal fix is that we should always extract all side effects from the index and length operands

https://github.com/dotnet/runtime/issues/84248 should help to mitigate a lot of regressions in that case (tried locally a while ago)

jakobbotsch:

Minor [diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=409052&view=ms.vss-build-web.run-extensions-tab).

EgorBo (approved):

PR #92123, opened 2023-09-15
JIT: Initialize jitstdout lazily

### LINKED ISSUE #91856: Crash when trying to load an Assembly in a hostfxr environment after calling FreeConsole()
### Description

Hi
I'm currently experiencing a crash in one of my applications which uses the hostfxr environment to load .NET Assemblies. The application is generally a GUI application, however we use the CONSOLE subsystem to have support for a command line interface. When run from the regular explorer, the application detects this and calls `FreeConsole()` to hide the console window.

This works fine for the most part, however we noticed that `FreeConsole()` apparently leaves the stdout, stderr and stdin handles in an invalid state which later on causes a crash [here](https://github.com/dotnet/runtime/blob/de0ab156194eb64deae0e1018db9a58f7b02f4a3/src/coreclr/jit/ee_il_dll.cpp#L92) when the JIT tries to duplicate the stdout handle. The code there seems to do some checks already but doesn't seem to catch this problem.

### Reproduction Steps

The latest nightly release of ImHex triggers this issue at launch on Windows 10 sometimes. It's not consistent but it happens often. 
It can be downloaded from https://imhex.download/#nightly

Alternatively, a simple application which uses the CONSOLE subsystem, calls `FreeConsole()` and then loads an Assembly through the hostfxr functions should reproduce the issue on Windows 10.

### Expected behavior

`stdout` should simply not be used in this case instead of causing a crash

### Actual behavior

The call to `_dup()` at https://github.com/dotnet/runtime/blob/de0ab156194eb64deae0e1018db9a58f7b02f4a3/src/coreclr/jit/ee_il_dll.cpp#L92 causes a abort to happen.

### Regression?

_No response_

### Known Workarounds

Interestingly enough, this issue seems to only affect Windows 10 (and potentially even older versions of Windows). It does not happen on Windows 11.

### Configuration

- .NET 7.0.400
- Windows 10
- x64
- It's specific to Windows 10 (and potentially before), not Windows 11

### Other information

- My application: 
https://github.com/WerWolv/ImHex
- The code that calls `FreeConsole()`: https://github.com/WerWolv/ImHex/blob/master/main/source/window/win_window.cpp#L238-L251
- The code that initializes the .NET runtime and loads an Assembly: https://github.com/WerWolv/ImHex/blob/master/plugins/script_loader/source/loaders/dotnet/dotnet_loader.cpp#L156
- Stacktrace:
image

jakobbotsch:
Avoid duplicating a handle and doing several I/O operations on the startup path. Fixes #91856 as a side effect.

### DIFF
--- a/src/coreclr/jit/ee_il_dll.cpp
+++ b/src/coreclr/jit/ee_il_dll.cpp
@@ -31,8 +31,6 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
 /*****************************************************************************/
 
-FILE* jitstdout = nullptr;
-
 ICorJitHost* g_jitHost        = nullptr;
 bool         g_jitInitialized = false;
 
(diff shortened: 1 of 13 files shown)


src/coreclr/jit/ee_il_dll.cpp
```diff
@@ -89,40 +96,58 @@ extern "C" DLLEXPORT void jitStartup(ICorJitHost* jitHost)
...
+
+        return observed;
+    }
+
+    return file;
+}
+
+FILE* jitstdout()
+{
+    if (s_jitstdout != nullptr)
```
jkotas, line 137: `s_jitstdout` needs to be volatile for this to work reliably. Otherwise, the C++ compiler can compile this as:

```
var temp = s_jitstdout;
if (s_jitstdout != null)
    return temp;
...
```

jakobbotsch: Ah, I see. Addressed

jakobbotsch:
cc @dotnet/jit-contrib PTAL @EgorBo. Am going to try to see if this can also make .NET 8.

Verified that the code no longer runs during startup/shutdown in release builds (unless e.g. `DOTNET_JitDisasm` is set), and manually tested a few stdout/JitStdOutFile scenarios on windows/ubuntu.

EgorBo (approved):

Issue #85181, opened 2023-04-21
F1 on our source generator errors goes nowhere because we don't set a HelpUri


danmoseley:
I got this
![image](https://user-images.githubusercontent.com/6385855/233733776-f9553850-f54c-4c05-9821-b79bab3efa32.png)

clicked on the SYSLIB1052 code, which is hyperlinked, and got an Oops page:

https://learn.microsoft.com/en-us/visualstudio/ide/not-in-toc/default?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(keywords_removed)%26rd%3Dtrue&view=vs-2022

Not sure if it's just this code.

teo-tsirpanis:
BTW clicking many newer compiler error codes also leads to a generic error page like this.

danmoseley:
@buyaa-n how do we handle ensuring that the SYSLIBxxx codes go someplace?

danmoseley:
should ours all pass [HelpLinkUri](https://github.com/dotnet/roslyn/blob/74f99e6cad8d288ff406f6faefd26faad7144442/src/Compilers/Core/Portable/Diagnostic/DiagnosticDescriptor.cs#L38) or should there be useful default behavior somewhere?

buyaa-n:
> how do we handle ensuring that the SYSLIBxxx codes go someplace?

I don't know how SYSLIBxxx links works, seems it should open https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib1050-1069, tagging @gewarren who created that page

gewarren:
None of the SYSLIB warning links seem to work, even the ones that have a dedicated page (I tried SYSLIB0011 for BinaryFormatter.Serialize). I think this is how the code analysis rules hook up to learn.microsoft.com: https://github.com/dotnet/roslyn-analyzers/blob/2420507acd5863bfe30a7dd97993c9362527045b/src/Utilities/Compiler/DiagnosticDescriptorHelper.cs#L36. @mavasani might be able to suggest a solution for the SYSLIB warnings?

mavasani:
> should ours all pass [HelpLinkUri](https://github.com/dotnet/roslyn/blob/74f99e6cad8d288ff406f6faefd26faad7144442/src/Compilers/Core/Portable/Diagnostic/DiagnosticDescriptor.cs#L38) or should there be useful default behavior somewhere?

[...]

PR #92134, opened 2023-09-15
Fix URL for syslib diagnostics

ericstj:
I was testing the fix for https://github.com/dotnet/runtime/issues/85181 and found that the links were off for syslib-diagnostics.  We didn't need to append the `.md`.  This will fix these.

@gewarren are you OK with this approach - or do you think it would be better to create fwlinks / aka.ms links?  Those could potentially handle changes to the target URL or could have a better default landing page if the diagnostic page didn't exist.  Not sure if it's worth the overhead though.  I guess the compiler itself isn't using FWLinks and that's good enough for them -- should be good for us too.

This needs backport to 8.0 as well.


gewarren (approved):
I would leave it like this, since it's what the compiler team does.

PR #92138, opened 2023-09-15
Add missing STJ xml docs

eiriktsarpalis:
Companion PR to https://github.com/dotnet/dotnet-api-docs/pull/9297

### DIFF
--- a/src/libraries/System.Text.Json/src/CompatibilitySuppressions.xml
+++ b/src/libraries/System.Text.Json/src/CompatibilitySuppressions.xml
@@ -58,54 +58,6 @@
     <IsBaselineSuppression>true</IsBaselineSuppression>
   </Suppression>
   <Suppression>
-    <DiagnosticId>CP0002</DiagnosticId>
-    <Target>M:System.Text.Json.Serialization.JsonObjectCreationHandlingAttribute.#ctor</Target>
-    <Left>ref/net462/System.Text.Json.dll</Left>
-    <Right>lib/net462/System.Text.Json.dll</Right>
-  </Suppression>
-  <Suppression>
-    <DiagnosticId>CP0002</DiagnosticId>
-    <Target>M:System.Text.Json.Serialization.JsonUnmappedMemberHandlingAttribute.#ctor</Target>
-    <Left>ref/net462/System.Text.Json.dll</Left>
-    <Right>lib/net462/System.Text.Json.dll</Right>
-  </Suppression>
-  <Suppression>
-    <DiagnosticId>CP0002</DiagnosticId>
-    <Target>M:System.Text.Json.Serialization.JsonObjectCreationHandlingAttribute.#ctor</Target>
-    <Left>ref/net8.0/System.Text.Json.dll</Left>
-    <Right>lib/net8.0/System.Text.Json.dll</Right>
-  </Suppression>
-  <Suppression>
-    <DiagnosticId>CP0002</DiagnosticId>
-    <Target>M:System.Text.Json.Serialization.JsonUnmappedMemberHandlingAttribute.#ctor</Target>
-    <Left>ref/net8.0/System.Text.Json.dll</Left>
-    <Right>lib/net8.0/System.Text.Json.dll</Right>
-  </Suppression>
-  <Suppression>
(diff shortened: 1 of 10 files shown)


src/libraries/System.Text.Json/src/CompatibilitySuppressions.xml
```diff
@@ -57,6 +57,54 @@
     <Right>lib/netstandard2.0/System.Text.Json.dll</Right>
     <IsBaselineSuppression>true</IsBaselineSuppression>
   </Suppression>
+  <Suppression>
```
eiriktsarpalis, line 60: Explicitly adding default constructors tripped off the API compat checker necessitating these suppressions. FYI @carlossanlop @ViktorHofer

src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonArray.cs
```diff
@@ -113,9 +113,10 @@ internal int GetElementIndex(JsonNode? node)
         }
 
         /// <summary>
-        /// Returns enumerator that wraps calls to <see cref="JsonNode.GetValue{T}"/>.
+        /// Returns an enumerable that wraps calls to <see cref="JsonNode.GetValue{T}"/>.
```
carlossanlop, line 116: This change didn't get reflected in the dotnet-api-docs PR, FYI. APIs with existing docs don't get updated by the porting tool.

carlossanlop (approved):
LGTM. Left a node, and also there were some suggestions in the dotnet-api-docs PR in case you want to modify them here too.

src/libraries/System.Text.Json/src/CompatibilitySuppressions.xml
ViktorHofer: That needs to be investigated. @eiriktsarpalis do you still have the error message available? I wonder if APICompat complains about the member being on the ref but not on the src or the other way around.

cc @ericstj as I'm out the next two weeks

ViktorHofer: Actually, I just looked at the reference source and the type defines an explicit constructor with parameters, so adding a default one is an API addition that you probably don't want to make. Or do you?

ViktorHofer (changes requested):
Marking as no-merge until we resolve the public API discussion

src/libraries/System.Text.Json/src/CompatibilitySuppressions.xml
ericstj: Yeah these are all ref/lib comparisons.  

> Explicitly adding default constructors

This isn't a case of explicit vs implicit.  Implicit default constructors [only exist if no other constructors are defined](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/instance-constructors#parameterless-constructors). In these cases you're adding a default constructor to types that didn't have an implicit one.  This a real error. 
image

If the API docs tooling is telling you this constructor exists -- it is mistaken.  Don't add this API.

ViktorHofer: > Yeah these are all ref/lib comparisons.

That's correct. I was asking about the error message as it provides information about if the error applies to the right or the left side. As this comparison is done in strict mode, the CP0002 error could apply to either missing API on the left or on the right side. Unfortunately, the XML doesn't provide that detail.

That said, in this case it's clear that this is about an API addition on the right side.

eiriktsarpalis: Thanks. The API docs tooling is actually correct -- it was complaining about a missing parameter that I mistook for a missing default constructor (similar to a few other types in the undoc list). I just reverted the unnecessary constructors and suppressions.

src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonArray.cs
eiriktsarpalis: Fixed.

ViktorHofer (approved):

PR #92142, opened 2023-09-15
6 more naive methods for Tensor Primitives.

michaelgsharp:
Added in naive implementations for
* CosineSimilarty
* Distance
* Dot
* Normalize
* SoftMax
* Sigmoid

@tannergooding @stephentoub @jeffhandley should we rename normalize as discussed since its not really the operation we are doing?

### DIFF
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.ComponentModel;
-
 namespace System.Numerics.Tensors
 {
     /// <summary>Performs primitive tensor operations over spans of memory.</summary>
@@ -260,22 +258,28 @@ namespace System.Numerics.Tensors
         /// <param name="x">The first tensor, represented as a span.</param>
         /// <param name="y">The second tensor, represented as a span.</param>
         /// <returns>The cosine similarity between the two vectors.</returns>
+        /// <exception cref="ArgumentException">Length of '<paramref name="x" />' must be same as length of '<paramref name="y" />'.</exception>
+        /// <exception cref="ArgumentException">'<paramref name="x" />' and '<paramref name="y" />' must not be empty.</exception>
         public static float CosineSimilarity(ReadOnlySpan<float> x, ReadOnlySpan<float> y)
         {
             if (x.Length != y.Length)
             {
                 ThrowHelper.ThrowArgument_SpansMustHaveSameLength();
             }
+            if (x.Length == 0 || y.Length == 0)
+            {
+                ThrowHelper.ThrowArgument_SpansMustBeNonEmpty();
+            }
 
-            var dotprod = 0f;
-            var magx = 0f;
-            var magy = 0f;
+            float dotprod = 0f;
+            float magx = 0f;
+            float magy = 0f;
 
             for (int i = 0; i < x.Length; i++)
             {
                 dotprod += x[i] * y[i];
-                magx += MathF.Pow(x[i], 2);
-                magy += MathF.Pow(y[i], 2);
+                magx += x[i] * x[i];
+                magy += y[i] * y[i];
             }
 
             return dotprod / (MathF.Sqrt(magx) * MathF.Sqrt(magy));
@@ -287,18 +291,25 @@ namespace System.Numerics.Tensors
         /// <param name="x">The first tensor, represented as a span.</param>
         /// <param name="y">The second tensor, represented as a span.</param>
         /// <returns>The Euclidean distance.</returns>
+        /// <exception cref="ArgumentException">Length of '<paramref name="x" />' must be same as length of '<paramref name="y" />'.</exception>
+        /// <exception cref="ArgumentException">'<paramref name="x" />' and '<paramref name="y" />' must not be empty.</exception>
         public static float Distance(ReadOnlySpan<float> x, ReadOnlySpan<float> y)
         {
             if (x.Length != y.Length)
             {
                 ThrowHelper.ThrowArgument_SpansMustHaveSameLength();
             }
+            if (x.Length == 0 || y.Length == 0)
+            {
+                ThrowHelper.ThrowArgument_SpansMustBeNonEmpty();
+            }
 
-            var distance = 0f;
+            float distance = 0f;
 
             for (int i = 0; i < x.Length; i++)
             {
-                distance += MathF.Pow(x[i] - y[i], 2);
+                float dist = x[i] - y[i];
+                distance += dist * dist;
             }
 
             return MathF.Sqrt(distance);
@@ -310,6 +321,7 @@ namespace System.Numerics.Tensors
         /// <param name="x">The first tensor, represented as a span.</param>
         /// <param name="y">The second tensor, represented as a span.</param>
         /// <returns>The dot product.</returns>
+        /// <exception cref="ArgumentException">Length of '<paramref name="x" />' must be same as length of '<paramref name="y" />'.</exception>
         public static float Dot(ReadOnlySpan<float> x, ReadOnlySpan<float> y) // BLAS1: dot
         {
             if (x.Length != y.Length)
@@ -317,7 +329,7 @@ namespace System.Numerics.Tensors
                 ThrowHelper.ThrowArgument_SpansMustHaveSameLength();
             }
 
-            var dotprod = 0f;
+            float dotprod = 0f;
 
             for (int i = 0; i < x.Length; i++)
             {
@@ -332,13 +344,13 @@ namespace System.Numerics.Tensors
         /// </summary>
         /// <param name="x">The first tensor, represented as a span.</param>
         /// <returns>The L2 norm.</returns>
-        public static float Normalize(ReadOnlySpan<float> x) // BLAS1: nrm2
+        public static float L2Normalize(ReadOnlySpan<float> x) // BLAS1: nrm2
         {
-            var magx = 0f;
+            float magx = 0f;
 
             for (int i = 0; i < x.Length; i++)
             {
-                magx += MathF.Pow(x[i], 2);
+                magx += x[i] * x[i];
             }
 
             return MathF.Sqrt(magx);
@@ -349,21 +361,27 @@ namespace System.Numerics.Tensors
         /// </summary>
         /// <param name="x">The first tensor, represented as a span.</param>
         /// <param name="destination">The destination tensor.</param>
+        /// <exception cref="ArgumentException">Destination is too short.</exception>
+        /// <exception cref="ArgumentException">'<paramref name="x" />' must not be empty.</exception>
         public static void SoftMax(ReadOnlySpan<float> x, Span<float> destination)
         {
             if (x.Length > destination.Length)
             {
                 ThrowHelper.ThrowArgument_DestinationTooShort();
             }
+            if (x.Length == 0)
+            {
+                ThrowHelper.ThrowArgument_SpansMustBeNonEmpty();
+            }
 
-            var expSum = 0f;
+            float expSum = 0f;
 
             for (int i = 0; i < x.Length; i++)
             {
                 expSum += MathF.Pow((float)Math.E, x[i]);
             }
 
-            for (int i = 0; i < destination.Length; i++)
+            for (int i = 0; i < x.Length; i++)
             {
                 destination[i] = MathF.Exp(x[i]) / expSum;
             }
@@ -374,11 +392,17 @@ namespace System.Numerics.Tensors
         /// </summary>
         /// <param name="x">The first tensor, represented as a span.</param>
         /// <param name="destination">The destination tensor.</param>
+        /// <exception cref="ArgumentException">Destination is too short.</exception>
+        /// <exception cref="ArgumentException">'<paramref name="x" />' must not be empty.</exception>
         public static void Sigmoid(ReadOnlySpan<float> x, Span<float> destination)
         {
             if (x.Length > destination.Length)
             {
                 ThrowHelper.ThrowArgument_DestinationTooShort();
+            }
+            if (x.Length == 0)
+            {
+                ThrowHelper.ThrowArgument_SpansMustBeNonEmpty();
             }
 
             for (int i = 0; i < x.Length; i++)
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
@@ -1,6 +1,8 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.ComponentModel;
+
 namespace System.Numerics.Tensors
 {
     /// <summary>Performs primitive tensor operations over spans of memory.</summary>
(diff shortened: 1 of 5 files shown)


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +255,136 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+        /// <param name="y">The second tensor, represented as a span.</param>
+        /// <returns>The cosine similarity between the two vectors.</returns>
+        public static float CosineSimilarity(ReadOnlySpan<float> x, ReadOnlySpan<float> y)
+        {
+            if (x.Length != y.Length)
+            {
+                ThrowHelper.ThrowArgument_SpansMustHaveSameLength();
+            }
+
+            var dotprod = 0f;
```
gfoidl, line 270: Don't use `var` in runtime-repo. Use the concrete type here instead.
Same on other places.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +255,136 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+            var magx = 0f;
+            var magy = 0f;
+
+            for (int i = 0; i < x.Length; i++)
+            {
+                dotprod += x[i] * y[i];
+                magx += MathF.Pow(x[i], 2);
+                magy += MathF.Pow(y[i], 2);
```
gfoidl, line 278:
```suggestion
                magx += x[i] * x[i];
                magy += y[i] * y[i];
```
? No need to call pow.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +255,136 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+            if (x.Length != y.Length)
+            {
+                ThrowHelper.ThrowArgument_SpansMustHaveSameLength();
+            }
+
+            var distance = 0f;
+
+            for (int i = 0; i < x.Length; i++)
+            {
+                distance += MathF.Pow(x[i] - y[i], 2);
```
gfoidl, line 301:
```suggestion
                float dist = x[i] - y[i];
                distance += dist * dist
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +255,136 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+        /// </summary>
+        /// <param name="x">The first tensor, represented as a span.</param>
+        /// <returns>The L2 norm.</returns>
+        public static float Normalize(ReadOnlySpan<float> x) // BLAS1: nrm2
+        {
+            var magx = 0f;
+
+            for (int i = 0; i < x.Length; i++)
+            {
+                magx += MathF.Pow(x[i], 2);
```
gfoidl, line 341:
```suggestion
                magx += x[i] * x[i];
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.ComponentModel;
```
stephentoub, line 4: What is this needed for?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +255,136 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+            }
+
+            var expSum = 0f;
+
+            for (int i = 0; i < x.Length; i++)
+            {
+                expSum += MathF.Pow((float)Math.E, x[i]);
+            }
+
+            for (int i = 0; i < destination.Length; i++)
```
stephentoub, line 366: This doesn't look right. Did you intend for this to be x.Length? Assuming it's wrong, that also suggests a test hole.

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
```diff
@@ -702,5 +702,123 @@ public static void TanhTensor_ThrowsForTooShortDestination(int tensorLength)
...
+        {
+            float[] x = CreateAndFillTensor(tensorLength);
+            float[] y = CreateAndFillTensor(tensorLength - 1);
+
+            Assert.Throws<ArgumentException>(() => TensorPrimitives.CosineSimilarity(x, y));
+        }
+
+        [Theory]
+        [InlineData(new float[] { 3, 2, 0, 5 }, new float[] { 1, 0, 0, 0 }, 0.49f)]
+        [InlineData(new float[] { 1, 1, 1, 1, 1, 0 }, new float[] { 1, 1, 1, 1, 0, 1 }, 0.80f)]
```
stephentoub, line 718: These tests will be insufficient for vectorized implementations as they're not large enough to trigger paths for some vector widths. It'd also be good to validate empty inputs.

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
```diff
@@ -702,5 +702,123 @@ public static void TanhTensor_ThrowsForTooShortDestination(int tensorLength)
...
+        }
+
+        [Theory]
+        [InlineData(new float[] { 3, 1, .2f }, new float[] { 0.8360188f, 0.11314284f, 0.05083836f })]
+        [InlineData(new float[] { 3, 4, 1 }, new float[] { 0.2594f, 0.7052f, 0.0351f })]
+        [InlineData(new float[] { 5, 3 }, new float[] { 0.8807f, 0.1192f })]
+        [InlineData(new float[] { 4, 2, 1, 9 }, new float[] { 0.0066f, 9.04658e-4f, 3.32805e-4f, 0.9920f})]
+        public static void SoftMax(float[] x, float[] expectedResult)
+        {
+            var dest = new float[x.Length];
```
stephentoub, line 790: It'd be good to add tests where the destination is longer the input. I expect if you do that, you'll find it fails with the current implementation.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +255,136 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
                 destination[i] = MathF.Tanh(x[i]);
             }
         }
+
+        /// <summary>Computes the cosine similarity between two non-zero vectors.</summary>
+        /// <param name="x">The first tensor, represented as a span.</param>
+        /// <param name="y">The second tensor, represented as a span.</param>
+        /// <returns>The cosine similarity between the two vectors.</returns>
```
stephentoub, line 262: We should document the ArgumentException that might be thrown (here and in the other methods).

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +255,136 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+
+            return dotprod;
+        }
+
+        /// <summary>
+        /// A mathematical operation that takes a vector and returns the L2 norm.
+        /// </summary>
+        /// <param name="x">The first tensor, represented as a span.</param>
+        /// <returns>The L2 norm.</returns>
+        public static float Normalize(ReadOnlySpan<float> x) // BLAS1: nrm2
```
stephentoub, line 335: L2Normalize? NormalizeL2?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
michaelgsharp: Shouldn't be needed, was probably added automatically when I did a quick fix or something. Removed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +255,136 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+            }
+
+            var dotprod = 0f;
+
+            for (int i = 0; i < x.Length; i++)
+            {
+                dotprod += x[i] * y[i];
+            }
+
+            return dotprod;
```
stephentoub, line 327: This will become a one-liner on top of the Aggregate helper in https://github.com/dotnet/runtime/pull/92154

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
michaelgsharp: Added the exception to all the docs.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +255,136 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+            }
+
+            var distance = 0f;
+
+            for (int i = 0; i < x.Length; i++)
+            {
+                distance += MathF.Pow(x[i] - y[i], 2);
+            }
+
+            return MathF.Sqrt(distance);
```
stephentoub, line 304: This will become a one-liner on top of the Aggregate helper in https://github.com/dotnet/runtime/pull/92154

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
michaelgsharp: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
michaelgsharp: Fixed

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
michaelgsharp: Fixed

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
michaelgsharp: Changed it to L2Normalize.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
michaelgsharp: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
michaelgsharp: Yeah, I meant it to be x and it was a whole, and I don't have any tests where the dest is longer than the input. Let me add some.

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
michaelgsharp: Yeah, I know they will be insufficient for the vectorized tests, but since none of these are yet I wasn't too worried. Would you like me to fix that now though or when we get the vectorized implementations in? Happy either way.

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
michaelgsharp: Yeah, it would have failed. Added in those tests and fixed it.

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
stephentoub: Let's get it merged and then we can add more.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +255,136 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+        public static float Normalize(ReadOnlySpan<float> x) // BLAS1: nrm2
+        {
+            var magx = 0f;
+
+            for (int i = 0; i < x.Length; i++)
+            {
+                magx += MathF.Pow(x[i], 2);
+            }
+
+            return MathF.Sqrt(magx);
```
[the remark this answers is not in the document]
michaelgsharp: Would you like me to wait till that PR is merged to make that change? or just do it after? Same question for the comment above.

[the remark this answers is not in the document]
stephentoub: I'll include it in that PR once this merges.

ericstj (approved):
LGTM

stephentoub (approved):
I'm going to merge this and then fix up a few things after rebasing my PR on it.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +253,162 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+            float magy = 0f;
+
+            for (int i = 0; i < x.Length; i++)
+            {
+                dotprod += x[i] * y[i];
+                magx += x[i] * x[i];
+                magy += y[i] * y[i];
+            }
+
+            return dotprod / (MathF.Sqrt(magx) * MathF.Sqrt(magy));
```
jeffhandley, line 285: If all values are 0 in one of the spans, I think that'd result in a DivideByZero here. Is that the desired behavior?

jeffhandley (commented):
Echoing the existing feedback about tests, we could add a lot more tests for all these methods to capture edge/corner cases and more exceptional data too.

_I meant to say, but we can add more tests after the RC2 milestone/snap_

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
tannergooding: Floating-point doesn't produce `DivideByZeroException`, it returns `NaN` (or `+/-Infinity`, depending on the exact values involved).

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +253,162 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+        /// <returns>The cosine similarity between the two vectors.</returns>
+        /// <exception cref="ArgumentException">Length of '<paramref name="x" />' must be same as length of '<paramref name="y" />'.</exception>
+        /// <exception cref="ArgumentException">'<paramref name="x" />' and '<paramref name="y" />' must not be empty.</exception>
+        public static float CosineSimilarity(ReadOnlySpan<float> x, ReadOnlySpan<float> y)
+        {
+            if (x.Length != y.Length)
+            {
+                ThrowHelper.ThrowArgument_SpansMustHaveSameLength();
+            }
+            if (x.Length == 0 || y.Length == 0)
```
tannergooding, line 269: We only need one of the checks here since we've already determined the lengths are equal.

That being said, is it desirable to throw here rather than allowing it to just return `NaN`?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +253,162 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+        /// <returns>The Euclidean distance.</returns>
+        /// <exception cref="ArgumentException">Length of '<paramref name="x" />' must be same as length of '<paramref name="y" />'.</exception>
+        /// <exception cref="ArgumentException">'<paramref name="x" />' and '<paramref name="y" />' must not be empty.</exception>
+        public static float Distance(ReadOnlySpan<float> x, ReadOnlySpan<float> y)
+        {
+            if (x.Length != y.Length)
+            {
+                ThrowHelper.ThrowArgument_SpansMustHaveSameLength();
+            }
+            if (x.Length == 0 || y.Length == 0)
```
tannergooding, line 302: Same here, the second check is redundant since we know the lengths are equivalent.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +253,162 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+
+            return dotprod;
+        }
+
+        /// <summary>
+        /// A mathematical operation that takes a vector and returns the L2 norm.
+        /// </summary>
+        /// <param name="x">The first tensor, represented as a span.</param>
+        /// <returns>The L2 norm.</returns>
+        public static float L2Normalize(ReadOnlySpan<float> x) // BLAS1: nrm2
```
tannergooding, line 347: Why is this `L2Normalize`? The standard API computes the "euclidean norm", that is the length or magnitude of the vector.

The name should likely just be `Norm`

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
stephentoub: These were already changed in my pr.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +253,162 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+            if (x.Length == 0)
+            {
+                ThrowHelper.ThrowArgument_SpansMustBeNonEmpty();
+            }
+
+            float expSum = 0f;
+
+            for (int i = 0; i < x.Length; i++)
+            {
+                expSum += MathF.Pow((float)Math.E, x[i]);
```
tannergooding, line 381: Why `MathF.Pow` and not `MathF.Exp`?

Why `(float)Math.E` and not `MathF.E`?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
jeffhandley: Never mind. Today I learned, "Dividing a floating-point value by zero doesn't throw an exception; it results in positive infinity, negative infinity, or not a number (NaN), according to the rules of IEEE 754 arithmetic."

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
stephentoub: @tannergooding, can you submit a cleanup PR with the changes you think should be made? Thanks.

PR #92145, opened 2023-09-15
add IO.Ports packaging for linux-musl

### LINKED ISSUE #63187: System.IO.Ports.SerialPort not working on linux-musl-arm64
We've tried to use SerialPort on linux-musl-arm64, but it doesn't work and I think the reason is that it is currently not supported. Is that right?

Do you have plans to add support in the near future?

wfurt:
fixes #63187


wfurt:
Any idea on this @hoyosjs, @ericstj and @ViktorHofer ?  This was attempt to fix  #63187 as we are missing packages but it seems like it is not as easy as I thought.

liu10725:
@wfurt Can you pass a compiled Linux-musl-arm64 "libSystem. IO. Ports. Native. so"? thanks

liu10725:
> @wfurt Can you pass a compiled Linux-musl-arm64 "libSystem. IO. Ports. Native. so"? thanks

Please upload a compiled so

wfurt:
I'm not sure if this is what you want but you can get everything produces by the PR build:

https://dev.azure.com/dnceng-public/public/_build/results?buildId=459685&view=artifacts&pathAsName=false&type=publishedArtifacts

find `library_bin_linux_musl_arm*`  and there should be `runtime.linux-musl-arm64.runtime.native.System.IO.Ports.9.0.0-ci.nupkg`


It would be _great_ to get independent validation.

liu10725:
@wfurt   I previously posted to Linux_ Musl_ Arm64 will encounter the following error

$ ldd libSystem.IO.Ports.Native.so
	ldd (0xffffbd9bc000)
	libc.so.6 => ldd (0xffffbd9bc000)
Error loading shared library ld-linux-aarch64.so.1: No such file or directory (needed by libSystem.IO.Ports.Native.so)


**But I found a new library file in your release directory and it has been perfectly fixed. I have successfully used this library in an openwrt system. Thank you**

libraries_bin_linux_musl_arm64_Release.zip\libraries_bin_linux_musl_arm64_Release\libraries_bin_linux_musl_arm64_Release.tar.gz\libraries_bin_linux_musl_arm64_Release.tar\bin\native\net9.0-linux-Release-arm64\libSystem.IO.Ports.Native.so

$ ldd libSystem.IO.Ports.Native.so
	ldd (0xffff8187f000)
	libc.musl-aarch64.so.1 => ldd (0xffff8187f000)

The problem has been fixed

wfurt:
thanks for confirmation @liu10725. Once merged, it should be available from package feed.

src/libraries/System.IO.Ports/pkg/runtime.native.System.IO.Ports.props
```diff
@@ -11,6 +11,7 @@
     <UseRuntimePackageDisclaimer>true</UseRuntimePackageDisclaimer>
     <!-- This is a native package and doesn't contain any ref/lib assets. -->
     <NoWarn>$(NoWarn);NU5128</NoWarn>
+    <DisablePackageBaselineValidation>true</DisablePackageBaselineValidation>
```
wfurt, line 14: This is similar to https://github.com/dotnet/runtime/blob/e0bed9410065d5e3554cbfc58e1ea1f1115c13e4/src/libraries/Microsoft.Bcl.Cryptography/src/Microsoft.Bcl.Cryptography.csproj#L10

We want to disable API compat check as we do not have anything to compare it it. And the binary runtimes really do not have any API surface.

ViktorHofer: Makes sense, please add this as an xml comment above the line.

ViktorHofer (approved):

akoeplinger (approved):

PR #92154, opened 2023-09-15
Add remaining set of TensorPrimitives APIs for .NET 8

stephentoub:
Represents the last planned .NET 8 surface area for TensorPrimitives (but not the last PR to improve it).

Adds non-vectorized implementations of:
- Max
- MaxMagnitude
- IndexOfMax
- IndexOfMaxMagnitude
- Min
- MinMagnitude
- IndexOfMin
- IndexOfMinMagnitude
- ConvertToHalf (only on .NET Core)
- ConvertToSingle (only on .NET Core)

Adds vectorized implementations of:
- Sum
- SumOfSquares
- SumOfMagnitudes
- Product
- ProductOfSums
- ProductOfDifferences

Vectorizes the implementations of:
- CosineSimilarity
- Distance
- L2Normalize
- Dot

Beyond vectorizing the non-vectorized ones, the vectorized implementations should be improved further, including:
- Handling alignment better
- Vectorizing the remainder that doesn't fit in a vector rather than falling back to scalar

### DIFF
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
@@ -533,7 +533,6 @@ namespace System.Numerics.Tensors
                 ThrowHelper.ThrowArgument_SpansMustBeNonEmpty();
             }
 
-            float result = float.PositiveInfinity;
             float resultMag = float.PositiveInfinity;
 
             for (int i = 0; i < x.Length; i++)
@@ -555,18 +554,16 @@ namespace System.Numerics.Tensors
 
                     if (currentMag < resultMag)
                     {
-                        result = current;
                         resultMag = currentMag;
                     }
                 }
                 else if (IsNegative(current))
                 {
-                    result = current;
                     resultMag = currentMag;
                 }
             }
 
-            return result;
+            return resultMag;
         }
 
         /// <summary>Computes the index of the maximum element in <paramref name="x"/>.</summary>
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
@@ -410,5 +410,448 @@ namespace System.Numerics.Tensors
                 destination[i] = 1f / (1 + MathF.Exp(-x[i]));
             }
         }
+
+        /// <summary>Computes the maximum element in <paramref name="x"/>.</summary>
+        /// <param name="x">The tensor, represented as a span.</param>
+        /// <returns>The maximum element in <paramref name="x"/>.</returns>
+        /// <exception cref="ArgumentException">Length of '<paramref name="x" />' must be greater than zero.</exception>
+        public static float Max(ReadOnlySpan<float> x)
+        {
+            if (x.IsEmpty)
+            {
+                ThrowHelper.ThrowArgument_SpansMustBeNonEmpty();
+            }
+
+            float result = float.NegativeInfinity;
+
+            for (int i = 0; i < x.Length; i++)
+            {
+                // This matches the IEEE 754:2019 `maximum` function.
+                // It propagates NaN inputs back to the caller and
+                // otherwise returns the greater of the inputs.
+                // It treats +0 as greater than -0 as per the specification.
+
+                float current = x[i];
+
+                if (current != result)
+                {
+                    if (float.IsNaN(current))
+                    {
+                        return current;
+                    }
+
+                    if (result < current)
+                    {
+                        result = current;
+                    }
+                }
+                else if (IsNegative(result))
+                {
+                    result = current;
+                }
+            }
+
+            return result;
+        }
+
+        /// <summary>Computes the minimum element in <paramref name="x"/>.</summary>
+        /// <param name="x">The tensor, represented as a span.</param>
+        /// <returns>The minimum element in <paramref name="x"/>.</returns>
+        /// <exception cref="ArgumentException">Length of '<paramref name="x" />' must be greater than zero.</exception>
+        public static float Min(ReadOnlySpan<float> x)
+        {
+            if (x.IsEmpty)
+            {
+                ThrowHelper.ThrowArgument_SpansMustBeNonEmpty();
+            }
+
+            float result = float.PositiveInfinity;
+
+            for (int i = 0; i < x.Length; i++)
+            {
+                // This matches the IEEE 754:2019 `minimum` function
+                // It propagates NaN inputs back to the caller and
+                // otherwise returns the lesser of the inputs.
+                // It treats +0 as greater than -0 as per the specification.
+
+                float current = x[i];
+
+                if (current != result)
+                {
+                    if (float.IsNaN(current))
+                    {
+                        return current;
+                    }
+
+                    if (current < result)
+                    {
+                        result = current;
+                    }
+                }
+                else if (IsNegative(current))
+                {
+                    result = current;
+                }
+            }
+
+            return result;
+        }
+
+        /// <summary>Computes the maximum magntude of any element in <paramref name="x"/>.</summary>
+        /// <param name="x">The tensor, represented as a span.</param>
+        /// <returns>The maximum magnitude of any element in <paramref name="x"/>.</returns>
+        /// <exception cref="ArgumentException">Length of '<paramref name="x" />' must be greater than zero.</exception>
+        public static float MaxMagnitude(ReadOnlySpan<float> x)
+        {
+            if (x.IsEmpty)
+            {
+                ThrowHelper.ThrowArgument_SpansMustBeNonEmpty();
+            }
+
+            float result = float.NegativeInfinity;
+            float resultMag = float.NegativeInfinity;
+
+            for (int i = 0; i < x.Length; i++)
+            {
+                // This matches the IEEE 754:2019 `maximumMagnitude` function.
+                // It propagates NaN inputs back to the caller and
+                // otherwise returns the input with a greater magnitude.
+                // It treats +0 as greater than -0 as per the specification.
+
+                float current = x[i];
+                float currentMag = Math.Abs(current);
+
+                if (currentMag != resultMag)
+                {
+                    if (float.IsNaN(currentMag))
+                    {
+                        return currentMag;
+                    }
+
+                    if (resultMag < currentMag)
+                    {
+                        result = current;
+                        resultMag = currentMag;
+                    }
+                }
+                else if (IsNegative(result))
+                {
+                    result = current;
+                    resultMag = currentMag;
+                }
+            }
+
+            return resultMag;
+        }
+
+        /// <summary>Computes the minimum magntude of any element in <paramref name="x"/>.</summary>
+        /// <param name="x">The tensor, represented as a span.</param>
+        /// <returns>The minimum magnitude of any element in <paramref name="x"/>.</returns>
+        /// <exception cref="ArgumentException">Length of '<paramref name="x" />' must be greater than zero.</exception>
+        public static float MinMagnitude(ReadOnlySpan<float> x)
+        {
+            if (x.IsEmpty)
+            {
+                ThrowHelper.ThrowArgument_SpansMustBeNonEmpty();
+            }
+
+            float result = float.PositiveInfinity;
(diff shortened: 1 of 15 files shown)


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -760,6 +1032,54 @@ public static partial class TensorPrimitives
...
+        {
+            public static float Invoke(float x) => x;
+            public static Vector128<float> Invoke(Vector128<float> x) => x;
+            public static Vector256<float> Invoke(Vector256<float> x) => x;
+#if NET8_0_OR_GREATER
+            public static Vector512<float> Invoke(Vector512<float> x) => x;
+#endif
+        }
+
+        private readonly struct LoadSqared : IUnaryOperator
```
gfoidl, line 1045: Super nit 😉
```suggestion
        private readonly struct LoadSquared : IUnaryOperator
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
```diff
@@ -390,6 +483,30 @@ public static partial class TensorPrimitives
             public Vector<float> Invoke(Vector<float> x, Vector<float> y, Vector<float> z) => (x * y) + z;
         }
 
+        private readonly struct LoadIdentity : IUnaryOperator
+        {
+            public float Invoke(float x) => x;
+            public Vector<float> Invoke(Vector<float> x) => x;
+        }
+
+        private readonly struct LoadSqared : IUnaryOperator
```
gfoidl, line 492:
```suggestion
        private readonly struct LoadSquared : IUnaryOperator
```

gfoidl (commented):
Just scrolled over it...

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +253,448 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+
+                        if (max < current)
+                        {
+                            result = i;
+                            max = current;
+                        }
+                    }
+                    else if (IsNegative(max))
+                    {
+                        result = i;
```
ericstj, line 472: This returns the last index of `+0` whereas for `NaN` we return the first index. Is that expected?  Same feedback for all the `IndexOf*` methods.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +253,448 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+                else if (IsNegative(current))
+                {
+                    result = current;
+                }
+            }
+
+            return result;
+        }
+
+        /// <summary>Computes the maximum magntude of any element in <paramref name="x"/>.</summary>
```
jeffhandley, line 343: Spelling:
```suggestion
        /// <summary>Computes the maximum magnitude of any element in <paramref name="x"/>.</summary>
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -253,5 +253,448 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
+                {
+                    result = current;
+                    resultMag = currentMag;
+                }
+            }
+
+            return resultMag;
+        }
+
+        /// <summary>Computes the minimum magntude of any element in <paramref name="x"/>.</summary>
```
jeffhandley, line 390: Spelling:
```suggestion
        /// <summary>Computes the minimum magnitude of any element in <paramref name="x"/>.</summary>
```

jeffhandley (approved):

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
stephentoub: I don't know what the desired behavior is, but under the assumption we want to return the first, I changed it.

ericstj (approved):
I have skimmed these where I understand them - did my best on the vectorized methods.  Can't find any issues. :shipit:

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -9,6 +9,336 @@ namespace System.Numerics.Tensors
...
+            {
+                ref float xRef = ref MemoryMarshal.GetReference(x);
+                ref float yRef = ref MemoryMarshal.GetReference(y);
+
+                Vector512<float> dotProductVector = Vector512<float>.Zero;
+                Vector512<float> xSumOfSquaresVector = Vector512<float>.Zero;
+                Vector512<float> ySumOfSquaresVector = Vector512<float>.Zero;
+
+                // Process vectors, summing their dot products and squares, as long as there's a vector's worth remaining.
+                int oneVectorFromEnd = x.Length - Vector<float>.Count;
```
gfoidl, line 77:
```suggestion
                int oneVectorFromEnd = x.Length - Vector512<float>.Count;
```
?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -9,6 +9,336 @@ namespace System.Numerics.Tensors
...
+            {
+                ref float xRef = ref MemoryMarshal.GetReference(x);
+                ref float yRef = ref MemoryMarshal.GetReference(y);
+
+                Vector256<float> dotProductVector = Vector256<float>.Zero;
+                Vector256<float> xSumOfSquaresVector = Vector256<float>.Zero;
+                Vector256<float> ySumOfSquaresVector = Vector256<float>.Zero;
+
+                // Process vectors, summing their dot products and squares, as long as there's a vector's worth remaining.
+                int oneVectorFromEnd = x.Length - Vector<float>.Count;
```
gfoidl, line 108:
```suggestion
                int oneVectorFromEnd = x.Length - Vector256<float>.Count;
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -9,6 +9,336 @@ namespace System.Numerics.Tensors
...
+                do
+                {
+                    Vector256<float> xVec = Vector256.LoadUnsafe(ref xRef, (uint)i);
+                    Vector256<float> yVec = Vector256.LoadUnsafe(ref yRef, (uint)i);
+
+                    dotProductVector += (xVec * yVec);
```
gfoidl, line 114: Here we have paranthesis, in the Vector512 path above not. Please make that consistent (either way).

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -9,6 +9,336 @@ namespace System.Numerics.Tensors
...
+            {
+                ref float xRef = ref MemoryMarshal.GetReference(x);
+                ref float yRef = ref MemoryMarshal.GetReference(y);
+
+                Vector128<float> dotProductVector = Vector128<float>.Zero;
+                Vector128<float> xSumOfSquaresVector = Vector128<float>.Zero;
+                Vector128<float> ySumOfSquaresVector = Vector128<float>.Zero;
+
+                // Process vectors, summing their dot products and squares, as long as there's a vector's worth remaining.
+                int oneVectorFromEnd = x.Length - Vector<float>.Count;
```
gfoidl, line 137:
```suggestion
                int oneVectorFromEnd = x.Length - Vector128<float>.Count;
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -9,6 +9,336 @@ namespace System.Numerics.Tensors
...
+                while (i <= oneVectorFromEnd);
+
+                // Sum the vector lanes into the scalar result.
+                dotProduct += Vector128.Sum(dotProductVector);
+                xSumOfSquares += Vector128.Sum(xSumOfSquaresVector);
+                ySumOfSquares += Vector128.Sum(ySumOfSquaresVector);
+            }
+
+            // Process any remaining elements past the last vector.
+            for (; (uint)i < (uint)x.Length; i++)
```
gfoidl, line 158:
```suggestion
            for (; i < x.Length; i++)
```
Are the casts needed here? Index access below is with `int`, so I don't see any benefit from the `uint` in the loop.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -9,6 +9,336 @@ namespace System.Numerics.Tensors
...
+                    i += Vector256<float>.Count;
+                }
+                while (i <= oneVectorFromEnd);
+
+                // Sum the vector lanes into the scalar result.
+                dotProduct += Vector256.Sum(dotProductVector);
+                xSumOfSquares += Vector256.Sum(xSumOfSquaresVector);
+                ySumOfSquares += Vector256.Sum(ySumOfSquaresVector);
+            }
+            else if (Vector128.IsHardwareAccelerated && x.Length >= Vector128<float>.Count)
```
gfoidl, line 127: Instead of `else if` this coulbe be an `if` checking against the remaining elements, in order to process more elements vectorized. I.e. for Vector256 there could remain 7 floats, which can then be processed via 1 Vector128 and 3 scalar loops. Similar to Vec512.

---

Or some more complicated code where the remainder is done vectorized too, by shifting and masking (zeroing) the duplicated vec-elements.
Sure, this can be done after that PR too...(with extensive measures if it's actually faster).

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
gfoidl: I wonder if there's  a way to test this. We access the elements in "unsafe" manner, so no bound checking done.

Say `Vector<float>.Count > Vector128<float>.Count`, so less than possible is done vectorized.
Say `Vector<float>.Count < Vector512<float>.Count`*, then it's read beyond the limits (maybe AV, but for spans they could be sliced from a bigger chunk, so AV is not very likely to a robust detection here).

The first case can be checked via review and perf-tests.
Is there a strategy for the latter case? Maybe an analyzer worth it (that flags uses of wrong VecXYZ inside an if-block)?
A more verbose variant would be to have read-/write-wrapper where the bounds are specified to use that for DEBUG builds, and for RELEASE the vanish.

\* ATM I don't know if `Vector<T>.Count` is Vec256.Count or does it use Vec512.Count? I know that there's an env-variable which can set the max bits for vector, but for the default value I don't remember.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: Technically the cast on the right shouldn't be necessary, but that's only a recent development, and so most of our code does so, and this is just staying consistent with that until they're all cleaned up. On the left hand side, unless the JIT can prove `i` isn't negative, it's necessary to enable it to avoid a bounds check (although there's another issue that prevents it from doing so with some loop shapes).

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: > Or some more complicated code where the remainder is done vectorized too

Right, I call that out in the PR description.

PR #92180, opened 2023-09-16
Update HttpClientJsonExtensions.Get.AsyncEnumerable.cs

### LINKED ISSUE #92179: `GetFromJsonAsAsyncEnumerable` extension default json serialize option is not the same as `GetFromJson`
### Description

`GetFromJsonAsAsyncEnumerable` extension default json serialize option is not the same as `GetFromJson`, for `GetFromJsonAsync<TValue>` etc extensions using the `new JsonSerializerOptions(JsonSerializerDefaults.Web)`, while the new `GetFromJsonAsAsyncEnumerable` extensions using the `JsonSerializerOptions.Default`

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.cs#L19

https://github.com/dotnet/runtime/blob/fa1bf3c0f9ec4e204203a49191170f9d53cd549a/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/JsonHelpers.cs#L15C10-L15C10

https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs#L138

https://github.com/dotnet/runtime/pull/89258/files#diff-096816e722c9256aa978e030242edf422dd8e41be616ab27bd7175cd522789e2R138

### Reproduction Steps

```c#

using var httpClient = new HttpClient();
httpClient.BaseAddress = new Uri("http://localhost:5297");
// we assume the response returned like
// [{"id":1,"title":"job_0 --- 09/16/2023 21:32:26 +08:00"},{"id":2,"title":"job_1 --- 09/16/2023 21:32:27 +08:00"}]
var stream = httpClient.GetFromJsonAsAsyncEnumerable<Job>("api/Jobs");
await foreach (var job in stream)
{
    Console.WriteLine(job);
    Console.WriteLine(DateTimeOffset.Now);
}

public sealed class Job
{
    public required int Id { get; set; }
    public required string Title { get; set; }
    public override string ToString() => $"JobId: {Id}, JobTitle: {Title}";
}
```

### Expected behavior

Think it should behave like `GetFromJsonAsync` to use the `JsonSerializerDefault.Web`

### Actual behavior

image

### Regression?

_No response_

### Known Workarounds

The `HttpContent` extension `ReadFromJsonAsAsyncEnumerable` is working

```c#
using var httpClient = new HttpClient();
httpClient.BaseAddress = new Uri("http://localhost:5297");
using var response = await httpClient.GetAsync("api/Jobs", HttpCompletionOption.ResponseHeadersRead);
ArgumentNullException.ThrowIfNull(response.Content);
var stream = response.Content.ReadFromJsonAsAsyncEnumerable<Job>();
await foreach (var job in stream)
{
    Console.WriteLine(job);
    Console.WriteLine(DateTimeOffset.Now);
}
```

### Configuration

.NET 8 RC 1

### Other information

_No response_

WeihanLi:
Update `jsonTypeInfo` for `ReadFromJsonAsAsyncEnumerable`

fixes #92179

### DIFF
--- a/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
+++ b/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
@@ -134,7 +134,7 @@ namespace System.Net.Http.Json
             Uri? requestUri,
             JsonSerializerOptions? options,
             CancellationToken cancellationToken)
-        {        
+        {
             return Core(client, requestUri, options, cancellationToken);
 
             static async IAsyncEnumerable<TValue?> Core(
--- a/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
+++ b/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
@@ -135,6 +135,11 @@ namespace System.Net.Http.Json
             JsonSerializerOptions? options,
             CancellationToken cancellationToken)
         {
+            if (client is null)
+            {
+                throw new ArgumentNullException(nameof(client));
+            }
+            
             return Core(client, requestUri, options, cancellationToken);
 
             static async IAsyncEnumerable<TValue?> Core(
--- a/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
+++ b/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
@@ -139,7 +139,7 @@ namespace System.Net.Http.Json
             {
                 throw new ArgumentNullException(nameof(client));
             }
-            
+
             return Core(client, requestUri, options, cancellationToken);
 
             static async IAsyncEnumerable<TValue?> Core(
--- a/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
+++ b/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
@@ -135,29 +135,9 @@ namespace System.Net.Http.Json
             JsonSerializerOptions? options,
             CancellationToken cancellationToken)
         {
-            if (client is null)
-            {
-                throw new ArgumentNullException(nameof(client));
-            }
+            var jsonTypeInfo = (JsonTypeInfo<TValue>)JsonHelpers.GetJsonTypeInfo(typeof(TValue), options);
 
-            return Core(client, requestUri, options, cancellationToken);
-
-            static async IAsyncEnumerable<TValue?> Core(
-                HttpClient client,
-                Uri? requestUri,
-                JsonSerializerOptions? options,
-                [EnumeratorCancellation] CancellationToken cancellationToken)
-            {
-                using HttpResponseMessage response = await client.GetAsync(requestUri, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
-                    .ConfigureAwait(false);
-                response.EnsureSuccessStatusCode();
-
-                await foreach (TValue? value in response.Content.ReadFromJsonAsAsyncEnumerable<TValue>(
-                    options, cancellationToken).ConfigureAwait(false))
-                {
-                    yield return value;
-                }
-            }
+            return FromJsonStreamAsyncCore(client, requestUri, jsonTypeInfo, cancellationToken);
         }
 
         private static IAsyncEnumerable<TValue?> FromJsonStreamAsyncCore<TValue>(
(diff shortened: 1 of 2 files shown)


src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
```diff
@@ -166,11 +183,8 @@ public static partial class HttpClientJsonExtensions
                     .ConfigureAwait(false);
                 response.EnsureSuccessStatusCode();
 
-                using Stream readStream = await GetHttpResponseStreamAsync(client, response, false, cancellationToken)
-                    .ConfigureAwait(false);
-
-                await foreach (TValue? value in JsonSerializer.DeserializeAsyncEnumerable<TValue>(
-                    readStream, jsonTypeInfo, cancellationToken).ConfigureAwait(false))
+                await foreach (TValue? value in response.Content.ReadFromJsonAsAsyncEnumerable<TValue>(
```
eiriktsarpalis, line 186: Please revert this change, it is unrelated to the fix at hand and the `GetHttpResponseStreamAsync` uses logic that is subtly different to reading directly from `HttpContent` values.

src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
```diff
@@ -135,12 +135,29 @@ public static partial class HttpClientJsonExtensions
...
+            if (client is null)
+            {
+                throw new ArgumentNullException(nameof(client));
+            }
 
-            var jsonTypeInfo = (JsonTypeInfo<TValue>)options.GetTypeInfo(typeof(TValue));
+            return Core(client, requestUri, options, cancellationToken);
 
-            return FromJsonStreamAsyncCore(client, requestUri, jsonTypeInfo, cancellationToken);
+            static async IAsyncEnumerable<TValue?> Core(
```
eiriktsarpalis, line 145: Likewise, this change would need to be reverted.

src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
```diff
@@ -135,12 +135,29 @@ public static partial class HttpClientJsonExtensions
             JsonSerializerOptions? options,
             CancellationToken cancellationToken)
         {
-            options ??= JsonSerializerOptions.Default;
-            options.MakeReadOnly();
+            if (client is null)
```
eiriktsarpalis, line 138: Private helpers should not be doing null validation. I would recommend replacing this will a `Debug.Assert(client != null);` statement and make the actual checks in the public methods that call into the helpers. This also avoids redundant checking that `client` is null in cases where the IAsyncEnumerable will be iterated over multiple times.

src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
```diff
@@ -135,12 +135,29 @@ public static partial class HttpClientJsonExtensions
...
+            {
+                throw new ArgumentNullException(nameof(client));
+            }
 
-            var jsonTypeInfo = (JsonTypeInfo<TValue>)options.GetTypeInfo(typeof(TValue));
```
eiriktsarpalis, line 141: The crux of the fix should be replacing the above with a call to the `JsonHelpers.GetJsonTypeInfo` method. See the `HttpContent` implementation for pointers.

src/libraries/System.Net.Http.Json/tests/FunctionalTests/HttpClientJsonExtensionsTests.cs
```diff
@@ -580,6 +580,25 @@ await foreach (string s in client.GetFromJsonAsAsyncEnumerable<string>("http://d
                 await Task.Delay(TimeSpan.FromMilliseconds(10));
             }
         }
+
+        [Fact]
+        public async Task GetFromJsonAsAsyncEnumerable_SerializerCaseInsensitiveTest()
```
eiriktsarpalis, line 585: Technically web defaults use camel casing, not case insensitive binding.
```suggestion
        public async Task GetFromJsonAsAsyncEnumerable_SerializerUsesCamelCase()
```

src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
WeihanLi: Reverted

src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
WeihanLi: Reverted

src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
WeihanLi: updated

src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Get.AsyncEnumerable.cs
WeihanLi: updated

WeihanLi:
Seemed the failed CI not related
image

eiriktsarpalis (approved):
Thanks!

PR #92185, opened 2023-09-16
Fix #91958: use `mach_timebase_info` to determine process time coefficient on macOS

### LINKED ISSUE #91958: Process::TotalProcessorTime is about 42 times lower than expected on ARM64 Mac
### Description

Let's consider this process on my system that has PID of 66358 (as shown by the Activity Monitor):
image

It has about 22 hours of CPU time.

### Reproduction Steps

Write and run this program:
```csharp
using System.Diagnostics;

var process = Process.GetProcessById(66358);
Console.WriteLine(process.TotalProcessorTime);
Console.WriteLine(process.TotalProcessorTime * 42);
```

### Expected behavior

This program should output about 22 hours in the first line, and some nonsensical value in the second line.

### Actual behavior

```
00:00:32.3739259
00:22:39.7048878
```

### Regression?

_No response_

### Known Workarounds

Multiply the result by 42 on ARM64 Mac :)

### Configuration

Macbook Pro M2 Max, ARM64
.NET SDK 7.0.202

### Other information

This is not the first time this API behaves problematically on macOS. The previous fix was https://github.com/dotnet/corefx/pull/37637

I suspect that the fix was incorrect: in reality, we should call `mach_timebase_info` to get the coefficient to convert the value returned by the underlying API to nanoseconds, and apparently that value may be system-dependent.

ForNeVeR:
Closes #91958.

I have compared the results of the process timing functions to the results of `ps` on my M2 MacBook device, and they seem to work properly after the changes.

The new tests were properly failing before the change (since we were returning values 42 times lower than the native results), and are, of course, green after the changes.


src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.OSX.cs
```diff
@@ -109,5 +110,16 @@ private static Interop.libproc.rusage_info_v3 GetCurrentProcessRUsage()
...
             return Interop.libproc.proc_pid_rusage(Environment.ProcessId);
         }
+
+        private static unsafe Interop.libSystem.mach_timebase_info_data_t GetTimeBase()
+        {
+            Interop.libSystem.mach_timebase_info_data_t timeBase = default;
+            var returnCode = Interop.libSystem.mach_timebase_info(&timeBase);
+            if (returnCode != 0)
+            {
+                throw new Win32Exception(returnCode);
```
ForNeVeR, line 120: Please check if this is correct.

src/libraries/System.Diagnostics.Process/tests/ProcessTests.Unix.cs
```diff
@@ -1016,3 +1049,13 @@ private static string StartAndReadToEnd(string filename, string[] arguments)
         }
     }
 }
+
+// TODO: Use the actual localization data in here?
```
ForNeVeR, line 1053: Please help me to figure out what to do with this.

Apparently, the `Interop.*.cs` files I've added to the test assembly relies on resources/localization, but I haven't figured out how to reuse the main projects' resources in tests.

I think it's possible to just duplicate the API I use in tests here and get rid of `SR`. But this will be quite a lot of duplication, since the structure returned by `proc_pid_rusage` is kinda long.

src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.OSX.cs
ForNeVeR: One possibility I see here would be to just return a coefficient of `1` in case of an error in this place, to fall back to the old behavior.

tmds: > One possibility I see here would be to just return an effective coefficient of numer: 1, denom: 1 in case of an error in this place, to fall back to the old behavior.

Yes, let's do this.

And, let's add a `Debug.Assert` so debug test runs will fail if this would ever return non-success.

Should we account for `mach_timebase_info` not existing on some systems?

am11: I think our minimum macOS support offsets this API’s minimum requirement.

ForNeVeR: Apparently [it was available at least in OS X 10.10.1](https://stackoverflow.com/a/31454136/2684760) back in 2015, and .NET only supports 10.15+, so we should be okay.

I'll apply the suggestion.

ForNeVeR: Done!

src/libraries/Common/src/Interop/OSX/Interop.libSystem.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class libSystem
+    {
+        [LibraryImport(Interop.Libraries.libSystem)]
+        public static unsafe partial int mach_timebase_info(mach_timebase_info_data_t* info);
+        public struct mach_timebase_info_data_t
```
adamsitnik, line 16: @jkoritzinsky @AaronRobinsonMSFT In terms of marshaller best practices, do we need to explicitly specify sequential layout for such structs?

```suggestion
        [StructLayout(LayoutKind.Sequential)]
        public struct mach_timebase_info_data_t
```

src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.OSX.cs
```diff
@@ -10,6 +10,7 @@ namespace System.Diagnostics
     public partial class Process
     {
         private const int NanosecondsTo100NanosecondsFactor = 100;
+        private static readonly Interop.libSystem.mach_timebase_info_data_t timeBase = GetTimeBase();
```
adamsitnik, line 13: It would be better to initialize this field lazily, when we need it for the first time. Otherwise this sys-call:
- may be called even if we don't need it (during type initializaiton)
- in theory it may fail and be very hard to handle properly

Since we use the following logic in 4 places:

```cs
return new TimeSpan(Convert.ToInt64($ulong * timeBase.numer / timeBase.denom / NanosecondsTo100NanosecondsFactor));
```

 we could introduce a helper method that could take care of everything, something like this:

```suggestion
        private static volatile uint s_timeBase_numer, s_timeBase_denom;
        
        private static TimeSpan Map(ulong sysTime)
        {
            uint denom = s_timeBase_denom;
            if (denom == default)
            {
                Interop.libSystem.mach_timebase_info_data_t timeBase = GetTimeBase();
                s_timeBase_denom = denom = timeBase.denom;
                s_timeBase_numer = timeBase.numer;
            }
            uint numer = s_timeBase_numer;
            
            return new TimeSpan(Convert.ToInt64(sysTime * numer / denom / NanosecondsTo100NanosecondsFactor));
        }
```

src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.OSX.cs
```diff
@@ -82,7 +83,7 @@ public TimeSpan UserProcessorTime
             {
                 EnsureState(State.HaveNonExitedId);
                 Interop.libproc.rusage_info_v3 info = Interop.libproc.proc_pid_rusage(_processId);
-                return new TimeSpan(Convert.ToInt64(info.ri_user_time / NanosecondsTo100NanosecondsFactor));
+                return new TimeSpan(Convert.ToInt64(info.ri_user_time * timeBase.numer / timeBase.denom / NanosecondsTo100NanosecondsFactor));
```
adamsitnik, line 86: thank you for finding and updating all use cases! 👍

src/libraries/System.Diagnostics.Process/tests/ProcessTests.Unix.cs
adamsitnik: That is a great question!

You should be able to reference the resource file in the test project similar to:

https://github.com/dotnet/runtime/blob/913231acc359c5e94d4397e8fe0c8a38ae6b0c5a/src/libraries/System.Formats.Tar/tests/System.Formats.Tar.Tests.csproj#L5

adamsitnik (commented):
Overall it looks good, but I found few places that could be polished a bit.

Thank you for your contribution @ForNeVeR !

src/libraries/Common/src/Interop/OSX/Interop.libSystem.cs
AaronRobinsonMSFT: Technically, no. Value types in .NET default to sequential layout. However, and this is annoying, there are some Roslyn warnings that are suppressed if one does explicitly mark the type with sequential layout. The reasoning here is historical, but the gist is if Roslyn complains about unreferenced fields, which can happen for types used in interop, then placing `StructLayout(LayoutKind.Sequential)` on the type will automatically suppress the warning.

The interop team's general guidance here has been to accept the defaults except where there is annoying friction with C# or where the tooling requires explicit details. This falls into the C# friction bucket, but only if a warning is emitted.

adamsitnik: @AaronRobinsonMSFT thank you for a very detailed answer!

src/libraries/System.Diagnostics.Process/tests/ProcessTests.Unix.cs
ForNeVeR: For some reason, this only started working for me after I added OSX-specific TFM to the test project. I guess it has something to do with how the resource layout works? Anyway, please take another look at the current solution. I have to say I am a bit confusing around this part.

src/libraries/Common/src/Interop/OSX/Interop.libSystem.cs
ForNeVeR: I can't see any related warnings in the compilation (and also we seem to have warnings as errors enabled in this part?).

Does this mean this attribute is unnecessary? I am totally okay with adding that if required. Though, yeah, we all know that sequential is the default struct layout 😅

src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.OSX.cs
ForNeVeR: Yeah, agree with your reasoning and applied a bit modified version of this snippet (the only change is in naming, I like `MapTime` a little bit better).

src/libraries/Common/src/Interop/OSX/Interop.libSystem.cs
AaronRobinsonMSFT: > Does this mean this attribute is unnecessary?

Yes.

ForNeVeR:
Build on CI is now failing due to something being wrong with `SR` on Windows. Have I broken that? I need help with resources, I don't think my changes to `StringResourcesPath` were right.

src/libraries/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests.csproj
```diff
@@ -2,9 +2,10 @@
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <IncludeRemoteExecutor>true</IncludeRemoteExecutor>
-    <TargetFrameworks>$(NetCoreAppCurrent)-windows;$(NetCoreAppCurrent)-unix;$(NetCoreAppCurrent)-browser</TargetFrameworks>
+    <TargetFrameworks>$(NetCoreAppCurrent)-windows;$(NetCoreAppCurrent)-unix;$(NetCoreAppCurrent)-browser;$(NetCoreAppCurrent)-osx</TargetFrameworks>
     <IgnoreForCI Condition="'$(TargetOS)' == 'browser'">true</IgnoreForCI>
     <EnableLibraryImportGenerator>true</EnableLibraryImportGenerator>
+    <StringResourcesPath>$(MSBuildProjectDirectory)\..\src\Resources\Strings.resx</StringResourcesPath>
```
am11, line 8: We have two options:

1. Include the two keys in test resources:
    ```xml
	  <data name="CantGetAllPids" xml:space="preserve">
	    <value>Could not get all running Process IDs.</value>
	  </data>
	  <data name="RUsageFailure" xml:space="preserve">
	    <value>Failed to set or retrieve rusage information. See the error code for OS-specific error information.</value>
	  </data>
    ```
    in:
    https://github.com/dotnet/runtime/blob/b4a3c9e6fbe09d2f52465e2189d57ef5c04b94c3/src/libraries/System.Diagnostics.Process/tests/Resources/Strings.resx#L1
    and delete this hard-coded line.

2. Include test resources alongside source ones -- separated by semicolon `;`:
    ```xml
		 <StringResourcesPath>$(MSBuildProjectDirectory)\Resources\Strings.resx;$(MSBuildProjectDirectory)\..\src\Resources\Strings.resx</StringResourcesPath>
    ```

    multiple resources seem to be supported.

Option 1 is probably much cleaner that avoids mixing stuff, but I'll defer to others. cc @ViktorHofer

src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.OSX.cs
```diff
@@ -109,5 +110,33 @@ private static Interop.libproc.rusage_info_v3 GetCurrentProcessRUsage()
...
+            uint denom = s_timeBase_denom;
+            if (denom == default)
+            {
+                Interop.libSystem.mach_timebase_info_data_t timeBase = GetTimeBase();
+                s_timeBase_denom = denom = timeBase.denom;
+                s_timeBase_numer = timeBase.numer;
+            }
+            uint numer = s_timeBase_numer;
+
+            return new TimeSpan(Convert.ToInt64(sysTime * numer / denom / NanosecondsTo100NanosecondsFactor));
```
danmoseley, line 125: Should it do the last division first, like here?

https://chromium.googlesource.com/chromium/src/+/refs/tags/58.0.3029.141/base/time/time_mac.cc#43

(I don't know what values it typically returns)

danmoseley: Oh, this says it gets the 42 from 125/3. Does 125 * ticks go over ulong max?

https://stackoverflow.com/questions/69318594/how-to-get-the-true-value-of-the-system-clock-in-raw-ticks-i-e-mach-absolute-t

danmoseley: Also I suppose you can store 1000 * denom 

I'm sure someone will point out that won't necessarily give the exact same result @tannergooding 🙂

src/libraries/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests.csproj
ViktorHofer: Agreed. Option 1 is cleaner while it adds some duplication.

src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.OSX.cs
ForNeVeR: That's a good point, but I was concerned about losing some precision due to doing division first. Though I can't wrap my head around it right now. Will try to do more thorough analysis later today.

ForNeVeR: Ok, here are my thoughts.

Typical tick values are pretty large actually. For comparison, let's consider a program that was working on a single-core CPU for a year.
```
> [TimeSpan]::FromDays(365).Ticks
315360000000000
> [long]::MaxValue
9223372036854775807
```

(here I consider `long` since `TimeSpan` takes `long` in its ctor, ot `ulong`)

This means we will overflow on a `numer` of `29248` (or if our tick value will get significantly bigger, i.e. we take 29000 years into account, or a CPU with 29k cores).

This is very far from being realistic, but it is also much closer than I expected. So, to be on the safe side, let's do the same as Chromium does: divide by our factor first, to get a hundred times more space, by losing some precision.

We may, of course, also do `value / (denom * 100) * numer` which is relatively the same, yet will lose a bit more precision as well.

danmoseley: I'm also OK with asserting that numer is reasonably small ( say < 1000) as an alternative, since we suspect that will always be true.

ForNeVeR: My current suggestion is to leave this as `sysTime / NanosecondsTo100NanosecondsFactor * numer / denom`.

I am not sure how useful an assertion would be in this code. Perhaps it'd be better to add `checked`, since we are concerned by overflow? We do not expect this code to be too performance sensitive, so `checked` might be good.

What do you think?

ForNeVeR:
I believe that all the existing feedback is addressed, so I'm waiting for further feedback. Thank you so much for your time, folks.

src/libraries/System.Diagnostics.Process/src/System/Diagnostics/Process.OSX.cs
danmoseley: Not my area, but I"m not sure an exception would be an improvement. I'd be interested in thoughts of a numeric expert like @tannergooding . I'll step aside for area owners to sign off overall...

ForNeVeR: I would say that an explicit exception might be an improvement (especially since it would be thrown from a member property such as `TotalProcessorTime` and not on class init).

But of course, I am ready to listen to other suggestions.

marek-safar:
@dotnet/area-system-diagnostics-process please review

EgorBo:
Ping @dotnet/area-system-diagnostics-process

adamsitnik (approved):
LGTM, thank you for your fix @ForNeVeR !

And apologies for the delay (I was on a parental leave).

src/libraries/System.Diagnostics.Process/tests/Resources/Strings.resx
```diff
@@ -120,4 +120,10 @@
   <data name="Argv_IncludeDoubleQuote" xml:space="preserve">
     <value>The argv[0] argument cannot include a double quote.</value>
   </data>
+  <data name="CantGetAllPids" xml:space="preserve">
```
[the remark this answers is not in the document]
adamsitnik: https://github.com/dotnet/runtime/pull/93539

ForNeVeR:
Thank you so much for help! ❤️

jeffhandley:
Backporting this fix to 8.0 since this was reported again in #98121.

PR #92187, opened 2023-09-17
Ensure the adapter name 100% matching when parsing proc/net/dev

### LINKED ISSUE #91982: Network Traffic Statistics on Linux

skyoxZ:
Fix #91982

* Add a mock adapter `wlan0a` in `FunctionalTests/NetworkFiles/dev` to reveal the issue.

* Add checking the adapter name.

* Remove unused variable `index`.

### DIFF
--- a/src/libraries/System.Net.NetworkInformation/src/System/Net/NetworkInformation/StringParsingHelpers.Statistics.cs
+++ b/src/libraries/System.Net.NetworkInformation/src/System/Net/NetworkInformation/StringParsingHelpers.Statistics.cs
@@ -415,7 +415,13 @@ namespace System.Net.NetworkInformation
                         Span<Range> pieces = stackalloc Range[18]; // [0]-[16] used, +1 to ensure any additional segment goes into [17]
                         ReadOnlySpan<char> lineSpan = line;
                         pieces = pieces.Slice(0, lineSpan.SplitAny(pieces, " :", StringSplitOptions.RemoveEmptyEntries));
-                        if (!lineSpan[pieces[0]].SequenceEqual(name)) continue; // Ensure the adapter name 100% matching
+
+                        if (!lineSpan[pieces[0]].SequenceEqual(name))
+                        {
+                            // The adapter name doesn't exactly match.
+                            continue;
+                        }
+
                         return new IPInterfaceStatisticsTable()
                         {
                             BytesReceived = ParseUInt64AndClampToInt64(lineSpan[pieces[1]]),
--- a/src/libraries/System.Net.NetworkInformation/src/System/Net/NetworkInformation/StringParsingHelpers.Statistics.cs
+++ b/src/libraries/System.Net.NetworkInformation/src/System/Net/NetworkInformation/StringParsingHelpers.Statistics.cs
@@ -414,7 +414,12 @@ namespace System.Net.NetworkInformation
                     if (line.Contains(name))
                     {
                         ReadOnlySpan<char> lineSpan = line;
-                        lineSpan.SplitAny(pieces, " :", StringSplitOptions.RemoveEmptyEntries);
+                        int pieceCount = lineSpan.SplitAny(pieces, " :", StringSplitOptions.RemoveEmptyEntries);
+
+                        if (pieceCount < 17)
+                        {
+                            continue;
+                        }
 
                         if (!lineSpan[pieces[0]].SequenceEqual(name))
                         {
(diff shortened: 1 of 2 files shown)


src/libraries/System.Net.NetworkInformation/src/System/Net/NetworkInformation/StringParsingHelpers.Statistics.cs
```diff
@@ -407,16 +407,15 @@ internal static IPInterfaceStatisticsTable ParseInterfaceStatisticsTableFromFile
             {
                 sr.ReadLine();
                 sr.ReadLine();
-                int index = 0;
                 while (!sr.EndOfStream)
                 {
                     string line = sr.ReadLine()!;
                     if (line.Contains(name))
                     {
```
MichalPetryka, line 414: Does it make sense to remove the Contains here since we're checking with SequenceEqual later again?

skyoxZ: I thought the `Contains` is cheap so let it do a precheck might have a better performance.

src/libraries/System.Net.NetworkInformation/src/System/Net/NetworkInformation/StringParsingHelpers.Statistics.cs
```diff
@@ -407,16 +407,15 @@ internal static IPInterfaceStatisticsTable ParseInterfaceStatisticsTableFromFile
...
-                        Span<Range> pieces = stackalloc Range[18]; // [0] skipped, [1]-[16] used, +1 to ensure any additional segment goes into [17]
+                        Span<Range> pieces = stackalloc Range[18]; // [0]-[16] used, +1 to ensure any additional segment goes into [17]
```
MihaZupan, line 415: This may now run multiple times. Can you please move the `stackalloc` out of the loop?

src/libraries/System.Net.NetworkInformation/src/System/Net/NetworkInformation/StringParsingHelpers.Statistics.cs
```diff
@@ -407,16 +407,15 @@ internal static IPInterfaceStatisticsTable ParseInterfaceStatisticsTableFromFile
...
                         ReadOnlySpan<char> lineSpan = line;
                         pieces = pieces.Slice(0, lineSpan.SplitAny(pieces, " :", StringSplitOptions.RemoveEmptyEntries));
-
+                        if (!lineSpan[pieces[0]].SequenceEqual(name)) continue; // Ensure the adapter name 100% matching
```
MihaZupan, line 418:
```suggestion

                        if (!lineSpan[pieces[0]].SequenceEqual(name))
                        {
                            // The adapter name doesn't exactly match.
                            continue;
                        }

```

skyoxZ:
Feedback addressed. I also removed `pieces.Slice` (technically it's not equivalent)

src/libraries/System.Net.NetworkInformation/src/System/Net/NetworkInformation/StringParsingHelpers.Statistics.cs
```diff
@@ -407,15 +407,20 @@ internal static IPInterfaceStatisticsTable ParseInterfaceStatisticsTableFromFile
...
+                Span<Range> pieces = stackalloc Range[18]; // [0]-[16] used, +1 to ensure any additional segment goes into [17]
                 while (!sr.EndOfStream)
                 {
                     string line = sr.ReadLine()!;
                     if (line.Contains(name))
                     {
-                        Span<Range> pieces = stackalloc Range[18]; // [0] skipped, [1]-[16] used, +1 to ensure any additional segment goes into [17]
                         ReadOnlySpan<char> lineSpan = line;
-                        pieces = pieces.Slice(0, lineSpan.SplitAny(pieces, " :", StringSplitOptions.RemoveEmptyEntries));
+                        lineSpan.SplitAny(pieces, " :", StringSplitOptions.RemoveEmptyEntries);
```
MihaZupan, line 417:
```suggestion
                        int pieceCount = lineSpan.SplitAny(pieces, " :", StringSplitOptions.RemoveEmptyEntries);

                        if (pieceCount < 17)
                        {
                            continue;
                        }
```

Since we're ignoring the count returned here, we could end up reading a range from a previous iteration if we encounter invalid input, whereas before we would have thrown an `IndexOutOfRangeException`.

MihaZupan (approved):
Thank you

PR #92191, opened 2023-09-17
Fix trimming of DebuggerDisplay with Name

arturek:
The `Name` property of the `DebuggerDisplay` attribute accepts the same format string as its `Value` property, but does not prevent trimming members it references. Thanks to this fix, members referenced by any of these two properties are not trimmed and can be displayed by a debugger.

Additionally, this change adds references of three projects to the illink solution file to fix its build errors.


vitek-karas (approved):
The change looks good - thanks!
@sbomer could you please also review this?

I would prefer to separate the solution file change to a different PR - as it's unrelated.

sbomer (commented):
Thank you! Should the `Type` property be given the same treatment?

arturek:
According to the [doc](https://learn.microsoft.com/en-us/visualstudio/debugger/using-the-debuggerdisplay-attribute?view=vs-2022#:~:text=They%20can%20be%20set%20to%20strings%20using%20the%20same%20syntax%20as%20the%20constructor.) this change should include `Type`. Thanks @sbomer for noticing it. I'll update the code and prepare another PR with the .sln file alone.

vitek-karas (approved):

vitek-karas:
@arturek Thanks a lot for the fix!

PR #92192, opened 2023-09-17
Improve the help message in build.ps1

arturek:
Prints the default architecture in the help message, which not always is obvious on Windows.

### DIFF
--- a/eng/build.ps1
+++ b/eng/build.ps1
@@ -30,7 +30,7 @@ function Get-Help() {
   Write-Host "Common settings:"
   Write-Host "  -arch (-a)                     Target platform: x86, x64, arm, arm64, or wasm."
   Write-Host "                                 Pass a comma-separated list to build for multiple architectures."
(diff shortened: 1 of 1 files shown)


eng/build.ps1
```diff
@@ -30,7 +30,7 @@ function Get-Help() {
   Write-Host "Common settings:"
   Write-Host "  -arch (-a)                     Target platform: x86, x64, arm, arm64, or wasm."
   Write-Host "                                 Pass a comma-separated list to build for multiple architectures."
-  Write-Host "                                 [Default: Your machine's architecture.]"
+  Write-Host ("                                 [Default: {0} (Depends on your machine's architecture.)]" -f [System.Runtime.InteropServices.RuntimeInformation]::ProcessArchitecture.ToString().ToLowerInvariant())
```
agocke, line 33: This doesn't look right. The machine architecture is not the same as the process architecture. Which one are we using?

arturek: My code is just a copy of the [initialization of this parameter](https://github.com/dotnet/runtime/blob/main/eng/build.ps1#L14). I haven't changed its description.

arturek: Thanks @hoyosjs . It took me a while to notice that my console unexpectedly defaults to x86 and that there's another version of the console which I should use.

akoeplinger (approved):

PR #92205, opened 2023-09-18
Use FMA in TensorPrimitives.CosineSimilarity

stephentoub:


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -80,9 +83,18 @@ private static float CosineSimilarityCore(ReadOnlySpan<float> x, ReadOnlySpan<fl
                     Vector512<float> xVec = Vector512.LoadUnsafe(ref xRef, (uint)i);
                     Vector512<float> yVec = Vector512.LoadUnsafe(ref yRef, (uint)i);
 
-                    dotProductVector += xVec * yVec;
-                    xSumOfSquaresVector += xVec * xVec;
-                    ySumOfSquaresVector += yVec * yVec;
+                    if (Avx512F.IsSupported)
```
stephentoub, line 86: Is this check needed inside of Vector512.IsHardwareAccelerated?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -111,9 +123,18 @@ private static float CosineSimilarityCore(ReadOnlySpan<float> x, ReadOnlySpan<fl
                     Vector256<float> xVec = Vector256.LoadUnsafe(ref xRef, (uint)i);
                     Vector256<float> yVec = Vector256.LoadUnsafe(ref yRef, (uint)i);
 
-                    dotProductVector += xVec * yVec;
-                    xSumOfSquaresVector += xVec * xVec;
-                    ySumOfSquaresVector += yVec * yVec;
+                    if (Fma.IsSupported)
```
stephentoub, line 126: Is this check needed inside of Vector256.IsHardwareAccelerated?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -140,9 +161,24 @@ private static float CosineSimilarityCore(ReadOnlySpan<float> x, ReadOnlySpan<fl
...
+                    }
+                    else if (AdvSimd.IsSupported)
+                    {
+                        dotProductVector = AdvSimd.FusedMultiplyAdd(dotProductVector, xVec, yVec);
+                        xSumOfSquaresVector = AdvSimd.FusedMultiplyAdd(xSumOfSquaresVector, xVec, xVec);
+                        ySumOfSquaresVector = AdvSimd.FusedMultiplyAdd(ySumOfSquaresVector, yVec, yVec);
+                    }
+                    else
+                    {
+                        dotProductVector += xVec * yVec;
```
EgorBo, line 178: do we care here about the fact these paths are not the same in terms of IEEE754?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
EgorBo: --instruction-set avx2 doesn't enable FMA but enables Vector256.IsHardwareAccelerated so yes, this check is needed as avx2 doesn't imply FMA

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
EgorBo: No, it's not needed here, Avx512F is basically the baseline

huoyaoyuan:
I wonder if we can have cross-platform FMA intrinsic since it's really a frequently used operation. Or are JIT allowed to translate Multiply(Add(a, b), c) node into FMA? FP precision can be a concern.

EgorBo:
> I wonder if we can have cross-platform FMA intrinsic since it's really a frequently used operation. Or are JIT allowed to translate Multiply(Add(a, b), c) node into FMA? FP precision can be a concern.

We can introduce a cross-plat helper for it, but it will be either "approximate" (meaning its return value won't be the same for the same input on all platforms) or its non-FMA implementation will be terribly slower than a*b+c, e.g.: https://godbolt.org/z/8G63e6sKv

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: > My general impression that FMA doesn't improve things a lot to trade determinism for

What determinism are you referring to? Do you mean across machines?

I thought it was desirable in terms of gaining a more accurate answer by not losing accuracy due to intermediate rounding.

If it's not beneficial, we can close this, though then I wonder why we've gone through the trouble of exposing all of these FMA primitives.

EgorBo: > What determinism are you referring to? Do you mean across machines?

Yep, however, if we can assume FMA is mostly always available we can just change the fallback to match its behavior - it will be slower but at least the result will be repeatable accross different platforms.

> If it's not beneficial

It's probably just my impression, but I can be wrong. Also, like you mentioned, it also provides better accuracy

tannergooding:
> I wonder if we can have cross-platform FMA intrinsic since it's really a frequently used operation

We already have this, it is `T.FusedMultiplyAdd`. We don't have it for SIMD yet, but that's because its not universally available.

The best option would be some kind of `MultiplyAddEstimate` (matching the other `*Estimate` functions) which allows it to be optimized to the more efficient of `(a * b) + c` or `fma(a, b, c)`

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: Right, the actual performance benefit of `FMA` is marginal outside of specific scenarios but it does change the maximum error of the 2 step result from `1 ULP` to `0.5 ULP` (Units in Last Place). This may or may not be a problem depending on whether its part of a converged result.

The main places it helps with performance is very heavy `multiply-add` chains where it can allow better utilization of the multiplier unit. It can cause a perf regression or no improvement in some cases like `(a * b) + (c * d)” where all data is simultaneously ready due to the data dependence it introduces on the FMA

stephentoub:
@tannergooding, is this a good change to make? If no, I'll close it. If yes, can you review? Thanks.

tannergooding (approved):
Changes look good/correct to me.

Whether or not we take this I think depends on if we're okay with subtly different behavior depending on the hardware being used.

I, personally, think that difference is ok for this scenario. We already will have some different behavior due to the underlying C runtime algorithm for various Math/MathF APIs. We could opt to provide a configuration switch allowing users to opt out of the faster in favor of the more precise if it becomes a concern.

stephentoub:
SGTM

PR #92212, opened 2023-09-18
[release/8.0] JIT: Initialize jitstdout lazily

jakobbotsch:
Backport of #92123 to release/8.0

## Customer Impact
The JIT may crash the process during startup in custom hosting scenarios where `stdout` is closed before the JIT is initialized. Customer reported in #91856.

## Testing
Manually verified that the JIT code making I/O operations on stdout no longer runs (unless needed, e.g. due to `DOTNET_JitDisasm` being set).

## Risk
Low, make some code not run during startup/shutdown, and make some mechanical changes in code that does not run in release builds.


EgorBo (approved):

jeffschwMSFT (approved):
approved. this can be merged when ready

PR #92215, opened 2023-09-18
[QUIC] Throw ODE if connection/listener is disposed

### LINKED ISSUE #75115: QuicListener.AcceptConnectionAsync rethrowing exceptions
`QuicListener.AcceptConnectionAsync` throws in the following scenarios (that I know of):

1. `QuicException` with `QuicError.OperationAborted`. Happens when the listener is disposed while waiting for a connection.
2. `ObjectDisposedException`. Happens when `AcceptConnectionAsync` is called after disposing the listener.
3. `AuthenticationException`. Happens when the client and server fail to negotiate a connection, e.g. server has an invalid certificate and the client rejects it.
4. `ArgumentException`. Happens when the options returned from the callback are invalid. e.g. a server certificate isn't configured in the callback.
5. Any exception is thrown in the callback. Happens when an exception is thrown in the callback and is then rethrown from `AcceptConnectionAsync`.

I think 5 is very broad. There should be a way to disambiguate a rethrown exception from important exceptions that are used to signal the health of the listener. e.g. someone could accidentally throw an ODE in the callback, which the code that calls the listener interprets as the listener is shut down and to stop listening for connections.

I think the exception from 5, and maybe also 3 and 4, should be wrapped in another exception that says an attempt to accept a connection failed. The inner exception then provides more detail.

ManickaP:
AcceptConnection/StreamAsync now throws ODE in case the listener/connection was stopped by DisposeAsync.

Addresses the point 1. from #75115, which is the last remaining, unfixed concern.

Fixes #75115

### DIFF
--- a/src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnection.cs
+++ b/src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnection.cs
@@ -382,6 +382,9 @@ public sealed partial class QuicConnection : IAsyncDisposable
             {
                 await stream.DisposeAsync().ConfigureAwait(false);
             }
+
+            // Propagate ODE if disposed in the meantime.
+            ObjectDisposedException.ThrowIf(_disposed == 1, this);
             // Propagate connection error if present.
             if (_acceptQueue.Reader.Completion.IsFaulted)
             {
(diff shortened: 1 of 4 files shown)


rzikm (approved):
LGTM!

wfurt (approved):
LGTM

ManickaP:
Outerloop failures unrelated.

Issue #88569, opened 2023-07-09
Port System.Net documentation for .NET 8.0 APIs


carlossanlop:
Below is the list of APIs that still show up as undocumented in dotnet-api-docs and were introduced in .NET 8.0.

Full porting instructions can be found in the [main issue](https://github.com/dotnet/runtime/issues/88561).

This task needs to be finished before the RC2 snap (September 18th).

| Summary  | Parameters | TypeParameters | ReturnValue | API                                                                                                                                                                                                                                                                                                                  |
|----------|------------|----------------|-------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Missing  | NA         | NA             | NA          | [F:System.Net.HttpStatusCode.UnprocessableContent](https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Net/HttpStatusCode.xml)                                                                                                                                                                            |
| Missing  | Missing    | NA             | Missing     | [M:System.Net.IPAddress.System#IFormattable#ToString(System.String,System.IFormatProvider)](https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Net/IPAddress.xml)                                                                                                                                        |
| Missing  | Missing    | NA             | Missing     | [M:System.Net.IPAddress.System#IParsable<System#Net#IPAddress>#Parse(System.String,System.IFormatProvider)](https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Net/IPAddress.xml)                                                                                                                        |
| Missing  | Missing    | NA             | Missing     | [M:System.Net.IPAddress.System#IParsable<System#Net#IPAddress>#TryParse(System.String,System.IFormatProvider,System.Net.IPAddress@)](https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Net/IPAddress.xml)                                                                                               |
| Missing  | Missing    | NA             | Missing     | [M:System.Net.IPAddress.System#ISpanFormattable#TryFormat(System.Span{System.Char},System.Int32@,System.ReadOnlySpan{System.Char},System.IFormatProvider)](https://github.com/dotnet/dotnet-api-docs/blob/main/xml/System.Net/IPAddress.xml)                                                                         |
[...]

PR #92223, opened 2023-09-18
Add missing System.Net.Http.Json comments

MihaZupan:
Contributes to #88569.

Ported to docs in https://github.com/dotnet/dotnet-api-docs/pull/9302

### DIFF
--- a/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Delete.cs
+++ b/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Delete.cs
@@ -84,7 +84,7 @@ namespace System.Net.Http.Json
         /// <param name="client">The client used to send the request.</param>
         /// <param name="requestUri">The Uri the request is sent to.</param>
         /// <param name="type">The type of the object to deserialize to and return.</param>
-        /// <param name="context">Source generated JsonSerializerContext used to control the deserialization behavior.</param>
+        /// <param name="context">The JsonSerializerContext used to control the deserialization behavior.</param>
         /// <param name="cancellationToken">A cancellation token that can be used by other objects or threads to receive notice of cancellation.</param>
         /// <returns>The task object representing the asynchronous operation.</returns>
         /// <exception cref="ArgumentNullException">The <paramref name="client"/> is <see langword="null"/>.</exception>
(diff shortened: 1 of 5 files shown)


src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpContentJsonExtensions.AsyncEnumerable.cs
```diff
@@ -60,17 +60,18 @@ public static partial class HttpContentJsonExtensions
...
-            return ReadFromJsonAsAsyncEnumerableCore(content, jsonTypeInfo, cancellationToken);
-        }
-
+        /// <summary>
+        /// Reads the HTTP content and returns the value that results from deserializing the content as
+        /// JSON in an async enumerable operation.
+        /// </summary>
+        /// <typeparam name="TValue">The target type to deserialize to.</typeparam>
+        /// <param name="content">The content to read from.</param>
+        /// <param name="jsonTypeInfo">Source generated JsonTypeInfo to control the behavior during deserialization.</param>
```
stephentoub, line 69: Must it be "source generated"?

MihaZupan: I suppose not, I can update the existing places that use the same wording.

wfurt (approved):
LGTM

eiriktsarpalis (commented):
Thanks!

src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/HttpClientJsonExtensions.Delete.cs
```diff
@@ -84,7 +84,7 @@ public static partial class HttpClientJsonExtensions
         /// <param name="client">The client used to send the request.</param>
         /// <param name="requestUri">The Uri the request is sent to.</param>
         /// <param name="type">The type of the object to deserialize to and return.</param>
-        /// <param name="context">Source generated JsonSerializerContext used to control the deserialization behavior.</param>
+        /// <param name="context">The JsonSerializerContext used to control the deserialization behavior.</param>
```
eiriktsarpalis, line 87: Technically a `JsonSerializerContext` could have been manually extended by the user and not by a source generator, but I think the change is ok otherwise.

PR #92229, opened 2023-09-18
Optimize Random{NumberGenerator}.GetItems for power-of-two choices

### LINKED ISSUE #92130: RandomNumberGenerator bad performance on GetString and possible reason
### Description

I was using .Net 7 and did not have available a way to generate a random string based on the Base64 url encoded characters and SO answers were showing low perf code. So I made one during this summer until I noticed yesterday that .Net 8 has a new `GetString` so I jumped into it . The one I have is faster up to 10 times for a 32 long string. Culprit and code below.

### Configuration
.Net 8 RC1, Windows 11 Pro 64 V 22H2, Thinkpad X1 Yoga

### Data for length = 11 (YouTube ids)

| Method     | Mean       | Error    | StdDev   | Median     | Gen0   | Allocated |
|----------- |-----------:|---------:|---------:|-----------:|-------:|----------:|
| GetString2 |   158.7 ns |  3.22 ns |  7.64 ns |   160.7 ns | 0.0114 |      48 B |
| GetString  | 1,002.2 ns | 20.08 ns | 46.93 ns | 1,027.0 ns | 0.0114 |      48 B |

### Analysis

Problem is calling `GetInt32` in the loop

```
       private static void GetItemsCore<T>(ReadOnlySpan<T> choices, Span<T> destination)
        {
            for (int i = 0; i < destination.Length; i++)
            {
                destination[i] = choices[GetInt32(choices.Length)];
            }
        }
```

### Limits
Limits of my approach. I must have a  `choices` string for which the length is a power of 2 <256. That is the case for `Length64String` and it is enough for my purposes and I believe for most web developpers.
I did not dig into your code to see if you have the same limitation, i.e. whether you need to abide to in order to ensure a uniform distribution.

### Solution

If you have that same limitation (I do not expect, a priori), then the API must be corrected to ensure uniformity.
If you do not have this limitation, I still believe it is worth having specialized methods to generate strings that are a power of 2 while < 256 or review the `GetItemsCore` to improve perfomance.

### Code Program.cs

```
using System.Security.Cryptography;

BenchmarkRunner.Run<RandomBase64StringGenerator>();

[MemoryDiagnoser]
public class RandomBase64StringGenerator
{
    const int length = 11; 

    [Benchmark]
    public void GetString2()
    {
       RandomGenerator.New64(length);
    }

    [Benchmark]
    public void GetString()
    {
        RandomNumberGenerator.GetString(RandomGenerator.Constants.Length64String,length);
    }
}

public sealed class RandomGenerator
{
    public class Constants
    {
        public const string Length64String = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_-";
    }

    const int _stackAllocThreshold = 32;

    public static string New64(int length = 11)
    {
        return GetRandomString(length, Constants.Length64String);
    }

    private static string GetRandomString(int length, ReadOnlySpan<char> characters)
    {
        // No plan to go beyond _stackAllocThreshold, so we throw
        if (length <= 0 || length > _stackAllocThreshold)
            throw new ArgumentException($"{nameof(length)} must be at least one" +
                $" and lower or equal to {_stackAllocThreshold}");

        Span<byte> bytes = stackalloc byte[length];
        Span<char> chars = stackalloc char[length];

        RandomNumberGenerator.Fill(bytes);

        int mask = 256 / characters.Length;

        for (var i = 0; i < bytes.Length; i++)
            chars[i] = characters[bytes[i] / mask];

        return new string(chars);
    }
}
```

stephentoub:
When possible, make a single call to get the required randomness rather than one per element.

Closes https://github.com/dotnet/runtime/issues/92130

```C#
using System.Security.Cryptography;

public class Tests
{
    private const int Length = 11;
    private static readonly string s_charChoices = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_-";
    private static ReadOnlySpan<int> s_intChoices => [42, 43, 44, 45, 46, 47, 48, 49];
    private static int[] s_dest = new int[Length];

    [Benchmark]
    public string RngGetString() => RandomNumberGenerator.GetString(s_charChoices, Length);

    [Benchmark]
    public void RandGetItems() => Random.Shared.GetItems(s_intChoices, s_dest);
}
```

| Method       | Toolchain         | Mean      | Ratio |
|------------- |------------------ |----------:|------:|
| RngGetString | \main\corerun.exe | 764.18 ns |  1.00 |
| RngGetString | \pr\corerun.exe   |  98.25 ns |  0.13 |
|              |                   |           |       |
| RandGetItems | \main\corerun.exe |  57.21 ns |  1.00 |
| RandGetItems | \pr\corerun.exe   |  16.56 ns |  0.29 |


src/libraries/System.Private.CoreLib/src/System/Random.cs
```diff
@@ -194,6 +196,38 @@ public void GetItems<T>(ReadOnlySpan<T> choices, Span<T> destination)
...
+            // - The number of choices is <= 256. This let's us get a single byte per choice.
+            // - The number of choices is a power of two. This let's us a byte and simply mask off
+            //   unnecessary bytes cheaply rather than needing to use rejection sampling.
+            // - The type of T doesn't contain any references and is larger than a byte in size.
+            //   This let's us write random bytes into the destination buffer rather than needing
+            //   a scratch buffer. we can use a single RNG.Fill to get all of the required randomness,
+            //   and then walk backwards through the destination, filling in the corresponding choice
+            //   based on the random byte at that position when interpreted as bytes. By walking backwards,
+            //   we won't overrite any random bytes that we still need to read.
+            if (!RuntimeHelpers.IsReferenceOrContainsReferences<T>() && Unsafe.SizeOf<T>() >= 2 &&
```
jkotas, line 210:
```suggestion
            if (!RuntimeHelpers.IsReferenceOrContainsReferences<T>() && Unsafe.SizeOf<T>() >= 1 &&
```

src/libraries/System.Private.CoreLib/src/System/Random.cs
```diff
@@ -194,6 +196,38 @@ public void GetItems<T>(ReadOnlySpan<T> choices, Span<T> destination)
...
+                BitOperations.IsPow2(choices.Length) && choices.Length <= 256)
+            {
+                // Reinterpret the bottom destination.Length bytes of the destination as our
+                // scratch buffer for randomness.
```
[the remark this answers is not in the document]
jkotas: We could generate a block of random bytes to a stack scratch buffer and fill the output one block at a time.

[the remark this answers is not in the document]
stephentoub: > We could generate a block of random bytes to a stack scratch buffer and fill the output one block at a time.

We can, it just means we either further limit where this can be applied to length stackalloc'able, or we pay the cost of ArrayPool for larger sizes.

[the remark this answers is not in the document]
jkotas: I meant something like this:
```csharp
Span<byte> randomBytes = stackalloc byte[512];
while (destination.Length > 0)
{
    if (destination.Length < randomBytes.Length)
    {
        randomBytes = randomBytes.Slice(0, destination.Length);
    }

    NextBytes(randomBytes);

    int mask = choices.Length - 1;
    for (int i = 0; i < randomBytes.Length; i++)
    {
        destination[i] = choices[randomBytes[i] & mask];
    }

    destination = destination.Slice(randomBytes.Length);
}
```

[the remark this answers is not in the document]
stephentoub: > I meant something like this:

I see, trading off multiple Fill/NextByte calls for a temporary array.

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/RandomNumberGenerator.cs
```diff
@@ -350,6 +353,38 @@ private static void GetHexStringCore(Span<char> destination, bool lowercase)
...
+            // do so only once rather than per element in a common case:
+            // - The number of choices is <= 256. This let's us get a single byte per choice.
+            // - The number of choices is a power of two. This let's us a byte and simply mask off
+            //   unnecessary bytes cheaply rather than needing to use rejection sampling.
+            // - The type of T doesn't contain any references and is larger than a byte in size.
+            //   This let's us write random bytes into the destination buffer rather than needing
+            //   a scratch buffer. we can use a single RNG.Fill to get all of the required randomness,
+            //   and then walk backwards through the destination, filling in the corresponding choice
+            //   based on the random byte at that position when interpreted as bytes. By walking backwards,
+            //   we won't overrite any random bytes that we still need to read.
```
vcsjones:
```suggestion
            //   we won't overwrite any random bytes that we still need to read.
```

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/RandomNumberGenerator.cs
```diff
@@ -350,6 +353,44 @@ private static void GetHexStringCore(Span<char> destination, bool lowercase)
...
+                // If T is unmanaged, we _could_ use the destination buffer itself, reinterpreted
+                // as a byte buffer, to hold the scratch bytes, but then we'd potentially be temporarily
+                // writing invalid Ts into the destination, which is a risk.
+                byte[]? arrayPoolArray = null;
+                Span<byte> randomBytes = destination.Length <= StackAllocThreshold ?
+                    stackalloc byte[StackAllocThreshold] :
+                    (arrayPoolArray = ArrayPool<byte>.Shared.Rent(destination.Length));
+                randomBytes = randomBytes.Slice(0, destination.Length);
+
+                Fill(randomBytes);
```
vcsjones, line 376: It will probably get inlined, but we can go directly to the internal implementation here.

```suggestion
                RandomNumberGeneratorImplementation.FillSpan(randomBytes);
```

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/RandomNumberGenerator.cs
```diff
@@ -350,6 +353,44 @@ private static void GetHexStringCore(Span<char> destination, bool lowercase)
...
+            // - The number of choices is <= 256. This let's us get a single byte per choice.
+            // - The number of choices is a power of two. This let's us use a byte and simply mask off
+            //   unnecessary bits cheaply rather than needing to use rejection sampling.
+            if (BitOperations.IsPow2(choices.Length) && choices.Length <= 256)
+            {
+                const int StackAllocThreshold = 256;
+
+                // Get a scratch buffer to fill with random bytes, either stack space if the
+                // destination is small enough, or an array pool array if it's larger.
+                // If T is unmanaged, we _could_ use the destination buffer itself, reinterpreted
```
vcsjones, line 367: `T` is `char` is probably the most interesting case to optimize. I would be in favor of keeping the re-interpret for special casing `char` since random strings are the most interesting case and the impetus for this API.

If we feel that adds too much complexity for a narrow case of "input len is pow2" then that's understandable too.

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/RandomNumberGenerator.cs
```diff
@@ -350,6 +353,44 @@ private static void GetHexStringCore(Span<char> destination, bool lowercase)
...
+                // as a byte buffer, to hold the scratch bytes, but then we'd potentially be temporarily
+                // writing invalid Ts into the destination, which is a risk.
+                byte[]? arrayPoolArray = null;
+                Span<byte> randomBytes = destination.Length <= StackAllocThreshold ?
+                    stackalloc byte[StackAllocThreshold] :
+                    (arrayPoolArray = ArrayPool<byte>.Shared.Rent(destination.Length));
```
vcsjones, line 373:
```suggestion
                    (arrayPoolArray = CryptoPool.Rent(destination.Length));
```

src/libraries/System.Private.CoreLib/src/System/Random.cs
```diff
@@ -194,6 +197,38 @@ public void GetItems<T>(ReadOnlySpan<T> choices, Span<T> destination)
...
 
+            // The most expensive part of this operation is the call to get random data. We can
+            // do so potentially many fewer times if:
+            // - the number of choices is <= 256. This let's us get a single byte per choice.
+            // - the number of choices is a power of two. This let's us use a byte and simply mask off
+            //   unnecessary bits cheaply rather than needing to use rejection sampling.
+            // In such a case, we can grab a bunch of random bytes in one call.
+            if (BitOperations.IsPow2(choices.Length) && choices.Length <= 256)
+            {
+                Span<byte> randomBytes = stackalloc byte[512]; // arbitrary size, a balance between stack consumed and number of random calls required
```
bartonjs, line 208: The line length with this comment post-statement feels a bit long.

```suggestion
                // This is an arbitrary size, a balance between stack consumed and number of random calls required
                Span<byte> randomBytes = stackalloc byte[512];
```

looks happier (more in line with the rest of the file (or, at least, PR))

bartonjs (approved):

src/libraries/System.Private.CoreLib/src/System/Random.cs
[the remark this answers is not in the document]
bartonjs: Yeah, when I did the AsBytes for GetString (in a comment in the reporting issue) I thought it was OK because no one other than the thread filling that string knew about the target memory.  In a chat I had with @vcsjones on Friday I said (or was convinced so hard by Kevin that I now recall having said it) that I wouldn't like that for GetItemsCore, but it was OK for GetString.

@jkotas 's approach here is more or less what I was thinking when I paged this out of active thinking on Friday.

Ponant:
I few remarks if you do not mind.

1) I tested @bartonjs method against mine and confirm it is faster, but it is mainly due to the use bitwise AND instead of integer division, and not theunsafe territory. At least on my machine.

2) I think the benefit of the pow2 non-negligible speedup should be documented in th emethods that use it such as GetString, perhaps...

3) I have an addional suggestion, which I think can be beneficial to users and which can be implemented in the same PR. But I do not know if oyu prefer that I create another issue or drop it here? it is about rewritting GetBytes to allow users to have a byte arrays where each byte is bound to [0,N] where N is a power of two. For N=256, we recover the case of the current getBytes. Let me know I will drop a code

stephentoub:
> I have an addional suggestion, which I think can be beneficial to users and which can be implemented in the same PR. But I do not know if oyu prefer that I create another issue or drop it here? it is about rewritting GetBytes to allow users to have a byte arrays where each byte is bound to [0,N] where N is a power of two. For N=256, we recover the case of the current getBytes. Let me know I will drop a code

This sounds very niche, and the savings is effectively an array access per element. I don't think it's worth special-casing without seeing more examples of where it would be really meaningful.

> I think the benefit of the pow2 non-negligible speedup should be documented in the methods that use it such as GetString, perhaps...

It's an implementation detail. We avoid documenting such optimization implementations, as we reserve the right to change them at any point.

vcsjones (approved):

Ponant:
> > I have an addional suggestion, which I think can be beneficial to users and which can be implemented in the same PR. But I do not know if oyu prefer that I create another issue or drop it here? it is about rewritting GetBytes to allow users to have a byte arrays where each byte is bound to [0,N] where N is a power of two. For N=256, we recover the case of the current getBytes. Let me know I will drop a code
> 
> This sounds very niche, and the savings is effectively an array access per element. I don't think it's worth special-casing without seeing more examples of where it would be really meaningful.
> 

A) Generate random booleans
B) Generate Random vectors in 4 dimensions (my case)

stephentoub:
> > > I have an addional suggestion, which I think can be beneficial to users and which can be implemented in the same PR. But I do not know if oyu prefer that I create another issue or drop it here? it is about rewritting GetBytes to allow users to have a byte arrays where each byte is bound to [0,N] where N is a power of two. For N=256, we recover the case of the current getBytes. Let me know I will drop a code
> > 
> > 
> > This sounds very niche, and the savings is effectively an array access per element. I don't think it's worth special-casing without seeing more examples of where it would be really meaningful.
> 
> A) Generate random booleans B) Generate Random vectors in 4 dimensions (my case)

I don't follow.  Are you proposing a new API? Or are you proposing that the implementation special-case `typeof(T) == typeof(bool)` and validates that `choices[0] == false` and `choices[1] == Unsafe.As<byte, bool>(1)`?  And for `typeof(T) == typeof(byte)`, looping through all of the choices as a precompute step to make sure that `choices[i] == (byte)i`?

GrabYourPitchforks:
@stephentoub Consider also duplicating the test here with one that passes a power-of-two sized values span. That way we get distribution tests across both implementations which now exist in this method.

https://github.com/dotnet/runtime/blob/0ad4e699b0a6d92c2d08f7b9b6393448ab8f310a/src/libraries/System.Security.Cryptography/tests/RandomNumberGeneratorTests.cs#L669-L675

Ponant:
The current `GetBytes` returns and array of bytes, where each byte can take any value from 0 to 255. My proposal is to have GetBytes to also allow me to return an array of bytes but where each byte is bound to [0,N] where N is a powere of two. For example, I can generate a byte array of bools  011100011010101 or random quaternions or any 4d vector, or octonions or above powers.

I am using something like that, it is essentially the same spirit as you do here with the new GetString in this PR, but for getting byte arrays. I am using something like that (checks on parameters omitted for brievety)

```
private static byte[] GetRandomBytes(int length, int maxNumOfBytes)
 {
        Span<byte> bytes = length <= _stackAllocThreshold ? stackalloc byte[length] : new byte[length];

        // Generate random bytes, each taking values between 0 and 255
        RandomNumberGenerator.Fill(bytes);

        // We do not need to loop if we cover all byte possibilities, since  RandomNumberGenerator.Fill does this 
        if (maxNumOfBytes != 256)
        {
            byte mask = (byte)(maxNumOfBytes - 1);

            // make each byte between 0 and maxNumOfBytes
            for (var i = 0; i < bytes.Length; i++)
                bytes[i] &= mask;

        }

        return bytes.ToArray();
}
```

stephentoub:
> The current GetBytes returns an array of bytes, where each byte can take any value from 0 to 255. My proposal is to have GetBytes to also allow me to return an array of bytes but where each byte is bound to [0,N] where N is a power of two.

That's a new API.  Feel free to open a separate proposal for one.  That won't be done as part of an optimization PR.

vcsjones:
> The current GetBytes returns an array of bytes, where each byte can take any value from 0 to 255. My proposal is to have GetBytes to also allow me to return an array of bytes but where each byte is bound to [0,N] where N is a power of two. For example, I can generate a byte array of bools 011100011010101 or random quaternions or any 4d vector, or octonions or above powers.

This would be a new [API proposal](https://github.com/dotnet/runtime/blob/main/docs/project/api-review-process.md).

We can discuss the merits and usefulness of the proposal there to keep this PR focused on the perf of existing APIs.

stephentoub:
@bartonjs, per your feedback on my recent PR, I've marked this as a breaking change to be documented.

stephentoub:
Opened https://github.com/dotnet/runtime/pull/108018 to revert the breaking change aspect.

PR #92241, opened 2023-09-18
Adding AVX512 path to Base64 encoding/Decoding

DeepakRajendrakumaran:
### Overview

This PR implements an AVX512 code path for Base64 encoding/Decoding. This is based on the work by _Wojciech Muła_ and _Daniel Lemire_. There is a fast `AVX512VBMI` path and the fallback uses `AVX512F/BW` - I'll refer to these as **VBMI_AVAILABLE** and **VBMI_UNAVAILABLE** here on. For performance purposes, this will be compared to an AVX2 implementation which will be referred to as **BASE_Version**
Reference for the algorithm:

- https://arxiv.org/pdf/1910.05109.pdf

This version uses intrinsics directly and not generic vector libraries due to lack of current support in JIt/Vector libraries to produce optimal code. Some additional support which would be required in order to use generic vector library for implementing this would be

1. Add `ShuffleUnsafe` for Vector512
2. Extend `Vector512.Shuffle()` to lower to intrinsics instead of going to fallback for more cases.
3. Expand `Vector512 `surface area to incorporate more high level functions

Even the current implementation can be further optimized by adding the `multishift()` implementation. This is a further optimization

### Generated code

(Will be focusing on the actual encoding/decoding code within the loop only)

### **Encoding**

**VBMI_AVAILABLE**

image

**VBMI_UNAVAILABLE**

image

**BASE_VERSION**
image

### **Decoding**

**VBMI_AVAILABLE**

image

**VBMI_UNAVAILABLE**

image

**BASE_VERSION**
image

### ON ICX -
**BASE_VERSION** vs **VBMI_UNAVAILABLE**

image

**BASE_VERSION** vs **VBMI_AVAILABLE**
image

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
@@ -68,7 +68,18 @@ namespace System.Buffers.Text
 
                 if (maxSrcLength >= 24)
                 {
-                    byte* end = srcMax - 45;
+                    byte* end = srcMax - 88;
+                    if (Vector512.IsHardwareAccelerated && Avx512BW.IsSupported && (end >= src))
+                    {
+                        Avx512Decode(ref src, ref dest, end, maxSrcLength, destLength, srcBytes, destBytes);
+
+                        if (src == srcEnd)
+                        {
+                            goto DoneExit;
+                        }
+                    }
+
+                    end = srcMax - 45;
                     if (Avx2.IsSupported && (end >= src))
                     {
                         Avx2Decode(ref src, ref dest, end, maxSrcLength, destLength, srcBytes, destBytes);
@@ -614,6 +625,137 @@ namespace System.Buffers.Text
 
             destIndex = localDestIndex;
             return status;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        [CompExactlyDependsOn(typeof(Avx512BW))]
+        [CompExactlyDependsOn(typeof(Avx512Vbmi))]
+        private static unsafe void Avx512Decode(ref byte* srcBytes, ref byte* destBytes, byte* srcEnd, int sourceLength, int destLength, byte* srcStart, byte* destStart)
+        {
+            // Reference for VBMI implementation : https://arxiv.org/pdf/1910.05109.pdf
+            // If we have AVX512 support, pick off 64 bytes at a time for as long as we can,
+            // but make sure that we quit before seeing any == markers at the end of the
+            // string. Also, because we write 16 zeroes at the end of the output, ensure
+            // that there are at least 22 valid bytes of input data remaining to close the
+            // gap. 64 + 2 + 22 = 88 bytes.
+            byte* src = srcBytes;
+            byte* dest = destBytes;
+
+            // The JIT won't hoist these "constants", so help it
+            // JIT will remove the Non VBMI path constants when VBMI is available and vice-versa.
+
+            // Vbmi path constants
+            Vector512<sbyte> vbmiLookup0 = Vector512.Create(
+                0x80808080, 0x80808080, 0x80808080, 0x80808080,
+                0x80808080, 0x80808080, 0x80808080, 0x80808080,
+                0x80808080, 0x80808080, 0x3e808080, 0x3f808080,
+                0x37363534, 0x3b3a3938, 0x80803d3c, 0x80808080).AsSByte();
+            Vector512<sbyte> vbmiLookup1 = Vector512.Create(
+                0x02010080, 0x06050403, 0x0a090807, 0x0e0d0c0b,
+                0x1211100f, 0x16151413, 0x80191817, 0x80808080,
+                0x1c1b1a80, 0x201f1e1d, 0x24232221, 0x28272625,
+                0x2c2b2a29, 0x302f2e2d, 0x80333231, 0x80808080).AsSByte();
+            Vector512<byte> vbmiPackedLanesControl = Vector512.Create(
+                0x06000102, 0x090a0405, 0x0c0d0e08, 0x16101112,
+                0x191a1415, 0x1c1d1e18, 0x26202122, 0x292a2425,
+                0x2c2d2e28, 0x36303132, 0x393a3435, 0x3c3d3e38,
+                0x00000000, 0x00000000, 0x00000000, 0x00000000).AsByte();
+
+            // Non Vbmi path constants
+            Vector512<sbyte> lutHi = Vector512.Create(
+                0x10, 0x10, 0x01, 0x02, 0x04, 0x08, 0x04, 0x08, 0x10, 0x10, 0x10, 0x10, 0x10, 0x10, 0x10, 0x10,
+                0x10, 0x10, 0x01, 0x02, 0x04, 0x08, 0x04, 0x08, 0x10, 0x10, 0x10, 0x10, 0x10, 0x10, 0x10, 0x10,
+                0x10, 0x10, 0x01, 0x02, 0x04, 0x08, 0x04, 0x08, 0x10, 0x10, 0x10, 0x10, 0x10, 0x10, 0x10, 0x10,
+                0x10, 0x10, 0x01, 0x02, 0x04, 0x08, 0x04, 0x08, 0x10, 0x10, 0x10, 0x10, 0x10, 0x10, 0x10, 0x10);
+
+            Vector512<sbyte> lutLo = Vector512.Create(
+                0x15, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x13, 0x1A, 0x1B, 0x1B, 0x1B, 0x1A,
+                0x15, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x13, 0x1A, 0x1B, 0x1B, 0x1B, 0x1A,
+                0x15, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x13, 0x1A, 0x1B, 0x1B, 0x1B, 0x1A,
+                0x15, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x11, 0x13, 0x1A, 0x1B, 0x1B, 0x1B, 0x1A);
+            Vector512<sbyte> mask2F = Vector512.Create((sbyte)'/');
+            Vector512<sbyte> lutShift = Vector512.Create(
+                0, 16, 19, 4, -65, -65, -71, -71, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 16, 19, 4, -65, -65, -71, -71, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 16, 19, 4, -65, -65, -71, -71, 0, 0, 0, 0, 0, 0, 0, 0,
+                0, 16, 19, 4, -65, -65, -71, -71, 0, 0, 0, 0, 0, 0, 0, 0);
+            Vector512<sbyte> packBytesInLaneMask = Vector512.Create(
+                        2, 1, 0, 6, 5, 4, 10, 9, 8, 14, 13, 12, -1, -1, -1, -1,
+                        18, 1, 0, 6, 5, 4, 10, 9, 8, 14, 13, 12, -1, -1, -1, -1,
+                        2, 1, 0, 6, 5, 4, 10, 9, 8, 14, 13, 12, -1, -1, -1, -1,
+                        2, 1, 0, 6, 5, 4, 10, 9, 8, 14, 13, 12, -1, -1, -1, -1);
+            Vector512<int> packLanesControl = Vector512.Create(0, 1, 2, 4, 5, 6, 8, 9, 10, 12, 13, 14, 0, 0, 0, 0);
+
+            // Common path constants
+            Vector512<sbyte> mergeConstant0 = Vector512.Create(0x01400140).AsSByte();
+            Vector512<short> mergeConstant1 = Vector512.Create(0x00011000).AsInt16();
+
+            // The fastest version of this algorithm requires AVX512VBMI support.
+            // The fallback path uses AVX512BW and AVX512F instructions.
+            do
+            {
+                AssertRead<Vector512<sbyte>>(src, srcStart, sourceLength);
+                Vector512<sbyte> str = Vector512.Load(src).AsSByte();
+
+                // Step 1: Translate encoded Base64 input to their original indices
+                // This step also checks for invalid inputs and exits.
+                // After this, we have indices which are verified to have upper 2 bits set to 0 in each byte.
+                // origIndex      = [...|00dddddd|00cccccc|00bbbbbb|00aaaaaa]
+                Vector512<sbyte> origIndex;
+                if (Avx512Vbmi.IsSupported)
+                {
+                    origIndex = Avx512Vbmi.PermuteVar64x8x2(vbmiLookup0, str, vbmiLookup1);
+                    ulong errorMask = (Vector512.BitwiseOr(origIndex.AsInt32(), str.AsInt32()).AsSByte()).ExtractMostSignificantBits();
+                    if (errorMask != 0)
+                    {
+                        break;
+                    }
+                }
+                else
+                {
+                    Vector512<sbyte> hiNibbles = Vector512.BitwiseAnd(Avx512F.ShiftRightLogical(str.AsInt32(), 4).AsSByte(), mask2F);
+
+                    Vector512<sbyte> loNibbles = Vector512.BitwiseAnd(str, mask2F);
+                    Vector512<sbyte> hi = Avx512BW.Shuffle(lutHi, hiNibbles);
+                    Vector512<sbyte> lo = Avx512BW.Shuffle(lutLo, loNibbles);
+                    if (!Vector512.EqualsAll(Vector512.BitwiseAnd(hi, lo), Vector512<sbyte>.Zero))
+                    {
+                        break;
+                    }
+                    Vector512<sbyte> eq2F = Vector512.Equals(str, mask2F);
+                    Vector512<sbyte> shift = Avx512BW.Shuffle(lutShift, Vector512.Add(eq2F, hiNibbles));
+                    origIndex = Vector512.Add(str, shift);
+                }
+
+
+
+                // Step 2: Now we need to reshuffle bits to remove the 0 bits.
+                // multiAdd1: [...|0000cccc|ccdddddd|0000aaaa|aabbbbbb]
+                Vector512<short> multiAdd1 = Avx512BW.MultiplyAddAdjacent(origIndex.AsByte(), mergeConstant0);
+                // multiAdd1: [...|00000000|aaaaaabb|bbbbcccc|ccdddddd]
+                Vector512<int> multiAdd2 = Avx512BW.MultiplyAddAdjacent(multiAdd1, mergeConstant1);
+
+                // Step 3: Pack 48 bytes
+                if (Avx512Vbmi.IsSupported)
+                {
+                    str = Avx512Vbmi.PermuteVar64x8(multiAdd2.AsByte(), vbmiPackedLanesControl).AsSByte();
+                }
+                else
+                {
+                    Vector512<int> packed = Avx512BW.Shuffle(multiAdd2.AsSByte(), packBytesInLaneMask).AsInt32();
+                    str = Avx512F.PermuteVar16x32(packed, packLanesControl).AsSByte();
+
+                }
+
+                AssertWrite<Vector512<sbyte>>(dest, destStart, destLength);
+                Vector512.Store(str.AsByte(), dest);
+                src += 64;
+                dest += 48;
+            }
+            while (src <= srcEnd);
+
+            srcBytes = src;
+            destBytes = dest;
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
--- a/src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
@@ -705,20 +705,20 @@ namespace System.Buffers.Text
                 if (Avx512Vbmi.IsSupported)
                 {
                     origIndex = Avx512Vbmi.PermuteVar64x8x2(vbmiLookup0, str, vbmiLookup1);
-                    ulong errorMask = (Vector512.BitwiseOr(origIndex.AsInt32(), str.AsInt32()).AsSByte()).ExtractMostSignificantBits();
-                    if (errorMask != 0)
+                    Vector512<sbyte> errorVec = (origIndex.AsInt32() | str.AsInt32()).AsSByte();
+                    if (errorVec.ExtractMostSignificantBits() != 0)
                     {
                         break;
                     }
                 }
                 else
                 {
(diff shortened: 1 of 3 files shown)


DeepakRajendrakumaran:
@BruceForstall @tannergooding @dotnet/avx512-contrib

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
```diff
@@ -616,6 +627,137 @@ private static OperationStatus DecodeWithWhiteSpaceFromUtf8InPlace(Span<byte> ut
...
+            {
+                AssertRead<Vector512<sbyte>>(src, srcStart, sourceLength);
+                Vector512<sbyte> str = Vector512.Load(src).AsSByte();
+
+                // Step 1: Translate encoded Base64 input to their original indices
+                // This step also checks for invalid inputs and exits.
+                // After this, we have indices which are verified to have upper 2 bits set to 0 in each byte.
+                // origIndex      = [...|00dddddd|00cccccc|00bbbbbb|00aaaaaa]
+                Vector512<sbyte> origIndex;
+                if (Avx512Vbmi.IsSupported)
```
EgorBo, line 705: Do we want to duplicate code for missing Avx512 subsets that much? Can we just guard the whole impl to require `Avx512Vbmi` and `Avx512BW` I doubt Base64's perf and large inputs are that important, IMO.

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
```diff
@@ -616,6 +627,137 @@ private static OperationStatus DecodeWithWhiteSpaceFromUtf8InPlace(Span<byte> ut
...
+                    if (errorMask != 0)
+                    {
+                        break;
+                    }
+                }
+                else
+                {
+                    Vector512<sbyte> hiNibbles = Vector512.BitwiseAnd(Avx512F.ShiftRightLogical(str.AsInt32(), 4).AsSByte(), mask2F);
+
+                    Vector512<sbyte> loNibbles = Vector512.BitwiseAnd(str, mask2F);
```
EgorBo, line 718: Here and in all other places - can we prefer `operator`s over static methods? like `vec1 & vec2`, etc.

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
```diff
@@ -68,7 +68,18 @@ private static unsafe OperationStatus DecodeFromUtf8(ReadOnlySpan<byte> utf8, Sp
 
                 if (maxSrcLength >= 24)
                 {
-                    byte* end = srcMax - 45;
+                    byte* end = srcMax - 88;
+                    if (Vector512.IsHardwareAccelerated && Avx512BW.IsSupported && (end >= src))
```
EgorBo, line 72: Afair, `Avx512BW` implies `Vector512.IsHardwareAccelerated` or the latter is needed because the analyzer will complain?

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Encoder.cs
```diff
@@ -226,6 +235,120 @@ public static unsafe OperationStatus EncodeToUtf8InPlace(Span<byte> buffer, int
...
+            // The JIT won't hoist these "constants", so help it
+            // JIT will remove the Non VBMI path constants when VBMI is available and vice-versa.
+
+            // VBMI path constants
+            Vector512<sbyte> shuffleVecVbmi = Vector512.Create(
+                0x01020001, 0x04050304, 0x07080607, 0x0a0b090a,
+                0x0d0e0c0d, 0x10110f10, 0x13141213, 0x16171516,
+                0x191a1819, 0x1c1d1b1c, 0x1f201e1f, 0x22232122,
+                0x25262425, 0x28292728, 0x2b2c2a2b, 0x2e2f2d2e).AsSByte();
+            Vector512<sbyte> vbmiLookup = Vector512.Create(
```
EgorBo, line 260: Can you replace this with `Vector512.Create("ABCDE.....9+/"u8);` ?

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
EgorBo: this also complicates test matrix

DeepakRajendrakumaran: This is a good question and I'm open to it. Would like to hear more input before makinga  call @tannergooding - thoughts?

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
DeepakRajendrakumaran: Vector512.IsHardwareAccelerated is there to weed out the scenario where throttling issues are there. Vector512.IsHardwareAccelerated returns false even if AVX512BW is supported if it has throttling issues like Cascade Lakes

EgorBo: ah, good point

danmoseley:
Do we need a new entry in the third party notices file?

EgorBo:
> Do we need a new entry in the third party notices file?

We should already have them if I am not mistaken (unless avx512 uses a different article)

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Encoder.cs
DeepakRajendrakumaran: Done

danmoseley:
We generally acknowledge significant reuse in the TPN file even if there's a link from the sources. I see https://github.com/dotnet/runtime/blob/17b60e35528ec0f1eba8a6606168412738bade59/THIRD-PARTY-NOTICES.TXT#L345  but I'm not sure it points to that pdf (eg it doesn't include Lemire in the list)

but, whatever @EgorBo recommends..

EgorBo:
I am not an expert in `THIRD-PARTY-NOTICES.TXT` 🙂 but seems like the link to that article is worth adding here https://github.com/dotnet/runtime/blob/main/THIRD-PARTY-NOTICES.TXT#L345

> There is an AVX512 based implementation in there - https://github.com/aklomp/base64/tree/master/lib/arch/avx512. 

Do we use any of the code from that repo? I see it has BSD-2 license

DeepakRajendrakumaran:
> I am not an expert in `THIRD-PARTY-NOTICES.TXT` 🙂 but seems like the link to that article is worth adding here https://github.com/dotnet/runtime/blob/main/THIRD-PARTY-NOTICES.TXT#L345
> 
> > There is an AVX512 based implementation in there - https://github.com/aklomp/base64/tree/master/lib/arch/avx512.
> 
> Do we use any of the code from that repo? I see it has BSD-2 license

Not directly but the logic(including shuffle constants are similar). But this implementation uses '_mm512_multishift_epi64_epi8' and that's not the one I'm using. This(https://github.com/WojciechMula/base64simd/blob/master/encode/encode.avx512vbmi.cpp) in particular is probably worth mentioning in hindsight(Since I used the non multishift version in my implementation with VBMI)

I used the below as resources to understand available implementations and available options(But they are all related). Which brings up the question which of these exactly I should be referencing

esp this for understanding : http://0x80.pl/notesen/2016-04-03-avx512-base64.html#id29

https://github.com/WojciechMula/base64simd/tree/master
https://arxiv.org/pdf/1910.05109.pdf
https://github.com/lemire/fastbase64/blob/master/src/fastavx512bwbase64.c
https://github.com/aklomp/base64/tree/master/lib/arch/avx512

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
DeepakRajendrakumaran: I modified the commit to have just the VBMI path after a conversation with Tanner

DeepakRajendrakumaran:
> > I am not an expert in `THIRD-PARTY-NOTICES.TXT` 🙂 but seems like the link to that article is worth adding here https://github.com/dotnet/runtime/blob/main/THIRD-PARTY-NOTICES.TXT#L345
> > > There is an AVX512 based implementation in there - https://github.com/aklomp/base64/tree/master/lib/arch/avx512.
> > 
> > 
> > Do we use any of the code from that repo? I see it has BSD-2 license
> 
> Not directly but the logic(including shuffle constants are similar). But this implementation uses '_mm512_multishift_epi64_epi8' and that's not the one I'm using. This(https://github.com/WojciechMula/base64simd/blob/master/encode/encode.avx512vbmi.cpp) in particular is probably worth mentioning in hindsight(Since I used the non multishift version in my implementation with VBMI)
> 
> I used the below as resources to understand available implementations and available options(But they are all related). Which brings up the question which of these exactly I should be referencing
> 
> esp this for understanding : http://0x80.pl/notesen/2016-04-03-avx512-base64.html#id29
> 
> https://github.com/WojciechMula/base64simd/tree/master https://arxiv.org/pdf/1910.05109.pdf https://github.com/lemire/fastbase64/blob/master/src/fastavx512bwbase64.c https://github.com/aklomp/base64/tree/master/lib/arch/avx512

@EgorBo I modified the reference to point to https://github.com/WojciechMula/base64simd/tree/master. Which has the closest versions to the implementation I went with. Will this require me adding anything to notice?

DeepakRajendrakumaran:
> > > I am not an expert in `THIRD-PARTY-NOTICES.TXT` 🙂 but seems like the link to that article is worth adding here https://github.com/dotnet/runtime/blob/main/THIRD-PARTY-NOTICES.TXT#L345
> > > > There is an AVX512 based implementation in there - https://github.com/aklomp/base64/tree/master/lib/arch/avx512.
> > > 
> > > 
> > > Do we use any of the code from that repo? I see it has BSD-2 license
> > 
> > 
> > Not directly but the logic(including shuffle constants are similar). But this implementation uses '_mm512_multishift_epi64_epi8' and that's not the one I'm using. This(https://github.com/WojciechMula/base64simd/blob/master/encode/encode.avx512vbmi.cpp) in particular is probably worth mentioning in hindsight(Since I used the non multishift version in my implementation with VBMI)
> > I used the below as resources to understand available implementations and available options(But they are all related). Which brings up the question which of these exactly I should be referencing
> > esp this for understanding : http://0x80.pl/notesen/2016-04-03-avx512-base64.html#id29
> > https://github.com/WojciechMula/base64simd/tree/master https://arxiv.org/pdf/1910.05109.pdf https://github.com/lemire/fastbase64/blob/master/src/fastavx512bwbase64.c https://github.com/aklomp/base64/tree/master/lib/arch/avx512
> 
> @EgorBo I modified the reference to point to https://github.com/WojciechMula/base64simd/tree/master. Which has the closest versions to the implementation I went with. Will this require me adding anything to notice?

Have updated THIRD PARTY NOTICE based on conversation with @tannergooding . Removing fallback avx512Bw path meant I had to add only 2 references.

EgorBo (approved):
Thanks! Looks way simpler now

DeepakRajendrakumaran:
@tannergooding @BruceForstall Any comments on this? I'd be great if we can move this forward this week,

BruceForstall:
I'm not the right person to review this. If @tannergooding can't review, maybe @stephentoub can review (or pick an appropriate reviewer).

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
```diff
@@ -616,6 +627,82 @@ private static OperationStatus DecodeWithWhiteSpaceFromUtf8InPlace(Span<byte> ut
...
+                0x191a1415, 0x1c1d1e18, 0x26202122, 0x292a2425,
+                0x2c2d2e28, 0x36303132, 0x393a3435, 0x3c3d3e38,
+                0x00000000, 0x00000000, 0x00000000, 0x00000000).AsByte();
+
+            Vector512<sbyte> mergeConstant0 = Vector512.Create(0x01400140).AsSByte();
+            Vector512<short> mergeConstant1 = Vector512.Create(0x00011000).AsInt16();
+
+            // This algorithm requires AVX512VBMI support.
+            // Vbmi was first introduced in CannonLake and is avaialable from IceLake on.
+            // This makes it okay to use Vbmi instructions since Vector512.IsHardwareAccelerated returns True only from IceLake on.
```
tannergooding, line 667: This comment isn't quite accurate.

`Vector512.IsHardwareAccelerated` can be made to return `true` for `Skylake-X` and up to before `IceLake` via an environment variable. This is why the caller has the check for `Vector512.IsHardwareAccelerated && Avx512Vbmi.IsSupported` and why this function has `[CompExactlyDependsOn(typeof(Avx512Vbmi))]`

We're fine with it not being usable pre IceLake since those often incur heavier downclocking and its unnecessary complexity for a non-default scenario.

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
```diff
@@ -616,6 +627,82 @@ private static OperationStatus DecodeWithWhiteSpaceFromUtf8InPlace(Span<byte> ut
...
+
+                // Step 1: Translate encoded Base64 input to their original indices
+                // This step also checks for invalid inputs and exits.
+                // After this, we have indices which are verified to have upper 2 bits set to 0 in each byte.
+                // origIndex      = [...|00dddddd|00cccccc|00bbbbbb|00aaaaaa]
+                Vector512<sbyte> origIndex;
+
+                origIndex = Avx512Vbmi.PermuteVar64x8x2(vbmiLookup0, str, vbmiLookup1);
+                Vector512<sbyte> errorVec = (origIndex.AsInt32() | str.AsInt32()).AsSByte();
+                if (errorVec.ExtractMostSignificantBits() != 0)
```
tannergooding, line 681: Are we specifically only caring about the most significant bits? Is this then more efficient than `errorVec != Vector512<sbyte>.Zero`?

-- I expect we're currently generating something like `kmovq; kortest` here, right?

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
```diff
@@ -616,6 +627,82 @@ private static OperationStatus DecodeWithWhiteSpaceFromUtf8InPlace(Span<byte> ut
...
+                // multiAdd1: [...|0000cccc|ccdddddd|0000aaaa|aabbbbbb]
+                Vector512<short> multiAdd1 = Avx512BW.MultiplyAddAdjacent(origIndex.AsByte(), mergeConstant0);
+                // multiAdd1: [...|00000000|aaaaaabb|bbbbcccc|ccdddddd]
+                Vector512<int> multiAdd2 = Avx512BW.MultiplyAddAdjacent(multiAdd1, mergeConstant1);
+
+                // Step 3: Pack 48 bytes
+                str = Avx512Vbmi.PermuteVar64x8(multiAdd2.AsByte(), vbmiPackedLanesControl).AsSByte();
+
+                AssertWrite<Vector512<sbyte>>(dest, destStart, destLength);
+                Vector512.Store(str.AsByte(), dest);
```
tannergooding, line 696: nit: Most other places in the JIT we do `str.Store(dest)` since its an extension method and can be accessed using instance syntax.

tannergooding (approved):
LGTM. Just a request to cleanup a couple minor things.

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
DeepakRajendrakumaran: > Are we specifically only caring about the most significant bits? Is this then more efficient than `errorVec != Vector512<sbyte>.Zero`?
> 
> -- I expect we're currently generating something like `kmovq; kortest` here, right?

Yeah - we only care about the high bit - https://github.com/WojciechMula/base64simd/blob/master/decode/lookup.avx512vbmi.cpp#L53C35-L53C54.

Based on disasm from below(and my recollection), its `vpord `-> `vpmovb2n `-> `kmovq`
image

tannergooding: 👍. Looks like we should probably ensure the pattern of `ExtractMostSignificantBits() != 0` is recognized so we don't have to move to the general purpose register, but otherwise looks as expected.

Would be nice if we had a tracking issue for it.

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
DeepakRajendrakumaran: > nit: Most other places in the JIT we do `str.Store(dest)` since its an extension method and can be accessed using instance syntax.

I might need to understand how isimdVector works a little better :)

image

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
DeepakRajendrakumaran: > 👍. Looks like we should probably ensure the pattern of `ExtractMostSignificantBits() != 0` is recognized so we don't have to move to the general purpose register, but otherwise looks as expected.
> 
> Would be nice if we had a tracking issue for it.

Sounds good - https://github.com/dotnet/runtime/issues/93932

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
```diff
@@ -616,6 +627,82 @@ private static OperationStatus DecodeWithWhiteSpaceFromUtf8InPlace(Span<byte> ut
...
+            // If we have AVX512 support, pick off 64 bytes at a time for as long as we can,
+            // but make sure that we quit before seeing any == markers at the end of the
+            // string. Also, because we write 16 zeroes at the end of the output, ensure
+            // that there are at least 22 valid bytes of input data remaining to close the
+            // gap. 64 + 2 + 22 = 88 bytes.
+            byte* src = srcBytes;
+            byte* dest = destBytes;
+
+            // The JIT won't hoist these "constants", so help it
+            // JIT will remove the Non VBMI path constants when VBMI is available and vice-versa.
```
MihaZupan, line 645:
```suggestion
```

Nit: outdated comment

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
```diff
@@ -616,6 +627,82 @@ private static OperationStatus DecodeWithWhiteSpaceFromUtf8InPlace(Span<byte> ut
...
+                AssertRead<Vector512<sbyte>>(src, srcStart, sourceLength);
+                Vector512<sbyte> str = Vector512.Load(src).AsSByte();
+
+                // Step 1: Translate encoded Base64 input to their original indices
+                // This step also checks for invalid inputs and exits.
+                // After this, we have indices which are verified to have upper 2 bits set to 0 in each byte.
+                // origIndex      = [...|00dddddd|00cccccc|00bbbbbb|00aaaaaa]
+                Vector512<sbyte> origIndex;
+
+                origIndex = Avx512Vbmi.PermuteVar64x8x2(vbmiLookup0, str, vbmiLookup1);
```
MihaZupan, line 679:
```suggestion
                Vector512<sbyte> origIndex = Avx512Vbmi.PermuteVar64x8x2(vbmiLookup0, str, vbmiLookup1);
```

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Encoder.cs
```diff
@@ -226,6 +235,82 @@ public static unsafe OperationStatus EncodeToUtf8InPlace(Span<byte> buffer, int
...
+            // Reference for VBMI implementation : https://github.com/WojciechMula/base64simd/tree/master/encode
+            // If we have AVX512 support, pick off 48 bytes at a time for as long as we can.
+            // But because we read 64 bytes at a time, ensure we have enough room to do a
+            // full 64-byte read without segfaulting.
+
+            byte* src = srcBytes;
+            byte* dest = destBytes;
+
+            // The JIT won't hoist these "constants", so help it
+            // JIT will remove the Non VBMI path constants when VBMI is available and vice-versa.
```
MihaZupan, line 252:
```suggestion
```

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Encoder.cs
```diff
@@ -226,6 +235,82 @@ public static unsafe OperationStatus EncodeToUtf8InPlace(Span<byte> buffer, int
...
+            AssertRead<Vector256<sbyte>>(src, srcStart, sourceLength);
+
+            // This algorithm requires AVX512VBMI support.
+            // Vbmi was first introduced in CannonLake and is avaialable from IceLake on.
+            // This makes it okay to use Vbmi instructions since Vector512.IsHardwareAccelerated returns True only from IceLake on.
+
+            // str = [...|PONM|LKJI|HGFE|DCBA]
+            Vector512<sbyte> str = Vector512.Load(src).AsSByte();
+
+            while (true)
```
MihaZupan, line 274: Could this be written as a do-while loop instead to avoid duplicating the read logic?

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
DeepakRajendrakumaran: I'm not sure I fully understand how this works

image

tannergooding: It's likely conflicting because `dest` is a `byte*` while `str` is a `Vector512<sbyte>` and so it can't resolve

`str.Store((sbyte*)dest)` should fix it, or `str.AsByte().Store(dest)`. The former is less IL, most notably.

DeepakRajendrakumaran: Ah..I messed up and was using `str.AsSbyte().Store(dest)` It's fixed now. Thank you

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
DeepakRajendrakumaran: Done

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
DeepakRajendrakumaran: Removed line 667

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Decoder.cs
DeepakRajendrakumaran: Do

> 

Done

src/libraries/System.Private.CoreLib/src/System/Buffers/Text/Base64Encoder.cs
DeepakRajendrakumaran: > Could this be written as a do-while loop instead to avoid duplicating the read logic?

Decided to stick to similar structure to avx2Encode - only the specific instructions used vary. But open to considering this

MihaZupan: Looks like in the case of `Avx2Encode`, the first load is actually different, so it makes sense to move it out of the loop there.

But it doesn't matter much either way, just a minor style preference since it removes a few redundant lines.

Issue #92202, opened 2023-09-18
Test failure Interop\\COM\\Dynamic\\Dynamic\\Dynamic.cmd


v-wenyuxu:
**Failed in:** [runtime-coreclr jitstress-random 20230917.1](https://dev.azure.com/dnceng-public/public/_build/results?buildId=409203&view=ms.vss-test-web.build-test-results-tab&runId=8894348&resultId=116613&paneView=debug)

**Failed tests:**
```
coreclr windows x64 Checked jitstress_random_2 @ Windows.10.Amd64.Open
    - Interop\\COM\\Dynamic\\Dynamic\\Dynamic.cmd
```

**Error message:**
```

Return code:      1
Raw output file:      C:\h\w\AF9F091C\w\AE880999\uploads\Reports\Interop.COM\Dynamic\Dynamic\Dynamic.output.txt
Raw output:
BEGIN EXECUTION
Copying 'C:\h\w\AF9F091C\p\CoreShim.dll'...
        1 file(s) copied.
 "C:\h\w\AF9F091C\p\corerun.exe" -p "System.Reflection.Metadata.MetadataUpdater.IsSupported=false" -p "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization=true"  Dynamic.dll 
Running BasicTest
Test Failure: System.NotSupportedException: Specified method is not supported.
   at System.Runtime.InteropServices.Variant.GetAccessor(VarEnum varType) in /_/src/libraries/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/ComInterop/Variant.Extended.cs:line 244
   at Microsoft.CSharp.RuntimeBinder.ComInterop.VariantBuilder.InitializeArgumentVariant(MemberExpression variant, Expression parameter) in /_/src/libraries/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/ComInterop/VariantBuilder.cs:line 80
   at Microsoft.CSharp.RuntimeBinder.ComInterop.ComInvokeBinder.GenerateTryBlock() in /_/src/libraries/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/ComInterop/ComInvokeBinder.cs:line 245
[...]

PR #50669, opened 2021-04-03
Fix contained LCL_VAR_ADDR in RMW.

sandreenko:
Call `emitIns_S_I` (stack immediate) to generate instruction like:
```
Generating: N008 (  1,  1) [000005] -c----------         t5 =    CNS_INT   long   123 REG NA $200
                                                              /--*  t4     long
                                                              +--*  t5     long
Generating: N010 (  8,  9) [000006] -c--G-------         t6 = *  ADD       long   REG NA <l:$241, c:$240>
Generating: N012 (  3,  4) [000000] Uc-----N----         t0 =    LCL_VAR_ADDR byref  V00 arg0          NA REG NA
                                                              /--*  t0     byref
                                                              +--*  t6     long
Generating: N014 (???,???) [000012] -A--G-------              *  STOREIND  long   REG NA

IN0001: 000009 add      qword ptr [rsp+08H], 123
```

Questions that I had:
1. should not it be fixed in `emitHandleMemOp`?
no, because `emitHandleMemOp` already has an instruction descriptor and it is wrong for this situation because it expects `A` meaning `dereference Address in a register`. We could change the fmt inside like we do for:
https://github.com/dotnet/runtime/blob/5265305d3121d2fce804948b9250b1b3a5f8a22d/src/coreclr/jit/emitxarch.cpp#L2931-L2932
but it would require a copy-paste of `emitIns_S_I` into this function.
Another argument is that we already treat `LCL_VAR_ADDR/LCL_FLD_ADDR` differently, to show such cases I have changed `OperIs(GT_LCL_VAR_ADDR, GT_LCL_FLD_ADDR)` to `OperIsLocalAddr()`.

A better long-term solution would be to rewrite `emitHandleMemOp` to create `id` but we don't have time for it now.

2. Do we need to call `genUpdateLife` for the local? 
It appears not to be necessary because the variable liveness does not change there so we can skip it.
However, we can do this, as you see `000000` is marked `VAR_USEASG` so call to `genUpdateLife(storeInd)` will give us the correct result, `UpdateLifeVar` just won't do anything with it. I could still add these calls if I hear an opinion that it will be less confusing.

3. Have I caught all places that need this fix? 
I checked all calls to `emitHandleMemOp` and found 1 more place that was fixed in this PR, others already have it or don't work with contained LCL_VAR_ADDR.

Fixes #41073 and #49880.

No spmi-diffs.

Issue #41073, opened 2020-08-20
contained `LCL_VAR_ADDR` is generated wrong in RMW.


sandreenko:
when we have a tree like:
```
N009 ( 15, 17) [000008] -A--G---R---              *  ASG       long   $VN.Void
N008 (  6,  7) [000007] *------N----              +--*  IND       long   <l:$241, c:$240>
N007 (  3,  5) [000001] ------------              |  \--*  LCL_VAR_ADDR   long V00 arg0          $100
N005 (  8,  9) [000006] ----G-------              \--*  ADD       long   <l:$241, c:$240>
N003 (  6,  7) [000004] *---G-------                 +--*  IND       long   <l:$180, c:$1c0>
N002 (  3,  5) [000003] ------------                 |  \--*  LCL_VAR _ADDR      long V00 arg0          $100
N004 (  1,  1) [000005] ------------                 \--*  CNS_INT   long   123 $200
```
jit recognizes it as `RWM` and mark `LCL_VAR_ADDR` as contained but then `emitter::emitInsRMW` calls `emitHandleMemOp` and it has no idea how to generate it right.

The current version fails with asserts added in https://github.com/dotnet/runtime/pull/39424/commits/cfd544a1176d51886156b56fbe21e0f1800ae685, the preview 5(release)  and 3.1 generate code like:
```
Generating: N012 (  3,  4) [000000] Uc-----N----         t0 =    LCL_VAR_ADDR byref  V00 arg0          NA REG NA
                                                              /--*  t0     byref
                                                              +--*  t6     long
Generating: N014 (???,???) [000012] -A--G-------              *  STOREIND  long   REG NA
IN0001:        add      qword ptr [0000H], 123
```
that triggers a null-dereference exception.

It can be generated by code like this:
```
Unsafe.As<float, int>(ref a) = Unsafe.As<float, int>(ref a) + 123;
```
[The full test source](https://github.com/dotnet/runtime/files/5100957/Runtime_39581.il.txt)

I think it doesn't hit the bar for 5.0, so I set 6.0 milestone unless I find more ways to expose this.

P.S. it took me 4 hours to create an IL project for 3.1 and I ended up using `ILGenerator` to generate the method in `csproj` because `ilproj` did not want to work, does anybody know how to create and run ilproj using release dotnet? If not please upvote https://developercommunity.visualstudio.com/idea/540810/add-microsoftnetsdkil-aka-ilproj-support-in-visual.html

<details>
[...]

PR #92252, opened 2023-09-18
JIT: add missing xarch RMW case

### LINKED ISSUE #92218: Incorrect NRE during operations on structures in method marked with AgressiveOptimization
### Description

Testing our code on .NET 8 we encountered getting NRE where it is not supposed to be. I managed to write a minimal sample for the test (below). The described behavior is repeated on Preview.7, RC.1, RC.2 (haven't tried other versions, but it works fine on .NET 7)

Apparently, the problem is in the generated JIT code, because with optimization enabled, the method generates an explicit reference to address 0.

```assembly
; Method MutableStructs.Program:TestMethod() (FullOpts)
       sub      rsp, 56
       xor      eax, eax
       mov      qword ptr [rsp+0x30], rax
       mov      qword ptr [rsp+0x28], rax

G_M000_IG02:                ;; offset=0x0010
       mov      qword ptr [rsp+0x30], 420
       mov      qword ptr [rsp+0x28], 42
       mov      ecx, 3
       call     [System.TimeSpan:op_UnaryNegation(System.TimeSpan):System.TimeSpan]
       mov      rcx, qword ptr [rsp+0x30]
       cmp      rcx, qword ptr [rsp+0x28]
       jl       SHORT G_M000_IG04
       align    [7 bytes for IG03]

G_M000_IG03:                ;; offset=0x0040
       add      qword ptr [0x0000], rax ;; <- NRE
       mov      rcx, qword ptr [rsp+0x30]
       cmp      rcx, qword ptr [rsp+0x28]
       jge      SHORT G_M000_IG03

G_M000_IG04:                ;; offset=0x0054
       add      rsp, 56
       ret      
; Total bytes of code: 89
```

image

### Reproduction Steps

```csharp

namespace MutableStructs;

public struct MutableStruct
{
    private long _internalValue;

    public long InternalValue
    {
        get => Volatile.Read(ref _internalValue);
        private set => Volatile.Write(ref _internalValue, value);
    }

    public void Add(long value) => AddInternal(value);
    private void AddInternal(long value) => InternalValue += value;
    public MutableStruct(long value) => InternalValue = value;
}

internal static class Program
{
    private static void Main() => TestMethod();

    [MethodImpl(MethodImplOptions.AggressiveOptimization)]
    private static void TestMethod()
    {
        var test = new MutableStruct(420);
        var from = new MutableStruct(42);

        var wrapper = -new TimeSpan(3);

        while (test.InternalValue >= from.InternalValue)
        {
            test.Add(wrapper.Ticks);  // <- NRE
        }
    }
}
```

### Expected behavior

The code is successfully executed

### Actual behavior

`Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at MutableStructs.Program.TestMethod()`

### Regression?

At least since .NET 7

### Known Workarounds

The behavior described is quite sensitive to a combination of factors. Small changes will normalize the situation: you can remove `Volatile`, or make `InternalValue`  an auto property, or not wrap `Add`/`AddInternal` calls, or use your own structure instead of `TimeSpan` - all of the above will remove the issue

### Configuration

.NET 8 RC.2
Windows 10 Pro 22H2 x64
12th Gen Intel(R) Core(TM) i9-12900K

### Other information

_No response_

AndyAyersMS:
Handle the case where we're indirectly updating a local with a value that is not a constant.

Fixes #92218.

### DIFF
--- a/src/coreclr/jit/emitxarch.cpp
+++ b/src/coreclr/jit/emitxarch.cpp
@@ -5485,6 +5485,13 @@ void emitter::emitInsRMW(instruction ins, emitAttr attr, GenTreeStoreInd* storeI
     {
         assert(!src->isContained()); // there must be one non-contained src
 
+        if (addr->isContained() && addr->OperIs(GT_LCL_ADDR))
+        {
+            GenTreeLclVarCommon* lclVar = addr->AsLclVarCommon();
+            emitIns_S_R(ins, attr, src->GetRegNum(), lclVar->GetLclNum(), lclVar->GetLclOffs());
+            return;
+        }
+
         // ind, reg
         id = emitNewInstrAmd(attr, offset);
         emitHandleMemOp(storeInd, id, emitInsModeFormat(ins, IF_ARD_RRD), ins);
--- a/src/tests/JIT/Regression/JitBlue/Runtime_92218/Runtime_92218.cs
+++ b/src/tests/JIT/Regression/JitBlue/Runtime_92218/Runtime_92218.cs
@@ -1,0 +1,40 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using Xunit;
+
+public struct MutableStruct
+{
+    private long _internalValue;
+
+    public long InternalValue
+    {
+        get => Volatile.Read(ref _internalValue);
+        private set => Volatile.Write(ref _internalValue, value);
+    }
+
+    public void Add(long value) => AddInternal(value);
+    private void AddInternal(long value) => InternalValue += value;
+    public MutableStruct(long value) => InternalValue = value;
+}
+
+public static class Runtime_92218
+{
+    [Fact]
+    [MethodImpl(MethodImplOptions.AggressiveOptimization)]
+    public static void Problem()
+    {
+        var test = new MutableStruct(420);
+        var from = new MutableStruct(42);
+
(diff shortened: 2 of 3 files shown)


AndyAyersMS:
@BruceForstall PTAL
cc @dotnet/jit-contrib

This is a regression from .NET 7, so we'll need to make the case for a backport.

BruceForstall (approved):
LGTM

Probably should run jitstress

jakobbotsch:
The issue looks like the same as #41073, where it seems we elected not to fix it in `emitHandleMemOp`. Does the reasoning given in its PR make sense for this situation too?

AndyAyersMS:
> The issue looks like the same as #41073, where it seems we elected not to fix it in `emitHandleMemOp`. Does the reasoning given in its PR make sense for this situation too?

My fix is in `emitInsRMW` so is consistent with #50669 (the fix for #41073), which added the `S_I` case.

jitstress failures also happening in main (#92202 ).

AndyAyersMS:
Build analysis doesn't seem to be tracking known jitstress issues, so will have to "merge on red".

jakobbotsch:
> Build analysis doesn't seem to be tracking known jitstress issues, so will have to "merge on red".

The issues with the JSON blobs need to have the "Known Build Error" label to be included in the analysis. I've added it to #92202

AndyAyersMS:
> > Build analysis doesn't seem to be tracking known jitstress issues, so will have to "merge on red".
> 
> The issues with the JSON blobs need to have the "Known Build Error" label to be included in the analysis. I've added it to #92202

Aha. Thanks!

Issue #92262, opened 2023-09-19
[wasm] [debugger] Evaluation of methods with object-param


ilonatommy:
New feature: support passing object parameters to methods evaluated in debugger.

Error: `The given key was not present in the dictionary.`

Sample test:
```

    public static class FastCheck
    {
        public class InstanceClass
        {
            public int number = 123;
        }

        public class MemberClass
        {
            public int Method(InstanceClass ic) => ic.number;
        }

        public static void run()
        {
            InstanceClass ic = new();
            MemberClass mc = new();
            Console.WriteLine("Break here");
        }
    }
```

```
        [Fact]
        public async Task EvaluateMethodsWithObjectParams() => await CheckInspectLocalsAtBreakpointSite(
            $"DebuggerTests.FastCheck", "run", 4, "DebuggerTests.FastCheck.run",
            $"window.setTimeout(function() {{ invoke_static_method ('[debugger-test] DebuggerTests.FastCheck:run'); 1 }})",
            wait_for_event_fn: async (pause_location) =>
            {
                var id = pause_location["callFrames"][0]["callFrameId"].Value<string>();
                await EvaluateOnCallFrameAndCheck(id,
                   ("mc.Method(ic)", TNumber(123)) // ToDo: FAILING
                );
           });
```
and uncommenting the failing test case in `EvaluateMethodsOnEnum` test.

PR #92263, opened 2023-09-19
[wasm] [debugger] Evaluation of methods on enums

### LINKED ISSUE #69299: [wasm] [debugger] Evaluation of methods on enums
Fix evaluation of:

`localEnum.toString()`, and `test.propEnum.ToString()`
and other possible methods called on valueType:enum.

Suggested here: https://github.com/dotnet/runtime/pull/66654#issuecomment-1125547986

ilonatommy:
Fixes https://github.com/dotnet/runtime/issues/69299.

- Adding tests for calling methods on enums.
- Fixing evaluation of `ToString` overload with parameters and potentially other functions that might have empty-named parameter handles in method signature.

Original error:
```
System.IndexOutOfRangeException: Index was outside the bounds of the array
```

Reason:
parameter handle array and `methodSignature.ParameterTypes` array were not in sync, handle array had 3 elements, one of them invalid (with empty name) while method signature types had 2 elements. When we ignore invalid handles, the logic works fine. https://learn.microsoft.com/en-us/dotnet/api/system.enum.tostring

Out of scope:
- passing objects as method arguments. "mc.valueTypeEnum.HasFlag(SampleEnum.no)" is not evaluated correctly because we do not have the mechanism for object-params. Delegated to issue https://github.com/dotnet/runtime/issues/92262

### DIFF
--- a/src/mono/wasm/debugger/DebuggerTestSuite/EvaluateOnCallFrame2Tests.cs
+++ b/src/mono/wasm/debugger/DebuggerTestSuite/EvaluateOnCallFrame2Tests.cs
@@ -742,7 +742,7 @@ namespace DebuggerTests
                 await EvaluateOnCallFrameAndCheck(id,
                    ("s_valueTypeEnum.ToString()", TString("no")),
                    ("mc.valueTypeEnum.ToString()", TString("yes"))
-                   // ("mc.valueTypeEnum.HasFlag(SampleEnum.no)", TBool(true)) // ToDo: FAILING
+                   // ("mc.valueTypeEnum.HasFlag(SampleEnum.no)", TBool(true)) // ToDo: https://github.com/dotnet/runtime/issues/92262
                 );
            });
     }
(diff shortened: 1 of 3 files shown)


src/mono/wasm/debugger/DebuggerTestSuite/EvaluateOnCallFrame2Tests.cs
```diff
@@ -731,5 +731,19 @@ public async Task EvaluateLocalObjectFromAssemblyNotRelatedButLoaded()
                     ("dt+1", "Cannot evaluate '(dt+1\n)': (2,9): error CS0019: Operator '+' cannot be applied to operands of type 'object' and 'int'")
                 );
            });
+
+        [Fact]
+        public async Task EvaluateMethodsOnEnum() => await CheckInspectLocalsAtBreakpointSite(
```
radical, line 736: IIUC, this PR is fixing the case of a type having a `ToString` override with parameters, and isn't specific to enums as such. So, essentially we are trying to invoke the no-params overload to `ToString`.

ilonatommy: I think you are correct. We had other tests with `ToString`:
https://github.com/dotnet/runtime/blob/5883b728c8b29a7026cf3a807a22928cee844bc5/src/mono/wasm/debugger/DebuggerTestSuite/EvaluateOnCallFrame2Tests.cs#L582
https://github.com/dotnet/runtime/blob/5883b728c8b29a7026cf3a807a22928cee844bc5/src/mono/wasm/debugger/DebuggerTestSuite/EvaluateOnCallFrame2Tests.cs#L613
but all of them are no-param method.
The name of test is correct, though and the PR proves that calling methods on enums is possible, so it really fixes the linked issue. I will update the PR's name.

thaystg (approved):

Issue #77536, opened 2022-10-27
[wasm][debugger] Support indexing by `object` schema


ilonatommy:
- [x] indexing with class instance - working, tests added in https://github.com/dotnet/runtime/pull/92268
- [ ] indexing with value type - https://github.com/dotnet/runtime/pull/92334
- [x] indexing with array - https://github.com/dotnet/runtime/pull/92337

thaystg:
This is fixed by the new debugger.

PR #92268, opened 2023-09-19
[wasm][debugger] Add tests for indexing by object schema

ilonatommy:

- Added test for indexing by object `Indexer`: passes without code changes.

Out of scope:
- Indexing with value type and array. Work is ongoing. The original issue will be edited to describe missing features and problems with implementations.


thaystg:
Hum, you added new lines to the test, it looks like it broke other 2 tests:
image

thaystg (approved):

PR #92269, opened 2023-09-19
Fix downlevel build break in TensorPrimitives

stephentoub:


steveisok (approved):

ericstj:
This needs more based on the changes I have in 8.0.  Should we bring those here too?  Maybe it makes sense to include `net6.0` in main so you have coverage of the frameworks building in 8.0.

ericstj (approved):

ericstj:
Hmm, this seems to have added prebuilts.   

>  error : 2 new pre-builts discovered! Detailed usage report can be found at /__w/1/s/artifacts/source-build/self/prebuilt-report/baseline-comparison.xml. See https://aka.ms/dotnet/prebuilts for guidance on what pre-builts are and how to eliminate them. Package IDs are: Microsoft.NETCore.App.Ref.6.0.20 Microsoft.NETCore.App.Ref.7.0.9

Maybe we should avoid building against 6.0 and 7.0 when building in sourcebuild?  @mthalman what do you reccomend?

ericstj:
I see, here's what we do for the other frameworks; https://github.com/dotnet/runtime/blob/d3da653d407bede37c51b3ed7ecac0498ae63e39/Directory.Build.props#L93-L98

I'll condition out the 6.0 and 7.0 so they aren't there for source build.

ericstj (approved):

stephentoub:
Thanks for the additional tweaks, @ericstj.

src/libraries/System.Numerics.Tensors/src/System.Numerics.Tensors.csproj
```diff
@@ -2,6 +2,7 @@
   
   <PropertyGroup>
     <TargetFrameworks>$(NetCoreAppCurrent);$(NetCoreAppPrevious);$(NetCoreAppMinimum);netstandard2.0;$(NetFrameworkMinimum)</TargetFrameworks>
+    <TargetFrameworks Condition="'$(DotNetBuildFromSource)' != 'true'">$(TargetFrameworks);net7.0;net6.0</TargetFrameworks>
```
ViktorHofer, line 5: I just saw this PR. I don't see an underlying issue that is fixed by this change, what was the motivation to add `net7.0` and `net6.0` TFMs to this library?

.NET 9 dotnet/runtime libraries packages support net8.0 (`NetMinimum`) and net9.0 (`NetCurrent`) only.

Looking at the latest package from main, I see that a warning is emitted at consumption time when targeting net6.0 or net7.0: https://dnceng.visualstudio.com/public/_artifacts/feed/dotnet9/NuGet/System.Numerics.Tensors/overview/9.0.0-alpha.1.23519.15

> System.Numerics.Tensors 9.0.0-alpha.1.23519.15 doesn't support $(TargetFramework) and has not been tested with it. Consider upgrading your TargetFramework to net8.0 or later. You may also set &lt;SuppressTfmSupportBuildWarnings&gt;true&lt;/SuppressTfmSupportBuildWarnings&gt; in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk.

PR #92277, opened 2023-09-19
Update Newtonsoft.Json from 13.0.1 to 13.0.3

mthalman:
.NET repos currently have a mix of Newtonsoft.Json versions that we're trying to get consolidated into the latest version. This is particularly necessary for the .NET's [source-build](https://github.com/dotnet/source-build) which can only reference one version of a dependency.

This was missed as part of the initial set of work to update all repos to 13.0.3.

Related to https://github.com/dotnet/source-build-externals/pull/195 and https://github.com/dotnet/source-build/issues/3629

### DIFF
--- a/eng/SourceBuildPrebuiltBaseline.xml
+++ b/eng/SourceBuildPrebuiltBaseline.xml
@@ -16,7 +16,12 @@
     <UsagePattern IdentityGlob="Microsoft.NETCore.App.Crossgen2.linux-x64/*8.*" />
     <UsagePattern IdentityGlob="Microsoft.NETCore.App.Runtime.linux-x64/*8.*" />
     <UsagePattern IdentityGlob="*Microsoft.DotNet.ILCompiler/*8.*" />
-    
+
+    <!-- This version is brought in transitively from NuGet.Packaging.6.2.4.
+         Once a newer version of NuGet.Packaging is referenced which has a
+         dependency on 13.0.3, this can be removed. -->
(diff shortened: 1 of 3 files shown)


ViktorHofer (approved):

ViktorHofer:
This needs to be forward ported into main as well. Btw I noticed that Arcade also needs a bump (in main).

carlossanlop (approved):
Do we need this in RC2 or can it wait? Asking because I have not snapped yet due to various blocking issues, and we could merge this and include it there.

mthalman:
@carlossanlop - Yes, RC2 would be good

mthalman:
Looks like there's a build issue though

ViktorHofer:
@mthalman please create the missing subscriptions (for both 8.0 and main if one of those doesn't exist).

mthalman:
> @mthalman please create the missing subscriptions (for both 8.0 and main if one of those doesn't exist).

Done

mthalman:
> According to CI, Newtonsoft.Json/13.0.1 is still somehow brought in.

This is due to a dependency on NuGet.Packaging.6.2.4 which still references 13.0.1. I'll add a prebuilt exclusion for this. Not an issue in full source-build since the NuGet.Packaging version would get lifted to the live version in that case.

mthalman:
Seems like a flaky test is blocking this.

carlossanlop:
I opened https://github.com/dotnet/runtime/issues/92295 to track the test failure. Merging now.

Issue #83203, opened 2023-03-09
MAUI project with non-ASCII project name cannot release to my Android phone


dellis1972:
### Description

Context https://i.azdo.io/1714603

When a user tries to use certain non-ascii characters for the project name AOT fails with the following error

```
[随机生成器.dll] Running: C:\a\_work\1\s\xamarin-android\bin\Release\dotnet\packs\Microsoft.NETCore.App.Runtime.AOT.win-x64.Cross.android-arm\8.0.0-preview.2.23127.4\Sdk\..\tools\mono-aot-cross.exe --response="C:\Users\cloudtest\AppData\Local\Temp\tmp2we1dw.tmp" (TaskId:285)
                     [随机生成器.dll] Using working directory: C:\a\_work\1\a\TestRelease\03-08_16.56.43\temp\BuildAotApplicationWithSpecialCharactersInProject随机生成器True\随机生成器 (TaskId:285)
                     [随机生成器.dll] Setting environment variables for execution: (TaskId:285)
                     [随机生成器.dll] 	MONO_PATH = C:\a\_work\1\a\TestRelease\03-08_16.56.43\temp\BuildAotApplicationWithSpecialCharactersInProject随机生成器True\随机生成器\obj\Release\android-arm\linked; (TaskId:285)
                     [随机生成器.dll] 	MONO_ENV_OPTIONS =  (TaskId:285)
                     [随机生成器.dll] Exit code: 1 (TaskId:285)
[...]

PR #90436, opened 2023-08-11
[Mono] Fix all the issues related to enabling WasmStripILAfterAOT

fanyang-mono:
This PR enables IL trim by default for WASM apps, by solving the following major problems:

- There are various opcode which was translated into a function call. I modified the `call` goto logic to check if there is an AOT'ed version of the callee method. If there is, invoke that. If there isn't, interp compile the callee method.
- Move the logic of checking if a method is supported a JIT/AOT call from interpreter callback to `mini-runtime.c`. Because the callback was initialized with the desired logic only when `mono_use_interpreter` is set to true. See the following code 
https://github.com/dotnet/runtime/blob/8d4f8b86a742caba0f7640c013a2c1cefe10b146/src/mono/mono/mini/mini-runtime.c#L4629-L4632
However, in `driver.c`, `mono_use_interpreter` was not set to true during AOT compile time, even though `interp` was passed in as an AOT option. It would be set to true during AOT runtime. Alter this behavior is probably not desirable. Thus, I choose to move the checking logic to `mini-runtime.c`
- Set the trimmed assemblies with the same timestamp as the original assemblies. So that it won't trigger unwanted rebuild or re-AOT
- Added implementation of `g_fopen` to properly handle unicode string on windows. 
- Interpreter can't do JIT call for methods which require synchronization. Remove those methods from trimmable method list.
- Disabled tailcall optimization when calling trimmed methods.
- Do jitcall when encountering virtual tailcalls.
- Change of ILStrip task:
  - The trimmed assemblies were saved in a new folder with the same name as the original assembly.
  - Created an output variable `UpdatedAssemblies` with the updated path of assemblies. The trimmed assemblies are updated with their new path.
  - Updated the WASM build toolchain to replace the content of `_WasmAssembliesInternal` with `UpdatedAssemblies`
- Added the input parameter `WasmRuntimeConfigFilePath` for `WasmAppBuilderBaseTask`. Got rid of the logic to calculate the path of `runtimeconfig.json` file. Use the value of `WasmRuntimeConfigFilePath` instead. (Fixes #93692)
- Found two issues related to wasm build process. @radical fixed them.
  - #93680
  - Fixes #93689
- Blazor app is able to packages the trimmed assemblies, when `WasmStripILAfterAOT` is enabled.
- Added tests for wasm template app and blazor template app 

**Library size change for top 3 largest libraries:**
- System.Private.CoreLib.dll.br: 452k -> 316K
- System.Text.Json.dll.br: 109k -> 81k
[...]

PR #92279, opened 2023-09-19
[mono][aot] Fix support for files with non-ascii characters on windows

BrzVlad:
Add g_fopen, g_unlink and g_rename which on windows do a utf8 to utf16 conversion and then call the corresponding wide char api.

Contributes to https://github.com/dotnet/runtime/issues/83203

### DIFF
--- a/src/mono/mono/eglib/gfile.c
+++ b/src/mono/mono/eglib/gfile.c
@@ -105,17 +105,15 @@ g_file_error_from_errno (gint err_no)
 	}
 }
 
-FILE *
-g_fopen (const char *path, const char *mode)
+FILE*
+g_fopen (const gchar *path, const gchar *mode)
 {
 	FILE *fp;
 
 	if (!path)
 		return NULL;
 
-#ifndef HOST_WIN32
-	fp = fopen (path, mode);
-#else
+#ifdef HOST_WIN32
 	gunichar2 *wPath = g_utf8_to_utf16 (path, -1, 0, 0, 0);
 	gunichar2 *wMode = g_utf8_to_utf16 (mode, -1, 0, 0, 0);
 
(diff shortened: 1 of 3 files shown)


src/mono/mono/eglib/gfile.c
```diff
@@ -155,3 +155,63 @@ g_file_set_contents (const gchar *filename, const gchar *contents, gssize length
...
 	return TRUE;
 }
+
+FILE*
+g_fopen (const gchar *path, const gchar *mode)
+{
+	FILE *fp;
+
+#ifdef HOST_WIN32
+	gunichar2 *wPath = g_utf8_to_utf16 (path, -1, 0, 0, 0);
```
fanyang-mono, line 165: One minor optimization that Johan suggested and I wasn't able to do it correctly is that if path only contains ascii characters, it could simply use `fopen`. I tried adding the optimization [here](https://github.com/dotnet/runtime/pull/90436/commits/ab352d4e322531705781727242881989648a4f74). But it failed the test on CI. I haven't figured out how to test it locally. So I took it out. If you could get it right, that would be great!

BrzVlad: I don't see this optimization having any impact in the real world, especially since we are dealing with file access which is bound to be quite slow. Unless the encoding conversion is craazy slow, which I doubt.

fanyang-mono: I don't really have any real world experience with the relevant API's. By looking at the code, we could avoid two calculations of `g_utf8_to_utf16`.
@lateralusX Do you have any insights here?

lateralusX: I have not done any measurements on the speed of encoding conversions on any platform and in relation to file IO its probably a small amount of time. The idea was to avoid doing the heap allocation and encoding conversion when not really needed (all characters in ascii format so standard function can be called) and since we have the logic put into a couple of g_ functions, the optimization will be isolated, simple and straightforward. I still think avoiding string encoding conversions and heap allocations when not needed is worthwhile, but I agree that in this case the frequency of calls coming into these functions will be low.

BrzVlad: @lateralusX I added the ascii string check to avoid utf16 conversion. Could you do a final review on this PR ?

lateralusX (approved):

PR #92284, opened 2023-09-19
Improve throughput / allocations of JsonNode.GetPath

stephentoub:
The current implementation is creating a `List<string>` and appending each segment to it, which in most of the cases is allocating a `string`. Then it iterates through that list in reverse order appending to a newly-created `StringBuilder`, which it then `ToString`s. In this change, it instead just uses `ValueStringBuilder`, appending to it as it goes.

In doing so, it does reverse the order of enumeration. Previously each node would effectively do:
```C#
void GetPath()
{
    AddNode(this);
    parent?.GetPath();
}
```
and now it's doing:
```C#
void GetPath()
{
    parent?.GetPath();
    AddNode(this);
}
```
While C# doesn't emit tail calls, with optimizations enabled, it's feasible the JIT might emit the recursive call as a jmp rather than a call, in which case it would avoid possible stack dives. However, that's not guaranteed, and doesn't happen today in tier 0 and other unoptimized code. On top of that, to get such a deep nesting in a JsonNode, you need to either go out of your way to create one manually using the JsonNode/Object/Array/Value constructors, or you need to use JsonSerializer.Deserializer, overriding its default MaxDepth, and in the case of a really deep input, it's also recursive and will stack overflow in smaller situations. I have a different version of this change that keeps the same ordering, passing around a span and a length separately, and prepending to the end of the span, but it results in more complicated code, so I'd prefer this variation that just uses ValueStringBuilder unless we have real concerns.

```C#
using System.Text.Json.Nodes;

[MemoryDiagnoser(false)]
public class Tests
{
    private JsonNode _json = JsonNode.Parse("""
        {
          "employees": {
            "Jane": {
              "age": 30,
              "position": "developer"
            }
          }
        }
        """)["employees"]["Jane"]["position"];

    [Benchmark]
    public string GetPath() => _json.GetPath();
}
```

| Method  | Toolchain         | Mean     | Ratio | Allocated | Alloc Ratio |
|-------- |------------------ |---------:|------:|----------:|------------:|
| GetPath | \main\corerun.exe | 222.3 ns |  1.00 |     488 B |        1.00 |
| GetPath | \pr\corerun.exe   | 113.8 ns |  0.51 |      72 B |        0.15 |


stephentoub:
@eiriktsarpalis, mind taking a look when you get a chance?

src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonArray.cs
```diff
@@ -202,15 +202,25 @@ internal void SetItem(int index, JsonNode? value)
...
             if (child != null)
             {
                 int index = List.IndexOf(child);
-                path.Add($"[{index}]");
+                Debug.Assert(index >= 0);
+
+                path.Append('[');
+#if NETCOREAPP
+                Span<char> chars = stackalloc char[JsonConstants.MaximumFormatUInt32Length];
+                ((uint)index).TryFormat(chars, out int charsWritten);
```
eiriktsarpalis, line 217: Should we be asserting that TryFormat returns true?

eiriktsarpalis (approved):

PR #92286, opened 2023-09-19
Fix Excessive Encoding in Test Logs

### LINKED ISSUE #92092: Runtime test output transcripts in the xml result files use excessive escaping rendering them basically unreadable
The result xml file generated by the new-style merged test wrappers includes stdout / stderr transcript for each test. I believe @jkoritzinsky initially implemented this functionality with the aim to help developers directly see the output transcript from the tests as an aid in issue investigation. At some point we modified this logic so that the output transcript is so heavily escaped it's basically unreadable, e.g. here:

<pre>
<test name="readytorun\determinism\crossgen2determinism\crossgen2determinism.cmd" type="OutOfProcessTest" method="readytorun\determinism\crossgen2determinism\crossgen2determinism.cmd" time="11.268134" result="Pass">
<output>
<![CDATA[ _x000D__x000A_Return_x0020_code:_x0020__x0020__x0020__x0020__x0020__x0020_0_x000D__x000A_Raw_x0020_output_x0020_file:_x0020__x0020__x0020__x0020__x0020__x0020_D:_x005C_git_x005C_runtime3_x005C_artifacts_x005C_tests_x005C_coreclr_x005C_windows.x64.Checked_x005C_readytorun_x005C_determinism_x005C_crossgen2determinism_x005C_output.txt_x000D__x000A_Raw_x0020_output:_x000D__x000A_BEGIN_x0020_EXECUTION_x000D__x000A__x000D__x000A_D:_x005C_git_x005C_runtime3_x005C_artifacts_x005C_tests_x005C_coreclr_x005C_windows.x64.Checked_x005C_readytorun_x005C_determinism_x005C_crossgen2determinism_x003E_set_x0020_DOTNETCLI_x003D_D:_x005C_git_x005C_runtime3_x005C_dotnet.cmd_x0020__x000D__x000A_Emitting_x0020_R2R_x0020_PE_x0020_file:_x0020_crossgen2smoke1.ildll_x000D__x000A_Emitting_x0020_map_x0020_file:_x0020_crossgen2smoke1.map_x000D__x000A_Emitting_x0020_R2R_x0020_PE_x0020_file:_x0020_crossgen2smoke2.ildll_x000D__x000A_Emitting_x0020_map_x0020_file:_x0020_crossgen2smoke2.map_x000D__x000A__x0020__x0022_D:_x005C_git_x005C_runtime3_x005C_artifacts_x005C_tests_x005C_coreclr_x005C_windows.x64.Checked_x005C_Tests_x005C_Core_Root_x005C_corerun.exe_x0022__x0020_-p_x0020__x0022_System.Reflection.Metadata.MetadataUpdater.IsSupported_x003D_false_x0022__x0020_-p_x0020__x0022_System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization_x003D_true_x0022__x0020__x0020_crossgen2determinism.dll_x0020__x000D__x000A_Files_x0020_of_x0020_length_x0020_139264_x0020_were_x0020_identical._x000D__x000A_Expected:_x0020_100_x000D__x000A_Actual:_x0020_100_x000D__x000A_END_x0020_EXECUTION_x0020_-_x0020_PASSED_x000D__x000A_PASSED_x000D__x000A_Test_x0020_Harness_x0020_Exitcode_x0020_is_x0020_:_x0020_0_x000D__x000A_To_x0020_run_x0020_the_x0020_test:_x000D__x000A_Set_x0020_up_x0020_CORE_ROOT_x0020_and_x0020_run._x000D__x000A__x003E__x0020_D:_x005C_git_x005C_runtime3_x005C_artifacts_x005C_tests_x005C_coreclr_x005C_windows.x64.Checked_x005C_readytorun_x005C_readytorun_x005C_.._x005C_determinism_x005C_crossgen2determinism_x005C_crossgen2determinism.cmd_x000D__x000A_ ]]>

</output>
</test>
</pre>

We seem to be escaping totally benign characters including spaces, CR/LF pairs or backslashes, increasing the size of the result files and reducing their value for developers, frankly speaking it would probably take me half an hour or so to turn the above into something reasonable. Jeremy speculated that we may have broken this with the xml checking / completion that was part of the Merge on Red project, assigning to Ivan to take a first look and we'll follow from there.

Thanks

Tomas

ivdiazsa:
Fixes #92092. At some point during the refactoring of the test logs generation systems, we added some encoding to the tests' recorded outputs. However, this encoding replaced every non-alphanumeric character with a hex code, including the most common ones like brackets, parentheses, and spaces. This, in turn, rendered the logs visually unappealing and very difficult to read, which complicated all test investigations. This PR fixes that and makes the logs easily readable again.

### DIFF
--- a/src/tests/Common/XUnitWrapperLibrary/TestSummary.cs
+++ b/src/tests/Common/XUnitWrapperLibrary/TestSummary.cs
@@ -2,9 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
-using System.IO;
 using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 using System.Text;
+using System.Xml;
+
 namespace XUnitWrapperLibrary;
 
 public class TestSummary
@@ -42,8 +45,13 @@ public class TestSummary
             testResultSb.Append($@"<test name=""{Name}"" type=""{ContainingTypeName}"""
                               + $@" method=""{MethodName}"" time=""{Duration.TotalSeconds:F6}""");
 
+            // It is possible for tests to have illegal XML characters in their output.
+            // So, we have to sanitize said output before writing it down to the XML log
+            // because otherwise, the python XML parser crashes. The way we clean it
+            // is by replacing said characters with their hexadecimal notation.
+
             string outputElement = !string.IsNullOrWhiteSpace(Output)
-                                 ? $"<output><![CDATA[{Output}]]></output>"
+                                 ? $"<output><![CDATA[{SanitizeOutput(Output)}]]></output>"
                                  : string.Empty;
 
             if (Exception is not null)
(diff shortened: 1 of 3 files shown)


trylek (approved):
Looks great to me, thanks Ivan for fixing this!

src/tests/Common/XUnitWrapperLibrary/TestSummary.cs
```diff
@@ -95,12 +102,66 @@ public string ToXmlString()
 
             return testResultSb.ToString();
         }
+
+        private string SanitizeOutput(string output)
+        {
+            Dictionary<char, string>? invalidChars = FindInvalidXMLChars(output);
```
danmoseley, line 108: Will this do something reasonable for surrogate pairs? They tend to appear in test output because they are interesting test inputs.

src/tests/Common/XUnitWrapperLibrary/TestSummary.cs
```diff
@@ -95,12 +102,66 @@ public string ToXmlString()
...
+
+            // All green, we can print the test's output as is.
+            if (invalidChars is null)
+                return output;
+
+            StringBuilder sanitizedOutput = new StringBuilder();
+            foreach (char ch in output)
+            {
+                string? invalidCharHexCode = string.Empty;
```
danmoseley, line 117: Why not just do a Replace on the string for each of the problematic characters rather than walking the string yourself? You know the string you are replacing with does not contain any invalid characters

danmoseley: Although I'm not sure why you don't just walk the steering checking each char is valid and replacing it on the fly rather than building up a dictionary

ivdiazsa: > Why not just do a Replace on the string for each of the problematic characters rather than walking the string yourself? You know the string you are replacing with does not contain any invalid characters

`String.Replace()` always creates a new string so it's inefficient to be creating new ones for each problematic character, especially considering some tests might have large outputs, and we have lots of tests to process.

ivdiazsa: > Although I'm not sure why you don't just walk the string checking each char is valid and replacing it on the fly rather than building up a dictionary first?

I did it to just return the output as is, in case there were no problematic characters, but yeah it might be simpler as you suggest. Will try it right now.

danmoseley: > String.Replace() always creates a new string so it's inefficient to be creating new ones for each problematic character

Right, for several replacements you want StringBuilder.Replace.

PR #92292, opened 2023-09-19
JIT: Make a quirk in block morphing more explicit

jakobbotsch:
Block morphing would create oddly typed trees (mixing up TYP_BYREF/TYP_I_IMPL, e.g. creating LCL_VAR<I_IMPL> for a TYP_BYREF typed local). The only effect of this was that it would avoid some constant propagation. Make this more explicit by setting GTF_DONT_CSE instead.


jakobbotsch:

[Minor handful of diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=411762&view=ms.vss-build-web.run-extensions-tab). Mostly in ARM64 where we in some cases did get some constant CSEs before that are now disabled. I don't think those CSEs are worth producing this questionable IR to avoid constant propagation in other cases.

BruceForstall (approved):

PR #92296, opened 2023-09-19
Build the PGO build legs through the global build job template

jkoritzinsky:
Contributes to moving our official builds to using single jobs per runtime + configuration + platform combo.

Also, remove the support for PGO instrumentation in the individual build jobs as we're only going to use it through the global build job in this change.

Official build at: https://dev.azure.com/dnceng/internal/_build/results?buildId=2271602&view=results


hoyosjs (approved):
Only question - does it preserve teh package names correctly?

jkoritzinsky:
Yep! The resulting package names are identical to the prior implementation.

PR #92300, opened 2023-09-19
HashSet.UnionWith preformance improvement

### LINKED ISSUE #92206: [API Proposal]: HashSet<T>.UnionWith should copy data from another HashSet<T>
### Background and motivation

I'll describe my specific case
```csharp
        
         HashSet<ITarget> _visibleTargetsSet = new HashSet<ITarget>();
         HashSet<ITarget> _registerdTargetsSet = new HashSet<ITarget>();
         HashSet<ITarget> _safeTargetsSet = new HashSet<ITarget>();
         List<ITarget> _visibleTargets = new List<ITarget>();

        private void SetupVisibleTargets()
        {
            _visibleTargetsSet.Clear();
            _visibleTargetsSet.UnionWith(_registeredTargetsSet);
            _visibleTargetsSet.ExceptWith(_safeTargetsSet);

            _visibleTargets.Clear();
            _visibleTargets.AddRange(_visibleTargetsSet);

            _safeTargetsSet.Clear();
        }
```
In this example, I find the difference between the sets of registered and visible targets. _safeTargetsSet is calculated somewhere else, and registerd targets are needed to stay immutable. I don't want to create new collections, so I use clear and use _safeTargetsSet every time i need a new data for _visibleTargets. The problem is that method UnionWith is just adding every item separately with coolision checks even if count is 0, unification occurs with HashSet<T> and comparers are equal.

### API Proposal

```csharp

public void UnionWith(IEnumerable<T> other)
        {
            if (other == null)
            {
                throw new ArgumentNullException("other");
            }
            HashSet<T> hashSet = other as HashSet<T>;
            if (hashSet != null && Count == 0 && AreEqualityComparersEqual(this, hashSet))
            {
                // Current method ConstructFrom could be used as reference
                CopyFrom(hashSet);
                return;
            }
            foreach (T item in other)
            {
                AddIfNotPresent(item);
            }
        }
```

### API Usage

```csharp

HashSet<int> c3 = new HashSet<int>();
...

void SomeMethod()
{
    var c1 = new HashSet<int>();
    var c2 = new HashSet<int>();
    for(int i = 0; i < 100; i ++)
        c1.Add(i);
    for(int i = 0; i < 50; i ++)
        c2.Add(i);
    c3.Clear();
    c3.UnionWith(c1);
    c3.ExceptWith(c2);
}
```

### Alternative Designs

It can be done with method overloading, not a current method change.

### Risks

Idk

Artromskiy:
Draft commit


Artromskiy:
This PR is related to issue #92206

eiriktsarpalis:
@Artromskiy something seems to be wrong with your PR branch -- the diff looks huge.

Artromskiy:
> @Artromskiy something seems to be wrong with your PR branch -- the diff looks huge.

Squashed and merged from master

Artromskiy:
@eiriktsarpalis  what should I do? Create a new PR or revert merged from master?

eiriktsarpalis:
It seems bce6e64c5c5b9059b55480991de3fad2b6b8ee01 was the culrpit. Perhaps reverting that might help.

eiriktsarpalis:
@Artromskiy I've squashed and rebased your changes to a single commit, which should get rid of some of the earlier merge artifacts. Is this now ready for review?

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/HashSet.cs
```diff
@@ -1068,6 +1056,69 @@ private int Initialize(int capacity)
...
+        private bool IsTrimNeeded()
+        {
+            return _count > 0 && _entries!.Length / _count > ShrinkThreshold;
+        }
+
+        /// <summary>
+        /// Initializes the HashSet from another HashSet with the same element type and equality comparer
+        /// </summary>
+        /// <param name="source"></param>
+        private void ConstructFrom(HashSet<T> source)
```
eiriktsarpalis, line 1072: Why has this method been moved from this original location? It's more difficult to review any changes that might have been made to it

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/HashSet.cs
```diff
@@ -1007,39 +982,52 @@ private void Resize(int newSize, bool forceNewHashCodes)
         /// </summary>
         public void TrimExcess()
         {
-            int capacity = Count;
-
-            int newSize = HashHelpers.GetPrime(capacity);
-            Entry[]? oldEntries = _entries;
-            int currentCapacity = oldEntries == null ? 0 : oldEntries.Length;
-            if (newSize >= currentCapacity)
+            int newCapacity = HashHelpers.GetPrime(_count);
```
eiriktsarpalis, line 985: This looks like an unrelated change?

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/HashSet.cs
```diff
@@ -110,7 +110,7 @@ public HashSet(IEnumerable<T> collection, IEqualityComparer<T>? comparer) : this
 
                 UnionWith(collection);
 
-                if (_count > 0 && _entries!.Length / _count > ShrinkThreshold)
+                if (IsTrimNeeded())
```
eiriktsarpalis, line 113: This looks like an unrelated change. Also, why does this need to be extracted to a helper method?

Artromskiy:
> @Artromskiy I've squashed and rebased your changes to a single commit, which should get rid of some of the earlier merge artifacts. Is this now ready for review?

I'm not sure it's passing tests. It only been used in my local project

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/HashSet.cs
Artromskiy: This check used in Initialization (when hashset created with another collection) and in ConstructFrom. I moved this to method to unify this check

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/HashSet.cs
Artromskiy: I changed Construct from method, as it always expanded prime to be larger than original collection

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/HashSet.cs
Artromskiy: Construct From moved from it's original location as it's not an initialization method anymore. Also CopyEntries separated from TrimExcess as it's needed by ConstructFrom

eiriktsarpalis: It doesn't technically need to be moved though. Keeping it at its originally location has the benefit of being easier to review and better preserving edit history.

Artromskiy: Got it, it won't happen again

stephentoub: Can you please move it back / revert it?

PR #92302, opened 2023-09-19
Use JsonSerializer to implement JsonNode.To{Json}String

stephentoub:
It already does all the caching and object reuse we want here.  Only thing we need to do here is ensure we're using appropriate options, which we lazily initialize.

```C#
using System.Text.Json.Nodes;

[MemoryDiagnoser(false)]
public class Tests
{
    private JsonNode _json = JsonNode.Parse("""
        {
          "employees": [
              {
                "Jane": {
                  "age": 30,
                  "position": "developer"
                }
              },
              {
                "John": {
                  "age": 30,
                  "position": "designer"
                }
              }]
        }
        """);

    [Benchmark] public new string ToString() => _json.ToString();
    [Benchmark] public string ToJsonString() => _json.ToJsonString();
}
```

| Method       | Toolchain         | Mean     | Ratio | Allocated | Alloc Ratio |
|------------- |------------------ |---------:|------:|----------:|------------:|
| ToString     | \main\corerun.exe | 618.5 ns |  1.00 |     600 B |        1.00 |
| ToString     | \pr\corerun.exe   | 577.5 ns |  0.95 |     448 B |        0.75 |
|              |                   |          |       |           |             |
| ToJsonString | \main\corerun.exe | 462.7 ns |  1.00 |     376 B |        1.00 |
| ToJsonString | \pr\corerun.exe   | 429.6 ns |  0.93 |     224 B |        0.60 |

### DIFF
--- a/src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonNode.To.cs
+++ b/src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonNode.To.cs
@@ -1,10 +1,21 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Threading;
+
 namespace System.Text.Json.Nodes
 {
     public abstract partial class JsonNode
     {
+        /// <summary>Lazily-initialized options that set the <see cref="JsonSerializerOptions.MaxDepth"/> to match the <see cref="JsonWriterOptions"/> default.</summary>
+        private static JsonSerializerOptions? s_unindentedToStringOptions;
+
+        /// <summary>
+        /// Lazily-initialized options that set the <see cref="JsonSerializerOptions.MaxDepth"/> to match the <see cref="JsonWriterOptions"/> default
+        /// and that sets <see cref="JsonSerializerOptions.WriteIndented"/> to true.
+        /// </summary>
+        private static JsonSerializerOptions? s_indentedToStringOptions;
+
         /// <summary>
         ///   Converts the current instance to string in JSON format.
         /// </summary>
@@ -12,8 +23,11 @@ namespace System.Text.Json.Nodes
         /// <returns>JSON representation of current instance.</returns>
         public string ToJsonString(JsonSerializerOptions? options = null)
         {
-            using PooledByteBufferWriter output = WriteToPooledBuffer(options, options?.GetWriterOptions() ?? default);
-            return JsonHelpers.Utf8GetString(output.WrittenMemory.Span);
+            options ??=
+                s_unindentedToStringOptions ??
+                Interlocked.CompareExchange(ref s_unindentedToStringOptions, new JsonSerializerOptions { MaxDepth = JsonWriterOptions.DefaultMaxDepth }, null) ??
+                s_unindentedToStringOptions;
+            return JsonSerializer.Serialize(this, options);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonNode.To.cs
```diff
@@ -1,19 +1,33 @@
...
         /// <returns>JSON representation of current instance.</returns>
         public string ToJsonString(JsonSerializerOptions? options = null)
         {
-            using PooledByteBufferWriter output = WriteToPooledBuffer(options, options?.GetWriterOptions() ?? default);
-            return JsonHelpers.Utf8GetString(output.WrittenMemory.Span);
+            options ??=
+                s_unindentedToStringOptions ??
+                Interlocked.CompareExchange(ref s_unindentedToStringOptions, new JsonSerializerOptions { MaxDepth = JsonWriterOptions.DefaultMaxDepth }, null) ??
+                s_unindentedToStringOptions;
+            return JsonSerializer.Serialize(this, options);
```
stephentoub, line 30: @eiriktsarpalis, this is running afoul of the trimmer. What's the right / best way to specify the necessary converters and silence the warnings?  Presumably I don't want to use the source generator in System.Text.Json itself to create a context for JsonNode.

src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonNode.To.cs
```diff
@@ -37,8 +51,13 @@ public override string ToString()
                 }
             }
 
-            using PooledByteBufferWriter output = WriteToPooledBuffer(writerOptions: new JsonWriterOptions { Indented = true });
-            return JsonHelpers.Utf8GetString(output.WrittenMemory.Span);
+#pragma warning disable CA1869 // TODO https://github.com/dotnet/roslyn-analyzers/issues/6957
+            JsonSerializerOptions options =
+                s_indentedToStringOptions ??
+                Interlocked.CompareExchange(ref s_indentedToStringOptions, new JsonSerializerOptions { MaxDepth = JsonWriterOptions.DefaultMaxDepth, WriteIndented = true, }, null) ??
+                s_indentedToStringOptions;
```
huoyaoyuan, line 58: I wonder the overhead of `Lazy<T>` here. More overhead on startup even with `PublicationOnly`? Another level of indirection on every access?

src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonNode.To.cs
eiriktsarpalis: I don't think using `JsonSerializer` is a viable solution here. It roots all reflection components and changes the behavior of the `ToJsonString()` method so that it defaults to reflection when no options parameter or an unconfigured options parameter is being passed.

> What's the right / best way to specify the necessary converters

The problem is that `JsonNode` (more specifically `JsonValue<T>`) can encapsulate arbitrary .NET values, as such there is no fixed set of contracts we need to supply to make the method happy.

I think you might be able to get similar performance gains if you rewrite the `WriteToPooledBuffer` method to use the https://github.com/dotnet/runtime/blob/e8c3052573d17ac1148ec394986046b29a4d7227/src/libraries/System.Text.Json/src/System/Text/Json/Writer/Utf8JsonWriterCache.cs class (which is what JsonSerializer is doing).

stephentoub: > I think you might be able to get similar performance gains if you rewrite the WriteToPooledBuffer method

That's what I had first, actually, but I simplified it down to just using JsonSerializer since it seemed like unnecessary duplication.  But if using JsonSerializer isn't viable due to trimming-related reasons, that's unfortunate but I can revert it back.

stephentoub: I'll close this and open a new PR from my other branch with those changes.

PR #92304, opened 2023-09-19
Avoid unnecessary array allocation in JsonHelpers.Utf8GetString on netstandard

stephentoub:


eiriktsarpalis (approved):
Good catch, thanks.

PR #92307, opened 2023-09-19
JIT: Reimport full spill clique for I_IMPL<->BYREF mismatches

jakobbotsch:
When the JIT sees code that pushes an untracked pointer from one branch,
and a tracked pointer from another, it will reimport successors that
were already imported under the wrong assumption of an untracked
pointer. However, like in the other cases of spill clique handling, it
should reimport the entire spill clique, including predecessors and other
successors of those; otherwise we can still end up with mistyped LCL_VAR
nodes.


src/coreclr/jit/importer.cpp
```diff
@@ -11387,19 +11387,17 @@ void Compiler::impImportBlock(BasicBlock* block)
...
-               change the temp to TYP_BYREF and reimport the successors.
-               Note: We should only allow this in unverifiable code.
-            */
+            // VC generates code where it pushes a byref from one branch, and an int (ldc.i4 0) from
+            // the other. This should merge to a byref in unverifiable code.
+            // However, if the branch which leaves the TYP_I_IMPL on the stack is imported first, the
+            // successor would be imported assuming there was a TYP_I_IMPL on
+            // the stack. Thus the value would not get GC-tracked. Hence,
+            // change the temp to TYP_BYREF and reimport the clique.
+            // Note: We should only allow this in unverifiable code.
```
SingleAccretion, line 11396: Nit: delete the `Note: ...`? We don't do verification anymore.

src/coreclr/jit/importer.cpp
```diff
@@ -11379,38 +11379,35 @@ void Compiler::impImportBlock(BasicBlock* block)
...
+            // However, if the branch which leaves the TYP_I_IMPL on the stack is imported first, the
+            // successor would be imported assuming there was a TYP_I_IMPL on
+            // the stack. Thus the value would not get GC-tracked. Hence,
+            // change the temp to TYP_BYREF and reimport the clique.
+            LclVarDsc* tempDsc = lvaGetDesc(tempNum);
+            if (tree->TypeIs(TYP_BYREF) && (tempDsc->TypeGet() == TYP_I_IMPL))
             {
-                lvaTable[tempNum].lvType = TYP_BYREF;
-                impReimportMarkSuccessors(block);
-                markImport = true;
```
SingleAccretion, line 11402: `impReimportMarkSuccessors` is now dead code.

jakobbotsch: Thanks, deleted

jakobbotsch:

[No diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=415510&view=ms.vss-build-web.run-extensions-tab). Failure is known according to build analysis.

AndyAyersMS (approved):

PR #92310, opened 2023-09-19
Change GetParametersNoCopy to GetParametersAsSpan and use it in a few more places

stephentoub:
The internal GetParametersNoCopy avoids a defensive ParameterInfo[] copy, but it requires the callers to promise not to mutate the array. I'm changing the method to return a `ReadOnlySpan<ParameterInfo>`, which means a caller can't mutate it without using unsafe code, and enabling the compiler to flag any misuse.  I've then used it in a few more places. We might subsequently want to consider exposing it publicly.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/DynamicMethod.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/DynamicMethod.cs
@@ -329,15 +329,10 @@ namespace System.Reflection.Emit
 
         public override MethodInfo GetBaseDefinition() => this;
 
-        public override ParameterInfo[] GetParameters()
-        {
-            ParameterInfo[] privateParameters = LoadParameters();
-            ParameterInfo[] parameters = new ParameterInfo[privateParameters.Length];
-            Array.Copy(privateParameters, parameters, privateParameters.Length);
-            return parameters;
-        }
+        public override ParameterInfo[] GetParameters() =>
+            GetParametersAsSpan().ToArray();
 
-        internal override ParameterInfo[] GetParametersNoCopy() => LoadParameters();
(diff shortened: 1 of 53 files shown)


jkotas:
Nit: `GetParametersAsSpan` may be more consistent with our naming convention for overloading on return value.

Existing art https://github.com/dotnet/runtime/blob/b049f42e5977ead2735402a79ad8247441cc3c7c/src/libraries/System.IO.Hashing/ref/System.IO.Hashing.cs#L14

jkotas (approved):
Thanks!

src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/DynamicMethod.cs
```diff
@@ -329,15 +329,10 @@ public override string ToString()
...
-            ParameterInfo[] privateParameters = LoadParameters();
-            ParameterInfo[] parameters = new ParameterInfo[privateParameters.Length];
-            Array.Copy(privateParameters, parameters, privateParameters.Length);
-            return parameters;
-        }
+        public override ParameterInfo[] GetParameters() =>
+            GetParametersAsSpan().ToArray();
 
-        internal override ParameterInfo[] GetParametersNoCopy() => LoadParameters();
+        internal override ReadOnlySpan<ParameterInfo> GetParametersAsSpan() => LoadParameters();
```
marek-safar, line 335: You could move LoadParameters implementation here

stephentoub: I could, but it would complicate DefineParameter, which uses members defined on the derived RuntimeParameterInfos it gets back from LoadParameters.

stephentoub:
Both failures are known and tracked. I ran extra platforms as well, and all of those also appear to be known, with the exception of some recently added LINQ tests failing on nativeaot for which I opened an issue.

PR #92313, opened 2023-09-19
Properly handle debugger-enumerating interior pointers and enregistered refs

### LINKED ISSUE #87239: ICorDebugGCReferenceEnum.Next fails to return stack references in .NET 8 preview 4
### Description

.NET 8 preview 4 seems to have a regression that ICorDebugGCReferenceEnum.Next is no longer returning stack references

### Reproduction Steps

1. Create the following program
2. Using Visual Studio, set a breakpoint where indicated
3. When the breakpoint is hit, take a heap snapshot from the Diagnostics tools window
4. View the snapshot

```C#
class MyStringHolder
{
    public readonly string Value;

    public MyStringHolder(string value)
    {
        Value = value;
    }
}

class Program
{
    public static void Main()
    {
        var bigString = new MyStringHolder(new string('*', 1000000));

        Console.WriteLine(bigString.Value.Length); // breakpoint here
    }
}
```

### Expected behavior

'MyStringHolder' should be shown as having the largest inclusive size

### Actual behavior

`ICorDebugGCReferenceEnum.Next` is not returning values for all of the stack variables (references with type `CorReferenceStack`). I see some from the hot reload thread, but none from the 'Main' function.

### Regression?

This worked correctly in .NET 8 preview 3

### Known Workarounds

_No response_

### Configuration

I am testing with .NET 8 preview 4 on Windows. Both x86 and x64. I didn't test any other architecture or OS.

### Other information

_No response_

leculver:
There is an ancient bug in the dac root walking code.  If we hit an interior pointer on a callstack that lives outside of the GC heap, we will report that pointer as-is to ICorDebug (specifically `CordbRefEnum::Next`).  This is despite the fact that ICorDebug *specifically* requests that the dac only enumerate pointers to real objects.  This non-gc pointer will be treated as a real object pointer and will cause a failed HRESULT in CordbRefEnum.  Since this call is wrapped with `IfFailThrow` it will halt all further processing of roots, leading to not fully capturing the object graph for memory analysis.

This code makes a small, targeted change to not enumerate non-GC pointers when the caller requests we specifically only enumerate pointers to real objects.

This is not a recent regression, but rather a bug that has likely existed since the original code was written.

FYI @asundheim.

### DIFF
--- a/src/coreclr/debug/daccess/dacdbiimpl.cpp
+++ b/src/coreclr/debug/daccess/dacdbiimpl.cpp
@@ -7781,8 +7781,9 @@ HRESULT DacStackReferenceWalker::Next(ULONG count, DacGcReference stackRefs[], U
         stackRefs[i].i64ExtraData = 0;
 
         const SOSStackRefData &sosStackRef = mList.Get(i);
-        if (sosStackRef.Address == 0 || sosStackRef.Flags & GC_CALL_INTERIOR)
+        if (sosStackRef.Flags & GC_CALL_INTERIOR | sosStackRef.Address == 0)
         {
+            // Direct pointer case - interior pointer, Frame ref, or enregistered var.
             stackRefs[i].pObject = CLRDATA_ADDRESS_TO_TADDR(sosStackRef.Object) | 1;
         }
         else
--- a/src/coreclr/debug/daccess/dacdbiimpl.cpp
+++ b/src/coreclr/debug/daccess/dacdbiimpl.cpp
@@ -7781,7 +7781,7 @@ HRESULT DacStackReferenceWalker::Next(ULONG count, DacGcReference stackRefs[], U
         stackRefs[i].i64ExtraData = 0;
 
         const SOSStackRefData &sosStackRef = mList.Get(i);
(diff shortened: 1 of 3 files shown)


tommcdon:
@hoyosjs

hoyosjs (approved):

src/coreclr/debug/daccess/dacdbiimpl.cpp
```diff
@@ -7781,8 +7781,9 @@ HRESULT DacStackReferenceWalker::Next(ULONG count, DacGcReference stackRefs[], U
         stackRefs[i].i64ExtraData = 0;
 
         const SOSStackRefData &sosStackRef = mList.Get(i);
-        if (sosStackRef.Flags & GC_CALL_INTERIOR)
+        if (sosStackRef.Flags & GC_CALL_INTERIOR | sosStackRef.Address == 0)
```
mikem8361, line 7784: Minor nit:  should this OR be `||`

hoyosjs, line 7784:
```suggestion
        if (sosStackRef.Flags & GC_CALL_INTERIOR || sosStackRef.Address == 0)
```

mikem8361 (approved):

PR #92315, opened 2023-09-20
Bring back CopyOutputSymbolsToPublishDirectory

### LINKED ISSUE #92188: CopyOutputSymbolsToPublishDirectory is not respected
It used to be the case that `CopyOutputSymbolsToPublishDirectory=false` would prevent copying PDB/symbols to the publish folder, but it looks like we dropped it in #89358.

Cc @agocke

agocke:
I accidentally removed this property from AOT compilation when adding support for Mac dsym bundles. This change re-enables support for suppressing debugging symbols in the output.

Fixes #92188

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
@@ -93,7 +93,10 @@
     <!-- replace apphost with binary we generated during native compilation -->
     <Delete Files="$(PublishDir)\$(TargetName)$(NativeBinaryExt)" />
     <Copy SourceFiles="$(NativeOutputPath)$(TargetName)$(NativeBinaryExt)" DestinationFolder="$(PublishDir)" />
+  </Target>
 
+  <Target Name="_CopyAotSymbols" AfterTargets="CopyNativeBinary"
+          Condition="'$(CopyOutputSymbolsToPublishDirectory)' == 'true'">
     <!-- dotnet CLI produces managed debug symbols, which we will delete and copy native symbols instead -->
     <Delete Files="$(PublishDir)\$(TargetName).pdb" />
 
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
```diff
@@ -93,7 +93,10 @@
     <!-- replace apphost with binary we generated during native compilation -->
     <Delete Files="$(PublishDir)\$(TargetName)$(NativeBinaryExt)" />
     <Copy SourceFiles="$(NativeOutputPath)$(TargetName)$(NativeBinaryExt)" DestinationFolder="$(PublishDir)" />
+  </Target>
 
+  <Target Name="_CopyAotSymbols" AfterTargets="CopyNativeBinary"
```
MichalStrehovsky, line 98:
```suggestion
  <Target Name="_CopyAotSymbols" AfterTargets="Publish"
```

It feels like someone deleting `CopyNativeBinary` without searching for all the places that refer to this string is a thing that could happen. This would prevent it.

MichalStrehovsky (approved):

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Publish.targets
eerhardt: > It feels like someone deleting CopyNativeBinary without searching for all the places that refer to this string is a thing that could happen. 

Tests would prevent that.

agocke: I will add a test for this in the SDK repo.

agocke: But it's a good suggestion regardless, why not parallelize?

PR #92316, opened 2023-09-20
JIT: Fixed incorrect reversed condition for GT

### LINKED ISSUE #92201: Test failure _Vector128_ro::JIT.HardwareIntrinsics.General._Vector128.Program.LessThanOrEqualAnySingle()
**Failed in:** [runtime-coreclr jitstress-random 20230917.1](https://dev.azure.com/dnceng-public/public/_build/results?buildId=409203&view=ms.vss-test-web.build-test-results-tab&runId=8894264&resultId=112889&paneView=debug)

**Failed tests:**
```
coreclr linux arm Checked jitstress_random_2 @ (Ubuntu.1804.Arm32.Open)Ubuntu.1804.Armarch.Open@mcr.microsoft.com/dotnet-buildtools/prereqs:ubuntu-18.04-helix-arm32v7
    - _Vector128_ro::JIT.HardwareIntrinsics.General._Vector128.Program.LessThanOrEqualAnySingle()
```

**Error message:**
```
One or more scenarios did not complete as expected.
```

**Stack trace:**
```
   at JIT.HardwareIntrinsics.General._Vector128.Program.LessThanOrEqualAnySingle() in /__w/1/s/artifacts/tests/coreclr/obj/AnyOS.x64.Checked/Managed/JIT/HardwareIntrinsics/General/Vector128/Vector128_ro/Vector128_ro/gen/LessThanOrEqualAny.Single.cs:line 52
   at Program.<<Main>$>g__TestExecutor727|0_728(StreamWriter tempLogSw, StreamWriter statsCsvSw, <>c__DisplayClass0_0&)
```

TIHan:
Resolves https://github.com/dotnet/runtime/issues/92201

`GT` was getting incorrectly reversed. Looks like an oversight, though it is surprising this wasn't caught sooner. Original code was here since 2018.

### DIFF
--- a/src/tests/JIT/Regression/JitBlue/Runtime_92201/Runtime_92201.csproj
+++ b/src/tests/JIT/Regression/JitBlue/Runtime_92201/Runtime_92201.csproj
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <Optimize>True</Optimize>
+    <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="$(MSBuildProjectName).cs" />
--- a/src/tests/JIT/Regression/JitBlue/Runtime_92201/Runtime_92201.csproj
+++ b/src/tests/JIT/Regression/JitBlue/Runtime_92201/Runtime_92201.csproj
@@ -7,6 +7,6 @@
 
     <CLRTestEnvironmentVariable Include="DOTNET_TieredCompilation" Value="0" />
     <CLRTestEnvironmentVariable Include="DOTNET_EnableHWIntrinsic" Value="0" />
-    <CLRTestEnvironmentVariable Include="DOTNET_JitStress" Value="9a" />
+    <CLRTestEnvironmentVariable Include="DOTNET_JitStressModeNames" Value="STRESS_BB_PROFILE STRESS_REVERSE_FLAG" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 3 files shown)


jakobbotsch (approved):
Add a test?

jakobbotsch:
> `GT` was getting incorrectly reversed. Looks like an oversight, though it is surprising this wasn't caught sooner. Original code was here since 2018.

It's probably very hard to end up with the `FGTU` unordered floating pointer greater than comparison in the first place, looks like the only possible way is if we reverse a `FLE` comparison in the first place, but then it's probably very unlikely for us to end up reversing it a second time to hit the bug.
So it might be very hard to come up with a test.

src/tests/JIT/Regression/JitBlue/Runtime_92201/Runtime_92201.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <Optimize>True</Optimize>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="$(MSBuildProjectName).cs" />
+
+    <CLRTestEnvironmentVariable Include="DOTNET_TieredCompilation" Value="0" />
+    <CLRTestEnvironmentVariable Include="DOTNET_EnableHWIntrinsic" Value="0" />
+    <CLRTestEnvironmentVariable Include="DOTNET_JitStress" Value="9a" />
```
jakobbotsch, line 10: Tests should  specify the exact right stress modes via `DOTNET_JitStressModeNames` as the exact set of enabled stress modes can otherwise end up changing in the future due to other changes. Also if possible it would be nice to reduce the test case to something minimal

TIHan: It would take some time to produce a minimal test and don't want to spend that much more time on it, but I can put in the stress mode names

TIHan: I cleaned some the test up.

Issue #75749, opened 2022-09-16
Port CoreCLR PAL to linux-riscv64


am11:
The first component in [coreclr port](https://github.com/dotnet/runtime/blob/main/docs/design/coreclr/botr/guide-for-porting.md#components-of-the-runtime-affected-by-a-port-to-a-new-architecture) is the abstraction layer located at `src/coreclr/pal`. That enables us to port other components of coreclr such as `src/coreclr/vm`, `src/coreclr/unwinder` and ultimately `src/coreclr/jit`.

We have riscv64 cross-toolchain with prerequisites set up for building the runtime. Some initial strokes to port coreclr are checked into the `main` branch (cmake configurations, libunwind, a few definitions in pal.h etc.).

Here is a gist for cross compilation:

```sh
#!/bin/sh

# on Ubuntu amd64 - with curl, jq and docker installed

tag="$(curl -sSL "https://raw.githubusercontent.com/dotnet/versions/main/build-info/docker/image-info.dotnet-dotnet-buildtools-prereqs-docker-main.json" |\
  jq -r '.repos[] | .images[] | .platforms [] | select(.dockerfile | contains("riscv64")).simpleTags[0]')"

URN="mcr.microsoft.com/dotnet-buildtools/prereqs:$tag"

cd runtime  # current main branch of dotnet/runtime

# Only build coreclr PAL and its tests:
docker run --rm -e ROOTFS_DIR=/crossrootfs/riscv64 -v"$(pwd)":/runtime "$URN" \
  /runtime/src/coreclr/build-runtime.sh -cross -a riscv64 -component paltests

# note: /crossrootfs/riscv64 has Debian sid bootstrap environment with llvm 14 and few other dependencies
# and the host has llvm 15 toolchain, which will be used by cmake for compilation with
# -sysroot=/crossrootfs/riscv64
#
# the relevant code to build this prerequisites docker image are:
#   * https://github.com/dotnet/arcade/blob/2e7521a/eng/common/cross/build-rootfs.sh#L155
#   * https://github.com/dotnet/arcade/blob/2e7521a/eng/common/cross/toolchain.cmake#L64
#   * https://github.com/dotnet/dotnet-buildtools-prereqs-docker/tree/main/src/ubuntu/22.04/cross/riscv64
```

[...]

PR #92324, opened 2023-09-20
[RISC-V] Set the HAVE_FUNCTIONAL_PTHREAD_ROBUST_MUTEXES_EXITCOD value to 0

JongHeonChoi:
The ```HAVE_FUNCTIONAL_PTHREAD_ROBUST_MUTEXES_EXITCODE``` value is also set to 0 in the ```VisionFive2``` target of ```RISCV64```.

cc @wscho77 @HJLeee @JongHeonChoi @t-mustafin @alpencolt @gbalykov @clamp03 @Sirntar @yurai007


eng/native/tryrun.cmake
```diff
@@ -170,8 +170,6 @@ else()
   message(FATAL_ERROR "Unsupported platform. OS: ${CMAKE_SYSTEM_NAME}, arch: ${TARGET_ARCH_NAME}")
 endif()
 
-if(TARGET_ARCH_NAME MATCHES "^(x86|x64|s390x|armv6|loongarch64|ppc64le)$")
+if(TARGET_ARCH_NAME MATCHES "^(x86|x64|s390x|armv6|loongarch64|riscv64|ppc64le)$")
   set_cache_value(HAVE_FUNCTIONAL_PTHREAD_ROBUST_MUTEXES_EXITCODE 0)
```
jkotas, line 174: Where is `HAVE_FUNCTIONAL_PTHREAD_ROBUST_MUTEXES_EXITCODE` used? Can we delete it instead?

jkotas, line 174: @kouvel Do we still need `HAVE_FUNCTIONAL_PTHREAD_ROBUST_MUTEXES`? It does not look right that it is architecture specific.

kouvel: It was introduced due to a couple of issues, one in Alpine Linux and one in glibc on ARM/ARM64, but I believe that both have since been fixed. I'm not sure why the cache value is being set based on architecture, it should probably be checked based on the target OS, which is currently done in `mutex.hpp` in the PAL.

It was noted in https://github.com/dotnet/runtime/pull/82380#discussion_r1128826231 that a PAL test for named mutexes fails when pthread robust mutexes are used on RISC-V, so I wonder if it actually works. @JongHeonChoi, have you checked if the corresponding test case in https://github.com/dotnet/runtime/blob/e235aeff6c822ad18a63b39819f61031a1ac95e2/src/coreclr/pal/src/configure.cmake#L1012-L1199 compiles and runs successfully on RISC-V? Probably would also need to check the PAL named mutex tests to make sure that it works.

JongHeonChoi: @kouvel Yes. I checked the return value of the test case in VisionFive2 target.

clamp03 (approved):

eng/native/tryrun.cmake
kouvel: Sounds good, thanks! I'll look into whether this check can be removed, as I'm not aware of any current cases where it would fail.

kouvel (approved):

wscho77 (approved):

gbalykov (approved):
LGTM. I've added pal tests results with this change too in https://github.com/dotnet/runtime/issues/75749#issuecomment-1729210218.

PR #92334, opened 2023-09-20
[wasm][debugger] Support indexing by value type schema

ilonatommy:

- Added test for indexing by enum and struct
- Enum indexing needed a fix. When indexing, `SyntaxNode` returns `IdentifierNameSyntax` value that is nested like this:
```
{
  "type": "object",
  "value": {
    "type": "object",
    "value": {
      "type": "number",
      "value": 0,
      "description": "0",
      "className": "Int32"
    },
    "description": "High",
    "className": "DebuggerTests.EvaluateLocalsWithIndexingTests.EnumIndexer",
    "objectId": "dotnet:valuetype:10",
    "isValueType": true,
    "isEnum": true
  },
  "description": "High",
  "className": "DebuggerTests.EvaluateLocalsWithIndexingTests.EnumIndexer",
  "objectId": "dotnet:valuetype:12",
  "isValueType": true,
  "isEnum": true
}
```
while in other scenarios, the nesting in enums do not occur and values look e.g. like this:
```
{
  "type": "object",
  "value": {
    "type": "number",
    "value": 0,
    "description": "0",
    "className": "Int32"
  },
  "description": "High",
  "className": "DebuggerTests.EvaluateLocalsWithIndexingTests.EnumIndexer",
  "objectId": "dotnet:valuetype:10",
  "isValueType": true,
  "isEnum": true
}
```
Because we want `ConvertJSToCSharpLocalVariableAssignment` to handle both cases, try-catch expression was added (checking type with "is" keyword does not work here as we are casting and exception happens in the cast). 

In `GetVariableDefinitions` we handle enum case separately as a workaround to the following: 
In the original version of code, we would produce variable definition that looks like this: `DebuggerTests.castToEnumTypeThatIsDefinedInExternalAssembly enumIdx = (DebuggerTests.castToEnumTypeThatIsDefinedInExternalAssembly) 0;`
Error produced when we use this approach implies that we are missing the assembly with Enum type definition passed as reference to the script: `The type or namespace name 'DebuggerTests' could not be found (are you missing a using directive or an assembly reference?)",`
Adding pdbs that are stored in `DebugStore.assemblies` to the Roslyn's script did not affect the error message, so I used the workaround of passing numeric value of enum instead.


src/mono/wasm/debugger/BrowserDebugProxy/EvaluateExpression.cs
```diff
@@ -277,8 +277,20 @@ public static string ConvertJSToCSharpLocalVariableAssignment(string idName, JTo
...
+                                // when indexing with enums, enum object is packed in another enum
+                                // which is not the case in other scenarios
+                                try
+                                {
+                                    valueRet = $"({typeRet}) {value["value"].Value<double>()}";
+                                }
+                                catch (InvalidCastException)
+                                {
+                                    // value["value"] is JObject with another "value" field
+                                    valueRet = $"({typeRet}) {value["value"]["value"].Value<double>()}";
```
thaystg, line 292: Should we check if this also throws an exception?

radical:
> ... while in other scenarios, the nesting in enums do not occur and values look ...

What are the cases when these two separate forms are emitted? Why not just one?

src/mono/wasm/debugger/BrowserDebugProxy/EvaluateExpression.cs
ilonatommy: I will add it, it might be easier to read the logs later. I also checked all the usages of this function and they are secured by try-catch on higher levels so it would have gotten caught.

ilonatommy:
Failures after merge with main:
The order of declaring indexing methods in `TestEvaluate` started to matter which should not happen.

Reason:
When we are checking method argument types in `GetParametersInfo`:
https://github.com/dotnet/runtime/blob/aba7c43598dc13e7e43843dceb57a397f13107cb/src/mono/wasm/debugger/BrowserDebugProxy/DebugStore.cs#L474
all types: `class Indexer`, `enum EnumIndexer` and `struct StructIndexer` are decoded as "object", even though `ElementType` defines ValueType and we would expect the two latter to return ValueType:
https://github.com/dotnet/runtime/blob/aba7c43598dc13e7e43843dceb57a397f13107cb/src/mono/wasm/debugger/BrowserDebugProxy/MonoSDBHelper.cs#L316.
As a consequence, when we evaluate indexing by struct: `f[StructIndexer]` and check compatibility of method `get_Items(Indexer)`, where indexObject is:
```
{
  "type": "object",
  "value": null,
  "description": "DebuggerTests.EvaluateLocalsWithIndexingTests.StructIndexer",
  "className": "DebuggerTests.EvaluateLocalsWithIndexingTests.StructIndexer",
  "objectId": "dotnet:valuetype:13",
  "isValueType": true
}
```
we are not able to detect type incompatibility (struct cannot be passed to `get_Index` with `Indexer` argument).
While similar case of passing struct into a method `get_Index[EnumIndexer]` fails on the debugger side, the above one passes, returning incorrect response:
```
{
   "type": "number",
   "value": 1.0,
   "description": "1",
   "className": "Object"
 }
```
instead of:
```
{
   "type": "boolean",
   "value": true,
   "description": "true"
 }
```

ilonatommy:
> What are the cases when these two separate forms are emitted? Why not just one?

Nested result:
- Evaluated expression: `f[enumIdx]` or `enumIdx` or `enumIdx.ToString()`
- How do we get this value? On `VisitInternal` we discover one `IdentifierNameSyntax` named `enumIdx`. We get its value on this line:
https://github.com/dotnet/runtime/blob/05730a37fff1fea8ac01cecc60b55f28da0b1db4/src/mono/wasm/debugger/BrowserDebugProxy/EvaluateExpression.cs#L430
by calling `ResolveIdentifier` -> `Resolve(string, CancellationToken)` -> `ResolveAsLocalOrThisMember` and retrieving the value from `scopeCache.Locals`. It is already nested.

Non-nested:
- I still cannot reproduce the expression when I got non-nested case. I will edit when I get more info.

ilonatommy:
> What are the cases when these two separate forms are emitted? Why not just one?

Update to this question after reverting the try-catch. Test that fails: `EvaluateMethodsOnPrimitiveTypesReturningObjects`.
Expression: `test.propString.Split('_', 3, System.StringSplitOptions.TrimEntries)`, the problematic value is `System.StringSplitOptions.TrimEntries`.

Reason: 
We are not nesting the results of `ResolveStaticMembersInStaticTypes`:
```
{
  "type": "object",
  "value": {
    "type": "number",
    "value": 2,
    "description": "2",
    "className": "Int32"
  },
  "description": "TrimEntries",
  "className": "System.StringSplitOptions",
  "objectId": "dotnet:valuetype:1",
  "isValueType": true,
  "isEnum": true
}
```
We could change the static members resolution behavior to be nested but it seems wrong and artificial. I think it will be better to go back to 2-version handling instead.

PR #92337, opened 2023-09-20
[wasm][debugger] Support indexing by array schema

ilonatommy:


src/mono/wasm/debugger/BrowserDebugProxy/MonoSDBHelper.cs
```diff
@@ -564,7 +564,7 @@ public override void Write(string val)
 
         public void WriteObj(DotnetObjectId objectId, MonoSDBHelper SdbHelper)
         {
-            if (objectId.Scheme == "object")
+            if (objectId.Scheme == "object" || objectId.Scheme == "array")
             {
                 Write(ElementType.Class, objectId.Value);
```
thaystg, line 569: I don't know how this is working.
If the scheme is array we should use:
`Write(ElementType.Array, objectId.Value);`

radical: > I don't know how this is working.

Tests did fail.

ilonatommy: These were tests not connected with array indexing, Thays is right that it's strange it did not fail. Version with `ElementType.Array` passes as well.

thaystg: Wait, from the Javascript point of view, the type is an object with a subtype array, that is how the Javascript works and how we send the information to the IDE to show it correctly.
From the Mono Runtime point of view, it's an array. The method is expecting an array as a parameter, the way that we are doing, we are saying to runtime that we are sending a "class" and then sending the id of an array, the runtime is decoding it correctly and this is why it's working when you use class, but I think we should fix runtime side.
We should add in this line: https://github.com/dotnet/runtime/blob/9c8ff4c1c12c488da660739333e36eb55afe9c56/src/mono/mono/component/debugger-agent.c#L5731C8-L5731C12 
`|| (type == MONO_TYPE_ARRAY)`

ilonatommy:
@thaystg, it should be the same as for object: `t->type` is `MONO_TYPE_VALUETYPE`. I updated the code, for the array test case it comes with: `t->type = 0x11, type = 0x14`.

thaystg (approved):

PR #92338, opened 2023-09-20
[Mono]: Add missing case for constrained gsharedvt call.

lateralusX:
https://github.com/dotnet/runtime/commit/1b788f4dc3e3a8829488e52c032ad6a70671e070 added a new value to our MonoRgctxInfoType enum type, but appears that all cases where not full adjusted. Running System.Buffers tests in full AOT hits the assert in `info_equal` about missing case, https://github.com/dotnet/runtime/blob/0dc5903679606b072adac70a268cdb77d1147b3e/src/mono/mono/mini/mini-generic-sharing.c#L2908.

This commit adds the new enum value and align handling similar to other cases added by [that commit](https://github.com/dotnet/runtime/commit/1b788f4dc3e3a8829488e52c032ad6a70671e070). After the fix System.Buffers test pass using full AOT.


lambdageek (approved):

lambdageek:
we should backport to net8.0

vargaz (approved):

lateralusX:
Failure should be unrelated, hitting this case before would have asserted and stopped process.

PR #92348, opened 2023-09-20
[browser] Fix SIMD+EH check

maraf:
- Before this change the `linkerWasmEnableSIMD` and `linkerWasmEnableEH` used in `configureRuntimeStartup` to check support for these features always had the default (`true`) value. Throwing assert error even when SIMD and EH was disabled for build.
- Call `cwraps.mono_wasm_abort` from `runtimeHelpers.abort` only after cwraps are ready (`onRuntimeInitializedAsync`).

### DIFF
--- a/src/mono/wasm/runtime/startup.ts
+++ b/src/mono/wasm/runtime/startup.ts
@@ -41,12 +41,7 @@ import { assertNoProxies } from "./gc-handles";
 const MONO_PTHREAD_POOL_SIZE = 4;
 
 export async function configureRuntimeStartup(): Promise<void> {
-    if (linkerWasmEnableSIMD) {
-        mono_assert(await loaderHelpers.simd(), "This browser/engine doesn't support WASM SIMD. Please use a modern version. See also https://aka.ms/dotnet-wasm-features");
-    }
-    if (linkerWasmEnableEH) {
-        mono_assert(await loaderHelpers.exceptions(), "This browser/engine doesn't support WASM exception handling. Please use a modern version. See also https://aka.ms/dotnet-wasm-features");
-    }
+    
 
     await init_polyfills_async();
 
@@ -388,7 +383,6 @@ async function mono_wasm_pre_init_essential_async(): Promise<void> {
     mono_log_debug("mono_wasm_pre_init_essential_async");
     Module.addRunDependency("mono_wasm_pre_init_essential_async");
 
-
     if (MonoWasmThreads) {
         preAllocatePThreadWorkerPool(MONO_PTHREAD_POOL_SIZE, runtimeHelpers.config);
     }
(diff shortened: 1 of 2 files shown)


src/mono/wasm/runtime/startup.ts
```diff
@@ -363,11 +367,10 @@ function mono_wasm_pre_init_essential(isWorker: boolean): void {
     }
 
     init_c_exports();
-    runtimeHelpers.mono_wasm_exit = cwraps.mono_wasm_exit;
+    runtimeHelpers.mono_wasm_exit = () => {
+        throw new Error("Mono shutdown");
+    };
     runtimeHelpers.abort = (reason: any) => {
```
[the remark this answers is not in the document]
pavelsavara: And perhaps those should be copied from `loaderHelper`

src/mono/wasm/runtime/startup.ts
```diff
@@ -41,12 +41,7 @@ import { assertNoProxies } from "./gc-handles";
 const MONO_PTHREAD_POOL_SIZE = 4;
 
 export async function configureRuntimeStartup(): Promise<void> {
-    if (linkerWasmEnableSIMD) {
-        mono_assert(await loaderHelpers.simd(), "This browser/engine doesn't support WASM SIMD. Please use a modern version. See also https://aka.ms/dotnet-wasm-features");
-    }
-    if (linkerWasmEnableEH) {
-        mono_assert(await loaderHelpers.exceptions(), "This browser/engine doesn't support WASM exception handling. Please use a modern version. See also https://aka.ms/dotnet-wasm-features");
-    }
+
```
radical, line 44:
```suggestion
```

radical:
> Before this change the linkerWasmEnableSIMD and linkerWasmEnableEH used in configureRuntimeStartup to check support for these features always had the default (true) value. Throwing assert error even when SIMD and EH was disabled for build.

So, this always failed on browsers that didn't support SIMD/EH, even when SIMD was disabled in the build? Can we test this with chrome/v8?

src/mono/wasm/runtime/startup.ts
```diff
@@ -470,6 +466,8 @@ async function instantiate_wasm_module(
 
         replace_linker_placeholders(imports);
         const assetToLoad = await loaderHelpers.wasmDownloadPromise.promise;
+
+        await ensureUsedWasmFeatures();
```
pavelsavara, line 470: we could actually bomb even before downloading the wasm file, few lines earlier.

maraf: I have fired it before awaiting wasm download, but awaited after. Makes sense?

maraf:
> So, this always failed on browsers that didn't support SIMD/EH, even when SIMD was disabled in the build? 
Exactly. 

> Can we test this with chrome/v8?
If we install Chrome 85 or Firefox 88, we can test it. I can look into it in a separate PR.

pavelsavara (approved):

radical (approved):

lewing:
closed in favor of https://github.com/dotnet/runtime/pull/92439

PR #92355, opened 2023-09-20
JIT: fix self-conflicting HFA arg prolog handling for arm64

### LINKED ISSUE #83167: Assertion failed 'regArgTab[argNum].slot == 2'
Seen in https://dev.azure.com/dnceng-public/public/_build/results?buildId=197177&view=logs&j=2f6a7d26-0d60-5ade-d191-981fe0847989&t=950a5a68-ae8a-5bb4-8cad-fb293bb99dc4&l=6041

(linux.arm64.Checked leg of runtime-extra-platforms)

```
  ILC: /__w/1/s/src/coreclr/jit/codegencommon.cpp:3957
  ILC: Assertion failed 'regArgTab[argNum].slot == 2' in 'System.Collections.Generic.GenericEqualityComparer`1[System.Numerics.Plane]:GetHashCode(System.Numerics.Plane):int:this' during 'Generate code' (IL size 24; hash 0x7039dd06; FullOpts)
  
  Aborted (core dumped)
/__w/1/s/artifacts/bin/coreclr/linux.arm64.Checked/build/Microsoft.NETCore.Native.targets(266,5): error MSB3073: The command ""/__w/1/s/artifacts/bin/coreclr/linux.arm64.Checked/x64/ilc/ilc" @"/__w/1/s/artifacts/obj/System.Numerics.Vectors.Tests/Release/net8.0/native/System.Numerics.Vectors.Tests.ilc.rsp"" exited with code 134. [/__w/1/s/src/libraries/System.Numerics.Vectors/tests/System.Numerics.Vectors.Tests.csproj]
```

Looks to be related to hardware intrinsics. Cc @tannergooding

AndyAyersMS:
Fix prolog handling in the case where the in-body destination register
for an HFA overlaps with one of the HFA argument registers. For instance
the HFA is passed in `s0-s3` and needs to end up in `v3`.

This requires special handling because the dependence analysis done in
`genFnPrologCalleeRegArgs` only tracks entire registers, not parts of
registers.

Fixes #83167

### DIFF
--- a/src/coreclr/jit/codegencommon.cpp
+++ b/src/coreclr/jit/codegencommon.cpp
@@ -2919,6 +2919,8 @@ void CodeGen::genFnPrologCalleeRegArgs(regNumber xtraReg, bool* pXtraRegClobbere
         bool writeThru;    // true if the argument gets homed to both stack and register
         bool processed;    // true after we've processed the argument (and it is in its final location)
         bool circular;     // true if this register participates in a circular dependency loop.
+        bool hfaConflict;  // arg is part of an HFA that will end up in the same register
+                           // but in a different slot (eg arg in s3 = v3.s[0], needs to end up in v3.s[3])
     } regArgTab[max(MAX_REG_ARG + 1, MAX_FLOAT_REG_ARG)] = {};
 
     unsigned   varNum;
@@ -3284,7 +3286,8 @@ void CodeGen::genFnPrologCalleeRegArgs(regNumber xtraReg, bool* pXtraRegClobbere
      * A circular dependency is a set of registers R1, R2, ..., Rn
      * such that R1->R2 (that is, R1 needs to be moved to R2), R2->R3, ..., Rn->R1 */
 
-    bool change = true;
+    bool change         = true;
+    bool hasHfaConflict = false;
     if (regArgMaskLive)
     {
         /* Possible circular dependencies still exist; the previous pass was not enough
@@ -3341,6 +3344,26 @@ void CodeGen::genFnPrologCalleeRegArgs(regNumber xtraReg, bool* pXtraRegClobbere
                     assert(regArgTab[argNum].slot <= (int)varDsc->lvHfaSlots());
                     assert(argNum > 0);
                     assert(regArgTab[argNum - 1].varNum == varNum);
+
+                    // If the field is passed in the same register as the destination,
+                    // but is in the wrong part of the register, mark it specially so later
+                    // we make sure to move it to the right spot before "freeing" the destination.
+                    //
+                    // If the arg is already in the right place we don't need to do anything special
+                    destRegNum = varDsc->GetRegNum();
+                    if (regNum == destRegNum)
+                    {
+                        const int slot = regArgTab[argNum].slot;
+                        if (slot != 1)
+                        {
+                            JITDUMP("HFA conflict; arg num %u needs to move from %s[%u] to %s[%u]\n", argNum,
+                                    getRegName(regNum), 0, getRegName(destRegNum), slot - 1);
+                            regArgTab[argNum].hfaConflict = true;
+
+                            // We'll need to do a special pass to resolve these
+                            hasHfaConflict = true;
+                        }
+                    }
                     regArgMaskLive &= ~genRegMask(regNum);
                     regArgTab[argNum].circular = false;
                     change                     = true;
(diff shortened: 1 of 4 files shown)


AndyAyersMS:
@jakobbotsch @BruceForstall went back and forth on this but this seems to be a plausible fix. 

I suspect we may not handle HFA doubles or HVAs properly though, so may want to tweak this some more.

How hard it would be to update the ABI tests to stress this bit of code better? Seems like if instead of hashing the args the in the callee the callee could permute them and call some other (noinline) method perhaps that would stress the various cases here better? Eg two swapped HFAs possibly with some other float args mixed in.

jakobbotsch:
> How hard it would be to update the ABI tests to stress this bit of code better? Seems like if instead of hashing the args the in the callee the callee could permute them and call some other (noinline) method perhaps that would stress the various cases here better? Eg two swapped HFAs possibly with some other float args mixed in.

This is essentially what the tailcall stressing mode already does, so you may try commenting the non HFA types in https://github.com/dotnet/runtime/blob/main/src/tests/JIT/Stress/ABI/Types.cs to force more frequent testing with HFAs. But from my tests it seemed like it was hard to get LSRA to allocate parameters in a way that created the hard cases, so maybe we would need a combination of this and a new stress mode.

What is the problematic case that happens here? I think I didn't quite understand it now that I read your comment "If the field is passed in the same register as the destination, but is in the wrong part of the register".

AndyAyersMS:
> What is the problematic case that happens here? I think I didn't quite understand it now that I read your comment "If the field is passed in the same register as the destination, but is in the wrong part of the register".

There is an HFA passed in `s0-s3` that needs to end up in `v3`. The current strategy for HFAs is to wait for the destination to be free and then merge everything into the destination when processing the first param reg. But here the HFA conflicts with itself so the destination never frees up. And we can't just ignore the dependence (that is pretend the destination is free) because the reassembly has to take care of moving `s3` first.

So the new codegen looks like:
```asmIN0008: 000000      stp     fp, lr, [sp, #-0x20]!
IN0009: 000004      mov     fp, sp
IN000a: 000008      mov     v3.s[3], v3.s[0]   // move s3 to the right spot, "freeing" v3
IN000b: 00000C      mov     v3.s[0], v0.s[0]
IN000c: 000010      mov     v3.s[1], v1.s[0]
IN000d: 000014      mov     v3.s[2], v2.s[0]

G_M8953_IG02:        ; offs=0x000018, size=0x001C, bbWeight=1, PerfScore 8.00, gcrefRegs=0000 {}, byrefRegs=0000 {}, BB01 [0000], byref

IN0001: 000018      mov     v0.16b, v3.16b
```
This reassembly for self-conflicts now happens in a prepass before the final param reg processing but after resolving circular dependences. The old code is still there to handle the non-self-conflicting cases.

jakobbotsch:
Ah, I see, ok then it makes sense to me that we may need some new code to handle this case.

AndyAyersMS:
Added a repro case that works with normal jitting.

@BruceForstall PTAL
cc @dotnet/jit-contrib 

(still might want to address other HFA types here...)

src/coreclr/jit/codegencommon.cpp
```diff
@@ -3341,6 +3344,26 @@ void CodeGen::genFnPrologCalleeRegArgs(regNumber xtraReg, bool* pXtraRegClobbere
...
+                    // If the field is passed in the same register as the destination,
+                    // but is in the wrong part of the register, mark it specially so later
+                    // we make sure to move it to the right spot before "freeing" the destination.
+                    //
+                    // If the arg is already in the right place we don't need to do anything special
+                    destRegNum = varDsc->GetRegNum();
+                    if (regNum == destRegNum)
+                    {
+                        const int slot = regArgTab[argNum].slot;
+                        if (slot != 1)
```
BruceForstall, line 3357: The condition above, `else if (regArgTab[argNum].slot == 1)` has already been checked. So won't this check always pass?

AndyAyersMS: Yes, looks like it will; let me change this to an assert.

BruceForstall (approved):
Thanks for the comments!

src/coreclr/jit/codegencommon.cpp
```diff
@@ -3823,6 +3846,56 @@ void CodeGen::genFnPrologCalleeRegArgs(regNumber xtraReg, bool* pXtraRegClobbere
         }
     }
 
+#if defined(TARGET_ARM64) && defined(FEATURE_SIMD)
+    // If we saw any hfa conflicts, handle those now
```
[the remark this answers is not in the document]
AndyAyersMS: Hopefully clearer now... I was sorely tempted to rewrite more of this method, but wanted something we could backport.

AndyAyersMS:
Thoughts on backporting this? Seems like this bug has been there forever, and no customer has reported hitting it.

BruceForstall:
> Thoughts on backporting this? Seems like this bug has been there forever, and no customer has reported hitting it.

At this stage of the release, because of the lack of customer reports (though perhaps something new is triggering it, like physical promotion?), and also because your code is a little uncertain about HVAs, I would lead against backporting it.

PR #92375, opened 2023-09-20
Allow multiple post-build steps and allow templated pre and post-build steps

jkoritzinsky:
Replace extraStepTemplate and extraStepParameters with a design similar to extraVariablesTemplates, but also allow non-templated steps and give the existing preBuildSteps customization the same options.

This is in preparation for merging our separate build jobs into individual per-runtime+target+configuration jobs.


eng/pipelines/common/global-build-job.yml
```diff
@@ -27,8 +27,7 @@ parameters:
   helixQueues: ''
   enablePublishTestResults: false
   testResultsFormat: ''
-  extraStepsTemplate: ''
-  extraStepsParameters: {}
+  postBuildSteps: []
```
radical, line 30: Should the name have a `Template` suffix like we have for `jobTemplate` in other places.

jkoritzinsky: I thought about it, but since we allow both template and non-template options, I wanted to match something closer to AzDO's naming. For plain AzDO YAML (without the absurd number of layers we have), the name of the parent is just `steps` can can be a template or a non-template.

radical:
runtime pipeline didn't start: `eng/pipelines/runtime.yml: (Line: 375, Col: 27, Idx: 14586) - (Line: 375, Col: 27, Idx: 14586): Mapping values are not allowed in this context.`

radical:
`runtime` pipeline seems to have failed to run with the same error as before.

eng/pipelines/coreclr/runtime-nativeaot-outerloop.yml
```diff
@@ -40,139 +40,276 @@ extends:
...
+              postBuildSteps:
+                - template: /eng/pipelines/coreclr/nativeaot-post-build-steps.yml
+                  parameters:
+                    creator: dotnet-bot
+                    testBuildArgs: 'nativeaot /p:IlcUseServerGc=false'
+                    liveLibrariesBuildConfig: Release
+              testRunNamePrefixSuffix: NativeAOT_Pri0_$(_BuildConfig)
+              extraVariablesTemplates:
+                - template: /eng/pipelines/common/templates/runtimes/test-variables.yml
+                  parameters:
```
agocke, line 200: Looks duplicated

jkoritzinsky: Removed the duplication

sbomer (approved):
LGTM, thanks!

hoyosjs (approved):
We don't need the equivalent change in the xplat job template? Essentially, we are moving most to the global build?

jkoritzinsky:
> We don't need the equivalent change in the xplat job template? Essentially, we are moving most to the global build?

Yep, my plan is to move most builds to the global build template model.

PR #92384, opened 2023-09-20
Ensure Bind can handle null from GetSection

### LINKED ISSUE #92213: NRE in ConfigurationBinder.BindInstance
### Description

This looks like a regression which happened between 8.0.0-rc.2.23431.9 (3c48925a6c1ab31865b4438a6cb88242d1a8fe4d) and 8.0.0-rc.2.23466.4 (287c10d2539d47268a1083c4d533cf84124900cf).

Updating the 1P internal codebase to the latest .NET Extensions, [several tests failed](https://domoreexp.visualstudio.com/R9/_build/results?buildId=23855283&view=ms.vss-test-web.build-test-results-tab&runId=186097722&paneView=debug&resultId=167506) with the following stack trace:
```
at Microsoft.Extensions.Configuration.ConfigurationBinder.BindInstance(Type type, BindingPoint bindingPoint, IConfiguration config, BinderOptions options, Boolean isParentCollection)
at Microsoft.Extensions.Configuration.ConfigurationBinder.BindProperty(PropertyInfo property, Object instance, IConfiguration config, BinderOptions options)
at Microsoft.Extensions.Configuration.ConfigurationBinder.BindProperties(Object instance, IConfiguration configuration, BinderOptions options)
at Microsoft.Extensions.Configuration.ConfigurationBinder.BindInstance(Type type, BindingPoint bindingPoint, IConfiguration config, BinderOptions options, Boolean isParentCollection)
at Microsoft.Extensions.Configuration.ConfigurationBinder.Bind(IConfiguration configuration, Object instance, Action`1 configureOptions)
at Microsoft.Extensions.Options.NamedConfigureFromConfigurationOptions`1.<>c__DisplayClass1_0.<.ctor>b__0(TOptions options)
at Microsoft.Extensions.Options.ConfigureNamedOptions`1.Configure(String name, TOptions options)
at Microsoft.Extensions.Options.OptionsFactory`1.Create(String name)
at Microsoft.Extensions.Options.UnnamedOptionsManager`1.get_Value()
```

### Reproduction Steps

I'm not familiar with this area of the runtime, nor I'm sufficiently familiar with the 1P codebase I'm updating to craft a standalone repro, so we may need to debug it interactively since this is an internal codebase.

### Expected behavior

No NRE, everything continues work as before :)

### Actual behavior

Tests fail

### Regression?

Yes

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

/cc: @geeknoid @joperezr

ericstj:
Fixes https://github.com/dotnet/runtime/issues/92213

IConfiguration instances may return a null value from GetSection. We were not handling this and would throw a NullReferenceException.

### DIFF
--- a/src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
+++ b/src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
@@ -298,6 +298,11 @@ namespace Microsoft.Extensions.Configuration
                 return;
             }
 
+            if (config == null)
+            {
+                return;
+            }
+
             var section = config as IConfigurationSection;
             string? configValue = section?.Value;
             if (configValue != null && TryConvertValue(type, configValue, section?.Path, out object? convertedValue, out Exception? error))
--- a/src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
+++ b/src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
@@ -298,7 +298,7 @@ namespace Microsoft.Extensions.Configuration
                 return;
             }
 
-            if (config == null)
+            if (config is null)
             {
                 return;
             }
--- a/src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.cs
+++ b/src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.cs
@@ -1768,7 +1768,7 @@ if (!System.Diagnostics.Debugger.IsAttached) { System.Diagnostics.Debugger.Launc
             Assert.Equal(0, options.OtherCodeNullable);
             Assert.Equal("default", options.OtherCodeString);
             Assert.Null(options.OtherCodeNull);
-            Assert.Null(options.OtherCodeUri);            
+            Assert.Null(options.OtherCodeUri);
         }
 
         [Fact]
@@ -2239,7 +2239,7 @@ if (!System.Diagnostics.Debugger.IsAttached) { System.Diagnostics.Debugger.Launc
                 Assert.Throws<ArgumentNullException>(() => configuration.GetValue(typeof(GeolocationClass), key, new GeolocationClass()));
                 Assert.Throws<ArgumentNullException>(() => configuration.GetValue(typeof(Geolocation), key));
                 Assert.Throws<ArgumentNullException>(() => configuration.GetValue(typeof(Geolocation), key, defaultValue: null));
-                Assert.Throws<ArgumentNullException>(() => configuration.GetValue(typeof(Geolocation), key, default(Geolocation)));    
+                Assert.Throws<ArgumentNullException>(() => configuration.GetValue(typeof(Geolocation), key, default(Geolocation)));
             }
         }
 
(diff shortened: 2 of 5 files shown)


src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
```diff
@@ -298,6 +298,11 @@ private static void BindProperty(PropertyInfo property, object instance, IConfig
                 return;
             }
 
+            if (config == null)
```
ericstj, line 301: Previously the entire block below was wrapped in a null check - https://github.com/dotnet/runtime/commit/35f043f414dbd234978f0eb694b47dfb42dd2031#diff-2e6f75e1d4f577e1c8c1b6a32d158ce7812a38ad8fcd3b4d762157b92c03e890L308-R315

Since this method already has multiple returns, I find it more readable to early-exit rather than count curly-brackets/indentation.

RussKie:
Thank you!

src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
tarekgh: sounds good to me.

RussKie, line 301: Should this be ❔
```suggestion
            if (config is null)
```

tarekgh (approved):
Added minor comment/suggestion, LGTM otherwise. Thanks for fixing this.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.cs
```diff
@@ -2404,5 +2405,27 @@ public void SharedChildInstance()
             config.GetSection("A").Bind(instance);
             Assert.Equal("localhost", instance.ConnectionString);
         }
+
+        [Fact]
+        public void CanBindToMockConfiugrationSection()
```
tarekgh, line 2410: Looks need to exclude the test for wasm?

ericstj:
@RussKie - if you wanted to workaround this you can register a mock section as the default return value for `GetSection` as follows:
```c#
mockConfSection.Setup(m => m.GetSection(It.IsAny<string>())).Returns(emptySection.Object);
```

That would ensure that GetSection doesn't return `null`.   We definitely want to fix this though, since it's a regression.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.cs
```diff
@@ -2404,5 +2405,27 @@ public void SharedChildInstance()
...
+        {
+            double expectedLatitude = 42.0002d;
```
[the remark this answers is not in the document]
ericstj: I thought of that, however I noticed that we already did this comparison in the same file: https://github.com/dotnet/runtime/blob/736dabeca728ccf8b911d96d1b4c575b4d0db7d2/src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.cs#L2131-L2132

I opted for not creating a new type 🤷‍♂️

RussKie:
> @RussKie Igor Velikorossov FTE - if you wanted to workaround this you can register a mock section as the default return value for `GetSection` as follows:
> 
> ```cs
> mockConfSection.Setup(m => m.GetSection(It.IsAny<string>())).Returns(emptySection.Object);
> ```
> 
> That would ensure that GetSection doesn't return `null`. We definitely want to fix this though, since it's a regression.

👍

src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
ericstj: Sure 👍

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.cs
ericstj: Yep will do this: https://github.com/dotnet/runtime/blob/736dabeca728ccf8b911d96d1b4c575b4d0db7d2/src/libraries/Microsoft.Extensions.Primitives/tests/CompositeChangeTokenTest.cs#L45-L46

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.cs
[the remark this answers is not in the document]
ericstj: I just went ahead and created a simple poco for this.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.cs
```diff
@@ -2404,5 +2405,29 @@ public void SharedChildInstance()
             config.GetSection("A").Bind(instance);
             Assert.Equal("localhost", instance.ConnectionString);
         }
+
+        // Moq heavily utilizes RefEmit, which does not work on most aot workloads
+        [ConditionalFact(typeof(PlatformDetection), nameof(PlatformDetection.IsReflectionEmitSupported))]
+        public void CanBindToMockConfiugrationSection()
```
stephentoub, line 2411:
```suggestion
        public void CanBindToMockConfigurationSection()
```

src/libraries/Microsoft.Extensions.Configuration.Binder/src/ConfigurationBinder.cs
```diff
@@ -298,6 +298,11 @@ private static void BindProperty(PropertyInfo property, object instance, IConfig
                 return;
             }
 
+            if (config is null)
+            {
+                return;
+            }
+
```
krwq, line 305: do we understand why this didn't repro earlier? it could mean there is some other bug as well

tarekgh: @krwq this is understood. Please have a look at the @ericstj comment https://github.com/dotnet/runtime/pull/92384#pullrequestreview-1636607530 which is pointing at the change that caused this regression.

ericstj: Yep, the null check was removed

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.cs
ericstj: So that didn't seem to be enough to fix it since the failure appears to be happening when the tool chain runs on the test.  I'll look closer at other tests.  I really don't want to disable the entire assembly, nor do I want to implement an verbose hand-written mock configuration for this.

ericstj: I tried a hand-written mock and was able to get it to work.  I'll use this method since adding the Moq dependency makes us lose test coverage.

tarekgh: Sounds good.

tarekgh (approved):

ericstj:
Remaining legs are outer loop build incorrectly triggered for all PRs

PR #92385, opened 2023-09-20
Provide ALC name in DispatchProxy for proxies in custom ALC

### LINKED ISSUE #82969: Provide ALC name in DispatchProxy for proxies in custom ALC
With https://github.com/dotnet/runtime/pull/62095 issues with proxies loaded from custom ALC is fixed and the ALC name was passed to the `AssemblyName` used for creating the Proxy type, but [that was causing issues for ALC that include paths in its name](https://github.com/dotnet/runtime/issues/80387). We reverted that part as it was causing a regression. 

Though passing ALC name to the proxy assembly could be useful for debugging purposes and @jkotas suggested a good way for fixing it in https://github.com/dotnet/runtime/pull/82807#discussion_r1125160686:

pedrobsaila:
Fixes #82969

### DIFF
--- a/src/coreclr/vm/assembly.cpp
+++ b/src/coreclr/vm/assembly.cpp
@@ -397,11 +397,6 @@ Assembly *Assembly::CreateDynamic(AssemblyBinder* pBinder, NativeAssemblyNamePar
     if (pAssemblyNameParts->_pName == NULL || pAssemblyNameParts->_pName[0] == '\0')
         COMPlusThrow(kArgumentException, W("ArgumentNull_AssemblyNameName"));
 
-    if (COMCharacter::nativeIsWhiteSpace(pAssemblyNameParts->_pName[0]))
-    {
-        COMPlusThrow(kArgumentException, W("InvalidAssemblyName"));
-    }
-
     // Set up the assembly manifest metadata
     // When we create dynamic assembly, we always use a working copy of IMetaDataAssemblyEmit
     // to store temporary runtime assembly information. This is to preserve the invariant that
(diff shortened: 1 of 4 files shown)


src/libraries/System.Reflection/tests/AssemblyNameTests.cs
```diff
@@ -86,7 +86,6 @@ public void Ctor_String_Public_Key(string name, string expectedName)
         [InlineData("", typeof(ArgumentException))]
         [InlineData("\0", typeof(ArgumentException))]
         [InlineData("\0a", typeof(ArgumentException))]
-        [InlineData("/a", typeof(FileLoadException))]
```
jkotas, line 89: These tests should be replaced by positive tests

src/coreclr/vm/assembly.cpp
```diff
@@ -397,10 +397,7 @@ Assembly *Assembly::CreateDynamic(AssemblyBinder* pBinder, NativeAssemblyNamePar
     if (pAssemblyNameParts->_pName == NULL || pAssemblyNameParts->_pName[0] == '\0')
         COMPlusThrow(kArgumentException, W("ArgumentNull_AssemblyNameName"));
 
-    if (COMCharacter::nativeIsWhiteSpace(pAssemblyNameParts->_pName[0])
-        || u16_strchr(pAssemblyNameParts->_pName, '\\') != NULL
-        || u16_strchr(pAssemblyNameParts->_pName, ':') != NULL
-        || u16_strchr(pAssemblyNameParts->_pName, '/') != NULL)
+    if (COMCharacter::nativeIsWhiteSpace(pAssemblyNameParts->_pName[0]))
```
jkotas, line 400: Delete this check as well?

jkotas (approved):
Thank you!

PR #92393, opened 2023-09-21
[llvm] Fix the IR generated for returning small valuetypes by value.

vargaz:
We were generating loads larger than the size of the alloca holding the valuetype.

### DIFF
--- a/src/mono/mono/mini/mini-llvm.c
+++ b/src/mono/mono/mini/mini-llvm.c
@@ -6007,7 +6007,23 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
 							retval = LLVMBuildInsertValue (builder, retval, element, i, "setret_simd_vtype_in_reg");
 						}
 					} else {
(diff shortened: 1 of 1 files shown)


jandupej:
Can we optimize this by using zero-extension for valuetypes with nice sizes (1,2,4 B) and falling back to this memory-based approach with other sizes?

lateralusX:
Moving the alloca into the entry_bb made wonders to the generated code, we now get the following assembler for the case where argument is null:

```
00007FF6BD348A70  test        rdx,rdx  
00007FF6BD348A73  je           07FF6BD348AC0h
...
00007FF6BD348AC0  xor         eax,eax  
00007FF6BD348AC2  add         rsp,28h  
00007FF6BD348AC6  ret
```

lateralusX (approved):
LGTM!

vargaz:
Does this need a backport ?

lateralusX:
I would say yes, we have repros so it can be hit, and if hit, side effects are hard to understand and track and can cause crashes since it can return uninitialized memory.

PR #92396, opened 2023-09-21
JIT: Fixed containment of STOREIND of HW intrinsics ConvertTo*

### LINKED ISSUE #92349: Incorrect value after truncating cast
### Description

When running a Release build, performing a `(ushort)(uint)` cast on `int` and storing the result in a `ushort*` seems to write out 32 bits instead of 16 bits.

### Reproduction Steps

1. Clone repo https://github.com/TechPizzaDev/SoftwareRasterizer
2. Checkout branch `net8-upgrade`
3. Download *Castle folder* from https://github.com/rawrunprotected/rasterizer/tree/master/SoftwareRasterizer and put next to built SoftwareRasterizer.exe
4. Run project **in Release mode** to see broken render (*it works in Debug mode*)
5. Remove `BUG` define (https://github.com/TechPizzaDev/SoftwareRasterizer/blob/318adc376bcdaee1b6c863dd9384871aed82897e/SoftwareRasterizer/Avx2Rasterizer.cs#L2)
6. Run project again to see correct render

### Expected behavior

Correct raster like in net7:
image

### Actual behavior

image

### Regression?

Regression in net8, worked on net7 and net6.

### Known Workarounds

Rewriting the code seems like the only workaround. The project also works in Debug mode.

### Configuration

.NET: 8.0.0-rc.1.23419.4
OS: Microsoft Windows 11 Home (10.0.22621 version 22621)
CPU: 11th Gen Intel(R) Core(TM) i3-1115G4 @ 3.00GHz
Arch: x64

#### Env variables
**DOTNET_EnableAVX512F=0**: Required since net8 seems to cause unexpected AccessViolationExceptions.

### Other information

`Sse41.MinHorizontal(Vector128<ushort>)` returns a `Vector128<ushort>` where the first `ushort` element is the minimum value from the vector, and the second `ushort` element is the index of the min element within the vector.

Problematic code: https://github.com/TechPizzaDev/SoftwareRasterizer/blob/318adc376bcdaee1b6c863dd9384871aed82897e/SoftwareRasterizer/Avx2Rasterizer.cs#L1262-L1264

TIHan:
Resolves https://github.com/dotnet/runtime/issues/92349 regarding the truncating CAST

## Description

As an example, the HW Intrinsic SIMD scalar `ConvertToInt32` will emit a `vmovd` instruction. This instruction can store a scalar value in memory or in a register, but only for 32 and 64 bits.

So, when we have something like this:
```csharp
// pValue is a 'ushort*'
*pValue = (ushort)(uint)Sse2.ConvertToInt32(newMinZ16);
```

## Acceptance Criteria

- [x] Add test case

### Note

We need to backport this to .NET 8.

### DIFF
--- a/src/coreclr/jit/gentree.cpp
+++ b/src/coreclr/jit/gentree.cpp
@@ -26448,3 +26448,38 @@ bool GenTree::CanDivOrModPossiblyOverflow(Compiler* comp) const
     // Not enough known information; therefore we might overflow.
     return true;
 }
+
+//------------------------------------------------------------------------
+// OperIsHWIntrinsicSIMDScalar: returns true if the given tree is GT_HWINTRINSIC and
+//                              is a SIMD scalar operation
+//
+// Return Value:
+//    true if the given tree is a hwintrinsic and a SIMD scalar operation
+//
+bool GenTree::OperIsHWIntrinsicSIMDScalar()
+{
+    if (!this->OperIsHWIntrinsic())
+        return false;
+
+#if defined(FEATURE_HW_INTRINSICS)
+    GenTreeHWIntrinsic* hwintrinsic = this->AsHWIntrinsic();
+    NamedIntrinsic      intrinsicId = hwintrinsic->GetHWIntrinsicId();
+
+#if defined(TARGET_AMD64)
+    HWIntrinsicCategory category = HWIntrinsicInfo::lookupCategory(hwintrinsic->GetHWIntrinsicId());
+
+    switch (category)
+    {
+        case HW_Category_SIMDScalar:
+            return true;
+
+        default:
+            break;
+    }
+#elif defined(TARGET_ARM64)
+    return HWIntrinsicInfo::SIMDScalar(intrinsicId);
+#endif // TARGET_ARM64 && !TARGET_AMD64
+#endif // FEATURE_HW_INTRINSICS
+
+    return false;
+}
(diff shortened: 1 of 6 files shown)


TIHan:
cc @tannergooding

jakobbotsch:
> We need to keep the cast operation for `int -> short`, otherwise we will get an incorrect result which is what #92349 was telling us.
> 
> I tried to see if there was something we could do in codegen, but it is a lot easier to handle in morph where we get rid of these CASTs. The simplest thing to do is check to see if we have a SIMD scalar operation and we are storing the result for a small type, then do not remove the CAST.

We should fix this in the backend. `STOREIND<ushort>` should never be storing more than two bytes. With this PR the bug still exists, it's just harder to expose (but can definitely be done, e.g. you can write IR manually that doesn't create these casts in the first place).

It looks like a bug in `ContainCheckStoreIndir`. This line:
https://github.com/dotnet/runtime/blob/aba7c43598dc13e7e43843dceb57a397f13107cb/src/coreclr/jit/lowerxarch.cpp#L6507

should be
```
isContainable = varTypeIsIntegral(simdBaseType) && (genTypeSize(src) == genTypeSize(node));
```

if `vmovd` has no 8/16 bit forms.

TIHan:
> We should fix this in the backend

Your fix works and is a lot simpler than what I had. Thank you.

jakobbotsch (commented):
Add a test?

TIHan:
@dotnet/jit-contrib @jakobbotsch @tannergooding This is ready, pending CI.

jakobbotsch (approved):

tannergooding (approved):

PR #92407, opened 2023-09-21
Convert all remaining tests in the Loader subtree to the merged model

trylek:


src/tests/Loader/CollectibleAssemblies/ByRefLocals/Unloaded.csproj
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <OutputType>Library</OutputType>
-    <DisableProjectBuild>true</DisableProjectBuild>
+    <RequiresProcessIsolation>true</RequiresProcessIsolation>
```
markples, line 5: Library and RPI seem contradictory (though I think this is the case where Library is the default but one of our files overwrites it when RPI is set)

src/tests/Loader/CollectibleAssemblies/Statics/Unloaded.csproj
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <OutputType>Library</OutputType>
-    <DisableProjectBuild>true</DisableProjectBuild>
+    <RequiresProcessIsolation>true</RequiresProcessIsolation>
```
markples, line 5: same

markples (approved):

src/tests/Loader/CollectibleAssemblies/ByRefLocals/Unloaded.csproj
trylek: I agree it's somewhat dirty but it was the easiest way to suppress the "duplicate project name" error - I can rename the two projects if you find that preferable but it will require changes to the test themselves as their source code hardcodes the assembly name. I'll be happy for any suggestions how to address this in a cleaner way.

markples: Oh.. can you confirm that this is what is happening - 

ByRefLocals and Statics both have (unrelated) libraries called Unloaded.  They work because they are marked RPI and get different directories and then msbuild copies their references into those directories.  (Otherwise, the merged test group would have two Unloaded libraries.)  However, the merged test group would still get two Unloaded libraries (the duplicates error), so this stops them from being included in that directory.  These are likely getting simple wrappers generated with zero actual tests in them (weird but harmless).

I'm definitely not sure about this, but perhaps they could be removed from the overall MergedWrapperProjectReference.  This may be a lurking problem if RPI is removed from the actual tests here (though maybe that's not possible due to the tests themselves) or if the output directory structure changes, so renaming one might be useful anyway.

trylek: OK, thanks for your feedback, I agree with you that it's probably going to be cleanest to deduplicate the name of the Unloaded library, I'll do that as the next commit.

src/tests/Loader/CollectibleAssemblies/Statics/Unloaded.csproj
trylek: Fixed in 14th commit.

src/tests/Loader/CollectibleAssemblies/ByRefLocals/Unloaded.csproj
trylek: Fixed in 14th commit.

trylek:
Mono errors in the recent runs were caused by a subtle bug in the method ```GetFilteredExcludeList``` at the end of ```issues.targets``` - in the presence of files with multiple dots in the name ("multiple extensions") is was dropping everything after the first dot. I have pushed the fix to issues.targets as the last commit on this PR, outerloop is green, I plan to merge this in if the Mono legs end up green after this change (not too hopeful, something seems to be crashing the LLVMAOT compiler, I'll look into that once the job finishes).

trylek:
Outerloop is green and runtime-extra-platforms doesn't seem to fare much worse than in a recent Steve Toub's PR so I'm about to merge this in.

PR #90253, opened 2023-08-09
[QUIC] `QuicStream` reading/writing work

ManickaP:
So far addressed:
- https://github.com/dotnet/runtime/issues/79911 - tested on repro (5+ mins), fixed
- https://github.com/dotnet/runtime/issues/82704 - tested on repro (5+ mins), fixed
- https://github.com/dotnet/runtime/issues/82523 - tested on repro (5+ mins), fixed

Remain as-is:
- https://github.com/dotnet/runtime/issues/79818
- https://github.com/dotnet/runtime/issues/77216

Closes #77216
Fixes #79911
Fixes #82704
Closes #79818
Fixes #82523

Also fixed problem with GC eating `QuicStream` when only either `ReadsClosed` or `WritesClosed` is awaited. Improved some logging. Added bunch of tests.

Affected tests were locally ran with iteration count 100, repro cases ran for 5+ minutes, ran ASP.NET Core tests locally and running stress test in docker locally now (result will be later).

cc @CarnaViire @rzikm @wfurt

PR #92410, opened 2023-09-21
[H/3] Remove waiting on tasks from H3 request stream dispose

### LINKED ISSUE #92390: HTTP/3: Hang when reading content or disposing response when SerializeToStreamAsync doesn't complete
### Description

Dependency update from https://github.com/dotnet/aspnetcore/pull/50362 caused new behavior in some tests.

Some tests hang on reading to the end of the response:

* `POST_ServerCompletesWithoutReadingRequestBody_ClientGetsResponse` hangs [here](https://github.com/dotnet/aspnetcore/blob/57d703606439dccabec92c2a0d5e4fa618985cd8/src/Servers/Kestrel/test/Interop.FunctionalTests/Http3/Http3RequestTests.cs#L437)

Some tests hang on disposing the response:

* `POST_ClientCancellationBidirectional_RequestAbortRaised(protocol: Http3)` hangs [here](https://github.com/dotnet/aspnetcore/blob/2ce54c68b2abfa66974a4e75cf80e61203180ce4/src/Servers/Kestrel/test/Interop.FunctionalTests/Http3/Http3RequestTests.cs#L1035)

Consistent between the tests is we have streaming request content via [StreamingHttpContent](https://github.com/dotnet/aspnetcore/blob/57d703606439dccabec92c2a0d5e4fa618985cd8/src/Servers/Kestrel/test/Interop.FunctionalTests/Http3/Http3RequestTests.cs#L35-L70). In the tests we're not completing the request stream so `SerializeToStreamAsync` doesn't exit.

Note: The implementation of `SerializeToStreamAsync` is [listening to the cancellation token](https://github.com/dotnet/aspnetcore/blob/57d703606439dccabec92c2a0d5e4fa618985cd8/src/Servers/Kestrel/test/Interop.FunctionalTests/Http3/Http3RequestTests.cs#L49-L52). I don't know if it was canceled before and isn't now, or if it should be canceled at all, but it isn't when I debug the tests.

It seems like wrong behavior for the request stream not being completed to cause a hang when reading to the end of response data (the server in `POST_ServerCompletesWithoutReadingRequestBody_ClientGetsResponse` has [sent its data and completed](https://github.com/dotnet/aspnetcore/blob/2ce54c68b2abfa66974a4e75cf80e61203180ce4/src/Servers/Kestrel/test/Interop.FunctionalTests/Http3/Http3RequestTests.cs#L403-L407)), or aborting the response via dispose.

### Reproduction Steps

See tests above.

### Expected behavior

Either:
* The cancellation token to `SerializeToStreamAsync` is canceled when the response finishes or the `HttpResponseMessage.Dispose()` is called, or
* `SerializeToStreamAsync` not exiting doesn't hang finishing reading the response or aborting

### Actual behavior

Hangs

### Regression?

Worked in .NET 8

### Known Workarounds

_No response_

### Configuration

.NET 9 latest from https://github.com/dotnet/aspnetcore/pull/50362

### Other information

_No response_

ManickaP:
The waiting on tasks was added due to HTTP stress crashing on `QuicStream` asserts:
https://github.com/dotnet/runtime/blob/9c8ff4c1c12c488da660739333e36eb55afe9c56/src/libraries/System.Net.Quic/src/System/Net/Quic/QuicStream.cs#L716-L717
H3 stream is issuing reads and writes on QuicStream in parallel. So if one ends on cancellation and decides to dispose the stream, the DisposeAsync might run in parallel with pending Write/Read.
Basically, this change reverts these to commits: https://github.com/dotnet/runtime/pull/90253/commits/d6f50614018da68494bd97672713b59be89c8512, https://github.com/dotnet/runtime/pull/90253/commits/f6261e94aaea6518ad2d1d5d35d6758877183c51 from https://github.com/dotnet/runtime/pull/90253.

Fixes #92390

The asserts still need to addressed, but we don't have a running stress atm.

cc @JamesNK

### DIFF
--- a/src/libraries/System.Net.Quic/src/System/Net/Quic/QuicStream.cs
+++ b/src/libraries/System.Net.Quic/src/System/Net/Quic/QuicStream.cs
@@ -713,8 +713,9 @@ public sealed partial class QuicStream
             await valueTask.ConfigureAwait(false);
         }
         Debug.Assert(_startedTcs.IsCompleted);
-        Debug.Assert(_receiveTcs.IsCompleted);
-        Debug.Assert(_sendTcs.IsCompleted);
+        // TODO: If there was and ongoing read/write when dispose was called, MsQuic events might be processed, but the continuations may not thus the state can still be in Ready.
+        //Debug.Assert(_receiveTcs.IsCompleted);
+        //Debug.Assert(_sendTcs.IsCompleted);
         _handle.Dispose();
 
         // TODO: memory leak if not disposed
--- a/src/libraries/Common/tests/System/Net/Http/Http3LoopbackStream.cs
+++ b/src/libraries/Common/tests/System/Net/Http/Http3LoopbackStream.cs
@@ -196,7 +196,7 @@ namespace System.Net.Test.Common
             }
         }
 
-        public async Task<byte[]> ReadRequestBodyAsync()
+        public async Task<byte[]> ReadRequestBodyAsync(int minimumBytes = -1)
         {
             var buffer = new MemoryStream();
 
@@ -211,6 +211,10 @@ namespace System.Net.Test.Common
                         break;
                     case null:
                         return buffer.ToArray();
+                }
+                if (minimumBytes >= 0 && buffer.Length >= minimumBytes)
+                {
+                    return buffer.ToArray();
                 }
             }
         }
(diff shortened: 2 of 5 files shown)


ManickaP:
I've run locally all the tests from https://github.com/dotnet/aspnetcore/issues/50833 with the fix and they pass promptly, no timeout, no hangs.
Running whole test suite reports some H2 tests failing, but they seem to be quarantined against different issues, so hopefully I did my due diligence properly.

JamesNK (approved):

src/libraries/System.Net.Quic/src/System/Net/Quic/QuicStream.cs
```diff
@@ -713,8 +713,9 @@ public override async ValueTask DisposeAsync()
             await valueTask.ConfigureAwait(false);
         }
         Debug.Assert(_startedTcs.IsCompleted);
-        Debug.Assert(_receiveTcs.IsCompleted);
-        Debug.Assert(_sendTcs.IsCompleted);
+        // TODO: If there was and ongoing read/write when dispose was called, MsQuic events might be processed, but the continuations may not thus the state can still be in Ready.
+        //Debug.Assert(_receiveTcs.IsCompleted);
```
wfurt, line 717: I'm not bug fan of submitting comments like this. same bellow.

PR #92414, opened 2023-09-21
Ensure that embedded broadcast checks the base type of the parent node

tannergooding:
`IsContainableHWIntrinsicOp` was only checking the base type of the broadcast node, which meant that scenarios such as `MultiplyLow(vector256Int16, Vector256.Create(int32).AsInt16())` would incorrectly try to contain the child node.

This ensures that we check both the base type of the broadcast (which indicates how many bytes it reads from memory) and the base type of the parent node (which indicates how many bytes it would read from memory).

### DIFF
--- a/src/coreclr/jit/lowerxarch.cpp
+++ b/src/coreclr/jit/lowerxarch.cpp
@@ -8205,7 +8205,7 @@ bool Lowering::IsContainableHWIntrinsicOp(GenTreeHWIntrinsic* parentNode, GenTre
             var_types parentBaseType = parentNode->GetSimdBaseType();
             var_types childBaseType  = hwintrinsic->GetSimdBaseType();
 
-            if (varTypeIsSmall(parentBaseType) || varTypeIsSmall(childBaseType))
+            if (varTypeIsSmall(parentBaseType) || (genTypeSize(parentBaseType) != genTypeSize(childBaseType)))
             {
                 // early return if either base type is not embedded broadcast compatible.
                 return false;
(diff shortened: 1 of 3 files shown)


src/coreclr/jit/lowerxarch.cpp
```diff
@@ -8202,18 +8202,20 @@ bool Lowering::IsContainableHWIntrinsicOp(GenTreeHWIntrinsic* parentNode, GenTre
         case NI_AVX2_BroadcastScalarToVector256:
         case NI_AVX512F_BroadcastScalarToVector512:
         {
-            var_types baseType = hwintrinsic->GetSimdBaseType();
-            if (varTypeIsSmall(baseType))
+            var_types parentBaseType = parentNode->GetSimdBaseType();
+            var_types childBaseType  = hwintrinsic->GetSimdBaseType();
+
+            if (varTypeIsSmall(parentBaseType) || varTypeIsSmall(childBaseType))
```
jakobbotsch, line 8208: Is this sufficient? For example, doesn't
```csharp
public static Vector256<long> Test(Vector256<long> x, ref int y)
{
    return Avx2.Add(x + x, Vector256.Create(y).AsInt64());
}
```
have the same problem and still pass these checks?

tannergooding: Yes, good point. Will do a check that the sizes match here.

There's a "todo" in that it is safe for some constants, but that's a different path and is already handled.

tannergooding: Fixed.

jakobbotsch (approved):

PR #92427, opened 2023-09-21
Add trim analysis warnings for delegate creation

### LINKED ISSUE #84918: [feature/damAnalyzer] Action delegate is not handled correctly in the dataflow
Ldfn calls should end up resolving into a method call that generated a warning for the analyzer. See for example [code](https://github.com/dotnet/linker/blob/5200664cf29083ca64a65c7b792eafffbb859690/test/Mono.Linker.Tests.Cases/DataFlow/AnnotatedMembersAccessedViaReflection.cs#L207-L211)

sbomer:
The ILLink analyzer was not producing warnings for annotated methods, lambdas, or local functions accessed for delegate creation.

Some ExpectedWarnings for ILLink/NativeAot were updated to avoid relying on the name of the compiler-generated methods for lambdas or local functions, because the analyzer reports the warnings as:

IL2111: Method 'lambda expression' with parameters... is accessed via reflection.

Fixes https://github.com/dotnet/runtime/issues/84918

### DIFF
--- a/src/tools/illink/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
+++ b/src/tools/illink/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Immutable;
 using System.Diagnostics;
+using ILLink.RoslynAnalyzer.TrimAnalysis;
 using ILLink.Shared.DataFlow;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.FlowAnalysis;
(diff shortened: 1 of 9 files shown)


src/tools/illink/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -421,6 +422,32 @@ public override TValue VisitExpressionStatement (IExpressionStatementOperation o
...
+				// lambda or local function has an annotated this parameter.
+				var instance = TopValue;
+				return HandleDelegateCreation (lambda.Symbol, instance, operation);
+			}
+
+			Debug.Assert (operation.Target is IMethodReferenceOperation);
+			if (operation.Target is not IMethodReferenceOperation methodReference)
+				return TopValue;
+
+			TValue instanceValue = Visit (methodReference.Instance, state);
```
vitek-karas, line 441: Will this actually report IL2026 if the target method has RUC on it?
We never visit the method itself, we only visit the instance.

But I guess the RUC analyzer will do that instead, right? (we should really merge the two into one... ;-))

vitek-karas (approved):
Other than the RUC question - this looks good.

src/tools/illink/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
vitek-karas: Nice :-)

PR #92429, opened 2023-09-21
[browser] move files around

pavelsavara:
Introducing directories 
- `core`
- `jiterpreter`
- `interop`

and moving TS files into them.


campersau:
I guess this is going to conflict with #91699

pavelsavara:
> I guess this is going to conflict with #91699

It was not that bad :)

maraf:
What about splitting the code by module?
- runtime
  - diagnostics
  - jiterpreter
  - ...
- loader
  - run.ts
  - ...
- emscripten
  - dotnet.lib.js
  - dotnet.pre.js
  - ...

What about moving typescript/js code to a separate folde (like client/client-side)?

pavelsavara:
> What about splitting the code by module?

That's exactly the idea, and we could do better than I did so far.

I did `core` for `dotnet.runtime.js` because I didn't like `runtime/runtime`  folder. We could rename the outer folder.

Rename of `es6` to `emscripten` is good idea, I forgot about that. But maybe we could do `native` instead.

pavelsavara:
I need to work on something else ... maybe later

PR #92430, opened 2023-09-21
HashAlgorithmName one-shots

### LINKED ISSUE #91407: [API Proposal]: Hash and HMAC one-shots for HashAlgorithName
### Background and motivation

Over the past few releases of .NET, we've added a number of "one shots" for cryptographic hashing and HMACs (#62489, #40012, #17590).

All of these one-shots live off of their respective type. For a SHA1 one-shot, use `SHA1`, for HMAC-SHA256, use `HMACSHA256.HashData`, etc.

There is one other mechanism for producing HMACs and hashes, which is `HashAlgorithmName`. This type is an identifier for the hash, but does not implement the hash itself. This type is fed in to places like `IncrementalHash`, `AsymmetricAlgorithm.{SignData,VerifyData}`, etc. This is useful when libraries want to allow the caller to specify the hash algorithm used.

If you want to use this in conjunction with the hashing one-shots, you currently need to switch over the `HashAlgorithmName`.

```C#
byte[] hash = myHashAlgorithmName.Name switch {
    "SHA1" => SHA1.HashData(data),
    "SHA256" => SHA256.HashData(data),
    // etc
    _ => throw new ArgumentException("helpful message"),
}
```

This has a number of undesirable properties:

1. With the introduction of SHA-3, the number of switch cases is now 8.
2. The internal hashing mechanism, `HashProviderDispenser`, actually prefers working with names. So this switch logic is inefficient, since we start out with a name, convert it to a static type, which then just changes it to the name. So the runtime can actually provide a better implementation by going directly to the internal `HashProviderDispenser`.
3. The switching over the names pays a native AOT size cost. If a library author implements this switch case, but their consumers only ever use `HashAlgorithmName.SHA256`, they end up rooting all of the static hash APIs, regardless of if they are used. This was identified in #91181

We have our own internal implementation of this called `HashOneShotHelpers`. It's used dozens of times, so there is a clear need for it ourselves, others will benefit from it.

### API Proposal

This proposal is identical to all of the hashing one-shots on SHA{n} and HMACSHA{n}, with a `HashAlgorithmName` as the first parameter.

```C#
namespace System.Security.Cryptography;

// Existing class
public static partial class CryptographicOperations {
        public static byte[] HmacData(HashAlgorithmName hashAlgorithm, byte[] key, byte[] source);
        public static byte[] HmacData(HashAlgorithmName hashAlgorithm, ReadOnlySpan<byte> key, ReadOnlySpan<byte> source);
        public static int HmacData(HashAlgorithmName hashAlgorithm, ReadOnlySpan<byte> key, ReadOnlySpan<byte> source, Span<byte> destination);
        public static bool TryHmacData(HashAlgorithmName hashAlgorithm, ReadOnlySpan<byte> key, ReadOnlySpan<byte> source, Span<byte> destination, out int bytesWritten);

        public static byte[] HmacData(HashAlgorithmName hashAlgorithm, byte[] key, Stream source);
        public static byte[] HmacData(HashAlgorithmName hashAlgorithm, ReadOnlySpan<byte> key, Stream source);
        public static int HmacData(HashAlgorithmName hashAlgorithm, ReadOnlySpan<byte> key, Stream source, Span<byte> destination);
        public static ValueTask<byte[]> HmacDataAsync(HashAlgorithmName hashAlgorithm, byte[] key, Stream source, CancellationToken cancellationToken = default);
        public static ValueTask<int> HmacDataAsync(HashAlgorithmName hashAlgorithm, ReadOnlyMemory<byte> key, Stream source, Memory<byte> destination, CancellationToken cancellationToken = default);
        public static ValueTask<byte[]> HmacDataAsync(HashAlgorithmName hashAlgorithm, ReadOnlyMemory<byte> key, Stream source, CancellationToken cancellationToken = default);

        public static byte[] HashData(HashAlgorithmName hashAlgorithm, byte[] source);
        public static byte[] HashData(HashAlgorithmName hashAlgorithm, ReadOnlySpan<byte> source);
        public static int HashData(HashAlgorithmName hashAlgorithm, ReadOnlySpan<byte> source, Span<byte> destination);
        public static bool TryHashData(HashAlgorithmName hashAlgorithm, ReadOnlySpan<byte> source, Span<byte> destination, out int bytesWritten);

        public static byte[] HashData(HashAlgorithmName hashAlgorithm, Stream source);
        public static int HashData(HashAlgorithmName hashAlgorithm, Stream source, Span<byte> destination);
        public static ValueTask<int> HashDataAsync(HashAlgorithmName hashAlgorithm, Stream source, Memory<byte> destination, CancellationToken cancellationToken = default);
        public static ValueTask<byte[]> HashDataAsync(HashAlgorithmName hashAlgorithm, Stream source, CancellationToken cancellationToken = default);
}
```

### API Usage

```csharp
public static void SomeLibraryMethod(HashAlgorithmName hashAlgorithm) {
    byte[] digest = CryptographicOperations.HashData(hashAlgorithm, "hash this"u8);
}
```

### Alternative Designs

_No response_

### Risks

_No response_

vcsjones:
This implements CryptographicOperations.{HashData,HmacData} and removes our internal helper in favor of using the public one.

Closes #91407.


src/libraries/System.Security.Cryptography/tests/HmacMD5Tests.cs
```diff
@@ -15,9 +15,11 @@ public class HmacMD5Tests : Rfc2202HmacTests<HmacMD5Tests.Traits>
         public sealed class Traits : IHmacTrait
         {
             public static bool IsSupported => true;
-            public static int HashSizeInBytes => HMACSHA1.HashSizeInBytes;
+            public static int HashSizeInBytes => HMACMD5.HashSizeInBytes;
```
vcsjones, line 18: This turned out to be a mistake in our HMACMD5 tests but it went unnoticed because the value is only used for "create a buffer that is at _least_ large enough". LEN(SHA1) > LEN(MD5) so it happened to work.  But let's fix it.

src/libraries/System.Security.Cryptography/tests/HmacTests.cs
vcsjones, line 1: The `HmacTests` are in desperate need of some refactoring and love in my opinion. I will look at that hopefully sometime for .NET 9. I would like to re-model the tests similarly to how we did SHAKE.

For now though, I decided to keep the changes small so as to not make too many test changes at the same time as product changes.

src/libraries/System.Security.Cryptography/tests/HashAlgorithmTestDriver.cs
vcsjones, line 1: Though the implementation is on `CryptographicOperations`, the tests are all part of the hashing / HMAC testing infrastructure. Otherwise we would be duplicating a lot of additional test logic, KATs, etc.

vcsjones:
These native AOT outerloop checks looks like they ran errantly according to https://github.com/dotnet/runtime/issues/92432, so though they are failing, I marked this ready for review.

bartonjs (approved):

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/CryptographicOperations.cs
```diff
@@ -58,5 +64,766 @@ public static void ZeroMemory(Span<byte> buffer)
...
+        ///   <paramref name="hashAlgorithm"/> specifies a hash algorithm not supported by the current platform.
+        /// </exception>
+        /// <exception cref="CryptographicException">
+        ///   <paramref name="hashAlgorithm"/> specifies an unknown hash algorithm.
+        /// </exception>
+        public static int HashData(HashAlgorithmName hashAlgorithm, ReadOnlySpan<byte> source, Span<byte> destination)
+        {
+            if (!TryHashData(hashAlgorithm, source, destination, out int bytesWritten))
+            {
+                throw new ArgumentException(SR.Argument_DestinationTooShort, nameof(destination));
```
stephentoub, line 153: This shows up at least 6 times. Might be worth creating a private static throw helper method for it, both to reduce the duplication and increase the liklihood this will be inlined.

vcsjones: There are some subtle differences between these. Some are on ROS, some are on ROM. That would mean using `.Span` on the memory, which can be a little expensive. Or we have two helpers, which I am not sure helps at that point.

Unless you meant for the helper to be

```C#
private static void CheckDestinationSize(int requiredSize, int destinationSize)
{
    if (destinationSize < requiredSize)
    {
        throw new ArgumentException(SR.Argument_DestinationTooShort, "destination");
    }
}

// Used like

CheckDestinationSize(hashLength, destination.Length);
```

stephentoub: I actually just meant:
```C#
private static void ThrowDestinationTooShort() => throw new ArgumentException(SR.Argument_DestinationTooShort, "destination");
```
but we could do what you suggest, too.

stephentoub:
Thanks.

PR #92435, opened 2023-09-21
Improve CborWriter buffer growth

### LINKED ISSUE #92425: CborWriter should use doubling for buffer growth
Currently, `CborWriter` increases its buffer 1024 bytes at a time:

https://github.com/dotnet/runtime/blob/51498964d35068ded252ec98baf4fe6ebf6a4612/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs#L210-L212

For large CBOR documents, this can be result in a lot of array resizing. We should instead change the strategy to do doubling.

MichalPetryka:
Fixes #92425.


src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
```diff
@@ -205,11 +205,19 @@ private void EnsureWriteCapacity(int pendingCount)
...
+            int currentCapacity = _buffer is null ? 0 : _buffer.Length;
+            int requiredCapacity = _offset + pendingCount;
+            if (currentCapacity < requiredCapacity)
             {
-                const int BlockSize = 1024;
-                int blocks = checked(_offset + pendingCount + (BlockSize - 1)) / BlockSize;
-                Array.Resize(ref _buffer, BlockSize * blocks);
+                int newCapacity = currentCapacity == 0 ? 1024 : currentCapacity * 2;
+                if ((uint)newCapacity > (uint)Array.MaxLength)
+                    newCapacity = Array.MaxLength;
```
vcsjones, line 214: This package builds for netstandard2.0. I don't believe `MaxLength` is available there.

hamarb123, line 214: would it be possible for it to get here when trying to expand to larger than `Array.MaxLength`, shouldn't there be some condition to throw here? e.g., you could just check if `requiredCapacity > Array.MaxLength` and throw if it is somewhere.

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
```diff
@@ -205,11 +205,19 @@ private void EnsureWriteCapacity(int pendingCount)
...
+                if (newCapacity < requiredCapacity)
+                    newCapacity = requiredCapacity;
+
+                byte[] newBuffer = new byte[newCapacity];
```
hamarb123, line 218: would it be worth using `GC.AllocateUninitializedArray` here?

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
MichalPetryka: AFAIR `new byte[]` will throw on such length.

hamarb123: ahh, I missed the next line lol - please close since I can't for some reason

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
MichalPetryka: I'll let the maintainers decide whether it makes sense here.

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
MichalPetryka: Ah right, should I replace it with some hardcoded constant like 1GB then?

hamarb123: I would replace it with the `Array.MaxValue` constant where it's unavailable; I think that's the correct value for basically all the runtimes we've had anyway isn't it?

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
```diff
@@ -205,11 +205,17 @@ private void EnsureWriteCapacity(int pendingCount)
...
+            int currentCapacity = _buffer is null ? 0 : _buffer.Length;
+            int requiredCapacity = _offset + pendingCount;
+            if (currentCapacity < requiredCapacity)
             {
-                const int BlockSize = 1024;
-                int blocks = checked(_offset + pendingCount + (BlockSize - 1)) / BlockSize;
-                Array.Resize(ref _buffer, BlockSize * blocks);
+                int newCapacity = currentCapacity == 0 ? 1024 : currentCapacity * 2;
+                if ((uint)newCapacity > (uint)Array.MaxLength || newCapacity < requiredCapacity)
+                    newCapacity = requiredCapacity;
```
bartonjs, line 214: Braces are always appreciated :)

```suggestion
                {
                    newCapacity = requiredCapacity;
                }
```

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
bartonjs: AllocateUninitializedArray would probably be OK, except it's not available for .NET Standard 2.0.  I don't think it's important enough here that it's worth a `#if`, so just always using `new byte[newCapacity]` seems fine.

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
vcsjones: For both .NET and .NET Framework, the array length limit when the array is a byte[] is 0x7FFFFFC7. It's probably reasonable to do 

```C#
const int MaxArrayLength = 0x7FFFFFC7; // Array.MaxLength
```

Or if we don't feel like burning it in as a const for the latest runtime:


```C#
#if NET
    int maxArrayLength = Array.Length;
#else
    int MaxArrayLength = 0x7FFFFFC7; // Array.MaxLength
#endif
```

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
```diff
@@ -205,11 +205,20 @@ private void EnsureWriteCapacity(int pendingCount)
...
-            if (_buffer is null || _buffer.Length - _offset < pendingCount)
+            int currentCapacity = _buffer is null ? 0 : _buffer.Length;
+            int requiredCapacity = _offset + pendingCount;
+            if (currentCapacity < requiredCapacity)
             {
-                const int BlockSize = 1024;
-                int blocks = checked(_offset + pendingCount + (BlockSize - 1)) / BlockSize;
-                Array.Resize(ref _buffer, BlockSize * blocks);
+                int newCapacity = currentCapacity == 0 ? 1024 : currentCapacity * 2;
+                const uint MaxArrayLength = 0x7FFFFFC7; // Array.MaxLength
```
eiriktsarpalis, line 213: Maybe add an assertion validating that we're picking the right constant.
```suggestion
                const uint MaxArrayLength = 0x7FFFFFC7; // Array.MaxLength
#if NETCOREAPP
                Debug.Assert(MaxArrayLength == Array.MaxLength);
#endif
```

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
```diff
@@ -205,11 +205,20 @@ private void EnsureWriteCapacity(int pendingCount)
...
+                if ((uint)newCapacity > MaxArrayLength || newCapacity < requiredCapacity)
```
eiriktsarpalis, line 214: This check seems to be redundant, either the doubling will exceed `MaxArrayLength` or it will overflow. The current initial capacity + doubling algorithm seems to suggest that the latter is going to happen.

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
eiriktsarpalis: Why not just use Array.Resize as before?

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
```diff
@@ -205,11 +205,20 @@ private void EnsureWriteCapacity(int pendingCount)
                 throw new OverflowException();
```
eiriktsarpalis, line 205: Apropos, I wrote a lot of this code and it hindsight it seems like I didn't account for not inlining throws in hot-path methods. We should consider refactoring into throw helpers at some point, though clearly not in scope for this PR.

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
MichalPetryka: > Presumably the original buffer is close to full so we're not saving a lot by only copying the written slice.

The intent was to avoid copying empty space here, since copying is really the biggest cost on resizing here.

eiriktsarpalis (approved):
Thanks

vcsjones:
@MichalPetryka Can you please fix the merge conflict?

MichalPetryka:
> Can you please fix the merge conflict?

Done

PR #92436, opened 2023-09-21
[wasm] Fix chrome update script to use 'chromium dash', instead of 'omahaproxy'

radical:
`https://chromiumdash.appspot.com` is the recommended way to get the chrome release versions now, instead of `http://omahaproxy.appspot.com`. This PR updates the task to do that.


maraf (approved):
LGTM

PR #92440, opened 2023-09-21
Improve getExactClasses to support classes as base types

### LINKED ISSUE #88547: NativeAOT: improve getExactClasses to support classes as base types
Currently, `getExactClasses` [gives up](https://github.com/dotnet/runtime/blob/b6266a6e4ceac8708304eed764e25777d08cf184/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs#L2213-L2217) if the base type is not an interface so if we have something like this it won't be optimized:

```cs

public class ClassA
{
    public virtual int GetValue() => 42;
}

public class ClassB : ClassA
{
    // we don't even need to override GetValue here
}

class MyClass
{
    static void Main()
    {
        Test(new ClassB());
    }

    static int Test(ClassA c) => c.GetValue();
}
```
Current codegen of `Test` on NativeAOT:
```asm
; Method MyClass:Test(ClassA):int (FullOpts)
       sub      rsp, 40
       mov      rax, qword ptr [rcx]
       call     [rax+30H]ClassA:GetValue():int:this
       nop      
       add      rsp, 40
       ret      
; Total bytes of code: 16
```
Expected codegen:
```asm
; Method MyClass:Test(ClassA):int (FullOpts)
       cmp      byte  ptr [rcx], cl ; nullcheck
       mov      eax, 42
       ret      
; Total bytes of code: 8
```

Presumably, this needs no changes on the jit side, only that `getExactClasses` should be improved on the [NAOT side](https://github.com/dotnet/runtime/blob/b6266a6e4ceac8708304eed764e25777d08cf184/src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs#L2213-L2217).

cc @dotnet/ilc-contrib

MichalStrehovsky:
Fixes #88547

```csharp

public class ClassA
{
    public virtual int GetValue() => 42;
}

public class ClassB : ClassA
{
    // we don't even need to override GetValue here
}

class MyClass
{
    static void Main()
    {
        Test(new ClassB());
    }

    static int Test(ClassA c) => c.GetValue();
}
```

Old codegen:

```asm
; Method MyClass:Test(ClassA):int (FullOpts)
       sub      rsp, 40
       mov      rax, qword ptr [rcx]
       call     [rax+30H]ClassA:GetValue():int:this
       nop
       add      rsp, 40
       ret
; Total bytes of code: 16
```

New codegen:

```asm
00007FF66DD8AFB0  cmp         dword ptr [rcx],ecx
00007FF66DD8AFB2  mov         eax,2Ah
00007FF66DD8AFB7  ret
```

Cc @dotnet/ilc-contrib @EgorBo

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILScanner.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILScanner.cs
@@ -411,8 +411,8 @@ namespace ILCompiler
             private HashSet<TypeDesc> _constructedTypes = new HashSet<TypeDesc>();
             private HashSet<TypeDesc> _canonConstructedTypes = new HashSet<TypeDesc>();
             private HashSet<TypeDesc> _unsealedTypes = new HashSet<TypeDesc>();
-            private Dictionary<TypeDesc, HashSet<TypeDesc>> _interfaceImplementators = new();
-            private HashSet<TypeDesc> _disqualifiedInterfaces = new();
+            private Dictionary<TypeDesc, HashSet<TypeDesc>> _implementators = new();
+            private HashSet<TypeDesc> _disqualifiedTypes = new();
 
             public ScannedDevirtualizationManager(NodeFactory factory, ImmutableArray<DependencyNodeCore<NodeFactory>> markedNodes)
             {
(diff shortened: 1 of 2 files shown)


EgorBo:
Love it! It should also pick up other jit opts like `if (obj is Class1)` or multiple-guesses GDV, etc which currently only kick in for interfaces. LGTM if CI passes 🙂

jkotas (approved):
Thanks

PR #89651, opened 2023-07-28
Build native windows-arm64 Mono cross-compilers

directhex:
This is a stub, but ought to build at least.

PR #92443, opened 2023-09-21
Use the native cross compilers on win-arm64

lewing:


steveisok:
This is pending win-arm64 cross compiler support from @directhex.

directhex:
It's done, but we have no means to CI effectively. That's the problem Sent from my iPhoneOn Sep 21, 2023, at 7:24 PM, Steve Pfister ***@***.***> wrote:﻿
This is pending win-arm64 cross compiler support from @directhex.

lewing:
are the packages being built and published? should we turn this on in head and let people try out the builds?  we could stand up a wbt win-arm64 lane for smoke tests if we have hosts

steveisok:
We will need to do an official build to validate. I do think there's value in having this in main.

lewing:
kicked off an official build, we'll see how it goes

lewing:
@steveisok looks green on the official build

radical (approved):

steveisok:
> @steveisok looks green on the official build, should we merge and try it out?

YOLO!

akoeplinger:
FYI this was already done in https://github.com/dotnet/runtime/pull/89651. The main commit shows 0 changes 😄

PR #92449, opened 2023-09-22
Simplify analyzer handling of properties/events

### LINKED ISSUE #83235: ILLink analyzer handling of IL3003 on properties and events
This is currently only about `RequiresAssemblyFiles` as it's allowed on properties and events. `RequiresUnreferencedCode` and `RequiresDynamicCode` are not allowed on properties and events (yet - https://github.com/dotnet/runtime/issues/83194).

Take this as an example:

```C#
class Base
{
    [RequiresAssemblyFiles("property")]
    public int MyProperty
    {
        get;
        [RequiresAssemblyFiles("setter")]
        set;
    }
}

class Derived
{
    [RequiresAssemblyFiles("property")]
    public int MyProperty
    {
        get;
        set; // IL3003 - the Derived.MyProperty.set doesn't have the attribute while Base.MyProperty.set does
    }
}
```

Basically the analyzer treats the property and the methods as separate entities and validates the mismatch of the attribute on each separately. But the semantics of the attribute are that if it's on the property it's as if it applies to both accessors.
So from a single-file compatibility, the above example is perfectly compatible, as on 4 accessors will produce IL3002 when called as all 4 fall under the property-level attribute.

There are other similar variations of this, where setting on property on Base and then setting it on both accessors in derived still causes warnings (and vice versa).

NativeAOT will implement this by looking at methods only, so the above will not produce a warning. As in the end the behavior on method level is what actually causes single-file incompatibility.

We should change the analyzer to this behavior as well - it should basically warn less. There's a small chance of breaks if somebody suppresses the warning using pragmas (which nobody should do on this warning really) and only scope it to the property declaration without including the accessors. I would be surprised if that actually happens even once.

sbomer:
Previously, the analyzer was effectively visiting referenced properties twice for each reference - it would visit both the referenced get/set method, and the property itself.

This change only visits the get/set method, and handles the property warnings there by treating the getter/setter as annotated when the property is annotated. This also allows some simplification of the override checking logic. Events get the same treatment.

Fixes https://github.com/dotnet/runtime/issues/83235

### DIFF
--- a/src/tools/illink/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
+++ b/src/tools/illink/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
@@ -54,14 +54,6 @@ namespace ILLink.RoslynAnalyzer
 					CheckMatchingAttributesInInterfaces (symbolAnalysisContext, typeSymbol);
 				}, SymbolKind.NamedType);
 
-
-				context.RegisterSymbolAction (symbolAnalysisContext => {
-					var propertySymbol = (IPropertySymbol) symbolAnalysisContext.Symbol;
-					if (AnalyzerDiagnosticTargets.HasFlag (DiagnosticTargets.Property)) {
-						CheckMatchingAttributesInOverrides (symbolAnalysisContext, propertySymbol);
-					}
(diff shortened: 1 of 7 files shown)


vitek-karas (approved):
Thanks!

PR #92452, opened 2023-09-22
Use snprintf in pal_icushim.c

jkoritzinsky:
When AddressSanitizer is enabled on macOS, the toolchain deprecates `sprintf`. Switch to `snprintf` for better bounds checking and to help fix the macOS ASAN build.

### DIFF
--- a/src/native/libs/System.Globalization.Native/pal_icushim.c
+++ b/src/native/libs/System.Globalization.Native/pal_icushim.c
@@ -33,6 +33,14 @@ FOR_ALL_ICU_FUNCTIONS
 
 #define MaxICUVersionStringLength 33
 
+#else
+
+#define VERSION_PREFIX_NONE ""
+#define VERSION_PREFIX_SUSE "suse"
+
+// .[suse]x.x.x, considering the max number of decimal digits for each component
+#define MaxICUVersionStringLength (sizeof(VERSION_PREFIX_SUSE) + 33)
+
 #endif
 
 static void* libicuuc = NULL;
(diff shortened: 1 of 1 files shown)


tarekgh:
@jkoritzinsky could you please fix the error `src/native/libs/System.Globalization.Native/pal_icushim.c(66,37): error GFC248B73: (NETCORE_ENGINEERING_TELEMETRY=Build) use of undeclared identifier 'MaxICUVersionStringLength'`?

tarekgh:
@jkoritzinsky does the AddressSanitizer is going to be enabled on .NET8.0 on macOS too? I am asking because we may need to port this to 8.0 release. 

CC @jkotas

jkoritzinsky:
No, I am not going to enable AddressSanitizer on the 8.0 branch.

src/native/libs/System.Globalization.Native/pal_icushim.c
```diff
@@ -55,24 +63,24 @@ static int FindSymbolVersion(int majorVer, int minorVer, int subVer, char* symbo
     if (dlsym(libicuuc, "u_strlen") == NULL)
     {
         // Now try just the _majorVer added
-        sprintf(symbolVersion, "_%d%s", majorVer, suffix);
-        sprintf(symbolName, "u_strlen%s", symbolVersion);
+        snprintf(symbolVersion, MaxICUVersionStringWithSuffixLength, "_%d%s", majorVer, suffix);
```
tarekgh, line 66: >MaxICUVersionStringWithSuffixLength [](http://example.com/codeflow?start=32&length=35)

`FindSymbolVersion` can be called with symbol version buffer of length `MaxICUVersionStringLength` which is smaller than `MaxICUVersionStringWithSuffixLength`. I know the old code was not protecting against buffer overflow so your changes is similar to what we had. But if we want to write it in correct reliable way, we should pass the symbol version length to here. I know this will touch some method declarations but will be safer I guess.

tarekgh (approved):
Thanks @jkoritzinsky!

PR #92462, opened 2023-09-22
Optimize capturing lambdas in `DependencyContextExtensions` and `DependencyContextJsonReader`

### LINKED ISSUE #87911: Optimize allocations of a capturing lambda in `GetRuntimeNativeAssets`
image

image

---

Also

image

This seems related to this lambda:

https://github.com/dotnet/runtime/blob/ce953d023f20964303cf761bf6e2e72d70bdf3d9/src/libraries/Microsoft.Extensions.DependencyModel/src/DependencyContextExtensions.cs#L35

karakasa:
close #87911 

I am kind of uncertain where to put the helper methods, as I don't want to pollute other namespaces but the methods' function are very generic and might be useful to others.


src/libraries/System.Linq/src/System/Linq/SelectManyWithParameterHelpers.cs
```diff
@@ -0,0 +1,78 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+
+namespace System.Linq
+{
+    internal static class SelectManyWithParameterHelpers
```
ericstj, line 9: This file should not be placed under `System.Linq` since it's not part of that assembly.

ericstj (commented):
It looks to me like the issue this issue attempts to address has an open question about the scenario being measured that notices/cares about the performance here.   @Youssef1313 can you chime in on that?  @ViktorHofer -- looks like you marked the original issue `help-wanted` can you have a look at this?

Youssef1313:
@ericstj I don't have much info other than seeing these allocations at the startup of an Uno Platform Skia application. From the trace screenshots posted in the issue, it seems like all Uno does is this:

https://github.com/unoplatform/uno/blob/8b88124a5ebdc61a782326d96ff32e5cba5d452a/src/Uno.UI.Runtime.Skia.Gtk/Rendering/OpenGLESRenderSurface.cs#L29-L32

Then Silk.NET calls into this code and then we see these allocations.

src/libraries/System.Linq/src/System/Linq/SelectManyWithParameterHelpers.cs
```diff
@@ -0,0 +1,78 @@
...
+            Func<TSource, TParam1, TParam2, IEnumerable<TResult>> func, TParam1 param1, TParam2 param2)
+        {
+            foreach (var elem in src)
+            {
+                foreach (var subElem in func(elem, param1, param2))
+                {
+                    yield return subElem;
+                }
+            }
+        }
```
ViktorHofer, line 76: Why do we need to differentiate between array and non array?

ViktorHofer:
I'm also surprised that this code is executed on the hot path. Nevertheless, the optimization looks trivial (replacing the lamda with a loop) so I'm OK with that. The extension helper file should live where it is used (in DependencyModel).

ericstj:
Here's the place in Silk.NET that's using this: https://github.com/dotnet/Silk.NET/blob/d4d55347bb5d4aa9f316b516abf05ed4d8a66bfe/src/Core/Silk.NET.Core/Loader/DefaultPathResolver.cs#L212

Probably better to use https://learn.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblydependencyresolver.resolveunmanageddlltopath?view=net-7.0

src/libraries/System.Linq/src/System/Linq/SelectManyWithParameterHelpers.cs
karakasa: The data source is Array as of now so I want to provide an optimized path for it. Meanwhile it would still work if the source becomes non-array in the future.

ViktorHofer: Where is the optimized path? The method bodies are identical.

karakasa: iirc `foreach` on arrays is compiled into `for(int i=0;i<...;i++)` while `foreach` on `(IEnumerable<T>)T[]` utilizes `SZGenericArrayEnumerator<T>` eventually.

ViktorHofer: IMO such a perf optimization isn't worth the cost of having two code paths.

src/libraries/Microsoft.Extensions.DependencyModel/Microsoft.Extensions.DependencyModel.sln
```diff
@@ -1,4 +1,8 @@
-﻿Microsoft Visual Studio Solution File, Format Version 12.00
+﻿
```
ViktorHofer: Please undo these file changes. We don't check sln changes in.

src/libraries/Microsoft.Extensions.DependencyModel/src/DependencyContextJsonReader.cs
```diff
@@ -741,9 +741,18 @@ private IEnumerable<Library> CreateLibraries(IEnumerable<TargetLibrary>? librari
...
+        }
+
+        private IEnumerable<Library> CreateLibrariesNotNull(IEnumerable<TargetLibrary> libraries, bool runtime, Dictionary<string, LibraryStub>? libraryStubs)
+        {
+            foreach (TargetLibrary library in libraries)
+            {
+                Library? createdLibrary = CreateLibrary(library, runtime, libraryStubs);
+                if (createdLibrary is not null)
+                    yield return createdLibrary;
+            }
```
ViktorHofer, line 755: Why not inline the foreach loop into L745 as the method is only used once?

karakasa: I would propose to make `CreateLibrariesNotNull` a local function so that no object is `new`ed if `libraries` is `null`, similar to the original logic.

ViktorHofer (approved):

PR #92465, opened 2023-09-22
Set stack size on MUSL with PT_GNU_STACK header

### LINKED ISSUE #72920: Set minimum stack size using PT_GNU_STACK header for Musl
See http://git.musl-libc.org/cgit/musl/commit/?id=7b3348a98c139b4b4238384e52d4b0eb237e4833

This can be done once the minimum target musl version is 1.21. We are on 1.20 right now (it is what CI uses at least). 

Cleanup with this issue number once implemented.

MichalStrehovsky:
Fixes #72920.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
@@ -23,6 +23,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <LinkerFlavor Condition="'$(LinkerFlavor)' == '' and '$(_targetOS)' == 'freebsd'">lld</LinkerFlavor>
     <LinkerFlavor Condition="'$(LinkerFlavor)' == '' and '$(_linuxLibcFlavor)' == 'bionic'">lld</LinkerFlavor>
     <LinkerFlavor Condition="'$(LinkerFlavor)' == '' and '$(_targetOS)' == 'linux'">bfd</LinkerFlavor>
+    <IlcDefaultStackSize Condition="'$(_linuxLibcFlavor)' == 'musl'">1572864</IlcDefaultStackSize>
   </PropertyGroup>
 
   <Target Name="SetupOSSpecificProps" DependsOnTargets="$(IlcDynamicBuildPropertyDependencies)">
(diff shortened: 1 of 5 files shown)


filipnavara (approved):

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
```diff
@@ -156,6 +156,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <LinkerArg Include="-Wl,-z,relro" Condition="'$(_IsApplePlatform)' != 'true'" />
       <!-- binskim warning BA3011 The BIND_NOW flag is missing -->
       <LinkerArg Include="-Wl,-z,now" Condition="'$(_IsApplePlatform)' != 'true'" />
+      <LinkerArg Include="-Wl,-z,stack-size=1572864" Condition="'$(_linuxLibcFlavor)' == 'musl'" />
```
jkotas, line 159:
```suggestion
      <LinkerArg Include="-Wl,-z,stack-size=$(IlcDefaultStackSize)" Condition="'$(IlcDefaultStackSize)' != ''" />
```
And set the default IlcDefaultStackSize for musl earlier in this file?

It would provide equivalent of `DOTNET_DefaultStackSize` config in case somebody needs it.

jkotas (approved):

src/native/libs/System.Native/CMakeLists.txt
```diff
@@ -4,12 +4,6 @@ if (NOT CLR_CMAKE_TARGET_MACCATALYST AND NOT CLR_CMAKE_TARGET_IOS AND NOT CLR_CM
     add_definitions(-DHAS_CONSOLE_SIGNALS)
 endif ()
 
-if (CLR_CMAKE_TARGET_LINUX_MUSL)
-    # Fix up the thread stack size for MUSL to more reasonable size.
-    # TODO: https://github.com/dotnet/runtimelab/issues/791
-    add_definitions(-DENSURE_PRIMARY_STACK_SIZE)
-endif ()
```
akoeplinger, line 11: we have the same in mono, I'm not sure I understand how this is fixed for everything but nativeaot?

jkotas: The issue was specific to native AOT.

We have the same `ENSURE_PRIMARY_STACK_SIZE` in CoreCLR too, however it is not straightforward to get rid of it there since CoreCLR allows configuring the stack size using env variable.

jkotas: Deleting the TODO in #92597

PR #92469, opened 2023-09-22
JIT: Support SIMD sizes for remainders in store decomposition

jakobbotsch:


jakobbotsch:

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=415621&view=ms.vss-build-web.run-extensions-tab).

src/coreclr/jit/promotiondecomposition.cpp
```diff
@@ -336,7 +337,29 @@ class DecompositionPlan
                         break;
 #endif
 
-                        // TODO-CQ: SIMD sizes
+#ifdef FEATURE_SIMD
+                    case 16:
```
EgorBo, line 341: Do we need SIMD12 and SIMD8 here?

EgorBo (approved):

PR #92470, opened 2023-09-22
Extend preinitialization interpreter

### LINKED ISSUE #78681: Allow modeling ref fields in the class constructor interpreter
To do https://github.com/dotnet/runtime/pull/78093#discussion_r1028400165 we need to be able to model Spans of things. We represents structs as byte arrays. Hard to do byrefs right now.

MichalStrehovsky:
Things that I added:

* Support for `typeof(T) == typeof(Bar)` (this will be useful later, we'll eventually be able to also freeze these).
* Support static interface method calls
* Constrained method calls on valuetypes
* More `ReadOnlySpan` construction patterns, `.Length`
* More indirect load/store support

Contributes to #78681. To full resolve this, we need to fix up things so we can answer `Sse2.IsSupported`.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
@@ -452,7 +452,7 @@ namespace ILCompiler
                             }
 
                             Value retVal;
-                            if ((!method.IsIntrinsic && !method.OwningType.IsIntrinsic) || !TryHandleIntrinsicCall(method, methodParams, out retVal))
+                            if (!method.IsIntrinsic || !TryHandleIntrinsicCall(method, methodParams, out retVal))
                             {
                                 recursionProtect ??= new Stack<MethodDesc>();
                                 recursionProtect.Push(methodIL.OwningMethod);
@@ -1699,13 +1699,6 @@ namespace ILCompiler
                         && (parameters[0] is RuntimeTypeValue || parameters[1] is RuntimeTypeValue):
                     {
                         retVal = ValueTypeValue.FromSByte(parameters[0] == parameters[1] ? (sbyte)1 : (sbyte)0);
-                        return true;
-                    }
-                case "get_IsSupported" when method.OwningType is MetadataType sse2Type
-                        && sse2Type.Name == "Sse2":
-                    {
-                        // TODO: hacky
-                        retVal = ValueTypeValue.FromSByte(sse2Type.Context.Target.Architecture == TargetArchitecture.X64 ? (sbyte)1 : (sbyte)0);
                         return true;
                     }
             }
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
@@ -779,7 +779,7 @@ namespace ILCompiler
                             }
                             else if (popped.ValueKind == StackValueKind.ByRef
                                 && (opcode == ILOpcode.conv_i || opcode == ILOpcode.conv_u)
-                                && reader.PeekILOpcode() is >= ILOpcode.ldind_i1 and <= ILOpcode.ldind_i)
+                                && (reader.PeekILOpcode() is >= ILOpcode.ldind_i1 and <= ILOpcode.ldind_ref) || reader.PeekILOpcode() == ILOpcode.ldobj)
                             {
                                 // In the interpreter memory model, there's no conversion from a byref to an integer.
                                 // Roslyn however sometimes emits a sequence of conv_u followed by ldind and we can
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
@@ -779,11 +779,15 @@ namespace ILCompiler
                             }
                             else if (popped.ValueKind == StackValueKind.ByRef
                                 && (opcode == ILOpcode.conv_i || opcode == ILOpcode.conv_u)
-                                && (reader.PeekILOpcode() is >= ILOpcode.ldind_i1 and <= ILOpcode.ldind_ref) || reader.PeekILOpcode() == ILOpcode.ldobj)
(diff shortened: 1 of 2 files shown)


stephentoub:
cc: @MihaZupan

MichalStrehovsky:
Because of 16f724f this no longer preinitializes `SearchValues.Create("Lol");`. I removed that because I want to add a more systematic solution to HW intrinsics IsSupported calls.

Longer SearchValues are going to also use Sse3.IsSupported and that one we can't hardcode. It needs to come out of how the compilation is configured.

What's here is ready for review. I don't want to mix that up with this PR.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
```diff
@@ -1425,6 +1483,27 @@ private Status TryScanMethod(MethodIL methodIL, Value[] parameters, Stack<Method
...
+                                TypeDesc type = methodIL.GetObject(reader.ReadILToken()) as TypeDesc;
+                                opcode = type.Category switch
+                                {
+                                    TypeFlags.SByte => ILOpcode.ldind_i1,
+                                    TypeFlags.Boolean or TypeFlags.Byte => ILOpcode.ldind_u1,
+                                    TypeFlags.Int16 => ILOpcode.ldind_i2,
+                                    TypeFlags.Char or TypeFlags.UInt16 => ILOpcode.ldind_u2,
+                                    TypeFlags.Int32 => ILOpcode.ldind_i4,
+                                    TypeFlags.UInt32 => ILOpcode.ldind_u4,
+                                    TypeFlags.Int64 or TypeFlags.UInt64 => ILOpcode.ldind_i8,
```
jkotas, line 1497: Add Float and Double? They are handled in the switch statement below.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
```diff
@@ -743,6 +777,16 @@ private Status TryScanMethod(MethodIL methodIL, Value[] parameters, Stack<Method
                                         return Status.Fail(methodIL.OwningMethod, opcode);
                                 }
                             }
+                            else if (popped.ValueKind == StackValueKind.ByRef
+                                && (opcode == ILOpcode.conv_i || opcode == ILOpcode.conv_u)
+                                && (reader.PeekILOpcode() is >= ILOpcode.ldind_i1 and <= ILOpcode.ldind_ref) || reader.PeekILOpcode() == ILOpcode.ldobj)
+                            {
+                                // In the interpreter memory model, there's no conversion from a byref to an integer.
+                                // Roslyn however sometimes emits a sequence of conv_u followed by ldind and we can
```
jkotas, line 785: What does Roslyn emit it for?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
```diff
@@ -743,6 +777,16 @@ private Status TryScanMethod(MethodIL methodIL, Value[] parameters, Stack<Method
                                         return Status.Fail(methodIL.OwningMethod, opcode);
                                 }
                             }
+                            else if (popped.ValueKind == StackValueKind.ByRef
+                                && (opcode == ILOpcode.conv_i || opcode == ILOpcode.conv_u)
+                                && (reader.PeekILOpcode() is >= ILOpcode.ldind_i1 and <= ILOpcode.ldind_ref) || reader.PeekILOpcode() == ILOpcode.ldobj)
```
jkotas, line 782:
```suggestion
                                && (reader.PeekILOpcode() is (>= ILOpcode.ldind_i1 and <= ILOpcode.ldind_ref) or ILOpcode.ldobj)
```
Pattern matching all the way.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
```diff
@@ -1464,10 +1543,56 @@ private Status TryScanMethod(MethodIL methodIL, Value[] parameters, Stack<Method
...
+
+                            StackEntry location = stack.Pop();
+                            if (location.ValueKind != StackValueKind.ByRef)
+                                ThrowHelper.ThrowInvalidProgramException();
+
+                            byte[] dest = ((ByRefValue)location.Value).PointedToBytes;
+                            int destOffset = ((ByRefValue)location.Value).PointedToOffset;
+                            byte[] src = ((ValueTypeValue)val).InstanceBytes;
+                            if (destOffset + src.Length > dest.Length)
+                                ThrowHelper.ThrowInvalidProgramException();
```
jkotas, line 1587: Should this be StatusFail? This is valid program, that just throws exception.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
MichalStrehovsky: Single/double added.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
MichalStrehovsky: Added as a comment inline.

MichalStrehovsky:
> Because of [16f724f](https://github.com/dotnet/runtime/commit/16f724fef434fb55ebcb604959c8a10c3696a5f6) this no longer preinitializes `SearchValues.Create("Lol");`. I removed that because I want to add a more systematic solution to HW intrinsics IsSupported calls.
> 
> Longer SearchValues are going to also use Sse3.IsSupported and that one we can't hardcode. It needs to come out of how the compilation is configured.
> 
> What's here is ready for review. I don't want to mix that up with this PR.

I dug into it some more. We'll be able to preinit short SearchValues strings.

Longer SearchValues strings are a lost cause. I have a couple extra changes (that I'll upstream separately) where I tried to make it work. It preinitialized a bunch of more stuff.

Eventually it runs into the issue that this init only code (that could be slow and it would be fine) is calling into our super optimized hardware intrinsic stuff (like `NonPackedContainsValueType` where I gave up). We'd need to extend the interpreter to interpret hardware intrinsics. We cannot lie about `IsSupported`/`IsHardwareAccelerated`, but then it leads to codepaths that actually use these. I don't see us implementing the septillions of various hardware intrinsics that we use all over corelib and keep adding on every release.

jkotas (approved):
Thank you!

Issue #89425, opened 2023-07-25
[browser] mono runtime already exited with 0


pavelsavara:
### Error Blob

Build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=350621&view=results
Log: https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-89387-merge-7052a2aebdb04a27be/WasmTestOnNodeJS-System.Net.WebSockets.Client.Tests/1/console.ec43186b.log?helixlogtype=result

```json
{
  "ErrorMessage": "mono runtime already exited with",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

### Reproduction Steps

wasm `System.Net.WebSockets.Client.Tests` nodeJS

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=350621
**Error message validated:** `mono runtime already exited with`
**Result validation: :white_check_mark:** Known issue matched with the provided build.
**Validation performed at:** 7/25/2023 9:31:59 AM UTC

### Report

|Build|Definition|Test|Pull Request|
|---|---|---|---|
|[457734](https://dev.azure.com/dnceng-public/public/_build/results?buildId=457734)|dotnet/runtime|[WasmTestOnNodeJS-System.Net.Http.Functional.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=457734&view=ms.vss-test-web.build-test-results-tab&runId=10309828&resultId=187499)|dotnet/runtime#94204|
|[457289](https://dev.azure.com/dnceng-public/public/_build/results?buildId=457289)|dotnet/runtime|[WasmTestOnNodeJS-System.Net.Http.Functional.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=457289&view=ms.vss-test-web.build-test-results-tab&runId=10297578&resultId=187499)|dotnet/runtime#94204|
|[456278](https://dev.azure.com/dnceng-public/public/_build/results?buildId=456278)|dotnet/runtime|[WasmTestOnNodeJS-System.Net.Http.Functional.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=456278&view=ms.vss-test-web.build-test-results-tab&runId=10269268&resultId=187506)|dotnet/runtime#94204|
#### Summary
[...]

PR #92479, opened 2023-09-22
[browser] mono runtime already exited with 0 - silence WS

pavelsavara:
Attempt on https://github.com/dotnet/runtime/issues/89425

### DIFF
--- a/src/mono/wasm/runtime/web-socket.ts
+++ b/src/mono/wasm/runtime/web-socket.ts
@@ -5,7 +5,7 @@ import MonoWasmThreads from "consts:monoWasmThreads";
 
 import { prevent_timer_throttling } from "./scheduling";
 import { Queue } from "./queue";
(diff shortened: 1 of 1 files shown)


lewing (approved):

radical (approved):

src/mono/wasm/runtime/web-socket.ts
```diff
@@ -58,17 +58,20 @@ export function ws_wasm_create(uri: string, sub_protocols: string[] | null, rece
     ws.binaryType = "arraybuffer";
     const local_on_open = () => {
         if (ws[wasm_ws_is_aborted]) return;
+        if (loaderHelpers.is_exited()) return;
```
radical, line 61: Should this be before `if (ws[wasm_ws_is_aborted]) return` too?

pavelsavara: doesn't really matter.

PR #86556, opened 2023-05-21
Convert.Try{From/To}HexString

hrrrrustic:
Close #78472

PR #92483, opened 2023-09-22
`Convert.ToHexStringLower`: lower variant for `Convert.ToHexString`

### LINKED ISSUE #60393: [API Proposal]: Add lowercase support for Convert.ToHexString
### Background and motivation

In some scenarios uppercase hex string is preferred, but in the others lowercase is **required**. See https://github.com/dotnet/runtime/pull/39702#issuecomment-828055896.

`Convert.ToHexString` Methods use `HexConverter.ToString` internally, which supports uppercase and lowercase. https://github.com/dotnet/runtime/blob/1ebec9a41191fcce94895408a4c1451da91b3877/src/libraries/System.Private.CoreLib/src/System/Convert.cs#L2930-L2938
But `HexConverter` is an internal class, and the case transforming in `HexConverter.ToString` is faster than calling `String.ToUpper()` again, because `HexConverter` use ssse3 or bit operations.

Add a new method will be simple because the internal class has supported it already.

### API Proposal

```C#
namespace System
{
    public static class Convert
    {
        public static string ToHexString(byte[] inArray, Casing casing = Casing.Upper);
        public static string ToHexString(byte[] inArray, int offset, int length, Casing casing = Casing.Upper);
        public static string ToHexString(ReadOnlySpan<byte> bytes, Casing casing = Casing.Upper);
    }
}
```     

### API Usage

```C#
var byteArray = GetByteArray();
var hexString = Convert.ToHexString(byteArray, Casing.Lower);
Console.WriteLine(hexString);
```     

### Risks

_No response_

determ1ne:
New methods:
```cs
namespace System
{
    public static class Convert
    {
        public static string ToHexStringLower(byte[] inArray);
        public static string ToHexStringLower(byte[] inArray, int offset, int length);
        public static string ToHexStringLower(ReadOnlySpan<byte> bytes);
    }
}
```

Close #60393.

`TryToHexStringLower` may be implemented after #86556 merged.

### DIFF
--- a/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs
+++ b/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs
@@ -1782,7 +1782,11 @@ namespace System.Diagnostics
             if (idData.Length != 16)
                 throw new ArgumentOutOfRangeException(nameof(idData));
 
+#if NET9_0_OR_GREATER
             return new ActivityTraceId(Convert.ToHexStringLower(idData));
+#else
+            return new ActivityTraceId(HexConverter.ToString(idData, HexConverter.Casing.Lower));
+#endif
         }
         public static ActivityTraceId CreateFromUtf8String(ReadOnlySpan<byte> idData) => new ActivityTraceId(idData);
 
@@ -1861,7 +1865,11 @@ namespace System.Diagnostics
                 span[1] = BinaryPrimitives.ReverseEndianness(span[1]);
             }
 
+#if NET9_0_OR_GREATER
             _hexString = Convert.ToHexStringLower(MemoryMarshal.AsBytes(span));
+#else
+            _hexString = HexConverter.ToString(MemoryMarshal.AsBytes(span), HexConverter.Casing.Lower);
+#endif
         }
 
         /// <summary>
@@ -1956,14 +1964,22 @@ namespace System.Diagnostics
         {
             ulong id;
             ActivityTraceId.SetToRandomBytes(new Span<byte>(&id, sizeof(ulong)));
+#if NET9_0_OR_GREATER
             return new ActivitySpanId(Convert.ToHexStringLower(new ReadOnlySpan<byte>(&id, sizeof(ulong))));
+#else
+            return new ActivitySpanId(HexConverter.ToString(new ReadOnlySpan<byte>(&id, sizeof(ulong)), HexConverter.Casing.Lower));
+#endif
         }
         public static ActivitySpanId CreateFromBytes(ReadOnlySpan<byte> idData)
         {
             if (idData.Length != 8)
                 throw new ArgumentOutOfRangeException(nameof(idData));
 
+#if NET9_0_OR_GREATER
             return new ActivitySpanId(Convert.ToHexStringLower(idData));
+#else
+            return new ActivitySpanId(HexConverter.ToString(idData, HexConverter.Casing.Lower));
+#endif
         }
         public static ActivitySpanId CreateFromUtf8String(ReadOnlySpan<byte> idData) => new ActivitySpanId(idData);
 
@@ -2031,7 +2047,11 @@ namespace System.Diagnostics
                 id = BinaryPrimitives.ReverseEndianness(id);
             }
 
+#if NET9_0_OR_GREATER
             _hexString = Convert.ToHexStringLower(new ReadOnlySpan<byte>(&id, sizeof(ulong)));
+#else
+            _hexString = HexConverter.ToString(new ReadOnlySpan<byte>(&id, sizeof(ulong)), HexConverter.Casing.Lower);
+#endif
         }
 
         /// <summary>
--- a/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs
+++ b/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs
@@ -1782,7 +1782,7 @@ namespace System.Diagnostics
             if (idData.Length != 16)
                 throw new ArgumentOutOfRangeException(nameof(idData));
 
-            return new ActivityTraceId(HexConverter.ToString(idData, HexConverter.Casing.Lower));
+            return new ActivityTraceId(Convert.ToHexStringLower(idData));
         }
         public static ActivityTraceId CreateFromUtf8String(ReadOnlySpan<byte> idData) => new ActivityTraceId(idData);
 
@@ -1861,7 +1861,7 @@ namespace System.Diagnostics
                 span[1] = BinaryPrimitives.ReverseEndianness(span[1]);
             }
 
-            _hexString = HexConverter.ToString(MemoryMarshal.AsBytes(span), HexConverter.Casing.Lower);
+            _hexString = Convert.ToHexStringLower(MemoryMarshal.AsBytes(span));
         }
 
         /// <summary>
@@ -1956,14 +1956,14 @@ namespace System.Diagnostics
         {
             ulong id;
             ActivityTraceId.SetToRandomBytes(new Span<byte>(&id, sizeof(ulong)));
-            return new ActivitySpanId(HexConverter.ToString(new ReadOnlySpan<byte>(&id, sizeof(ulong)), HexConverter.Casing.Lower));
+            return new ActivitySpanId(Convert.ToHexStringLower(new ReadOnlySpan<byte>(&id, sizeof(ulong))));
         }
         public static ActivitySpanId CreateFromBytes(ReadOnlySpan<byte> idData)
         {
             if (idData.Length != 8)
                 throw new ArgumentOutOfRangeException(nameof(idData));
 
-            return new ActivitySpanId(HexConverter.ToString(idData, HexConverter.Casing.Lower));
+            return new ActivitySpanId(Convert.ToHexStringLower(idData));
         }
         public static ActivitySpanId CreateFromUtf8String(ReadOnlySpan<byte> idData) => new ActivitySpanId(idData);
 
@@ -2031,7 +2031,7 @@ namespace System.Diagnostics
                 id = BinaryPrimitives.ReverseEndianness(id);
             }
 
-            _hexString = HexConverter.ToString(new ReadOnlySpan<byte>(&id, sizeof(ulong)), HexConverter.Casing.Lower);
+            _hexString = Convert.ToHexStringLower(new ReadOnlySpan<byte>(&id, sizeof(ulong)));
         }
 
         /// <summary>
--- a/src/libraries/System.Private.CoreLib/src/System/Convert.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Convert.cs
@@ -3091,7 +3091,7 @@ namespace System
                 charsWritten = 0;
                 return true;
             }
-            else if (source.Length > int.MaxValue / 2 || destination.Length > source.Length  * 2)
+            else if (source.Length > int.MaxValue / 2 || destination.Length > source.Length * 2)
             {
                 charsWritten = 0;
                 return false;
@@ -3168,7 +3168,7 @@ namespace System
                 charsWritten = 0;
                 return true;
             }
-            else if (source.Length > int.MaxValue / 2 || destination.Length > source.Length  * 2)
+            else if (source.Length > int.MaxValue / 2 || destination.Length > source.Length * 2)
             {
                 charsWritten = 0;
                 return false;
--- a/src/libraries/System.Private.CoreLib/src/System/Convert.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Convert.cs
@@ -3153,5 +3153,30 @@ namespace System
 
             return HexConverter.ToString(bytes, HexConverter.Casing.Lower);
         }
+
+        /// <summary>
+        /// Converts a span of 8-bit unsigned integers to its equivalent span representation that is encoded with lowercase hex characters.
+        /// </summary>
+        /// <param name="source">A span of 8-bit unsigned integers.</param>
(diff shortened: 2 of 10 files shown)


determ1ne:
@dotnet-policy-service agree

stephentoub:
It'd be good to include in this PR updates to places elsewhere in the repo we can switch to using the new public API, e.g.
https://github.com/dotnet/runtime/blob/0841678135c9edd25bc807bb8aaef4b7801890b4/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/AuthenticationHelper.Digest.cs#L233

vcsjones:
The final approved API has additional overloads. See @bartonjs's comment here: https://github.com/dotnet/runtime/issues/60393#issuecomment-1730147073. It would be good to get `TryToHexStringLower` with the rest of them.

buyaa-n:
> It'd be good to include in this PR updates to places elsewhere in the repo we can switch to using the new public API, e.g.
> 
> https://github.com/dotnet/runtime/blob/0841678135c9edd25bc807bb8aaef4b7801890b4/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/AuthenticationHelper.Digest.cs#L233

@determ1ne could you apply the feedback? In addition to above case, I see 5 hits in: https://github.com/dotnet/runtime/blob/52bd95caa3563bb889548e489762388d6842dca4/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L1785 and one hit in: https://github.com/dotnet/runtime/blob/52bd95caa3563bb889548e489762388d6842dca4/src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/SignedXmlDebugLog.cs#L200

Otherwise, the PR looks good, thank you!

determ1ne:
@stephentoub, @buyaa-n I tried to patch all `HexConverter.ToString` to `Convert.ToHexString` or `Convert.ToHexStringLower`, please check if the new commit is appropriate.

buyaa-n:
> @stephentoub, @buyaa-n I tried to patch all `HexConverter.ToString` to `Convert.ToHexString` or `Convert.ToHexStringLower`, please check if the new commit is appropriate.

Thanks, looks good to me, you have got a conflict in `src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicTlsSecret.cs`

> Close https://github.com/dotnet/runtime/issues/60393.
> 
> TryToHexStringLower may be implemented after https://github.com/dotnet/runtime/pull/86556 merged.

This should be contributing to https://github.com/dotnet/runtime/issues/60393 not closing, should be closed when `TryToHexStringLower` implemented

determ1ne:
@buyaa-n I've implemented `Convert.TryToHexStringLower` and the commits were rebased.

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs
```diff
@@ -1782,7 +1782,7 @@ public static ActivityTraceId CreateFromBytes(ReadOnlySpan<byte> idData)
             if (idData.Length != 16)
                 throw new ArgumentOutOfRangeException(nameof(idData));
 
-            return new ActivityTraceId(HexConverter.ToString(idData, HexConverter.Casing.Lower));
+            return new ActivityTraceId(Convert.ToHexStringLower(idData));
```
buyaa-n, line 1785: > error CS0117: 'Convert' does not contain a definition for 'ToHexStringLower'

Ah, looks this project builds for netstandard, .Netframework and previous .NET core releases either need to #if-def to only build on Core current release, or maybe bettor to revert changes.

src/libraries/System.Private.CoreLib/src/System/Convert.cs
```diff
@@ -3102,5 +3101,82 @@ public static bool TryToHexString(ReadOnlySpan<byte> source, Span<char> destinat
...
+            }
+            else if (source.Length > int.MaxValue / 2 || destination.Length > source.Length  * 2)
+            {
+                charsWritten = 0;
+                return false;
+            }
+
+            HexConverter.EncodeToUtf16(source, destination, HexConverter.Casing.Lower);
+            charsWritten = source.Length  * 2;
+            return true;
```
buyaa-n, line 3179:
```suggestion
            else if (source.Length > int.MaxValue / 2 || destination.Length > source.Length * 2)
            {
                charsWritten = 0;
                return false;
            }

            HexConverter.EncodeToUtf16(source, destination, HexConverter.Casing.Lower);
            charsWritten = source.Length * 2;
            return true;
```

src/libraries/System.Runtime/ref/System.Runtime.cs
```diff
@@ -1285,6 +1285,10 @@ public static partial class Convert
         public static string ToHexString(byte[] inArray, int offset, int length) { throw null; }
         public static string ToHexString(System.ReadOnlySpan<byte> bytes) { throw null; }
         public static bool TryToHexString(System.ReadOnlySpan<byte> source, System.Span<char> destination, out int charsWritten) { throw null; }
+        public static string ToHexStringLower(byte[] inArray) { throw null; }
+        public static string ToHexStringLower(byte[] inArray, int offset, int length) { throw null; }
+        public static string ToHexStringLower(System.ReadOnlySpan<byte> bytes) { throw null; }
+        public static bool TryToHexStringLower(System.ReadOnlySpan<byte> source, System.Span<char> destination, out int charsWritten) { throw null; }
```
buyaa-n, line 1291: APIs should be added alphabetically, please move to the bottom after `TryToBase64Chars(...)`, unrelated with your change but please move the `public static bool TryToHexString(System.ReadOnlySpan<byte> source, System.Span<char> destination, out int charsWritten) { throw null; }` too

src/libraries/System.Runtime/tests/System.Runtime.Extensions.Tests/System/Convert.ToHexString.cs
```diff
@@ -106,5 +136,13 @@ public static unsafe void ToHexString(byte[] input, string expected)
...
             Assert.Equal(expected, actual);
         }
+
+        [Theory]
+        [MemberData(nameof(ToHexStringTestData))]
+        public static unsafe void ToHexStringLower(byte[] input, string expected)
+        {
+            string actual = Convert.ToHexStringLower(input);
+            Assert.Equal(expected.ToLower(), actual);
+        }
```
buyaa-n, line 146: Could you add a unit test for `TryToHexStringLower(...)`?

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs
determ1ne: > > error CS0117: 'Convert' does not contain a definition for 'ToHexStringLower'
> 
> Ah, looks this project builds for netstandard, .Netframework and previous .NET core releases either need to #if-def to only build on Core current release, or maybe bettor to revert changes.

Should I use `#if NET9_0_OR_GREATER`?

src/libraries/System.Private.CoreLib/src/System/Convert.cs
determ1ne: Code snippet copied from `Convert.TryToHexString`, will remove the space for both.

src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs
buyaa-n: Yep, thanks

src/libraries/System.Private.CoreLib/src/System/Convert.cs
```diff
@@ -3099,7 +3098,84 @@ public static bool TryToHexString(ReadOnlySpan<byte> source, Span<char> destinat
...
+        /// <param name="charsWritten">When this method returns, contains the number of chars that were written in <paramref name="destination"/>.</param>
+        /// <returns>true if the conversion was successful; otherwise, false.</returns>
+        public static bool TryToHexStringLower(ReadOnlySpan<byte> source, Span<char> destination, out int charsWritten)
+        {
+            if (source.Length == 0)
+            {
+                charsWritten = 0;
+                return true;
+            }
+            else if (source.Length > int.MaxValue / 2 || destination.Length > source.Length  * 2)
```
buyaa-n, line 3171: NIT: missed this
```suggestion
            else if (source.Length > int.MaxValue / 2 || destination.Length > source.Length * 2)
```

src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Utils.cs
```diff
@@ -720,10 +720,17 @@ internal static X509Certificate2Collection BuildBagOfCerts(KeyInfoX509Data keyIn
...
+        internal static string EncodeHexString(byte[] sArray)
+        {
+            return Convert.ToHexString(sArray);
+        }
+#else
         internal static string EncodeHexString(byte[] sArray)
         {
             return HexConverter.ToString(sArray);
         }
+#endif
```
buyaa-n, line 733: This doesn't seem need an if-def, it was added same time as `Convert.ToHexString(...)` added, was it failing in other builds?
```suggestion
        internal static string EncodeHexString(byte[] sArray)
        {
            return Convert.ToHexString(sArray);
        }
```

src/libraries/System.Security.Cryptography.Pkcs/src/Internal/Cryptography/PkcsHelpers.cs
```diff
@@ -364,7 +364,12 @@ public static string ToSerialString(this byte[] serialBytes)
             return ToUpperHexString(serialBytes);
         }
 
-#if NETCOREAPP || NETSTANDARD2_1
+#if NET5_0_OR_GREATER
+        private static string ToUpperHexString(ReadOnlySpan<byte> ba)
+        {
+            return Convert.ToHexString(ba);
+        }
+#elif NETCOREAPP || NETSTANDARD2_1
```
buyaa-n, line 372: I think anything except in [System/Diagnostics/Activity.cs](https://github.com/dotnet/runtime/pull/92483/files#diff-1f21ee16d832bed839eb66474c6e2d0ee2d2775070643086b12f2c0aed014e5b) was failing on other platforms, did they?

src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/SignedXmlDebugLog.cs
```diff
@@ -197,7 +197,11 @@ private static string FormatBytes(byte[]? bytes)
             if (bytes == null)
                 return NullString;
 
+#if NET9_0_OR_GREATER
+            return Convert.ToHexStringLower(bytes);
+#else
             return HexConverter.ToString(bytes, HexConverter.Casing.Lower);
+#endif
```
buyaa-n, line 204: Same here, was this causing a build failure?

src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Utils.cs
determ1ne: From [nuget](https://www.nuget.org/packages/System.Security.Cryptography.Xml/#supportedframeworks-body-tab) the `System.Security.Cryptography.Xml 8.0.0` package is still targeting .net framework, while `Convert.ToHexString` is added in net5.0, so I'm not sure if it will build (or works).

src/libraries/System.Security.Cryptography.Pkcs/src/Internal/Cryptography/PkcsHelpers.cs
determ1ne: same here, `System.Security.Cryptography.Pkcs 8.0.0` is still targeting net46*.

src/libraries/System.Private.CoreLib/src/System/Convert.cs
determ1ne: thanks, resolved in [e56b06a](https://github.com/dotnet/runtime/pull/92483/commits/e56b06ae2f28d494c626abcf409ce33411046953)

src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Utils.cs
determ1ne: Please check the CI result https://github.com/dotnet/runtime/pull/92483/checks?check_run_id=18807551051 , removing the if-def will fail the build.

src/libraries/System.Security.Cryptography.Pkcs/src/Internal/Cryptography/PkcsHelpers.cs
buyaa-n: Well, these files doesn't included in .net framework builds, but it could built for netstandard https://github.com/dotnet/runtime/blob/40d9bd2ae3f06cdaff269a53f0c90d48dff7ed70/src/libraries/System.Security.Cryptography.Xml/src/System.Security.Cryptography.Xml.csproj#L55-L58

src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Utils.cs
buyaa-n: Thanks for checking, the file doesn't included in .netframework build, probably failing on netstandard

buyaa-n (approved):
LGMT, thank you @determ1ne ,

The failure unrelated and known.

cincuranet:
Maybe improvements in `System.Diagnostics.Perf_Activity.ActivityAllocations` (the only suspect in diff): https://github.com/dotnet/perf-autofiling-issues/issues/24761

Issue #92278, opened 2023-09-19
AspnetCore application tries to load Microsoft.DiaSymReader.Native.amd64.dll at runtime on centos linux


neyromant:
### Description

I ran my aspnet core application on centos linux using the LL_DEBUG=libs command and found that it was unsuccessfully trying to load the Microsoft.DiaSymReader.Native.amd64.dll library. This does not lead to any visible consequences, but I would like to understand the reasons for this behavior.
After all, Microsoft.DiaSymReader.Native.amd64.dll is a windows-only library

Can someone tell me what is the reason for this behavior and how to avoid it?

Thanks in advance for any answer.

### Reproduction Steps

Build self contained asp.net core app, run it on linux centos by LL_DEBUG=libs ./my-app

### Expected behavior

An application does not try to load Microsoft.DiaSymReader.Native.amd64.dll on linux

### Actual behavior

Application tries to load Microsoft.DiaSymReader.Native.amd64.dll on linux

### Configuration

.net 7
centos 9
x64

tommcdon:
@neyromant thanks for reporting this issue! I completely agree we do not need to probe for `Microsoft.DiaSymReader.Native.<arch>.dll` on Linux.  This is the code is used for displaying line numbers in exception callstacks.  Here is the code that attempts to load the Windows DLL:

https://github.com/dotnet/runtime/blob/034d27f4a9420e10601447d0385c88de9b4a2925/src/coreclr/vm/ceeload.cpp#L2262-L2273

And if `GetISymUnmanagedReaderNoThrow` returns NULL (as it should on Linux), we will fall back to loading portable PDB's which is supported on all platforms:

https://github.com/dotnet/runtime/blob/034d27f4a9420e10601447d0385c88de9b4a2925/src/coreclr/vm/debugdebugger.cpp#L570-L592

I feel we could #ifdef this code out on Linux as we do not support Windows PDB's on Linux.  Feel free to open a PR for review otherwise we will leave it on the backlog for .NET 9.

neyromant:
@tommcdon thanks for you answer! I made a [draft pull request](https://github.com/dotnet/runtime/pull/92492). Could you take a look to see if I understood your idea correctly?

jkotas:
Fixed by #92492

neyromant:
@tommcdon @jkotas Please tell me,  is it possible to bring this fix to net7 or at least 8? What should I do for this?

jkotas:
[...]

PR #92492, opened 2023-09-22
Avoiding trying to load the native library Microsoft.DiaSymReader.Native.<arch>.dll on Linux

neyromant:
Hi, this is attempt to fix problem #92278


teo-tsirpanis (commented):
How about you bail out at the beginning of the method?

jkotas:
All code related to `Microsoft.DiaSymReader` should be under `FEATURE_ISYM_READER`. Looks like there is a bug in that `FEATURE_ISYM_READER` is defined on all platforms. It should be defined on Windows only.

Could you please try to fix this by defining `FEATURE_ISYM_READER` on Windows only here: https://github.com/dotnet/runtime/blob/2000251605d7978979d1fef101673335513bd765/src/coreclr/clrdefinitions.cmake#L137

neyromant:
@jkotas thank you. I added a condition for adding a definition only for windows

jkotas (approved):
LGTM! Thank you

src/coreclr/clrdefinitions.cmake
```diff
@@ -134,7 +134,9 @@ endif (CLR_CMAKE_TARGET_WIN32 AND (CLR_CMAKE_TARGET_ARCH_AMD64 OR CLR_CMAKE_TARG
 if(FEATURE_INTERPRETER)
   add_compile_definitions(FEATURE_INTERPRETER)
```
jkotas, line 135:
```suggestion
    add_compile_definitions(FEATURE_INTERPRETER)
```
Fix indentation

neyromant:
> LGTM! Thank you

@jkotas Thank you! 
But after merge main branch for resolve conflicts some tests, even those that previously completed successfully, failed.
Tell me, please, what to do?

jkotas:
The failure is not related to this PR. I have opened https://github.com/dotnet/runtime/issues/92518 on it.

tommcdon (approved):
Looks good to me.  Thanks for the contribution!

PR #92505, opened 2023-09-22
Fix asserts about arm64 hardware intrinsic register selection

### LINKED ISSUE #91209: Assertion failed 'targetReg != op2Reg' during 'Generate code'
```c#
// Found by Antigen

using System;
using System.Numerics;
public class TestClass
{
    public struct S1
    {
        public struct S1_D1_F2
        {
        }
    }
    static Vector64<int> s_v64_int_22 = Vector64.Create(6);
    static Vector64<uint> s_v64_uint_23 = Vector64.Create((uint)5);
    static S1 s_s1_52 = new S1();
    float float_63 = 5.125f;
    Vector64<int> v64_int_72 = Vector64.Create(5);
    S1.S1_D1_F2 s1_s1_d1_f2_101 = new S1.S1_D1_F2();
    static int s_loopInvariant = 1;
    public S1.S1_D1_F2 Method3(ref Vector64<uint> p_v64_uint_155, out short p_short_156, float p_float_157, S1 p_s1_158, bool p_bool_159)
    {
        unchecked
        {
            p_short_156 = 15&4;
            int __loopvar3 = 15-4;
            while (15>4)
            {
                if (__loopvar3 > s_loopInvariant)
                    break;
                switch (Vector64.Dot(AdvSimd.MultiplyAddByScalar(s_v64_int_22, s_v64_int_22, s_v64_int_22), v64_int_72 = v64_int_72 & s_v64_int_22))
                {
                    case 2:
                    {
                        break;
                    }
                    case -2:
                    {
                        break;
                    }
                    default:
                    {
                        break;
                    }
                }
            }
            return s1_s1_d1_f2_101;
        }
    }
    public void Method0()
    {
        unchecked
        {
            short short_211 = -5;
            float float_215 = 6.0416665f;
            S1.S1_D1_F2 s1_s1_d1_f2_220 = new S1.S1_D1_F2();
            s1_s1_d1_f2_220 = Method3(ref s_v64_uint_23, out short_211, float_63 -= float_215 *= 15/4, s_s1_52, 15==4);
            return;
        }
    }
    {
        new TestClass().Method0();
    }
}
/* 
Environment:

set COMPlus_TieredCompilation=0

Assert failure(PID 11200 [0x00002bc0], Thread: 10704 [0x29d0]): Assertion failed 'targetReg != op2Reg' in 'TestClass:Method3(byref,byref,float,TestClass+S1,bool):TestClass+S1+S1_D1_F2:this' during 'Generate code' (IL size 83; hash 0x8f2c057a; FullOpts)
    File: D:\a\_work\1\s\src\coreclr\jit\hwintrinsiccodegenarm64.cpp Line: 320
    Image: E:\kpathak\CORE_ROOT\corerun.exe
*/
```

BruceForstall:
There are a bunch of asserts in arm64 hardware intrinsics codegen, for intrinsics with read-modify-write behavior, that the target register is not the same as the non-RMW operand registers, since we do a `mov` to the target register which could trash those argument registers. However, if the target register and the `op1` register are the same, then no `mov` is necessary, and also, the register allocator should already have ensured proper lifetime conflict resolution.

So, put the asserts under a check that the `mov` is required.

Fixes #91209

### DIFF
--- a/src/coreclr/jit/hwintrinsiccodegenarm64.cpp
+++ b/src/coreclr/jit/hwintrinsiccodegenarm64.cpp
@@ -287,10 +287,13 @@ void CodeGen::genHWIntrinsic(GenTreeHWIntrinsic* node)
             {
                 if (isRMW)
                 {
-                    assert(targetReg != op2Reg);
-                    assert(targetReg != op3Reg);
+                    if (targetReg != op1Reg)
+                    {
+                        assert(targetReg != op2Reg);
+                        assert(targetReg != op3Reg);
 
-                    GetEmitter()->emitIns_Mov(INS_mov, emitTypeSize(node), targetReg, op1Reg, /* canSkip */ true);
+                        GetEmitter()->emitIns_Mov(INS_mov, emitTypeSize(node), targetReg, op1Reg, /* canSkip */ true);
+                    }
 
                     HWIntrinsicImmOpHelper helper(this, intrin.op4, node);
 
@@ -317,10 +320,13 @@ void CodeGen::genHWIntrinsic(GenTreeHWIntrinsic* node)
             {
                 if (isRMW)
                 {
-                    assert(targetReg != op2Reg);
-                    assert(targetReg != op3Reg);
+                    if (targetReg != op1Reg)
+                    {
+                        assert(targetReg != op2Reg);
+                        assert(targetReg != op3Reg);
 
-                    GetEmitter()->emitIns_Mov(INS_mov, emitTypeSize(node), targetReg, op1Reg, /* canSkip */ true);
+                        GetEmitter()->emitIns_Mov(INS_mov, emitTypeSize(node), targetReg, op1Reg, /* canSkip */ true);
+                    }
 
                     GetEmitter()->emitIns_R_R_R_I(ins, emitSize, targetReg, op2Reg, op3Reg, 0, opt);
                 }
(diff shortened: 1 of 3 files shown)


BruceForstall:
@kunalspathak @tannergooding PTAL
cc @dotnet/jit-contrib

BruceForstall:
No need to back-port this to .NET 8 since it's zero diff and the change only affects DEBUG asserts.

src/coreclr/jit/hwintrinsiccodegenarm64.cpp
```diff
@@ -287,10 +287,13 @@ void CodeGen::genHWIntrinsic(GenTreeHWIntrinsic* node)
             {
                 if (isRMW)
                 {
-                    assert(targetReg != op2Reg);
-                    assert(targetReg != op3Reg);
+                    if (targetReg != op1Reg)
```
kunalspathak, line 290: Not sure how this solves the problem. We `emitIns_Mov` already skips the move (see `/* canSkip */ true`) if `targetReg == op1Reg`.

BruceForstall: That's correct (that the `mov` is already elided). It solves the problem by putting the asserts inside the `if`.

kunalspathak: Ah, I missed the part where you said: "So, put the asserts under a check that the mov is required.".  But regardless, I think we should fix the problem of updating `delayRegFree` properly based on `lastUse` information.

kunalspathak (approved):

src/coreclr/jit/hwintrinsiccodegenarm64.cpp
BruceForstall: > But regardless, I think we should fix the problem of updating `delayRegFree` properly based on `lastUse` information.

I agree. But I'm not sure it's related to this fix: we still should be able to generate `mla` with all the same registers as arguments (however weird a case that would be).

PR #47046, opened 2021-01-15
Dprintf optimization

PeterSolMS:
Here's the initial implementation for putting the stress log into a memory mapped file and piping the dprintf calls to the stress log.

It's much faster than the dprintf implementation we normally and doesn't take a lock.

There is logic inside StressLog.cpp to switch to the memory mapped file when COMPLUS_StressLogFilename is set. Here's a set of environment variables I've used:

COMPlus_StressLogFilename=GCPerfSimStresslog.log
COMPlus_gcServer=1
COMPlus_LogLevel=9
COMPlus_StressLog=1
COMPlus_StressLogSize=10000000
COMPlus_TotalStressLogSize=1000000000

So this creates a max of 64 GB of stress log.

This mechanism also works for clrgc.dll, which accounts for some complications.

I have an analyzer to go with this that I'm still working on...

PR #92506, opened 2023-09-22
Memory mapped stresslog for NativeAOT

cshung:
This is a WIP PR to port https://github.com/dotnet/runtime/pull/47046 and subsequent changes to NativeAOT so that we can collect a vast amount of stresslog in case we need to.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/stressLog.cpp
+++ b/src/coreclr/nativeaot/Runtime/stressLog.cpp
@@ -30,6 +30,7 @@
 #include "threadstore.h"
 #include "threadstore.inl"
 #include "thread.inl"
+#include "RhConfig.h"
 
 template<typename T> inline T VolatileLoad(T const * pt) { return *(T volatile const *)pt; }
 template<typename T> inline void VolatileStore(T* pt, T val) { *(T volatile *)pt = val; }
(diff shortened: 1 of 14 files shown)


mikem8361:
This will take some work in the native DCR to display memory based stress logs.  Since you are bumping the data contract version in your other PR, then the DCR can at least detect the new format.  There will probably be additions needed to the contract which I can't spell out because I don't understand the changes enough to say.

cshung:
> This will take some work in the native DCR to display memory-based stress logs. 

The debugger should be able to, for the most part, ignore the fact that the `ThreadStressLog` and the `StressLogChunk` objects are allocated from the memory backed by a memory-mapped file. It doesn't need to decode the `StressLogHeader` either since it can reach those objects through other means.

It would be nice if we could make sure this potentially huge memory-mapped region won't get into the crash dump. In my typical usage, I map 32G address space into that file and I'd love to not have a dump file containing that 32G after I have already got a 32G stress log file.

src/coreclr/nativeaot/Runtime/stressLog.cpp
```diff
@@ -80,10 +81,125 @@ const static unsigned __int64 RECYCLE_AGE = 0x40000000L;        // after a billi
 
 /*********************************************************************************/
 
+#ifdef MEMORY_MAPPED_STRESSLOG
```
jkotas, line 84: What would it take to refactor the stresslog implementation to get rid of this duplicated copy and have one implementation that is shared by all runtimes?

cshung: It is indeed a good win from an engineering perspective. I envision if we had a libStresslog, it could be repurposed for other projects as well. 

I did consider that option when I started, but I decided that it was too costly for the quality week. From an implementation perspective, I imagine it will follow a similar path to what we did with eventpipe.

jkotas: I would rather we take the time to do this right than introducing more duplicated code.

PR #92510, opened 2023-09-22
JitDump improvements and other minor cleanup

BruceForstall:
A set of minor cleanups, especially textual typos, grammar, and name/case consistency.

Improvements are made to register and variable liveness dumping. Previously, we would display the old and new live sets. This can be annoying as you often end up figuring out the change in the live sets in your head. So, make it explicit and dump out the actual change: `+` means added, `-` means removed.

E.g., before:
```
Live regs: 0000020000280000 {x19 x21 d9} => 0000820000280000 {x19 x21 d9 d15}
Live vars after [000099]: {V00 V11 V198} => {V00 V01 V11 V198}
```

after:
```
Live regs: 0000020000280000 {x19 x21 d9} + {d15} => 0000820000280000 {x19 x21 d9 d15}
Live vars after [000099]: {V00 V11 V198} +{V01} => {V00 V01 V11 V198}
```

Other things to note:
1. In the LSRA allocation tables, local var intervals for local vars 0-9 were dumped as `V0` to `V9`. Fix this to be `V00` to `V09` to match the `V%02u` format we use everywhere else, so it's easier to search for `V01` (for example) everywhere.
2. Be consistent with the casing of `GenTree`
3. Fix the LSRA `Reuse` event alignment
4. For debug info range output, preface the output with "Debug:" so it's easier to find. Also, include the variable name in the live range update output.
5. Use `HasBase()` and `HasIndex()` more places
6. If a GenTree node has both `GTF_SPILL` and `GTF_SPILLED` set, use the flag character `#` instead of `z` (which indicates `GTF_SPILLED`).
7. If a GenTree node has `IsUnusedValue()` true, use `uXXX` for the node result instead of `tXXX`, to make it obvious. (Note, however, that LSRA doesn't set the unused value bit on the spill/unspill nodes it creates, and apparently the bit isn't checked after that point, so this output change isn't quite as useful as it could be.)

### DIFF
--- a/src/coreclr/jit/codegen.h
+++ b/src/coreclr/jit/codegen.h
@@ -1051,11 +1051,11 @@ protected:
 #endif // !defined(TARGET_64BIT)
 
     //-------------------------------------------------------------------------
-    // genUpdateLifeStore: Do liveness udpate after tree store instructions
+    // genUpdateLifeStore: Do liveness update after tree store instructions
     // were emitted, update result var's home if it was stored on stack.
     //
     // Arguments:
-    //     tree        -  Gentree node
+    //     tree        -  GenTree node
     //     targetReg   -  of the tree
     //     varDsc      -  result value's variable
     //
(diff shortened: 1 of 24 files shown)


BruceForstall:
@dotnet/jit-contrib PTAL

jakobbotsch (approved):
LGTM. Any idea what's up with the linux-arm64 TP diffs? (Though I'm not complaining 😄)

BruceForstall:
[No asm diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=419216&view=ms.vss-build-web.run-extensions-tab), but some not insignificant linux-arm64 tpdiff improvements

> Any idea what's up with the linux-arm64 TP diffs?

No... Maybe for some reason the compiler optimizes "HasIndex/HasBase" better than "Index/Base"? Or, there was one case of rearranging LSRA work below a "return" -- maybe that "return" case is common on linux-arm64?

PR #88840, opened 2023-07-13
Ensure that INumberBase implements IUtf8SpanFormattable

tannergooding:
Since `IUtf8SpanFormattable` is a net new interface in .NET 8, then we need to ensure that `INumberBase` implements it in the same release we can't hit a diamond problem in the future.

PR #86875, opened 2023-05-29
Expose IUtf8SpanParsable and implement it on the primitive numeric types

tannergooding:
This makes progress towards https://github.com/dotnet/runtime/issues/81500

Still not covered are:
* BigInteger
* Boolean
* Complex
* Char (explicit API surface to avoid boxing)
* DateOnly
* DateTime
* DateTimeOffset
* Enum
* Guid
* IPAddress
* IPNetwork
* TimeOnly
* TimeSpan
* Version

PR #92511, opened 2023-09-22
Add missing ArrayPool returns to failure branches

xparadoxical:
#86875 added code that rents arrays from the ArrayPool but doesn't return them in failure branches. #88840 fixed one occurence of this and this PR fixes the other 6, both in Try and throwing methods.


stephentoub (approved):
LGTM.

@tannergooding, thoughts on whether this is important to backport to 8?

Issue #90993, opened 2023-08-23
[API Proposal]: Add `Type.IsReferenceType`


Neme12:
### Background and motivation

`Type` already has the `IsValueType` property, but there's no `IsReferenceType` property. This is necessary because it's not equivalent to just `!type.IsValueType`, and that's because there are types that are neither value or reference types, such as unconstrained generic parameters.

### API Proposal

```diff
 namespace System;

 public abstract partial class Type
 {
     // Existing APIs for reference:
     public bool IsValueType { get; }
     protected virtual bool IsValueTypeImpl();

     // Proposed APIs:
+    public bool IsReferenceType { get; }
+    protected virtual bool IsReferenceTypeImpl();
 }
```

### API Usage

```csharp
if (type.IsReferenceType)
{
    // do something
}
```

steveharter:
ECMA-335 states that at the highest level there are only value types and reference types, so `IsReferenceType` would be the same as `!ValueType`.

For unconstrained generic parameters, yes that is an edge case and I don't think we want to add `IsReferenceType` just to return `false` for that and for `IsValueType`. We already have `IsConstructedGenericType` or `IsGenericTypeDefinition`.

steveharter:
Setting to "future" for now; I believe we'll end up closing this.

cc @jkotas @MichalStrehovsky

Neme12:
It's not just generics, there are also by-ref types. It's not that easy to check whether something is a regular reference type. I think it should be easier. Why not add the property, it feels natural to have it next to the existing `IsValueType`.

En3Tho:
I believe it's better to collect concrete examples with possible workarounds to see if there are inconsistencies that really get into the eye or workarounds that are completely non intuitive or something like that.

Otherwise this is just easily done via extensions method in your own codebase.

jkotas:
Could you please share example of existing code on this repos or elsewhere on github that would benefit from this API? I would not be opposed to introducing a convenience API like this if we have a proof of its usefulness. 

There are multiple conflicting definitions of "reference type". The C# definition is different from ECMA-335 definition. It would be best for the proposal show implementation of the API using existing properties. It would clarify its exact behavior and it is something that we would need to the baseline `System.Type` implementation.

> IsValueTypeImpl

Nit: I do not think that this would need to through public property + protected implementation split. This split is historic artifact, we have not done it for newer System.Type APIs.

Neme12:
> It would be best for the proposal to show implementation of the API using existing properties.

[...]

PR #92512, opened 2023-09-23
Convert a few reflection FCalls to C#

jkotas:


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Type.Internal.cs
```diff
@@ -1,67 +0,0 @@
...
-
-using System.Diagnostics;
-using System.Reflection;
-
-using Internal.Runtime.Augments;
-
-namespace System
-{
-    //
-    // This file contains methods on Type that are internal to the framework.
```
jkotas, line 12: This historic split does not make sense anymore.

Wraith2:
The name `IsActualValueType` seems strange. What is actual about a type which makes it not possible to use `IsValueType` and is there a way to make that distinction clearer?

jkotas:
`IsActualValueType` matches the existing `IsActualEnum` name. The public `Type.IsValueType` and `Type.IsEnum` properties have a questionable behavior for generic parameters: They return true for the generic parameters in some cases (https://github.com/dotnet/runtime/issues/90993#issuecomment-1704014144). It is a behavior one pretty much never wants and it leads to bugs. Also, this quirk makes the API slower. 

More self-descripting name would be `IsValueTypeWithoutTheGenericParameterQuirk` and `IsEnumWithoutTheGenericParameterQuirk` that is a mouthful. I am going to add better comment.

src/coreclr/System.Private.CoreLib/src/System/RuntimeType.CoreCLR.cs
```diff
@@ -3412,6 +3425,30 @@ internal unsafe bool IsActualEnum
...
         }
 
+        internal new unsafe bool IsInterface
+        {
+            get
+            {
+                TypeHandle th = GetNativeTypeHandle();
+
+                bool isInterface = !th.IsTypeDesc && th.AsMethodTable()->IsInterface;
+                GC.KeepAlive(this);
```
EgorBo, line 3435: I wonder if JIT is allowed to fold `typeof(T).IsInterface` without intrinsics at all - the RuntimeType object is on the NonGC heap, `GetNativeTypeHandle()` is basically MethodTable which is mostly immutable 🙂

EgorBo (approved):

src/coreclr/System.Private.CoreLib/src/System/RuntimeType.CoreCLR.cs
jkotas: Yeah, it would be possible in theory. I am not sure whether it would make things simpler overall.

PR #92514, opened 2023-09-23
Improve nullability check for generic `.ctor` parameters

### LINKED ISSUE #92487: `NullabilityInfoContext` reporting incorrect nullability for `ParameterInfo` of generic types
### Description

When it comes to generic types, it is not possible for the `NullabilityInfoContext` class to determine the nullability of members whose type is a generic parameter -- in such cases `NullabilityInfoContext` will fall back to reporting the members as being `Nullable`. This works fine for the cases of `PropertyInfo` and `FieldInfo`, however I found a case where the nullability of a `ParameterInfo` is reported erroneously as `NotNull`.

### Reproduction Steps

```C#
Type type = typeof(MyPoco<string>);
var ctx = new NullabilityInfoContext();

// Property nullability -- works as expected
PropertyInfo prop = type.GetProperties()[0];
NullabilityInfo propNullability = ctx.Create(prop);
Console.WriteLine(propNullability.ReadState); // Nullable

// Parameter nullability -- ReadState & WriteState reported as NotNull
var param = type .GetConstructors()[0].GetParameters()[0];
var info = ctx.Create(param);
Console.WriteLine(info.ReadState); // NotNull
Console.WriteLine(info.WriteState); // NotNull

public class MyPoco<T>
{
    public MyPoco(T value) => Value = value;
    public T Value { get; }
}
```

### Expected behavior

Parameter nullability should be reported as `Nullable`, similar to the property.

### Actual behavior

Reported as `NotNull`.

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

Adding a `notnull` constraint makes nullability information being reported as expected:

```C#
var ctx = new NullabilityInfoContext();

Type type = typeof(MyPoco<string>);
PropertyInfo prop = type.GetProperties()[0];
NullabilityInfo propNullability = ctx.Create(prop);
Console.WriteLine(propNullability.ReadState); // NotNull

var param = typeof(MyPoco<string>).GetConstructors()[0].GetParameters()[0];
var info = ctx.Create(param);
Console.WriteLine(info.ReadState); // NotNull
Console.WriteLine(info.WriteState); // NotNull

public class MyPoco<T> where T : notnull
{
    public MyPoco(T value) => Value = value;
    public T Value { get; }
}
```

karakasa:
Fix #92487 

`NullabilityInfoContext.CheckParameterMetadataType` didn't check for `.ctor` so nullable parameters in `.ctor` was incorrectly reported as `NotNull`.

Is the test sufficient?

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Reflection/NullabilityInfoContext.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Reflection/NullabilityInfoContext.cs
@@ -89,9 +89,7 @@ namespace System.Reflection
             ParameterInfo? metaParameter;
             MemberInfo metaMember;
 
-            MemberInfo member = parameter.Member;
-
-            switch (member)
+            switch (parameter.Member)
             {
                 case ConstructorInfo ctor:
                     var metaCtor = (ConstructorInfo)GetMemberMetadataDefinition(ctor);
@@ -102,14 +100,7 @@ namespace System.Reflection
                 case MethodInfo method:
                     MethodInfo metaMethod = GetMethodMetadataDefinition(method);
                     metaMember = metaMethod;
-                    if (string.IsNullOrEmpty(parameter.Name))
-                    {
-                        metaParameter = metaMethod.ReturnParameter;
-                    }
-                    else
-                    {
-                        metaParameter = GetMetaParameter(metaMethod, parameter);
-                    }
+                    metaParameter = string.IsNullOrEmpty(parameter.Name) ? metaMethod.ReturnParameter : GetMetaParameter(metaMethod, parameter);
                     break;
 
                 default:
--- a/src/libraries/System.Private.CoreLib/src/System/Reflection/NullabilityInfoContext.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Reflection/NullabilityInfoContext.cs
@@ -86,35 +86,56 @@ namespace System.Reflection
 
         private void CheckParameterMetadataType(ParameterInfo parameter, NullabilityInfo nullability)
         {
-            if (parameter.Member is MethodInfo method)
-            {
-                MethodInfo metaMethod = GetMethodMetadataDefinition(method);
-                ParameterInfo? metaParameter = null;
-                if (string.IsNullOrEmpty(parameter.Name))
-                {
-                    metaParameter = metaMethod.ReturnParameter;
-                }
-                else
-                {
-                    ReadOnlySpan<ParameterInfo> parameters = metaMethod.GetParametersAsSpan();
-                    for (int i = 0; i < parameters.Length; i++)
-                    {
-                        if (parameter.Position == i &&
-                            parameter.Name == parameters[i].Name)
-                        {
-                            metaParameter = parameters[i];
-                            break;
-                        }
-                    }
-                }
+            ParameterInfo? metaParameter;
+            MemberInfo metaMember;
 
-                if (metaParameter != null)
-                {
-                    CheckGenericParameters(nullability, metaMethod, metaParameter.ParameterType, parameter.Member.ReflectedType);
-                }
+            MemberInfo member = parameter.Member;
+
+            switch (member)
+            {
+                case ConstructorInfo ctor:
+                    var metaCtor = (ConstructorInfo)GetMemberMetadataDefinition(ctor);
+                    metaMember = metaCtor;
+                    metaParameter = GetMetaParameter(metaCtor, parameter);
+                    break;
+
+                case MethodInfo method:
+                    var metaMethod = GetMethodMetadataDefinition(method);
+                    metaMember = metaMethod;
+                    if (string.IsNullOrEmpty(parameter.Name))
+                    {
+                        metaParameter = metaMethod.ReturnParameter;
+                    }
+                    else
+                    {
+                        metaParameter = GetMetaParameter(metaMethod, parameter);
+                    }
+                    break;
+
+                default:
+                    return;
+            }
+
+            if (metaParameter != null)
+            {
+                CheckGenericParameters(nullability, metaMember, metaParameter.ParameterType, parameter.Member.ReflectedType);
             }
         }
 
+        private static ParameterInfo? GetMetaParameter(MethodBase metaMethod, ParameterInfo parameter)
+        {
+            ReadOnlySpan<ParameterInfo> parameters = metaMethod.GetParametersAsSpan();
+            for (int i = 0; i < parameters.Length; i++)
+            {
+                if (parameter.Position == i &&
+                    parameter.Name == parameters[i].Name)
+                {
+                    return parameters[i];
+                }
+            }
+
+            return null;
+        }
         private static MethodInfo GetMethodMetadataDefinition(MethodInfo method)
         {
             if (method.IsGenericMethod && !method.IsGenericMethodDefinition)
--- a/src/libraries/System.Private.CoreLib/src/System/Reflection/NullabilityInfoContext.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Reflection/NullabilityInfoContext.cs
@@ -100,7 +100,7 @@ namespace System.Reflection
                     break;
 
                 case MethodInfo method:
-                    var metaMethod = GetMethodMetadataDefinition(method);
+                    MethodInfo metaMethod = GetMethodMetadataDefinition(method);
                     metaMember = metaMethod;
                     if (string.IsNullOrEmpty(parameter.Name))
                     {
--- a/src/libraries/System.Runtime/tests/System/Reflection/NullabilityInfoContextTests.cs
+++ b/src/libraries/System.Runtime/tests/System/Reflection/NullabilityInfoContextTests.cs
@@ -1318,48 +1318,51 @@ namespace System.Reflection.Tests
             Assert.Equal(NullabilityState.NotNull, info.GenericTypeArguments[2].GenericTypeArguments[1].ReadState);
         }
 
+        private static Type EnsureReflection([DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] Type type) => type;
+
         public static IEnumerable<object[]> TestCtorParametersData() => new object[][]
         {
-            [typeof(GenericTypeWithCtor<>), NullabilityState.Nullable, NullabilityState.Nullable],
-            [typeof(GenericTypeWithCtor<int>), NullabilityState.NotNull, NullabilityState.NotNull],
-            [typeof(GenericTypeWithCtor<int?>), NullabilityState.Nullable, NullabilityState.Nullable],
-            [typeof(GenericTypeWithCtor<object>), NullabilityState.Nullable, NullabilityState.Nullable],
+            [EnsureReflection(typeof(GenericTypeWithCtor<>)), NullabilityState.Nullable, NullabilityState.Nullable],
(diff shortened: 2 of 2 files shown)


src/libraries/System.Runtime/tests/System/Reflection/NullabilityInfoContextTests.cs
```diff
@@ -1599,4 +1619,9 @@ public class TypeWithPropertiesNestingItsGenericTypeArgument<T>
         public Tuple<int, int?, Tuple<T?>>? Deep4 { get; set; }
         public Tuple<int, int, Tuple<T, int>?>? Deep5 { get; set; }
     }
+
+    public class GenericTypeWithNullableCtor<T>
```
eiriktsarpalis, line 1623: Nit: nothing "nullable" about the type per se:
```suggestion
    public class GenericTypeWithCtor<T>
```

src/libraries/System.Runtime/tests/System/Reflection/NullabilityInfoContextTests.cs
```diff
@@ -1599,4 +1619,9 @@ public class TypeWithPropertiesNestingItsGenericTypeArgument<T>
...
+    {
+        public GenericTypeWithNullableCtor(T value) { }
+    }
```
eiriktsarpalis, line 1626: Could you also include a second type:
```C#
public class GenericNonNullableTypeWithCtor<T> with T : notnull
{
    public GenericNonNullableTypeWithCtor(T value) { }
}
```
and a test confirming that the parameter of `GenericNonNullableTypeWithCtor<object>` is `NotNull`?

src/libraries/System.Private.CoreLib/src/System/Reflection/NullabilityInfoContext.cs
```diff
@@ -86,35 +86,47 @@ public NullabilityInfo Create(ParameterInfo parameterInfo)
...
+        }
 
-                if (metaParameter != null)
+        private static ParameterInfo? GetMetaParameter(MethodBase metaMethod, ParameterInfo parameter)
+        {
+            ReadOnlySpan<ParameterInfo> parameters = metaMethod.GetParametersAsSpan();
+            for (int i = 0; i < parameters.Length; i++)
+            {
+                if (parameter.Position == i &&
+                    parameter.Name == parameters[i].Name)
```
eiriktsarpalis, line 122: Assuming the `metaMethod` we're passing is the correct one, is the second check even necessary?

karakasa: It’s the original logic.

I don’t believe the second check is necessary but I left it there in case of edge cases I don’t know.

eiriktsarpalis (commented):
Do we have equivalent testing for generic methods as well? i.e.
```C#
public class ClassWithGenericMethods
{
    public void GenericMethod<T>(T value) { }
    public void GenericNotNullMethod<T>(T value) where T : notnull { }
}
```

karakasa:
> Do we have equivalent testing for generic methods as well? i.e.

I cannot find one. There're tests for methods inside a generic class with (or without) `notnull` constraints but no generic methods with generic parameters. Let me add some tests.

karakasa:
For reference:

image

<details>
<summary>code</summary>
<pre>
    public class GenericTypeWithCtor<T>
    {
        public GenericTypeWithCtor(T value) { }
    }
    public class GenericTypeWithCtorDisallowAttribute<T>
    {
        public GenericTypeWithCtorDisallowAttribute([DisallowNull] T value) { }
    }
    public class GenericTypeWithCtorMaybeAttribute<T>
    {
        public GenericTypeWithCtorMaybeAttribute([MaybeNull] out T value) { value = default; }
    }
    public class GenericTypeWithCtorAllowAttribute<T>
    {
        public GenericTypeWithCtorAllowAttribute([AllowNull] T value) { }
    }
    public class GenericNonNullableTypeWithCtor<T> where T : notnull
    {
        public GenericNonNullableTypeWithCtor(T value) { }
    }
    public class GenericNonNullableTypeWithCtorDisallowAttribute<T> where T : notnull
    {
        public GenericNonNullableTypeWithCtorDisallowAttribute([DisallowNull] T value) { }
    }
    public class GenericNonNullableTypeWithCtorMaybeAttribute<T> where T : notnull
    {
        public GenericNonNullableTypeWithCtorMaybeAttribute([MaybeNull] out T value) { value = default; }
    }
    public class GenericNonNullableTypeWithCtorAllowAttribute<T> where T : notnull
    {
        public GenericNonNullableTypeWithCtorAllowAttribute([AllowNull] T value) { }
    }
</pre>
</details>

eiriktsarpalis:
Wow, thanks for taking the time to write this down. It looks reasonable to me, but defer to @buyaa for a final decision. 

(Also, TIL that the nullability attributes had effect on `Nullable<T>` members)

src/libraries/System.Runtime/tests/System/Reflection/NullabilityInfoContextTests.cs
```diff
@@ -1317,6 +1317,120 @@ public void TestNullabilityInfoCreationOnPropertiesWithNestedGenericTypeArgument
...
+            [typeof(GenericNonNullableTypeWithCtor_Maybe<>), NullabilityState.Nullable, NullabilityState.NotNull],
+            [typeof(GenericNonNullableTypeWithCtor_Maybe<int>), NullabilityState.NotNull, NullabilityState.NotNull],
+            [typeof(GenericNonNullableTypeWithCtor_Maybe<object>), NullabilityState.Nullable, NullabilityState.NotNull],
+
+            [typeof(GenericNonNullableTypeWithCtor_Allow<>), NullabilityState.NotNull, NullabilityState.Nullable],
+            [typeof(GenericNonNullableTypeWithCtor_Allow<int>), NullabilityState.NotNull, NullabilityState.NotNull],
+            [typeof(GenericNonNullableTypeWithCtor_Allow<object>), NullabilityState.NotNull, NullabilityState.Nullable],
+        };
+
+        [Theory]
```
buyaa-n, line 1360:
```suggestion
        [Theory]
        [SkipOnMono("Nullability attributes trimmed on Mono")]
```

src/libraries/System.Runtime/tests/System/Reflection/NullabilityInfoContextTests.cs
```diff
@@ -1317,6 +1317,120 @@ public void TestNullabilityInfoCreationOnPropertiesWithNestedGenericTypeArgument
...
+                [typeof(ClassWithGenericMethods_Allow), MethodNullable, typeof(int?), NullabilityState.Nullable, NullabilityState.Nullable],
+                [typeof(ClassWithGenericMethods_Allow), MethodNullable, typeof(object), NullabilityState.Nullable, NullabilityState.Nullable],
+
+                [typeof(ClassWithGenericMethods_Allow), MethodNonNullable, null, NullabilityState.NotNull, NullabilityState.Nullable],
+                [typeof(ClassWithGenericMethods_Allow), MethodNonNullable, typeof(int), NullabilityState.NotNull, NullabilityState.NotNull],
+                [typeof(ClassWithGenericMethods_Allow), MethodNonNullable, typeof(object), NullabilityState.NotNull, NullabilityState.Nullable],
+            };
+        } 
+
+        [Theory]
```
buyaa-n, line 1418:
```suggestion
        [Theory]
        [SkipOnMono("Nullability attributes trimmed on Mono")]
```

buyaa-n:
The nullability reference list looks about right to me, thank you @karakasa, I see the PR is still in draft, what is missing to do?

karakasa:
> missing to do?

hi, @buyaa-n 

I believe the test failure is due to `MakeGenericType` lacks metadata, not limited to Mono. 

Should we just skip it on AOT, or make it work through rd.xml, or through put calls in the code?

```
[FAIL] System.Reflection.Tests.NullabilityInfoContextTests.TestMethodsWithGenericParameters(type: typeof(System.Reflection.Tests.ClassWithGenericMethods), methodName: "GenericMethod", genericParameterType: typeof(int?), expectedRead: Nullable, expectedWrite: Nullable)
System.NotSupportedException : 'System.Reflection.Tests.ClassWithGenericMethods.GenericMethod[System.Nullable`1[System.Int32]](System.Nullable`1[System.Int32])' is missing native code. MethodInfo.MakeGenericMethod() is not compatible with AOT compilation. Inspect and fix AOT related warnings that were generated when the app was published. For more information see https://aka.ms/nativeaot-compatibility
```

buyaa-n:
Yes, the failure is not on Mono, I guess Mono doesn't trip attributes of public types, but we had similar issue on Mono for other tests, if you look up some tests has `[SkipOnMono("Nullability attributes trimmed on Mono")]` which are not failing on AOT too. UI would just add the same attributes, but you can skip on AOT only, up to you

buyaa-n:
> or make it work through rd.xml, or through put calls in the code

I don't know if that would work, if you think that could work, then go ahead please, CC @jkotas @MichalStrehovsky

jkotas:
You can get the instantiated MethodInfo in AOT-compatible way by creating delegate and getting the method info from the delegate.

src/libraries/System.Runtime/tests/System/Reflection/NullabilityInfoContextTests.cs
```diff
@@ -1370,60 +1373,83 @@ public void TestCtorParameters(Type type, NullabilityState expectedRead, Nullabi
...
+            [(GenericMethodRef<object>)ClassWithGenericMethods_Maybe.GenericNotNullMethod, NullabilityState.Nullable, NullabilityState.NotNull],
+
+            [(GenericMethod<int>)ClassWithGenericMethods_Allow.GenericMethod, NullabilityState.NotNull, NullabilityState.NotNull],
+            [(GenericMethod<int?>)ClassWithGenericMethods_Allow.GenericMethod, NullabilityState.Nullable, NullabilityState.Nullable],
+            [(GenericMethod<object>)ClassWithGenericMethods_Allow.GenericMethod, NullabilityState.Nullable, NullabilityState.Nullable],
+
+            [(GenericMethod<int>)ClassWithGenericMethods_Allow.GenericNotNullMethod, NullabilityState.NotNull, NullabilityState.NotNull],
+            [(GenericMethod<object>)ClassWithGenericMethods_Allow.GenericNotNullMethod, NullabilityState.NotNull, NullabilityState.Nullable],
+
+            // Specialized delegates are required due to CS8622
```
buyaa-n, line 1436: NIT, do not understand the comment: 
1. Wasn't the delegate version needed because of AOT issue? Looks `CS8622` is a nullability warning
2. Why the comment added here (in between the test data creation) I would assume it applies to all of the test data, probably better to be added above the delegates declaration?

buyaa-n (approved):
Thank you @karakasa, one small question, overall LGTM.

src/libraries/System.Runtime/tests/System/Reflection/NullabilityInfoContextTests.cs
karakasa: > 2\. Why the comment added here (in between the test data creation)

Some of the `ClassWithGenericMethods_Disallow` methods would generate CS8622 warnings if they are converted into delegate types without `[DisallowNull]`. I may put `#pragma warning disable CS8622` around the code instead, but I think it's better to have another annotated delegate type.

image

buyaa-n: Ah, got it, thank you!

Issue #92491, opened 2023-09-22
[macOS] NativeAOT unable to produce executables when using the Xcode 15 toolchain


lambdageek:
When the new linker from Xcode 15 is used on object files produced by NativeAOT, the Xcode linker produces artifacts that crash during the first stack walk.

As a workaround it is possible to install Xcode 14.3.1 in parallel and set it as the default toolchain using `xcode-select`.

This affects arm64 and x64 Macs when using the `-p:PublishAot=true` option with .NET 8

It also affects the experimental ios NativeAOT support.

There is already a PR to fix the the underlying issue here: 
* https://github.com/dotnet/llvm-project/pull/471 (for net9)
* https://github.com/dotnet/llvm-project/pull/472 (for net8.0-rc2 or net8.0-rtm)

I'm opening this issue mostly for tracking purposes

lambdageek:
Do we need to fix net7, too?

filipnavara:
macOS and iOS was not supported for NativeAOT on . NET 7, so no

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

filipnavara:
Another Xcode 15 breakage was reported in https://github.com/dotnet/runtime/issues/92297. Preliminary analysis suggests that we hit the same bug with non-`MH_SUBSECTIONS_VIA_SYMBOLS` object files getting the unwind tables corrupted. In this case, however, it's for the assembly code in both CoreCLR and NativeAOT runtime. For NativeAOT we are much less likely to observe this as a problem but it's easily observable on CoreCLR.

A possible fix is to update the assembly code to use `MH_SUBSECTIONS_VIA_SYMBOLS` and thus avoid hitting the linker bug. I have done an experiment with that (https://github.com/filipnavara/runtime/commit/aba0b02bb5312904d694bd077f1b853538fa7b8b) and now I am waiting for a feedback.

Crucially, I didn't audit Mono code or other components that may run into the same issue.

lambdageek:
> Crucially, I didn't audit Mono code or other components that may run into the same issue.

What would we need to audit:
* handwritten asm files comprising libmonosgen or other platform libraries;
* the aot object files that are linked using the platform toolchain linker when targeting apple platforms
anything else?

And I suppose @dalexsoto the macios team should likewise look at their code?

@steveisok do we have any xcode15 testing lanes?

filipnavara:
> handwritten asm files comprising libmonosgen or other platform libraries;

[...]

Issue #92297, opened 2023-09-19
Exception unwinding broken with Xcode 15 linker


vcsjones:
Recently I've started observing a strange behavior when attempting to run tests where first chance exceptions seem to be crashing the test runner.

For example, running: `./dotnet.sh test src/libraries/System.Formats.Cbor/tests`

Produces:

```
The active test run was aborted. Reason: Test host process crashed : Unhandled exception. System.Formats.Cbor.CborContentException: The current CBOR map contains an incomplete key/value pair.
   at System.Formats.Cbor.CborReader.PeekStateCore() in /Users/vcsjones/Projects/runtime/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Reader/CborReader.PeekState.cs:line 82
   at System.Formats.Cbor.CborReader.SkipNextNode(Int32& depth) in /Users/vcsjones/Projects/runtime/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Reader/CborReader.SkipValue.cs:line 70
   at System.Formats.Cbor.CborReader.SkipToAncestor(Int32 depth, Boolean disableConformanceModeChecks) in /Users/vcsjones/Projects/runtime/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Reader/CborReader.SkipValue.cs:line 51
```

I don't see any reason this exception should crash the test runner.

It's not just in S.F.Cbor. For Tar:

```
 Skipped System.Formats.Tar.Tests.TarWriter_WriteEntry_File_Tests.CreateEntryFromFileOwnedByNonExistentUser [1 ms]
The active test run was aborted. Reason: Test host process crashed : Unhandled exception. Unhandled exception. System.IO.IOException: The file '/var/folders/gc/qfr6cnsj3p1fc1944dkn1t6h0000gn/T/33hxgrfd.1wv/file.tar' already exists.
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirError) in /Users/vcsjones/Projects/runtime/src/libraries/Common/src/Interop/Unix/Interop.IOErrors.cs:line 16
```

I have my suspicion that the issue is due to upgrade Xcode to Xcode 15 (However I am on macOS 13 still).

Currently, I can't run any tests.

```
./dotnet.sh --info
.NET SDK:
 Version:   8.0.100-rc.1.23415.11
 Commit:    12f1505f6e

Runtime Environment:
 OS Name:     Mac OS X
[...]

PR #92301, opened 2023-09-19
Refactor out TargetFramework concept from Microsoft.Interop.SourceGeneration

jkoritzinsky:
Refactor the concept of TargetFramework out of Micrsoft.Interop.SourceGeneration. Replace it with a CodeEmitOptions struct for us to introduce emit switches based on TFM features (in particular, the ability to skip initialization of locals was only introduced in .NET 5, so we need to have a switch for it). As all other use cases of the TFM are downlevel-support-specific in LibraryImportGenerator, move the TFM types into LibraryImportGenerator.

PR #92520, opened 2023-09-23
Make CoreCLR/NativeAOT assembly compile with .subsections_via_symbols on Apple platforms

filipnavara:
Related to #92297 #92491

The new Xcode 15 linker mishandles object files without the MH_SUBSECTIONS_VIA_SYMBOLS flag (`.subsections_via_symbols` assembly code). It ends up treating the whole object file as one function and produces unwind info only for the first function in the file.

### DIFF
--- a/src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.S
+++ b/src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.S
@@ -295,7 +295,7 @@ LEAF_END RhpAssignRef, _TEXT
 
 #ifndef LSE_INSTRUCTIONS_ENABLED_BY_DEFAULT
         PREPARE_EXTERNAL_VAR_INDIRECT_W g_cpuFeatures, 16
-        tbz    w16, #ARM64_ATOMICS_FEATURE_FLAG_BIT, CmpXchgRetry
+        tbz    w16, #ARM64_ATOMICS_FEATURE_FLAG_BIT, LOCAL_LABEL(CmpXchgRetry)
 #endif
 
         mov    x10, x2
@@ -315,7 +315,7 @@ LOCAL_LABEL(CmpXchgRetry):
 
         // Current value matches comparand, attempt to update with the new value.
         stlxr   w12, x1, [x0]
-        cbnz    w12, CmpXchgRetry
+        cbnz    w12, LOCAL_LABEL(CmpXchgRetry)
 #endif
 
 LOCAL_LABEL(DoCardsCmpXchg):
@@ -330,9 +330,9 @@ LOCAL_LABEL(CmpXchgNoUpdate):
         mov     x0, x10
 
 #ifndef LSE_INSTRUCTIONS_ENABLED_BY_DEFAULT
-        tbnz    w16, #ARM64_ATOMICS_FEATURE_FLAG_BIT, NoBarrierCmpXchg
+        tbnz    w16, #ARM64_ATOMICS_FEATURE_FLAG_BIT, LOCAL_LABEL(NoBarrierCmpXchg)
         InterlockedOperationBarrier
-NoBarrierCmpXchg:
+LOCAL_LABEL(NoBarrierCmpXchg):
 #endif
         ret     lr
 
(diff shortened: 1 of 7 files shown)


src/coreclr/nativeaot/Runtime/arm64/WriteBarriers.S
```diff
@@ -224,9 +224,11 @@ LEAF_END RhpByRefAssignRefArm64, _TEXT
 
         PREPARE_EXTERNAL_VAR_INDIRECT g_highest_address, x12
         ccmp    x14, x12, #0x2, hs
-        blo     C_FUNC(RhpAssignRefArm64)
+        bhs     LOCAL_LABEL(NotInHeap)
 
-NotInHeap:
+        b       C_FUNC(RhpAssignRefArm64)
```
filipnavara, line 229: Conditional jump to non-local lables are not allowed in subsections-via-symbols mode. I updated 3 places to use non-conditional branches instead.

Most of the diff is just decorating local labels with `LOCAL_LABEL`.

jkotas (approved):
LGTM. Thank you!

jkotas:
We depend on assembly code to stay together here: https://github.com/dotnet/runtime/blob/5dfd8056825822d3349af3122a3e78584b1b64b1/src/coreclr/vm/amd64/JitHelpers_Fast.asm#L54-L56

Can `.subsections_via_symbols` break this guarantee by reordering methods inside the file?

filipnavara:
> Can `.subsections_via_symbols` break this guarantee by reordering methods inside the file?

In theory it probably can. I'll check what the linker happens to do in practice. I don't think it does reordering, and we don't use "dead strip" or "identical code folding" options that could make this a real problem.

(good catch!)

filipnavara:
I am blocked by unrelated build breakage (related to PR #92301 and discussed there). I don't feel overly confident in the results of the analysis (with and without `-dead_strip`) that I gathered so far, so I will have to resolve the build breakage first to proceed further.

Meanwhile, I checked that some other platforms (eg. arm32) split of this code into `patchedcode.S` file. I suppose we don't depend on unwind information for these particular functions, especially since they are patched at runtime. Would it make sense to do the split for arm64 and amd64 too and turn off `.subsections_via_symbols` for them? That way we would get the original ordering/preservation guarantees and only lose the unwind information for this specific code [with the Xcode 15 linker].

filipnavara:
Checked output on arm64 so far, both in the default configuration and with added `-dead_strip`, with the Xcode 15 default linker and the legacy one (`-ld_classic` switch). In all cases the code was preserved and didn't get reordered. So, for strictly practical point of view, the generated output is correct.

For the theoretical point of view, I think the linker is allowed to do some reordering. The comments in the old linker specifically mention "Atoms in the same section from the same .o file should be contiguous and sequenced in the same order they were in the .o file," but it's not obvious how hard the guarantee is and whether it applies only to the initial order or whether it can be affected by some subsequent optimization. The only two reordering optimizations that I am aware of are a) reordering initialiser/terminator functions to the start/end of code section, and b) reordering DATA section to group together data modified by the `dyld` dynamic linker. Neither of those optimizations applies here.

That leaves us with several options:
1) Do nothing. The downside is that a future linker update may start doing some optimization that breaks the code.
2) Split the patchable code into separate `patchedcode.S` file and disable `.subsections_via_symbols` for it individually. The new linker would lose the unwind info for this specific code but we may not need it anyway. I'd need someone more knowledgable to confirm or deny that assumption. The upside of that approach is that we would still communicate the correct semantics to the linker so it's less likely to be accidentally broken. The downside is that we would still hit the Xcode 15 linker bug but only for a very limited part of code where it may not have consequences.
3) Use the `-order_file` linker switch and explicitly enforce the order of the symbols inside the patchable section. This is potentially an unchartered territory where we may hit another linker bug minefield. It also introduces a maintenance burden where the order file needs to be kept up-to-date with the assembly code.
4) Detect Xcode version and force the use of classic linker with `-ld_classic` switch if Xcode 15 is used. This complicates the build process and it would only buy us some limited amount of time. Apple publicly stated that they intend to remove the old linker in future Xcode update.

*Interestingly, `-dead_strip` does strip about 120Kb of code on debug libcoreclr.dylib build, so it might be worth checking out separately as a size optimization.*

filipnavara:
One more thing, `JIT_UpdateWriteBarrierState` is inside the patched section for .asm (Windows) and outside for .S (non-Windows). Is that intentional?

jkotas:
> One more thing, JIT_UpdateWriteBarrierState is inside the patched section for .asm (Windows) and outside for .S (non-Windows). Is that intentional?

I do not see a good reason for it. I think it should be outside the patched section in both cases.

janvorli (commented):
LGTM, thank you!

PR #92534, opened 2023-09-23
Improve and unify debug views of dictionaries.

### LINKED ISSUE #88736: Debugger display of dictionary types and `DebuggerDisplayAttribute.Name`
[DebuggerDisplayAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debuggerdisplayattribute.-ctor?view=net-7.0) has `Name` and `Type` properties. We never use these.

A place `Name` is useful is collection items. It can be used to display a custom name instead of an index in a debugger window.

Before:

image

After:

image

This seems like an improvement to me. It's easier to see what the keys and values are.

I was wondering why [IDictionaryDebugView](https://github.com/dotnet/runtime/blob/1b2664df569cbe103811377b1a98e744f1afe159/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/IDictionaryDebugView.cs) and other dictionary type proxies don't use this feature. Right now the key and value are squashed together in the debugger, and a meaningless index is displayed as the name.

arturek:
The change allows generic and non-generic dictionaries to be displayed in the same way: with separate columns for keys and values with an ability to expand each column.

Currently non-generic dictionaries are displayed using two columns but without an option to expand keys or values: 
Hashtable debug view

Generic dictionaries are displayed as ordered lists of key/value pairs with a drill-down functionality: 
Dictionary debug view

The PR unifies the way both kinds of dictionaries are displayed by a debugger.  Dictionaries are shown using two-column format (which IMO better represents a dictionary) and both keys and values can be expanded:
New debug view of dictionaries

Fixes #88736


src/libraries/System.Private.CoreLib/src/System/Collections/Generic/IDictionaryDebugView.cs
```diff
@@ -17,12 +17,17 @@ public IDictionaryDebugView(IDictionary<K, V> dictionary)
...
 
         [DebuggerBrowsable(DebuggerBrowsableState.RootHidden)]
-        public KeyValuePair<K, V>[] Items
+        public DebugViewDictionaryItem<K, V>[] Items
         {
             get
             {
-                KeyValuePair<K, V>[] items = new KeyValuePair<K, V>[_dict.Count];
-                _dict.CopyTo(items, 0);
+                var keyValuePairs = new KeyValuePair<K, V>[_dict.Count];
```
eiriktsarpalis, line 24: Is the intermediate KeyValuePair array necessary here? Seems it could be avoided.

eiriktsarpalis (commented):
Awesome, thanks! We should investigate and check if other included dictionary types need the same treatment (including the ones in `System.Collections.Immutable` and `System.Collections.Frozen`).

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/IDictionaryDebugView.cs
arturek: The only way I found to avoid creating this temporary array was to enumerate the dictionary. However, such a change slowed down the code by at least 30%. I think this would affect the debugger UI responsiveness more than the array allocation. Have you thought about some other solution?

eiriktsarpalis: Right, I incorrectly assumed this was only targeting `Dictionary<K,V>` and missed it's targeting the interface. Looks good then.

eiriktsarpalis:
@arturek this seems to be causing test failures in CI, could you take a look?

arturek:
@eiriktsarpalis
Immutable and frozen dictionaries use their own debugger proxy implementation to cache their enumerated items. That proxy displays items the same way as a generic dictionary does it now. I can make updates to them after this PR is completed. Other types of generic dictionaries (in `System.Collections`, `System.Collections.Concurrent`, `System.Collections.NonGeneric` and `System.Private.CoreLib`) are updated by this PR. I hope there are no dictionaries defined in other places.

I'll fix the tests. I thought they are unrelated but now I can see the reason they are failing.

eiriktsarpalis:
Still seeing a number of failing tests, e.g.
```
  Starting:    System.Collections.Immutable.Tests (parallel test collections = on, max threads = 2)
    System.Collections.Tests.DebugView_Tests.TestDebuggerAttributes_Null(obj: []) [FAIL]
      System.InvalidOperationException : Expected one DebuggerTypeProxyAttribute on System.Collections.ListDictionaryInternal.
      Stack Trace:
        /_/src/libraries/Common/tests/System/Diagnostics/DebuggerAttributes.cs(129,0): at System.Diagnostics.DebuggerAttributes.FindAttribute(Type type, Type attributeType)
        /_/src/libraries/Common/tests/System/Diagnostics/DebuggerAttributes.cs(141,0): at System.Diagnostics.DebuggerAttributes.GetProxyType(Type type, Type[] genericTypeArguments)
        /_/src/libraries/Common/tests/System/Collections/DebugView.Tests.cs(166,0): at System.Collections.Tests.DebugView_Tests.TestDebuggerAttributes_Null(Object obj)
    System.Collections.Tests.DebugView_Tests.TestDebuggerAttributes_Null(obj: [System.Collections.DictionaryEntry, System.Collections.DictionaryEntry]) [FAIL]
      System.InvalidOperationException : Expected one DebuggerTypeProxyAttribute on System.Collections.ListDictionaryInternal.
      Stack Trace:
        /_/src/libraries/Common/tests/System/Diagnostics/DebuggerAttributes.cs(129,0): at System.Diagnostics.DebuggerAttributes.FindAttribute(Type type, Type attributeType)
        /_/src/libraries/Common/tests/System/Diagnostics/DebuggerAttributes.cs(141,0): at System.Diagnostics.DebuggerAttributes.GetProxyType(Type type, Type[] genericTypeArguments)
        /_/src/libraries/Common/tests/System/Collections/DebugView.Tests.cs(166,0): at System.Collections.Tests.DebugView_Tests.TestDebuggerAttributes_Null(Object obj)
    System.Collections.Tests.DebugView_Tests.TestDebuggerAttributes_Dictionary(obj: [], expected: []) [FAIL]
      System.InvalidOperationException : Expected one DebuggerDisplayAttribute on System.Collections.ListDictionaryInternal.
      Stack Trace:
        /_/src/libraries/Common/tests/System/Diagnostics/DebuggerAttributes.cs(129,0): at System.Diagnostics.DebuggerAttributes.FindAttribute(Type type, Type attributeType)
        /_/src/libraries/Common/tests/System/Diagnostics/DebuggerAttributes.cs(115,0): at System.Diagnostics.DebuggerAttributes.ValidateDebuggerDisplayReferences(Object obj)
        /_/src/libraries/Common/tests/System/Collections/DebugView.Tests.cs(140,0): at System.Collections.Tests.DebugView_Tests.TestDebuggerAttributes_Dictionary(IDictionary obj, KeyValuePair`2[] expected)
```

arturek:
@eiriktsarpalis:
I'm working on it. The latest issues are quite unexpected.

The tests you mentioned failed because `ListDictionaryInternal` doesn't have `DebuggerTypeProxyAttribute`, but this attribute was [added a month ago](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Collections/ListDictionaryInternal.cs). They also fail because they use the previous definition of the `IDictionaryDebugView` instead of the one modified by this PR. What's even stranger, these tests pass on my machine when run in the System.Collection solution, but not in System.Collection.Immutable (and earlier I used only the former solution to verify my fixes before committing them).

I'll look at it tomorrow.

arturek:
@eiriktsarpalis 
I updated tests and they almost pass. The only failure reason is "_System.Net.Quic.QuicException : The connection timed out from inactivity_" which doesn't seem related to my changes.
I couldn't find out what exactly preprocessor symbols `NETFRAMEWORK` and `NETCOREAPP` represent, and which of them would be better in this case, but current code seems to work as expected.

src/libraries/Common/tests/System/Collections/DebugView.Tests.cs
```diff
@@ -11,14 +12,87 @@ namespace System.Collections.Tests
 {
     public class DebugView_Tests
     {
-        public static IEnumerable<object[]> TestDebuggerAttributes_Inputs()
+        private static IEnumerable<object[]> TestDebuggerAttributes_GenericDictionaries()
```
eiriktsarpalis, line 15: Curious, what necessitated the splitting of the test inputs into separate methods?

src/libraries/Common/tests/System/Collections/DebugView.Tests.cs
```diff
@@ -29,40 +103,84 @@ public static IEnumerable<object[]> TestDebuggerAttributes_Inputs()
...
         [ConditionalTheory(typeof(PlatformDetection), nameof(PlatformDetection.IsDebuggerTypeProxyAttributeSupported))]
-        [MemberData(nameof(TestDebuggerAttributes_Inputs))]
-        public static void TestDebuggerAttributes(object obj)
+        [MemberData(nameof(TestDebuggerAttributes_InputsPresentedAsDictionary))]
+        public static void TestDebuggerAttributes_Dictionary(IDictionary obj, KeyValuePair<string, string>[] expected)
+        {
+            DebuggerAttributes.ValidateDebuggerDisplayReferences(obj);
+            DebuggerAttributeInfo info = DebuggerAttributes.ValidateDebuggerTypeProxyProperties(obj);
+            PropertyInfo itemProperty = info.Properties.Single(pr => pr.GetCustomAttribute<DebuggerBrowsableAttribute>().State == DebuggerBrowsableState.RootHidden);
+            var itemArray = itemProperty.GetValue(info.Instance) as Array;
```
eiriktsarpalis, line 172: Please replace `var` keywords with explicit type names.

src/libraries/Common/tests/System/Diagnostics/DebuggerAttributes.cs
```diff
@@ -86,18 +93,52 @@ public static IEnumerable<PropertyInfo> GetDebuggerVisibleProperties(Type debugg
...
             {
-                throw new InvalidOperationException($"Expected one DebuggerTypeProxyAttribute on {type}.");
+                CustomAttributeData[] attributes = t.GetTypeInfo().CustomAttributes
+                    .Where(a => a.AttributeType == attributeType)
+                    .ToArray();
+                if (attributes.Length != 0)
+                {
+                    if (attributes.Length > 1)
+                    {
+                        throw new InvalidOperationException($"Expected one {attributeType.Name} on {type} but found more.");
```
eiriktsarpalis, line 131: Should this be an assert instead?

src/libraries/System.Collections.NonGeneric/tests/SortedListTests.cs
```diff
@@ -223,19 +223,21 @@ public void DebuggerAttribute_NormalList()
         {
             var list = new SortedList() { { "a", 1 }, { "b", 2 } };
             DebuggerAttributeInfo debuggerAttribute = DebuggerAttributes.ValidateDebuggerTypeProxyProperties(list);
-            PropertyInfo infoProperty = debuggerAttribute.Properties.Single(property => property.Name == "Items");
-            object[] items = (object[])infoProperty.GetValue(debuggerAttribute.Instance);
-            Assert.Equal(list.Count, items.Length);
+            PropertyInfo itemProperty = debuggerAttribute.Properties.Single(pr => pr.GetCustomAttribute<DebuggerBrowsableAttribute>().State == DebuggerBrowsableState.RootHidden);
+            var itemArray = itemProperty.GetValue(debuggerAttribute.Instance) as Array;
```
eiriktsarpalis, line 227: var -> type name please.

src/libraries/Common/tests/System/Diagnostics/DebuggerAttributes.cs
```diff
@@ -86,18 +93,52 @@ public static IEnumerable<PropertyInfo> GetDebuggerVisibleProperties(Type debugg
...
+
+        private static CustomAttributeData FindAttribute(Type type, Type attributeType)
         {
-            // Get the DebuggerTypeProxyAttribute for obj
-            CustomAttributeData[] attrs =
-                type.GetTypeInfo().CustomAttributes
-                .Where(a => a.AttributeType == typeof(DebuggerTypeProxyAttribute))
-                .ToArray();
-            if (attrs.Length != 1)
+            for (var t = type; t != null; t = t.BaseType)
```
eiriktsarpalis, line 122: var -> type name please

src/libraries/System.ObjectModel/tests/ReadOnlyDictionary/ReadOnlyDictionaryTests.cs
```diff
@@ -239,8 +239,8 @@ public static void DebuggerAttributeTests()
             DebuggerAttributes.ValidateDebuggerDisplayReferences(dict);
             DebuggerAttributeInfo info = DebuggerAttributes.ValidateDebuggerTypeProxyProperties(dict);
             PropertyInfo itemProperty = info.Properties.Single(pr => pr.GetCustomAttribute<DebuggerBrowsableAttribute>().State == DebuggerBrowsableState.RootHidden);
-            KeyValuePair<int, int>[] pairs = itemProperty.GetValue(info.Instance) as KeyValuePair<int, int>[];
-            Assert.Equal(dict, pairs);
+            var itemArray = itemProperty.GetValue(info.Instance) as Array;
```
eiriktsarpalis, line 242: var -> type name

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/DebugViewDictionaryItem.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System.Diagnostics;
+
+namespace System.Collections.Generic
+{
+    /// <summary>
+    /// Defines a key/value pair for displaying an item of a dictionary by a debugger.
+    /// </summary>
+    [DebuggerDisplay("{Value}", Name = "[{Key}]")]
+    internal readonly struct DebugViewDictionaryItem<K, V>
```
eiriktsarpalis, line 12: Generic naming conventions:
```suggestion
    internal readonly struct DebugViewDictionaryItem<TKey, TValue>
```

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/IDictionaryDebugView.cs
```diff
@@ -17,12 +17,17 @@ public IDictionaryDebugView(IDictionary<K, V> dictionary)
         }
 
         [DebuggerBrowsable(DebuggerBrowsableState.RootHidden)]
-        public KeyValuePair<K, V>[] Items
+        public DebugViewDictionaryItem<K, V>[] Items
```
eiriktsarpalis, line 20: Any chance you could rename the type parameters to be `TKey` and `TValue` here as well?

src/libraries/Common/tests/System/Collections/DebugView.Tests.cs
arturek: Previously the list contained only generic dictionaries and lists, and all of them were displayed and tested the same way. I added non-generic dictionaries which used a different method of testing.  Depending on the .Net version (Framwork or Core) generic dictionaries now behave similar to elements of either the first or the second group. That's why now there are three groups with two testing methods.

src/libraries/Common/tests/System/Collections/DebugView.Tests.cs
```diff
@@ -11,14 +12,87 @@ namespace System.Collections.Tests
...
+                    new ("[\"b\"]", "\"B\""),
+                }
+            };
+#endif
+        }
+
+        private static IEnumerable<object[]> TestDebuggerAttributes_ListInputs()
         {
-            yield return new object[] { new Dictionary<int, string>() };
-            yield return new object[] { new HashSet<string>() };
```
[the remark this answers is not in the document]
arturek: Thanks. I lost this line while splitting the list.  Fixed.
I checked again whether it was the only missing element and I added one more test case for a non-empty generic SortedList.

src/libraries/Common/tests/System/Diagnostics/DebuggerAttributes.cs
arturek: IMO helper methods such as this one should not determine the outcome of a test (unless it's their main purpose). Test methods can handle exceptions (or not) based on their requirements. These tests worked this way previously, I haven't change it.

src/libraries/Common/tests/System/Diagnostics/DebuggerAttributes.cs
arturek: Done.

src/libraries/Common/tests/System/Collections/DebugView.Tests.cs
arturek: Done.

src/libraries/System.Collections.NonGeneric/tests/SortedListTests.cs
arturek: Done.

src/libraries/System.ObjectModel/tests/ReadOnlyDictionary/ReadOnlyDictionaryTests.cs
arturek: Done.

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/DebugViewDictionaryItem.cs
arturek: Done.

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/IDictionaryDebugView.cs
arturek: Done.

eiriktsarpalis (commented):
Almost forgot, but would it be possible to include this change in immutable collections as well? If you prefer to do this in a future PR that's also fine.

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/DebugViewDictionaryItem.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            Key = keyValue.Key;
+            Value = keyValue.Value;
+        }
+
+        [DebuggerBrowsable(DebuggerBrowsableState.Collapsed)]
+        public TKey Key { get; init; }
+
+        [DebuggerBrowsable(DebuggerBrowsableState.Collapsed)]
+        public TValue Value { get; init; }
```
stephentoub, line 30:
```suggestion
        [DebuggerBrowsable(DebuggerBrowsableState.Collapsed)]
        public TKey Key { get; }

        [DebuggerBrowsable(DebuggerBrowsableState.Collapsed)]
        public TValue Value { get; }
```

src/libraries/System.Private.CoreLib/src/System/Collections/Hashtable.cs
```diff
@@ -1385,7 +1382,7 @@ public override void OnDeserialization(object? sender)
                 // call OnDeserialization on our parent table.
             }
 
-            internal override KeyValuePairs[] ToKeyValuePairsArray()
+            internal override DebugViewDictionaryItem<object, object?>[] ToKeyValuePairsArray()
```
stephentoub, line 1385: Nit: this is no longer a "KeyValuePairsArray"

src/libraries/System.Private.CoreLib/src/System/Collections/Generic/DebugViewDictionaryItem.cs
arturek: Done

src/libraries/System.Private.CoreLib/src/System/Collections/Hashtable.cs
arturek: Done.

arturek:
@eiriktsarpalis 
> Almost forgot, but would it be possible to include this change in immutable collections as well? If you prefer to do this in a future PR that's also fine.

It's on my list. I'd prefer to do it in a separate PR just to close this phase and keep scope of this PR smaller.

eiriktsarpalis (approved):
Thanks!

PR #92538, opened 2023-09-23
Implement InitialCapacity on CborWriter.

### LINKED ISSUE #91978: [API Proposal]: InitialCapacity for CborWriter
### Background and motivation

The `CborWriter` class currently has an internal buffer that grows as data is written to the buffer.

Callers that write large CBOR documents may end up spending a significant amount time constantly re-allocating and copying the buffer as it grows. This was observed in #91969. In this scenario, 98% of the time spent was in `Array.Resize`.

Callers that know before hand that they are going to write a large CBOR document currently don't have an easy way to give a hint that the internal buffer should be larger than the default.

Conversely, callers that know they are going to create small CBOR documents may give a smaller-than-the-default hint to avoid over-allocating the internal buffer.

This is similar to [what we did for AsnWriter](https://github.com/dotnet/runtime/issues/69573) except now we are doing it for `CborWriter`.

### API Proposal

```C#
namespace System.Formats.Cbor;

public partial class CborWriter {
    public CborWriter(int initialCapacity, CborConformanceMode conformanceMode = CborConformanceMode.Strict, bool convertIndefiniteLengthEncodings = false, bool allowMultipleRootLevelValues = false);
}
```

### API Usage

```csharp
CborWriter writer = new CborWriter(initialCapacity: 10000000);
// write 1000000 bytes of stuff to the writer.
```

### Alternative Designs

_No response_

### Risks

_No response_

vcsjones:
Closes #91978

### DIFF
--- a/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
+++ b/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
@@ -3,6 +3,7 @@
 
 using System.Buffers;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Diagnostics;
 using System.Runtime.InteropServices;
 
@@ -11,9 +12,10 @@ namespace System.Formats.Cbor
     /// <summary>A writer for Concise Binary Object Representation (CBOR) encoded data.</summary>
     public partial class CborWriter
     {
+        private const int DefaultCapacitySentinel = -1;
         private static readonly ArrayPool<byte> s_bufferPool = ArrayPool<byte>.Create();
 
-        private byte[] _buffer = null!;
+        private byte[] _buffer;
         private int _offset;
 
         private Stack<StackFrame>? _nestedDataItems;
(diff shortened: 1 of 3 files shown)


src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.cs
```diff
@@ -60,14 +62,42 @@ public partial class CborWriter
...
             CborConformanceModeHelpers.Validate(conformanceMode);
 
             ConformanceMode = conformanceMode;
             ConvertIndefiniteLengthEncodings = convertIndefiniteLengthEncodings;
             AllowMultipleRootLevelValues = allowMultipleRootLevelValues;
             _definiteLength = allowMultipleRootLevelValues ? null : (int?)1;
+
+            _buffer = initialCapacity switch
+            {
+                DefaultCapacitySentinel or 0 => null!,
```
eiriktsarpalis, line 97: This align the field with its nullability annotation and remove the need for null checks in the resize method:
```suggestion
                DefaultCapacitySentinel or 0 => Array.Empty<byte>(),
```

src/libraries/System.Formats.Cbor/tests/Writer/CborWriterTests.cs
```diff
@@ -332,7 +333,36 @@ public static void WriteEncodedValue_ValidPayloadWithTrailingBytes_ShouldThrowAr
...
         public static void InvalidConformanceMode_ShouldThrowArgumentOutOfRangeException(CborConformanceMode mode)
         {
-            Assert.Throws<ArgumentOutOfRangeException>(() => new CborWriter(conformanceMode: mode));
+            Assert.Throws<ArgumentOutOfRangeException>("conformanceMode", () => new CborWriter(conformanceMode: mode));
+        }
+
+        [Theory]
+        [InlineData(-2)]
+        [InlineData(int.MinValue)]
+        public static void InvalidInitialCapacity_ShouldThrowArgumentOutOfRangeException(int capacity)
```
eiriktsarpalis, line 342: Do we want to test for OOM in cases where `initialCapacity` is too large?

vcsjones: In my opinion, no, for two reasons. The first is that doesn't seem like a concern of `CborWriter`'s tests. I don't think the writer should be testing how array allocation works, or doesn't work, in a way that it unconcerned with.

Second: We.. had.. some tests like this in S.S.Cryptography and I think we ended up disabling a few of the tests because some platforms (tvOS?) would crash or behave erratically during an OOM.

eiriktsarpalis (approved):

PR #92539, opened 2023-09-23
Use invariant culture for CBOR

filipnavara:
The tests fail otherwise on my macOS Sonoma machine in English / Czech / Saudi Arabia culture settings:

image

cc @vcsjones

### DIFF
--- a/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Reader/CborReader.Tag.cs
+++ b/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Reader/CborReader.Tag.cs
@@ -74,7 +74,7 @@ namespace System.Formats.Cbor
                 // TODO determine if conformance modes should allow inexact date sting parsing
                 if (!DateTimeOffset.TryParseExact(dateString, CborWriter.Rfc3339FormatString, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out DateTimeOffset result))
                 {
-                    throw new CborContentException(SR.Cbor_Reader_InvalidDateTimeEncoding + " " + dateString);
+                    throw new CborContentException(SR.Cbor_Reader_InvalidDateTimeEncoding);
                 }
 
                 return result;
--- a/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Reader/CborReader.Tag.cs
+++ b/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Reader/CborReader.Tag.cs
@@ -72,9 +72,9 @@ namespace System.Formats.Cbor
                 string dateString = ReadTextString();
 
                 // TODO determine if conformance modes should allow inexact date sting parsing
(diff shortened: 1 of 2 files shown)


src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Reader/CborReader.Tag.cs
```diff
@@ -72,9 +72,9 @@ public DateTimeOffset ReadDateTimeOffset()
                 string dateString = ReadTextString();
 
                 // TODO determine if conformance modes should allow inexact date sting parsing
-                if (!DateTimeOffset.TryParseExact(dateString, CborWriter.Rfc3339FormatString, null, DateTimeStyles.RoundtripKind, out DateTimeOffset result))
+                if (!DateTimeOffset.TryParseExact(dateString, CborWriter.Rfc3339FormatString, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out DateTimeOffset result))
                 {
-                    throw new CborContentException(SR.Cbor_Reader_InvalidDateTimeEncoding);
+                    throw new CborContentException(SR.Cbor_Reader_InvalidDateTimeEncoding + " " + dateString);
```
filipnavara, line 77:
```suggestion
                    throw new CborContentException(SR.Cbor_Reader_InvalidDateTimeEncoding);
```

vcsjones:
Removed "tests" from description so it's clear this is a product code change. Looks good to me, but this could be a breaking change that needs to get documented.

Since we were specifying the format string explicitly:

https://github.com/dotnet/runtime/blob/d45854c970da6dc92f03c59a9550549e3c9281b3/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Tag.cs#L136

How does changing the culture to invariant affect the produced date strings?

filipnavara:
> Removed "tests" from description so it's clear this is a product code change. Looks good to me, but this could be a breaking change that needs to get documented.

Thanks for correcting it.

> Since we were specifying the format string explicitly:
> https://github.com/dotnet/runtime/blob/d45854c970da6dc92f03c59a9550549e3c9281b3/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Tag.cs#L136
> 
> How does changing the culture to invariant affect the produced date strings?

I think the culture I use has some unusual representation of the time zone suffixes but I am honestly not sure. I set it up at some point to debug a similar problem and then forgot about it...

filipnavara:
Actually, the reason why it fails is even funnier:

```csharp
var dto = DateTimeOffset.Now;
Console.WriteLine(dto.ToString("yyyy-MM-ddTHH:mm:ss.FFFFFFFK"));
Console.WriteLine(dto.ToString("yyyy-MM-ddTHH:mm:ss.FFFFFFFK", System.Globalization.CultureInfo.InvariantCulture));
```

outputs

```
1445-03-08T21:47:52.688385+02:00
2023-09-23T21:47:52.688385+02:00
```

Non-Gregorian calendar.

vcsjones:
> Non-Gregorian calendar.

Ah. Yeah. That'd do it.

stephentoub (approved):

PR #92540, opened 2023-09-23
JIT: Register field seqs for GC pointer constants in side table

jakobbotsch:
fgValueNumberTreeConst would not register the field sequences in the side table for TYP_BYREF/TYP_REF. That can lead to spurious diffs after copy propagation kicks in.

### DIFF
--- a/src/coreclr/jit/compiler.h
+++ b/src/coreclr/jit/compiler.h
@@ -5102,6 +5102,9 @@ public:
     // Assign the proper value number to the tree
     void fgValueNumberTreeConst(GenTree* tree);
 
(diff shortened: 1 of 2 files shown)


jakobbotsch:
@MihuBot

jakobbotsch:

[SPMI diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=417020&view=ms.vss-build-web.run-extensions-tab), [JIT diffs](https://github.com/MihuBot/runtime-utils/issues/183)

EgorBo (approved):

PR #92521, opened 2023-09-23
Make `TypeDescriptor` thread safe with custom providers by enlarging `lock` region

karakasa:
### Summary

Fix #92394 and add a test.

[see comment](https://github.com/dotnet/runtime/issues/92394#issuecomment-1732252426) for the reason of race conditions.

The fix is conservative as it just extends the range of `lock` statement. If the performance decrease in this PR is acceptable, I would recommend enlarging `lock` region because it is unlikely to result in bugs.

I did code another aggressive approach (#92548) that minimizes lock use but it seems causing stack overflow in certain conditions so I'm not gonna PR that.

### Risk

There would be performance impact due to the extended use of `lock`.

The bottleneck would be querying many to-be-created `TypeDescriptionNode`s at the same time - as they are forced to be created one by one - which may decrease startup perf. Otherwise the impact should be minimal because most queries quit at the first `if` block.

PR #92548, opened 2023-09-24
Make `TypeDescriptor` thread safe with custom providers and minimize lock use

### LINKED ISSUE #92394: TypeDescriptor.GetProperties(object instance) is not thread-safe
### Description

If multiple threads call TypeDescriptor.GetProperties(object instance) at the same time, they can get different results if the instance has a TypeDescriptorProvider attribute.  I believe this is because of missing locks around reading the WeakHashTables.

### Reproduction Steps

```using System;
using System.ComponentModel;
using System.Threading;

namespace TypeDescriptionProvider_GetProperties_repro
{
	public class SomeTypeProvider : TypeDescriptionProvider
	{
		public static ThreadLocal<bool> Constructed = new ThreadLocal<bool>();
		public static ThreadLocal<bool> GetPropertiesCalled = new ThreadLocal<bool>();
		private class CTD : ICustomTypeDescriptor
		{
			public AttributeCollection GetAttributes() => AttributeCollection.Empty;
			public string? GetClassName() => null;
			public string? GetComponentName() => null;
			public TypeConverter GetConverter() => new TypeConverter();
			public EventDescriptor? GetDefaultEvent() => null;
			public PropertyDescriptor? GetDefaultProperty() => null;
			public object? GetEditor(Type editorBaseType) => null;
			public EventDescriptorCollection GetEvents() => EventDescriptorCollection.Empty;
			public EventDescriptorCollection GetEvents(Attribute[]? attributes) => EventDescriptorCollection.Empty;

			public PropertyDescriptorCollection GetProperties()
			{
				GetPropertiesCalled.Value = true;
				return PropertyDescriptorCollection.Empty;
			}

			public PropertyDescriptorCollection GetProperties(Attribute[]? attributes)
			{
				throw new NotImplementedException();
			}

			public object? GetPropertyOwner(PropertyDescriptor? pd) => null;
		}
		public override ICustomTypeDescriptor GetTypeDescriptor(Type objectType, object? instance)
		{
			Constructed.Value = true;
			return new CTD();
		}
	}

	[TypeDescriptionProvider(typeof(SomeTypeProvider))]
	public sealed class SomeType
	{
		public int SomeProperty { get; set; }
	}

	public static class Program
	{
		public static void ConcurrentTest(SomeType instance)
		{
			var properties = TypeDescriptor.GetProperties(instance);
			Console.WriteLine($"Constructed = {SomeTypeProvider.Constructed.Value} GetPropertiesCalled={SomeTypeProvider.GetPropertiesCalled.Value} properties.Count = {properties.Count} (should be 0)");
		}

		public static void Main()
		{
			const int threadCount = 10;
			// Uncomment and it will get the correct properties for all threads
			// TypeDescriptor.GetProperties(new SomeType());
			using (var finished = new CountdownEvent(threadCount))
			{
				SomeType[] instances = new SomeType[threadCount];
				for (var i = 0; i < threadCount; i++)
				{
					instances[i] = new SomeType();
				}
				for (var i = 0; i < threadCount; i++)
				{
					int _i = i;
					new Thread(() =>
					{
						ConcurrentTest(instances[_i]);
						finished.Signal();
					}).Start();
				}

				finished.Wait();
			}
		}
	}
}

### Expected behavior

Every thread should call the provider and get an empty list of properties.

### Actual behavior

Some threads will not call the provider and get the default properties based on reflection

### Regression?

Not that I know.

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

karakasa:
### Summary

Fix #92394 

Similar to #92521 but the PR takes a more aggressive approach. The PR creates specialized versions of `AddProvider` and `NodeFor(Type)` for use in `CheckDefaultProvider` to minimize lock. Though the side effects are not fully reviewed due to control flow change.

### Risk

The performance should be close to, or better than, the before-PR implementation because the `lock` statement doesn't change.

The PR might introduce bugs as it puts `s_defaultProviders[type] = null` **after** `AddProvider`.

### DIFF
--- a/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/TypeDescriptor.cs
+++ b/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/TypeDescriptor.cs
@@ -190,6 +190,30 @@ namespace System.ComponentModel
         }
 
         /// <summary>
+        /// Specialized version for calls from <see cref="CheckDefaultProvider(Type)"/>.
+        /// Code is adjusted to minimize lock in the aforementioned method.
+        /// </summary>
+        /// <param name="provider"></param>
+        /// <param name="type"></param>
+        private static void AddProvider_FewerLock(TypeDescriptionProvider provider, Type type)
+        {
+            Debug.Assert(provider != null);
+            Debug.Assert(type != null);
+
+            lock (s_providerTable)
+            {
+                // Get the root node, hook it up, and stuff it back into
+                // the provider cache.
+                TypeDescriptionNode node = NodeFor_LockFreeWithCreateDelegator(type);
+                var head = new TypeDescriptionNode(provider) { Next = node };
+                s_providerTable[type] = head;
+                s_providerTypeTable.Clear();
+            }
+
+            Refresh(type);
+        }
+
+        /// <summary>
         /// Adds a type description provider that will be called on to provide
         /// type information for a single object instance. A provider added
         /// using this method will never have its CreateInstance method called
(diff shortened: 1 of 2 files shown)


karakasa:
causing stack overflow.

PR #92550, opened 2023-09-24
Fix memory leak in MemoryCache

verdie-g:
`WeakReference<T>.TryGetTarget` apparently returns false in the target object's finalizer. Here is an example demonstrating that:
```cs
CreateTest();
GC.Collect(GC.MaxGeneration);
GC.WaitForPendingFinalizers();

void CreateTest() => _ = new Test();

class Test
{
    private readonly WeakReference<Test> _wr;

    public Test()
    {
        _wr = new WeakReference<Test>(this);
        Console.WriteLine(_wr.TryGetTarget(out _)); // True
    }

    ~Test()
    {
        Console.WriteLine(_wr.TryGetTarget(out _)); // False
    }
}
```

In MemoryCache that would cause the Stats in the _allStats list to never be cleaned out.
```cs
MemoryCache c = new(new MemoryCacheOptions { TrackStatistics = true });
for (int i = 0; i < 100; i += 1)
{
    CreateThread();
}

for (int i = 0; i < 10; i += 1)
{
    Thread.Sleep(100);
    GC.Collect(GC.MaxGeneration);
    GC.WaitForPendingFinalizers();
}

Console.WriteLine(c.GetCurrentStatistics());

void CreateThread()
{
    Thread t = new(() => { c.Get(""); });
    t.Start();
    t.Join();
}
```
At the end of this program, `c._allStats.Count == 100`.

### DIFF
--- a/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
+++ b/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
@@ -389,10 +389,9 @@ namespace Microsoft.Extensions.Caching.Memory
             {
                 for (int i = 0; i < _allStats.Count; i++)
                 {
-                    if (_allStats[i].TryGetTarget(out Stats? stats) && stats == current)
+                    if (_allStats[i].TryGetTarget(out Stats? stats))
                     {
                         _allStats.RemoveAt(i);
-                        break;
                     }
                 }
 
--- a/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
+++ b/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
@@ -392,6 +392,7 @@ namespace Microsoft.Extensions.Caching.Memory
                     if (!_allStats[i].TryGetTarget(out Stats? stats))
                     {
                         _allStats.RemoveAt(i);
+                        i--;
                     }
                 }
 
--- a/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
+++ b/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
@@ -389,7 +389,7 @@ namespace Microsoft.Extensions.Caching.Memory
             {
                 for (int i = 0; i < _allStats.Count; i++)
                 {
-                    if (_allStats[i].TryGetTarget(out Stats? stats))
+                    if (!_allStats[i].TryGetTarget(out Stats? stats))
                     {
                         _allStats.RemoveAt(i);
                     }


src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
```diff
@@ -389,10 +389,9 @@ private void RemoveFromStats(Stats current)
             {
                 for (int i = 0; i < _allStats.Count; i++)
                 {
-                    if (_allStats[i].TryGetTarget(out Stats? stats) && stats == current)
+                    if (_allStats[i].TryGetTarget(out Stats? stats))
                     {
                         _allStats.RemoveAt(i);
-                        break;
```
verdie-g, line 395: I've removed the `break` as a defensive mechanism. If the code is bug-free, each `RemoveFromStats` should remove a single element from `_allStats`.

src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
```diff
@@ -389,10 +389,9 @@ private void RemoveFromStats(Stats current)
             {
                 for (int i = 0; i < _allStats.Count; i++)
                 {
-                    if (_allStats[i].TryGetTarget(out Stats? stats) && stats == current)
+                    if (_allStats[i].TryGetTarget(out Stats? stats))
```
jkotas, line 392: This is going to remove all elements with non-null target from the collection and the only elements left in the collection are going to be the ones with null target. Is it the desired behavior?

src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
jkotas: > it could also remove elements from threads that recently died too.

Only if the element was not at `i+1`th position. `RemoveAt` is going to move it to `i`th position and the next iteration is going to skip it.

src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
verdie-g: Good catch, fixed. I didn't add any test because they are complex to write givin this code path is in a finalizer.

src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
verdie-g: Good catch. No strong opinion about that behavior. Maybe I should just comment what is the expected behavior?

Jozkee: You could `i--;` after `RemoveAt(i);` to not miss it.

jozkee (commented):
Do you see any downsides to use this aggressive approach on `RemoveFromStats`?

jozkee (approved):
Thanks, @verdie-g.

PR #92555, opened 2023-09-24
Split off patched code into separate .S file and disable subsections-via-symbols for it

### LINKED ISSUE #92554: Should `JIT_UpdateWriteBarrierState` be in the patchable section or not?
In the .S file (non-Windows) the function is defined outside of `JIT_PatchedCodeStart` block:

https://github.com/dotnet/runtime/blob/bbe5c66c50201816c14149fd3f259db86bc1f493/src/coreclr/vm/arm64/asmhelpers.S#L208

In the .asm file (Windows) the function is defined inside the patchable block:

https://github.com/dotnet/runtime/blob/bbe5c66c50201816c14149fd3f259db86bc1f493/src/coreclr/vm/arm64/asmhelpers.asm#L260-L278

Which one is correct?

Ref:  https://github.com/dotnet/runtime/pull/92520#issuecomment-1732518100

filipnavara:
Follow up to #92520.
Fixes #92554


jkotas:
I think we want to take this for .NET 8 together with #92520.

@janvorli Do you agree? If yes, could you please do a second review this PR and #92520 ?

janvorli:
I like moving the patchable code to a separate asm file for the sake of unification with the platforms that already had that and also for the sake of easy reasoning on which code is actually patched.
As for excluding functions in this file from the `.subsections_via_symbols` treatment though, I wonder if the missing unwind info could hurt debugging when stepping through one of these functions. It seems to me that the probability of the functions getting reordered is very low. The doc for the directive says that it can split the sections of the object file into individual **blocks** that can be trimmed if not used. I don't see how that changes the situation w.r.t. function ordering. I am not aware of any explicit guarantee that functions in an .obj file don't get reordered and yet we have been relying on that. 
So, I'd prefer using the directive unconditionally.
As for putting this to 8, based on my reasoning above, I think we don't necessarily need to port it to 8.

janvorli:
@filipnavara I am also very curious about how you have figured out that the problem can be fixed with the `.subsections_via_symbols` directive. I was hit by it on Friday and I wasn't able to find any reference to people hitting similar issue. I got as far as figuring out that only the first function in the .obj got unwind info (in fact, the second one was there too, but with a broken unwind info and there was no further ones), but I wasn't able to figure out a remedy.

janvorli (approved):
LGTM modulo the comment on conditional application of the `.subsections_via_symbols`

filipnavara:
> @filipnavara I am also very curious about how you have figured out that the problem can be fixed with the `.subsections_via_symbols` directive. I was hit by it on Friday and I wasn't able to find any reference to people hitting similar issue. I got as far as figuring out that only the first function in the .obj got unwind info (in fact, the second one was there too, but with a broken unwind info and there was no further ones), but I wasn't able to figure out a remedy.

TL;DR: Pure luck. 

I was aware of the missing `.subsections_via_symbols` flag in ObjWriter since I investigated separate issue related to `-dead_strip` linker switch occasionally causing stack overflows with the old Xcode linker. When I was updating the ObjWriter in C# prototype last week I fixed it without giving it much of a thought. At the same time, I switched to Xcode 15 final build and started hitting the problem. At first, I thought that it was some local change, but I soon realized that it's the new linker, and curiously it happened to be fixed by the very last change I did in the ObjWriter prototype.

Another clue was that the linker also incorrectly reports the undefined function names when removing some of the runtime .o files from the linker command line. Instead of reporting one of the ILC generated methods it consistently reported the first one.

jkotas:
> It seems to me that the probability of the functions getting reordered is very low.

It can be reordered for hot/cold splitting using PGO data or using heuristics based on call graph. I know for sure that the linker on Windows would do that. I am not sure about the macOS linker.

filipnavara:
> It can be reordered for hot/cold splitting using PGO data or using heuristics based on the call graph. I know for sure that the linker on Windows would do that. I am not sure about the macOS linker.

Old ld64 doesn't do that even if the format was designed to allow it. The new linker is undocumented black box but I very much doubt it does any such reordering. There are no command line options for PGO, except for explicit order files which existed since the original macOS 10 linker.

filipnavara:
I addressed the feedback. Should we merge it?

jkotas:
@filipnavara Thank you!

PR #92561, opened 2023-09-24
Remove some nullchecks for frozen objects

EgorBo:
This PR drops redundant nullchecks on top of frozen objects, e.g.:
```cs
bool Test() => typeof(int).IsInterface;
```

```diff
; Method Program:Test():ubyte:this (FullOpts)
       sub      rsp, 40
       mov      rcx, 0x2BC000053F0      ; 'System.Int32'
-      cmp      dword ptr [rcx], ecx
       call     [System.RuntimeType:get_IsInterface():ubyte:this]
       nop      
       add      rsp, 40
       ret      
-; Total bytes of code: 28
+; Total bytes of code: 26
```

### DIFF
--- a/src/coreclr/jit/valuenum.cpp
+++ b/src/coreclr/jit/valuenum.cpp
@@ -1635,7 +1635,7 @@ bool ValueNumStore::IsKnownNonNull(ValueNum vn)
         return false;
     }
 
-    if (IsVNObjHandle(vn))
+    if (IsVNHandle(vn))
     {
         assert(CoercedConstantValue<size_t>(vn) != 0);
         return true;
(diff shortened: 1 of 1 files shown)


src/coreclr/jit/valuenum.cpp
```diff
@@ -1634,6 +1634,13 @@ bool ValueNumStore::IsKnownNonNull(ValueNum vn)
     {
         return false;
     }
+
+    if (IsVNObjHandle(vn))
```
SingleAccretion, line 1638: Why just `ObjHandle`?

I think it could be any handle, as in `fgAddrCouldBeNull`.

EgorBo: Good point! since none of them are allowed to be zero.

EgorBo:
@dotnet/jit-contrib ptal, simple change, https://dev.azure.com/dnceng-public/public/_build/results?buildId=417342&view=ms.vss-build-web.run-extensions-tab

jakobbotsch (approved):

PR #92562, opened 2023-09-24
Add ObjectContentType to getObjectContent

EgorBo:
This PR adds `ObjectContentType* pType` output arg to `getObjectContent` JIT-EE API. So jit no longer need to assume something is a class handle - EE will tell it. Also, this essentially now marks `RuntimeType::m_handle` as CNS_INT [clshandle] for frozen objects - it might allow some potential foldings.

Should also allow some future experiments and proper support for this API on NativeAOT side.


EgorBo:
@MihuBot

EgorBo:
@MihuBot

EgorBo:
FYI: @jkotas just a quick impl to fold accesses to MethodTable (currently, only to its first field dwFlags) -- it leads to constant foldings for e.g. `typeof(T).IsInterface` and catches more cases where we can optimize `IsValueType` but overall diffs are not particulary exciting so I am closing. It also assumes that MethodTable.dwFlags are immutable which might be a fragile statement?

jkotas:
Thank you for giving it a shot!

> It also assumes that MethodTable.dwFlags are immutable which might be a fragile statement?

`dwFlags` happens to be immutable today, but there are other MethodTable fields that are mutable. We would need to be able to communicate this to the JIT somehow.

> support for this API on NativeAOT side.

It would be complicated to make this work well with AOT. AOT compute the MethodTable content only after all code generation is done. We would need to partially compute the MethodTable and communicate to the JIT which parts it is ok to depend on.

PR #92570, opened 2023-09-25
Reduce System.Data.Odbc build configurations

jkotas:
Contributes to #53900

### DIFF
--- a/src/libraries/Common/src/Interop/Unix/Interop.Odbc.cs
+++ b/src/libraries/Common/src/Interop/Unix/Interop.Odbc.cs
@@ -1,0 +1,39 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Libraries
+    {
+        internal const string Odbc32 = "libodbc";
+    }
+
+    internal static partial class Odbc
+    {
+        private static IntPtr DllImportResolver(string libraryName, Assembly assembly, DllImportSearchPath? searchPath)
+        {
+            if (libraryName == Libraries.Odbc32)
+            {
+                if (OperatingSystem.IsMacOS() || OperatingSystem.IsIOS() || OperatingSystem.IsTvOS() || OperatingSystem.IsWatchOS())
+                {
+                    libraryName = "libodbc.2.dylib";
+                }
+                else
+                {
+                    libraryName = "libodbc.so.2";
+                }
+                return NativeLibrary.Load(libraryName, assembly, default);
+            }
+            return default;
+        }
+
+        static Odbc()
+        {
+            NativeLibrary.SetDllImportResolver(Assembly.GetExecutingAssembly(), DllImportResolver);
+        }
+    }
+}
--- a/src/libraries/Common/src/Interop/Unix/Interop.Odbc.cs
+++ b/src/libraries/Common/src/Interop/Unix/Interop.Odbc.cs
@@ -14,26 +14,25 @@ internal static partial class Interop
 
     internal static partial class Odbc
     {
-        private static IntPtr DllImportResolver(string libraryName, Assembly assembly, DllImportSearchPath? searchPath)
+        internal static string GetNativeLibraryName()
         {
-            if (libraryName == Libraries.Odbc32)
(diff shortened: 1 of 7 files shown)


src/libraries/Common/src/Interop/Unix/Interop.Odbc.cs
```diff
@@ -0,0 +1,39 @@
...
+        internal const string Odbc32 = "libodbc";
+    }
+
+    internal static partial class Odbc
+    {
+        private static IntPtr DllImportResolver(string libraryName, Assembly assembly, DllImportSearchPath? searchPath)
+        {
+            if (libraryName == Libraries.Odbc32)
+            {
+                if (OperatingSystem.IsMacOS() || OperatingSystem.IsIOS() || OperatingSystem.IsTvOS() || OperatingSystem.IsWatchOS())
```
danmoseley, line 21: What is the status of WatchOS support? As we seem to not check for it in places we check for tvOS eg

https://github.com/dotnet/runtime/blob/8c566d4ffe865c6cc311eecf92b075863f0f0944/src/libraries/Microsoft.Extensions.FileProviders.Physical/src/PhysicalFilesWatcher.cs#L92

jkotas: We are not consistent. For example, we check for WatchOS here: https://github.com/dotnet/runtime/blob/8c566d4ffe865c6cc311eecf92b075863f0f0944/src/libraries/System.Net.Quic/src/System/Net/Quic/Interop/msquic.cs#L151

In practice, I do not think that this library ever works or gets used on any Apple device OSes (iOS, tvOS, watchOS), so this is all theoretical concern.

src/libraries/Common/src/Interop/Unix/Interop.Odbc.cs
```diff
@@ -0,0 +1,38 @@
...
+    internal static partial class Libraries
+    {
+        internal const string Odbc32 = "libodbc";
+    }
+
+    internal static partial class Odbc
+    {
+        internal static string GetNativeLibraryName()
+        {
+            if (OperatingSystem.IsMacOS() || OperatingSystem.IsIOS() || OperatingSystem.IsTvOS() || OperatingSystem.IsWatchOS())
```
roji, line 19: nit: can be expression-bodied method with conditional (also below)

roji (approved):
LGTM, just checking - this doesn't actually remove support for any Unix OSs (e.g. freebsd), just reorganizes the build configs right?

ViktorHofer:
Would it make sense to also collapse windows and unix configurations together so that we don't need the PNSE throwing assembly build configuration anymore and reduce the inner builds further down from 3 to 1?

ViktorHofer (approved):

src/libraries/Common/src/Interop/Unix/Interop.Odbc.cs
danmoseley: opened a bug to consider holistically.

src/libraries/Common/src/Interop/Unix/Interop.Odbc.cs
jkotas: Yeah, I wrote them as expression-bodies method originally, but I did not like how the long condition looked. I think the current shape is easier to understand.

jkotas:
> Would it make sense to also collapse windows and unix configurations together

You can always replace build time checks with runtime checks. We tend to have Windows/Unix split in many places, so one more or less does not make a huge difference. Also, there are OSes that are neither Windows nor Unix where the merged Windows/Unix implementation may cause observable changes compared to the current PNSE implementation. 

> do these change represent a pattern that we can follow for other libraries that would benefit from reduced build platforms?

Yes, this pattern can be used to avoid explosion of Unix build flavor due to native library differences. This pattern was used by System.Drawing.Common. The only other library where this can be used is System.DirectoryServices.Protocols.

src/libraries/Common/src/Interop/Unix/Interop.Odbc.cs
steveisok: Speculating, but these checks for watchOS may have been added before we said it was out of scope. 

I don't think it would cost a lot to take it out. I don't have a strong opinion either way.

danmoseley: @steveisok tracked by https://github.com/dotnet/runtime/issues/92601

PR #92581, opened 2023-09-25
Fix CFG test helpers defined in TEST section

### LINKED ISSUE #92553: Is the `_TEST` section a typo?
https://github.com/dotnet/runtime/blob/bbe5c66c50201816c14149fd3f259db86bc1f493/src/coreclr/vm/amd64/jithelpers_fast.S#L523-L531

Is this really supposed to be a `_TEST` section or should it be `_TEXT`? cc @jakobbotsch

jakobbotsch:
Other versions of this file (.asm, arm64 versions) are already correct.

Fix #92553

### DIFF
--- a/src/coreclr/vm/amd64/jithelpers_fast.S
+++ b/src/coreclr/vm/amd64/jithelpers_fast.S
@@ -520,12 +520,12 @@ LOCAL_LABEL(ProbeLoop):
 
 LEAF_END_MARKED JIT_StackProbe, _TEXT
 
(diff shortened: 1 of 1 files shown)


filipnavara (approved):

EgorBo (approved):

EgorBo:
Any idea why this didn't fail on CI - are these used?

jakobbotsch:
> Any idea why this didn't fail on CI - are these used?

I don't think putting these in a different section causes any issues.

PR #92597, opened 2023-09-25
Delete TODO that is not relevant

jkotas:


jkotas:
Context: https://github.com/dotnet/runtime/pull/92465#pullrequestreview-1642572154

This is hard do something about unless we delete the ability to configure the stack size using env variable from CoreCLR and move the responsibility of setting the default stack size to host.

akoeplinger (approved):
Thank you!

SamMonoRT (approved):

PR #92599, opened 2023-09-25
Remove Unwrap flag from UniqueComInterfaceMarshaller

jtschuster:
The Unwrap flag only has effect when UniqueInstance is not set. To avoid confusion from anyone referencing this code, we should remove it here.

See [here](https://github.com/dotnet/runtime/blob/eb4c875340cb3647d43df53eca1d42f50df5fda5/src/coreclr/vm/interoplibinterface_comwrappers.cpp#L863) for the code that unwraps a ComObject, which is guarded by a check to ensure UniqueInstance is not set.

NativeAOT needed to move the Unwrap code to inside the `!UniqueInstance` to match behavior of CoreCLR. This should only be noticeable when using ComWrappers to wrap an unwrap the same object in the same NativeAOT instance. In-Proc COM with different servers and clients won't hit this behavior, so I'm not sure it's worth backporting to 8.

### DIFF
--- a/src/tests/Interop/COM/ComWrappers/API/Program.cs
+++ b/src/tests/Interop/COM/ComWrappers/API/Program.cs
@@ -9,6 +9,7 @@ namespace ComWrappersTests
     using System.Diagnostics;
     using System.Runtime.CompilerServices;
     using System.Runtime.InteropServices;
+    using System.Runtime.InteropServices.Marshalling;
 
     using ComWrappersTests.Common;
     using TestLibrary;
@@ -188,14 +189,14 @@ namespace ComWrappersTests
             var testObjUnwrapped = wrappers.GetOrCreateObjectForComInstance(comWrapper, CreateObjectFlags.Unwrap);
             Assert.Same(testObj, testObjUnwrapped);
 
-            // UniqueInstance and Unwrap should always be a new ComObject, never unwrapped
+            // UniqueInstance and Unwrap should always be a new com object, never unwrapped
             var testObjUniqueUnwrapped = wrappers.GetOrCreateObjectForComInstance(comWrapper, CreateObjectFlags.Unwrap | CreateObjectFlags.UniqueInstance);
-            Assert.True(testObjUniqueUnwrapped is ComObject)
             Assert.NotSame(testObj, testObjUniqueUnwrapped);
 
             // Release the wrapper
             int count = Marshal.Release(comWrapper);
-            Assert.Equal(0, count);
+            // Expect 1 reference remaining from UniqueInstance
+            Assert.Equal(1, count);
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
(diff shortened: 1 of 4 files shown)


AaronRobinsonMSFT:
This seems like goodness - thanks! We should have a test though that validates this experience. We have one for `UniqueInstance` and we have an `Unwrap` test. One could imagine we want an explicit test that passes both, but asserts the instance is never unwraped.

src/tests/Interop/COM/ComWrappers/API/Program.cs
```diff
@@ -188,9 +189,14 @@ public void ValidateComInterfaceCreationRoundTrip()
             var testObjUnwrapped = wrappers.GetOrCreateObjectForComInstance(comWrapper, CreateObjectFlags.Unwrap);
             Assert.Same(testObj, testObjUnwrapped);
 
+            // UniqueInstance and Unwrap should always be a new com object, never unwrapped
+            var testObjUniqueUnwrapped = wrappers.GetOrCreateObjectForComInstance(comWrapper, CreateObjectFlags.Unwrap | CreateObjectFlags.UniqueInstance);
+            Assert.NotSame(testObj, testObjUniqueUnwrapped);
```
AaronRobinsonMSFT, line 194: Why not bring back the assert for `ComObject`?

src/tests/Interop/COM/ComWrappers/API/Program.cs
```diff
@@ -188,9 +189,14 @@ public void ValidateComInterfaceCreationRoundTrip()
...
+
             // Release the wrapper
             int count = Marshal.Release(comWrapper);
-            Assert.Equal(0, count);
+            // Expect 1 reference remaining from UniqueInstance
+            Assert.Equal(1, count);
```
AaronRobinsonMSFT, line 199: I would revert this change and use the `ComWrappers.FinalRelease()` to reduce that reference count.

Tests that validate the count drops to 0 are really important so we ensure we create stable and predictable systems.

src/tests/Interop/COM/ComWrappers/API/Program.cs
jtschuster: The ComWrappers implementation returns an `ITestObjectWrapper`. I suppose it probably should check to make sure it is that type rather than the original object.

src/tests/Interop/COM/ComWrappers/API/Program.cs
jtschuster: Added a FinalRelease on the ITestObjectWrapper to do this.

src/tests/Interop/COM/ComWrappers/API/Program.cs
AaronRobinsonMSFT: Oh right. This is doing everything manually. Thanks.

AaronRobinsonMSFT (approved):

jtschuster:
Looks like NativeAOT is unwrapping the object when both flags are set, but CoreCLR isn't.

jkoritzinsky (approved):

PR #92600, opened 2023-09-25
Add illink analyzer support for field/property initializers

sbomer:
The dataflow analyzer was exiting early for cases where the owning symbol was not an `IMethodSymbol`. This meant we weren't running dataflow analysis for field and property initializers.

This fixes it by allowing through cases where the owning symbol is not an `IMethodSymbol`, and adding testcases to validate that we don't hit asserts in the code paths that only light up for methods.

### DIFF
--- a/src/tools/illink/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
+++ b/src/tools/illink/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
@@ -505,6 +505,9 @@ namespace ILLink.RoslynAnalyzer.DataFlow
 
 		public override TValue VisitFlowAnonymousFunction (IFlowAnonymousFunctionOperation operation, LocalDataFlowState<TValue, TValueLattice> state)
 		{
+			// The containing symbol of a lambda is either another method, or a field (for field initializers).
+			// For property initializers, the containing symbol is the compiler-generated backing field.
+			// For property accessors, the containing symbol is the accessor method.
 			Debug.Assert (operation.Symbol.ContainingSymbol is IMethodSymbol or IFieldSymbol);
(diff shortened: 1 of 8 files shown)


src/tools/illink/test/Mono.Linker.Tests/TestCasesRunner/ResultChecker.cs
```diff
@@ -880,7 +880,7 @@ void VerifyLoggedMessages (AssemblyDefinition original, TrimmingTestLogger logge
 											break;
 										}
 										if (memberDefinition.Name == ".ctor" &&
```
vitek-karas, line 882: Nit: This should probably also have `actualName.StartsWith (expectedMember.DeclaringType.FullName)` just like the static `.cctor` above.

src/tools/illink/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -504,7 +505,7 @@ public override TValue VisitObjectCreation (IObjectCreationOperation operation,
 
 		public override TValue VisitFlowAnonymousFunction (IFlowAnonymousFunctionOperation operation, LocalDataFlowState<TValue, TValueLattice> state)
 		{
-			Debug.Assert (operation.Symbol.ContainingSymbol is IMethodSymbol);
+			Debug.Assert (operation.Symbol.ContainingSymbol is IMethodSymbol or IFieldSymbol);
```
vitek-karas, line 508: Why not property as well? What does this translate to:
```csharp
class Helper
{
    public Action<int> DoIt => (x) => Console.WriteLine(x);
}
```

I guess it could be a getter with the body as well...

vitek-karas (approved):

PR #92603, opened 2023-09-25
Include info about system call errors in some exceptions from operating on named mutexes

### LINKED ISSUE #89090: Include info about system call errors in exceptions from creating/opening a named mutex
And some info about relevant arguments to help with diagnosing the issue.

kouvel:
- Added new PAL APIs for creating and opening mutexes that take a string buffer for system call error info. These are called with a stack-allocated buffer and upon error the system call errors are appended to the exception message.
- When there is a system call failure that leads to the PAL API failing, some info is appended to the error string, including the system call, relevant arguments, return value, and `errno`
- `chmod` on OSX seemingly can be interrupted by signals, fixed to retry. Also fixed a couple other small things.

Fixes https://github.com/dotnet/runtime/issues/89090


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.NativeAot.cs
```diff
@@ -210,6 +210,27 @@ public static void SetLastPInvokeError(int error)
             PInvokeMarshal.t_lastError = error;
         }
 
+        /// <summary>
+        /// Begins tracking system call errors on the calling thread during PAL APIs. The system call errors may include
```
jkotas, line 214: Is this going to work well with thread suspension? You may get number of extra PAL API calls if a thread suspension happens during the PInvoke transitions. It may be confusing to see these unrelated calls as part of the error message.

src/coreclr/pal/src/sharedmemory/sharedmemory.cpp
```diff
@@ -221,16 +289,32 @@ int SharedMemoryHelpers::Open(LPCSTR path, int flags, mode_t mode)
...
+            sharedMemoryError = SharedMemoryError::OutOfMemory;
+            break;
 
         default:
-            throw SharedMemoryException(static_cast<DWORD>(SharedMemoryError::IO));
+            sharedMemoryError = SharedMemoryError::IO;
+            break;
+    }
+
+    if (sharedMemoryError != SharedMemoryError::NameTooLong)
```
janvorli, line 306: Is this `if` necessary? It seems we should not be getting that and if we do, we would like to know why.

src/coreclr/pal/src/synchmgr/wait.cpp
```diff
@@ -229,6 +229,7 @@ SignalObjectAndWait(
         bAlertable ? "TRUE" : "FALSE");
 
     CPalThread *thread = InternalGetCurrentThread();
+
```
janvorli, line 232: A nit - there is a couple of files with the only change being an added empty line. Can you please remove them?

src/coreclr/pal/src/thread/thread.cpp
```diff
@@ -2699,6 +2701,96 @@ CPalThread::GetCachedStackLimit()
...
+    if (thread == nullptr || !thread->m_isTrackingSystemCallErrors)
+    {
+        return;
+    }
+
+    const int BufferSize = 256;
+    char *buffer = thread->m_systemCallErrors;
+    if (buffer == nullptr)
+    {
+        buffer = new(std::nothrow) char[BufferSize];
```
janvorli, line 2733: The PAL code should not use the new operator other than the placement one. Please use `InternalNew` instead. Using the new operator ends up calling the operator defined in coreclr runtime, which is not expected and can cause crashes (such issue was fixed in https://github.com/dotnet/coreclr/pull/18457).

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.NativeAot.cs
kouvel: Good point, maybe the tracking can be paused/resumed around suspension, I'll update.

jkotas: I think it may be better to do the tracking for the mutex implementation only (pass the error tracing context explicitly instead of implicitly as thread static).

kouvel: Currently only APIs relevant to named mutexes track errors. I thought it could be expanded to include other APIs if needed. Passing in a tracing context would work too, but what would be the advantage of doing that (wouldn't it end up getting stored on the thread anyway)?

src/coreclr/pal/src/sharedmemory/sharedmemory.cpp
kouvel: The `NameTooLong` case already returns `ERROR_FILENAME_EXCED_RANGE`, for which a `PathTooLongException` exception is thrown, which seemed sufficient to me, but I guess it would be nice to see what the path was. I'll update.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.NativeAot.cs
kouvel: An alternative I considered is to use separate APIs for the mutex-specific ones that include something like a tracing context, but that wouldn't cover waits and it wouldn't be easily expandable. I felt this solution would be simpler and reusable.

jkotas: We have been working on making the Win32 PAL to disappear. The machine-wide mutex is one of the few remaining substantial pieces remaining in the Win32 PAL. Let's consider what the picture may look like if it was moved to some combination of System.Native and C#. It is unlikely we would want to build System.Native-wide syscall tracing mechanism, there is no need for it and it would just add overhead. It would likely be error reporting specific for the machine-wide mutex operations.

> wouldn't it end up getting stored on the thread anyway

In general, it is better to pass things around explicitly instead of implicitly via thread statics when there is a choice. If you have a strong preference for thread statics here, I can live with it.

src/libraries/System.Private.CoreLib/src/System/Threading/Mutex.Windows.cs
```diff
@@ -82,10 +109,41 @@ private static OpenExistingResult OpenExistingWorker(string name, out Mutex? res
...
+            finally
+            {
+                systemCallErrors = Marshal.EndTrackingSystemCallErrors(getSystemCallErrors: !success);
+            }
+
+            if (!success)
+            {
+                SystemException? innerEx =
+                    string.IsNullOrEmpty(systemCallErrors) ? null : GetInnerExceptionForSystemCallErrors(systemCallErrors);
+                throw new ApplicationException(SR.Arg_SynchronizationLockException, innerEx);
```
jkotas, line 128: Nothing else in the repo uses SystemException for this purpose. The error details are included in the error message in situations like this. Prior art:

https://github.com/dotnet/runtime/blob/b945aeda8a827a54eaf4e91183cf3725c1e090fa/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/NativeLibrary.NativeAot.Unix.cs#L43-L45

https://github.com/dotnet/runtime/blob/b945aeda8a827a54eaf4e91183cf3725c1e090fa/src/libraries/Common/src/Interop/Unix/Interop.IOErrors.cs#L181-L183

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.NativeAot.cs
kouvel: Ok how about the following for now?
- Create separate qcalls for creating/opening/releasing mutexes on Unix
- Refactor Mutex.Windows.cs into common/Windows/Unix portions
- Pass in something to the Unix APIs to track system call errors

jkotas: Sounds good to me.

> Mutex.Windows.cs into common/Windows/Unix portions

Nit: Mutex.Windows.cs is <100 lines and the common part is going to be a small fraction of it. To keep things simple, I would just create `Mutex.CoreCLR.Windows.cs` under src\coreclr\system.private.corelib and leave Mutex.Windows.cs alone (maybe just delete `TARGET_UNIX` ifdefs from it since they are not going to be needed anymore).

kouvel: > Nit: Mutex.Windows.cs is <100 lines and the common part is going to be a small fraction of it. To keep things simple, I would just create `Mutex.CoreCLR.Windows.cs` under src\coreclr\system.private.corelib and leave Mutex.Windows.cs alone (maybe just delete `TARGET_UNIX` ifdefs from it since they are not going to be needed anymore).

It seems like we can just fork the API call and keep the common portion mostly the same between runtimes on Windows, I've copied the code for now in the latest commit.

src/coreclr/pal/src/misc/utils.cpp
```diff
@@ -366,3 +366,61 @@ BOOL IsRunningOnMojaveHardenedRuntime()
 }
 
 #endif // __APPLE__
+
+const char *GetFriendlyErrorCodeString(int errorCode, char (&rawErrorCodeStringBuffer)[RawErrorCodeStringBufferSize])
```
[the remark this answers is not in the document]
kouvel: It looks like `strerrorname_np` doesn't exist everywhere, I added a check for it and retained the switch as a fallback. Also changed to use `strerror` for unknown errors.

src/coreclr/pal/src/sharedmemory/sharedmemory.cpp
kouvel: It looks like the exception thrown already includes the mutex name, that should be sufficient for this case.

src/coreclr/pal/src/thread/thread.cpp
kouvel: After updating based on other feedback, `new` is not used anymore and a stack-allocated buffer is passed around instead.

kouvel:
Addressed feedback and fixed a few minor things.

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
```diff
@@ -2148,6 +2148,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)System\Runtime\InteropServices\StandardOleMarshalObject.Windows.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)System\Security\SecureString.Windows.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)System\Threading\LowLevelMonitor.Windows.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)System\Threading\Mutex.Windows.cs" Condition="'$(FeatureCoreCLR)' != 'true'" />
```
jkotas, line 2151: Why can't we use this file for CoreCLR? I think it does exactly what we want.

src/libraries/Common/src/System/IO/Win32Marshal.cs
```diff
@@ -3,6 +3,7 @@
 
 using System.Diagnostics;
 using System.Runtime.InteropServices;
+using Microsoft.Win32.SafeHandles;
```
jkotas, line 6: This looks unused.

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
kouvel: I made a copy of it in CoreCLR as per https://github.com/dotnet/runtime/pull/92603#discussion_r1336464382. I can modify the version in libraries and remove the copy instead if that's preferable, that's kind of what I wanted to do.

jkotas: I meant to make a copy for Unix + CoreCLR PAL config only. Keep Windows intact.

jkotas: Ah, my comment was confusing since I have suggested the `Mutex.CoreCLR.Windows.cs` name. It is probably not the best name for the CoreCLR PAL config. We use Windows to mean CoreCLR Win32 PAL in some places and it is a constant source of confusion.

jkotas: I would keep Mutex.CoreCLR.Unix.cs, fold Mutex.CoreCLR.cs into it, and delete Mutex.CoreCLR.Windows.cs and this condition.

The rest of the changes LGTM. Sorry about the confusion.

kouvel: Ok makes sense.

kouvel:
Updated OP to reflect the current state

src/coreclr/System.Private.CoreLib/src/System/Threading/Mutex.CoreCLR.Unix.cs
```diff
@@ -0,0 +1,158 @@
...
+                CreateMutexCore(0, initiallyOwned, name, AccessRights, out int errorCode, out string? errorDetails);
+
+            if (mutexHandle.IsInvalid)
+            {
+                mutexHandle.SetHandleAsInvalid();
+#if TARGET_UNIX || TARGET_BROWSER || TARGET_WASI
+                if (errorCode == Interop.Errors.ERROR_FILENAME_EXCED_RANGE)
+                    // On Unix, length validation is done by CoreCLR's PAL after converting to utf-8
+                    throw new ArgumentException(SR.Argument_WaitHandleNameTooLong, nameof(name));
+#endif
```
jkotas, line 32:
```suggestion
                if (errorCode == Interop.Errors.ERROR_FILENAME_EXCED_RANGE)
                    // On Unix, length validation is done by CoreCLR's PAL after converting to utf-8
                    throw new ArgumentException(SR.Argument_WaitHandleNameTooLong, nameof(name));
```

src/coreclr/System.Private.CoreLib/src/System/Threading/Mutex.CoreCLR.Unix.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                myHandle.Dispose();
+
+#if TARGET_UNIX || TARGET_BROWSER || TARGET_WASI
+                if (errorCode == Interop.Errors.ERROR_FILENAME_EXCED_RANGE)
+                {
+                    // On Unix, length validation is done by CoreCLR's PAL after converting to utf-8
+                    throw new ArgumentException(SR.Argument_WaitHandleNameTooLong, nameof(name));
+                }
+#endif
```
jkotas, line 64:
```suggestion
                if (errorCode == Interop.Errors.ERROR_FILENAME_EXCED_RANGE)
                {
                    // On Unix, length validation is done by CoreCLR's PAL after converting to utf-8
                    throw new ArgumentException(SR.Argument_WaitHandleNameTooLong, nameof(name));
                }
```

src/coreclr/pal/src/synchobj/mutex.cpp
```diff
@@ -529,39 +577,76 @@ OpenMutexA (
...
+  dwSystemCallErrorsBufferSize -- Size of the buffer pointed to by lpSystemCallErrors in bytes.
+
+  See MSDN docs on OpenMutexW for all other parameters.
 --*/
 
 HANDLE
 PALAPI
-OpenMutexW(
+PAL_OpenMutexW(
        IN DWORD dwDesiredAccess,
```
jkotas, line 612: We can omit the `dwDesiredAccess`and `bInheritHandle` arguments from the `PAL_` version. They are not used.

src/coreclr/pal/src/synchobj/mutex.cpp
```diff
@@ -92,33 +93,74 @@ static CAllowedObjectTypes aotAnyMutex(anyMutexTypeIds, ARRAY_SIZE(anyMutexTypeI
...
+  dwSystemCallErrorsBufferSize -- Size of the buffer pointed to by lpSystemCallErrors in bytes.
+
+  See MSDN docs on CreateMutexW for all other parameters.
 --*/
 
 HANDLE
 PALAPI
-CreateMutexW(
+PAL_CreateMutexW(
     IN LPSECURITY_ATTRIBUTES lpMutexAttributes,
```
jkotas, line 128: Similar, we can omit the `lpMutexAttributes` argument from the `PAL_` version.

jkotas (approved):
Thank you!

src/coreclr/System.Private.CoreLib/src/System/Threading/Mutex.CoreCLR.Unix.cs
```diff
@@ -0,0 +1,154 @@
...
+        ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+        // Unix-specific implementation
+
+        private const int SystemCallErrorsBufferSize = 256;
+
+        private static unsafe SafeWaitHandle CreateMutexCore(
+            nint mutexAttributes,
+            bool initialOwner,
+            string? name,
+            uint desiredAccess,
```
jkotas, line 95: This argument is unused.

src/coreclr/System.Private.CoreLib/src/System/Threading/Mutex.CoreCLR.Unix.cs
```diff
@@ -0,0 +1,135 @@
...
+            result = new Mutex(myHandle);
+            return OpenExistingResult.Success;
+        }
+
+        // Note: To call ReleaseMutex, you must have an ACL granting you
+        // MUTEX_MODIFY_STATE rights (0x0001). The other interesting value
+        // in a Mutex's ACL is MUTEX_ALL_ACCESS (0x1F0001).
+        public void ReleaseMutex()
+        {
+            if (!Interop.Kernel32.ReleaseMutex(SafeWaitHandle))
```
danmoseley, line 75: is error code ever worth getting / exposting from releasemutex? since you're preserving errors in other places.

kouvel: Natural errors aren't expected from `ReleaseMutex`. There can be errors from acquiring the lock (waiting on a mutex) and that's not covered, but most of the issues are likely to be found when creating/opening a mutex. Covering waits is more involved.

PR #92604, opened 2023-09-25
Update the Windows ARM64 unwinder

janvorli:
This change updates the Windows ARM64 unwinder to match the current state in Windows. It contains a fix for a bug that is needed as a prerequisite for a .NET issue fix.

The change omits handling of the MSFT_OP_EC_CONTEX opcode that is specific to the arm64ec and that is not relevant to .NET. Also, the pointer authentication stripping opcode is implemented as dummy, since it is also not relevant to .NET.

### DIFF
--- a/src/coreclr/unwinder/arm64/unwinder.cpp
+++ b/src/coreclr/unwinder/arm64/unwinder.cpp
@@ -1666,7 +1666,7 @@ Routine Description:
 Arguments:
 
     ControlPcRva - Supplies the address where control left the specified
-        function, as an offset relative to the IamgeBase.
+        function, as an offset relative to the ImageBase.
 
     ImageBase - Supplies the base address of the image that contains the
         function being unwound.
@@ -1726,7 +1726,7 @@ Return Value:
     UNREFERENCED_PARAMETER(UnwindFlags);
 
     //
-    // Unless a special frame is enountered, assume that any unwinding
+    // Unless a special frame is encountered, assume that any unwinding
     // will return us to the return address of a call and set the flag
     // appropriately (it will be cleared again if the special cases apply).
     //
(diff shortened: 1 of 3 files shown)


src/coreclr/unwinder/arm64/unwinder.cpp
```diff
@@ -657,20 +1711,22 @@ Return Value:
...
     ULONG_PTR UnwindCodesEndPtr;
     ULONG_PTR UnwindDataPtr;
     ULONG UnwindIndex;
     ULONG UnwindWords;
 
+    UNREFERENCED_PARAMETER(UnwindFlags);
+
     //
-    // Unless a special frame is encountered, assume that any unwinding
+    // Unless a special frame is enountered, assume that any unwinding
```
jkotas, line 1729: Typo

src/coreclr/unwinder/arm64/unwinder.cpp
```diff
@@ -1281,278 +2567,155 @@ Return Value:
...
+        return STATUS_SUCCESS;
     }
 
     //
-    // Second case is more complex: we must step along each operation
-    // to ensure it should be executed.
+    // Make sure out-of-bound stack accesses don't send us into an infinite
+    // unwinding loop.
     //
+#ifdef HOST_WINDOWS
```
jkotas, line 2627: Should this rather be `#if 0` for our copy?

src/coreclr/unwinder/arm64/unwinder.cpp
```diff
@@ -1633,78 +2776,49 @@ BOOL DacUnwindStackFrame(T_CONTEXT *pContext, T_KNONVOLATILE_CONTEXT_POINTERS* p
 }
 
 #if defined(HOST_UNIX)
+
+#undef PRUNTIME_FUNCTION
+//#undef PKNONVOLATILE_CONTEXT_POINTERS
```
jkotas, line 2781: Delete?

jkotas (approved):
Thanks!

Issue #92572, opened 2023-09-25
Test failure: JIT/Regression/JitBlue/Runtime_91209/Runtime_91209/Runtime_91209.cmd


BruceForstall:
arm64, jitstress_isas_nohwintrinsic

https://dev.azure.com/dnceng-public/public/_build/results?buildId=416992&view=ms.vss-test-web.build-test-results-tab&runId=9130856&paneView=debug&resultId=109848

```
set DOTNET_TieredCompilation=0
set DOTNET_EnableHWIntrinsic=0
```

```
21:10:07.093 Running test: JIT/Regression/JitBlue/Runtime_91209/Runtime_91209/Runtime_91209.cmd

Return code:      1
Raw output file:      C:\h\w\B1000A13\w\B1C10A18\uploads\JitBlue\Runtime_91209\Runtime_91209\output.txt
Raw output:
BEGIN EXECUTION
 "C:\h\w\B1000A13\p\corerun.exe" -p "System.Reflection.Metadata.MetadataUpdater.IsSupported=false" -p "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization=true"  Runtime_91209.dll 
Test1
System.PlatformNotSupportedException: Operation is not supported on this platform.
   at System.Runtime.Intrinsics.Arm.AdvSimd.MultiplyAddByScalar(Vector64`1 addend, Vector64`1 left, Vector64`1 right)
   at Runtime_91209.Problem1()
   at Runtime_91209.Test1()
   at __GeneratedMainWrapper.Main()
Expected: 100
Actual: 101
END EXECUTION - FAILED
FAILED
Test failed. Trying to see if dump file was created in C:\cores since 9/23/2023 9:10:07 PM
Test Harness Exitcode is : 1
To run the test:
Set up CORE_ROOT and run.
> C:\h\w\B1000A13\w\B1C10A18\e\JIT\Regression\Regression_2\../JitBlue/Runtime_91209/Runtime_91209/Runtime_91209.cmd
Xunit.Sdk.TrueException: 
Return code:      1
Raw output file:      C:\h\w\B1000A13\w\B1C10A18\uploads\JitBlue\Runtime_91209\Runtime_91209\output.txt
Raw output:
BEGIN EXECUTION
[...]

PR #92607, opened 2023-09-25
Ensure HW intrinsics enabled when running arm64 longmul tests.

### LINKED ISSUE #92571: Test failure: JIT/opt/Multiply/MultiplyLongOps/MultiplyLongOps.cmd
arm64, jitstress_isas_nohwintrinsic

https://dev.azure.com/dnceng-public/public/_build/results?buildId=416992&view=ms.vss-test-web.build-test-results-tab&runId=9130856&paneView=debug

```
set DOTNET_EnableHWIntrinsic=0
```

```
21:10:14.673 Running test: JIT/opt/Multiply/MultiplyLongOps/MultiplyLongOps.cmd

Assert failure(PID 848 [0x00000350], Thread: 7956 [0x1f14]): Assertion failed 'compiler->compIsaSupportedDebugOnly(HWIntrinsicInfo::lookupIsa(intrin.id))' in 'MultiplyLongOpsTest:smnegl_single_cast(int,int,long)' during 'Generate code' (IL size 15; hash 0xe1d89229; FullOpts)

    File: D:\a\_work\1\s\src\coreclr\jit\hwintrinsiccodegenarm64.cpp Line: 210
    Image: C:\h\w\B1000A13\p\corerun.exe

__tmp0_MultiplyLongOps.cs:19:11: error: ARM64: expected string not found in input
 //ARM64: {{^ *}} smaddl {{x[0-9]+}}, {{w[0-9]+}}, {{w[0-9]+}}, {{x[0-9]+}}{{$}}
          ^
__jit_disasm.out:23:73: note: scanning from here
; BEGIN METHOD MultiplyLongOpsTest:smaddl_single_cast(int,int,long,long)
                                                                        ^
__jit_disasm.out:27:2: note: possible intended match here
 stp x19, x20, [sp, #0x10]
 ^

Input file: __jit_disasm.out
Check file: __tmp0_MultiplyLongOps.cs

-dump-input=help explains the following input dump.
...
... lots more ...
```

New test from https://github.com/dotnet/runtime/pull/91886

@c272 @kunalspathak

c272:
Currently, when jitstress runs the HWIntrinsics disabled environment for JIT opts, arm64 longmul tests will fall over as it must have hardware intrinsics enabled to detect the correct assembly output (as per #92571).

This patch simply forces on DOTNET_EnableHWIntrinsic to ensure the test will always function correctly. I'm not sure this is a complete solution to the problem (not sure what category this test should be in, and I'm not familiar enough with the `HWIntrinsics` test templating to move it to there quickly if need be), but this should work.

Let me know what you think.
cc: @kunalspathak

Fixes: https://github.com/dotnet/runtime/issues/92571

### DIFF
--- a/src/tests/JIT/opt/Multiply/MultiplyLongOps.csproj
+++ b/src/tests/JIT/opt/Multiply/MultiplyLongOps.csproj
@@ -14,5 +14,6 @@
 
     <CLRTestEnvironmentVariable Include="DOTNET_TieredCompilation" Value="0" />
     <CLRTestEnvironmentVariable Include="DOTNET_JITMinOpts" Value="0" />
+    <CLRTestEnvironmentVariable Include="DOTNET_EnableHWIntrinsic" Value="1" />
   </ItemGroup>
 </Project>


c272:
Doesn't look like the test failures here are the fault of this patch (the modified test itself seems to be passing)? Failures seem to be in `readytorun`.

kunalspathak:
Failure seems related to https://github.com/dotnet/runtime/issues/92572.

kunalspathak (approved):
LGTM

PR #92608, opened 2023-09-25
CI: Don't run runtime.yml on perf pipeline specific changes

radical:
.. and the same for `dotnet-linker-tests` pipeline. The perf pipeline
specific files are used only by `dotnet-runtime-perf`, and
`runtime-wasm-perf` which currently do not run on PRs.

### DIFF
--- a/eng/pipelines/runtime-linker-tests.yml
+++ b/eng/pipelines/runtime-linker-tests.yml
@@ -22,6 +22,7 @@ trigger:
     - eng/pipelines/coreclr/templates/build-perf*
     - eng/pipelines/coreclr/templates/perf-job.yml
     - eng/pipelines/coreclr/templates/run-perf*
+    - eng/testing/performance/*
 
 schedules:
   - cron: "0 7,19 * * *" # run at 7:00 and 19:00 (UTC) which is 23:00 and 11:00 (PST).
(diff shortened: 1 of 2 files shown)


eng/pipelines/runtime-linker-tests.yml
```diff
@@ -18,6 +18,10 @@ trigger:
     - LICENSE.TXT
     - PATENTS.TXT
     - THIRD-PARTY-NOTICES.TXT
+    - eng/pipelines/coreclr/perf*yml
+    - eng/pipelines/coreclr/templates/build-perf*
+    - eng/pipelines/coreclr/templates/perf-job.yml
+    - eng/pipelines/coreclr/templates/run-perf*
```
LoopedBard3, line 25: I think we can also include the below path in the ignored paths, it contains some of the .proj files and the performance-setup files.
```suggestion
    - eng/testing/performance/*
    
```

eng/pipelines/runtime.yml
```diff
@@ -19,6 +19,10 @@ trigger:
     - LICENSE.TXT
     - PATENTS.TXT
     - THIRD-PARTY-NOTICES.TXT
+    - eng/pipelines/coreclr/perf*yml
+    - eng/pipelines/coreclr/templates/build-perf*
+    - eng/pipelines/coreclr/templates/perf-job.yml
+    - eng/pipelines/coreclr/templates/run-perf*
```
LoopedBard3, line 26: Same as above: I think we can also include the below path in the ignored paths, it contains some of the .proj files and the performance-setup files.
```suggestion
    - eng/testing/performance/*
    
```

LoopedBard3 (approved):

PR #92611, opened 2023-09-25
Port SpinWait(int iterations) to C#

EgorBo:
It seems to be simple unless I am missing something. Presumably, it should allow gc to interrupt at any point while previously we had to enter preemptive mode for too many iterations and could slightly stall gc in coop.

Also, I heard modern ARM cpus should use ISB instead of Yield

### DIFF
--- a/src/coreclr/System.Private.CoreLib/src/System/Threading/Thread.CoreCLR.cs
+++ b/src/coreclr/System.Private.CoreLib/src/System/Threading/Thread.CoreCLR.cs
@@ -145,18 +145,16 @@ namespace System.Threading
             // void YieldProcessorNormalized(const YieldProcessorNormalizationInfo &normalizationInfo, unsigned int count)
 
             // Prevent overflow on the multiply below on 32bit
-            if (IntPtr.Size == 4)
+#if TARGET_32BIT
+            // Copied from YieldProcessorNormalization (yieldprocessornormalized.h)
+            const int targetNsPerNormalizedYield = 37;
+            const int maxYieldsPerNormalizedYield = targetNsPerNormalizedYield * 10;
+            const int maxCount = int.MaxValue / maxYieldsPerNormalizedYield;
+            if (iterations > maxCount)
             {
-                // Copied from YieldProcessorNormalization (yieldprocessornormalized.h)
-                const int targetNsPerNormalizedYield = 37;
-                const int maxYieldsPerNormalizedYield = targetNsPerNormalizedYield * 10;
-
-                const int maxCount = int.MaxValue / maxYieldsPerNormalizedYield;
-                if (iterations > maxCount)
-                {
-                    iterations = maxCount;
-                }
+                iterations = maxCount;
             }
+#endif
 
             nint n = (nint)iterations * GetYieldsPerNormalizedYield;
             Debug.Assert(n != 0);
--- a/src/coreclr/System.Private.CoreLib/src/System/Threading/Thread.CoreCLR.cs
+++ b/src/coreclr/System.Private.CoreLib/src/System/Threading/Thread.CoreCLR.cs
@@ -377,8 +377,7 @@ namespace System.Threading
         }
 
         /// <summary>
-        /// Max value to be passed into <see cref="SpinWait(int)"/> for optimal delaying. This value is normalized to be
-        /// appropriate for the processor.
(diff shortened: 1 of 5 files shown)


EgorBo:
Ah, I probably have to keep the callback for RISC-V/LA, etc

EgorBo:
Looks like Rust switched to ISB over YIELD: https://github.com/rust-lang/rust/commit/c064b6560b7ce0adeb9bbf5d7dcf12b1acb0c807

EgorBo:
@jkotas @kouvel should be ready for review unless I am missing something.

Should/can it be shared with NAOT (my understanding that NAOT doesn't calculate optimal number of yields for current cpu and has some hard-coded values instead)

jkotas:
What does it help with? If you are worried about the GC suspension starvation, it would be less code duplication to fix it by moving the check for coop threshold to managed code, like it is done in native AOT: https://github.com/dotnet/runtime/blob/867f5d483e340af033f2ba83d29b21bab0e3bcb3/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Thread.NativeAot.cs#L346-L353

jkotas:
> Looks like Rust switched to ISB over YIELD: https://github.com/rust-lang/rust/commit/c064b6560b7ce0adeb9bbf5d7dcf12b1acb0c807

I do not think that the justification in this PR applies to our use. We do not care about long the yield takes - the calibration of the spin loops will take care of yield taking less time.

We do care about power saving and hyper-threading core scheduling effects that this PR has no data about.

jkotas:
> Yield is a sort of nop

The Arm manual suggests that the hw implementation may take advantage of the hint. So it can be nop on one Arm hw, but actual useful hint on a different Arm hw. It does not make it easier to measure.

> ISB is closer to x64's pause so should be better energy-wise.

Wall clock time != energy consumption. It would be a good idea to check about this with Arm folks.

src/coreclr/System.Private.CoreLib/src/System/Threading/Thread.CoreCLR.cs
```diff
@@ -126,7 +128,53 @@ private void StartCallback()
...
+                    // TODO: on some CPUs it's better to call ISB here
+                    ArmBase.Yield();
+                }
+                else if (X86Base.IsSupported)
+                {
+                    X86Base.Pause();
+                }
+                else
+                {
+                    throw new PlatformNotSupportedException();
```
danmoseley, line 174: Will this normally be elided by the jit?

EgorBo: Yep, this was mostly needed to satisfy our SIMD analyzer that got confused in this method

jkotas:
> > What does it help with? If you are worried about the GC suspension starvation, it would be less code duplication to fix it by moving the check for coop threshold to managed code, like it is done in native AOT:

> Ah, ok, will change to that

Do you plan to work on this? If not, I am going to do it - it will help to reduce overhead for small number of spin counts.

EgorBo:
> > > What does it help with? If you are worried about the GC suspension starvation, it would be less code duplication to fix it by moving the check for coop threshold to managed code, like it is done in native AOT:
> 
> > Ah, ok, will change to that
> 
> Do you plan to work on this? If not, I am going to do it - it will help to reduce overhead for small number of spin counts.

Ah, sure feel free, I am not sure I understood what you meant after looking at it

kunalspathak:
> I'll ask the ARM folks about it, I thought that Yield is a sort of nop while ISB is closer to x64's pause so should be better energy-wise.

@a74nh @TamarChristinaArm - do you have more information on this?

a74nh:
Yes, in current hardware, yield is mostly a nop.
 
Switching to an ISB will create a stall and therefore reduce power. Also, it should improve spin lock performance. Architecturally, it isn’t the best solution in the long term as it’s just relying on a side effect of the ISB forcing the cpu flush.
 
An alternative would be to create a data dependency between the loads (i.e. between the data returned from one load, and the address of the next) which would essentially create a pause which a full cpu flush. However, that’s not been tested in anger – maybe something for the new year.
 
Future hardware will have [WFET](https://developer.arm.com/documentation/ddi0596/2021-12/Base-Instructions/WFET--Wait-For-Event-with-Timeout-) and should be a better fit, but that’s not available yet.

Issue #92219, opened 2023-09-18
Remaining .NET 8 work for TensorPrimitives


stephentoub:
### RC2
- [x] Add naive implementations of `Add`, `AddMultiply`, `Cosh`, `Divide`, `Exp`, `Log`, `Multiply`, `MultiplyAdd`, `Negate`, `Subtract`, `Sinh`, `Tanh` - @tannergooding
- [x] Add naive implementations of `CosineSimilarity`, `Distance`, `Dot`, `Norm`, `Sigmoid`, `SoftMax` - @michaelgsharp  
- [x] Add naive implementations of `Max/Min{Magnitude}`, `IndexOfMax/Min{Magnitude}`, `ConvertToHalf/Single`, `Log2` - @stephentoub 
- [x] Add vectorized implementations of `Abs`, `Sum`, `SumOfSquares`, `SumOfMagnitudes`, `Product`, `ProductOfSums`, `ProductOfDifferences` - @stephentoub
- [x] Vectorize `Add`, `Subtract`, `Multiply`, `Divide`, `Negate`, `AddMultiply`, `MultiplyAdd` - @stephentoub
- [x] Backport merged functionality to release/8.0-rc2 - @michaelgsharp 
- [x] Rename `L2Normalize` to `Norm`? - @stephentoub

### GA
- [x] Consider source rename of `netcore.cs` and `netstandard.cs` since they don't only apply to those target framework identifiers? per @stephentoub it maps to our existing naming conventions
- [x] Ensure packages set up to ship XML docs correctly.
- [x] Review/update XML comments for API docs - @stephentoub
- [x] Update the API proposal issue to reflect the new APIs we added since that issue was approved - @stephentoub
- [x] Add content to the RC2 release notes with call to action to try out preview package - @luisquintanilla 
  - [x] Collect benchmarks to include in the post - @michaelgsharp 
- [x] Make it safe across all operations to use an input span as a destination, as well, test, and document - @stephentoub
- [x] Update `CosineSimilarity` implementations to vectorize remainder processing - @stephentoub
- [x] Update `Aggregate` helper to vectorize remainder processing (`Distance`, `Dot`, `Norm`, `Sum`, `SumOfSquares`, `SumOfMagnitudes`, `Product`, `ProductOfSums`, `ProductOfDifferences`) - @stephentoub
- [x] Vectorize `Exp` - @tannergooding
- [x] Vectorize `Softmax` - @stephentoub
[...]

PR #92618, opened 2023-09-26
Vectorize TensorPrimitives.Min/Max{Magnitude}

stephentoub:

### DIFF
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
@@ -368,44 +368,46 @@ namespace System.Numerics.Tensors
                 // Load the first vector as the initial set of results, and bail immediately
                 // to scalar handling if it contains any NaNs (which don't compare equally to themselves).
                 Vector512<float> resultVector = Vector512.LoadUnsafe(ref xRef, 0), current;
-                if (Vector512.EqualsAll(resultVector, resultVector))
+                if (!Vector512.EqualsAll(resultVector, resultVector))
                 {
-                    int oneVectorFromEnd = x.Length - Vector512<float>.Count;
-
-                    // Aggregate additional vectors into the result as long as there's at least one full vector left to process.
-                    i = Vector512<float>.Count;
-                    do
-                    {
-                        // Load the next vector, and early exit on NaN.
-                        current = Vector512.LoadUnsafe(ref xRef, (uint)i);
-                        if (!Vector512.EqualsAll(current, current))
-                        {
-                            goto Scalar;
-                        }
-
-                        resultVector = TMinMax.Invoke(resultVector, current);
-                        i += Vector512<float>.Count;
-                    }
-                    while (i <= oneVectorFromEnd);
-
-                    // If any elements remain, handle them in one final vector.
-                    if (i != x.Length)
-                    {
-                        current = Vector512.LoadUnsafe(ref xRef, (uint)(x.Length - Vector512<float>.Count));
-                        if (!Vector512.EqualsAll(current, current))
-                        {
-                            goto Scalar;
-                        }
-
-                        resultVector = TMinMax.Invoke(resultVector, current);
-                    }
-
-                    // Aggregate the lanes in the vector to create the final scalar result.
-                    return TMinMax.Invoke(resultVector);
+                    return GetFirstNaN(resultVector);
                 }
+
+                int oneVectorFromEnd = x.Length - Vector512<float>.Count;
+
+                // Aggregate additional vectors into the result as long as there's at least one full vector left to process.
+                i = Vector512<float>.Count;
+                do
+                {
+                    // Load the next vector, and early exit on NaN.
+                    current = Vector512.LoadUnsafe(ref xRef, (uint)i);
+                    if (!Vector512.EqualsAll(current, current))
+                    {
+                        return GetFirstNaN(current);
+                    }
+
+                    resultVector = TMinMax.Invoke(resultVector, current);
+                    i += Vector512<float>.Count;
+                }
+                while (i <= oneVectorFromEnd);
+
+                // If any elements remain, handle them in one final vector.
+                if (i != x.Length)
+                {
+                    current = Vector512.LoadUnsafe(ref xRef, (uint)(x.Length - Vector512<float>.Count));
+                    if (!Vector512.EqualsAll(current, current))
+                    {
+                        return GetFirstNaN(current);
+                    }
+
+                    resultVector = TMinMax.Invoke(resultVector, current);
+                }
+
+                // Aggregate the lanes in the vector to create the final scalar result.
+                return TMinMax.Invoke(resultVector);
             }
-            else
 #endif
+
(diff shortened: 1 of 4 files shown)


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1075,6 +1241,55 @@ private static Vector512<float> FusedMultiplyAdd(Vector512<float> x, Vector512<f
         }
 #endif
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static float HorizontalAggregate<TAggregate>(Vector128<float> x) where TAggregate : struct, IBinaryOperator
+        {
+            float f = x[0];
+            for (int i = 1; i < Vector128<float>.Count; i++)
```
EgorBo, line 1248: Vectorization via hadd is left for future improvements or it's not IEEE754 friendly?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1139,49 +1354,367 @@ public static Vector512<float> Invoke(Vector512<float> x, Vector512<float> y)
...
+            public static float Invoke(float result, float current) =>
+                current == result ?
+                    (IsNegative(result) ? current : result) :
+                    (current > result ? current : result);
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public static Vector128<float> Invoke(Vector128<float> result, Vector128<float> current) =>
+                Vector128.ConditionalSelect(Vector128.Equals(result, current),
+                    Vector128.ConditionalSelect(IsNegative(result), current, result),
+                    Vector128.Max(result, current));
```
EgorBo, line 1386: nit: afair, should be enough to use just `Vector128.Max` on arm as far as I remember, it handles -0.0/0.0

EgorBo (approved):
LGTM, but leaving an approval for @tannergooding who is an expert in IEEE754

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: hadd handles min/max?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -348,9 +344,179 @@ private static float CosineSimilarityCore(ReadOnlySpan<float> x, ReadOnlySpan<fl
...
+            float result = x[0];
+            int i = 0;
+
+#if NET8_0_OR_GREATER
+            if (Vector512.IsHardwareAccelerated && x.Length >= Vector512<float>.Count * 2)
+            {
+                ref float xRef = ref MemoryMarshal.GetReference(x);
+
+                // Load the first vector as the initial set of results, and bail immediately
+                // to scalar handling if it contains any NaNs (which don't compare equally to themselves).
```
tannergooding, line 373: Why does this need to bail to the scalar path?

If we find `NaN`, we return `NaN`. We then have the option of returning the exact `NaN` (best practice) or a generic `NaN` (allowed by the IEEE 754 spec).

So we could either do:
```csharp
if (!Vector512.EqualsAll(resultVector, resultVector))
{
    return float.NaN;
}
```

or, something like this (which follows the "best practice"):
```csharp
if (!Vector512.EqualsAll(resultVector, resultVector))
{
    Vector512<float> vectorMask = Vector512.Equals(resultVector, resultVector);
    ulong bitMask = vectorMask.ExtractMostSignificantBits();

    int index = BitOperations.TrailingZeroCount(bitMask);
    return resultVector[index];
}
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: It doesn't, but much like `Sum` is a specialized aggregate for addition, there are often optimized ways to handle aggregates for other operations as well.

The general pattern is typically:
```csharp
operation(vector.GetLower(), vector.GetUpper());
```

until you hit `V128`, and then you do:
```csharp
var tmp = Vector128.Shuffle(vector, ...); // swap even/odd indices
operation(vector, tmp);
// repeat up to log2(count) times, so 0 more times for 64-bit T, 1 more for 32-bit T, 2 more for 16-bit T, 3 more for 8-bit T
```

EgorBo: Oh sorry I was referring to Sum operation, although, I guess it should be possible to do Max horizontally as well?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: Right. `Arm64` has an `IEEE 754` compliant `min`, `max`, `minMagnitude`, and `maxMagnitude` instructions directly available.

x64 does not, but you can do optimized alternative sequences for some scenarios.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: It depends on the workload.

There are a lot of workloads that will avoid `NaN` entirely. There are other workloads that may use `NaN` as part of representing sparse or complex data.

I think it'd be fine to not handle it in the first vectorization pass here. But it'd probably be worth explicitly handling longer term.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: > I was referring to Sum operation

That's already handled here:
https://github.com/dotnet/runtime/blob/aec06846449c331532022b9c01874c80e5a35fc6/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs#L1087-L1091

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: > I think it'd be fine to not handle it in the first vectorization pass here. But it'd probably be worth explicitly handling longer term.

If it's worth doing longer term, I'll just do it now. Why put off for tomorrow, and all that.

stephentoub:
Feedback addressed:
- Used AdvSimd.Max/Min where applicable (we'll see how CI feels about it)
- Removed iteration from HorizontalAggregate
- Avoided scalar path for returning a found NaN

I rebased to pick up and merge with #92575

PR #92627, opened 2023-09-26
[release/8.0][wasm] Fix regressed file sizes for blazor

radical:
[wasm] build: Revert to older behavior for WasmNativeStrip

The earlier change was done in 26ae09784c16dfa6707cd73eca2aeb00a28741d0,
which changed to pass `-g` to the link step also. But that resulted in
increased native file sizes.

Changed sizes for the `minimum blazor template - publish` scenario:

---|run from rc2 HEAD (KiB)|With the change (KiB)|% change
--|--|--|--
SOD - Minimum Blazor Template - Publish|7918|7705|-2.69
Total Uncompressed _framework|4254|4104|-3.53
Aggregate - .js|540|437|-19.07
Aggregate - .wasm|3711|3663|-1.29
pub/wwwroot/_framework/dotnet.js|35|35|0.00
pub/wwwroot/_framework/dotnet.native.8.0.0-VERSION.js|234|131|-44.02
pub/wwwroot/_framework/dotnet.native.wasm|1170|1122|-4.10
pub/wwwroot/_framework/dotnet.runtime.8.0.0-VERSION.js|216|217|0.46

## Customer impact

Improved download size for the blazor app.

## Testing

Sizes confirmed with a `dotnet-runtime-perf` run for `blazor_scenarios`. Correctness verified by unit tests.

## Risk

Low. This is reverting to older behavior to change the symbols being included in the output.

### Details

- [wasm] Add getters for __indirect_function_table, and wasmMemory
- [wasm] cleanup corresponding tests
- [wasm] Remove WasmNativeStrip from wasm.proj as it will have no effect

Issue: https://github.com/dotnet/perf-autofiling-issues/issues/20642

### DIFF
--- a/src/mono/wasm/Wasm.Build.Tests/WasmNativeDefaultsTests.cs
+++ b/src/mono/wasm/Wasm.Build.Tests/WasmNativeDefaultsTests.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.IO;
+using System.Text.RegularExpressions;
 using Xunit;
 using Xunit.Abstractions;
 
@@ -12,6 +13,7 @@ namespace Wasm.Build.Tests
 {
     public class WasmNativeDefaultsTests : TestMainJsTestBase
     {
+        private static Regex s_regex = new("\\*\\* WasmBuildNative:.*");
         public WasmNativeDefaultsTests(ITestOutputHelper output, SharedBuildPerTestClassFixture buildContext)
             : base(output, buildContext)
         {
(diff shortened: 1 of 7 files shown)


pavelsavara:
> pub/wwwroot/_framework/dotnet.js 	35722.000 bytes 	35838.000 bytes

This is size change bit suspicious, because your changes should have no way how to influence that.
Most likely it means you are not comparing the same base commit. Do the other size changes have the same problem ?

radical:
> > pub/wwwroot/_framework/dotnet.js 	35722.000 bytes 	35838.000 bytes
> 
> This is size change bit suspicious, because your changes should have no way how to influence that. Most likely it means you are not comparing the same base commit. Do the other size changes have the same problem ?

Yeah, it's not comparing exactly the same base commit, because we don't have any runs for 8.0 or 8.0-rc2 branches, so I have to compare against the last run that we have from rc1.

Essentially, the difference is that we don't pass `-g` to link step now.

radical:
I can do a manual run on the current HEAD, and then we could compare.

radical:
@pavelsavara @lewing I have updated the results.

pavelsavara (approved):

radical:
@pavelsavara good catch, I fixed the percentages now.

lewing (approved):

radical:
@carlossanlop this is ready for merge.

PR #92635, opened 2023-09-26
JIT: Remove CallArgABIInformation::IsStruct

jakobbotsch:
Since we store signature types now this bit is no longer necessary.


jakobbotsch:
cc @dotnet/jit-contrib

[No diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=419101&view=ms.vss-build-web.run-extensions-tab)

BruceForstall (approved):

PR #92630, opened 2023-09-26
[wasm][debugger] Support multidimensional indexing of object scheme

ilonatommy:
Fixes https://github.com/dotnet/runtime/issues/76062.

Out of scope:
- indexing `valuetype` scheme https://github.com/dotnet/runtime/pull/92637

PR #92637, opened 2023-09-26
[wasm][debugger] Support indexing of valuetype scheme

ilonatommy:
Follow-up for https://github.com/dotnet/runtime/pull/92630.

Changes in tests: 
- We will be using similar mechanisms for struct like we do for class - refactored `TestEvaluate` class into `CommonCollections`, `ClassWithIndexers` and `StructWithIndexers` to avoid code duplication. `f` was an ambiguous name, now we have c for class and s for struct.
- We need a test case with both: indexers and inline arrays on the same object, so inline arrays from array tests got exposed and added to class/struct indexers.
- Added a check for invoking methods on inline arrays - works fine.
- Multidimensional indexing of value types.

Changes in logic:
- Added checking `LiteralExpressionSyntax` arguments for compatibility - this revealed a small overlooking in `CheckParameterCompatibility` switch case -  when the method expects `ElementType.Boolean` we cannot send string or char type as a param, it's not compatible.
- Unified indexing resolving mechanism for objects and value types, now both use `get_Items` method (`InvokeGetItemOnJObject`).
- Added receiving inline arrays on value types in the runtime, we used to read their sizes but then ignore reading the contents.


thaystg (approved):

src/mono/mono/component/debugger-agent.c
```diff
@@ -5353,14 +5353,15 @@ decode_vtype (MonoType *t, MonoDomain *domain, gpointer void_addr, gpointer void
...
+	int inlineArray = -1;
 
 	/* is_enum, ignored */
 	decode_byte (buf, &buf, limit);
 	if (CHECK_PROTOCOL_VERSION(2, 61))
 		decode_byte (buf, &buf, limit);
 	klass = decode_typeid (buf, &buf, limit, &d, &err);
 	if (CHECK_PROTOCOL_VERSION(2, 65))
-		decode_int (buf, &buf, limit); //ignore inline array
+		inlineArray = decode_int (buf, &buf, limit);
```
radical, line 5364: What is this value? Is it the size/number of elements in the inline array? Naming it accordingly would be good.

src/mono/mono/component/debugger-agent.c
```diff
@@ -5385,6 +5386,12 @@ decode_vtype (MonoType *t, MonoDomain *domain, gpointer void_addr, gpointer void
 		if (err != ERR_NONE)
 			return err;
 		nfields --;
+		if (CHECK_PROTOCOL_VERSION(2, 66) && inlineArray > 0)
```
radical, line 5389: This is checking for `inlineArray > 0`, but the `for` loop uses only values from index `[1]` onwards.
Why is the first one skipped? And the condition should be changed to be `> 1` too?

src/mono/mono/component/debugger-agent.c
ilonatommy: True, renaming to `inlineArraySize`

src/mono/mono/component/debugger-agent.c
ilonatommy: Changing to `i=0` does not break any tests. @thaystg, why?

thaystg: We are following the same thing that is done in the buffer_add_value_full.
The first element of the array is already sent here:
https://github.com/dotnet/runtime/blob/0c81f494f5664a462a26efdce0044875a22480f3/src/mono/mono/component/debugger-agent.c#L5269
Then we send the next the elements from 1 to array size.
And the same is done in the decode, on line 5385 we are decoding the first element, then we are decoding the next.

If you change it to i=0, and the first element has a value different of the second element you would see an error in the tests. I think!

src/mono/wasm/debugger/BrowserDebugProxy/MemberReferenceResolver.cs
```diff
@@ -408,6 +408,19 @@ async Task<JObject> ResolveAsInstanceMember(ArraySegment<string> parts, JObject
...
+                        {
+                            try
+                            {
+                                JObject vtResult = await InvokeGetItemOnJObject(rootObject, valueType.TypeId, objectId, elementIdxInfo, token);
+                                if (vtResult != null)
+                                    return vtResult;
+                            }
+                            catch (Exception)
+                            {
+                                logger.LogDebug($"Failed indexing '{objectId.Value}' with an indexer.");
```
radical, line 421: maybe ".. with a valuetype indexer."?

src/mono/wasm/debugger/BrowserDebugProxy/MemberReferenceResolver.cs
```diff
@@ -408,6 +408,19 @@ async Task<JObject> ResolveAsInstanceMember(ArraySegment<string> parts, JObject
...
                         var typeInfo = await context.SdbAgent.GetTypeInfo(valueType.TypeId, token);
+                        if (valueType.InlineArray == null)
+                        {
+                            try
+                            {
+                                JObject vtResult = await InvokeGetItemOnJObject(rootObject, valueType.TypeId, objectId, elementIdxInfo, token);
+                                if (vtResult != null)
+                                    return vtResult;
+                            }
+                            catch (Exception)
```
radical, line 419: What exception gets thrown here? I mean, what exactly fails, and when?

src/mono/wasm/debugger/BrowserDebugProxy/MemberReferenceResolver.cs
```diff
@@ -543,6 +533,43 @@ async Task<ElementIndexInfo> GetElementIndexInfo(List<JObject> nestedIndexers)
...
+            if (methodIds == null || methodIds.Length == 0)
+                throw new InvalidOperationException($"Type '{rootObject?["className"]?.Value<string>()}' cannot be indexed.");
+            var type = rootObject?["type"]?.Value<string>();
+
+            // ToDo: optimize the loop by choosing the right method at once without trying out them all
+            for (int i = 0; i < methodIds.Length; i++)
+            {
+                MethodInfoWithDebugInformation methodInfo = await context.SdbAgent.GetMethodInfo(methodIds[i], token);
+                ParameterInfo[] paramInfo = methodInfo.GetParametersInfo();
+                if (paramInfo.Length == elementIdxInfo.DimensionsCount)
```
radical, line 555: Invert the condition to `continue` early, saving on the nesting.

src/mono/wasm/debugger/BrowserDebugProxy/MemberReferenceResolver.cs
```diff
@@ -543,6 +533,43 @@ async Task<ElementIndexInfo> GetElementIndexInfo(List<JObject> nestedIndexers)
...
+                {
+                    try
+                    {
+                        if (!CheckParametersCompatibility(paramInfo, elementIdxInfo.Indexers))
+                            continue;
+                        ArraySegment<byte> buffer = await WriteIndexObjectAsIndices(objectId, elementIdxInfo.Indexers, paramInfo);
+                        JObject getItemRetObj = await context.SdbAgent.InvokeMethod(buffer, methodIds[i], token);
+                            return (JObject)getItemRetObj["value"];
```
radical, line 563: indent

src/mono/wasm/debugger/BrowserDebugProxy/MemberReferenceResolver.cs
```diff
@@ -543,6 +533,43 @@ async Task<ElementIndexInfo> GetElementIndexInfo(List<JObject> nestedIndexers)
...
+            DotnetObjectId objectId,
+            ElementIndexInfo elementIdxInfo,
+            CancellationToken token)
+        {
+            int[] methodIds = await context.SdbAgent.GetMethodIdsByName(typeId, "get_Item", BindingFlags.Default, token);
+            if (methodIds == null || methodIds.Length == 0)
+                throw new InvalidOperationException($"Type '{rootObject?["className"]?.Value<string>()}' cannot be indexed.");
+            var type = rootObject?["type"]?.Value<string>();
+
+            // ToDo: optimize the loop by choosing the right method at once without trying out them all
```
radical, line 550: What is needed, or missing to do this?

radical: Your idea here is the better approach.

src/mono/wasm/debugger/tests/debugger-test/debugger-evaluate-test.cs
```diff
@@ -508,58 +508,72 @@ public class EvaluateLocalsWithIndexingTests
...
-            public bool this[string key] => key.Length > 3;
-            public int this[double key] => (int)key;
-            public int this[float key] => (int)key;
-            public int this[decimal key] => (int)key;
-            public int this[Indexer indexer] => indexer.index;
-            public char this[char[] arr] => arr.Length == 0 ? '0' : arr[0];
-
-            public double this[int key1, double key2] => key1 + key2;
-            public string this[char key1, string key2, string key3] => $"{key1}-{key2}-{key3}";
+            public Dictionary<string, bool> indexedByStr = new Dictionary<string, bool>() { { "1", true }, { "111", false }, { "true", true} };
```
radical, line 521: note: be careful with changing existing test code, as it can be easy to break tests even when we think we changed the line numbers correctly, especially if they end up affecting line numbers used for breakpoints for instance. Such a change might not break a test immediately, but it might now be stopping on the incorrect line opening a potential for the test not doing the intended checks.

src/mono/wasm/debugger/BrowserDebugProxy/MemberReferenceResolver.cs
ilonatommy: Reason why it might be not possible:
1) If the parameter is required (as opposed to default) then we don't have information about its type. Even though we know what types were passed to the method, we cannot tell if this method will work well with them till we ask the runtime to `CMD_VM_INVOKE_METHOD` and it will throw / return a result.
2) Sometimes we even are not sure what type was passed to the method, see https://github.com/dotnet/runtime/issues/93057.
It would be good to try if something changed but please, not in this PR. I will see if I can make it in a separate one.

PR #92651, opened 2023-09-26
JIT: Handle remainder accesses more precisely in physical promotion liveness

jakobbotsch:
The liveness pass in physical promotion will currently handle any struct
LCL_FLD access of a physically promoted struct as accessing the
remainder. However, if the LCL_FLD only touches promoted fields then the
remainder is not actually used. There was a TODO around this which this
PR fixes as I stumbled upon a case this would improve.

### DIFF
--- a/src/coreclr/jit/jitstd/vector.h
+++ b/src/coreclr/jit/jitstd/vector.h
@@ -145,6 +145,7 @@ public:
 
     // cctors
     vector(const vector& vec);
+    vector(vector&& vec);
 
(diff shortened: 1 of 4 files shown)


jakobbotsch:

[Minor diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=419353&view=ms.vss-build-web.run-extensions-tab).

The linux x64 helix queue seems to be in a bad state, hence the timeouts...

AndyAyersMS (approved):

Issue #92633, opened 2023-09-26
`UnsafeAccessor` throws `MissingFieldException` on generic field (`T`)


AoshiW:
### Description

`UnsafeAccessor` throws `MissingFieldException` on generic field (`T`) in generic type,  where genetic `T` is:
- `class`  (e.g. `string`)
- `struct` that has `class` as a generic parameter  (e.g. `ArraySegment<string>`)

### Reproduction Steps

```cs
[UnsafeAccessor(UnsafeAccessorKind.Field, Name = "_items")]
static extern ref string[] GetItems(List<string> list);

GetItems(new());
```

### Expected behavior

getting field

### Actual behavior

throws exception
```
System.MissingFieldException: Field not found: 'System.Collections.Generic.List`1._items'.
```

### Configuration

.NET SDK: 8.0.100-rc.1.23455.8

jkotas:
cc @AaronRobinsonMSFT

AaronRobinsonMSFT:
Boo. For my original implementation of the signature match this worked. When I switched to the more robust, but stricter, built-in element compare this support was lost. Le sigh. There is little chance this sort of fix meets the servicing bar. I will take a little bit to try and make it work, but this is likely a .NET 9 fix.

AaronRobinsonMSFT:
Yep, this is a future fix. If the `T` is anywhere in the field or method signature this is going to fail.

Let me add some tests for this.

AaronRobinsonMSFT:
With how this is currently implemented there is some overlap with https://github.com/dotnet/runtime/issues/89439. The core issue being how we parse and compare signatures.

AaronRobinsonMSFT:
I've added disabled tests in https://github.com/dotnet/runtime/pull/92657 and fixed some issues with NAOT.

AaronRobinsonMSFT:
I'm going to close this as duplicate since this work will be included in https://github.com/dotnet/runtime/issues/89439. I've already got the static/instance fields working and have re-enabled the failing tests in my [working branch](https://github.com/AaronRobinsonMSFT/runtime/tree/generic_unsafe_accessors).

PR #92657, opened 2023-09-26
Add tests for `UnsafeAccessor` on fields on generic types

AaronRobinsonMSFT:
Add tests for the currently unsupported scenario described in https://github.com/dotnet/runtime/issues/92633.

Fix ILC to handle compiling `UnsafeAccessorsTests`.

The `UnsafeAccessorsTests` project now compiles and passes on CoreCLR and NAOT.

### DIFF
--- a/src/tests/baseservices/compilerservices/UnsafeAccessors/UnsafeAccessorsTests.cs
+++ b/src/tests/baseservices/compilerservices/UnsafeAccessors/UnsafeAccessorsTests.cs
@@ -43,7 +43,7 @@ static unsafe class UnsafeAccessorsTests
         private void _mvv() {}
 
         // The "init" is important to have here - custom modifier test.
-        // The signature of set_Prop is 
+        // The signature of set_Prop is
         // instance void modreq([System.Runtime]System.Runtime.CompilerServices.IsExternalInit) set_Prop ( string 'value')
         private string Prop { get; init; }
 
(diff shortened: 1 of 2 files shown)


src/tests/baseservices/compilerservices/UnsafeAccessors/UnsafeAccessorsTests.cs
```diff
@@ -188,6 +215,23 @@ public static void Verify_AccessFieldClass()
         extern static ref string GetPrivateField(UserDataClass d);
     }
 
+    [Fact]
+    [ActiveIssue("https://github.com/dotnet/runtime/issues/92633")]
```
jkotas, line 219: Do these tests fail on all runtimes or just on CoreCLR? If the tests fail on CoreCLR only, we can add `TestRuntimes.CoreCLR` filter to the ActiveIssue.

src/tests/baseservices/compilerservices/UnsafeAccessors/UnsafeAccessorsTests.cs
```diff
@@ -85,6 +85,33 @@ struct UserDataValue
         public string GetFieldValue() => _f;
     }
 
+    class UserDataGenericClass<T>
+    {
+        public const string StaticGenericFieldName = nameof(_GF);
+        public const string GenericFieldName = nameof(_gf);
+        public const string StaticGenericMethodName = nameof(_GM);
```
jkotas, line 92: Do you plan to add method tests to this PR too so that these strings and methods are used?

src/tests/baseservices/compilerservices/UnsafeAccessors/UnsafeAccessorsTests.cs
AaronRobinsonMSFT: They fail on all runtimes.

src/tests/baseservices/compilerservices/UnsafeAccessors/UnsafeAccessorsTests.cs
AaronRobinsonMSFT: These will be used when we enable more, not in this PR. I didn't use them right here as there is no need, but they represent the pattern. I didn't see any reason to not include them because that is what they will be. I'd prefer to leave them there instead of kicking off the CI again.

vitek-karas (approved):

PR #92660, opened 2023-09-26
Enable all interop generators if any are enabled

### LINKED ISSUE #92655: Interop source generators failing build in dotnet/runtime when latest SDK is installed
### Description

When building for older frameworks the latest interop source generator is used, which ends up forcing the latest shared library `Microsoft.Interop.SourceGeneration.dll` while still using older source generators.  This makes the older source generators break if there are breaking changes to the internal `Microsoft.Interop.SourceGeneration.dll`.

### Reproduction Steps

Build C:\src\dotnet\runtime\src\libraries\Common\tests\TestUtilities\TestUtilities.csproj

### Expected behavior

Builds successfully

### Actual behavior

Build fails with
> CSC : error CS8784: Generator 'VtableIndexStubGenerator' failed to initialize. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'TypeLoadException' with message 'Could not load type 'Microsoft.Interop.TargetFramework' from assembly 'Microsoft.Interop.SourceGeneration, Version=42.42.42.42, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.' 

### Regression?

Yes, 

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

I attempted to fix this in https://github.com/dotnet/runtime/pull/92648 before fully understanding the problem.

jtschuster:
It looks like we enable only a few of the live generators at a time when we should enable all if there are any that reference Microsoft.Interop.SourceGeneration

Fixes https://github.com/dotnet/runtime/issues/92655 on my machine

### DIFF
--- a/eng/generators.targets
+++ b/eng/generators.targets
@@ -56,13 +56,15 @@
     <ProjectReference Include="$(LibrariesProjectRoot)System.Runtime.InteropServices\gen\LibraryImportGenerator\LibraryImportGenerator.csproj"
                       ReferenceOutputAssembly="false"
                       OutputItemType="Analyzer"
-                      SetConfiguration="Configuration=$(LibrariesConfiguration)"
-                      Condition="@(EnabledGenerators->AnyHaveMetadataValue('Identity', 'LibraryImportGenerator'))" />
+                      SetConfiguration="Configuration=$(LibrariesConfiguration)" />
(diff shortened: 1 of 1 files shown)


jkoritzinsky:
I have an alternative fix in https://github.com/dotnet/runtime/pull/92661 that changes the build to not reference the live generators when building for a non-live TFM.

jkoritzinsky:
Closing as we merged #92661 instead.

PR #92661, opened 2023-09-26
Change how test assemblies opt-in to LibraryImportGenerator usage

### LINKED ISSUE #92655: Interop source generators failing build in dotnet/runtime when latest SDK is installed

jkoritzinsky:
Change test and test support assemblies to be auto-opted in to referencing LibraryImportGenerator instead of manually enabling it.

By changing this to an auto-opt-in, our TFM-based filtering logic kicks in and ensures that each project references the version of LibraryImportGenerator that corresponds to its TFM (as intended).

Fixes https://github.com/dotnet/runtime/issues/92655

### DIFF
--- a/eng/generators.targets
+++ b/eng/generators.targets
@@ -17,7 +17,8 @@
                        Condition="'$(EnableLibraryImportGenerator)' == '' and
                                   (
                                     '$(IsSourceProject)' == 'true' or
-                                    '$(IsTestProject)' == 'true'
+                                    '$(IsTestProject)' == 'true' or
+                                    '$(IsTestSupportProject)' == 'true'
                                   ) and
                                   '$(MSBuildProjectExtension)' == '.csproj' and
                                   (
--- a/eng/generators.targets
+++ b/eng/generators.targets
@@ -8,14 +8,17 @@
   <ItemGroup>
     <EnabledGenerators Include="LibraryImportGenerator" Condition="'$(EnableLibraryImportGenerator)' == 'true'" />
     <!-- If the current project is not System.Private.CoreLib, we enable the LibraryImportGenerator source generator
-         when the project is a C# source project that:
+         when the project is a C# source or test project that:
          - doesn't target the a TFM that includes LibraryImportGenerator or
          - doesn't reference the live targeting pack (i.e. when inbox) and
            - references System.Private.CoreLib, or
            - references System.Runtime.InteropServices -->
     <EnabledGenerators Include="LibraryImportGenerator"
                        Condition="'$(EnableLibraryImportGenerator)' == '' and
-                                  '$(IsSourceProject)' == 'true' and
+                                  (
+                                    '$(IsSourceProject)' == 'true' or
+                                    '$(IsTestProject)' == 'true'
+                                  ) and
                                   '$(MSBuildProjectExtension)' == '.csproj' and
                                   (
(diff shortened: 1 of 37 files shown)


eng/generators.targets
```diff
@@ -8,14 +8,18 @@
...
          - doesn't reference the live targeting pack (i.e. when inbox) and
            - references System.Private.CoreLib, or
            - references System.Runtime.InteropServices -->
     <EnabledGenerators Include="LibraryImportGenerator"
                        Condition="'$(EnableLibraryImportGenerator)' == '' and
-                                  '$(IsSourceProject)' == 'true' and
+                                  (
+                                    '$(IsSourceProject)' == 'true' or
+                                    '$(IsTestProject)' == 'true' or
+                                    '$(IsTestSupportProject)' == 'true'
```
ViktorHofer, line 21: This is either true for projects that target a non .NETCoreApp TFM or ones that disable implicit framework references and then reference System.Runtime.InteropServices or CoreLib. So essentially this logic only applies to our .NET Framework TFMs in test projects.

Is that intention? Just want to better understand the change.

jkoritzinsky: Yes, this should only be true for .NET Framework or Standard TFMs in tests.

ViktorHofer (approved):

ericstj (approved):
I pulled this down and confirm it fixes https://github.com/dotnet/runtime/issues/92655.  

> By changing this to an auto-opt-in, our TFM-based filtering logic kicks in and ensures that each project references the version of LibraryImportGenerator that corresponds to its TFM (as intended).

Confirmed that this is happening.  Indeed, this is much better behavior.  I'm glad we're not depending on using the newer source generator when targeting older frameworks which carry it.

Thank you @jkoritzinsky

jtschuster (approved):

jkoritzinsky:
Wasm failures are specific to this PR. Need to figure out what's causing it.

jkoritzinsky:
Okay for some reason I can't find any other CI run that has these particular jobs running. I'm going to file a known issue.

radical:
> Wasm failures are specific to this PR. Need to figure out what's causing it.

This fixes the issue - [`c40b2dc` (#92741)](https://github.com/dotnet/runtime/pull/92741/commits/c40b2dc683faea95b846db588763aea025adade3) which got introduced here.

PR #92666, opened 2023-09-26
Make it possible to preinitialize HW intrinsic IsSupported

MichalStrehovsky:
* Move the IL rewriting for HW intrinsice `IsSuported` calls to `ILProvider` from `RyuJitCompilation`
* Also rewrite constant true/false

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler/Program.cs
+++ b/src/coreclr/tools/aot/ILCompiler/Program.cs
@@ -20,6 +20,7 @@ using Internal.TypeSystem;
 using Internal.TypeSystem.Ecma;
 
 using ILCompiler.Dataflow;
+using ILCompiler.DependencyAnalysis;
 using ILLink.Shared;
 
 using Debug = System.Diagnostics.Debug;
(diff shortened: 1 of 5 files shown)


src/coreclr/tools/aot/ILCompiler/Program.cs
```diff
@@ -143,6 +144,11 @@ public int Run()
             if (typeSystemContext.InputFilePaths.Count == 0)
                 throw new CommandLineException("No input files specified");
 
+            ilProvider = new HardwareIntrinsicILProvider(
+                instructionSetSupport,
+                new ExternSymbolMappedField(typeSystemContext.GetWellKnownType(WellKnownType.Int32), "g_cpuFeatures"),
```
jkotas, line 149: Nit: It looks a bit odd to see this low-level implementation detail in the program Main. I think it would be ok for `HardwareIntrinsicILProvider` to have the `g_cpuFeatures` field name hardcoded. It has the bit masks hardcoded, so it can have the field name hardcoded too.

jkotas (approved):
Thanks

src/coreclr/tools/aot/ILCompiler/Program.cs
MichalStrehovsky: It's actually making things consistent with `g_compilerEmbeddedSettingsBlob` and `g_compilerEmbeddedKnobsBlob` a couple dozen lines below.

PR #92667, opened 2023-09-26
[wasm] GetChromeVersions: Fix fetching v8 version given a chrome version

radical:


src/tasks/WasmBuildTasks/GetChromeVersions.cs
```diff
@@ -9,14 +9,15 @@
...
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Microsoft.Build.Framework;
 
 using MBU = Microsoft.Build.Utilities;
 
 namespace Microsoft.WebAssembly.Build.Tasks;
 
-public class GetChromeVersions : MBU.Task
+public partial class GetChromeVersions : MBU.Task
```
ilonatommy, line 20: Where is the other part of this partial class?
Edit: okay, I see it's a requirement of `GeneratedRegexAttribute` to have it partial.

src/tasks/WasmBuildTasks/GetChromeVersions.cs
```diff
@@ -192,9 +196,45 @@ private async Task<(ChromeVersionSpec version, string baseSnapshotUrl)> FindVers
                                             foundRelease.version,
                                             foundRelease.chromium_main_branch_position.ToString(),
                                             foundRelease.version);
+        string foundV8Version = await FindV8VersionFromChromeVersion(foundRelease.version).ConfigureAwait(false);
         string? baseSnapshotUrl = await FindSnapshotUrlFromBasePositionAsync(versionSpec, throwIfNotFound: true)
                                         .ConfigureAwait(false);
+
+        versionSpec = versionSpec with { v8_version = foundV8Version };
```
ilonatommy, line 203: That's fine. Why don't we change the order of lines and put v8 version info into the record at once, with all the other values? But only if you will be making other serious changes, it's a nit.

ilonatommy (approved):

src/tasks/WasmBuildTasks/GetChromeVersions.cs
radical: Good catch! Fixed.

PR #92668, opened 2023-09-26
Remove all PGO assets except for the runtime PGO archive.

jkoritzinsky:
Removes duplicate PGO assets.


hoyosjs (approved):

steveisok:
@jkoritzinsky I think this should be backported to 8 as well.

PR #92675, opened 2023-09-26
Convert SpinWait to QCall

jkotas:


src/coreclr/vm/comsynchronizable.cpp
```diff
@@ -1044,29 +1044,8 @@ FCIMPL1(void, ThreadNative::SpinWait, int iterations)
...
 
-    //
-    // If we're not going to spin for long, it's ok to remain in cooperative mode.
-    // The threshold is determined by the cost of entering preemptive mode; if we're
-    // spinning for less than that number of cycles, then switching to preemptive
-    // mode won't help a GC start any faster.
-    //
-    if (iterations <= 100000)
-    {
-        YieldProcessorNormalized(iterations);
```
jkotas, line 1055: This path was missing a GC probe that makes it unfriendly to GC suspension. It has proper GC probe now thanks to SuppressGCTransition.

EgorBo: Do I understand it correctly that a non-interruptable loop with potentially 1023 iterations is not an issue for gc starving?

EgorBo (approved):
Ah, interesting trick with invoking the same method but with different gc mode

src/coreclr/vm/comsynchronizable.cpp
jkotas: Right, our does say that starving the GC suspension for 1 microsecond is ok: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.suppressgctransitionattribute

jkotas: (1023 iterations of pause may be more than that, but it does not really matter exactly. It just needs to be in the right ballpark.)

PR #91229, opened 2023-08-28
Add new Pipeline for Running Libs Tests with TestReadyToRun

ivdiazsa:
Follows issue #85417. This is a follow up to PR #80946, which added the flag `-p:TestReadyToRun` for running libraries tests with a crossgen'd runtime. This PR takes this new functionality and creates a new pipeline to have a way of regularly and consistently testing the libraries using _ReadyToRun_, extending our R2R/Crossgen2 coverage.

PR #88991, opened 2023-07-17
Update targetingpacks.targets

ViktorHofer:
The .NET 8 Preview 6 SDK has the features required to simplify the targetingpacks.targets logic.

PR #92677, opened 2023-09-26
Use a different crossgen2 when running crossgen2 during our build than the crossgen2 that we are shipping

jkoritzinsky:
Instead of using the same crossgen2 project when we are building the product as we are shipping, repurpose the _crossarch copy of the build to be used in all scenarios. This avoids a race condition in our build where building the runtime and tool packs can fail because Microsoft.NETCore.App.Runtime is trying to R2R all assemblies while Microsoft.NETCore.App.Crossgen2 is re-publishing crossgen2 as a single-file self-contained app.

We already do this for our cross-builds. This just brings the non-cross builds to use the same pattern to remove the race condition.

This is required for the VMR work to put the whole product build into a single job. For whatever reason, that configuration hits the race condition very consistently.

### DIFF
--- a/eng/targetingpacks.targets
+++ b/eng/targetingpacks.targets
@@ -35,7 +35,7 @@
                              LatestRuntimeFrameworkVersion="$(ProductVersion)"
                              RuntimeFrameworkName="$(LocalFrameworkOverrideName)"
                              RuntimePackNamePatterns="$(LocalFrameworkOverrideName).Runtime.**RID**"
-                             RuntimePackRuntimeIdentifiers="linux-arm;linux-arm64;linux-musl-arm64;linux-musl-x64;linux-x64;osx-x64;rhel.6-x64;tizen.4.0.0-armel;tizen.5.0.0-armel;win-arm64;win-x64;win-x86;linux-musl-arm;osx-arm64;maccatalyst-x64;maccatalyst-arm64;linux-s390x;linux-bionic-arm;linux-bionic-arm64;linux-bionic-x64;linux-bionic-x86"
+                             RuntimePackRuntimeIdentifiers="linux-arm;linux-arm64;linux-musl-arm64;linux-musl-x64;linux-x64;osx-x64;rhel.6-x64;tizen.4.0.0-armel;tizen.5.0.0-armel;win-arm64;win-x64;win-x86;linux-musl-arm;osx-arm64;maccatalyst-x64;maccatalyst-arm64;linux-s390x;linux-bionic-arm;linux-bionic-arm64;linux-bionic-x64;linux-bionic-x86;freebsd-x64;freebsd-arm64"
                              TargetFramework="$(NetCoreAppCurrent)"
                              TargetingPackName="$(LocalFrameworkOverrideName).Ref"
                              TargetingPackVersion="$(ProductVersion)"
--- a/eng/targetingpacks.targets
+++ b/eng/targetingpacks.targets
@@ -112,6 +112,10 @@
       <PackageReference Remove="@(PackageReference)"
                         Condition="'$(UsePackageDownload)' != 'true' and $([System.String]::Copy('%(Identity)').StartsWith('$(LocalFrameworkOverrideName).Runtime'))" />
       <PackageDownload Remove="@(PackageDownload)"
+                       Condition="'$(UsePackageDownload)' == 'true' and $([System.String]::Copy('%(Identity)').StartsWith('$(LocalFrameworkOverrideName).Crossgen2'))" />
+      <PackageReference Remove="@(PackageReference)"
+                        Condition="'$(UsePackageDownload)' != 'true' and $([System.String]::Copy('%(Identity)').StartsWith('$(LocalFrameworkOverrideName).Crossgen2'))" />
+      <PackageDownload Remove="@(PackageDownload)"
                        Condition="'$(UsePackageDownload)' == 'true' and $([System.String]::Copy('%(Identity)').EndsWith('Microsoft.DotNet.ILCompiler'))" />
       <!-- Always remove the PackageReference items as some of the packages are only via PackageReference -->
       <PackageReference Remove="@(PackageReference)"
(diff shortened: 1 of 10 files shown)


jkoritzinsky:
> Sounds like a sequencing problem. Why not declare a dependency on whatever state is needed for Microsoft.NETCore.App.Runtime's R2R operation?

Both Microsoft.NETCore.App.Runtime and Microsoft.NETCore.App.Crossgen2 both try to use the same project, but they build it with different global properties. We could try to hack something together to sequence them, but going this path makes us more consistent across platforms and avoids adding additional layers of complicated infra to fix complicated infra instead of just building something simple.

jkoritzinsky:
WBT failures are https://github.com/dotnet/runtime/issues/92732#event-10493266156

ViktorHofer:
> Both Microsoft.NETCore.App.Runtime and Microsoft.NETCore.App.Crossgen2 both try to use the same project, but they build it with different global properties

Building the same project with different global properties usually results in binclashes and worst case, race conditions. Is that still the case with these changes?

jkoritzinsky:
There is still technically a race condition between the initial build of `crossgen2.csproj` and `Microsoft.NETCore.App.Crossgen2.sfxproj`, but due to how we order subsets, they would never be built in parallel. At some point we should fix the crossgen2 build to not have the race condition at all, but that fix will be much more invasive and require more design work.

agocke:
Rather than have the "plain" project have a different name, should we just take the existing project, copy and rename it to "publish", and then use the "inbuild" as the regular project?

I think we would be able to delete a lot of the weird conditions we have in publishing just to deal with the alternate global settings.

jkoritzinsky:
I can try that out.

jkoritzinsky:
I've updated this PR to try the new approach. There's now 3 variants of the crossgen build, each for a different scenario. They all build the same sources, but there's some minor differences in their builds for their particular use cases.

- crossgen2_inbuild.csproj
  - Built as a framework-dependent executable. Used during the build process on the host machine.
  - Uses the cross-component jit builds.
  - Targets NetCoreAppToolCurrent as it runs with the LKG SDK.
- crossgen2.csproj
  - Built as a framework-dependent executable. Used on Helix machines for ReadyToRun testing
  - Uses the regular (target) jit builds.
  - Targets NetCoreAppToolCurrent as it is built before we have built the ref-pack.
- crossgen2_publish.csproj
  - Referenced by Microsoft.NETCore.App.Crossgen2.sfxproj. Contains all of the publishing logic. Is only ever published, never just built.
  - Uses the regular (target) jit builds.
  - Targets NetCoreAppCurrent to automatically pick up the live-built Ref and Runtime packs and integrate with the rest of the repo's support for targeting the live build.

We can likely get rid of `crossgen2.csproj` and use `crossgen2_publish.csproj` in its place, but that would require significant build work to reorganize the build to publish a crossgen2 build with the live runtime. I think this work is worthwhile, but we should simplify the test build and the AzDO pipelines before we do this work (as that is the majority of the cost here).

eng/targetingpacks.targets
```diff
@@ -111,6 +111,10 @@
                        Condition="'$(UsePackageDownload)' == 'true' and $([System.String]::Copy('%(Identity)').StartsWith('$(LocalFrameworkOverrideName).Runtime'))" />
       <PackageReference Remove="@(PackageReference)"
                         Condition="'$(UsePackageDownload)' != 'true' and $([System.String]::Copy('%(Identity)').StartsWith('$(LocalFrameworkOverrideName).Runtime'))" />
+      <PackageDownload Remove="@(PackageDownload)"
```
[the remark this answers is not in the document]
jkoritzinsky: I'd rather do that in a follow-up PR. There's a lot of cleanup that we can do in this file and I'd rather do it all at once in a focused effort. This PR is blocking the "coreclr single job" work I'm doing, so I want to keep this focused.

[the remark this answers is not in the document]
ViktorHofer: I attempted that months ago: https://github.com/dotnet/runtime/pull/88991. I initially implemented the `EnableRuntimePackDownload` switch incorrectly in the SDK which I then fixed during .NET 8.

Based on the PR failure, the DNNE.targets require some of the pack still to be downloaded. We probably need to add a PackageDownload item for the test using it.

eng/targetingpacks.targets
```diff
@@ -111,6 +111,10 @@
...
+                       Condition="'$(UsePackageDownload)' == 'true' and $([System.String]::Copy('%(Identity)').StartsWith('$(LocalFrameworkOverrideName).Crossgen2'))" />
+      <PackageReference Remove="@(PackageReference)"
+                        Condition="'$(UsePackageDownload)' != 'true' and $([System.String]::Copy('%(Identity)').StartsWith('$(LocalFrameworkOverrideName).Crossgen2'))" />
```
ViktorHofer, line 117: cc @ivdiazsa (https://github.com/dotnet/runtime/pull/91229)

src/installer/pkg/sfx/Microsoft.NETCore.App/Microsoft.NETCore.App.Crossgen2.sfxproj
```diff
@@ -28,36 +28,22 @@
...
 
+  <ItemGroup>
+    <ProjectReference
+      Include="$(RepoRoot)src/coreclr/tools/aot/crossgen2/crossgen2_publish.csproj"
+      AdditionalProperties="_IsPublishing=true
+        ;RuntimeIdentifier=$(PackageRID)
+        ;NativeAotSupported=$(NativeAotSupported)
+        ;CoreCLRArtifactsPath=$(CoreCLRArtifactsPath)
+        ;ObjCopyName=$(ObjCopyName)
+        ;R2ROverridePath=$(MSBuildThisFileDirectory)ReadyToRun.targets"
```
ViktorHofer, line 39: Do we still need to pass these properties in? Can't we just set them directly in crossgen2_publish.csproj now that that project file is only responsible for publishing?

agocke (approved):
Looks great! We could probably add a “builddependsonpublish” target in the publish project since we now think it’s only to be used for publishing.

src/installer/pkg/sfx/Microsoft.NETCore.App/Microsoft.NETCore.App.Crossgen2.sfxproj
jkoritzinsky: We can probably inline some of them, but I don't think we can inline all of them. I think `_IsPublishing` needs to be set as a global property. I'll look at that in a follow-up PR.

jkoritzinsky:
> Looks great! We could probably add a “builddependsonpublish” target in the publish project since we now think it’s only to be used for publishing.

I'll do some cleanup after I merge this in to make it a little more self-descriptive. Since this is blocking the "live illink" PR as well, I'm going to merge it in now.

PR #90283, opened 2023-08-09
Fixing COM Aggregation and XAML reference tracking on .NET AOT

manodasanW:
There were blocks put in place in AOT for COM aggregation and XAML reference tracking wasn't implemented.  This PR adds support for both of them by porting the CoreCLR native code that existed to C# for AOT support.  With these changes, the ComWrappers API tests and weak reference tests all pass.

Fixes #89713
Fixes #85137

PR #92679, opened 2023-09-26
Define bool as Interop.BOOL to prevent upper bytes setting native bool

yowl:
This PR fixes an issue with NativeAOT-LLVM and possibly other native AOT targets where the bool passed to `RhpCollect` was not passed as `Interop.BOOL` leading to the upper bytes being observed as part of the native bool.

cc @SingleAccretion 
cc @AaronRobinsonMSFT 
cc @jkotas

### DIFF
--- a/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/InternalCalls.cs
+++ b/src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/InternalCalls.cs
@@ -62,12 +62,12 @@ namespace System.Runtime
         [RuntimeExport("RhCollect")]
         internal static void RhCollect(int generation, InternalGCCollectionMode mode, bool lowMemoryP = false)
         {
-            RhpCollect(generation, mode, lowMemoryP);
+            RhpCollect(generation, mode, lowMemoryP ? Interop.BOOL.TRUE : Interop.BOOL.FALSE);
         }
 
         [DllImport(Redhawk.BaseName)]
         [UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvCdecl) })]
-        private static extern void RhpCollect(int generation, InternalGCCollectionMode mode, bool lowMemoryP);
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/Runtime.Base/src/System/Runtime/InternalCalls.cs
```diff
@@ -62,12 +62,12 @@ internal static class InternalCalls
...
         internal static void RhCollect(int generation, InternalGCCollectionMode mode, bool lowMemoryP = false)
         {
-            RhpCollect(generation, mode, lowMemoryP);
+            RhpCollect(generation, mode, lowMemoryP ? Interop.BOOL.TRUE : Interop.BOOL.FALSE);
         }
 
         [DllImport(Redhawk.BaseName)]
         [UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvCdecl) })]
-        private static extern void RhpCollect(int generation, InternalGCCollectionMode mode, bool lowMemoryP);
+        private static extern void RhpCollect(int generation, InternalGCCollectionMode mode, Interop.BOOL lowMemoryP);
```
AaronRobinsonMSFT, line 70: The default marshalling behavior in CoreCLR is to marshal that as an `int`, but I don't know about NAOT. Regardless of that, this will make the signature blittable which means inlinable and easier to work with.

Thanks!

AaronRobinsonMSFT (approved):

AaronRobinsonMSFT:
@jkotas I signed off because from an interop perspective this is all goodness. Is NAOT different in this regard? Given @yowl statement it sounds like NAOT might be different?

jkotas:
NAOT should not be any different. I think the problem is that the the CoreLib is marked with `DisableRuntimeMarshalling`. It means that `bool` in DllImport is going to byte-sized bool without any marshalling. An alternative way to fix this would be to turn the P/Invoke into LibraryImport.

jkotas:
This is bug introduced by #90283. We should backport it to release/8.0.

jkotas (approved):

PR #92680, opened 2023-09-27
Add plain Vector128 support to IndexOfAnyAsciiSearcher

MihaZupan:
Contributes to https://github.com/dotnet/perf-autofiling-issues/issues/21818

I believe this should be an okay implementation correctness-wise, but it's up to Mono implementations how efficiently they can handle the `Vector128` paths (e.g. efficiently handing `Vector128.Shuffle` with non-const indices).

@fanyang-mono could you please check how this impacts Mono perf?
We shouldn't merge this without good validation as it will affect most uses of `SearchValues`.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/SearchValues/ProbabilisticWithAsciiCharSearchValues.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/SearchValues/ProbabilisticWithAsciiCharSearchValues.cs
@@ -20,7 +20,7 @@ namespace System.Buffers
 
         public ProbabilisticWithAsciiCharSearchValues(scoped ReadOnlySpan<char> values)
         {
-            Debug.Assert(IndexOfAnyAsciiSearcher.IsVectorizationSupported);
+            Debug.Assert(Vector128.IsHardwareAccelerated);
             Debug.Assert(values.ContainsAnyInRange((char)0, (char)127));
 
             IndexOfAnyAsciiSearcher.ComputeAsciiState(values, out _asciiState);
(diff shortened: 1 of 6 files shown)


fanyang-mono:
For initial validation, please run affected microbenchmarks from https://github.com/dotnet/perf-autofiling-issues/issues/21818 using this script https://github.com/dotnet/performance/blob/main/scripts/benchmarks_local.py

You should run them against both Mono JIT and Mono interpreter. It could be achieved by passing `MonoJIT` or `MonoInterpreter` to the switch `--run-type`

MihaZupan:
@MihuBot

src/libraries/System.Private.CoreLib/src/System/SearchValues/ProbabilisticWithAsciiCharSearchValues.cs
```diff
@@ -42,7 +42,7 @@ internal override int IndexOfAny(ReadOnlySpan<char> span)
 
             // We check whether the first character is ASCII before calling into IndexOfAnyAsciiSearcher
             // in order to minimize the overhead this fast-path has on non-ASCII texts.
-            if (IndexOfAnyAsciiSearcher.IsVectorizationSupported && span.Length >= Vector128<short>.Count && char.IsAscii(span[0]))
```
tannergooding, line 45: I think we need to keep `IsSupported` checks to ensure trimming works as expected on platforms without SIMD support.

MihaZupan: Is that needed even though all its ctor uses are guarded behind it?

tannergooding: R2R has some special rules for various ISA checks and I'm pretty sure we expect them to be used even when all callers are expected to hit a given path.

CC. @davidwrighton

Issue #92624, opened 2023-09-26
new FileStream with too-large preallocation hint does not throw on Windows


danmoseley:
This test assumes creating a 1TB file will fail -- this is why the first run test fails here, because I have 1.5TB free. edit: turns out it fails anyway on Windows

Note the 2nd flavor of the test passes here -- perhaps because the first file wasn't deleted (?) but I would expect dispose to have happened between them and cleaned up the test directory. edit: it's because the finalizer hadn't run before dispose.

```xml
      <test name="System.IO.Tests.FileStream_ctor_options.WhenDiskIsFullTheErrorMessageContainsAllDetails(mode: Create)" type="System.IO.Tests.FileStream_ctor_options" method="WhenDiskIsFullTheErrorMessageContainsAllDetails" time="0.0389214" result="Fail">
        <traits>
          <trait name="category" value="outerloop" />
        </traits>
        <failure exception-type="Xunit.Sdk.ThrowsException">
          <message><![CDATA[Assert.Throws() Failure\r\nExpected: typeof(System.IO.IOException)\r\nActual:   (No exception was thrown)]]></message>
          <stack-trace><![CDATA[   at System.IO.Tests.FileStream_ctor_options.WhenDiskIsFullTheErrorMessageContainsAllDetails(FileMode mode) in C:\git\runtime\src\libraries\System.IO.FileSystem\tests\FileStream\ctor_options.cs:line 169
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr) in C:\git\runtime\src\libraries\System.Private.CoreLib\src\System\Reflection\MethodBaseInvoker.cs:line 157]]></stack-trace>
        </failure>
      </test>
      <test name="System.IO.Tests.FileStream_ctor_options.WhenDiskIsFullTheErrorMessageContainsAllDetails(mode: CreateNew)" type="System.IO.Tests.FileStream_ctor_options" method="WhenDiskIsFullTheErrorMessageContainsAllDetails" time="0.0028693" result="Pass">
        <traits>
          <trait name="category" value="outerloop" />
        </traits>
      </test>
```

It should try to create a 100TB file or something, and we can meet back here in 2028. 😄

danmoseley:
[...]

PR #92682, opened 2023-09-27
Improve disk full test

danmoseley:
Relates to https://github.com/dotnet/runtime/issues/92624

1. Ensure that if the file is somehow created, it is deleted by the end of the test (in the base class dispose) rather than only if the finalizer has run first.
2. Change to creating 2^^60 byte file rather than 2^^40 to make sure it isn't created. (I'm sure someone will read this in 10 years and laugh)
3. Disable for Windows because it doesn't throw at present. Tracking with above.
4. Remove OuterLoop as there's no reason for it to be: it won't create the file.

On my Linux box the tests fail within a few 100us - the size doesn't matter.

### DIFF
--- a/src/libraries/System.IO.FileSystem/tests/FileStream/ctor_options.cs
+++ b/src/libraries/System.IO.FileSystem/tests/FileStream/ctor_options.cs
@@ -152,7 +152,7 @@ namespace System.IO.Tests
             }
         }
 
-        [OuterLoop("Might allocate 1 TB file if there is enough space on the disk")]
+        [ActiveIssue("https://github.com/dotnet/runtime/issues/92624", TestPlatforms.Windows)]
         // macOS fcntl doc does not mention ENOSPC error: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fcntl.2.html
         // But depending on the OS version, it might actually return it.
         // Since we don't want to have unstable tests, it's better to not run it on macOS at all.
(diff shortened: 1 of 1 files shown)


src/libraries/System.IO.FileSystem/tests/FileStream/ctor_options.cs
```diff
@@ -162,21 +162,23 @@ public void PreallocationSize(FileAccess access, FileMode mode)
...
-
-            // ensure it was NOT created
-            bool exists = File.Exists(filePath);
-            if (exists)
+            try
             {
-                File.Delete(filePath);
+                // not using Assert.Throws because in the event of failure, we want to dispose, so the next test isn't affected
+                using FileStream fs = CreateFileStream(filePath, mode, FileAccess.Write, FileShare.None, bufferSize: 1, FileOptions.None, tooMuch);
+                Assert.Fail($"Expected to throw IOException, {fs.Length}");
```
krwq, line 172:
```suggestion
                Assert.Throws(() =>
                {
                    using FileStream fs = CreateFileStream(filePath, mode, FileAccess.Write, FileShare.None, bufferSize: 1, FileOptions.None, tooMuch);
                });
```

won't this work though?

krwq (approved):

src/libraries/System.IO.FileSystem/tests/FileStream/ctor_options.cs
```diff
@@ -162,21 +162,25 @@ public void PreallocationSize(FileAccess access, FileMode mode)
...
-            Assert.Contains(filePath, ex.Message);
-            Assert.Contains(tooMuch.ToString(), ex.Message);
-
-            // ensure it was NOT created
-            bool exists = File.Exists(filePath);
-            if (exists)
+            try
+            {
+                // not using Assert.Throws because in the event of failure, we want to dispose, so the next test isn't affected
+                Assert.Throws(() =>
```
Jozkee, line 171:
```suggestion
                Assert.Throws<IOException>(() =>
```
This would fix the build. You should also extract the other asserts from the try-catch.

I would consider reverting to the previous commit instead 😄.

jozkee (changes requested):

src/libraries/System.IO.FileSystem/tests/FileStream/ctor_options.cs
danmoseley: reverted to what it was originally as I think it's fine. can you signoff @Jozkee ?

danmoseley:
created known issue https://github.com/dotnet/runtime/issues/93323

jozkee (approved):

src/libraries/System.IO.FileSystem/tests/FileStream/ctor_options.cs
stephentoub, line 172: Why not use Assert.Throws as @krwq suggested? e.g.
```C#
IOException ex = Assert.Throws<IOException>(() => CreateFileStream(filePath, mode, FileAccess.Write, FileShare.None, bufferSize: 1, FileOptions.None, tooMuch));
Assert.Contains(filePath, ex.Message);
Assert.Contains(tooMuch.ToString(), ex.Message);
Assert.False(File.Exists(filePath));
```
?

danmoseley: If the test fails because the file is created, there is a race between the finalizer releasing the lock on it and the base class attempting to delete the file.

Now it's creating a colossal file, that's implausible unless there's some partial create mode in some future test configuration. But this way ensures the stream is disposed in that case.

In general we make reasonable attempts to not allow one test failure to trigger another, although of course that is not always feasible. That reduces noise in CI.

danmoseley: That's my recollection of why I did this - what am I missing? But happy to change as you propose if you wish.

Jozkee: IOW: The file created here will very likely occupy the whole disk in most configs. We need to dispose it, _even if this test fails_, to avoid causing out-of-disk errors to other tests.

PR #88488, opened 2023-07-06
fix IsMutuallyAuthenticated with NegotiateClientCertificateAsync

wfurt:
contributes to  #65563 and  #75545.

The problem is that with existing code we clean `_selectedClientCertificate` if the selected certificate was attached to credentials but not used. So later it can be provided via renegotiation or PHA for TLS 1.3 but we already lost the link.

The fix is to keep  `_selectedClientCertificate` untouched once client certificate was selected and used but guard `LocalServerCertificate` with the extra check that was added by  #79898.

PR #79128, opened 2022-12-02
fix SslStream.IsMutuallyAuthenticated with cached credentials

wfurt:
fixes #65563

We can query Schannel after handshake is done to see if certificate was used. This is no-op for test of the platforms at the moment.
The test is essentially repro @rzikm put together during investigation. 

not 100% about android. cc: @simonrozsival

PR #68009, opened 2022-04-14
re-enable TLS 13 tests on Windows

wfurt:
While the underlying issue lives in Schannel and it is being fixed in Windows the root analyses points to certificates with CAPI provider. That works fine for previous TLS versions but it fails for TLS 1.3 as that is somehow different. 

This change brings new pfx files from https://github.com/dotnet/runtime-assets/pull/234
With updated data (essentially still same certificate) everything seems to be stable. 

This is only fix for the 
```
---- System.ComponentModel.Win32Exception : The Local Security Authority cannot be contacted
```

other (recent server 2022) failures are unrelated to this.

fixes #58927

PR #63945, opened 2022-01-18
Fix IsMutuallyAuthenticated on Linux and OSX

rzikm:
fixes #40314

This PR builds on top https://github.com/dotnet/runtime/pull/63200. It utilizes newly introduced certificate callbacks/breaks to signal to SecureChannel that credentials refresh is needed. Thus the connection establishment uses the same workflow as on Windows (Start as anonymous, refresh credentials if requested).

I have noticed another similar bug (at least it seems like a bug to me) with `IsMutuallyAuthenticated`: If Client first connects to a server which requires client cert, and later, it connects to a server which does not require client cert, the second time the `IsMutuallyAuthenticated` property is still true on client side. This bug is reproducible on all platforms AFAICT by flipping the `InlineData` attributes on `SslStreamCredentialCacheTest.SslStream_SameCertUsedForClientAndServer_Ok` test --> To be reported separately.

PR #92684, opened 2023-09-27
[release/6.0] fix IsMutuallyAuthenticated on SslStream

wfurt:
This is backport of PR #88488 and PR #79128 and parts of PR #63945. 
It also brings spirit of test-only PR #68009 to get test coverage for TLS 1.3.  

This only covers Windows to minimize the code delta i.e. it does not bring all the changes from PR #63945 to cover Linux & macOS.

## Customer Impact

The property `IsMutuallyAuthenticated` on `SslStream` indicates if mutual TLS authentication is performed with client certificate. Current 6.0 implementation can get confused in several cases, so the value is unreliable for security audits.

## Testing

This brings all the current tests from 8.0 branch.
Customer validated on private bits in production - neither functional, nor perf regression.

## Risk

Medium.
While the change is quite large, it should be specific just to that property i.e. it should not impact TLS handshake or any other I/O on `SslStream`. Since the `IsMutuallyAuthenticated` is already unreliable this should bring it up to 8.0 code base to fix all known cases when it is incorrect. To reduce complexity, this fixes only Windows as macOS & Linux changes from PR #68009 had more significant impact on functionality and flow.


rzikm (approved):
LGTM if CI is green

carlossanlop:
> LGTM if CI is green

@wfurt please send an email to Tactics requesting approval and add the `servicing-consider` label. I couldn't find an email yet. We still have time to include this in the November release.

karelz:
@carlossanlop we will bring it in for December. We need to prepare also 7.0 backport - fixing only 6.0 would be weird. And as you see from the delta, it is rather involved change, so I don't want to rush it.

carlossanlop:
Friendly reminder: If you'd like this to be included in the December release, please merge it before Tuesday November 14th EOD (Code Complete).

karelz:
Thanks @carlossanlop we want to get validation on privates before we send it to Tactics. We will miss also December release.

src/libraries/Common/src/Interop/Windows/SspiCli/SecuritySafeHandles.cs
```diff
@@ -320,10 +321,15 @@ public new IntPtr DangerousGetHandle()
...
     {
+#pragma warning disable 0649
+         // This is used only by SslStream but it is included elsewhere
+         public X509Certificate? LocalCertificate;
+ #pragma warning restore 0649
         public SafeFreeCredential_SECURITY() : base() { }
 
         protected override bool ReleaseHandle()
         {
+            LocalCertificate?.Dispose();
```
stephentoub, line 332: Same question as on the 7.0 PR

stephentoub (approved):

src/libraries/Common/src/Interop/Windows/SspiCli/SecuritySafeHandles.cs
rzikm: Let's keep the disucssion here for both PRs.

I don't see anything in the dispose implementation nor transitively called methods that could throw. The Reset method which is called there is virtual, *but* we explicitly make sure we assign only `X509Certificate2` instances to the field.

wfurt: This is what current `main` does. If we are concern we may wrap it In `try/catch` or simply re-order the code. In either case my preference would be to keep it consistent across all supported branches to make future back ports easier

src/libraries/Common/src/Interop/Windows/SspiCli/SSPIWrapper.cs
```diff
@@ -426,27 +426,41 @@ private static unsafe int EncryptDecryptHelper(OP op, ISSPIInterface secModule,
...
         }
 
+        public static bool QueryContextAttributes_SECPKG_ATTR_REMOTE_CERT_CONTEXT(ISSPIInterface secModule, SafeDeleteContext securityContext, out SafeFreeCertContext? certContext)
+            => QueryCertContextAttribute(secModule, securityContext, Interop.SspiCli.ContextAttribute.SECPKG_ATTR_REMOTE_CERT_CONTEXT, out certContext);
+
+        public static bool QueryContextAttributes_SECPKG_ATTR_LOCAL_CERT_CONTEXT(ISSPIInterface secModule, SafeDeleteContext securityContext, out SafeFreeCertContext? certContext)
+            => QueryCertContextAttribute(secModule, securityContext, Interop.SspiCli.ContextAttribute.SECPKG_ATTR_LOCAL_CERT_CONTEXT, out certContext);
+
+        public static bool QueryContextAttributes_SECPKG_ATTR_REMOTE_CERT_CHAIN(ISSPIInterface secModule, SafeDeleteContext securityContext, out SafeFreeCertContext? certContext)
+            => QueryCertContextAttribute(secModule, securityContext, Interop.SspiCli.ContextAttribute.SECPKG_ATTR_REMOTE_CERT_CHAIN, out certContext);
```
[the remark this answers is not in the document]
wfurt: updated

karelz:
Approved by Tactics (@SteveMCarroll) on 1/9 via email. Adding Servicing-approved label accordingly.

PR #92686, opened 2023-09-27
Avoid membarrier on lower Android versions

### LINKED ISSUE #92196: [nativeaot] SYS_SECCOMP signal on Android 9.0 and below
### Description

I am porting a project to android using native aot and dotnet SDK 8.0.100-rc.1
One user reported a crash on his Blackview Max 1 running android 8.0, the CPU of the device is a MediaTek Helio P23 (arm64).
I compiled an hello world console application and asked them to run it in termux, it crashes with the same error.

I don't have a device to personally reproduce the issue.

### Reproduction Steps

Create an hello world project following the native aot docs and try to run on a mediatek CPU
```
dotnet new console --aot
dotnet publish -r linux-bionic-arm64 -p:DisableUnsupportedError=true -p:PublishAotUsingRuntimePack=true
```

This is the exact binary i used for this test: [helloandroid.zip](https://github.com/dotnet/runtime/files/12643256/helloandroid.zip)

### Expected behavior

The application prints hello world.

### Actual behavior

The application is terminated with 
> Fatal signal 31 (SIGSYS), code 1 in tid 7752 (helloandroid), pid 7752 (helloandroid)

Full logcat output 
```
Build fingerprint: 'Blackview/MAX1_EEA/MAX1:8.1.0/O11019/1566373453:user/release-keys'
09-17 16:58:26.530 F/DEBUG   ( 7755): Revision: '0'
09-17 16:58:26.530 F/DEBUG   ( 7755): ABI: 'arm64'
09-17 16:58:26.530 F/DEBUG   ( 7755): pid: 7752, tid: 7752, name: helloandroid  >>> ./helloandroid <<<
09-17 16:58:26.530 F/DEBUG   ( 7755): signal 31 (SIGSYS), code 1 (SYS_SECCOMP), fault addr --------
09-17 16:58:26.530 F/DEBUG   ( 7755): Cause: seccomp prevented call to disallowed arm64 system call 0
09-17 16:58:26.530 F/DEBUG   ( 7755):     x0   0000000000000000  x1   0000000000000000  x2   000000000000001d  x3   0000000000000032
09-17 16:58:26.530 F/DEBUG   ( 7755):     x4   8080808080808080  x5   fefefefefefefeff  x6   fefefefefefefeff  x7   7f7f7f7f7f7f7f7f

09-17 16:58:26.530 F/DEBUG   ( 7755):     x8   000000000000011b  x9   bdc23b16125e035d  x10  0000000000000032  x11  0000000000000000
09-17 16:58:26.530 F/DEBUG   ( 7755):     x12  0000000000000001  x13  0000000000000000  x14  0000000000000000  x15  0000007f05200000
09-17 16:58:26.530 F/DEBUG   ( 7755):     x16  0000005654385aa8  x17  0000007f056ad510  x18  0000007f05766000  x19  0000007f05a14a40
09-17 16:58:26.530 F/DEBUG   ( 7755):     x20  0000005654395000  x21  0000000000000001  x22  0000007f05a14a40  x23  0000000000000000
09-17 16:58:26.530 F/DEBUG   ( 7755):     x24  0000000000000000  x25  0000000000000000  x26  0000000000000000  x27  0000000000000000
09-17 16:58:26.530 F/DEBUG   ( 7755):     x28  0000000000000000  x29  0000007fe6e29160  x30  00000056542fefe0
09-17 16:58:26.531 F/DEBUG   ( 7755):     sp   0000007fe6e290d0  pc   0000007f056ad530  pstate 0000000000000000

09-17 16:58:26.530 W/crash_dump64( 7755): type=1400 audit(0.0:140): avc: denied { search } for name="com.termux" dev="dm-2" ino=508640 scontext=u:r:crash_dump:s0:c512,c768 tcontext=u:object_r:app_data_file:s0:c512,c768 tclass=dir permissive=0
09-17 16:58:26.532 F/DEBUG   ( 7755): 
09-17 16:58:26.532 F/DEBUG   ( 7755): backtrace:
09-17 16:58:26.532 F/DEBUG   ( 7755):     #00 pc 000000000001d530  /system/lib64/libc.so (syscall+32)
09-17 16:58:26.532 F/DEBUG   ( 7755):     #01 pc 000000000003ffdc  /data/data/com.termux/files/home/helloandroid (offset 0x85000)
```

I can't seem to make sense out of this stack trace since pc for the stack frame 01 doesn't even seem to point to .text but i'm not an android developer.

I can run the same binary on several snapdragon devices and it works as expected.

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

- Built on linux x64 using SDK 8.0.100-rc.1 and ndk 25.0.8775105
- Attempted to run on a Blackview Max 1
   - MediaTek Helio P23 MT6763T
   - Android 8.0 
   - Stock rom
- Seems to be specific to this configuration

### Other information

_No response_

MichalStrehovsky:
Hopefully fixes #92196.

I don't actually have an ARM64 device with an old Android version so I can't testing it actually fixes the problem, but it's a plausible fix.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/gc/unix/gcenv.unix.cpp
+++ b/src/coreclr/gc/unix/gcenv.unix.cpp
@@ -168,6 +168,17 @@ enum membarrier_cmd
 
 bool CanFlushUsingMembarrier()
 {
+
+#ifdef TARGET_ANDROID
+    // Avoid calling membarrier on older Android versions where membarrier
+    // may be barred by seccomp causing the process to be killed.
+    int apiLevel = android_get_device_api_level();
+    if (apiLevel < __ANDROID_API_Q__)
+    {
+        return false;
+    }
+#endif
+
     // Starting with Linux kernel 4.14, process memory barriers can be generated
     // using MEMBARRIER_CMD_PRIVATE_EXPEDITED.
 


filipnavara:
There's a second check here:
https://github.com/dotnet/runtime/blob/ca8d6f0420096831766ec11c7d400e4f7ccc7a34/src/coreclr/pal/src/thread/process.cpp#L2578

MichalStrehovsky:
> There's a second check here:
> 
> https://github.com/dotnet/runtime/blob/ca8d6f0420096831766ec11c7d400e4f7ccc7a34/src/coreclr/pal/src/thread/process.cpp#L2578

That's in CoreCLR PAL that NativeAOT doesn't use. Would be nice to fix, but not in a PR we might backport to 8.0.

jkotas (approved):

src/coreclr/gc/unix/gcenv.unix.cpp
```diff
@@ -168,6 +168,17 @@ enum membarrier_cmd
 
 bool CanFlushUsingMembarrier()
 {
+
+#ifdef TARGET_ANDROID
+    // Avoid calling membarrier on older Android versions where membarrier
+    // may be barred by seccomp causing the process to be killed.
+    int apiLevel = android_get_device_api_level();
+    if (apiLevel < __ANDROID_API_Q__)
```
MichalPetryka, line 176: Maybe check if `seccomp` is enabled for the process first to unblock the usage on older versions?

MichalStrehovsky: ART doesn't bother with these old versions either. I only see it as more opportunities for bugs. We already have a problem testing _this_ fix.

MichalStrehovsky:
@jkotas do you have any opinion about backporting this? It feels like low risk to backport (it is all under android ifdefs), but if I have to write "Testing: none" in the bug template, I don't see this getting approved.

jkotas:
The fix can be validated by the OP by picking up nightly .NET 9 packages.

If we hear that the fix works but depending on the .NET 9 preview package is not a viable workaround, we should be able to do a backport.

MichalStrehovsky:
> The fix can be validated by the OP by picking up nightly .NET 9 packages.
> 
> If we hear that the fix works but depending on the .NET 9 preview package is not a viable workaround, we should be able to do a backport.

I realized the crummy Android tablet in my household matches the specs that hit the bug. I was able to validate that I hit the issue on a Hello World without this fix, and bits from release/8.0 with this diff backported make the issue go away.

Given the market share of Android <= 9.0 && >= 8.0 (the hole between when seccomp was enabled and when membarrier was allowed, about [6% of addressable market](https://apilevels.com/)) and very low risk I'll ask for a backport.

PR #92689, opened 2023-09-27
Make the Environments class partial

erikjanwestendorp:
Not sure if this is a desired addition, but maybe it's a good idea to make the Environments.cs class partial so that it's easy (and straightforward) extendable? For example adding a 'Test' environment would be easier 😄


huoyaoyuan:
> but maybe it's a good idea to make the Environments.cs class partial so that it's easy (and straightforward) extendable?

It wouldn't. `partial` is only effective within one compilation, which typically produces one dll assembly. It has no effect for shipped dll.

teo-tsirpanis:
This change won't have the effect you think. `partial` means that the type can have additional members _in the same assembly_. For extending members across assemblies see the extensions proposal of the C# language (dotnet/csharplang#5497).

stephentoub:
Closing for cited reasons.

PR #92694, opened 2023-09-27
[RISC-V] Fix target type unsignedness detection in genFloatToIntCast()

tomeksowi:
Use `varTypeIsUnsigned` which checks for VTF_UNS on cast **target** type classification.

Fixes TryConvertToSaturatingUInt64Test and TryConvertToTruncatingUInt64Test from System.Runtime.Numerics.ComplexTests_GenericMath.

Part of #84834

### DIFF
--- a/src/coreclr/jit/codegenriscv64.cpp
+++ b/src/coreclr/jit/codegenriscv64.cpp
@@ -570,11 +570,11 @@ void CodeGen::genFuncletProlog(BasicBlock* block)
         // fiFrameType constraints:
         assert(frameSize < -2048);
 
-        offset       = -frameSize - genFuncletInfo.fiSP_to_FPRA_save_delta;
-        int SP_delta = roundUp((UINT)offset, STACK_ALIGN);
-        offset       = SP_delta - offset;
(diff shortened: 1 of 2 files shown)


src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -3340,7 +3340,7 @@ void CodeGen::genFloatToIntCast(GenTree* treeNode)
     emitAttr dstSize = EA_ATTR(genTypeSize(dstType));
     noway_assert((dstSize == EA_4BYTE) || (dstSize == EA_8BYTE));
 
-    bool        IsUnsigned = treeNode->gtFlags & GTF_UNSIGNED;
+    bool        IsUnsigned = varTypeIsUnsigned(dstType);
```
jakobbotsch, line 3343: Can you please fix the naming convention now that you are here?

tomeksowi: Sure, fixed other locals as well.

jakobbotsch (approved):

PR #92695, opened 2023-09-27
Fix Common.Tests.GetPrettyName_CannotRead_ReturnsNull test for root user

sirntar:
That test was meant to change privileges of file to `0` and then try to read file, which should resolve in System.Exception, but instead it reads file normally.

I modified test, so it will just check whether file permissions has been set to `None` if user have root permissions, because in that case kernel would ignore privileges check, so the test would fail.

Of course there is an option to change code in corelib instead of test, but I'm not sure if root should be artificially restricted.

cc @tomeksowi @wscho77 @HJLeee @JongHeonChoi @t-mustafin @gbalykov @clamp03 @yurai007 @jkotas

### DIFF
--- a/src/libraries/Common/tests/Tests/Interop/OSReleaseTests.cs
+++ b/src/libraries/Common/tests/Tests/Interop/OSReleaseTests.cs
@@ -53,7 +53,7 @@ namespace Common.Tests
             string path = CreateTestFile();
             File.SetUnixFileMode(path, UnixFileMode.None);
 
-            // If user have root permissions, kernel doesn't care about access priviliges,
+            // If user have root permissions, kernel doesn't care about access privileges,
             // so there is no point in expecting System.Exception
             if (!Environment.IsPrivilegedProcess)
             {
--- a/src/libraries/Common/tests/Tests/Interop/OSReleaseTests.cs
+++ b/src/libraries/Common/tests/Tests/Interop/OSReleaseTests.cs
@@ -50,19 +50,38 @@ namespace Common.Tests
         [Fact, PlatformSpecific(TestPlatforms.Linux)]
         public void GetPrettyName_CannotRead_ReturnsNull()
         {
+            // This test only applies for non-root users.
+            if (Environment.IsPrivilegedProcess)
+            {
+                return;
+            }
+
+            string path = CreateTestFile();
+            File.SetUnixFileMode(path, UnixFileMode.None);
+
+            Assert.ThrowsAny<Exception>(() => File.ReadAllText(path)); 
+
+            string? name = Interop.OSReleaseFile.GetPrettyName(path);
+            Assert.Null(name);
+        }
+
+        [Fact, PlatformSpecific(TestPlatforms.Linux)]
+        public void GetPrettyName_NonePrivileges_CanRead_ReturnsNull()
+        {
+            // This test only applies for root users.
+            if (!Environment.IsPrivilegedProcess)
+            {
+                return;
+            }
+
             string path = CreateTestFile();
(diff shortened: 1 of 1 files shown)


am11:
(Change is not specific to RISC-V)

cc @wfurt

src/libraries/Common/tests/Tests/Interop/OSReleaseTests.cs
```diff
@@ -52,7 +52,17 @@ public void GetPrettyName_CannotRead_ReturnsNull()
...
             File.SetUnixFileMode(path, UnixFileMode.None);
-            Assert.ThrowsAny<Exception>(() => File.ReadAllText(path));
+
+            // If user have root permissions, kernel doesn't care about access priviliges,
+            // so there is no point in expecting System.Exception
+            if (!Environment.IsPrivilegedProcess)
+            {
+                Assert.ThrowsAny<Exception>(() => File.ReadAllText(path)); 
+            }
+            else
```
wfurt, line 62: I would probably make the test conditional and if needed add separate set of tests for privilege access. I would expect more tests that checks permissions.

But in general I think is is ok if just want to get unblocked to run tests as root.

Sirntar: I wasn't certain whether I should modify the test itself or the core library functionality, so this solution was more of a quick fix.

I like your approach of creating a separate set of tests specifically for privileged users. Do you think it's worth implementing in this case? If so, I can start working on that solution tomorrow.

wfurt: perhaps @dotnet/area-system-io  can provide more guidance. I personally feel the quick fix + follow-up would be best.

src/libraries/Common/tests/Tests/Interop/OSReleaseTests.cs
```diff
@@ -52,7 +52,17 @@ public void GetPrettyName_CannotRead_ReturnsNull()
         {
             string path = CreateTestFile();
             File.SetUnixFileMode(path, UnixFileMode.None);
-            Assert.ThrowsAny<Exception>(() => File.ReadAllText(path));
+
+            // If user have root permissions, kernel doesn't care about access priviliges,
```
danmoseley, line 56:
```suggestion
            // If user have root permissions, kernel doesn't care about access privileges,
```

danmoseley: (nit, just merged in case that saves you time)

Sirntar:
@dotnet-policy-service agree company="Samsung"

src/libraries/Common/tests/Tests/Interop/OSReleaseTests.cs
```diff
@@ -50,9 +50,38 @@ public void GetPrettyName_NoFile_ReturnsNull()
...
+
+            string? name = Interop.OSReleaseFile.GetPrettyName(path);
+            Assert.Null(name);
+        }
+
+        [Fact, PlatformSpecific(TestPlatforms.Linux)]
+        public void GetPrettyName_NonePrivileges_CanRead_ReturnsNull()
+        {
+            // This test only applies for root users.
+            if (!Environment.IsPrivilegedProcess)
```
tomeksowi, line 72: Perhaps put `[ConditionalFact(typeof(PlatformDetection), nameof(PlatformDetection.IsPrivilegedProcess))]` on the test signatures? (I think it works with `Environment` too).

tomeksowi (approved):

wfurt (approved):
LGTM

PR #92696, opened 2023-09-27
[wasm] Disable MetricsSupport feature by default

radekdoulik:
This should improve the startup perf and size.


radical:
I was doing this just for the `blazorminapp` in `dotnet/performance` - https://github.com/dotnet/performance/pull/3367 . But this would supersede it.

Are you able to see the change have any change in file size etc?

radekdoulik:
> I was doing this just for the `blazorminapp` in `dotnet/performance` - [dotnet/performance#3367](https://github.com/dotnet/performance/pull/3367) . But this would supersede it.
> 
> Are you able to see the change have any change in file size etc?

Not locally. Hope to see some difference in the https://radekdoulik.github.io/WasmPerformanceMeasurements/ once it lands in main.

lewing (approved):

PR #87045, opened 2023-06-02
Enable identical code folding on Linux

MichalStrehovsky:
This instructs the linker to deduplicate identical sections on Linux. We were previously not doing this for two reasons:

1. We weren't passing the command line switch.
2. We had trouble getting linker to actually fold things. Linux linkers are particularly picky on what they're willing to fold. We've been historically putting these things to RW sections (dotnet/corert#686) and that doesn't help. Looking at C++, it looks like it places vtables into `.text` section, so let's just do the same.

I also found out why `--foldmethodbodies` doesn't work outside Windows - we place managed code into a `__managedcode` section and Linux linkers won't fold sections that have names that are valid C identifiers (would need to name this `.managedcode` like on Windows) - https://github.com/dotnet/llvm-project/blob/c01ca3bc8a420b3796d816c43bdd06e337c72ea6/lld/ELF/ICF.cpp#L192. Not addressing that part because the whole thing looks like a hairy yak and this option is not supported anyway.

Cc @dotnet/ilc-contrib

PR #92705, opened 2023-09-27
[CI ONLY] [NativeAOT] ObjWriter in C#

filipnavara:
Ref: #77178

This is reimplementation of NativeAOT ObjWriter in pure C# instead of depending on LLVM. It implements Mach-O, ELF, COFF object file emitter with DWARF and CodeView debugging information. Only x64 and arm64 targets are implemented to cover officially supported platforms. Certain features are not implemented yet, e.g. COMDAT in ELF. Other features like DWARF debugging info generation are currently slower than the previous implementation. A limited testing was done on osx-arm64, win-x64, and linux-x64. Previous version of the branch was also tested on osx-x64, win-arm64, and linux-arm64.

**Caveat: This is NOT for review, the draft PR is opened specifically to run smoke tests only. The code was rebased over current main branch and updated to reflect most ObjWriter changes from the past year (both on the runtime repo side and the LLVM fork repo). The performance and structure of the code is not the final shape and I expect to rewrite certain parts before submitting this for actual review.**

cc @TIHan

Tracking list of issues found by the CI:
- [x] linux-x64: `System.NotSupportedException: Unsupported relocation: IMAGE_REL_TLSGD`
- [x] linux-arm64: `relocation R_AARCH64_TLSDESC_ADR_PAGE21 cannot be used against symbol 'tls_InlinedThreadStatics'`
- [x] linux-arm64: `libunwind: malformed DW_CFA_register DWARF unwind, reg too big`
- [x] linux-arm64: Different pointer encoding is used in EH frames which seems to cause issues
- [x] iOS/tvOS: `warning: ignoring file /tmp/helix/working/B5450932/w/C6420AD0/e/publish/native/iOS.Device.Aot.Test.o, building for iOS-arm64 but attempting to link with file built for unknown-unsupported file format ( 0x7F 0x45 0x4C 0x46 0x02 0x01 0x01 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 )`
- [x] win-x64: `Framework.lib(System.Private.CoreLib.obj) : fatal error LNK1243: invalid or corrupt file: COMDAT section 0x20001 associated with following section 0x0` (MultiModule test)
- [x] win-arm64: `CoreFXTestLibrary.AssertTestException: Assert.AreEqual: Expected: [System.Func2[CommonType10[],CommonType10[]]]. Actual: [System.Func2[CommonType10[],CommonType10[]]].`


filipnavara:
Tracking list of issues found by the CI:
- [ ] linux-x64: `System.NotSupportedException: Unsupported relocation: IMAGE_REL_TLSGD`
- [ ] linux-arm64: `relocation R_AARCH64_TLSDESC_ADR_PAGE21 cannot be used against symbol 'tls_InlinedThreadStatics'`
- [ ] iOS/tvOS: `warning: ignoring file /tmp/helix/working/B5450932/w/C6420AD0/e/publish/native/iOS.Device.Aot.Test.o, building for iOS-arm64 but attempting to link with file built for unknown-unsupported file format ( 0x7F 0x45 0x4C 0x46 0x02 0x01 0x01 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00 )`

filipnavara:
Still not sure what is going on with the win-arm64 tests. They pass locally on my Windows Dev Kit 2023.

am11:
This one looks a bit less mysterious:

```
Running Test: ExistingInstantiations.Test.TestWithExistingInst
Caught Unexpected exception:System.NotSupportedException: Cannot retrieve a MethodInfo for this delegate because the method it targeted was not enabled for metadata.
   at Internal.Reflection.Extensions.NonPortable.DelegateMethodInfoRetriever.GetDelegateMethodInfo(Delegate) + 0x29c
   at ExistingInstantiations.Test.TestWithExistingInst() + 0x518
   at CoreFXTestLibrary.Internal.Runner.RunTestMethod(TestInfo) + 0x364
   at CoreFXTestLibrary.Internal.Runner.RunTest(TestInfo) + 0x20
---- Test FAILED ---------------
```

perhaps `TestReverseLookupsWithArrayArg` failure has the same underlying reason?

BTW, CI is using precisely this toolchain version:

```
  -- The C compiler identification is MSVC 19.36.32537.0
  -- The CXX compiler identification is MSVC 19.36.32537.0
```
to cross-compile for arm64 on x64: `build.cmd -ci -arch arm64 -os windows  -s clr.aot+host.native+libs+tools.illink -c Release -rc Release -lc Release -hc Release`

src/coreclr/tools/aot/ILCompiler.Compiler/ILCompiler.Compiler.csproj
```diff
@@ -586,8 +586,16 @@
     <Compile Include="Compiler\DependencyAnalysis\Target_LoongArch64\LoongArch64ReadyToRunHelperNode.cs" />
     <Compile Include="Compiler\DependencyAnalysis\Target_LoongArch64\LoongArch64ReadyToRunGenericHelperNode.cs" />
     <Compile Include="Compiler\DependencyAnalysis\Target_LoongArch64\LoongArch64UnboxingStubNode.cs" />
+    <Compile Include="Compiler\ObjectWriter\DwarfBuilder.cs" />
+    <Compile Include="Compiler\ObjectWriter\DwarfCie.cs" />
+    <Compile Include="Compiler\ObjectWriter\DwarfEhFrame.cs" />
+    <Compile Include="Compiler\ObjectWriter\DwarfFde.cs" />
+    <Compile Include="Compiler\ObjectWriter\DwarfNative.cs" />
     <Compile Include="Compiler\ObjectWriter\LlvmObjectWriter.cs" />
```
SingleAccretion, line 599: I know this is `[NO REVIEW]`, but this bit of feedback is perhaps more appropriate for this stage...

`LlvmObjectWriter` would soft-conflict with the downstream [`LLVMObjectWriter.cs`](https://github.com/dotnet/runtimelab/blob/feature/NativeAOT-LLVM/src/coreclr/tools/aot/ILCompiler.LLVM/CodeGen/LLVMObjectWriter.cs). It would be nicer from our downstream perspective for it to be named something more distinctive, e. g. `NativeLlvmObjectWriter`.

filipnavara:
> This one looks a bit less mysterious:

Thanks, @am11. I will check that one. I will likely wait till tomorrow to have direct access to the Win/ARM machine instead of just RDP-over-RDP-over-Tailscale. I have an uncommitted code where I match more closely the COFF output from the old ObjWriter so it's diffable to certain extent, and easier to spot differences. The DynamicGenerics test produces high number of sections and relocations and triggers the "big obj" code paths which didn't get much testing, so that's my primary suspicion.

src/coreclr/tools/aot/ILCompiler.Compiler/ILCompiler.Compiler.csproj
filipnavara: Thanks for the heads-up. I will check what is done downstream and try to avoid any conflicts.

TIHan:
Thank you for making the PR @filipnavara .

I'm out for a bit and hopefully will look at this next week. So far it looks promising.

filipnavara:
> Still not sure what is going on with the win-arm64 tests. They pass locally on my Windows Dev Kit 2023.

Turns out the error was somewhere between the chair and the keyboard. I had a wrong branch checked out.

filipnavara:
I found the issue in the COFF/ARM64 code. I was incorrectly ignoring the addend for `IMAGE_REL_BASED_ARM64_PAGEOFFSET_12A` (and it was actually an `Assert` on debug build). This broke fat pointers that use the `+2` address offset to be distinguished from real pointers. The same issue didn't happen for Mach-O since it rewrites this particular relocation into a more complex one. On ELF we emit the `RELA` section with addend inside the relocations, not inside the code, so it happened to work. COFF is the only platform that embeds this kind of relocation in the data.

While searching for the root cause I found couple more issues that are mostly harmless but should be fixed nevertheless. I'll clean up the code and commit it soon.

src/coreclr/tools/aot/external/LibObjectFile/changelog.md
am11, line 1: For trailing whitespace errors:

```sh
# add dotnet remote (once)
git remote add dotnet https://github.com/dotnet/runtime

git fetch dotnet

# remove trailing whitespace only from files modified by the delta (mac's sed)
git show --name-only --pretty="" HEAD...dotnet/main |\
    xargs -I{} sh -c "test -f {} && sed -i '' 's/[[:space:]]*$//' {}"

# (with linux/ gnu-sed, remove the space between -i and '')
```

filipnavara: They come from the LibObjectFile dependency. I am intentionally ignoring it until we decide how/if we want/need to consume that code. I'm very likely going to rewrite the DWARF debugging info emitter next, and then I will start evaluating the pros and cons of those dependencies.

am11: One option is PackageReference, just like the Cecil dependency is handled in   illink. However, rewrite might help get rid of this whole libobjfile. 😅

filipnavara:
I think the code now reached a point where it passes the smoke tests for the supported platforms, which was the primary purpose of this PR. I'll keep it open for the moment, but it served its purpose.

Aside from some general structural improvements these are the areas I intend to explore next (in no particular order):
- Emitting DWARF debugging info eagerly and directly into corresponding sections without the intermediate representation.
- Resolve relative relocations within same section before writing them down the object files. This should have no effect on output after linking. It does, however, produce slightly smaller object files, and it may avoid hitting limits on the number of relocations in certain formats.
- Optimize the string table building to take advantage of suffix matches. All the formats use some variation of string table with null-terminated string referenced by offset. It's possible to generate smaller object files by producing `_unwind0_XYZ` at offset N, and then reusing the suffix `XYZ` at offset N+9 as a different string.
- Explore generating XDATA/PDATA unwinding info in COFF to be closer to the structure that MSVC produces.
- Implement COMDAT handling for ELF file format. It's the one big missing thing for feature parity.

src/coreclr/tools/aot/external/LibObjectFile/changelog.md
filipnavara: I love a lot of things about LibObjectFile, and naturally I also appreciate my own Mach-O libraries. The prototype would have never been written without them. That said, the write-only nature of ObjWriter means that we use only a tiny slice of these libraries, and it may be more manageable to just do everything ourselves. I don't have a good answer for that yet.

Currently LibObjectFile is used for emitting the DWARF debugging info, and that's pretty complex piece of code. Once again, we don't need to read or validate the DWARF information in the compiler though, so half of the code is unused or worse, produces a performance overhead. LibObjectFile implementation is not the most efficient in terms of performance and memory usage. I'm more than happy to contribute back to resolve these issues if it ends up making sense. I definitely don't suffer from the "not invented here" syndrome. That said, the experience with producing CodeView debugging information and my improved understanding of the DWARF format convinced me that it may be viable to write a readable and efficient code to produce the DWARF information in a fast write-only fashion.

am11: Agreed, I think reusing community libraries is encouraged in modern .NET ecosystem, including the runtime repo. Also, (I'm not sure if in the future it may even be possible) replacing llvm-libunwind with managed implementation using these abstractions for NativeAOT could be one reason to expand their usage to read & write. So there could be more to justify their inclusion. 👍

am11:
According to `dotnet-trace` on linux-arm64. ilc.dll spends > 11 seconds in

```
libobjectfile!LibObjectFile.Elf.ElfWriter`1[LibObjectFile.Elf.ElfEncoderDirect].WriteSections()
```

during System.Runtime.Tests publishing (and output object size is 273M vs. main's 244M). With published-ilc, this test OOMs (code 137, the current CI failure).

[speedscope-trace.zip](https://github.com/dotnet/runtime/files/12777316/speedscope-trace.zip)

filipnavara:
@am11 Thanks for looking into it, really appreciated!

I didn't focus on the performance and memory usage outside of isolated scenarios. I profiled some code paths for COFF and CodeView but there's very little overlap with what ELF and DWARF does. I'll check the trace you provided and do some profiling on my side as well.

The output size difference is expected. It's mostly caused by extra relocations in the output and non-optimized string table (as mentioned in the "future work" list above). There's also some difference in debugging info size but not nearly as big.

Notably, the memory usage of the DWARF debugging info emitter is pretty high. It's not easy to refactor without significant changes to LibObjectFile. It operates on a "document" model (akin to `JsonDocument` but binary and multiple cross-linked documents), while we really need something closer to the "writer" model (akin to `Utf8JsonWriter`). There's already an internal model present in the compiler and it is memory intensive to convert from one model to the other. Additionally, the DWARF model in LibObjectFile has additional abstraction for relocations which are then converted to ELF relocations. This all adds up both in memory and time profiles.

I did some comparisons with the debugging info turned off, and the results were largely favorable to the C# implementation. That said, I did it primarily on Mach-O, not ELF.

filipnavara:
> speedscope-trace.zip

Wow, I totally didn't expect `ElfObjectFile.Verify` to take 5 times as much as `ElfObjectFile.WriteSections`. I knew `ElfStringTable.GetOrCreateIndex` is extremely inefficient but this exceeded all my expectations. It tries to do suffix matching but in the process creates a ton of strings. Given all the time spent in the method I suspect the memory usage culprit could be there as well.

filipnavara:
Since some people are apparently following the PR and looking at some of the performance issues, I made an isolated sample showing the problem with `ElfStringTable`: https://github.com/filipnavara/StringTableBenchmark

I used three different algorithms to build the string table:
- **Naive** produces string table with no suffix deduplication (similar to the code used for COFF)
- **MultiKeySort** sorts the input strings and then produces optimized string table with deduplicated suffixes (similar to the LLVM string table builder)
- **ElfStringTable** uses LibObjectFile's `ElfStringTable`, which incrementally builds the table and uses dictionary to deduplicate suffixes

Note that `ElfStringTable` technically provides an incremental API where the full set of string is not known beforehand. However, in reality it's pre-populated with the full string set anyway, and this property can be used for optimizing the common case without changing the external API.

Without further ado, here are the results on my MacBook Air M1 for a same string set taken from `UseSystemResourceKeys.o` in the smoke tests. The set contains slightly less than 25k strings where many can be suffix compressed.

| Method         | Mean       | Error     | StdDev    | Gen0       | Gen1       | Gen2      | Allocated |
|--------------- |-----------:|----------:|----------:|-----------:|-----------:|----------:|----------:|
| ElfStringTable | 350.838 ms | 6.8811 ms | 8.1914 ms | 38000.0000 | 21000.0000 | 7000.0000 | 316.81 MB |
| MultiKeySort   |  11.456 ms | 0.0469 ms | 0.0439 ms |  1062.5000 |  1046.8750 |  968.7500 |   4.87 MB |
| Naive          |   2.374 ms | 0.0060 ms | 0.0050 ms |  1042.9688 |  1015.6250 |  996.0938 |   5.82 MB |

You can clearly see that the approach used by `ElfStringTable` doesn't scale well.

am11:
Great optimizations! linux-arm64 object size is 241M vs. 244M on main, and CI leg isn't jamming. :)

agocke:
@filipnavara really good info. Agreed that ElfStringTable isn't looking great

filipnavara:
> @filipnavara really good info. Agreed that ElfStringTable isn't looking great

I committed an experimental fix and it passed the CI. I'll submit it upstream and then focus on the further work I mentioned above.

TIHan:
@filipnavara , this is looking really good and glad others were able to look at it. The performance improvements do look great.

How far do you think your solution is from matching the existing functionality? It looks like the smoke tests are passing on all the platforms. @agocke , are there any scenarios that we need to cover that are not covered by the tests?

filipnavara:
> I think the big thing is that we generally don't put externals into runtime itself for larger dependencies.

I agree in general. I would like to postpone this step until we decide whether LibObjectFile is really the right fit. I want to finish the DWARF experiments first to make up my mind on that. It should not take too much time.

(If you feel like there's anything that would be easier to communicate over a call we can arrange it.)

am11:
> ELF is missing COMDAT support.

@filipnavara, out of curiosity, is it so that COMDATs are currently given no special treatment in ELF emitter (but the data is emitted in the object nonetheless) and the plan is to explore https://github.com/dotnet/runtime/issues/77491, or implement LLVM's COMDAT handling counterpart as good/bad as theirs? Since COMDATs [are not natural to Unix](https://maskray.me/blog/2021-07-25-comdat-and-section-group), can we avoid spilling COMDAT down to objwriter from higher level context using some neutral semantics and let each writer choose from what's available on their native stack ([ICF](https://github.com/dotnet/runtime/pull/87045), `SHF_GNU_RETAIN` etc.)?

Please CMIIW, this sounds like something which could be focused on and explored independently from this initial port, since it would bring additional wins (on top of the wins we are already seeing in this PR vs. `main`) and it's not about the missing data.

filipnavara:
> ...or implement LLVM's COMDAT handling counterpart as good/bad as theirs?

The ILC compiler can run in many modes and some of them depend on the COMDAT to work properly (eg. the multi-module mode with pre-compiled framework and duplicate methods between object files), while others use it just as size a optimization (foldable data sections, unboxing stubs IIRC). The size optimization would be visible in the linked file and not the intermediate object file.

I intend to get it to the same shape as the LLVM ObjWriter, which means emitting some section groups with the `GRP_COMDAT` flag. I didn't look at the details for almost a year, so I need to refresh my memory on what exactly is currently emitted.

The code for COMDAT is already present for COFF, so I don't expect it to be particularly hard. I mostly ignored it until now since it was not strictly necessary but I want to make sure that I don't introduce any size regressions (and if I do, I want to understand where they are, and why they are there).

filipnavara:
> ELF is missing COMDAT support.

Minor update:
- The code that emits ELF groups with the COMDAT flag is added.
- DWARF debugging info used incorrect addressing and didn't emit correct relocations for static data variables, so I fixed it to follow something that's closer to clang DWARF-4 output. This is ELF specific, since Mach-O doesn't use COMDAT sections, and it doesn't use relocations for DWARF either (linker produces the table that maps addresses to the original .o file).
- It uncovered multiple unsupported scenarios in LibObjectFile which happen when there's more than 0xff00 sections in the output file. This happens even for the smoke tests when they get compiled with `--methodbodyfolding`. For example, the `Reflection_FromUsage` test generates around 82k sections. I fixed few of the scenarios, but there are few more that still result in silent corruption (symbol table truncates the section numbers instead of using extension table).

TIHan:
This is great @filipnavara . So awesome how well this is coming along.

filipnavara:
Small update, the PR is not abandoned even if I didn't commit anything for a while.

Last week I fixed issues in LibObjectFile related to large number of sections, and couple of related performance issues. The was reviewed and new version of LibObjectFile was released on the weekend. I'll update this PR with this version, and that should hopefully wrap up the "Implement COMDAT handling for ELF file format" item on the TODO list.

I am also about 70% done with rewriting the DWARF code. I keep it on a [separate branch] (https://github.com/filipnavara/runtime/tree/objwriter-dwarf-wip) where I frequently "force push". It's very much work in progress but it works for some simple ELF+DWARF use cases and generates all the same information that was produced by the LLVM ObjWriter.

agocke:
@filipnavara no rush. Lemme know if you need help with anything

filipnavara:
Looking into the Windows failures now. I can reproduce them locally. They are not caused by the very last commit, and the previous CI results were already wiped. It happens right after dehydration in some array allocation. All the NativeAOT smoke tests still pass. I don't know exactly how and when the crossgen2 compilation started failing, and I was under the impression that the live ILCompiler was previously not used for this step while it is definitely used now.

filipnavara:
For posterity, the crossgen2 error was actually caused by the last commit ([Add generic StringTableBuilder and use it in COFF](https://github.com/dotnet/runtime/pull/92705/commits/cc7b405bd7d4068ceb89e8937db6a80a52262755)) but it was in a very non-obvious way. I tried to revert that commit but it still failed because I missed that the crossgen2.obj file was not rebuilt.

The COFF format has plenty of quirks. It originally supported only 8 characters for section name. That was later extended to support longer names by using a `/DDDDDDD` notation. `/` is a prefix that tells that the real name is in a string table and it's followed by 7 decimal digits representing the offset into the string table. So, what happens when the string is at offset larger than 9999999? My code incorrectly truncated it.

There are two ways to fix it. The obvious and easy one is to place the section names at the beginning of the string table to avoid the large offsets. Turns out that there's an extension of the extension of the COFF format to reference higher offsets, and that's the other way. The section name can also use the `//BBBBBB` notation, where `//` is the prefix and the rest is an offset expressed as base64 number for a total maximum offset of 64^6 or 0xfffffffff (incl. zero offset). This extended notation works for up to 64 GiB string table which should be enough for everyone.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ObjectWriter/CoffObjectWriter.cs
```diff
@@ -0,0 +1,1126 @@
...
+                Span<byte> stringTableSize = stackalloc byte[4];
+                BinaryPrimitives.WriteUInt32LittleEndian(stringTableSize, Size);
+                stream.Write(stringTableSize);
+                base.Write(stream);
+            }
+        }
+
+        private static class JamCrc32
+        {
+            private static uint[] Table =
```
am11, line 1063: nit: lookup table can be replace with:

```c#
using System.Numerics;
...
private static readonly uint[] s_crcLookup = Crc32ReflectedTable.Generate(0xEDB88320u);
```

```csproj
<Compile Include="$(LibrariesProjectRoot)Common\src\System\Numerics\Crc32ReflectedTable.cs" />
```

filipnavara: Good idea, I will test it!

filipnavara: The polynomial is different, `0x04C11DB7` according to my notes, and there may be some other small differences. I'll try to make a test project to replicate the table with `Crc32ReflectedTable.Generate`.

am11: I tested it on fiddle: https://dotnetfiddle.net/SJW4qF :)

am11: aside:
```c#
System.IO.Hashing.Crc32 c = new();
c.Append(stream);
var result = ~BitConverter.ToUInt32(c.GetCurrentHash());
```
result matches JAMCRC from https://crccalc.com but `class JamCrc32` results are a bit off. If it's a bug, we can replace the whole class?

filipnavara: `System.IO.Hashing.Crc32` uses different intial state (~0) than the COFF CRC (0). That's why it cannot be used directly.

src/coreclr/tools/aot/ILCompiler.Compiler/ILCompiler.Compiler.csproj
```diff
@@ -22,6 +22,8 @@
     <ProjectReference Include="..\ILCompiler.DependencyAnalysisFramework\ILCompiler.DependencyAnalysisFramework.csproj" />
     <ProjectReference Include="..\ILCompiler.MetadataTransform\ILCompiler.MetadataTransform.csproj" />
     <ProjectReference Include="..\ILCompiler.TypeSystem\ILCompiler.TypeSystem.csproj" />
+        
+    <ProjectReference Include="..\external\Melanzana\Melanzana.MachO\Melanzana.MachO.csproj" />
```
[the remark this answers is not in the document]
filipnavara: Thanks. I'm still going to discuss how to handle ingesting this library (or creating a minimal subset and including it directly).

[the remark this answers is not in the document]
am11: BTW, awesome work flattening ElfObjectWriter and removing LibOjbectFile dependency. 👍

src/coreclr/tools/aot/ILCompiler.Compiler/ILCompiler.Compiler.csproj
filipnavara: Renamed to LegacyObjectWriter.cs for now.

filipnavara:
I didn't post an update in a long while and I feel like a one is in order.

I've rewritten all the ELF, DWARF and Mach-O code not to depend on external libraries. This is a decision that can be re-evaluated later if there are significant benefits of doing so (eg. better unit tests or code sharing with other .NET components). Object files usually use a very simple subset of the full file format. The extra cost of Mach-O / ELF binary writing code is about 350 lines of code for each format (not including constants) compared to code referencing the external library.

DWARF is now a mostly direct translation of the ObjWriter code with minor tweaks to make it more maintainable (eg. automatic tracking of children in Debug Information Entries and writing the NUL terminators).

I've addressed all but one of the TODOs from https://github.com/dotnet/runtime/pull/92705#issuecomment-1741746859. The leftover is to optimize out same-section relocations from the intermediate object files. I don't really consider it to be a blocker. It's something that would be nice to have but doesn't affect the resulting executable in any way.

In the next few weeks I plan to add comments and do general code cleanup. I'll open a new PR for review once that is done.

am11:
> It's something that would be nice to have but doesn't affect the resulting executable in any way.

The final linkage will certainly take care of resolving local relocs and stripping out unused symbols. One thing I noticed the difference in object .o size of `dotnet new webapiaot` template app goes from 53M to 58M on linux-x64 (while it shaves off 1-2 seconds from `Generating native code` step, based on 10 builds average). Out of curiosity, is the size increase only because of local relocs or is there something else in play as well? Maybe those items can be tracked in a milestone:future issue.

filipnavara:
> Out of curiosity, is the size increase only because of local relocs or is there something else in play as well?

Most likely it’s the local relocs. I will check. Thanks for bringing it up.

filipnavara:
> One thing I noticed the difference in object .o size of `dotnet new webapiaot` template app goes from 53M to 58M on linux-x64 (while it shaves off 1-2 seconds from `Generating native code` step, based on 10 builds average)

Can you retry the comparison with the latest commit please?

am11:
Size is now 56M (with `DOTNET_USE_LLVM_OBJWRITER=1` it's 53M). On osx-arm64, results show improvements: was 47M, now 41M. 🙂 
(CI failure looks related to the last change, the DWARF exception suggests perhaps we need to update the debugging info?)

filipnavara:
Yeah, the test failures are definitely related 😅 . I reused some experimental code I had laying around but I forgot to use the latest version.

filipnavara:
> Size is now 56M (with `DOTNET_USE_LLVM_OBJWRITER=1` it's 53M). On osx-arm64, results show improvements: was 47M, now 41M.

Unfortunately, the improvements in osx-arm64 are a bug. Mach-O uses the subsections-via-symbols feature, so we need to preserve the relocations. The dead code elimination in the linker is allowed to rewrite the section contents and the relative offsets would change.

I tried to reproduce in on linux-x64 to see if I can explain the size difference but unfortunately, I hit an unrelated bug (https://github.com/dotnet/runtime/issues/95852).

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ObjectWriter/CodeView/CodeViewSymbolsBuilder.cs
```diff
@@ -238,7 +238,7 @@ private CodeViewRegister GetCVRegNum(uint regNum)
 
                 foreach (var sequencePoint in sequencePoints)
                 {
-                    if (lastFileName == null || lastFileName != sequencePoint.FileName)
+                    if (lastFileName is null || lastFileName != sequencePoint.FileName)
```
am11, line 241: If you prefer pattern matching terse syntax, we can reduce the verbosity a bit in those long RelocType.XX conditions:

```c#
using static ILCompiler.DependencyAnalysis.RelocType;
...
if (relocType is IMAGE_REL_BASED_ARM64_BRANCH26 or IMAGE_REL_BASED_ARM64_PAGEBASE_REL21 or
     IMAGE_REL_BASED_ARM64_PAGEOFFSET_12A or IMAGE_REL_AARCH64_TLSLE_ADD_TPREL_HI12 or ..)
...
```
(the duplication due to `IMAGE_REL_` prefix is enough as is 😅)

filipnavara: I used several different code styles as I progressed, sometimes intentionally, sometimes as a consequence of reusing existing code... I am generally open to ideas how to keep the code as terse and readable as possible ;-)

PR #92708, opened 2023-09-27
Fold GC.KeepAlive for frozen objects

EgorBo:
Basically, remove `GC.KeepAlive` used for frozen(nongc) objects if JIT can see it.
```cs
void Test()
{
    string s = "hello world";
    GC.KeepAlive(s);
}
```
Codegen diff:
```diff
; Method Runtime_90508:Test():this (FullOpts)
-      mov      rax, 0x22800008D58      ; 'hello world'
       ret    
```

This is mainly needed for code paths in RuntimeType e.g. https://github.com/dotnet/runtime/blob/main/src/coreclr/System.Private.CoreLib/src/System/RuntimeType.CoreCLR.cs#L3381 (where `this` could be a frozen object)

### DIFF
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -5698,8 +5698,6 @@ Compiler::fgWalkResult Compiler::optVNConstantPropCurStmt(BasicBlock* block,
                                                           GenTree*    tree,
                                                           bool*       changed)
 {
-    *changed = false;
-
     // Don't perform const prop on expressions marked with GTF_DONT_CSE
     // TODO-ASG: delete.
     if (!tree->CanCSE())
@@ -5820,33 +5818,15 @@ Compiler::fgWalkResult Compiler::optVNConstantPropCurStmt(BasicBlock* block,
 //    block   - The block that contains the statement that contains the tree.
 //    stmt    - The statement node in which the "tree" is present.
 //    tree    - The currently visited tree node.
-//    changed - [OUT] changes were made.
 //
 // Return Value:
 //    Returns the standard visitor walk result.
 //
-Compiler::fgWalkResult Compiler::optVNGenericPropCurStmt(BasicBlock* block,
-                                                         Statement*  stmt,
-                                                         GenTree*    tree,
-                                                         bool*       changed)
+Compiler::fgWalkResult Compiler::optVNGenericPropCurStmt(BasicBlock* block, Statement* stmt, GenTree* tree)
 {
-    *changed = false;
-
-    switch (tree->OperGet())
+    if (tree->OperIs(GT_KEEPALIVE) && vnStore->IsVNObjHandle(tree->gtGetOp1()->gtVNPair.GetConservative()))
     {
-        case GT_KEEPALIVE:
-        {
-            if (vnStore->IsVNObjHandle(tree->gtGetOp1()->gtVNPair.GetConservative()))
-            {
-                optAssertionProp_Update(gtNewNothingNode(), tree, stmt);
-                *changed = true;
-                return WALK_CONTINUE;
-            }
-        }
-        break;
-
-        default:
-            return WALK_CONTINUE;
+        optAssertionProp_Update(gtNewNothingNode(), tree, stmt);
     }
     return WALK_CONTINUE;
 }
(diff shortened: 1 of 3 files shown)


src/coreclr/jit/assertionprop.cpp
```diff
@@ -2562,6 +2562,14 @@ GenTree* Compiler::optVNConstantPropOnTree(BasicBlock* block, GenTree* tree)
     ValueNumPair vnPair = tree->gtVNPair;
     ValueNum     vnCns  = vnStore->VNConservativeNormalValue(vnPair);
 
+    // Replace KEEPALIVE(ObjHandle) with NO-OP as the object handle is always
+    // kept alive by the runtime.
```
SingleAccretion, line 2566: Can this be done in the general walk over nodes (`optAssertionProp`) instead of here?

Nominally this is really a morph transformation, but we cannot use VNs easily in morph.

src/coreclr/jit/assertionprop.cpp
```diff
@@ -2562,6 +2562,14 @@ GenTree* Compiler::optVNConstantPropOnTree(BasicBlock* block, GenTree* tree)
...
+    if (tree->OperIs(GT_KEEPALIVE))
+    {
+        const ValueNumPair opVnp = tree->gtGetOp1()->gtVNPair;
+        return opVnp.BothEqual() && vnStore->IsVNObjHandle(opVnp.GetLiberal()) ? gtNewNothingNode() : nullptr;
```
SingleAccretion, line 2570: I think you can just check the conservative VN.

src/coreclr/jit/assertionprop.cpp
EgorBo: @SingleAccretion looks like `optAssertionProp` is never fired for my example because it has zero assertions? so sounds like that path misses opportunities

src/coreclr/jit/assertionprop.cpp
EgorBo: Addressed

src/coreclr/jit/assertionprop.cpp
EgorBo: I've added a separate path for morph

EgorBo: Ah, interesting idea, I have more ideas for such generic prop. @SingleAccretion can you take a look at the impl I've just pushed

src/coreclr/jit/assertionprop.cpp
```diff
@@ -5801,6 +5809,48 @@ Compiler::fgWalkResult Compiler::optVNConstantPropCurStmt(BasicBlock* block, Sta
...
+{
+    *changed = false;
+
+    switch (tree->OperGet())
+    {
+        case GT_KEEPALIVE:
+        {
+            if (vnStore->IsVNObjHandle(tree->gtGetOp1()->gtVNPair.GetConservative()))
+            {
+                optAssertionProp_Update(gtNewNothingNode(), tree, stmt);
```
SingleAccretion, line 5841: You can `tree->gtBashToNOP` in-place and avoid allocating a new tree.

src/coreclr/jit/assertionprop.cpp
```diff
@@ -5852,18 +5902,24 @@ void Compiler::optVnNonNullPropCurStmt(BasicBlock* block, Statement* stmt, GenTr
...
-    Compiler*                   pThis = pData->pThis;
+    auto*     pData = (VNAssertionPropVisitorInfo*)data->pCallbackData;
+    Compiler* pThis = pData->pThis;
 
     pThis->optVnNonNullPropCurStmt(pData->block, pData->stmt, *ppTree);
 
-    return pThis->optVNConstantPropCurStmt(pData->block, pData->stmt, *ppTree);
+    bool         changed;
+    fgWalkResult walkResult = pThis->optVNConstantPropCurStmt(pData->block, pData->stmt, *ppTree, &changed);
+    if (!changed && (walkResult == WALK_CONTINUE))
```
SingleAccretion, line 5918: Is there ever a case when `walkResult` other than `WALK_CONTINUE` will be returned when `changed` is `false`?

src/coreclr/jit/assertionprop.cpp
```diff
@@ -5852,18 +5902,24 @@ void Compiler::optVnNonNullPropCurStmt(BasicBlock* block, Statement* stmt, GenTr
...
+    {
+        walkResult = pThis->optVNGenericPropCurStmt(pData->block, pData->stmt, *ppTree, &changed);
```
SingleAccretion, line 5920: `changed` is not used after this point, so does `optVNGenericPropCurStmt` need to set it?

SingleAccretion (commented):
Structure LGTM, let's see what TP diffs say.

src/coreclr/jit/assertionprop.cpp
EgorBo: not now, tried to make it more agile, probably can convert to an assert for now untill it's changed (for TP)

src/coreclr/jit/assertionprop.cpp
EgorBo: made it for consistency, no strong opinion on it, can delete in sake of TP

EgorBo:
Closing, diffs are too small to worse tp impact

Issue #92586, opened 2023-09-25
JIT: LSRA loses track of anti-preferences


jakobbotsch:
LSRA does not carry any explicit anti preference information for intervals, they are all essentially merged into `Interval::registerPreferences`, with additional special cased behavior through `Interval::preferCalleeSave`.
The result is that LSRA may quickly lose track of anti preferences resulting from kills in the live range of an interval. Consider the following:
```csharp
[MethodImpl(MethodImplOptions.NoInlining)]
private static unsafe void Foo(long rcx, long rdx, ref long p)
{
    long v1 = rcx;
    long v2 = rdx;
    p = p < 0 ? v1 : v2;
    Bar(v2, v1);
}

[MethodImpl(MethodImplOptions.NoInlining)]
private static void Bar(long rcx, long rdx)
{
}
```

We generate:
```asm
; Assembly listing for method Program:Foo(long,long,byref) (FullOpts)
; Emitting BLENDED_CODE for X64 with AVX - Windows
; FullOpts code
; optimized code
; rsp based frame
; partially interruptible
; No PGO data
; Final local variable assignments
;
;  V00 arg0         [V00,T01] (  3,  3   )    long  ->  rcx         single-def
;  V01 arg1         [V01,T02] (  3,  3   )    long  ->  rdx         single-def
;  V02 arg2         [V02,T00] (  3,  3   )   byref  ->   r8         single-def
;  V03 loc0         [V03,T04] (  3,  3   )    long  ->  [rsp+0x20]  spill-single-def
;  V04 loc1         [V04,T05] (  3,  3   )    long  ->  rdx         single-def
;  V05 OutArgs      [V05    ] (  1,  1   )  struct (32) [rsp+0x00]  do-not-enreg[XS] addr-exposed "OutgoingArgSpace"
;* V06 tmp1         [V06    ] (  0,  0   )   byref  ->  zero-ref   
;  V07 tmp2         [V07,T03] (  3,  3   )   byref  ->   r8         single-def
;  V08 tmp3         [V08,T06] (  2,  2   )    long  ->  rax        
;
; Lcl frame size = 40

G_M47833_IG01:  ;; offset=0x0000
       sub      rsp, 40
						;; size=4 bbWeight=1 PerfScore 0.25
G_M47833_IG02:  ;; offset=0x0004
       mov      qword ptr [rsp+0x20], rcx
       cmp      qword ptr [r8], 0
       mov      rax, rdx
       cmovl    rax, rcx
[...]

PR #92709, opened 2023-09-27
LSRA: Honor the register aversion decision during allocation

kunalspathak:
Taken from idea by @jakobbotsch in https://github.com/dotnet/runtime/issues/92586#issuecomment-1733667033 and slightly tweaked it.


src/coreclr/jit/lsra.cpp
```diff
@@ -11519,6 +11519,14 @@ void LinearScan::RegisterSelection::reset(Interval* interval, RefPosition* refPo
     candidates  = refPosition->registerAssignment;
     preferences = currentInterval->registerPreferences;
 
+    regMaskTP updatedPreferences = (preferences & ~currentInterval->registerAversion);
+    if (updatedPreferences != RBM_NONE)
+    {
+        // registerAversion is the best-effort opportunity. Do not update the preference
+        // if registerAversion contains all the registers present in preferences.
+        preferences = updatedPreferences;
+    }
```
jakobbotsch, line 11528: This defeats a lot of the purpose of having this -- I would expect the anti preferences to be more important than the preferences as allocating an anti-preferenced register is (always?) going to result in a spill, while allocating an unpreferred register is just going to result in a reg-reg move. Does this PR with this change fix my example in #92586?

kunalspathak: The reason we can get to `(preferences & ~currentInterval->antiPreferences) == RBM_NONE` that your patch has is because during `mergeRegisterPreferences()` we do not stop the update from happening if the `registerPreferences` contains registers we want to avoid. The right way to do this is to in addition of tracking `registerAversion` make sure `registerPreferences` is up to date with that information throughout the allocation, not just do it during allocation of that interval/RefPosition. Reason being, we check if the interval's assigned physical register (`physReg`) is already part of `registerPreferences` and if yes, we skip the allocation and just assign the same register at the RefPosition. This happens regardless of if that register is present in `registerAversion` mask. By making sure we never put register in `registerPreferences` that are already in `registerAversion` we avoid cases like following:

For `V01`, if we do not keep `registerPreferences` up to date, we end up with:
```
Interval  1: (V01) ref RefPositions {#0@0 #3@9 #18@35 #40@53} physReg:rdx Preferences=[rdx r8] Aversions=[rcx rdx]
```

When we allocate for `RefPosition #0`, we see `rdx` is in `registerPreferences` and assign it.

image

By keeping things up to date, we get nice diff for such scenarios:

image

Now, I still have a check of `RBM_NONE` which I ideally won't to get rid of and just add an `assert(updatedPreferences != RBM_NONE)`, but I need to see if there are any corner JitStressRegs cases where this is not correct.

> Does this PR with this change fix my example in https://github.com/dotnet/runtime/issues/92586?

It does:
```asm
       mov      rax, rcx
       cmp      qword ptr [r8], 0
       mov      rcx, rdx
       cmovl    rcx, rax
       mov      qword ptr [r8], rcx
       mov      rcx, rdx
       mov      rdx, rax
       call     [helloworld._92586:Bar(long,long)]
       nop
```

jakobbotsch: > during mergeRegisterPreferences() we do not stop the update from happening if the registerPreferences contains registers we want to avoid

IMO that behavior is fine and better than trying to keep `registerPreferences` as a mix of preferences and anti preferences during build, where both preferences and anti preferences may be found in any order. It seems simpler to me to only combine `preferences` and `antiPreferences` at the last step right before we need it (in effect making both sets "add-only" during the build step). But maybe it complicates things during allocation stage given that we also want to update the preferences there.

> Now, I still have a check of RBM_NONE which I ideally won't to get rid of and just add an assert(updatedPreferences != RBM_NONE), but I need to see if there are any corner JitStressRegs cases where this is not correct.

I suppose it depends on what the model is... for example, I would think that `preferences & ~antiPreferences == RBM_NONE` just means that at some point the interval would prefer to be in registers that at other points the interval would prefer not to be in. That seems like a perfectly reasonable situation that can occur.

> It does:

With the change from https://github.com/dotnet/runtime/pull/92709/commits/985c96a8e19c80f8a6bbe832030ee623f7e0878a I also think that this case shouldn't occur anymore, but the comment in `reset` is confusing to me since both preferences and anti preferences are best-effort and I think that anti preferences are more important.

We should check memory impact of this change too since adding a field to `Interval` is pretty much equivalent to adding a field to `GenTree`.

kunalspathak: > We should check memory impact of this change too since adding a field to Interval is pretty much equivalent to adding a field to GenTree.

Can you remind me how to do it? I know that we spit that information as part of JITDump, but how do we get the aggregate data?

jakobbotsch: Flip the define here to enable memory stats in release builds:
https://github.com/dotnet/runtime/blob/dd8e9aa8b5cee831ed3885921b79baddc40c2184/src/coreclr/jit/jit.h#L511

Then set `DOTNET_JitMemStats=1` and run superpmi.exe with release JIT for some collection (without parallelization). It should output the aggregate results at the end.

kunalspathak: For benchmarks.run collection. Let's see how much asmdiff improvements we see and should justify the cost.

```diff
-          LSRA_Interval |  123169280 |   2.67%
+         LSRA_Interval |  135486208 |   2.92%
```

jakobbotsch: I'm personally all for taking the memory hit to improve LSRA...

kunalspathak: Alright, so I have updated the logic to what you have proposed, but we still need to make sure that we honor the `aversion` even while updating the preferences as done in [here](https://github.com/dotnet/runtime/pull/92709/files#diff-f7c3782214008ef38ca061ec749219208db22fb6f7608182695c74913765d529R2329-R2335). Doing that gives us some more improvements as seen below for real-world win/x64 collection:

```
[14:12:51] 126 contexts with diffs (69 improvements, 16 regressions, 41 same size)
[14:12:51]   -272/+36 bytes
```

jakobbotsch: Nice. There seems to be a somewhat significant TP hit, especially in MinOpts... is it just from the change in `reset`?

It would also be good to check the memory impact for MinOpts.

kunalspathak: > There seems to be a somewhat significant TP hit, especially in MinOpts

It is impacted more because of changes in `mergePreferences` which is called both by `associateRefPositionWithInterval` and `updateRegisterPreferences`. I could avoid calling `updateRegisterPreferences` at 2 places where we update `registerAversion`, but not sure if it is worth spreading the logic at other places.

```
?associateRefPosWithInterval@LinearScan@@AEAAXPEAVRefPosition@@@Z                    : 13402194 : +5.14%  : 58.86% : +0.0829%
??$select@$0A@@RegisterSelection@LinearScan@@QEAA_KPEAVInterval@@PEAVRefPosition@@@Z : 6813596  : +0.76%  : 29.93% : +0.0421%
?newInterval@LinearScan@@AEAAPEAVInterval@@W4var_types@@@Z                           : 2273723  : +2.56%  : 9.99%  : +0.0141%
?updateRegisterPreferences@Interval@@QEAAX_K@Z                                       : 228420   : +35.73% : 1.00%  : +0.0014%
?mergeRegisterPreferences@Interval@@QEAAX_K@Z                                        : 38714    : NA      : 0.17%  : +0.0002%
```


asp.net collection tier0 MinOpts collection of windows/x64:

```diff
-         LSRA_Interval |  181897840 |   7.39%
+         LSRA_Interval |  200087624 |   8.07%
```

jakobbotsch (approved):
LGTM. The TP cost / memory impact is perhaps a bit high for the diffs we're getting, but given that linux-x64 host sees TP improvements that might not reflect much. I'll leave it up to you.

I think there's some follow-up simplifications we can make, e.g. `mergeRegisterPreferences` tries to guess when the mask represents aversions/anti-preferences. I don't think we need that kind of guessing anymore, so likely we can simplify a lot of stuff there.

kunalspathak:
> I think there's some follow-up simplifications we can make, e.g. `mergeRegisterPreferences` tries to guess when the mask represents aversions/anti-preferences. I don't think we need that kind of guessing anymore, so likely we can simplify a lot of stuff there.

Given that this PR is a bit old, I would like to do that as a follow-up PR. I need to see how other minor changes that I called out in https://github.com/dotnet/runtime/pull/92709#discussion_r1386849360 could change things.

PR #92711, opened 2023-09-27
[wasm] Supress policheck warning in blazor-sample

radekdoulik:
Replace the offending part in the layout name. I think the suit-spade is false positive, I used just sp in place of spade to silence it.


src/mono/sample/wasm/blazor-frame/wwwroot/css/bootstrap-icons/bootstrap-icons.json
```diff
@@ -1181,8 +1181,8 @@
   "suit-diamond": 62876,
   "suit-heart-fill": 62877,
   "suit-heart": 62878,
-  "suit-spade-fill": 62879,
-  "suit-spade": 62880,
+  "suit-sp-fill": 62879,
```
SamMonoRT, line 1184: I suggest update to spade-symbol instead of spade. Also update heart diamond etc to add '-symbol' to it to make it uniform. Then if it creates a new issue (after you merge), we can close as By Design.

SamMonoRT (approved):

PR #92712, opened 2023-09-27
Fix LLVMAOT Mono runtime variant official build to produce correctly named runtime packs

akoeplinger:
In https://github.com/dotnet/runtime/commit/75ee623b8f0350a4b4be86fa71745a74beb059d1 the condition in `src/installer/pkg/sfx/Microsoft.NETCore.App/Microsoft.NETCore.App.Runtime.props` got changed from checking `MonoBundleLLVMOptimizer` to `MonoAOTEnableLLVM` but we weren't setting that property in runtime-official.yml so both jobs produced runtime packs with the same suffix, resulting in the artifact uploads randomly overwriting each other.

Triggered a test official build here: https://dev.azure.com/dnceng/internal/_build/results?buildId=2277856&view=results

### DIFF
--- a/eng/pipelines/runtime-llvm.yml
+++ b/eng/pipelines/runtime-llvm.yml
@@ -119,7 +119,7 @@ extends:
             testGroup: innerloop
             nameSuffix: AllSubsets_Mono_LLVMAOT
             buildArgs: -s mono+libs+host+packs -c $(_BuildConfig)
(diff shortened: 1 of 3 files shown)


steveisok (approved):

steveisok:
Merging early / on red to help unblock the official build

PR #92715, opened 2023-09-27
Vectorize TensorPrimitives.ConvertToHalf

stephentoub:

### DIFF
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
@@ -26,10 +26,301 @@ namespace System.Numerics.Tensors
                 ThrowHelper.ThrowArgument_DestinationTooShort();
             }
 
-            for (int i = 0; i < source.Length; i++)
+            ref float sourceRef = ref MemoryMarshal.GetReference(source);
+            ref ushort destinationRef = ref Unsafe.As<Half, ushort>(ref MemoryMarshal.GetReference(destination));
+            int i = 0, twoVectorsFromEnd;
+
+#if NET8_0_OR_GREATER
+            if (Vector512.IsHardwareAccelerated)
             {
-                destination[i] = (Half)source[i];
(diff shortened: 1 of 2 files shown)


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -26,10 +26,301 @@ public static void ConvertToHalf(ReadOnlySpan<float> source, Span<Half> destinat
...
+
+            while (i < source.Length)
+            {
+                Unsafe.Add(ref destinationRef, i) = BitConverter.HalfToUInt16Bits((Half)Unsafe.Add(ref sourceRef, i));
+                i++;
+            }
+
+            // This implements a vectorized version of the `explicit operator Half(float value) operator`.
+            // See detailed description of the algorithm used here:
+            //     https://github.com/dotnet/runtime/blob/ca8d6f0420096831766ec11c7d400e4f7ccc7a34/src/libraries/System.Private.CoreLib/src/System/Half.cs#L606-L714
```
EgorBo, line 134: Presuambly, we don't support that but x64/arm should already have something to accelerate it, like VCVTNEPS2BF16

EgorBo: I filed an issue for scalars on arm64: https://github.com/dotnet/runtime/issues/92718 for vectors if I understand it correctly it needs SVE. Not sure about X64, if modern cpus already support AVX512_BF16 we can go ahead and file an api proposal? @tannergooding and then use here

tannergooding: Yeah, no acceleration today. It's been put off since it likely requires `TYP_HALF` in the JIT and potentially some answers around what it means for `Vector.IsHardwareAccelerated` to report true, while `Vector<Half>` is not accelerated on most hardware.

There's a few different ways to think about it, and we do have `Vector<T>.IsSupported`, but it might end up inconsistent with how other types work.

tannergooding (approved):

PR #92717, opened 2023-09-27
JIT: Promote size-wise improvements in physical promotion

jakobbotsch:
I hit the following case:
```
Evaluating access byref @000
  Single write-back cost: 3
  Write backs: 0
  Read backs: 0
  Estimated cycle improvement: 0 cycles per invocation
  Estimated size improvement: 2 bytes
Disqualifying replacement
```

These cases happen when the blocks that have candidates for promotion in them have bbWeight equal to 0.

If we estimate a size improvement without a cycle improvement it still makes sense to promote a replacement. More generally, a large size improvement can make up for a small cycle regression, so add a heuristic similar to the existing one for this. I've set it to be quite conservative: we require 100 bytes of size improvement before we allow 1 cycle of regression. This is enough to handle the common case where the cycle improvement is 0 due to the bbWeight = 0.

### DIFF
--- a/src/coreclr/jit/promotion.cpp
+++ b/src/coreclr/jit/promotion.cpp
@@ -802,14 +802,27 @@ public:
         if ((cycleImprovementPerInvoc > 0) &&
             ((cycleImprovementPerInvoc * ALLOWED_SIZE_REGRESSION_PER_CYCLE_IMPROVEMENT) >= -sizeImprovement))
         {
-            JITDUMP("  Promoting replacement\n\n");
+            JITDUMP("  Promoting replacement (cycle improvement)\n\n");
+            return true;
+        }
+
+        // Similarly, even for a cycle-wise regression, if we see a large size
+        // wise improvement we may want to promote. The main case is where all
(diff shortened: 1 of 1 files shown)


jakobbotsch:
I looked at the win-x86 regressions. The diff is very skewed as a result of a single 50 byte regression in `System.IO.StreamWriter:Flush` -- that context is present in the collection 1190 times, so the regression is repeated that many times. (as a side note: for the context I looked at we have `fgCalledCount = 3` and mark some blocks as rarely run based on this very thin PGO data). The regression itself is that we promote but then end up keeping the promoted fields on the stack -- the fields are used around some `Debug.Assert` string interpolation that wouldn't be present in release SPCs.

jakobbotsch:

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=421028&view=ms.vss-build-web.run-extensions-tab). See my comment above for the reason for the win-x86 regressions... They are overstated and I think we can ignore them.

BruceForstall (approved):

PR #92719, opened 2023-09-27
WIP: [ios][tests] Trim on helix for libraries and functional tests on CI

steveisok:
Work in conjunction with https://github.com/dotnet/runtime/pull/92645. Instead of trimming on build machine, do it on helix.


kotlarmilos:
I'm currently working on trimming for the runtime tests. It might be good to wait for this to be merged before proceeding. Assuming that the publish is done on build machine, how do you plan to invoke the trimmer?

I tried to hook it into the pipeline after the build target, which is the default one for the runtime tests, but it seems the `PrepareForILLink` has an indirect dependency on one of the publish targets (`_ComputeAssembliesToPostprocessOnPublish`).

PR #92724, opened 2023-09-27
 Report Requires diagnostics from dataflow analyzer

sbomer:
This moves the logic for generating diagnostics about RequiresUnreferencedCode,
RequiresAssemblyFiles, and RequiresDynamicCode attributes from the separate
Requires* analyzers into the DynamicallyAccessedMembersAnalyzer.

This includes logic for warning on access to attributed members, not the warnings
about mismatching attributes on base/override methods. The override validation
logic is still handled by the respective analyzers.

The DynamicallyAccessedMembersAnalyzer is now turned on by any of the settings
EnableTrimAnalyzer, EnableAotAnalyzer, or EnableSingleFileAnalyzer.

A new type RequiresAnalyzerContext is used to cache per-compilation state of
the analyzers, to avoid recomputing it for each trim analysis pattern that
might potentially produce a warning. The context only stores state for enabled
analyzers, based on the MSBuild property values.

There are a few minor differences in the warning behavior with this change:
- Implicit calls to annotated base ctors warn because they are visible in the CFG
- Unused local functions aren't discovered by the analysis, so don't warn
  (matching linker/AOT behavior)


src/tools/illink/src/ILLink.RoslynAnalyzer/RequiresAnalyzerBase.cs
```diff
@@ -54,63 +55,11 @@ public override void Initialize (AnalysisContext context)
...
-					var objectCreation = (IObjectCreationOperation) operationContext.Operation;
-					var ctor = objectCreation.Constructor;
-					if (ctor is not null) {
-						CheckCalledMember (operationContext, ctor, incompatibleMembers);
-					}
-				}, OperationKind.ObjectCreation);
-
 				context.RegisterOperationAction (operationContext => {
 					var fieldReference = (IFieldReferenceOperation) operationContext.Operation;
 					CheckCalledMember (operationContext, fieldReference.Field, incompatibleMembers);
```
vitek-karas, line 60: This seems to be the last peice which needs CheckCalledMember (and all the infra that brings in). Would it make sense to move this to the DAM analyzer as well? I wonder how this plays with the rest of it. Don't we see field references in code in assignments and such anyway?

src/tools/illink/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -70,11 +70,16 @@ public sealed class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
...
+		protected override bool CreateSpecialIncompatibleMembersDiagnostic (
+			IOperation operation,
+			ImmutableArray<ISymbol> specialIncompatibleMembers,
+			ISymbol member,
+			out Diagnostic? incompatibleMembersDiagnostic)
 		{
+			incompatibleMembersDiagnostic = null;
 			// Some RUC-annotated APIs are intrinsically handled by the trimmer
 			if (member is IMethodSymbol method && Intrinsics.GetIntrinsicIdForMethod (new MethodProxy (method)) != IntrinsicId.None) {
 				return true;
```
vitek-karas, line 85: This is unfortunate - that said we have a similar hack in the trimmer. I guess improvement in the future - how to handle RUC and similar correctly with the intrinsics handling.

vitek-karas (approved):
Looks good - the only problem I see is the potentially problematic suppression of warnings due to RUC/RDC/RAF in presense of intrinsic handling of AOT specific behaviors. But I don't think we have such a case yet, so it doesn't block functionality... yet.

src/tools/illink/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
vitek-karas: That is more the point of the comment here: https://github.com/dotnet/runtime/pull/92724#discussion_r1355251455

This one is about the fact that we have a "External" way of suppressing RUC warnings on intrinsics. In here, we check if a method is intrinsic and then trust the other analyzer it will do its job instead. In linker we use this giant enum which describes which pattern the warning came from, and then based on that we suppress the "general" RUC processing and rely on the method body scanner instead.

NativeAOT is the only implementation where this weird approach is not used - there it's all part of the main "handle call" functionality (still not ideal where it is, but it's much better), and we don't need a way to determine if something will be handled by the intrinsics or other mechanics to decide if we should apply RUC logic or not.

This is definitely not something to fix in this PR though.

src/tools/illink/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisFieldAccessPattern.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices.ComTypes;
+using ILLink.Shared;
+using ILLink.Shared.DataFlow;
+using ILLink.Shared.TrimAnalysis;
+using ILLink.Shared.TypeSystemProxy;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Operations;
+using Microsoft.CodeAnalysis.Diagnostics;
+using MultiValue = ILLink.Shared.DataFlow.ValueSet<ILLink.Shared.DataFlow.SingleValue>;
```
vitek-karas, line 16: Nit - these seem excessive - could be probably trimmed down.

src/tools/illink/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimAnalysisReflectionAccessPattern.cs
```diff
@@ -34,23 +35,23 @@ namespace ILLink.RoslynAnalyzer.TrimAnalysis
...
 		}
 
-		public TrimAnalysisReflectionAccessPattern Merge (ValueSetLattice<SingleValue> lattice, TrimAnalysisReflectionAccessPattern other)
-		{
-			Debug.Assert (Operation == other.Operation);
-			Debug.Assert (SymbolEqualityComparer.Default.Equals (ReferencedMethod, other.ReferencedMethod));
-
-			return new TrimAnalysisReflectionAccessPattern (
-				ReferencedMethod,
-				lattice.Meet (Instance, other.Instance),
```
vitek-karas, line 44: The Instance property doesn't seem to be used - we should remove it. It was the sole reason for having a Merge... without it there's really no reason to have Merge.

vitek-karas (approved):

vitek-karas:
Looks good - thanks!

PR #92726, opened 2023-09-27
Revert "Remove Latin1CharSearchValues (#91884)"

MihaZupan:
This reverts #91884.
Contributes to https://github.com/dotnet/perf-autofiling-issues/issues/21818
Contributes to https://github.com/dotnet/perf-autofiling-issues/issues/21828

In that PR I measured that the throughput of `IndexOfAny` is practically the same between `Latin1CharSearchValues` and its alternative, `ProbabilisticMap`. While that does seem to be the case, `IndexOfAnyExcept` can be much worse with the prob map, as we currently fall back to an `O(n * m)` loop.

I'm reverting the change until we have better alternatives in place.
Runtimes without explicit ISA support, such as some Mono variants with plain `Vector128`, should benefit from changes like https://github.com/dotnet/runtime/pull/92680 for ASCII search values (assuming optimal codegen).
We can also look into using a faster than `O(n * m)` fallback for the prob map case, but that may come with other downsides like a higher memory footprint (e.g. allocating a bitmap for all values).

cc: @stephentoub

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
+++ b/src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
@@ -337,7 +337,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CompareInfo.iOS.cs" Condition="'$(IsiOSLike)' == 'true'" />
     <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CompareOptions.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CultureData.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CultureData.Browser.cs" Condition="'$(TargetsBrowser)' == 'true'" />
+    <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CultureData.Browser.cs" Condition="'$(TargetsBrowser)' == 'true'"/>
     <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CultureData.Icu.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CultureData.iOS.cs" Condition="'$(IsiOSLike)' == 'true'" />
     <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CultureData.Nls.cs" />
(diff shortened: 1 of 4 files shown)


src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
```diff
@@ -337,7 +337,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CompareInfo.iOS.cs" Condition="'$(IsiOSLike)' == 'true'" />
     <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CompareOptions.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CultureData.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CultureData.Browser.cs" Condition="'$(TargetsBrowser)' == 'true'" />
+    <Compile Include="$(MSBuildThisFileDirectory)System\Globalization\CultureData.Browser.cs" Condition="'$(TargetsBrowser)' == 'true'"/>
```
stephentoub, line 340: You could probably avoid reverting this line :)

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
```diff
@@ -1305,7 +1306,7 @@
     <Compile Include="$(CommonPath)Interop\Browser\Interop.CompareInfo.cs" Condition="'$(TargetsBrowser)' == 'true'">
       <Link>Common\Interop\Interop.CompareInfo.cs</Link>
     </Compile>
-    <Compile Include="$(CommonPath)Interop\Browser\Interop.Calendar.cs" Condition="'$(TargetsBrowser)' == 'true'">
+     <Compile Include="$(CommonPath)Interop\Browser\Interop.Calendar.cs" Condition="'$(TargetsBrowser)' == 'true'">
```
stephentoub, line 1309: Ditto

stephentoub (approved):

fanyang-mono (approved):

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
MihaZupan: Right, I just did a lazy `git revert` 😄
Re-reverted these now.

PR #92731, opened 2023-09-27
Correctly set sendTrustList flag when saving credentials to cache

rzikm:
This PR fixes a case when custom trust with `sendInHandshake` is provided to `SslStreamCertificateContext` on the server side. We incorrectly cache the credentials as if the `sendInHandshake` was set to `false`, leading to the allocation of new SCHANNEL credentials object for each incoming connection.


wfurt (approved):
LGTM

avparuch (approved):

rzikm:
CI failure is unrelated.

PR #92739, opened 2023-09-27
Add more static preinitialization support

### LINKED ISSUE #78681: Allow modeling ref fields in the class constructor interpreter

MichalStrehovsky:
Resolves #78681 (or "there's nothing else we'd be willing to do for it"). This is the rest of what I implemented trying to get `SearchValues.Create(someVeryLongString)` to preinitialize. It doesn't actually enable more `SearchValues` scenarios beyond those that already worked because I eventually hit codepaths that would require us to implement hardware intrinsics support in the interpreter. The `SearchValues` scenarios that we do support were implemented in #92470 and #92666. But since I already implemented this, here it is; maybe it will be useful for something else:

* Support for modelling `Span<X>`, including creating it from stackalloc
* Support for math/comparisons with native int
* `Unsafe.Add`

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
@@ -755,15 +755,17 @@ namespace ILCompiler
 
                     case ILOpcode.conv_i:
                     case ILOpcode.conv_u:
+                    case ILOpcode.conv_i1:
                     case ILOpcode.conv_i2:
                     case ILOpcode.conv_i4:
                     case ILOpcode.conv_i8:
+                    case ILOpcode.conv_u1:
                     case ILOpcode.conv_u2:
                     case ILOpcode.conv_u4:
                     case ILOpcode.conv_u8:
                         {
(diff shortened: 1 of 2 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
```diff
@@ -707,6 +759,7 @@ private Status TryScanMethod(MethodIL methodIL, Value[] parameters, Stack<Method
                     case ILOpcode.conv_i4:
                     case ILOpcode.conv_i8:
                     case ILOpcode.conv_u2:
+                    case ILOpcode.conv_u4:
```
jkotas, line 762: It looks like that places like this one will end up handling all integer types in the fullness of time. Would it make sense to add them all and be done with it instead of adding one at a time?

(IIRC, we have discussed in one of the earlier preinit PRs.)

jkotas (approved):

PR #89587, opened 2023-07-27
Make config binding gen incremental

layomia:
Fixes https://github.com/dotnet/runtime/issues/83534. https://github.com/dotnet/runtime/pull/89226 is a related change wrt. perf.

This PR aims to solve the incremental generation perf issue in the config binding generator, so that diffs in input compilations that don't change prior effective inputs the generator (or the resulting output) do not cause regeneration.

It also moves Roslyn helpers, generator helpers, and the `TypeRef` identity token to common locations to be shared by the JSON and config binding generators.

PR #92740, opened 2023-09-27
Reenable pragma suppressions for config binder gen (+ JSON)

layomia:
- Prototyping a fix. They were disabled in https://github.com/dotnet/runtime/pull/89587, which (temporarily, hopefully) sacrificed the UX over favor of incremental generation.
- Workaround till we implement a separate analyzer to handle diagnostics (https://github.com/dotnet/runtime/issues/92509).
- Proposing to address in 8.0 for config binder since all diagnostics are issued in source.

### DIFF
--- a/src/libraries/Common/src/SourceGenerators/DiagnosticInfo.cs
+++ b/src/libraries/Common/src/SourceGenerators/DiagnosticInfo.cs
@@ -4,7 +4,9 @@
 using System;
 using System.Linq;
 using System.Numerics.Hashing;
+using System.Reflection;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
 
 namespace SourceGenerators;
 
@@ -14,6 +16,9 @@ namespace SourceGenerators;
 /// </summary>
 internal readonly struct DiagnosticInfo : IEquatable<DiagnosticInfo>
 {
+    private static readonly Lazy<CSharpSyntaxTree?> s_dummySyntaxTree = new Lazy<CSharpSyntaxTree?>(GetDummySyntaxTree);
+    private static readonly Lazy<FieldInfo?> s_sourceTreeBackingFieldInfo = new Lazy<FieldInfo?>(GetSourceTreeBackingFieldInfo);
+
     public DiagnosticDescriptor Descriptor { get; private init; }
     public object?[] MessageArgs { get; private init; }
     public Location? Location { get; private init; }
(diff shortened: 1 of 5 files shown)


src/libraries/Common/src/SourceGenerators/DiagnosticInfo.cs
```diff
@@ -30,8 +35,35 @@ public static DiagnosticInfo Create(DiagnosticDescriptor descriptor, Location? l
...
+        {
+            if (sourceLocation is null)
+            {
+                return null;
+            }
+
+            Location trimmedLocation = Location.Create(sourceLocation.SourceTree?.FilePath ?? "", sourceLocation.SourceSpan, sourceLocation.GetLineSpan().Span);
+
+            if (sourceLocation.IsInSource &&
+                !trimmedLocation.IsInSource &&
```
eiriktsarpalis, line 48: What determines whether `trimmedLocation.IsInSource` is true?

src/libraries/Common/src/SourceGenerators/DiagnosticInfo.cs
```diff
@@ -30,8 +35,35 @@ public static DiagnosticInfo Create(DiagnosticDescriptor descriptor, Location? l
...
+                s_sourceTreeBackingFieldInfo.Value is FieldInfo sourceTreeField &&
+                s_dummySyntaxTree.Value is CSharpSyntaxTree syntaxTree)
```
eiriktsarpalis, line 50: If the instance is only used to check if the `DummySyntaxTree` type exists, why not just store a `Type?` instance?

src/libraries/Common/src/SourceGenerators/DiagnosticInfo.cs
```diff
@@ -57,4 +89,34 @@ public Diagnostic CreateDiagnostic()
...
     }
+
+    private static FieldInfo? GetSourceTreeBackingFieldInfo()
+    {
+        FieldInfo? info;
+
+        try
+        {
+            FieldInfo[] fields = typeof(Location).GetFields(BindingFlags.NonPublic);
+            info = typeof(Location).GetField("<SourceTree>k__BackingField", BindingFlags.NonPublic | BindingFlags.Instance);
```
eiriktsarpalis, line 100: I'm not comfortable taking a dependency on this type of implementation detail, particularly in product code. We should instead work with @dotnet/roslyn to come up with a better solution for storing `Location` instances in incremental values that doesn't encapsulate the `Compilation` instance.

cc @stephentoub

PR #76904, opened 2022-10-11
Check if interval interferes with the call

kunalspathak:
After building intervals, check if any of them interferes with a call. In other words, scan all the intervals and check if there are refpositions that are created before a call and used after the call. If yes, then update such intervals preference to use `callee-save` registers. To accomplish this, I keep track of `LsraLocation` just before adding `RefTypeKill` and also record the killmask. After building intervals, go through all the call locations and see if any of the interval collides and if yes, then set its `preferCalleeSave = true`.

I am assuming a TP cost to do this because we will be doing non-linear search, but hopefully, it will be not as bad, and we do get some good results out of it. I do add various checks to skip iteration if we know that scanning it won't make sense. We cannot do this at the time of building interval, because we do not have information about the future uses of the interval under consideration.

Related conversation: https://github.com/dotnet/runtime/pull/75565#issuecomment-1275032313

PR #81242, opened 2023-01-26
JIT: Preference LIR edges away from killed registers

jakobbotsch:
We already seem to be able to handle this via allocation heuristics in most cases, but this gets a few more.

PR #92744, opened 2023-09-27
LSRA: PreferCalleeSave for non-localVar intervals

kunalspathak:
Idea from https://github.com/dotnet/runtime/pull/76904, but just with lower TP cost.


jakobbotsch:
Is this different from #81242?

kunalspathak:
> Is this different from #81242?

Didn't realize you had this already. Wondering why this was not merged. The idea is essentially same, but the way I was doing it in https://github.com/dotnet/runtime/pull/76904 was TP heavy, so this should have lower TP impact. Compared to what you are doing, we might over mark `preferCalleeSave` to some intervals that has `lastUse` before the call and new definition after the call. Probably ok to not mark such intervals as `preferCalleeSave`? 

```
RefTypeDef
RefTypeUse
RefTypeUse
RefTypeKill
RefTypeKill
...
RefTypeDef
RefTypeUse
```

jakobbotsch:
> Wondering why this was not merged.

I think I wanted to investigate why this was a problem for floats but not for integer intervals, but I never got around to doing that. It's sort of odd that even without this preferencing we can allocate integer intervals without hitting the issue. Also IIRC the diffs were somewhat mixed.

This fix seems fine to me too, but the nice part about doing it from `buildKillPositionsForNode` is that we handle the local intervals and the LIR temp intervals from the same place. Not sure which one is best TP wise, though.

kunalspathak:
The results are mixed, and for some collection worse. There is around 0.5% regression in TP. 

image

The approach I am taking here is if an interval ever interferes with call, I am marking it as `preferCalleeSave` (which is too aggressive)  vs. what @jakobbotsch is doing in https://github.com/dotnet/runtime/pull/81242 to just mark intervals whose definition is before call and use is after the call (LIR temps). I looked at the regressions and they are mostly coming from the extra prolog/epilog save restore of callee-save registers because they are preferred more now. We might see benefits of eliminating the memory access from the inner blocks from this and move that cost to prolog/epilog.

image

I tried something in the middle, where we mark those intervals to be `preferCalleeSave` whose def (and zero or more uses) is before the call and `lastUse` is after the call. That too gives mixed results, where the regressions are mostly from prolog/epilog save/restore.

PR #92748, opened 2023-09-28
Account port number already included within server string

### LINKED ISSUE #59701: Linux: Instantiating a LdapConnection with a server string including a port number in .Net6.0-rc1 System.DirectoryServices.Protocols throws later an error at a SendRequest
### Description
If you instantiate a LdapConnection with a string consisting from a server name and a port number e.g.
```
LdapConnection ldap = new LdapConnection("ldap.example.com:636");
```  
then a later method call like 
```
DirectoryResponse dResp = ldap.SendRequest(request);
```
yields
```
Exception has occurred: CLR/System.DirectoryServices.Protocols.LdapException
An unhandled exception of type 'System.DirectoryServices.Protocols.LdapException' occurred in System.DirectoryServices.Protocols.dll: 'A bad parameter was passed to a routine.'
   at System.DirectoryServices.Protocols.LdapConnection.Connect()
   at System.DirectoryServices.Protocols.LdapConnection.SendRequestHelper(DirectoryRequest request, Int32& messageID)
   at System.DirectoryServices.Protocols.LdapConnection.SendRequest(DirectoryRequest request, TimeSpan requestTimeout)
   at System.DirectoryServices.Protocols.LdapConnection.SendRequest(DirectoryRequest request)
```

However when instantiating the LdapConnection in this manner 
```
LdapConnection ldap = new LdapConnection(new LdapDirectoryIdentifier("ldap.example.com",636));
```
then same SendRequest works without exceptions.

Documentation at

[.Net Documentation System_DirectoryServices_Protocols_LdapConnection__ctor_System_String_](https://docs.microsoft.com/en-us/dotnet/api/system.directoryservices.protocols.ldapconnection.-ctor?view=dotnet-plat-ext-5.0#System_DirectoryServices_Protocols_LdapConnection__ctor_System_String_)

states that

server String
A string specifying the server which can be a domain name, LDAP server name or dotted strings representing the IP address of the LDAP server. Optionally, this parameter may also include a port number, separated from the right end of the string by a colon (:).

### Configuration

.Net6-rc1 Linux Ubuntu 18.04 x86-64

### Regression?

Don't know.

buyaa-n:
On Linux: Instantiating a LdapConnection with a server string including a port number throws exception when call `LdapConnection.SendRequest(...)`, this works fine on Windows. Turns out on Linux it always appends the port number to the server string when populating the uri string, fixed this issue

Fixes https://github.com/dotnet/runtime/issues/59701

### DIFF
--- a/src/libraries/System.DirectoryServices.Protocols/tests/DirectoryServicesProtocolsTests.cs
+++ b/src/libraries/System.DirectoryServices.Protocols/tests/DirectoryServicesProtocolsTests.cs
@@ -75,31 +75,10 @@ namespace System.DirectoryServices.Protocols.Tests
             // Does not throw
         }
 
-
-        private LdapConnection GetConnectionWithServerNameAndPort()
-        {
-            LdapDirectoryIdentifier directoryIdentifier = new LdapDirectoryIdentifier($"{LdapConfiguration.Configuration.ServerName}:{LdapConfiguration.Configuration.Port}", true, false);
-            NetworkCredential credential = new NetworkCredential(LdapConfiguration.Configuration.UserName, LdapConfiguration.Configuration.Password);
-
-            LdapConnection connection = new LdapConnection(directoryIdentifier, credential)
-            {
-                AuthType = AuthType.Basic
-            };
-
-            // Set server protocol before bind; OpenLDAP servers default
-            // to LDAP v2, which we do not support, and will return LDAP_PROTOCOL_ERROR
-            connection.SessionOptions.ProtocolVersion = 3;
-            connection.SessionOptions.SecureSocketLayer = LdapConfiguration.Configuration.UseTls;
-            connection.Bind();
-
-            connection.Timeout = new TimeSpan(0, 3, 0);
-            return connection;
-        }
-
         [ConditionalFact(nameof(IsLdapConfigurationExist))]
         public void TestServerWithPortNumber()
         {
-            using LdapConnection connection = GetConnectionWithServerNameAndPort();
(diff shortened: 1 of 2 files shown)


src/libraries/System.DirectoryServices.Protocols/tests/DirectoryServicesProtocolsTests.cs
```diff
@@ -774,13 +786,25 @@ private SearchResultEntry SearchUser(LdapConnection connection, string rootDn, s
             return null;
         }
 
+        private LdapConnection GetConnection(string server)
+        {
+            LdapDirectoryIdentifier directoryIdentifier = new LdapDirectoryIdentifier(server, true, false);
```
danmoseley, line 791: Nit, when passing true or false to a method it's nice to name the parameters so it's way to see what they are

src/libraries/System.DirectoryServices.Protocols/tests/DirectoryServicesProtocolsTests.cs
```diff
@@ -74,6 +74,18 @@ public void TestUnavailableNonCriticalExtension()
             _ = (SearchResponse) connection.SendRequest(searchRequest);
             // Does not throw
         }
+
+        [ConditionalFact(nameof(IsLdapConfigurationExist))]
```
danmoseley, line 78:
```suggestion
        [ConditionalFact(nameof(LdapConfigurationExists))]
```

danmoseley: (Exists already)

danmoseley (approved):

PR #92749, opened 2023-09-28
Flesh out TensorPrimitives XML docs

stephentoub:

### DIFF
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
@@ -146,7 +146,7 @@ namespace System.Numerics.Tensors
         /// If a value is equal to <see cref="float.NaN"/>, the result stored into the corresponding destination location is also NaN.
         /// </para>
         /// <para>
-        /// The angles in x must be in radians. Multiply by <see cref="MathF.PI"/>/180 to convert degrees to radians.
+        /// The angles in x must be in radians. Use <see cref="M:System.Single.DegreesToRadians"/> or multiply by <see cref="MathF.PI"/>/180 to convert degrees to radians.
         /// </para>
         /// <para>
         /// This method may call into the underlying C runtime or employ instructions specific to the current architecture. Exact results may differ between different
@@ -1076,7 +1076,7 @@ namespace System.Numerics.Tensors
         /// the corresponding destination location is set to that value.
         /// </para>
         /// <para>
-        /// The angles in x must be in radians. Multiply by <see cref="MathF.PI"/>/180 to convert degrees to radians.
+        /// The angles in x must be in radians. Use <see cref="M:System.Single.DegreesToRadians"/> or multiply by <see cref="MathF.PI"/>/180 to convert degrees to radians.
         /// </para>
         /// <para>
         /// This method may call into the underlying C runtime or employ instructions specific to the current architecture. Exact results may differ between different
@@ -1252,7 +1252,7 @@ namespace System.Numerics.Tensors
         /// If a value is equal to <see cref="float.NaN"/>, the corresponding destination location is set to NaN.
         /// </para>
         /// <para>
-        /// The angles in x must be in radians. Multiply by <see cref="MathF.PI"/>/180 to convert degrees to radians.
+        /// The angles in x must be in radians. Use <see cref="M:System.Single.DegreesToRadians"/> or multiply by <see cref="MathF.PI"/>/180 to convert degrees to radians.
         /// </para>
         /// <para>
         /// This method may call into the underlying C runtime or employ instructions specific to the current architecture. Exact results may differ between different
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
@@ -19,14 +19,10 @@ namespace System.Numerics.Tensors
         /// </para>
         /// <para>
         /// If a value is equal to <see cref="float.NegativeInfinity"/> or <see cref="float.PositiveInfinity"/>, the result stored into the corresponding destination location is set to <see cref="float.PositiveInfinity"/>.
-        /// If a value is equal to <see cref="float.NaN"/>, the result stored into the corresponding destination location is set to <see cref="float.NaN"/>.
+        /// If a value is equal to <see cref="float.NaN"/>, the result stored into the corresponding destination location is the original NaN value with the sign bit removed.
         /// </para>
         /// <para>
         /// <paramref name="x"/> and <paramref name="destination"/> may not overlap; if they do, behavior is undefined.
-        /// </para>
-        /// <para>
-        /// This method may call into the underlying C runtime or employ instructions specific to the current architecture. Exact results may differ between different
-        /// operating systems or architectures.
         /// </para>
         /// </remarks>
         public static void Abs(ReadOnlySpan<float> x, Span<float> destination) =>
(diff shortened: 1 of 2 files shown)


stephentoub:
Note that in the doc comments, I noted that it's not safe to use an input span as an output span. I think that's a restriction we should remove, so that you can easily do in-place manipulations. We actually do already support this on some operations, but I left the comment on them for consistency. Now that the remainder processing support was merged, we can update the rest of the operations to make this safe, and then we can update those docs to clarify that overlaps between inputs and destination is safe iff they start at the same memory location.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -6,72 +6,177 @@ namespace System.Numerics.Tensors
...
+        /// </para>
+        /// <para>
+        /// If a value is equal to <see cref="float.NegativeInfinity"/> or <see cref="float.PositiveInfinity"/>, the result stored into the corresponding destination location is set to <see cref="float.PositiveInfinity"/>.
+        /// If a value is equal to <see cref="float.NaN"/>, the result stored into the corresponding destination location is set to <see cref="float.NaN"/>.
+        /// </para>
+        /// <para>
+        /// <paramref name="x"/> and <paramref name="destination"/> may not overlap; if they do, behavior is undefined.
+        /// </para>
+        /// <para>
+        /// This method may call into the underlying C runtime or employ instructions specific to the current architecture. Exact results may differ between different
```
tannergooding, line 28: Abs is one we can (and should) guarantee is the same everywhere (and which already should be consistent)

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -6,72 +6,177 @@ namespace System.Numerics.Tensors
...
+        /// <remarks>
+        /// <para>
+        /// This method effectively computes <c><paramref name="destination" />[i] = MathF.Abs(<paramref name="x" />[i])</c>.
+        /// </para>
+        /// <para>
+        /// The absolute value of a <see cref="float"/> is its numeric value without its sign. For example, the absolute value of both 1.2e-03 and -1.2e03 is 1.2e03.
+        /// </para>
+        /// <para>
+        /// If a value is equal to <see cref="float.NegativeInfinity"/> or <see cref="float.PositiveInfinity"/>, the result stored into the corresponding destination location is set to <see cref="float.PositiveInfinity"/>.
+        /// If a value is equal to <see cref="float.NaN"/>, the result stored into the corresponding destination location is set to <see cref="float.NaN"/>.
```
tannergooding, line 22: Abs is a case where the spec requires NaN return the same bit pattern with the sign bit removed. It is one of the few cases we aren’t allowed to normalize

tannergooding: Notably, there is never a reason to normalize for Abs either, since it’s more efficient to simply strip the sign bit and that’s what all platforms do already

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -6,72 +6,177 @@ namespace System.Numerics.Tensors
...
+        /// This method effectively computes <c><paramref name="destination" />[i] = <paramref name="x" />[i] + <paramref name="y" />[i]</c>.
+        /// </para>
+        /// <para>
+        /// <paramref name="x"/> and <paramref name="y"/> may overlap, but neither may overlap with <paramref name="destination"/>; if they do, behavior is undefined.
+        /// </para>
+        /// <para>
+        /// If either of the element-wise input values is <see cref="float.NaN"/>, the resulting element-wise value is <see cref="float.NaN"/>.
+        /// </para>
+        /// <para>
+        /// This method may call into the underlying C runtime or employ instructions specific to the current architecture. Exact results may differ between different
```
tannergooding, line 52: Add, Subtract, Multiply, and Divide are likewise cases where all platforms should be trivially deterministic with no differences in result

we should never defer to the C runtime for them

tannergooding: What might vary is the exact bit pattern for some NaN, as given NaN + NaN, it’s not guaranteed which NaN you get back

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -6,72 +6,177 @@ namespace System.Numerics.Tensors
...
+        /// This method effectively computes <c><paramref name="destination" />[i] = (<paramref name="x" />[i] + <paramref name="y" />) * <paramref name="multiplier" />[i]</c>.
+        /// </para>
+        /// <para>
+        /// <paramref name="x"/> and <paramref name="multiplier"/> may overlap, but neither may overlap with <paramref name="destination"/>; if they do, behavior is undefined.
+        /// </para>
+        /// <para>
+        /// If any of the element-wise input values is <see cref="float.NaN"/>, the resulting element-wise value is <see cref="float.NaN"/>.
+        /// </para>
+        /// <para>
+        /// This method may call into the underlying C runtime or employ instructions specific to the current architecture. Exact results may differ between different
```
tannergooding, line 150: AddMultiply and MultiplyAdd are cases we should determine if we want to guarantee determinism or if we want to give the flexibility to use a fused operation internally, this would give a more accurate result across both operations and be within 1 ULP of the two separate operations

I’d lean towards determinism and that we should expose a separate MultiplyAddEstimate method which allows the optimization via explicit opt-in, as we’ve done elsewhere for math apis

stephentoub: Should I interpret this as "we should revert https://github.com/dotnet/runtime/pull/92205"?

tannergooding: CosineSimilarity is one of the more complex APIs where we can’t easily guarantee it today, so I think it’s fine there.

That’s different from the more primitive API, like MultiplyAdd, where it’s trivial to make the determinism guarantee and where we don’t really lose any perf by doing so.

tannergooding: Long term, I’d like us to move towards cross platform determinism everywhere, but that’s a long goal and not something to worry about for this release where differences are already guaranteed to exist.

So I think we only need to be concerned with the places where we can trivially provide perf and determinism

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -6,72 +6,153 @@ namespace System.Numerics.Tensors
...
+        /// </para>
+        /// <para>
+        /// <paramref name="x"/> and <paramref name="destination"/> may not overlap; if they do, behavior is undefined.
+        /// </para>
+        /// <para>
+        /// If a value is equal to <see cref="float.NegativeInfinity"/> or <see cref="float.PositiveInfinity"/>, the result stored into the corresponding destination location is set to <see cref="float.PositiveInfinity"/>.
+        /// If a value is equal to <see cref="float.NaN"/>, the result stored into the corresponding destination location is also NaN.
+        /// </para>
+        /// <para>
+        /// The angles in x must be in radians. Multiply by <see cref="MathF.PI"/>/180 to convert degrees to radians.
```
tannergooding, line 149: In .NET 8, we have the new float.DegreesToRadians and inverse api

tannergooding (approved):

Issue #65864, opened 2022-02-24
[wasm][debugger] Todo in expression evaluation


radical:
- [ ] incorrect argument types in method calls
- [ ] surface exceptions thrown from property getters, and method calls
- [ ] surface a NRE for `foo.NullObject.Prop`
- [ ] Calling methods with `params`
- [ ] methods with named arguments
- [ ] Extension methods
- [ ] Go through [SyntaxKind](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.syntaxkind) and cover unsupported cases of arguments passed to methods.

thaystg:
This is fixed by the new debugger.

PR #92754, opened 2023-09-28
[wasm][debugger] Support passing negative/positive numbers to methods.

ilonatommy:
Passing negative number generates different `ExpressionSyntax` type than passing a positive number. Because we classify method arguments by casting them to Syntax types, we were not handling negative case at all (throwing `Unable to evaluate method 'GetDefaultAndRequiredParam'. Unable to write into binary writer, not recognized expression type: PrefixUnaryExpressionSyntax"`).
Same with positive numbers, passing `+2` should behave the same as passing `2`.
`SyntaxKind` has a lot of possible values (https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.syntaxkind) and we probably are not handling all of them yet. Logging this into https://github.com/dotnet/runtime/issues/65864.

### DIFF
--- a/src/mono/wasm/debugger/BrowserDebugProxy/MonoSDBHelper.cs
+++ b/src/mono/wasm/debugger/BrowserDebugProxy/MonoSDBHelper.cs
@@ -757,7 +757,7 @@ namespace Microsoft.WebAssembly.Diagnostics
                     Write(ElementType.U2, (int)us);
                     break;
                 case short s:
-                    Write(ElementType.I2, (uint)s);
+                    Write(ElementType.I2, (uint)s  * coeff);
                     break;
                 case uint ui:
                     Write(ElementType.U4, ui);
--- a/src/mono/wasm/debugger/BrowserDebugProxy/MonoSDBHelper.cs
+++ b/src/mono/wasm/debugger/BrowserDebugProxy/MonoSDBHelper.cs
@@ -650,51 +650,37 @@ namespace Microsoft.WebAssembly.Diagnostics
             return false;
         }
 
+        public bool WriteConst(PrefixUnaryExpressionSyntax constValue)
+        {
+            switch (constValue.Kind())
+            {
+                case SyntaxKind.UnaryMinusExpression:
+                {
+                    switch (constValue.Operand)
+                    {
+                        case LiteralExpressionSyntax les:
+                        {
+                            return WriteNumber(les.Token.Value, convertToNegative: true);
+                        }
+                        default:
+                        {
+                            // not supported yet
+                            break;
+                        }
+                    }
+                    break;
+                }
+            }
+            return false;
+        }
+
         public async Task<bool> WriteConst(LiteralExpressionSyntax constValue, MonoSDBHelper SdbHelper, CancellationToken token)
         {
             switch (constValue.Kind())
             {
                 case SyntaxKind.NumericLiteralExpression:
                 {
-                    switch (constValue.Token.Value) {
-                        case double d:
-                            Write(ElementType.R8, d);
-                            break;
-                        case float f:
-                            Write(ElementType.R4, f);
-                            break;
-                        case long l:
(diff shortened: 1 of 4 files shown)


thaystg (approved):

src/mono/wasm/debugger/BrowserDebugProxy/MonoSDBHelper.cs
```diff
@@ -728,6 +730,51 @@ public async Task<bool> WriteConst(LiteralExpressionSyntax constValue, MonoSDBHe
...
+                    Write(ElementType.U1, (int)b);
+                    break;
+                case sbyte sb:
+                    Write(ElementType.I1, (uint)sb);
+                    break;
+                case ushort us:
+                    Write(ElementType.U2, (int)us);
+                    break;
+                case short s:
+                    Write(ElementType.I2, (uint)s);
```
radical, line 760: Isn't `* coeff` needed here too?

src/mono/wasm/debugger/BrowserDebugProxy/MonoSDBHelper.cs
```diff
@@ -728,6 +730,51 @@ public async Task<bool> WriteConst(LiteralExpressionSyntax constValue, MonoSDBHe
             return false;
         }
 
+        public bool WriteNumber(object number, bool convertToNegative=false)
```
radical, line 733: Why is `convertToNegative` needed here, instead of changing the `number` as needed when calling this method?

radical: nvm!

src/mono/wasm/debugger/BrowserDebugProxy/MonoSDBHelper.cs
radical: And add the test cases for all the types which support a negative number.

ilonatommy: Good catch

src/mono/wasm/debugger/DebuggerTestSuite/EvaluateOnCallFrame2Tests.cs
```diff
@@ -499,6 +499,8 @@ public async Task EvaluateLocalObjectFromAssemblyNotRelatedButLoaded()
                     ("test.GetDefaultAndRequiredParam(3, 2)", TNumber(5)),
                     ("test.GetDefaultAndRequiredParam(3, +2)", TNumber(5)),
                     ("test.GetDefaultAndRequiredParam(3, -2)", TNumber(1)),
+                    ("test.GetDefaultAndRequiredParam(-123l, -1.1f)", TNumber("-124.1", isDecimal: true)), // long, float
+                    ("test.GetDefaultAndRequiredParam(-0.23)", TNumber("-32768.23", isDecimal: true)), // double, short
```
radical, line 503: This is missing the `short` argument.

radical (approved):
Other than the feedback on the test, LGTM 👍

src/mono/wasm/debugger/DebuggerTestSuite/EvaluateOnCallFrame2Tests.cs
radical:
```csharp
                    ("test.GetDefaultAndRequiredParam(3, +2)", TNumber(5)),
                    ("test.GetDefaultAndRequiredParam(3, -2)", TNumber(1)),
                    ("test.GetDefaultAndRequiredParam(-123l, -1.1f)", TNumber("-124.1", isDecimal: true)),
```

- These 3 are not testing optional parameters.
- Also, for the cases that call `GetDefaultAndRequiredParam` - add comments mentioning what optional type is being tested.

Do the changes in this PR also affect passing a negative number as an argument to a method?

PR #92755, opened 2023-09-28
Use UnsafeAccessor in JSHostImplementation instead of reflection

akoeplinger:
Resolves a FIXME comment that points to a closed issue.

### DIFF
--- a/src/libraries/System.Runtime.InteropServices.JavaScript/src/System/Runtime/InteropServices/JavaScript/JSHostImplementation.cs
+++ b/src/libraries/System.Runtime.InteropServices.JavaScript/src/System/Runtime/InteropServices/JavaScript/JSHostImplementation.cs
@@ -327,31 +327,20 @@ namespace System.Runtime.InteropServices.JavaScript
             ThreadJsOwnedObjects.Clear();
         }
 
-        private static FieldInfo? thread_id_Field;
-        private static FieldInfo? external_eventloop_Field;
+        [UnsafeAccessor(UnsafeAccessorKind.Field, Name = "external_eventloop")]
+        private static extern ref bool GetThreadExternalEventloop(Thread @this);
 
-        // FIXME: after https://github.com/dotnet/runtime/issues/86040 replace with
-        // [UnsafeAccessor(UnsafeAccessorKind.Field, Name="external_eventloop")]
-        // static extern ref bool ThreadExternalEventloop(Thread @this);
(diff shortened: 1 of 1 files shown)


jkotas (approved):

src/libraries/System.Runtime.InteropServices.JavaScript/src/System/Runtime/InteropServices/JavaScript/JSHostImplementation.cs
```diff
@@ -327,31 +327,20 @@ public static void UninstallWebWorkerInterop()
...
+        private static extern ref long GetThreadNativeThreadId(Thread @this);
+
         public static IntPtr GetNativeThreadId()
         {
-            if (thread_id_Field == null)
-            {
-                thread_id_Field = typeof(Thread).GetField("thread_id", BindingFlags.NonPublic | BindingFlags.Instance)!;
-            }
-            return (int)(long)thread_id_Field.GetValue(Thread.CurrentThread)!;
+            return (int)GetThreadNativeThreadId(Thread.CurrentThread);
```
MichalPetryka, line 343:
```suggestion
            return (nint)GetThreadNativeThreadId(Thread.CurrentThread);
```
Should this truncate the top bits of the ID?

akoeplinger: hmm good question, I wonder why we didn't use `long` here instead, @pavelsavara ?

stephentoub (approved):

src/libraries/System.Runtime.InteropServices.JavaScript/src/System/Runtime/InteropServices/JavaScript/JSHostImplementation.cs
pavelsavara: Because int64 doesn't exist on wasm imports/exports. 
our wasm is 32bit platform, so this is int32 size.
If you make it int64, you need to marshal it as `BigInt` (which is heap object) or as double (which is int52 at most).
Do we expect to have more IDs than 2^31 ?

I don't remember the details of when we are passing this to/from JS at the moment.
Feel free to add another FIXME for this and I will re-evaluate that later.

akoeplinger: ok, thanks for the details, I think it is fine for now :)

PR #92760, opened 2023-09-28
[llvm] Avoid zero extending non-negative constant array indexes, its …

vargaz:
…not needed and it prevents abcrem from working.


vargaz:
This fixes compilation time problems on the HugeArray test.

@steveisok

lambdageek (approved):

lambdageek:
How's the compilation time on HugeArray1.cs ? that one is about 9 times bigger than HugeArray.cs

vargaz:
1m42s with this change.

SamMonoRT:
Should we backport for 8.0 GA ?

vargaz:
Don't think it is that important.

PR #92764, opened 2023-09-28
Delete misc unnecessary code

jkotas:


src/coreclr/vm/ceeload.h
```diff
@@ -79,21 +79,6 @@ class EnCEEClassData;
...
-#if defined(HOST_AMD64)
-#define NATIVE_SYMBOL_READER_DLL W("Microsoft.DiaSymReader.Native.amd64.dll")
-#elif defined(HOST_X86)
-#define NATIVE_SYMBOL_READER_DLL W("Microsoft.DiaSymReader.Native.x86.dll")
-#elif defined(HOST_ARM)
-#define NATIVE_SYMBOL_READER_DLL W("Microsoft.DiaSymReader.Native.arm.dll")
-#elif defined(HOST_ARM64)
-#define NATIVE_SYMBOL_READER_DLL W("Microsoft.DiaSymReader.Native.arm64.dll")
-#elif defined(HOST_LOONGARCH64)
-#define NATIVE_SYMBOL_READER_DLL W("Microsoft.DiaSymReader.Native.loongarch64.dll")
```
jkotas, line 92: This is Windows specific, so it only needs to have architectures supported by Windows. Also, this definition does not need to be in a header file.

src/coreclr/nativeaot/Runtime/inc/MethodTable.h
```diff
@@ -260,22 +260,6 @@ class MethodTable
         return (m_uFlags & OptionalFieldsFlag) != 0;
     }
 
-    bool IsEquivalentTo(MethodTable * pOtherEEType)
```
jkotas, line 263: @MichalStrehovsky Applies https://github.com/dotnet/runtime/pull/91921 in the unmanaged runtime.

stephentoub (approved):

PR #92672, opened 2023-09-26
Update TensorPrimitives aggregations to vectorize handling of remaining elements

stephentoub:
Contributes to https://github.com/dotnet/runtime/issues/92219

PR #92765, opened 2023-09-28
Avoid large table as part of TensorPrimitives remainder mask

stephentoub:
As discussed in https://github.com/dotnet/runtime/pull/92672#pullrequestreview-1648754274


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
```diff
@@ -640,11 +640,14 @@ private static unsafe Vector<float> LoadRemainderMaskSingleVector(int validItems
...
-            return AsVector(
-                ref Unsafe.As<uint, float>(ref MemoryMarshal.GetReference(RemainderUInt32Mask_16x16)),
-                (validItems * 16) + (16 - Vector<float>.Count));
+            return Vector.ConditionalSelect(
+                (Vector<float>)Vector.GreaterThan(AsVector(ref MemoryMarshal.GetReference<int>(s_0through15)), new Vector<int>(Vector<int>.Count - 1 - validItems)),
+                ~Vector<float>.Zero,
+                Vector<float>.Zero);
         }
 
+        private static readonly int[] s_0through15 = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
```
EgorBo, line 649: Does it need to be a `ROS<int>` ?

EgorBo (approved):

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
stephentoub: This is in the netstandard build where CreateSpan doesn't exist. As a result, if this returns a span, Roslyn will generate a lazily-initialized array.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1404,22 +1404,25 @@ private static Vector512<float> FusedMultiplyAdd(Vector512<float> x, Vector512<f
...
+                Vector256<float>.AllBitsSet,
+                Vector256<float>.Zero);
 
 #if NET8_0_OR_GREATER
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private static unsafe Vector512<float> LoadRemainderMaskSingleVector512(int validItems) =>
-            Vector512.LoadUnsafe(
-                ref Unsafe.As<uint, float>(ref MemoryMarshal.GetReference(RemainderUInt32Mask_16x16)),
-                (uint)(validItems * 16)); // all sixteen floats in the row
+            Vector512.ConditionalSelect(
```
EgorBo, line 1422: So for AVX512 it'd looked like 
```cs
var mmask16 mask = (1 << n) - 1;
var maskedAnd = Avx512.And(vec1, vec2, mask);
```
but we decided not to expose mask registers so if we expose these as a public API, we can intrinsify at leas the AVX512 version to be cheap 🙂

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1404,22 +1404,25 @@ private static Vector512<float> FusedMultiplyAdd(Vector512<float> x, Vector512<f
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private static unsafe Vector128<float> LoadRemainderMaskSingleVector128(int validItems) =>
-            Vector128.LoadUnsafe(
-                ref Unsafe.As<uint, float>(ref MemoryMarshal.GetReference(RemainderUInt32Mask_16x16)),
-                (uint)((validItems * 16) + 12)); // last four floats in the row
+            Vector128.ConditionalSelect(
+                Vector128.LessThan(Vector128.Create(3, 2, 1, 0), Vector128.Create(validItems)).AsSingle(),
+                Vector128<float>.AllBitsSet,
+                Vector128<float>.Zero);
```
tannergooding, line 1410: This is going to be slower and result in a 128-bit constant (or more for other sizes) emitted at runtime per method this gets inlined into.

I think the table is ultimately better and we can get rid of it longer term using a proper JIT intrinsic.

-----------

If we really don't want the table, then achieving this with just a broadcast + comparison should be sufficient, since `LessThan` already produces a per-element mask. You can use `GreaterThanOrEqual` if you need the mask inverted

stephentoub: Ok, I'll stick with the table. I was on the fence anyway. Just don't like its bulkiness.

tannergooding: That is, since `TrailingMask` wants to handle only trailing elements, it will skip processing the first `n` items that have already been processed.

So if `2` items remain, you have `3 < 2, 2 < 2, 1 < 2, 0 < 2` which produces `Zero, Zero, AllBitsSet, AllBitsSet` already

Issue #92766, opened 2023-09-28
JIT: LowerHWIntrinsicCreate can create illegal IR


jakobbotsch:
```scala
LowerHWIntrinsicCreate before:
N002 (  1,  1) [000017] -----------                   t17 =    LCL_VAR   byref  V01 arg0         u:1 $81
                                                            ┌──▌  t17    byref  
N004 (  3,  4) [000054] -c---------                   t54 = ▌  LEA(b+4)  byref 
                                                            ┌──▌  t54    byref  
N005 (  4,  3) [000019] n---GO-----                   t19 = ▌  IND       float  <l:$203, c:$202>
                                                            ┌──▌  t3     float  
               [000089] ---XG------                   t89 = ▌  HWINTRINSIC simd16 float CreateScalarUnsafe
               [000090] -c---------                   t90 =    CNS_INT   int    1
N003 (  1,  2) [000004] -----------                    t4 =    CNS_DBL   float  0.0000000000000000 $240
                                                            ┌──▌  t89    simd16 
                                                            ├──▌  t90    int    
                                                            ├──▌  t4     float  
               [000091] ---XG------                   t91 = ▌  HWINTRINSIC simd16 float Insert
               [000092] -c---------                   t92 =    CNS_INT   int    2
N004 (  1,  2) [000005] -----------                    t5 =    CNS_DBL   float  0.0000000000000000 $240
                                                            ┌──▌  t91    simd16 
                                                            ├──▌  t92    int    
                                                            ├──▌  t5     float  
               [000093] ---XG------                   t93 = ▌  HWINTRINSIC simd16 float Insert
               [000094] -c---------                   t94 =    CNS_INT   int    3
N005 (  1,  2) [000006] -----------                    t6 =    CNS_DBL   float  0.0000000000000000 $240
                                                            ┌──▌  t93    simd16 
                                                            ├──▌  t94    int    
                                                            ├──▌  t6     float  
N006 (  7,  9) [000011] ---XG------                   t11 = ▌  HWINTRINSIC simd16 float Insert $2c0
                                                            ┌──▌  t11    simd16 
N007 ( 11, 12) [000014] DA-XG------                         ▌  STORE_LCL_VAR simd16 V07 tmp4         d:1 $1c1
               [000085] -----------                            IL_OFFSET void   INLRT @ 0x03D[--]
N001 (  1,  2) [000016] -----------                   t16 =    CNS_DBL   float  0.0000000000000000 $240
[...]

Issue #92537, opened 2023-09-23
JIT: Invalid smnegl transformation


jakobbotsch:
```csharp
// Generated by Fuzzlyn v1.6 on 2023-09-23 15:15:18
// Run on Arm64 Windows
// Seed: 17929576410580344290
// Reduced from 324.8 KiB to 0.4 KiB in 00:04:43
// Debug: Outputs -2472766155
// Release: Outputs 1822201141

public class C0
{
    public uint F4;
    public C0(uint f4)
    {
        F4 = f4;
    }
}

public class Program
{
    public static C0 s_8 = new C0(2472766155U);
    public static ushort[] s_39 = new ushort[]{1};
    public static void Main()
    {
        uint vr2 = s_8.F4;
        long vr0 = -(s_39[0] * (long)vr2);
        System.Console.WriteLine(vr0);
    }
}
```

The codegen is
```asm
G_M27646_IG03:  ;; offset=0x001C
            movz    x0, #0x1D78      // data for Program:s_8
            movk    x0, #0x7A00 LSL #16
            movk    x0, #657 LSL #32
            ldr     x1, [x0]
            ldr     w1, [x1, #0x08]
            ldr     x0, [x0, #0x08]
            ldr     w2, [x0, #0x08]
            cmp     w2, #0
            bls     G_M27646_IG06
            ldrh    w0, [x0, #0x10]
            smnegl  x0, w0, w1
            movz    x1, #0x5B90      // code for System.Console:WriteLine(long)
            movk    x1, #0x8153 LSL #16
            movk    x1, #0x7FFE LSL #32
            ldr     x1, [x1]
            blr     x1
```

This looks like a regression introduced by #91886, cc @c272 @kunalspathak

jakobbotsch:
@c272 @kunalspathak Can you please take a look at this? We hit tons of bad code gen problems in the fuzzers because of this issue.

kunalspathak:
Will take a look.

PR #92768, opened 2023-09-28
JIT: Reorder indirs on arm64 to make them amenable to ldp optimization

### LINKED ISSUE #64815: Arm64: Evaluate if it is possible to combine subsequent field loads in a single load
Evaluate to see how feasible it would be to combine loads of subsequent fields using `ldp` instead of loading them separately during the use. It cannot be considered as a peep-hole optimization, but an analysis is needed in earlier phases to check around for consecutive field loads and if found one, combine them to a single load.

```c#

class Body { public double x, y, z, vx, vy, vz, mass; }
...
foreach (var b in bodies) {
    b.x += dt * b.vx; b.y += dt * b.vy; b.z += dt * b.vz;
}
```

Below code is generated for the loop that deals with multiplication of `double`.
- `#1` can be combined into `ldp dX, dY, [x3, #8] `
- `#2` can be combined into `ldp dX, dY, [x3, #32] `
- `#3` can be combined into `stp dX, dY, [x3, #8] `

```asm
G_M56457_IG05:              ;; offset=0114H
        D37D7C43          ubfiz   x3, x2, #3, #32
        91004063          add     x3, x3, #16
        F8636803          ldr     x3, [x0, x3]
        FD400470          ldr     d16, [x3,#8]   ; <-- #1
        FD401071          ldr     d17, [x3,#32]  ; <-- #2
        1E710811          fmul    d17, d0, d17
        1E712A10          fadd    d16, d16, d17
        FD000470          str     d16, [x3,#8]   ; <-- #3
        FD400870          ldr     d16, [x3,#16]  ; <-- #1
        FD401471          ldr     d17, [x3,#40]  ; <-- #2
        1E710811          fmul    d17, d0, d17
        1E712A10          fadd    d16, d16, d17
        FD000870          str     d16, [x3,#16]  ; <-- #3
        FD400C70          ldr     d16, [x3,#24]
        FD401871          ldr     d17, [x3,#48]
        1E710811          fmul    d17, d0, d17
        1E712A10          fadd    d16, d16, d17
        FD000C70          str     d16, [x3,#24]
        11000442          add     w2, w2, #1
        6B02003F          cmp     w1, w2
        54FFFD8C          bgt     G_M56457_IG05

```

Reference: https://godbolt.org/z/9jY5hYnoa

category:implementation
theme:codegen
skill-level:intermediate
cost:medium
impact:medium

jakobbotsch:
This PR adds an ARM64 optimization in lowering that tries to reorder indirections to be near each other, making them amenable to `ldp` optimization done in the backend.

Example:
```csharp
public struct Point
{
    public long X, Y;

    public static bool Equals(in Point left, in Point right)
    {
        return (left.X == right.X) & (left.Y == right.Y);
    }
}
```

Before:
```asm
G_M28859_IG02:  ;; offset=0x0008
            ldr     x2, [x0]
            ldr     x3, [x1]
            ldr     x0, [x0, #0x08]
            ldr     x1, [x1, #0x08]
            cmp     x2, x3
            ccmp    x0, x1, 0, eq
            cset    x0, eq
```

After:
```asm
G_M28859_IG02:  ;; offset=0x0008
            ldp     x2, x0, [x0]
            ldp     x3, x1, [x1]
            cmp     x2, x3
            ccmp    x0, x1, 0, eq
            cset    x0, eq
```

The transformation works by moving the trees that are not part of the data flow of the second indirection past it (after doing the necessary interference checks to prove whether this reordering is ok). For example, for the case above, the jitdump illustrates what happens:

```scala
[000009] and [000002] are indirs off the same base with offsets +004 and +000
  ..and they are amenable to ldp optimization
  ..and it is close by. Trying to move the following range (where * are nodes part of the data flow):

   N001 (  1,  1) [000000] -----------                    t0 =    LCL_VAR   byref  V00 arg0         u:1 $80
                                                               ┌──▌  t0     byref  
1. N002 (  3,  2) [000002] ---XG------                    t2 = ▌  IND       int    <l:$200, c:$201>
   N003 (  1,  1) [000003] -----------                    t3 =    LCL_VAR   byref  V01 arg1         u:1 $81
                                                               ┌──▌  t3     byref  
   N004 (  3,  2) [000005] ---XG------                    t5 = ▌  IND       int    <l:$202, c:$203>
                                                               ┌──▌  t2     int    
                                                               ├──▌  t5     int    
   N005 ( 10,  5) [000006] ---XG------                    t6 = ▌  EQ        int    <l:$207, c:$206>
*  N006 (  1,  1) [000007] -----------                    t7 =    LCL_VAR   byref  V00 arg0         u:1 (last use) $80
                                                               ┌──▌  t7     byref  
*  N008 (  3,  4) [000018] -c---------                   t18 = ▌  LEA(b+4)  byref 
                                                               ┌──▌  t18    byref  
2. N009 (  4,  3) [000009] n---GO-----                    t9 = ▌  IND       int    <l:$208, c:$209>

Have conservative interference with last indir. Trying a smarter interference check...
Interference checks passed. Moving nodes that are not part of data flow tree

Result:

   N001 (  1,  1) [000000] -----------                    t0 =    LCL_VAR   byref  V00 arg0         u:1 $80
                                                               ┌──▌  t0     byref  
1. N002 (  3,  2) [000002] ---XG------                    t2 = ▌  IND       int    <l:$200, c:$201>
*  N006 (  1,  1) [000007] -----------                    t7 =    LCL_VAR   byref  V00 arg0         u:1 (last use) $80
                                                               ┌──▌  t7     byref  
*  N008 (  3,  4) [000018] -c---------                   t18 = ▌  LEA(b+4)  byref 
                                                               ┌──▌  t18    byref  
2. N009 (  4,  3) [000009] n---GO-----                    t9 = ▌  IND       int    <l:$208, c:$209>
   N003 (  1,  1) [000003] -----------                    t3 =    LCL_VAR   byref  V01 arg1         u:1 $81
                                                               ┌──▌  t3     byref  
   N004 (  3,  2) [000005] ---XG------                    t5 = ▌  IND       int    <l:$202, c:$203>
                                                               ┌──▌  t2     int    
                                                               ├──▌  t5     int    
   N005 ( 10,  5) [000006] ---XG------                    t6 = ▌  EQ        int    <l:$207, c:$206>
```
In this case we needed to reorder the indirection with the `[000005] IND` and `[000006] EQ` nodes.

The optimization also supports reordering stores with loads to unlock opportunities like #64815. To do that there is a (somewhat ad hoc) alias analysis based on `TYP_REF` and locals. For example, for the case in #64815:
```csharp
public class Body { public double x, y, z, vx, vy, vz, mass; }

public static void Foo(double dt, Body[] bodies)
{
    foreach (var b in bodies)
    {
        b.x += dt * b.vx;
        b.y += dt * b.vy;
        b.z += dt * b.vz;
    }
}
```

the jit dump is:
```scala
[000033] and [000029] are indirs off the same base with offsets +040 and +016
  ..but at wrong offset
[000033] and [000021] are indirs off the same base with offsets +040 and +032
  ..and they are amenable to ldp optimization
  ..and it is close by. Trying to move the following range (where * are nodes part of the data flow):

   N006 (  1,  1) [000019] -----------                   t19 =    LCL_VAR   ref    V04 loc2         u:2 <l:$202, c:$142>
                                                               ┌──▌  t19    ref    
   N008 (  3,  4) [000075] -c---------                   t75 = ▌  LEA(b+32) byref 
                                                               ┌──▌  t75    byref  
1. N009 (  4,  3) [000021] n---GO-----                   t21 = ▌  IND       double <l:$484, c:$485>
                                                               ┌──▌  t18    double 
                                                               ├──▌  t21    double 
   N010 ( 10,  8) [000022] ----GO-----                   t22 = ▌  MUL       double <l:$489, c:$488>
                                                               ┌──▌  t17    double 
                                                               ├──▌  t22    double 
   N011 ( 19, 15) [000023] ---XGO-----                   t23 = ▌  ADD       double <l:$48d, c:$48c>
   N012 (  1,  1) [000014] -----------                   t14 =    LCL_VAR   ref    V04 loc2         u:2 <l:$202, c:$142>
                                                               ┌──▌  t14    ref    
   N014 (  3,  4) [000071] -c---------                   t71 = ▌  LEA(b+8)  byref 
                                                               ┌──▌  t71    byref  
                                                               ├──▌  t23    double 
   N015 ( 24, 19) [000025] nA-XGO-----                         ▌  STOREIND  double <l:$206, c:$205>
                  [000116] -----------                            IL_OFFSET void   INLRT @ 0x01F[E-]
   N005 (  1,  1) [000030] -----------                   t30 =    LCL_VAR   double V00 arg0         u:1 $100
*  N006 (  1,  1) [000031] -----------                   t31 =    LCL_VAR   ref    V04 loc2         u:2 <l:$202, c:$142>
                                                               ┌──▌  t31    ref    
*  N008 (  3,  4) [000081] -c---------                   t81 = ▌  LEA(b+40) byref 
                                                               ┌──▌  t81    byref  
2. N009 (  4,  3) [000033] n---GO-----                   t33 = ▌  IND       double <l:$492, c:$493>

Have conservative interference with last indir. Trying a smarter interference check...
Cannot interfere with [000025] since they are off the same local V04 and indir range [040..048) does not interfere with store range [008..016)
Interference checks passed. Moving nodes that are not part of data flow tree

Result:

   N006 (  1,  1) [000019] -----------                   t19 =    LCL_VAR   ref    V04 loc2         u:2 <l:$202, c:$142>
                                                               ┌──▌  t19    ref    
   N008 (  3,  4) [000075] -c---------                   t75 = ▌  LEA(b+32) byref 
                                                               ┌──▌  t75    byref  
1. N009 (  4,  3) [000021] n---GO-----                   t21 = ▌  IND       double <l:$484, c:$485>
*  N006 (  1,  1) [000031] -----------                   t31 =    LCL_VAR   ref    V04 loc2         u:2 <l:$202, c:$142>
                                                               ┌──▌  t31    ref    
*  N008 (  3,  4) [000081] -c---------                   t81 = ▌  LEA(b+40) byref 
                                                               ┌──▌  t81    byref  
2. N009 (  4,  3) [000033] n---GO-----                   t33 = ▌  IND       double <l:$492, c:$493>
                                                               ┌──▌  t18    double 
                                                               ├──▌  t21    double 
   N010 ( 10,  8) [000022] ----GO-----                   t22 = ▌  MUL       double <l:$489, c:$488>
                                                               ┌──▌  t17    double 
                                                               ├──▌  t22    double 
   N011 ( 19, 15) [000023] ---XGO-----                   t23 = ▌  ADD       double <l:$48d, c:$48c>
   N012 (  1,  1) [000014] -----------                   t14 =    LCL_VAR   ref    V04 loc2         u:2 <l:$202, c:$142>
                                                               ┌──▌  t14    ref    
   N014 (  3,  4) [000071] -c---------                   t71 = ▌  LEA(b+8)  byref 
                                                               ┌──▌  t71    byref  
                                                               ├──▌  t23    double 
   N015 ( 24, 19) [000025] nA-XGO-----                         ▌  STOREIND  double <l:$206, c:$205>
                  [000116] -----------                            IL_OFFSET void   INLRT @ 0x01F[E-]
   N005 (  1,  1) [000030] -----------                   t30 =    LCL_VAR   double V00 arg0         u:1 $100
```

In this case we needed to reorder with a much larger range of trees -- the `[000022]` to `[000030]` trees needed to be moved past the indirection. In this case that involves proving that the `[000033]` indir does not interfere with the `[000025]` store.

I want to look into handling `LCL_FLD` and stores (`STOREIND/STORE_LCL_FLD`) as a follow-up.

Fix #64815


kunalspathak:
Can it get the example in https://github.com/dotnet/runtime/issues/64815?

jakobbotsch:
@kunalspathak Sadly no, due to the store -- our interference checking isn't smart enough to realize that
```asm
        FD000470          str     d16, [x3,#8]
        FD400870          ldr     d16, [x3,#16]
```
are ok to reorder. It gives up here:

https://github.com/dotnet/runtime/blob/968ca812ab719737f41e2532c48e4071b1915b7d/src/coreclr/jit/lower.cpp#L8055-L8067

It's probably not that difficult to make this particular check smarter (of course we could also make `InterferesWith` itself smarter, but that's a harder/more expensive)

jakobbotsch:
@kunalspathak I pushed a new commit that makes the interference check more clever. It means we get the case in that issue now, diff:
```diff
@@ -1,20 +1,18 @@
-G_M62947_IG03:  ;; offset=0x0020
+G_M62947_IG03:  ;; offset=0x001C
             ldr     x3, [x0, w1, UXTW #3]
-            ldr     d16, [x3, #0x08]
-            ldr     d17, [x3, #0x20]
-            fmul    d17, d0, d17
-            fadd    d16, d16, d17
+            ldp     d16, d17, [x3, #0x08]
+            ldp     d18, d19, [x3, #0x20]
+            fmul    d18, d0, d18
+            fadd    d16, d16, d18
             str     d16, [x3, #0x08]
-            ldr     d16, [x3, #0x10]
-            ldr     d17, [x3, #0x28]
-            fmul    d17, d0, d17
-            fadd    d16, d16, d17
+            fmul    d16, d0, d19
+            fadd    d16, d17, d16
             str     d16, [x3, #0x10]
             ldr     d16, [x3, #0x18]
             ldr     d17, [x3, #0x30]
             fmul    d17, d0, d17
             fadd    d16, d16, d17
             str     d16, [x3, #0x18]
             add     w1, w1, #1
             cmp     w2, w1
             bgt     G_M62947_IG03
```

jakobbotsch:
Of course we need the `STOREIND` counterpart to also get the `stp` :-)

jakobbotsch:
Some data on why cases get rejected (1 = rejected due to that reason):
```scala
Volatile
     <=          0 ===>   37438 count ( 99% of total)
      1 ..       1 ===>     166 count (100% of total)

Too far
     <=          0 ===>   32988 count ( 88% of total)
      1 ..       1 ===>    4450 count (100% of total)

Interference
     <=          0 ===>   28934 count ( 87% of total)
      1 ..       1 ===>    4054 count (100% of total)
```

This is over libraries_tests.run. Could try to adjust the distance we search to see if we can get some more of those "Too far" cases...

jakobbotsch:
With 16 distance on smoke_tests:
```scala
Volatile
     <=          0 ===>    3829 count ( 99% of total)
      1 ..       1 ===>      10 count (100% of total)

Too far
     <=          0 ===>    3684 count ( 96% of total)
      1 ..       1 ===>     145 count (100% of total)

Interference
     <=          0 ===>    3513 count ( 95% of total)
      1 ..       1 ===>     171 count (100% of total)
```

With 32:
```scala
Too far
     <=          0 ===>    3796 count ( 99% of total)
      1 ..       1 ===>      33 count (100% of total)

Interference
     <=          0 ===>    3543 count ( 93% of total)
      1 ..       1 ===>     253 count (100% of total)
```

So that allowed another 112 cases to pass onwards to the interference analysis, but there we rejected 82 of them... so probably not worth it to increase it.

jakobbotsch:
cc @dotnet/jit-contrib PTAL @kunalspathak @BruceForstall 

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=428690&view=ms.vss-build-web.run-extensions-tab). There is a small TP impact of having to store and search for candidate indirections.

Fuzzlyn failure was #92537. libraries-jitstress failures look like an infra issue and a known issue.

src/coreclr/jit/lower.cpp
```diff
@@ -8179,8 +8185,354 @@ void Lowering::LowerIndir(GenTreeIndir* ind)
...
+//
+// Returns:
+//    True if the optimization was successful.
+//
+bool Lowering::OptimizeForLdp(GenTreeIndir* ind)
+{
+    if (!ind->TypeIs(TYP_INT, TYP_LONG, TYP_DOUBLE, TYP_SIMD16))
+    {
+        return false;
+    }
```
jakobbotsch, line 8217: We need to fix #92766 to add `TYP_FLOAT` here.

EgorBo: Are small types supported?

src/coreclr/jit/lower.cpp
```diff
@@ -8179,8 +8185,354 @@ void Lowering::LowerIndir(GenTreeIndir* ind)
...
+bool Lowering::TryMakeIndirsAdjacent(GenTreeIndir* prevIndir, GenTreeIndir* indir)
+{
+    if ((indir->gtFlags & GTF_IND_VOLATILE) != 0)
+    {
+        JITDUMP("  ..but second indir is volatile\n");
+        return false;
+    }
+
+    GenTree* cur = prevIndir;
+    for (int i = 0; i < 16; i++)
```
EgorBo, line 8296: what is 16?

EgorBo:
Nice diffs!

src/coreclr/jit/lower.cpp
neon-sunset: Perhaps this needs to have `TYP_SIMD8` too? (on a rare occasion there is Vector64-based code for ARM64)

BruceForstall:
Looks like there are some interesting regressions where spilling happens, and where the `ldp` opt. doesn't actually occur.

src/coreclr/jit/lower.cpp
kunalspathak: > Perhaps this needs to have `TYP_SIMD8` too? (on a rare occasion there is Vector64-based code supprted by ARM64)

True, wondering if `TYP_SIMD8` was considered in this list?

src/coreclr/jit/sideeffects.h
```diff
@@ -150,6 +150,7 @@ class AliasSet final
     void AddNode(Compiler* compiler, GenTree* node);
     bool InterferesWith(const AliasSet& other) const;
     bool InterferesWith(const NodeInfo& node) const;
+    bool WritesLocal(unsigned lclNum);
```
BruceForstall, line 153: can it be `const`?

src/coreclr/jit/lower.cpp
```diff
@@ -8179,8 +8185,354 @@ void Lowering::LowerIndir(GenTreeIndir* ind)
...
+{
+    if (!ind->TypeIs(TYP_INT, TYP_LONG, TYP_DOUBLE, TYP_SIMD16))
+    {
+        return false;
+    }
+
+    target_ssize_t offs   = 0;
+    GenTree*       addr   = ind->Addr();
+    FieldSeq*      fldSeq = nullptr;
+    comp->gtPeelOffsets(&addr, &offs, &fldSeq);
```
BruceForstall, line 8222: `fldSeq` is unused here; don't compute it?

src/coreclr/jit/lower.cpp
```diff
@@ -8179,8 +8185,354 @@ void Lowering::LowerIndir(GenTreeIndir* ind)
...
+                    // *(x+4), *(x+8), *(x+0), *(x+12)
+                    prev.Indir = nullptr;
+                    result     = true;
+                    break;
+                }
+                break;
+            }
+            else
+            {
+                JITDUMP("  ..but at wrong offset\n");
```
kunalspathak, line 8267:
```suggestion
                JITDUMP("  ..but at non-adjacent offset\n");
```

src/coreclr/jit/lower.cpp
```diff
@@ -8179,8 +8185,354 @@ void Lowering::LowerIndir(GenTreeIndir* ind)
...
+    {
+        return false;
+    }
+
+    bool result = false;
+
+    SavedIndir newInd;
+    newInd.Indir    = ind;
+    newInd.AddrBase = addr->AsLclVarCommon();
+    newInd.Offset   = offs;
```
BruceForstall, line 8234: Add a convenience constructor?

src/coreclr/jit/lower.cpp
```diff
@@ -8179,8 +8185,354 @@ void Lowering::LowerIndir(GenTreeIndir* ind)
...
+// Arguments:
+//    ind   - Indirection to record and to try to move.
+//    next  - [out] Next node to lower. Only written if the optimization was successful.
+//
+// Returns:
+//    True if the optimization was successful.
+//
+bool Lowering::TryMakeIndirsAdjacent(GenTreeIndir* prevIndir, GenTreeIndir* indir)
+{
+    if ((indir->gtFlags & GTF_IND_VOLATILE) != 0)
```
kunalspathak, line 8289: Is this check needed even for the `prevIndir`?

src/coreclr/jit/lower.cpp
kunalspathak: Write a comment or convert `16` into a variable like `maxDistanceToSearch` to explain the purpose of this loop.

src/coreclr/jit/lower.cpp
```diff
@@ -8179,8 +8185,354 @@ void Lowering::LowerIndir(GenTreeIndir* ind)
...
+    }
+
+    bool result = false;
+
+    SavedIndir newInd;
+    newInd.Indir    = ind;
+    newInd.AddrBase = addr->AsLclVarCommon();
+    newInd.Offset   = offs;
+
+    int maxCount = min(m_blockIndirs.Height(), 8);
```
BruceForstall, line 8236: Why 8? (comment)

src/coreclr/jit/lower.cpp
```diff
@@ -8179,8 +8185,354 @@ void Lowering::LowerIndir(GenTreeIndir* ind)
...
+    MarkTree(indir);
+
+    INDEBUG(dumpWithMarks());
+    JITDUMP("\n");
+
+    m_scratchSideEffects.Clear();
+
+    for (GenTree* cur = prevIndir->gtNext; cur != indir; cur = cur->gtNext)
+    {
+        if (cur->IsCall() || (cur->OperIsStoreBlk() && (cur->AsBlk()->gtBlkOpKind == GenTreeBlk::BlkOpKindHelper)))
```
kunalspathak, line 8349: I am wondering if we should move this check in the `for-loop` above where you check the distance between the `prevIndir` and `indir`.

src/coreclr/jit/lower.cpp
```diff
@@ -8179,8 +8185,354 @@ void Lowering::LowerIndir(GenTreeIndir* ind)
...
+    return result;
+}
+
+//------------------------------------------------------------------------
+// TryMakeIndirsAdjacent: Try to prove that it is legal to move an indirection
+// to be adjacent to a previous indirection. If successful, perform the move.
+//
+// Arguments:
+//    ind   - Indirection to record and to try to move.
+//    next  - [out] Next node to lower. Only written if the optimization was successful.
```
BruceForstall, line 8282: These don't match the function signature

src/coreclr/jit/lower.cpp
BruceForstall: (please comment)

src/coreclr/jit/lower.cpp
```diff
@@ -8179,8 +8185,354 @@ void Lowering::LowerIndir(GenTreeIndir* ind)
...
+        if (cur->IsCall() || (cur->OperIsStoreBlk() && (cur->AsBlk()->gtBlkOpKind == GenTreeBlk::BlkOpKindHelper)))
+        {
+            JITDUMP("[Heuristic] Giving up due to node that kills registers [%06u]\n", Compiler::dspTreeID(cur));
+            UnmarkTree(indir);
+            return false;
+        }
+
+        if ((cur->gtLIRFlags & LIR::Flags::Mark) != 0)
+        {
+            // Part of data flow of 'indir', so we will be moving past this node.
```
kunalspathak, line 8358:
```suggestion
            // Part of data flow of 'indir', so we will be moving it upward past this node.
```

src/coreclr/jit/lower.cpp
```diff
@@ -8179,8 +8185,354 @@ void Lowering::LowerIndir(GenTreeIndir* ind)
...
+            {
+                JITDUMP("Indir [%06u] interferes with [%06u]\n", Compiler::dspTreeID(indir), Compiler::dspTreeID(cur));
+                UnmarkTree(indir);
+                return false;
+            }
+        }
+    }
+
+    JITDUMP("Interference checks passed. Moving nodes that are not part of data flow tree\n\n",
+            Compiler::dspTreeID(indir));
```
BruceForstall, line 8474: The `dspTreeID` isn't displayed (missing `[%06u]`)

src/coreclr/jit/lower.cpp
```diff
@@ -8179,8 +8185,354 @@ void Lowering::LowerIndir(GenTreeIndir* ind)
...
+void Lowering::UnmarkTree(GenTree* node)
+{
+    node->gtLIRFlags &= ~LIR::Flags::Mark;
+    node->VisitOperands([=](GenTree* op) {
+        UnmarkTree(op);
+        return GenTree::VisitResult::Continue;
+    });
+}
+
+#endif
```
BruceForstall, line 8534:
```suggestion
#endif // TARGET_ARM64
```

BruceForstall (commented):
Very cool optimization. Just a few nits/questions/comments.

kunalspathak (changes requested):

src/coreclr/jit/promotiondecomposition.cpp
```diff
@@ -1204,26 +1204,44 @@ void Compiler::gtPeelOffsets(GenTree** addr, target_ssize_t* offset, FieldSeq**
...
+                *fldSeq = m_fieldSeqStore->Append(intCon->gtFieldSeq, *fldSeq);
+                *addr   = op2;
+            }
+            else
+            {
+                break;
+            }
         }
-        else if (op1->IsCnsIntOrI() && !op1->AsIntCon()->IsIconHandle())
+        else if ((*addr)->OperIs(GT_LEA))
```
kunalspathak, line 1235: possibly this is adding to x64 TP. Have it under `#ifdef TARGET_ARM64`?

src/coreclr/jit/lower.cpp
jakobbotsch: > Are small types supported?

Nope, there is no small variant of `ldp` I believe.

> Perhaps this needs to have TYP_SIMD8 too? (on a rare occasion there is Vector64-based code supprted by ARM64)
> 
> True, wondering if TYP_SIMD8 was considered in this list?

Yep, this can have `TYP_SIMD8` -- I forgot about those, thanks for spotting this.

src/coreclr/jit/sideeffects.h
jakobbotsch: Yep, good catch.

src/coreclr/jit/lower.cpp
jakobbotsch: Made it optional.

src/coreclr/jit/lower.cpp
jakobbotsch: Made it optional.

src/coreclr/jit/lower.cpp
jakobbotsch: Added a comment.

src/coreclr/jit/lower.cpp
jakobbotsch: Fixed.

src/coreclr/jit/lower.cpp
jakobbotsch: Seems reasonable -- moved it.

FWIW, we might be able to remove this heuristic after we merge #92744

src/coreclr/jit/lower.cpp
jakobbotsch: Fixed

src/coreclr/jit/promotiondecomposition.cpp
jakobbotsch: I think it's ok to pay a small amount of TP to keep this generic helper consistent across the platforms.

src/coreclr/jit/lower.cpp
jakobbotsch: Added a comment. The 16 here is the max distance between two candidate indirections where we will attempt to prove that moving the latter one is ok.

src/coreclr/jit/lower.cpp
jakobbotsch: Changed

src/coreclr/jit/lower.cpp
jakobbotsch: No, not for correctness, but I moved the check into `OptimizeForLdp` anyway which will have the effect of checking it on both indirs, since we won't be able to combine them if either is volatile.

src/coreclr/jit/lower.cpp
jakobbotsch: The suggestion doesn't seem quite right to me, but I rephrased it in a different way which hopefully makes more sense.

jakobbotsch:
> Looks like there are some interesting regressions where spilling happens, and where the `ldp` opt. doesn't actually occur.

Some of the failures to combine `ldp` was because we had already combined the previous indir with another one. Added an early-out in this case to skip adding the "middle" indir to the `m_blockIndirs` list.

Remaining regressions seem to be mostly about register allocation, but I think it's expected to see some regressions due to suboptimal register allocation when doing this kind of reordering (since the transformations lengthens the lifetime of the moved LIR temp).

kunalspathak (approved):
LGTM.

AndyAyersMS:
Do you have a sense that moving the second indir "earlier" is better than moving the first indir "later"? (or even moving both of them somewhere in between)?

I believe think it's generally desirable to do loads as soon as you know you're going to need them so as to hide load latency, so I think moving the second one up is the right approach, but there may be cases where there is less pressure if we moved the first one later, or if the loads are not on any critical path maybe the CPU would be better off spending limited resources on something else.

Conventional wisdom is that "scheduling" is no longer really critical with today's highly OOO CPUs so perhaps none of this matters and compacting the code better is the key aspect.

jakobbotsch:
> Do you have a sense that moving the second indir "earlier" is better than moving the first indir "later"? (or even moving both of them somewhere in between)?

The decision here was mainly driven by the fact that implementation wise, it is much easier to move the second indirection and its operands backwards, compared to trying to find all transitive users (including through locals) of the first indirection and moving them forwards past the second indirection.

jakobbotsch:
superpmi-diffs failure is the "out of space" one.

Issue #92767, opened 2023-09-28
CommandLine.dll: Internal assembly name doesn't match the assembly name


kotlarmilos:
## Description

There is a mismatch between the `InternalName` and `OriginalFilename` values in the assembly and its actual name. While it's expected to be `CommandLine.dll`, it is `commandline.dll`.

Here are details from the assembly:
```
FileVersion2.0.275.0@InternalNamecommandline.dll(LegalCopyright HOriginalFilenamecommandline.dll4ProductVersion2.0.275<
Assembly Version2.0.275.0àÄ6
```

This mismatch causes a runtime issue in Mono embedding scenarios on apple mobile platforms. When the assembly is referenced by another, it cannot be loaded, resulting in the error:
```
Failed to load AOT module 'aot-instances' while running in aot-only mode because a dependency cannot be found or it is out of date.
```

## Repro steps
Mismatch in Assembly and OriginalFilename Causes Runtime Issue in Mono Embedding on Apple Mobile Platforms
1. Build the repository:
```
./build.sh mono+libs -os ios -arch arm64
```

2. Compile the tests:
```
./src/tests/build.sh  ci os ios arm64 Debug -mono tree baseservices/mono /p:ArchiveTests=true /p:DevTeamProvisioning=- /p:RunAOTCompilation=true /p:MonoForceInterpreter=false /p:LibrariesConfiguration=Debug /bl
```

## Workaround

If the assembly is not used, the parameter `/p:PublishTrimmed=true` should be passed to remove the assembly from the app bundle. 

An alternative is to pass the `--aot-lazy-assembly-load` runtime parameter, which shouldn't load all referenced assemblies during the startup. However, if the the referenced assembly is used, the runtime will abort because these kinds of errors can only be handled when loading the AOT image, not later.

PR #92772, opened 2023-09-28
[mono][ios] Apple mobile template allows passing process arguments to the runtime

kotlarmilos:
## Description

This PR introduces changes to the apple mobile template which allows passing process arguments to the runtime. ~~The `mono_ios_runtime_init` function is updated to take process arguments (`argc` and `argv`) and invokes `mono_jit_parse_options` function.~~ The `mono_ios_runtime_init` function already retrieves the arguments and they are passed to the `mono_jit_parse_options`.

This PR is the result of testing the `--aot-lazy-assembly-load` runtime parameter. The goal was to identify a potential workaround for https://github.com/dotnet/runtime/issues/92767.


steveisok:
@kotlarmilos do we need to modify anything in xharness to provide a way to shovel args?

kotlarmilos:
> @kotlarmilos do we need to modify anything in xharness to provide a way to shovel args?

It should be already possible to pass extra arguments to the iOS app bundle: 
https://github.com/dotnet/xharness/blob/2b1d423ce08e1ed78c0a821d0850e0f5ab3b193a/src/Microsoft.DotNet.XHarness.Apple/AppOperations/AppTester.cs#L566-L570

kotlarmilos:
@ivanpovazan @akoeplinger Please provide feedback when you have a chance, not urgent :)

src/tasks/AppleAppBuilder/Templates/runtime.h
```diff
@@ -4,6 +4,6 @@
 #ifndef runtime_h
 #define runtime_h
 
-void mono_ios_runtime_init (void);
+void mono_ios_runtime_init (int argc, char* argv[]);
```
ivanpovazan, line 7: nit: aligning declaration and definition signatures
```suggestion
void mono_ios_runtime_init (int argc, char** argv);
```

src/tasks/AppleAppBuilder/Templates/runtime.m
```diff
@@ -375,8 +375,17 @@ static bool is_pinvoke_override_library (const char* libraryName)
     mono_set_crash_chaining (TRUE);
 
     if (wait_for_debugger) {
-        char* options[] = { "--debugger-agent=transport=dt_socket,server=y,address=0.0.0.0:55556" };
-        mono_jit_parse_options (1, options);
+        argc++;
+        char** copy_argv = (char**) malloc (argc * sizeof (char*));
+        memcpy (copy_argv, argv, (argc - 1) * sizeof (char*));
+        copy_argv [argc - 1] = strdup ("--debugger-agent=transport=dt_socket,server=y,address=0.0.0.0:55556");
```
ivanpovazan, line 381: nit: I think it is a bit more clear when the arg count increment is done after copying, eg:
```c
        char** argv_ext = (char**) malloc ((argc + 1) * sizeof (char*));
        // copy passed args
        memcpy (argv_ext, argv, argc * sizeof (char*));
        // add an extra arg
        argv_ext [argc] = strdup ("--debugger-agent=transport=dt_socket,server=y,address=0.0.0.0:55556");
        argc++;
```

ivanpovazan (approved):
LGTM! 
Left only two smaller suggestions.

akoeplinger (approved):

src/tasks/AppleAppBuilder/Templates/runtime.m
```diff
@@ -255,7 +255,7 @@ static bool is_pinvoke_override_library (const char* libraryName)
 #endif
 
 void
-mono_ios_runtime_init (void)
+mono_ios_runtime_init (int argc, char** argv)
```
ivanpovazan, line 258: I just bumped into this: if we pass the arguments here, do we still need to do it again on line: https://github.com/dotnet/runtime/pull/92772/files#diff-6cef647106df3b37e45599e150f97a7dc3678f2606723b1967dbda7d1f0a42b9R281 
?

kotlarmilos: Good catch, let's simplify it. Since the `mono_ios_runtime_init` exits I suggest retrieving arguments within the function.

PR #46436, opened 2020-12-28
Bug fix issue 30351

NewellClark:
This fixes [issue 30351](https://github.com/dotnet/runtime/issues/30351). I noticed there weren't any unit tests for System.CodeDom.IndentedTextWriter, so I went ahead and added those as well.

Issue #30351, opened 2019-07-24
IndentedTextWriter Async methods don't indent properly


markrendle:
[System.CodeDom.Compiler.IndentedTextWriter](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.Extensions/src/System/CodeDom/Compiler/IndentedTextWriter.cs) wrapped around a `StringWriter` doesn't indent properly when using the `WriteLineAsync` methods.

- First write is not indented
- Writes following empty lines (i.e. from `await WriteLineAsync();`) are not indented
- Lines following lines with content are indented correctly.

Gist: https://gist.github.com/markrendle/01b8b70d2b7c4f023bf196e96562c14a

N.B.: I know there's no point using `*Async` on a `StringWriter`, but the code in question is used with both `StringWriter` and `StreamWriter`.

N.B. 2: I'd just submit a PR to fix this, but I don't know if that would constitute a breaking change.

Using .NET Core 3.0 Preview 7 win-x64.

carlossanlop:
Triage: @markrendle would you be interested in submitting a PR to get this fixed?

carlossanlop:
Assigning to @samtrion

PR #92791, opened 2023-09-28
IndentedTextWriter doesn't use indentation on the first line

### LINKED ISSUE #92039: IndentedTextWriter doesn't use indentation on the first line.
### Description

`IndentedTextWriter` doesn't emit indentation on the first line when `Indent` is nonzero.

### Reproduction Steps

```cs
[Test]
public void IndentedTextWriterBug()
{
	using StringWriter stringWriter = new();
	using IndentedTextWriter indentedTextWriter = new(stringWriter);
	indentedTextWriter.Indent++;
	indentedTextWriter.Write("");
	Assert.That(stringWriter.ToString(), Is.EqualTo(IndentedTextWriter.DefaultTabString));
	// Fails because the string is actually ""
}
```

### Expected behavior

Indentation should be emitted on any line if `Indent` is nonzero when beginning that line.

### Actual behavior

Indentation is not emitted in the first line.

### Regression?

I think it's been this way forever. Git blame showed that most of the file hasn't been changed for years, including the relevant field.

### Known Workarounds

* I can write the indentation myself for the first line.
* I can use reflection to set `_tabsPending` to true before using the `IndentedTextWriter`.

```cs
typeof(IndentedTextWriter).GetField("_tabsPending", BindingFlags.NonPublic | BindingFlags.Instance)!.SetValue(indentedTextWriter, true);
```

### Configuration

.NET 7

Windows 10 x64

### Other information

Initializing `_tabsPending` to `true` should solve this issue.

pedrobsaila:
Fixes #92039


buyaa-n (approved):
LGTM, thank you @pedrobsaila

buyaa-n:
This bug fix is somewhat breaking, though not sure if we want to add breaking change doc. The indentation issue first reported in https://github.com/dotnet/runtime/issues/30351 where it is mentioned that we would like to accept fix even if it was breaking. The issue is resolved with https://github.com/dotnet/runtime/pull/46436 but it only fixed the async Write indentation part.  We did not add a breaking change doc for that fix, but this one might be more breaking. 

Please let me know what you think @ericstj @jeffhandley @steveharter

PR #92794, opened 2023-09-28
[release/6.0] Update Ubuntu 16.04 amd64 queues to 22.04

carlossanlop:
Also use the queues with the *.svc suffix

### DIFF
--- a/eng/pipelines/coreclr/templates/helix-queues-setup.yml
+++ b/eng/pipelines/coreclr/templates/helix-queues-setup.yml
@@ -66,9 +66,9 @@ jobs:
     # Linux musl x64
     - ${{ if eq(parameters.platform, 'Linux_musl_x64') }}:
       - ${{ if eq(variables['System.TeamProject'], 'public') }}:
-        - (Alpine.314.Amd64.Open)ubuntu.2204.amd64.open@mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.14-helix-amd64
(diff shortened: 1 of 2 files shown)


agocke:
Am I reading this right that this is only the docker host container that's changing?

carlossanlop:
@agocke According to the email, the ask is to only bump the 1604 ones. Did I miss something?

dougbu (approved):
I like it :grin:

should have mentioned in #92796 as well: using `.svc` queues makes sense across the board. but @ilyas1974 probably won't get too annoyed if some are left behind.

similarly, you should be able to use ubuntu.2204.* queues everywhere you're testing in a Docker container. but we're not currently asking for everyone to stop using ubuntu.1804.*.

carlossanlop:
Ah! Good to know, @dougbu. I suspect that's what @agocke was going to say next. 😄 

I can do that in a separate PR. For now I only wanted to address the main request.

agocke:
What I was asking is: I read `(Alpine.314.Amd64.Open)ubuntu.2204.amd64.open.svc@mcr.microsoft.com/dotnet-buildtools/prereqs:alpine-3.14-helix-amd64` as saying, "run an alpine docker image on top of an ubuntu.22.04" host OS. So, the actual code running in this test is still seeing the same Alpine image as before, and it's just a question of which OS the kick-off code sees.

Is that right?

dougbu:
> Is that right?

Yes.

jkoritzinsky (approved):

agocke (approved):
Great! LGTM

PR #92797, opened 2023-09-28
Always emit jump from hot finally block to cold target

amanasifkhalid:
On some platforms, `CodeGen::genCallFinally()` will remove the jump between a finally block in a call-finally pair to its jump target if the target is its immediate successor in the block list (in other words, we just fall through). However, if we are doing hot/cold splitting, it is possible for the finally block to be the last hot block, and its target the first cold block. Thus, if the two are contiguous in the block list but in separate regions, we must always emit a jump.

### DIFF
--- a/src/coreclr/jit/codegenxarch.cpp
+++ b/src/coreclr/jit/codegenxarch.cpp
@@ -252,7 +252,8 @@ BasicBlock* CodeGen::genCallFinally(BasicBlock* block)
 #endif // JIT32_GCENCODER
 
         // Now go to where the finally funclet needs to return to.
-        if (block->bbNext->bbJumpDest == block->bbNext->bbNext)
+        if ((block->bbNext->bbJumpDest == block->bbNext->bbNext)
+            && !compiler->fgInDifferentRegions(block->bbNext, block->bbNext->bbJumpDest))
         {
             // Fall-through.
             // TODO-XArch-CQ: Can we get rid of this instruction, and just have the call return directly
--- a/src/coreclr/jit/codegenxarch.cpp
+++ b/src/coreclr/jit/codegenxarch.cpp
@@ -252,8 +252,8 @@ BasicBlock* CodeGen::genCallFinally(BasicBlock* block)
 #endif // JIT32_GCENCODER
 
         // Now go to where the finally funclet needs to return to.
-        if ((block->bbNext->bbJumpDest == block->bbNext->bbNext)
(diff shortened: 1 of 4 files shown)


amanasifkhalid:
cc @dotnet/jit-contrib

src/coreclr/jit/codegenxarch.cpp
```diff
@@ -252,7 +252,8 @@ BasicBlock* CodeGen::genCallFinally(BasicBlock* block)
 #endif // JIT32_GCENCODER
 
         // Now go to where the finally funclet needs to return to.
-        if (block->bbNext->bbJumpDest == block->bbNext->bbNext)
+        if ((block->bbNext->bbJumpDest == block->bbNext->bbNext)
+            && !compiler->fgInDifferentRegions(block->bbNext, block->bbNext->bbJumpDest))
```
AndyAyersMS, line 256: Should we revise this (and the "copies" in other archs) to factor out some of these repeated indirections, eg this seems more readable to me:
```suggestion
       BasicBlock* const nextBlock = block->bbNext;
       BasicBlock* const jumpDest  = nextBlock->bbJumpDest;
        if ((jumpDest == nextBlock->bbNext)
            && !compiler->fgInDifferentRegions(nextBlock, jumpDest))
```
Likely the compiler is doing this behind the scenes anyways so this is more about readabilty than about perf.

AndyAyersMS: Also note there are uses of `block->bbNext` higher up that we can capture too.

AndyAyersMS (commented):
Fix seems fine, just a suggestion on the code.

src/coreclr/jit/codegenxarch.cpp
amanasifkhalid: Good idea, I'll push a revision

amanasifkhalid:
Thank you @AndyAyersMS ; I implemented your suggestion.

AndyAyersMS (approved):

amanasifkhalid:
@AndyAyersMS sorry I was a bit overzealous the first time with your suggestion: I was checking the next block's jump destination too early in genCallEarly(), creating a potential null pointer dereference for call-finally pairs that don't return. Now CI is clean, aside from a Pri0 test timing out on osx x64 (seems unrelated?).

AndyAyersMS (approved):
Yeah "retless call" case is a bit odd and rare, but it can happen.

Issue #92236, opened 2023-09-18
Test failure: Loader\\binding\\tracing\\BinderTracingTest.ResolutionFlow\\BinderTracingTest.ResolutionFlow.cmd


BruceForstall:
pipeline: runtime-coreclr jitstress-isas-x86
job: coreclr windows x86 Checked jitstress_isas_x86_nosse3_4 @ Windows.10.Amd64.Open

hang? hit a 30m timeout.

https://dev.azure.com/dnceng-public/public/_build/results?buildId=408841&view=ms.vss-test-web.build-test-results-tab&runId=8882878&paneView=dotnet-dnceng.dnceng-build-release-tasks.helix-test-information-tab&resultId=120623

```
set DOTNET_TieredCompilation=0
set DOTNET_EnableSSE3_4=0
```

```
  Starting:    Loader.binding.XUnitWrapper (parallel test collections = on, max threads = 4)
   Loader.binding.XUnitWrapper: [Long Running Test] 'Loader\binding\tracing\BinderTracingTest.ResolutionFlow\BinderTracingTest.ResolutionFlow.cmd', Elapsed: 00:02:12
   Loader.binding.XUnitWrapper: [Long Running Test] 'Loader\binding\tracing\BinderTracingTest.ResolutionFlow\BinderTracingTest.ResolutionFlow.cmd', Elapsed: 00:04:12
   Loader.binding.XUnitWrapper: [Long Running Test] 'Loader\binding\tracing\BinderTracingTest.ResolutionFlow\BinderTracingTest.ResolutionFlow.cmd', Elapsed: 00:06:12
   Loader.binding.XUnitWrapper: [Long Running Test] 'Loader\binding\tracing\BinderTracingTest.ResolutionFlow\BinderTracingTest.ResolutionFlow.cmd', Elapsed: 00:08:12
   Loader.binding.XUnitWrapper: [Long Running Test] 'Loader\binding\tracing\BinderTracingTest.ResolutionFlow\BinderTracingTest.ResolutionFlow.cmd', Elapsed: 00:10:12
   Loader.binding.XUnitWrapper: [Long Running Test] 'Loader\binding\tracing\BinderTracingTest.ResolutionFlow\BinderTracingTest.ResolutionFlow.cmd', Elapsed: 00:12:12
[...]

PR #92802, opened 2023-09-29
Add stress mode for arm/arm64 function fragment splitting

BruceForstall:
Set JitSplitFunctionSize to either 4 or 200 under new STRESS_UNWIND mode.

### DIFF
--- a/src/coreclr/jit/compiler.h
+++ b/src/coreclr/jit/compiler.h
@@ -9897,6 +9897,7 @@ public:
         STRESS_MODE(NO_OLD_PROMOTION) /* Do not use old promotion */                            \
         STRESS_MODE(PHYSICAL_PROMOTION) /* Use physical promotion */                            \
         STRESS_MODE(PHYSICAL_PROMOTION_COST)                                                    \
+        STRESS_MODE(UNWIND) /* stress unwind info; e.g., create function fragments */           \
                                                                                                 \
         /* After COUNT_VARN, stress level 2 does all of these all the time */                   \
                                                                                                 \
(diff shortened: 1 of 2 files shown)


BruceForstall:
Out of disk space errors. Try again.

BruceForstall:
jitstress failure is in Loader/binding/tracing/BinderTracingTest.ResolutionFlow/BinderTracingTest.ResolutionFlow.sh: https://github.com/dotnet/runtime/issues/92236

replay failure looks like infra (and in any case not arm/arm64)?
```
C:\h\w\A5450979\w\9B6908B2\e>C:\python3\python.exe C:\h\w\A5450979\p\superpmi_replay.py -jit_directory C:\h\w\A5450979\p -arch x64 -platform linux -partition 3 -partition_count 3 -log_directory C:\h\w\A5450979\w\9B6908B2\uploads 
['linux-x64-3' END OF WORK ITEM LOG: Command timed out, and was killed]
```

BruceForstall:
@kunalspathak PTAL
cc @dotnet/jit-contrib

kunalspathak (approved):
I am assuming this is added to catch issues that usually was just showing up in Antigen? Also, don't think we need to update any pipeline with this new stress mode, right? It gets enabled automatically.

BruceForstall:
> I am assuming this is added to catch issues that usually was just showing up in Antigen? Also, don't think we need to update any pipeline with this new stress mode, right? It gets enabled automatically.

Yes in both cases.

I had intended to add this stress mode many years ago when fragments were introduced, but never did. (There might be an issue somewhere but I couldn't find it.)

PR #92803, opened 2023-09-29
Allow `key#value` for superpmi JIT option specification

BruceForstall:
superpmi.py will pass this through from the `-jitoption` / `-base_jit_option` / `-diff_jit_option` to superpmi.exe `-jitoption` and `-jit2option`.

Currently, the format is `key=value`. I wrap invocation of superpmi.py with Windows batch file scripting, which has an annoying problem of "eating" the equals size `=`. This works around that problem. I can't think of any case where `#` is needed in a key or value, hence that choice as an additional option.


BruceForstall:
Hmmmm.... CI didn't even run the 'runtime' pipeline (and build).

jakobbotsch (approved):

PR #92807, opened 2023-09-29
Clear up some confusion with emitAttr

BruceForstall:
Sometimes `emitAttr` is expected to be just a size, sometimes an attribute with size and GC/pointer bits. Clear up a few cases where there was confusion with what was expected, e.g., difference between declaration and definition parameter names.

Add an `EA_IS_SIZE()` macro that is usable for asserting that an `emitAttr` is exactly a size with no other bits set.


BruceForstall:
Not sure it's worth doing anything here without thinking about a broader convention.

PR #83945, opened 2023-03-26
Unroll SequenceEqual(ref byte, ref byte, nuint) in JIT

EgorBo:
Unroll `SequenceEqual` for constant length `[1..15]` (will add SIMD separately if this lands) for both x64 and arm64.
Example (utf8 literal):

```csharp
bool Test1(ReadOnlySpan<byte> data) => "hello world"u8.SequenceEqual(data);

bool Test2(Span<byte> data) => data.StartsWith("test"u8);
```
Codegen diff: https://www.diffchecker.com/E1laymuB/

## Limitations

Unfortunately, it works only when a constant span (either RVA or e.g. `data.Slice(0, 10)`) is on the left. It happens because we use left span's Length here:
https://github.com/dotnet/runtime/blob/922c75c837fc95092e1d9e433a087ac62082953d/src/libraries/System.Private.CoreLib/src/System/MemoryExtensions.cs#L1437

In theory, JIT is smart enough to perform things like:
```c
if (x == 42)
{
    Foo(x); // x will be replaced with 42
}
```
via AssertProp, but in this case it's a bit more complicated than that. Perhaps, we can assist it with `IsKnowConstant`. Or we can use RHS span's length instead if we think that a constant span is more likely to appear on the right side.
Works for `StartsWith`.

## Motivation

Mainly, [these comparisons](https://github.com/TechEmpower/FrameworkBenchmarks/blob/ec8216a383161f4b0f294006a653dad147a49487/frameworks/CSharp/aspnetcore/PlatformBenchmarks/BenchmarkApplication.cs#L82-L138) in TE.

## Benchmarks
```csharp
[Benchmark]
public int TE_Json()
{
    return GetRequestType("/json"u8);
}

[Benchmark]
public int TE_Plaintext()
{
    return GetRequestType("/plaintext"u8);
}

public static class Paths
{
    public static ReadOnlySpan<byte> Json => "/json"u8;
    public static ReadOnlySpan<byte> Plaintext => "/plaintext"u8;
}

[MethodImpl(MethodImplOptions.NoInlining)]
private static int GetRequestType(ReadOnlySpan<byte> path)
{
    // Simulate TE scenario
    if (path.Length == 10 && Paths.Plaintext.SequenceEqual(path))
    {
        return 1;
    }
    else if (path.Length == 5 && Paths.Json.SequenceEqual(path))
    {
        return 2;
    }
    return 3;
}

static byte[] data1 = new byte[100];
static byte[] data2 = new byte[100];

[Benchmark]
[...]

Issue #89722, opened 2023-07-31
Suboptimal codegen for Pairwise vector comparison


EgorBo:
```csharp
bool Test1(ref byte a, ref byte b)
{
    var a1 = Vector128.LoadUnsafe(ref a);
    var a2 = Vector128.LoadUnsafe(ref a, 16);
    var b1 = Vector128.LoadUnsafe(ref b);
    var b2 = Vector128.LoadUnsafe(ref b, 16);
    return a1 == b1 && a2 == b2;
}

// This produces better codegen (that we expect for Test1 too):
bool Test2(ref byte a, ref byte b)
{
    var a1 = Vector128.LoadUnsafe(ref a);
    var a2 = Vector128.LoadUnsafe(ref a, 16);
    var b1 = Vector128.LoadUnsafe(ref b);
    var b2 = Vector128.LoadUnsafe(ref b, 16);
    return ((a1 ^ b1) | (a2 ^ b2)) == Vector128<byte>.Zero;
}
```
Codegen:
```asm
; Method P:Test1(byref,byref):bool:this (FullOpts)	
       vzeroupper 	
       vmovups  xmm0, xmmword ptr [rdx]	
       vmovups  xmm1, xmmword ptr [rdx+10H]	
       vmovups  xmm2, xmmword ptr [r8]	
       vmovups  xmm3, xmmword ptr [r8+10H]	
       vpcmpeqb xmm0, xmm0, xmm2	
       vpmovmskb eax, xmm0	
       xor      ecx, ecx	
       vpcmpeqb xmm0, xmm1, xmm3	
       vpmovmskb edx, xmm0	
       cmp      edx, 0xFFFF	
       sete     dl	
       movzx    rdx, dl	
       cmp      eax, 0xFFFF	
       mov      eax, edx	
       cmovne   eax, ecx	
       ret      	
; Total bytes of code: 64


; Method P:Test2(byref,byref):bool:this (FullOpts)
       vzeroupper 
       vmovups  xmm0, xmmword ptr [rdx]
       vpxor    xmm0, xmm0, xmmword ptr [r8]
       vmovups  xmm1, xmmword ptr [rdx+10H]
       vpxor    xmm1, xmm1, xmmword ptr [r8+10H]
       vpor     xmm0, xmm0, xmm1
       vptest   xmm0, xmm0
       sete     al
       movzx    rax, al
       ret      
; Total bytes of code: 39
```
`Test1` is expected to produce the same codegen as `Test2`

PR #92810, opened 2023-09-29
JIT: Optimize SequenceEqual to use ccmp on ARM64

jakobbotsch:
In the original PR we could not get this this working due to some conservative interference. This now does the right thing with #92710 merged.

Also change LowerCallMemcmp/LowerCallMemmove to return next node to lower just to align it a bit more with other functions.

```csharp
bool Test1(ReadOnlySpan<byte> data) => "hello world"u8.SequenceEqual(data);
```

Before:
```asm
G_M52062_IG01:  ;; offset=0x0000
            stp     fp, lr, [sp, #-0x10]!
            mov     fp, sp
G_M52062_IG02:  ;; offset=0x0008
            movz    x2, #0x20D4      // data for <PrivateImplementationDetails>:430646847E70344C09F58739E99D5BC96EAC8D5FE7295CF196B986279876BF9B
            movk    x2, #0xE179 LSL #16
            movk    x2, #554 LSL #32
            cmp     w1, #11
            bne     G_M52062_IG05
						;; size=20 bbWeight=1 PerfScore 3.00
G_M52062_IG03:  ;; offset=0x001C
            ldr     x1, [x2]
            ldr     x2, [x2, #0x03]
            ldr     x3, [x0]
            ldr     x0, [x0, #0x03]
            eor     x1, x1, x3
            eor     x0, x2, x0
            orr     x0, x1, x0
            cmp     x0, #0
            cset    x0, eq
						;; size=36 bbWeight=0.48 PerfScore 6.96
G_M52062_IG04:  ;; offset=0x0040
            ldp     fp, lr, [sp], #0x10
            ret     lr
G_M52062_IG05:  ;; offset=0x0048
            mov     w0, wzr
            b       G_M52062_IG04
						;; size=8 bbWeight=0 PerfScore 0.00
```

After:
```asm
G_M52062_IG01:  ;; offset=0x0000
            stp     fp, lr, [sp, #-0x10]!
            mov     fp, sp
G_M52062_IG02:  ;; offset=0x0008
            movz    x2, #0x20D4      // data for <PrivateImplementationDetails>:430646847E70344C09F58739E99D5BC96EAC8D5FE7295CF196B986279876BF9B
            movk    x2, #0x63E LSL #16
            movk    x2, #445 LSL #32
            cmp     w1, #11
            bne     G_M52062_IG05
						;; size=20 bbWeight=1 PerfScore 3.00
G_M52062_IG03:  ;; offset=0x001C
            ldr     x1, [x2]
            ldr     x2, [x2, #0x03]
            ldr     x3, [x0]
            ldr     x0, [x0, #0x03]
            cmp     x1, x3
            ccmp    x2, x0, 0, eq
            cset    x0, eq
						;; size=28 bbWeight=0.48 PerfScore 6.48
G_M52062_IG04:  ;; offset=0x0038
            ldp     fp, lr, [sp], #0x10
            ret     lr
G_M52062_IG05:  ;; offset=0x0040
            mov     w0, wzr
            b       G_M52062_IG04
						;; size=8 bbWeight=0 PerfScore 0.00
```

Benchmarks from #83945 on my M1:
|       Method |        Job |               Toolchain |      Mean |     Error |    StdDev | Ratio |
|------------- |----------- |------------------------ |----------:|----------:|----------:|------:|
|      TE_Json | Job-RFUZAM |      /Core_Root/corerun | 1.9945 ns | 0.0100 ns | 0.0083 ns |  0.98 |
|      TE_Json | Job-KSXFQI | /Core_Root_base/corerun | 2.0339 ns | 0.0011 ns | 0.0009 ns |  1.00 |
|              |            |                         |           |           |           |       |
| TE_Plaintext | Job-RFUZAM |      /Core_Root/corerun | 0.7460 ns | 0.0086 ns | 0.0080 ns |  0.99 |
| TE_Plaintext | Job-KSXFQI | /Core_Root_base/corerun | 0.7515 ns | 0.0008 ns | 0.0008 ns |  1.00 |
|              |            |                         |           |           |           |       |
|    Equals_15 | Job-RFUZAM |      /Core_Root/corerun | 0.5063 ns | 0.0008 ns | 0.0008 ns |  0.87 |
|    Equals_15 | Job-KSXFQI | /Core_Root_base/corerun | 0.5799 ns | 0.0006 ns | 0.0006 ns |  1.00 |

### DIFF
--- a/src/coreclr/jit/lower.cpp
+++ b/src/coreclr/jit/lower.cpp
@@ -1798,11 +1798,12 @@ GenTree* Lowering::AddrGen(void* addr)
 //
 // Arguments:
 //    tree - GenTreeCall node to replace with STORE_BLK
+//    next - [out] Next node to lower if this function returns true
 //
 // Return Value:
-//    nullptr if no changes were made
+//    false if no changes were made
 //
-GenTree* Lowering::LowerCallMemmove(GenTreeCall* call)
+bool Lowering::LowerCallMemmove(GenTreeCall* call, GenTree** next)
 {
     JITDUMP("Considering Memmove [%06d] for unrolling.. ", comp->dspTreeID(call))
     assert(comp->lookupNamedIntrinsic(call->gtCallMethHnd) == NI_System_Buffer_Memmove);
@@ -1812,7 +1813,7 @@ GenTree* Lowering::LowerCallMemmove(GenTreeCall* call)
     if (comp->info.compHasNextCallRetAddr)
     {
         JITDUMP("compHasNextCallRetAddr=true so we won't be able to remove the call - bail out.\n")
-        return nullptr;
+        return false;
     }
 
     GenTree* lengthArg = call->gtArgs.GetUserArgByIndex(2)->GetNode();
@@ -1857,7 +1858,8 @@ GenTree* Lowering::LowerCallMemmove(GenTreeCall* call)
 
             JITDUMP("\nNew tree:\n")
             DISPTREE(storeBlk);
-            return storeBlk;
+            *next = srcBlk;
+            return true;
         }
         else
         {
@@ -1868,7 +1870,7 @@ GenTree* Lowering::LowerCallMemmove(GenTreeCall* call)
     {
         JITDUMP("size is not a constant.\n")
     }
-    return nullptr;
+    return false;
 }
 
 //------------------------------------------------------------------------
@@ -1877,11 +1879,12 @@ GenTree* Lowering::LowerCallMemmove(GenTreeCall* call)
 //
 // Arguments:
 //    tree - GenTreeCall node to unroll as memcmp
+//    next - [out] Next node to lower if this function returns true
 //
 // Return Value:
-//    nullptr if no changes were made
+//    false if no changes were made
 //
-GenTree* Lowering::LowerCallMemcmp(GenTreeCall* call)
+bool Lowering::LowerCallMemcmp(GenTreeCall* call, GenTree** next)
 {
     JITDUMP("Considering Memcmp [%06d] for unrolling.. ", comp->dspTreeID(call))
     assert(comp->lookupNamedIntrinsic(call->gtCallMethHnd) == NI_System_SpanHelpers_SequenceEqual);
@@ -1891,13 +1894,13 @@ GenTree* Lowering::LowerCallMemcmp(GenTreeCall* call)
     if (!comp->opts.OptimizationEnabled())
     {
         JITDUMP("Optimizations aren't allowed - bail out.\n")
-        return nullptr;
+        return false;
     }
 
     if (comp->info.compHasNextCallRetAddr)
     {
         JITDUMP("compHasNextCallRetAddr=true so we won't be able to remove the call - bail out.\n")
-        return nullptr;
+        return false;
     }
 
     GenTree* lengthArg = call->gtArgs.GetUserArgByIndex(2)->GetNode();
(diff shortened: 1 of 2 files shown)


jakobbotsch:

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=422680&view=results)

src/coreclr/jit/lower.cpp
```diff
@@ -2020,51 +2023,77 @@ GenTree* Lowering::LowerCallMemcmp(GenTreeCall* call)
...
 
                     BlockRange().InsertAfter(rArgClone, l1Indir, l2Offs, l2AddOffs, l2Indir);
                     BlockRange().InsertAfter(l2Indir, r1Indir, r2Offs, r2AddOffs, r2Indir);
-                    BlockRange().InsertAfter(r2Indir, lXor, rXor, resultOr, zeroCns);
-                    BlockRange().InsertAfter(zeroCns, result);
+
+#ifdef TARGET_ARM64
+                    if (!varTypeIsSIMD(loadType))
+                    {
+                        // ARM64 will get efficient ccmp codegen if we emit the normal thing:
```
EgorBo, line 2043: I remember we discussed this and it didn't want to emit ccmp due to some side effects or something like that (probably my PR that re-ordered INDirs fixed that?)

src/coreclr/jit/lower.cpp
```diff
@@ -2020,51 +2023,77 @@ GenTree* Lowering::LowerCallMemcmp(GenTreeCall* call)
...
+                        //
+                        // bool result = (*(int*)leftArg == *(int)rightArg) & (*(int*)(leftArg + 1) == *(int*)(rightArg
+                        // +
+                        // 1))
+
+                        GenTree* eq1 = newBinaryOp(comp, GT_EQ, TYP_INT, l1Indir, r1Indir);
+                        GenTree* eq2 = newBinaryOp(comp, GT_EQ, TYP_INT, l2Indir, r2Indir);
+                        result       = newBinaryOp(comp, GT_AND, TYP_INT, eq1, eq2);
+
+                        BlockRange().InsertAfter(r2Indir, eq1, eq2, result);
```
EgorBo, line 2053: Ideally, it should work for SIMD as well and then be optimized: https://github.com/dotnet/runtime/issues/89722

src/coreclr/jit/lower.cpp
```diff
@@ -2020,51 +2023,77 @@ GenTree* Lowering::LowerCallMemcmp(GenTreeCall* call)
...
+                    *next = lArgClone;
+
                     GenTree* l1Indir   = comp->gtNewIndir(loadType, lArgUse.Def());
                     GenTree* r1Indir   = comp->gtNewIndir(loadType, rArgUse.Def());
-                    GenTree* lXor      = newBinaryOp(comp, GT_XOR, actualLoadType, l1Indir, r1Indir);
                     GenTree* l2Offs    = comp->gtNewIconNode(cnsSize - loadWidth, TYP_I_IMPL);
                     GenTree* l2AddOffs = newBinaryOp(comp, GT_ADD, lArg->TypeGet(), lArgClone, l2Offs);
                     GenTree* l2Indir   = comp->gtNewIndir(loadType, l2AddOffs);
-                    GenTree* r2Offs    = comp->gtCloneExpr(l2Offs); // offset is the same
+                    GenTree* r2Offs    = comp->gtNewIconNode(cnsSize - loadWidth, TYP_I_IMPL);
```
EgorBo, line 2033: motivation? 🙂

EgorBo (approved):

src/coreclr/jit/lower.cpp
jakobbotsch: I left a TODO-CQ about it below, presumably both the int/SIMD paths can benefit on xarch (while only SIMD on arm64).
I was thinking we could model it after/combine it with `TryLowerAndOrToCCMP`, although I can see your prototype did it on tree form which probably makes it simpler

src/coreclr/jit/lower.cpp
jakobbotsch: No strong motivation... just that we were already creating all the other constants from scratch so it looked a bit out of place to me to clone one lone one. I can revert it if you prefer

EgorBo: I am fine with this, was just wondering if you were improving TP impact or something like that

PR #92820, opened 2023-09-29
Enable TensorPrimitives to perform in-place operations

stephentoub:
Some operations would produce incorrect results if the same span was passed as both an input and an output.  When vectorization was employed but the span's length wasn't a perfect multiple of a vector, we'd do the standard trick of performing one last operation on the last vector's worth of data; however, that relies on the operation being idempotent, and if a previous operation has overwritten input with a new value due to the same memory being used for input and output, some operations won't be idempotent.  This fixes that by masking off the already processed elements.  It adds tests to validate in-place use works, and it updates the docs to carve out this valid overlapping.

### DIFF
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
@@ -23,6 +23,9 @@ namespace System.Numerics.Tensors
         /// <para>
         /// This method effectively computes <c><paramref name="destination" />[i] = (Half)<paramref name="source" />[i]</c>.
         /// </para>
+        /// <para>
+        /// <paramref name="source"/> and <paramref name="destination"/> must not overlap. If they do, behavior is undefined.
+        /// </para>
         /// </remarks>
         public static void ConvertToHalf(ReadOnlySpan<float> source, Span<Half> destination)
         {
@@ -47,6 +50,9 @@ namespace System.Numerics.Tensors
         /// <remarks>
         /// <para>
         /// This method effectively computes <c><paramref name="destination" />[i] = (float)<paramref name="source" />[i]</c>.
+        /// </para>
+        /// <para>
+        /// <paramref name="source"/> and <paramref name="destination"/> must not overlap. If they do, behavior is undefined.
         /// </para>
         /// </remarks>
         public static void ConvertToSingle(ReadOnlySpan<Half> source, Span<float> destination)
@@ -519,7 +525,10 @@ namespace System.Numerics.Tensors
                         return GetFirstNaN(current);
                     }
 
-                    result = TMinMax.Invoke(result, current);
+                    result = Vector512.ConditionalSelect(
+                        Vector512.Equals(LoadRemainderMaskSingleVector512(x.Length - i), Vector512<float>.Zero),
+                        result,
+                        TMinMax.Invoke(result, current));
                 }
 
                 // Aggregate the lanes in the vector to create the final scalar result.
@@ -565,7 +574,10 @@ namespace System.Numerics.Tensors
                         return GetFirstNaN(current);
                     }
 
-                    result = TMinMax.Invoke(result, current);
+                    result = Vector256.ConditionalSelect(
+                        Vector256.Equals(LoadRemainderMaskSingleVector256(x.Length - i), Vector256<float>.Zero),
+                        result,
+                        TMinMax.Invoke(result, current));
                 }
 
                 // Aggregate the lanes in the vector to create the final scalar result.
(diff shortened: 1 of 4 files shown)


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -672,7 +687,10 @@ private static float CosineSimilarityCore(ReadOnlySpan<float> x, ReadOnlySpan<fl
                     if (i != x.Length)
                     {
                         uint lastVectorIndex = (uint)(x.Length - Vector512<float>.Count);
-                        TUnaryOperator.Invoke(Vector512.LoadUnsafe(ref xRef, lastVectorIndex)).StoreUnsafe(ref dRef, lastVectorIndex);
+                        Vector512.ConditionalSelect(
+                            Vector512.Equals(LoadRemainderMaskSingleVector512(x.Length - i), Vector512<float>.Zero),
+                            Vector512.LoadUnsafe(ref dRef, lastVectorIndex),
```
tannergooding, line 692: Is it worth avoiding this load by caching the last vector computed from the main loop?

tannergooding (approved):

EgorBo:
Any particular reason we can't handle overlapped pointers? We can throw an exception (thec check is cheap, isn't) or just use scalar fallback?

tannergooding:
There's little benefit and a lot of drawbacks, even for scalar code paths.

If your output overlaps your input after the input start, then you have a self mutating algorithm even for the scalar side.

So it's easier just to leave it as UB and tell users "don't do that".

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: I'm not following; where did we already load this vector?

tannergooding: We're basically backtracking by `Count - remainder`, doing an operation, and then blending the already computed entries with the new entries before storing.

Say we have a total number `19` inputs. We'd do 4 full loop iterations (`0-3`, `4-7`, `8-11`, `12-15`).

We'd then have `3` remaining, so we backtrack by 1 and reprocess `15-18`. To support exact overlapping input/output, we need to preserve the already computed `15` which already exists as the last element from the previous loop iteration.

So, rather than doing another memory access, we could just cache the last stored vector from the loop, shuffle the elements left by 3, and use that in the conditional select.

EgorBo:
Burst solved this via `[NoAlias]` attributes: https://docs.unity3d.com/Packages/com.unity.burst@1.4/manual/docs/OptimizationGuidelines-Aliasing.html

It seems like silent UB is the worst option honestly

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: That might be something worth doing here, but it might also be fine to postpone that for "modern .NET only" when we can get some helper API in place to make that process simpler.

tannergooding:
> An exception telling me "you should go and make a copy of your data" sounds way better than a silent UB to me.

We could explicitly throw, yes. I don't have a preference.

> I don't see any problems with that. Thats how auto-vectorization works in C++, memset,memcpy etc - we check whether pointers don't overlap (if we can't tell that from a general alias analysis) and do simd, otherwise - scalar fallback.

A lot of those are explicitly UB (particularly `memcpy`) or only have a single input (`memset`). `memmove` is one of the only places where overlapping input/output is well defined and only because it can be well-defined since it's a single input/output and only doing memory copying, so the output overwriting inputs isn't a concern (because it is spec'd to work as if a copy to an intermediate buffer happened first).

EgorBo:
> A lot of those are explicitly UB (particularly memcpy) 

That's why in many places `memcpy` is just an alias of `memmove` for safety reasons.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: Ok, let's leave it as is for now.

stephentoub:
This PR isn't introducing the undefined behavior, and rather is making some more defined, so I'm going to merge, and we can subsequently decide to follow-up with either leaving it as is, throwing for any other kinds of overlap, or making it work.

tannergooding:
> That's why in many places memcpy is just an alias of memmove for safety reasons. 

Places doing it for convenience or simplicity doesn't mean it's something that can be relied upon.

The do it because it's trivial in the case of single input/output to handle it using SIMD with a minor additional code path that remains accelerated. It basically just requires caching an extra value in a temporary SIMD register.

For cases like `Abs` or similar, that would likewise be trivial. For cases like `Add`, that becomes immensely complex due to having to consider all the ways the destination could overlap with either input.

I'd be in favor of simply throwing if the UB is that much of a concern. It's what we do for a few other APIs.

PR #92821, opened 2023-09-29
Call fgRenumberBlocks in optIfConversion

### LINKED ISSUE #92818: The JIT uses an redundant rbp register
### Description

The JIT uses an redundant `rbp` register.

Code:
```c#
public class C
{
    public static int M1(int x, int y)
    {
        return (x == 0 && y == 0) ? 1 : 0;
    }

    public static bool M2(int x, int y)
    {
        return (x == 0 && y == 0);
    }
}
```

ASM:
```asm
// coreclr trunk-20230907+17b60e35528ec0f1eba8a6606168412738bade59

C:M1(int,int):int (FullOpts):
       push     rbp
       mov      rbp, rsp
       or       edi, esi
       sete     al
       movzx    rax, al
       pop      rbp
       ret      

C:M2(int,int):ubyte (FullOpts):
       or       edi, esi
       sete     al
       movzx    rax, al
       ret      
```

[godbolt](https://godbolt.org/z/1r9Mbqjz5)

EgorBo:
Closes https://github.com/dotnet/runtime/issues/92818

TP/Size diffs CI test, no need in code review

### DIFF
--- a/src/coreclr/jit/ifconversion.cpp
+++ b/src/coreclr/jit/ifconversion.cpp
@@ -793,5 +793,10 @@ PhaseStatus Compiler::optIfConversion()
     }
 #endif
 
+    if (madeChanges)
+    {
+        fgRenumberBlocks();
+    }
+
     return madeChanges ? PhaseStatus::MODIFIED_EVERYTHING : PhaseStatus::MODIFIED_NOTHING;
 }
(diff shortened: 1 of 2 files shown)


EgorBo:
Diffs are not too big but there is n TP regression: https://dev.azure.com/dnceng-public/public/_build/results?buildId=423271&view=ms.vss-build-web.run-extensions-tab

`IfConvert` phases collapsed some conditions into a branchless form and it left `fgBbCount` not updated - it then resulted in unnecessary RBP-based frame due to block count (see ). Typically, we call `fgUpdateChangedFlowGraph` in such cases, but it was a bit overkill TP-wise here.

I tried to make it more generic and do as part of some "Optimize Layout" phase but it resulted in TP regressions.

@dotnet/jit-contrib @AndyAyersMS PTAL

AndyAyersMS (approved):
You should update the PR title...

Issue #85755, opened 2023-05-04
SPMI: Give specific context examples for tpdiff


jakobbotsch:
Running `tpdiff` should give a "jit-analyze"-like result of the throughput results, i.e. specific context examples of regressions/improvements.

I think there is potential to clean up superpmi.exe as part of this work. Today we produce several .csv files to communicate the information back: base metrics, diff metrics and the diffs info. The problem is that the "diffs info" only writes a row for each context with ASM diffs, so it is not useful for this purpose.
We could subsume these three mechanisms if we introduce a new "--output-csv" argument and always output a row into this file for every context. superpmi.py would then be responsible for aggregating the metrics from this file and and could furthermore use it to produce the tpdiff examples.

jakobbotsch:
I've put this in future for now, but I will try to see if I can get to this in 8.0 time frame.

jakobbotsch:
After #92824 this should be a straightforward extension to the report generation.

PR #92824, opened 2023-09-29
SPMI: Simplify and improve reporting of context information

jakobbotsch:
Currently we have multiple separate mechanisms to report information
back from superpmi.exe:
1. -baseMetricsSummary/metricsSummary, which outputs a .csv file with
   aggregated statics for all contexts from the perspective of the base
   JIT
2. -diffMetricsSummary, which is the corresponding for the diff JIT
   when diffing
3. -diffsInfo, which during diffing will output a .csv with individual
   rows for every context that had diffs in it

This PR replaces these three mechanisms with a -details argument. When
passed, superpmi.exe will write a .csv file to the specified path that
contains a row for every context.

The arg is supported in both replay and diff mode but creates .csv files
with slightly different formats for these. For replays the header output
is:
```
Context,Context size,Result,MinOpts,Size,Instructions\n
```

For diffs the output is:
```
Context,Context size,Base result,Diff result,MinOpts,Has diff,Base size,Diff size,Base instructions,Diff instructions\n
```

superpmi.py is changed to utilize this new output instead, which
involves computing some of the same details we were getting from the
metrics summaries before.

Prerequisite for https://github.com/dotnet/runtime/issues/85755

No observable superpmi.py changes are expected.


jakobbotsch:

Ran two test superpmi-diffs runs [with](https://dev.azure.com/dnceng-public/public/_build/results?buildId=423222&view=results) and [without](https://dev.azure.com/dnceng-public/public/_build/results?buildId=423372&view=results) any diffs.

BruceForstall (approved):
LGTM

PR #92826, opened 2023-09-29
Clean up the crossgen2_publish project and local/live packs references

jkoritzinsky:
Remove some properties in the `crossgen2_publish.csproj` ProjectReference that we can define in the project. Resolve various TODOs in targetingpacks.targets to use SDK features and remove some of our custom code. Move the "use the locally-built crossgen2" logic into the `eng/` folder and set it up to be automatically imported when a project is targeting the locally-built runtime.


ivdiazsa:
This overall looks good to me and will be a nice cleanup of our general infra. I'm seeing that some builds are failing because they can't find the `apphost` however. That's actually one of the problems I was facing with my TestR2R Libraries project that I've yet to resolve. But I think that once that's fixed, this should be ready to merge.

eng/readytorun.targets
```diff
@@ -0,0 +1,36 @@
...
+      <OptimizationMibcFiles Include="$(CoreCLRArtifactsPath)StandardOptimizationData.mibc" Condition="'$(DotNetBuildFromSource)' != 'true'" />
+    </ItemGroup>
+
+    <ItemGroup>
+      <PublishReadyToRunPgoFiles Condition="'$(IncludePgoFiles)' == 'true'" Include="$(CoreCLRArtifactsPath)StandardOptimizationData.mibc"/>
+      <PublishReadyToRunCrossgen2ExtraArgsList Condition="'$(IncludePgoFiles)' == 'true'" Include="--embed-pgo-data"/>
+    </ItemGroup>
+
+    <PropertyGroup>
+      <Crossgen2Dll>$(CoreCLRArtifactsPath)\$(BuildArchitecture)\crossgen2\crossgen2.dll</Crossgen2Dll>
```
ivdiazsa, line 18: NIT/Question: What's the criteria to define whether we write paths with backslashes, forward slashes, or .NET's `Path` API? I think it would be best to gradually standardize it for maximum safety. While I prefer forward slashes, I believe using the `Path` API would be best to avoid potential compatibility pitfalls.

src/installer/pkg/sfx/Microsoft.NETCore.App/ReadyToRun.targets
```diff
@@ -1,51 +0,0 @@
...
-
-      <PublishReadyToRunUseCrossgen2>true</PublishReadyToRunUseCrossgen2>
-      <PublishReadyToRunCrossgen2ExtraArgs>@(PublishReadyToRunCrossgen2ExtraArgsList)</PublishReadyToRunCrossgen2ExtraArgs>
-      <PublishReadyToRunPerfmapFormatVersion Condition="'$(PublishReadyToRunPerfmapFormatVersion)' == ''">1</PublishReadyToRunPerfmapFormatVersion>
-
-      <ScriptExt>.sh</ScriptExt>
-      <ScriptExt Condition="'$(OS)' == 'Windows_NT'">.cmd</ScriptExt>
-    </PropertyGroup>
-
-    <ItemGroup Condition="'$(RuntimeFlavor)' != 'Mono'">
```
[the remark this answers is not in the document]
jkoritzinsky: This section was for crossgen1, which we no longer produce.

eng/targetingpacks.targets
```diff
@@ -20,11 +20,9 @@
                             '$(TargetFrameworkVersion)' == 'v$(NetCoreAppCurrentVersion)'">
```
ViktorHofer, line 20: Remove changes from this file. I'm already taking care of them via https://github.com/dotnet/runtime/pull/88991/

eng/readytorun.targets
```diff
@@ -0,0 +1,36 @@
...
+
+      <ScriptExt>.sh</ScriptExt>
+      <ScriptExt Condition="'$(OS)' == 'Windows_NT'">.cmd</ScriptExt>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <Crossgen2Tool Include="$(Crossgen2Dll)"
+                     TargetArch="$(TargetArchitecture)"
+                     TargetOS="$(TargetOS)"
+                     DotNetHostPath="$(RepoRoot)/dotnet$(ScriptExt)"
```
ViktorHofer, line 32:
```suggestion
                     DotNetHostPath="$(DotNetTool)"
```

eng/readytorun.targets
```diff
@@ -0,0 +1,36 @@
...
+
+    <PropertyGroup>
+      <Crossgen2Dll>$(CoreCLRArtifactsPath)\$(BuildArchitecture)\crossgen2\crossgen2.dll</Crossgen2Dll>
+
+      <PublishReadyToRunUseCrossgen2>true</PublishReadyToRunUseCrossgen2>
+      <PublishReadyToRunCrossgen2ExtraArgs>@(PublishReadyToRunCrossgen2ExtraArgsList)</PublishReadyToRunCrossgen2ExtraArgs>
+      <PublishReadyToRunPerfmapFormatVersion Condition="'$(PublishReadyToRunPerfmapFormatVersion)' == ''">1</PublishReadyToRunPerfmapFormatVersion>
+
+      <ScriptExt>.sh</ScriptExt>
+      <ScriptExt Condition="'$(OS)' == 'Windows_NT'">.cmd</ScriptExt>
```
ViktorHofer, line 25:
```suggestion
```

eng/targetingpacks.targets
jkoritzinsky: I need to keep the import of the ReadyToRun.targets file as your PR does not handle actually wiring up the in-build crossgen2 invocation, but I'll remove the rest of the changes.

src/libraries/System.Runtime.InteropServices/tests/TestAssets/NativeExports/NativeExports.csproj
```diff
@@ -23,16 +23,12 @@
...
-                      ExcludedRuntimeIdentifiers="android"
-                      AppHostPackNamePattern="Microsoft.NETCore.App.Host.**RID**"
-                      AppHostPackVersion="7.0.0"
-                      AppHostRuntimeIdentifiers="linux-arm;linux-arm64;linux-musl-arm64;linux-musl-x64;linux-x64;osx-x64;rhel.6-x64;tizen.4.0.0-armel;tizen.5.0.0-armel;win-arm64;win-x64;win-x86;linux-musl-arm;osx-arm64;linux-s390x;linux-bionic-arm;linux-bionic-arm64;linux-bionic-x64;linux-bionic-x86"
-                      TargetFramework="$(NetCoreAppCurrent)"
-                      Condition="'@(KnownAppHostPack)' == '' or !@(KnownAppHostPack->AnyHaveMetadataValue('TargetFramework', '$(NetCoreAppCurrent)'))" />
+  <ItemGroup Condition="'$(UseLocalAppHostPack)' != 'true' and '$(NetCoreAppToolCurrent)' != '$(NetCoreAppCurrent)'">
+    <KnownAppHostPack Include="@(KnownAppHostPack->WithMetadataValue('Identity', 'Microsoft.NETCore.App')->WithMetadataValue('TargetFramework', '$(NetCoreAppToolCurrent)'))">
+      <TargetFramework>$(NetCoreAppCurrent)</TargetFramework>
+    </KnownAppHostPack>
```
ViktorHofer, line 31: nit:

```suggestion
    <KnownAppHostPack Include="@(KnownAppHostPack->WithMetadataValue('Identity', 'Microsoft.NETCore.App')->WithMetadataValue('TargetFramework', '$(NetCoreAppToolCurrent)'))"
                      TargetFramework="$(NetCoreAppCurrent)" />
```

eng/readytorun.targets
ViktorHofer: In this case, it would be better to use MSBuild's normalization function though as it automatically takes care of using the right directory separators (which makes the path look correct in a binlog).

```suggestion
      <Crossgen2Dll>$([MSBuild]::NormalizePath('$(CoreCLRArtifactsPath)', '$(BuildArchitecture)', 'crossgen2', 'crossgen2.dll'))</Crossgen2Dll>
```

src/installer/pkg/sfx/Microsoft.NETCore.App/Microsoft.NETCore.App.Crossgen2.sfxproj
```diff
@@ -31,13 +31,8 @@
...
       Include="$(RepoRoot)src/coreclr/tools/aot/crossgen2/crossgen2_publish.csproj"
-      AdditionalProperties="_IsPublishing=true
-        ;RuntimeIdentifier=$(PackageRID)
-        ;NativeAotSupported=$(NativeAotSupported)
-        ;CoreCLRArtifactsPath=$(CoreCLRArtifactsPath)
-        ;ObjCopyName=$(ObjCopyName)
-        ;R2ROverridePath=$(MSBuildThisFileDirectory)ReadyToRun.targets"
-      Targets="Publish;PublishItemsOutputGroup"
+      AdditionalProperties="CoreCLRArtifactsPath=$(CoreCLRArtifactsPath)
+        ;ObjCopyName=$(ObjCopyName)"
```
ViktorHofer, line 35: The problem with additional properties is that they flow to the entire referenced project graph. I.e. with https://github.com/dotnet/runtime/pull/91468, crossgen2_publish.csproj now references ILLink.Tasks.csproj which references Mono.Linker.csproj with those global properties set. Other build parts reference these projects as well, but without these global properties set, which results in the following error:

> /Users/runner/work/1/s/.dotnet/sdk/8.0.100-rc.1.23415.11/Sdks/Microsoft.NET.Sdk/targets/Microsoft.PackageDependencyResolution.targets(266,5): error NETSDK1047: Assets file '/Users/runner/work/1/s/artifacts/obj/Mono.Linker/project.assets.json' doesn't have a target for 'net8.0/osx-arm64'. Ensure that restore has run and that you have included 'net8.0' in the TargetFrameworks for your project. You may also need to include 'osx-arm64' in your project's RuntimeIdentifiers. [/Users/runner/work/1/s/src/tools/illink/src/linker/Mono.Linker.csproj]

jkoritzinsky: With this change, we don’t pass the RID through AdditionalProperties and these properties shouldn’t influence anything in the restore process, so hopefully that fixes this problem.

ViktorHofer: These additional global properties still flow to the linker projects which cause double builds, double writes and in the worst case races. We should avoid that.

jkoritzinsky: I'll see if I can get rid of the rest of them.

eng/readytorun.targets
```diff
@@ -0,0 +1,33 @@
+<Project>
+  <PropertyGroup>
+    <!-- Only use mibc files if UsingToolIbcOptimization is false. Allows enabling/disabling using ibc instead of mibc data -->
+    <IncludePgoFiles Condition="'$(UsingToolIbcOptimization)' != 'true' and '$(EnableNgenOptimization)' == 'true' and '$(DotNetBuildFromSource)' != 'true'">true</IncludePgoFiles>
+  </PropertyGroup>
+
+  <Target Name="ResolveReadyToRunCompilers" DependsOnTargets="ResolveRuntimeFilesFromLocalBuild">
+    <ItemGroup>
+      <OptimizationMibcFiles Include="$(CoreCLRArtifactsPath)StandardOptimizationData.mibc" Condition="'$(DotNetBuildFromSource)' != 'true'" />
+    </ItemGroup>
```
ViktorHofer, line 10: AFAIK this item is only used by crossgen-corelib.proj. Wouldn't it make more sense to move it there?

eng/targetingpacks.targets
```diff
@@ -185,4 +160,6 @@
                         Condition="'%(RuntimeFramework.FrameworkName)' == '$(LocalFrameworkOverrideName)'" />
     </ItemGroup>
   </Target>
+
+  <Import Project="$(MSBuildThisFileDirectory)readytorun.targets" Condition="'$(UseLocalTargetingRuntimePack)' == 'true'" />
```
ViktorHofer, line 164: The targetingpacks.targets file is designed to be self-contained. Can we just inline the target and make `Microsoft.NETCore.App.Runtime.sfxproj` import targetingpacks.targets? Alternatively, if we don't want that, can we move this import out of this file, similar to how we use the live linker (just another import next to targetingpacks.targets)?

eng/readytorun.targets
jkoritzinsky: It’s used when publishing the shared framework as a whole as well.

eng/targetingpacks.targets
jkoritzinsky: We can’t as we need to override the target, not sequence before or after it. I can move it out, but I wanted to have it trigger on the same properties that are set in this file so that if you’re using the live runtime, you get the live R2R experience as well.

eng/codeOptimization.targets
```diff
@@ -27,4 +27,17 @@
     </PropertyGroup>
   </Target>
 
+  <Target Name="AddReadyToRunPgoOptions" DependsOnTargets="ResolveRuntimeFilesFromLocalBuild" BeforeTargets="ResolveReadyToRunCompilers">
```
ViktorHofer, line 30: Do we actually depend on that target `ResolveRuntimeFilesFromLocalBuild` here?

eng/targetingpacks.targets
```diff
@@ -147,6 +147,17 @@
     </ItemGroup>
   </Target>
 
+  <!-- Use local Crossgen2 pack for NetCoreAppCurrent. -->
+  <Target Name="UpdateLocalCrossgen2Pack"
+          Condition="'$(UseLocalCrossgen2Pack)' == 'true'"
+          AfterTargets="ResolveFrameworkReferences">
+    <ItemGroup>
+      <ResolvedCrossgen2Pack PackageDirectory="$(CoreCLRArtifactsPath)\$(BuildArchitecture)\crossgen2\"
```
ViktorHofer, line 155: Let's not hardcode this path here and use a property instead. I noticed people copying this file to use a local build tree which wouldn't work anymore then.

```suggestion
      <ResolvedCrossgen2Pack PackageDirectory="$(Crossgen2Dir)"
```

eng/targetingpacks.targets
```diff
@@ -147,6 +147,17 @@
...
+                             Condition="$([System.String]::new('%(Identity)').StartsWith('$(LocalFrameworkOverrideName).Crossgen2'))" />
```
ViktorHofer, line 156: Are you sure this is necessary? This item should only contain Crossgen2 framework pack items.

eng/targetingpacks.targets
```diff
@@ -147,6 +147,17 @@
...
+    </ItemGroup>
+  </Target>
+  
+
```
ViktorHofer, line 160:
```suggestion
```

eng/targetingpacks.targets
```diff
@@ -172,4 +183,15 @@
                         Condition="'%(RuntimeFramework.FrameworkName)' == '$(LocalFrameworkOverrideName)'" />
     </ItemGroup>
   </Target>
+
+  <Target Name="UpdateLocalCrossgen2Tool"
```
ViktorHofer, line 187: Please add a comment above this target

eng/targetingpacks.targets
ViktorHofer: Btw, what made me not pursue this further is that the local build layout doesn't group the binaries under a "tools" folder which the `ResolveReadyToRunCompilers` task depends on?

ViktorHofer, line 155: One more comment, please also set `NuGetPackageVersion="$(ProductVersion)"` just for completeness and parity with the other items.

jkoritzinsky: This PR also updates `crossgen2_inbuild.targets` to put it under a `tools` folder.  I need to update one more place I realized I missed related to this.

eng/testing/tests.singlefile.targets
```diff
@@ -142,19 +142,6 @@
...
-      <Crossgen2ArtifactPath>$(CoreCLRCrossgen2Dir)crossgen2$(ExeSuffix)</Crossgen2ArtifactPath>
-    </PropertyGroup>
-
-    <ItemGroup>
-      <Crossgen2CurrentTool Include="@(Crossgen2Tool->'$(Crossgen2ArtifactPath)')" />
-      <Crossgen2Tool Remove="@(Crossgen2Tool)" />
-      <Crossgen2Tool Include="@(Crossgen2CurrentTool)" />
-    </ItemGroup>
-  </Target>
-
```
jkoritzinsky, line 157: @ivdiazsa I've removed this target now that we automatically hook up the Crossgen2 tool when referencing the live build.

ViktorHofer (approved):

PR #92830, opened 2023-09-29
[linux-bionic] Skip failing NTLM tests

steveisok:
Some NTLM tests are still running and failing even though libSystem.Net.Security.Native isn't being built for linux-bionic. This PR skips them all.

### DIFF
--- a/src/libraries/System.Net.Security/tests/UnitTests/NegotiateAuthenticationTests.cs
+++ b/src/libraries/System.Net.Security/tests/UnitTests/NegotiateAuthenticationTests.cs
@@ -32,6 +32,7 @@ namespace System.Net.Security.Tests
         }
 
         [Fact]
+        [SkipOnPlatform(TestPlatforms.LinuxBionic, "The test is specific to GSSAPI / Managed implementations of NegotiateAuthentication")]
         public void RemoteIdentity_ThrowsOnUnauthenticated()
         {
             NegotiateAuthenticationClientOptions clientOptions = new NegotiateAuthenticationClientOptions { Credential = s_testCredentialRight, TargetName = "HTTP/foo" };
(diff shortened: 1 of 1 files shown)


wfurt:
should we include the managed implementation? That should just work without GSSAP AFAIK.
cc: @filipnavara

steveisok:
> should we include the managed implementation? That should just work without GSSAP AFAIK. cc: @filipnavara

Ah, yes, I think we should.

filipnavara:
We should but there was not any product-level check that we are running on linux-bionic.

am11:
> We should but there was not any product-level check that we are running on linux-bionic.

It is possible to test:
```c#
if (RuntimeInformation.RuntimeIdentifier.StartsWith("linux-bionic")) ForBionic();
else if (RuntimeInformation.RuntimeIdentifier.StartsWith("linux-musl")) ForMusl();
else ForGlibC();
```

filipnavara:
> It is possible to test:

Hmm. I though there's a reason why the [test check](https://github.com/dotnet/runtime/blob/8dad137b7a51adba94c7224768a4c5c76b017f4d/src/libraries/Common/tests/TestUtilities/System/PlatformDetection.Unix.cs#L171-L184) is implemented differently. If `RuntimeInformation.RuntimeIdentifier` returns the correct information then we likely just need to modify this condition:

https://github.com/dotnet/runtime/blob/8dad137b7a51adba94c7224768a4c5c76b017f4d/src/libraries/System.Net.Security/src/System/Net/NegotiateAuthenticationPal.Unix.cs#L25

and add `|| (OperationSystem.IsLinux() && RuntimeInformation.RuntimeIdentifier.StartsWith("linux-bionic")`.

adamsitnik (changes requested):
Please solve the merge conflict.

I am requesting changes mostly so this PR does not show up at the next Stale PRs email ;)

steveisok:
For whatever reason, this doesn't appear to be failing in CI any longer.

PR #92833, opened 2023-09-29
Remove `CompilationProvider` dependency from the source generator.

eiriktsarpalis:
The incremental values defined in the Regex source generator have a dependency on `CompilationProvider` which means that any update to the compilation object will trigger a re-evaluation of the `CompilationData` value. This PR refactors the implementation somewhat so that compilation data is obtained from the semantic model handed by the `SyntaxProvider`, meaning that the source generator should only trigger in cases where source containing `GeneratedRegexAttribute` attribute declarations is being updated.


stephentoub:
This looks fine. Just to make sure I understand, though, the bulk of the work was already being avoided, yes?  The majority of the work in the source generator is on the actual code generation for the regexes, which shouldn't have been based on this input.

stephentoub (approved):

eiriktsarpalis:
Correct, the original CompilationData incremental value would end up being cached so it wouldn't trigger additional computation downstream.

PR #92835, opened 2023-09-29
Apply sequence equality comparison to the final Regex incremental value.

eiriktsarpalis:
ImmutableArray uses reference equality semantics, so two arrays containing equal values will always trigger source compilation downstream. Updates the incremental value to use reference equality comparison.

### DIFF
--- a/src/libraries/System.Text.RegularExpressions/gen/RegexGenerator.cs
+++ b/src/libraries/System.Text.RegularExpressions/gen/RegexGenerator.cs
@@ -125,7 +125,10 @@ namespace System.Text.RegularExpressions.Generator
                 })
 
                 // Combine all of the generated text outputs into a single batch. We then generate a single source output from that batch.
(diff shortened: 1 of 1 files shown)


stephentoub (approved):

stephentoub:
Thanks. How did you a) come across the need for these and b) validate them?

eiriktsarpalis:
It's just something I noticed when auditing the implementation. I could try adding an incremental caching unit test suite similar to the LibraryImportGenerator or STJ to make sure this works as expected E2E.

eiriktsarpalis:
@stephentoub would we want to service this for .NET 8? Without this change the final incremental value is always going to use reference equality so it can never cache results.

stephentoub:
.NET 8 is as good or better than .NET 7 here, right?  I'd say it's probably not worthwhile unless we hear any noise about it.

PR #92838, opened 2023-09-29
Throw exception in TensorPrimitives for unsupported span overlaps

stephentoub:

### DIFF
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
@@ -1,6 +1,9 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
 namespace System.Numerics.Tensors
 {
     /// <summary>Performs primitive tensor operations over spans of memory.</summary>
@@ -10,6 +13,7 @@ namespace System.Numerics.Tensors
         /// <param name="x">The tensor, represented as a span.</param>
         /// <param name="destination">The destination tensor, represented as a span.</param>
         /// <exception cref="ArgumentException">Destination is too short.</exception>
+        /// <exception cref="ArgumentException"><paramref name="x"/> and <paramref name="destination"/> reference overlapping memory locations and do not begin at the same location.</exception>
         /// <remarks>
         /// <para>
         /// This method effectively computes <c><paramref name="destination" />[i] = MathF.Abs(<paramref name="x" />[i])</c>.
@@ -21,10 +25,6 @@ namespace System.Numerics.Tensors
         /// If a value is equal to <see cref="float.NegativeInfinity"/> or <see cref="float.PositiveInfinity"/>, the result stored into the corresponding destination location is set to <see cref="float.PositiveInfinity"/>.
         /// If a value is equal to <see cref="float.NaN"/>, the result stored into the corresponding destination location is the original NaN value with the sign bit removed.
         /// </para>
-        /// <para>
-        /// <paramref name="x"/> may overlap with <paramref name="destination" />, but only if the input and the output span begin at the same memory
-        /// location; otherwise, behavior is undefined. It is safe, for example, to use the same span for all span parameters.
-        /// </para>
         /// </remarks>
         public static void Abs(ReadOnlySpan<float> x, Span<float> destination) =>
             InvokeSpanIntoSpan<AbsoluteOperator>(x, destination);
@@ -35,14 +35,11 @@ namespace System.Numerics.Tensors
         /// <param name="destination">The destination tensor, represented as a span.</param>
         /// <exception cref="ArgumentException">Length of <paramref name="x" /> must be same as length of <paramref name="y" />.</exception>
         /// <exception cref="ArgumentException">Destination is too short.</exception>
+        /// <exception cref="ArgumentException"><paramref name="x"/> and <paramref name="destination"/> reference overlapping memory locations and do not begin at the same location.</exception>
+        /// <exception cref="ArgumentException"><paramref name="y"/> and <paramref name="destination"/> reference overlapping memory locations and do not begin at the same location.</exception>
         /// <remarks>
         /// <para>
         /// This method effectively computes <c><paramref name="destination" />[i] = <paramref name="x" />[i] + <paramref name="y" />[i]</c>.
-        /// </para>
-        /// <para>
-        /// <paramref name="x"/> and <paramref name="y"/> may overlap arbitrarily, but they may only overlap with <paramref name="destination" />
-        /// if the input and the output span begin at the same memory location; otherwise, behavior is undefined. It is safe, for example,
-        /// to use the same span for any subset of the span parameters, such as to perform an in-place operation.
         /// </para>
         /// <para>
         /// If either of the element-wise input values is equal to <see cref="float.NaN"/>, the resulting element-wise value is also NaN.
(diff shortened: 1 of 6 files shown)


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -1306,12 +1238,25 @@ public static void Tanh(ReadOnlySpan<float> x, Span<float> destination)
...
             {
                 destination[i] = MathF.Tanh(x[i]);
             }
         }
 
+        /// <summary>Throws an exception if the <paramref name="input"/> and <paramref name="output"/> spans overlap and don't begin at the same memory location.</summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static void ValidateInputOutputSpanNonOverlapping(ReadOnlySpan<float> input, Span<float> output)
+        {
+            if (!Unsafe.AreSame(ref MemoryMarshal.GetReference(input), ref MemoryMarshal.GetReference(output)) &&
```
EgorBo, line 1253: so there are cases where it's ok to have input and output to be the same reference? Is it true for all APIs?

EgorBo (approved):
Thanks! Feels safer now 🙂

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
EgorBo: btw, I'll take a look what we can improve in codegen for `MemoryExtensions.Overlaps`

stephentoub: > so there are cases where it's ok to have input and output to be the same reference? Is it true for all APIs?

Yes, all of the APIs support in-place now.

tannergooding, line 1253: Is this overly conservative?

That is, since the length of `input` is exact and determines how big `output` needs to be. But the length of `output` is not bounded, I think this would fail for a case such as `input.Length == 5` and `output.Length == 6` but where only element 5 overlaps with element 0 of `input`.

So, I think instead we need to slice `output` to be the same length as `input`.

stephentoub: Is that really worthwhile adding even more cost for? I agree what you cite is possible, but I'm not sure in what situation it would really matter.

tannergooding: Possibly not, but it's also not a lot of cost here.

We've already validated that `output.Length >= input.Length`, so `Overlaps` is already doing more than necessary here.

We could just have our own check here that is basically:
```csharp
private static bool Overlaps<T>(ReadOnlySpan<T> input, Span<T> output)
{
    ref T inputBase = ref MemoryMarshal.GetReference(input);
    ref T outputBase = ref MemoryMarshal.GetReference(output);

    if (Unsafe.AreSame(ref inputBase, ref outputBase))
    {
        // Same address, can never overwrite an input
        return false;
    }

    int length = input.Length;
    Debug.Assert(output.Length >= length);

    if (Unsafe.IsAddressLessThan(ref inputBase, ref outputBase))
    {
        // input starts in memory before output, so we have a bad overlap if `inputBase + length` is greater than
        // outputBase, since that means writing the first output will overwrite some latter input
        return Unsafe.IsAddressGreaterThan(ref Unsafe.Add(ref inputBase, length), ref outputBase);
    }

    // output starts in memory before input, this is technically safe since we will always have loaded more elements
    // from input than will be written to output and therefore can never overwrite something we still need to process
    // but, we don't support that today.

    Debug.Assert(Unsafe.IsAddressLessThan(ref outputBase, ref inputBase));
    return Unsafe.IsAddressGreaterThan(ref Unsafe.Add(ref outputBase, length), ref inputBase);
}
```

This gives us very nice and small codegen with no more than 2 comparisons for any path. It even makes it very easy to support the other safe kind of overlapping in the future if we desired (which, if we did support, would get this down to 1 branch and no more than 2 comparisons).

EgorBo: I'd leave as is and improve BCL's Overlap instead, e.g. it doesn't need those fast paths for empty spans IMO. With a little massage it shows decent codegen:
```
public static unsafe bool Overlaps<T>(thhis ReadOnlySpan<T> span, ReadOnlySpan<T> other)
{
    ref T pSpan = ref MemoryMarshal.GetReference(span);
    ref T pOther = ref MemoryMarshal.GetReference(other);
    ref T spanEnd = ref Unsafe.AddByteOffset(ref pSpan, (nint)span.Length * sizeof(T));
    ref T otherEnd = ref Unsafe.AddByteOffset(ref pOther, (nint)other.Length * sizeof(T));
    return Unsafe.IsAddressGreaterThan(ref spanEnd, ref pOther) &&
           Unsafe.IsAddressGreaterThan(ref otherEnd, ref pSpan);
}
```
```asm
; Method C:Test(System.ReadOnlySpan`1[float],System.ReadOnlySpan`1[float]):ubyte (FullOpts)
       mov      rax, bword ptr [rcx]
       mov      ecx, dword ptr [rcx+0x08]
       mov      r8, bword ptr [rdx]
       mov      edx, dword ptr [rdx+0x08]
       lea      rdx, bword ptr [r8+4*rdx]
       lea      rcx, bword ptr [rax+4*rcx]
       cmp      rcx, r8
       jbe      SHORT G_M25501_IG04
       cmp      rdx, rax
       seta     al
       movzx    rax, al
       jmp      SHORT G_M25501_IG05
G_M25501_IG04:  ;; offset=0x0024
       xor      eax, eax
G_M25501_IG05:  ;; offset=0x0026
       ret      
; Total bytes of code: 39
```

tannergooding: Sounds reasonable. The only downside is it makes it harder to support the "safe" overlapping scenario if we wanted it. But that's also not a huge deal.

jkotas: > return Unsafe.IsAddressGreaterThan(ref sEnd, ref oStart) && Unsafe.IsAddressGreaterThan(ref oEnd, ref sStart);

This does not look reliable. The GC can kick between the two `Unsafe.IsAddressGreaterThan` calls and reorder the spans in memory.

It is important to access the snap refs exactly once like what the current implementation does. (There are also other less performant options possible - pinning, etc.)

EgorBo: Ah, totally forgot about that, good point. Still, sounds like https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/MemoryExtensions.cs#L3020-L3023 this path is not necessary?

jkotas: I think deleting that condition would change the observable behavior. You can have an empty span that points into the middle of non-empty span. The API returns false in this case today. Deleting that condition would make the API to return true, in some of these cases.

I agree that it may be possible to tweak the implementation of Overlaps to have a slightly better perf.

tannergooding: I believe we can also special case our own scenario needed here, given we already know lengths match, to do a much simpler set of checks.

We might need to utilize `ByteOffset`, but in general it should be possible to get it down to no more than 2 branches/3 comparisons (or 1 branch, 2 comparisons if we allow the safe overlapping scenario).

stephentoub: I'll merge this to get the behavior in place. Can you follow-up with a change to the helper if that's desirable?

PR #92839, opened 2023-09-29
Remove fgUpdateFlowGraph from optOptimizeFlow

EgorBo:
Almost no diffs locally but with TP improvements.

### DIFF
--- a/src/coreclr/jit/optimizer.cpp
+++ b/src/coreclr/jit/optimizer.cpp
@@ -5350,11 +5350,8 @@ PhaseStatus Compiler::optOptimizeFlow()
     noway_assert(opts.OptimizationEnabled());
     noway_assert(fgModified == false);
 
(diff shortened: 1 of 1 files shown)


EgorBo:
Almost no size diffs (a few improvements in fact) with TP wins: https://dev.azure.com/dnceng-public/public/_build/results?buildId=423327&view=ms.vss-build-web.run-extensions-tab

PTAL @dotnet/jit-contrib

AndyAyersMS (approved):
Interesting. How did you find this?

I wonder what else we can stop doing.

EgorBo:
> Interesting. How did you find this?
> 
> I wonder what else we can stop doing.

Mostly just experimenting, we can probably also remove `fgReorderBlocks` from optimize flow before all the loop opts to recognize more loops - but that needs careful output diff analysis as those are big

PR #92841, opened 2023-09-29
Preinitialize pop/switch/Type.IsValueType

MichalStrehovsky:
These showed up in ASP.NET Stage1 and were low hanging enough.

Cc @dotnet/ilc-contrib


jkotas (approved):
Thanks

Issue #83988, opened 2023-03-27
[API Proposal] Add persist able AssemblyBuilder implementation


buyaa-n:
Contributes to [Support equivalent of AssemblyBuilder.Save to save in-memory IL to an assembly](https://github.com/dotnet/runtime/issues/62956) 

As per https://github.com/dotnet/runtime/issues/62956 we plan to a add managed implementation of `AssemblyBuilder` that could save in-memory IL to an assembly file/stream.

## Proposed design 1
Initial version would only support Save, no Run, therefore the factory method not accepting `AssemblyBuilderAccess` that defines the access mode for the assembly. It is the current prototype design.

```diff
namespace System.Reflection.Emit
{
+    public sealed class AssemblyFileBuilder : AssemblyBuilder
+    {
+        internal AssemblyFileBuilder(AssemblyName name, IEnumerable<CustomAttributeBuilder>? assemblyAttributes) { }

         // New static methods without AssemblyBuilderAccess, because initial version will only support Save
+        public static AssemblyFileBuilder DefineDynamicAssembly(AssemblyName name) { throw null; }
+        public static AssemblyFileBuilder DefineDynamicAssembly(AssemblyName name, Enumerable<CustomAttributeBuilder>? assemblyAttributes) { throw null; }
	
        // New instance methods
+        public void Save(Stream stream) { }
+        public void Save(string assemblyFileName) { } // same as in .NET Framework
+    }
}
```
### Usage example
```cs
using System.Reflection.Emit;

public class MyType
{
    public void MyMethod(AssemblyName assemblyName, string fileName)
    {
	AssemblyFileBuilder builder = AssemblyFileBuilder.DefineDynamicAssembly(assemblyName);
	// ... Add module and other members 
	builder.Save(fileName);
    }
}
```
## Alternative design 2:
The new persisted AssemblyBuilder type implementation will not exposed, instead we add `Save` into the abstract class AssemblyBuilder and `protected abstract SaveCore` method for implementation. Use the same factory method for creating AssemblyBuilder. This need for using reflection in order to create persisted AssemblyBuilder. 
```diff
namespace System.Reflection.Emit
{
    public abstract class AssemblyBuilder : Assembly
    {
+        public void Save(Stream stream) { }
+        public void Save(string assemblyFileName) { }
+        protected abstract void SaveCore(Stream stream);
    }

    [Flags]
    public enum AssemblyBuilderAccess
    {
        Run = 1,
+	Save = 2,
[...]

PR #92846, opened 2023-09-30
Minimal ILGenerator implementation for new AssemblyBuilder.Save

buyaa-n:
Basically, replacing the ILGenerator.Emit overloads implementaiton with `InstructionEncoder` which will be written into the file when saving the methods in a type using [MethodBodyStreamEncoder](https://learn.microsoft.com/dotnet/api/system.reflection.metadata.ecma335.methodbodystreamencoder)

Contributes to https://github.com/dotnet/runtime/issues/83988

### DIFF
--- a/src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
+++ b/src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
@@ -20,6 +20,7 @@ namespace System.Reflection.Emit
         internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
         {
             _methodBuilder = methodBuilder;
+            // For compat, runtime implementation doesn't throw for negative or zero value.
             _builder = new BlobBuilder(Math.Max(size, DefaultSize));
             _il = new InstructionEncoder(_builder, new ControlFlowBuilder());
         }
@@ -177,7 +178,6 @@ namespace System.Reflection.Emit
         {
             // Puts the opcode onto the IL stream followed by the metadata token
             // represented by str.
-
             ModuleBuilder modBuilder = (ModuleBuilder)_methodBuilder.Module;
             int tempVal = modBuilder.GetStringMetadataToken(str);
             _il.OpCode((ILOpCode)opcode.Value);
--- a/src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
+++ b/src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
@@ -1,0 +1,203 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Buffers.Binary;
+using System.Reflection.Metadata;
+using System.Reflection.Metadata.Ecma335;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace System.Reflection.Emit
+{
+    internal sealed class ILGeneratorImpl : ILGenerator
+    {
+        private const int DefaultSize = 16;
+        private readonly MethodBuilder _methodBuilder;
+        private readonly BlobBuilder _builder;
+        private readonly InstructionEncoder _il;
+        private bool _hasDynamicStackAllocation;
+
+        internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
+        {
+            _methodBuilder = methodBuilder;
+            _builder = new BlobBuilder(Math.Max(size, DefaultSize));
+            _il = new InstructionEncoder(_builder, new ControlFlowBuilder());
+        }
+
+        internal InstructionEncoder Instructions => _il;
+        internal bool HasDynamicStackAllocation => _hasDynamicStackAllocation;
+
+        public override int ILOffset => _il.Offset;
+
+        public override void BeginCatchBlock(Type? exceptionType) => throw new NotImplementedException();
+        public override void BeginExceptFilterBlock() => throw new NotImplementedException();
+        public override Label BeginExceptionBlock() => throw new NotImplementedException();
+        public override void BeginFaultBlock() => throw new NotImplementedException();
+        public override void BeginFinallyBlock() => throw new NotImplementedException();
+        public override void BeginScope() => throw new NotImplementedException();
+        public override LocalBuilder DeclareLocal(Type localType, bool pinned) => throw new NotImplementedException();
+        public override Label DefineLabel() => throw new NotImplementedException();
+
+        public override void Emit(OpCode opcode)
+        {
+            if (opcode == OpCodes.Localloc)
+            {
+                _hasDynamicStackAllocation = true;
+            }
+            _il.OpCode((ILOpCode)opcode.Value);
+        }
+
+        public override void Emit(OpCode opcode, byte arg)
+        {
+            _il.OpCode((ILOpCode)opcode.Value);
+            _builder.WriteByte(arg);
+        }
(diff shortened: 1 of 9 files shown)


src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -0,0 +1,203 @@
...
+        private const int DefaultSize = 16;
+        private readonly MethodBuilder _methodBuilder;
+        private readonly BlobBuilder _builder;
+        private readonly InstructionEncoder _il;
+        private bool _hasDynamicStackAllocation;
+
+        internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
+        {
+            _methodBuilder = methodBuilder;
+            _builder = new BlobBuilder(Math.Max(size, DefaultSize));
```
danmoseley, line 23: Should it throw on size < 1?

buyaa-n: For compat with the existing runtime ILGenerator:
https://github.com/dotnet/runtime/blob/a590cb4cfb9f1a66c043476695fd0e79835842eb/src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeILGenerator.cs#L90

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ModuleBuilderImpl.cs
```diff
@@ -180,11 +180,18 @@ internal void AppendMetadata()
...
         }
 
-        private void WriteMethods(TypeBuilderImpl typeBuilder, List<GenericTypeParameterBuilderImpl> genericParams)
+        private void WriteMethods(TypeBuilderImpl typeBuilder, List<GenericTypeParameterBuilderImpl> genericParams, MethodBodyStreamEncoder methodBodyEncoder)
         {
             foreach (MethodBuilderImpl method in typeBuilder._methodDefinitions)
             {
-                MethodDefinitionHandle methodHandle = AddMethodDefinition(method, method.GetMethodSignatureBlob(), _nextParameterRowId);
+                int offset = -1;
+                ILGeneratorImpl? il = method.ILGeneratorImpl;
```
steveharter, line 188: Check for `!method.IsAbstract` before setting variable for `method.ILGeneratorImpl;`?

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
jeffhandley: That might be worth a comment. It looks like the scenario is covered in a unit test, but the test could use a comment too.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ModuleBuilderImpl.cs
```diff
@@ -230,6 +237,14 @@ private void WriteMethods(TypeBuilderImpl typeBuilder, List<GenericTypeParameter
             }
         }
 
+        private static int AddMethodBody(MethodBuilderImpl method, ILGeneratorImpl il, MethodBodyStreamEncoder methodBodyEncoder) =>
+            methodBodyEncoder.AddMethodBody(
+                instructionEncoder: il.Instructions,
+                maxStack: 8, // TODO
```
steveharter, line 243: Per offline discussion, we should add an implementation to calculate this with this PR.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/MethodBuilderImpl.cs
```diff
@@ -98,7 +111,23 @@ protected override ParameterBuilder DefineParameterCore(int position, ParameterA
...
+        protected override ILGenerator GetILGeneratorCore(int size)
+        {
+            if (IsGenericMethod && !IsGenericMethodDefinition)
+            {
+                throw new InvalidOperationException();
+            }
+
+            if ((_methodImplFlags & MethodImplAttributes.CodeTypeMask) != MethodImplAttributes.IL ||
+                (_methodImplFlags & MethodImplAttributes.Unmanaged) != 0 ||
+                (_attributes & MethodAttributes.PinvokeImpl) != 0)
```
jkotas, line 123: The CoreLib implementation also has a check for DllImport here. Is it intentionally omitted?

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ModuleBuilderImpl.cs
jkotas: Do we even need `method.IsAbstract` check here? It does not look like right to throw away the IL body.

buyaa-n: Good point, the runtime implementation throws while populating the method bodies on `TypeBuilder.CreateTypeInfo()`:
https://github.com/dotnet/runtime/blob/087ca4662309f3db40b3096cbf89853d9c292e16/src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeTypeBuilder.cs#L1696
As we are not populating the `TypeInfo` similarly, probably we should check it and throw in `MethodBuilder.GetILGenerator()`?

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/MethodBuilderImpl.cs
buyaa-n: We are setting the attribute when `DllImportAttribute` is set, no need to keep another boolean flag for this https://github.com/dotnet/runtime/blob/42a6776e6cb1569fe277181f424ccbbe243cf386/src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/MethodBuilderImpl.cs#L111-L114

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ModuleBuilderImpl.cs
jkotas: I agree.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ModuleBuilderImpl.cs
steveharter: If we can set `maxstack == the number of opcodes` for now that should protect us against any `InvalidProgramException` thrown by the JIT etc.

buyaa-n: I can do that until I have got a better handling

steveharter (approved):
LGTM pending the discussion on a value for `maxstack` that shouldn't result in an exception if exceeded.

buyaa-n:
The failures not related and known

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -46,6 +49,10 @@ public override void Emit(OpCode opcode)
                 _hasDynamicStackAllocation = true;
             }
             _il.OpCode((ILOpCode)opcode.Value);
+
+            // TODO for now only count the Opcodes emitted, in order to correctly we might need to make below API public
```
buyaa-n, line 53:
```suggestion
            // TODO: for now only count the Opcodes emitted, in order to calculate it correctly we might need to make internal Opcode APIs public
```

PR #92852, opened 2023-09-30
JIT: Merge stores

EgorBo:
This PR merges consecutive stores, e.g.:
```cs
static void Test(char[] arr)
{
    if (arr.Length >= 4)
    {
        arr[0] = 'T';
        arr[1] = 'R';
        arr[2] = 'U';
        arr[3] = 'E';
    }
}
```

```diff
; Assembly listing for method C:Test(ushort[]) (FullOpts)
       mov      eax, dword ptr [rcx+0x08]
       cmp      eax, 4
       jl       SHORT G_M28053_IG04
-      mov      word  ptr [rcx+0x10], 84
-      mov      word  ptr [rcx+0x12], 82
-      mov      word  ptr [rcx+0x14], 85
-      mov      word  ptr [rcx+0x16], 69
+      mov      rax, 0x45005500520054
+      mov      qword ptr [rcx+0x10], rax
G_M28053_IG04:
       ret      
-; Total bytes of code 33
+; Total bytes of code 23
```

Another example:
```cs
void Test(Foo foo)
{
    foo.A = 0;
    foo.B = 0;
    foo.C = 0;
    foo.D = 0;
}

class Foo { public byte A, B, C, D; }
```
```diff
; Method Prog:Test(Foo):this (FullOpts)
-      mov      byte  ptr [rdx+0x08], 0
-      mov      byte  ptr [rdx+0x09], 0
-      mov      byte  ptr [rdx+0x0A], 0
-      mov      byte  ptr [rdx+0x0B], 0
+      xor      eax, eax
+      mov      dword ptr [rdx+0x08], eax
       ret      
-; Total bytes of code: 17
+; Total bytes of code: 6
```

All unaligned/nonatomic concerns were addressed in https://github.com/dotnet/runtime/pull/92939 (and SIMD support)


stephentoub:
Nice. Does this work with span in addition to array?

EgorBo:
> Nice. Does this work with span in addition to array?

Yes! Should work in a general case, structs, pointers, etc. E.g. for spans:
```cs
void Foo(Span<int> span)
{
    span[2] = 1000000;
    span[1] = 2000000;
}
```
```diff
; Method C:Foo(System.Span`1[int]):this (FullOpts)
       sub      rsp, 40
       mov      rax, bword ptr [rdx]
       mov      ecx, dword ptr [rdx+0x08]
       cmp      ecx, 2
       jbe      SHORT G_M18523_IG04
-      mov      dword ptr [rax+0x08], 0xF4240
-      mov      dword ptr [rax+0x04], 0x1E8480
+      mov      rcx, 0xF4240001E8480
+      mov      qword ptr [rax+0x04], rcx
       add      rsp, 40
       ret      
G_M18523_IG04:  ;; offset=0x0022
       call     CORINFO_HELP_RNGCHKFAIL
       int3     
; Total bytes of code: 40
```
(doesn't look to be useful in this example, but still).

Initially, I was optimizing this case:
```cs
static decimal Test() => 95.5M;
```
```diff
; Assembly listing for method C:Test():System.Decimal (Tier1)
-      mov      dword ptr [rcx], 0x10000
-      xor      eax, eax
-      mov      dword ptr [rcx+0x04], eax
+      mov      qword ptr [rcx], 0x10000
       mov      qword ptr [rcx+0x08], 955
       mov      rax, rcx
       ret      
; Total bytes of code 23,
```
And in ^ case we can probably even merge two 8-byte stores into one SIMD store.

EgorBo:
Made changes to support SIMD as well, but will do it in a separate PR, e.g.:

image

EgorBo:
[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=423852&view=ms.vss-build-web.run-extensions-tab) - no TP regressions. Actual diffs are 2-3x bigger with scaled indices and simd support but simd is a bit tricky so I didn't include it in the initial version.

@jakobbotsch @SingleAccretion @dotnet/jit-contrib PTAL whether I am missing something obvious or not.

ShreyasJejurkar:
Was just wondering what `rust` emits for the same code, it's doing below. Any idea why it's different? 
https://godbolt.org/z/doeaKn8WP

image

EgorBo:
> Was just wondering what `rust` emits for the same code, it's doing below. Any idea why it's different? https://godbolt.org/z/doeaKn8WP
> 
> image

char is 4 bytes in Rust, so 4*4 = 16 Bytes, hence, SIMD.
in .NET it's 2 bytes: 2*4 = 8 byres, onle 8-byte long is enough.

ShreyasJejurkar:
> in .NET it's 2 bytes: 2 * 4 = 8 bytes -> 8-byte long is enough

oh, that's why it's not SIMD in this case, just plain instruction?

src/coreclr/jit/lower.cpp
```diff
@@ -7802,6 +7802,313 @@ void Lowering::ContainCheckBitCast(GenTree* node)
...
+        }
+    }
+    else if (ind->Addr()->OperIs(GT_LCL_VAR))
+    {
+        // Address is just a local, no offset, scale is 1
+        data->baseAddr = ind->Addr();
+        data->index    = nullptr;
+        data->scale    = 1;
+        data->offset   = 0;
+    }
```
jakobbotsch, line 7884: These presumably need checks for address exposure, otherwise the first `STOREIND` could modify one of the locals used in a later one.

src/coreclr/jit/lower.cpp
```diff
@@ -7802,6 +7802,313 @@ void Lowering::ContainCheckBitCast(GenTree* node)
...
+        {
+            return;
+        }
+
+        // Both should be unused
+        LIR::Use use;
+        if (BlockRange().TryGetUse(ind, &use) || BlockRange().TryGetUse(prevInd, &use))
+        {
+            return;
+        }
```
jakobbotsch, line 8017: `GT_STOREIND` is not a value, it cannot have a use

src/coreclr/jit/lower.cpp
```diff
@@ -7802,6 +7802,313 @@ void Lowering::ContainCheckBitCast(GenTree* node)
...
+
+        // Update offset to be the minimum of the two
+        addr->SetOffset(min(prevData.offset, currData.offset));
+
+        // Update type for both STOREIND and val
+        ind->gtType         = newType;
+        ind->Data()->gtType = newType;
+
+        // We currently only support these constants for val
+        // Will be extended to GT_CNS_VEC in future.
```
jakobbotsch, line 8082:
```suggestion
```

src/coreclr/jit/lower.cpp
```diff
@@ -7802,6 +7802,313 @@ void Lowering::ContainCheckBitCast(GenTree* node)
...
+    //   STOREIND(short)
+    //     ...addr nodes...
+    //   STOREIND(short) <-- we're here
+    //
+    // First we merge two 'short' stores, then we merge the result with the 'int' store
+    // to get a single store of 8 bytes.
+    do
+    {
+        // This check is not really needed, just for better throughput.
+        // We only support these types for the initial version.
```
jakobbotsch, line 7976:
```suggestion
```

jakobbotsch (commented):
Looks fine to me with a couple of minor things. 

Does it help in micro benchmarks too?

EgorBo:
> Does it help in micro benchmarks too?

I don't expect too much, probably better with upcoming SIMD, but e.g.:
```cs
public class MainClass
{
    public Test[] _arr = Enumerable.Range(0, 1000).Select(_ => new Test()).ToArray();

    [Benchmark]
    public void Test()
    {
        foreach (var test in _arr)
        {
            test.a = 0;
            test.b = 0;
            test.c = 0;
            test.d = 0;
        }
    }
}

public class Test
{
    public int a, b, c, d;
}
```

| Method | Toolchain                   | Mean     | Error    | StdDev  | Ratio |
|------- |---------------------------- |---------:|---------:|--------:|------:|
| Test   | \Core_Root\corerun.exe      | 321.6 ns | 36.47 ns | 2.00 ns |  0.66 |
| Test   | \Core_Root_base\corerun.exe | 487.5 ns | 19.92 ns | 1.09 ns |  1.00 |

src/coreclr/jit/lower.cpp
EgorBo: Converted to an assert, just in case 🙂

src/coreclr/jit/lower.cpp
EgorBo: Thanks! Addressed, luckily it seems it didn't affect the diffs

src/coreclr/jit/lower.cpp
```diff
@@ -7802,6 +7802,312 @@ void Lowering::ContainCheckBitCast(GenTree* node)
...
+
+        if (!found)
+        {
+            // Unexpected node in-between.
+            return false;
+        }
+        prev = prev->gtPrev;
+    }
+
+    *prevTree = prev;
```
jakobbotsch, line 7921: I think most of this logic could be simplified by using `BlockRange::GetTreeRange` (if the range is not closed return false). The logic here is sort of a manual implementation of what that function does generally.

I mainly ask because I was thinking about what happens when the first node of the tree is the first node in the block. Doesn't `prev` become `nullptr` here?

src/coreclr/jit/lower.cpp
```diff
@@ -7802,6 +7802,312 @@ void Lowering::ContainCheckBitCast(GenTree* node)
...
+        {
+            return;
+        }
+
+        // Get coalescing data for the previous STOREIND
+        GenTreeStoreInd* prevInd = prevTree->AsStoreInd();
+        if (!GetStoreCoalescingData(comp, prevInd->AsStoreInd(), &prevData, &prevTree) || (prevTree == nullptr))
+        {
+            return;
+        }
```
jakobbotsch, line 8013: Doesn't this mean we skip cases where the two indirs are at the beginning of the BB?

src/coreclr/jit/lower.cpp
jakobbotsch: (It's checked by `LIR::Range::CheckLIR`, but an assert seems fine)

src/coreclr/jit/lower.cpp
EgorBo: Ah, nice! I wasn't aware of it. Addressed! It saved a ton.

src/coreclr/jit/lower.cpp
EgorBo: Fixed by moving to `BlockRange::GetTreeRange`

EgorBo: (very likely the first NODE is an IL_OFFSET anyway, but I guess it's not guaranteed)

src/coreclr/jit/lower.cpp
```diff
@@ -7802,6 +7802,285 @@ void Lowering::ContainCheckBitCast(GenTree* node)
...
+
+        // It's not a STOREIND - bail out.
+        if ((prevTree == nullptr) || !prevTree->OperIs(GT_STOREIND))
+        {
+            return;
+        }
+
+        // Get coalescing data for the previous STOREIND
+        GenTreeStoreInd* prevInd = prevTree->AsStoreInd();
+        if (!GetStoreCoalescingData(comp, prevInd->AsStoreInd(), &prevData) || (prevTree == nullptr))
```
jakobbotsch, line 7976:
```suggestion
        if (!GetStoreCoalescingData(comp, prevInd->AsStoreInd(), &prevData))
```

jakobbotsch (approved):

EgorBo:
Failure is https://github.com/dotnet/runtime/issues/92930

PR #92228, opened 2023-09-18
Add package readmes (continued)

ViktorHofer:
Continuation of https://github.com/dotnet/runtime/pull/91210

One of the top customer problems that package consumers are facing is lack of documentation. As such, we are driving an effort to increase the adoption and quality of NuGet package READMEs.

### Ask

To lead by example and ensure that our packages have the maximum impact, we would greatly appreciate your support in increasing the adoption and quality of NuGet package READMEs.

### Why

The README file is an essential part of your package as it provides important information to users and helps them understand what the package is and what it does quickly. Also, README is the first things for users when they view your package on NuGet.org and soon other tooling, it is crucial for package authors to write and include high-quality READMEs for their packages.

### How

This PR initializes READMEs for all the below listed packages. Please directly push to this branch and also use this PR by your POD area team to review changes. For a great example, please see [`System.Text.Json`'s package README](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/PACKAGE.md). Expect to spend ~30min of your time per README. Please try to keep documentation as minimal as possible to avoid duplicating information that is already available on docs.microsoft.com.
 
## Packages
_This list is sorted alphabetically._

| Package | | Status |
| :--- | :---: | :---: |
| Microsoft.Bcl.Cryptography | [Edit](https://github.com/dotnet/runtime/edit/PackageReadmesContinued/src/libraries/Microsoft.Bcl.Cryptography/src/PACKAGE.md) | ✅ |
| Microsoft.Extensions.Caching.Abstractions | [Edit](https://github.com/dotnet/runtime/edit/PackageReadmesContinued/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/PACKAGE.md) | ✅ |
| Microsoft.Extensions.DependencyInjection.Specification.Tests | [Edit](https://github.com/dotnet/runtime/edit/PackageReadmesContinued/src/libraries/Microsoft.Extensions.DependencyInjection.Specification.Tests/src/PACKAGE.md) | |
| Microsoft.Extensions.FileProviders.Abstractions | [Edit](https://github.com/dotnet/runtime/edit/PackageReadmesContinued/src/libraries/Microsoft.Extensions.FileProviders.Abstractions/src/PACKAGE.md) | ✅ |
[...]

PR #92859, opened 2023-10-01
Update package description used types for System.IO.Hashing

xoofx:
While installing the NuGet package, the recently added types were not added to the csproj/NuGet description. This PR fixes this.

Should be probably backported to .NET 8 branch?

### DIFF
--- a/src/libraries/System.IO.Hashing/src/System.IO.Hashing.csproj
+++ b/src/libraries/System.IO.Hashing/src/System.IO.Hashing.csproj
@@ -8,7 +8,11 @@
 
 Commonly Used Types:
 System.IO.Hashing.Crc32
-System.IO.Hashing.XxHash32</PackageDescription>
+System.IO.Hashing.Crc64
+System.IO.Hashing.XxHash128
+System.IO.Hashing.XxHash3
+System.IO.Hashing.XxHash32
+System.IO.Hashing.XxHash64</PackageDescription>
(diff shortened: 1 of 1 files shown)


stephentoub:
@ViktorHofer, what's the guidance on which / how many types to list in the commonly used section?

ViktorHofer:
> @ViktorHofer, what's the guidance on which / how many types to list in the commonly used section?

System.Text.Json's package readme lists seven "main types": https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/PACKAGE.md#main-types

That list of types should highlight the most important types from the package so that consumers can directly use them in code or look them up on docs.microsoft.com. Enumerating all public types isn't the intention here. Hope that helps.

FWIW with the shift to package readmes, we stopped enumerating these types in the package description (not consistent yet). I see that System.IO.Hashing doesn't have a package readme yet.

stephentoub:
Thanks.  Given that direction, seems like we should instead just be rewriting the readme, yes?

ViktorHofer:
> Given that direction, seems like we should instead just be rewriting the readme, yes?

Yes, we should create a package readme for this package and update the package description. We could do that as part of https://github.com/dotnet/runtime/pull/92228 or via a separate PR.

xoofx:
All good then if it is going to be handled more nicely, thanks!

Issue #81023, opened 2023-01-23
TaskCancelledException in HttpConnectionPool


sksk571:
### Description

We have a small web application that started sporadically throwing TaskCancelledException's after migration from .NET5 to .NET6 and later .NET7. The application consists of just one controller method that retrieves image from Amazon S3, resizes it and saves a resized copy back to S3. The application works fine under .NET5, but under .NET6 and .NET7 it sporadically throws thousands of TaskCancelledException errors from method HttpConnectionPool+HttpConnectionWaiter`1+<WaitForConnectionAsync>

### Configuration

- Version: .NET6+
- OS version: Alpine Linux 3.16
- Architecture: x64

### Regression?

This is a regression in .NET6+ compared to .NET5.

### Data

Dump files captured during one of the incidents contain thousands of async calls looking like
```
.00007f0f9ec03878 (0) System.Net.Http.HttpConnectionPool+HttpConnectionWaiter`1+<WaitForConnectionAsync>d__5[[System.__Canon, System.Private.CoreLib]]
..00007f0f9ec03918 (4) System.Net.Http.HttpConnectionPool+<SendWithVersionDetectionAndRetryAsync>d__84
...00007f0f9ec03a38 (0) System.Net.Http.DiagnosticsHandler+<SendAsyncCore>d__10
....00007f0f9ec03b00 (0) System.Net.Http.HttpClient+<<SendAsync>g__Core|83_0>d
.....00007f0f9ec03bc0 (0) Amazon.Runtime.HttpWebRequestMessage+<GetResponseAsync>d__20
```
and
```
.00007f0f9ec1c840 (1) System.Net.Http.HttpConnectionPool+<ConnectToTcpHostAsync>d__99
..00007f0f9ec1c910 (0) System.Net.Http.HttpConnectionPool+<ConnectAsync>d__98
...00007f0f9ec1c9d8 (0) System.Net.Http.HttpConnectionPool+<CreateHttp11ConnectionAsync>d__100
....00007f0f9ec1ca98 (0) System.Net.Http.HttpConnectionPool+<AddHttp11ConnectionAsync>d__74
```

The events from a trace captured using System.Net.Http provider:
[dotnet_20230119_154543_2847c867.9.excel.xlsx](https://github.com/dotnet/runtime/files/10480348/dotnet_20230119_154543_2847c867.9.excel.xlsx)

### Analysis

[...]

PR #34633, opened 2020-04-07
NameResolutionPal.Unix enabled async name resolution

gfoidl:
Fixes https://github.com/dotnet/runtime/issues/33378

Also

* added basic tests in `System.Net.NameResolution.Pal.Tests`
* same order of methods as in the Windows-Pal

---
Edit: this issue is fixed.

Marking this PR as draft, as there is some bug anywhere which I can't find.
To check the native implementation 8c955bc4bd1d0dbe4b2823f4327ad52bf9856fd7 adds some (c)tests -- this will be reverted later -- and they pass without any problem.
The managed implementation is analogous to the windows' one.

The bug causes either a hang or a segfault.
For the latter I can see GDB Output (see [core-dump.txt](https://github.com/dotnet/runtime/files/4443296/core-dump.txt)).

Maybe it's a dumb failure on my side, which I don't see atm, or something different.

If the test (added with 4614267d909794899188d28ae05bddcdd53e7b08) doesn't hang, the native side it will call into the OS with return code `0`. The crash happens anywhere after the OS-call and the invocation of the callback. Unfortunately I'm not able to debug it further (due to lack of knowledge here deep down that stack).

PR #92862, opened 2023-10-01
Start new per-host queue when an async over sync Dns call is cancelled

### LINKED ISSUE #92045: System.Net.Dns.RunAsync() does not decrement telemetry for canceled requests
If a call that ends up in `Dns.RunAsync()` is canceled, nothing ever calls `NameResolutionTelemetry.AfterResolution()` and so the variable `NameResolutionTelemetry._currentLookups` remains forever incremented by one.

The unmatched `BeforeResolution()` is called in `Dns.RunAsync()` here: https://github.com/dotnet/runtime/blob/9101b853a1f5194de3e7bc495454a57df0da300e/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs#L652C16-L652C16

image

### LINKED ISSUE #92054: Socket.ConnectAsync(EndPoint, CT) does not pass CT to DNS query, enabling code with many concurrent connections to the same DNS name to reach a state where no further connections get established because all DNS queries time out
Consider some code that is trying to establish 1000s of TCP connections to the same URL (e.g. a load test tool trying to simulate 8000 users, each with its own connection).

The internet is not perfect, so some (or even many) of these connections may fail and need to be retried. For example, the OS may fail to resolve a DNS query for whatever reason (maybe it got lost in the network, maybe the OS itself just has a 16-entry buffer and drops requests if too many are concurrently executing).

`Socket.ConnectAsync(EndPoint, CancellationToken)` takes a CancellationToken:

https://github.com/dotnet/runtime/blob/62e50b68d30c9638d3648bea9c1e575e84216f05/src/libraries/System.Net.Sockets/src/System/Net/Sockets/Socket.Tasks.cs#L86

One would expect to be able to use this to cancel connection attempts on timeout. Indeed, this is where the `CancellationToken` constructed from `SocketHttpHandler.ConnectTimeout` ends up in when we try to establish HTTP connections.

But let's zoom in on the DNS part of the connection attempt:

```
>	System.Net.NameResolution.dll!System.Net.Dns.GetHostEntryOrAddressesCoreAsync(string hostName, bool justReturnParsedIp, bool throwOnIIPAny, bool justAddresses, System.Net.Sockets.AddressFamily family, System.Threading.CancellationToken cancellationToken) Line 501	C#
 	System.Net.NameResolution.dll!System.Net.Dns.GetHostAddressesAsync(string hostNameOrAddress, System.Net.Sockets.AddressFamily family, System.Threading.CancellationToken cancellationToken) Line 241	C#
 	System.Net.Sockets.dll!System.Net.Sockets.SocketAsyncEventArgs.DnsConnectAsync(System.Net.DnsEndPoint endPoint, System.Net.Sockets.SocketType socketType, System.Net.Sockets.ProtocolType protocolType) Line 667	C#
 	System.Net.Sockets.dll!System.Net.Sockets.Socket.ConnectAsync(System.Net.Sockets.SocketAsyncEventArgs e, bool userSocket, bool saeaCancelable) Line 2697	C#
 	System.Net.Sockets.dll!System.Net.Sockets.Socket.AwaitableSocketAsyncEventArgs.ConnectAsync(System.Net.Sockets.Socket socket) Line 1182	C#
 	System.Net.Sockets.dll!System.Net.Sockets.Socket.ConnectAsync(System.Net.EndPoint remoteEP, System.Threading.CancellationToken cancellationToken) Line 101	C#
 	System.Net.Http.dll!System.Net.Http.HttpConnectionPool.ConnectToTcpHostAsync(string host, int port, System.Net.Http.HttpRequestMessage initialRequest, bool async, System.Threading.CancellationToken cancellationToken) Line 1600	C#
 	System.Net.Http.dll!System.Net.Http.HttpConnectionPool.ConnectAsync(System.Net.Http.HttpRequestMessage request, bool async, System.Threading.CancellationToken cancellationToken) Line 1526	C#
 	System.Net.Http.dll!System.Net.Http.HttpConnectionPool.CreateHttp11ConnectionAsync(System.Net.Http.HttpRequestMessage request, bool async, System.Threading.CancellationToken cancellationToken) Line 1631	C#
 	System.Net.Http.dll!System.Net.Http.HttpConnectionPool.AddHttp11ConnectionAsync(System.Net.Http.HttpConnectionPool.RequestQueue<System.Net.Http.HttpConnection>.QueueItem queueItem) Line 480	C#
 	System.Net.Http.dll!System.Net.Http.HttpConnectionPool.CheckForHttp11ConnectionInjection.AnonymousMethod__77_0((System.Net.Http.HttpConnectionPool thisRef, System.Net.Http.HttpConnectionPool.RequestQueue<System.Net.Http.HttpConnection>.QueueItem queueItem) state) Line 541	C#
```

One of the first things `Socket.ConnectAsync()` does it call `AwaitableSocketAsyncEventArgs.ConnectAsync()` to start the process of connecting (and after that waits for it to complete, with proper cancellation handling for the wait): https://github.com/dotnet/runtime/blob/62e50b68d30c9638d3648bea9c1e575e84216f05/src/libraries/System.Net.Sockets/src/System/Net/Sockets/Socket.Tasks.cs#L100

This just sets up some bookeeping and takes us to an internal overload of `Socket.ConnectAsync()`: https://github.com/dotnet/runtime/blob/62e50b68d30c9638d3648bea9c1e575e84216f05/src/libraries/System.Net.Sockets/src/System/Net/Sockets/Socket.cs#L2724

If we are dealing with a `DnsEndPoint` (i.e. the URL had a hostname instead of IP address), this now calls `DnsConnectAsync()`:

https://github.com/dotnet/runtime/blob/62e50b68d30c9638d3648bea9c1e575e84216f05/src/libraries/System.Net.Sockets/src/System/Net/Sockets/Socket.cs#L2762

`DnsConnectAsync()` will try to grab a CancellationToken for the DNS query from a variable called `_multipleConnectCancellation`. I do not understand what that is but in my experiments the value is simply null, so we end up not passing any CancellationToken to `Dns.GetHostAddressesAsync()`. The actual CancellationToken accompanying our `Socket.ConnectAsync()` we left behind several method calls ago.

https://github.com/dotnet/runtime/blob/62e50b68d30c9638d3648bea9c1e575e84216f05/src/libraries/System.Net.Sockets/src/System/Net/Sockets/SocketAsyncEventArgs.cs#L649-L659

image

In other words, even if a socket connection attempt is cancelled (e.g. due to timeout), the DNS query it triggered remains active. 

On its own, it just sounds inefficienct. However, When we combine this with https://github.com/dotnet/runtime/pull/49171, we get in trouble. We can now get into the following situation:

* A large queue of pending connection attempts has built up in this one-at-a-time DNS queue (e.g. because there was a delay and OS resolver needed 5-10 seconds of retries for a single request, while the app at the sime time was continuing in its job of creating 1000s of connections).
* Because the original slow request took a long time, the majority of the queue may already have timed out! The DNS requests are still in the queue and will still be executed because the CancellationToken is not passed to `Dns.GetHostAddressesAsync()`.
* Some of these DNS queries may also be slow (after all, there are bound to be some failures in thousands of them) and again require 5+ seconds of waiting before OS resolver gets a response after retries (or never?).
* Meanwhile, the HttpConnectionPool keeps retrying and making new connections but they all keep timing out because the DNS queries are all still sitting in the queue by the time the timeout elapses and the CancellationToken (which DNS is not listening to) is signaled.

In a sample scenario with 8000 HttpClients, after the initial slow DNS resolve starts the described chain of events, I am seeing average DNS resolve durations of 30-40 seconds (this includes the time spent in the queue).

Repro app: https://github.com/sandersaares/whatiswrongwithyourdog-dns-does-not-get-canceled/blob/main/Program.cs

Running this on a new D32 VM on Ubuntu 22 with .NET 7, I soon (depends on run, but no more than 30 seconds) start to see output similar to:

```
azureuser@hack01:~/33/net7.0$ dotnet ConsoleApp33.dll
Successfully completed 0 requests, last one took 0.0 ms and finished 4.6 ms ago. 0 failures seen. 0/8000 tasks successfully completed at least one request.
Successfully completed 62 requests, last one took 902.8 ms and finished 77.6 ms ago. 0 failures seen. 62/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 49.5 ms ago. 0 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 1049.7 ms ago. 6 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 2049.7 ms ago. 12 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 3049.4 ms ago. 15 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 4048.9 ms ago. 39 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 5049.6 ms ago. 184 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 6049.8 ms ago. 330 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 7049.8 ms ago. 468 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 8050.0 ms ago. 611 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 9049.9 ms ago. 763 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 10049.0 ms ago. 904 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 11049.7 ms ago. 1034 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 12049.6 ms ago. 1171 failures seen. 68/8000 tasks successfully completed at least one request.
Successfully completed 68 requests, last one took 1906.8 ms and finished 13049.9 ms ago. 1313 failures seen. 68/8000 tasks successfully completed at least one request.
```

None of the fresh (not yet connected) tasks at this point will successfully make an HTTP request - all retries will fail because they will all time out while sitting in the DNS queue.

`dotnet-counters` shows the queue size in this particular case to be over 1 minute!

```
[System.Net.NameResolution]
    Average DNS Lookup Duration (ms)                                  67,039.097
    Current DNS Lookups                                                7,932
    DNS Lookups Requested                                             22,315
```

Related to #81023 but different, as in my case I have a forever-loop of failed requests because the app keeps retrying fast enough to maintain critical mass in the DNS request queue.

antonfirsov:
We had several customer reports of #81023 -- long running `getaddrinfo` lookups stalling `Dns.*Async` calls as a result of #49171. This is a simple mitigation of the problem by starting a new per-hostname Task queue in case a cancellation fires and also fixing #92054 to make sure the `CancellationToken` flows down to `GetHostAddressesAsync`.

The downside of this mitigation is that it actually needs a cancellation (~`ConnectTimeout`) to work. IMO this is still better than not having a solution to the problem at .NET level at all. We could implement a more sophisticated fix (see #92863), but I'm assuming that we are looking for a low-risk change to service .NET 8 and maybe earlier versions.

Fixes #92054
Fixes #92045
Mitigates #81023 (doesn't fully fix the issue)

### DIFF
--- a/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
+++ b/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
@@ -666,8 +666,8 @@ namespace System.Net
                     {
                         using (cancellationToken.UnsafeRegister(static s =>
                         {
-                            // In case a request is cancelled, make sure we start queue for this key.
-                            // This helps avoiding congestion on stalled or long running requests.
+                            // In case a request is cancelled, start a new queue for this key.
+                            // This helps avoiding congestion when a getaddrinfo call runs for too long.
                             lock (s_tasks)
                             {
                                 s_tasks.Remove(s!);
--- a/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
+++ b/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
@@ -641,9 +641,7 @@ namespace System.Net
         /// Since this is doing synchronous work on a thread pool thread, we want to limit how many threads end up being
         /// blocked.  We could employ a semaphore to limit overall usage, but a common case is that DNS requests are made
         /// for only a handful of endpoints, and a reasonable compromise is to ensure that requests for a given host are
-        /// serialized.  Once the data for that host is cached locally by the OS, the subsequent requests should all complete
-        /// very quickly, and if the head-of-line request is taking a long time due to the connection to the server, we won't
-        /// block lots of threads all getting data for that one host.  We also still want to issue the request to the OS, rather
+        /// serialized. We also still want to issue the request to the OS, rather
         /// than having all concurrent requests for the same host share the exact same task, so that any shuffling of the results
         /// by the OS to enable round robin is still perceived.
         /// </remarks>
@@ -668,6 +666,8 @@ namespace System.Net
                     {
                         using (cancellationToken.UnsafeRegister(static s =>
                         {
+                            // In case a request is cancelled, make sure we start queue for this key.
+                            // This helps avoiding congestion on stalled or long running requests.
                             lock (s_tasks)
                             {
                                 s_tasks.Remove(s!);
@@ -691,16 +691,17 @@ namespace System.Net
                 }, key, cancellationToken, TaskContinuationOptions.DenyChildAttach, TaskScheduler.Default);
 
                 // If it's possible the task may end up getting canceled, it won't have a chance to remove itself from
-                // the dictionary if it is canceled, so use a separate continuation to do so.
+                // the dictionary and report AfterResolution() if it is canceled, so use a separate continuation to do so.
                 if (cancellationToken.CanBeCanceled)
                 {
-                    task.ContinueWith((task, key) =>
+                    task.ContinueWith(delegate
                     {
+                        NameResolutionTelemetry.Log.AfterResolution(key, startingTimestamp, false);
                         lock (s_tasks)
                         {
                             ((ICollection<KeyValuePair<object, Task>>)s_tasks).Remove(new KeyValuePair<object, Task>(key!, task));
                         }
-                    }, key, CancellationToken.None, TaskContinuationOptions.OnlyOnCanceled | TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
+                    }, CancellationToken.None, TaskContinuationOptions.OnlyOnCanceled | TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
                 }
 
                 // Finally, store the task into the dictionary as the current task for this key.
--- a/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
+++ b/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
@@ -666,7 +666,16 @@ namespace System.Net
                     Debug.Assert(!Monitor.IsEntered(s_tasks));
                     try
                     {
-                        return func(key, startingTimestamp);
+                        using (cancellationToken.UnsafeRegister(static s =>
+                        {
+                            lock (s_tasks)
+                            {
+                                s_tasks.Remove(s!);
+                            }
+                        }, key))
+                        {
+                            return func(key, startingTimestamp);
+                        }
                     }
                     finally
                     {
--- a/src/libraries/System.Net.Sockets/src/System/Net/Sockets/Socket.Tasks.cs
+++ b/src/libraries/System.Net.Sockets/src/System/Net/Sockets/Socket.Tasks.cs
@@ -97,7 +97,7 @@ namespace System.Net.Sockets
 
             saea.RemoteEndPoint = remoteEP;
 
-            ValueTask connectTask = saea.ConnectAsync(this);
+            ValueTask connectTask = saea.ConnectAsync(this, saeaCancelable: cancellationToken.CanBeCanceled);
             if (connectTask.IsCompleted || !cancellationToken.CanBeCanceled)
             {
                 // Avoid async invocation overhead
@@ -1202,11 +1202,11 @@ namespace System.Net.Sockets
                     ValueTask.FromException<int>(CreateException(error));
             }
 
-            public ValueTask ConnectAsync(Socket socket)
+            public ValueTask ConnectAsync(Socket socket, bool saeaCancelable)
             {
                 try
                 {
-                    if (socket.ConnectAsync(this, userSocket: true, saeaCancelable: false))
+                    if (socket.ConnectAsync(this, userSocket: true, saeaCancelable: saeaCancelable))
                     {
                         return new ValueTask(this, _mrvtsc.Version);
                     }


src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
```diff
@@ -641,9 +641,7 @@ private static bool LogFailure(object hostNameOrAddress, long? startingTimestamp
         /// Since this is doing synchronous work on a thread pool thread, we want to limit how many threads end up being
         /// blocked.  We could employ a semaphore to limit overall usage, but a common case is that DNS requests are made
         /// for only a handful of endpoints, and a reasonable compromise is to ensure that requests for a given host are
-        /// serialized.  Once the data for that host is cached locally by the OS, the subsequent requests should all complete
-        /// very quickly, and if the head-of-line request is taking a long time due to the connection to the server, we won't
-        /// block lots of threads all getting data for that one host.  We also still want to issue the request to the OS, rather
```
antonfirsov, line 646: > Once the data for that host is cached locally by the OS, the subsequent requests should all complete very quickly, and if the head-of-line request is taking a long time due to the connection to the server, we won't block lots of threads all getting data for that one host

Essentially, this turned out to be a false assumption. In Ubuntu there is no caching by default, and each `getaddrinfo` call will send an A & AAAA DNS request and take 5-50ms.

stephentoub: > This turned out to be a false assumption. In Ubuntu there is no caching by default, and each getaddrinfo call sends an A & AAAA DNS request and takes 5-50ms.

On what are you basing this? On Ubuntu for:
```C#
[Benchmark]
public object GetHostEntry() => Dns.GetHostEntry("www.microsoft.com");
```
I get:
|         Method |     Mean |
|--------------- |---------:|
| GetHostEntry | 498.9 us |

antonfirsov:
```C#
[Benchmark]
public void GetHostAddresses() => Dns.GetHostAddresses("fb.lt.main.ml.mdn.skype.net");
```

averages to `24.84 ms` in my local 22.04 environment and to `9.051 ms` in the Azure D32 VM from #92054.

MihaZupan: > `fb.lt.main.ml.mdn.skype.net`

That name resolves through a CNAME of `fb.lt.main.ml.mdn.trafficmanager.net`, which advertises a CNAME for `fileserver-backend-lb.lt-westus-01.ic3-middle-lane-main-lt.westus-test.cosmic-int.office.net` with a 0 second TTL. So my guess is the OS is just honoring the TTLs in this case.

The `www.microsoft.com` example above on the other hand uses non-0 TLLs.

stephentoub: You can also check whether DNS caching in your system is enabled, e.g.
```sh
systemctl is-active systemd-resolved
```
My point is simply that stating that it's a false assumption that OSes cache DNS information is itself a false assumption. Every major OS I'm aware of is capable of doing DNS caching. Whether it's enabled is a different question, though to my knowledge it is by default in Ubuntu 22.04.

antonfirsov: I was wrong thinking that answers are never cached on Ubuntu by default. However, to me the comment suggests that we can generally rely on short completion times because of the caching, which is also not true, eg. as Miha pointed out answers with TTL=0 are not cached, which then increases the probability of hitting #81023.

I will add the comment back & augment it.

src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
```diff
@@ -682,16 +691,17 @@ private static Task<TResult> RunAsync<TResult>(Func<object, long, TResult> func,
                 }, key, cancellationToken, TaskContinuationOptions.DenyChildAttach, TaskScheduler.Default);
 
                 // If it's possible the task may end up getting canceled, it won't have a chance to remove itself from
-                // the dictionary if it is canceled, so use a separate continuation to do so.
+                // the dictionary and report AfterResolution() if it is canceled, so use a separate continuation to do so.
                 if (cancellationToken.CanBeCanceled)
                 {
-                    task.ContinueWith((task, key) =>
+                    task.ContinueWith(delegate
```
stephentoub, line 697: Previously this wasn't allocating a new delegate and a new closure on every call. Now it is, as it's closing over the key and the task and the startingTimestamp. It'll even result in that allocation even if !CanBeCanceled, because state it's closing over is at the top-level method scope.

src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
```diff
@@ -666,7 +664,18 @@ private static Task<TResult> RunAsync<TResult>(Func<object, long, TResult> func,
                     Debug.Assert(!Monitor.IsEntered(s_tasks));
                     try
                     {
-                        return func(key, startingTimestamp);
+                        using (cancellationToken.UnsafeRegister(static s =>
+                        {
+                            // In case a request is cancelled, start a new queue for this key.
+                            // This helps avoiding congestion when a getaddrinfo call runs for too long.
```
stephentoub, line 670: What's the rationale for why a cancellation request means we no longer want to serialize?

src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
stephentoub: At a minimum you can pass the required state in via the object state parameter. That'll require a tuple allocation/boxing, but the one allocation for that is still better than the two allocations for the closure and delegate. And it'll make it specific to the CanBeCanceled case, whereas with how it's currently written the closure allocation will happen first thing in the method body.

src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
stephentoub: One of the original motivations for doing this queueuing was a DNS request that could get stalled, and thus all requests for that DNS endpoint would end up blocking all threads in the pool.  Doesn't this effectively put us back into the same place?

antonfirsov: From what I see, #48566 doesn't talk about stalled (up to several seconds) requests but about slow (tens of milliseconds?) requests using up the ThreadPool threads. This does not remove the serialization of those. The queue is only being reset when a cancellation kicks in which is typically triggered by a `ConnectTimeout`, which is usually in a multiple seconds range. If the customer mentioned in #34633 had stalled requests, they would have likely came back to us reporting #81023 much earlier.

cc @davidfow if you can recall how long did the slow requests take in #34633.  Without that I would assume it's a case similar to what we discussed in https://github.com/dotnet/runtime/pull/92862#discussion_r1342186061 and "slow" means that the requests are in the 5-50 ms range.

stephentoub: Right, so if every request for a given destination ends up taking a longer period of time and ends up being canceled as a result, every cancellation is going to cause subsequent requests to run concurrently with ones already there, and if it happens enough, won't it similarly spread out over all of the pool threads?

> when a cancellation kicks in which is typically triggered by a ConnectTimeout

That's one source. The Dns APIs that support cancellation are public.

antonfirsov: > if every request for a given destination ends up taking a longer period of time and ends up being canceled as a result, every cancellation is going to cause subsequent requests to run concurrently with ones already there, and if it happens enough, won't it similarly spread out over all of the pool threads?

Yes, but how likely is this scenario, and is it more probable than hitting #81023? To me this looks like having all requests to the same host stalled for a long enough period to starve the ThreadPool is a catsrophic event which would have serious impact on the application's funcionality even without the pool starvation, and the customer has to go and fix the environmental setup anyways. Occasional stalled requests are more common IMO, and the problem is that they are currently killing a possibly long chain of follow-up requests.

> That's one source. The Dns APIs that support cancellation are public.

IMO this doesn't really change things. Regularly occuring cases of cancellation are be most likely timeouts, and cancelling a significant percentage of DNS lookups by setting an overly short timeout doesn't look like a sane scenario to me.

stephentoub: > Occasional stalled requests are more common IMO

What is the cause of the stalled requests you're seeing that makes this the case?

antonfirsov: In #92054 we see the OS resolver (succesfully) retrying a query after the default `timeout` of 5 seconds, and `getaddrinfo` is blocking for this 5sec period. The retries are sporadic and do not come in bursts. They are caused by a packet loss in my understanding, which shouldn't be very uncommon.

I don't have such confident information from the other cases, because this is typcially happening in production systems in a sporadic manner. Currently I'm working with @matt-augustine checking if a private build of the changes in this PR helps with [their symptoms](https://github.com/dotnet/runtime/issues/81023#issuecomment-1730141380). Would that bring enough confidence for you that the change is net positive?

antonfirsov:
Currently https://github.com/dotnet/runtime/issues/81023 is no longer blocking the customers who originally reported the issue. Closing the PR in favor of a broader investigation into potential DNS improvements for 9.0.

PR #92864, opened 2023-10-01
Remove base type rooting for types in rooted assemblies

### LINKED ISSUE #92271: NativeAOT's implementation of `TrimmerRootAssembly` is too aggressive
For example:

TestApp.csproj (normal console app)
```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
  </PropertyGroup>

  <ItemGroup>
    <ProjectReference Include="..\TestLib\TestLib.csproj" />
    <TrimmerRootAssembly Include="TestLib" />
  </ItemGroup>

</Project>
```

And then the `TestLib` contains
```csharp
using System.Runtime.Loader;

namespace TestLib;

internal class Derived : AssemblyLoadContext
{
}
```

Now running (on the TestApp):
```
dotnet publish --self-contained /p:PublishTrimmed=true
```

The output will contain `AssemblyLoadContext` class in CoreLib, but for example the `AssemblyLoadContext.GetAssemblyName` method will be missing from it (since it's not used anywhere).

Now publishing the same app with NativeAOT:
```
dotnet publish /p:PublishAot=true
```

Using sizoscope tells us that `AssemblyLoadContext.GetAssemblyName` is actually preserved in the app.

In short, the NativeAOT implementation of assembly rooting seems to be rooting too much. For example it seems to fully root all base types of the types from the target assembly, including all members on such base types.

This is not a problem necessarily for cases where `TrimmerRootAssembly` is used as a workaround to trim incompatibilities (it's adds more code), but it can matter more when it's used in a test app to [validate trim/AOT compatibility of a library](https://review.learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?branch=main&branchFallbackFrom=pr-en-us-36487&tabs=net8plus&pivots=dotnet-8-0#csproj-file).

It would also be nice to make the behavior consistent between `illink` and `ilc`.

vitek-karas:
When rooting entire assemblies NativeAOT also roots all base types of all the types in such assembly. Regardless of which assembly the base type comes from. Historically this was necessary to make certain combinations of generic instantiations to work, but that's no longer the case. The compiler has better mechanism of tracking necessary things for generics now.

This rooting behavior brings in more code than necessary. This is specifically problematic when using the assembly rooting as a mechanism to test library trim/AOT compatibility. If the tested library depends on another library which has some incompatible code in it, but it's not used, ideally the compiler should remove the unused code and thus not see the incompatibilities. Rooting entire base types sometimes breaks that behavior and produces unnecessary warnings.

The product change is really just "don't root the base type", all of the rest of the change is tests. Modified existing test to validate more things around rooting behavior across assemblies. And then infrastructure changes to make it possible to use this test from NativeAOT.

This brings the behavior of NativeAOT and trimmer much closer with regard to assembly rooting.

Fixes https://github.com/dotnet/runtime/issues/92271.

### DIFF
--- a/src/libraries/System.Security.Cryptography/tests/DSATests.cs
+++ b/src/libraries/System.Security.Cryptography/tests/DSATests.cs
@@ -171,7 +171,7 @@ namespace System.Security.Cryptography.Tests
             public override void ImportParameters(DSAParameters parameters) => _dsa.ImportParameters(parameters);
             public override bool VerifySignature(byte[] rgbHash, byte[] rgbSignature) => _dsa.VerifySignature(rgbHash, rgbSignature);
             protected override byte[] HashData(Stream data, HashAlgorithmName hashAlgorithm) =>
-                (byte[])_dsa.GetType().GetMethod(
+                (byte[])typeof(DSA).GetMethod(
                     nameof(HashData),
                     BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance,
                     null,
@@ -179,7 +179,7 @@ namespace System.Security.Cryptography.Tests
                     null)
                 .Invoke(_dsa, new object[] { data, hashAlgorithm });
             protected override byte[] HashData(byte[] data, int offset, int count, HashAlgorithmName hashAlgorithm) =>
-                (byte[])_dsa.GetType().GetMethod(
+                (byte[])typeof(DSA).GetMethod(
                     nameof(HashData),
                     BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance,
                     null,
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/RootingHelpers.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/RootingHelpers.cs
@@ -40,15 +40,6 @@ namespace ILCompiler
                 rootProvider.AddReflectionRoot(type, reason);
             }
 
-            // Also root base types. This is so that we make methods on the base types callable.
-            // This helps in cases like "class Foo : Bar<int> { }" where we discover new
-            // generic instantiations.
-            TypeDesc baseType = type.BaseType;
-            if (baseType != null)
-            {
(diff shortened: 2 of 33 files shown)


MichalStrehovsky (approved):
Looks good but tests might need adjusting. Please also run nativeaot-outerloop before mergimg

vitek-karas:
I changed it to only "Stop" the old behavior for rooted assemblies. rd.xml rooting will remain as before.
It doesn't feel worth to possible break people with this change for rd.xml.

Updated the trimming tests accordingly.

vitek-karas:
@MichalStrehovsky could you please take another look - specifically at the test changes.

src/libraries/System.Security.Cryptography/tests/DSATests.cs
```diff
@@ -171,15 +171,15 @@ protected override void Dispose(bool disposing)
             public override void ImportParameters(DSAParameters parameters) => _dsa.ImportParameters(parameters);
             public override bool VerifySignature(byte[] rgbHash, byte[] rgbSignature) => _dsa.VerifySignature(rgbHash, rgbSignature);
             protected override byte[] HashData(Stream data, HashAlgorithmName hashAlgorithm) =>
-                (byte[])_dsa.GetType().GetMethod(
+                (byte[])typeof(DSA).GetMethod(
```
MichalStrehovsky, line 174: There are multiple implementations of DSA/ECDSA so this is technically getting a different method, but since it's virtual it probably doesn't matter. Just wanted to point out for double checking.

MichalStrehovsky (approved):
Still looks good, thanks!

src/libraries/System.Security.Cryptography/tests/DSATests.cs
vitek-karas: Thanks - I was aware of that, but as you stated, the method is declared as virtual on the `DSA` class, so unless some derived class does `new` this will work (and `new` would be really weird for a protected virtual method). If this was in product I would be much more worried, but for tests I think this is perfectly OK.

PR #92871, opened 2023-10-02
[browser][nodejs] keep runtime alive for JSExport calls

### LINKED ISSUE #92713: [wasm] Invoking async [JSExport] and [JSImport] kills the process on node, but works in browser
Given following in C#:
```csharp

public static partial class Program
{
    public static void Main () { }

    [JSExport]
    public static Task TestAsyncVoid () => Task.Delay(1);
}
```

When invoking the method in node:
```js
const runtime = await dotnet.withExitCodeLogging().withConfig(config).create();
await runtime.runMain("EntryAssembly.dll", []);
const exports = await runtime.getAssemblyExports("EntryAssembly.dll");
await exports.Program.TestAsyncVoid();
console.log("finished");
```
— runtime exits without printing "finished" with the following log:
```
WASM EXIT undefined
WASM EXIT 0
```

Same story with async `[JSImport]`'s. Basically, any kind of concurrency doesn't work in node.

I've also tried testing with bun, but it behaved the same.

At the same time all this works fine in browser.

Is there something I'm missing? Maybe some kind of switch or option when creating the runtime?

Here is the repro (breaks on async calls in node, but works fine in browser): https://github.com/Elringus/JSInteropTest

.NET 8 RC1
NodeJS v20.5.1

pavelsavara:
fixes https://github.com/dotnet/runtime/issues/92713

### DIFF
--- a/src/mono/wasm/runtime/startup.ts
+++ b/src/mono/wasm/runtime/startup.ts
@@ -5,7 +5,7 @@ import MonoWasmThreads from "consts:monoWasmThreads";
 import WasmEnableLegacyJsInterop from "consts:wasmEnableLegacyJsInterop";
 
 import { DotnetModuleInternal, CharPtrNull } from "./types/internal";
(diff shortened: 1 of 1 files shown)


maraf (approved):

pavelsavara:
Known issue https://github.com/dotnet/runtime/issues/92885

PR #92875, opened 2023-10-02
Fix zero/sign extend casts on arm64

### LINKED ISSUE #92865: JIT Arm64: Incorrectly optimizes out narrowing
### Description

Arm64 JIT in release mode eliminates a long to int conversion when used to access an array, even though the upper bits of the long may contain data. This causes an AccessViolationException to occur.

### Reproduction Steps

Run the test case (reduced from [WattleScript](https://github.com/WattleScript/wattlescript/blob/af640a2a91512f21603d91ffd1cd756a7c5bd552/src/WattleScript.Interpreter/Execution/VM/Instruction.cs)) as a console app in release mode `dotnet run -c Release`

```cs
struct TwoValues
{
    ulong _data;
    public int NumVal => (int) (_data >> 7);
    
    public TwoValues(int numval)
    {
    	_data = (ulong) (0xFFFUL << 39) |
		((ulong) (uint) numval) << 7;
    }
}

static class MainClass
{
	static readonly string[] strings = {
		"hello",
		"world",
		"37",
		"times"
	};
	static TwoValues[] data = new TwoValues[6];
	{
		data[3] = new TwoValues(0);
		data[4] = new TwoValues(1);
		Run(3);
		Run(4);
	}
	
	static void Run(int index)
	{
		Console.WriteLine(strings[data[index].NumVal]);
	}
}
```

### Expected behavior

Program should print:

hello
world

x86_64 release mode behaviour, and all platforms debug behaviour

### Actual behavior

Program generates access violation exception

### Regression?

Worked ok with .NET 6

### Known Workarounds

Narrow type manually with `& 0xFFFFFFFF`

### Configuration

Ubuntu 23.04
dotnet SDK 7.0.111
arm64 (ARMv8.0, Raspberry Pi 400)

### Other information

_No response_

EgorBo:
Fixes https://github.com/dotnet/runtime/issues/92865


src/coreclr/jit/codegenarmarch.cpp
```diff
@@ -1706,17 +1708,45 @@ void CodeGen::genCodeForIndexAddr(GenTreeIndexAddr* node)
...
+        {
+#if defined(TARGET_ARM64)
+            if (scale <= 4)
+            {
+                // target = base + index<<scale
+                GetEmitter()->emitIns_R_R_R_I(INS_add, emitActualTypeSize(node), node->GetRegNum(), base->GetRegNum(),
+                                              indexReg, scale, INS_OPTS_UXTW);
+            }
+            else
+#endif
```
jakobbotsch, line 1721: The entire `!index->TypeIs(TYP_I_IMPL)` case can be under this #ifdef.

jakobbotsch (approved):

src/coreclr/jit/codegenarmarch.cpp
jakobbotsch: (As it is it looks like there is a small arm32 bug, since it will enter the if for a small-typed load)

PR #92399, opened 2023-09-21
[wasm] Having space in path fails builds

ilonatommy:
Investigating https://github.com/dotnet/runtime/issues/92335.

- Change `wbt` -> `wbt artifacts` e.g. `artifacts/bin/Wasm.Build.Tests/Release/net8.0/browser-wasm/wbt/blz_checkfingerprinting_Debug_gqyxbvpe_hfq` -> `artifacts/bin/Wasm.Build.Tests/Release/net8.0/browser-wasm/wbt artifacts/blz_checkfingerprinting_Debug_gqyxbvpe_hfq`
The change is not in the immediate parent directory but still fails BuildPublishTests.
Repro:
publish with AOT on Windows/Linux/iOS.

error (Windows):
```
C:\Program Files\dotnet\packs\Microsoft.NET.Runtime.WebAssembly.Sdk\8.0.0-rc.2.23479.6\Sdk\WasmApp.Native.targets(682,5
): error : Precompiling failed for C:\repos\with spaces\obj\Release\net8.0\wasm\for-publish\aot
-in\aot-instances.dll with exit code 1. [C:\repos\with spaces\withspaces.csproj]
C:\Program Files\dotnet\packs\Microsoft.NET.Runtime.WebAssembly.Sdk\8.0.0-rc.2.23479.6\Sdk\WasmApp.Native.targets(682,5
): error : Can not open image C:\repos\with [C:\repos\with spaces\withs
paces.csproj]
```
in the default `WasmDedup=true` is caused by incorrect argv parsing in `driver.c` for arguments with spaces.

When dedup disabled or on Linux, error is:
```
# project path: /workspaces/spaces/with spaces/withSpaces.csproj
/usr/local/dotnet/7.0.306/packs/Microsoft.NET.Runtime.WebAssembly.Sdk/7.0.13/Sdk/WasmApp.Native.targets(326,5): error : clang version 15.0.0
/usr/local/dotnet/7.0.306/packs/Microsoft.NET.Runtime.WebAssembly.Sdk/7.0.13/Sdk/WasmApp.Native.targets(326,5): error : Target: wasm32-unknown-emscripten
/usr/local/dotnet/7.0.306/packs/Microsoft.NET.Runtime.WebAssembly.Sdk/7.0.13/Sdk/WasmApp.Native.targets(326,5): error : Thread model: posix
[...]

Issue #89073, opened 2023-07-18
icudt_EFIGS.dat blocked by firewall


socreative:
We have integrated the dotnet.js Wasm runtime (.NET 8) into our existing Angular front-end app but we're having issue with the **icudt_EFIGS.dat** file being blocked by the firewall. 

`Failed to load resource: the server responded with a status of 403 (forbidden)`

Is there a workaround for this? A way to use another extension for that file maybe?

am11:
One approach is to use custom file https://www.meziantou.net/reducing-blazor-application-size-by-providing-custom-icu-data.htm#use-custom-icu-data (e.g. by downloading the prebuilt file in on your build server and pointing to it in:
`<Target Name="UseCustomICU" AfterTargets="ResolveRuntimePackAssets">`).

radical:
cc @ilonatommy @maraf

maraf:
@ts-indikaf The `ReferenceCopyLocalPaths` needs to be put in `<ItemGroup>`, because it is an item collection.

ilonatommy:
> We have integrated the dotnet.js Wasm runtime (.NET 8) into our existing Angular front-end app but we're having issue with the **icudt_EFIGS.dat** file being blocked by the firewall.

Does it also happen for full ICU (**icudt.dat**)? To switch on full ICU, add `<WasmIncludeFullIcuData>true</WasmIncludeFullIcuData>` to pure wasm browser app or `<BlazorWebAssemblyLoadAllGlobalizationData>true</BlazorWebAssemblyLoadAllGlobalizationData>` to Blazor wasm app.

ilonatommy:
> Is there a workaround for this? A way to use another extension for that file maybe?

The only other extension that ICU theoretically supports is `.so`. Before we investigate this option: would your firewall block it as well?

ts-indikaf:
@ilonatommy ` .so` files are okay but `.dll` are blocked. Thanks

maraf:
@ts-indikaf You can use a JavaScript API that was added in .NET 8. I have create a sample https://github.com/maraf/dotnet-wasm-rename-icu

ts-indikaf:
@maraf - Many thanks for the solution. Unfortunately, it throws the following run time error from the dotnet.js file. I'm on dotnet8 preview6 and I have installed wasm-tools as well.
![image](https://github.com/dotnet/runtime/assets/66067000/8fa06366-f82e-4cff-ab8d-e9688be69113)

[...]

PR #92880, opened 2023-10-02
[wasm] Fix AOT for projects with space in path/name

maraf:
Relates to https://github.com/dotnet/runtime/issues/89073 https://github.com/dotnet/runtime/pull/92399


lewing (dismissed):
Just don't try %20 in the path

Issue #92873, opened 2023-10-02
[mono][jit] Zero SIMD vectors with movi on arm64


jandupej:
The reverse engineering document (https://dougallj.github.io/applecpu/firestorm.html) does not indicate that M1 recognizes `eor v0,v0,v0` as a zeroing idiom (like an x64 CPU would). Instead, there is the `movi` instruction (https://developer.arm.com/documentation/dui0801/l/A64-SIMD-Vector-Instructions/MOVI--vector---A64-?lang=en) that can initialize all lanes of a SIMD vector. Dougall Johnson's document shows that for zero, this is internally resolved with register renaming.

Also use this for `OP_XONES`.

EgorBo:
Oh, interesting, looks like `mov x, #0` is also better than `mov x, xzr` on Apple, e.g.: https://godbolt.org/z/6rh4rn5G3

PR #92882, opened 2023-10-02
[mono][jit] Arm64 SIMD regs are now zeroed with movi instead of eor

jandupej:
Dougall Johnson's reverse engineering document (https://dougallj.github.io/applecpu/firestorm.html) shows that Apple M1 may not recognize `eor v0,v0,v0` as a zeroing idiom like x64 CPUs would. The instruction would thus have to be executed normally, taking up an execution port and a latency of 2 (perhaps even waiting for previous register value). Instead there is `movi` which can broadcast an immediate Byte across a vector register. This instruction is handled by register renaming, as per the document. This PR moves vector zeroing (an initialization to all 1) to `movi`.

Addresses https://github.com/dotnet/runtime/issues/92873

### DIFF
--- a/src/mono/mono/mini/mini-arm64.c
+++ b/src/mono/mono/mini/mini-arm64.c
@@ -3933,13 +3933,15 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
 
 		case OP_XZERO:
 			if (ins->klass && mono_class_value_size (ins->klass, NULL) == 8)
-				arm_neon_eor_8b (code, dreg, dreg, dreg);
+				arm_neon_movi_b (code, VREG_LOW, dreg, 0);
 			else
-				arm_neon_eor_16b (code, dreg, dreg, dreg);
+				arm_neon_movi_b (code, VREG_FULL, dreg, 0);
 			break;
 		case OP_XONES:
-			arm_neon_eor_16b (code, dreg, dreg, dreg);
-			arm_neon_not_16b (code, dreg, dreg);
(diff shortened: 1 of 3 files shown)


vargaz (approved):

src/mono/mono/mini/mini-arm64.c
```diff
@@ -3933,13 +3933,15 @@ mono_arch_output_basic_block (MonoCompile *cfg, MonoBasicBlock *bb)
...
-				arm_neon_eor_16b (code, dreg, dreg, dreg);
+				arm_neon_movi_b (code, VREG_FULL, dreg, 0);
 			break;
 		case OP_XONES:
-			arm_neon_eor_16b (code, dreg, dreg, dreg);
-			arm_neon_not_16b (code, dreg, dreg);
+			if (ins->klass && mono_class_value_size (ins->klass, NULL) == 8)
+				arm_neon_movi_b (code, VREG_LOW, dreg, 0xff);
+			else
+				arm_neon_movi_b (code, VREG_FULL, dreg, 0xff);
```
matouskozak, line 3944: Could this be replaced by `arm_neon_movi_b (code, get_vector_size_macro (ins), dreg, 0xff)`?

matouskozak (approved):
LGTM!

src/mono/mono/mini/mini-arm64.c
fanyang-mono: +1

jandupej: Absolutely!

jandupej:
CI green, merging.

PR #92889, opened 2023-10-02
Implement overrides for Span and ReadOnlySpan on SslStream

### LINKED ISSUE #79072: SslStream does not have Read override with Span
I bump to it when investigating #78586
The exception looks like:
```
Unhandled exception. System.NotSupportedException:  This method may not be called when another read operation is pending.
   at System.Net.Security.SslStream.ReadAsyncInternal[TIOAdapter](Memory`1 buffer, CancellationToken cancellationToken)
   at System.Runtime.CompilerServices.PoolingAsyncValueTaskMethodBuilder`1.StateMachineBox`1.System.Threading.Tasks.Sources.IValueTaskSource<TResult>.GetResult(Int16 token)
   at System.Net.Security.SslStream.Read(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.Stream.Read(Span`1 buffer)
   at System.Net.Http.HttpConnection.Read(Span`1 destination)
   at System.Net.Http.HttpConnection.ChunkedEncodingReadStream.Read(Span`1 buffer)
   at System.Net.Http.HttpBaseStream.Read(Byte[] buffer, Int32 offset, Int32 count)
```

Because the overload is missing, we hit the base class that would rent buffer, read to it and then copy data over to the span.

https://github.com/dotnet/runtime/blob/067e43a1363f6d1748acc679f1804cd48299ba89/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs#L790-L808

rzikm:
Closes #79072.

Used same technique as used in System.Security.Cryptography to bridge Span<byte> code to Memory<byte> code.

First commit is a version without `PointerMemoryManager` pooling, and would incur allocation of the memory manager on each Read(Span<byte>) and Write(ROS<byte>) call.

Second commit introduces the pooling. Currently, the `PointerMemoryManager<T>` type does not support replacing the inner pointer and it did not sound like a safe feature in general, so I made a copy for internal SslStream use.

### DIFF
--- a/src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
+++ b/src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
@@ -955,6 +955,7 @@ namespace System.Net.Security
 
         // (non-generic) copy of the PointerMemoryManager<T> which supports resetting the stored
         // pointer to allow pooling its instances instead of allocating a new one per Read/Write call.
+        // The memory ponted to by the intenal poiner is assumed to be externally pinned (or naive memory).
         internal sealed unsafe class PoolingPointerMemoryManager : MemoryManager<byte>
         {
             private void* _pointer;
@@ -977,11 +978,13 @@ namespace System.Net.Security
 
             public override MemoryHandle Pin(int elementIndex = 0)
             {
-                throw new NotSupportedException();
+                // memory assumed to be pinned already
+                return new MemoryHandle(_pointer, default, null);
             }
 
             public override void Unpin()
             {
+                // nop
             }
         }
     }
--- a/src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
+++ b/src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
@@ -717,7 +717,7 @@ namespace System.Net.Security
             }
         }
 
-        private static unsafe PoolingPointerMemoryManager RentPointerMemoryManager(ref PoolingPointerMemoryManager? field, void* pointer, int length)
+        private static unsafe PoolingPointerMemoryManager RentPointerMemoryManager(ref PoolingPointerMemoryManager? field, byte* pointer, int length)
         {
             // we get null when called for the first-time, or concurrent read or write operation
             var manager = Interlocked.Exchange(ref field, null) ?? new PoolingPointerMemoryManager();
@@ -958,14 +958,14 @@ namespace System.Net.Security
         // The memory ponted to by the intenal poiner is assumed to be externally pinned (or naive memory).
         internal sealed unsafe class PoolingPointerMemoryManager : MemoryManager<byte>
         {
-            private void* _pointer;
+            private byte* _pointer;
             private int _length;
 
             protected override void Dispose(bool disposing)
             {
             }
 
-            public void Reset(void* pointer, int length)
+            public void Reset(byte* pointer, int length)
             {
                 _pointer = pointer;
                 _length = length;
@@ -979,7 +979,7 @@ namespace System.Net.Security
             public override MemoryHandle Pin(int elementIndex = 0)
             {
                 // memory assumed to be pinned already
-                return new MemoryHandle(_pointer, default, null);
+                return new MemoryHandle(_pointer + elementIndex, default, null);
             }
 
             public override void Unpin()
--- a/src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
+++ b/src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
@@ -762,10 +762,10 @@ namespace System.Net.Security
             // Otherwise, fall back to reading a byte via Read, the same way Stream.ReadByte does.
             // This allocation is unfortunate but should be relatively rare, as it'll only occur once
             // per buffer fill internally by Read.
-            Span<byte> oneByte = stackalloc byte[1];
-            int bytesRead = Read(oneByte);
+            byte oneByte = default;
+            int bytesRead = Read(new Span<byte>(ref oneByte));
             Debug.Assert(bytesRead == 0 || bytesRead == 1);
-            return bytesRead == 1 ? oneByte[0] : -1;
+            return bytesRead == 1 ? oneByte : -1;
         }
 
         public override unsafe int Read(Span<byte> buffer)
@@ -798,12 +798,7 @@ namespace System.Net.Security
             return vt.GetAwaiter().GetResult();
         }
 
-        public override void WriteByte(byte value)
-        {
-            Span<byte> oneByte = stackalloc byte[1];
-            oneByte[0] = value;
-            Write(oneByte);
-        }
+        public override void WriteByte(byte value) => Write(new ReadOnlySpan<byte>(ref value));
 
         public override unsafe void Write(ReadOnlySpan<byte> buffer)
         {
--- a/src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
+++ b/src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
@@ -726,6 +726,12 @@ namespace System.Net.Security
             return manager;
         }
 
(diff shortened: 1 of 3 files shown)


src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
```diff
@@ -748,6 +762,27 @@ public override int ReadByte()
...
+
+                try
+                {
+                    ValueTask<int> vt = ReadAsyncInternal<SyncReadWriteAdapter>(memoryManager.Memory, default(CancellationToken));
+                    Debug.Assert(vt.IsCompleted, "Sync operation must have completed synchronously");
+                    return vt.GetAwaiter().GetResult();
+                }
+                finally
+                {
+                    _readPointerMemoryManager = memoryManager;
```
stephentoub, line 781: I suggest having a ReturnPointerMemoryManager that first does Reset(null, 0) before storing it back into the field.

src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
```diff
@@ -748,6 +762,27 @@ public override int ReadByte()
             return bytesRead == 1 ? oneByte[0] : -1;
         }
 
+        public override unsafe int Read(Span<byte> buffer)
+        {
+            ThrowIfExceptionalOrNotAuthenticated();
+
+            fixed (byte* ptr = &MemoryMarshal.GetReference(buffer))
+            {
+                var memoryManager = RentPointerMemoryManager(ref _readPointerMemoryManager, ptr, buffer.Length);
```
stephentoub, line 771: Nit:
var => PoolingPointerMemoryManager

Same below

src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
```diff
@@ -888,5 +944,35 @@ private static void ThrowNotAuthenticated()
         {
             throw new InvalidOperationException(SR.net_auth_noauth);
         }
+
+        internal sealed unsafe class PoolingPointerMemoryManager : MemoryManager<byte>
```
stephentoub, line 948: Can you add a comment about why this is necessary (and the other options we considered and dismissed for now)?

src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
```diff
@@ -748,6 +762,27 @@ public override int ReadByte()
             return bytesRead == 1 ? oneByte[0] : -1;
         }
 
+        public override unsafe int Read(Span<byte> buffer)
```
stephentoub, line 765: Should the fallback paths in Read/WriteByte be updated to use the span-based APIs now that they're using this approach?

src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
```diff
@@ -742,12 +762,33 @@ public override int ReadByte()
             // Otherwise, fall back to reading a byte via Read, the same way Stream.ReadByte does.
             // This allocation is unfortunate but should be relatively rare, as it'll only occur once
             // per buffer fill internally by Read.
-            byte[] oneByte = new byte[1];
-            int bytesRead = Read(oneByte, 0, 1);
+            Span<byte> oneByte = stackalloc byte[1];
```
stephentoub, line 765: This doesn't need a stackalloc. It can use a local byte and span's ctor, eg https://github.com/dotnet/runtime/blob/6b25c3ba1a349b22e70c5418b5b05e545e1e2271/src/libraries/System.IO.Compression.Brotli/src/System/IO/Compression/dec/BrotliStream.Decompress.cs#L46-L48

src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
```diff
@@ -757,6 +798,34 @@ public override int Read(byte[] buffer, int offset, int count)
             return vt.GetAwaiter().GetResult();
         }
 
+        public override void WriteByte(byte value)
+        {
+            Span<byte> oneByte = stackalloc byte[1];
```
stephentoub, line 803: Ditto

stephentoub:
Thanks.

Can you write/run some perf tests to validate this? There are enough variables involved here that it's hard to say for sure whether this is a win just by eyeballing it.

stephentoub (approved):
Assuming perf improves, the changes lgtm.

The increased size of SslStream itself isn't ideal, but I don't have a better option. We could subsequently explore the tradeoffs involved in pooling these across instances instead.

rzikm:
```
        [Benchmark(OperationsPerInvoke = ReadWriteIterations)]
        [BenchmarkCategory(Categories.NoAOT)]
        public void ReadWrite_Sync()
        {
            Span<byte> clientBuffer = _largeClientBuffer;
            Span<byte> serverBuffer = _largeServerBuffer;
            for (int i = 0; i < ReadWriteIterations; i++)
            {
                _sslClient.Write(clientBuffer);

                int totalRead = 0;
                Span<byte> buff = serverBuffer;
                while (totalRead < clientBuffer.Length)
                {
                    int bytesRead = _sslServer.Read(buff);
                    totalRead += bytesRead;
                    buff = buff.Slice(bytesRead);
                }
            }
        }
```
| Method         | Job        | Toolchain                                                                                               | Mean     | Error    | StdDev   | Median   | Min      | Max      | Ratio | Allocated | Alloc Ratio |
|--------------- |----------- |-------------------------------------------------------------------------------------------------------- |---------:|---------:|---------:|---------:|---------:|---------:|------:|----------:|------------:|
| ReadWrite_Sync | Job-JUJLNX | \CoreRun-79072-SslStream-does-not-have-Read-override-with-Span-windows-x64-net8.0\Core_Root\corerun.exe | 23.11 us | 0.195 us | 0.182 us | 23.15 us | 22.65 us | 23.38 us |  0.98 |         - |          NA |
| ReadWrite_Sync | Job-ARTEFN | \CoreRun-main-windows-x64-net8.0\Core_Root\corerun.exe                                                  | 23.49 us | 0.246 us | 0.231 us | 23.48 us | 22.94 us | 23.83 us |  1.00 |         - |          NA |

I expected a bit more of an impact, but it is there. It has been consistently 1-2% better across multiple runs and different benchmarks.

stephentoub (approved):

rzikm:
There are some related test failures, need to investigate

```
 System.Net.Security.Tests.SslStreamMemoryConformanceTests.ReadWrite_Success_Large(mode: SyncSpan, writeSize: 10485760, startWithFlush: True) [FAIL]
      Showing first 10 differences
        Position 32768: Expected: 43, Actual: 111
        Position 32769: Expected: 248, Actual: 178
        Position 32770: Expected: 78, Actual: 47
        Position 32771: Expected: 28, Actual: 72
        Position 32772: Expected: 125, Actual: 39
        Position 32773: Expected: 10, Actual: 81
        Position 32774: Expected: 16, Actual: 144
        Position 32775: Expected: 85, Actual: 170
        Position 32776: Expected: 10, Actual: 128
        Position 32777: Expected: 0, Actual: 86
      Total number of differences: 10412466 out of 10485760
```

src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs
```diff
@@ -888,5 +952,40 @@ private static void ThrowNotAuthenticated()
...
+
+            public override Span<byte> GetSpan()
+            {
+                return new Span<byte>(_pointer, _length);
+            }
+
+            public override MemoryHandle Pin(int elementIndex = 0)
+            {
+                // memory assumed to be pinned already
+                return new MemoryHandle(_pointer, default, null);
```
stephentoub, line 982: This isn't factoring in elementIndex.

rzikm: Thanks, I completely overlooked this.

Interestingly enough, I was unable to reproduce the failures locally (on the same OS).

rzikm:
Test failures are unrelated.

PR #92897, opened 2023-10-02
This vectorizes TensorPrimitives.Log2

tannergooding:

### DIFF
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
@@ -926,6 +926,7 @@ namespace System.Numerics.Tensors
         private readonly struct Log2Operator : IUnaryOperator
         {
             public bool CanVectorize => false;
+
             public float Invoke(float x) => Log2(x);
 
             public Vector<float> Invoke(Vector<float> x)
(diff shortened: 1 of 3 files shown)


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
```diff
@@ -903,24 +904,40 @@ public Vector<float> Invoke(Vector<float> x, Vector<float> y)
...
         private readonly struct AbsoluteOperator : IUnaryOperator
         {
+            public bool CanVectorize => true;
             public float Invoke(float x) => MathF.Abs(x);
             public Vector<float> Invoke(Vector<float> x) => Vector.Abs(x);
         }
 
+        private readonly struct Log2Operator : IUnaryOperator
+        {
+            public bool CanVectorize => false;
```
stephentoub, line 928: Total nit:
```suggestion
            public bool CanVectorize => false;

```

stephentoub (approved):
Nice.

stephentoub (approved):

stephentoub:
Looks like are some NaN-related test failures

tannergooding:
> Looks like there are some NaN-related test failures

Should be handled now. The issue was that `vrs4_log2f` was deferring the special cases to the scalar `log2f` (which I missed initially): https://github.com/amd/aocl-libm-ose/blob/aocl-4.1/src/optimized/vec/vrs4_log2f.c#L183-L185

The actual logic required to handle them as vectorized isn't terribly complicated, so I ported that: https://github.com/amd/aocl-libm-ose/blob/aocl-4.1/src/optimized/log2f.c#L173-L205

Basically just some additional ternaries in the case where such a value is passed in.

tannergooding:
Will do

PR #92898, opened 2023-10-02
[wasm] Perf pipeline - fix blazor_scenarios run for hybrid globalization

radical:
Fails with:
```
Traceback (most recent call last):
  File "/mnt/vss/_work/1/s/Payload/performance/scripts/ci_setup.py", line 487, in <module>
    __main(sys.argv[1:])
  File "/mnt/vss/_work/1/s/Payload/performance/scripts/ci_setup.py", line 483, in __main
    main(CiSetupArgs(**vars(args)))
  File "/mnt/vss/_work/1/s/Payload/performance/scripts/ci_setup.py", line 411, in main
    dotnet_version = dotnet.get_dotnet_version(target_framework_moniker, args.cli) if args.dotnet_versions == [] else args.dotnet_versions[0]
  File "/mnt/vss/_work/1/s/Payload/performance/scripts/dotnet.py", line 581, in get_dotnet_version
    raise RuntimeError(
RuntimeError: Unable to determine the .NET SDK used for net8.0
```

This is because the new definition didn't copy over the `--dotnet-versions
8.0.0` workaround needed for now. And this runs only once a week, so it
was discovered on Oct 2(Monday) even though it was merged on Sep
26(Friday).- [wasm] Perf pipeline - fix blazor_scenarios run for hybrid globalization

- [wasm] Perf: run the hybrid-globalization job on runtime-wasm-perf also, for validation

This was added in https://github.com/dotnet/runtime/pull/89825 .

### DIFF
--- a/eng/pipelines/coreclr/perf-wasm-jobs.yml
+++ b/eng/pipelines/coreclr/perf-wasm-jobs.yml
@@ -191,6 +191,8 @@ jobs:
         projectFile: blazor_perf.proj
         runKind: blazor_scenarios
         runJobTemplate: /eng/pipelines/coreclr/templates/run-scenarios-job.yml
+        # For working with a newer sdk, and previous tfm (eg. 9.0 sdk, and net8.0 tfm)
+        additionalSetupParameters: '--dotnetversions 8.0.0' # passed to performance-setup.sh
         logicalmachine: 'perftiger'
         downloadSpecificBuild: ${{ parameters.downloadSpecificBuild }}
         perfForkToUse: ${{ parameters.perfForkToUse }}
--- a/eng/pipelines/coreclr/perf-wasm-jobs.yml
+++ b/eng/pipelines/coreclr/perf-wasm-jobs.yml
@@ -174,7 +174,7 @@ jobs:
         downloadSpecificBuild: ${{ parameters.downloadSpecificBuild }}
         perfForkToUse: ${{ parameters.perfForkToUse }}
 
-- ${{if and(ne(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'Schedule')) }}:
+- ${{if or(and(ne(variables['System.TeamProject'], 'public'), in(variables['Build.Reason'], 'Schedule')), in(variables['Build.DefinitionName'], 'runtime-wasm-perf')) }}:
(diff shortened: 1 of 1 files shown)


radical:
@LoopedBard3 the original PR seems to have missed testing the new job on wasm.

LoopedBard3 (approved):
LGTM, assuming the additional Build.DefinitionName check is to more easily allow testing.

radical:
> LGTM, assuming the additional Build.DefinitionName check is to more easily allow testing.

Yep, it is. We have the `runtime-wasm-perf` pipeline to run on PRs, and it does sanity checking essentially.

radical:
https://dev.azure.com/dnceng-public/public/_build/results?buildId=424830&view=results

LoopedBard3:
> @LoopedBard3 the original PR seems to have missed testing the new job on wasm.

My understanding is that the hybrid globalization primarily affects the size of built apps, so it was not included for the Wasm Microbenchmark runs as we do not have SOD measurements there. If we think that there are benchmarks that the HybridGlobalization will impact, it should definitely add Wasm runs with the HG enabled.

radical:
> > @LoopedBard3 the original PR seems to have missed testing the new job on wasm.
> 
> My understanding is that the hybrid globalization primarily affects the size of built apps, so it was not included for the Wasm Microbenchmark runs as we do not have SOD measurements there. If we think that there are benchmarks that the HybridGlobalization will impact, it should definitely add Wasm runs with the HG enabled.

I meant, I think we missed running the new job (added in `perf-wasm-jobs.yml`) on CI, because it fails without the workaround.

LoopedBard3:
> > > @LoopedBard3 the original PR seems to have missed testing the new job on wasm.
> > 
> > 
> > My understanding is that the hybrid globalization primarily affects the size of built apps, so it was not included for the Wasm Microbenchmark runs as we do not have SOD measurements there. If we think that there are benchmarks that the HybridGlobalization will impact, it should definitely add Wasm runs with the HG enabled.
> 
> I meant, I think we missed running the new job (added in `perf-wasm-jobs.yml`) on CI for the original PR, because it fails without the workaround.

Ah, I see. That makes sense, I think that the change from 8 to 9 came after the branch was made which also likely contributed 👍.

PR #92908, opened 2023-10-02
JIT: Make BasicBlock::bbJumpKind private

amanasifkhalid:
This is the beginning of a larger effort to disallow the use of BBJ_NONE (reserved for basic blocks that fall through) before the current method's block layout is finalized.

@AndyAyersMS PTAL, cc @dotnet/jit-contrib

### DIFF
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -5260,7 +5260,7 @@ public:
     {
         ASSERT_TP pAssertionOut;
 
-        if (predBlock->bbJumpKind == BBJ_COND && (predBlock->bbJumpDest == block))
+        if (predBlock->getBBJumpKind() == BBJ_COND && (predBlock->bbJumpDest == block))
         {
             pAssertionOut = mJumpDestOut[predBlock->bbNum];
 
@@ -5460,7 +5460,7 @@ ASSERT_TP* Compiler::optComputeAssertionGen()
 
             printf(FMT_BB " valueGen = ", block->bbNum);
             optPrintAssertionIndices(block->bbAssertionGen);
-            if (block->bbJumpKind == BBJ_COND)
+            if (block->getBBJumpKind() == BBJ_COND)
             {
                 printf(" => " FMT_BB " valueGen = ", block->bbJumpDest->bbNum);
                 optPrintAssertionIndices(jumpDestGen[block->bbNum]);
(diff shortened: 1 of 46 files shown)


src/coreclr/jit/block.h
```diff
@@ -702,8 +702,26 @@ struct BasicBlock : private LIR::Range
...
 
+public:
+    BBjumpKinds getBBJumpKind() const
+    {
+        return bbJumpKind;
+    }
+
+    void setBBJumpKind(BBjumpKinds kind DEBUG_ARG(Compiler* comp))
+    {
+#ifdef DEBUG
```
amanasifkhalid, line 716: I'm probably getting ahead of myself here, but I'm thinking we will need to track whether BBJ_NONE is allowed or not in the current Compiler's state, so passing in a pointer to it will enable us to assert this state. I avoided making this a noway_assert to minimize TP diffs, and made `comp` a debug arg to avoid unused variable warnings in Release builds (though this approach also has the negative effect of removing this assert in Checked builds).

src/coreclr/jit/assertionprop.cpp
```diff
@@ -5260,7 +5260,7 @@ class AssertionPropFlowCallback
     {
         ASSERT_TP pAssertionOut;
 
-        if (predBlock->bbJumpKind == BBJ_COND && (predBlock->bbJumpDest == block))
+        if (predBlock->getBBJumpKind() == BBJ_COND && (predBlock->bbJumpDest == block))
```
EgorBo, line 5263: Looks like most uses can be replaced with an existing `block->KindIs(...)` (it also supports 'params').

Also, should it be Get or get?

amanasifkhalid: Thank you for catching that -- I'll push a revision converting all the bbJumpKind comparisons to KindIs().

I'm not sure what the case should be. Some BasicBlock methods (like `KindIs`) use Pascal case, while others (like `makeBlockHot`) use camel case.

SingleAccretion: The convention for _all_ new code in the Jit is to use PascalCase, except in LSRA (where camelCase is used consistently).

src/coreclr/jit/block.h
SingleAccretion: I would personally recommend using a verifier pattern (i. e. adding an assert somewhere in fgdiagnostic.cpp) for checking this property instead of the setters. It will make for less debug-only code overall.

amanasifkhalid: I like the cleanliness of your suggestion, but if we move this check to somewhere like `Compiler::fgDebugCheckBBlist` instead of checking every time `bbJumpKind` is updated, then is there any need to make `bbJumpKind` private? @AndyAyersMS

SingleAccretion: > is there any need to make bbJumpKind private

Not strictly, but it is still an improvement.

https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/clr-jit-coding-conventions.md

> 15.5.1 Public data members
> Do not declare public data members. Instead, public accessor functions should be exposed to access class members.

AndyAyersMS:
SPMI replay error is infrastructure related:
```
OSError: [Errno 28] No space left on device
```
so maybe retry this bit?

amanasifkhalid:
Looks like the infrastructure issue didn't repro; CI is clean now.

AndyAyersMS (approved):
I skimmed through the diffs but did not check them all carefully.

Issue #92915, opened 2023-10-02
Investigate improving JIT heuristics with machine learning


AndyAyersMS:
### Overview 

There is now fairly substantial literature on improving compiler optimization by leveraging machine learning (ML). A comprehensive survey compiled by [Zheng Wang](https://github.com/zwang4) can be found here: https://github.com/zwang4/awesome-machine-learning-in-compilers.

Here's a closely relevant paper and github repo: [MLGO: a Machine Learning Guided Compiler
Optimizations Framework](https://arxiv.org/pdf/2101.04808.pdf). Like our efforts below it leverages a Policy Gradient algorithm (reinforcement Learning).
* https://github.com/google/ml-compiler-opt

[Several years ago](https://github.com/AndyAyersMS/PerformanceExplorer/blob/master/notes/notes-aug-2016.md) we attempted to leverage ML to create better inline heuristics. That experiment was largely a failure, at least as far as using ML to predict if an inline would improve performance. But one of the key speculations from that work was that lack of PGO was to blame. Now that we have PGO, it is a good time to revisit this area to see if it PGO was indeed the key missing ingredient.

### Proposed Work

During the .NET 9 product cycle, we plan to investigate applying ML techniques to heuristics used by the JIT. The items below represent the initial steps of this investigation. This list will change and grow as the work progresses.

We also want to tackle a relatively simple problem, at least initially. Thus our initial effort will be to try and refine and improve the heuristic the JIT uses for Common Subexpression Elimination (aka CSE):

- [ ] Study the current CSE heuristic and try and understand what it is modelling. Identify apparent shortcomings and limitations.
  - See [initial notes](https://github.com/dotnet/runtime/issues/92915#issuecomment-1743925194) below.
- [x] Develop tooling for varying which CSEs the JIT can perform, either in concert with the current heuristic or independent of it. Measure the impact variation of this on key JIT metrics (performance, code size, jit time, etc).
  - https://github.com/dotnet/runtime/pull/92918
  - https://github.com/dotnet/jitutils/pull/381
  - https://github.com/dotnet/jitutils/pull/383
- [ ] See how well PerfScores can be used as a proxy for measuring actual performance. Ideally, we can leverage PerfScores to avoid needing to run benchmarks repeatedly. But if not, perhaps we can still rely on PerfScores to limit or prioritize the runs needed.
  - See [notes](https://github.com/dotnet/runtime/issues/92915#issuecomment-1744042491) below for some initial data.
[...]

PR #92918, opened 2023-10-03
JIT: Allow experimenting with different CSE subsets

AndyAyersMS:
Introduce two config vars for experimenting with CSEs:
* `JitCSEHash`: identifies a method for CSE experimentatin
* `JitCSEMask`: specifies a bitmask of allowed CSEs (by "attempt")

When the hash is nonzero, any method whose hash matches will perform only the subset of CSEs specified by the mask (up to 32 CSEs).

Also introduce a config var to dump the total number of CSEs to either the assemby listing footer or the one-liner from the disassembly summary.
* `JitMetrics`

This can perhaps be generalized eventually to report more metrics and perhaps to report them back to SPMI when it is the jit host.

Finally, note CSE lcl vars that represent hoisted trees and or are "multiple-def" CSEs in the local var table.

Contributes to #92915.

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -2677,6 +2677,13 @@ void Compiler::compInitOptions(JitFlags* jitFlags)
         verboseDump = (JitConfig.JitDumpTier0() > 0);
     }
 
+    // Optionally suppress dumping OSR jit requests.
+    //
+    if (verboseDump && jitFlags->IsSet(JitFlags::JIT_FLAG_OSR))
+    {
+        verboseDump = (JitConfig.JitDumpOSR() > 0);
+    }
+
     // Optionally suppress dumping except for a specific OSR jit request.
     //
     const int dumpAtOSROffset = JitConfig.JitDumpAtOSROffset();
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -5229,10 +5229,12 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
         INDEBUG(sprintf_s(debugPart, 128, ", hash=0x%08x%s", info.compMethodHash(), compGetStressMessage()));
 
         char metricPart[128] = {0};
+#ifdef DEBUG
         if (JitConfig.JitMetrics() > 0)
         {
-            INDEBUG(sprintf_s(metricPart, 128, ", perfScore=%.2f, numCse=%u", info.compPerfScore, optCSEcount));
+            sprintf_s(metricPart, 128, ", perfScore=%.2f, numCse=%u", info.compPerfScore, optCSEcount);
         }
+#endif
 
         const bool hasProf = fgHaveProfileData();
         printf("%4d: JIT compiled %s [%s%s%s%s, IL size=%u, code size=%u%s%s]\n", methodsCompiled, fullName,
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -5224,11 +5224,11 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
         char metricPart[128] = {0};
         if (JitConfig.JitMetrics() > 0)
         {
-            INDEBUG(sprintf_s(debugPart, 128, ", perfScore=%.2f, numCse=%u", info.compPerfScore, optCSEcount));
+            INDEBUG(sprintf_s(metricPart, 128, ", perfScore=%.2f, numCse=%u", info.compPerfScore, optCSEcount));
         }
 
         const bool hasProf = fgHaveProfileData();
-        printf("%4d: JIT compiled %s [%s%s%s%s, IL size=%u, code size=%u%s,%s]\n", methodsCompiled, fullName,
+        printf("%4d: JIT compiled %s [%s%s%s%s, IL size=%u, code size=%u%s%s]\n", methodsCompiled, fullName,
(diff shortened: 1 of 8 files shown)


AndyAyersMS:
cc @dotnet/jit-contrib

src/coreclr/jit/jitconfigvalues.h
```diff
@@ -376,6 +376,13 @@ CONFIG_INTEGER(JitConstCSE, W("JitConstCSE"), 0)
 #define CONST_CSE_ENABLE_ALL 3
 #define CONST_CSE_ENABLE_ALL_NO_SHARING 4
 
+// Allow fine-grained controls of CSEs done in a particular method
+CONFIG_INTEGER(JitCSEHash, W("JitCSEHash"), 0)
+CONFIG_INTEGER(JitCSEMask, W("JitCSEMask"), 0)
+
+// Enable metric output in jit disasm & elsewhere
+CONFIG_INTEGER(JitMetrics, W("JitMetrics"), 0)
+
```
BruceForstall, line 385: These are available in Release. Makes sense for JitMetrics, but do you want that for the JitCSE* variables?

I suggest documenting the JitCSE* variables in more detail (_how_ do you set JitCSEMask?)

src/coreclr/jit/compiler.cpp
```diff
@@ -5221,10 +5221,16 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
...
 
+        char metricPart[128] = {0};
+        if (JitConfig.JitMetrics() > 0)
+        {
+            INDEBUG(sprintf_s(debugPart, 128, ", perfScore=%.2f, numCse=%u", info.compPerfScore, optCSEcount));
+        }
+
         const bool hasProf = fgHaveProfileData();
-        printf("%4d: JIT compiled %s [%s%s%s%s, IL size=%u, code size=%u%s]\n", methodsCompiled, fullName,
+        printf("%4d: JIT compiled %s [%s%s%s%s, IL size=%u, code size=%u%s,%s]\n", methodsCompiled, fullName,
```
BruceForstall, line 5231:
```suggestion
        printf("%4d: JIT compiled %s [%s%s%s%s, IL size=%u, code size=%u%s%s]\n", methodsCompiled, fullName,
```

src/coreclr/jit/compiler.cpp
```diff
@@ -5221,10 +5221,16 @@ void Compiler::compCompile(void** methodCodePtr, uint32_t* methodCodeSize, JitFl
         char debugPart[128] = {0};
         INDEBUG(sprintf_s(debugPart, 128, ", hash=0x%08x%s", info.compMethodHash(), compGetStressMessage()));
 
+        char metricPart[128] = {0};
+        if (JitConfig.JitMetrics() > 0)
+        {
+            INDEBUG(sprintf_s(debugPart, 128, ", perfScore=%.2f, numCse=%u", info.compPerfScore, optCSEcount));
```
BruceForstall, line 5227:
```suggestion
            INDEBUG(sprintf_s(metricPart, 128, ", perfScore=%.2f, numCse=%u", info.compPerfScore, optCSEcount));
```

src/coreclr/jit/optcse.cpp
```diff
@@ -3318,6 +3323,21 @@ class CSE_Heuristic
...
 
 #ifdef DEBUG
+
+            if (doCSE)
+            {
+                const int attempt = m_pCompiler->optCSEattempt++;
+
+                if (m_pCompiler->info.compMethodHash() == (unsigned)JitConfig.JitCSEHash())
+                {
+                    doCSE = ((1ULL << attempt) & ((unsigned long long)JitConfig.JitCSEMask())) != 0;
```
BruceForstall, line 3333: Should you check that `attempt` is in range? i.e.,
```
if (attempt >= 32) { // or `sizeof(unsigned int) * BITS_PER_BYTE`
  doCSE = false;
} else {
...
}
```

Presumably JitCSEMask is a 32-bit int.

src/coreclr/jit/compiler.cpp
AndyAyersMS: Thanks, I added this but wasn't using it for anything, so didn't notice it was messed up.

src/coreclr/jit/jitconfigvalues.h
AndyAyersMS: Sure, will do.

AndyAyersMS:
@BruceForstall think I addressed your feedback, feel free to take another look.

src/coreclr/jit/jitconfigvalues.h
jakobbotsch: It doesn't look like `JitMetrics` actually works in release (even though it is made available there).

AndyAyersMS: Ah, let me move it under debug for now too.

src/coreclr/jit/optcse.cpp
```diff
@@ -3318,6 +3323,30 @@ class CSE_Heuristic
...
+            if (doCSE)
+            {
+                const int attempt = m_pCompiler->optCSEattempt++;
+
+                if (m_pCompiler->info.compMethodHash() == (unsigned)JitConfig.JitCSEHash())
+                {
+                    // We can only mask the first 32 CSE attempts, so suppress anything beyond that.
+                    // Note methods with > 32 CSEs are currently quite rare.
+                    //
+                    if (attempt > 32)
```
BruceForstall, line 3336:
```suggestion
                    if (attempt >= 32)
```

otherwise if `attempt == 32` you do (int) `1 << 32` which is undefined?

BruceForstall (approved):

AndyAyersMS:
SPMI failures are "no space on device". No diffs expected and no diffs in the runs that succeeded.

PR #92922, opened 2023-10-03
Fix issue with reflection invoke and pointer types

MichalStrehovsky:
We should not try to `RhBox` these. Tests will be added when #10842 is fixed. The behavior we have now is not the CoreCLR behavior.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/DynamicInvokeInfo.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/DynamicInvokeInfo.cs
@@ -769,7 +769,7 @@ namespace System.Reflection
                         Debug.Assert(type.IsPointer);
                         obj = Pointer.Box((void*)Unsafe.As<byte, IntPtr>(ref obj.GetRawData()), type);
                     }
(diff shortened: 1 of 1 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/DynamicInvokeInfo.cs
```diff
@@ -769,7 +769,7 @@ private unsafe void CopyBackToArray(ref object? src, object?[] dest)
                         Debug.Assert(type.IsPointer);
                         obj = Pointer.Box((void*)Unsafe.As<byte, IntPtr>(ref obj.GetRawData()), type);
                     }
-                    if ((transform & Transform.FunctionPointer) != 0)
+                    else if ((transform & Transform.FunctionPointer) != 0)
```
jkotas, line 772: Nit: This would produce a bit smaller code:
```
else
{
   obj = RuntimeImports.RhBox(
       ((transform & Transform.FunctionPointer) != 0) ? EETypePtr.EETypePtrOf<IntPtr>() : argumentInfo.Type,
       ref obj.GetRawData());
}
```

jkotas (approved):
Thanks

PR #92923, opened 2023-10-03
`readonly`-ify fields based on whole program view

MichalStrehovsky:
Marking fields `readonly` has a couple benefits:

* Codegen can assume the value doesn't change and can inline the values into code.
* In the static constructor interpreter, we allow reading readonly fields of other classes. So `class A { public static int X = 123 } class B { static int Y = A.X }` - `A` can be preinitialized at compile time, but `B` cannot. However: `class A { public static readonly int X = 123 } class B { static int Y = A.X }` both `A` and `B` can preinitialize at compile time.

In this PR, I'm adding infrastructure to collect information about field writes during whole program analysis. This collects information about fields that are effectively read-only. Then during compilation, we use this data instead of the `IsInitOnly` metadata. The big advantage of this approach (as opposed to e.g. a Roslyn analyzer) is that we look at the app after trimming - some fields might be writable, but if we discarded the code that writes them, they're read only.

This is also a correctness fix - the compiler would previously assume preinitializing instances of classes with all-readonly instance fields is safe, but one can set them with reflection and this would lead to a crash in the write barrier. Not that anyone should be reflection-setting readonly fields outside of deserialization scenarios.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/IL/ILImporter.Scanner.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/IL/ILImporter.Scanner.cs
@@ -974,21 +974,23 @@ namespace Internal.IL
             _isReadOnly = true;
         }
 
-        private void ImportFieldAccess(int token, bool isStatic, bool write, string reason)
+        private void ImportFieldAccess(int token, bool isStatic, bool? write, string reason)
         {
             var field = (FieldDesc)_methodIL.GetObject(token);
             var canonField = (FieldDesc)_canonMethodIL.GetObject(token);
 
             _compilation.NodeFactory.MetadataManager.GetDependenciesDueToAccess(ref _dependencies, _compilation.NodeFactory, _canonMethodIL, canonField);
 
-            if (write)
+            // Write will be null for ld(s)flda. Consider address loads write unless they were
+            // for initonly fields. We'll trust the initonly that this is not a write.
+            if (write == null)
+                write = !field.IsInitOnly;
+
+            if (write.Value)
             {
                 bool isInitOnlyWrite = field.OwningType == _methodIL.OwningMethod.OwningType
                     && ((field.IsStatic && _methodIL.OwningMethod.IsStaticConstructor)
                         || (!field.IsStatic && _methodIL.OwningMethod.IsConstructor));
-
-                // Consider initonly statics always safe; Roslyn does generate ldflda for these.
-                isInitOnlyWrite |= field.IsStatic && field.IsInitOnly;
 
                 if (!isInitOnlyWrite)
                 {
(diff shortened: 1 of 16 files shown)


src/coreclr/tools/aot/ILCompiler.Compiler/IL/ILImporter.Scanner.cs
```diff
@@ -974,13 +974,36 @@ private void ImportReadOnlyPrefix()
             _isReadOnly = true;
         }
 
-        private void ImportFieldAccess(int token, bool isStatic, string reason)
+        private void ImportFieldAccess(int token, bool isStatic, bool? write, string reason)
```
SingleAccretion, line 977: Does this analysis interact correctly with unsafe code?

E. g. `&s + <offset> = <value>;`

MichalStrehovsky: It doesn't, and neither with interop.

jkotas: This should not be a problem in practice, given what we use the result of the analysis for today. Correct?

MichalStrehovsky: After 865875849d186d08a071f62aaa84c51cf30f966a I can't think of a problematic case - this should not be making any new assumptions.

src/tests/nativeaot/SmokeTests/TrimmingBehaviors/DeadCodeElimination.cs
```diff
@@ -378,6 +380,100 @@ public static void Run()
         }
     }
```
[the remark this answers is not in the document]
MichalStrehovsky: Test added.

src/coreclr/tools/aot/ILCompiler.Compiler/IL/ILImporter.Scanner.cs
SingleAccretion: Are the consumers restricted enough that even if the field is only written to via pointer arithmetic, things still work?

```cs
[StructLayout(LayoutKind.Explicit)]
class Xyz
{
    [FieldOffset(0)]
    public int FieldOne;
    [FieldOffset(4)]
    public int FieldTwo;
}

Xyz xyz = { ... };
Unsafe.AddByteOffset(ref xyz.FieldOne, 4) = { ... };
```

MichalStrehovsky: We go over all fields on the class and check for read only on each. The check is in the preexisting code that we're switching from believing isliteral to using this analysis (that's the correctness fix part from the pr description).

jkotas (approved):
Nice!

PR #92926, opened 2023-10-03
Add testing for #92539.

eiriktsarpalis:
Adding testing for the fix in #92539. Companion PR to #92924.

### DIFF
--- a/src/libraries/System.Formats.Cbor/tests/Reader/CborReaderTests.Tag.cs
+++ b/src/libraries/System.Formats.Cbor/tests/Reader/CborReaderTests.Tag.cs
@@ -2,8 +2,11 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Numerics;
+using System.Threading;
+using Microsoft.DotNet.RemoteExecutor;
 using Test.Cryptography;
 using Xunit;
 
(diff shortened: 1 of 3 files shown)


eiriktsarpalis:
Companion PR to https://github.com/dotnet/runtime/pull/92924

src/libraries/System.Formats.Cbor/tests/Reader/CborReaderTests.Tag.cs
```diff
@@ -192,6 +195,31 @@ public static void ReadDateTimeOffset_SingleValue_HappyPath(string expectedValue
             Assert.Equal(expectedValue.Offset, result.Offset);
         }
 
+        [SkipOnTargetFramework(TargetFrameworkMonikers.NetFramework)]
```
vcsjones, line 198:
```suggestion
        [SkipOnTargetFramework(TargetFrameworkMonikers.NetFramework, "Explain why this test cannot run on .NET Framework")]
```

src/libraries/System.Formats.Cbor/tests/Writer/CborWriterTests.Tag.cs
```diff
@@ -88,6 +91,30 @@ public static void WriteDateTimeOffset_SingleValue_HappyPath(string valueString,
             AssertHelper.HexEqual(expectedHexEncoding.HexToByteArray(), encoding);
         }
 
+        [SkipOnTargetFramework(TargetFrameworkMonikers.NetFramework)]
```
vcsjones, line 94:
```suggestion
        [SkipOnTargetFramework(TargetFrameworkMonikers.NetFramework, "Explain why this test cannot run on .NET Framework")]
```

vcsjones (approved):
Looks good to me with a non-blocking suggestion.

src/libraries/System.Formats.Cbor/tests/Reader/CborReaderTests.Tag.cs
eiriktsarpalis: It's because of the remote executor, which has been giving me trouble with netfx in certain CI legs. At this point I just disable it from the start. I'll remove the suppression to see what happens.

vcsjones: That's fine - just noting that the attribute has a `reason` parameter that I think we should populate if we decide to keep the attribute.

jeffhandley (approved):

Issue #92080, opened 2023-09-14
CBOR does not handle canonical NaN correctly


vcsjones:
### Description

Since System.Formats.Cbor adheres to RFC 7049, I don't believe `NaN` is handled correctly.

The [RFC states](https://www.rfc-editor.org/rfc/rfc7049#section-2.3) for canonical CBOR:

> Also, there are many representations for NaN.  If NaN is an allowed value, it must always be represented as 0xf97e00.

`CborWriter` will encode it as 0xF9FE00.

Similarly, `CborReader` will not throw for any non-canonically encoded `NaN`.

### Reproduction Steps

To reproduce the `CborWriter` issue:

```C#
using System;
using System.Formats.Cbor;

CborWriter writer = new(CborConformanceMode.Canonical);
writer.WriteSingle(float.NaN);
Console.WriteLine(Convert.ToHexString(writer.Encode())); // Outputs F9FE00
```

-----

To reproduce the `CborReader` issue:

```C#
byte[] canonical = Convert.FromHexString("F97E00");
byte[] nonCanonical = Convert.FromHexString("F9FE00");
CborReader canonicalReader = new CborReader(canonical, CborConformanceMode.Canonical);
CborReader nonCanonicalReader = new CborReader(nonCanonical, CborConformanceMode.Canonical);
Console.WriteLine(float.IsNaN(canonicalReader.ReadSingle()));
Console.WriteLine(float.IsNaN(nonCanonicalReader.ReadSingle()));
```

### Expected behavior

`CborWriter` encodes `NaN` as 0xf97e00.

----

`CborReader` should throw when decoding a `NaN` that is non-canonically encoded.

### Actual behavior

`CborWriter` incorrectly encodes `NaN` as 0xF9FE00 when in canonical encoding.

----

`CborReader` incorrectly permits all encodings of `NaN` when in canonical encoding.

### Regression?

No.

### Configuration

Host:
  Version:      8.0.0-rc.1.23419.4
  Architecture: arm64
  Commit:       92959931a3
  RID:          osx-arm64

Using 7.0.0 of System.Formats.Cbor package.

vcsjones:
This actually might be an issue in the definition of `Half.NaN`.

I would expect this to print the same thing:

```C#
Half h = Half.Zero / Half.Zero;
Console.WriteLine(BitConverter.HalfToInt16Bits(h).ToString("X"));
[...]

PR #92934, opened 2023-10-03
CBOR Writer: Use canonical NaN representation for NaN values

tomeksowi:
[RFC 7049 (CBOR)](https://www.rfc-editor.org/rfc/rfc7049) specifies "If NaN is an allowed value, it must always be represented as 0xf97e00". The only exception I made is when the user explicitly requests precision (FP size) is preserved.

The problem occurred for x86, C# defines NaN as 0.0/0.0 which yields -NaN on x86 FP units, which gets encoded as 0xf9fe00.

Fixes issue #92080

Also, use positive NaNs on RISC-V for FP ops crossgen-evaluated when host is x86.

Part of #84834

### DIFF
--- a/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
+++ b/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
@@ -7,6 +7,10 @@ namespace System.Formats.Cbor
 {
     public partial class CborWriter
     {
+        // Canonical NaN representations as per RFC 7049
+        private static readonly float CanonicalNaNSingle = CborHelpers.Int32BitsToSingle(0x7fc00000);
+        private static readonly double CanonicalNaNDouble = BitConverter.Int64BitsToDouble(0x7ff8000000000000);
+
         // Implements major type 7 encoding per https://tools.ietf.org/html/rfc7049#section-2.1
 
         /// <summary>Writes a single-precision floating point number (major type 7).</summary>
@@ -59,7 +63,7 @@ namespace System.Formats.Cbor
         {
             if (float.IsNaN(value))
             {
-                value = CborHelpers.Int32BitsToSingle(0x7fc00000); // canonical NaN as per RFC 7049
+                value = CanonicalNaNSingle;
             }
             EnsureWriteCapacity(1 + sizeof(float));
             WriteInitialByte(new CborInitialByte(CborMajorType.Simple, CborAdditionalInfo.Additional32BitData));
@@ -72,7 +76,7 @@ namespace System.Formats.Cbor
         {
             if (double.IsNaN(value))
             {
-                value = BitConverter.Int64BitsToDouble(0x7ff8000000000000); // canonical NaN as per RFC 7049
+                value = CanonicalNaNDouble;
             }
             EnsureWriteCapacity(1 + sizeof(double));
             WriteInitialByte(new CborInitialByte(CborMajorType.Simple, CborAdditionalInfo.Additional64BitData));
--- a/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
+++ b/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
@@ -18,6 +18,11 @@ namespace System.Formats.Cbor
         /// <para>The written data is not accepted under the current conformance mode.</para></exception>
         public void WriteSingle(float value)
         {
+            if (float.IsNaN(value))
+            {
+                value = BitConverter.Int32BitsToSingle(0x7fc00000); // canonical NaN as per RFC 7049
+            }
+
             if (!CborConformanceModeHelpers.RequiresPreservingFloatPrecision(ConformanceMode) &&
                  TryConvertSingleToHalf(value, out var half))
             {
@@ -38,6 +43,11 @@ namespace System.Formats.Cbor
         /// <para>The written data is not accepted under the current conformance mode.</para></exception>
         public void WriteDouble(double value)
         {
+            if (double.IsNaN(value))
+            {
+                value = BitConverter.Int64BitsToDouble(0x7ff8000000000000); // canonical NaN as per RFC 7049
+            }
+
             if (!CborConformanceModeHelpers.RequiresPreservingFloatPrecision(ConformanceMode) &&
                  TryConvertDoubleToSingle(value, out float single))
             {
--- a/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
+++ b/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
@@ -7,9 +7,9 @@ namespace System.Formats.Cbor
 {
     public partial class CborWriter
     {
-        // Canonical NaN representations as per RFC 7049
-        private static readonly float CanonicalNaNSingle = CborHelpers.Int32BitsToSingle(0x7fc00000);
-        private static readonly double CanonicalNaNDouble = BitConverter.Int64BitsToDouble(0x7ff8000000000000);
+        // CBOR RFC 8949 says: if NaN is an allowed value, and there is no intent to support NaN payloads or signaling NaNs, the protocol needs to pick a single representation, typically 0xf97e00. If that simple choice is not possible, specific attention will be needed for NaN handling.
+        // In this implementation "that simple choice is not possible" for CTAP2 mode (RequiresPreservingFloatPrecision), in which "representations of any floating-point values are not changed".
+        private const ushort PositiveQNaNBitsHalf = 0x7e00;
 
         // Implements major type 7 encoding per https://tools.ietf.org/html/rfc7049#section-2.1
 
@@ -61,10 +61,6 @@ namespace System.Formats.Cbor
         }
         private void WriteSingleCore(float value)
         {
-            if (float.IsNaN(value))
-            {
-                value = CanonicalNaNSingle;
-            }
             EnsureWriteCapacity(1 + sizeof(float));
             WriteInitialByte(new CborInitialByte(CborMajorType.Simple, CborAdditionalInfo.Additional32BitData));
             CborHelpers.WriteSingleBigEndian(_buffer.AsSpan(_offset), value);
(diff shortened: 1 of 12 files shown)


src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
```diff
@@ -18,6 +18,11 @@ public partial class CborWriter
         /// <para>The written data is not accepted under the current conformance mode.</para></exception>
         public void WriteSingle(float value)
         {
+            if (float.IsNaN(value))
```
vcsjones, line 21: I see that `WriteSingle` and `WriteDouble` are covered, but not `WriteHalf`

https://github.com/dotnet/runtime/blob/88f25474334bfb0da10e0b5acb0219a64a04828b/src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.netcoreapp.cs#L20-L27

I think the appropriate change is to leave WriteSingle and WriteDouble alone, and fix WriteHalf to say "If this is NaN, then write it with positive NaN"

tomeksowi: > I think the appropriate change is to leave WriteSingle and WriteDouble alone, and fix WriteHalf to say "If this is NaN, then write it with positive NaN"

It wouldn't cover writing a float/double NaN and requiring preserving precision (they should be positive too as per RFC 7049).

You're right about WriteHalf, though.

tomeksowi: I pushed the positive NaN reassignment to Write(Single|Double)Core because most of the cases are handled by WriteHalf, that way code for Write(Double|Single) is left alone. Besides, it's more of an representation thing so it's probably better to handle it just before writing.

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
```diff
@@ -57,6 +57,10 @@ public void WriteDouble(double value)
         }
         private void WriteSingleCore(float value)
         {
+            if (float.IsNaN(value))
+            {
+                value = CborHelpers.Int32BitsToSingle(0x7fc00000); // canonical NaN as per RFC 7049
```
krwq, line 62: could these use constants? CanonicalFloat defined here or just float.NaN or something? `CborHelpers.Int32BitsToSingle(0x7fc00000);` and `BitConverter.Int64BitsToDouble(0x7ff8000000000000)` seem constant or am I missing something?

src/coreclr/jit/valuenum.cpp
```diff
@@ -106,7 +106,7 @@ struct DoubleTraits
 template <typename TFp, typename TFpTraits>
 TFp FpAdd(TFp value1, TFp value2)
 {
-#if defined(TARGET_ARMARCH) || defined(TARGET_LOONGARCH64)
```
eiriktsarpalis, line 109: cc @dotnet/jit-contrib

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
tomeksowi: Sure, I put them in named constants.

`float.NaN` is defined as `0.0f / 0.0f` which doesn't yield a consistent bit pattern across platforms (it's negative on x86, positive elsewhere).

jakobbotsch (approved):
The JIT change LGTM.

src/libraries/System.Formats.Cbor/tests/Writer/CborWriterTests.Simple.netcoreapp.cs
```diff
@@ -18,7 +18,7 @@ public partial class CborWriterTests
         [InlineData(0.00006103515625, "f90400")]
         [InlineData(-4.0, "f9c400")]
         [InlineData(float.PositiveInfinity, "f97c00")]
-        [InlineData(float.NaN, "f9fe00")]
```
eiriktsarpalis, line 21: Could you also add a few reader tests validating that the previously used encoding is still readable as NaN?

eiriktsarpalis (approved):

vcsjones (approved):

src/coreclr/jit/valuenum.cpp
tannergooding: Why is any change to valuenum.cpp needed here for the cbor fix?

This seems like an independent change related to constant folding which should be in it's own PR.

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
```diff
@@ -7,6 +7,10 @@ namespace System.Formats.Cbor
 {
     public partial class CborWriter
     {
+        // Canonical NaN representations as per RFC 7049
+        private static readonly float CanonicalNaNSingle = CborHelpers.Int32BitsToSingle(0x7fc00000);
+        private static readonly double CanonicalNaNDouble = BitConverter.Int64BitsToDouble(0x7ff8000000000000);
```
tannergooding, line 12: nit: I think it would be better to simply calls these what they actually are, which is `PositiveQNaN`.

The concern being that they are canonical in the context of CBOR, but not more generally, and someone may get the wrong idea especially if just generally searching for things related to NaN in the repo.

Calling it what it is, a `+QNaN` and then leaving a comment that it's used because `RFC 7049` expects it as it's own canonical representation avoids the ambiguity, while potentially providing additional clarity and allowing us to clearly indicate why we don't use `float.NaN`

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
```diff
@@ -57,6 +61,10 @@ public void WriteDouble(double value)
         }
         private void WriteSingleCore(float value)
         {
+            if (float.IsNaN(value))
+            {
+                value = CanonicalNaNSingle;
+            }
```
tannergooding, line 67: This isn't guaranteed to work across all possible runtimes/compilers/future versions. The IEEE 754 spec allows for a `NaN` to be normalized to any representation (and while we aren't likely to do that ourselves, it is very much legal to do so).

Rather, instead, you should have something like:
```csharp
if (float.IsNaN(value))
{
    WriteInt32Core(PositiveQNaNBits);
    return;
}

...
```

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
```diff
@@ -130,7 +142,7 @@ public void WriteSimpleValue(CborSimpleValue value)
         private static bool TryConvertDoubleToSingle(double value, out float result)
         {
             result = (float)value;
-            return BitConverter.DoubleToInt64Bits(result) == BitConverter.DoubleToInt64Bits(value);
+            return double.IsNaN(value) || BitConverter.DoubleToInt64Bits(result) == BitConverter.DoubleToInt64Bits(value);
```
tannergooding, line 145: Is this simply trying to see if a given `double` can be represented as a `float` without loss in precision?

Is there a non-canonicalization mode where the `NaN` payload bits are necessary to consider?

tomeksowi: > Is this simply trying to see if a given `double` can be represented as a `float` without loss in precision?

That's what the hitherto bit pattern comparison was trying to do. That failed for FPUs with lax NaN propagation rules (e.g. RISC-V) because the conversion to float drops the sign and payload.

> Is there a non-canonicalization mode where the `NaN` payload bits are necessary to consider?

[RFC 8949](https://www.rfc-editor.org/rfc/rfc8949) offers an easy way out by saying "we don't support payloads":
*If NaN is an allowed value, and there is no intent to support NaN payloads or signaling NaNs, the protocol needs to pick a single representation, typically 0xf97e00. If that simple choice is not possible, specific attention will be needed for NaN handling.*

However, CborConformanceMode.Ctap2CanonicalMode "that simple choice is not possible" as it says: *representations of any floating-point values are not changed*. For that mode we don't enter TryConvertDoubleToSingle and friends so no problem there. But I guess in Write(Single|Double)Core and WriteHalf the condition should be:
```cs
if (float.IsNaN(value) && !CborConformanceModeHelpers.RequiresPreservingFloatPrecision(ConformanceMode))
{
    value = CanonicalNaNSingle;
}
```
Agreed?

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
tomeksowi: WriteInt32Core doesn't exist but I agree, it shouldn't go through an FPU.

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.cs
tomeksowi: OK

src/coreclr/jit/valuenum.cpp
tomeksowi: OK, #93098

tannergooding (approved):

src/libraries/System.Formats.Cbor/src/System/Formats/Cbor/Writer/CborWriter.Simple.netcoreapp.cs
```diff
@@ -21,7 +21,14 @@ public void WriteHalf(Half value)
         {
             EnsureWriteCapacity(1 + sizeof(short));
             WriteInitialByte(new CborInitialByte(CborMajorType.Simple, CborAdditionalInfo.Additional16BitData));
-            BinaryPrimitives.WriteHalfBigEndian(_buffer.AsSpan(_offset), value);
+            if (Half.IsNaN(value) && !CborConformanceModeHelpers.RequiresPreservingFloatPrecision(ConformanceMode))
```
eiriktsarpalis, line 24: I'm not sure I understand why the second check is necessary. It is meant to check if the current conformance mode permits using lower-precision representation, however `Half` is the lowest one available per the spec.

tomeksowi: Half NaN may also have a sign and payload and CTAP2 mode description says *representations of any floating-point values are not changed*. So I opted for writing the input verbatim.

Issue #92877, opened 2023-10-02
JsonObjectCreationHandling.Populate not working with parameterized constructors


eiriktsarpalis:
Thanks, this looks like a real bug. FWIW it's not specific to primary constructors or records, this fails too:

```C#
using System.Text.Json.Serialization;
using System.Text.Json;

var singleUserJson = """{"Username":"Filip","PhoneNumbers":["123456"]}""";
var userFromJson = JsonSerializer.Deserialize<User>(singleUserJson);
Console.WriteLine(userFromJson.PhoneNumbers.Count); // 0

public class User
{
    public User(string name)
        => Name = name;

    public string Name { get; }

    [JsonObjectCreationHandling(JsonObjectCreationHandling.Populate)]
    public List<string> PhoneNumbers { get; } = new();
}
```

_Originally posted by @eiriktsarpalis in https://github.com/dotnet/docs/issues/37329#issuecomment-1742795375_

fekberg:
FYI behaves the same even if you mark the parameter as optional (default) value.

krwq:
The bug is here:
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Converters/Object/ObjectWithParameterizedConstructorConverter.cs#L338

It should be `CanDeserializeOrPopulate` - possibly more places need similar change.

eiriktsarpalis:
On closer inspection, the problem appears to be less trivial than I originally thought. The converter used for serializing types with parameterized constructors will deserialize both constructor parameters and properties _before_ it instantiates the deserialized object. 

This is not accidental, instantiating the object _depends_  on all constructor parameters being deserialized, but at the same time deserializing properties with populate semantics _depends_ on the object being instantiated ahead of time, so the two requirements are fundamentally in conflict. 

I can't think of any good fixes other than buffering the object's JSON payload and replaying deserialization for properties _after_ instantiation has completed, however that would require a substantial rearchitecting of the parameterized constructor implementation which is probably too risky for a .NET 8 servicing update. My recommendation is to simply mark populate semantics as unsupported in .NET 8 and work on a proper fix in future releases.

krwq:
Current problem can be worked around like this:

```csharp
using System.Text.Json.Serialization;
using System.Text.Json;
using System.Text.Json.Serialization.Metadata;

JsonSerializerOptions options = new()
{
    TypeInfoResolver = new DefaultJsonTypeInfoResolver()
[...]

PR #92937, opened 2023-10-03
Throw when applying `JsonObjectHandling.Populate` to types with parameterized constructors.

eiriktsarpalis:
This PR completely disables support for `JsonObjectHandling.Populate` on types using parameterized constructors, which is currently not working as expected. Should be backported to .NET 8. We will follow up with a proper fix in Future releases.

Contributes to https://github.com/dotnet/runtime/issues/92877.

### DIFF
--- a/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Metadata/DefaultJsonTypeInfoResolver.Helpers.cs
+++ b/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Metadata/DefaultJsonTypeInfoResolver.Helpers.cs
@@ -38,12 +38,20 @@ namespace System.Text.Json.Serialization.Metadata
         private static JsonTypeInfo CreateTypeInfoCore(Type type, JsonConverter converter, JsonSerializerOptions options)
         {
             JsonTypeInfo typeInfo = JsonTypeInfo.CreateJsonTypeInfo(type, converter, options);
-            typeInfo.NumberHandling = GetNumberHandlingForType(typeInfo.Type);
-            typeInfo.PreferredPropertyObjectCreationHandling = GetObjectCreationHandlingForType(typeInfo.Type);
 
-            if (typeInfo.Kind == JsonTypeInfoKind.Object)
+            if (GetNumberHandlingForType(typeInfo.Type) is { } numberHandling)
             {
-                typeInfo.UnmappedMemberHandling = GetUnmappedMemberHandling(typeInfo.Type);
+                typeInfo.NumberHandling = numberHandling;
+            }
+
+            if (GetObjectCreationHandlingForType(typeInfo.Type) is { } creationHandling)
+            {
+                typeInfo.PreferredPropertyObjectCreationHandling = creationHandling;
+            }
+
+            if (GetUnmappedMemberHandling(typeInfo.Type) is { } unmappedMemberHandling)
+            {
+                typeInfo.UnmappedMemberHandling = unmappedMemberHandling;
             }
 
             typeInfo.PopulatePolymorphismMetadata();
--- a/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Metadata/JsonTypeInfo.cs
+++ b/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Metadata/JsonTypeInfo.cs
@@ -552,17 +552,14 @@ namespace System.Text.Json.Serialization.Metadata
             {
                 VerifyMutable();
 
-                if (value is not null)
+                if (Kind != JsonTypeInfoKind.Object)
                 {
-                    if (Kind != JsonTypeInfoKind.Object)
-                    {
-                        ThrowHelper.ThrowInvalidOperationException_JsonTypeInfoOperationNotPossibleForKind(Kind);
-                    }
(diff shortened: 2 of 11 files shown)


src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Metadata/DefaultJsonTypeInfoResolver.Helpers.cs
```diff
@@ -38,12 +38,20 @@ internal static MemberAccessor MemberAccessor
         private static JsonTypeInfo CreateTypeInfoCore(Type type, JsonConverter converter, JsonSerializerOptions options)
         {
             JsonTypeInfo typeInfo = JsonTypeInfo.CreateJsonTypeInfo(type, converter, options);
-            typeInfo.NumberHandling = GetNumberHandlingForType(typeInfo.Type);
-            typeInfo.PreferredPropertyObjectCreationHandling = GetObjectCreationHandlingForType(typeInfo.Type);
 
-            if (typeInfo.Kind == JsonTypeInfoKind.Object)
+            if (GetNumberHandlingForType(typeInfo.Type) is { } numberHandling)
```
eiriktsarpalis, line 42: Not directly related to the change at hand, but this addresses an inconsistency in how attribute configuration is mapped to contract metadata.

src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Metadata/JsonTypeInfo.cs
```diff
@@ -552,17 +552,14 @@ internal void ValidateCanBeUsedForPropertyMetadataSerialization()
             {
                 VerifyMutable();
 
-                if (value is not null)
+                if (Kind != JsonTypeInfoKind.Object)
```
eiriktsarpalis, line 555: Similarly, this is a change to make the property setter more consistent with the other enum policy setters.

src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Metadata/DefaultJsonTypeInfoResolver.Helpers.cs
```diff
@@ -38,12 +38,20 @@ internal static MemberAccessor MemberAccessor
...
+            }
+
+            if (GetObjectCreationHandlingForType(typeInfo.Type) is { } creationHandling)
+            {
+                typeInfo.PreferredPropertyObjectCreationHandling = creationHandling;
+            }
+
+            if (GetUnmappedMemberHandling(typeInfo.Type) is { } unmappedMemberHandling)
+            {
+                typeInfo.UnmappedMemberHandling = unmappedMemberHandling;
```
krwq, line 54: doesn't this change the behavior of unmapped member handling for collections and dictionaries

src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Metadata/DefaultJsonTypeInfoResolver.Helpers.cs
krwq: IMO this was more readable before, why do we avoid calling the setter?

eiriktsarpalis: We don't want to call the setter if not attribute is specified. This is consistent with what we do in the source generator.

src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Metadata/DefaultJsonTypeInfoResolver.Helpers.cs
eiriktsarpalis: How so? It only avoids setting the property if no attribute is specified.

krwq (changes requested):
I'd prefer we took workaround approach and created proper fix for this: https://github.com/dotnet/runtime/issues/92877#issuecomment-1744999977

eiriktsarpalis:
@krwq see my comment in https://github.com/dotnet/runtime/issues/92877#issuecomment-1745117280. The problem lies with the async implementation of the converter which currently cannot be replayed like the sync variant is doing.

src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Metadata/DefaultJsonTypeInfoResolver.Helpers.cs
krwq: UnmappedMemberHandling was set for `typeInfo.Kind == JsonTypeInfoKind.Object` only now it's always

krwq (approved):

src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Metadata/DefaultJsonTypeInfoResolver.Helpers.cs
eiriktsarpalis: Correct, the idea is to force an error if users try to apply the attribute on invalid kinds.

Issue #76503, opened 2022-10-02
SIMD store coalescing and atomicity guarantees


SingleAccretion:
Reading the new [memory model document](https://github.com/dotnet/runtime/pull/75790), we can see:

> Memory accesses to *properly aligned* data of primitive types are always atomic. The value that is observed is always a result of complete read and write operations.

However, take this sample:
```cs
[MethodImpl(MethodImplOptions.NoInlining)]
private static void Problem(float[] p, Vector4 a)
{
    p[0] = a.X;
    p[1] = a.Y;
    p[2] = a.Z;
    p[3] = a.W;
}
```
It can emit a store 128 bits in size:
```asm
IN0003: 00000D vmovupd  xmm0, xmmword ptr [rdx]
IN0004: 000011 vmovupd  xmmword ptr [rcx+10H], xmm0
```
x64/x86 does not guarantee these stores will be atomic, so it is questionable whether this transformation is correct. Of course, it is a bit hard to imagine this being an actual problem in practice.

Bonus bug: the transform in question doesn't look at volatility, so it will coalesce volatile stores, dropping any barriers (compiler or harware) in the process, though this is a bit harder to reproduce in C#.

EgorBo:
JIT doesn't do store coalescing today except, I guess, this SIMD case but I stink there are lots of hidden perf fruits there so it will be sad to give up on that due to this hard requirement.

tannergooding:
For reference, from `Intel® 64 and IA-32 Architectures Software Developer’s Manual - 8.1.1 Guaranteed Atomic Operation`:
> The Intel486 processor (and newer processors since) guarantees that the following basic memory operations will 
always be carried out atomically:
• Reading or writing a byte.
• Reading or writing a word aligned on a 16-bit boundary.
• Reading or writing a doubleword aligned on a 32-bit boundary.
>
> The Pentium processor (and newer processors since) guarantees that the following additional memory operations 
will always be carried out atomically:
• Reading or writing a quadword aligned on a 64-bit boundary.
• 16-bit accesses to uncached memory locations that fit within a 32-bit data bus.
>
> The P6 family processors (and newer processors since) guarantee that the following additional memory operation 
will always be carried out atomically:
• Unaligned 16-, 32-, and 64-bit accesses to cached memory that fit within a cache line.
>
[...]

Issue #51638, opened 2021-04-21
Question: Is coreclr's use of SIMD-accelerated initobj legal for ref types T?


GrabYourPitchforks:
This is extracted out of a discussion whose salient points are https://github.com/dotnet/runtime/pull/51534#issuecomment-823856835 and https://github.com/dotnet/runtime/pull/51534#issuecomment-824182239.

In a nutshell, when the JIT sees an _initobj_ instruction, it will attempt to perform SIMD writes to the destination address, This is true whether or not the target type contains gcrefs. See the below example for a demonstration.

```cs
using System;

public class C {
    internal struct MyStruct
    {
        object _data0;
        object _data1;
    }
    
    internal class MyClass1
    {
        MyStruct _s;
        
        void Clear()
        {
            _s = default;
        }
    }
    
    internal class MyClass2
    {
        object _dataX;
        MyStruct _s;
        
        void Clear()
        {
            _s = default;
        }
    }
}
```

```asm
; Core CLR v5.0.421.11614 on amd64

MyClass1.Clear()
    L0000: vzeroupper
    L0003: vxorps xmm0, xmm0, xmm0
    L0007: vmovdqu [rcx+8], xmm0
    L000c: ret

MyClass2.Clear()
    L0000: vzeroupper
    L0003: vxorps xmm0, xmm0, xmm0
    L0007: vmovdqu [rcx+0x10], xmm0
    L000c: ret
```

In this example, references to `MyStruct` are only guaranteed 8-byte aligned, _not_ 16-byte aligned. So it's possible that the _vmovdqu_ instruction could straddle a cache line boundary or a page boundary. The implicit assumption here is that even if the SIMD write as a whole is not atomic, each individual native word-sized element of the vector is published atomically (since the target address is native word-aligned).

Per https://github.com/dotnet/runtime/pull/51534#discussion_r616875863, this assumption is valid for ARM64. But per https://github.com/dotnet/runtime/pull/51534#issuecomment-823856835, this assumption might not be valid for x86 / x64 / ARM32.

So, the larger question: is the current JIT behavior legal such that no other thread will ever see a torn write to a gcref field, or is the JIT relying on undocumented behavior? And if we're relying on a potentially invalid assumption, do we need to change our behavior?

tannergooding:
Notably Intel documents the following in "Volume 3A. 8.1.1 Guaranteed Atomic Operations"

[...]

PR #92939, opened 2023-10-03
Merged stores: Fix alignment-related issues and enable SIMD where possible

EgorBo:
Adjust rules when we can use unaligned stores for merged ones. Also, enable 2xLONG/REF -> SIMD. And 2xSIMD to wider SIMD.

**Wider scalar primitives for naturally aligned data of primitives  (>1B):**

Target memory | Crosses cache-line</br>boundary? | x64* | arm64
-- | -- | -- | --
Global memory | Yes | 🚫 | 🚫
  | No | ✅ | 🚫
Local memory (not exposed) | Yes | ✅ | ✅
  | No | ✅ | ✅

**SIMD for for naturally aligned data of primitives (>1B):**

| Target memory              | Known alignment  | x64*   | arm64 |
|----------------------------|------------------|--------|-------|
| Global memory              | 1B (aka unknown) | 🚫      | 🚫     |
|                            | 8B (most common)              | 🚫      | ✅     |
|                            | 16B (rare**)             | ✅(AVX+) | ✅     |
| Local memory (not exposed) | 1B               | ✅      | ✅     |
|                            | 8B               | ✅      | ✅     |
|                            | 16B              | ✅ | ✅     |

_* both Intel and AMD_
_** it's very unlikely JIT can assume 16-byte alignment currently anyhow_

PS: Merged stores are conservatively disabled on LA64 and RISC-V

Per "Arm Architecture Reference Manual":
```
* Writes from SIMD and floating-point registers of a 128-bit value that is 64-bit aligned in memory
  are treated as a pair of single - copy atomic 64 - bit writes.
```

@tannergooding said that x64 with AVX promises atomicy for 16B for 16B aligned data - so far it seems to be the only thing x64 can guarantee to us.

Related issues: https://github.com/dotnet/runtime/issues/76503, https://github.com/dotnet/runtime/issues/51638,

### DIFF
--- a/src/coreclr/jit/gentree.h
+++ b/src/coreclr/jit/gentree.h
@@ -487,6 +487,8 @@ enum GenTreeFlags : unsigned int
                                               //           This flag is useful in cases where it is required to generate register
                                               //           indirect addressing mode. One such case is virtual stub calls on xarch.
     GTF_IND_UNALIGNED           = 0x02000000, // OperIsIndir() -- the load or store is unaligned (we assume worst case alignment of 1 byte)
+    GTF_IND_KNOWN_UNALIGNED     = 0x00100000, // GT_IND -- same as GTF_IND_UNALIGNED (it's expected to be set) with a legal allowance
+                                              //           to use wider loads/stores where needed.
     GTF_IND_INVARIANT           = 0x01000000, // GT_IND -- the target is invariant (a prejit indirection)
     GTF_IND_NONNULL             = 0x00400000, // GT_IND -- the indirection never returns null (zero)
     GTF_IND_INITCLASS           = 0x00200000, // OperIsIndir() -- the indirection requires preceding static cctor
--- a/src/coreclr/jit/gentree.h
+++ b/src/coreclr/jit/gentree.h
@@ -487,8 +487,7 @@ enum GenTreeFlags : unsigned int
                                               //           This flag is useful in cases where it is required to generate register
                                               //           indirect addressing mode. One such case is virtual stub calls on xarch.
     GTF_IND_UNALIGNED           = 0x02000000, // OperIsIndir() -- the load or store is unaligned (we assume worst case alignment of 1 byte)
-    GTF_IND_KNOWN_UNALIGNED     = 0x00100000, // GT_IND -- same as GTF_IND_UNALIGNED (it's expected to be set) with a legal allowance
-                                              //           to use wider loads/stores where needed.
+    GTF_IND_ALLOW_NON_ATOMIC    = 0x00100000, // GT_IND -- this memory access does not need be atomic
     GTF_IND_INVARIANT           = 0x01000000, // GT_IND -- the target is invariant (a prejit indirection)
     GTF_IND_NONNULL             = 0x00400000, // GT_IND -- the indirection never returns null (zero)
     GTF_IND_INITCLASS           = 0x00200000, // OperIsIndir() -- the indirection requires preceding static cctor
--- a/src/coreclr/jit/gentree.h
+++ b/src/coreclr/jit/gentree.h
@@ -487,10 +487,10 @@ enum GenTreeFlags : unsigned int
                                               //           This flag is useful in cases where it is required to generate register
                                               //           indirect addressing mode. One such case is virtual stub calls on xarch.
     GTF_IND_UNALIGNED           = 0x02000000, // OperIsIndir() -- the load or store is unaligned (we assume worst case alignment of 1 byte)
-    GTF_IND_ALLOW_NON_ATOMIC    = 0x00100000, // GT_IND -- this memory access does not need be atomic
     GTF_IND_INVARIANT           = 0x01000000, // GT_IND -- the target is invariant (a prejit indirection)
     GTF_IND_NONNULL             = 0x00400000, // GT_IND -- the indirection never returns null (zero)
     GTF_IND_INITCLASS           = 0x00200000, // OperIsIndir() -- the indirection requires preceding static cctor
+    GTF_IND_ALLOW_NON_ATOMIC    = 0x00100000, // GT_IND -- this memory access does not need to be atomic
 
     GTF_IND_COPYABLE_FLAGS = GTF_IND_VOLATILE | GTF_IND_NONFAULTING | GTF_IND_UNALIGNED | GTF_IND_INITCLASS,
     GTF_IND_FLAGS = GTF_IND_COPYABLE_FLAGS | GTF_IND_NONNULL | GTF_IND_TGT_NOT_HEAP | GTF_IND_TGT_HEAP | GTF_IND_INVARIANT,
--- a/src/coreclr/jit/lower.cpp
+++ b/src/coreclr/jit/lower.cpp
@@ -7928,6 +7928,17 @@ void Lowering::LowerStoreIndirCoalescing(GenTreeStoreInd* ind)
         return;
     }
 
+    // TODO-ARM64-CQ: revisit TYP_REF if we find a case where it's beneficial.
+    // The algorithm does support TYP_REF (with null value), but it seems to be not worth
+    // it on ARM64 where it's pretty efficient to do "stp xzr, xzr, [addr]" to clear two
+    // items at once. Although, it may be profitable to do "stp q0, q0, [addr]"
+    //
+    // Other types are added for better throughput to early out.
+    if (ind->TypeIs(TYP_REF, TYP_BYREF, TYP_STRUCT, TYP_FLOAT, TYP_DOUBLE))
+    {
+        return;
+    }
+
     // We're going to do it in a loop while we see suitable STOREINDs to coalesce.
     // E.g.: we have the following LIR sequence:
     //
--- a/src/coreclr/jit/lower.cpp
+++ b/src/coreclr/jit/lower.cpp
@@ -8036,17 +8036,35 @@ void Lowering::LowerStoreIndirCoalescing(GenTreeStoreInd* ind)
                 return;
             }
 
+            int startOffset = min(prevData.offset, currData.offset) % genTypeSize(ind);
+
 #ifdef TARGET_XARCH
             // Check whether data can be guaranteed to be in a cache line.
-            int startOffset = min(prevData.offset, currData.offset) % genTypeSize(ind);
-            if ((startOffset + (genTypeSize(ind) * 2)) > TARGET_POINTER_SIZE)
+            if ((startOffset + genTypeSize(ind) * 2) > TARGET_POINTER_SIZE)
+            {
+                // NOTE: We can use SIMD if we can assume 16-bytes alignment (on Intel/AMD cpus with AVX+),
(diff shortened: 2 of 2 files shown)


tannergooding:
> @tannergooding said that x64 with AVX promises atomicy for 16B for 16B aligned data - so far it seems to be the only thing x64 can guarantee to us.

Note this is from `9.1.1 Guaranteed Atomic Operations` in `Intel® 64 and IA-32 Architectures Software Developer’s Manual; Volume 3 (3A, 3B, 3C, & 3D): System Programming Guide`

tannergooding:
> The Intel486 processor (and newer processors since) guarantees that the following basic memory operations will
always be carried out atomically:
• Reading or writing a byte.
• Reading or writing a word aligned on a 16-bit boundary.
• Reading or writing a doubleword aligned on a 32-bit boundary.
>
>The Pentium processor (and newer processors since) guarantees that the following additional memory operations
will always be carried out atomically:
• Reading or writing a quadword aligned on a 64-bit boundary.
• 16-bit accesses to uncached memory locations that fit within a 32-bit data bus.
>
>The P6 family processors (and newer processors since) guarantee that the following additional memory operation
will always be carried out atomically:
• Unaligned 16-, 32-, and 64-bit accesses to cached memory that fit within a cache line.
>
>Processors that enumerate support for Intel® AVX (by setting the feature flag CPUID.01H:ECX.AVX[bit 28]) guarantee
that the 16-byte memory operations performed by the following instructions will always be carried out atomically:
• MOVAPD, MOVAPS, and MOVDQA.
• VMOVAPD, VMOVAPS, and VMOVDQA when encoded with VEX.128.
• VMOVAPD, VMOVAPS, VMOVDQA32, and VMOVDQA64 when encoded with EVEX.128 and k0 (masking
disabled).
>
>(Note that these instructions require the linear addresses of their memory operands to be 16-byte aligned.)
Accesses to cacheable memory that are split across cache lines and page boundaries are not guaranteed to be
atomic by the Intel Core 2 Duo, Intel Atom, Intel Core Duo, Pentium M, Pentium 4, Intel Xeon, P6 family, Pentium,
and Intel486 processors. The Intel Core 2 Duo, Intel Atom, Intel Core Duo, Pentium M, Pentium 4, Intel Xeon, and
P6 family processors provide bus control signals that permit external memory subsystems to make split accesses
atomic; however, nonaligned data accesses will seriously impact the performance of the processor and should be
avoided.
>
> Except as noted above, an x87 instruction or an SSE instruction that accesses data larger than a quadword may be
implemented using multiple memory accesses. If such an instruction stores to memory, some of the accesses may
complete (writing to memory) while another causes the operation to fault for architectural reasons (e.g., due an
page-table entry that is marked “not present”). In this case, the effects of the completed accesses may be visible
to software even though the overall instruction caused a fault. If TLB invalidation has been delayed (see Section
4.10.4.4), such page faults may occur even if all accesses are to the same page.

src/coreclr/jit/lower.cpp
```diff
@@ -8003,6 +8011,42 @@ void Lowering::LowerStoreIndirCoalescing(GenTreeStoreInd* ind)
...
+            int startOffset = min(prevData.offset, currData.offset) % genTypeSize(ind);
+            if ((startOffset + (genTypeSize(ind) * 2)) > TARGET_POINTER_SIZE)
+            {
+                return;
+            }
+#else
+            return;
+#endif
+        }
+
```
EgorBo, line 8049: @SingleAccretion @jakobbotsch addressed the concerns about atomicity for primitives we discussed offline.

src/coreclr/jit/gentree.h
```diff
@@ -487,6 +487,8 @@ enum GenTreeFlags : unsigned int
                                               //           This flag is useful in cases where it is required to generate register
                                               //           indirect addressing mode. One such case is virtual stub calls on xarch.
     GTF_IND_UNALIGNED           = 0x02000000, // OperIsIndir() -- the load or store is unaligned (we assume worst case alignment of 1 byte)
+    GTF_IND_KNOWN_UNALIGNED     = 0x00100000, // GT_IND -- same as GTF_IND_UNALIGNED (it's expected to be set) with a legal allowance
+                                              //           to use wider loads/stores where needed.
```
SingleAccretion, line 491: `IND_KNOWN_UNALIGNED` is a pretty confusing name in my opinion. We don't actually know if it's unaligned dynamically.
```suggestion
    GTF_IND_ALLOW_NON_ATOMIC    = 0x00100000, // GT_IND -- this memory access does not need be atomic
```
Nit: the flags are sorted by value. It would be good to maintain that.
Nit: you should add this to `GTF_IND_FLAGS`.

EgorBo: Makes sense! Will apply once I get the diffs, want to know how bad they're 😄

EgorBo: Ah, the diffs are not too bad https://dev.azure.com/dnceng-public/public/_build/results?buildId=426134&view=ms.vss-build-web.run-extensions-tab, initial improvements were +400kb per platform.

EgorBo:
@jakobbotsch @dotnet/jit-contrib PTAL, diffs: https://dev.azure.com/dnceng-public/public/_build/results?buildId=427041&view=results (regression as expected because it made the whole https://github.com/dotnet/runtime/pull/92852 algorithm more conservative, but the initial diffs were -400kb so most wins are expected to remain, obviously, most base addresses are TYP_REF like Jakob predicted).

Wins on ARM64 due better SIMD guarantees.

src/coreclr/jit/lower.cpp
```diff
@@ -7920,6 +7928,17 @@ void Lowering::LowerStoreIndirCoalescing(GenTreeStoreInd* ind)
...
+    // TODO-ARM64-CQ: revisit TYP_REF if we find a case where it's beneficial.
+    // The algorithm does support TYP_REF (with null value), but it seems to be not worth
+    // it on ARM64 where it's pretty efficient to do "stp xzr, xzr, [addr]" to clear two
+    // items at once. Although, it may be profitable to do "stp q0, q0, [addr]"
+    //
+    // Other types are added for better throughput to early out.
+    if (ind->TypeIs(TYP_REF, TYP_BYREF, TYP_STRUCT, TYP_FLOAT, TYP_DOUBLE))
+    {
+        return;
+    }
```
jakobbotsch, line 7940: IMO it would be better to write this as a check for the supported types, instead of as a check for unsupported types.

src/coreclr/jit/lower.cpp
```diff
@@ -8003,6 +8016,64 @@ void Lowering::LowerStoreIndirCoalescing(GenTreeStoreInd* ind)
...
+#ifdef TARGET_XARCH
+            // Check whether data can be guaranteed to be in a cache line.
+            // Only AMD/Intel can guarantee that unaligned writes are atomic if they don't cross the
+            // cache line boundary.
+            if ((ptrAlignOffset + genTypeSize(ind) * 2) > TARGET_POINTER_SIZE)
+            {
+                // NOTE: We can use SIMD if we can assume 16-bytes alignment (on Intel/AMD cpus with AVX+),
+                // but it's very unlikely we can detect that from surrounding code since GC only offers 8-bytes one.
+                return;
+            }
```
jakobbotsch, line 8055: I'm not quite sure I understand this check, I would expect it to be
```cpp
bool isCombinedIndirAtomic = (min(prevData.offset, curData.offset) % (genTypeSize(ind) * 2)) == 0;
```

and then additionally allowing the `ind->TypeIs(TYP_LONG, TYP_REF)` cases for arm64 as below.

src/coreclr/jit/lower.cpp
EgorBo: Changed

src/coreclr/jit/lower.cpp
EgorBo: @jakobbotsch ah.. hm.. can you check my recent commit if I addressed it?

Also, I realized I no longer need to set GTF_UNALIGNED on the read since we maintain the alignment.

jakobbotsch: Looks better now, but thinking about it again I think we also need a `genTypeSize(ind) < TARGET_POINTER_SIZE` to take the base alignment into account (if I understand currently the code now might combine 2x `TYP_I_IMPL` into `TYP_SIMD16` on xarch)

jakobbotsch: Sorry, I guess the check actually needs to be `!ind->TypeIs(TYP_I_IMPL)`

EgorBo: Ah, indeed, it could also enable 2xINT to LONG on 32bit.

I think `genTypeSize(ind) < TARGET_POINTER_SIZE` reads better to me

src/coreclr/jit/lower.cpp
```diff
@@ -8015,32 +8077,84 @@ void Lowering::LowerStoreIndirCoalescing(GenTreeStoreInd* ind)
...
+                    }
+                    break;
+                }
+                return;
+
+#if defined(TARGET_AMD64)
+            case TYP_SIMD16:
+                // Check for AVX to present and PreferredVectorByteLength
+                if (comp->compOpportunisticallyDependsOn(InstructionSet_AVX) &&
+                    (comp->getPreferredVectorByteLength() >= 32))
```
jakobbotsch, line 8119: Are both checks necessary?

jakobbotsch (approved):

src/coreclr/jit/lower.cpp
EgorBo: The first one check whether ISA is available to use. The 2nd checks whether we're allowed to use vectors of the given width. `comp->compOpportunisticallyDependsOn(InstructionSet_AVX)` doesn't take `PreferredVectorByteLength` into account because there 128-bit/scalar instructions they can provide which still can be used.

jakobbotsch: I see, I thought `getPreferredVectorByteLength() == 32` implied support for the ISA

EgorBo: UPD: it does, I was wrong

EgorBo:
Improved [Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=428700&view=ms.vss-build-web.run-extensions-tab)

kunalspathak:
> Improved [Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=428700&view=ms.vss-build-web.run-extensions-tab) on arm64

seems there are more regressions on linux/windows x64. Do we know why?

image

kunalspathak:
> > Improved [Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=428700&view=ms.vss-build-web.run-extensions-tab) on arm64
> 
> seems there are more regressions on linux/windows x64. Do we know why?

image

EgorBo:
> seems there are more regressions on linux/windows x64. Do we know why?

these are reverted improvements from https://github.com/dotnet/runtime/pull/92852 because they turned out to be not legal (but fortunately, most improvements remained)

EgorBo:
x86 SPMI jobs failed with timeout/"no space left", I'll check other runs

PR #92941, opened 2023-10-03
Condition the use of NetCoreAppPrevious TFM

ViktorHofer:
NuGet doesn't support duplicate TFMs in the TargetFrameworks string. Condition the use of NetCoreAppPrevious TFMs until NuGet supports that (which is planned afaik).

This fixes a NRE thrown by NuGet inside Visual Studio.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -89,8 +89,9 @@
 
     <!-- The previous supported .NET version. -->
     <NetCoreAppPreviousVersion>8.0</NetCoreAppPreviousVersion>
-    <NetCoreAppPrevious>net$(NetCoreAppPreviousVersion)</NetCoreAppPrevious>
-    <NetCoreAppPrevious Condition="'$(DotNetBuildFromSource)' == 'true'">$(NetCoreAppCurrent)</NetCoreAppPrevious>
+    <!-- For the .NET 9 release, set NetCoreAppPrevious to empty as NuGet has issues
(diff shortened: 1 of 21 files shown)


ericstj:
I see - so for projects that use just `$(NetCoreAppPrevious)` we didn't have to do anything since an extra `;` is allowed, but for any project that used a suffix we had to condition it.

ericstj (approved):

carlossanlop (approved):
Let me try to explain the change to see if I got it: NetCoreAppPrevious is now set to empty to avoid causing trouble to NuGet. It is now removed from all the csprojs that used to mention it. Only if the user explicitly sets a value for it, then it will automatically propagate to those csprojs. Is this correct?

A couple of questions:
- How did we learn about this problem? There's no linked issue so I'm curious.
- What is NuGet doing wrong with NetCoreAppPrevious?
- I assume this needs backporting to net8.0?

ViktorHofer:
> How did we learn about this problem? There's no linked issue so I'm curious.

https://github.com/dotnet/runtime/discussions/92126 made me play with VS and I noticed that restore is completely broken.

> I assume this needs backporting to net8.0?

No. For .NET 8, `NetCoreAppPrevious` is set to `net7.0` and `NetCoreAppMinimum` to `net6.0` so they aren't the duplicate TFMs. For .NET 9 (main branch), these two properties were set to the same TFM. Reason for that is that the previous version and the minimum version are the same thing (net8.0) because .NET 9 only supports net8.0 and net9.0.

PR #92955, opened 2023-10-03
[release/7.0] Fixes not find assembly and pdb if it's not using the default path

thaystg:
## Customer Impact
Fixes https://github.com/dotnet/runtime/issues/93016
When a customer is setting another path for the assets during the Blazor start the debugger is not working.
```
await Blazor.start({
    loadBootResource: function (type: string, name: string, defaultUri: string, integrity: string) {
        return `${defaultUri.replace("_framework", "assets/dotnet")}?${integrity}`;
    }
});
```

## Testing
Manually tested.

## Risk
Low. Only moving the try catch.

### DIFF
--- a/src/mono/wasm/debugger/BrowserDebugProxy/DebugStore.cs
+++ b/src/mono/wasm/debugger/BrowserDebugProxy/DebugStore.cs
@@ -1491,50 +1491,24 @@ namespace Microsoft.WebAssembly.Diagnostics
 
             foreach (string url in asm_files)
             {
-                try
-                {
-                    string candidate_pdb = Path.ChangeExtension(url, "pdb");
-                    string pdb = pdb_files.FirstOrDefault(n => n == candidate_pdb);
+                string candidate_pdb = Path.ChangeExtension(url, "pdb");
+                string pdb = pdb_files.FirstOrDefault(n => n == candidate_pdb);
 
-                    steps.Add(
-                        new DebugItem
-                        {
-                            Url = url,
-                            Data = Task.WhenAll(MonoProxy.HttpClient.GetByteArrayAsync(url, token), pdb != null ? MonoProxy.HttpClient.GetByteArrayAsync(pdb, token) : Task.FromResult<byte[]>(null))
-                        });
-                }
-                catch (Exception e)
-                {
-                    if (tryUseDebuggerProtocol)
+                steps.Add(
+                    new DebugItem
                     {
-                        try
-                        {
-                            string unescapedFileName = Uri.UnescapeDataString(url);
-                            steps.Add(
-                            new DebugItem
-                            {
-                                Url = url,
-                                Data = context.SdbAgent.GetBytesFromAssemblyAndPdb(Path.GetFileName(unescapedFileName), token)
-                            });
-                        }
-                        catch (Exception ex)
-                        {
-                            logger.LogDebug($"Failed to get bytes using debugger protocol {url} ({ex.Message})");
-                        }
-                    }
-                    else
-                    {
-                        logger.LogDebug($"Failed to read {url} ({e.Message})");
-                    }
-                }
+                        Url = url,
+                        Data = Task.WhenAll(MonoProxy.HttpClient.GetByteArrayAsync(url, token), pdb != null ? MonoProxy.HttpClient.GetByteArrayAsync(pdb, token) : Task.FromResult<byte[]>(null))
+                    });
             }
 
             foreach (DebugItem step in steps)
             {
                 AssemblyInfo assembly = null;
+                byte[][] bytes;
                 try
                 {
-                    byte[][] bytes = await step.Data.ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


src/mono/wasm/debugger/BrowserDebugProxy/DebugStore.cs
```diff
@@ -1544,7 +1518,23 @@ public async IAsyncEnumerable<SourceFile> Load(SessionId id, string[] loaded_fil
...
+                            assembly = new AssemblyInfo(monoProxy, id, step.Url, bytes[0], bytes[1], logger, token);
+                        }
+                        else
+                        {
+                            logger.LogDebug($"Failed to read {step.Url} ({e.Message})");
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        logger.LogError($"Failed to load {step.Url} ({ex.Message})");
```
dlemstra, line 1536: Wondering why `ex` is not passed as an argument to `LogError`? This might provide the user with more information?

src/mono/wasm/debugger/BrowserDebugProxy/DebugStore.cs
```diff
@@ -1544,7 +1518,23 @@ public async IAsyncEnumerable<SourceFile> Load(SessionId id, string[] loaded_fil
...
+                    {
+                        if (tryUseDebuggerProtocol)
+                        {
+                            string unescapedFileName = Uri.UnescapeDataString(step.Url);
+                            bytes = await context.SdbAgent.GetBytesFromAssemblyAndPdb(Path.GetFileName(unescapedFileName), token);
+                            assembly = new AssemblyInfo(monoProxy, id, step.Url, bytes[0], bytes[1], logger, token);
+                        }
+                        else
+                        {
+                            logger.LogDebug($"Failed to read {step.Url} ({e.Message})");
```
radical, line 1531: I think he meant https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions.logerror?view=dotnet-plat-ext-7.0#microsoft-extensions-logging-loggerextensions-logerror(microsoft-extensions-logging-ilogger-system-exception-system-string-system-object()) . Does the error show up in VS? how?

src/mono/wasm/debugger/BrowserDebugProxy/DebugStore.cs
```diff
@@ -1544,7 +1518,23 @@ public async IAsyncEnumerable<SourceFile> Load(SessionId id, string[] loaded_fil
...
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        logger.LogError($"Failed to load {step.Url} ({ex})");
```
radical, line 1536: I think he meant https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions.logerror?view=dotnet-plat-ext-7.0#microsoft-extensions-logging-loggerextensions-logerror(microsoft-extensions-logging-ilogger-system-exception-system-string-system-object()) . Does the error show up in VS? how?

src/mono/wasm/debugger/BrowserDebugProxy/DebugStore.cs
```diff
@@ -1544,7 +1518,23 @@ public async IAsyncEnumerable<SourceFile> Load(SessionId id, string[] loaded_fil
                 }
                 catch (Exception e)
                 {
-                    logger.LogError($"Failed to load {step.Url} ({e.Message})");
+                    try
+                    {
+                        if (tryUseDebuggerProtocol)
+                        {
+                            string unescapedFileName = Uri.UnescapeDataString(step.Url);
+                            bytes = await context.SdbAgent.GetBytesFromAssemblyAndPdb(Path.GetFileName(unescapedFileName), token);
```
radical, line 1526: `ConfigureAwait(false)`

src/mono/wasm/debugger/BrowserDebugProxy/DebugStore.cs
dlemstra: That's what I meant. I would expect that to should up in the output of the console application that is being launched when running the debugger? Now also wondering if the `e` should also be passed to the `logger.LogDebug` method.

dlemstra:
And thanks for this quick fix thaystg. Happy that you could reproduce this error and find what was causing this so quickly. I am wondering if extra unit tests should be added for this?

thaystg:
> And thanks for this quick fix thaystg. Happy that you could reproduce this error and find what was causing this so quickly. I am wondering if extra unit tests should be added for this?
This is a scenario that has a different implementation for .net8, which doesn't matter the file url, I don't think we need unit tests for it.

radical (approved):

carlossanlop:
@thaystg has all feedback been addressed? If yes, can you please send an email to Tactics requesting approval?

carlossanlop:

Issue #92944, opened 2023-10-03
System.Diagnostics.Tests.ProcessTests.TestCheckChildProcessUserAndGroupIds failing Subset assertion


ericstj:
## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=425348
Build error leg or test failing: System.Diagnostics.Tests.ProcessTests.TestCheckChildProcessUserAndGroupIds
Pull request: https://github.com/dotnet/runtime/pull/92730

## Error Message

Similar issue reported in past as https://github.com/dotnet/runtime/issues/28922

```
Child exception:
  Xunit.Sdk.SubsetException: Assert.Subset() Failure
Expected: HashSet<UInt32> [12, 33, 61, 80, 98, ...]
Actual:   HashSet<UInt32> [80, 12, 20, 61, 701, ...]
   at System.Diagnostics.Tests.ProcessTests.CheckUserAndGroupIds(String userId, String groupId, String groupIdsJoined, String checkGroupsExact) in /_/src/libraries/System.Diagnostics.Process/tests/ProcessTests.Unix.cs:line 571
   at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args) in /_/src/mono/System.Private.CoreLib/src/System/Reflection/MethodBaseInvoker.Mono.cs:line 22
   at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr) in /_/src/libraries/System.Private.CoreLib/src/System/Reflection/MethodBaseInvoker.cs:line 178

Child process:
  System.Diagnostics.Process.Tests, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51 System.Diagnostics.Tests.ProcessTests Int32 CheckUserAndGroupIds(System.String, System.String, System.String, System.String)

Child arguments:
  504, 80, 12,33,61,80,98,100,204,250,395,398,399,400,701,702,703, False
```

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json
{
[...]

PR #92960, opened 2023-10-03
Adding a vectorized implementation of TensorPrimitives.Log

tannergooding:

### DIFF
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
@@ -2795,6 +2795,71 @@ namespace System.Numerics.Tensors
 #if NET8_0_OR_GREATER
             public static Vector512<float> Invoke(Vector512<float> x)
             {
+                Vector512<float> specialResult = x;
+
+                // x is subnormal or infinity or NaN
+                Vector512<uint> specialMask = Vector512.GreaterThanOrEqual(x.AsUInt32() - Vector512.Create(V_MIN), Vector512.Create(V_MAX - V_MIN));
+
+                if (specialMask != Vector512<uint>.Zero)
+                {
+                    // float.IsZero(x) ? float.NegativeInfinity : x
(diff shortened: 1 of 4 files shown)


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2579,6 +2579,226 @@ public static Vector512<float> Invoke(Vector512<float> x, Vector512<float> y)
...
+                    specialResult,
+                    n * Vector256.Create(V_LN2) + q
+                );
+            }
+
+#if NET8_0_OR_GREATER
+            public static Vector512<float> Invoke(Vector512<float> x)
+            {
+            }
+#endif
```
stephentoub, line 2799: This doesn't look right :smile:

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2579,6 +2579,291 @@ public static Vector512<float> Invoke(Vector512<float> x, Vector512<float> y)
 #endif
         }
 
+        private readonly struct LogOperator : IUnaryOperator
+        {
+            // This code is based on `vrs4_logf` from amd/aocl-libm-ose
```
stephentoub, line 2584: We already have the right callouts in the thirdparty file, right?

stephentoub (approved):

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: Yep! We have the general encompassing license notice here: https://github.com/dotnet/runtime/blob/main/THIRD-PARTY-NOTICES.TXT#L1139-L1164

And then the explicit callout like this in each function that's ported something, so users have something to refer back to/cross reference more easily.

stephentoub:
Failure is https://github.com/dotnet/runtime/issues/92944

PR #92799, opened 2023-09-28
runtime-wasm-perf: add triggers for PRs

radical:
This is useful to prevent perf pipeline from breaking when changes are
made in `dotnet/runtime`.

PR #92964, opened 2023-10-03
[IGNORE][wasm] testing perf pipeline

radical:


radical:
@dougbu `runtime-wasm-perf` was getting triggered for all PRs. Then I merged https://github.com/dotnet/runtime/pull/92799 with `pr:` section, and `trigger: none`. Now, the pipeline is not getting triggered for PRs by default, like this one which has a change in one the included paths.

Do I need a `trigger:` also? Looking at the docs `trigger` seems to be for running off the repo branches like a rolling build, and `pr` is for running the pipeline on PRs. Any idea what I'm missing here?

dougbu:
@radical @AlitzelMendez re-ticked the `Require a team member's comment before building a pull request` checkbox; I'm not sure why. @AlitzelMendez❔

that checkbox is off again

PR #92966, opened 2023-10-03
Replace CMake Object Interfaces with CMake 3.20 Objects

### LINKED ISSUE #92787: Remove the `_obj` suffix on our CMake targets and remove the "interface" targets now that `target_link_libraries` can link to object libraries.

ivdiazsa:
Resolves #92787. Now that we have set 3.20 as the minimum version of **CMake** in the repo, we no longer need to separate `OBJECT` and `INTERFACE` targets in the _CMakeLists.txt_ files. This PR removes said `INTERFACE` targets, and adjusts the `OBJECT` ones accordingly, in order to take advantage of 3.20's ability to link directly to object libraries with `target_link_libraries()`.


jkoritzinsky:
The CG2 failures in the outerloop are unrelated. I'll take a look separately.

jkoritzinsky (approved):

ivdiazsa:
Merging this now since failures are unrelated and are being addressed in another PR.

PR #92971, opened 2023-10-03
Fix the path for crossgen2 for the outerloop cg2 legs

jkoritzinsky:
Fixes failures in the R2R-CG2 legs in the outerloop pipeline.


hoyosjs (approved):

jkoritzinsky:
Tests are running in all legs now. Merging this to get the pipeline back to at least this state.

PR #91808, opened 2023-09-08
[mono] Second try at using C11 atomics

lambdageek:
* For each platform decide if we will use C11 standard atomics, Win32 API atomics, GCC atomic intrinsics or emulated atomics.

   To use C11 atomics we generally want them to be lock-free for the primitive types we care about (that is, `ATOMIC_LONG_LONG_LOCK_FREE == 2` not `1` and similarly for other macros) because otherwise we cannot be sure if the atomic might use a global lock in which case we could have thread suspend problems if both the GC and the rest of the runtime use an atomic at the same time.

* On win32, use the win32 atomics until MSVC atomics support is not experimental, or we update our build to pass `/experimental:c11atomics`; and also until we build our C++ code with C++23 or later (otherwise MSVC will complain about including stdatomic.h)

* If the header gets included while we're generating offsets using `offsets-tool.py`, pretend we're using emulated atomics.  On some Linux configurations, the libclang that we use ends up picking up the platform `atomic.h` header, not the clang one, and then errors out on their underlying implementation.

* Replace the use of `bool` in some macros in Mono - it will expand to `_Bool` in C11 and mess things up.

   Use `boolean` instead

   Fixes https://github.com/dotnet/runtime/issues/91779

PR #92974, opened 2023-10-03
Implement Interlocked for small types.

### LINKED ISSUE #64658: [API Proposal]: Interlocked.CompareExchange for byte/short/sbyte/ushort
### Background and motivation

We have a few flags fields in the compiler that can be set by multiple threads, used as part of our symbol models. As we avoid locks whenever possible, we generally use a loop on `Interlocked.CompareExchange`, under the assumption it will succeed eventually or the value will have been computed and set by some other thread. However, because `CompareExchange` only supports `int`, `long`, `uint`, and `ulong`  integer values, these fields must be at least 1 `int` wide. This doesn't sound like much, but as the compiler needs to load many, many symbols, those few extra bits can add up. I'd like to be able to use `byte`s where possible for these values. I only really want `byte`/`ushort` overloads, but added the signed variants for completeness.

### API Proposal

```C#
namespace System.Threading
{
    public static class Interlocked
    {
        public static byte CompareExchange(ref byte location1, byte value, byte comparand);
        public static sbyte CompareExchange(ref sbyte location1, sbyte value, sbyte comparand);
        public static short CompareExchange(ref short location1, short value, short comparand);
        public static ushort CompareExchange(ref ushort location1, ushort value, ushort comparand);
    }
}
```

### API Usage

```C#
        public static bool Set(ref byte flags, byte toSet)
        {
            byte oldState, newState;
            do
            {
                oldState = flags;
                newState = oldState | toSet;
                if (newState == oldState)
                {
                    return false;
                }
            }
            while (Interlocked.CompareExchange(byte flags, newState, oldState) != oldState);
            return true;
        }
```

### Alternative Designs

_No response_

### Risks

_No response_

MichalPetryka:
Implements Interlocked.CompareExchange/Exchange for byte/sbyte/short/ushort.

Uses an intrinsic on CoreCLR x64 and ARM64 and fallbacks for other platforms.

Fixes #64658.


src/libraries/System.Threading/tests/InterlockedTests.cs
```diff
@@ -240,6 +272,54 @@ public void InterlockedExchange_BoxedObject()
...
+        {
+            ushort value = 42;
+
+            Assert.Equal(42u, Interlocked.CompareExchange(ref value, 12345, 41));
+            Assert.Equal(42u, value);
+
+            Assert.Equal(42u, Interlocked.CompareExchange(ref value, 12345, 42));
+            Assert.Equal(12345u, value);
+        }
+
```
stephentoub, line 322: I think it'd be good to add some tests that use our BoundedMemory helper to ensure we're not touching memory outside of the single value being exchanged. Some of the proposed implementations in the past would have erroneously done so.

MichalPetryka: I'll take a look tomorrow, opened the PR after running the tests on linux arm64 to see how other platforms react with the CI for now.

src/mono/mono/utils/atomic.h
```diff
@@ -266,6 +303,24 @@ static inline gint32 mono_atomic_dec_i32(volatile gint32 *val)
 	return gcc_sync_sub_and_fetch (val, 1);
 }
 
+static inline guint8 mono_atomic_xchg_u8(volatile guint8 *val, guint8 new_val)
+{
+	guint8 old_val;
```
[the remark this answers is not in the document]
MichalPetryka: For now I wanted to keep the new intrinsics same as the old one.
I assume there might be some historic reasons with old compilers that cause things to be this way, so I don't want to touch it without maintainers asking me to do so.

src/libraries/System.Threading/tests/InterlockedTests.cs
MichalPetryka: Done.

MichalPetryka:
With this the fallbacks will be used again for the small types on RISC-V, cc @tomeksowi in case you want to add intrinsics for those too since you've recently added intrinsics for the existing overloads in #92102.

src/mono/mono/utils/atomic.h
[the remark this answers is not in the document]
lambdageek: > I am wondering. Why are we not using the `__atomic_exchange_n` bultin from gcc instead of this loop? (This question also applies to the other xchg implementations)

Mostly because mono is old - this code is older than gcc's newer atomics.  The old sync intrinsics didn't have the right operation.

I have a https://github.com/dotnet/runtime/pull/91808 open to will switch us over to C11 atomics on platforms where that makes sense.  But I'm not sure there's a lot of motivation to touch this older implementation as long as GCC doesn't ditch the legacy sync atomics.

src/mono/mono/utils/atomic.h
```diff
@@ -457,6 +512,8 @@ static inline void mono_atomic_store_i64(volatile gint64 *dst, gint64 val)
 
 #define WAPI_NO_ATOMIC_ASM
 
+extern guint8 mono_atomic_cas_u8(volatile guint8 *dest, guint8 exch, guint8 comp);
```
lambdageek, line 515: It's interesting that we don't need implementations for these in atomic.c. I guess we don't have any platforms with emulated atomics anymore

MichalPetryka:
@MihuBot

MichalPetryka:
@MihuBot -arm

tomeksowi:
> With this the fallbacks will be used again for the small types on RISC-V, cc @tomeksowi in case you want to add intrinsics for those too since you've recently added intrinsics for the existing overloads in #92102.

Thanks for the heads-up. RISC-V atomics don't come in short-int variants so there would have to be additional masking, in that case CAS blows up to a sequence of dozen instructions or so (which clang even outlines on -Os), Interlocked.Exchange which normally is one instruction must be a full lr-sc loop for short ints.

I will put them on my todo but won't work on them anytime soon as we're focusing more on stability now. IMHO there's little gain over just calling these fallbacks.

EgorBo:
/azp run runtime-nativeaot-outerloop

(on @MichalPetryka's request)

lambdageek (approved):
Mono changes LGTM

lambdageek:
Created https://github.com/dotnet/runtime/issues/93488 for mono follow-up work to actually support these operations as intrinsics in the mono JIT/AOT and interpreter

MichalStrehovsky (approved):
Native AOT specific changes LGTM.

mangod9:
@stephentoub, assume you are good with the overall change? If so we can merge.

MichalPetryka:
> If so we can merge.

This needs a review on the CoreCLR JIT changes too.

JulieLeeMSFT:
cc @EgorBo, PTAL.

src/coreclr/System.Private.CoreLib/src/System/Threading/Interlocked.CoreCLR.cs
```diff
@@ -42,6 +42,24 @@ public static partial class Interlocked
...
 
         #region Exchange
+        /// <summary>Sets a 8-bit unsigned integer to a specified value and returns the original value, as an atomic operation.</summary>
+        /// <param name="location1">The variable to set to the specified value.</param>
+        /// <param name="value">The value to which the <paramref name="location1"/> parameter is set.</param>
+        /// <returns>The original value of <paramref name="location1"/>.</returns>
+        /// <exception cref="NullReferenceException">The address of location1 is a null pointer.</exception>
+        [Intrinsic]
+        [MethodImpl(MethodImplOptions.InternalCall)]
+        public static extern byte Exchange(ref byte location1, byte value);
```
AaronRobinsonMSFT, line 52: Instead of creating an FCall here, I think we should consider the cost of a QCall and applying `SuppressGCTransition`. Unless there is a clear measurable benefit for an FCall we should be deferring to QCalls.

AaronRobinsonMSFT: > I feel like that could be done in a separate PR to not add any additional parts to this PR.

Sure. It is early enough in .NET 9, that that is a fair approach. Look forward to that PR.

MichalPetryka: > Unless there is a clear measurable benefit for an FCall we should be deferring to QCalls.

As for this, the fallback path should only be used on RISC-V, ARM32 and WASM (with NativeAOT-LLVM) AFAIR. So none of those are currently of huge performance concern.

jkotas: Converting this method to QCall would leave a lot of perf on the table (the `ref` argument would require marshalling among other things).

It agre that the FCalls added in this PR are JIT intrinsics on x64 and arm64, but there are other pre-existing perf-critical System.Interlocked FCalls that are not JIT intrinsics and that won't be easy to turn into JIT intrinsics.

We should be considering all System.Interlocked FCalls - including the perf critical ones - for any conversions.

AaronRobinsonMSFT: > Converting this method to QCall would leave a lot of perf on the table (the ref argument would require marshalling among other things).

For this it would be a `fixed` statement and then call with a pointer. I agree there is cost but I think understanding how much is perhaps worth the investigation for something as minor as an interlocked op on a byte or short. Obviously not in this PR, but it does seem a bit hand wavey to say perf-critical without a known benchmark to follow up on.

jkotas: I agree that any changes in the pattern should be measured. I am pretty sure that the QCall overhead would be prohibitive for `Interlocked.CompareExchange(ref object, object, object)` and `Interlocked.CompareExchange(ref T, T, T)`.

stephentoub:
> @stephentoub, assume you are good with the overall change?

Yes.  Subsequently, we should go through use of Interlocked in the core libraries looking for opportunities to use these.  There are at least a handful of places we use Interlocked with an int but could get away with it being a byte, and in some cases we may be able to shave off a few bytes of allocation.

MichalPetryka:
> should go through use of Interlocked in the core libraries looking for opportunities to use these

Worth noting that doing so would permanently pessimize the code performance on RISC-V.

jkotas:
> Worth noting that doing so would permanently pessimize the code performance on RISC-V.

I would not say permanently. Over time, RISC-V capabilities are likely to catch up to be on parity with other mainstream architectures.

MichalPetryka:
> I would not say permanently. Over time, RISC-V capabilities are likely to catch up to be on parity with other mainstream architectures.

RISC-V does not expose fast 1B/2B atomics in their Atomics extension, only 4B/8B ones. So unless RISC-V gets a new extension for small type atomics, the performance will be worse with those overloads.

stephentoub:
From my perspective, if we're not comfortable using these, we shouldn't be exposing them. And conversely if we choose to expose them, we should be comfortable using them.

MichalPetryka:
> From my perspective, if we're not comfortable using these, we shouldn't be exposing them. And conversely if we choose to expose them, we should be comfortable using them.

Since only a single arch might have potential performance issues (I have no way to test how much worse the small type overloads are there, the Samsung RISC-V folks would need to do so) and since modern technologies like WASM or even Linux Kernel rely on small atomics (the kernel emulates them with 4B ones as far as I've seen), I wouldn't consider this to be blocking.

tomeksowi:
> unless RISC-V gets a new extension for small type atomics

This extension is [Zabha](https://github.com/riscv/riscv-zabha). Not yet ratified though, target date is Q1'25.

Klotzi111:
> Yes. Subsequently, we should go through use of Interlocked in the core libraries looking for opportunities to use these. There are at least a handful of places we use Interlocked with an int but could get away with it being a byte, and in some cases we may be able to shave off a few bytes of allocation.

Just to clarify: Should these usages in the core library be in this PR or in a separate one?  
If in a separate one. Can this PR being merged than?

EgorBo:
I was playing with your branch locally via IL tests:
```
.assembly _ {}
.assembly extern System.Runtime {}
.assembly extern System.Runtime.Extensions {}
.assembly extern System.Console {}
.assembly extern System.Threading {}

.class private auto ansi '<Module>' {}

.class private auto ansi sealed beforefieldinit Program
    extends [System.Runtime]System.Object
{
    .field private int16 'field'
    .field private int16 'field2'

    .method private hidebysig static void Main () cil managed 
    {
        .maxstack 8
        .entrypoint

        newobj instance void Program::.ctor()
        
        dup
        ldc.i4.s -128
        stfld int16 Program::'field'
        
        dup
        ldc.i4 0
        stfld int16 Program::'field2'

        dup
        callvirt instance int16 Program::Test()
        call void [System.Console]System.Console::WriteLine(int32)
        
        ldfld int16 Program::'field'
        call void [System.Console]System.Console::WriteLine(int32)
        ret
    }

    .method instance int16 Test () noinlining
    {
        .maxstack 8

        ldarg.0
        ldflda int16 Program::'field'

        ldarg.0
        ldfld int16 Program::'field2'
        call int16 [System.Threading]System.Threading.Interlocked::Exchange(int16&, int16)
        ret
    }

    .method public hidebysig specialname rtspecialname 
        instance void .ctor () cil managed 
    {
        .maxstack 8
        ldarg.0
        call instance void [System.Runtime]System.Object::.ctor()
        ret
    }
}
```

Can you help me to understand why this program prints
```
65408
0
```
it was supposed to be
```
-128
0
```

EgorBo (approved):
JIT side LGTM, checked locally that the recent changes fix my test ^

MichalPetryka:
@jkotas Do you have any objections with merging this?

src/coreclr/inc/winwrap.h
```diff
@@ -254,18 +254,32 @@ WszCreateProcess(
     LPPROCESS_INFORMATION lpProcessInformation
     );
 
-#if defined(HOST_X86) && defined(_MSC_VER)
+#if defined(_MSC_VER)
 
 //
-// Windows SDK does not use intrinsics on x86. Redefine the interlocked operations to use intrinsics.
+// Redefine the interlocked operations to intrinsics
+// to avoid missing InterlockedCompareExchange8 and Windows 8 requirement.
```
jkotas, line 261: What is the Windows 8 requirement that this is talking about?

src/coreclr/inc/winwrap.h
```diff
@@ -254,18 +254,32 @@ WszCreateProcess(
...
 //
 
 #include "intrin.h"
 
-#define InterlockedIncrement            _InterlockedIncrement
-#define InterlockedDecrement            _InterlockedDecrement
+#define InterlockedExchange8            _InterlockedExchange8
+#define InterlockedCompareExchange8     _InterlockedCompareExchange8
```
jkotas, line 267: If this is working around missing InterlockedCompareExchange8, the workaround should just have InterlockedCompareExchange8 - the rest is unnecessary.

Do you happen to know if there is a bug opened on InterlockedCompareExchange8 missing? If not, we should open one.

src/coreclr/inc/winwrap.h
```diff
@@ -254,18 +254,32 @@ WszCreateProcess(
...
+#define InterlockedCompareExchange16    _InterlockedCompareExchange16
 #define InterlockedExchange             _InterlockedExchange
 #define InterlockedCompareExchange      _InterlockedCompareExchange
+
+#endif // _MSC_VER
+
+#if defined(HOST_X86) && defined(_MSC_VER)
+
+//
+// Windows SDK does not use intrinsics on x86. Redefine the interlocked operations to use intrinsics.
```
jkotas, line 278: As far as I can tell, this is fixed in Windows SDK now and this workaround can be deleted.

src/coreclr/inc/winwrap.h
MichalPetryka: According to the documentation, [InterlockedExchange8](https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedexchange8) and [InterlockedCompareExchange16](https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchange16) are exposed only on Windows 8+. My reasoning was that even though Windows 7 is not supported anymore, working this around here makes more sense than totally breaking the API there especially since we need to that already for the completely missing [InterlockedCompareExchange8](https://godbolt.org/z/9bWjvq91W).

src/coreclr/inc/winwrap.h
MichalPetryka: > Do you happen to know if there is a bug opened on InterlockedCompareExchange8 missing? If not, we should open one.

Haven't checked if there's one since I don't really know where WinAPI issues go, this should probably be fixed there too but using it then would still not be safe since it'd only be guaranteed to be present in the future OS version DLLs.

src/coreclr/inc/winwrap.h
jkotas: InterlockedExchange8 and InterlockedCompareExchange16 are always inlined as compiler intrinsics. The underlying OS does not matter. 

My guess is that nothing new is mentioned in the docs as supported on Win7, even when it is works there just fine.

I would keep this as simple as possible.

src/coreclr/inc/winwrap.h
jkotas: > Haven't checked if there's one since I don't really know where WinAPI issues go

https://github.com/microsoft/WindowsAppSDK seems to be the place to report Windows SDK issues on github.

src/coreclr/nativeaot/Runtime/portable.cpp
```diff
@@ -368,6 +368,18 @@ COOP_PINVOKE_HELPER(Object *, RhpCheckedXchg, (Object ** location, Object * valu
     return ret;
 }
 
+COOP_PINVOKE_HELPER(uint8_t, RhpLockCmpXchg8, (uint8_t * location, uint8_t value, uint8_t comparand))
+{
+    ASSERT_UNCONDITIONALLY("NYI");
```
jkotas, line 373: This will create work for wasm in runtimelab once these APIs get used. Can this be implemented in this PR? (It should be just copy & paste in a few places.)

src/coreclr/inc/winwrap.h
MichalPetryka: It seems like `InterlockedExchange8` [is missing too on ARM64 MSVC](https://godbolt.org/z/v39Trfae6).

> InterlockedExchange8 and InterlockedCompareExchange16 are always inlined as compiler intrinsics. The underlying OS does not matter.

Since the docs say `This function is implemented using a compiler intrinsic where possible.`, I assumed the compiler might emit a call to a method on some platforms.
Even then, I don't think doing this selectively makes sense here, if some intrinsics require redefining to work properly, it makes more sense to consistently redefine all of them in such case.

src/coreclr/nativeaot/Runtime/portable.cpp
MichalPetryka: I've partially reintroduced the helper methods after @SingleAccretion asked me to do so for WASM (and RISC-V) but I've left them NYI on purpose since they require the raw assembly helpers I've deleted here on XArch and ARM64 paths (after making them always expand) for debugging support and implementing them here without adding those would create an untested code path which would still require changes on the LLVM port for those WASM helpers.

src/coreclr/inc/winwrap.h
jkotas: > It seems like InterlockedExchange8 [is missing too on ARM64 MSVC](https://godbolt.org/z/v39Trfae6).

The define for InterlockedExchange8 can stay then.

I see these defines as workarounds for Windows SDK bugs. Once these bugs are fixed, these workarounds can be deleted.

We just need the minimum required to work around the bugs in the current Windows SDK. We do not need to grow this into something complete or permanent.

src/coreclr/nativeaot/Runtime/portable.cpp
jkotas: Ok, these can be implemented in separate PR.

SingleAccretion: FWIW, the downstream perspective here is that the small-typed atomics will be implemented as must-expand intrinsics (like `int` and `long` ones are, currently). I have added a note about this to https://github.com/dotnet/runtimelab/issues/2472.

src/coreclr/inc/winwrap.h
jkotas: Opened https://github.com/microsoft/WindowsAppSDK/issues/4074

src/coreclr/inc/winwrap.h
jkotas: I have pushed a commit to keep make the workaround minimal and reference the issue.

jkotas (approved):
The changes outside the JIT look good to me. Thank you!

src/coreclr/inc/winwrap.h
MichalPetryka: > I have pushed a commit to keep make the workaround minimal and reference the issue.

I think you missed `InterlockedExchange8` there.

jkotas: `InterlockedExchange8` is there and it works fine as far as I can tell.

jkotas:
Build errors in tests:
```
src/tests/JIT/Intrinsics/Interlocked.cs(16,27): error XUW1001: Projects in merged tests group should not have entry points. Convert to Facts or Theories. [/__w/1/s/src/tests/JIT/Intrinsics/Interlocked_r.csproj] [/__w/1/s/src/tests/build.proj]
```

MichalPetryka:
> Build errors in tests:
> 
> ```
> src/tests/JIT/Intrinsics/Interlocked.cs(16,27): error XUW1001: Projects in merged tests group should not have entry points. Convert to Facts or Theories. [/__w/1/s/src/tests/JIT/Intrinsics/Interlocked_r.csproj] [/__w/1/s/src/tests/build.proj]
> ```

Will fix this soon, I'm getting a new machine set up currently.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Interlocked.cs
```diff
@@ -11,16 +11,44 @@ public static partial class Interlocked
...
+            return CompareExchange(ref location1, value, comparand);
+#else
+            return RuntimeImports.InterlockedCompareExchange(ref location1, value, comparand);
+#endif
+        }
+
         [Intrinsic]
         public static int CompareExchange(ref int location1, int value, int comparand)
         {
+#if TARGET_AMD64 || TARGET_ARM64
```
jkotas, line 37:
```suggestion
#if TARGET_X86 || TARGET_AMD64 || TARGET_ARM64 || TARGET_RISCV64
```
Nit: It would be nice to make these ifdefs match the CoreCLR w/ JIT variant, so that there is no confusion about why they are different.

(I know that native AOT is not - yet - supported on RiscV.)

jkotas:
Linux x86 build break:
```
  ld.lld: error: undefined symbol: __atomic_exchange_8
```

MichalPetryka:
@jkotas Looks like there are some new pipelines for Linux x86 and they apparently lack `__atomic_exchange_8`, what's the preferred way of solving this?

jkotas:
>  they apparently lack __atomic_exchange_8, what's the preferred way of solving this?

Is this expected to be always expanded as compiler intrinsic on Linux x86? My guess is that the problem is that the intrinsic definition is off and it is not getting expanded for some reason.

MichalPetryka:
> > they apparently lack __atomic_exchange_8, what's the preferred way of solving this?
> 
> Is this expected to be always expanded as compiler intrinsic on Linux x86? My guess is that the problem is that the intrinsic definition is off and it is not getting expanded for some reason.

I'm pretty sure it is expanded, the code is just compiled for all platforms there.

MichalPetryka:
> Linux x86 build break:
> 
> ```
>   ld.lld: error: undefined symbol: __atomic_exchange_8
> ```

I've just realised, this is erroring on the 64bit exchange. Why is this only failing here and not on main then?

jkotas:
> I've just realised, this is erroring on the 64bit exchange. Why is this only failing here and not on main then?

That is a good question. This change should not be introducing any dependencies on 64bit exchange, right?

src/coreclr/pal/src/CMakeLists.txt
```diff
@@ -347,6 +347,13 @@ if(CLR_CMAKE_TARGET_SUNOS)
   )
 endif(CLR_CMAKE_TARGET_SUNOS)
 
+if (NOT HAVE_BUILTIN_GNUC_ATOMICS)
+  target_link_libraries(coreclrpal
+    PUBLIC
+    atomic
```
jkotas, line 353: If this is really needed, we would need to understand licensing and whether it is introducing any new pre-requisites.

MichalPetryka:
> > I've just realised, this is erroring on the 64bit exchange. Why is this only failing here and not on main then?
> 
> That is a good question. This change should not be introducing any dependencies on 64bit exchange, right?

None that I know of.

src/coreclr/pal/src/CMakeLists.txt
MichalPetryka: Googling around seems to indicate `libatomic` is required for 64bit atomics on some 32bit platforms, AFAIR it's licensed the same as `libgcc` (which we already depend on for most platforms) and both should be installed together on most systems.

I'm wondering why x86 seems to still fail with it though.

src/coreclr/inc/winwrap.h
```diff
@@ -254,112 +254,14 @@ WszCreateProcess(
...
-    )
-{
-    return((PVOID)(LONG_PTR)_InterlockedCompareExchange((LONG volatile *)Destination, (LONG)(LONG_PTR)ExChange, (LONG)(LONG_PTR)Comperand));
-}
-
-#endif // HOST_X86 && _MSC_VER
-
-#if defined(HOST_X86) & !defined(InterlockedIncrement64)
-
-// Interlockedxxx64 that do not have intrinsics are only supported on Windows Server 2003
```
jkotas, line 299: I am wondering whether the removal of this Windows Server 2003 workaround can be introducing the libatomic dependendency.

MichalPetryka: I thought this header is windows only?

jkotas: CoreCLR Win32 PAL is a partial Windows emulator. A lot of things that look Windows-specific in CoreCLR are actually cross-plat.

Submitted #97358 to check

jkotas: Yes, it is the problem. You can add the `#if defined(HOST_X86) & !defined(InterlockedIncrement64)` block back to fix the Linux x86 build breaks and revert the libatomic linking.

jkotas: https://github.com/dotnet/runtime/pull/97358 is merged.

MichalPetryka: I'll update the PR and fix the tests later today.

src/coreclr/pal/src/configure.cmake
```diff
@@ -78,6 +78,18 @@ int main(int argc, char **argv) {
...
+  unsigned long long u8 = 0;
+  unsigned int u4 = 0;
+  unsigned short u2 = 0;
+  unsigned char u1 = 0;
+  __atomic_exchange_n(&u8, 1, __ATOMIC_ACQ_REL);
+  __atomic_exchange_n(&u4, 1, __ATOMIC_ACQ_REL);
+  __atomic_exchange_n(&u2, 1, __ATOMIC_ACQ_REL);
+  return __atomic_exchange_n(&u1, 1, __ATOMIC_ACQ_REL);
+}" HAVE_BUILTIN_GNUC_ATOMICS)
+
```
jkotas, line 92:
```suggestion
```

src/coreclr/pal/src/CMakeLists.txt
```diff
@@ -347,6 +347,13 @@ if(CLR_CMAKE_TARGET_SUNOS)
...
+  )
+endif()
+
```
jkotas, line 356:
```suggestion
```

MichalPetryka:
@jkotas I've realised there's a potential issue with atomics as FCalls: some compilers might implement them with locking on a mutex which AFAIR is not okay inside of an FCall. Checking with [CLang on godbolt makes it seem to be the case on arm32](https://godbolt.org/z/5jnKGh875).

jkotas:
It is a problem with `__atomic_exchange_n` that is used for InterlockedExchange. `__atomic_exchange_n` is also the one that caused troubles on Linux x86 before.

Other operations are implemented using `__sync_val_...` intrinsics. Do `__sync_val_...` intrinsics have the same problem?

If `__sync_val_...` intrinsics are fine:
- Use `__sync_swap` to implement InterlockedExchange when compiling using LLVM. We do that here already: https://github.com/dotnet/runtime/blob/d3583b8360071725f525f292e8aeafa1f867a341/src/coreclr/nativeaot/Runtime/unix/PalRedhawkInline.h#L30-L34
- Implement InterlockedExchange using `__sync_val_compare_and_swap` when not compiling using LLVM. Similar to what I have done to fix the Linux x86 break: https://github.com/dotnet/runtime/blob/d3583b8360071725f525f292e8aeafa1f867a341/src/coreclr/pal/inc/pal.h#L3645-L3661

MichalPetryka:
ARM32 failure seems like an unrelated JIT bug with `long`s.

jkotas:
> I've opened a new issue to track this as this PR doesn't introduce the issue: https://github.com/dotnet/runtime/issues/97452.

It does not introduce it, but it makes the problem worse. Could you please replace the new uses of `__atomic_exchange_n` with what I have suggested? It should address the issue. (Bonus points for fixing the existing uses too.)

MichalPetryka:
> > I've opened a new issue to track this as this PR doesn't introduce the issue: #97452.
> 
> It does not introduce it, but it makes the problem worse. Could you please replace the new uses of `__atomic_exchange_n` with what I have suggested? It should address the issue. (Bonus points for fixing the existing uses too.)

As I've noted in that issue, those other APIs also use helper calls so I'd assume they have the same issue.

MichalPetryka:
@MihuBot

MichalPetryka:
@jkotas I think the arm32 atomic helper issue should be fixed in a separate PR (I can take a look at using the safe LLVM helpers later today) so this should be ready to merge.

jkotas:
> Turns out a small bug in the helpers somehow trashed the stack.

What was this bug?

MichalPetryka:
> > Turns out a small bug in the helpers somehow trashed the stack.
> 
> What was this bug?

A missing cast to `UINT8` from `CHAR` which was returned by `InterlockedExchange8`.

jkotas (approved):
Thank you for making this happen!

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Interlocked.cs
```diff
@@ -11,16 +11,44 @@ public static partial class Interlocked
...
+#else
+            return RuntimeImports.InterlockedCompareExchange(ref location1, value, comparand);
+#endif
+        }
+
         [Intrinsic]
         public static int CompareExchange(ref int location1, int value, int comparand)
         {
+#if TARGET_X86 || TARGET_AMD64 || TARGET_ARM64 || TARGET_RISCV64
+            return CompareExchange(ref location1, value, comparand);
```
VSadov, line 38: This is a self-referential "must expand" intrinsic now.

PR #92981, opened 2023-10-04
[wasm] Interpreter automatic PGO

kg:
This PR adds automatic PGO for interpreter tiering. The idea is that by maintaining a list of which interpreter methods were tiered during execution, we can skip directly to generating optimized code on future runs. This will get the application into a steady high-performance state faster on future runs, while also reducing the amount of time/memory we spend on code generation (since normally we generate unoptimized code, then generate optimized code).

The infrastructure for this feature is not platform gated, but the only platform that has an implementation for loading/saving the PGO data right now is the browser, and the runtime option controlling it is only default-on for the browser.

Web application code can either use `INTERNAL.interpgo_load_data` and `INTERNAL.interpgo_save_data` to manually control the loading/saving of the data, or can use the new `withInterpreterPgo(true, autoSaveDelay)` builder method to turn on automatic mode. Automatic mode will attempt to load the data during startup, and will automatically save it after a delay (at which point your application should be in something approximating a steady-state and critical code has already tiered.) The current web implementation repurposes the cache storage code from pavel's memory snapshot feature to store the pgo table in the cache next to the snapshot(s).

This PR also adds basic (controlled by a `#define` and off by default) timing instrumentation for generate_code, because in my testing the browser's profiler was producing wildly incorrect timing data for interpreter codegen. This should make it easy to evaluate how well the feature is working since you can just look at the output.

One interesting finding: The list of method hashes generated on a first run is bigger than the list generated on future runs - I believe this is because once a full set of methods is tiered, we no longer need the tiered version of some of them because the hottest methods have gotten inlined into their callers. So on a second run the list of hashes stabilizes into a smaller list.

In a couple local test runs of browser-bench with this active and without, the time spent in generate_code was 3600ms without interpgo and 3209ms with interpgo, so the improvement over the course of a full run of an application with lots of generated code should be worthwhile. For smaller applications it's below the noise floor, I haven't been able to measure it successfully at that scale.

### DIFF
--- a/src/mono/mono/mini/interp/transform.c
+++ b/src/mono/mono/mini/interp/transform.c
@@ -34,6 +34,7 @@
 #include "interp.h"
 #include "transform.h"
 #include "tiering.h"
+#include "interp-pgo.h"
 
 #if HOST_BROWSER
 #include "jiterpreter.h"
@@ -4831,6 +4832,8 @@ generate_code (TransformData *td, MonoMethod *method, MonoMethodHeader *header, 
 	gboolean generate_enc_seq_points_without_debug_info = FALSE;
 	InterpBasicBlock *exit_bb = NULL;
 
+	mono_interp_pgo_generate_start ();
+
 	original_bb = bb = mono_basic_block_split (method, error, header);
 	goto_if_nok (error, exit);
 	g_assert (bb);
@@ -8247,12 +8250,17 @@ generate_code (TransformData *td, MonoMethod *method, MonoMethodHeader *header, 
 		}
 	}
 
+	if (td->optimized && !inlining)
+		mono_interp_pgo_method_was_tiered (method);
+
 exit_ret:
 	g_free (arg_locals);
 	g_free (local_locals);
 	mono_bitset_free (il_targets);
 	mono_basic_block_free (original_bb);
 	td->dont_inline = g_list_remove (td->dont_inline, method);
+
+	mono_interp_pgo_generate_end ();
 
 	return ret;
 exit:
--- a/src/mono/mono/mini/interp/transform.c
+++ b/src/mono/mono/mini/interp/transform.c
@@ -4832,8 +4832,6 @@ generate_code (TransformData *td, MonoMethod *method, MonoMethodHeader *header, 
 	gboolean generate_enc_seq_points_without_debug_info = FALSE;
 	InterpBasicBlock *exit_bb = NULL;
 
-	mono_interp_pgo_generate_start ();
-
 	original_bb = bb = mono_basic_block_split (method, error, header);
 	goto_if_nok (error, exit);
 	g_assert (bb);
@@ -8260,8 +8258,6 @@ exit_ret:
 	mono_basic_block_free (original_bb);
 	td->dont_inline = g_list_remove (td->dont_inline, method);
 
-	mono_interp_pgo_generate_end ();
-
 	return ret;
 exit:
 	ret = FALSE;
@@ -11135,6 +11131,7 @@ generate (MonoMethod *method, MonoMethodHeader *header, InterpMethod *rtm, MonoG
 	}
 
 retry:
+	mono_interp_pgo_generate_start ();
 	memset (&transform_data, 0, sizeof(transform_data));
 	td = &transform_data;
 
@@ -11339,6 +11336,7 @@ exit:
 		g_array_free (td->line_numbers, TRUE);
 	g_slist_free (td->imethod_items);
 	mono_mempool_destroy (td->mempool);
+	mono_interp_pgo_generate_end ();
 	if (retry_compilation)
 		goto retry;
 }
--- a/src/mono/sample/wasm/browser-bench/main.js
+++ b/src/mono/sample/wasm/browser-bench/main.js
@@ -13,7 +13,7 @@ let legacyExportTargetInt;
 let jsExportTargetInt;
 let legacyExportTargetString;
 let jsExportTargetString;
-let _jiterpreter_dump_stats;
+let _jiterpreter_dump_stats, _interp_pgo_save_data;
 
 function runLegacyExportInt(count) {
     for (let i = 0; i < count; i++) {
@@ -63,10 +63,9 @@ function importTargetThrows(value) {
 class MainApp {
     async init({ getAssemblyExports, setModuleImports, BINDING, INTERNAL }) {
         const exports = await getAssemblyExports("Wasm.Browser.Bench.Sample.dll");
-        INTERNAL.jiterpreter_apply_options({
-            enableStats: true
-        });
+        // Capture these two internal APIs for use at the end of the benchmark run
         _jiterpreter_dump_stats = INTERNAL.jiterpreter_dump_stats.bind(INTERNAL);
+        _interp_pgo_save_data = INTERNAL.interp_pgo_save_data.bind(INTERNAL);
         runBenchmark = exports.Sample.Test.RunBenchmark;
         setTasks = exports.Sample.Test.SetTasks;
         setExclusions = exports.Sample.Test.SetExclusions;
@@ -117,7 +116,6 @@ class MainApp {
             console.log(resultString);
         }
 
-        _jiterpreter_dump_stats();
         document.getElementById("out").innerHTML += "Finished";
         const r1 = await fetch("/results.json", {
             method: 'POST',
@@ -129,6 +127,21 @@ class MainApp {
             body: document.getElementById("out").innerHTML
         });
         console.log("post request complete, response: ", r2);
+
+        // If the jiterpreter's statistics are enabled, make sure we dump them at the
+        //  end of the benchmark run to have a definitive version of them that captures
+        //  everything, since the automatic ones are triggered at a basically random
+        //  interval
+        if (_jiterpreter_dump_stats)
+            _jiterpreter_dump_stats();
+
+        // We explicitly save the interpreter PGO table after a run is complete
+        //  in order to capture all the methods that were tiered while running benchmarks
+        // The normal automatic save-on-startup-timeout behavior is insufficient for
+        //  scenarios like this since we can't predict how long the benchmarks will
+        //  take, and "startup" is spread over the whole run instead of the beginning
+        if (_interp_pgo_save_data)
+            await _interp_pgo_save_data(true);
     }
 
     yieldBench() {
@@ -179,11 +192,27 @@ class MainApp {
                 ? 'unique/' + guid + '/' + page
                 : page;
 
+            let resolved = false;
             promise = new Promise(resolve => { promiseResolve = resolve; })
             window.resolveAppStartEvent = function (event) {
-                if (!eventName || event == eventName)
+                if (!eventName || event == eventName) {
+                    resolved = true;
                     promiseResolve();
+                }
             }
+
+            // The appstart event may never fire if something goes wrong, in that case
+            //  we want to time out within a reasonable period of time so that the run
+            //  of browser-bench will eventually complete instead of just freezing.
+            setTimeout(function () {
+                if (resolved)
+                    return;
+                console.error(`waitFor ${eventName} timed out`);
+                promiseResolve();
+            // Make sure this timeout is big enough that it won't cause measurements
+            //  to be truncated! i.e. "Blazor Reach managed cold" is nearly 10k in some
+            //  configurations right now
+            }, 20000);
 
             document.body.appendChild(this._frame);
             await promise;
@@ -208,7 +237,15 @@ try {
     globalThis.mainApp.SetFramePage = globalThis.mainApp.setFramePage.bind(globalThis.mainApp);
 
     const runtime = await dotnet
+        // We enable jiterpreter stats so that in local runs you can open the devtools
+        //  console to see statistics on how much code it generated and whether any new opcodes
+        //  are causing traces to fail to compile
         .withRuntimeOptions(["--jiterpreter-stats-enabled"])
+        // We enable interpreter PGO so that you can exercise it in local tests, i.e.
+        //  run browser-bench one, then refresh the tab to measure the performance improvement
+        //  on the second run of browser-bench. The overall speed of the benchmarks won't
+        //  improve much, but the time spent generating code during the run will go down
+        .withInterpreterPgo(true, 30)
         .withElementOnExit()
         .withExitCodeLogging()
         .create();
--- a/src/mono/wasm/Wasm.Build.Tests/Templates/InterpPgoTests.cs
+++ b/src/mono/wasm/Wasm.Build.Tests/Templates/InterpPgoTests.cs
@@ -9,6 +9,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using System;
 using System.Diagnostics;
+using Microsoft.Playwright;
 
 #nullable enable
 
@@ -72,7 +73,7 @@ public class InterpPgoTests : WasmTemplateTestBase
         using var runCommand = new RunCommand(s_buildEnv, _testOutput)
                                     .WithWorkingDirectory(_projectDir!);
         await using var runner = new BrowserRunner(_testOutput);
-        var url = await runner.StartServerAndGetUrl(runCommand, $"run --no-silent -c {config} --no-build --project {projectFile} --forward-console");
+        var url = await runner.StartServerAndGetUrlAsync(runCommand, $"run --no-silent -c {config} --no-build --project {projectFile} --forward-console");
         IBrowser browser = await runner.SpawnBrowserAsync(url);
         IBrowserContext context = await browser.NewContextAsync();
 
--- a/src/mono/wasm/Wasm.Build.Tests/Templates/InterpPgoTests.cs
+++ b/src/mono/wasm/Wasm.Build.Tests/Templates/InterpPgoTests.cs
@@ -1,0 +1,123 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.IO;
+using Wasm.Build.Tests;
+using Xunit;
+using Xunit.Abstractions;
+using System.Threading;
+using System.Threading.Tasks;
+using System;
+using System.Diagnostics;
+
+#nullable enable
+
+namespace Wasm.Build.Templates.Tests;
+
+public class InterpPgoTests : WasmTemplateTestBase
+{
+    public InterpPgoTests(ITestOutputHelper output, SharedBuildPerTestClassFixture buildContext)
+        : base(output, buildContext)
+    {
+    }
+
+    [Theory]
+    // Interpreter PGO is not meaningful to enable in debug builds - tiering is inactive there so all methods
+    //  would get added to the PGO table instead of just hot ones.
+    [InlineData("Release")]
+    public async Task FirstRunGeneratesTableAndSecondRunLoadsIt(string config)
+    {
+        // We need to invoke Greeting enough times to cause BCL code to tier so we can exercise interpreter PGO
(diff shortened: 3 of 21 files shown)


src/mono/mono/mini/interp/interpgo.c
```diff
@@ -0,0 +1,395 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
vargaz, line 1: Might be better to call this interp-pgo.c to be similar to the other files.

src/mono/mono/mini/interp/interpgo.h
```diff
@@ -0,0 +1,16 @@
+#ifndef __MONO_MINI_INTERPGO_H__
```
vargaz, line 1: Ditto.

src/mono/mono/mini/interp/interpgo.c
vargaz: Also, the function names should be interppgo or interp_pgo, not interpgo.

src/mono/mono/mini/interp/interpgo.c
```diff
@@ -0,0 +1,395 @@
...
+	uint32_t size, capacity;
+	mono_mutex_t mutex;
+} interpgo_table;
+
+// loaded_table is the table we loaded from persistent storage at startup (if any),
+//  while building_table is the table we built during the current run. we store these
+//  separately so that we don't have to maintain a sorted table (for bsearch) on an
+//  ongoing basis.
+static interpgo_table *loaded_table = NULL, *building_table = NULL;
+
```
vargaz, line 163: There is no need to NULL out static variables.

src/mono/mono/mini/interp/interpgo.c
```diff
@@ -0,0 +1,395 @@
...
+	return (result != NULL);
+}
+
+static void
+table_add (interpgo_table **table_variable, uint8_t hash[MM3_HASH_BYTE_SIZE]) {
+	interpgo_table *table = *table_variable, *temp;
+	// If we don't have a table yet, allocate one
+	if (!table) {
+		table = g_malloc0 (sizeof (interpgo_table));
+		// We may have lost a race against another thread that just allocated a table
```
vargaz, line 198: table->mutex is not initialized ?

src/mono/mono/mini/interp/interpgo.c
```diff
@@ -0,0 +1,395 @@
...
+	mono_qsort (table->data, table->size / MM3_HASH_BYTE_SIZE, MM3_HASH_BYTE_SIZE, hash_comparer);
+}
+
+static void
+compute_method_hash (MonoMethod *method, uint8_t outbuf[MM3_HASH_BYTE_SIZE]) {
+	// method token + image guid
+	size_t size = sizeof(uint32_t) + 16;
+	uint32_t *inbuf = alloca (size);
+	// method tokens are globally unique within a given assembly
+	inbuf[0] = mono_method_get_token (method);
```
vargaz, line 243: This is not true for generic instances.

src/mono/mono/mini/interp/interpgo.c
kg: i've never seen any code initialize our mutexes. size/capacity are initialized on-demand later guarded by the mutex

src/mono/mono/mini/interp/interpgo.c
kg: OK, so I need to hash the generic arguments? Or is it more complex than that?

src/mono/wasm/runtime/types/index.ts
```diff
@@ -93,6 +93,16 @@ export type MonoConfig = {
...
+    /**
+     * If true, a list of the methods optimized by the interpreter will be saved and used for faster startup
+     *  on future runs of the application
+     */
+    interpreterPgo?: boolean,
+    /**
+     * Configures how long to wait before saving the interpreter PGO list. If your application takes
+     *  a while to start you should adjust this value.
+     */
+    interpreterPgoSaveDelay?: number,
```
pavelsavara, line 105: please run `npm run rollup` or msbuild in debug mode to re-generate and update `dotnet.d.ts`. Don't do it by hands, please.

src/mono/wasm/test-main.js
```diff
@@ -290,9 +293,10 @@ function configureRuntime(dotnet, runArgs) {
         }
     }
     if (ENVIRONMENT_IS_WEB) {
-        if (runArgs.memorySnapshot) {
+        if (runArgs.memorySnapshot)
             dotnet.withStartupMemoryCache(true);
-        }
+        if (runArgs.interpreterPgo)
```
pavelsavara, line 298: similar to `memorySnapshot` the `interpreterPgo` needs multiple runs to take any effect.
For `memorySnapshot` this script would do  "dryRun" to create such run. But it would NOT execute any user code and will be stopped very early, because that's enough.

I think that means that in a unit test, `interpreterPgo` would never have chance to store the stats and subsequenty real run would not have it available.

I think we have following options
1) don't test interpreterPgo in unit tests, but find another way
2) make `dryRun` to run longer
3) make `dryRun` to run to store stats much earlier, together with memory snapshot (but there was not much managed code running at that point)
4) use some pre-loaded stats. Such stats would not match the exact hash of the binaries, tho

src/mono/wasm/runtime/types/index.ts
```diff
@@ -195,7 +205,7 @@ export interface AssetEntry {
     /**
      * the integrity hash of the asset (if any)
      */
-    hash?: string | null | ""; // 
+    hash?: string | null | ""; //
```
pavelsavara, line 208: please drop `//`

src/mono/wasm/runtime/types/index.ts
```diff
@@ -207,13 +217,13 @@ export interface AssetEntry {
     /**
      * If true, an attempt will be made to load the asset from each location in MonoConfig.remoteSources.
      */
-    loadRemote?: boolean, // 
+    loadRemote?: boolean, //
```
pavelsavara, line 220: please drop `//`

src/mono/wasm/runtime/interpgo.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (!cacheKey)
+        return;
+
+    await storeCacheEntry(cacheKey, memory, "application/octet-stream");
+
+    cleanupCache(tablePrefix, cacheKey); // no await
+}
+
+export async function interpgo_save_data () {
+    const cacheKey = await getCacheKey(tablePrefix);
```
pavelsavara, line 30: since this is scheduled 15 seconds later, you can use `loaderHelpers.is_runtime_running` or `loaderHelpers.is_exited` and skip storing it if it's not running. In particular, storing snapshot when something failed (during startup) is probably not good.

src/mono/wasm/runtime/cwraps.ts
```diff
@@ -148,9 +148,9 @@ const fn_signatures: SigLine[] = [
     [true, "mono_jiterp_tlqueue_clear", "void", ["number"]],
     [true, "mono_jiterp_begin_catch", "void", ["number"]],
     [true, "mono_jiterp_end_catch", "void", []],
-
+    [true, "mono_interpgo_load_table", "number", ["number", "number"]],
+    [true, "mono_interpgo_save_table", "number", ["number", "number"]],
     ...diagnostics_cwraps,
-    ...legacy_interop_cwraps
```
pavelsavara, line 153: `legacy_interop_cwraps` need to stay, right ?

src/mono/sample/wasm/browser-bench/main.js
```diff
@@ -203,6 +218,7 @@ try {
 
     const runtime = await dotnet
         .withRuntimeOptions(["--jiterpreter-stats-enabled"])
+        .withInterpreterPgo(true, 30)
```
pavelsavara, line 221: when benchmark is running in frame, we are starting many instances per second. They are not and I think they should not live 30 seconds. Effectively, not testing interpgo? 

In which case it would be better just disable it for the "appstart" scenarios.

pavelsavara: cc @radekdoulik

src/mono/mono/mini/interp/interpgo.c
pavelsavara: or skip generic instances ? Are they usually hot ?

kg: some generic instances are very hot in the BCL, like the ones used in string/array searches

src/mono/sample/wasm/browser-bench/main.js
kg: Interpgo works for successive browser-bench runs locally, you're correct that it doesn't work for appstart but it wouldn't anyway (I tested and nothing in appstart is hot enough to tier). So this is mostly here to allow people to test it locally and measure its impact on a suitably large piece of code.

src/mono/wasm/runtime/cwraps.ts
kg: Yes, good catch. I'm not sure what happened here.

src/mono/wasm/test-main.js
kg: I don't think it's possible to exercise this in traditional unit tests, but we could probably exercise it in a WBT test of some sort.

src/mono/mono/mini/interp/interpgo.c
vargaz: It can hash the full method name for example, but maybe its good enough this way.

lambdageek: If it's a wrapper method, I think the token is either shared with the method that it is wrapping, or it is zero.  maybe we don't mind the collisions?

lambdageek: Can we share the AOT compiler's logic for uniquely identifying methods?

src/mono/mono/mini/interp/interpgo.c
```diff
@@ -0,0 +1,395 @@
...
+	// Atomically swap the new table in
+	interpgo_table *old_table = mono_atomic_xchg_ptr ((volatile gpointer*)&loaded_table, result);
+
+	if (old_table) {
+		// If we had an old table before this, we need to lock it and then get rid of the buffer
+		mono_os_mutex_lock (&old_table->mutex);
+		if (old_table->data)
+			g_free (old_table->data);
+		old_table->data = NULL;
+		mono_os_mutex_unlock (&old_table->mutex);
```
lambdageek, line 327: Why not just have a global mutex and double-checked locking? Is there a particular reason why this has to be atomic_xchg?

src/mono/mono/mini/interp/interpgo.c
```diff
@@ -0,0 +1,395 @@
...
+	result->data = g_malloc0 (data_size);
+	g_assert ((int64_t)size < (int64_t)data_size);
+	result->size = size;
+	result->capacity = data_size;
+	memcpy (result->data, data + sizeof (uint32_t), result->size);
+
+	// Atomically swap the new table in
+	interpgo_table *old_table = mono_atomic_xchg_ptr ((volatile gpointer*)&loaded_table, result);
+
+	if (old_table) {
```
lambdageek, line 321: Under what circumstances do we expect the table to already be there? If another thread loads it?  Or do we get an updated table under some circumstances (maybe I missed where that happens)?

If it's a race with other threads, the usual lock-free approach would be a CAS with the expected value being NULL, and then dispose of the new table, not the old one.

src/mono/mono/mini/interp/interpgo.c
```diff
@@ -0,0 +1,395 @@
...
+	memcpy (data, &table->size, sizeof (uint32_t));
+	memcpy (data + sizeof (uint32_t), table->data, table->size);
+	mono_os_mutex_unlock (&table->mutex);
+	return 0;
+}
+
+#endif // HOST_BROWSER
+
+// These are outside the guarded region because MSVC fails compilation if any .c file
+//  contains no functions
```
lambdageek, line 364: we have a macro `MONO_EMPTY_SOURCE_FILE(interpgo)` to placate MSVC

src/mono/mono/mini/interp/interpgo.c
vargaz: The aot compiler/runtime has a mono_aot_method_hash () which returns a good hash.

src/mono/mono/mini/interp/interpgo.c
kg: I ended up at this incrementally, I think you're right that a global mutex is probably better.

src/mono/mono/mini/interp/interpgo.c
kg: Yeah in multithreaded scenarios it's possible that multiple threads could try to load a table. We could make attempts after the first fail if we wanted.

src/mono/mono/mini/interp/interpgo.c
```diff
@@ -0,0 +1,395 @@
...
+	return FALSE;
+}
+
+void
+mono_interpgo_method_was_tiered (MonoMethod *method) {
+	if (!mono_opt_interpgo_recording)
+		return;
+
+	if (method->wrapper_type != MONO_WRAPPER_NONE)
+		return;
```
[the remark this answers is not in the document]
kg: They're already unconditionally optimized, and it's hard to come up with a hash for them
image

src/mono/mono/mini/interp/interpgo.c
kg: mono_aot_method_hash looks comprehensive but also looks really expensive, and i'm not sure how well it will avoid collisions. do you still want me to use it?

src/mono/wasm/Wasm.Build.Tests/Templates/InterpPgoTests.cs
```diff
@@ -0,0 +1,124 @@
...
+            _testOutput.WriteLine("/// Starting server");
+
+            // Create a single browser instance and single context to host all our pages.
+            // If we don't do this, each page will have its own unique cache and the table won't be loaded.
+            using var runCommand = new RunCommand(s_buildEnv, _testOutput)
+                                        .WithWorkingDirectory(_projectDir!);
+            await using var runner = new BrowserRunner(_testOutput);
+            var url = await runner.StartServerAndGetUrl(runCommand, $"run --no-silent -c {config} --no-build --project {projectFile} --forward-console");
+            var browser = await runner.SpawnBrowserAsync(url);
+            var context = await browser.NewContextAsync();
```
radical, line 77: nit: Suggest using specific types because these aren't common to WBT, and seeing the types clearly would help with reading them.
```suggestion
            IBrowser browser = await runner.SpawnBrowserAsync(url);
            IBrowserContext context = await browser.NewContextAsync();
```

src/mono/wasm/Wasm.Build.Tests/WasmTemplateTestBase.cs
```diff
@@ -132,4 +132,16 @@ private static void UpdateRuntimeconfigTemplateForNode(string projectDir)
         else
             projectProvider.AssertBundle(buildArgs, buildProjectOptions);
     }
+
+    public const string DefaultRuntimeAssetsRelativePath = "./_framework/";
+
+    public void UpdateBrowserMainJs(Func<string, string> transform, string targetFramework, string runtimeAssetsRelativePath = DefaultRuntimeAssetsRelativePath)
```
radical, line 138: nit:
```suggestion
    protected const string DefaultRuntimeAssetsRelativePath = "./_framework/";

    protected void UpdateBrowserMainJs(Func<string, string> transform, string targetFramework, string runtimeAssetsRelativePath = DefaultRuntimeAssetsRelativePath)
```

radical (approved):
WBT changes look good.

src/mono/wasm/runtime/startup.ts
```diff
@@ -557,9 +562,25 @@ async function mono_wasm_before_memory_snapshot() {
...
+
     endMeasure(mark, MeasuredBlock.memorySnapshot);
 }
 
+async function maybeSaveInterpPgoTable () {
+    // If the application exited abnormally, don't save the table. It probably doesn't contain useful data,
+    //  and saving would overwrite any existing table from a previous successful run.
+    // We treat exiting with a code of 0 as equivalent to if the app is still running - it's perfectly fine
+    //  to save the table once main has returned, since the table can still make future runs faster.
+    if ((loaderHelpers.exitCode !== undefined) && (loaderHelpers.exitCode !== 0))
```
pavelsavara, line 578: web apps never exit

kg: I don't understand, wasn't this your suggestion?

src/mono/wasm/runtime/startup.ts
```diff
@@ -557,9 +562,25 @@ async function mono_wasm_before_memory_snapshot() {
         runtimeHelpers.storeMemorySnapshotPending = false;
     }
 
+    if (runtimeHelpers.config.interpreterPgo) {
+        if (typeof (globalThis.setTimeout) === "function")
```
pavelsavara, line 566: this is overly defensive, I think

src/mono/sample/wasm/browser-bench/main.js
pavelsavara: could you please add comment explaining that ?

src/mono/wasm/runtime/startup.ts
pavelsavara: my bad, I misread the logic. 
It means take snapshot if it's still running or if it exited with zero, right ?

src/mono/wasm/Wasm.Build.Tests/Templates/InterpPgoTests.cs
```diff
@@ -0,0 +1,124 @@
...
+                // Verify that no PGO table was located in cache
+                Assert.Contains("Failed to load interp_pgo table", output);
+                // Verify that the table was saved after the app ran
+                Assert.Contains("Saved interp_pgo table", output);
+                // Verify that a specific method was tiered by the Greeting calls and recorded by PGO
+                Assert.Contains("added System.Runtime.CompilerServices.Unsafe:Add<byte> (byte&,int) to table", output);
+            }
+
+            {
+                _testOutput.WriteLine("/// Second run");
```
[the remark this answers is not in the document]
kg: Each page is basically a separate tab as I understand it

src/mono/wasm/runtime/startup.ts
kg: Correct, we don't want to save a PGO table for a run that asserted or failed

pavelsavara (approved):

src/mono/mono/mini/interp/transform.c
```diff
@@ -4815,6 +4816,8 @@ generate_code (TransformData *td, MonoMethod *method, MonoMethodHeader *header,
 	gboolean inlining = td->method != method;
 	InterpBasicBlock *exit_bb = NULL;
 
+	mono_interp_pgo_generate_start ();
```
BrzVlad, line 4819: Why aren't these methods in `generate` directly ?

kg: They were originally browser-only until I realized I could make them all portable. If you want I can move those into transform.c

BrzVlad: I meant that `generate_code` is called recursively for every method that is inlined in the current method compilation. And it seems to me that you want to call this method only once per method compilation, in `generate` method.

kg: I misunderstood, great point. Would that mean I could get rid of the recursion handling too?

BrzVlad: yes

BrzVlad (approved):

PR #92982, opened 2023-10-04
Consolidate some TensorPrimitivesTests logic around special values

stephentoub:


src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
```diff
@@ -57,6 +57,66 @@ private static unsafe float MathFMinMagnitude(float x, float y)
...
+
+        private static unsafe float Int32ToSingle(int i) => *(float*)&i;
+
+        private static float AnotherSingleNaN = Int32ToSingle(-8388607);
+
+        /// <summary>Loads a variety of special values (e.g. NaN) into random positions in <paramref name="x"/>.</summary>
+        private static void SetSpecialValues(Span<float> x)
+        {
+            // NaN
+            x[s_random.Next(x.Length)] = float.NaN;
```
tannergooding, line 71: There's at least 5 different interesting `NaN` that would be good to cover here...

In particular, the 4 `NaN` kinds and the inclusive range of their bit patterns are:
* `-qNaN` - `[0xFFC0_0000, 0xFFFF_FFFF]`
* `+qNaN` - `[0x7FC0_0000, 0x7FFF_FFFF]`
* `-sNaN` - `[0xFF00_0001, 0xFFBF_FFFF]`
* `+sNaN` - `[0x7F00_0001, 0x7FBF_FFFF]`

Where for historical reasons, the canonical forms of each of these have historically been:
`-qNaN` - `0xFFC0_0000`
`+qNaN - `0x7FC0_0000`
`-sNaN` - `0xFFA0_0000`
`+sNaN - `0x7FA0_0000`

-- This is because for `qNaN`, it's the bit that distinguishes it as quiet and leaves the payload empty. While for `sNaN`, it's using the most significant payload bit, since at least one payload bit needs to be set to distinguish it from infinity.

The 5th explicitly interesting value is `AllBitsSet`, which is `0xFFFF_FFFF` (a `-qNaN`). We could additionally test random payload bit patterns, but they are less interesting.

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
```diff
@@ -57,6 +57,66 @@ private static unsafe float MathFMinMagnitude(float x, float y)
...
+            x[s_random.Next(x.Length)] = float.PositiveInfinity;
+            x[s_random.Next(x.Length)] = float.NegativeInfinity;
+
+            // +Zero, -Zero
+            x[s_random.Next(x.Length)] = +0.0f;
+            x[s_random.Next(x.Length)] = -0.0f;
+
+            // +Epsilon, -Epsilon
+            x[s_random.Next(x.Length)] = +float.Epsilon;
+            x[s_random.Next(x.Length)] = -float.Epsilon;
```
tannergooding, line 84: `Epsilon` is the smallest `subnormal`, (bit pattern `0x0000_0001` and `0x8000_0001`).

It's also interesting to cover the:
* largest `subnormal` (bit pattern `0x007F_FFFF` and `0x807F_FFFF`)
* smallest `normal` (bit pattern `0x0080_0000` and `0x8080_0000`)
* largest `normal` (bit pattern `0x7F7F_FFFF` and `0x8F7F_FFFF` --- `MaxValue` and `MinValue`)

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
tannergooding, line 71: Do we have any concern around `s_random.Next()` overwriting a previous random?

tannergooding (approved):

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
stephentoub: My thinking is that we're relying on this being used enough that we'll have good enough coverage of everything.  Alternatively, we could test out all combinations.  I've been thinking we start with the former and graduate to the latter if/when we want the 100% coverage.

stephentoub:
Thanks.  Good feedback. I'm going to merge this to submit a PR that depends on it, and I'll address adding the additional values in that one.

PR #92985, opened 2023-10-04
Make thread pool thread timeouts configurable

kouvel:
- Added two config options, one that configures the worker and wait thread timeouts, and another that enables keeping some number of worker threads alive after they are created
- This enables services that take periodic traffic to keep some worker threads around for better latency, while allowing extra threads to time out as appropriate for the service

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Threading/PortableThreadPool.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Threading/PortableThreadPool.cs
@@ -13,7 +13,6 @@ namespace System.Threading
     /// </summary>
     internal sealed partial class PortableThreadPool
     {
-        private const int ThreadPoolThreadTimeoutMs = 20 * 1000; // If you change this make sure to change the timeout times in the tests.
         private const int SmallStackSizeBytes = 256 * 1024;
 
         private const short MaxPossibleThreadCount = short.MaxValue;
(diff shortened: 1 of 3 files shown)


src/libraries/System.Private.CoreLib/src/System/Threading/PortableThreadPool.cs
```diff
@@ -40,6 +39,23 @@ internal sealed partial class PortableThreadPool
...
+
+            // The amount of time in milliseconds a thread pool thread waits without having done any work before timing out and
+            // exiting. Set to -1 to disable the timeout. Applies to worker threads and wait threads. Also see the
+            // ThreadsToKeepAlive config value for relevant information.
+            int timeoutMs =
+                AppContextConfigHelper.GetInt32Config(
+                    "System.Threading.ThreadPool.ThreadTimeoutMs",
+                    "DOTNET_ThreadPool_ThreadTimeoutMs",
+                    DefaultThreadPoolThreadTimeoutMs);
+            return timeoutMs >= -1 ? timeoutMs : DefaultThreadPoolThreadTimeoutMs;
```
mangod9, line 56: is timeout of 0 valid?

kouvel: It's valid in that a thread would wait on the semaphore with a zero-timeout and likely exit if the semaphore is not signaled. It's probably not a good setting to use though, including any very low timeout values, as they could lead to threads exiting very quickly, or a fairly busy loop in the rare case that IO is pending on the thread.

eduardo-vp (approved):

PR #92987, opened 2023-10-04
[7.0] Make thread pool thread timeouts configurable

kouvel:
- Port of https://github.com/dotnet/runtime/pull/92985 to 7.0 with some changes to accommodate differences in the thread pool implementations. This version covers both the native and managed thread pools, including the native IOCP thread pool.
- Added two config options, one that configures the worker and wait thread timeouts, and another that enables keeping some number of worker threads alive after they are created
- This enables services that take periodic traffic to keep some worker threads around for better latency, while allowing extra threads to time out as appropriate for the service

## Customer Impact

These config capabilities were requested by a 1p customer for a scenario where a service takes work in waves and is experiencing noticeably high latency with thread pool worker and IOCP threads being torn down and recreated between waves. The config vars enable the service to keep some number of threads always alive, and to increase the timeout for any extra threads that are created.

## Regression?

No

## Testing

Verified the default behavior hasn't changed when not configured, and verified using events that the config values are working as expected.

## Risk

Low. The change is a bit larger in .NET 7 and .NET 6 because it also covers the native worker and IOCP thread pools. The mechanism for configuring is slightly different to allow both the native and managed thread pools to use the same config vars, though similar to before / other config vars.


mangod9 (approved):

jeffschwMSFT (approved):
approved. we will take for consideration in 7.0.x

PR #86170, opened 2023-05-12
Use /INCREMENTAL:NO for Debug

kunalspathak:
Stop doing `/INCREMENTAL:YES` 

Fixes https://github.com/dotnet/runtime/issues/64082.

PR #92993, opened 2023-10-04
JIT: Export debug helpers

jakobbotsch:
A few weeks ago these helper stopped working for me.  This exports them in debug builds and makes them work again (as a side bonus, it also means they now work in checked builds).

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -9201,6 +9201,14 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 */
 
+#ifndef DBGFUNC
+#ifdef _MSC_VER
+#define DBGFUNC __declspec(dllexport)
+#else
+#define DBGFUNC
+#endif
+#endif
+
 /*****************************************************************************/
 /* The following functions are intended to be called from the debugger, to dump
  * various data structures.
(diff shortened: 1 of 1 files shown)


EgorBo (approved):

jakobbotsch:
> Did you investigate why this stopped working?

I haven't, no... that might be a good idea. Presumably it's either due to a VS toolset update or because we changed some compiler options.

src/coreclr/jit/compiler.cpp
```diff
@@ -9201,6 +9201,14 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 */
 
+#ifndef DBGFUNC
+#ifdef _MSC_VER
+#define DBGFUNC __declspec(dllexport)
```
BruceForstall, line 9206: We don't use `__declspec(dllexport)` elsewhere; we use the ClrJit.exports and ClrJit.PAL.exports files.

I notice that some (but not all?) such functions are marked  `extern "C"`. Also seems weird that `DllMain` isn't in the ClrJit.exports file but maybe it's special?

BruceForstall:
> A few weeks ago these helper stopped working for me.

In what way are they not working?

Do the symbols (and functions) just not exist in the image? The debugger should be able to find them even if they are not exported. Are the symbols being C++ name mangled and the debugger doesn't understand that?

BruceForstall:
fwiw, I have a local runtime build based on 9/28 sources (1cf79260b6f0684defb8a4a3aaf27c1806b55bbf) running in VS 2022 Version 17.8.0 Preview 2.0 and I can invoke these within the VS QuickWatch window (where I normally use them) fine (and they are not DLL exports).

jakobbotsch:
> In what way are they not working?
> 
> Do the symbols (and functions) just not exist in the image? The debugger should be able to find them even if they are not exported. Are the symbols being C++ name mangled and the debugger doesn't understand that?
> 
> fwiw, I have a local runtime build based on 9/28 sources (https://github.com/dotnet/runtime/commit/1cf79260b6f0684defb8a4a3aaf27c1806b55bbf) running in VS 2022 Version 17.8.0 Preview 2.0 and I can invoke these within the VS QuickWatch window (where I normally use them) fine (and they are not DLL exports).

The symbols work fine after an initial build through build.cmd, but when I rebuild the clrjit project from Visual Studio they stop working:

image
(I rebuilt clrjit and relaunched a debugging session between these two.)

jakobbotsch:
It seems like the functions are not present at all in the clrjit.dll built through Visual Studio. Let me try a few months old commit to see if this is a change made by us or something in VS.

src/coreclr/jit/compiler.cpp
jakobbotsch: I checked and MSVC just combines the exports, so both these functions and the ones from the file end up exported. Since these are only exported in debug builds I think it's better to keep it out of the file.

We don't need `extern "C"` on these, the debugger has no problem with demangling them.

jakobbotsch:
I have the same problem even if I go back to 2b11093 that is 5 months old, so seems like something has changed around VS here.

jakobbotsch:
Hmm, looks like /OPT:REF is enabled in the VS project options:

image

If I flip this flag then they work properly. It looks like we add that linker flag unconditionally for debug here:

https://github.com/dotnet/runtime/blob/ffa4ef75ec135a64a30e14df5c2afca340ee1203/eng/native/configurecompiler.cmake#L121

but we have been adding that for a while now (since #86170), and this only broke for me recently. Regardless, let me just fix this then.

jakobbotsch:
Superseded by #93053

BruceForstall:
ok, I implemented the above: https://github.com/dotnet/runtime/pull/93086

PR #92994, opened 2023-10-04
Try rooting less stuff for reflectable method signatures

MichalStrehovsky:
I don't think we need a working `MethodTable` for everything in the method signature.

@Sergio0694 also noticed `IUnknown**` roots `IUnknown`. This doesn't address that (just reference types), but maybe non-ref pointers could be workable in a similar way (we mostly go for IntPtr?).

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/DynamicInvokeInfo.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/DynamicInvokeInfo.cs
@@ -72,20 +72,21 @@ namespace System.Reflection
                 {
                     Transform transform = default;
 
-                    Type argumentType = parameters[i].ParameterType;
+                    var argumentType = (RuntimeType)parameters[i].ParameterType;
                     if (argumentType.IsByRef)
                     {
                         _needsCopyBack = true;
                         transform |= Transform.ByRef;
-                        argumentType = argumentType.GetElementType()!;
+                        argumentType = (RuntimeType)argumentType.GetElementType()!;
                     }
                     Debug.Assert(!argumentType.IsByRef);
 
-                    EETypePtr eeArgumentType = argumentType.TypeHandle.ToEETypePtr();
-
-                    if (eeArgumentType.IsValueType)
+                    // This can return a null MethodTable for reference types.
+                    // The compiler makes sure it returns a non-null MT for everything else.
+                    EETypePtr eeArgumentType = argumentType.ToEETypePtrMayBeNull();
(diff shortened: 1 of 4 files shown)


MichalStrehovsky:
> @Sergio0694 also noticed `IUnknown**` roots `IUnknown`. This doesn't address that (just reference types), but maybe non-ref pointers could be workable in a similar way (we mostly go for IntPtr?).

So the pointers are a lot trickier and I'd need to add even more special casing for them on the classlib side. It would probably not bring any meaningful savings for normal .NET apps, but bring a lot of risk that we regress something (we'd need to duplicate a bunch of intricate rules such as "method that takes a pointer to SomeEnum can also be invoked by `Pointer.Box`ing `int*`", etc.).

I'm already a bit uncomfortable with this change that limits it to reference types only, but it does save 0.1% on ASP.NET scenarios. We'd get triple that if we could also do this for value types, but it doesn't look likely we can. Pointers would be noise-level savings.

@jkotas no hard feelings if you say doing this for reference types is not worth it either. It does add a bit of mental overhead to `DynamicInvokeInfo` and 0.1% might not be worth the trouble. I was hoping for more. It could be more for _some_ apps.

Cc @dotnet/ilc-contrib

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/DynamicInvokeInfo.cs
```diff
@@ -636,6 +638,14 @@ private void ThrowForArgCountMismatch()
...
 
+                    // Destination type can be null if we don't have a MethodTable for this type. This means one cannot
+                    // possibly pass a valid non-null object instance here.
+                    if (dstEEType.IsNull)
+                    {
+                        Debug.Assert(i != 0 || _isStatic);
+                        throw InvokeUtils.CreateChangeTypeArgumentException(srcEEType, Method.GetParametersAsSpan()[i - (_isStatic ? 0 : 1)].ParameterType, destinationIsByRef: false);
+                    }
+
                     if (!(srcEEType.RawValue == dstEEType.RawValue ||
```
jkotas, line 649: I would move the null check to be after the exact type match. Something like:

if (srcEEType.RawValue != dstEEType.RawValue)
{
    if (dstEEType.IsNull)
    {
        ...
    }

    if (!(RuntimeImports.AreTypesAssignable(srcEEType, dstEEType) ...

}

jkotas:
> It does add a bit of mental overhead to DynamicInvokeInfo

I do not think that this mental overhead is a big deal. I see this as small change giving us small improvement.

jkotas (approved):

PR #92997, opened 2023-10-04
SPMI: Fix superpmi.py result assert

jakobbotsch:
The string is actually "Error", not "Failure"


jakobbotsch:
cc @dotnet/jit-contrib, simple change

amanasifkhalid (approved):

PR #92998, opened 2023-10-04
JIT: Add ordering side effects in fgMorphExpandInstanceField

### LINKED ISSUE #92990: jitstress failing widely with NRE
There are many NRE failures in the latest jitstress/libraries-jitstress runs.

jakobbotsch:
In #92710 I removed an ordering side effect set on a LCL_VAR node in fgMorphExpandInstanceField, since the side effect is meaningless on a local. However, I did not realize that setting the ordering effect was important because propagating it to parent nodes was actually necessary.

This PR explicitly sets the ordering effect on the ADD<byref> nodes created, since forming these and reporting them to GC is illegal if the base is invalid.

Additionally, it sets the ordering effect on the consuming indirection; if that indirection made use of the fact that the FIELD_ADDR is non-null to become non-faulting, then it has an ordering dependency once we expand the FIELD_ADDR to a control-flow based null-check. This is not actually necessary since we always set end up with an ADD<byref> that propagates it, but that seems more like a coincidence than anything else.

Fix #92990

### DIFF
--- a/src/coreclr/jit/compiler.h
+++ b/src/coreclr/jit/compiler.h
@@ -5917,8 +5917,8 @@ private:
     // small; hence the other fields of MorphAddrContext.
     struct MorphAddrContext
     {
-        size_t m_totalOffset = 0;     // Sum of offsets between the top-level indirection and here (current context).
-        bool   m_used        = false; // Whether this context was used to elide a null check.
+        GenTreeIndir* Consumer = nullptr; // Indirection using this address.
+        size_t        TotalOffset = 0;    // Sum of offsets between the top-level indirection and here (current context).
     };
 
 #ifdef FEATURE_SIMD
--- a/src/coreclr/jit/compiler.h
+++ b/src/coreclr/jit/compiler.h
@@ -5917,8 +5917,8 @@ private:
     // small; hence the other fields of MorphAddrContext.
     struct MorphAddrContext
     {
-        GenTreeIndir* Consumer = nullptr; // Indirection using this address.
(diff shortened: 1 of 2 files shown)


jakobbotsch:

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=427080&view=ms.vss-build-web.run-extensions-tab)

src/coreclr/jit/compiler.h
```diff
@@ -5917,8 +5917,8 @@ class Compiler
     // small; hence the other fields of MorphAddrContext.
     struct MorphAddrContext
     {
-        size_t m_totalOffset = 0;     // Sum of offsets between the top-level indirection and here (current context).
-        bool   m_used        = false; // Whether this context was used to elide a null check.
+        GenTreeIndir* Consumer = nullptr; // Indirection using this address.
+        size_t        TotalOffset = 0;    // Sum of offsets between the top-level indirection and here (current context).
```
BruceForstall, line 5921: nit: why change the naming style? We aren't very consistent with https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/clr-jit-coding-conventions.md#8.7, but we are pretty consistent to not capitalize local/field names.

BruceForstall (approved):
LGTM with one nit

src/coreclr/jit/compiler.h
jakobbotsch: I reverted this... mainly I think the `m_` prefix on fields meant to be used publicly looks off (I like the convention, but only for private fields, and I thought that's only where it applied).

PR #93000, opened 2023-10-04
[browser] streaming compile wasm bit earlier

pavelsavara:
This PR moves start of streaming compilation of WASM module into `dotnet.js` loader.
`dotnet.native.wasm` file could start to be compiled even before download and compilation of `dotnet.runtime.js` and `dotnet.native.js` is done.

This brings bit more of the logic (and moves few bytes of download size) into the loader.

Let's measure if this is good idea or not ...


pavelsavara:
WBT failures are unrelated

radekdoulik (approved):

PR #93002, opened 2023-10-04
[mono][jit] Internal memcpy uses SIMD

jandupej:
Copying large enough structs byval is now done in SIMD if alignment allows or CPU does not mind unaligned access.

### DIFF
--- a/src/mono/mono/mini/memory-access.c
+++ b/src/mono/mono/mini/memory-access.c
@@ -162,7 +162,7 @@ MONO_DISABLE_WARNING(4127) /* conditional expression is constant */
 	}
 
 copy_16:
-#if defined(TARGET_ARM64) && !defined(DISABLE_SIMD)
+#if (defined(TARGET_ARM64) || defined(TARGET_AMD64)) && !defined(DISABLE_SIMD)
 	while (size >= 16) {
 		cur_reg = alloc_xreg (cfg);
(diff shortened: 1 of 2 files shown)


src/mono/mono/mini/memory-access.c
```diff
@@ -130,6 +130,12 @@ mini_emit_memcpy (MonoCompile *cfg, int destreg, int doffset, int srcreg, int so
 	}
 
 	g_assert (align > 0);
+	
+#ifndef NO_UNALIGNED_ACCESS	
+	// Unaligned access is not a problem on this platform. Use the longest
+	// possible vector.
+	goto copy_16;
```
vargaz, line 137: Note that the cpu allowing unaligned access does not mean that its going to be fast.

jandupej: What I'm trying to do is optimize memcpying types like `Matrix4x4` which have proper alignment, but do not register as SIMD types. Other than checking class name against a known list, do you see a way of detecting those?

vargaz: The function has an align argument which is supposed to be the guaranteed alignment of src/dest.

jandupej:
It appears that allowing unaligned writes could break ordering if the struct sits squarely on a cache line boundary, which we cannot prevent. Closing.

PR #93003, opened 2023-10-04
JIT: Fix invalid smnegl transformation

### LINKED ISSUE #92537: JIT: Invalid smnegl transformation

jakobbotsch:
The newly introduced TryLowerNegToMulLongOp would always lower to a signed multiplication, even for unsigned multiplications.

Also apply a couple of other fixes to the transformation:
* Replace uses of GenTree::ReplaceWith with LIR::Use::ReplaceWith
* Add some necessary interference checks

Fix #92537


kunalspathak:
> The newly introduced TryLowerNegToMulLongOp would always lower to a signed multiplication, even for unsigned multiplications.

I pointed that out in the PR https://github.com/dotnet/runtime/pull/91886#issuecomment-1715629341 but seems like we still missed some cases.

kunalspathak:
Interestingly there are no diffs, so seems we don't have coverage for this scenario. Can you please add a test case too?

jakobbotsch:
> Interestingly there are no diffs, so seems we don't have coverage for this scenario. Can you please add a test case too?

Added the test, can you PTAL?

kunalspathak (approved):
LGTM

kunalspathak:
superpmi failures are known.

PR #93010, opened 2023-10-04
[browser] streamline Task/Promise marshalling

pavelsavara:
This streamlines passing Promise/Task from two cross-boundary calls per promise argument to one.
- none when it's passed
- one when it's resolved, rejected or garbage collected

## Changes
- drop `create_task_callback`
- simplify `complete_task`
- simplify `mono_wasm_marshal_promise` and rename it to `mono_wasm_resolve_or_reject_promise`
- introduced `GCVHandle`
    - it's like `GCHandle` in that it keeps proxy of C# object alive while JS owns it.
    - but it's allocated ahead of time on JS side
- introduced `JSVHandle`
    - it's like `JSHandle` in that it keeps proxy of JS object alive while C# owns it.
    - but it's allocated ahead of time on C# side
- improved `GetTaskResultDynamic` to handle `void` Task
- introduced internal `JSObject.DisposeLocal`
    - we can dispose both sides during resolve/reject call without extra dispose call

## Interop "stack frame"
- fixed `IntPtr` and handles to be `I32` rather than `U32`
- changed `MarshalerType` from `I32` to `U8` size
- added `MarshalerType.TaskResolved` and `MarshalerType.TaskRejected`
    - this allows us to pass better metadata in the same `JSMarshalerArgument` slot
- moved `JSMarshalerArgument.ElementType` to different offset
    - to not clash with handles
- `JSMarshalerArgument.ElementType` now contains type of the promise's result
    - this is necessary for marshaling with dynamic type
- changed `JSBindingType` layout accordingly. 
    - In future this could also help with nested generic types.
- bumped schema version to `2`

## Other
- better type for `WeakRefInternal<T>`
- interop unit test no longer `ForceDisposeProxies` in the middle of the program because it can't balance C# side of handles


src/mono/wasm/runtime/loader/logging.ts
```diff
@@ -94,15 +94,17 @@ export function setup_proxy_console(id: string, console: Console, origin: string
     const consoleUrl = `${origin}/console`.replace("https://", "wss://").replace("http://", "ws://");
 
     consoleWebSocket = new WebSocket(consoleUrl);
-    consoleWebSocket.addEventListener("open", () => {
+    consoleWebSocket.onopen = () => {
```
maraf, line 97: What is the reason for this change? `addEventListener` is generally preferred.

maraf (approved):
Looks great! 👍

src/mono/wasm/runtime/loader/logging.ts
pavelsavara: during mono_exit we want to un-install the handler, but we don't have the original. This helps. Is there better way ?

maraf: Let's not use lambdas if need to remove them

pavelsavara: done

radical:
Re-triggered the builds to pick up the build fixes from https://github.com/dotnet/runtime/pull/93412 .

pavelsavara:
I moved logging changes to https://github.com/dotnet/runtime/pull/93472

radical:
Please avoid merging on red when the failure is not known, and especially when the failures are in wasm jobs. If merging any way, then please add an explanation for that.
https://github.com/dotnet/runtime/issues/93583

PR #93012, opened 2023-10-04
Fix conflicting test assembly name in classloader tests

akoeplinger:
dd52 was building test.dll and dev10_897464 was building Test.dll which caused assembly resolving conflicts:

> warning MSB3243: No way to resolve conflict between "Test, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null" and "test, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null". Choosing "Test, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null" arbitrarily.

Fixed by making the test assembly names unique for each test.


trylek (approved):
Sorry about the overlooking. Thanks for fixing!

PR #93018, opened 2023-10-04
Vectorize TensorPrimitives.Exp

tannergooding:


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
```diff
@@ -923,6 +923,19 @@ public Vector<float> Invoke(Vector<float> x, Vector<float> y)
...
 
+        private readonly struct ExpOperator : IUnaryOperator
+        {
+            public bool CanVectorize => false;
+
+            public float Invoke(float x) => MathF.Exp(x);
+
+            public Vector<float> Invoke(Vector<float> x)
+            {
+                // Vectorizing requires shift right support, which is .NET 7 or later
```
stephentoub, line 934: I don't see "ShiftRight" or ">>" in the implementation above. Did I miss it?

stephentoub (approved):

PR #93021, opened 2023-10-04
System.Formats.Cbor package's readme

illialarka:
Adding PACKAGE.md for System.Formats.Cbor package.

Progress towards #92228.

### DIFF
--- a/src/libraries/System.Formats.Cbor/src/PACKAGE.md
+++ b/src/libraries/System.Formats.Cbor/src/PACKAGE.md
@@ -30,8 +30,7 @@ The main types provided by this library are:
 
 <!-- Links to further documentation. Remove conceptual documentation if not available for the library. -->
 
-* [Conceptual documentation](https://learn.microsoft.com/en-us/dotnet/standard/serialization/**LIBRARYNAME**/overview)
-* [API documentation](https://learn.microsoft.com/en-us/dotnet/api/**LIBRARYNAME**)
+* [API documentation](https://learn.microsoft.com/en-us/dotnet/api/system.formats.cbor)
 
 ## Related Packages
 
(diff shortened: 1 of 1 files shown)


src/libraries/System.Formats.Cbor/src/PACKAGE.md
```diff
@@ -2,41 +2,121 @@
...
 
 <!-- The related packages associated with this package -->
 
+.NETFramework 4.6.2:
+
+* Provides HashCode types: [Microsoft.Bcl.HashCode](https://www.nuget.org/packages/Microsoft.Bcl.HashCode/)
+* Resource pooling: [System.Buffers](https://www.nuget.org/packages/System.Buffers/)
+* Efficient memory representation: [System.Memory](https://www.nuget.org/packages/System.Memory/)
+* Provides functionality over pointers: [System.Runtime.CompilerServices.Unsafe](https://www.nuget.org/packages/System.Runtime.CompilerServices.Unsafe/)
+* Provides underlying type for tuples: [System.ValueTuple](https://www.nuget.org/packages/System.ValueTuple/)
```
bartonjs, line 102: What makes these "related"?  Based on the fact that it's using TFMs here suggests it's "dependencies", but that's already data that gets surfaced, so why would it be in the description?

I'd, personally, expect this to be more for "this is the primitive layer, but if you want a friendly, opinionated view of a thing that's easier to work with, see [something else]".  I don't think there's anything like that with CBOR, so I'd, personally, say there are no "related" packages.

illialarka: I was looking at this section more like "dependencies" packages for some reason. I am aware that user can see dependencies clicking on "Dependencies" tab.

Overall, you are right! Perhaps, I have misunderstood the purpose of this section. Are you okay if I remove the section completely? I can not list any related packages.

krwq (approved):
LGTM with suggestion

ViktorHofer (approved):

Issue #93023, opened 2023-10-04
[wasm] WBT tests failing when to compile `aot-instances.dll.bc` when using `-O0`


radical:
## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=427185
Build error leg or test failing: Wasm.Build.Tests.SatelliteAssembliesTests.CheckThatSatelliteAssembliesAreNotAOTed(buildArgs: BuildArgs { ProjectName = placeholder, Config = Release, AOT = True, ProjectFileContents = placeholder, ExtraBuildArgs =  }, id: \"kbid5vmh_0lm\")
Pull request: https://github.com/dotnet/runtime/pull/93001

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json
{
  "ErrorMessage": "",
  "ErrorPattern": "aot-instances.dll.bc.*returned 137",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=427185
**Error message validated:** `aot-instances.dll.bc.*returned 137`
**Result validation: :white_check_mark:** Known issue matched with the provided build.
**Validation performed at:** 10/4/2023 7:07:10 PM UTC

radical:
Fixed by https://github.com/dotnet/runtime/pull/93026 .

PR #93026, opened 2023-10-04
[wasm] WBT: Avoid using `-O0` in tests

radical:
Some WBT tests started failing while compiling `aot-instances.dll.bc`
when compiled with `-O0`. So, we avoid using that where possible.

Fixes issue: https://github.com/dotnet/runtime/issues/93023


ilonatommy (approved):
Is this change temporary till we fix the cause?

radical:
> Is this change temporary till we fix the cause?

No. We want to avoid having `-O0` in the tests because it can end up hitting memory limits any away, especially with AOT, and dedup.

Issue #93020, opened 2023-10-04
JIT: Flow Graph Modernization and Improved Block Layout


AndyAyersMS:
### Overview

The current block layout algorithm in the JIT is based on local permutations of block order. It is complicated and likely far from optimal. We would like to improve the overall block layout algorithm used by the JIT, in particular adopting a global cost-minimizing approach to layout&mdash;for instance, one in the style of Young et. al.'s [Near-optimal intraprocedural branch alignment](https://dl.acm.org/doi/abs/10.1145/258915.258932). Additional complexities arise in our case because of various EH reporting requirements, so the JIT cannot freely reorder all blocks, but we should be able to apply the global ordering techniques within EH regions.

Before we can tackle this problem there are several important (and sizeable) prerequisites, which we can lump together as "flow graph modernization." There are a lot of details here, but at a high level:
* Update the flow graph to make fall-through behavior explicit for most phases (until layout). This means disallowing BBJ_NONE and adding explicit fall through successors for BBJ_COND and BBJ_SWITCH. More on this below.
* Defer most block reordering work until late in the JIT phase pipeline (ideally, perhaps, waiting until after LSRA has run, so we can properly position any new blocks it introduces). 
* Leverage the work done in .NET 8 to make flow edges persistent and ensure that those edges accurately describe successor likelihoods. Possibly make successor edge enumeration a first-class concept.

It is not yet clear how much progress we can make during .NET 9. The list of items below is preliminary and subject to change.

#### Motivation

Past studies have shown that the two phases that benefit most from block-level PGO data are inlining and block layout. In a previous compiler project, the net benefit from PGO was on the order of 15%, with about 12% attributable to inlining, and 2% to improved layout.

The current JIT is likely seeing a much smaller benefit from layout. The goal here is to ensure that we are using the accurate PGO data to make informed decisions about the ordering of blocks, with the hope of realizing perhaps a 1 or 2% net benefit across a wide range of applications (with some benefiting much more, and others, not at all).

#### Flow Graph Modernization

- [ ] No implicit fall through (until layout)
  - [x] Hide various fields on `BasicBlock` behind setters/getters (`bbNext`, `bbJumpKind`, `bbJumpTarget`, ...)
    - https://github.com/dotnet/runtime/pull/92908
    - https://github.com/dotnet/runtime/pull/93032
    - https://github.com/dotnet/runtime/pull/93152
    - https://github.com/dotnet/runtime/pull/93415
[...]

PR #93032, opened 2023-10-04
JIT: Make BasicBlock::bbPrev and bbNext private

amanasifkhalid:
Followup to #92908, and next step for #93020. CC @dotnet/jit-contrib.


BruceForstall:
[No asm diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=427804&view=ms.vss-build-web.run-extensions-tab) but some TP diffs.

BruceForstall:
Seems like we should work to avoid too many `GetBBNext()` calls in the code (maybe as a follow-up?)

E.g., instead of `(block->bbNext == nullptr)` => a new function `block->IsLast()`

`block->GetBBNext() == compiler->fgFirstColdBlock` => `block->IsLastHotBlock(compiler)` (true also for no cold section and `block->IsEnd()`?) Similarly, `block->IsFirstColdBlock(compiler)`?

src/coreclr/jit/block.h
```diff
@@ -508,10 +508,49 @@ struct BasicBlock : private LIR::Range
...
+    void SetBBPrev(BasicBlock* prev)
+    {
+        bbPrev = prev;
+        if (prev)
+        {
+            prev->bbNext = this;
+        }
+    }
+
+    BasicBlock* GetBBNext() const
```
SingleAccretion, line 548: While we're renaming things, we could give these more "obvious" names, e. g. `GetNextBlock`.

amanasifkhalid:
> [No asm diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=427804&view=ms.vss-build-web.run-extensions-tab) but some TP diffs (maybe because SetBBPrev() does more than what it replaces?)

We previously had BasicBlock::setNext() that updated bbNext, and also null-checked the new next block to see if it could update its bbPrev; most of the time, we'd just directly update bbNext, and would sparingly call setNext() if we also intended to update the next block's bbPrev. Would it make sense to not have SetBBNext() do this additional work, and keep setNext() around to do it? I think this would reduce/eliminate the TP diffs, but we'd lose the guarantee that bbNext and bbNext->bbPrev are consistent within SetBBNext().

> Seems like we should work to avoid too many GetBBNext() calls in the code (maybe as a follow-up?)

I'm happy to add helper methods for checking bbNext/bbPrev. This is for readability, since GetBBNext() should be getting inlined, right?

amanasifkhalid:
> E.g., instead of `(block->bbNext == nullptr)` => a new function `block->IsLast()`, and `block->bbPrev == nullptr` => `block->IsFirst()`.

Would it be preferable to check against fgFirstBB and fgLastBB instead of adding new methods? Ditto checking if block == fgFirstColdBlock instead of a new method (though for checking if a block is the last hot block, I agree we'd need a new method to avoid calling GetBBNext()).

BruceForstall:
> Would it make sense to not have SetBBNext() do this additional work, and keep setNext() around to do it?

No, it's probably best to leave it as you have it now.

> I'm happy to add helper methods for checking bbNext/bbPrev. This is for readability, since GetBBNext() should be getting inlined, right?

Yes.

> Would it be preferable to check against fgFirstBB and fgLastBB instead of adding new methods? Ditto checking if block == fgFirstColdBlock instead of a new method (though for checking if a block is the last hot block, I agree we'd need a new method to avoid calling GetBBNext()).

We don't always have a `Compiler*` available that can be used to get `fgFirstBB`/`fgLastBB`.

For `fgFirstColdBlock` we have no other way to check except using that variable, which then requires a `Compiler*`.

AndyAyersMS:
I suspect over time we're going to see more and more cases where BB operations are contextual and need access to the compiler instance. At various times I've wondered if blocks should just point back to some parent object (either the compiler or some sort of "flow graph" object, but that seems wasteful. Given that blocks are all the same size we could perhaps play tricks to factor out common fields, if we say allocated them contiguously. For now, I think providing the context when needed via args is ok.

One issue with either a context arg or a parent object is that when we inline we do some funny things as there are two compiler objects around, and so it's possible to consult the wrong one....

amanasifkhalid:
@BruceForstall @AndyAyersMS PTAL. CI is clean, but [small TP diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=428856&view=ms.vss-build-web.run-extensions-tab). I'm planning on making the jump target members private next; if we can assert the jump kind is valid every time we try to read the jump destination, then maybe we can remove code that sets the jump destination to null when changing the jump kind. This might make up for the TP regression in Release builds.

AndyAyersMS (approved):
This LGTM. Thanks!

src/coreclr/jit/block.cpp
```diff
@@ -186,6 +186,21 @@ FlowEdge* Compiler::BlockPredsWithEH(BasicBlock* blk)
     return res;
 }
 
+//------------------------------------------------------------------------
+// IsLastHotBlock: see if this is the last block before the cold section
```
BruceForstall, line 190: Should we have a corresponding `IsFirstColdBlock(Compiler*)`?

src/coreclr/jit/block.h
```diff
@@ -1435,10 +1476,10 @@ class BasicBlockIterator
     {
         assert(m_block != nullptr);
         // Check that we haven't been spliced out of the list.
-        assert((m_block->bbNext == nullptr) || (m_block->bbNext->bbPrev == m_block));
-        assert((m_block->bbPrev == nullptr) || (m_block->bbPrev->bbNext == m_block));
+        assert((m_block->IsLast()) || m_block->Next()->PrevIs(m_block));
+        assert((m_block->IsFirst()) || m_block->Prev()->NextIs(m_block));
```
BruceForstall, line 1480: With the function, we can remove an unnecessary set of parens
```suggestion
        assert(m_block->IsLast() || m_block->Next()->PrevIs(m_block));
        assert(m_block->IsFirst() || m_block->Prev()->NextIs(m_block));
```

src/coreclr/jit/codegencommon.cpp
```diff
@@ -5210,7 +5210,7 @@ void CodeGen::genReserveEpilog(BasicBlock* block)
 
     assert(block != nullptr);
     const VARSET_TP& gcrefVarsArg(GetEmitter()->emitThisGCrefVars);
-    bool             last = (block->bbNext == nullptr);
+    bool             last = (block->IsLast());
```
BruceForstall, line 5213:
```suggestion
    bool             last = block->IsLast();
```
or just substitute it in the call below and eliminate the `last` var

src/coreclr/jit/codegencommon.cpp
```diff
@@ -5257,7 +5257,7 @@ void CodeGen::genReserveFuncletEpilog(BasicBlock* block)
 
     JITDUMP("Reserving funclet epilog IG for block " FMT_BB "\n", block->bbNum);
 
-    bool last = (block->bbNext == nullptr);
+    bool last = (block->IsLast());
```
BruceForstall, line 5260: ditto

src/coreclr/jit/codegenlinear.cpp
```diff
@@ -547,10 +546,10 @@ void CodeGen::genCodeForBBlist()
 
         /* Is this the last block, and are there any open scopes left ? */
 
-        bool isLastBlockProcessed = (block->bbNext == nullptr);
+        bool isLastBlockProcessed = (block->IsLast());
```
BruceForstall, line 549:
```suggestion
        bool isLastBlockProcessed = block->IsLast();
```

src/coreclr/jit/codegenlinear.cpp
```diff
@@ -547,10 +546,10 @@ void CodeGen::genCodeForBBlist()
...
         if (block->isBBCallAlwaysPair())
         {
-            isLastBlockProcessed = (block->bbNext->bbNext == nullptr);
+            isLastBlockProcessed = (block->Next()->IsLast());
```
BruceForstall, line 552:
```suggestion
            isLastBlockProcessed = block->Next()->IsLast();
```

src/coreclr/jit/codegenlinear.cpp
```diff
@@ -615,7 +614,7 @@ void CodeGen::genCodeForBBlist()
             // Note: we may be generating a few too many NOPs for the case of call preceding an epilog. Technically,
             // if the next block is a BBJ_RETURN, an epilog will be generated, but there may be some instructions
             // generated before the OS epilog starts, such as a GS cookie check.
-            if ((block->bbNext == nullptr) || !BasicBlock::sameEHRegion(block, block->bbNext))
+            if ((block->IsLast()) || !BasicBlock::sameEHRegion(block, block->Next()))
```
BruceForstall, line 617:
```suggestion
            if (block->IsLast() || !BasicBlock::sameEHRegion(block, block->Next()))
```

src/coreclr/jit/codegenlinear.cpp
```diff
@@ -818,7 +817,7 @@ void CodeGen::genCodeForBBlist()
             GetEmitter()->emitLoopAlignment(DEBUG_ARG1(block->KindIs(BBJ_ALWAYS)));
         }
 
-        if ((block->bbNext != nullptr) && (block->bbNext->isLoopAlign()))
+        if (!block->IsLast() && (block->Next()->isLoopAlign()))
```
BruceForstall, line 820:
```suggestion
        if (!block->IsLast() && block->Next()->isLoopAlign())
```

src/coreclr/jit/compiler.cpp
```diff
@@ -5291,11 +5291,11 @@ PhaseStatus Compiler::placeLoopAlignInstructions()
             }
         }
 
-        if ((block->bbNext != nullptr) && (block->bbNext->isLoopAlign()))
+        if (!block->IsLast() && (block->Next()->isLoopAlign()))
```
BruceForstall, line 5294:
```suggestion
        if (!block->IsLast() && block->Next()->isLoopAlign())
```

src/coreclr/jit/fgbasic.cpp
```diff
@@ -5641,24 +5635,20 @@ void Compiler::fgMoveBlocksAfter(BasicBlock* bStart, BasicBlock* bEnd, BasicBloc
     {
         printf("Relocated block%s [" FMT_BB ".." FMT_BB "] inserted after " FMT_BB "%s\n", (bStart == bEnd) ? "" : "s",
                bStart->bbNum, bEnd->bbNum, insertAfterBlk->bbNum,
-               (insertAfterBlk->bbNext == nullptr) ? " at the end of method" : "");
+               (insertAfterBlk->IsLast()) ? " at the end of method" : "");
```
BruceForstall, line 5638:
```suggestion
               insertAfterBlk->IsLast() ? " at the end of method" : "");
```

BruceForstall (commented):
Found a few nits that could be updated. Can't say I looked at every case.

src/coreclr/jit/block.cpp
amanasifkhalid: I can add that in for consistency.

src/coreclr/jit/block.h
amanasifkhalid: Thanks for catching all these nits (sorry, these kinds of PRs seem especially tedious to review).

amanasifkhalid:
@BruceForstall I think I fixed all the style issues introduced (plus a couple extra I found along the way).

BruceForstall (approved):
LGTM. Thanks!

PR #93035, opened 2023-10-04
[mono][hotreload] Adding test case failing

thaystg:
If add a field and do not assign anything to it the field->type is NULL.


lambdageek:
Fix is in #93558

PR #93038, opened 2023-10-04
Split the MarshalAsParser into two classes

jkoritzinsky:
One class will handle the raw parsing of the MarshalAs attribute.

The other class handles resolving a MarshalAs attribute into the custom marshaller type that will provide the marshalling based on our generator's rules.

This split makes it easier for other generator models to reuse our parsing logic without having to also reuse our custom marshaller resolution logic. In particular, a hypothetical IDispatch-focused generator would want to marshal all values into `VARIANT` values, so they may want to reuse the marshaller attribute parsing logic, but not the custom marshallers directly.

### DIFF
--- a/src/libraries/System.Runtime.InteropServices/gen/Microsoft.Interop.SourceGeneration/MarshalAsParser.cs
+++ b/src/libraries/System.Runtime.InteropServices/gen/Microsoft.Interop.SourceGeneration/MarshalAsParser.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Interop
     /// <summary>
     /// Simple User-application of System.Runtime.InteropServices.MarshalAsAttribute
     /// </summary>
-    public sealed record MarshalAsInfo(
+    public abstract record MarshalAsInfo(
         UnmanagedType UnmanagedType,
         CharEncoding CharEncoding) : MarshallingInfoStringSupport(CharEncoding)
     {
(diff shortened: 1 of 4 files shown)


src/libraries/System.Runtime.InteropServices/gen/Microsoft.Interop.SourceGeneration/MarshalAsParser.cs
```diff
@@ -21,18 +21,25 @@ public sealed record MarshalAsInfo(
...
+    public sealed record MarshalAsArrayInfo(
+        UnmanagedType UnmanagedType,
+        CharEncoding CharEncoding,
+        UnmanagedType ArraySubType): MarshalAsInfo(UnmanagedType, CharEncoding);
+
     /// <summary>
-    /// This class suppports parsing a System.Runtime.InteropServices.MarshalAsAttribute.
+    /// This class suppports parsing a System.Runtime.InteropServices.MarshalAsAttribute into a <see cref="MarshalAsInfo"/>.
     /// </summary>
     public sealed class MarshalAsAttributeParser : IMarshallingInfoAttributeParser, IUseSiteAttributeParser
```
jtschuster, line 36: Is this still used as a `IMarshallingInfoAttributeParser` now?

jtschuster (approved):

src/libraries/System.Runtime.InteropServices/gen/Microsoft.Interop.SourceGeneration/MarshalAsParser.cs
jkoritzinsky: It's intended to optionally be used that way. We don't use it in the same way currently, but we delegate to this implementation from the other one through the interface.

PR #93039, opened 2023-10-04
[mono][hotreload] Generate seq_points if we don't have pdb information from HotReload

thaystg:
Sometimes when we apply changes on runtime side we don't receive the PDB information, so we should try to generate the seq points using the same rules used by coreclr.

Maybe in the future we can use these rules always and then avoid runtime loading the PDB information???

### DIFF
--- a/src/mono/mono/metadata/mono-debug.c
+++ b/src/mono/mono/metadata/mono-debug.c
@@ -1127,6 +1127,21 @@ mono_debug_enabled (void)
 	return mono_debug_format != MONO_DEBUG_FORMAT_NONE;
 }
 
+
+//Returns true if the method has updates but doesn't have ppdb information then we should generate the seq points using the coreclr rules
+mono_bool
+mono_debug_generate_enc_seq_points_without_debug_info (MonoDebugMethodInfo *minfo)
+{
+	MonoImage* img = m_class_get_image (minfo->method->klass);
+	if (img->has_updates) {
+		guint32 idx = mono_metadata_token_index (minfo->method->token);
+		MonoDebugInformationEnc *mdie = (MonoDebugInformationEnc *) mono_metadata_update_get_updated_method_ppdb (img, idx);
+		if (mdie == NULL)
+			return TRUE;
+	}
+	return FALSE;
+}
+
 void
 mono_debug_get_seq_points (MonoDebugMethodInfo *minfo, char **source_file, GPtrArray **source_file_list, int **source_files, MonoSymSeqPoint **seq_points, int *n_seq_points)
 {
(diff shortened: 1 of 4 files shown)


src/mono/mono/mini/interp/transform.c
```diff
@@ -5084,6 +5087,15 @@ generate_code (TransformData *td, MonoMethod *method, MonoMethodHeader *header,
 			WRITE64_INS (td->last_ins, 0, &counter);
 		}
 
+		if (generate_enc_seq_points_without_debug_info)
+		{
+			if ((*td->ip == CEE_NOP) ||
+				(*td->ip == CEE_CALLVIRT) ||
+				(*td->ip == CEE_CALLI) ||
+				(*td->ip == CEE_CALL) ||
+				(GPTRDIFF_TO_INT (td->sp - td->stack) == 0))
```
lambdageek, line 5096: Can this be a helper function.  Also add a comment that this is matching CoreCLR's algorithm for adding sequence points

src/mono/mono/metadata/mono-debug.c
```diff
@@ -1127,6 +1127,21 @@ mono_debug_enabled (void)
 	return mono_debug_format != MONO_DEBUG_FORMAT_NONE;
 }
 
+
+//Returns true if the method has updates but doesn't have ppdb information then we should generate the seq points using the coreclr rules
+mono_bool
+mono_debug_generate_enc_seq_points_without_debug_info (MonoDebugMethodInfo *minfo)
+{
+	MonoImage* img = m_class_get_image (minfo->method->klass);
+	if (img->has_updates) {
```
lambdageek, line 1136:
```suggestion
	if (G_UNLIKELY (img->has_updates)) {
```

src/mono/mono/mini/interp/transform.c
```diff
@@ -5084,6 +5087,15 @@ generate_code (TransformData *td, MonoMethod *method, MonoMethodHeader *header,
 			WRITE64_INS (td->last_ins, 0, &counter);
 		}
 
+		if (generate_enc_seq_points_without_debug_info)
```
lambdageek, line 5090:
```suggestion
		if (G_UNLIKELY (generate_enc_seq_points_without_debug_info))
```

src/native/public/mono/metadata/details/mono-debug-functions.h
```diff
@@ -7,6 +7,7 @@
 #endif
 
 MONO_API_FUNCTION(mono_bool, mono_debug_enabled, (void))
+MONO_API_FUNCTION(mono_bool, mono_debug_generate_enc_seq_points_without_debug_info, (MonoDebugMethodInfo* minfo))
```
lambdageek, line 10: I don't think this should be a public API.

lambdageek (commented):
Mostly looks ok, but I don't think we need to add a new public API for this.

lambdageek (approved):

PR #93045, opened 2023-10-05
CI: Do not run `CrossDac` job for mono, or wasm changes.

radical:


radical:
@jkoritzinsky the new crossdac job is running unconditionally. So, I added a condition to run it for non-mono, and non-wasm changes. I don't know what all this depends on - maybe `SetPathVars_coreclr.containsChanges` would be enough? In any case, this does not need to run for wasm.

@steveisok Is this needed for mono?

hoyosjs (approved):

jkoritzinsky (approved):
This job should be built for any set of changes that run tests with CoreCLR. The build is more for better diagnostics than any build validation.

radical:
The NativeAOT failure is helix failing, and unrelated to this PR.

PR #93051, opened 2023-10-05
Fix passing CMAKE_BUILD_TYPE in runtime tests using Ninja

akoeplinger:
I noticed that we were building the native assets of runtime tests always in Debug mode on Windows when using Ninja, even though we passed CMAKE_BUILD_TYPE=Release.

This results in warnings like this about mismatching debug/release msvc runtime when building with NativeAOT (which uses release msvc runtime in Release config):

> [warning]LIBCMT.lib(initializers.obj)(0,0): warning LNK4098: defaultlib 'libcmtd.lib' conflicts with use of other libs; use /NODEFAULTLIB:library

It turned out that this is because we were accidentally passing CMAKE_SYSTEM_VERSION and CMAKE_BUILD_TYPE as a single string, instead of multiple. This resulted in the CMAKE_SYSTEM_VERSION being set like this in CMakeCache.txt:

```
CMAKE_SYSTEM_VERSION:UNINITIALIZED=10.0 -DCMAKE_BUILD_TYPE=Release
```

CMake then defaulted CMAKE_BUILD_TYPE to Debug.

This uncovered an MSVC bug when compiling some sources with the MSVC version used on CI in /O2 mode. This is fixed in 19.38+, disabled optimization in these files on older versions to workaround.

### DIFF
--- a/src/tests/Interop/DisabledRuntimeMarshalling/DisabledRuntimeMarshallingNative.cpp
+++ b/src/tests/Interop/DisabledRuntimeMarshalling/DisabledRuntimeMarshallingNative.cpp
@@ -3,6 +3,11 @@
 
 #include <xplatform.h>
 
+// MSVC versions before 19.38 generate incorrect code for this file when compiling with /O2 
+#if defined(_MSC_VER) && (_MSC_VER < 1938)
+#pragma optimize("", off)
+#endif
+
 struct StructWithShortAndBool
 {
     bool b;
--- a/src/tests/Interop/PInvoke/Array/MarshalArrayAsField/LPArrayNative/MarshalArrayByValArrayNative.cpp
+++ b/src/tests/Interop/PInvoke/Array/MarshalArrayAsField/LPArrayNative/MarshalArrayByValArrayNative.cpp
@@ -7,6 +7,11 @@
 
 using namespace std;
 
+// MSVC versions before 19.38 generate incorrect code for this file when compiling with /O2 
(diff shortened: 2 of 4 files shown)


akoeplinger:
Hmm looks like this causes one test failure:
```
      ============ [TakeFloatArraySeqStructByVal] ============
      WARNING: Test error - int __cdecl Equals<float>(float *,int,float *,int)
      Array Element Not Equal: index: 1Expected: 100
      Actual: 101
      END EXECUTION - FAILED
      FAILED
```

jkoritzinsky (approved):

jkoritzinsky:
I’ll take a look and see if I can figure it out.

akoeplinger:
Actually, looks like the failure I posted above is specific to the NativeAOT job, on the JIT job it fails in `CStyle_Array_Float`

akoeplinger:
If I replace the MarshalArrayLPArrayNative.dll and MarshalArrayByValArrayNative.dll files in the Helix artifacts with my local build then the error goes away.

akoeplinger:
Looks like a compiler bug on the MSVC used on CI to me, if I decompile the `TakeFloatArraySeqStructByVal` function using Ghidra the broken one looks like this, it's assigning the wrong values to the array:

image


... while the working local build of the .dll looks as expected:

image

MichalPetryka:
> /O3 then the MSVC versions

MSVC doesn't have O3 so it ignores that.

PR #93053, opened 2023-10-05
Pass /OPT:NOREF for DEBUG VS builds

jakobbotsch:
/OPT:REF will cause the linker to optimize out unused functions/statics, which is not desirable when debugging where you may want to be able to use and inspect those members from the debugger.

See #92993 for some more context.


BruceForstall (approved):

jkoritzinsky (approved):
cc: @jtschuster

PR #93062, opened 2023-10-05
SPMI: Fix wrong SPMI replay result check

jakobbotsch:
Fix a bug introduced in #92824 that would cause SPMI to not show replay failures.


jakobbotsch:
cc @dotnet/jit-contrib

AndyAyersMS (approved):

PR #93068, opened 2023-10-05
Move internal BitOperations.LeadingZeroCount({U}Int128)

stephentoub:
With no other signed overloads exposed, it's easy to accidentally use these for int/long/etc.  They exist only for the implementation of {U}Int128 at this point, so just move them to be in those types.


src/libraries/System.Private.CoreLib/src/System/Int128.cs
```diff
@@ -742,12 +742,19 @@ public static (Int128 Quotient, Int128 Remainder) DivRem(Int128 left, Int128 rig
...
         /// <inheritdoc cref="IBinaryInteger{TSelf}.LeadingZeroCount(TSelf)" />
         public static Int128 LeadingZeroCount(Int128 value)
+            => value.LeadingZeroCount();
+
+        /// <summary>Computes the number of leading zero bits in this value.</summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private int LeadingZeroCount()
         {
+            Int128 value = this;
+
```
jkotas, line 752:
```suggestion
```
Can we fold the private method into the public method? I do not see the reason for the split.

stephentoub: The public method returns Int128, but there are other methods on these types that want an int; seemed unnecessary to roundtrip through Int128 to get it.

jkotas: > The public method returns Int128

That is a weird design. Was it intentional?

jkotas (approved):

src/libraries/System.Private.CoreLib/src/System/Int128.cs
tannergooding: The public methods have to account for being used in a generic context, where you often expect or require `T in, T out` and where type changes can lead to additional work or complexity.

This then also matches how most hardware behaves when their is direct support for the operation over a given `T`.

src/libraries/System.Private.CoreLib/src/System/Int128.cs
```diff
@@ -742,12 +742,19 @@ public static (Int128 Quotient, Int128 Remainder) DivRem(Int128 left, Int128 rig
 
         /// <inheritdoc cref="IBinaryInteger{TSelf}.LeadingZeroCount(TSelf)" />
         public static Int128 LeadingZeroCount(Int128 value)
+            => value.LeadingZeroCount();
+
+        /// <summary>Computes the number of leading zero bits in this value.</summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private int LeadingZeroCount()
         {
+            Int128 value = this;
```
tannergooding, line 751: Why do we need to make a copy of `this`, rather than just reading `_lower` and `_upper` directly?

tannergooding: and why does it need to be an instance method at all?

stephentoub: > Why do we need to make a copy of this, rather than just reading _lower and _upper directly?

I did that to ensure there's a consistent view of the data, so that even if there's a tear while reading the value, subsequent operations on it will be consistent, e.g. the check on _upper and then the use of _upper.  If we don't think that's important, I can remove the local.

> and why does it need to be an instance method at all?

Because you can't overload by return type, so we'd need to name it something different if it's taking the same arguments as the existing method.  Is that what you'd prefer?

tannergooding: I think having a different name, like `LeadingZeroCountCore` would be fine here. That allows it to be `static`, allows the minor optimization of not needing the JIT to promote and remove `upper` since it's unused, etc.

It being instance and forcing the value to become address taken and then additionally inserting a copy feels like its create more work and is likely to de-optimize things in the long run.

jkotas (approved):

tannergooding (approved):

PR #93071, opened 2023-10-05
JIT: Support retbuf optimization for non 'lvIsTemp' locals

jakobbotsch:
* Handle liveness for the LCL_ADDR definitions when we get to the call
* Remove lvIsTemp check from retbuf optimization


jakobbotsch:
This had nice improvements, but also a large number of regressions. I'm hoping that switching retbufs to be handled by physical promotion will improve on the results, but I need to do that work first.

PR #93079, opened 2023-10-05
Get rid of static constructor in Task<T>

lambdageek:
Because the `default(TResult)` task was cached in in the `Task<TResult>` class, every instantiation would run the static constructor.

With this change, the cached value is stored in
`TaskCache.CachedDefaultTask<TResult>` instead, and `Task<TResult>` no longer needs to execute a static constructor.

This is expected to improve startup time for applications that do a lot of async work.  (But no measurements have been done).  Validated that ``Task`1::.cctor`` does not exist in the IL after this change

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Future.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Future.cs
@@ -58,9 +58,6 @@ namespace System.Threading.Tasks
     [DebuggerDisplay("Id = {Id}, Status = {Status}, Method = {DebuggerDisplayMethodDescription}, Result = {DebuggerDisplayResultDescription}")]
     public class Task<TResult> : Task
     {
-        /// <summary>A cached task for default(TResult).</summary>
-        internal static readonly Task<TResult> s_defaultResultTask = TaskCache.CreateCacheableTask<TResult>(default);
-
         private static TaskFactory<TResult>? s_Factory;
 
         // The value itself, if set.
--- a/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/TaskCache.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/TaskCache.cs
@@ -40,5 +40,14 @@ namespace System.Threading.Tasks
 
             return tasks;
         }
+
+        /// <summary>
+        /// A cached task for default(TResult).
+        /// </summary>
+        internal static class CachedDefaultValueTask<TResult>
+        {
+            internal static Task<TResult> s_defaultResultTask = CreateCacheableTask<TResult>(default);
+        }
+
     }
 }
--- a/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/TaskCache.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/TaskCache.cs
@@ -44,7 +44,7 @@ namespace System.Threading.Tasks
         /// <summary>
         /// A cached task for default(TResult).
         /// </summary>
-        internal static class CachedDefaultValueTask<TResult>
+        internal static class CachedDefaultTask<TResult>
         {
             internal static readonly Task<TResult> s_defaultResultTask = CreateCacheableTask<TResult>(default);
         }
--- a/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/TaskCache.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/TaskCache.cs
@@ -46,7 +46,7 @@ namespace System.Threading.Tasks
         /// </summary>
         internal static class CachedDefaultValueTask<TResult>
         {
-            internal static Task<TResult> s_defaultResultTask = CreateCacheableTask<TResult>(default);
+            internal static readonly Task<TResult> s_defaultResultTask = CreateCacheableTask<TResult>(default);
         }
 
     }
--- a/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs
@@ -1041,7 +1041,7 @@ namespace System.IO
                     Task.CompletedTask;
 
             public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback? callback, object? state) =>
-                TaskToAsyncResult.Begin(Task<int>.s_defaultResultTask, callback, state);
+                TaskToAsyncResult.Begin(TaskCache.CachedDefaultValueTask<int>.s_defaultResultTask, callback, state);
 
             public override int EndRead(IAsyncResult asyncResult) =>
                 TaskToAsyncResult.End<int>(asyncResult);
--- a/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/IO/Stream.cs
@@ -1041,7 +1041,7 @@ namespace System.IO
                     Task.CompletedTask;
 
             public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback? callback, object? state) =>
-                TaskToAsyncResult.Begin(TaskCache.CachedDefaultValueTask<int>.s_defaultResultTask, callback, state);
+                TaskToAsyncResult.Begin(TaskCache.CachedDefaultTask<int>.s_defaultResultTask, callback, state);
 
             public override int EndRead(IAsyncResult asyncResult) =>
                 TaskToAsyncResult.End<int>(asyncResult);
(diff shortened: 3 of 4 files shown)


lambdageek:
@stephentoub I was looking at logs of the methods that Mono AOT was loading for a ios app and noticed that `Task<XYZ>::.cctor` was appearing frequently during startup and felt motivated to make it go away.  I haven't tried measuring the impact yet, but I kind of want to explore this.  Do you think that, in principle, this is beneficial?

src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/TaskCache.cs
```diff
@@ -40,5 +40,14 @@ private static Task<int>[] CreateInt32Tasks()
 
             return tasks;
         }
+
+        /// <summary>
+        /// A cached task for default(TResult).
+        /// </summary>
+        internal static class CachedDefaultValueTask<TResult>
+        {
+            internal static Task<TResult> s_defaultResultTask = CreateCacheableTask<TResult>(default);
```
stephentoub, line 49:
```suggestion
            internal static readonly Task<TResult> s_defaultResultTask = CreateCacheableTask<TResult>(default);
```

src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/TaskCache.cs
```diff
@@ -40,5 +40,14 @@ private static Task<int>[] CreateInt32Tasks()
 
             return tasks;
         }
+
+        /// <summary>
+        /// A cached task for default(TResult).
+        /// </summary>
+        internal static class CachedDefaultValueTask<TResult>
```
stephentoub, line 47: Does this introduce any perf issues, in particular for reference type TResults, where it's now slower due to one generic context calling into a different generic context?

stephentoub, line 47: Nit: CachedDefaultValueTask => CachedDefaultTask... otherwise it can be read as "cached default ValueTask" rather than "cached "default value" task"

lambdageek: @stephentoub There is no such situation in the existing code, as far as I can tell.

The old `Task<TResult>.s_DefaultResultTask` was only used in 3 places:
1. in Stream.cs we had `Task<int>.s_DefaultResultTask` which is a concerete instantiation
2. in Task.cs we had two uses in `Task.FromResult<TResult>(TResult result)`:
  * `if (result is null) return Task<TResult>.s_DefaultResultTask` was already a generic context calling into a different generic context.  Same as the new code `TaskCache.CachedDefaultTask<TResult>.s_DefaultResultTask`
  * `if (/* result is the same size as one of several primitive value types and the bitpattern matches the default value of that primitive value type */) return Task<TResult>.s_DefaultResultTask;` - again this is already one generic context calling into another generic context, although they're not reference types.

I'll try a microbenchmark on `Task.FromResult<T>(default(T))` for some different types.

lambdageek: Added a microbenchmark:

```csharp
[GenericTypeArguments(typeof(byte))]
[GenericTypeArguments(typeof(int))]
[GenericTypeArguments(typeof(string))]
[GenericTypeArguments(typeof(Type))]
[BenchmarkCategory(Categories.Libraries)]
public class Perf_FromResult<T>
{
    [Benchmark]
    public void FromResultDefault()
    {
        Task<T> d = Task.FromResult<T>(default(T));
    }
}
```

(below "runtime-bugs" is baseline, and "runtime" is the PR)

Results for `string`:

| Method            | Job        | Toolchain                                                                                                | Mean     | Error     | StdDev    | Median   | Min      | Max      | Ratio | MannWhitney(5%) | RatioSD | Allocated | Alloc Ratio |
|------------------ |----------- |--------------------------------------------------------------------------------------------------------- |---------:|----------:|----------:|---------:|---------:|---------:|------:|---------------- |--------:|----------:|------------:|
| FromResultDefault | Job-KNJPPY | /runtime-bugs/artifacts/bin/testhost/net9.0-osx-Release-arm64/shared/Microsoft.NETCore.App/9.0.0/corerun | 6.648 ns | 0.1802 ns | 0.2075 ns | 6.568 ns | 6.466 ns | 7.146 ns |  1.00 | Base            |    0.00 |         - |          NA |
| FromResultDefault | Job-SJEICH | /runtime/artifacts/bin/testhost/net9.0-osx-Release-arm64/shared/Microsoft.NETCore.App/9.0.0/corerun      | 6.358 ns | 0.0123 ns | 0.0109 ns | 6.356 ns | 6.344 ns | 6.379 ns |  0.96 | Same            |    0.03 |         - |          NA |

results for `System.Type`:

| Method            | Job        | Toolchain                                                                                                | Mean     | Error     | StdDev    | Median   | Min      | Max      | Ratio | MannWhitney(5%) | Allocated | Alloc Ratio |
|------------------ |----------- |--------------------------------------------------------------------------------------------------------- |---------:|----------:|----------:|---------:|---------:|---------:|------:|---------------- |----------:|------------:|
| FromResultDefault | Job-KNJPPY | /runtime-bugs/artifacts/bin/testhost/net9.0-osx-Release-arm64/shared/Microsoft.NETCore.App/9.0.0/corerun | 6.391 ns | 0.0072 ns | 0.0067 ns | 6.390 ns | 6.384 ns | 6.405 ns |  1.00 | Base            |         - |          NA |
| FromResultDefault | Job-SJEICH | /runtime/artifacts/bin/testhost/net9.0-osx-Release-arm64/shared/Microsoft.NETCore.App/9.0.0/corerun      | 6.368 ns | 0.0136 ns | 0.0113 ns | 6.365 ns | 6.354 ns | 6.398 ns |  1.00 | Same            |         - |          NA |

src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Future.cs
```diff
@@ -58,9 +58,6 @@ namespace System.Threading.Tasks
     [DebuggerDisplay("Id = {Id}, Status = {Status}, Method = {DebuggerDisplayMethodDescription}, Result = {DebuggerDisplayResultDescription}")]
     public class Task<TResult> : Task
     {
-        /// <summary>A cached task for default(TResult).</summary>
```
jkotas, line 61: This generates before-field-init constructor. The constructor should be running lazily only when the field is accessed.

jkotas: Is this working around Mono running before-field-init constructors earlier than required? It would be better to fix that in Mono for all before-field-init constructors.

jkotas:
> I haven't tried measuring the impact yet,

I expect that it is going to regress static footprint of apps that use async heavily.

src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Future.cs
lambdageek: > This generates before-field-init constructor. The constructor should be running lazily only when the field is accessed. I do not see how this helps.

It's making the AOT compiler compile a cctor for every instance of Task<T>, even if it is never used

lambdageek:
> > I haven't tried measuring the impact yet,
> 
> I expect that it is going to regress static footprint of apps that use async heavily (on CoreCLR/NativeAOT at least).

I expect the exact opposite. I'll look at aot size tomorrow

src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Future.cs
lambdageek: > Is this working around Mono running before-field-init constructors earlier than required? It would be better to fix that in Mono for all before-field-init constructors.

That is a good point, after we make that fix, this won't show up in startup traces anymore.

I still feel this might be a trimming/AOT win. But I guess it depends on how prevalent `Task.FromResult` is at different instantiations.

I'll close this tomorrow if I don't see an AOT impact

lambdageek:
@jkotas Here's the final binaries for the [aspnet/Benchmarks "TodosApi"](https://github.com/aspnet/Benchmarks/tree/5e95f153591537206f717cfcba2603beed23e88b/src/BenchmarksApps/TodosApi) with .NET 8 RC1, a .NET 9 baseline and this PR published with NativeAot in release mode.  (I verified that the Ilc .rsp file points to the CoreLib from the relevant build of the runtime):

```console
-rwxr-xr-x  1 alklig  staff  24772904 Oct  6 12:24 TodosApi-net8.0-rc1
-rwxr-xr-x  1 alklig  staff  25103592 Oct  6 12:32 TodosApi-net9.0-PR
-rwxr-xr-x  1 alklig  staff  25103608 Oct  6 12:31 TodosApi-net9.0-baseline
```

This PR is a couple of bytes smaller.

My intuition is that `Task.FromResult<T>` is pretty rare in user code - which would mostly use `async/await` and it is pretty rare in libraries and frameworks - which are more likely to use a `TaskCompletionSource<T>`

jkotas:
It is interesting that the AOT size is same for your test. 16 bytes is not meaningful difference. I think your intuition about Task.FromResult<T>` being used rarely in the core libraries is right. However, it seems to be used often enough to make this change a wash.

PR #38993, opened 2020-07-09
Allow building libs tests for all configurations

ViktorHofer:
Building libraries tests and passing in the -allconfigurations switch
results in the libraries packages being built and tested. Conditioning
the packages build/test on the /p:TestPackages switch.

PR #93082, opened 2023-10-05
Convert AllConfigurations and NETFX jobs to use the global build templates

jkoritzinsky:
Moving these two jobs that have minor customizations to use the global build template + post build steps to reduce complexity in the libraries build and test jobs. Also clean up some places that were still on extraStepsTemplate or were attempting to be (but had typos).

### DIFF
--- a/eng/pipelines/runtime.yml
+++ b/eng/pipelines/runtime.yml
@@ -1128,6 +1128,7 @@ extends:
                 parameters:
                   creator: dotnet-bot
                   testRunNamePrefixSuffix: NET48_$(_BuildConfig)
+                  extraHelixArguments: /p:BuildTargetFramework=net48
             condition: >-
               or(
                 eq(dependencies.evaluate_paths.outputs['SetPathVars_libraries.containsChange'], true),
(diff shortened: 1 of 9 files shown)


eng/pipelines/runtime.yml
```diff
@@ -1107,32 +1108,44 @@ extends:
...
-          jobTemplate: /eng/pipelines/libraries/build-job.yml
+          jobTemplate: /eng/pipelines/common/global-build-job.yml
           buildConfig: ${{ variables.debugOnPrReleaseOnRolling }}
           platforms:
           - windows_x64
           jobParameters:
-            framework: allConfigurations
-            runTests: true
-            useHelix: false
+            buildArgs: -restore -build -test -s tools+libs+libs.tests -allConfigurations -c $(_BuildConfig) /p:TestAssemblies=false /p:TestPackages=true
```
ViktorHofer, line 1146: This might have already been like this before but I think that `-restore -build -test` can be collapsed to `-test` as `-restore -build` is the default when no action is passed in: https://github.com/dotnet/runtime/blob/66f4a1df77a0a979a0b9c0941700eaf2ee808de5/eng/build.ps1#L251C33-L251C33

ViktorHofer: and btw, at some point we should add another subset for those allconfigurations packages tests instead of relying on msbuild properties. cc @ericstj

ericstj: File an issue?  Not sure I have a strong opinion here so long as we still do it in CI.  Here's the PR that added that: https://github.com/dotnet/runtime/pull/38993

ViktorHofer (approved):

PR #93084, opened 2023-10-05
JIT: Remove duplicated stores

EgorBo:
```cs
void Test(char[] arr)
{
    arr[0] = 'a';
    arr[0] = 'b';
}
```
Codegen diff:
```diff
; Method Program:Test(ushort[]):this (FullOpts)
G_M45342_IG01:  ;; offset=0x0000
       sub      rsp, 40
						;; size=4 bbWeight=1 PerfScore 0.25

G_M45342_IG02:  ;; offset=0x0004
       cmp      dword ptr [rdx+0x08], 0
       jbe      SHORT G_M45342_IG04
       add      rdx, 16
-      mov      word  ptr [rdx], 97
       mov      word  ptr [rdx], 98
						;; size=20 bbWeight=1 PerfScore 6.25

G_M45342_IG03:  ;; offset=0x0018
       add      rsp, 40
       ret      
						;; size=5 bbWeight=1 PerfScore 1.25

G_M45342_IG04:  ;; offset=0x001D
       call     CORINFO_HELP_RNGCHKFAIL
       int3     
						;; size=6 bbWeight=0 PerfScore 0.00
-; Total bytes of code: 35
+; Total bytes of code: 30
```

Presumably, VNBasedDeadStoreRemoval wasn't able to handle these because of bound checks.


SingleAccretion:
> VNBasedDeadStoreRemoval

This phase only operates on local stores. The same principle could in theory be applied to memory stores, but would be much more complicated (and slower).

EgorBo:
Suprisingly, it found quite a few diffs: https://dev.azure.com/dnceng-public/public/_build/results?buildId=429076&view=ms.vss-build-web.run-extensions-tab

@dotnet/jit-contrib @jakobbotsch PTAL

src/coreclr/jit/lower.cpp
```diff
@@ -8007,8 +8000,23 @@ void Lowering::LowerStoreIndirCoalescing(GenTreeStoreInd* ind)
...
         }
 
-        // Offset has to match the size of the type. We don't support the same or overlapping offsets.
-        if (abs(prevData.offset - currData.offset) != (int)genTypeSize(prevData.targetType))
+        const int offset = abs(prevData.offset - currData.offset);
+
+        // Offset has to match the size of the type.
+        if (offset != (int)genTypeSize(prevData.targetType))
+        {
+            if (offset == 0)
```
kunalspathak, line 8008: This is basically doing:

```c++
if (prevData.offset == currData.offset)
{
  continue;
}
```

I didn't understand the comment and need for `if (offset != (int)genTypeSize(prevData.targetType))`.

kunalspathak: you could have the `if (offset != (int)genTypeSize(prevData.targetType))` condition separately.

if (prevData.offset == currData.offset)
{
  continue;
}
if (offset != (int)genTypeSize(prevData.targetType))
{
  return;
}

EgorBo: Ah, you're right - addressed

kunalspathak (approved):

jakobbotsch (approved):

PR #93086, opened 2023-10-05
Update JIT debug helpers

BruceForstall:
1. Make them available in builds linked using /OPT:REF (currently, Checked builds (on Windows); they are not available at all in Release builds).
2. Add a bunch of missing documentation.
3. Add `cTreeRange`
4. Make the newer dumpers more consistent with the rest by dumping a sequence number.
5. Add `dbLoop` for `dFindLoop`

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -9216,6 +9216,7 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
  *      cStmt,       dStmt          : Display a Statement (call gtDispStmt()).
  *      cTree,       dTree          : Display a tree (call gtDispTree()).
  *      cTreeLIR,    dTreeLIR       : Display a tree in LIR form (call gtDispLIRNode()).
+ *      cTreeRange,  dTreeRange     : Display a range of trees in LIR form (call gtDispLIRNode()).
  *      cTrees,      dTrees         : Display all the trees in a function (call fgDumpTrees()).
  *      cEH,         dEH            : Display the EH handler table (call fgDispHandlerTab()).
  *      cVar,        dVar           : Display a local variable given its number (call lvaDumpEntry()).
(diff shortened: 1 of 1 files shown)


BruceForstall:
@jakobbotsch PTAL
cc @dotnet/jit-contrib

src/coreclr/jit/compiler.cpp
```diff
@@ -9230,417 +9231,276 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
...
+// `/INCLUDE` switch, and potentially makes them easier to find by a debugger.
+//
+// On x64, it appears the names are not decorated (which is actually unexpected). However, on x86, the names
+// are decorated for `__stdcall` with a leading `_` and trailing `@N` for `N` bytes of arguments. We could
+// make the functions all `__cdecl` to simplify, with only a leading `_`. For now, only do the "include"
+// thing for x64.
+
+#define DBGAPI extern "C"
+
+#if defined(HOST_AMD64) && defined(_MSC_VER)
```
jakobbotsch, line 9268: I assume it's disabled on x86 due to the weird mangling done there, but what about arm64?

src/coreclr/jit/compiler.cpp
```diff
@@ -9230,417 +9231,276 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
...
+//
+// Mark them `export "C"` so the names are not C++ name mangled. This makes it easier to refer to them in the
+// `/INCLUDE` switch, and potentially makes them easier to find by a debugger.
+//
+// On x64, it appears the names are not decorated (which is actually unexpected). However, on x86, the names
+// are decorated for `__stdcall` with a leading `_` and trailing `@N` for `N` bytes of arguments. We could
+// make the functions all `__cdecl` to simplify, with only a leading `_`. For now, only do the "include"
+// thing for x64.
+
+#define DBGAPI extern "C"
```
jakobbotsch, line 9266: Alternatively you can use a single extern "C" block enclosing the functions below to avoid putting this on all functions manually.

src/coreclr/jit/compiler.cpp
jakobbotsch: If this indeed is disabled on x86 for that reason I think you can fix it by switching these functions to stdcall (or cdecl, the opposite of the default we're using). In that case it's easier to keep the DBGAPI macro around.

BruceForstall: Yes, it's due to different mangling on x86. The mangling is due to the calling convention; we default to stdcall. We could have a separate set of pragmas for x86 mangling. It would be easier to support cdecl since it doesn't include the argument size (so isn't fragile to function signature changes). Then, using `DBGAPI` to specify cdecl could be used. I presume it would work for native arm64 because that and x64 only have a single calling convention so don't need to mangle.

src/coreclr/jit/compiler.cpp
BruceForstall: Yeah, I'm not fond of that due to non-local (not as easily visible) context effects.

BruceForstall:
@jakobbotsch I updated it to work for x86/arm64 (though I didn't test it on arm64, but it does build)

jakobbotsch (approved):

PR #93088, opened 2023-10-05
Fix Options Source Gen Trimming Issues

### LINKED ISSUE #92327: Source generator for options validation leads to trimming warnings/errors in AOT
### Description

If you follow the [guide in the docs](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#source-generator-1), you get trimming warnings related to data annotations. In my own program, these warnings lead to errors.

### Reproduction Steps

* Create a new app: `dotnet new webapiaot`
* Add *NuGet.config*:

    ```xml
    <?xml version="1.0" encoding="utf-8"?>
    <configuration>
      <packageSources>
        <add key="dotnet8" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" />
      </packageSources>
    </configuration>
    ```

* Add nightly package references in *.csproj*:

    ```xml
    <ItemGroup>
      <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.0-rc.2.23466.4" />
      <PackageReference Include="Microsoft.Extensions.Options" Version="8.0.0-rc.2.23466.4" />
    </ItemGroup>
    ```

* Copy sample code [from the docs](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#source-generator-1) into your *Program.cs*
* Build app: `dotnet run build`

### Expected behavior

Program compiles without error and warnings.

### Actual behavior

You get a bunch of trimming warnings related to data annotations:

```txt
/tmp/optionsvalidator/Microsoft.Extensions.Options.SourceGeneration/Microsoft.Extensions.Options.Generators.OptionsValidatorGenerator/Validators.g.cs(50): Trim analysis warning IL2026: SecondValidatorNoNamespace.Validate(String,SecondModelNoNamespace): Using member 'System.ComponentModel.DataAnnotations.ValidationContext.ValidationContext(Object)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The Type of instance cannot be statically discovered and the Type's properties can be trimmed. [/tmp/optionsvalidator/optionsvalidator.csproj]
/tmp/optionsvalidator/Microsoft.Extensions.Options.SourceGeneration/Microsoft.Extensions.Options.Generators.OptionsValidatorGenerator/Validators.g.cs(17): Trim analysis warning IL2026: FirstValidatorNoNamespace.Validate(String,FirstModelNoNamespace): Using member 'System.ComponentModel.DataAnnotations.ValidationContext.ValidationContext(Object)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. The Type of instance cannot be statically discovered and the Type's properties can be trimmed. [/tmp/optionsvalidator/optionsvalidator.csproj]
/tmp/optionsvalidator/Microsoft.Extensions.Options.SourceGeneration/Microsoft.Extensions.Options.Generators.OptionsValidatorGenerator/Validators.g.cs(73): Trim analysis warning IL2026: __OptionValidationStaticInstances.<Validators_g>F1872690B80E3D4EBF0BB4EEFEFFBD255387F31D123553B47D3D3ED7D782424E7____Attributes..cctor(): Using member 'System.ComponentModel.DataAnnotations.MinLengthAttribute.MinLengthAttribute(Int32)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Uses reflection to get the 'Count' property on types that don't implement ICollection. This 'Count' property may be trimmed. Ensure it is preserved. [/tmp/optionsvalidator/optionsvalidator.csproj]
```

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

* Runtime: Microsoft.AspNetCore.App 8.0.0-rc.1.23421.29
* SDK: 8.0.100-rc.1.23455.8
* OS: Linux

### Other information

_No response_

tarekgh:
Fixes #92327

This change addresses issues that arise when using the Options source generator in AOT builds. The trimming linker generates warnings due to the use of Validation attributes that rely on reflection. Additionally, the source generator emits code that uses `ValidationContext`, which has constructors annotated with the `RequiredUnreferencedCode` attribute, leading to linker warnings. It's important to note that the `ValidationContext` object created in this context is exclusively used within the `Validator.TryValidateValue` API call.

The solution consists of two parts:

- We generate code to create the `ValidationContext` object and initialize the display and member names within it. This ensures the object's safe use because the reflection code path is not executed during this time. Consequently, we suppress the trimming linker warnings associated with this part.
- For attributes such as `MaxLengthAttribute`, `MinLengthAttribute`, `LengthAttribute`, `CompareAttribute`, and `RangeAttribute`, which rely on reflection, we generate replacement attributes. These replacement attributes are strongly typed and entirely avoid the use of reflection.

These changes address the issues and improve the overall safety and performance of the code in AOT buildings.


src/libraries/Microsoft.Extensions.Options/gen/Parser.cs
```diff
@@ -467,7 +470,81 @@ private List<ValidatedMember> GetMembersToValidate(ITypeSymbol modelType, bool s
...
+                        }
+                    }
+                    else if (SymbolEqualityComparer.Default.Equals(attributeType, _symbolHolder.MinLengthAttributeSymbol))
+                    {
+                        if (!CacheAttributesData(memberType, _symbolHolder.MinLengthAttributeSymbol.Name, location, attributeType, ref attributeFullQualifiedName))
+                        {
+                            continue;
+                        }
+                    }
+                    else if (_symbolHolder.MinLengthAttributeSymbol is not null &&
```
jeffhandley, line 488: Only this symbol has the `_symbolHolder... is not null` check. Is that intentional?

src/libraries/Microsoft.Extensions.Options/gen/Parser.cs
```diff
@@ -467,7 +470,81 @@ private List<ValidatedMember> GetMembersToValidate(ITypeSymbol modelType, bool s
...
+                    }
+                    else if (SymbolEqualityComparer.Default.Equals(attributeType, _symbolHolder.CompareAttributeSymbol))
+                    {
+                        ImmutableArray<IParameterSymbol> constructorParameters = attribute.AttributeConstructor?.Parameters ?? ImmutableArray<IParameterSymbol>.Empty;
+                        if (constructorParameters.Length == 1 && constructorParameters[0].Name == "otherProperty" && constructorParameters[0].Type.SpecialType == SpecialType.System_String)
+                        {
+                            _commonUtilities.EnsureTrackingAttribute(_symbolHolder.CompareAttributeSymbol.Name, createValue: true, out HashSet<object>? trackedTypeList);
+                            trackedTypeList!.Add((modelType.ToDisplayString(SymbolDisplayFormat.FullyQualifiedFormat), (string)attribute.ConstructorArguments[0].Value!));
+                            attributeFullQualifiedName = $"{Emitter.StaticGeneratedValidationAttributesClassesNamespace}.{Emitter.StaticAttributeClassNamePrefix}{_commonUtilities.Suffix}_{_symbolHolder.CompareAttributeSymbol.Name}";
+                        }
```
jeffhandley, line 504: Suggestion: Refactor this out to a `TrackCompareAttributeIfNeeded` (or similar) helper method that contains the logic for tracking the attribute if needed.

src/libraries/Microsoft.Extensions.Options/gen/Parser.cs
```diff
@@ -467,7 +470,81 @@ private List<ValidatedMember> GetMembersToValidate(ITypeSymbol modelType, bool s
...
+                            typeSymbol = ((INamedTypeSymbol)typeSymbol).TypeArguments[0];
+                        }
+
+                        if (argumentSpecialType != SpecialType.None &&
+                            CommonUtilities.IsConvertibleBasicType(typeSymbol) &&
+                            (constructorParameters.Length != 3 || typeSymbol.SpecialType == argumentSpecialType)) // When type is provided as a parameter, it has to match the property type.
+                        {
+                            _commonUtilities.EnsureTrackingAttribute(_symbolHolder.RangeAttributeSymbol.Name, createValue: false, out HashSet<object>? trackedTypeList);
+                            attributeFullQualifiedName = $"{Emitter.StaticGeneratedValidationAttributesClassesNamespace}.{Emitter.StaticAttributeClassNamePrefix}{_commonUtilities.Suffix}_{_symbolHolder.RangeAttributeSymbol.Name}";
+                        }
```
jeffhandley, line 544: Suggestion: Refactor this out to a `TrackRangeAttributeIfNeeded` (or similar) helper method that contains the logic for tracking the attribute if needed.

src/libraries/Microsoft.Extensions.Options/gen/Parser.cs
```diff
@@ -567,6 +644,28 @@ private List<ValidatedMember> GetMembersToValidate(ITypeSymbol modelType, bool s
             return null;
         }
 
+        private bool CacheAttributesData(ITypeSymbol memberType, string attributeName, Location location, ITypeSymbol attributeType, ref string attributeFullQualifiedName)
```
jeffhandley, line 647: Recommendation: The name of this method doesn't tell me what the return value means. I think the method's purpose is something along the lines of: `AttributeIsTrackedForSubstitution` or something like that. A `true` return value would make the `continue` statements above clearer too, such that "if this attribute needs to be substituted, I'll skip over it."

You could also move the "if MaxLength or MinLength or Length" logic into this method and then rename it to `TrackLengthBasedAttributeIfNeeded` or something like that. With that if/else logic outside this method, it's a bit repetitive at the call site, and there's an implicit contract that this method will only be called for one of those 3 attributes.

src/libraries/Microsoft.Extensions.Options/tests/SourceGeneration.Unit.Tests/Baselines/DataAnnotationAttributesWithParams.g.cs
```diff
@@ -0,0 +1,137 @@
...
+            {
+                length = stringValue.Length;
+            }
+            else if (value is System.Collections.ICollection collectionValue)
+            {
+                length = collectionValue.Count;
+            }
+            else
+            {
+                throw new global::System.InvalidCastException($"The field of type {value.GetType()} must be a string, array or ICollection type.");
```
jeffhandley, line 130: Nit (Oxford comma)... although it looks like this might be preexisting.

```suggestion
                throw new global::System.InvalidCastException($"The field of type {value.GetType()} must be a string, array, or ICollection type.");
```

src/libraries/Microsoft.Extensions.Options/tests/SourceGeneration.Unit.Tests/Baselines/DataAnnotationAttributesWithParams.g.cs
```diff
@@ -0,0 +1,137 @@
...
+    {
+    }
+}
+namespace __OptionValidationGeneratedAttributes
+{
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Extensions.Options.SourceGeneration", "42.42.42.42")]
+    [global::System.AttributeUsage(global::System.AttributeTargets.Property | global::System.AttributeTargets.Field | global::System.AttributeTargets.Parameter, AllowMultiple = false)]
+    file class __SourceGen__LengthAttribute : global::System.ComponentModel.DataAnnotations.ValidationAttribute
+    {
+        private static string DefaultErrorMessageString => "The field {0} must be a string or collection type with a minimum length of '{1}' and maximum length of '{2}'.";
```
jeffhandley, line 100: There's some inconsistency on these messages. Some say, "string, array or ICollection type," while others say, "string or collection type."

tarekgh: I picked up the message from the defined attribute resources to match the runtime behavior.

jeffhandley:
It's kind of a bummer that we have to special-case the built-in validators to pull this off, but it makes sense for .NET 8. This will address the vast majority of instances where AOT would have been blocked for customers. If we hear feedback about ecosystem or customer attributes that need this same type of substitution treatment, we can explore approaches for a general-purpose substitution.

src/libraries/Microsoft.Extensions.Options/tests/SourceGeneration.Unit.Tests/Baselines/DataAnnotationAttributesWithParams.g.cs
jeffhandley: All right; we must have had the inconsistencies before then. If I understand correctly, the source generator will now bake in the English message without support for localization. That's also fine.

src/libraries/Microsoft.Extensions.Options/gen/Parser.cs
tarekgh: `LengthAttribute` is introduced in 8.0 so if we are generating code on down-level, its symbol can be null. This is why I am making it optional existence in the symbol holder. But I had typo you caught as I was checking MinLengthAttributeSymbol instead. most likely I did that during some refactoring I was doing before submitting the PR. Thanks for the catch!

src/libraries/Microsoft.Extensions.Options/gen/CommonUtilities.cs
```diff
@@ -0,0 +1,83 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+
+namespace Microsoft.Extensions.Options.Generators
+{
+    internal sealed class CommonUtilities
```
eiriktsarpalis, line 13: Nit: Most "utilities" classes I'm aware of are typically collections of static helper methods. Perhaps it could be named differently to convey that it's an instance containing state?

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
```diff
@@ -209,10 +604,17 @@ private void GenModelSelfValidationIfNecessary(ValidatedModel modelToValidate)
             OutLn($"/// <returns>Validation result.</returns>");
             OutGeneratedCodeAttribute();
 
+            // We disable the warning on `new ValidationContext(object)` usage as we use it in a safe way that not require executing the reflection code.
+            // This is done by initializing the DisplayName in the context which is the part trigger reflection if it is not initialized.
+            OutLn($"#if NET5_0_OR_GREATER // UnconditionalSuppressMessage is supported on .NET 5.0 and later");
```
eiriktsarpalis, line 609: Is it not possible to read the target framework from the `Compilation` object and emit the suppression conditionally? I would assume this should work even in multi-targeted projects since the source generator runs independently for every target.

eiriktsarpalis (commented):
I'm guessing this will be backported to 8?

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
tarekgh: I am thinking of the library author scenario who produce library which is be used later in AOT app. what do you think about that?

tarekgh:
> I'm guessing this will be backported to 8?

Yes.

src/libraries/Microsoft.Extensions.Options/gen/Parser.cs
tarekgh: I have done the refactoring and renaming. Please let me know if you have any more comments about that. Thanks!

src/libraries/Microsoft.Extensions.Options/tests/SourceGeneration.Unit.Tests/Baselines/DataAnnotationAttributesWithParams.g.cs
tarekgh: > If I understand correctly, the source generator will now bake in the English message without support for localization.

That is right. But note currently we don't localize the libraries resources in net core. So, the behavior will be same between the runtime and the source gen for now :-)

jeffhandley (approved):
Looks good to me, save @eiriktsarpalis's question about the target framework handling.

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
tarekgh: I applied the suggestion

eiriktsarpalis (approved):

PR #93089, opened 2023-10-05
Fix a memory leak in runtime interop stubs when using an array of structs of types that use old-style managed marshalers

jkoritzinsky:
When we don't have a managed value in a ClearNative call, the managed pointer to the managed field with a null pointer instead of skipping the ClearNative call.

The initial approach was taken in https://github.com/dotnet/coreclr/pull/27562 to avoid bad GC pointers, but this approach also avoids bad GC pointers and correctly frees the memory.

Validated locally. I'm also going to validate offline with a 1st-party customer that reported this issue and then backport to 6.0-8.0.


jkotas:
Can we write a test for this?

jkoritzinsky:
I have an idea for one. I'll try it out tomorrow.

jkoritzinsky:
@jkotas I've constructed a test using the "Fixed Array" marshaller with a COM object to give me a testable lifetime hook. Sadly my initial idea of using `ICustomMarshaler` doesn't work as we disallow `ICustomMarshaler` marshalling on fields.

src/coreclr/vm/ilmarshalers.h
```diff
@@ -3138,39 +3138,33 @@ class ILMngdMarshaler : public ILMarshaler
     void EmitClearNative(ILCodeStream* pslILEmit) override
     {
         WRAPPER_NO_CONTRACT;
-        ILCodeLabel* pNoManagedValueLabel = nullptr;
         if (IsFieldMarshal(m_dwMarshalFlags))
         {
-            pNoManagedValueLabel = pslILEmit->NewCodeLabel();
+            ILCodeLabel* pHasManagedValueLabel = pslILEmit->NewCodeLabel();
```
jkotas, line 3143: I am not able to figure out why we need this at all. The managed home does not seem to be used by the clear native worker methods. Is it used later?

Asking differently: What is going to break if this whole `if (IsFieldMarshal(m_dwMarshalFlags))` block is deleted?

jkoritzinsky: We had a problem in checked CoreCLR builds and CoreCLR builds with GCStress where (without this block), the OBJECTREF that was created for the interior pointer to a null managed value would cause an AV. For example, let's take the following struct:

```csharp
struct S
{
	public string s;
	public object[] arr;
}
```

For this struct, we'd end up passing a byref to a pointer to `arr` as the managed value to the underlying managed marshaller implementation. For marshallers that are implemented in native code, we'd end up creating an OBJECTREF around it. In a checked CoreCLR build with GCStress, we would then try to validate the OBJECTREF. This would fail with an AV for null managed struct pointers, as the pointer to `arr` is 4 or 8 (depending on the platform), and the OBJECTREF validation logic would try to dereference the pointer to validate the MethodTable and would crash. I believe the GC was also failing for the same reason when trying to scan the GCPROTECTED roots.

Without this fix, we avoid this crash by (erroneously) not calling the ClearNative or ClearNativeContents methods. With this fix, this block turns these "almost null" pointers into null pointers so the rest of the logic around null byrefs/OBJECTREFs works correctly.

Alternatively, we can update the marshallers to not take the managed value for the "clear native " step. I'm okay doing this in main (and was planning on doing that as a follow-up to this work), but I would prefer to do this targeted fix downlevel as I believe it has less risk.

jkotas: > For marshallers that are implemented in native code, we'd end up creating an OBJECTREF around it. In a checked CoreCLR build with GCStress, we would then try to validate the OBJECTREF.

Where does this happen? I see the affected ClearNative/ClearNativeContent methods 
 take `OBJECTREF*` argument in native code, but I do not see this argument used in any of the ClearNative/ClearNativeContent methods.

jkotas: > This would fail with an AV for null managed struct pointers, as the pointer to arr is 4 or 8 (depending on the platform),

Where exactly are these bogus byrefs created? This code should prevent us from creating the bogus byrefs - it is effectively duplicate of the normalization in this PR.
 https://github.com/dotnet/runtime/blob/f21a2666c577306e437f80fe934d76cdb15072a5/src/coreclr/vm/ilmarshalers.h#L998-L1004

AaronRobinsonMSFT: Agreed. @jkoritzinsky Let's chat offline I think both @jkotas and I are missing something.

src/tests/Interop/ArrayMarshalling/StructArray/MarshalStructArrayTest.cs
```diff
@@ -0,0 +1,105 @@
...
+    private static ComInterfaceEntry* CreateVtableEntry()
+    {
+        ComInterfaceEntry* entry = (ComInterfaceEntry*)NativeMemory.AllocZeroed((nuint)sizeof(ComInterfaceEntry));
+        entry->IID = Guid.NewGuid();
+        nint* vtable = (nint*)NativeMemory.Alloc((nuint)(sizeof(void*) * 4));
+        GetIUnknownImpl(out vtable[0], out vtable[1], out vtable[2]);
+        vtable[3] = (nint)(delegate* unmanaged<nint, int>)&NativeMethodImpl;
+        entry->Vtable = (nint)vtable;
+        return entry;
+    }
```
AaronRobinsonMSFT, line 82: I don't think this is needed. We can return a `null` vtable with length `0`.

src/tests/Interop/ArrayMarshalling/StructArray/MarshalStructArrayTest.cs
```diff
@@ -0,0 +1,105 @@
...
+            var (weakRef, obj) = CreateComObject();
+            StructWithObjectArrayField[] arr = [ new() { objs = [obj] } ];
+            MarshalStructArrayNative.ForwardToCallback(arr, &Callback);
+            return (weakRef, new WeakReference(obj));
+        }
+
+        [UnmanagedCallersOnly]
+        static void Callback(void* arg)
+        {
+        }
```
AaronRobinsonMSFT, line 41: I'm assume we can assert `arg` is non-null?

src/coreclr/vm/ilmarshalers.h
jkoritzinsky: Looking at this more, you are both correct. We can just remove this block. I'll do that and re-validate the fix.

src/tests/Interop/ArrayMarshalling/StructArray/MarshalStructArrayTest.cs
AaronRobinsonMSFT: You can return `null` if the count is `0`.

https://github.com/dotnet/runtime/blob/6296aac7e05b7db72e1cea58a7d59ed65cb2f815/src/coreclr/vm/interoplibinterface_comwrappers.cpp#L722-L726

src/tests/Interop/ArrayMarshalling/StructArray/MarshalStructArrayTest.cs
```diff
@@ -0,0 +1,92 @@
...
+    protected override void ReleaseObjects(IEnumerable objects)
+    {
+        throw new NotImplementedException();
+    }
+    
+    [UnmanagedCallersOnly]
+    static int NativeMethodImpl(nint thisPtr)
+    {
+        return 0;
+    }
```
AaronRobinsonMSFT, line 91:
```suggestion
```

AaronRobinsonMSFT:
@jkoritzinsky Can you also please run this test a dozen times or so on a Checked build with `DOTNET_GCStress=C`?

jkoritzinsky:
Validated that the test passes with GCStress=C.

src/tests/Interop/ArrayMarshalling/StructArray/MarshalStructArrayTest.cs
```diff
@@ -0,0 +1,86 @@
+using System;
```
jkotas, line 1: License header

src/tests/Interop/ArrayMarshalling/StructArray/MarshalStructArrayTest.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        // Try to GC 10 times until the built-in RCW is collected.
+        for (int i = 0; i < 10 && wrapperWeakRef.IsAlive; i++)
+        {
+            GC.Collect();
+        }
+
+        Assert.False(wrapperWeakRef.IsAlive);
+
+        // Run two more GCs, one to collect the ComWrappers MOW and one to collect the underlying weak reference.
```
jkotas, line 24: Tests that depend on object getting collected are often unreliable.

It may be better to directly construct IUnknown vtable with just Release method filled in and verify that the Release method was called.

src/tests/Interop/ArrayMarshalling/StructArray/MarshalStructArrayNative.cpp
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+//
+
+#include <xplatform.h>
+#include <platformdefines.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+extern "C" DLL_EXPORT void STDMETHODCALLTYPE ForwardToCallback(void* value, void(STDMETHODCALLTYPE* callback)(void*))
```
jkotas, line 10: Would `Marshal.DestroyStructucture` hit the buggy path? `Marshal.DestroyStructucture` may be simpler and more direct way to test it.

jkotas (approved):

src/tests/Interop/ArrayMarshalling/StructArray/MarshalStructArrayNative.cpp
jkoritzinsky: I'll check and see if I can do it with that instead. That would be significantly cleaner.

src/tests/Interop/ArrayMarshalling/StructArray/MarshalStructArrayTest.cs
jkoritzinsky: We'll still need to rely on the GC above for the built-in RCW, but I should be able to reduce the complexity here.

jkoritzinsky: I might be able to do this without any GC requirements. I'll see what I can do.

src/tests/Interop/ArrayMarshalling/StructArray/MarshalStructArrayNative.cpp
jkotas, line 10: Delete - no longer needed?

src/tests/Interop/ArrayMarshalling/StructArray/MarshalStructArrayTest.cs
```diff
@@ -0,0 +1,102 @@
...
+public struct StructWithObjectArrayField
+{
+    [MarshalAs(UnmanagedType.ByValArray, SizeConst = 1)]
+    public object[] objs;
+}
+
+public static unsafe class MarshalStructArrayNative
+{
+    [DllImport(nameof(MarshalStructArrayNative))]
+    public static extern void ForwardToCallback(StructWithObjectArrayField[] arr, delegate* unmanaged<void*, void> cb);
```
jkotas, line 57: Dtto

PR #93090, opened 2023-10-06
Handle binding to initialized abstract members

### LINKED ISSUE #92137: ConfigurationBinder source generator throws on an abstract member rather than binding to existing value
### Description

When encountering an abstract type in an object graph the configuration generator tries to create it and throws.

This is inconsistent with the runtime binder, which will use an existing instance.

I wonder if we should warn on this or not.  Definitely if the abstract type is the root *and* we're generating `Get` calls that's a problem.  Anywhere else it could be supported if the member is already initialized - since we won't need to create one in the binder.

### Reproduction Steps

Run attached repro project:
[bindAbstract.zip](https://github.com/dotnet/runtime/files/12622040/bindAbstract.zip)

Also shown here: https://gist.github.com/ericstj/121fe52f9a1b5115b246c08605c685c0

To see expected behavior remove `EnableConfigurationBindingGenerator`.  You'll see the output
```
options.Abstract.GetHashCode() before bind: 55915408
options.Abstract.GetHashCode() after bind: 55915408
John Doe
```
Note that the runtime binder didn't rely on seeing the derived type and creating a new one -- this is apparent by seeing that the object is the same before and after then bind.

### Expected behavior

Repro runs successfully and sets the members on the existing instance of the derived type.  Not creating a new instance nor throwing.

### Actual behavior

Warning emitted by the source generator:
```
C:\scratch\bindAbstract\Program.cs(11,1): warning SYSLIB1100: Cannot create instance of type 'MyAbstractClass' because
it is missing a public instance constructor. (https://learn.microsoft.com/dotnet/fundamentals/syslib-diagnostics/syslib
1100.md)
```

And exception thrown at runtime:
```
Unhandled exception. System.InvalidOperationException: Cannot create instance of type '<global namespace>.MyAbstractClass' because it is missing a public instance constructor.
   at Microsoft.Extensions.Configuration.Binder.SourceGeneration.<BindingExtensions_g>FC276FD7C89D1ED0B2B3E96BAC8F157FABB86A642EAABC54639B357A857C01D08__BindingExtensions.BindCore(IConfiguration configuration, MyOptions& instance, Boolean defaultValueIfNotFound, BinderOptions binderOptions) in C:\scratch\bindAbstract\Microsoft.Extensions.Configuration.Binder.SourceGeneration\Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator\BindingExtensions.g.cs:line 144
   at Microsoft.Extensions.Configuration.Binder.SourceGeneration.<BindingExtensions_g>FC276FD7C89D1ED0B2B3E96BAC8F157FABB86A642EAABC54639B357A857C01D08__BindingExtensions.Bind_MyOptions(IConfiguration configuration, Object instance) in C:\scratch\bindAbstract\Microsoft.Extensions.Configuration.Binder.SourceGeneration\Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator\BindingExtensions.g.cs:line 48
   at Program.<Main>$(String[] args) in C:\scratch\bindAbstract\Program.cs:line 11
```

### Regression?

_No response_

### Known Workarounds

Disable generator.

### Configuration

_No response_

### Other information

_No response_

ericstj:
Fixes https://github.com/dotnet/runtime/issues/92137

If an abstract member is encountered in an object graph we should only throw if it's not already initialized.

### DIFF
--- a/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
+++ b/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
@@ -1047,15 +1047,13 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
                     else
                     {
                         Debug.Assert(!type.IsValueType);
-
+                        EmitStartBlock($"if ({instanceToBindExpr} is not null)");
+                        EmitBindCoreCall();
+                        EmitEndBlock();
                         if (type is ObjectSpec { InitExceptionMessage: string exMsg })
                         {
+                            EmitStartBlock("else");
                             _writer.WriteLine($@"throw new {Identifier.InvalidOperationException}(""{exMsg}"");");
-                        }
-                        else
-                        {
-                            EmitStartBlock($"if ({instanceToBindExpr} is not null)");
-                            EmitBindCoreCall();
                             EmitEndBlock();
                         }
                     }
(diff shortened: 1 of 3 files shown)


tarekgh (approved):
LGTM!

PR #93092, opened 2023-10-06
[mono] Enable branch optimizations when using LLVM.

vargaz:
Doing these optimizations early enables the removal of some dead code which decreases the amount of generic instances the aot compiler generates since the aot compiler will no longer encounter the methods called in the dead code.

### DIFF
--- a/src/mono/mono/mini/mini-codegen.c
+++ b/src/mono/mono/mini/mini-codegen.c
@@ -2165,8 +2165,9 @@ mono_local_regalloc (MonoCompile *cfg, MonoBasicBlock *bb)
 
 MONO_RESTORE_WARNING
 
+/* Returns -1 if opcode is not a conditional */
 CompRelation
-mono_opcode_to_cond (int opcode)
+mono_opcode_to_cond_unchecked (int opcode)
 {
 	switch (opcode) {
 	case OP_CEQ:
@@ -2286,10 +2287,21 @@ mono_opcode_to_cond (int opcode)
 	case OP_CMOV_LGT_UN:
 		return CMP_GT_UN;
 	default:
+		return (CompRelation)-1;
+	}
+}
+
+CompRelation
+mono_opcode_to_cond (int opcode)
+{
+	CompRelation rel = mono_opcode_to_cond_unchecked (opcode);
+
+	if (rel == (CompRelation)-1) {
 		printf ("%s\n", mono_inst_name (opcode));
 		g_assert_not_reached ();
 		return (CompRelation)0;
 	}
+	return rel;
 }
 
 CompRelation
--- a/src/mono/mono/mini/mini-llvm.c
+++ b/src/mono/mono/mini/mini-llvm.c
@@ -6145,7 +6145,10 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
 				/* The comparison result is not needed */
 				continue;
 
-			rel = mono_opcode_to_cond (ins->next->opcode);
+			rel = mono_opcode_to_cond_unchecked (ins->next->opcode);
+			if (rel == (CompRelation)-1)
+				/* The following branch etc. was optimized away */
+				continue;
 
 			if (ins->opcode == OP_ICOMPARE_IMM) {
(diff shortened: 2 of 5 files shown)


vargaz:
This saves about 100k on a wasm hello world.

vargaz:
Failures are unrelated.

src/mono/mono/mini/mini-codegen.c
```diff
@@ -2166,7 +2166,7 @@ mono_local_regalloc (MonoCompile *cfg, MonoBasicBlock *bb)
 MONO_RESTORE_WARNING
 
 CompRelation
-mono_opcode_to_cond (int opcode)
+mono_try_opcode_to_cond (int opcode)
```
fanyang-mono, line 2169: The method names is a little confusing to me. I feel these ones are better:
`mono_try_opcode_to_cond` -> 'mono_opcode_to_cond'
`mono_opcode_to_cond` -> `mono_opcode_to_cond_with check`

src/mono/mono/mini/mini-codegen.c
```diff
@@ -2286,10 +2286,21 @@ mono_opcode_to_cond (int opcode)
...
+
+CompRelation
+mono_opcode_to_cond (int opcode)
+{
+	CompRelation rel = mono_try_opcode_to_cond (opcode);
+
+	if (rel == (CompRelation)-1) {
 		printf ("%s\n", mono_inst_name (opcode));
 		g_assert_not_reached ();
 		return (CompRelation)0;
```
fanyang-mono, line 2301: This lane will never be reached.

vargaz: Some compilers can't detect that and they will error out.

src/mono/mono/mini/mini-llvm.c
```diff
@@ -12961,8 +12963,15 @@ emit_method_inner (EmitContext *ctx)
 		bb = (MonoBasicBlock*)g_ptr_array_index (bblock_list, bb_index);
 
 		// Prune unreachable mono BBs.
-		if (!(bb == cfg->bb_entry || bb->in_count > 0))
+		if (!(bb == cfg->bb_entry || bb->in_count > 0)) {
+			LLVMBasicBlockRef target_bb = ctx->bblocks [bb->block_num].call_handler_target_bb;
+			if (target_bb) {
+				LLVMBuilderRef target_builder = create_builder (ctx);
+				LLVMPositionBuilderAtEnd (target_builder, target_bb);
+				LLVMBuildUnreachable (target_builder);
```
lambdageek, line 12971: We always have unique `call_handler_target_bb`s? they dont' get shared?

Also at this point in emit we didn't write any other code into `target_bb`, right? so the unreachable will be the first thing in there?

vargaz: If target_bb is dead, then nothing is going to be added to call_handler_bb, so it will become invalid IR. This didn't happen before because the branch opt pass was not enabled. Will add a comment.

src/mono/mono/mini/mini-codegen.c
lambdageek: nit: Now that we're C11 we can probably redefine `G_GNUC_NORETURN` as

```c
#if __STDC_VERSION__ >= 202311L
#define G_GNUC_NORETURN [[noreturn]]
#else
#define G_GNUC_NORETURN _Noreturn
#endif
```

and then all compilers will know that `mono_assertion_message_unreachable` doesn't return

src/mono/mono/mini/mini-codegen.c
vargaz: Renamed the try version to unchecked, didn't want to rename all callers of the old one.

src/mono/mono/mini/mini-codegen.c
lambdageek: https://github.com/dotnet/runtime/pull/93131

lambdageek (approved):

vargaz:
Failures are unrelated.

PR #93093, opened 2023-10-06
Vectorize TensorPrimitives.Tanh/Cosh/Sinh

stephentoub:
Tanh and Cosh are based on AOCL-LibM.

AOCL-LibM doesn't appear to have a sinh implementation, so this Sinh is just based on the sinh formula based on exp(x).

I also augmented the tests further, including:
- Added more tests for sinh/cosh/tanh
- Add an equality routine that supports comparing larger values with a tolerance
- Tightened the tolerance for most functions
- Changed some tests to be theories to be consistent with style elsewhere in the tests
- Fixed some use of Math to be MathF


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2992,6 +2992,183 @@ public static Vector512<float> Invoke(Vector512<float> x)
 #endif
         }
 
+        /// <summary>MathF.Sinh(x)</summary>
+        private readonly struct SinhOperator : IUnaryOperator
+        {
+            public static float Invoke(float x) => MathF.Sinh(x);
+            public static Vector128<float> Invoke(Vector128<float> x) => (ExpOperator.Invoke(x) - ExpOperator.Invoke(-x)) / Vector128.Create(2f);
```
stephentoub, line 2999: Is there anything better to do than the standard formula? I didn't see a sinh implementation in the amd code.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2992,6 +2992,183 @@ public static Vector512<float> Invoke(Vector512<float> x)
...
+            public static Vector128<float> Invoke(Vector128<float> x)
+            {
+                Vector128<uint> ux = x.AsUInt32() & Vector128.Create(SIGN_MASK);
+                if (Vector128.GreaterThanAny(ux, Vector128.Create(ARG_MAX)))
+                {
+                    return Vector128.Create(
+                        MathF.Cosh(x.GetElement(0)),
+                        MathF.Cosh(x.GetElement(1)),
+                        MathF.Cosh(x.GetElement(2)),
+                        MathF.Cosh(x.GetElement(3)));
```
stephentoub, line 3047: The amd implementation falls back to scalar above the threshold. Do we want to try to do better, e.g. use conditional selects to implement the scalar handling of these outliers?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2992,6 +2992,183 @@ public static Vector512<float> Invoke(Vector512<float> x)
...
+
+            public static float Invoke(float x) => MathF.Tanh(x);
+
+            public static Vector128<float> Invoke(Vector128<float> x)
+            {
+                Vector128<uint> ux = x.AsUInt32();
+                Vector128<uint> sign = ux & Vector128.Create(~V4_TANHF_SIGN_MASK);
+
+                ux &= Vector128.Create(V4_TANHF_SIGN_MASK);
+                if (Vector128.GreaterThanAny(ux, Vector128.Create(V4_TANHF_ARG_MAX)))
```
stephentoub, line 3117: The amd implementation looks wrong here. From my reading, it's falling back to the scalar implementation when it's less than or equal to the maximum rather than when it's greater than the maximum (it has a comment that says it's for when it's greater). I assume greater is correct here?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2992,6 +2992,183 @@ public static Vector512<float> Invoke(Vector512<float> x)
...
+                {
+                    return Vector128.Create(
+                        MathF.Tanh(x.GetElement(0)),
+                        MathF.Tanh(x.GetElement(1)),
+                        MathF.Tanh(x.GetElement(2)),
+                        MathF.Tanh(x.GetElement(3)));
+                }
+
+                Vector128<float> y = ux.AsSingle();
+                Vector128<float> z = ExpOperator.Invoke(Vector128.Create(-2f) * y) - Vector128.Create(1f);
```
stephentoub, line 3127: The amd implementation does `-1f` but it has a comment saying it's doing `+1f`. I'm assuming the `-1f` is correct.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2992,6 +2992,183 @@ public static Vector512<float> Invoke(Vector512<float> x)
...
+                Vector128<uint> sign = ux & Vector128.Create(~V4_TANHF_SIGN_MASK);
+
+                ux &= Vector128.Create(V4_TANHF_SIGN_MASK);
+                if (Vector128.GreaterThanAny(ux, Vector128.Create(V4_TANHF_ARG_MAX)))
+                {
+                    return Vector128.Create(
+                        MathF.Tanh(x.GetElement(0)),
+                        MathF.Tanh(x.GetElement(1)),
+                        MathF.Tanh(x.GetElement(2)),
+                        MathF.Tanh(x.GetElement(3)));
```
stephentoub, line 3123: As with cosh, the amd implementation has a scalar fallback here. Do we want to try to do better? The maximum value is ~8; with common usage of Tanh, will we expect values > 8 (or < -8) to be common? That's appx where tanh levels off at 1 / -1.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2992,6 +2992,183 @@ public static Vector512<float> Invoke(Vector512<float> x)
...
+                return Vector512.Create(HALFV).AsSingle() * (z + Vector512.Create(INVV2).AsSingle() * 1f / z);
+            }
+#endif
+        }
+
+        /// <summary>MathF.Tanh(x)</summary>
+        private readonly struct TanhOperator : IUnaryOperator
+        {
+            // This code is based on `vrs4_tanhf` from amd/aocl-libm-ose
+            // Copyright (C) 2008-2022 Advanced Micro Devices, Inc. All rights reserved.
```
stephentoub, line 3088: The copyright for LogOperator below:
https://github.com/dotnet/runtime/pull/93093/files#diff-49f5fe82910e47b0f111b2a18f248ac3ee5e2d3d1b113deea0a8c87907ace9adR3176
says 2018-2019... should all of these be updated?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
lilinus: Don't know about other implementations but exp(x)-exp(y) is not good numerically when x and y are similar (around 0).

Better way to calculate would perhaps be the form: sinh(x) = (exp(2*x) - 1) / (2 * exp(x)) where we use expm1 for the numerator (if it exists as an operator here).

In other words expm1(2*x) / (2*x);

tannergooding: `aocl-libm` doesn't have a vectorized version, but does have a scalar version which could be vectorized: https://github.com/amd/aocl-libm-ose/blob/aocl-4.1/src/optimized/sinhf.c

-- Notably, the scalar version of coshf looks more accurate, where-as the vectorized version looks like it estimates a little bit more for speed. The branches are relegated to special inputs (like NaN) which should be rare, and the handling of very large/small values (which should also be rare). We should be able to, in general, isolate those "special cases" to a single branch so the core stays fast for the main cases and the `ConditionalSelect` is rarely ever needed (much as we've done for other algorithms so far).

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: I think that would be desirable. The outliers are expected to be rare, but they also aren't terrible complicated to handle. It's largely just a set of 2-4 `ConditionalSelects` that we can avoid for most inputs.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: Potentially? The root license for the entire repo is 2008-2022.

It looks like the per file headers are then mostly when started vs last touched within that 14 year range.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: Yes. This looks like a bug

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: I wouldn't expect them to be common, but it's a case we can support with Vectorization without it hurting performance by having the branch so the additional logic + masking is predicted not taken and not executed unless required.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: -1 is correct.

The algorithm is $`{\frac {e^{2x} - 1}{e^{2x} + 1}}`$

It is then breaking this down into $`{\frac {-(e^{2|x|} - 1)}{({e^{2|x|} + 1) + 2}}}`$, which functionally computes `-tanh(x)`. It then toggles the sign of `x`, giving it the correct value.

Not quite sure why they did it this way since it could've been done the "normal" way without the need to toggle and without the need to `(z -1) + 2`. -- There's theoretically some very small rounding difference, but it doesn't look like something they're explicitly accounting for here.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2992,6 +2993,163 @@ public static Vector512<float> Invoke(Vector512<float> x)
...
+            private const uint SIGN_MASK = 0x7FFFFFFF;
+            private const uint LOGV = 0x3f317300;
+            private const uint HALFV = 0x3f800074;
+            private const uint INVV2 = 0x3e7ffe30;
+
+            public static float Invoke(float x) => MathF.Cosh(x);
+
+            public static Vector128<float> Invoke(Vector128<float> x)
+            {
+                Vector128<float> y = (x.AsUInt32() & Vector128.Create(SIGN_MASK)).AsSingle();
```
tannergooding, line 3029:
```suggestion
                Vector128<float> y = Vector128.Abs(x);
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2992,6 +2993,163 @@ public static Vector512<float> Invoke(Vector512<float> x)
...
+            // if ( asint(x) > infinity) return x with overflow exception and
+            // return x.
+            // if x is NaN then raise invalid FP operation exception and return x.
+            //
+            // coshf = v/2 * exp(x - log(v)) where v = 0x1.0000e8p-1
+
+            private const uint SIGN_MASK = 0x7FFFFFFF;
+            private const uint LOGV = 0x3f317300;
+            private const uint HALFV = 0x3f800074;
+            private const uint INVV2 = 0x3e7ffe30;
```
tannergooding, line 3023: nit: I slightly prefer seeing the floating-point value involved, since its not a bitwise manipulation:
```suggestion
            private const float LOGV = 0.693161f;
            private const float HALFV = 1.0000138f;
            private const float INVV2 = 0.24999309f;
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2992,6 +2993,163 @@ public static Vector512<float> Invoke(Vector512<float> x)
...
+
+            public static float Invoke(float x) => MathF.Sinh(x);
+
+            public static Vector128<float> Invoke(Vector128<float> x)
+            {
+                Vector128<uint> ux = x.AsUInt32();
+                Vector128<uint> sign = ux & Vector128.Create(~SIGN_MASK);
+                Vector128<float> y = (ux & Vector128.Create(SIGN_MASK)).AsSingle();
```
tannergooding, line 3068: Same suggestion, so it performs better on Arm64:
```suggestion
                Vector128<float> y = Vector128.Abs(x);
```

`sign` is probably fine as is, but if you move it down next to where it's needed in the `xor`, some hardware can fold it down to a combined instruction that does `(x & y) ^ z`

tannergooding (approved):

PR #93098, opened 2023-10-06
[RISC-V] Use standard NaNs in FP ops for cross-compiling RISC-V on x86

tomeksowi:
Use positive NaNs on RISC-V for FP ops crossgen-evaluated when host is x86.

Split from #92934

Part of #84834


tomeksowi:
Build failures seem to be caused by problems with remote downloads, not the code change itself.

tomeksowi:
@jakobbotsch PR ready for review

jakobbotsch (approved):

PR #93101, opened 2023-10-06
Fix "warning : Reference to undeclared extern assembly 'mscorlib'" in tst.il

akoeplinger:
The test was missing the reference.


jkoritzinsky (approved):

trylek (approved):
Thanks for fixing this!

PR #93115, opened 2023-10-06
Avoid a few array allocations with DefaultBinder

stephentoub:
In places where we're not handing out a `T[]` and thus actually need a `T[]`, we can instead just get a span from the `ListBuilder<T>` being used.


src/libraries/System.Private.CoreLib/src/System/RuntimeType.cs
```diff
@@ -967,5 +967,102 @@ private CheckValueStatus TryChangeType(ref object? value, ref bool copyBack)
 
             return CheckValueStatus.ArgumentException;
         }
+
+        // Helper to build lists of MemberInfos. Special cased to avoid allocations for lists of one element.
```
jkotas, line 971: Move this into a separate .cs file and de-dup it with the nativeaot copy?

stephentoub: Sure. Do we want to keep the nativeaot's DEBUG-only logic?

jkotas: Either way is fine with me.

stephentoub: Subsequently we might also want to dedup ListBuilder and ValueListBuilder. The main thing the former provides is a non-allocating slot for a single entry, which ValueListBuilder can also achieve via the scratch span passed in, and that can be created now via an inlinearray, a collection expression that itself uses an inlinearray, or a span wrapping a single local.

jkotas (approved):
Thanks

PR #93116, opened 2023-10-06
Optimize Ordinal.EqualsIgnoreCase_Vector with AVX2 and AVX512

### LINKED ISSUE #92497: Add AVX/AVX512 to Ordinal.EqualsIgnoreCase
`Ordinal.EqualsIgnoreCase` currently has a SIMDified path only for `Vector128`: `EqualsIgnoreCase_Vector128` and since we're observing this call in traces on hot path for one of our 1P customers, it should makes sense to add AVX and AVX2 here. For that, we need to re-use existing `Ascii.EqualsIgnoreCase` API that already does it, the problem that that API gives up on on non-ASCII input while we want to know the exact position where we gave up in order to switch to the slow fallback - we need to modify that API to return it.

image

EgorBo:
Closes https://github.com/dotnet/runtime/issues/92497

Currently, `str1.Equals(str2, StringComparison.OrdinalIgnoreCase)` is only accelerated with `Vector128`, this PR adds `Vector256` and `Vector512` via `ISimdVector` abstraction - the best part that there is no code duplication.

```cs
public class Benchmarks
{
    static readonly string s1 = @"C:\prj\runtime\src\coreclr\System.Private.CoreLib\System.Private.CoreLib.csproj";
    static readonly string s2 = @"C:\prj\runtime\src\coreclr\System.Private.CoreLib\system.Private.CoreLib.csproj";

    [Benchmark]
    public bool EqualsIgnoreCase()
    {
        return s1.Equals(s2, StringComparison.OrdinalIgnoreCase);
    }
}
```

| Method           | Toolchain             | Mean      | Error    | StdDev    | Ratio |
|----------------- |---------------------- |----------:|---------:|----------:|------:|
| EqualsIgnoreCase | Core_Root\corerun.exe | 10.610 ns | 0.1701 ns | 0.1591 ns |  1.00 |
| EqualsIgnoreCase | Core_Root_PR\corerun.exe |  **5.878 ns** | 0.0354 ns | 0.0313 ns |  0.55 |

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Globalization/Ordinal.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Globalization/Ordinal.cs
@@ -89,6 +89,9 @@ namespace System.Globalization
             nuint i = 0;
             TVector vec1;
             TVector vec2;
+            TVector loweringMask = TVector.Create(0x20);
+            TVector vecA = TVector.Create('a');
+            TVector vecZMinusA = TVector.Create('z' - 'a');
             do
             {
                 vec1 = TVector.LoadUnsafe(ref Unsafe.As<char, ushort>(ref charA), i);
@@ -99,9 +102,18 @@ namespace System.Globalization
                     goto NON_ASCII;
                 }
 
-                if (!Utf16Utility.VectorOrdinalIgnoreCaseAscii(vec1, vec2))
+                TVector notEquals = ~TVector.Equals(vec1, vec2);
+
+                if (notEquals != TVector.Zero)
                 {
-                    return false;
+                    // not exact match
+
+                    vec1 |= loweringMask;
+                    vec2 |= loweringMask;
+                    if (TVector.GreaterThanAny((vec1 - vecA) & notEquals, vecZMinusA) || vec1 != vec2)
+                    {
+                        return false; // first input isn't in [A-Za-z], and not exact match of lowered
+                    }
                 }
 
                 i += (nuint)TVector.Count;
--- a/src/libraries/System.Private.CoreLib/src/System/Globalization/Ordinal.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Globalization/Ordinal.cs
@@ -78,39 +78,40 @@ namespace System.Globalization
             return OrdinalCasing.CompareStringIgnoreCase(ref strA, lengthA, ref strB, lengthB);
         }
 
-        private static bool EqualsIgnoreCase_Vector128(ref char charA, ref char charB, int length)
+        private static bool EqualsIgnoreCase_Vector<TVector>(ref char charA, ref char charB, int length)
+            where TVector : struct, ISimdVector<TVector, ushort>
         {
-            Debug.Assert(length >= Vector128<ushort>.Count);
-            Debug.Assert(Vector128.IsHardwareAccelerated);
+            Debug.Assert(length >= TVector.Count);
+            Debug.Assert(TVector.IsHardwareAccelerated);
 
             nuint lengthU = (nuint)length;
-            nuint lengthToExamine = lengthU - (nuint)Vector128<ushort>.Count;
+            nuint lengthToExamine = lengthU - (nuint)TVector.Count;
             nuint i = 0;
-            Vector128<ushort> vec1;
-            Vector128<ushort> vec2;
+            TVector vec1;
+            TVector vec2;
             do
             {
-                vec1 = Vector128.LoadUnsafe(ref charA, i);
-                vec2 = Vector128.LoadUnsafe(ref charB, i);
+                vec1 = TVector.LoadUnsafe(ref Unsafe.As<char, ushort>(ref charA), i);
+                vec2 = TVector.LoadUnsafe(ref Unsafe.As<char, ushort>(ref charB), i);
 
-                if (!Utf16Utility.AllCharsInVector128AreAscii(vec1 | vec2))
+                if (!Utf16Utility.AllCharsInVectorAreAscii(vec1 | vec2))
                 {
                     goto NON_ASCII;
                 }
 
-                if (!Utf16Utility.Vector128OrdinalIgnoreCaseAscii(vec1, vec2))
+                if (!Utf16Utility.VectorOrdinalIgnoreCaseAscii(vec1, vec2))
                 {
                     return false;
                 }
 
-                i += (nuint)Vector128<ushort>.Count;
+                i += (nuint)TVector.Count;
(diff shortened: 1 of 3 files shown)


src/libraries/System.Private.CoreLib/src/System/Globalization/Ordinal.cs
```diff
@@ -78,39 +78,40 @@ internal static int CompareStringIgnoreCaseNonAscii(ref char strA, int lengthA,
...
                 }
 
-                if (!Utf16Utility.Vector128OrdinalIgnoreCaseAscii(vec1, vec2))
+                if (!Utf16Utility.VectorOrdinalIgnoreCaseAscii(vec1, vec2))
                 {
                     return false;
                 }
 
-                i += (nuint)Vector128<ushort>.Count;
+                i += (nuint)TVector.Count;
```
neon-sunset, line 107: This might noticeably regress comparisons for `length < TVector.Count` when AVX512 fully lights up, and a lot of string comparisons might fall under such a scenario (< 32 chars).

Naturally, this poses a challenge of expressing TVector vectorized core, with TVector / 2 (e.g. AVX512 -> AVX2) fallback to ensure ~10-30 character strings are compared quickly.

I'm not sure what would the good catch-all solution for such problem look like given how most vectorized loops are currently implemented in CoreLib, but it might become more of a problem in the future, so I wanted to highlight this.

EgorBo: The implementation can be updated to handle trailing elements via SIMD too

EgorBo: Data alignment is worth doing but it's sadly quite verbose to do. It's especially beneficial to do for string opts because chars in string objects are only 4-byte aligned.

MichalPetryka: > Data alignment is worth doing but it's sadly quite verbose to do. It's especially beneficial to do for string opts because chars in string objects are only 4-byte aligned.

Would we pin to guarantee we're keeping alignment for the whole loop or would we just align at the start and hope the GC doesn't move stuff if we'd be adding that to SIMD loops in .Net?

EgorBo: > > Data alignment is worth doing but it's sadly quite verbose to do. It's especially beneficial to do for string opts because chars in string objects are only 4-byte aligned.
> 
> Would we pin to guarantee we're keeping alignment for the whole loop or would we just align at the start and hope the GC doesn't move stuff if we'd be adding that to SIMD loops in .Net?

I guess pinning is the only option

EgorBo:
@stephentoub @tannergooding PTAL, some notes:
1) I copied parts of `Ascii.EqualsIgnoreCase` here, I can't just invoke `Ascii.EqualsIgnoreCase` because we need to fallback to ICU in case of non-ascii, while `Ascii.EqualsIgnoreCase` just bails out. I'll think of how I can unify them separately.
2) Trailing elements are now handled with SIMD (used to be scalar) so all benchmarks for Length=8 (chars) and higher show improvements.

Had to replace == with Equals due to https://github.com/dotnet/runtime/issues/93174

EgorBo:
Ping, also cc @adamsitnik @MihaZupan who authored/contributed Ascii.cs where I copied this code from

src/libraries/System.Private.CoreLib/src/System/Globalization/Ordinal.cs
```diff
@@ -78,39 +78,73 @@ internal static int CompareStringIgnoreCaseNonAscii(ref char strA, int lengthA,
             return OrdinalCasing.CompareStringIgnoreCase(ref strA, lengthA, ref strB, lengthB);
         }
 
-        private static bool EqualsIgnoreCase_Vector128(ref char charA, ref char charB, int length)
+        private static bool EqualsIgnoreCase_Vector<TVector>(ref char charA, ref char charB, int length)
```
MihaZupan, line 81: Not specific to just this PR, but would the expectation be that the different `ISimdVector` impl variants are inlined into the caller?
Or should we be spamming them with aggressive inlining / counting on PGO to decide to avoid the extra calls into the correct helper?

src/libraries/System.Private.CoreLib/src/System/Globalization/Ordinal.cs
```diff
@@ -78,39 +78,73 @@ internal static int CompareStringIgnoreCaseNonAscii(ref char strA, int lengthA,
...
+            TVector vec2;
+            TVector loweringMask = TVector.Create(0x20);
+            TVector vecA = TVector.Create('a');
+            TVector vecZMinusA = TVector.Create('z' - 'a');
             do
             {
-                vec1 = Vector128.LoadUnsafe(ref charA, i);
-                vec2 = Vector128.LoadUnsafe(ref charB, i);
+                vec1 = TVector.LoadUnsafe(ref Unsafe.As<char, ushort>(ref charA), i);
+                vec2 = TVector.LoadUnsafe(ref Unsafe.As<char, ushort>(ref charB), i);
```
MihaZupan, line 97: Should we add an `internal static TVector<ushort> LoadUnsafe(ref char source, nuint elementOffset)` helper like we have for `Vector128/256/512` to avoid reintroducing all the `Unsafe.As`s?

MihaZupan (approved):
This LGTM

src/libraries/System.Private.CoreLib/src/System/Globalization/Ordinal.cs
EgorBo: > Should we add an `internal static TVector<ushort> LoadUnsafe(ref char source, nuint elementOffset)` helper like we have for `Vector128/256/512` to avoid reintroducing all the `Unsafe.As`s?

You will have to make it part of the public ISimdVector interface to work (like, a separate API accepting char), that's why I didn't even try 🙂

src/libraries/System.Private.CoreLib/src/System/Globalization/Ordinal.cs
EgorBo: I think it depends on a case? in this case the original function was not inlineable too

src/libraries/System.Private.CoreLib/src/System/Globalization/Ordinal.cs
MihaZupan: Ah right, nvm then

PR #93119, opened 2023-10-06
[QUIC] Cleaned up TlsSecret and added test.

ManickaP:
Added test for SSLKEYLOGFILE, fixed missing writing of the secret and removed direct reference from safe handle (now held by IDisposable).

### DIFF
--- a/src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicTlsSecret.cs
+++ b/src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicTlsSecret.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 #if DEBUG
+using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Text;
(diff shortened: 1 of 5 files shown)


rzikm (approved):
LGTM

src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicTlsSecret.cs
```diff
@@ -19,88 +20,83 @@ internal sealed class MsQuicTlsSecret : IDisposable
...
+        NativeMemory.Clear(_tlsSecrets, (nuint)sizeof(QUIC_TLS_SECRETS));
     }
 
     public unsafe void Dispose()
     {
-        if (_tlsSecrets != null)
+        if (_tlsSecrets is not null)
         {
-            NativeMemory.Clear(_tlsSecrets, (nuint)sizeof(QUIC_TLS_SECRETS));
+            NativeMemory.Free(_tlsSecrets);
```
wfurt, line 98: I'm not sure this is safe. It feels like we can free the memory while MsQuic still may be trying to write to that. That was one reason why originally the native memory was tight to the MsQuic handle lifecycle.

ManickaP: The dispose is called when the connection handle gets released, not from `QuicConnection.DisposeAsync`.

PR #93121, opened 2023-10-06
ECMA335 NaN encoding test fix

tomeksowi:
The test expected an x86 NaN bit pattern which is negative (NaN in C# is defined as 0.0 / 0.0).

[EMCA-335 standard](https://www.ecma-international.org/wp-content/uploads/ECMA-335_6th_edition_june_2012.pdf) *does not specify how to access the exact bit pattern of NaNs that are created, nor the behavior when converting a NaN between 32-bit and 64-bit representation. All of this behavior is deliberately left implementation-specific.*

So fix the test by comparing the encoding to native NaN bit pattern.

Part of #84834

### DIFF
--- a/src/libraries/System.Reflection.Metadata/tests/Metadata/Ecma335/Encoding/InstructionEncoderTests.cs
+++ b/src/libraries/System.Reflection.Metadata/tests/Metadata/Ecma335/Encoding/InstructionEncoderTests.cs
@@ -182,16 +182,20 @@ namespace System.Reflection.Metadata.Ecma335.Tests
             il.LoadConstantR4(float.PositiveInfinity);
             il.LoadConstantR4(float.Epsilon);
 
-            AssertEx.Equal(new byte[]
+            var expectedEncoding = new byte[]
             {
                 (byte)ILOpCode.Ldc_r4, 0x00, 0x00, 0x00, 0x00,
                 (byte)ILOpCode.Ldc_r4, 0xFF, 0xFF, 0x7F, 0x7F,
                 (byte)ILOpCode.Ldc_r4, 0xFF, 0xFF, 0x7F, 0xFF,
-                (byte)ILOpCode.Ldc_r4, 0x00, 0x00, 0xC0, 0xFF,
+                (byte)ILOpCode.Ldc_r4, 0xCC, 0xCC, 0xCC, 0xCC,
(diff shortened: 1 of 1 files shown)


src/libraries/System.Reflection.Metadata/tests/Metadata/Ecma335/Encoding/InstructionEncoderTests.cs
```diff
@@ -182,16 +182,20 @@ public void LoadConstantR4()
...
                 (byte)ILOpCode.Ldc_r4, 0xFF, 0xFF, 0x7F, 0x7F,
                 (byte)ILOpCode.Ldc_r4, 0xFF, 0xFF, 0x7F, 0xFF,
-                (byte)ILOpCode.Ldc_r4, 0x00, 0x00, 0xC0, 0xFF,
+                (byte)ILOpCode.Ldc_r4, 0xCC, 0xCC, 0xCC, 0xCC,
                 (byte)ILOpCode.Ldc_r4, 0x00, 0x00, 0x80, 0xFF,
                 (byte)ILOpCode.Ldc_r4, 0x00, 0x00, 0x80, 0x7F,
                 (byte)ILOpCode.Ldc_r4, 0x01, 0x00, 0x00, 0x00
-            }, builder.ToArray());
+            };
+            // float.NaN may differ across architectures, in particular it's negative on x86 and positive elsewhere
```
jkotas, line 195: `float.NaN` is a constant defined in platform-neutral reference assembly. The platform-neutral reference assemblies should be exactly same regardless of the platform that they are built on.

It may be better to compensate for the non-deterministic NaN encoding in Roslyn. Roslyn should produce the exact same binary for given C# file, regardless of the platform that the binary is built on.

tannergooding: If Roslyn isn't already doing this, it's likely a determinism bug and something they'd want to fix.

The JIT may also canonicalize in some cases, such as for constant folding.

tomeksowi: I fixed the issue with importing in #93285, so closing this PR.

I would still appreciate an answer to understand NaNs better.

PR #93124, opened 2023-10-06
Improve performance of re-used symmetric one shots

vcsjones:
Today, when you use the symmetric one-shots to decrypt or decrypt, it creates a new lite cipher, uses it, then clean it up every time.

(Context: "lite cipher" is the internal facade over the platform's native encryption components)

Creating a lite cipher is expensive from various things like creating handles, validating keys and inputs, and simply the natural of the algorithm itself from activities like key expansion.

This change allows caching the lite cipher for the duration of the key, so long as the key doesn't change. If the key changes, then the lite cipher is invalidated.

The scenario this improves is this:

```C#
using Aes aes = aes.Create();
aes.Key = GetTheKey();
byte[] iv1 = RandomNumberGenerator.GetBytes(16);
byte[] iv2 = RandomNumberGenerator.GetBytes(16);
byte[] ciphertext1 = aes.EncryptCbc(plaintext1, iv1);
byte[] ciphertext2 = aes.EncryptCbc(plaintext2, iv2);
// N or more calls to EncryptCbc / DecryptCbc, etc.
```

Because the one-shots were previously more-or-less fully contained, they were thread safe - such that calling the one shots from a single instance of SymmetricAlgorithm would work and produce correct results. This is trait I would like to keep and I am not comfortable breaking. Because of this then, the caching uses interlocked exchanges of handles. Since we are only caching one instance at a given time, using an instance from multiple threads may not benefit from the cache. I think this is reasonable to keep the caching mechanism simple, and there is an easy developer workaround: pool an instance of the SymmetricAlgorithm.

This does mean though that there is a performance penalty, albeit a small one, for when the one-shot is used exactly once. The extra bookkeeping for the caching is wasted. However, it is far outweighed if you use it more than once.

Full data is below, but some examples:

Aes.EncryptCbc with a single call when from 579.0ns to 606.2ns.
Aes.EncryptCbc with three calls when from 1,395.5ns to 757.2ns.

TripleDES.EncryptCbc with a single call when from 11,311.6ns to 11,294.2ns (this "improvement" is within error)
TripleDES.EncryptCbc with three calls when from 33,826.3ns to 13,004.6ns.

The performance improvements can be substantial when it's used multiple times. For single uses, the performance loss is two `Interlocked.Exchange` calls.

This pull request addresses ECB and CBC only. Absent is CFB. CFB, because of feedback size, could be addressed differently, which is why I am leaving it out of this pull request. ECB and CBC are the far more common modes as well.

Benchmarks are in this Gist so they don't wrap funny in this comment. https://gist.github.com/vcsjones/b79889c4c6196bec50d783ff6e9aba0d

[CBC benchmark source](https://github.com/vcsjones/DotNetCryptoBenches/blob/main/src/CbcOneShot.cs)
[ECB benchmark source](https://github.com/vcsjones/DotNetCryptoBenches/blob/main/src/EcbOneShot.cs)

### DIFF
--- a/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesImplementation.cs
+++ b/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesImplementation.cs
@@ -9,7 +9,12 @@ namespace System.Security.Cryptography
 {
     internal sealed partial class AesImplementation : Aes
     {
-        private ILiteSymmetricCipher? _encryptCbcLiteHash, _decryptCbcLiteHash, _encryptEcbLiteHash, _decryptEcbLiteHash;
+        private const int BitsPerByte = 8;
+
+        private ILiteSymmetricCipher? _encryptCbcLiteHash;
+        private ILiteSymmetricCipher? _decryptCbcLiteHash;
+        private ILiteSymmetricCipher? _encryptEcbLiteHash;
+        private ILiteSymmetricCipher? _decryptEcbLiteHash;
 
         public sealed override ICryptoTransform CreateDecryptor()
         {
@@ -274,7 +279,5 @@ namespace System.Security.Cryptography
             Interlocked.Exchange(ref _encryptEcbLiteHash, null)?.Dispose();
             Interlocked.Exchange(ref _decryptEcbLiteHash, null)?.Dispose();
         }
-
-        private const int BitsPerByte = 8;
     }
 }
--- a/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesImplementation.cs
+++ b/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesImplementation.cs
@@ -43,7 +43,7 @@ namespace System.Security.Cryptography
 
         protected sealed override void Dispose(bool disposing)
         {
-            InvalidOneShotAlgorithms();
+            InvalidateOneShotAlgorithms();
             base.Dispose(disposing);
         }
 
@@ -52,7 +52,7 @@ namespace System.Security.Cryptography
             set
             {
                 base.Key = value;
-                InvalidOneShotAlgorithms();
+                InvalidateOneShotAlgorithms();
             }
         }
 
@@ -267,7 +267,7 @@ namespace System.Security.Cryptography
             return result;
         }
 
-        private void InvalidOneShotAlgorithms()
+        private void InvalidateOneShotAlgorithms()
         {
             Interlocked.Exchange(ref _encryptCbcLiteHash, null)?.Dispose();
             Interlocked.Exchange(ref _decryptCbcLiteHash, null)?.Dispose();
--- a/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesImplementation.cs
+++ b/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesImplementation.cs
@@ -1,12 +1,16 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Diagnostics;
+using System.Threading;
 using Internal.Cryptography;
 
 namespace System.Security.Cryptography
 {
     internal sealed partial class AesImplementation : Aes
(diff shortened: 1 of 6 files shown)


src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesImplementation.cs
```diff
@@ -1,12 +1,16 @@
...
 
+using System.Diagnostics;
+using System.Threading;
 using Internal.Cryptography;
 
 namespace System.Security.Cryptography
 {
     internal sealed partial class AesImplementation : Aes
     {
+        private ILiteSymmetricCipher? _encryptCbcLiteHash, _decryptCbcLiteHash, _encryptEcbLiteHash, _decryptEcbLiteHash;
```
bartonjs, line 12: I'd do these as 4 separate lines.  Assuming any code written in the last 5 years has stacked them together, I'll accept it, though.

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesImplementation.cs
```diff
@@ -69,19 +80,16 @@ protected sealed override void Dispose(bool disposing)
...
-            {
-                return UniversalCryptoOneShot.OneShotEncrypt(cipher, paddingMode, plaintext, destination, out bytesWritten);
-            }
+                encrypting: true,
+                paddingMode,
+                plaintext,
+                destination,
+                UniversalCryptoOneShot.OneShotEncrypt,
+                out bytesWritten);
         }
```
bartonjs, line 98: This pattern repeats over, and over, and over.  (12 times, if my math's right)  Is there any way we can reduce the copy-pasta?  (The best I've come up with is to make a class `CachedKeyState` and defer these methods into _cachedKey.TryEncryptEcbCore (et al)... but that doesn't feel obviously better)

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesImplementation.cs
vcsjones: > . Assuming any code written in the last 5 years has stacked them together

No time to start doing things right like the present. 😏

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesImplementation.cs
vcsjones: It's a method call but they all pass in different parameters between encrypting, mode, etc. The introduction of `OneShotTransformation` _was_ my attempt at reducing duplication.

On top of this relying heavily on partials to avoid every allocation as possible, I am inclined to say "I think this looks fine". I also don't know that we want to burn an allocation on an abstraction for otherwise readable code.

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/AesImplementation.cs
```diff
@@ -39,28 +43,35 @@ public sealed override void GenerateKey()
...
+            InvalidateOneShotAlgorithms();
             base.Dispose(disposing);
         }
 
+        public override byte[] Key
+        {
+            set
+            {
+                base.Key = value;
+                InvalidateOneShotAlgorithms();
```
bartonjs, line 60: * This instance has been used, so there's a cached lite-cipher for the mode we care about.
* Thread A calls into set_Key, hasn't yet made it to updating the field
* Thread B starts encrypting (or decrypting, whatever), something long.  Gigabyteish.  Still has k0
* Thread A makes it out to InvalidateOneShotAlgorithms.  get_Key will now return k1
* Thread B finishes.  The field is null, so it puts back the instance it used at k0.
* As long as parallel ops never happen again, the cache stays at k0 from one-borrow one-return.

Right?

vcsjones: Ah, right, yep.

vcsjones:
Okay, going to close this for now, as the threading part needs a bit of work. Will re-open when that's in better shape, with new benchmarks.

PR #93125, opened 2023-10-06
Replace confusing `new T[] { ... }` optimization with collection expressions

stephentoub:
We rely heavily on the compiler rewriting:
```C#
ReadOnlySpan<T> span = new T[] { /* const data */ };
```
into variants that don't actually allocate and instead blit the data into the assembly data section and create a span that points directly to that data. But the syntax is very confusing and looks like it allocates on every use.

Now that we have collection expressions and they handle the same optimizations (and more), we can avoid the confusion and simplify the code by switching to use them. This does so for all the uses I could find with some greping around.


jkotas (approved):
Nice!

src/libraries/Common/src/System/Security/Cryptography/Asn1/OaepParamsAsn.xml.cs
```diff
@@ -11,11 +11,11 @@ namespace System.Security.Cryptography.Asn1
     [StructLayout(LayoutKind.Sequential)]
     internal partial struct OaepParamsAsn
     {
-        private static ReadOnlySpan<byte> DefaultHashFunc => new byte[] { 0x30, 0x09, 0x06, 0x05, 0x2B, 0x0E, 0x03, 0x02, 0x1A, 0x05, 0x00 };
+        private static ReadOnlySpan<byte> DefaultHashFunc => [0x30, 0x09, 0x06, 0x05, 0x2B, 0x0E, 0x03, 0x02, 0x1A, 0x05, 0x00];
```
tannergooding, line 14: Styling question/preference....

With `new byte[] { x, y, z }` we almost always have the space between the brace and the first entry. Should we do the same here and consistently have `[ x, y, z ]`?

-- My own preference is yes, I find it immensely more readable to have it separated; but understand not everyone may agree

tannergooding (approved):

src/libraries/Common/src/System/Security/Cryptography/Asn1/OaepParamsAsn.xml.cs
danmoseley: Me too 🙂 

ideally there's a role that we can use to enforce whichever way we prefer..

tannergooding: I talked with @CyrusNajmabadi on Discord and apparently no space is the Roslyn default (for various good reasons around matching tuples, list patterns, positional patterns, etc).

I still don't like it, but I'd guess its a losing battle 😄 
-- My brain likes to visually distinguish indices/parameters (`f(x, y)` or `f[x, y]`) from sets/collections (`{ x, y }` or `[ x, y ]`)

gfoidl: Interestingly the [docs use the space too](https://github.com/dotnet/docs/blob/6ccfd5d56617f07c11966e0410e5686625b3e954/docs/csharp/language-reference/operators/snippets/shared/CollectionExpressionExamples.cs#L42) 😃.

vcsjones, line 14: The `.xml.cs` files are generated. These changes will get undone the next time the generator runs.

We should fix the generator so the changes are not undone the next time it runs. The line you are looking for is:

https://github.com/dotnet/runtime/blob/53eaf7d0077e9f815acc970b852fbe79feab4782/src/libraries/Common/src/System/Security/Cryptography/Asn1/asn.xslt#L256

It should be changed to

```plain
private static ReadOnlySpan&lt;byte&gt; <xsl:call-template name="DefaultValueField"/> =&gt; [<xsl:value-of select="@defaultDerInit"/>];
```

vcsjones: The "ASN regenerated files" build errors are occurring because the generator is not matching what is checked in. Once the generator and source agree the build errors should go away.

stephentoub: Thanks.

CyrusNajmabadi:
Woohoo.  Did you use the analyzer/fixer here?  Or do this manually?

stephentoub:
> Woohoo. Did you use the analyzer/fixer here? Or do this manually?

Manual (regex). The analyzer doesn't flag these, or at least it doesn't in the version I'm using, e.g. for:
```C#
internal class Program
{
    private static void Main()
    {
        ReadOnlySpan<int> span = new int[] { 1, 2, 3, 4, 5, 6, 7, 8, 9 };
        Console.WriteLine(span[0]);
    }
}
```
no suggestion or fix is offered.

src/libraries/Common/src/System/Security/Cryptography/RsaPaddingProcessor.cs
```diff
@@ -10,62 +10,62 @@ namespace System.Security.Cryptography
...
-            {
+        private static ReadOnlySpan<byte> DigestInfoSha3_512 =>
+            [
                 0x30, 0x51, 0x30, 0x0D, 0x06, 0x09, 0x60, 0x86, 0x48,
                 0x01, 0x65, 0x03, 0x04, 0x02, 0x0A, 0x05, 0x00, 0x04,
                 0x40,
-            };
+            ];
 
-        private static ReadOnlySpan<byte> EightZeros => new byte[8];
```
stephentoub, line 68: Note that this one was actually allocating on each call :(

src/libraries/Common/src/System/Security/Cryptography/Asn1/OaepParamsAsn.xml.cs
danmoseley: > for various good reasons around matching tuples, list patterns, positional patterns,

I'm curious what this means -- do you mean their formatter can't easily distinguish when a space would be needed and not, so it consistently doesn't use a space?

src/libraries/Common/src/System/Security/Cryptography/RsaPaddingProcessor.cs
danmoseley: Is that necessarily bad? I know you and others have pushed back on caching memory that would never leave Gen0 (I don't know this code OC)

stephentoub: > Is that necessarily bad? Often we avoid caching memory that would never leave Gen0 (I don't know this code OC)

If this had instead been:
```C#
private static ReadOnlySpan<byte> EightZeros => new byte[8] { 0, 0, 0, 0, 0, 0, 0, 0 };
```
there wouldn't have been any allocation, not even on first use: it would have simply been eight zero'd bytes in the assembly data, and the span would have pointed directly to that memory.
[SharpLab](https://sharplab.io/#v2:EYLgtghglgdgPgYhgVwDaosVBTABAEygGdMcBYAKAAEAmARkqoGZcq6A2Vm3AYVwG9KuYbgAOAJygA3CABc8bTgCVsEfAHkYqAJ4BlURBgAeYNvkA+XAEFcAXksxsAd1yn5AbQAcAXQDcQkQlpOQUOXBU1TR19QxMzbEsAITsHZ1d4r28BXAAGABpcgvzCkuKy3ABffwoKoA)

And, yes, a property that looks like it should be non-allocating but that allocates on every access is bad :)

src/libraries/Common/src/System/Security/Cryptography/Asn1/OaepParamsAsn.xml.cs
CyrusNajmabadi: We mean that this is consistent with tuples and list patterns. The existing features this is modeled after.

src/libraries/Common/src/System/Security/Cryptography/RsaPaddingProcessor.cs
vcsjones: Oh, wow, TIL.

stephentoub: We should just fix that in Roslyn. I don't see a good reason for these two to exhibit such a codegen difference. (Though moving forward we'll use that syntax much less and so hopefully won't trip over it again.)
cc: @jcouv

CyrusNajmabadi:
> Manual (regex). The analyzer doesn't flag these, or at least it doesn't in the version I'm using (Version 17.9.0 Preview 1.0 [34203.221.main]), e.g. for:

I'll add support for that case.

PR #93147, opened 2023-10-06
[release/6.0] Fix a memory leak in runtime interop stubs when using an array of structs of types that use old-style managed marshalers

jkoritzinsky:
Fixes Issue (Issue was reported by a 1st party over email)

main PR #93089

# Description

When passing an array of structs that contains an array field marshalled as a SAFEARRAY or "ByValArray" field to a P/Invoke or a COM interop call, the native array and its contents are not released. This PR changes the marshalling logic such that the native array and contents will be released when they should be.

# Customer Impact

Customers without this fix that have interop code with this shape will have unmanaged memory leaks.

# Regression

This is a regression that was introduced in .NET 5.

# Testing

This fix has been manually validated in main and a unit test has been added to the PR in main. This fix has also manually been validated in this branch with the customer and it fixes their scenario.

# Risk

The risk for this PR is low. The logic here will only be hit in these two scenarios, and the expected and documented behavior is that we do not leak this memory. It is very unlikely that customers have taken a dependency on this memory being leaked.

### DIFF
--- a/src/coreclr/vm/ilmarshalers.h
+++ b/src/coreclr/vm/ilmarshalers.h
@@ -3134,39 +3134,33 @@ protected:
     void EmitClearNative(ILCodeStream* pslILEmit) override
     {
         WRAPPER_NO_CONTRACT;
-        ILCodeLabel* pNoManagedValueLabel = nullptr;
         if (IsFieldMarshal(m_dwMarshalFlags))
         {
-            pNoManagedValueLabel = pslILEmit->NewCodeLabel();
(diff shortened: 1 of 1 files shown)


jeffschwMSFT (approved):
approved. please get a code review. we will take for consideration in 6.0.x

jkotas (approved):

jkoritzinsky:
Failures are unrelated (also showing up in release/6.0-staging without this change).

AmeerAssi:
@jkoritzinsky thanks for doing this fix. Can I get a drop of this change to test and see if it fixes the leak I am experiencing?
or do you know when it is planned to be released?

PR #93160, opened 2023-10-07
Fix Decimal literal formatting in source generators

### LINKED ISSUE #92773: Configuration Binding Generator Initializes Suffixed Default Values Incorrectly
### Description

When configuration binding generation is enabled by means of:

```xml
<EnableConfigurationBindingGenerator>true</EnableConfigurationBindingGenerator>
<Features>$(Features);InterceptorsPreview</Features>
```

…a project can fail to compile with this error message:

```
/app/repro/Microsoft.Extensions.Configuration.Binder.SourceGeneration/Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator/BindingExtensions.g.cs(87,25): e
rror CS0664: Literal of type double cannot be implicitly converted to type 'decimal'; use an 'M' suffix to create a literal of this type [/app/repro/repro.csproj]
```

### Reproduction Steps

Given the following project file:

```xml

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <EnableConfigurationBindingGenerator>true</EnableConfigurationBindingGenerator>
    <Features>$(Features);InterceptorsPreview</Features>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Extensions.Hosting" Version="8.0.0-rtm.23477.9" />
    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.0-rtm.23477.9" />
  </ItemGroup>

</Project>
```

…and the following Program.cs:

```csharp

var x = c.GetSection("Anything").Get<Values>();
Console.WriteLine(x);

readonly record struct Values(int A, string B, decimal C = 2.30m);
```

…run `dotnet build` via SDK version "8.0.100-rtm.23477.37".

### Expected behavior

The compilation succeeds, and the value is bound from configuration.

### Actual behavior

Compilation fails with the following error:

```
/app/repro/Microsoft.Extensions.Configuration.Binder.SourceGeneration/Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator/BindingExtensions.g.cs(87,25): e
rror CS0664: Literal of type double cannot be implicitly converted to type 'decimal'; use an 'M' suffix to create a literal of this type [/app/repro/repro.csproj]
```

The line in question in the generated code appears here in a method named `InitializeValues` (but really `Initialize{typename}`):

```csharp
decimal C = 2.30;
if (configuration["C"] is string value2)
{
    C = ParseDecimal(value2, () => configuration.GetSection("C").Path);
}
```

### Regression?

I'm afraid I don't know.

### Known Workarounds

_No response_

### Configuration

.NET Version: 8.0.100-rtm.23477.37
OS and version: macOS 13.5.1 (22G90)
Architecture: x64
Specific?: No, I don't think so.
Blazor: ❌

### Other information

You know what, I just checked initializing `float C = 2.30f` as well, and it behaves the same.

```csharp
float C = 2.3;
if (configuration["C"] is string value2)
{
    C = ParseFloat(value2, () => configuration.GetSection("C").Path);
}
```

ericstj:
Fixes https://github.com/dotnet/runtime/issues/92773

Both ConfigurationBinder and Json were not handling decimal values correctly.  This shares and updates our workaround method for formatting these.


src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
```diff
@@ -0,0 +1,68 @@
...
+    // Standard format for double and single on non-inbox frameworks.
+    public const string DoubleFormatString = "G17";
+    public const string SingleFormatString = "G9";
+
+    // Format a literal in C# format -- works around https://github.com/dotnet/roslyn/issues/58705
+    public static string FormatLiteral(object? value, TypeRef type)
+    {
+        if (value == null)
+        {
+            return $"default({type.FullyQualifiedName})";
```
eiriktsarpalis, line 22: Nit: wouldn't it be preferable to just use `null` for the case of nullable types?

src/libraries/System.Text.Json/tests/Common/TestClasses/TestClasses.Constructor.cs
```diff
@@ -1972,6 +1972,34 @@ public void VerifyMinimal()
...
         }
     }
+    public class Class_With_Parameters_Default_Values
+    {
+        public int I { get; }
+        public float F { get;}
+        public double D { get;}
+        public decimal M { get;}
+
+        public Class_With_Parameters_Default_Values(int i = 21, float f = 42.0f, double d = 3.14159, decimal m = 3.1415926535897932384626433M)
```
eiriktsarpalis, line 1982: Could you also add a char parameter?

eiriktsarpalis: Could you also add nullable variants of the same parameter types?

eiriktsarpalis (approved):

src/libraries/System.Text.Json/tests/Common/TestClasses/TestClasses.Constructor.cs
eiriktsarpalis: Not sure if we're testing for this already elsewhere, but could you also add an optional enum parameter?

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
```diff
@@ -0,0 +1,68 @@
...
+namespace SourceGenerators;
+
+internal static class CSharpSyntaxUtilities
+{
+    // Standard format for double and single on non-inbox frameworks.
+    public const string DoubleFormatString = "G17";
+    public const string SingleFormatString = "G9";
+
+    // Format a literal in C# format -- works around https://github.com/dotnet/roslyn/issues/58705
+    public static string FormatLiteral(object? value, TypeRef type)
```
eiriktsarpalis, line 18: Wouldn't we be able to `SymbolDisplay.FormatPrimitive` here instead?

eiriktsarpalis: Ah nevermind, I just what the linked issue was about.

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        if (value == null)
+        {
+            return $"default({type.FullyQualifiedName})";
+        }
+
+        if (type.TypeKind is TypeKind.Enum)
```
eiriktsarpalis, line 25: Nit: this check is probably redundant since the default value is always provided as the underlying type.

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
```diff
@@ -0,0 +1,68 @@
...
+                return $"{@double.ToString(DoubleFormatString, CultureInfo.InvariantCulture)}D";
+            case float.NegativeInfinity:
+                return "float.NegativeInfinity";
+            case float.PositiveInfinity:
+                return "float.PositiveInfinity";
+            case float.NaN:
+                return "float.NaN";
+            case float @float:
+                return $"{@float.ToString(SingleFormatString, CultureInfo.InvariantCulture)}F";
+            case decimal.MaxValue:
```
eiriktsarpalis, line 53: Are the `MinValue`/`MaxValue` cases strictly necessary? The underlying numeric representations can still be used as literals and we're not special casing `MinValue`/`MaxValue` in the other numeric types.

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
tarekgh: I am seeing we used `null` before.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.cs
```diff
@@ -1064,6 +1064,9 @@ public void BindsToClassConstructorParametersWithDefaultValues()
             Assert.Equal("John", testOptions.ClassWhereParametersHaveDefaultValueProperty.Name);
             Assert.Equal("123, Abc St.", testOptions.ClassWhereParametersHaveDefaultValueProperty.Address);
             Assert.Equal(42, testOptions.ClassWhereParametersHaveDefaultValueProperty.Age);
+            Assert.Equal(42.0f, testOptions.ClassWhereParametersHaveDefaultValueProperty.F);
+            Assert.Equal(3.14159, testOptions.ClassWhereParametersHaveDefaultValueProperty.D);
+            Assert.Equal(3.1415926535897932384626433M, testOptions.ClassWhereParametersHaveDefaultValueProperty.M);
```
tarekgh, line 1069: I assume producing the same formatted precisions on the .NET Framework. right?

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
tarekgh, line 18: >FormatLiteral [](http://example.com/codeflow?start=25&length=13)

do we want to support Half type too which has precision of 5?

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
```diff
@@ -0,0 +1,68 @@
...
+                return "decimal.MaxValue";
+            case decimal.MinValue:
+                return "decimal.MinValue";
+            case decimal @decimal:
+                return $"{@decimal.ToString(CultureInfo.InvariantCulture)}M";
```
tarekgh, line 58: >"{@decimal.ToString(CultureInfo.InvariantCulture)}M"; [](http://example.com/codeflow?start=24&length=53)

worth it a comment telling the default formatting precision for decimal is 29? just in case if anyone wonder why we didn't use Gxx for decimal too.

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
tarekgh: I think Eric copied the code from text.json which doing the same. But still a good question why we did that in the first place?

tarekgh (approved):
I have left a couple of questions. LGTM otherwise.

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
ericstj: I was basing these cases of what Roslyn special cased here: 
https://github.com/dotnet/roslyn/blob/0caa880a161057559d1dcaad890e4053f520136e/src/Compilers/CSharp/Portable/SymbolDisplay/ObjectDisplay.cs#L494-L513

We can test `half` and see how it works.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.cs
ericstj: Yes, these tests passed for me on both .NETFramework and .NETCore.  I suspect that's because whatever rounding that might happen happens the same way on both sides.

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
ericstj: This was existing, but I'll double check it by adding a test and removing if not needed.

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
ericstj: Correct, I'm just moving this from JSON.  I believe this was covered here https://github.com/dotnet/roslyn/issues/58705 - C# actually doesn't like the numerical representation for these.

eiriktsarpalis: > C# actually doesn't like the numerical representation for these.

It works so long as you slap an 'M' suffix at its end.

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
ericstj: As best I can tell this is a gap in the language.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#6454-real-literals

Half literals are not supported.  We cannot emit a value for `Half` here as one doesn't exist today:
https://sharplab.io/#v2:EYLgtghglgdgPgAQEwEYCwAoTBtAsgTwEEAXYgJymAFdiBTACnoAkIAbAMwEoBmAOhQAsnALqYE3AATIJAYQkBvTBOUSAxgHsYAZ2ISWHNQH0AFhIC8EvoIDcSleKkCJuZm3YTTF1xx78hCiQBfO2UHBCdcFG93Twlo30EuAOCMFSlJcOckaI9zSz8k+SCQ9MdnbhzYq38iwLTMFMwxDKRnIlIKajoJEAkSckoaWkxFVPtJAn7Oocq3ThLRtLSmPOM3WzHlFPrN0v13FaKAc1pia2KsDECgA

I filed https://github.com/dotnet/roslyn/issues/70300 track this gap in the language.

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
```diff
@@ -0,0 +1,68 @@
...
+            case float.NegativeInfinity:
+                return "float.NegativeInfinity";
+            case float.PositiveInfinity:
+                return "float.PositiveInfinity";
+            case float.NaN:
+                return "float.NaN";
+            case float @float:
+                return $"{@float.ToString(SingleFormatString, CultureInfo.InvariantCulture)}F";
+            case decimal.MaxValue:
+                return "decimal.MaxValue";
```
tannergooding, line 54: If `Min`/`MaxValue` are here, why not the same for cases like `double` or `float`?

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            case string @string:
+                return SymbolDisplay.FormatLiteral(@string, quote: true); ;
+            case char @char:
+                return SymbolDisplay.FormatLiteral(@char, quote: true);
+            case double.NegativeInfinity:
+                return "double.NegativeInfinity";
+            case double.PositiveInfinity:
+                return "double.PositiveInfinity";
+            case double.NaN:
```
tannergooding, line 41: Will this actually match?

Does C# emit `value.Equals(double.NaN)`?

ericstj: Another existing case from Json.  C# emits:
```C#
if (obj is double && double.IsNaN((double)obj))
```
https://sharplab.io/#v2:EYLgtghglgdgPgAQEwEYCwAoBBmABM3AYVwG9NcL89gB7GgG1wFkAKASlPMu5uACsApgGMALrgBuEegFcBuALy4AJjWnB6AgHQA5CNoDcXbhSPGAzgHcoIoQAsWkmQLanuZDMc+4hEM3JVqGjp6IK5e3AgA7LgiAE6yhh5eAL5hJkmeUbgAZlJ+idypGMlAA

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
ericstj: This was already existing in Json -- my understanding is that it was part of the workaround for https://github.com/dotnet/roslyn/issues/58705

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
ericstj: I see - so probably my fix to use suffixes can remove this special case.

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
ericstj: I think we can remove this special case, mentioned https://github.com/dotnet/runtime/pull/93160/files/c99abe6de78ea5cb63ddad48de7014ec7f6768a0#r1350960069

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
ericstj: We used this same `default` value before: https://github.com/dotnet/runtime/pull/93160/files#diff-b7da6b90947be2bdebc35f367d9dc2c5bb66dd3be881af8a2d6cbb9c18108a7fL1358

src/libraries/System.Text.Json/tests/Common/TestClasses/TestClasses.Constructor.cs
ericstj: Added these cases

src/libraries/Common/src/SourceGenerators/CSharpSyntaxUtilities.cs
ericstj: It's also redundant since none of the enum types were covered by the switch statement and it would have fallen into default case which calls the same.

ericstj:
We might need to service this.  Looks like without this we cannot handle enums with default values which is a bit more common than decimal literals.

PR #93169, opened 2023-10-07
Fix TensorPrimitives.IndexOfXx corner-case when first element is seed value

stephentoub:
IndexOfXx would return -1 in cases when it should have returned 0 with the first element matching, e.g. when the sole element in an IndexOfMax was float.NegativeInfinity.

Found as part of adding more tests for Min/Max{Magnitude} to validate they match their IndexOfXx variants.


src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
```diff
@@ -1431,7 +1454,27 @@ public static void MaxMagnitude_Tensor(int tensorLength)
...
+        [Theory]
+        [MemberData(nameof(TensorLengths))]
+        public static void MaxMagnitude_Tensor_SpecialValues(int tensorLength)
+        {
+            using BoundedMemory<float> x = CreateAndFillTensor(tensorLength);
+
+            RunForEachSpecialValue(() =>
+            {
+                float maxMagnitude = x[0];
+                foreach (float i in x.Span)
```
ericstj, line 1470: nit:
```suggestion
                foreach (float f in x.Span)
```

I see we have this existing convention in other magnitude tests, but it didn't make sense to me.  Why do we use `i` to represent a float?

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
```diff
@@ -1431,7 +1454,27 @@ public static void MaxMagnitude_Tensor(int tensorLength)
...
+                {
+                    maxMagnitude = MathFMaxMagnitude(maxMagnitude, i);
```
ericstj, line 1472: nit:
```suggestion
                    maxMagnitude = MathFMaxMagnitude(maxMagnitude, f);
```

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
```diff
@@ -1720,7 +1784,27 @@ public static void MinMagnitude_Tensor(int tensorLength)
...
+        [Theory]
+        [MemberData(nameof(TensorLengths))]
+        public static void MinMagnitude_Tensor_SpecialValues(int tensorLength)
+        {
+            using BoundedMemory<float> x = CreateAndFillTensor(tensorLength);
+
+            RunForEachSpecialValue(() =>
+            {
+                float minMagnitude = x[0];
+                foreach (float i in x.Span)
```
ericstj, line 1800: nit:
```suggestion
                foreach (float f in x.Span)
```

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
```diff
@@ -1720,7 +1784,27 @@ public static void MinMagnitude_Tensor(int tensorLength)
...
+                {
+                    minMagnitude = MathFMinMagnitude(minMagnitude, i);
```
ericstj, line 1802: nit:
```suggestion
                    minMagnitude = MathFMinMagnitude(minMagnitude, f);
```

ericstj (approved):
LGTM.  Just a couple nits.

PR #93176, opened 2023-10-07
Fix clsHnd in impImportCall

### LINKED ISSUE #93174: ISimdVector issue (GVM resolution doesn't report [Intrinsic]?)
```cs
private static bool Foo1<TVector>(ref int t1, ref int t2)
    where TVector : struct, ISimdVector<TVector, int>
{
    TVector vec1 = TVector.LoadUnsafe(ref t1);
    TVector vec2 = TVector.LoadUnsafe(ref t2);
    return vec1 == vec2;
}

private static bool Foo2<TVector>(ref int t1, ref int t2)
    where TVector : struct, ISimdVector<TVector, int>
{
    TVector vec1 = TVector.LoadUnsafe(ref t1);
    TVector vec2 = TVector.LoadUnsafe(ref t2);
    return vec1.Equals(vec2);
}
```
Current codegen for `FooX<Vector128<int>>(...)`:
```asm
; Assembly listing for method Foo1
G_M44695_IG01:  ;; offset=0x0000
       sub      rsp, 72
       vzeroupper 
G_M44695_IG02:  ;; offset=0x0007
       vmovups  xmm0, xmmword ptr [rcx]
       vmovaps  xmmword ptr [rsp+0x30], xmm0
       vmovups  xmm0, xmmword ptr [rdx]
       vmovaps  xmmword ptr [rsp+0x20], xmm0
       mov      rax, qword ptr [rsp+0x30]
       mov      qword ptr [rsp+0x18], rax
       mov      rax, qword ptr [rsp+0x20]
       mov      qword ptr [rsp+0x10], rax
       mov      eax, dword ptr [rsp+0x18]
       cmp      dword ptr [rsp+0x10], eax
       jne      SHORT G_M44695_IG08
G_M44695_IG03:  ;; offset=0x0039
       mov      eax, dword ptr [rsp+0x1C]
       cmp      dword ptr [rsp+0x14], eax
       jne      SHORT G_M44695_IG08
G_M44695_IG04:  ;; offset=0x0043
       mov      rax, qword ptr [rsp+0x38]
       mov      qword ptr [rsp+0x08], rax
       mov      rax, qword ptr [rsp+0x28]
       mov      qword ptr [rsp], rax
       mov      eax, dword ptr [rsp+0x08]
       cmp      dword ptr [rsp], eax
       jne      SHORT G_M44695_IG06
G_M44695_IG05:  ;; offset=0x005F
       mov      eax, dword ptr [rsp+0x0C]
       cmp      dword ptr [rsp+0x04], eax
       jne      SHORT G_M44695_IG06
       mov      eax, 1
       jmp      SHORT G_M44695_IG07
G_M44695_IG06:  ;; offset=0x0070
       xor      eax, eax
G_M44695_IG07:  ;; offset=0x0072
       jmp      SHORT G_M44695_IG09
G_M44695_IG08:  ;; offset=0x0074
       xor      eax, eax
G_M44695_IG09:  ;; offset=0x0076
       add      rsp, 72
       ret      
; Total bytes of code 123


; Assembly listing for method Foo2
G_M34100_IG01:  ;; offset=0x0000
       vzeroupper 
G_M34100_IG02:  ;; offset=0x0003
       vmovups  xmm0, xmmword ptr [rcx]
       vpcmpd   k1, xmm0, xmmword ptr [rdx], 4
       kortestb k1, k1
       sete     al
       movzx    rax, al
G_M34100_IG03:  ;; offset=0x0018
       ret      
; Total bytes of code 25
```

Both `Foo1` and `Foo2` are expected to produce the same codegen. It looks like `vec1 == vec2;` was inlined into slow path of `Vector128.op_Equality` as if `[Intrinsic]` didn't exist on it. `Equals` works fine because it's a non-Intrinsic wrapper around intrinsic ==.

cc @tannergooding @dotnet/jit-contrib 

We need to fix this issue before we start adopting `ISimdVector`

EgorBo:
Fixes https://github.com/dotnet/runtime/issues/93174

### DIFF
--- a/src/coreclr/jit/importercalls.cpp
+++ b/src/coreclr/jit/importercalls.cpp
@@ -191,9 +191,14 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
             }
         }
 
-        clsHnd = pResolvedToken->hClass;
-
+        clsHnd   = pResolvedToken->hClass;
         clsFlags = callInfo->classFlags;
+
+        // if clsHnd is an interface and method is static, try to get the actual implementation class
+        if ((clsFlags & CORINFO_FLG_INTERFACE) != 0 && (clsFlags & CORINFO_FLG_INTERFACE) != 0)
+        {
+            clsHnd = info.compCompHnd->getMethodClass(methHnd);
+        }
 
 #ifdef DEBUG
         // If this is a call to JitTestLabel.Mark, do "early inlining", and record the test attribute.
--- a/src/coreclr/jit/compiler.h
+++ b/src/coreclr/jit/compiler.h
@@ -2810,7 +2810,6 @@ public:
     GenTreeHWIntrinsic* gtNewScalarHWIntrinsicNode(
         var_types type, GenTree* op1, GenTree* op2, GenTree* op3, NamedIntrinsic hwIntrinsicID);
     CorInfoType getBaseJitTypeFromArgIfNeeded(NamedIntrinsic       intrinsic,
-                                              CORINFO_CLASS_HANDLE clsHnd,
                                               CORINFO_SIG_INFO*    sig,
                                               CorInfoType          simdBaseJitType);
 
--- a/src/coreclr/jit/hwintrinsic.cpp
+++ b/src/coreclr/jit/hwintrinsic.cpp
@@ -359,7 +359,6 @@ const TernaryLogicInfo& TernaryLogicInfo::lookup(uint8_t control)
 //
 // Arguments:
 //    intrinsic       -- id of the intrinsic function.
-//    clsHnd          -- class handle containing the intrinsic function.
 //    method          -- method handle of the intrinsic function.
 //    sig             -- signature of the intrinsic call.
 //    simdBaseJitType -- Predetermined simdBaseJitType, could be CORINFO_TYPE_UNDEF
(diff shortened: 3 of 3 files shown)


EgorBo:
@MihuBot

EgorBo:
@MihuBot

src/coreclr/jit/importercalls.cpp
```diff
@@ -191,10 +191,15 @@ var_types Compiler::impImportCall(OPCODE                  opcode,
             }
         }
 
-        clsHnd = pResolvedToken->hClass;
-
+        clsHnd   = pResolvedToken->hClass;
         clsFlags = callInfo->classFlags;
 
+        // if clsHnd is an interface and method is static, try to get the actual implementation class
+        if ((clsFlags & CORINFO_FLG_INTERFACE) != 0 && (clsFlags & CORINFO_FLG_INTERFACE) != 0)
```
jkotas, line 198: There are other situations where `getCallInfo` can resolve the method on a different class. Would it be better to return the class handle of the resolved method in `CORINFO_CALL_INFO`?

jkotas: (Also, the condition has typo - it checks `CORINFO_FLG_INTERFACE` twice.)

EgorBo: @jkotas do you mean getCallInfo should patch input resolveToken->hClass? Because it seems that JIT mostly access hClass directly from resolveToken for methods in importer

jkotas: I meant that there should be a new `CORINFO_CLASS_HANDLE hClass; //target class handle` field in `CORINFO_CALL_INFO`.

Patching `CORINFO_RESOLVED_TOKEN` would be a hack. `CORINFO_RESOLVED_TOKEN` is meant to be "in" argument for `getCallInfo`.

> it seems that JIT mostly access hClass directly from resolveToken for methods in importer

I would expect that some places want the `hClass` of the method referenced by the IL and some places want the `hClass` of the method that the VM resolved the method to. I would not be surprised if there are subtle bugs today like the one you are trying to fix.

PR #93178, opened 2023-10-07
Add missing readonly modifiers to GCHandle and DependentHandle

Sergio0694:
This PR sprinkles a bunch of missing `readonly` modifiers to APIs of `GCHandle` and `DependentHandle` that are not actually modifying instance state. This allows Roslyn to skip some unnecessary defensive copies when invoking any of these APIs over values stored in either readonly fields, or from members marked as readonly (from mutable struct types).

### DIFF
--- a/src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
+++ b/src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
@@ -81,9 +81,9 @@ namespace System.Runtime
         /// <exception cref="InvalidOperationException">
         /// Thrown if <see cref="IsAllocated"/> is <see langword="false"/> or if the input value is not <see langword="null"/>.</exception>
         /// <remarks>This property is thread-safe.</remarks>
-        public readonly object? Target
+        public object? Target
         {
-            get
+            readonly get
             {
                 IntPtr handle = _handle;
 
@@ -118,9 +118,9 @@ namespace System.Runtime
         /// </remarks>
         /// <exception cref="InvalidOperationException">Thrown if <see cref="IsAllocated"/> is <see langword="false"/>.</exception>
         /// <remarks>This property is thread-safe.</remarks>
-        public readonly object? Dependent
+        public object? Dependent
         {
-            get
+            readonly get
             {
                 IntPtr handle = _handle;
 
--- a/src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
+++ b/src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
@@ -71,7 +71,7 @@ namespace System.Runtime
         /// <see cref="DependentHandle(object?, object?)"/> and has not yet been disposed.
         /// </summary>
         /// <remarks>This property is thread-safe.</remarks>
-        public bool IsAllocated => (nint)_handle != 0;
+        public readonly bool IsAllocated => (nint)_handle != 0;
 
         /// <summary>
         /// Gets or sets the target object instance for the current handle. The target can only be set to a <see langword="null"/> value
@@ -81,7 +81,7 @@ namespace System.Runtime
         /// <exception cref="InvalidOperationException">
         /// Thrown if <see cref="IsAllocated"/> is <see langword="false"/> or if the input value is not <see langword="null"/>.</exception>
         /// <remarks>This property is thread-safe.</remarks>
-        public object? Target
+        public readonly object? Target
         {
             get
             {
@@ -118,7 +118,7 @@ namespace System.Runtime
         /// </remarks>
         /// <exception cref="InvalidOperationException">Thrown if <see cref="IsAllocated"/> is <see langword="false"/>.</exception>
         /// <remarks>This property is thread-safe.</remarks>
-        public object? Dependent
+        public readonly object? Dependent
         {
             get
             {
@@ -154,7 +154,7 @@ namespace System.Runtime
         /// <returns>The values of <see cref="Target"/> and <see cref="Dependent"/>.</returns>
         /// <exception cref="InvalidOperationException">Thrown if <see cref="IsAllocated"/> is <see langword="false"/>.</exception>
         /// <remarks>This property is thread-safe.</remarks>
-        public (object? Target, object? Dependent) TargetAndDependent
+        public readonly (object? Target, object? Dependent) TargetAndDependent
         {
             get
             {
@@ -176,7 +176,7 @@ namespace System.Runtime
         /// </summary>
         /// <returns>The target object instance, if present.</returns>
         /// <remarks>This method mirrors <see cref="Target"/>, but without the allocation check.</remarks>
-        internal object? UnsafeGetTarget()
+        internal readonly object? UnsafeGetTarget()
         {
             return InternalGetTarget(_handle);
         }
@@ -191,7 +191,7 @@ namespace System.Runtime
         /// The signature is also kept the same as the one for the internal call, to improve the codegen.
         /// Note that <paramref name="dependent"/> is required to be on the stack (or it might not be tracked).
         /// </remarks>
-        internal object? UnsafeGetTargetAndDependent(out object? dependent)
+        internal readonly object? UnsafeGetTargetAndDependent(out object? dependent)
         {
             return InternalGetTargetAndDependent(_handle, out dependent);
         }
@@ -200,7 +200,7 @@ namespace System.Runtime
         /// Sets the dependent object instance for the current handle to <see langword="null"/>.
         /// </summary>
         /// <remarks>This method mirrors the <see cref="Target"/> setter, but without allocation and input checks.</remarks>
-        internal void UnsafeSetTargetToNull()
+        internal readonly void UnsafeSetTargetToNull()
         {
             InternalSetTargetToNull(_handle);
         }
@@ -209,7 +209,7 @@ namespace System.Runtime
         /// Sets the dependent object instance for the current handle.
         /// </summary>
         /// <remarks>This method mirrors <see cref="Dependent"/>, but without the allocation check.</remarks>
-        internal void UnsafeSetDependent(object? dependent)
+        internal readonly void UnsafeSetDependent(object? dependent)
         {
             InternalSetDependent(_handle, dependent);
         }
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/GCHandle.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/GCHandle.cs
@@ -76,7 +76,7 @@ namespace System.Runtime.InteropServices
         }
 
         // Target property - allows getting / updating of the handle's referent.
-        public object? Target
+        public readonly object? Target
         {
             get
             {
@@ -103,7 +103,7 @@ namespace System.Runtime.InteropServices
         /// Retrieve the address of an object in a Pinned handle.  This throws
         /// an exception if the handle is any type other than Pinned.
         /// </summary>
-        public IntPtr AddrOfPinnedObject()
+        public readonly IntPtr AddrOfPinnedObject()
         {
             // Check if the handle was not a pinned handle.
             // You can only get the address of pinned handles.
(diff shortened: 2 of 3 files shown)


jkotas:
> values stored in either readonly fields

Storing GCHandles in readonly fields is a pit of failure. Consider:

```
using System.Runtime.InteropServices;

var t = new Test();
t.Dispose();

class Test
{
    readonly GCHandle _handle = GCHandle.Alloc(null);

    public void Dispose()
    {
         Console.WriteLine(_handle.IsAllocated);

         _handle.Free();

         // This prints `true` even though the handle was just freed!
         Console.WriteLine(_handle.IsAllocated);
    }
}
```

This change does not hurt, but I would avoid storing the GCHandles in `readonly` fields in the first place.

Sergio0694:
@jkotas I agree that's not a good idea, the real scenario I was thinking about (and the reason why I made this PR) was more about allowing instance methods touching GC handles stored on mutable fields to be marked as readonly without causing defensive copies (eg. if they're just accessing the target of the handle). As a side note I still wish there was some way (or an analyzer perhaps) to spot defensive copies. It's so easy to accidentally invoke a non readonly member on a value that triggers a defensive copy for one reason or another 🥲

src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
```diff
@@ -81,7 +81,7 @@ public DependentHandle(object? target, object? dependent)
         /// <exception cref="InvalidOperationException">
         /// Thrown if <see cref="IsAllocated"/> is <see langword="false"/> or if the input value is not <see langword="null"/>.</exception>
         /// <remarks>This property is thread-safe.</remarks>
-        public object? Target
+        public readonly object? Target
```
AaronRobinsonMSFT, line 84: This property has a setter. Marking it `readonly` is very odd.

src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
```diff
@@ -118,7 +118,7 @@ public DependentHandle(object? target, object? dependent)
         /// </remarks>
         /// <exception cref="InvalidOperationException">Thrown if <see cref="IsAllocated"/> is <see langword="false"/>.</exception>
         /// <remarks>This property is thread-safe.</remarks>
-        public object? Dependent
+        public readonly object? Dependent
```
AaronRobinsonMSFT, line 121: Has setter.

src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
```diff
@@ -200,7 +200,7 @@ public DependentHandle(object? target, object? dependent)
         /// Sets the dependent object instance for the current handle to <see langword="null"/>.
         /// </summary>
         /// <remarks>This method mirrors the <see cref="Target"/> setter, but without allocation and input checks.</remarks>
-        internal void UnsafeSetTargetToNull()
+        internal readonly void UnsafeSetTargetToNull()
```
AaronRobinsonMSFT, line 203: Marking a setter `readonly` is a busted mechanic. I don't see how this is correct.

src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
```diff
@@ -209,7 +209,7 @@ internal void UnsafeSetTargetToNull()
         /// Sets the dependent object instance for the current handle.
         /// </summary>
         /// <remarks>This method mirrors <see cref="Dependent"/>, but without the allocation check.</remarks>
-        internal void UnsafeSetDependent(object? dependent)
+        internal readonly void UnsafeSetDependent(object? dependent)
```
AaronRobinsonMSFT, line 212: See above.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/GCHandle.cs
```diff
@@ -76,7 +76,7 @@ public void Free()
         }
 
         // Target property - allows getting / updating of the handle's referent.
-        public object? Target
+        public readonly object? Target
```
AaronRobinsonMSFT, line 79: Has setter.

src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
Sergio0694: I'm going to reply to this since it's the first comment, but it applies to all others below as well, since it's the same logic. The `readonly` modifier here simply means "this member is not mutating any instance state". Which _is_ true: none of these methods it mutating the `GCHandle` instance (including all the setter accessors as well). All they're doing is just mutating some _other_ property somewhere else, which is fine. I don't see how this would be any different than, say, some value type "builder type" that exposes some `Write`/`Append` method which is also marked `readonly` — as long as that method is, say, appending an item to a wrapped `List<T>` object (or something like that), the method is in fact not touching any instance state. This is just about avoid defensive copies from the C# compiler, but `readonly` per se doesn't mean you can make assumptions on whether a member is mutating _state_ in general. Does that make sense?

stephentoub: I'm also hesitant to mark the setter as readonly. Removing `readonly` is paramount to a breaking change, which means we're promising here that the setter won't ever mutate the instance state of this struct. While that's the case today, making this readonly is exposing an implementation detail.

stephentoub:
Can you highlight any places where a) this avoids defensive copies, b) those copies involve additional cost, and c) that cost is measurable in any meaningful way?

src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
AaronRobinsonMSFT: Let me call out the two phrases in Steve's response that reflect how I reviewed the pattern in this PR.

> mark the setter as readonly

Setters, by definition, mutate. Marking it `readonly` is a code smell and begs the following question: "if this is 'read only' what exactly is this doing?" 

> making this readonly is exposing an implementation detail

This relates to the former statement. Again, setters mutate. There is nothing more to say. If this one doesn't that is interesting and likely also because of some subtle detail that should not be assumed to be the best way or the concern of the caller. This is particularly true in the NAOT world where more and more runtime components are moving to managed code and where expressing implementation details at the managed boundary is a recipe for future pain.

Sergio0694: If setters being readonly are too odd, maybe I could just mark the getters as so? I'm still not entirely clear I follow the "setters by definition mutate" argument, given that the point here is specifically about indicating that the instance state doesn't mutate, and not a out _any_ state mutating (see my previous example about array builder types), but regardless, I agree it's not really worth spending too much time over. Perfectly happy to update this PR to just mark the getters, or to abandon it if that's the preferred direction.

AaronRobinsonMSFT: > maybe I could just mark the getters as so?

That is a reasonable expectation and one I would be interested in seeing.

> I'm still not entirely clear I follow the "setters by definition mutate" argument, given that the point here is specifically about indicating that the instance state doesn't mutate, and not a out any state mutating

That is an implementation detail. We aren't expressing how this particular setter is implemented as it is odd, setters set a value on the instance as a concept. In this case the setter does some other hidden behavior but the vast majority of the time setters change state.

AaronRobinsonMSFT: Also, please please please... I can't say this enough... please respond to https://github.com/dotnet/runtime/pull/93178#issuecomment-1752141194. Too many times we are sent "optimizations" and they either aren't or have zero measurable impact. Let's make sure we are seeing improvements in some way.

src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
```diff
@@ -71,7 +71,7 @@ public DependentHandle(object? target, object? dependent)
         /// <see cref="DependentHandle(object?, object?)"/> and has not yet been disposed.
         /// </summary>
         /// <remarks>This property is thread-safe.</remarks>
-        public bool IsAllocated => (nint)_handle != 0;
+        public readonly bool IsAllocated => (nint)_handle != 0;
```
lambdageek, line 74: I understand that this is preventing a defensive copy at the IL level, but does this actually make any difference at all in the AOT or JIT?  I assume all the AOT compilers will optimize out extra copies since it is evident to them that `this` is not modified.  So that leaves RyuJIT. Perhaps it would be better to focus on additional analysis there to eliminate unnecessary copies.

tannergooding: I wholeheartedly agree on some of the below cases, around setters, that it's likely confusing to mark them `readonly` even if they are only mutating indirect or static state.

But for cases like this, where its a trivial instance method that will never mutate. We should really just do the right thing and make everything simpler for Roslyn and the JIT.

Because even if the JIT is able to do all the right things, it has to do more to make it happen. That additional work can also sometimes block other optimizations due to other limits being hit around number of locals, total amount of IL inlined, etc. So it's ultimately better for us to mark cases like this correctly, which then also allows good interplay with other language features, analyzers, etc.

src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
tannergooding: I agree that marking it on the `setter` is confusing and likely non-beneficial. It happening to only mutate indirect or static state doesn't mean it will always do that or that all possible implementations may do that.

The `getter` we should really just mark as `readonly` though. It's essentially a wrapper around a trivial field read and will likely never need to mutate. So it makes everyone's lives better by us doing the "right thing" and annotating it.

src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
AaronRobinsonMSFT: > But for cases like this, where its a trivial instance method that will never mutate.

Agree, it is for clarity and a getter would seem to, in general, be a non-mutating operation.

src/coreclr/System.Private.CoreLib/src/System/Runtime/DependentHandle.cs
jkotas: > It's essentially a wrapper around a trivial field read and will likely never need to mutate

If we ever introduce GC mode with read barriers, the getter is going to mutate auxiliary state just like the setter does.

GCHandle is a wrapper around IntPtr that is readonly except for the `Free` method. This implementation detail is implicitly exposed already. Changing it would be a major breaking change. I expect that there is a ton of code out there that depends on `GCHandle.Target` setter to not mutate the GCHandle instance itself and only mutate the auxiliary state.

stephentoub:
>> "Can you highlight any places where a) this avoids defensive copies, b) those copies involve additional cost, and c) that cost is measurable in any meaningful way?"

> I can't, no. This is more of a "conceptual" change (ie. adding readonly here just seems correct on its own, and I'm not proposing the change specifically to fix some performance issue I noticed). At least in simple tests, the JIT seems to produce equivalent codegen. However, the IL is certainly more bloated without readonly (ie. it has one extra local and whatnot), so this still just seems to "make things simpler" for the JIT, like @tannergooding also mentioned.

Thanks. It's not worth making the public API more confusing to make the JIT's life a little bit easier, especially when it's already doing just fine here.

AaronRobinsonMSFT:
@jkotas and @stephentoub After the most recent updates, I don't see any obvious concerns with these changes. Do either of you concerns with going forward with these changes? An alternative would be to focus on https://github.com/dotnet/runtime/issues/94134 as opposed to making changes to `GCHandle`.

jkotas (approved):
Looks fine to me.

AaronRobinsonMSFT (approved):

PR #93179, opened 2023-10-07
SPMI: Fix 'misses' fraction computation and calculate total contexts correctly

jakobbotsch:
* The misses fraction was computed as misses / diffed, which can produce odd results (such as > 100% of contexts being misses). Compute it as misses / total.
* Fix a few places computing total contexts as "successful + failing". After the recent aggregation change the "failing" category no longer includes misses, so this has been updated to "successful + failing + misses".
* Aggregate "diffed contexts" as we want this in a few places (previously it was approximated as "successful compiles" in the diff, which is usually the same thing, but not exactly)
* Remove a bunch of unnecessary int(x) calls after recent changes to do aggregation in superpmi.py

### DIFF
--- a/src/coreclr/scripts/superpmi.py
+++ b/src/coreclr/scripts/superpmi.py
@@ -1441,13 +1441,13 @@ def print_superpmi_result(return_code, coreclr_args, base_metrics, diff_metrics)
     elif return_code == 2:
         logging.warning("Asm diffs found")
     elif return_code == 3:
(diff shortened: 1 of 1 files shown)


jakobbotsch:
E.g. for @EgorBo's PR at #93176 
Before:
image


After:
image

jakobbotsch:

BruceForstall (approved):

Issue #93167, opened 2023-10-07
8.0.0-rc.1 Library not loaded: /System/Library/Frameworks/CryptoKit.framework/Versions/A/CryptoKit


zhenlinyang:
### Description

System.DllNotFoundException: Unable to load shared library 'libSystem.Security.Cryptography.Native.Apple' or one of its dependencies. In order to help diagnose loading problems, consider setting the DYLD_PRINT_LIBRARIES environment variable: 

dlopen(???/dotnet-sdk/shared/Microsoft.NETCore.App/8.0.0-rc.1.23419.4/libSystem.Security.Cryptography.Native.Apple.dylib, 1): Library not loaded: /System/Library/Frameworks/CryptoKit.framework/Versions/A/CryptoKit

macOS Mojave 10.14.6

### Reproduction Steps

Use ```System.Security.Cryptography.MD5.Create();``` in ```macOS Mojave 10.14.6```.

### Expected behavior

Same as .Net 7.0, can be used normally.

### Actual behavior

System.DllNotFoundException: Unable to load shared library 'libSystem.Security.Cryptography.Native.Apple' or one of its dependencies. In order to help diagnose loading problems, consider setting the DYLD_PRINT_LIBRARIES environment variable: 

dlopen(???/dotnet-sdk/shared/Microsoft.NETCore.App/8.0.0-rc.1.23419.4/libSystem.Security.Cryptography.Native.Apple.dylib, 1): Library not loaded: /System/Library/Frameworks/CryptoKit.framework/Versions/A/CryptoKit

### Configuration

8.0.0-rc.1.23419.4 error
7.0.x correct

vcsjones:
> macOS Mojave 10.14.6

[.NET 8 requires macOS 10.15 or later.](https://github.com/dotnet/core/blob/main/release-notes/8.0/supported-os.md#macos)

macOS 10.14 does not include the CryptoKit library that .NET 8 uses for cryptography purposes.

vcsjones:
The installer should be blocking installation on 10.14. Are you saying that you were able to run the pkg installer on 10.14 and it did not block installation?

https://github.com/dotnet/runtime/blob/f21a2666c577306e437f80fe934d76cdb15072a5/src/installer/pkg/sfx/bundle/shared-framework-distribution-template-x64.xml#L10

> I recommend that .Net 8 be compatible with macOS 10.14.

[...]

PR #93180, opened 2023-10-07
Update comment with min OS version locations to update

akoeplinger:
Discovered another version we need to keep in sync in https://github.com/dotnet/runtime/issues/93167

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -63,8 +63,6 @@
     <AndroidApiLevelMin>21</AndroidApiLevelMin>
     <iOSVersionMin>11.0</iOSVersionMin>
     <tvOSVersionMin>11.0</tvOSVersionMin>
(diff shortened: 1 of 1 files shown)


jkotas (approved):

Directory.Build.props
```diff
@@ -58,6 +58,7 @@
         - src/installer/pkg/sfx/bundle/shared-framework-distribution-template-x64.xml
         - src/installer/pkg/sfx/bundle/shared-framework-distribution-template-arm64.xml
         - src/tasks/MobileBuildTasks/Apple/AppleProject.cs
+        - dotnet/installer repo > src/redist/targets/GeneratePKG.targets
```
am11, line 61: Remove the unused `watchOSVersionMin` and `watchOS64_32VersionMin` and add them when needed.

akoeplinger: Done

PR #93187, opened 2023-10-08
[LoongArch64] Fix some issues for CoreRoot test cases:

LuckyXu-HF:

Fix some issues for CoreRoot test cases:
* fix the cast of floating-point types to uint32 in volatileLocal1.dll.
* fix the register conflict for large offset addresses of OSR root frames when `DOTNET_JitStressRegs=2` in more_tailcalls.sh.

### DIFF
--- a/src/coreclr/jit/codegencommon.cpp
+++ b/src/coreclr/jit/codegencommon.cpp
@@ -5848,6 +5848,7 @@ void CodeGen::genFnProlog()
     const bool isOSRx64Root = false;
 #endif // TARGET_AMD64
 
+#ifndef TARGET_LOONGARCH64
     tempMask = initRegs & ~excludeMask & ~regSet.rsMaskResvd;
 
     if (tempMask != RBM_NONE)
@@ -5869,6 +5870,7 @@ void CodeGen::genFnProlog()
             initReg  = genRegNumFromMask(tempMask);
         }
     }
+#endif
 
 #if defined(TARGET_AMD64)
     // For x64 OSR root frames, we can't use any as of yet unsaved
--- a/src/coreclr/jit/codegencommon.cpp
+++ b/src/coreclr/jit/codegencommon.cpp
@@ -5848,7 +5848,6 @@ void CodeGen::genFnProlog()
     const bool isOSRx64Root = false;
 #endif // TARGET_AMD64
 
-#ifndef TARGET_LOONGARCH64
     tempMask = initRegs & ~excludeMask & ~regSet.rsMaskResvd;
 
     if (tempMask != RBM_NONE)
@@ -5870,7 +5869,6 @@ void CodeGen::genFnProlog()
             initReg  = genRegNumFromMask(tempMask);
         }
     }
-#endif
 
 #if defined(TARGET_AMD64)
     // For x64 OSR root frames, we can't use any as of yet unsaved
@@ -5887,6 +5885,13 @@ void CodeGen::genFnProlog()
     if (isRoot && compiler->opts.IsOSR())
     {
         initReg = REG_IP1;
+    }
+#elif defined(TARGET_LOONGARCH64)
+    // For LoongArch64 OSR root frames, we may need a scratch register for large
+    // offset addresses. Use a register that won't be allocated.
+    if (isRoot && compiler->opts.IsOSR())
+    {
+        initReg = REG_SCRATCH;
     }
 #endif
 
--- a/src/coreclr/jit/codegenloongarch64.cpp
+++ b/src/coreclr/jit/codegenloongarch64.cpp
@@ -3754,6 +3754,13 @@ void CodeGen::genFloatToIntCast(GenTree* treeNode)
         GetEmitter()->emitIns_R_R(ins1, dstSize, tmpReg, tmpReg);
         GetEmitter()->emitIns_R_R(ins2, dstSize, treeNode->GetRegNum(), tmpReg);
 
+        if (dstType == TYP_UINT)
+        {
+            GetEmitter()->emitIns_R_R_I(INS_addu16i_d, EA_PTRSIZE, REG_RA, REG_R0, -32768);
+            GetEmitter()->emitIns_R_R_I(INS_bne, EA_PTRSIZE, treeNode->GetRegNum(), REG_RA, (2 << 2));
+            GetEmitter()->emitIns_R_R_I(INS_ori, dstSize, treeNode->GetRegNum(), REG_R0, 0);
+        }
+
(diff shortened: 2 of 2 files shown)


src/coreclr/jit/codegencommon.cpp
```diff
@@ -5869,6 +5870,7 @@ void CodeGen::genFnProlog()
             initReg  = genRegNumFromMask(tempMask);
         }
     }
+#endif
```
LuckyXu-HF, line 5873: For LoongArch64, it’s ok to keep `initReg = REG_SCRATCH` here.

LuckyXu-HF:
cc @shushanhf
Could you please review this PR? Thanks.

src/coreclr/jit/codegencommon.cpp
jakobbotsch: I do not see why this code should not run on LA64. Sounds like LA64 is not marking the reserved registers correctly.

shushanhf: Hi, @jakobbotsch 
I try to explain it.
The `tempMash` is only used here to update the `initReg`.

I think these lines are old codes which only valid for x86/amd64.

Maybe it's best to refactor these codes based on the ISA is RISC or CISC.

shushanhf: for arm64, the initial value of 'initReg' is REG_SCRATCH (in fact is REG_R9), while the `REG_PINVOKE_FRAME` is also REG_R9.
This is the root case why arm64 has to be updated liking the x86/amd64 when ` if (isRoot && compiler->opts.IsOSR())`.

for LA64, the `REG_SCRATCH` and `REG_PINVOKE_FRAME` are different.

shushanhf: For RISC architectures liking LoongArch64, ARM64, RISCV64,   it's easy to avoid the conflict.

shushanhf: @jakobbotsch 
which test case can test the `REG_PINVOKE_FRAME` ?

jakobbotsch: Does the code cause any problems on LA64? It logically looks fine to me -- it is picking a register that it knows will be trashed so it can be used for initing.

Does it have to do with OSR? I can see that ARM64 and x64 have special handling for OSR below (that overrides `initReg` again) -- maybe add an LA64 case there as well? I would prefer it over fixing the same problem in a different way for LA64.

I agree it's likely that ARM64 does not _need_ this code, but `excludeMask` and `regSet.rsMaskResvd` can have registers in them for several reasons, and I am not confident those cannot end up with `REG_SCRATCH` in them in some cases where we then need to compensate for this.

shushanhf: > Does the code cause any problems on LA64? It logically looks fine to me -- it is picking a register that it knows will be trashed so it can be used for initing.

The test cases of CoreRoot is OK for LA64.
But I just want to know the codes
https://github.com/dotnet/runtime/blob/613e4fecde90a05d1df0cda69392434514ac18f5/src/coreclr/jit/morph.cpp#L2734-L2738
how to return the `nonStdRegNum` as `REG_PINVOKE_FRAME`  ??
I tested the test caseds of CoreRoot and the library‘s tests, there is no this case on LA64.

 
> Does it have to do with OSR? 

Yes, the OSR for LA64 is tested OK.

LuckyXu-HF: > Does the code cause any problems on LA64? It logically looks fine to me -- it is picking a register that it knows will be trashed so it can be used for initing.
> 
> Does it have to do with OSR? I can see that ARM64 and x64 have special handling for OSR below (that overrides `initReg` again) -- maybe add an LA64 case there as well? I would prefer it over fixing the same problem in a different way for LA64.
> 
> I agree it's likely that ARM64 does not _need_ this code, but `excludeMask` and `regSet.rsMaskResvd` can have registers in them for several reasons, and I am not confident those cannot end up with `REG_SCRATCH` in them in some cases where we then need to compensate for this.

Yes, these codes will cause the register conflict for large offset addresses of OSR root frames when `DOTNET_JitStressRegs=2` in `more_tailcalls.sh`.

The `initReg` was modified by `line:5870` from `REG_SCRATCH` to `REG_A0` on LA64. (ARM64 also has the same action, but ARM64 instruction `ldr` can handle bigger offset, and ARM64 has special handling for OSR below, so ARM64 didn’t meet the problem.)

Currently we didn’t meet any problem test cases with this solution (including CoreRoot testcases、CoreRoot stress tests、libs.tests、aspnetcore testcases). This solution can get a little bit gain for LA64.

jakobbotsch: `REG_PINVOKE_FRAME` looks unused on 64-bit platforms. That doesn't mean that LSRA cannot assign `REG_SCRATCH` to something else, however (at least on arm64. I don't know about LA64.)

> ARM64 has special handling for OSR below, so ARM64 didn’t meet the problem

Please switch LA64 to be similar to x64 and ARM64. x64 and ARM64 both override `initReg` below for OSR methods.

shushanhf: > I am suggesting to do the following:
> 
> 1. Remove `#ifndef TARGET_LOONGARCH64` here

By the way, I think these old codes about the conflics 
https://github.com/dotnet/runtime/blob/613e4fecde90a05d1df0cda69392434514ac18f5/src/coreclr/jit/codegencommon.cpp#L5891-L5895

might be cased here
https://github.com/dotnet/runtime/blob/613e4fecde90a05d1df0cda69392434514ac18f5/src/coreclr/jit/codegencommon.cpp#L6234-L6242

But these is not used by LA64 and the LA64 can hold the large offset.


> 2. Add `#ifdef TARGET_LOONGARCH64` below and handle OSR methods like arm64 and x64 is handled (`if (isRoot && compiler->opts.IsOSR()) { initReg = REG_SCRATCH; }`)
> 
> The point is to align the platforms with the same behavior. It will make it easier in the future if we decide to clean this up a bit.

OK. We will update it.

jakobbotsch: I agree that the assert looks outdated. Feel free to update it to `defined(TARGET_X86) || defined(TARGET_ARM)`.

LuckyXu-HF: Already updated, @shushanhf  @jakobbotsch  thank you!

jakobbotsch (approved):
Thank you for addressing the feedback.

PR #93190, opened 2023-10-08
Improve literal-after-loop regex optimization

stephentoub:
Regex currently has an optimization that looks to see whether the pattern begins with a set loop followed by some literal, in which case it can optimize the search for matches by searching for the literal and then walking backwards through the starting set.  However, it's missing a handful of cases we can easily support:
- It currently gives up if the set loop is wrapped in an atomic and/or a capture.
- It currently gives up if the literal is a set that's wrapped in an atomic, capture, concatenate, loop, or lazy loop.
- If the set loop is followed by an ignore-case string, it currently only searches for the starting set of that string, rather than more of it.
- If the literal is a set, we'd only examine it if it was exactly one iteration (RegexNodeKind.Set) rather than a loop with at least one iteration.

This fixes all of those issues, such that the optimization extends to more patterns. In our regex database, there are currently 189 patterns that lead to using this optimization. With this change, that increases to 331.

Based on benchmark from https://github.com/BurntSushi/rebar#ruff-noqa:
```C#
using System.Text.RegularExpressions;

[HideColumns("start", "end", "Error", "StdDev", "Job")]
public partial class Tests
{
    private static readonly Regex s_regex = new Regex("(\\s*)((?i:# noqa)(?::\\s?(([A-Z]+[0-9]+(?:[,\\s]+)?)+))?)", RegexOptions.Compiled);
    private static string s_haystack = new HttpClient().GetStringAsync("https://raw.githubusercontent.com/BurntSushi/rebar/master/benchmarks/haystacks/wild/cpython-226484e4.py").Result;

    [Benchmark]
    public int LineByLine()
    {
        int total = 0;
        foreach (ReadOnlySpan<char> line in s_haystack.AsSpan().EnumerateLines())
        {
            total += s_regex.Count(line);
        }
        return total;
    }

    [Benchmark]
    public int All() => s_regex.Count(s_haystack);
}
```

| Method     | Toolchain         | Mean      | Ratio |
|----------- |------------------ |----------:|------:|
| LineByLine | \main\corerun.exe | 308.76 ms |  1.00 |
| LineByLine | \pr\corerun.exe   |  66.82 ms |  0.22 |
|            |                   |           |       |
| All        | \main\corerun.exe | 278.16 ms |  1.00 |
| All        | \pr\corerun.exe   |  23.11 ms |  0.08 |


src/libraries/System.Text.RegularExpressions/tests/UnitTests/RegexFindOptimizationsTests.cs
```diff
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Diagnostics;
```
buyaa-n, line 4: NIT: this seems not used

buyaa-n (approved):
Looks good at my level

src/libraries/System.Text.RegularExpressions/tests/FunctionalTests/Regex.Groups.Tests.cs
```diff
@@ -958,6 +958,16 @@ private static IEnumerable<(CultureInfo Culture, string Pattern, string Input, R
                 yield return (enUS, @"(?((\w{3}))\1\1|no)", "no", RegexOptions.None, new string[] { "no", "" });
             }
 
+            // Special cases involving starting position search optimizations
+            yield return (enUS, @"(\d*)(hello)(\d*)", "123hello456", RegexOptions.None, new string[] { "123hello456", "123", "hello", "456" });
+            yield return (enUS, @"((\d*))[AaBbCc](\d*)", "1b", RegexOptions.None, new string[] { "1b", "1", "1", "" });
```
danmoseley, line 963: why the nested captures in particular? 

is it worth doing a few other variations, eg., `@"(\d*)(hello)(\d*)", "hello"` `@"(\w*)(hello)(\w*)", "hello"`.. no reason you regressed these, but basically throwing in 5-10 variations while you're here

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexPrefixAnalyzer.cs
```diff
@@ -816,37 +820,84 @@ public static (RegexNode LoopNode, (char Char, string? String, char[]? Chars) Li
...
-                    chars = chars.Slice(0, RegexCharClass.GetSetChars(nextChild.Str!, chars));
-                    if (!chars.IsEmpty)
+            // If the resulting node is a set with at least one iteration, we can search for it.
+            if (nextChild.IsSetFamily &&
+                !RegexCharClass.IsNegated(nextChild.Str!) &&
+                (nextChild.Kind is RegexNodeKind.Set || nextChild.M >= 1))
+            {
+                Span<char> chars = stackalloc char[5]; // maximum number of chars optimized by IndexOfAny
+                chars = chars.Slice(0, RegexCharClass.GetSetChars(nextChild.Str!, chars));
+                if (!chars.IsEmpty)
```
danmoseley, line 889: is the empty test here a runtime improvement vs just letting the foreach skip?

it seems to me it just causes the empty check twice:
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgAYBYAKDIEYabi6kACYlVgQQAoBlAA7YAdgB4wAC2xQAfK0nTcAShoBvGq02sAlgDNWPAIQKouAHQBJXAFF8AjAE8V1La3UvXW3dBjZJBk3kdYXkpUyVNDU9Nd2jogGEIYVwIABsYMwB1KG0MGB4wJQBuKOiAX1LWCupqphZ2TgAhfiExEzkTZTVK71g/CQCwoO0QzojWSti4xOS0jOzc/MKSj01q2tXWIA===

```asm
    L0007: mov esi, [rcx+8]
    L000a: test esi, esi
    L000c: je short L002f
    L000e: mov rdi, [rcx]
    L0011: xor ebx, ebx
    L0013: test esi, esi
    L0015: jle short L002f
```
vs 
```asm
    L0007: mov rsi, [rcx]
    L000a: mov edi, [rcx+8]
    L000d: xor ebx, ebx
    L000f: test edi, edi
    L0011: jle short L002b
```

danmoseley: maybe a potential JIT improvement.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexPrefixAnalyzer.cs
```diff
@@ -816,37 +820,84 @@ public static (RegexNode LoopNode, (char Char, string? String, char[]? Chars) Li
...
             {
-                case RegexNodeKind.One when !RegexCharClass.CharInClass(nextChild.Ch, firstChild.Str!):
-                    return (firstChild, (nextChild.Ch, null, null));
+                // The literal can be searched for as either a single char or as a string.
+                // But we need to make sure that its starting character isn't part of the preceding
+                // set, as then we can't know for certain where the set loop ends.
+                return
+                    RegexCharClass.CharInClass(prefix[0], firstChild.Str!) ? null :
+                    prefix.Length == 1 ? (firstChild, (prefix[0], null, StringComparison.Ordinal, null)) :
+                    (firstChild, ('\0', prefix, StringComparison.Ordinal, null));
```
danmoseley, line 838: trying to figure out from the caller, but can this line just return null rather than `(firstChild, ('\0', prefix, StringComparison.Ordinal, null)`

in general I find it hard to reason when we return complicated ad-hoc structs, a named struct might be easier to follow (not new in this PR)

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexPrefixAnalyzer.cs
stephentoub: > is the empty test here a runtime improvement vs just letting the foreach skip?

It's not an efficiency thing: it's necessary for correctness. If there are too many chars in the set to fit into the specified chars span, GetSetChars returns 0, which means we can't trust the contents of chars.

src/libraries/System.Text.RegularExpressions/tests/FunctionalTests/Regex.Groups.Tests.cs
stephentoub: > why the nested captures in particular?

Because the implementation traverses down through nested captures and atomics.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexPrefixAnalyzer.cs
stephentoub: > but can this line just return null rather than (firstChild, ('\0', prefix, StringComparison.Ordinal, null)

No... it needs to tell the caller what the prefix is.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexPrefixAnalyzer.cs
danmoseley: I guess IsEmpty could move down to just around https://github.com/dotnet/runtime/pull/93190/files/6aad45619b9884cd7cd379f4bdd15564a16f8d4b#diff-3a7a4260e3a2cd96d2beb16ed4ec8ac508678b139c67f70a37f81be6433cf610R899
then if the foreach ends up returning IsEmpty is never called.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexPrefixAnalyzer.cs
```diff
@@ -816,37 +820,84 @@ public static (RegexNode LoopNode, (char Char, string? String, char[]? Chars) Li
...
+
+                if (RegexCharClass.ParticipatesInCaseConversion(ordinalCaseInsensitivePrefix[0]))
+                {
+                    if (RegexCharClass.CharInClass((char)(ordinalCaseInsensitivePrefix[0] | 0x20), firstChild.Str!) ||
+                        RegexCharClass.CharInClass((char)(ordinalCaseInsensitivePrefix[0] & ~0x20), firstChild.Str!))
+                    {
+                        return null;
+                    }
+                }
+                else if (RegexCharClass.CharInClass(ordinalCaseInsensitivePrefix[0], firstChild.Str!))
```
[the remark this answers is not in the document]
stephentoub: That would only apply to the contents of the if block above, as if the starting character doesn't participate in case conversion, it may not be ASCII. At which point it just moves two CharInClass calls and adds a third to the new method. For now I'll leave it as is... if we find ourselves needing to do the same exact thing elsewhere, we can refactor it then.

src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/RegexPrefixAnalyzer.cs
stephentoub: I don't understand... move down to where?  Are you saying put the foreach loop above the IsEmpty check? If so, what benefit does that provide?

danmoseley: yes .well, not a meaningful one, but in the case chars is not empty, and it goes into the foreach, and returns from the method from within that, then this IsEmpty check wouldn't execute, whereas right now it does either way. am I spacing?

stephentoub: > but in the case chars is not empty, and it goes into the foreach, and returns from the method from within that, then this IsEmpty check wouldn't execute, whereas right now it does either way

And if the foreach loop moved earlier, then you're checking its count and if it's 0 then you're checking IsEmpty, so with both orderings in one case you're checking twice and in the other case once.

danmoseley: Yeah in the current case if empty there is one check, otherwise two. With this change, if non empty and returning from within loop one check, otherwise two. I guess I was implicitly assuming non empty was more common, for no reason.

Anyway this matters zero, I was just curious what mistake I was making...

danmoseley: https://github.com/dotnet/runtime/issues/93708

Issue #93188, opened 2023-10-08
[mono][aot] Precompiling failed for System.Private.CoreLib.dll with exit code 134


kotlarmilos:
## Build Information
Build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=430969&view=logs&j=bb18f708-6684-5381-2c5c-c995f6debaa3&t=644a1944-383a-5910-653c-070240832801
Build error leg or test failing: android-x64 Release AllSubsets_Mono / iossimulator-x64 Release AllSubsets_Mono

```
Mono Ahead of Time compiler - compiling assembly /__w/1/s/artifacts/bin/Android.Device_Emulator.Aot_Llvm.Test/Release/net9.0/android-x64/publish/System.Private.CoreLib.dll
AOTID 1AE7737B-90F2-AECE-8C52-4D7EB44CEE71
* Assertion at /__w/1/s/src/mono/mono/mini/mini-llvm.c:7820, condition `values [ins->sreg1]' not met
```

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json
{
  "ErrorMessage": "condition `values [ins->sreg1]' not met",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=430969
**Error message validated:** `condition `values [ins->sreg1]' not met`
**Result validation: :white_check_mark:** Known issue matched with the provided build.
**Validation performed at:** 10/8/2023 8:12:32 AM UTC

### Report
|Build|Definition|Step Name|Console log|Pull Request|
|---|---|---|---|---|
|[431543](https://dev.azure.com/dnceng-public/public/_build/results?buildId=431543)|dotnet/runtime|Build product|[Log](https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_apis/build/builds/431543/logs/763)||
|[431495](https://dev.azure.com/dnceng-public/public/_build/results?buildId=431495)|dotnet/runtime|Build product|[Log](https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_apis/build/builds/431495/logs/629)|dotnet/runtime#92471|
[...]

PR #93195, opened 2023-10-09
[mono] Revert part of 93092 since it causes a regression when aot-ing corlib.

vargaz:
It causes a regression: https://github.com/dotnet/runtime/issues/93188.


vargaz:
Reverting until the problem can be fixed.

kotlarmilos (approved):

PR #93200, opened 2023-10-09
Add SuppressSingleFileAnalysisWarnings option to ILC and use it in runtime tests

akoeplinger:
ILC emitted warnings about single file analysis issues without a way to suppress them like we have for trimming/aot analysis.

Added that option and suppress these warnings in the runtime tests (we already disable the roslyn analyzers).

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
@@ -51,6 +51,10 @@ The .NET Foundation licenses this file to you under the MIT license.
   <PropertyGroup Condition="'$(SuppressAotAnalysisWarnings)' == 'true'">
     <EnableAotAnalyzer Condition="'$(EnableAotAnalyzer)' == ''">false</EnableAotAnalyzer>
   </PropertyGroup>
+  
+  <PropertyGroup Condition="'$(SuppressSingleFileAnalysisWarnings)' == 'true'">
+    <EnableSingleFileAnalyzer Condition="'$(EnableSingleFileAnalyzer)' == ''">false</EnableSingleFileAnalyzer>
+  </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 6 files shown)


src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -51,6 +51,10 @@ The .NET Foundation licenses this file to you under the MIT license.
   <PropertyGroup Condition="'$(SuppressAotAnalysisWarnings)' == 'true'">
     <EnableAotAnalyzer Condition="'$(EnableAotAnalyzer)' == ''">false</EnableAotAnalyzer>
   </PropertyGroup>
+  
+  <PropertyGroup Condition="'$(SuppressSingleFileAnalysisWarnings)' == 'true'">
```
MichalStrehovsky, line 55: Placing this in this file means that SuppressSingleFileAnalysisWarnings will only disable the single file analyzer for projects that do PublishAot. We'd ideally also respect this for PublishSingleFile.

I wonder if we should interpret these SuppressXXXWarnings in the Roslyn analyzer itself instead of turning SuppressXXXWarnings into EnableXXXAnalyzer in various targets (currently SuppressAot disables the analyzer in AOT targets, SuppressTrim disables the analyzer in ILLinker targets, and SuppressSingleFile doesn't really have suitable targets).

@sbomer @vitek-karas for thoughts.

vitek-karas: I definitely agree it should apply no matter which way single-file analysis is triggered. Technically doing this in MSBuild is a bit "Faster" as it avoid loading the analyzer completely in some cases, but it adds complexity for us. Doing this in the analyzer does feel like a simpler solution, but I'll leave that up to Sven as he knows the ins and outs of the MSBuilds parts of this much better.

stephentoub:
@akoeplinger, is this still relevant?

akoeplinger:
Yes this is still an issue but I had no time to get back to it, will close for now.

PR #93214, opened 2023-10-09
Align data in Buffer.Memmove for arm64

EgorBo:
Currently, we [never switch](https://github.com/dotnet/runtime/blob/9f4884c1aa724e96f665c04d3b4be96125213ffc/src/libraries/System.Private.CoreLib/src/System/Buffer.Unix.cs#L8-L13) to the native `memmove` on ARM64 for Mac/Linux. Instead of enabling it back, I tested manual alignment for `dest` and for most cases saw small wins or the same performance as the native `memmove` on Apple M2 (macOS) and Ampere (Linux).

What's more important, it shows nice wins compared to the current impl when data is not 16 bytes aligned (even just 8-byte alignment is expensive) which is quite a common case since GC only offers 8-byte alignment and e.g. String's data is always 4-byte aligned, etc. A benchmark on Apple M2:

```cs
static byte[] _src = new byte[10000];
static byte[] _dst = new byte[10000];

[Benchmark]
public void CopyTo_align4() => _src.AsSpan(4).CopyTo(_dst.AsSpan(4));
```
Presumably, it should also help with the noise in microbenchmarks since GC gives us 8-byte alignment which can be sometimes naturally aligned to 16 bytes or may be not, depends on the Moon's phase.

| Method        | Toolchain               | Mean     | Error   | StdDev  | Ratio |
|-------------- |------------------------ |---------:|--------:|--------:|------:|
| CopyTo_align4 | /Core_Root_base/corerun | 195.1 ns | 0.75 ns | 0.70 ns |  1.68 |
| CopyTo_align4 | /Core_Root_pr/corerun   | **116.2 ns** | 0.54 ns | 0.50 ns |  1.00 |

I was using the following benchmark to get exact alignment in my tests:
```cs
static Benchmarks()
{
    byte* align64_src = (byte*)NativeMemory.AlignedAlloc(1_000_000, 64);
    byte* align64_dst = (byte*)NativeMemory.AlignedAlloc(1_000_000, 64);
    _align4_src = align64_src + 4;
    _align4_dst = align64_dst + 4;
}

[Benchmark]
[Arguments(...)]
public void CopyTo_align4(int len) =>
    new Span<byte>(_align4_src, len).CopyTo(new Span<byte>(_align4_dst, len));
```

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Buffer.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Buffer.cs
@@ -248,7 +248,7 @@ namespace System
             }
 
 #if HAS_CUSTOM_BLOCKS
-            if (len > 128)
+            if (len >= 256)
             {
                 // Try to opportunistically align the destination below. The input isn't pinned, so the GC
                 // is free to move the references. We're therefore assuming that reads may still be unaligned.
@@ -261,6 +261,7 @@ namespace System
                 dest = ref Unsafe.Add(ref dest, misalignedElements);
                 len -= misalignedElements;
             }
+#endif
 
             // Copy 64-bytes at a time until the remainder is less than 64.
             // If remainder is greater than 16 bytes, then jump to MCPY00. Otherwise, unconditionally copy the last 16 bytes and return.
--- a/src/libraries/System.Private.CoreLib/src/System/Buffer.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Buffer.cs
@@ -246,25 +246,23 @@ namespace System
             //
             // TODO: Consider doing the same on x86/AMD64 for V256 and V512
 #if HAS_CUSTOM_BLOCKS && TARGET_ARM64
-            if (len >= 512 && Vector128.IsHardwareAccelerated)
+            if (Vector128.IsHardwareAccelerated && len >= 512)
             {
-                const nuint alignment = 16;
-
                 // Try to opportunistically align the destination below. The input isn't pinned, so the GC
                 // is free to move the references. We're therefore assuming that reads may still be unaligned.
                 //
                 // dest is more important to align than src because an unaligned store is more expensive
                 // than an unaligned load.
-                nuint misalignedElements = (nuint)Unsafe.AsPointer(ref dest) & (alignment - 1);
+                nuint misalignedElements = (nuint)Unsafe.AsPointer(ref dest) & (Vector128.Size - 1);
                 if (misalignedElements != 0)
                 {
                     // E.g. if misalignedElements is 4, it means we need to use a scalar loop
                     // for 16 - 4 = 12 elements till we're aligned to 16b boundary.
-                    misalignedElements = alignment - misalignedElements;
+                    misalignedElements = Vector128.Size - misalignedElements;
                     nuint offset = 0;
                     do
                     {
-                        // For large misalignments on x64 we might want to use smaller SIMD here.
+                        // For large misalignment on x64 we might want to use smaller SIMD here.
                         Unsafe.Add(ref dest, offset) = Unsafe.Add(ref src, offset);
                         offset++;
                     }
--- a/src/libraries/System.Private.CoreLib/src/System/Buffer.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Buffer.cs
@@ -245,7 +245,7 @@ namespace System
             // 512 is an arbitrary threshold picked for Ampere and Apple M1.
             //
             // TODO: Consider doing the same on x86/AMD64 for V256 and V512
-#if HAS_CUSTOM_BLOCKS && TARGET_ARM64
+#if HAS_CUSTOM_BLOCKS
             if (Vector128.IsHardwareAccelerated && len >= 512)
             {
                 // Try to opportunistically align the destination below. The input isn't pinned, so the GC
@@ -253,25 +253,11 @@ namespace System
                 //
                 // dest is more important to align than src because an unaligned store is more expensive
                 // than an unaligned load.
-                nuint misalignedElements = (nuint)Unsafe.AsPointer(ref dest) & (Vector128.Size - 1);
-                if (misalignedElements != 0)
-                {
-                    // E.g. if misalignedElements is 4, it means we need to use a scalar loop
-                    // for 16 - 4 = 12 elements till we're aligned to 16b boundary.
-                    misalignedElements = Vector128.Size - misalignedElements;
-                    nuint offset = 0;
-                    do
-                    {
-                        // For large misalignment on x64 we might want to use smaller SIMD here.
-                        Unsafe.Add(ref dest, offset) = Unsafe.Add(ref src, offset);
-                        offset++;
-                    }
-                    while (offset != misalignedElements);
-
-                    src = ref Unsafe.Add(ref src, misalignedElements);
-                    dest = ref Unsafe.Add(ref dest, misalignedElements);
-                    len -= misalignedElements;
-                }
+                nuint misalignedElements = Vector128.Size - (nuint)Unsafe.AsPointer(ref dest) & (Vector128.Size - 1);
+                Unsafe.As<byte, Block16>(ref dest) = Unsafe.As<byte, Block16>(ref src);
+                src = ref Unsafe.Add(ref src, misalignedElements);
+                dest = ref Unsafe.Add(ref dest, misalignedElements);
+                len -= misalignedElements;
             }
 #endif
             // PInvoke to the native version when the copy length exceeds the threshold.
--- a/src/libraries/System.Private.CoreLib/src/System/Buffer.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Buffer.cs
@@ -241,30 +241,34 @@ namespace System
             return;
 
         MCPY05:
-            // Misaligned access via SIMD is especially expensive on ARM64 on large data.
-            // 512 is an arbitrary threshold picked for Ampere and Apple M1.
-            //
-            // TODO: Consider doing the same on x86/AMD64 for V256 and V512
+            // PInvoke to the native version when the copy length exceeds the threshold.
+            if (len > MemmoveNativeThreshold)
+            {
+                goto PInvoke;
+            }
+
 #if HAS_CUSTOM_BLOCKS
-            if (Vector128.IsHardwareAccelerated && len >= 512)
+            if (len >= 128)
             {
                 // Try to opportunistically align the destination below. The input isn't pinned, so the GC
                 // is free to move the references. We're therefore assuming that reads may still be unaligned.
                 //
                 // dest is more important to align than src because an unaligned store is more expensive
                 // than an unaligned load.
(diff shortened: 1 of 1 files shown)


EgorBo:
@jkotas PTAL

jkotas:
> https://github.com/dotnet/runtime/blob/9f4884c1aa724e96f665c04d3b4be96125213ffc/src/libraries/System.Private.CoreLib/src/System/Buffer.Unix.cs#L10

Would it be better to fix this TODO instead? What is the perf of your microbenchmark with `MemmoveNativeThreshold = 2048`?

src/libraries/System.Private.CoreLib/src/System/Buffer.cs
```diff
@@ -240,6 +241,39 @@ internal static void Memmove(ref byte dest, ref byte src, nuint len)
             return;
 
         MCPY05:
+            // Misaligned access via SIMD is especially expensive on ARM64 on large data.
+            // 512 is an arbitrary threshold picked for Ampere and Apple M1.
+            //
+            // TODO: Consider doing the same on x86/AMD64 for V256 and V512
+#if HAS_CUSTOM_BLOCKS && TARGET_ARM64
+            if (Vector128.IsHardwareAccelerated && len >= 512)
```
jkotas, line 249: The rest of the method does not use Vector128. It is odd to control this behavior by `Vector128.IsHardwareAccelerated`.

src/libraries/System.Private.CoreLib/src/System/Buffer.cs
```diff
@@ -240,6 +241,39 @@ internal static void Memmove(ref byte dest, ref byte src, nuint len)
...
+                        offset++;
+                    }
+                    while (offset != misalignedElements);
+
+                    src = ref Unsafe.Add(ref src, misalignedElements);
+                    dest = ref Unsafe.Add(ref dest, misalignedElements);
+                    len -= misalignedElements;
+                }
+            }
+#endif
```
MichalPetryka, line 276:
```suggestion
#if HAS_CUSTOM_BLOCKS
            if (Vector128.IsHardwareAccelerated && len >= 128)
            {
                // Try to opportunistically align the destination below. The input isn't pinned, so the GC
                // is free to move the references. We're therefore assuming that reads may still be unaligned.
                //
                // dest is more important to align than src because an unaligned store is more expensive
                // than an unaligned load.
                nuint misalignedElements = (nuint)Unsafe.AsPointer(ref dest) & (Vector128.Size - 1);
				Vector128.LoadUnsafe(ref src).StoreUnsafe(ref dest);
                src = ref Unsafe.Add(ref src, misalignedElements);
                dest = ref Unsafe.Add(ref dest, misalignedElements);
                len -= misalignedElements;
            }
#endif
```
I feel like there's no sense to be so restrictive here, 128 should probably already have benefit here and from what I've seen XArch benefits from alignment too (especially on AMD CPUs) and there's no reason not to use a single SIMD load to get the alignment here

EgorBo: XArch won't benefit from aligning to 16 bytes, only SSE-only, the rest will likely regress.

EgorBo:
> > https://github.com/dotnet/runtime/blob/9f4884c1aa724e96f665c04d3b4be96125213ffc/src/libraries/System.Private.CoreLib/src/System/Buffer.Unix.cs#L10
> 
> Would it be better to fix this TODO instead? What is the perf of your microbenchmark with `MemmoveNativeThreshold = 2048`?

Do you mean to jump into native for 512 uncoditionally? I assume it was disabled for a reason and on some platforms it's slow?

src/libraries/System.Private.CoreLib/src/System/Buffer.cs
MichalPetryka: IIRC @tannergooding wanted to port the memmove code to crossplatform intrinsics and I feel like getting such alignment only makes sense on platforms that do support SIMD.

EgorBo:
> What is the perf of your microbenchmark with MemmoveNativeThreshold = 2048

pretty much the same, sometimes the managed impl is a bit faster. But it's on Apple M2 where memmove is quite optimized, I am not so sure about other platforms/distros, e.g. I've seen a glibc version that didn't try to align data.

jkotas:
> I assume it was disabled for a reason and on some platforms it's slow?

Yes, it was slow during the initial Arm64 bring up when we run on glibc without any Arm64 specific optimizations. I would expect that memcpy is optimized on all current Arm64 systems that we run on. It is what the TODO alludes to.

src/libraries/System.Private.CoreLib/src/System/Buffer.cs
```diff
@@ -240,6 +241,25 @@ internal static void Memmove(ref byte dest, ref byte src, nuint len)
...
+#if HAS_CUSTOM_BLOCKS
+            if (Vector128.IsHardwareAccelerated && len >= 512)
+            {
+                // Try to opportunistically align the destination below. The input isn't pinned, so the GC
+                // is free to move the references. We're therefore assuming that reads may still be unaligned.
+                //
+                // dest is more important to align than src because an unaligned store is more expensive
+                // than an unaligned load.
+                nuint misalignedElements = Vector128.Size - (nuint)Unsafe.AsPointer(ref dest) & (Vector128.Size - 1);
+                Unsafe.As<byte, Block16>(ref dest) = Unsafe.As<byte, Block16>(ref src);
```
MichalPetryka, line 257:
```suggestion
                Unsafe.WriteUnaligned<Block16>(ref dest, Unsafe.ReadUnaligned<Block16>(ref src));
```

src/libraries/System.Private.CoreLib/src/System/Buffer.cs
jkotas: > e.g. mono interpreter

This code is not used by mono interpreter.

src/libraries/System.Private.CoreLib/src/System/Buffer.cs
EgorBo: The current impl is using `Unsafe.As` in the method everywhere so I didn't change it, presumably if an unaligned read is a problem then `HAS_CUSTOM_BLOCKS` should not be defined for such a platform (e.g. it currently is not defined for RISC-V and ARM32)

EgorBo:
Enabled for XARCH as well since the suggested way to align is cheaper. I am seeing wins even on XArch now for misaligned access:

| Method        | Job        | Toolchain                   | len  | Mean      |
|-------------- |----------- |---------------------------- |----- |----------:|
| CopyTo_align4 | Job-VLSCTC | \Core_Root\corerun.exe      | 256  |  2.284 ns |
| CopyTo_align4 | Job-BCQRRZ | \Core_Root_base\corerun.exe | 256  |  2.299 ns |
|               |            |                             |      |           |
| CopyTo_align4 | Job-VLSCTC | \Core_Root\corerun.exe      | 512  |  2.986 ns |
| CopyTo_align4 | Job-BCQRRZ | \Core_Root_base\corerun.exe | 512  |  5.534 ns |
|               |            |                             |      |           |
| CopyTo_align4 | Job-VLSCTC | \Core_Root\corerun.exe      | 1024 |  5.717 ns |
| CopyTo_align4 | Job-BCQRRZ | \Core_Root_base\corerun.exe | 1024 |  8.299 ns |
|               |            |                             |      |           |
| CopyTo_align4 | Job-VLSCTC | \Core_Root\corerun.exe      | 2048 | 11.466 ns |
| CopyTo_align4 | Job-BCQRRZ | \Core_Root_base\corerun.exe | 2048 | 16.798 ns |
|               |            |                             |      |           |
| CopyTo_align4 | Job-VLSCTC | \Core_Root\corerun.exe      | 4096 | 26.465 ns |
| CopyTo_align4 | Job-BCQRRZ | \Core_Root_base\corerun.exe | 4096 | 27.076 ns |

Len = 4096 is done in native.

src/libraries/System.Private.CoreLib/src/System/Buffer.cs
```diff
@@ -246,6 +247,29 @@ internal static void Memmove(ref byte dest, ref byte src, nuint len)
...
+#if HAS_CUSTOM_BLOCKS
+            if (len >= 128)
+            {
+                // Try to opportunistically align the destination below. The input isn't pinned, so the GC
+                // is free to move the references. We're therefore assuming that reads may still be unaligned.
+                //
+                // dest is more important to align than src because an unaligned store is more expensive
+                // than an unaligned load.
+#if TARGET_AMD64
+                // Align to 64 bytes for AMD64 (important for 512bit SIMD)
```
jkotas, line 259: I think we should always align to 64 bytes here to keep it simple and match Block64 used below.

EgorBo: Yeah, also wanted to do that, it probably will be a sub-ns less efficient for arm64 since it doesn't benefit from alignment higher than 16 bytes but it's not worth the complexity

MichalPetryka: Does it maybe make sense to bump the minimum size for this to 256 in this case?

EgorBo: Will measure and adjust

EgorBo:
For the reference, codegen of block copy:
```cs
[StructLayout(LayoutKind.Sequential, Size = 64)]
private struct Block64 { }

static void CopyBlock64(ref byte dest, ref byte src)
{
    Unsafe.As<byte, Block64>(ref dest) = Unsafe.As<byte, Block64>(ref src);
}
```
AVX512 CPU:
```asm
       vmovdqu32 zmm0, zmmword ptr [rdx]
       vmovdqu32 zmmword ptr [rcx], zmm0
```
AVX CPU:
```asm
       vmovdqu  ymm0, ymmword ptr [rdx]
       vmovdqu  ymmword ptr [rcx], ymm0
       vmovdqu  ymm0, ymmword ptr [rdx+0x20]
       vmovdqu  ymmword ptr [rcx+0x20], ymm0
```
SSE2 CPU:
```asm
       movups   xmm0, xmmword ptr [rdx]
       movups   xmmword ptr [rcx], xmm0
       movups   xmm0, xmmword ptr [rdx+0x10]
       movups   xmmword ptr [rcx+0x10], xmm0
       movups   xmm0, xmmword ptr [rdx+0x20]
       movups   xmmword ptr [rcx+0x20], xmm0
       movups   xmm0, xmmword ptr [rdx+0x30]
       movups   xmmword ptr [rcx+0x30], xmm0
```
ARM64 NEON CPU:
```asm
            ldp     q16, q17, [x1]
            stp     q16, q17, [x0]
            ldp     q16, q17, [x1, #0x20]
            stp     q16, q17, [x0, #0x20]
```

Two notes: 
1) it probably makes sense for JIT to use at least 2 regs for BLK unrolling for better pipelining (or even 4 on some x-archs).
2) AVX-512 CPUs might benefit from Block128 copy or even Block256 (My Ryzen with AVX512 won't benefit from it because its AVX-512 is basically 2xAVX2)

jkotas (approved):
Thanks

EgorBo:
Hopefully, will make benchmarks like this more stable:

image

PR #93220, opened 2023-10-09
[iOS][non-icu] HybridGlobalization use system ICU

mkhamoyan:
With these changes we will not load `ICU` anymore for `Apple mobile platforms` and will use system `ICU` from `Apple sdk`.
This will make `HybridGlobalization` as default and only option for `Apple mobile platforms`.
As a follow up, we'll allow also to link in `ICU`.

Below is size comparison for `System.Globalization.Tests` app.

<span><span class="ui-provider a b c d e f g h i j k l m n o p q r s t u v w x y z ab ac ae af ag ah ai aj ak" dir="ltr">
 -- | Hybrid mode | Loading ICU
-- | -- | --
app size | **37,8 MB** | **43,1 MB**
ipa size | **12,7 MB**  | **14.7 MB**

</span></span>

This will give us **2 MB** improvement.


src/tests/FunctionalTests/iOS/Device/AOT/iOS.Device.Aot.Test.csproj
```diff
@@ -9,6 +9,7 @@
     <ExpectedExitCode>42</ExpectedExitCode>
     <SelfContained>true</SelfContained>
     <UseConsoleUITemplate>true</UseConsoleUITemplate>
+    <HybridGlobalization>true</HybridGlobalization>
```
steveisok, line 12: I think this should be assumed with your changes.

mkhamoyan: In FunctionalTests looks like `GlobalizationMode.Hybrid` doesn't return true , I tried adding here, but yes this was unnecessary, still same issue.

steveisok:
> /azp run runtime-extra-platforms

If you just want to run this against all apple platforms you can azp run runtime-ioslike

src/libraries/System.Collections.NonGeneric/tests/CaseInsensitiveComparerTests.cs
```diff
@@ -64,7 +64,7 @@ public void Ctor_CultureInfo_Compare(object a, object b, int expected)
             }
         }
 
-        [Fact]
+        [ConditionalFact(typeof(PlatformDetection), nameof(PlatformDetection.IsNotHybridGlobalizationOnOSX))]
```
ilonatommy, line 67: What's the reason for blocking? Is `CaseInsensitiveComparer` not supported on iOS? If it is, shouldn't it be marked as active issue that we want to investigate later? I think that the same for all the other conditional blocking. If we plan to investigate them, it's better to mark them with ActiveIssues. If we plan to drop the support for them, the conditional fact/theory + mention in the docs on changed behavior is fine.

src/libraries/System.Private.CoreLib/src/System/Globalization/GlobalizationMode.cs
```diff
@@ -13,7 +13,9 @@ internal static partial class GlobalizationMode
         private static partial class Settings
         {
             internal static bool Invariant { get; } = AppContextConfigHelper.GetBooleanConfig("System.Globalization.Invariant", "DOTNET_SYSTEM_GLOBALIZATION_INVARIANT");
-#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS || TARGET_BROWSER
+#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS
+            internal static bool Hybrid { get; } = true;
```
ilonatommy, line 17: This does not look right. We ignore the user's setting now if they would like to disable HG (keep in mind that while setting HG as default it has certain differences in behavior that users might not want to accept and they would prefer to switch HG off). Use `DOTNET_SYSTEM_GLOBALIZATION_HYBRID` to propagate the MSBuild setting to managed code.

src/mono/sample/iOS-NativeAOT/Program.csproj
```diff
@@ -4,6 +4,7 @@
     <OutputPath>bin</OutputPath>
     <IntermediateOutputPath>$(MSBuildThisFileDirectory)/obj/</IntermediateOutputPath>
     <TargetFramework>$(NetCoreAppCurrent)</TargetFramework>
+    <HybridGlobalization>true</HybridGlobalization>
```
ilonatommy, line 7: It should be redundant, as it's set by default now. In case you wanted to keep the sample in the same shape as it was before,
```suggestion
    <HybridGlobalization>false</HybridGlobalization> 
```

src/tests/FunctionalTests/Directory.Build.props
```diff
@@ -38,6 +39,7 @@
     <EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization Condition="'$(EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization)' == ''">false</EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization>
     <EnableUnsafeBinaryFormatterSerialization Condition="'$(EnableUnsafeBinaryFormatterSerialization)' == ''">false</EnableUnsafeBinaryFormatterSerialization>
     <InvariantGlobalization Condition="'$(TargetOS)' == 'maccatalyst' and '$(InvariantGlobalization)' == ''">true</InvariantGlobalization>
+    <HybridGlobalization Condition="'$(HybridGlobalization)' == 'true'">true</HybridGlobalization>
```
ilonatommy, line 42: This does not look intentional.

ilonatommy (commented):
I mostly read C# part

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
```diff
@@ -92,7 +92,8 @@ The .NET Foundation licenses this file to you under the MIT license.
       <NativeLibrary Include="$(IntermediateOutputPath)libs/System.Globalization.Native/build/libSystem.Globalization.Native.a" />
       <DirectPInvoke Include="libSystem.Globalization.Native" />
       <StaticICULibs Include="-Wl,-Bstatic" Condition="'$(StaticExecutable)' != 'true'" />
-      <StaticICULibs Include="-licuio -licutu -licui18n -licuuc -licudata -lstdc++" />
+      <StaticICULibs Include="-licuio -licutu -lstdc++" />
```
MichalStrehovsky, line 95: Does/will HybridGlobalization work on Linux?

This is all under a `StaticICULinking=-true` condition that only works on Linux (ref: https://github.com/dotnet/runtime/issues/94951#issuecomment-1817640360).

eng/testing/tests.ioslike.targets
```diff
@@ -157,6 +157,9 @@
     <WriteLinesToFile File="$(PublishDir)xunit-excludes.txt" Lines="$(XunitExcludesTxtFileContent)" Overwrite="true" />
 
     <PropertyGroup>
+      <!-- Default for now. If we can remove our ICU and rely on the system one, then HybridGlobalization becomes the default and we can remove checks for it -->
+      <HybridGlobalization>true</HybridGlobalization>
```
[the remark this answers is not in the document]
mkhamoyan: This is for tests, we want to run all our tests on ios with Hybrid mode enabled

src/libraries/Common/tests/Tests/System/StringTests.cs
```diff
@@ -1033,24 +1033,26 @@ public static void CompareToNoMatch_StringComparison()
...
-                        firstSpan.CompareTo(secondSpan, StringComparison.CurrentCultureIgnoreCase));
-                    Assert.Equal(
-                        string.Compare(firstSpan.ToString(), secondSpan.ToString(), StringComparison.InvariantCulture),
-                        firstSpan.CompareTo(secondSpan, StringComparison.InvariantCulture));
-                    Assert.Equal(
-                        string.Compare(firstSpan.ToString(), secondSpan.ToString(), StringComparison.InvariantCultureIgnoreCase),
-                        firstSpan.CompareTo(secondSpan, StringComparison.InvariantCultureIgnoreCase));
+                    if (PlatformDetection.IsNotHybridGlobalizationOnOSX)
+                    {
+                        // Due to differences in the implementation, the exact result of CompareTo will not necessarily match with string.Compare.
```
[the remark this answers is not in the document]
mkhamoyan: This comment is not added by me https://github.com/dotnet/runtime/blob/main/src/libraries/Common/tests/Tests/System/StringTests.cs#L1036

src/libraries/System.Collections.NonGeneric/tests/CaseInsensitiveComparerTests.cs
mkhamoyan: This is Ctor_CultureInfo_Compare_TurkishI() it will not be supported, will check/update these test cases in case will be investigates will add ActiveIssue, if no will update the doc.

src/libraries/System.Private.CoreLib/src/System/Globalization/GlobalizationMode.cs
mkhamoyan: This is done intentionally as we were facing issue with some FunctionalTests.
This PR does not allow for iOS to have Hybrid as switch, instead it will work only with Hybrid mode.
The reason is in runtime we need to duplicate a lot of files and build 2 runtimes (1. Loading ICU libs 2. Hybrid using system ICU) to be able to have Hybrid mode as switch. We decided for now to make changes the way to have only one runtime with Hybrid mode.

src/mono/sample/iOS-NativeAOT/Program.csproj
mkhamoyan: I'm not sure I follow, where it is set by default?

src/tests/FunctionalTests/Directory.Build.props
mkhamoyan: It is, this is the section for iOS ([here](https://github.com/dotnet/runtime/blob/main/src/tests/FunctionalTests/Directory.Build.props#L35)) and is needed to run all FunctionalTests on hybrid mode.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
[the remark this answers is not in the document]
mkhamoyan: Will update this part, looks like this part is only for Linux.

src/tests/FunctionalTests/Directory.Build.props
mkhamoyan: As discussed offline, this part is not needed, will update, thanks!

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
```diff
@@ -63,6 +63,7 @@
     <ILLinkSubstitutionsXmls Include="$(ILLinkSharedDirectory)ILLink.Substitutions.NoX86Intrinsics.xml" Condition="'$(SupportsX86Intrinsics)' != 'true'" />
     <ILLinkSubstitutionsXmls Include="$(ILLinkSharedDirectory)ILLink.Substitutions.NoWasmIntrinsics.xml" Condition="'$(SupportsWasmIntrinsics)' != 'true'" />
     <ILLinkSubstitutionsXmls Include="$(ILLinkSharedDirectory)ILLink.Substitutions.OSX.xml" Condition="'$(IsOSXLike)' == 'true'" />
+    <ILLinkSubstitutionsXmls Include="$(ILLinkSharedDirectory)ILLink.Substitutions.iOS.xml" Condition="'$(IsiOSLike)' == 'true'" />
```
rolfbjarne, line 66: Would it make sense to put the substitutions in `ILLink.Substitutions.OSX.xml` since the condition to include that file is the same?

rolfbjarne (commented):
If I understand this correctly, these changes only only affect users if they enable HybridGlobalization - if they disable HybridGlobalization, they'll get the same behavior as before?

steveisok:
> If I understand this correctly, these changes only only affect users if they enable HybridGlobalization - if they disable HybridGlobalization, they'll get the same behavior as before?

How we are rolling, at least initially, is to have this be the only option. As a follow up, we'll allow you to link in ICU.

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
mkhamoyan: > Would it make sense to put the substitutions in `ILLink.Substitutions.OSX.xml` since the condition to include that file is the same?

IsiOSLike is only for iOS, tvOS and MacCatalyst

rolfbjarne: ah, sorry, I misread that

src/mono/sample/iOS-NativeAOT/Program.csproj
akoeplinger: we only support Hybrid now so we don't need to explicitly set it to true.

mkhamoyan: After having this line https://github.com/dotnet/runtime/pull/93220/files#diff-2884e8c39a88d26cb57ef06adca164f61238c4ad1c593d8e38bcb167212faa38R17 yes these changes are redundant, will update them.

src/native/libs/System.Globalization.Native/pal_casing.m
```diff
@@ -6,13 +6,13 @@
 #include "pal_errors.h"
 
 #import <Foundation/Foundation.h>
+#include <unicode/uchar.h>
```
akoeplinger, line 9: this should already be included by pal_icushim_internal.h

src/native/libs/System.Globalization.Native/pal_common.c
```diff
@@ -1,6 +1,9 @@
 #include <stdint.h>
 #include "pal_icushim_internal.h"
 #include "pal_casing.h"
+#if defined(TARGET_MACCATALYST) || defined(TARGET_IOS) || defined(TARGET_TVOS)
+#include <unicode/uchar.h>
```
akoeplinger, line 5: same here, this should've been included in pal_icushim_internal.h

src/native/libs/System.Globalization.Native/pal_collation.m
```diff
@@ -64,6 +64,9 @@ static NSStringCompareOptions ConvertFromCompareOptionsToNSStringCompareOptions(
     if (comparisonOptions & IgnoreWidth)
         options |= NSWidthInsensitiveSearch;
 
+    if (comparisonOptions & IgnoreKanaType)
+        options |= NSCaseInsensitiveSearch | NSDiacriticInsensitiveSearch | NSWidthInsensitiveSearch;
```
akoeplinger, line 68: this doesn't look correct to me, ignoring Kana looks more complicated and we have custom code in e.g. FillIgnoreKanaRules.

src/native/libs/System.Globalization.Native/pal_casing.m
mkhamoyan: It is not included by `pal_icushim_internal.h`. In  `pal_icushim_internal.h` are included our ICU headers for OSX platform but not for `TARGET_MACCATALYST`/`TARGET_IOS`/`TARGET_TVOS`.
Now I include from system ICU only the ones that will be needed in `.c` file.

mkhamoyan: I will move all these includes to `pal_icushim_internal.h` to make it more clear.

rolfbjarne:
> > If I understand this correctly, these changes only only affect users if they enable HybridGlobalization - if they disable HybridGlobalization, they'll get the same behavior as before?
> 
> How we are rolling, at least initially, is to have this be the only option. As a follow up, we'll allow you to link in ICU.

Would it make sense to show a warning or error if someone tries to disabled HybridGlobalization then (or maybe a warning if someone tries to enabled it - since it's redundant - and error if someone tries to disable it)?

mkhamoyan:
> > > If I understand this correctly, these changes only only affect users if they enable HybridGlobalization - if they disable HybridGlobalization, they'll get the same behavior as before?
> > 
> > 
> > How we are rolling, at least initially, is to have this be the only option. As a follow up, we'll allow you to link in ICU.
> 
> Would it make sense to show a warning or error if someone tries to disabled HybridGlobalization then (or maybe a warning if someone tries to enabled it - since it's redundant - and error if someone tries to disable it)?

Yes, that makes sense, will add error/warning.

src/native/libs/System.Globalization.Native/pal_collation.m
mkhamoyan: I created separate PR for IgnoreKanaType changes https://github.com/dotnet/runtime/pull/96002, please take a look. Once it will be merged will bring that changes to this PR also.

src/mono/CMakeLists.txt
```diff
@@ -747,10 +747,9 @@ elseif(HOST_BROWSER)
   set(STATIC_ICU 1)
   set(ICU_LIBS "icucore")
 elseif(HOST_IOS OR HOST_TVOS OR HOST_MACCAT)
-  set(ICU_FLAGS "-DTARGET_UNIX -DU_DISABLE_RENAMING -Wno-reserved-id-macro -Wno-documentation -Wno-documentation-unknown-command -Wno-switch-enum -Wno-covered-switch-default -Wno-extra-semi-stmt -Wno-unknown-warning-option -Wno-deprecated-declarations")
+  set(ICU_FLAGS "-DTARGET_UNIX -DTARGET_IOS -DTARGET_TVOS -DTARGET_MACCATALYST -DU_DISABLE_RENAMING -Wno-reserved-id-macro -Wno-documentation -Wno-documentation-unknown-command -Wno-switch-enum -Wno-covered-switch-default -Wno-extra-semi-stmt -Wno-unknown-warning-option -Wno-deprecated-declarations")
```
akoeplinger, line 750: is it ok if we pass `-DTARGET_IOS -DTARGET_TVOS -DTARGET_MACCATALYST` unconditionally? I'd have expected we only pass on the respective platform

src/native/libs/System.Globalization.Native/pal_icushim_internal.h
```diff
@@ -21,6 +21,12 @@
 
 // All ICU headers need to be included here so that all function prototypes are
 // available before the function pointers are declared below.
+#if defined(__APPLE__) && (TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS)
+#include <unicode/uchar.h>
+#include <unicode/uidna.h>
+#include <unicode/utypes.h>
+#endif
+#if !defined(__APPLE__) || (defined(__APPLE__) && !(TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS))
```
akoeplinger, line 29: couldn't we just do #else here?

src/native/libs/System.Globalization.Native/pal_locale.m
```diff
@@ -61,7 +60,7 @@ static void GetParent(const char* localeID, char* parent, int32_t parentCapacity
     int32_t i;
 
     if (localeID == NULL)
-        localeID = [NSLocale systemLocale].localeIdentifier.UTF8String;
+        localeID = [NSLocale currentLocale].localeIdentifier.UTF8String;
```
akoeplinger, line 63: this seems unrelated to the system ICU change?

src/tests/FunctionalTests/Directory.Build.props
```diff
@@ -5,6 +5,7 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <IsTestProject Condition="'$(IsTestProject)' == ''">true</IsTestProject>
     <IsFunctionalTest>true</IsFunctionalTest>
+    <HybridGlobalization Condition="'$(TargetOS)' == 'ios' or '$(TargetOS)' == 'iossimulator' or '$(TargetOS)' == 'tvos' or '$(TargetOS)' == 'tvossimulator' or '$(TargetOS)' == 'maccatalyst'">true</HybridGlobalization>
```
akoeplinger, line 8: can we use `TargetsAppleMobile` here?

akoeplinger (approved):
Couple questions, looks good otherwise. Thanks!

src/native/libs/System.Globalization.Native/pal_locale.m
mkhamoyan: yes, will revert it back.

src/tests/FunctionalTests/Directory.Build.props
mkhamoyan: Updated, thanks.

src/mono/CMakeLists.txt
mkhamoyan: Updated, thanks.

mkhamoyan:
Failures are not related

PR #93197, opened 2023-10-09
Arm64: Implement LoadAndInsertScalar APIs

kunalspathak:
Implements below 51 APIs:

```c#
public static unsafe (Vector64<byte> Value1, Vector64<byte> Value2) LoadAndInsertScalar((Vector64<byte>, Vector64<byte>) values, [ConstantExpectedAttribute(Max = (byte)(7))] byte index, byte* address);
public static unsafe (Vector64<sbyte> Value1, Vector64<sbyte> Value2) LoadAndInsertScalar((Vector64<sbyte>, Vector64<sbyte>) values, [ConstantExpectedAttribute(Max = (byte)(7))] byte index, sbyte* address);
public static unsafe (Vector64<short> Value1, Vector64<short> Value2) LoadAndInsertScalar((Vector64<short>, Vector64<short>) values, [ConstantExpectedAttribute(Max = (byte)(3))] byte index, short* address);
public static unsafe (Vector64<ushort> Value1, Vector64<ushort> Value2) LoadAndInsertScalar((Vector64<ushort>, Vector64<ushort>) values, [ConstantExpectedAttribute(Max = (byte)(3))] byte index, ushort* address);
public static unsafe (Vector64<int> Value1, Vector64<int> Value2) LoadAndInsertScalar((Vector64<int>, Vector64<int>) values, [ConstantExpectedAttribute(Max = (byte)(1))] byte index, int* address);
public static unsafe (Vector64<uint> Value1, Vector64<uint> Value2) LoadAndInsertScalar((Vector64<uint>, Vector64<uint>) values, [ConstantExpectedAttribute(Max = (byte)(1))] byte index, uint* address);
public static unsafe (Vector64<float> Value1, Vector64<float> Value2) LoadAndInsertScalar((Vector64<float>, Vector64<float>) values, [ConstantExpectedAttribute(Max = (byte)(1))] byte index, float* address);
public static unsafe (Vector64<byte> Value1, Vector64<byte> Value2, Vector64<byte> Value3) LoadAndInsertScalar((Vector64<byte>, Vector64<byte>, Vector64<byte>) values, [ConstantExpectedAttribute(Max = (byte)(7))] byte index, byte* address);
[...]

PR #93223, opened 2023-10-09
Implement StoreSelectedScalar for Arm64

SwapnilGaikwad:
Contribute towards https://github.com/dotnet/runtime/issues/84510.

``` c#
// AdvSimd
// ST2 (single structure)
public static unsafe void StoreSelectedScalar64x2(byte*   address, (Vector64<byte>   Value1, Vector64<byte>   Value2) value, byte index);
public static unsafe void StoreSelectedScalar64x2(sbyte*  address, (Vector64<sbyte>  Value1, Vector64<sbyte>  Value2) value, byte index);
public static unsafe void StoreSelectedScalar64x2(short*  address, (Vector64<short>  Value1, Vector64<short>  Value2) value, byte index);
public static unsafe void StoreSelectedScalar64x2(ushort* address, (Vector64<ushort> Value1, Vector64<ushort> Value2) value, byte index);
public static unsafe void StoreSelectedScalar64x2(int*    address, (Vector64<int>    Value1, Vector64<int>    Value2) value, byte index);
public static unsafe void StoreSelectedScalar64x2(uint*   address, (Vector64<uint>   Value1, Vector64<uint>   Value2) value, byte index);
public static unsafe void StoreSelectedScalar64x2(float*  address, (Vector64<float>  Value1, Vector64<float>  Value2) value, byte index);

// ST3 (single structure)
public static unsafe void StoreSelectedScalar64x3(byte*   address, (Vector64<byte>   Value1, Vector64<byte>   Value2, Vector64<byte>   Value3) value, byte index);
public static unsafe void StoreSelectedScalar64x3(sbyte*  address, (Vector64<sbyte>  Value1, Vector64<sbyte>  Value2, Vector64<sbyte>  Value3) value, byte index);
public static unsafe void StoreSelectedScalar64x3(short*  address, (Vector64<short>  Value1, Vector64<short>  Value2, Vector64<short>  Value3) value, byte index);
public static unsafe void StoreSelectedScalar64x3(ushort* address, (Vector64<ushort> Value1, Vector64<ushort> Value2, Vector64<ushort> Value3) value, byte index);
public static unsafe void StoreSelectedScalar64x3(int*    address, (Vector64<int>    Value1, Vector64<int>    Value2, Vector64<int>    Value3) value, byte index);
public static unsafe void StoreSelectedScalar64x3(uint*   address, (Vector64<uint>   Value1, Vector64<uint>   Value2, Vector64<uint>   Value3) value, byte index);
public static unsafe void StoreSelectedScalar64x3(float*  address, (Vector64<float>  Value1, Vector64<float>  Value2, Vector64<float>  Value3) value, byte index);

// ST4 (single structure)
public static unsafe void StoreSelectedScalar64x4(byte*   address, (Vector64<byte>   Value1, Vector64<byte>   Value2, Vector64<byte>   Value3, Vector64<byte>   Value4) value, byte index);
public static unsafe void StoreSelectedScalar64x4(sbyte*  address, (Vector64<sbyte>  Value1, Vector64<sbyte>  Value2, Vector64<sbyte>  Value3, Vector64<sbyte>  Value4) value, byte index);
public static unsafe void StoreSelectedScalar64x4(short*  address, (Vector64<short>  Value1, Vector64<short>  Value2, Vector64<short>  Value3, Vector64<short>  Value4) value, byte index);
public static unsafe void StoreSelectedScalar64x4(ushort* address, (Vector64<ushort> Value1, Vector64<ushort> Value2, Vector64<ushort> Value3, Vector64<ushort> Value4) value, byte index);
public static unsafe void StoreSelectedScalar64x4(int*    address, (Vector64<int>    Value1, Vector64<int>    Value2, Vector64<int>    Value3, Vector64<int>    Value4) value, byte index);
public static unsafe void StoreSelectedScalar64x4(uint*   address, (Vector64<uint>   Value1, Vector64<uint>   Value2, Vector64<uint>   Value3, Vector64<uint>   Value4) value, byte index);
public static unsafe void StoreSelectedScalar64x4(float*  address, (Vector64<float>  Value1, Vector64<float>  Value2, Vector64<float>  Value3, Vector64<float>  Value4) value, byte index);


// AdvSimd.Arm64
// ST2 (single structure)
public static unsafe void StoreSelectedScalar128x2(byte*   address, (Vector128<byte>   Value1, Vector128<byte>   Value2) value, byte index);
public static unsafe void StoreSelectedScalar128x2(sbyte*  address, (Vector128<sbyte>  Value1, Vector128<sbyte>  Value2) value, byte index);
public static unsafe void StoreSelectedScalar128x2(short*  address, (Vector128<short>  Value1, Vector128<short>  Value2) value, byte index);
public static unsafe void StoreSelectedScalar128x2(ushort* address, (Vector128<ushort> Value1, Vector128<ushort> Value2) value, byte index);
public static unsafe void StoreSelectedScalar128x2(int*    address, (Vector128<int>    Value1, Vector128<int>    Value2) value, byte index);
public static unsafe void StoreSelectedScalar128x2(uint*   address, (Vector128<uint>   Value1, Vector128<uint>   Value2) value, byte index);
public static unsafe void StoreSelectedScalar128x2(long*   address, (Vector128<long>   Value1, Vector128<long>   Value2) value, byte index);
public static unsafe void StoreSelectedScalar128x2(ulong*  address, (Vector128<ulong>  Value1, Vector128<ulong>  Value2) value, byte index);
public static unsafe void StoreSelectedScalar128x2(float*  address, (Vector128<float>  Value1, Vector128<float>  Value2) value, byte index);
public static unsafe void StoreSelectedScalar128x2(double* address, (Vector128<double> Value1, Vector128<double> Value2) value, byte index);

// ST3 (single structure)
public static unsafe void StoreSelectedScalar128x3(byte*   address, (Vector128<byte> Value1,    Vector128<byte> Value2,      Vector128<byte> Value3), byte index);
public static unsafe void StoreSelectedScalar128x3(sbyte*  address, (Vector128<sbyte> Value1,   Vector128<sbyte> Value2,     Vector128<sbyte> Value3), byte index);
public static unsafe void StoreSelectedScalar128x3(short*  address, (Vector128<short> Value1,   Vector128<short> Value2,     Vector128<short> Value3), byte index);
public static unsafe void StoreSelectedScalar128x3(ushort*  address, (Vector128<ushort> Value1,  Vector128<ushort> Value2,    Vector128<ushort> Value3), byte index);
public static unsafe void StoreSelectedScalar128x3(int*  address, (Vector128<int> Value1,     Vector128<int> Value2,       Vector128<int> Value3), byte index);
public static unsafe void StoreSelectedScalar128x3(uint*  address, (Vector128<uint> Value1,    Vector128<uint> Value2,      Vector128<uint> Value3), byte index);
public static unsafe void StoreSelectedScalar128x3(long*  address, (Vector128<long> Value1,    Vector128<long> Value2,      Vector128<long> Value3), byte index);
public static unsafe void StoreSelectedScalar128x3(ulong*  address, (Vector128<ulong> Value1,    Vector128<ulong> Value2,      Vector128<ulong> Value3), byte index);
public static unsafe void StoreSelectedScalar128x3(float*  address, (Vector128<float> Value1,   Vector128<float> Value2,     Vector128<float> Value3), byte index);
public static unsafe void StoreSelectedScalar128x3(double*  address, (Vector128<double> Value1,   Vector128<double> Value2,     Vector128<double> Value3), byte index);

// ST4 (single structure)
public static unsafe void StoreSelectedScalar128x4(byte*   address, (Vector128<byte> Value1,        Vector128<byte> Value2,      Vector128<byte> Value3,      Vector128<byte> Value4), byte index);
public static unsafe void StoreSelectedScalar128x4(sbyte*   address, (Vector128<sbyte> Value1,       Vector128<sbyte> Value2,     Vector128<sbyte> Value3,     Vector128<sbyte> Value4), byte index);
public static unsafe void StoreSelectedScalar128x4(short*   address, (Vector128<short> Value1,       Vector128<short> Value2,     Vector128<short> Value3,     Vector128<short> Value4), byte index);
public static unsafe void StoreSelectedScalar128x4(ushort*   address, (Vector128<ushort> Value1,      Vector128<ushort> Value2,    Vector128<ushort> Value3,    Vector128<ushort> Value4), byte index);
public static unsafe void StoreSelectedScalar128x4(int*   address, (Vector128<int> Value1,         Vector128<int> Value2,       Vector128<int> Value3,       Vector128<int> Value4), byte index);
public static unsafe void StoreSelectedScalar128x4(uint*   address, (Vector128<uint> Value1,        Vector128<uint> Value2,      Vector128<uint> Value3,      Vector128<uint> Value4), byte index);
public static unsafe void StoreSelectedScalar128x4(long*   address, (Vector128<long> Value1,        Vector128<long> Value2,      Vector128<long> Value3,      Vector128<long> Value4), byte index);
public static unsafe void StoreSelectedScalar128x4(ulong*   address, (Vector128<ulong> Value1,        Vector128<ulong> Value2,      Vector128<ulong> Value3,      Vector128<ulong> Value4), byte index);
public static unsafe void StoreSelectedScalar128x4(float*   address, (Vector128<float> Value1,       Vector128<float> Value2,     Vector128<float> Value3,     Vector128<float> Value4), byte index);
public static unsafe void StoreSelectedScalar128x4(double*   address, (Vector128<double> Value1,       Vector128<double> Value2,     Vector128<double> Value3,     Vector128<double> Value4), byte index);      

```


SwapnilGaikwad:
This patch would emit the following assembly sequence

<details>
<summary>1. When the index is a constant value</summary>

```c#
AdvSimd.StoreSelectedScalar64x2((Int32*)_dataTable.outArrayPtr, (AdvSimd.LoadVector64((Int32*)(_dataTable.inArray1Ptr)), AdvSimd.LoadVector64((Int32*)(_dataTable.inArray2Ptr))), 1);
```

```
...
            stp     d8, d16, [fp, #0x10]	// [V18 tmp17], [V19 tmp18]
            ldr     q16, [fp, #0x10]	// [V04 tmp3]
            ldr     q17, [fp, #0x18]	// [V04 tmp3+0x08]
            st2     {v16.s, v17.s}[1], [x20]
...
```
</details>


<details>
<summary>2. When index is not a constant</summary>

```c#
AdvSimd.StoreSelectedScalar64x2((Int32*)_dataTable.outArrayPtr, (AdvSimd.LoadVector64((Int32*)(_dataTable.inArray1Ptr)), AdvSimd.LoadVector64((Int32*)(_dataTable.inArray2Ptr))), elemIndex);
```

```
...
 mov     x0, x21
movz    x2, #0xC8A0      // code for System.Runtime.Intrinsics.Arm.AdvSimd:StoreSelectedScalar64x2(ulong,System.ValueTuple`2[System.Runtime.Intrinsics.Vector64`1[int],System.Runtime.Intrinsics.Vector64`1[int]],ubyte)
movk    x2, #0x5DB8 LSL #16
movk    x2, #0xFFFF LSL #32
ldr     x2, [x2]
...

; Assembly listing for method System.Runtime.Intrinsics.Arm.AdvSimd:StoreSelectedScalar64x2(ulong,System.ValueTuple`2[System.Runtime.Intrinsics.Vector64`1[short],System.Runtime.Intrinsics.Vector64`1[short]],ubyte) (FullOpts)

G_M60890_IG01:  ;; offset=0x0000
            stp     fp, lr, [sp, #-0x20]!
            mov     fp, sp
            stp     d0, d1, [fp, #0x10]	// [V01 arg1], [V01 arg1+0x08]
						;; size=12 bbWeight=1 PerfScore 2.50
G_M60890_IG02:  ;; offset=0x000C
            ldr     q16, [fp, #0x10]	// [V01 arg1]
            ldr     q17, [fp, #0x18]	// [V01 arg1+0x08]
            uxtb    w1, w1
            adr     x2, [G_M60890_IG03]
            add     x2, x2, x1,  LSL #3
            br      x2
						;; size=24 bbWeight=1 PerfScore 7.00
G_M60890_IG03:  ;; offset=0x0024
            st2     {v16.h, v17.h}[0], [x0]
            b       G_M60890_IG07
G_M60890_IG04:  ;; offset=0x002C
            st2     {v16.h, v17.h}[1], [x0]
            b       G_M60890_IG07
G_M60890_IG05:  ;; offset=0x0034
            st2     {v16.h, v17.h}[2], [x0]
            b       G_M60890_IG07
G_M60890_IG06:  ;; offset=0x003C
            st2     {v16.h, v17.h}[3], [x0]
						;; size=4 bbWeight=1 PerfScore 1.00
G_M60890_IG07:  ;; offset=0x0040
            ldp     fp, lr, [sp], #0x20
            ret     lr

```
</details>

<details>
<summary>3. When the index is variable and out of bounds</summary>
The current test crashes with an error.

```
Illegal instruction (core dumped)
```
 Not sure whether it's the desired behaviour. Also, I haven't considered the possibility of creating a ROP gadget with the sequence for Case 2 above.  The sequence of instruction where value of an index can be used control the jump target.
</details>

kunalspathak:
As seen in https://github.com/dotnet/runtime/pull/93197, the API names should be just `StoreSelectedScalar` because you can deduce the shape and tuple size from the parameters.

src/coreclr/jit/emitarm64.cpp
```diff
@@ -5590,10 +5590,23 @@ void emitter::emitIns_R_R_I(
             isLdSt     = true;
             break;
 
+        case INS_st2:
```
kunalspathak, line 5593: could you possibly combine this with the `case INS_ld1, INS_st1` below?

src/coreclr/jit/hwintrinsicarm64.cpp
```diff
@@ -1738,16 +1740,29 @@ GenTree* Compiler::impSpecialIntrinsic(NamedIntrinsic        intrinsic,
             break;
         }
 
+        case NI_AdvSimd_StoreSelectedScalar64x2:
```
kunalspathak, line 1743: nit: rearrange to group the similar ones next to each other.

```c++
        case NI_AdvSimd_StoreVector64x2:
        case NI_AdvSimd_Arm64_StoreVector128x2:
        case NI_AdvSimd_StoreSelectedScalar64x2:
        case NI_AdvSimd_Arm64_StoreSelectedScalar128x2:
```

src/coreclr/jit/hwintrinsicarm64.cpp
```diff
@@ -1738,16 +1740,29 @@ GenTree* Compiler::impSpecialIntrinsic(NamedIntrinsic        intrinsic,
...
-            assert(sig->numArgs == 2);
             assert(retType == TYP_VOID);
 
             CORINFO_ARG_LIST_HANDLE arg1     = sig->args;
             CORINFO_ARG_LIST_HANDLE arg2     = info.compCompHnd->getArgNext(arg1);
             var_types               argType  = TYP_UNKNOWN;
             CORINFO_CLASS_HANDLE    argClass = NO_CLASS_HANDLE;
+            const bool isSingleStructStore = sig->numArgs == 3;
+
+            if (isSingleStructStore)
```
kunalspathak, line 1756: These might need changes once we rename the APIs to `StoreSelectedScalar`.

src/coreclr/jit/hwintrinsiclistarm64.h
```diff
@@ -473,6 +473,7 @@ HARDWARE_INTRINSIC(AdvSimd,       SignExtendWideningUpper,
...
 HARDWARE_INTRINSIC(AdvSimd,       StoreSelectedScalar,                                              -1,      3,      true,  {INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1},         HW_Category_MemoryStore,           HW_Flag_BaseTypeFromSecondArg|HW_Flag_HasImmediateOperand|HW_Flag_SIMDScalar|HW_Flag_SpecialCodeGen)
+HARDWARE_INTRINSIC(AdvSimd,       StoreSelectedScalar64x2,                                           8,      3,      true,  {INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_invalid,        INS_invalid,        INS_st2,            INS_invalid},     HW_Category_MemoryStore,           HW_Flag_BaseTypeFromFirstArg|HW_Flag_SpecialImport|HW_Flag_SpecialCodeGen|HW_Flag_NeedsConsecutiveRegisters|HW_Flag_SIMDScalar|HW_Flag_HasImmediateOperand)
```
kunalspathak, line 476:
```suggestion
HARDWARE_INTRINSIC(AdvSimd,       StoreSelectedScalar64x2,                                           8,      3,      true,  {INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_invalid,        INS_invalid,        INS_st2,            INS_invalid},     HW_Category_MemoryStore,           HW_Flag_BaseTypeFromSecondArg|HW_Flag_HasImmediateOperand|HW_Flag_SIMDScalar|HW_Flag_SpecialCodeGen|HW_Flag_SpecialImport|HW_Flag_NeedsConsecutiveRegisters)
```

src/coreclr/jit/hwintrinsiclistarm64.h
```diff
@@ -658,6 +659,7 @@ HARDWARE_INTRINSIC(AdvSimd_Arm64, StorePair,
...
 HARDWARE_INTRINSIC(AdvSimd_Arm64, StorePairNonTemporal,                                             -1,      3,      true,  {INS_stnp,           INS_stnp,           INS_stnp,           INS_stnp,           INS_stnp,           INS_stnp,           INS_stnp,           INS_stnp,           INS_stnp,           INS_stp},         HW_Category_MemoryStore,           HW_Flag_BaseTypeFromSecondArg|HW_Flag_SpecialCodeGen)
+HARDWARE_INTRINSIC(AdvSimd_Arm64, StoreSelectedScalar128x2,                                         16,      3,      true,  {INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_st2},         HW_Category_MemoryStore,           HW_Flag_BaseTypeFromFirstArg|HW_Flag_SpecialImport|HW_Flag_SpecialCodeGen|HW_Flag_NeedsConsecutiveRegisters|HW_Flag_SIMDScalar|HW_Flag_HasImmediateOperand)
```
kunalspathak, line 662: same here.

kunalspathak (changes requested):
I will double check about the immediate operand out of range scenario.

kunalspathak:
> I will double check about the immediate operand out of range scenario.

Alright, so if you follow https://github.com/dotnet/runtime/pull/93197, the `immOp` is handled in it.

kunalspathak:
> 3. When the index is variable and out of bounds

They should throw `IndexOutOfBoundException`

src/coreclr/jit/emitarm64.cpp
SwapnilGaikwad: Done

src/coreclr/jit/hwintrinsicarm64.cpp
SwapnilGaikwad: Done

src/coreclr/jit/hwintrinsiclistarm64.h
SwapnilGaikwad: Done

src/coreclr/jit/hwintrinsiclistarm64.h
SwapnilGaikwad: Done

src/coreclr/jit/hwintrinsicarm64.cpp
SwapnilGaikwad: Done

kunalspathak (changes requested):

src/coreclr/jit/emitarm64.cpp
```diff
@@ -5593,7 +5593,6 @@ void emitter::emitIns_R_R_I(
         case INS_ld2:
         case INS_ld3:
         case INS_ld4:
-        case INS_st2:
```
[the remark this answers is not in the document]
SwapnilGaikwad: ooh right, `st2` needs to have this check.

src/coreclr/jit/hwintrinsic.cpp
```diff
@@ -1132,6 +1132,46 @@ GenTree* Compiler::impHWIntrinsic(NamedIntrinsic        intrinsic,
         }
     }
 
+#ifdef TARGET_ARM64
+    if ((intrinsic == NI_AdvSimd_StoreSelectedScalar) || (intrinsic == NI_AdvSimd_Arm64_StoreSelectedScalar))
```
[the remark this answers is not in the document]
SwapnilGaikwad: Sure, I'll check this out. Just followed the approach for `LoadAndInsertScalar` intrinsic 🙂 .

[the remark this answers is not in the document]
kunalspathak: Right, we have to do it here because it returns `TYP_STRUCT`, but `StoreSelectedScalar` can be handled in `impSpecialIntrinsic` method I think.

src/tests/JIT/HardwareIntrinsics/Arm/Shared/StoreSelScalarx2Test.template
```diff
@@ -0,0 +1,313 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
```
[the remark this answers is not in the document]
SwapnilGaikwad: Done

src/coreclr/jit/hwintrinsiccodegenarm64.cpp
```diff
@@ -810,6 +810,30 @@ void CodeGen::genHWIntrinsic(GenTreeHWIntrinsic* node)
...
+            case NI_AdvSimd_StoreSelectedScalar64x4:
+            case NI_AdvSimd_Arm64_StoreSelectedScalar128x2:
+            case NI_AdvSimd_Arm64_StoreSelectedScalar128x3:
+            case NI_AdvSimd_Arm64_StoreSelectedScalar128x4:
+            {
+                assert(intrin.op2->OperIsFieldList());
+
+                GenTreeFieldList* fieldList  = intrin.op2->AsFieldList();
+                GenTree*          firstField = fieldList->Uses().GetHead()->GetNode();
+                op2Reg                       = firstField->GetRegNum();
```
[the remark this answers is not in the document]
SwapnilGaikwad: Done

src/coreclr/jit/hwintrinsiclistarm64.h
```diff
@@ -476,6 +476,9 @@ HARDWARE_INTRINSIC(AdvSimd,       SignExtendWideningUpper,
...
 HARDWARE_INTRINSIC(AdvSimd,       StoreSelectedScalar,                                              -1,      3,      true,  {INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1},         HW_Category_MemoryStore,           HW_Flag_BaseTypeFromSecondArg|HW_Flag_HasImmediateOperand|HW_Flag_SIMDScalar|HW_Flag_SpecialCodeGen)
+HARDWARE_INTRINSIC(AdvSimd,       StoreSelectedScalar64x2,                                           8,      3,      true,  {INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_st2,            INS_invalid,        INS_invalid,        INS_st2,            INS_invalid},     HW_Category_MemoryStore,           HW_Flag_BaseTypeFromFirstArg|HW_Flag_HasImmediateOperand|HW_Flag_SIMDScalar|HW_Flag_SpecialCodeGen|HW_Flag_SpecialImport|HW_Flag_NeedsConsecutiveRegisters)
```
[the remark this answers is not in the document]
SwapnilGaikwad: Now we combined all `NxM` into StoreSelectedScalar so we can just work with its name. Happy to change it if you prefer the separate entries for each `M`.

src/coreclr/jit/hwintrinsic.cpp
```diff
@@ -1132,6 +1132,46 @@ GenTree* Compiler::impHWIntrinsic(NamedIntrinsic        intrinsic,
...
+                    intrinsic =
+                        sizeBytes == 8 ? NI_AdvSimd_StoreSelectedScalar64x4 : NI_AdvSimd_Arm64_StoreSelectedScalar128x4;
+                    break;
+                default:
+                    assert("unsupported");
+            }
+        }
+    }
+#endif
+
```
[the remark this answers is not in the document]
SwapnilGaikwad: Done

src/coreclr/jit/hwintrinsicarm64.cpp
```diff
@@ -1746,22 +1747,71 @@ GenTree* Compiler::impSpecialIntrinsic(NamedIntrinsic        intrinsic,
...
+                    simdBaseJitType = strip(info.compCompHnd->getArgType(sig, arg2, &tmpClass));
+
+                    if (simdBaseJitType == CORINFO_TYPE_PTR)
+                    {
+                        simdBaseJitType = info.compCompHnd->getChildType(argClass, &tmpClass);
+                    }
+                }
+            }
+
+            if (op2->TypeGet() == TYP_STRUCT)
```
kunalspathak, line 1797: i think it will be good to separate out `*StoreSelectedScalar` and `*StoreVector64` because `StoreVector` doesn't have an equivalent of `fieldCout == 1`, so don't want to penalize it with the extra checks.

src/coreclr/jit/hwintrinsicarm64.cpp
```diff
@@ -1746,22 +1747,71 @@ GenTree* Compiler::impSpecialIntrinsic(NamedIntrinsic        intrinsic,
...
+                unsigned int         sizeBytes   = 0;
+                CORINFO_FIELD_HANDLE fieldHandle = info.compCompHnd->getFieldInClass(argClass, 0);
+                CorInfoType          fieldType   = info.compCompHnd->getFieldType(fieldHandle, &structType);
+                simdBaseJitType                  = getBaseJitTypeAndSizeOfSIMDType(structType, &sizeBytes);
+                if (simdBaseJitType == CORINFO_TYPE_UNDEF) // the argument is not a vector
+                {
+                    CORINFO_CLASS_HANDLE tmpClass;
+                    simdBaseJitType = strip(info.compCompHnd->getArgType(sig, arg2, &tmpClass));
+
+                    if (simdBaseJitType == CORINFO_TYPE_PTR)
```
kunalspathak, line 1790: this is little odd to have. Can you please check if this can be extracted in other common path?

kunalspathak (changes requested):

src/coreclr/jit/hwintrinsicarm64.cpp
SwapnilGaikwad: Cool, I'll separate them.

src/coreclr/jit/hwintrinsiclistarm64.h
```diff
@@ -475,7 +475,7 @@ HARDWARE_INTRINSIC(AdvSimd,       SignExtendWideningLower,
...
-HARDWARE_INTRINSIC(AdvSimd,       StoreSelectedScalar,                                              -1,      3,      true,  {INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1},         HW_Category_MemoryStore,           HW_Flag_BaseTypeFromSecondArg|HW_Flag_HasImmediateOperand|HW_Flag_SIMDScalar|HW_Flag_SpecialCodeGen)
+HARDWARE_INTRINSIC(AdvSimd,       StoreSelectedScalar,                                               8,      3,      true,  {INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1,            INS_st1},         HW_Category_MemoryStore,           HW_Flag_BaseTypeFromFirstArg|HW_Flag_HasImmediateOperand|HW_Flag_SIMDScalar|HW_Flag_SpecialCodeGen|HW_Flag_SpecialImport|HW_Flag_NeedsConsecutiveRegisters)
```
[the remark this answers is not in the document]
SwapnilGaikwad: Sure, I'll update it.

src/coreclr/jit/hwintrinsic.cpp
```diff
@@ -1182,6 +1182,12 @@ GenTree* Compiler::impHWIntrinsic(NamedIntrinsic        intrinsic,
         immOp = impStackTop(1).val;
         assert(HWIntrinsicInfo::isImmOp(intrinsic, immOp));
     }
+    else if (intrinsic == NI_AdvSimd_StoreSelectedScalar || intrinsic == NI_AdvSimd_Arm64_StoreSelectedScalar)
```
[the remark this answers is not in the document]
SwapnilGaikwad: True. Added an explicit condition to assert the immediate used in`StoreSelectedScalar`. I can move those asserts after setting the if you prefer so.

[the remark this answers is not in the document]
SwapnilGaikwad: Not sure whether I misunderstood your [suggestion](https://github.com/dotnet/runtime/pull/93223#discussion_r1360749482).

src/coreclr/jit/hwintrinsiclistarm64.h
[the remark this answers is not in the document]
SwapnilGaikwad: Done

src/coreclr/jit/hwintrinsic.cpp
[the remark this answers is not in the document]
SwapnilGaikwad: Moved the check for `ImmOp` to `hwintrinsicarm64.cpp` and got rid of this check altogether.

src/coreclr/jit/hwintrinsicarm64.cpp
SwapnilGaikwad: Done

src/coreclr/jit/hwintrinsicarm64.cpp
SwapnilGaikwad: Refactored it all, now getting the `simdBaseJitType` from the first arg so got quite simpler.

src/coreclr/jit/hwintrinsicarm64.cpp
```diff
@@ -1747,21 +1748,34 @@ GenTree* Compiler::impSpecialIntrinsic(NamedIntrinsic        intrinsic,
...
             CORINFO_CLASS_HANDLE    argClass = NO_CLASS_HANDLE;
-
             argType             = JITtype2varType(strip(info.compCompHnd->getArgType(sig, arg2, &argClass)));
             op2                 = impPopStack().val;
             unsigned fieldCount = info.compCompHnd->getClassNumInstanceFields(argClass);
-            argType             = JITtype2varType(strip(info.compCompHnd->getArgType(sig, arg1, &argClass)));
-            op1                 = getArgForHWIntrinsic(argType, argClass);
 
-            assert(op2->TypeGet() == TYP_STRUCT);
+            if (op2->TypeGet() == TYP_STRUCT)
```
kunalspathak, line 1762: I don't think you need this check since this is just for `StoreVector64x2` and `StoreVector128x2`. This case should be untouched now.

src/coreclr/jit/hwintrinsiccodegenarm64.cpp
```diff
@@ -512,7 +512,10 @@ void CodeGen::genHWIntrinsic(GenTreeHWIntrinsic* node)
                 break;
         }
 
-        assert(ins != INS_invalid);
+        if (!(intrin.id == NI_AdvSimd_StoreSelectedScalar || intrin.id == NI_AdvSimd_Arm64_StoreSelectedScalar))
```
kunalspathak, line 515: The logic of detecting instruction should be moved in the above switch-case so when we leave, this assert holds true. Then below, you can just call `emitIns()`.

kunalspathak (changes requested):

src/coreclr/jit/hwintrinsicarm64.cpp
SwapnilGaikwad: True... rolled back all the changes to StoreVectorNx2.

src/coreclr/jit/hwintrinsiccodegenarm64.cpp
SwapnilGaikwad: Done. Also moved updating op2Reg to avoid duplication. Not sure if it's ideal though.

kunalspathak (approved):
LGTM. I will merge it once CI is green. Thanks!

kunalspathak:
@SwapnilGaikwad - seems there are some test failures.

SwapnilGaikwad:
> @SwapnilGaikwad - seems there are some test failures.

I'll take a look.

SwapnilGaikwad:
Hi @kunalspathak , the tests failed because for `StoreSelectedScalar` with `Vector128` (storing multiple structures) and a variable index, we needed an switch table that can take indices up to 15. Currently, it generated ladder only up to index 7.

The main reason for this was the refactoring that uses `simdSize` from the `hwintrinsiclistarm64.cpp` to determine the size of the vector length. For the intrinsics in `AdvSimd` class we set it to 8 and for `AdvSimd.Arm64` it's 16. 
In the main branch, we have `StoreSelectedScalar` using a single register (emitting ST1) for both `Vector128` and `Vector64` in `AdvSimd` class. For those `Vector128` variants, we ended up with the switch table that can support indices from 0 to 7. The test was using index 15 that resulted in executing an illegal instruction and crashed. All the remaining `Vector128` variants (ST2/3/4) are in `AdvSimd.Arm64` class so they worked fine.

Instead of putting a different mechanism to determine the vector size, this patch moves the `Vector128` variant for single structure stores emitting ST1 from `AdvSimd` to `AdvSimd.Arm64` class. They may belong in `AdvSimd.Arm64` because  [VSTR](https://developer.arm.com/documentation/ddi0406/c/Application-Level-Architecture/Instruction-Details/Alphabetical-list-of-instructions/VSTR) and [VST1](https://developer.arm.com/documentation/ddi0406/c/Application-Level-Architecture/Instruction-Details/Alphabetical-list-of-instructions/VST1--multiple-single-elements-) instructions that they intended to emit on Arm32 bit architecture support data sizes up to 64 bits. Not sure how are they expected to work on Arm32. Thus, it makes sense to put only `Vector64` variants in `AdvSimd`.

This breaks the backward compatibility. The only usages for `StoreSelectedScalar` in SDK are with `Vector64` but not aware of it's usage outside. Kindly let me know if you prefer support `Vector128` variants back in `AdvSimd`. That may need some extra handling while importing the intrinsics.

kunalspathak:
> This breaks the backward compatibility.

Indeed and we do not want to do it although we are not sure if anyone is consuming it.  I do know that there are lot of other APIs though that operates on `Vector128` and should be under `AdvSimd.Arm64` ideally.  @tannergooding - do you know why there were added in `AdvSimd`? 

> The test was using index 15 that resulted in executing an illegal instruction and crashed

Ideally, we should not crash with that but should instead throw `InvalidIndexException`. Actually, we should also add some test cases around invalid index. I don't see it currently in the test template.

SwapnilGaikwad:
> Ideally, we should not crash with that but should instead throw `InvalidIndexException`. Actually, we should also add some test cases around invalid index. I don't see it currently in the test template.

Agreed. I never got to ensure from the assembly that it throws `InvalidIndexException`. 🤔 


<details>
<summary>RunBasicScenario_UnsafeRead</summary>

```
; Assembly listing for method JIT.HardwareIntrinsics.Arm.StoreSelectedScalarTest__StoreSelectedScalar_Vector128_Byte_15:RunBasicScenario_UnsafeRead():this (FullOpts)
; Emitting BLENDED_CODE for generic ARM64 - Unix
; FullOpts code
; optimized code
; fp based frame
; partially interruptible
; No PGO data
; 6 inlinees with PGO data; 27 single block inlinees; 6 inlinees without PGO data
; Final local variable assignments
;
;  V00 this         [V00,T02] (  4,  4   )     ref  ->  x19         this class-hnd single-def <JIT.HardwareIntrinsics.Arm.StoreSelectedScalarTest__StoreSelectedScalar_Vector128_Byte_15>
;# V01 OutArgs      [V01    ] (  1,  1   )  struct ( 0) [sp+0x00]  do-not-enreg[XS] addr-exposed "OutgoingArgSpace"
;  V02 tmp1         [V02,T14] (  2,  4   )    long  ->  x21         "impAppendStmt"
;  V03 tmp2         [V03,T15] (  2,  4   )    long  ->  x21         "impAppendStmt"
;  V04 tmp3         [V04,T24] (  2,  2   )     ref  ->   x0         class-hnd single-def "Inline return value spill temp" <System.String>
;* V05 tmp4         [V05,T25] (  0,  0   )     int  ->  zero-ref    single-def "Inline stloc first use temp"
;* V06 tmp5         [V06,T26] (  0,  0   )     int  ->  zero-ref    single-def "Inline stloc first use temp"
;  V07 tmp6         [V07,T01] (  7,  6.77)     ref  ->  x20         class-hnd exact single-def "Inline stloc first use temp" <System.String>
;* V08 tmp7         [V08    ] (  0,  0   )   ubyte  ->  zero-ref    single-def "Inline return value spill temp"
;* V09 tmp8         [V09    ] (  0,  0   )   ubyte  ->  zero-ref    "Inline return value spill temp"
;* V10 tmp9         [V10    ] (  0,  0   )   ubyte  ->  zero-ref    single-def "Inline return value spill temp"
;* V11 tmp10        [V11    ] (  0,  0   )   ubyte  ->  zero-ref    "Inlining Arg"
;* V12 tmp11        [V12    ] (  0,  0   )   ubyte  ->  zero-ref    "Inlining Arg"
;* V13 tmp12        [V13    ] (  0,  0   )   ubyte  ->  zero-ref    "Inlining Arg"
;  V14 tmp13        [V14,T20] (  2,  3.70)   byref  ->   x0         single-def "Inlining Arg"
;  V15 tmp14        [V15,T21] (  2,  3.70)   byref  ->   x1         single-def "Inlining Arg"
;* V16 tmp15        [V16    ] (  0,  0   )   ubyte  ->  zero-ref    "Inlining Arg"
;* V17 tmp16        [V17    ] (  0,  0   )   ubyte  ->  zero-ref    "Inlining Arg"
;* V18 tmp17        [V18    ] (  0,  0   )   ubyte  ->  zero-ref    "Inlining Arg"
;  V19 tmp18        [V19,T12] (  2,  4   )   byref  ->   x0         single-def "Inlining Arg"
;  V20 tmp19        [V20,T13] (  2,  4   )   byref  ->   x1         single-def "Inlining Arg"
;  V21 tmp20        [V21,T00] (  4,  8   )   byref  ->  x21         single-def "Inlining Arg"
;* V22 tmp21        [V22    ] (  0,  0   )    long  ->  zero-ref    ld-addr-op "Inline stloc first use temp"
;  V23 tmp22        [V23,T16] (  2,  4   )    long  ->   x0         "Inlining Arg"
;  V24 tmp23        [V24,T06] (  3,  6   )    long  ->   x1         "Inlining Arg"
;  V25 tmp24        [V25,T03] (  3,  6   )   byref  ->  x22         single-def "Inlining Arg"
;* V26 tmp25        [V26    ] (  0,  0   )    long  ->  zero-ref    ld-addr-op "Inline stloc first use temp"
;  V27 tmp26        [V27,T17] (  2,  4   )    long  ->   x0         "Inlining Arg"
;  V28 tmp27        [V28,T07] (  3,  6   )    long  ->   x1         "Inlining Arg"
;  V29 tmp28        [V29,T04] (  3,  6   )   byref  ->  x21         single-def "Inlining Arg"
;* V30 tmp29        [V30    ] (  0,  0   )    long  ->  zero-ref    ld-addr-op "Inline stloc first use temp"
;  V31 tmp30        [V31,T18] (  2,  4   )    long  ->   x0         "Inlining Arg"
;  V32 tmp31        [V32,T08] (  3,  6   )    long  ->   x1         "Inlining Arg"
;  V33 tmp32        [V33,T05] (  3,  6   )   byref  ->  x20         single-def "Inlining Arg"
;* V34 tmp33        [V34    ] (  0,  0   )    long  ->  zero-ref    ld-addr-op "Inline stloc first use temp"
;  V35 tmp34        [V35,T19] (  2,  4   )    long  ->  x22         "Inlining Arg"
;  V36 tmp35        [V36,T09] (  3,  6   )    long  ->  x20         "Inlining Arg"
;  V37 tmp36        [V37,T22] (  3,  3   )     ref  ->  x23         class-hnd exact single-def "Inline stloc first use temp" <ubyte[]>
;  V38 tmp37        [V38,T23] (  3,  3   )     ref  ->   x0         class-hnd exact single-def "Inline stloc first use temp" <ubyte[]>
;* V39 tmp38        [V39    ] (  0,  0   )    long  ->  zero-ref    "Inlining Arg"
;* V40 tmp39        [V40    ] (  0,  0   )   byref  ->  zero-ref    "Inlining Arg"
;* V41 tmp40        [V41    ] (  0,  0   )   byref  ->  zero-ref    "Inlining Arg"
;* V42 tmp41        [V42    ] (  0,  0   )   byref  ->  zero-ref    "Inlining Arg"
;  V43 cse0         [V43,T10] (  5,  5   )   byref  ->  x20         "CSE - aggressive"
;  V44 cse1         [V44,T11] (  9,  4.16)     ref  ->   x1         multi-def "CSE - moderate"
;
; Lcl frame size = 8

G_M44535_IG01:  ;; offset=0x0000
            stp     fp, lr, [sp, #-0x40]!
            stp     x19, x20, [sp, #0x18]
            stp     x21, x22, [sp, #0x28]
            str     x23, [sp, #0x38]
            mov     fp, sp
            mov     x19, x0
						;; size=24 bbWeight=1 PerfScore 5.00
G_M44535_IG02:  ;; offset=0x0018
            mov     w0, #47
            bl      System.String:FastAllocateString(int):System.String
            mov     x20, x0
            cbnz    x20, G_M44535_IG04
						;; size=16 bbWeight=1 PerfScore 3.00
G_M44535_IG03:  ;; offset=0x0028
            movz    x1, #0xE008
            movk    x1, #0x39F4 LSL #16
            movk    x1, #0xFFFF LSL #32
            mov     x0, x1
            movz    x2, #0xCF18      // code for System.Diagnostics.Debug:Fail(System.String,System.String)
            movk    x2, #0x4582 LSL #16
            movk    x2, #0xFFFF LSL #32
            ldr     x2, [x2]
            blr     x2
						;; size=36 bbWeight=0.46 PerfScore 3.47
G_M44535_IG04:  ;; offset=0x004C
            ldr     w0, [x20, #0x08]
            cmp     w0, #20
            bge     G_M44535_IG06
						;; size=12 bbWeight=0.92 PerfScore 4.16
G_M44535_IG05:  ;; offset=0x0058
            movz    x1, #0xE008
            movk    x1, #0x39F4 LSL #16
            movk    x1, #0xFFFF LSL #32
            mov     x0, x1
            movz    x2, #0xCF18      // code for System.Diagnostics.Debug:Fail(System.String,System.String)
            movk    x2, #0x4582 LSL #16
            movk    x2, #0xFFFF LSL #32
            ldr     x2, [x2]
            blr     x2
						;; size=36 bbWeight=0.46 PerfScore 3.47
G_M44535_IG06:  ;; offset=0x007C
            add     x0, x20, #12
            movz    x1, #0xFFA4
            movk    x1, #0x39F4 LSL #16
            movk    x1, #0xFFFF LSL #32
            ldp     q16, q17, [x1]
            ldr     q18, [x1, #0x18]
            stp     q16, q17, [x0]
            str     q18, [x0, #0x18]
            ldr     w0, [x20, #0x08]
            sub     w0, w0, #20
            cmp     w0, #27
            bge     G_M44535_IG08
						;; size=48 bbWeight=0.92 PerfScore 13.87
G_M44535_IG07:  ;; offset=0x00AC
            movz    x1, #0xE008
            movk    x1, #0x39F4 LSL #16
            movk    x1, #0xFFFF LSL #32
            mov     x0, x1
            movz    x2, #0xCF18      // code for System.Diagnostics.Debug:Fail(System.String,System.String)
            movk    x2, #0x4582 LSL #16
            movk    x2, #0xFFFF LSL #32
            ldr     x2, [x2]
            blr     x2
						;; size=36 bbWeight=0.46 PerfScore 3.47
G_M44535_IG08:  ;; offset=0x00D0
            add     x0, x20, #52
            movz    x1, #0xFF54
            movk    x1, #0x39F4 LSL #16
            movk    x1, #0xFFFF LSL #32
            ldp     q16, q17, [x1]
            ldr     q18, [x1, #0x20]
            ldr     q19, [x1, #0x26]
            stp     q16, q17, [x0]
            str     q18, [x0, #0x20]
            str     q19, [x0, #0x26]
            mov     x0, x20
            movz    x1, #0xD248      // code for System.Console:WriteLine(System.String)
            movk    x1, #0x45C9 LSL #16
            movk    x1, #0xFFFF LSL #32
            ldr     x1, [x1]
            blr     x1
            add     x20, x19, #40
            mov     x21, x20
            ldrsb   wzr, [x21]
            add     x0, x21, #32
            movz    x1, #0xE280      // code for System.Runtime.InteropServices.GCHandle:AddrOfPinnedObject():long:this
            movk    x1, #0x459E LSL #16
            movk    x1, #0xFFFF LSL #32
            ldr     x1, [x1]
            blr     x1
            ldr     x1, [x21, #0x10]
            add     x0, x0, x1
            sub     x0, x0, #1
            sub     x1, x1, #1
            bic     x21, x0, x1
            mov     x22, x20
            add     x0, x22, #24
            movz    x1, #0xE280      // code for System.Runtime.InteropServices.GCHandle:AddrOfPinnedObject():long:this
            movk    x1, #0x459E LSL #16
            movk    x1, #0xFFFF LSL #32
            ldr     x1, [x1]
            blr     x1
            ldr     x1, [x22, #0x10]
            add     x0, x0, x1
            sub     x0, x0, #1
            sub     x1, x1, #1
            bic     x0, x0, x1
            ldr     q16, [x0]
            st1     {v16.b}[15], [x21]
            mov     x21, x20
            add     x0, x21, #24
            movz    x1, #0xE280      // code for System.Runtime.InteropServices.GCHandle:AddrOfPinnedObject():long:this
            movk    x1, #0x459E LSL #16
            movk    x1, #0xFFFF LSL #32
            ldr     x1, [x1]
            blr     x1
            ldr     x1, [x21, #0x10]
            add     x0, x0, x1
            sub     x0, x0, #1
            sub     x1, x1, #1
            bic     x21, x0, x1
            add     x0, x20, #32
            movz    x1, #0xE280      // code for System.Runtime.InteropServices.GCHandle:AddrOfPinnedObject():long:this
            movk    x1, #0x459E LSL #16
            movk    x1, #0xFFFF LSL #32
            ldr     x1, [x1]
            blr     x1
            mov     x22, x0
            ldr     x20, [x20, #0x10]
            movz    x0, #0x1EF0
            movk    x0, #0x45BC LSL #16
            movk    x0, #0xFFFF LSL #32
            mov     x1, #16
            bl      CORINFO_HELP_NEWARR_1_VC
            mov     x23, x0
            movz    x0, #0x1EF0
            movk    x0, #0x45BC LSL #16
            movk    x0, #0xFFFF LSL #32
            mov     x1, #1
            bl      CORINFO_HELP_NEWARR_1_VC
            ldr     q16, [x21]
            str     q16, [x23, #0x10]
            add     x2, x22, x20
            sub     x2, x2, #1
            sub     x1, x20, #1
            bic     x2, x2, x1
            ldrb    w2, [x2]
            strb    w2, [x0, #0x10]
            ldrb    w2, [x0, #0x10]
            mov     x0, x19
            mov     x1, x23
            movz    x3, #0xFF48
            movk    x3, #0x39F4 LSL #16
            movk    x3, #0xFFFF LSL #32
            movz    x4, #0xD6E0      // code for JIT.HardwareIntrinsics.Arm.StoreSelectedScalarTest__StoreSelectedScalar_Vector128_Byte_15:ValidateResult(ubyte[],ubyte,System.String):this
            movk    x4, #0x45C9 LSL #16
            movk    x4, #0xFFFF LSL #32
            ldr     x4, [x4]
            blr     x4
						;; size=376 bbWeight=1 PerfScore 99.00
G_M44535_IG09:  ;; offset=0x0248
            ldr     x23, [sp, #0x38]
            ldp     x21, x22, [sp, #0x28]
            ldp     x19, x20, [sp, #0x18]
            ldp     fp, lr, [sp], #0x40
            ret     lr
						;; size=20 bbWeight=1 PerfScore 6.00
G_M44535_IG10:  ;; offset=0x025C
            bl      CORINFO_HELP_RNGCHKFAIL
            brk_unix #0
						;; size=8 bbWeight=0 PerfScore 0.00

; Total bytes of code 612, prolog size 20, PerfScore 202.64, instruction count 153, allocated bytes for code 612 (MethodHash=52aa5208) for method JIT.HardwareIntrinsics.Arm.StoreSelectedScalarTest__StoreSelectedScalar_Vector128_Byte_15:RunBasicScenario_UnsafeRead():this (FullOpts)
; ============================================================
```
</details>



<details>
<summary>StoreSelectedScalar</summary>

```
; Assembly listing for method System.Runtime.Intrinsics.Arm.AdvSimd+Arm64:StoreSelectedScalar(ulong,System.Runtime.Intrinsics.Vector128`1[ubyte],ubyte) (FullOpts)
; Emitting BLENDED_CODE for generic ARM64 - Unix
; FullOpts code
; optimized code
; fp based frame
; partially interruptible
; No PGO data
; Final local variable assignments
;
;  V00 arg0         [V00,T00] (  3,  3   )    long  ->   x0         single-def
;  V01 arg1         [V01,T02] (  3,  3   )  simd16  ->   d0         HFA(simd16)  single-def <System.Runtime.Intrinsics.Vector128`1[ubyte]>
;  V02 arg2         [V02,T01] (  3,  3   )   ubyte  ->   x1         single-def
;# V03 OutArgs      [V03    ] (  1,  1   )  struct ( 0) [sp+0x00]  do-not-enreg[XS] addr-exposed "OutgoingArgSpace"
;
; Lcl frame size = 0

G_M30164_IG01:  ;; offset=0x0000
            stp     fp, lr, [sp, #-0x10]!
            mov     fp, sp
G_M30164_IG02:  ;; offset=0x0008
            uxtb    w1, w1
            adr     x2, [G_M30164_IG03]
            add     x2, x2, x1,  LSL #3
            br      x2
						;; size=16 bbWeight=1 PerfScore 3.00
G_M30164_IG03:  ;; offset=0x0018
            st1     {v0.b}[0], [x0]
            b       G_M30164_IG19
G_M30164_IG04:  ;; offset=0x0020
            st1     {v0.b}[1], [x0]
            b       G_M30164_IG19
G_M30164_IG05:  ;; offset=0x0028
            st1     {v0.b}[2], [x0]
            b       G_M30164_IG19
G_M30164_IG06:  ;; offset=0x0030
            st1     {v0.b}[3], [x0]
            b       G_M30164_IG19
G_M30164_IG07:  ;; offset=0x0038
            st1     {v0.b}[4], [x0]
            b       G_M30164_IG19
G_M30164_IG08:  ;; offset=0x0040
            st1     {v0.b}[5], [x0]
            b       G_M30164_IG19
G_M30164_IG09:  ;; offset=0x0048
            st1     {v0.b}[6], [x0]
            b       G_M30164_IG19
G_M30164_IG10:  ;; offset=0x0050
            st1     {v0.b}[7], [x0]
            b       G_M30164_IG19
G_M30164_IG11:  ;; offset=0x0058
            st1     {v0.b}[8], [x0]
            b       G_M30164_IG19
G_M30164_IG12:  ;; offset=0x0060
            st1     {v0.b}[9], [x0]
            b       G_M30164_IG19
G_M30164_IG13:  ;; offset=0x0068
            st1     {v0.b}[10], [x0]
            b       G_M30164_IG19
G_M30164_IG14:  ;; offset=0x0070
            st1     {v0.b}[11], [x0]
            b       G_M30164_IG19
G_M30164_IG15:  ;; offset=0x0078
            st1     {v0.b}[12], [x0]
            b       G_M30164_IG19
G_M30164_IG16:  ;; offset=0x0080
            st1     {v0.b}[13], [x0]
            b       G_M30164_IG19
G_M30164_IG17:  ;; offset=0x0088
            st1     {v0.b}[14], [x0]
            b       G_M30164_IG19
G_M30164_IG18:  ;; offset=0x0090
            st1     {v0.b}[15], [x0]
						;; size=4 bbWeight=1 PerfScore 1.00
G_M30164_IG19:  ;; offset=0x0094
            ldp     fp, lr, [sp], #0x10
            ret     lr

; Total bytes of code 156, prolog size 8, PerfScore 53.10, instruction count 39, allocated bytes for code 156 (MethodHash=aade8a2b) for method System.Runtime.Intrinsics.Arm.AdvSimd+Arm64:StoreSelectedScalar(ulong,System.Runtime.Intrinsics.Vector128`1[ubyte],ubyte) (FullOpts)
; ============================================================
```
</details>

kunalspathak:
Just to confirm, are you unblocked or wanted some clarification on this PR?

SwapnilGaikwad:
> Just to confirm, are you unblocked or wanted some clarification on this PR?

I am not sure whether you would like the approach in the current version that breaks the backward compatibility but brings APIs in a consistent state.  Or you would like to keep `StoreSelectedScalar` from a single `Vector128` in `AdvSimd` class as it is in the `main` branch? If we want to split the `StoreSelectedScalar` in `AdvSimd` and `AdvSimd.Arm64` then we would need additional mechanism to disambiguate them which might not be elegant.

kunalspathak:
We should not break the backward compatibility and that's how the existing [LoadAndInsertScalar](https://github.com/dotnet/runtime/pull/93197) APIs are implemented. They are overridden while keeping name same.

kunalspathak (commented):
Changes LGTM. 

What does the code and behavior looks like for following?
1. out of bounds constant
2. out of bounds variable
3. in bound variable

SwapnilGaikwad:
> Changes LGTM.
> 
> What does the code and behavior looks like for following?
> 
> 1. out of bounds constant
 - Based on the tests in the our test templates, RunBasicScenario_Load, RunReflectionScenario_UnsafeRead, RunLclVarScenario_UnsafeRead, RunClassFldScenario, RunStructLclFldScenario throw
    ```
    System.IndexOutOfRangeException: Index was outside the bounds of the array.
    ```
 - While RunBasicScenario_UnsafeRead segfaults with the following error
    ```
    Fatal error. System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception.
    Segmentation fault (core dumped)
    ```
> 2. out of bounds variable

- Same as case 1. The RunBasicScenario_Load, RunReflectionScenario_UnsafeRead, RunLclVarScenario_UnsafeRead, RunClassFldScenario, RunStructLclFldScenario throw
    ```
    System.IndexOutOfRangeException: Index was outside the bounds of the array.
    ```
- RunBasicScenario_UnsafeRead segfaults with the following error
    ```
    Fatal error. System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception.
    Segmentation fault (core dumped)
    ```
> 3. in bound variable

- generates switch ladder and jumps to the variant of instruction with correct index

kunalspathak:
> RunBasicScenario_UnsafeRead segfaults with the following error

Interesting. What is the cause of it?

SwapnilGaikwad:
Now we get out of range exception for all the invalid scenarios (with constant or variable index) with all the test cases.
```
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
```


> I never got to ensure from the assembly that it throws `InvalidIndexException`. 🤔

This mystery is also resolved now. There was a check for array out of bounds missing. Now added it. Can be confirmed from the assembly for `StoreSelectedScalar` as well.

```
...
G_M56847_IG04:
      ldr     x0, [fp, #0x38]	// [V00 arg0]
      ldr     q16, [fp, #0x20]	// [V01 arg1]
      ldr     w1, [fp, #0x1C]	// [V02 arg2]
      uxtb    w1, w1
      cmp     w1, #4
      blo     G_M56847_IG05
      bl      CORINFO_HELP_THROW_ARGUMENTOUTOFRANGEEXCEPTION
...
```
<details>
<summary>Full assembly</summary>

```
; Assembly listing for method System.Runtime.Intrinsics.Arm.AdvSimd:StoreSelectedScalar(ulong,System.Runtime.Intrinsics.Vector128`1[int],ubyte) (MinOpts)
; Emitting BLENDED_CODE for generic ARM64 - Unix
; MinOpts code
; debuggable code
; fp based frame
; fully interruptible
; No PGO data
; Final local variable assignments
;
;  V00 arg0         [V00    ] (  1,  1   )    long  ->  [fp+0x38]  do-not-enreg[]
;  V01 arg1         [V01    ] (  1,  1   )  simd16  ->  [fp+0x20]  HFA(simd16)  do-not-enreg[S] <System.Runtime.Intrinsics.Vector128`1[int]>
;  V02 arg2         [V02    ] (  1,  1   )   ubyte  ->  [fp+0x1C]  do-not-enreg[]
;# V03 OutArgs      [V03    ] (  1,  1   )  struct ( 0) [sp+0x00]  do-not-enreg[XS] addr-exposed "OutgoingArgSpace"
;
; Lcl frame size = 48

G_M56847_IG01:  ;; offset=0x0000
            stp     fp, lr, [sp, #-0x40]!
            mov     fp, sp
            str     x0, [fp, #0x38]	// [V00 arg0]
            str     w1, [fp, #0x1C]	// [V02 arg2]
            str     q0, [fp, #0x20]	// [V01 arg1]
						;; size=20 bbWeight=1 PerfScore 4.50
G_M56847_IG02:  ;; offset=0x0014
            movz    x0, #0x45A8
            movk    x0, #0x4004 LSL #16
            movk    x0, #0xFFFF LSL #32
            ldr     w0, [x0]
            cbz     w0, G_M56847_IG04
						;; size=20 bbWeight=1 PerfScore 5.50
G_M56847_IG03:  ;; offset=0x0028
            bl      CORINFO_HELP_DBG_IS_JUST_MY_CODE
						;; size=4 bbWeight=0.50 PerfScore 0.50
G_M56847_IG04:  ;; offset=0x002C
            ldr     x0, [fp, #0x38]	// [V00 arg0]
            ldr     q16, [fp, #0x20]	// [V01 arg1]
            ldr     w1, [fp, #0x1C]	// [V02 arg2]
            uxtb    w1, w1
            cmp     w1, #4
            blo     G_M56847_IG05
            bl      CORINFO_HELP_THROW_ARGUMENTOUTOFRANGEEXCEPTION
						;; size=28 bbWeight=1 PerfScore 9.00
G_M56847_IG05:  ;; offset=0x0048
            ldr     w1, [fp, #0x1C]	// [V02 arg2]
            uxtb    w1, w1
            adr     x2, [G_M56847_IG06]
            add     x2, x2, x1,  LSL #3
            br      x2
						;; size=20 bbWeight=1 PerfScore 5.00
G_M56847_IG06:  ;; offset=0x005C
            st1     {v16.s}[0], [x0]
            b       G_M56847_IG10
G_M56847_IG07:  ;; offset=0x0064
            st1     {v16.s}[1], [x0]
            b       G_M56847_IG10
G_M56847_IG08:  ;; offset=0x006C
            st1     {v16.s}[2], [x0]
            b       G_M56847_IG10
G_M56847_IG09:  ;; offset=0x0074
            st1     {v16.s}[3], [x0]
						;; size=4 bbWeight=1 PerfScore 1.00
G_M56847_IG10:  ;; offset=0x0078
            nop
            nop
						;; size=8 bbWeight=1 PerfScore 1.00
G_M56847_IG11:  ;; offset=0x0080
            ldp     fp, lr, [sp], #0x40
            ret     lr

; Total bytes of code 136, prolog size 20, PerfScore 48.10, instruction count 34, allocated bytes for code 136 (MethodHash=5b1221f0) for method System.Runtime.Intrinsics.Arm.AdvSimd:StoreSelectedScalar(ulong,System.Runtime.Intrinsics.Vector128`1[int],ubyte) (MinOpts)
; ============================================================
```
</details>

Interestingly, for some tests I got `IndexOutOfRangeException` earlier because the test was executing without an error but failing in the validation stage where it was accessing the out of range index. Only the first test was failing with an illegal instruction probably because of the code layout. Invalid index tried to jump to location that did not contain a valid index. For the further tests, the jump target at invalid index contained a valid instruction so it did not throw any exception during the execution.

kunalspathak:
Seems the indices of few test cases needs adjustment for them to not throw Exception.

SwapnilGaikwad:
> Seems the indices of few test cases needs adjustment for them to not throw Exception.

Had to move the range check after recomputing the `simdSize` 🤦 .

src/coreclr/jit/hwintrinsicarm64.cpp
```diff
@@ -1786,6 +1787,65 @@ GenTree* Compiler::impSpecialIntrinsic(NamedIntrinsic        intrinsic,
...
+            else
+            {
+                // While storing from a single vector, both Vector128 and Vector64 API calls are in AdvSimd class.
+                // Thus, we get simdSize as 8 for both of the calls. We re-calculate that simd size for such API calls.
+                getBaseJitTypeAndSizeOfSIMDType(argClass, &simdSize);
+            }
+
+            assert(HWIntrinsicInfo::isImmOp(intrinsic, op3));
+            HWIntrinsicInfo::lookupImmBounds(intrinsic, simdSize, simdBaseType, &immLowerBound, &immUpperBound);
+            op3     = addRangeCheckForHWIntrinsic(op3, immLowerBound, immUpperBound);
```
kunalspathak, line 1831: Should use `addRangeCheckIfNeeded` so we can also get some extra asserts checks present in it.

SwapnilGaikwad: cool... When I had added `addRangeCheckIfNeeded`, the native image generation step was failing on the following assert. May need some tweaks to the assert. I'll check this.
```
Assertion failed '!immOp->IsCnsIntOrI()' in 'System.Text.Unicode.Utf8Utility:TranscodeToUtf8(ulong,int,ulong,int,byref,byref):int' during 'Importation' (IL size 2297; hash 0x85587dde; FullOpts)
```

kunalspathak: Looks like `StoreSelectedScalar` for single struct needs to be handled properly. https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Text/Unicode/Utf8Utility.Transcoding.cs,1001

SwapnilGaikwad: you're right.. needed to move the range check after recomputing the `simdSize` 🙂. 
Is there anything we need to do to push this PR further?

kunalspathak: Seems https://github.com/dotnet/runtime/pull/93223/commits/142bcc08b2f9c23d558c6ea0ab63383ed66d5238 introduced these failures on Mono because it touches the existing `StoreSelectedScalar` API. AFAIK, they worked fine before your commit. I do not know much on how to debug/repro this, but can you double check?

```
/tmp/helix/working/B11909CC/p/build/apple/AppleBuild.targets(242,5): error : Precompiling failed for /tmp/helix/working/B11909CC/w/ADA40973/e/publish/System.Private.CoreLib.dll with exit code 139. [/private/tmp/helix/working/B11909CC/w/ADA40973/e/publish/ProxyProjectForAOTOnHelix.proj]
/tmp/helix/working/B11909CC/p/build/apple/AppleBuild.targets(242,5): error :  [/private/tmp/helix/working/B11909CC/w/ADA40973/e/publish/ProxyProjectForAOTOnHelix.proj]
```

SwapnilGaikwad:
> can you double check?

Sure! It's tricky to debug, and it isn't obvious why  compilation of `System.Runtime.Serialization.SerializationGuard` would fail. I'll try rebasing/merging main, if the error persists then can debug further.

kunalspathak:
> > can you double check?
> 
> Sure! It's tricky to debug, and it isn't obvious why compilation of `System.Runtime.Serialization.SerializationGuard` would fail. I'll try rebasing/merging main, if the error persists then can debug further.

From what I remember, these tests were passing earlier and failing with the recent commit and are in arm64, so most likly related. I could be wrong.

SwapnilGaikwad:
> > > can you double check?
> > 
> > 
> > Sure! It's tricky to debug, and it isn't obvious why compilation of `System.Runtime.Serialization.SerializationGuard` would fail. I'll try rebasing/merging main, if the error persists then can debug further.
> 
> From what I remember, these tests were passing earlier and failing with the recent commit and are in arm64, so most likly related. I could be wrong.

You're right, it seems to be related to the arg check. Not sure how to debug it though. I'll try to see if I can reproduce the failure. If that doesn't progress, should I put the explicit check as before?

kunalspathak:
> If that doesn't progress, should I put the explicit check as before?

I would say just inspect the code changes and see if the existing `SelectedScalar` with single vector is unchanged. That should fix the problem.

kunalspathak:
@SwapnilGaikwad - can you resolve the merge conflicts?

fanyang-mono (approved):
The change to Mono codebase looks good to me.

kunalspathak:
> The change to Mono codebase looks good to me.

@fanyang-mono - was the change essentially to make sure that mono continue to ignore newer APIs whose 2nd parameter is tuple and only support the existing `StoreSelectedScalar` API?

fanyang-mono:
> > The change to Mono codebase looks good to me.
> 
> @fanyang-mono - was the change essentially to make sure that mono continue to ignore newer APIs whose 2nd parameter is tuple and only support the existing `StoreSelectedScalar` API?

Unfortunately, yes, Mono is not super friendly with handling API with 4 and more input arguments. We need to add additional code to handle those cases, rather than automatically handled by the machinery like Vector64/128.

Issue #93229, opened 2023-10-09
Microsoft.Gen.OptionsValidation.Test.EmitterTests.TestEmitter fails occasionally


akoeplinger:
There seems to be some non-deterministic behavior since the assert fails like this which suggests the classes were generated in a different order:

```
Assert.Equal() Failure
                                 ↓ (pos 128490)
Expected: ··· class __SourceGen__MinLengthAttribute : global::System.Compo···
Actual:   ··· class __SourceGen__RangeAttribute : global::System.Component···
                                 ↑ (pos 128490)

   at Microsoft.Gen.OptionsValidation.Test.EmitterTests.TestEmitter() in /_/src/libraries/Microsoft.Extensions.Options/tests/SourceGenerationTests/EmitterTests.cs:line 56
--- End of stack trace from previous location ---
```

## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=431517
Build error leg or test failing: Microsoft.Gen.OptionsValidation.Test.EmitterTests.TestEmitter
Pull request: https://github.com/dotnet/runtime/pull/93201

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json
{
  "ErrorMessage": "",
  "ErrorPattern": "Actual:.*class __SourceGen__RangeAttribute",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=431517
**Error message validated:** `Actual:.*class __SourceGen__RangeAttribute`
**Result validation: :white_check_mark:** Known issue matched with the provided build.
**Validation performed at:** 10/9/2023 3:03:46 PM UTC

### Report

|Build|Definition|Test|Pull Request|
|---|---|---|---|
|[432677](https://dev.azure.com/dnceng-public/public/_build/results?buildId=432677)|dotnet/runtime|[Microsoft.Gen.OptionsValidation.Test.EmitterTests.TestEmitter](https://dev.azure.com/dnceng-public/public/_build/results?buildId=432677&view=ms.vss-test-web.build-test-results-tab&runId=9582492&resultId=114378)|dotnet/runtime#93037|
[...]

PR #93231, opened 2023-10-09
Do not clone empty arrays in CloneByteArray

vcsjones:
`CloneByteArray` is used heavily in S.S.Cryptography to create defensive copies of byte arrays. This is to avoid exposing any kind of mutability to callers.

`Clone` however appears to [create a new `byte[]`][lab] for empty byte arrays every time. Empty arrays are not uncommon. It could be for a CNG property, it could be for key parameters, or other `AsnEncodedData`.

If its empty, just return the identity to save the allocation.

[lab]: https://sharplab.io/#v2:C4LgTgrgdgPgAgJgIwFgBQ6BGBPYBTAbQF0ACMEgXhKjwHcSd8CAGIgbi10NIGNKSAFI24BKMADoAwgBsA9jQEiOaYcRI8E/IV2JipchUvTo4SAJwCASngBmeMHig88AUQCOEAIbSAzgLAANOoiRmimFtZ2Dk6uHt5+AIJgYJ7Y4i4AtgAOwNgAPMIAfIpBPCHK4Va29o7O7l6+AjylCOVAA

### DIFF
--- a/src/libraries/Common/src/System/Security/Cryptography/Helpers.cs
+++ b/src/libraries/Common/src/System/Security/Cryptography/Helpers.cs
@@ -48,7 +48,7 @@ namespace Internal.Cryptography
             return src switch
             {
                 null => null,
-                [] => src,
+                { Length: 0 } => src,
                 _ => (byte[])src.Clone(),
             };
         }
--- a/src/libraries/Common/src/System/Security/Cryptography/Helpers.cs
+++ b/src/libraries/Common/src/System/Security/Cryptography/Helpers.cs
@@ -45,12 +45,12 @@ namespace Internal.Cryptography
         [return: NotNullIfNotNull(nameof(src))]
         public static byte[]? CloneByteArray(this byte[]? src)
         {
-            if (src == null)
+            return src switch
             {
-                return null;
-            }
-
-            return (byte[])(src.Clone());
+                null => null,
+                [] => src,
+                _ => (byte[])src.Clone(),
+            };
         }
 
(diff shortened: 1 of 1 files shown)


filipnavara (approved):

src/libraries/Common/src/System/Security/Cryptography/Helpers.cs
```diff
@@ -45,12 +45,12 @@ internal static partial class Helpers
...
         {
-            if (src == null)
+            return src switch
             {
-                return null;
-            }
-
-            return (byte[])(src.Clone());
+                null => null,
+                [] => src,
```
vcsjones, line 51:
```suggestion
                { Length: 0 } => src,
```

bartonjs (approved):

vcsjones:
Failure is tracked in https://github.com/dotnet/runtime/issues/93229. Merging.

src/libraries/Common/src/System/Security/Cryptography/Helpers.cs
```diff
@@ -45,12 +45,12 @@ internal static partial class Helpers
...
+            return src switch
             {
-                return null;
-            }
-
-            return (byte[])(src.Clone());
+                null => null,
+                { Length: 0 } => src,
+                _ => (byte[])src.Clone(),
+            };
```
stephentoub, line 53: FWIW, it could also have been:
```C#
return src is not null ? src.AsSpan().ToArray() : null;
```
and let span's ToArray handle the empty case via Array.Empty.

PR #93238, opened 2023-10-09
Make a copy of assembly path.

### LINKED ISSUE #93127: Assembly.Location is sometimes corrupted on .NET 8
### Description

The `Assembly.Location` property sometimes (5%-50% of the time) returns a corrupted string (most commonly "/", but sometimes other values that appear to be a truncated version of the real path with one extra character appended). This only seems to happen for the System.Net.Sockets.dll assembly and only with `COMPlus_PerfMapEnabled=1` and `LANG=sv_SE.UTF-8`. It did not happen on .NET 6.0, 3.1 or earlier versions with the same code. It does not always happen on .NET 8.

Sometimes the real application also crashes with a segfault around the same point in the code somewhere in libcoreclr.so, but I have not been able to get a proper stack trace or repro the segfault. I cannot be sure if it's part of the same problem, but happy to share a core dump privately.

### Reproduction Steps

Extract [locationbug.zip](https://github.com/dotnet/runtime/files/12832897/locationbug.zip), edit go.sh to set `DOTNET_ROOT` to your .NET 8 SDK directory, make it executable and run it.

The main method code:

```csharp
	private static void Main(string[] args)
	{
		const string hostname = "www.example.com"; // Host needs to resolve, but does not need to be a MongoDB server

		new MongoClient(new MongoClientSettings	{ Servers = new[] {new MongoServerAddress(hostname)} });
		new HttpClient().GetAsync("http://" + hostname);

		var socketAssembly = typeof(Socket).Assembly;
		Console.Error.WriteLine($"Socket assembly location: {socketAssembly.Location}");

		if (!socketAssembly.Location.EndsWith(".dll", StringComparison.Ordinal))
			throw new ApplicationException($"Unexpected location \"{socketAssembly.Location}\" for assembly FullName={socketAssembly.FullName}");
	}
```

The Mongo and HttpClient code looks bizarre, but it does not happen without those! (The real application does not use HttpClient, but actually connects to a MongoDB server using MongoClient.)

### Expected behavior

> Socket assembly location: .../dotnet/sdk-8.0-rc1/shared/Microsoft.NETCore.App/8.0.0-rc.1.23419.4/System.Net.Sockets.dll

This is the output most of the time.

### Actual behavior

Sometimes:

> Socket assembly location: /
> Unhandled exception. System.ApplicationException: Unexpected location "/" for assembly FullName=System.Net.Sockets, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a

Sometimes this happens on the first attempt, sometimes it takes 50 or so.

### Regression?

Yes, worked on .NET 6.0, 3.1 and earlier. (Not tested on 7.0.)

### Known Workarounds

Don't set `COMPlus_PerfMapEnabled=1` or `LANG=sv_SE.UTF-8`.
It also seems to be very sensitive to the order of statements.

### Configuration

.NET SDK 8.0.100-rc.1.23463.5
Linux x64
Reproduced on Linux Mint 20 VM (kernel 5.4.0-163-generic) and multiple Ubuntu 22.04.3 servers (kernels 5.15.0-86-generic and 6.2.0-1013-aws).

### Other information

_No response_

AaronRobinsonMSFT:
Fixes #93127

### DIFF
--- a/src/coreclr/vm/perfinfo.cpp
+++ b/src/coreclr/vm/perfinfo.cpp
@@ -52,7 +52,8 @@ void PerfInfo::LogImage(PEAssembly* pPEAssembly, CHAR* guid)
     SString value;
     value.Printf("%s%c%s%c%p", path.GetUTF8(), sDelimiter, guid, sDelimiter, baseAddr);
 
-    WriteLine(SL("ImageLoad"), value);
+    SString command{ SString::Literal, "ImageLoad" };
+    WriteLine(command, value);
 }
 
 // Writes a command line, with "type" being the type of command, with "value" as the command's corresponding instructions/values. This is to be used to log specific information, e.g. LogImage
--- a/src/coreclr/vm/perfinfo.cpp
+++ b/src/coreclr/vm/perfinfo.cpp
@@ -32,8 +32,8 @@ void PerfInfo::LogImage(PEAssembly* pPEAssembly, CHAR* guid)
         PRECONDITION(guid != nullptr);
     } CONTRACTL_END;
 
-    SString value;
-    const SString& path = pPEAssembly->GetPath();
+    // Nothing to log if the assembly path isn't present.
+    SString path = pPEAssembly->GetPath();
     if (path.IsEmpty())
     {
         return;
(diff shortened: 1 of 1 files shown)


jkotas (approved):
Thanks

elinor-fung (approved):

PR #93242, opened 2023-10-09
Sanitize Reg Entries in SerialPort.GetPortNames (Issue 93240)

### LINKED ISSUE #93240: SerialPort.GetPortNames not handling poor registry data
### Description

Working with a device using a SiliconLabs CP210x USB To UART bridge (device is a M-Bus Mikro-Master) on Windows 11.

When I use `SerialPort.GetPortNames` the name is reported as:

"COM9\0\0\0\0\0\0\0\0춖쪆\uffffɆ\u0004\0\0懛睄\uffff초ㄐ龜\uffff춖쪆\uffff\0\0\0\0颰뻺鴇\uffff\0\0\0\0춖쪆\uffff\u0004\0\0\0\u0002\0\0\0䚠뻯鴇\uffff\0\0\0\0\u0001\0\0\0춖쪆\uffff\u0002\0\0\0\u0004\0\0\0䚠뻯鴇\uffff\0\0\0\0䚠뻯鴇\uffff麌籀\uffff\0\0쪆\uffff춖쪆\uffff\0\0\0\0ᾷ沰誗\uffff䱓䉁\uffff\0\0\0\0컰ㄐ龜\uffff䬖闎\uffffỷ"

Device manager lists it plainly as "`"COM9"

Looking at the source, the code reads the value from the registry.

RegEdit displays the value clearly as just COM9.

However, when I duplicate the code to read the registry entry with a `GetValue` call, I see the same problematic value returned.

### Reproduction Steps

Install the SiLabs driver.
Plug in the device
Query port names via `SerialPort.GetPortNames()

### Expected behavior

Return "COM9"

### Actual behavior

Returns "COM9\0\0\0\0\0\0\0\0춖쪆\uffffɆ\u0004\0\0懛睄\uffff초ㄐ龜\uffff춖쪆\uffff\0\0\0\0颰뻺鴇\uffff\0\0\0\0춖쪆\uffff\u0004\0\0\0\u0002\0\0\0䚠뻯鴇\uffff\0\0\0\0\u0001\0\0\0춖쪆\uffff\u0002\0\0\0\u0004\0\0\0䚠뻯鴇\uffff\0\0\0\0䚠뻯鴇\uffff麌籀\uffff\0\0쪆\uffff춖쪆\uffff\0\0\0\0ᾷ沰誗\uffff䱓䉁\uffff\0\0\0\0컰ㄐ龜\uffff䬖闎\uffffỷ"

### Regression?

Unknown

### Known Workarounds

Don't use SerialPort.GetPortNames, but instead read the registry manually and sanitize the values yourself

### Configuration

_No response_

### Other information

_No response_

ctacke:
Fixes: https://github.com/dotnet/runtime/issues/93240

Edit (@krwq): Using correct syntax so that issue linking works

### DIFF
--- a/src/libraries/System.IO.Ports/src/System/IO/Ports/SerialPort.Win32.cs
+++ b/src/libraries/System.IO.Ports/src/System/IO/Ports/SerialPort.Win32.cs
@@ -26,8 +26,10 @@ namespace System.IO.Ports
                     string[] result = serialKey.GetValueNames();
                     for (int i = 0; i < result.Length; i++)
                     {
-                        // Replace the name in the array with its value.
-                        result[i] = (string)serialKey.GetValue(result[i]);
+                        // Replace the name in the array with its value, trimming at the first null character if it exists
+                        var temp = (string)serialKey.GetValue(result[i]);
+                        var end = temp.IndexOf('\0');
+                        result[i] = temp.Substring(0, end < 0 ? temp.Length : end);
                     }
                     return result;
                 }


src/libraries/System.IO.Ports/src/System/IO/Ports/SerialPort.Win32.cs
```diff
@@ -26,8 +26,10 @@ public static string[] GetPortNames()
                     string[] result = serialKey.GetValueNames();
                     for (int i = 0; i < result.Length; i++)
                     {
-                        // Replace the name in the array with its value.
-                        result[i] = (string)serialKey.GetValue(result[i]);
+                        // Replace the name in the array with its value, trimming at the first null character if it exists
+                        var temp = (string)serialKey.GetValue(result[i]);
+                        var end = temp.IndexOf('\0');
```
adamsitnik, line 31: I am not sure whether we should try to workaround the `null` characters here. Ideally [RegistryKey.GetValueNames](https://github.com/dotnet/runtime/blob/aea45f7aaa8f73cb2a585d8b14ce66634d58cb68/src/libraries/Microsoft.Win32.Registry/src/Microsoft/Win32/RegistryKey.cs#L839) should do it out of the box.

@dotnet/area-microsoft-win32 thoughts?

adamsitnik (commented):
@ctacke thank you for your contribution!

The main owner of System.IO.Ports is currently offline, in the meantime I've tagged the owners of `Microsoft.Win32.Registry` to verify whether it should be fixed here or in the API that IO Ports is simply calling.

src/libraries/System.IO.Ports/src/System/IO/Ports/SerialPort.Win32.cs
wfurt: GetValueNames looks like better choice to me. can you please give it try @ctacke ?

ctacke: I'm not sure what the suggestion is.  The Value Name is not the issue, the Value itself is.  I can't "try GetValueNames" as a resolution, because it doesn't give the data I need.

You could certainly argue that `Registry.GetValue` should be stopping at the first null, but I don't know if there are other instances where a null-containing string might be valid data.  If null-containing strings are generally considered invalid data for registry entries, then yes, the bug would be in `Registry.GetValue`

wfurt: I may be wrong but it seems like this is what the method does e.g. terminates on null (while the `GetValue` does not)
https://learn.microsoft.com/en-us/dotnet/api/microsoft.win32.registrykey.getvaluenames?view=net-8.0#microsoft-win32-registrykey-getvaluenames

Perhaps @adamsitnik  can explain more.

krwq:
@ctacke could you please investigate if the recommended API is suitable for usage? I don't have any hardware where I could try this out

Issue #92975, opened 2023-10-03
Tracking issue for remaining AssemblyBuilder.Save work in .NET 9


buyaa-n:
See https://github.com/dotnet/runtime/issues/62956 for work done in .NET 8 for adding `AssemblyBuilder.Save(string/stream)` support. Remaining work include:
- [x] Add new `ILGenerator` implementation
  - [x] [Initial ILGenerator implementation, save minimal/simple IL](https://github.com/dotnet/runtime/pull/92846).
  - [x] [Calculate the maxstack appropriately](https://github.com/dotnet/runtime/pull/93244)
  - [x] [Add support for emitting a call to method, constructor, field, type references](https://github.com/dotnet/runtime/pull/94116)  
  - [x] [Add local variables support](https://github.com/dotnet/runtime/pull/93809)
  - [x] [Add label (branching) support](https://github.com/dotnet/runtime/pull/93565)
  - [x] [Add Try/Catch/Finally/Filter/Fault block support](https://github.com/dotnet/runtime/pull/94366)
  - [x] [Add support for remaining Emit APIs (EmitCall, EmitCalli, Emit Signatures)](https://github.com/dotnet/runtime/pull/96362)
- [x] Add support for remaining members
  - [x] [Add `ConstructorBuilderImpl` and save it to file/stream (this should include `Emit(OpCode opcode, ConstructorInfo con)` implementation)](https://github.com/dotnet/runtime/pull/94732)
  - [x] [Add `PropertyBuilderImp`l and save it to file/stream](https://github.com/dotnet/runtime/pull/95303)
  - [x] [Add `EventBuildterImpl` and save it to file/stream](https://github.com/dotnet/runtime/pull/95936)
  - [x] [Add implementation for remaining unimplemented APIs in `AssemblyBuilder`/`ModuleBuilder` (`CreateGlobalFuntions`, `DefineGlobalMethod` etc)](https://github.com/dotnet/runtime/pull/96805)
  - [x] [Add implementation for remaining unimplemented APIs for a `MethodBuilder`/`TypeBuilder`/`FieldBuilder` (like `SignatureCallingConvention`, `*RequiredCustomModifiers`, `*OptionalCustomModifiers`)](https://github.com/dotnet/runtime/pull/96805)
- [x] Add [approved public APIs](https://github.com/dotnet/runtime/issues/83988): https://github.com/dotnet/runtime/pull/97177
- [x] Entry point support
[...]

Issue #88102, opened 2023-06-27
Native crash on System.Text.Json.SourceGeneration.Roslyn4.4.Tests net8.0-linux-Debug-arm64-Mono_release


Jozkee:
## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=322652
Build error leg or test failing: System.Text.Json.SourceGeneration.Roslyn4.4.Tests.WorkItemExecution
Pull request: https://github.com/dotnet/runtime/pull/85378
Console log: https://[helixre8s23ayyeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-85378-merge-9106d2de4c074ca4a0/System.Text.Json.SourceGeneration.Roslyn4.4.Tests/1/console.e281cc02.log?helixlogtype=result](https://helixre8s23ayyeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-85378-merge-9106d2de4c074ca4a0/System.Text.Json.SourceGeneration.Roslyn4.4.Tests/1/console.e281cc02.log?helixlogtype=result)

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json
{
  "ErrorMessage": "Segmentation fault      (core dumped) \"$RUNTIME_PATH/dotnet\" exec --runtimeconfig System.Text.Json.SourceGeneration.Roslyn4.4.Tests.runtimeconfig.json --depsfile System.Text.Json.SourceGeneration.Roslyn4.4.Tests.deps.json xunit.console.dll System.Text.Json.SourceGeneration.Roslyn4.4.Tests.dll -xml testResults.xml -nologo -nocolor -notrait category=IgnoreForCI -notrait category=OuterLoop -notrait category=failing $RSP_FILE",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

### Report

|Build|Definition|Test|Pull Request|
|---|---|---|---|
[...]

Issue #88027, opened 2023-06-26
Test failure BasicEventSourceTests.TestsManifestGeneration.Test_EventSource_EtwManifestGenerationRollover


v-wenyuxu:
**Failed in:** [runtime-coreclr libraries-pgo 20230625.1](https://dev.azure.com/dnceng-public/public/_build/results?buildId=320072&view=ms.vss-test-web.build-test-results-tab&runId=6562706&resultId=134419&paneView=debug)

**Failed tests:**
```
net8.0-windows-Release-arm64-CoreCLR_checked-defaultpgo-Windows.11.Arm64.Open
   - BasicEventSourceTests.TestsManifestGeneration.Test_EventSource_EtwManifestGenerationRollover
```

**Error message:**
```
Microsoft.DotNet.RemoteExecutor.RemoteExecutionException : Remote process failed with an unhandled exception.
```

**Stack trace:**
```
 Child exception:
  System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Diagnostics.Tracing.Session.TraceEventSession.get_ProviderNameToGuid()
   at Microsoft.Diagnostics.Tracing.Session.TraceEventProviders.GetProviderGuidByName(String name)
   at Microsoft.Diagnostics.Tracing.Session.TraceEventSession.EnableProvider(String providerName, TraceEventLevel providerLevel, UInt64 matchAnyKeywords, TraceEventProviderOptions options)
   at BasicEventSourceTests.TestsManifestGeneration.<Test_EventSource_EtwManifestGenerationRollover>b__3_0() in /_/src/libraries/System.Diagnostics.Tracing/tests/BasicEventSourceTest/TestsManifestGeneration.Etw.cs:line 98
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr) in /_/src/libraries/System.Private.CoreLib/src/System/Reflection/MethodInvoker.cs:line 59

Child process:
  System.Diagnostics.Tracing.Tests, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51 BasicEventSourceTests.TestsManifestGeneration Void <Test_EventSource_EtwManifestGenerationRollover>b__3_0()


```

v-wenyuxu:
[...]

PR #93244, opened 2023-10-09
Add initial MaxStack calculation

### LINKED ISSUE #93419: [API Proposal]: Make OpCode.StackChange() public
### Background and motivation

[OpCodes](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/OpCodes.cs#L258) represents MS IL instructions that used for emitting IL with [ILGenerator](https://learn.microsoft.com/dotnet/api/system.reflection.emit.ilgenerator). An `OpCode` instruction may pop value(s) from stack as an operand of the instruction and also may push value(s) into the stack as a result.  This stack change is needed for calculating [MaxStack](https://learn.microsoft.com/dotnet/api/system.reflection.methodbody.maxstacksize) value that is essential for populating method body. I could use the   `StackBehaviourPop` and `StackBehaviourPush` properties of `OpCode` to calculate this, but there is an internal method that gives this info directly:
https://github.com/dotnet/runtime/blob/946c5245aea149d9ece53fd0debc18328c29083b/src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/Opcode.cs#L48-L49
Making this public will be useful for calculating `MaxStack` faster
Related to https://github.com/dotnet/runtime/pull/93244/files#r1353013060

### API Proposal

```diff
namespace System.Reflection.Emit

public readonly partial struct OpCode : System.IEquatable<System.Reflection.Emit.OpCode>
{
    ...
    public StackBehaviour StackBehaviourPop { get }
    public StackBehaviour StackBehaviourPush { get }
    ...
+   public int EvaluationStackDelta { get }
}
```
### Alternative Designs

Other naming options: `EvaluationStackDelta`, `StackSizeDelta`, `StackTransitionDelta` or `StackDepthDelta` as it is showing the difference of evaluation stack transition.

### API Usage

```csharp
    private void UpdateStackSize(OpCode opCode)
    {
        _currentStack += opCode.EvaluationStackDelta;
        _maxStackSize = Math.Max(_maxStackSize, _currentStack);
    }
```

buyaa-n:
This is initial MaxStack calculation for simple (no branching) instructions, basically `UpdateStackSize(OpCode opCode)` will be called for each `OpCode` emit and updates the `_currentStack` based on its `StackBehaviourPush`, `StackBehaviourPop` values.
```cs
        private void UpdateStackSize(OpCode opCode)
        {
            _currentStack += GetStackChangeFor(opCode.StackBehaviourPush) + GetStackChangeFor(opCode.StackBehaviourPop);

            if (_currentStack > _maxStackSize)
            {
                _maxStackSize = _currentStack;
            }
        }
```
This calculation will be updated as needed when branching IL.Emit overloads implemented. 

As I cannot use the internal [OpCode.StackChange()](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/Opcode.cs#L48-L49) method created `int GetStackChangeFor(StackBehaviour stackBehaviour)` method

Fixes https://github.com/dotnet/runtime/issues/93419


src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -42,40 +41,88 @@ internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
...
+            UpdateStackSize(opcode);
+        }
 
-            // TODO: for now only count the Opcodes emitted, in order to calculate it correctly we might need to make internal Opcode APIs public
-            // https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/Opcode.cs#L48
-            _maxStackSize++;
+        public override void Emit(OpCode opcode)
+        {
+            EmitOpcode(opcode);
         }
```
AaronRobinsonMSFT, line 103:
```suggestion
        public override void Emit(OpCode opcode) => EmitOpcode(opcode);
```

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -86,98 +133,91 @@ public override void Emit(OpCode opcode, int arg)
...
+                    EmitOpcode(arg switch
                     {
-                        0 => ILOpCode.Ldarg_0,
-                        1 => ILOpCode.Ldarg_1,
-                        2 => ILOpCode.Ldarg_2,
-                        _ => ILOpCode.Ldarg_3,
+                        0 => OpCodes.Ldarg_0,
+                        1 => OpCodes.Ldarg_1,
+                        2 => OpCodes.Ldarg_2,
+                        _ => OpCodes.Ldarg_3,
```
AaronRobinsonMSFT, line 167: This condition will handle all negative numbers too, is that what we want? I think an exception or some other other mode would be more appropriate.

buyaa-n: I thought when casting to `uint` negative values would become positive large number something that > 3, no?

AaronRobinsonMSFT: Yes, that is true. My statement was more related to the notion of all unknown values being converted to `Ldarg_3`. I think `Ldarg_0` or an exception being the failure mode makes more sense.

buyaa-n: I am OK with defaulting to `Ldarg_0`, though this is how the existing CoreLib `ILGenerator` does, maybe we should keep it as is for compat, what you think?

AaronRobinsonMSFT: Boo. Your instinct here is correct - keep the same behavior.

src/libraries/System.Reflection.Emit/tests/PersistableAssemblyBuilder/AssemblySaveILGeneratorTests.cs
```diff
@@ -168,5 +168,81 @@ public void MultipleTypesWithMultipleMethods()
...
+        }
+
+        private MethodInfo LoadILGenerator_GetMaxStackSizeMethod()
+        {
+            Type ilgType = Type.GetType("System.Reflection.Emit.ILGeneratorImpl, System.Reflection.Emit", throwOnError: true)!;
+            return ilgType.GetMethod("GetMaxStackSize", BindingFlags.NonPublic | BindingFlags.Instance, Type.EmptyTypes);
+        }
     }
-}
+}
```
buyaa-n, line 248:
```suggestion
}
```

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -42,40 +41,82 @@ internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
         public override LocalBuilder DeclareLocal(Type localType, bool pinned) => throw new NotImplementedException();
         public override Label DefineLabel() => throw new NotImplementedException();
 
-        public override void Emit(OpCode opcode)
+        private static int GetStackChangeFor(StackBehaviour stackBehaviour) =>
+            stackBehaviour switch
+            {
+                StackBehaviour.Pop0
+                or StackBehaviour.Push0 => 0,
```
stephentoub, line 48: I'm finding the formatting here difficult to read. Can you at least add a blank line after each line that has a => in it, in order to separate the groupings?

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -42,40 +41,82 @@ internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
...
+                or StackBehaviour.Popref_popi_popref => -3,
+                StackBehaviour.Push1
+                or StackBehaviour.Pushi
+                or StackBehaviour.Pushi8
+                or StackBehaviour.Pushr4
+                or StackBehaviour.Pushr8
+                or StackBehaviour.Pushref
+                or StackBehaviour.Varpush => 1,
+                StackBehaviour.Push1_push1 => 2,
+                _ => throw new InvalidOperationException()
```
buyaa-n, line 76:
```suggestion
                or StackBehaviour.Push0 => 0,
                
                StackBehaviour.Pop1
                or StackBehaviour.Popi
                or StackBehaviour.Popref
                or StackBehaviour.Varpop => -1,
                
                StackBehaviour.Pop1_pop1
                or StackBehaviour.Popi_pop1
                or StackBehaviour.Popi_popi
                or StackBehaviour.Popi_popi8
                or StackBehaviour.Popi_popr4
                or StackBehaviour.Popi_popr8
                or StackBehaviour.Popref_pop1
                or StackBehaviour.Popref_popi => -2,
                
                StackBehaviour.Popi_popi_popi
                or StackBehaviour.Popref_popi_pop1
                or StackBehaviour.Popref_popi_popi
                or StackBehaviour.Popref_popi_popi8
                or StackBehaviour.Popref_popi_popr4
                or StackBehaviour.Popref_popi_popr8
                or StackBehaviour.Popref_popi_popref => -3,
                
                StackBehaviour.Push1
                or StackBehaviour.Pushi
                or StackBehaviour.Pushi8
                or StackBehaviour.Pushr4
                or StackBehaviour.Pushr8
                or StackBehaviour.Pushref
                or StackBehaviour.Varpush => 1,
                
                StackBehaviour.Push1_push1 => 2,
                
                _ => throw new InvalidOperationException()
```

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -42,40 +41,88 @@ internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
         public override LocalBuilder DeclareLocal(Type localType, bool pinned) => throw new NotImplementedException();
         public override Label DefineLabel() => throw new NotImplementedException();
 
-        public override void Emit(OpCode opcode)
+        private static int GetStackChangeFor(StackBehaviour stackBehaviour) =>
```
buyaa-n, line 44: If we make the below API public, we don't need this `GetStackChangeFor` method and calculating the `StackBehaviourPush`, `StackBehaviourPop` diff with  `GetStackChangeFor(opCode.StackBehaviourPush) + GetStackChangeFor(opCode.StackBehaviourPop)` https://github.com/dotnet/runtime/blob/715293eeeddf76593679fba7ab8cfd451980225b/src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/Opcode.cs#L48-L49

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -42,40 +41,88 @@ internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
...
+                or StackBehaviour.Pushi
+                or StackBehaviour.Pushi8
+                or StackBehaviour.Pushr4
+                or StackBehaviour.Pushr8
+                or StackBehaviour.Pushref
+                or StackBehaviour.Varpush => 1,
+                
+                StackBehaviour.Push1_push1 => 2,
+                
+                _ => throw new InvalidOperationException()
```
buyaa-n, line 82: Oops spacing
```suggestion

                StackBehaviour.Pop1
                or StackBehaviour.Popi
                or StackBehaviour.Popref
                or StackBehaviour.Varpop => -1,

                StackBehaviour.Pop1_pop1
                or StackBehaviour.Popi_pop1
                or StackBehaviour.Popi_popi
                or StackBehaviour.Popi_popi8
                or StackBehaviour.Popi_popr4
                or StackBehaviour.Popi_popr8
                or StackBehaviour.Popref_pop1
                or StackBehaviour.Popref_popi => -2,

                StackBehaviour.Popi_popi_popi
                or StackBehaviour.Popref_popi_pop1
                or StackBehaviour.Popref_popi_popi
                or StackBehaviour.Popref_popi_popi8
                or StackBehaviour.Popref_popi_popr4
                or StackBehaviour.Popref_popi_popr8
                or StackBehaviour.Popref_popi_popref => -3,

                StackBehaviour.Push1
                or StackBehaviour.Pushi
                or StackBehaviour.Pushi8
                or StackBehaviour.Pushr4
                or StackBehaviour.Pushr8
                or StackBehaviour.Pushref
                or StackBehaviour.Varpush => 1,

                StackBehaviour.Push1_push1 => 2,

                _ => throw new InvalidOperationException()
```

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
buyaa-n: As per offline discussion created [API proposal](https://github.com/dotnet/runtime/issues/93419) for making this public, updated the PR accordingly

src/libraries/System.Reflection.Emit/tests/PersistableAssemblyBuilder/AssemblySaveILGeneratorTests.cs
```diff
@@ -168,5 +168,81 @@ public void MultipleTypesWithMultipleMethods()
...
+
+                MethodInfo getMaxStackSizeMethod = LoadILGenerator_GetMaxStackSizeMethod();
+                Assert.Equal(9, getMaxStackSizeMethod.Invoke(il1, new object[0]));
+                Assert.Equal(3, getMaxStackSizeMethod.Invoke(il2, new object[0]));
+
+                Assembly assemblyFromDisk = AssemblySaveTools.LoadAssemblyFromPath(file.Path);
+                Type typeFromDisk = assemblyFromDisk.Modules.First().GetType("MyType");
+                MethodBody body1 = typeFromDisk.GetMethod("Method1").GetMethodBody();
+                MethodBody body2 = typeFromDisk.GetMethod("Method2").GetMethodBody();
+                Assert.Equal(9, body1.MaxStackSize);
```
steveharter, line 237: Having an aggregate comparison for stack size like this is fine, but may not detect bugs at an opcode level since, for example, an off-by-one in one case may be offset by an opposite off-by-negative-one case. Should we check the current stack size after each Emit(opcode) by using reflection to investigate the current stack size in the generator (`m_curDepth`) or is this redundant in general due to other tests?

steveharter (approved):
LGTM pending API approval.

steveharter:
Note there are some unaccounted test failures. Perhaps rebase\merge to latest since I see some recent JIT fixes went in.

src/libraries/System.Reflection.Emit/tests/PersistableAssemblyBuilder/AssemblySaveILGeneratorTests.cs
buyaa-n: > Should we check the current stack size after each Emit(opcode) by using reflection to investigate the current stack size in the generator (m_curDepth) or is this redundant in general due to other tests?

We can check stack size after each Emit(opcode), but might be redundant, especially in this PR only simple Emit operations covered that changes the stack linearly. My plan was keep checking the `MaxStackSize` for next PRs that add more complicated Emit overloads. It might be good to test current stack size for those Emit overloads that change the stack differently.

src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/Opcode.cs
```diff
@@ -45,7 +45,7 @@ internal OpCode(OpCodeValues value, int flags)
         internal bool EndsUncondJmpBlk() =>
             (m_flags & EndsUncondJmpBlkFlag) != 0;
 
-        internal int StackChange() =>
+        public int EvaluationStackDelta =>
```
jkotas, line 48: This needs doc comment.

src/libraries/System.Reflection.Primitives/ref/System.Reflection.Primitives.cs
```diff
@@ -33,6 +33,7 @@ public enum FlowControl
         public override bool Equals([System.Diagnostics.CodeAnalysis.NotNullWhenAttribute(true)] object? obj) { throw null; }
         public bool Equals(System.Reflection.Emit.OpCode obj) { throw null; }
         public override int GetHashCode() { throw null; }
+        public int EvaluationStackDelta { get { throw null; } }
```
jkotas, line 36: Is this order produced by the ref assembly generator?
 
If you have not done it already, you may want to regenerate this ref assembly source (https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/updating-ref-source.md).

src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/Opcode.cs
buyaa-n: Yes, thank you!

buyaa-n: Added

src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/Opcode.cs
```diff
@@ -45,7 +45,13 @@ internal OpCode(OpCodeValues value, int flags)
         internal bool EndsUncondJmpBlk() =>
             (m_flags & EndsUncondJmpBlkFlag) != 0;
 
-        internal int StackChange() =>
+        /// <summary>
+        /// The value of how the IL instruction changes the evaluation stack.
+        /// </summary>
+        /// <remarks>
+        /// The difference between how many elements are popped from the stack and how many are pushed onto the stack as a result of the IL instruction.
```
jkotas, line 52: This needs to mention what this API returns when this is unknown (e.g. for `call` instruction)

jkotas:
Add test for the new API under https://github.com/dotnet/runtime/blob/fe043b489f69a888ad2efd97fbc177c1f25e541d/src/libraries/System.Reflection.Emit/tests/OpCodesTests.cs

src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/Opcode.cs
```diff
@@ -45,7 +45,15 @@ internal OpCode(OpCodeValues value, int flags)
...
+        /// For some IL instructions like <see cref="OpCodes.Call"/> this value is not fixed and depends on the called reference signature, for example if it is
```
jkotas, line 53: You are right that the actual value is not fixed, but this method is going to return 0 in this case. It means that 0 returned by this method means either zero delta or unknown delta.

Would it make more sense to throw an exception for unknown delta instead of returning 0?

buyaa-n: > Would it make more sense to throw an exception for unknown delta instead of returning 0?

That would change the existing `RuntimeILGenerator` behavior, I think I should note that in that case user should not rely on this value for calculating `StackSize` nor for calculating `MaxStack`.

buyaa-n: > That would change the existing `RuntimeILGenerator` behavior 

Actually, it's not have to affect `RuntimeILGenerator` as we can handle the throwing in the new ILGenerator implementation, though not sure such special casing is worth, don't you think adding a `NOTE` in the doc could be enough?

buyaa-n: > Actually, it's not have to affect `RuntimeILGenerator` as we can handle the throwing in the new ILGenerator implementation, 

I take it back, handling the throwing in the new ILGenerator doesn't make sense 😄

src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/Opcode.cs
```diff
@@ -45,7 +45,15 @@ internal OpCode(OpCodeValues value, int flags)
...
+        /// calling a method, depends on the method signature (push count depends on the returning a value, pop count depends how many parameters passed).
```
buyaa-n, line 54: How is this look?
```suggestion
        /// For some IL instructions like <see cref="OpCodes.Call"/> this value is not fixed and depends on the called reference signature. 
        /// In this case you should not rely on this value for calculating stack size and/or max stack, instead need to evaluate the reference signature.
        /// For example, in case it is calling a method, need to evaluate the method signature, the push count depends on the returning value,
        /// the pop count depends on how many parameters passed.
```

jkotas: In the first sentence, "this value" means "The difference between how many elements are popped from the stack and how many are pushed onto the stack".

In the second sentence, "this value" means "the value returned by this property".

Avoid saying "this value" for clarity?

Also, I think we should document what the function returns when the value is not fixed.

buyaa-n:
Failures unrelated and already reported in https://github.com/dotnet/runtime/issues/88102, https://github.com/dotnet/runtime/issues/88027

Issue #92917, opened 2023-10-03
Rename `dns.lookupS.duration` to `dns.lookup.duration`


antonfirsov:
The name implemented in #89813 is result of a misunderstanding. https://github.com/lmolkova/semantic-conventions/blob/dotnet8-metrics/docs/dotnet/dotnet-dns-metrics.md#metric-dnslookupsduration has a typo in the title, but lower in the table it's correctly spelled as `dns.lookup.duration` which is a name that aligns with OTel conventions.

Issue #92837, opened 2023-09-29
`dns.lookup.duration` should report `error.type`


lmolkova:
### Description

Currently `dns.lookup.duration` only reports duration of he successful lookups, but failed lookups are quite interesting as well (count, rate, duration, failure reason). 

Having `error.type` attribute on the `dns.lookup.duration` metric and reporting it when DNS lookup fails would provide all the details.

Adding an attribute later is considered a breaking change on OTel, since it increases the number of time series in the metric and breaks alerts and dashboards (so creating it as a bug).

### Reproduction Steps

N/A 

https://github.com/dotnet/runtime/blob/e89daba754e84dba3a86c92e8c2a714a75a2f0d0/src/libraries/System.Net.NameResolution/src/System/Net/NameResolutionMetrics.cs#L24

### Expected behavior

The metric should have `error.type` attribute nad report duration of both, failed and successfully lookups.

### Actual behavior

Metric only reports successful lookups

MihaZupan:
> Currently `dns.lookup.duration` only reports duration of he successful lookups

Where are you seeing that? We're using the same code for both success & failure cases, and have tests for the failure case here
https://github.com/dotnet/runtime/blob/d1972fe4ec6b58097b9c7641308f6827e356ab2a/src/libraries/System.Net.NameResolution/tests/FunctionalTests/MetricsTest.cs#L45

> Having `error.type` attribute on the `dns.lookup.duration` metric and reporting it when DNS lookup fails would provide all the details.

What sort of information would you expect to see there?

lmolkova:
> Where are you seeing that? We're using the same code for both success & failure cases, and have tests for the failure case here

Thanks for the clarification! So it records all durations regardless of the result - I didn't realize that.

In this case, users are just not able to separate successful attempts from failed ones.
I'd like to see an exception type captured in `error.type` (or, if there is any enum-like list of DNS errors this would also be a great option).

antonfirsov:
@karelz after discussing with @noahfalk and @lmolkova, we think the addition of `error.type` is something we should get into 8.0:

- Attribute additions are breaking changes according to otel, making the negative impact higher if we add it in future releases.
- It's easy / low risk to implement

antonfirsov:
Linking the related aspnet issue: https://github.com/dotnet/aspnetcore/issues/51029

MihaZupan:
> Attribute additions are breaking changes according to otel, making the negative impact higher if we add it in future releases

[...]

PR #93254, opened 2023-10-09
Adjust the DNS lookup duration metric

antonfirsov:
- #92917
- #92837

This is a necessary change  to ensure the metric is in sync with the spec and the OTel recommendations. An 8.0 backport is planned to avoid future breaking changes.

### DIFF
--- a/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
+++ b/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
@@ -29,7 +29,7 @@ namespace System.Net
                 throw;
             }
 
-            NameResolutionTelemetry.Log.AfterResolution(string.Empty, startingTimestamp, errorType: null);
+            NameResolutionTelemetry.Log.AfterResolution(string.Empty, startingTimestamp);
 
             if (NetEventSource.Log.IsEnabled()) NetEventSource.Info(null, name);
             return name;
@@ -400,7 +400,7 @@ namespace System.Net
                 throw;
             }
 
-            NameResolutionTelemetry.Log.AfterResolution(hostName, startingTimestamp, errorType: null);
+            NameResolutionTelemetry.Log.AfterResolution(hostName, startingTimestamp);
 
             return result;
         }
@@ -440,7 +440,7 @@ namespace System.Net
                 throw;
             }
 
-            NameResolutionTelemetry.Log.AfterResolution(address, startingTimestamp, errorType: null);
+            NameResolutionTelemetry.Log.AfterResolution(address, startingTimestamp);
 
             // Do the forward lookup to get the IPs for that host name
             startingTimestamp = NameResolutionTelemetry.Log.BeforeResolution(name);
@@ -470,7 +470,7 @@ namespace System.Net
                 throw;
             }
 
-            NameResolutionTelemetry.Log.AfterResolution(name, startingTimestamp, errorType: null);
+            NameResolutionTelemetry.Log.AfterResolution(name, startingTimestamp);
 
             // One of three things happened:
             // 1. Success.
@@ -594,32 +594,22 @@ namespace System.Net
             static async Task<T> CompleteAsync(Task task, string hostName, long startingTimestamp)
             {
                 _ = NameResolutionTelemetry.Log.BeforeResolution(hostName);
-                string? errorType = null;
+                Exception? exception = null;
                 try
                 {
                     return await ((Task<T>)task).ConfigureAwait(false);
                 }
                 catch (Exception ex)
                 {
-                    errorType = GetErrorType(ex);
+                    exception = ex;
                     throw;
                 }
                 finally
                 {
-                    NameResolutionTelemetry.Log.AfterResolution(hostName, startingTimestamp, errorType);
+                    NameResolutionTelemetry.Log.AfterResolution(hostName, startingTimestamp, exception);
                 }
             }
         }
-
-        private static string GetErrorType(Exception exception) => (exception as SocketException)?.SocketErrorCode switch
-        {
-            SocketError.HostNotFound => "host_not_found",
-            SocketError.TryAgain => "try_again",
-            SocketError.AddressFamilyNotSupported => "address_family_not_supported",
-            SocketError.NoRecovery => "no_recovery",
-
-            _ => exception.GetType().Name
-        };
 
         private static IPHostEntry CreateHostEntryForAddress(IPAddress address) =>
             new IPHostEntry
@@ -643,8 +633,7 @@ namespace System.Net
 
         private static bool LogFailure(object hostNameOrAddress, long? startingTimestamp, Exception exception)
         {
-            string errorType = GetErrorType(exception);
-            NameResolutionTelemetry.Log.AfterResolution(hostNameOrAddress, startingTimestamp, errorType);
+            NameResolutionTelemetry.Log.AfterResolution(hostNameOrAddress, startingTimestamp, exception);
             return false;
         }
 
--- a/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
+++ b/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
@@ -577,7 +577,7 @@ namespace System.Net
         }
 
         private static Task<T>? GetAddrInfoWithTelemetryAsync<T>(string hostName, bool justAddresses, AddressFamily addressFamily, CancellationToken cancellationToken)
-     where T : class
+            where T : class
         {
             long startingTimestamp = Stopwatch.GetTimestamp();
             Task? task = NameResolutionPal.GetAddrInfoAsync(hostName, justAddresses, addressFamily, cancellationToken);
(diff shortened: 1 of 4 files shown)


src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
```diff
@@ -577,7 +577,7 @@ private static Task GetHostEntryOrAddressesCoreAsync(string hostName, bool justR
         }
 
         private static Task<T>? GetAddrInfoWithTelemetryAsync<T>(string hostName, bool justAddresses, AddressFamily addressFamily, CancellationToken cancellationToken)
-             where T : class
+     where T : class
```
MihaZupan, line 580: Nit: the indentation looks off

src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
```diff
@@ -594,19 +594,33 @@ private static Task GetHostEntryOrAddressesCoreAsync(string hostName, bool justR
...
 
+        private static string GetErrorType(Exception exception) => (exception as SocketException)?.SocketErrorCode switch
+        {
+            SocketError.HostNotFound => "host_not_found",
+            SocketError.TryAgain => "try_again",
+            SocketError.AddressFamilyNotSupported => "address_family_not_supported",
+            SocketError.NoRecovery => "no_recovery",
+
+            _ => exception.GetType().Name
+        };
```
antonfirsov, line 622: I picked the codes we can realistically get from the PAL by looking at our [mapping in the Unix PAL](https://github.com/dotnet/runtime/blob/94f212275b2f51ca67025d677d7d5c5bc75f670f/src/libraries/System.Net.NameResolution/src/System/Net/NameResolutionPal.Unix.cs#L23-L46), and [`getaddrinfo` return values on Windows](https://learn.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getaddrinfo#return-value).
There are some other codes that may theoretically occur in practice. Ie. seeing `InvalidArgument` , `TypeNotFound` or `SocketNotSupported` means a programming error in our PAL if I'm not mistaken.

MihaZupan (approved):

JamesNK (approved):
Change looks good. Details about the error attribute need to be added to https://github.com/dotnet/docs/pull/37213 and/or the OTEL semantic conventions

src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
```diff
@@ -594,19 +594,33 @@ private static Task GetHostEntryOrAddressesCoreAsync(string hostName, bool justR
...
                 }
                 finally
                 {
-                    NameResolutionTelemetry.Log.AfterResolution(hostName, startingTimestamp, successful: result is not null);
+                    NameResolutionTelemetry.Log.AfterResolution(hostName, startingTimestamp, errorType);
                 }
             }
         }
 
+        private static string GetErrorType(Exception exception) => (exception as SocketException)?.SocketErrorCode switch
```
stephentoub, line 614: These strings are specifically for metrics, right? Please add a comment on the method to that effect.  Where do these result strings come from? Are they strings dictated by otel, or are you just snake_casing the names from the enum?

src/libraries/System.Net.NameResolution/src/System/Net/NameResolutionMetrics.cs
```diff
@@ -13,15 +13,25 @@ internal static class NameResolutionMetrics
...
         private static readonly Histogram<double> s_lookupDuration = s_meter.CreateHistogram<double>(
-            name: "dns.lookups.duration",
+            name: "dns.lookup.duration",
             unit: "s",
             description: "Measures the time taken to perform a DNS lookup.");
 
         public static bool IsEnabled() => s_lookupDuration.Enabled;
 
-        public static void AfterResolution(TimeSpan duration, string hostName)
+        public static void AfterResolution(TimeSpan duration, string hostName, string? errorType)
```
stephentoub, line 22: You're explicitly passing null for errorType in multiple places. You could make it optional here with a default null value and then avoid saying `, errorType: null` at each of those call sites.

src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
antonfirsov: > These strings are specifically for metrics, right?

Yes.

> Where do these result strings come from?

They are defined by us, by mapping enum values to snake-cased strings which is an OTel convention. [We do the same in HTTP metrics](https://github.com/dotnet/runtime/blob/1d760d6de8250eecb5b3711957e0393a796c89b5/src/libraries/System.Net.Http/src/System/Net/Http/Metrics/MetricsHandler.cs#L157-L173).

stephentoub: >  which is an OTel convention

Why don't we do the same for the `_ =>` case then? i.e. it seems like our general goal here is to take any SocketError value or Exception name, and snake case it, and we're then just optimizing a few cases that we expect to be most common by hardcoding the transformation. In the name of robustness, shouldn't we be dynamically creating the snak_cased version of the resulting SocketError if there is one (outside of the range we've hardcoded) or for the exception type?

src/libraries/System.Net.NameResolution/src/System/Net/NameResolutionMetrics.cs
```diff
@@ -19,19 +19,29 @@ internal static class NameResolutionMetrics
...
         }
+
+        private static string GetErrorType(Exception exception) => (exception as SocketException)?.SocketErrorCode switch
+        {
+            SocketError.HostNotFound => "host_not_found",
+            SocketError.TryAgain => "try_again",
+            SocketError.AddressFamilyNotSupported => "address_family_not_supported",
+            SocketError.NoRecovery => "no_recovery",
+
+            _ => exception.GetType().Name
```
JamesNK, line 44: Full name is used in other tags like this.

karelz (approved):

src/libraries/System.Net.NameResolution/src/System/Net/NameResolutionMetrics.cs
antonfirsov: Good catch! This is an overlook, [Liudmila's spec](https://github.com/lmolkova/semantic-conventions/blob/dotnet8-metrics/docs/dotnet/dotnet-http-metrics.md#metric-httpclientrequestduration) on `http.client.request.duration` requires us to use the full name:

> When error.type attribute is reported, it contains one of [HTTP Request errors](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpRequestError.cs), a full exception type, or a string representation of received status code.

We should update this for consistence (with the spec and aspnet) as part of the sync-up effort we are doing now, I will add the change of the HTTP tag value to #93255.

MihaZupan: Was the intention behind "full exception type" actually the full .NET type name, or more so along the lines of "use HttpRequestException instead of just 'Http' or 'HttpException'"?

IMO `Name` feels appropriate here, I don't think `System.Threading.Tasks.TaskCanceledException` buys us anything over `TaskCanceledException`.

src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
stephentoub: And when that mapping is augmented in the future, we're going to remember to update this random method as well?

src/libraries/System.Net.NameResolution/src/System/Net/NameResolutionMetrics.cs
antonfirsov: I tend to agree. Will not touch the existing behavior in the PRs, opened https://github.com/dotnet/runtime/issues/93302.

src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
noahfalk: > it seems like our general goal here is to take any SocketError value or Exception name, and snake case it

I'd frame the goal as disambiguating the most common sources of error that users would appreciate not having grouped together when analyzing failures. I don't think it is a goal that every possible SocketError has been disambiguated, that new future SocketError values are automatically disambiguated, or that exception names are transformed into snake case. The hard-coded list Anton has here seems fine to me. If we ever changed it we'd probably want that to be an intentional choice so we can warn users that their queries might need to be updated.

noahfalk (approved):
LGTM. I commented on the other issue that I do think FullName would be preferable for exception types for consistency across other .NET metrics.

antonfirsov:
CI failures are unrelated (eg. #93413).

lewing:
I opened https://github.com/dotnet/runtime/pull/93418 to revert this because it broke the build

karelz:
Looks like it didn't break the build after all - see explanation in https://github.com/dotnet/runtime/pull/93418#issuecomment-1760530850

Issue #88901, opened 2023-07-14
Consider running debug builds of analyzers/source generators in some build leg


vitek-karas:
Analyzers are compiled against a specific version of the C# compiler. As such they can only recognize the shapes of the code that version of the compiler supports. With a new version of the compiler new shapes become possible, sometimes these are subtle changes, sometimes they mean adding new node types to representation or new enum values to properties. Since we can't update analyzers in sync with the compiler, the analyzer has to be written such that it can survive most of these changes without too much harm (typically it would simply ignore the new shape).

But that means that we don't get to be notified if such a new shape start occurring and can't fix the analyzer unless we have some other reason to look into that case. There will be cases where the analyzer doesn't work the way it should simply because it wasn't updated to match the latest compiler behavior.

We considered adding a warning into the analyzer to ask the user to file an issue if this happens, but that comes with rather severe downsides:
* If this happens, it's not an actionable warning (other than filing the bug), the only solution is to suppress the warning code
* Lot of users treat warnings as errors, so this becomes a blocking problem for them if it happens
* We're putting burden on our users due to our inability to keep things in sync

(See the discussion in https://github.com/dotnet/runtime/pull/88836 for some more detail)

For the illink analyzer we decided to change these cases into assertions. But that means we would like to run the debug build of the analyzer on as much code as possible so that we catch these things internally when they happen.

I think it would make sense to do this for all the analyzers where it's possible - there are assertions in their code bases as well which are only exercised on targeted unit tests, but basically never on real-world code.

The same thing applies to source generators as well obviously.

/cc @sbomer @agocke @jkoritzinsky

jkoritzinsky:
We moved all of our assertions into throws (in both debug and release) quite a while ago, so this doesn't apply for the interop generators.

PR #93259, opened 2023-10-09
Reference live ILLink analyzer

sbomer:
In https://github.com/dotnet/runtime/pull/91233 I inadvertently turned off the trim analyzer for libraries, because the analyzer  bits don't flow to the consuming project with the ILLink.Tasks ProjectReference. This adds a separate ProjectReference to use the live analyzer. This addresses https://github.com/dotnet/runtime/issues/88901 for the ILLink analyzers.

Using the live analyzer uncovered some bugs that are fixed in this change:
- Invalid assert for field initializers with `throw` statements
- Invalid cast to `IMethodSymbol` for the ContainingSymbol of a local. Locals can occur in field initializers, when created from out params.
- Wrong warning code produced for assignment to annotated property in an initializer. This was being treated as a call to the setter instead of an assignment to the underlying field:
   `Type AnnotatedPropertyWithSetter { get; set; } = GetUnknown ();`
   (IIRC I hit this one when investigating some related behavior, not via a new warning produced when analyzing runtime)
- Invalid assert for delegate creation where the delegate is created from anything other than a method (for example, a field that has an Action or Func type).


eng/liveILLink.targets
```diff
@@ -39,6 +39,12 @@
         <SetTargetFramework Condition="'$(MSBuildRuntimeType)' == 'Core'">TargetFramework=$(NetCoreAppToolCurrent)</SetTargetFramework>
         <SetTargetFramework Condition="'$(MSBuildRuntimeType)' != 'Core'">TargetFramework=$(NetFrameworkToolCurrent)</SetTargetFramework>
     </ProjectReference>
+
+    <!-- Need to reference the analyzer project separately, because there's no easy way to get it as a transitive reference of ILLink.Tasks.csproj -->
+    <ProjectReference Include="$(_ILLinkTasksSourceDir)..\ILLink.RoslynAnalyzer\ILLink.RoslynAnalyzer.csproj"
+                      ReferenceOutputAssembly="false"
+                      OutputItemType="Analyzer"
```
ViktorHofer, line 46: Please also set PrivateAssets and Private same as the above ProjectReference sets.

src/tools/illink/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -190,6 +195,11 @@ TValue ProcessSingleTargetAssignment (IOperation targetOperation, ISimpleAssignm
 						// For now, just don't warn. https://github.com/dotnet/linker/issues/2731
```
vitek-karas, line 195:
```suggestion
						// For now, just don't warn. https://github.com/dotnet/runtime/issues/93277
```

I moved the issue to runtime - and we should look into fixing this.

vitek-karas (approved):

eng/liveILLink.targets
ViktorHofer: Those would probably also benefit from `Private="false"`. That hinders msbuild from copying the source generator into the lib/app output which is unnecessary work. Not 100% sure about PrivateAssets though. Maybe that metadata is by default true when `ReferenceOutputAssembly=false`.

vitek-karas (approved):

eng/liveILLink.targets
ViktorHofer: OK, I just verified that `PrivateAssets` isn't required here. The ILLink.RoslynAnalyzer.csproj dependency doesn't show up as a package dependency with your changes, i.e. for System.Text.Json:

image

ViktorHofer (approved):
Also verified that this change doesn't require updating our solution files again in src/libraries.

PR #93264, opened 2023-10-09
Skip Jitted Methods Counting Test when DOTNET_EnableHWIntrinsic=0

### LINKED ISSUE #93261: Jitted Methods Counting Test Fails When Hardware Intrinsics are Disabled
Tracking issue following up #93163.

The above-mentioned issue exposed a bug with the _Jitted Methods Counting Test_, regarding Hardware Intrinsics. When `DOTNET_EnableHWIntrinsic=0`, there is a much higher JIT activity, which in turn returns a much higher amount of jitted methods than the test was designed for. As per @jkotas' feedback, `DOTNET_EnableHWIntrinsic=0` has a similar effect to `DOTNET_ReadyToRun=0`, so this test ought not be run in those cases.

For general information and context, here's how to build and run the test locally.

```bash
# Build the runtime and libraries.

./build.sh --subset clr+libs --runtimeConfiguration Checked --librariesConfiguration Release

# Build the tests and core root.

src/tests/build.sh -x64 -checked -generatelayoutonly
src/tests/build.sh -x64 -checked -test:readytorun/JittedMethodsCountingTest/JittedMethodsCountingTest.csproj

# Setup and run the tests.

cd artifacts/tests/coreclr/linux.x64.Checked
export CORE_ROOT=$(pwd)/Tests/Core_Root
./readytorun/JittedMethodsCountingTest/JittedMethodsCountingTest.sh
```

ivdiazsa:
Resolves Issue #93261.

Recently, a new regression uncovered an untreated case in the Jitted Methods Counting Test. Due to its purpose and how it works, this test is incompatible when Hardware Intrinsics are disabled (i.e., `DOTNET_EnableHWIntrinsic=0`). This PR tells it to skip when such environment is detected.

### DIFF
--- a/src/tests/readytorun/JittedMethodsCountingTest/JittedMethodsCountingTest.cs
+++ b/src/tests/readytorun/JittedMethodsCountingTest/JittedMethodsCountingTest.cs
@@ -23,18 +23,8 @@ public class JittedMethodsCountingTest
             return 100;
         }
 
-        // This test is currently not compatible with the R2R-CG2 pipelines on Arm64.
-        // if (IsRunCrossgen2Set() && IsRunningOnARM64())
-        // {
-        //     Console.WriteLine("\nThis test is currently unsupported on ARM64 when"
-        //                       + " RunCrossGen2 is enabled. Skipping...\n");
-        //     return 100;
-        // }
-
         Console.WriteLine("\nHello World from Jitted Methods Counting Test!");
-
-        // Get the total amount of Jitted Methods.
-        long jits = JitInfo.GetCompiledMethodCount(false);
+        long jits = JitInfo.GetCompiledMethodCount(currentThread: false);
 
         Console.WriteLine("Number of Jitted Methods in App: {0} - Max Threshold: {1}\n",
                           jits,
--- a/src/tests/readytorun/JittedMethodsCountingTest/JittedMethodsCountingTest.cs
+++ b/src/tests/readytorun/JittedMethodsCountingTest/JittedMethodsCountingTest.cs
@@ -14,14 +14,16 @@ public class JittedMethodsCountingTest
     [Fact]
     public static int TestEntryPoint()
     {
-        // If DOTNET_ReadyToRun is disabled, then this test ought to be skipped.
-        if (!IsReadyToRunEnvSet())
+        // If either DOTNET_ReadyToRun or DOTNET_EnableHWIntrinsics are disabled
+        // (i.e. set to "0"), then this test ought to be skipped.
+        if (!IsReadyToRunEnabled() || !IsHardwareIntrinsicsEnabled())
         {
-            Console.WriteLine("\nThis test is only supported in ReadyToRun scenarios."
-                              + " Skipping...\n");
+            Console.WriteLine("\nThis test is only supported in ReadyToRun scenarios"
+                              + " with Hardware Intrinsics enabled. Skipping...\n");
             return 100;
         }
 
+        // This test is currently not compatible with the R2R-CG2 pipelines on Arm64.
         if (IsRunCrossgen2Set() && IsRunningOnARM64())
         {
             Console.WriteLine("\nThis test is currently unsupported on ARM64 when"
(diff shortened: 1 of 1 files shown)


trylek (approved):
LGTM, thanks!

src/tests/readytorun/JittedMethodsCountingTest/JittedMethodsCountingTest.cs
```diff
@@ -14,14 +14,16 @@ public class JittedMethodsCountingTest
...
+        if (!IsReadyToRunEnabled() || !IsHardwareIntrinsicsEnabled())
         {
-            Console.WriteLine("\nThis test is only supported in ReadyToRun scenarios."
-                              + " Skipping...\n");
+            Console.WriteLine("\nThis test is only supported in ReadyToRun scenarios"
+                              + " with Hardware Intrinsics enabled. Skipping...\n");
             return 100;
         }
 
+        // This test is currently not compatible with the R2R-CG2 pipelines on Arm64.
```
jkotas, line 26: This comment is not very useful. It does not explain why it is not compatible. Good comments need to explain "why", not just "what".

src/tests/readytorun/JittedMethodsCountingTest/JittedMethodsCountingTest.cs
```diff
@@ -31,13 +33,17 @@ public static int TestEntryPoint()
 
         Console.WriteLine("\nHello World from Jitted Methods Counting Test!");
 
+        // Get the total amount of Jitted Methods.
```
jkotas, line 36: This comment is not useful. It is the exact equivalent of the code that follows.

jkotas, line 36:
```suggestion
```
This comment is not useful. It is the exact equivalent of the code that follows.

src/tests/readytorun/JittedMethodsCountingTest/JittedMethodsCountingTest.cs
```diff
@@ -31,13 +33,17 @@ public static int TestEntryPoint()
...
         long jits = JitInfo.GetCompiledMethodCount(false);
```
jkotas, line 37:
```suggestion
        long jits = JitInfo.GetCompiledMethodCount(currentThread: false);
```
This is an edit that would make this code easier to follow. It is not obvious by just looking at the code what the `false` argument means.

src/tests/readytorun/JittedMethodsCountingTest/JittedMethodsCountingTest.cs
ivdiazsa: It seems that the problem with this configuration is no longer there. Will remove the check altogether.

src/tests/readytorun/JittedMethodsCountingTest/JittedMethodsCountingTest.cs
```diff
@@ -14,30 +14,26 @@ public class JittedMethodsCountingTest
...
+        // If either DOTNET_ReadyToRun or DOTNET_EnableHWIntrinsics are disabled
+        // (i.e. set to "0"), then this test ought to be skipped.
+        if (!IsReadyToRunEnabled() || !IsHardwareIntrinsicsEnabled())
         {
-            Console.WriteLine("\nThis test is only supported in ReadyToRun scenarios."
-                              + " Skipping...\n");
-            return 100;
-        }
-
-        if (IsRunCrossgen2Set() && IsRunningOnARM64())
```
jkotas, line 25: Also delete `IsRunCrossgen2Set` and `IsRunningOnARM64` methods

src/tests/readytorun/JittedMethodsCountingTest/JittedMethodsCountingTest.cs
```diff
@@ -14,30 +14,26 @@ public class JittedMethodsCountingTest
...
-        {
-            Console.WriteLine("\nThis test is currently unsupported on ARM64 when"
-                              + " RunCrossGen2 is enabled. Skipping...\n");
+            Console.WriteLine("\nThis test is only supported in ReadyToRun scenarios"
+                              + " with Hardware Intrinsics enabled. Skipping...\n");
             return 100;
         }
 
         Console.WriteLine("\nHello World from Jitted Methods Counting Test!");
+        long jits = JitInfo.GetCompiledMethodCount(currentThread: false);
```
jkotas, line 28:
```suggestion

        long jits = JitInfo.GetCompiledMethodCount(currentThread: false);
```
Formating nit

ivdiazsa:
Failures are unrelated to this fix. Merging the PR now.

PR #93285, opened 2023-10-10
[RISC-V] Fix dropping NaN sign and payload when importing float constants

tomeksowi:
RISC-V float-to-double conversion canonicalizes the constant NaN value. Use soft conversion to double.

Since gtNewDconNode is used in quite a few places, the offending conversion is silent (float to double), and most programmers are unaware of NaN propagation issues on uncommon platforms like RISC-V, introduce factory functions dedicated to float and double that do the necessary conversions and make unintended upcasts more difficult.

Also fix printing NaN constants in dumps.

This fixes System.Reflection.Metadata.Ecma335.Tests.InstructionEncoderTests.LoadConstantR4.

Part of #84834

### DIFF
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -2643,7 +2643,7 @@ GenTree* Compiler::optVNConstantPropOnTree(BasicBlock* block, GenTree* tree)
 
                     case TYP_DOUBLE:
                         // Same sized reinterpretation of bits to double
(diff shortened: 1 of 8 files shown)


jakobbotsch (approved):
LGTM. I suspect this isn't the only change that will need to be made if we really want to have consistent cross-platform NaN behavior for all cases in the JIT for RISC-V. I sort of worry that it would require a (larger) software float library, which I'm not sure how happy we would be about accepting -- but let's cross that bridge if we get to it.

PR #93289, opened 2023-10-10
Use collection expression in foreach in CreateManifestAndDescriptors

stephentoub:
Avoids an unnecessary string[] allocation. It'll instead result in an inline array of the elements.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs
@@ -3163,9 +3163,9 @@ namespace System.Diagnostics.Tracing
 
                 // Collect task, opcode, keyword and channel information
 #if FEATURE_MANAGED_ETW_CHANNELS && FEATURE_ADVANCED_MANAGED_ETW_CHANNELS
-                foreach (var providerEnumKind in new string[] { "Keywords", "Tasks", "Opcodes", "Channels" })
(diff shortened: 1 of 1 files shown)


src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -3163,9 +3163,9 @@ private static bool AttributeTypeNamesMatch(Type attributeType, Type reflectedAt
 
                 // Collect task, opcode, keyword and channel information
 #if FEATURE_MANAGED_ETW_CHANNELS && FEATURE_ADVANCED_MANAGED_ETW_CHANNELS
-                foreach (var providerEnumKind in new string[] { "Keywords", "Tasks", "Opcodes", "Channels" })
+                foreach (var providerEnumKind in (ReadOnlySpan<string>)["Keywords", "Tasks", "Opcodes", "Channels"])
```
MihaZupan, line 3166: > `(ReadOnlySpan<string>)`

Is the cast in places like this something that we'll be able to omit in future C# versions?

MihaZupan (approved):

src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs
stephentoub: Likely. Collection expressions _currently_ don't have a natural type, so they currently can't be used in places where no target type is specified.

tarekgh (approved):
CC @davmason

Issue #92784, opened 2023-09-28
Atomics are missing from `OperExceptions`


SingleAccretion:
All the store-like atomics are missing from `GenTree::OperExceptions`, even as they can throw NREs. I noticed this when implementing downstream support for them (and, in fact, this bug blocks that support).

The plan for fixing is as follows:
1) Make all atomics (including `GT_CMPXCHG`) `GenTreeIndir`-derived.
2) This will automatically "do the right thing" in a lot of places; however, all uses of `OperIsIndir` will need to be audited.
3) Add the missing opcodes to `GenTree::OperExceptions`.

EgorBo:
> Make all atomics (including GT_CMPXCHG) GenTreeIndir-derived.

Probably will need some special caseing to skip address mode recognition

PR #93295, opened 2023-10-10
Make atomics `GenTreeIndir`-derived

SingleAccretion:
This is the first (zero-diff, hence quirks) part of fixing #92784.

We make all atomics `OperIsIndir` - they are a lot like stores and should share the common machinery like `GTF_IND_NONFAULTING` handling seamlessly this way.


SingleAccretion:
@dotnet/jit-contrib

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=434375&view=ms.vss-build-web.run-extensions-tab) - none in code, as expected.

There is an anomalous TP regression on Windows x64; it is not yet clear why that is - it does not reproduce locally. I will need to pull the latest `main` and recheck, however, it may also be due to the difference in MSVC versions. In either case, I believe it to be due to some different inlining decisions.

jakobbotsch (approved):

PR #93296, opened 2023-10-10
Improve a vector implementation to support alignment and non-temporal tores

tannergooding:


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1220,102 +1222,876 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+
+                return;
+            }
+
+            // This is the software fallback when no acceleration is available
+            // It requires no branches to hit
+
+            SoftwareFallback(ref xRef, ref dRef, remainder);
+
+            static void SoftwareFallback(ref float xRef, ref float dRef, nuint length)
```
stephentoub, line 1296: Why separate SoftwareFallback out into its own method (which presumably isn't going to get inlined automatically)?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1220,102 +1222,876 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                Vector128<float> beg = TUnaryOperator.Invoke(Vector128.LoadUnsafe(ref xRef));
+                Vector128<float> end = TUnaryOperator.Invoke(Vector128.LoadUnsafe(ref xRef, remainder - (uint)(Vector128<float>.Count)));
+
+                if (remainder > (uint)(Vector128<float>.Count * 8))
+                {
+                    // Pinning is cheap and will be short lived for small inputs and unlikely to be impactful
+                    // for large inputs (> 85KB) which are on the LOH and unlikely to be compacted.
+
+                    fixed (float* px = &xRef)
+                    fixed (float* pd = &dRef)
```
stephentoub, line 1319: What's the thinking for using fixed and pointers instead of refs? The same situation that results in alignment changing is also the one that results in pinning being expensive.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: If the dispatch method is "too large" or has "too many locals", the JIT will generate worse code.

In practice, only one of these branches will be "live" and we'll either have just the `SoftwareFallback` or just a given `Vectorized###` and `Vectorized###Small`.

We could likely mark this one as `AggressiveInlining`, but it's probably more profitable to mark `InvokeSpanIntoSpan` as such so the argument validation is inlined into the root entry point and then the dispatch to the core handler (`SoftwareFallback`, `Vectorized128`, etc) is the call instead

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1220,102 +1222,876 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                                vector2.StoreAlignedNonTemporal(dPtr + (uint)(Vector128<float>.Count * 1));
+                                vector3.StoreAlignedNonTemporal(dPtr + (uint)(Vector128<float>.Count * 2));
+                                vector4.StoreAlignedNonTemporal(dPtr + (uint)(Vector128<float>.Count * 3));
+
+                                // We load, process, and store the next four vectors
+
+                                vector1 = TUnaryOperator.Invoke(Vector128.Load(xPtr + (uint)(Vector128<float>.Count * 4)));
+                                vector2 = TUnaryOperator.Invoke(Vector128.Load(xPtr + (uint)(Vector128<float>.Count * 5)));
+                                vector3 = TUnaryOperator.Invoke(Vector128.Load(xPtr + (uint)(Vector128<float>.Count * 6)));
+                                vector4 = TUnaryOperator.Invoke(Vector128.Load(xPtr + (uint)(Vector128<float>.Count * 7)));
```
stephentoub, line 1377: Is this manual unrolling going to increase code size problematically?  We have some operators whose Invoke methods are marked as AggressiveInlining. (Related question is whether more or less of them should be AggressiveInlining.)

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: > If the dispatch method is "too large" or has "too many locals", the JIT will generate worse code.

Shouldn't that only matter after it's already trimmed away all but the relevant scalar/128/256/512 path?

>  but it's probably more profitable to mark InvokeSpanIntoSpan as such so the argument validation is inlined into the root entry point

That would likely put all of that at the user's call site, since most of the calls in our methods to InvokeSpanIntoSpan are just the call.  That doesn't seem particularly desirable.

EgorBo:
If a function is idempotent (which most of them are) it's enough a single load to align data (like I did in https://github.com/dotnet/runtime/pull/93214/files) without any unrolled loops, etc - presumably, you can expose a `bool IsIdempotent {get;}` property in these abstraction and only align those for simplicity

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
EgorBo: Without pinning you then can not use APIs like `StoreAlignedNonTemporal`

tannergooding: Using fixed is required to support non-temporal stores, which significantly improves perf for large inputs.

Pinning itself is cheap, it is just spilling an address to the stack. The potential impact of pinning is that it can negatively impact the GC's ability to perform compaction of memory.

For "small" data, we avoid pinning entirely. Where small is currently `>= 128` bytes for V128, `>= 256` bytes for V256, `>= 512` bytes for V512.

For medium sized data (`<= 85KB`, so on the regular GC heap), the pin is going to be short and the biggest risk is if we repeatedly pin the data across many calls.

For large sized data (`> 85KB`, on the LOH), allocations are already unlikely to be compacted and so the pinning shouldn't be an issue.

stephentoub: Ah, ok. As part of making this non-draft, it'd be nice if the comment explained that.

stephentoub: > The potential impact of pinning is that it can negatively impact the GC's ability to perform compaction of memory.

Right, that's why I was saying the situation that would cause the negative impact with pinning is the same situation that would cause the negative impact with assuming a particular alignment with refs (since you could all of a sudden become unaligned).

Egor's explanation about StoreAlignedNonTemporal explains the rationale.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: > Is this manual unrolling going to increase code size problematically

It shouldn't. It allows us to take advantage of pipelining and avoid the overhead of the loop (which can be significant for older platforms). The general standard for best perf (and for various memcpy or other basic algorithms) is to unroll like this.

> We have some operators whose Invoke methods are marked as AggressiveInlining

We probably want to consider if they should be marked that way or not. For basic operations, like `Abs`, we're already doing a single instruction each. For more complex functions like `Cosh`, we probably shouldn't be. Cases like `Min` or `Max` are probably right on the line and could go either way based on the hardware.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: > Shouldn't that only matter after it's already trimmed away all but the relevant scalar/128/256/512 path?

Depends on a few factors and the JIT is getting better about it, but it still has issues at times.

> That would likely put all of that at the user's call site, since most of the calls in our methods to InvokeSpanIntoSpan are just the call. That doesn't seem particularly desirable.

It depends on where the inlining happens. I doubt we'd end up with the `InvokeSpanIntoSpan(...)` inlined to the user call site. Rather instead we'd probably have it inlined into `Abs(...)` and the user call go to that instead.

But, it probably doesn't matter too much and we can mark `SoftwareFallback` and `Vector###Small` as `AggressiveInlining` if you prefer that. It's just beneficial to not have them in the main body so the main dispatch algorithm can participate in normal inlining heuristics around constant inputs and the like.

stephentoub: > But, it probably doesn't matter too much and we can mark SoftwareFallback and Vector###Small as AggressiveInlining if you prefer that.

I don't currently see a downside to doing so (since that's the only call site, and if that code path is compiled in, you're going to call the method), and I can imagine it could save a couple nanoseconds, which could be measurable for small inputs.  So my default would be to mark it as such, unless perf tests show that it actually doesn't matter.

tannergooding: Marked with AggressiveInlining.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
gfoidl: > Where small is currently `>= 128` bytes for V128, `>= 256` bytes for V256, `>= 512` bytes for V512.

@tannergooding you have typos in here, should be `<=`.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1220,102 +1222,892 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
 
             ValidateInputOutputSpanNonOverlapping(x, destination);
 
+            // Since every branch has a cost and since that cost is
```
stephentoub, line 1225: Is the intent to do the other methods in this PR, or are you planning to do them all in a subsequent PR?

tannergooding: Subsequent PR. Wanted to get the initial code pattern up before doing “everything”

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1220,102 +1222,892 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                //
+                // Unless the original length was an exact multiple of Count, then we'll
+                // end up reprocessing a couple elements in case 1 for end. We'll also
+                // potentially reprocess a few elements in case 0 for beg, to handle any
+                // data before the first aligned address.
+
+                nuint endIndex = remainder;
+                remainder = (remainder + (uint)(Vector128<float>.Count - 1)) & (nuint)(-Vector128<float>.Count);
+
+                switch (remainder / (uint)(Vector128<float>.Count))
```
stephentoub, line 1449: Is https://github.com/dotnet/runtime/pull/93346/files#r1355278777 relevant here at all?

tannergooding: I don't think so. Native memcpy algorithms have similar jump tables, explicitly, because the number of remaining is unpredictable.

That is, the entire point is the number of remaining blocks can't be predicted. But we need to have handling for each. Looping is undesirable because the branch predictor gets it wrong (both by default because loops are predicted to be taken at least 6-8 times, but also over time because the number of iterations is not consistent).

A fallthrough jump table, such as this, allows us to have consistent jump target addresses, where we use the index as an offset to jump. Even though the branch itself is unpredictable it is unconditional and the destinations are often already decoded or partially decoded due to being immediately after the jump which allows for reduce cost. This makes it better than many conditional branches which can likewise not be predicted.

Using such jump tables is actually explicitly covered in the Intel/AMD optimization manuals, typically as part of the vectorization logic. The current May 2023 copy of `Intel® 64 and IA-32 Architectures Optimization Reference Manual` covers this:
> After an initial unaligned
store, and adjusting the destination pointer to be 32-byte aligned, the residual phase follows the same
consideration as described in Section 15.16.3.1, which may employ a large jump table to handle each
residual value scenario with minimal branching, depending on the amount of unrolled 32B-aligned
stores. The main loop is a simple YMM register to 32-byte-aligned store operation, which can deliver
close to 30 bytes per cycle for lengths more than a thousand byte. The limiting factor here is due to each
256-bit VMOVDQA store consists of a store_address and a store_data micro-op flow. Only port 4 is available
to dispatch the store_data micro-op each cycle.

Among other such references
> On Ivy Bridge microarchitecture, a REP MOVSB implementation of memcpy can achieve throughput at
slightly better than the 128-bit SIMD implementation when copying thousands of bytes. However, if the
size of copy operation is less than a few hundred bytes, the REP MOVSB approach is less efficient than the
explicit residual copy technique described in phase 2 of Section 15.16.3.1. This is because handling 1-
127 residual copy length (via jump table or switch/case, and is done before the main loop) plus one or
two 8x16B iterations incurs less branching overhead than the hardware provided micro-op flows. For the
grueling implementation details of 128-bit SIMD implementation of memcpy(), one can look up from the
archived sources of open source library such as GLibC.

or the discussions on how the instruction cache works
> For code entering the Decoded ICache, each unconditional branch is the last micro-op occupying a
Decoded ICache Way. Therefore, only three unconditional branches per a 32 byte aligned chunk can
enter the Decoded ICache.
Unconditional branches are frequent in jump tables and switch declarations. Below are examples for
these constructs, and methods for writing them so that they fit in the Decoded ICache.
Compilers create jump tables for C++ virtual class methods or DLL dispatch tables. Each unconditional
branch consumes five bytes; therefore up to seven of them can be associated with a 32-byte chunk. Thus
jump tables may not fit in the Decoded ICache if the unconditional branches are too dense in each
32Byte-aligned chunk. This can cause performance degradation for code executing before and after the
branch table.
The solution is to add multi-byte NOP instructions among the branches in the branch table. This may
increases code size and should be used cautiously. However, these NOPs are not executed and therefore
have no penalty in later pipe stages.
Switch-Case constructs represents a similar situation. Each evaluation of a case condition results in an
unconditional branch. The same solution of using multi-byte NOP can apply for every three consecutive
unconditional branches that fits inside an aligned 32-byte chunk.

tannergooding:
Here are the before/after results for `Abs`, where `Abs1` is the previous implementation and `Abs2` is the implementation from this PR. We are faster across the board and more than 2x faster for large inputs, with the gains really showing past 256 elements (1024 bytes).

| Method | TensorLength | Mean          | Error       | StdDev      |
|------- |------------- |--------------:|------------:|------------:|
| Abs1   | 1            |      2.609 ns |   0.0611 ns |   0.0572 ns |
| Abs2   | 1            |      2.562 ns |   0.0518 ns |   0.0485 ns |
| Abs1   | 2            |      3.142 ns |   0.0487 ns |   0.0456 ns |
| Abs2   | 2            |      2.563 ns |   0.0121 ns |   0.0101 ns |
| Abs1   | 3            |      3.309 ns |   0.0294 ns |   0.0229 ns |
| Abs2   | 3            |      2.742 ns |   0.0171 ns |   0.0133 ns |
| Abs1   | 4            |      3.090 ns |   0.0496 ns |   0.0464 ns |
| Abs2   | 4            |      2.572 ns |   0.0187 ns |   0.0166 ns |
| Abs1   | 5            |      5.198 ns |   0.0190 ns |   0.0148 ns |
| Abs2   | 5            |      3.133 ns |   0.0788 ns |   0.0774 ns |
| Abs1   | 6            |      5.275 ns |   0.0790 ns |   0.0739 ns |
| Abs2   | 6            |      3.098 ns |   0.0782 ns |   0.0732 ns |
| Abs1   | 7            |      5.322 ns |   0.0916 ns |   0.0857 ns |
| Abs2   | 7            |      3.164 ns |   0.0544 ns |   0.0509 ns |
| Abs1   | 8            |      2.936 ns |   0.0553 ns |   0.0517 ns |
| Abs2   | 8            |      2.573 ns |   0.0442 ns |   0.0414 ns |
| Abs1   | 9            |      5.454 ns |   0.0910 ns |   0.0852 ns |
| Abs2   | 9            |      3.165 ns |   0.0536 ns |   0.0501 ns |
| Abs1   | 10           |      5.431 ns |   0.0675 ns |   0.0631 ns |
| Abs2   | 10           |      3.230 ns |   0.0308 ns |   0.0288 ns |
| Abs1   | 11           |      5.369 ns |   0.0447 ns |   0.0418 ns |
| Abs2   | 11           |      3.199 ns |   0.0443 ns |   0.0393 ns |
| Abs1   | 12           |      5.571 ns |   0.0593 ns |   0.0555 ns |
| Abs2   | 12           |      3.211 ns |   0.0320 ns |   0.0299 ns |
| Abs1   | 13           |      5.585 ns |   0.0491 ns |   0.0459 ns |
| Abs2   | 13           |      3.262 ns |   0.0295 ns |   0.0276 ns |
| Abs1   | 14           |      5.766 ns |   0.0550 ns |   0.0514 ns |
| Abs2   | 14           |      3.264 ns |   0.0292 ns |   0.0273 ns |
| Abs1   | 15           |      5.865 ns |   0.0454 ns |   0.0354 ns |
| Abs2   | 15           |      3.254 ns |   0.0451 ns |   0.0377 ns |
| Abs1   | 16           |      2.790 ns |   0.0207 ns |   0.0193 ns |
| Abs2   | 16           |      4.984 ns |   0.0479 ns |   0.0448 ns |
| Abs1   | 32           |      6.515 ns |   0.0886 ns |   0.0785 ns |
| Abs2   | 32           |      6.012 ns |   0.1203 ns |   0.1067 ns |
| Abs1   | 64           |     17.126 ns |   0.1647 ns |   0.1540 ns |
| Abs2   | 64           |     15.343 ns |   0.3172 ns |   0.3115 ns |
| Abs1   | 128          |     38.650 ns |   0.7464 ns |   0.6982 ns |
| Abs2   | 128          |     37.150 ns |   0.4134 ns |   0.3665 ns |
| Abs1   | 256          |     82.113 ns |   0.5631 ns |   0.4702 ns |
| Abs2   | 256          |     51.379 ns |   0.2897 ns |   0.2710 ns |
| Abs1   | 512          |    169.559 ns |   0.7197 ns |   0.6732 ns |
| Abs2   | 512          |     78.668 ns |   0.4391 ns |   0.4107 ns |
| Abs1   | 1024         |    336.791 ns |   1.9587 ns |   1.8321 ns |
| Abs2   | 1024         |    133.500 ns |   0.7219 ns |   0.6400 ns |
| Abs1   | 2048         |    698.482 ns |   1.3343 ns |   1.1142 ns |
| Abs2   | 2048         |    244.729 ns |   1.6655 ns |   1.5579 ns |
| Abs1   | 4096         |  1,411.587 ns |   5.8918 ns |   5.2229 ns |
| Abs2   | 4096         |    471.249 ns |   1.2273 ns |   1.0880 ns |
| Abs1   | 65536        | 22,871.118 ns |  97.4602 ns |  91.1643 ns |
| Abs2   | 65536        |  7,306.615 ns |  32.3811 ns |  30.2893 ns |
| Abs1   | 131072       | 45,773.050 ns | 244.2154 ns | 228.4393 ns |
| Abs2   | 131072       | 18,461.823 ns | 105.6972 ns |  93.6978 ns |

tannergooding:
Similar results for `Exp`

| Method | TensorLength | Mean           | Error         | StdDev        |
|------- |------------- |---------------:|--------------:|--------------:|
| Exp1   | 1            |       5.224 ns |     0.1122 ns |     0.1050 ns |
| Exp2   | 1            |       4.863 ns |     0.0823 ns |     0.0770 ns |
| Exp1   | 2            |       7.359 ns |     0.1578 ns |     0.1689 ns |
| Exp2   | 2            |       6.872 ns |     0.0975 ns |     0.0912 ns |
| Exp1   | 3            |       9.315 ns |     0.1091 ns |     0.0852 ns |
| Exp2   | 3            |       9.018 ns |     0.1643 ns |     0.1537 ns |
| Exp1   | 4            |       4.820 ns |     0.1094 ns |     0.1075 ns |
| Exp2   | 4            |       4.863 ns |     0.1028 ns |     0.0962 ns |
| Exp1   | 5            |       9.103 ns |     0.0672 ns |     0.0629 ns |
| Exp2   | 5            |       6.563 ns |     0.1169 ns |     0.1036 ns |
| Exp1   | 6            |       9.280 ns |     0.2022 ns |     0.2076 ns |
| Exp2   | 6            |       6.685 ns |     0.1498 ns |     0.1538 ns |
| Exp1   | 7            |       9.111 ns |     0.0848 ns |     0.0793 ns |
| Exp2   | 7            |       6.660 ns |     0.1413 ns |     0.1321 ns |
| Exp1   | 8            |       5.006 ns |     0.0963 ns |     0.0901 ns |
| Exp2   | 8            |       4.848 ns |     0.1006 ns |     0.0941 ns |
| Exp1   | 9            |      10.272 ns |     0.1189 ns |     0.1112 ns |
| Exp2   | 9            |       7.593 ns |     0.0876 ns |     0.0820 ns |
| Exp1   | 10           |      10.188 ns |     0.0713 ns |     0.0596 ns |
| Exp2   | 10           |       7.581 ns |     0.0935 ns |     0.0875 ns |
| Exp1   | 11           |      10.274 ns |     0.1070 ns |     0.0949 ns |
| Exp2   | 11           |       7.631 ns |     0.1412 ns |     0.1320 ns |
| Exp1   | 12           |      10.246 ns |     0.0755 ns |     0.0670 ns |
| Exp2   | 12           |       7.555 ns |     0.0742 ns |     0.0658 ns |
| Exp1   | 13           |      10.270 ns |     0.1266 ns |     0.1184 ns |
| Exp2   | 13           |       7.651 ns |     0.0916 ns |     0.0857 ns |
| Exp1   | 14           |      10.252 ns |     0.1509 ns |     0.1411 ns |
| Exp2   | 14           |       7.585 ns |     0.1017 ns |     0.0951 ns |
| Exp1   | 15           |      10.428 ns |     0.0935 ns |     0.0875 ns |
| Exp2   | 15           |       7.795 ns |     0.0801 ns |     0.0749 ns |
| Exp1   | 16           |       7.271 ns |     0.0845 ns |     0.0749 ns |
| Exp2   | 16           |      19.323 ns |     0.3291 ns |     0.3078 ns |
| Exp1   | 32           |      18.716 ns |     0.2705 ns |     0.2530 ns |
| Exp2   | 32           |      25.631 ns |     0.3860 ns |     0.3610 ns |
| Exp1   | 64           |      48.860 ns |     0.4013 ns |     0.3753 ns |
| Exp2   | 64           |      50.285 ns |     0.9766 ns |     0.9135 ns |
| Exp1   | 128          |     109.969 ns |     1.6557 ns |     1.5487 ns |
| Exp2   | 128          |     110.952 ns |     1.4802 ns |     1.3846 ns |
| Exp1   | 256          |     230.547 ns |     2.4931 ns |     2.0819 ns |
| Exp2   | 256          |     193.106 ns |     0.8509 ns |     0.7105 ns |
| Exp1   | 512          |     477.460 ns |     5.6110 ns |     5.2486 ns |
| Exp2   | 512          |     317.518 ns |     5.5631 ns |     5.2037 ns |
| Exp1   | 1024         |     977.141 ns |    13.0740 ns |    12.2295 ns |
| Exp2   | 1024         |     573.851 ns |     7.7362 ns |     7.2364 ns |
| Exp1   | 2048         |   1,931.447 ns |    12.4202 ns |    10.3714 ns |
| Exp2   | 2048         |   1,017.003 ns |    19.3625 ns |    20.7177 ns |
| Exp1   | 4096         |   3,892.225 ns |    27.1610 ns |    22.6807 ns |
| Exp2   | 4096         |   1,975.208 ns |    26.4692 ns |    23.4642 ns |
| Exp1   | 65536        |  61,938.616 ns |   354.4827 ns |   296.0091 ns |
| Exp2   | 65536        |  32,390.331 ns |   590.1723 ns |   552.0475 ns |
| Exp1   | 131072       | 125,798.377 ns | 2,260.7971 ns | 2,114.7510 ns |
| Exp2   | 131072       |  61,126.610 ns | 1,091.3727 ns | 1,020.8707 ns |

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -12,6 +12,8 @@ namespace System.Numerics.Tensors
 {
     public static partial class TensorPrimitives
     {
+        private const nuint NonTemporalByteThreshold = 256 * 1024;
```
stephentoub, line 15: Can you add a comment about what this is and how the value was chosen?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1220,102 +1222,892 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+
+                            remainder -= misalignment;
+                        }
+
+                        Vector128<float> vector1;
+                        Vector128<float> vector2;
+                        Vector128<float> vector3;
+                        Vector128<float> vector4;
+
+                        if (canAlign && (remainder > (NonTemporalByteThreshold / sizeof(float))))
```
stephentoub, line 1354: We have two checks on canAlign here.  Would it help to eliminate a branch if this were instead structured as:
```C#
if (canAlign)
{
    ... // do alignment

    if (remainder > (NonTemporalByteThreshold / sizeof(float)))
    {
        ... // handle non-temporal path
        goto AdjustingRefs;        
    }
}
... // what's currently in the else block
```
?

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
```diff
@@ -20,7 +20,7 @@ public static partial class TensorPrimitivesTests
             TensorLengths.Concat(new object[][] { [0] });
 
         public static IEnumerable<object[]> TensorLengths =>
-            from length in Enumerable.Range(1, 128)
+            from length in Enumerable.Range(1, 256)
```
stephentoub, line 23: This is good, but it's also going to double the number of test cases we're running.  Not a big deal except that on netfx it seems to struggle with the theory count.  We should probably subsequently change some of those theories to be loops, e.g. instead of:
```C#
[Theory]
[MemberData(nameof(TensorLengths))]
public void Foo(int tensorLength)
{
    ...
}
```
do:
```C#
[Fact]
public void Foo()
{
    foreach (int tensorLength in TensorLengths)
    {
        ...
    }
}
```

stephentoub (approved):

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding:
```suggestion
        /// <summary>Defines the threshold, in bytes, at which non-temporal stores will be used.</summary>
        /// <remarks>
        ///     A non-temporal store is one that allows the CPU to bypass the cache when writing to memory.
        ///
        ///     This can be beneficial when working with large amounts of memory where the writes would otherwise
        ///     cause large amounts of repeated updates and evictions. The hardware optimization manuals recommend
        ///     the threshold to be roughly half the size of the last level of on-die cache -- that is, if you have approximately
        ///     4MB of L3 cache per core, you'd want this to be approx. 1-2MB, depending on if hyperthreading was enabled.
        ///
        ///     However, actually computing the amount of L3 cache per core can be tricky or error prone. Native memcpy
        ///     algorithms use a constant threshold that is typically around 256KB and we match that here for simplicity. This
        ///     threshold accounts for most processors in the last 10-15 years that had approx. 1MB L3 per core and support
        ///     hyperthreading, giving a per core last level cache of approx. 512KB.
        /// </remarks>
        private const nuint NonTemporalByteThreshold = 256 * 1024;
```

src/libraries/System.Numerics.Tensors/tests/TensorPrimitivesTests.cs
tannergooding: I'll handle this in the immediately following PR covering binary/ternary.

lewing:
This appears to have broken the outerloop tests

tannergooding:
@lewing is there an issue or a link that can be shared so it can be investigated?

lewing:
https://github.com/dotnet/runtime/runs/17652611530

tannergooding:
https://github.com/dotnet/runtime/pull/93412 fixes the issue.

Another PR went in that removed `unsafe` and it didn't conflict, so the nothing caught the issue.

PR #93320, opened 2023-10-10
Rename JitDisablePgo to JitDisablePGO

EgorBo:
Everywhere else in env vars we use upper case for "PGO", e.g. `TieredPGO`, `WritePGOData`, etc.

The problem that env vars are case-sensitive on non-Windows so it's better to maintain a consistency. E.g. currently aspnet/Benchmarks uses it with PGO [here](https://github.com/aspnet/Benchmarks/blob/993e524a4357be6faf8b6460b0a54fe9950af1ef/build/pgo-scenarios.yml#L55) and it explains why it's not respected on Linux here:
image

I don't think it will impact customers as these are internal and for Dynamic PGO we already have public runtimeconfig/msbuild knobs to disable PGO.

Grep: https://grep.app/search?q=JitDisablePgo

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -2738,9 +2738,9 @@ void Compiler::compInitOptions(JitFlags* jitFlags)
         }
         // Optionally, disable use of profile data.
         //
-        else if (JitConfig.JitDisablePgo() > 0)
+        else if (JitConfig.JitDisablePGO() > 0)
         {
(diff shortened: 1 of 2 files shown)


EgorBo:
@AndyAyersMS @dotnet/jit-contrib PTAL

AndyAyersMS (approved):

EgorBo:
FYI @sebastienros since it's going to "regress" on graphs :-) (not sure if you triage regressions on that tab)

Issue #93302, opened 2023-10-10
[Metrics] `error.type` doesn't report `Exception.FullName`


antonfirsov:
`http.client.request.duration`'s `error.type` contains `Exception.Name` for cases which do not map to `HttpRequestError` values, which is inconsistent with aspnet's implementation that reports `Exception.FullName` -- see https://github.com/dotnet/aspnetcore/pull/51084.

The [wording of the .NET Metrics spec](https://github.com/lmolkova/semantic-conventions/blob/dotnet8-metrics/docs/dotnet/dotnet-http-metrics.md#metric-httpclientrequestduration), is somewhat ambivalent:

> When error.type attribute is reported, it contains one of [HTTP Request errors](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Net.Http/src/System/Net/Http/HttpRequestError.cs), a full exception type, or a string representation of received status code.

As https://github.com/dotnet/runtime/pull/93254#discussion_r1352865229 pointed out, this could also mean to *"use HttpRequestException instead of just 'Http' or 'HttpException'"*. Moreover, reporting the full name doesn't bring any value, the set of exceptions we can see here is limited, collisions are not expected. 

I assume this is not the case with aspnet, where exceptions may come from an arbitrarily large set. My recommendation would be to leave and document the current behaviors as-is.

@JamesNK @noahfalk we should make a decision ASAP.

lmolkova:
No opinion here, just adding a few points:
- changing attribute value later would NOT be breaking from OTel perspective
- it might(?) be possible to get custom exceptions if users provide custom certificate check callback (maybe there are other cases?)

noahfalk:
My preference would be to use FullName. I recognize it doesn't provide much value in terms of error disambiguation on this metric, but I do think it could provide value for consistency across metrics. If an engineer was writing queries to process metrics from different parts of .NET it would be a simpler mental model for them if the exception names reported in error.type are always FullName rather than some metrics use Name and others use FullName.

JamesNK:
+1 for consistency.

antonfirsov:
Ok. Gonna add this to #93254, and raise a separate PR for the HTTP part.

MihaZupan:
Are we saying `FullName` is the right choice here? We could also change ASP.NET to use `Name` to get consistency instead :)

antonfirsov:
[...]

PR #93322, opened 2023-10-11
Report Exception.FullName for `http.client.request.duration/error.type`

antonfirsov:
Address #93302 as agreed. This should be backported to 8.0.


karelz (approved):

MihaZupan (approved):

antonfirsov:
The `enterprise-linux` build failure is unrelated.

PR #93336, opened 2023-10-11
CI Test: Enable /O2 for MSVC Release

EgorBo:
Motivation:

image


EgorBo:
Looks correct, going to close then since I can't prove it's useful via TP diffs.
We run TP diffs with Native PGO being disabled so it could be the reason, but then it makes it harder to prove it's a useful change.

PR #93344, opened 2023-10-11
JIT: Fix exponential blowup of memory dependency arrays in VNForMapSelectWork

### LINKED ISSUE #93342: JIT: Exponential blowup in VNForMapSelectWork
The following program hits OOM on x86 and requires ~16 GB of memory to compile on x64. In checked builds we hit an assert while trying to grow an array above 2^32.
The cause is unbounded exponential growth of the memory dependency sets in `VNForMapSelectWork`.
Introduced by #89241. The case was reported by @punchready on the community discord (the example program is based on the reported example).

```csharp
public class Program
{
    [MethodImpl(MethodImplOptions.NoOptimization)]
    public static void Main()
    {
        new Program().Foo();
    }

    public int foo;
    public int bar;
    public int baz;

    [MethodImpl(MethodImplOptions.AggressiveOptimization)]
    private void Foo()
    {
        if (foo == 1)
        {
            bar += 11;
            baz += 11;
        }
        if (foo == 2)
            bar += 12;
        if (foo == 3)
            bar += 13;
        if (foo == 4)
            bar += 14;
        if (foo == 5)
            bar += 15;
        if (foo == 6)
            bar += 16;
        if (foo == 7)
            bar += 17;
        if (foo == 8)
            bar += 18;
        if (foo == 9)
            bar += 19;
        if (foo == 10)
            bar += 20;
        if (foo == 11)
            bar += 21;
        if (foo == 12)
            bar += 22;
        if (foo == 13)
            bar += 23;
        if (foo == 14)
            bar += 24;
        if (foo == 15)
            bar += 25;
        if (foo == 16)
            bar += 26;
        if (foo == 17)
            bar += 27;
        if (foo == 18)
            bar += 28;
        if (foo == 19)
            bar += 29;
        if (foo == 20)
            bar += 30;
        if (foo == 21)
            bar += 31;
        if (foo == 22)
            bar += 32;
        if (foo == 23)
            bar += 33;
        if (foo == 24)
            bar += 34;
        if (foo == 25)
            bar += 35;
        if (foo == 26)
            bar += 36;
        if (foo == 27)
            bar += 37;
        if (foo == 28)
            bar += 38;
        if (foo == 29)
            bar += 39;
        if (foo == 30)
            bar += 40;
        if (foo == 31)
            bar += 41;
        if (foo == 32)
            bar += 42;
        if (foo == 33)
            bar += 43;
        if (foo == 34)
            bar += 44;
        if (foo == 35)
            bar += 45;
        if (foo == 36)
            bar += 46;
        if (foo == 37)
            bar += 47;

        bar = baz;
    }
}

```

The problem is that when we evaluate the last `baz` access we repeatedly combine memory dependencies of the predecessors without removing the accumulating duplicates.

jakobbotsch:
Switch to sets when accumulating the memory dependencies found in
`VNForMapSelectWork`. Otherwise we might duplicate each memory dependency
for every path back to the store that induced it, and there can be an
exponential number of those.

Fix #93342


jakobbotsch:

[No diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=434628&view=ms.vss-build-web.run-extensions-tab)

Some odd MinOpts TP diffs that I wouldn't expect, I want to check why tomorrow.

src/coreclr/jit/valuenum.cpp
```diff
@@ -2907,19 +2980,16 @@ ValueNum ValueNumStore::VNForMapSelectInner(ValueNumKind vnk, var_types type, Va
...
 // Arguments:
-//    alloc      - Allocator to use if memory is required.
+//    comp       - Compiler instance
 //    deps       - Array stack containing the memory dependencies.
-//    startIndex - Start index into 'deps' of memory dependencies.
 //
-void ValueNumStore::MapSelectWorkCacheEntry::SetMemoryDependencies(CompAllocator         alloc,
-                                                                   ArrayStack<ValueNum>& deps,
-                                                                   unsigned              startIndex)
+void ValueNumStore::MapSelectWorkCacheEntry::SetMemoryDependencies(Compiler* comp, SmallValueNumSet& set)
```
AndyAyersMS, line 2986: Update arg comment just above

AndyAyersMS (approved):
Changes LGTM, one minor comment nit.

Couldn't the `MapSelectWorkCacheEntry` hold a `SmallValueNumSet` too? Or do you think that is not worth it because that set is fixed and array rep is more space efficient?

jakobbotsch:
> Couldn't the `MapSelectWorkCacheEntry` hold a `SmallValueNumSet` too? Or do you think that is not worth it because that set is fixed and array rep is more space efficient?

Ah, yes it can, since we are creating a separate set for every recursive call now. I'll make that change and perhaps shrink the default size a bit depending on the memory overhead.

jakobbotsch:
> Couldn't the MapSelectWorkCacheEntry hold a SmallValueNumSet too? Or do you think that is not worth it because that set is fixed and array rep is more space efficient?

Hmm, it has some significant ish memory overhead (1.3% more VN memory, 0.16% more memory total) since `MapSelectWorkCacheEntry` ends up being larger, and there's no simple way to avoid that that I can see.
I think I'll stick with the current code. That also feels a bit safer since we've gone through the internal testing with it now.

jakobbotsch:
> Some odd MinOpts TP diffs that I wouldn't expect, I want to check why tomorrow.

Seems like MSVC stopped inlining something in `fgMorphArgs` causing these differences:
```scala
Base: 447610812, Diff: 447723040, +0.0251%

?fgMorphArgs@Compiler@@AEAAPEAUGenTreeCall@@PEAU2@@Z                              : 67415 : +4.22% : 60.07% : +0.0151%
?AddFinalArgsAndDetermineABIInfo@CallArgs@@QEAAXPEAVCompiler@@PEAUGenTreeCall@@@Z : 43536 : +1.79% : 38.79% : +0.0097%
memcpy                                                                            : 1052  : +0.25% : 0.94%  : +0.0002%
memset                                                                            : 225   : +0.01% : 0.20%  : +0.0001%
```

Apparently MSVC decided it is no longer profitable to inline `fgMakeOutgoingStructArgCopy` based on these VN changes: https://www.diffchecker.com/GP97PRNx/
Not really sure how that works out, but given that inline decisions are likely completely changed by native PGO, it feels safe to ignore.

Issue #75594, opened 2022-09-14
Suboptimal x64 codegen for signed Math.BigMul


tevador:
### Description

The JIT generates suboptimal x64 code for `Math.BigMul(long, long, out long)`.

### Configuration

* .NET 6
* Windows 10
* AMD Ryzen CPU (x64)

### Data

Source code:

```csharp
[MethodImpl(MethodImplOptions.NoInlining | MethodImplOptions.AggressiveOptimization)]
public static void TestBigMul2(ref ulong x, ref ulong y)
{
	x = Math.BigMul(x, y, out y);
}

[MethodImpl(MethodImplOptions.NoInlining | MethodImplOptions.AggressiveOptimization)]
public static void TestBigMul1(ref long x, ref long y)
{
	x = Math.BigMul(x, y, out y);
}
```

Results in the following machine code:

```asm
TestBigMul1(ref ulong, ref ulong):
 push        rax  
 mov         rax,qword ptr [rcx]  
 mov         qword ptr [rsp+18h],rdx  
 mov         r8,qword ptr [rdx]  
 lea         r9,[rsp]  
 mov         rdx,rax  
 mulx        rax,r10,r8  
 mov         qword ptr [r9],r10  
 mov         rdx,qword ptr [rsp]  
 mov         r8,qword ptr [rsp+18h]  
 mov         qword ptr [r8],rdx  
 mov         qword ptr [rcx],rax  
 add         rsp,8  
 ret  

TestBigMul1(ref long, ref long):
 push        rax  
 mov         rax,qword ptr [rcx]  
 mov         qword ptr [rsp+18h],rdx  
 mov         r8,qword ptr [rdx]  
 lea         r9,[rsp]  
 mov         rdx,rax  
 mulx        rdx,r10,r8  
 mov         qword ptr [r9],r10  
 mov         r9,qword ptr [rsp]  
 mov         r10,qword ptr [rsp+18h]  
 mov         qword ptr [r10],r9  
 mov         r9,rax  
 sar         r9,3Fh  
 and         r9,r8  
 sub         rdx,r9  
 sar         r8,3Fh  
 and         rax,r8  
 sub         rdx,rax  
 mov         qword ptr [rcx],rdx  
 add         rsp,8  
 ret
```

### Analysis

The unsigned overload uses a single `mulx` instruction as expected.

[...]

PR #93345, opened 2023-10-11
Use BigMul for 32x32=64 in decimal

lilinus:
No need to have a separate implementation in DecCalc

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Decimal.DecCalc.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Decimal.DecCalc.cs
@@ -184,20 +184,7 @@ namespace System
 
             private static void UInt64x64To128(ulong a, ulong b, ref DecCalc result)
             {
-                ulong low = UInt32x32To64((uint)a, (uint)b); // lo partial prod
-                ulong mid = UInt32x32To64((uint)a, (uint)(b >> 32)); // mid 1 partial prod
-                ulong high = UInt32x32To64((uint)(a >> 32), (uint)(b >> 32));
-                high += mid >> 32;
-                low += mid <<= 32;
-                if (low < mid)  // test for carry
-                    high++;
-
-                mid = UInt32x32To64((uint)(a >> 32), (uint)b);
-                high += mid >> 32;
-                low += mid <<= 32;
-                if (low < mid)  // test for carry
-                    high++;
-
+                ulong high = Math.BigMul(a, b, out ulong low);
                 if (high > uint.MaxValue)
                     Number.ThrowOverflowException(SR.Overflow_Decimal);
                 result.Low64 = low;
--- a/src/libraries/System.Private.CoreLib/src/System/Decimal.DecCalc.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Decimal.DecCalc.cs
@@ -177,6 +177,11 @@ namespace System
                 return (uint)(BitConverter.DoubleToUInt64Bits(d) >> 52) & 0x7FFu;
             }
 
+            private static ulong UInt32x32To64(uint a, uint b)
+            {
+                return (ulong)a * (ulong)b;
+            }
+
             private static void UInt64x64To128(ulong a, ulong b, ref DecCalc result)
             {
                 ulong high = Math.BigMul(a, b, out ulong low);
@@ -376,7 +381,7 @@ namespace System
 
                 // Compute full remainder, rem = dividend - (quo * divisor).
                 //
-                ulong prod = quo * (den & uint.MaxValue); // quo * lo divisor
+                ulong prod = UInt32x32To64(quo, (uint)den); // quo * lo divisor
                 num -= prod;
 
                 if (num > ~prod)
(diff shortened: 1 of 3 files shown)


src/libraries/System.Private.CoreLib/src/System/Decimal.DecCalc.cs
```diff
@@ -184,20 +184,7 @@ private static ulong UInt32x32To64(uint a, uint b)
 
             private static void UInt64x64To128(ulong a, ulong b, ref DecCalc result)
             {
-                ulong low = UInt32x32To64((uint)a, (uint)b); // lo partial prod
```
tannergooding, line 187: Can you delete `UInt32x32To64` as well?

src/libraries/System.Private.CoreLib/src/System/Decimal.DecCalc.cs
```diff
@@ -184,20 +184,7 @@ private static ulong UInt32x32To64(uint a, uint b)
 
             private static void UInt64x64To128(ulong a, ulong b, ref DecCalc result)
```
tannergooding, line 185: This can be replaced, technically speaking, by `Math.BigMul(ulong, ulong, out ulong)` as well.

Then it can also be deleted.

src/libraries/System.Private.CoreLib/src/System/Decimal.DecCalc.cs
lilinus: Could yes, but:
- Unfortunately does not exist any BigMul method for two uint32.
- It is used in a lot of places where narrowing/casting from ulongs to uints are done, and the code is pretty convoluted already. This shows intent quite nicely

src/libraries/System.Private.CoreLib/src/System/Decimal.DecCalc.cs
lilinus: Yeah sure, NB that overflow check and insertion to DecCalc result are still in body below. Do we prefer to move those around to where this is now called or should i keep this method?

tannergooding: Probably better to keep the extra checks centralized here and just defer the algorithm to Math.BigMul

lilinus:
@dotnet-policy-service agree

src/libraries/System.Private.CoreLib/src/System/Decimal.DecCalc.cs
```diff
@@ -381,7 +376,7 @@ private static uint Div96By64(ref Buf12 bufNum, ulong den)
 
                 // Compute full remainder, rem = dividend - (quo * divisor).
                 //
-                ulong prod = UInt32x32To64(quo, (uint)den); // quo * lo divisor
+                ulong prod = quo * (den & uint.MaxValue); // quo * lo divisor
```
tannergooding, line 379: This probably needs a comment on what it's doing.

But, notably, this may regress 32-bit platforms as it will now do a more expensive `64x64=64` multiplication, rather than doing the cheaper `32x32=64`.

In general an internal `Math.BigMul(uint a, uint b, out uint low)` could be defined that users `Bmi2.MultiplyNoFlags`, `ArmBase.MultiplyHigh`, and otherwise falls back to the naive algorithm of `(ulong)a * b`

lilinus: Moved all the stuff into an internal `ulong Math.BigMul(uint, uint)` (since we had a `long Math.BigMul(int, int)`). I tried using initrinsic for x86 on 32 bit. Doesn't seem to exist a `ArmBase.MultiplyHigh` yet.

tannergooding (approved):
LGTM. Thanks!

tannergooding:
Closing and reopening to requeue CI and try to get the tests all passing. They look unrelated.

src/libraries/System.Private.CoreLib/src/System/Decimal.DecCalc.cs
pentp: Did you measure the performance of this change? Last time I checked, a similar change caused a significant performance regression because the codegen for `Bmi2.MultiplyNoFlags` is suboptimal.

src/libraries/System.Private.CoreLib/src/System/Math.cs
```diff
@@ -155,6 +155,19 @@ internal static void ThrowNegateTwosCompOverflow()
             throw new OverflowException(SR.Overflow_NegateTwosCompNum);
         }
 
+        internal static unsafe ulong BigMul(uint a, uint b)
+        {
+#if TARGET_32BIT
+            if (Bmi2.IsSupported)
+            {
+                uint low;
+                uint high = Bmi2.MultiplyNoFlags(a, b, &low);
```
pentp, line 164: Did you measure the performance for this? It's very likely to be significantly worse than the previous codegen that the JIT automatically generated for uint*uint->ulong multiplication on x86.

lilinus: Sorry no i didn't know how to do benchmarks. Probably we should just remove the intrinsic block for x86 in BigMul methods if they are slow? Same for 64 x 64 = 128

pentp: There's an existing issue for MULX coden quality: #11782
And a different related issue: #75594

pentp: And a 3rd issue that is very closely related to this `DecCalc` code: #58263

Issue #76067, opened 2022-09-23
Suboptimal ASM emmited for Vector256<T>.Zero and Vector128<T>.Zero


israellot:
When using Vector256<byte>.Zero, I would expect it to be kept in a fixed register and reused. Instead, what I see is a `vxorps ` operation emitted every time. 

AVX2 has 16 YMM registers.

Bellow is one example. I can get the desired behavior by forcing a zero vector variable instead of using Vector256<byte>.Zero;
Assigning `Vector256<byte>.Zero`  to a variable alone does not do the trick. Only the extra xor operation ensures it stays in a fixed register.

```
var byteVector = Vector256.LoadUnsafe<byte>(ref spanRef);
           
var low = Avx2.UnpackLow(byteVector, Vector256<byte>.Zero);
var high = Avx2.UnpackHigh(byteVector, Vector256<byte>.Zero);

var added = Avx2.Add(low.AsInt16(), high.AsInt16());

added = Avx2.HorizontalAdd(added, Vector256<short>.Zero);
added = Avx2.HorizontalAdd(added, Vector256<short>.Zero);
added = Avx2.HorizontalAdd(added, Vector256<short>.Zero);

//ASM

mov      rax, bword ptr [rcx]
vmovdqu  ymm0, ymmword ptr[rax]
vxorps   ymm1, ymm1, ymm1
vpunpcklbw ymm1, ymm0, ymm1
vxorps   ymm2, ymm2, ymm2
vpunpckhbw ymm0, ymm0, ymm2
vpaddw   ymm0, ymm1, ymm0
vxorps   ymm1, ymm1, ymm1
vphaddw  ymm0, ymm0, ymm1
vxorps   ymm1, ymm1, ymm1
vphaddw  ymm0, ymm0, ymm1
vxorps   ymm1, ymm1, ymm1
vphaddw  ymm0, ymm0, ymm1
```

```
var byteVector = Vector256.LoadUnsafe<byte>(ref spanRef);

var zero = Vector256<byte>.Zero;
zero = Avx2.Xor(zero, zero); //forces fixed register
            
var low = Avx2.UnpackLow(byteVector, zero);
var high = Avx2.UnpackHigh(byteVector, zero);

var added = Avx2.Add(low.AsInt16(), high.AsInt16());
added = Avx2.HorizontalAdd(added, zero.AsInt16());
added = Avx2.HorizontalAdd(added, zero.AsInt16());
[...]

PR #93346, opened 2023-10-11
Improve SpanHelpers.ClearWithReferences for arm64

EgorBo:
Follow up to https://github.com/dotnet/runtime/pull/93214, now for `Span<GC>.Clear();`

I noticed that the existing implementation is slightly sub-optimal on arm: it's relying on complex addressing modes + is not aligned to 16 bytes. Turns out, 16-byte alignment makes a lot of sense even if we don't use SIMD - `stp` (two 8-byte stores) benefits from it as well.

[Standalone benchmark](https://gist.github.com/EgorBo/397fa64ec1ce28d9f8b3240d83730869) (for simpler tests)

**Apple M2 Max** (also tested on Ampere):
```
| Method     | len  | Mean       | Ratio |
|----------- |----- |-----------:|------:|-
| Clear_Base | 1    |   1.199 ns |  1.00 |
| Clear_PR   | 1    |   1.186 ns |  0.99 |
|            |      |            |       |
| Clear_Base | 5    |   1.215 ns |  1.00 |
| Clear_PR   | 5    |   1.215 ns |  1.00 |
|            |      |            |       |
| Clear_Base | 8    |   1.617 ns |  1.00 |
| Clear_PR   | 8    |   1.269 ns |  0.78 |
|            |      |            |       |
| Clear_Base | 16   |   2.405 ns |  1.00 |
| Clear_PR   | 16   |   2.061 ns |  0.86 |
|            |      |            |       |
| Clear_Base | 64   |   8.727 ns |  1.00 |
| Clear_PR   | 64   |   5.415 ns |  0.62 |
|            |      |            |       |
| Clear_Base | 512  |  70.081 ns |  1.00 |
| Clear_PR   | 512  |  35.482 ns |  0.51 |
|            |      |            |       |
| Clear_Base | 1024 | 140.080 ns |  1.00 |
| Clear_PR   | 1024 |  72.254 ns |  0.52 |
|            |      |            |       |
| Clear_Base | 4096 | 559.551 ns |  1.00 |
| Clear_PR   | 4096 | 333.439 ns |  0.60 |
```

**Ryzen 7950X:**
```
| Method     | len  | Mean        | Ratio |
|----------- |----- |------------:|------:|-
| Clear_Base | 1    |   0.3854 ns |  1.00 |
| Clear_PR   | 1    |   0.3798 ns |  0.99 |
|            |      |             |       |
| Clear_Base | 5    |   0.5310 ns |  1.00 |
| Clear_PR   | 5    |   0.3907 ns |  0.74 |
|            |      |             |       |
| Clear_Base | 8    |   0.8995 ns |  1.00 |
| Clear_PR   | 8    |   0.5264 ns |  0.59 |
|            |      |             |       |
| Clear_Base | 16   |   1.4398 ns |  1.00 |
| Clear_PR   | 16   |   1.2039 ns |  0.84 |
|            |      |             |       |
| Clear_Base | 64   |   5.1603 ns |  1.00 |
| Clear_PR   | 64   |   5.1023 ns |  0.99 |
|            |      |             |       |
| Clear_Base | 512  |  46.0601 ns |  1.00 |
| Clear_PR   | 512  |  45.9100 ns |  1.00 |
|            |      |             |       |
| Clear_Base | 1024 |  94.9471 ns |  1.00 |
| Clear_PR   | 1024 |  94.7812 ns |  1.00 |
|            |      |             |       |
| Clear_Base | 4096 | 380.4915 ns |  1.00 |
| Clear_PR   | 4096 | 375.3246 ns |  0.99 |
```
For X64, it slightly improves small inputs. It seems like if we pin the input and use SIMD we can see >2X improvement for large inputs but I was not sure about doing pinning (from GC's point of view, atomicity is fine) here and left as is, cc @jkotas


src/libraries/System.Private.CoreLib/src/System/SpanHelpers.cs
```diff
@@ -337,75 +337,112 @@ public static unsafe void ClearWithReferences(ref IntPtr ip, nuint pointerSizeLe
...
-        Write2To3:
-            Debug.Assert(pointerSizeLength >= 2);
-
-            // Write first two and last one.
-            Unsafe.Add(ref ip, 1) = default;
-            Unsafe.Add(ref Unsafe.Add(ref ip, (nint)pointerSizeLength), -1) = default;
-
-        Write1:
-            Debug.Assert(pointerSizeLength >= 1);
+                // This switch is expected to be lowered to a jump table
```
jkotas, line 342: This works great for microbenchmarks, but it works poorly for real work loads where the jump target is poorly predicated. We used to have switch table like this in managed memcopy originally, but it was later removed.

EgorBo: > This works great for microbenchmarks, but it works poorly for real work loads where the jump target is poorly predicated. We used to have switch table like this in managed memcopy originally, but it was later removed.

Should I revert to the previous shape? I was mostly interested in aligning to 16 bytes (it's the reason I am seeing nice results on arm64) and slightly different addressing mode in the main loop, to make it foldable into `stp`

EgorBo:
Codegen on arm64:
```
; Method Prog:ClearWithReferences2(byref,ulong) (FullOpts)
G_M64195_IG01:  ;; offset=0x0000
            stp     fp, lr, [sp, #-0x10]!
            mov     fp, sp
						;; size=8 bbWeight=8 PerfScore 12.00

G_M64195_IG02:  ;; offset=0x0008
            mov     x2, x1
            cmp     x2, #8
            bhi     G_M64195_IG12
						;; size=12 bbWeight=8 PerfScore 16.00

G_M64195_IG03:  ;; offset=0x0014
            cmp     w2, #8
            bhi     G_M64195_IG12
            mov     w1, w2
            adr     x2, [@RWD00]
            ldr     w2, [x2, x1, LSL #2]
            adr     x3, [G_M64195_IG02]
            add     x2, x2, x3
            br      x2
						;; size=32 bbWeight=4 PerfScore 30.00

G_M64195_IG04:  ;; offset=0x0034
            str     xzr, [x0]
            b       G_M64195_IG15
            align   [0 bytes for IG13]
            align   [0 bytes]
            align   [0 bytes]
            align   [0 bytes]
						;; size=8 bbWeight=0.50 PerfScore 1.00

G_M64195_IG05:  ;; offset=0x003C
            stp     xzr, xzr, [x0]
            b       G_M64195_IG15
						;; size=8 bbWeight=0.50 PerfScore 1.00

G_M64195_IG06:  ;; offset=0x0044
            stp     xzr, xzr, [x0]
            str     xzr, [x0, #0x10]
            b       G_M64195_IG15
						;; size=12 bbWeight=0.50 PerfScore 1.50

G_M64195_IG07:  ;; offset=0x0050
            stp     xzr, xzr, [x0]
            stp     xzr, xzr, [x0, #0x10]
            b       G_M64195_IG15
						;; size=12 bbWeight=0.50 PerfScore 1.50

G_M64195_IG08:  ;; offset=0x005C
            stp     xzr, xzr, [x0]
            stp     xzr, xzr, [x0, #0x10]
            str     xzr, [x0, #0x20]
            b       G_M64195_IG15
						;; size=16 bbWeight=0.50 PerfScore 2.00

G_M64195_IG09:  ;; offset=0x006C
            stp     xzr, xzr, [x0]
            stp     xzr, xzr, [x0, #0x10]
            stp     xzr, xzr, [x0, #0x20]
            b       G_M64195_IG15
						;; size=16 bbWeight=0.50 PerfScore 2.00

G_M64195_IG10:  ;; offset=0x007C
            stp     xzr, xzr, [x0]
            stp     xzr, xzr, [x0, #0x10]
            stp     xzr, xzr, [x0, #0x20]
            str     xzr, [x0, #0x30]
            b       G_M64195_IG15
						;; size=20 bbWeight=0.50 PerfScore 2.50

G_M64195_IG11:  ;; offset=0x0090
            stp     xzr, xzr, [x0]
            stp     xzr, xzr, [x0, #0x10]
            stp     xzr, xzr, [x0, #0x20]
            stp     xzr, xzr, [x0, #0x30]
            b       G_M64195_IG15
						;; size=20 bbWeight=0.50 PerfScore 2.50

G_M64195_IG12:  ;; offset=0x00A4
            lsr     x2, x1, #3
						;; size=4 bbWeight=4 PerfScore 4.00

G_M64195_IG13:  ;; offset=0x00A8
            stp     xzr, xzr, [x0]
            stp     xzr, xzr, [x0, #0x10]
            stp     xzr, xzr, [x0, #0x20]
            stp     xzr, xzr, [x0, #0x30]
            add     x0, x0, #64
            sub     x2, x2, #1
            cbnz    x2, G_M64195_IG13
						;; size=28 bbWeight=32 PerfScore 192.00

G_M64195_IG14:  ;; offset=0x00C4
            and     x1, x1, #7
            b       G_M64195_IG02
						;; size=8 bbWeight=4 PerfScore 6.00

G_M64195_IG15:  ;; offset=0x00CC
            ldp     fp, lr, [sp], #0x10
            ret     lr
RWD00  	dd	000000C4h ; case G_M64195_IG15
       	dd	0000002Ch ; case G_M64195_IG04
       	dd	00000034h ; case G_M64195_IG05
       	dd	0000003Ch ; case G_M64195_IG06
       	dd	00000048h ; case G_M64195_IG07
       	dd	00000054h ; case G_M64195_IG08
       	dd	00000064h ; case G_M64195_IG09
       	dd	00000074h ; case G_M64195_IG10
       	dd	00000088h ; case G_M64195_IG11
; Total bytes of code: 212
```

Previous codegen:
```asm
; Method Prog:ClearWithReferences(byref,ulong) (FullOpts)
G_M45361_IG01:  ;; offset=0x0000
            stp     fp, lr, [sp, #-0x10]!
            mov     fp, sp

G_M45361_IG02:  ;; offset=0x0008
            cmp     x1, #8
            blo     G_M45361_IG04
            align   [0 bytes for IG03]
            align   [0 bytes]
            align   [0 bytes]
            align   [0 bytes]

G_M45361_IG03:  ;; offset=0x0010
            lsl     x2, x1, #3
            add     x3, x0, x2
            str     xzr, [x3, #-0x08]
            add     x3, x0, x2
            str     xzr, [x3, #-0x10]
            add     x3, x0, x2
            str     xzr, [x3, #-0x18]
            add     x3, x0, x2
            str     xzr, [x3, #-0x20]
            add     x3, x0, x2
            str     xzr, [x3, #-0x28]
            add     x3, x0, x2
            str     xzr, [x3, #-0x30]
            add     x3, x0, x2
            str     xzr, [x3, #-0x38]
            add     x2, x0, x2
            str     xzr, [x2, #-0x40]
            sub     x1, x1, #8
            cmp     x1, #8
            bhs     G_M45361_IG03
						;; size=80 bbWeight=4 PerfScore 60.00

G_M45361_IG04:  ;; offset=0x0060
            cmp     x1, #4
            bhs     G_M45361_IG07

G_M45361_IG05:  ;; offset=0x0068
            cmp     x1, #2
            bhs     G_M45361_IG08
            cbnz    x1, G_M45361_IG09
						;; size=12 bbWeight=0.50 PerfScore 1.25

G_M45361_IG06:  ;; offset=0x0074
            ldp     fp, lr, [sp], #0x10
            ret     lr
						;; size=8 bbWeight=0.50 PerfScore 1.00

G_M45361_IG07:  ;; offset=0x007C
            stp     xzr, xzr, [x0, #0x10]
            lsl     x2, x1, #3
            add     x3, x0, x2
            str     xzr, [x3, #-0x18]
            add     x2, x0, x2
            str     xzr, [x2, #-0x10]
						;; size=24 bbWeight=0.50 PerfScore 2.50

G_M45361_IG08:  ;; offset=0x0094
            str     xzr, [x0, #0x08]
            lsl     x2, x1, #3
            add     x1, x0, x2
            str     xzr, [x1, #-0x08]
						;; size=16 bbWeight=0.50 PerfScore 1.75

G_M45361_IG09:  ;; offset=0x00A4
            str     xzr, [x0]
						;; size=4 bbWeight=0.50 PerfScore 0.50

G_M45361_IG10:  ;; offset=0x00A8
            ldp     fp, lr, [sp], #0x10
            ret     lr
						;; size=8 bbWeight=0.50 PerfScore 1.00
; Total bytes of code: 176
```

src/libraries/System.Private.CoreLib/src/System/SpanHelpers.cs
EgorBo: Btw, presumably, PGO can re-shuffle the switch (or convert to if-else) based on profile

jkotas: > Should I revert to the previous shape?

Yes, unless you can prove that the switch with jump table is better for real workloads.

For reference, https://github.com/dotnet/coreclr/pull/9786 is the PR that got rid of the switch from memory copy. It went through extensive validation.

src/libraries/System.Private.CoreLib/src/System/SpanHelpers.cs
```diff
@@ -406,8 +420,35 @@ public static unsafe void ClearWithReferences(ref IntPtr ip, nuint pointerSizeLe
 
             // Write only element.
             ip = default;
+            return;
+
+        LargeInput:
+            // Branch-less alignment to 64 bytes.
+            Unsafe.As<nint, Block64>(ref ip) = default;
```
jkotas, line 427: This makes fragile assumption about the instruction that the JIT is going to compile this into. It is valid for the JIT to compile this into one-byte-at-time assignment.

src/libraries/System.Private.CoreLib/src/System/SpanHelpers.cs
tannergooding: > This works great for microbenchmarks, but it works poorly for real work loads where the jump target is poorly predicated. We used to have switch table like this in managed memcopy originally, but it was later removed.

@jkotas, I'm not sure what you mean here, could you elaborate?

The hardware optimization manuals and underlying native memcpy algorithms across all 3 major implementations (MSVC, GCC, Clang; for x64 and Arm64) use a jump table explicitly for this reason. Because an unconditional and unpredictable branch, is better than a tight tree of conditional and still unpredictable branches.

The premise is that a fallthrough jump table allows a tight, unrolled version of handling for the trailing elements to exist. This then requires 1 unpredicted branch to decide which of the immediately following and equidistant switch cases needs to be jumped to. These cases are often already in the instruction/decode cache due to being in the same or following 32-byte code window.

EgorBo:
I have mixed results with this changes, I'll wait for fixes for https://github.com/dotnet/runtime/issues/93382 (needed for the remainder) and https://github.com/dotnet/runtime/issues/76067#issuecomment-1761889198 (needed for the main loop) and try again after

src/libraries/System.Private.CoreLib/src/System/SpanHelpers.cs
jkotas: https://github.com/dotnet/coreclr/pull/9786 was done by Intel engineers to address bottlenecks observed in real workloads. Yes, it goes contrary to the guidance you have quoted.

Possible explanation: Conditional jumps and indirect jumps use two different predictors. The relative performance of conditional jumps vs. indirect jump depends on relative pressure on each predictor. .NET code tends to have more indirect jumps than an average code. It means that the indirect jump predictor sees more pressure in .NET code and it is more profitable to depend on indirect jump predictor in .NET memcpy implementation.

PR #93349, opened 2023-10-11
Skip non virtual generic methods in build_imt_slots

bholmes:
When processing the methods of an interface, check that a generic method has the virtual flag to avoid corrupting the vtable.

### DIFF
--- a/src/mono/mono/metadata/object.c
+++ b/src/mono/mono/metadata/object.c
@@ -1549,8 +1549,10 @@ build_imt_slots (MonoClass *klass, MonoVTable *vt, gpointer* imt, GSList *extra_
 			}
 			method = mono_class_get_method_by_index (iface, method_slot_in_interface);
 			if (method->is_generic) {
-				has_generic_virtual = TRUE;
-				vt_slot ++;
+				if (m_method_is_virtual (method)) {
+					has_generic_virtual = TRUE;
+					vt_slot ++;
+				}
 				continue;
 			}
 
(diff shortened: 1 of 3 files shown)


src/mono/mono/metadata/object.c
```diff
@@ -1576,7 +1578,7 @@ build_imt_slots (MonoClass *klass, MonoVTable *vt, gpointer* imt, GSList *extra_
 			for (method_slot_in_interface = 0; method_slot_in_interface < mcount; method_slot_in_interface++) {
 				MonoMethod *method = mono_class_get_method_by_index (iface, method_slot_in_interface);
 
-				if (method->is_generic)
+				if (method->is_generic && m_method_is_virtual(method))
 					has_generic_virtual = TRUE;
```
bholmes, line 1582: Pretty sure this is dead code, but I wanted to highlight the potential problem.

vargaz (approved):

lambdageek:
I think this is preempted by #94437 - just need to try the test from this PR to validate that it's the same codepath.  Sorry it took us a while to revisit this.

src/mono/mono/metadata/object.c
lambdageek: yea, I ended up deleting this in #93349 - it was only used for COM interop

lambdageek:
Yea this is the same issue.  It's been fixed by various backports in the January servicing releases (8.0.101, 7.0.15, 6.0.26)

Thanks for looking into it, @bholmed.

Issue #93314, opened 2023-10-02
Name of WebAssembly tools in VS setup doesn't deliniate what a versionless description is


timheuer:
### Description

![image](https://github.com/dotnet/runtime/assets/4821/2af919fe-8171-45e8-a865-baabb1a4aa4e)
In this view, it is not clear what the versionless is.  Feedback from customers and internal has been confusion.  Just put the version number always

### Reproduction Steps

Run VS Setup and look for WebAssembly component

### Expected behavior

See specific versions

### Actual behavior

Two are versioned, one is not

marcpopMSFT:
The string is actually generated from the workload manifest description of the workload here: https://github.com/dotnet/runtime/blob/main/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Current.Manifest/WorkloadManifest.json.in#L8. It probably makes sense to put the version even in the current manifest to make it easier to differentiate in VS.

marcpopMSFT:
@steveisok

radical:
The string in the screenshot is different from the manifest json though:
`NET 7.0 WebAssembly build tools` vs `.NET WebAssembly build tools for net7.0`(https://github.com/dotnet/runtime/blob/main/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.net7.Manifest/WorkloadManifest.json.in#L8).
Is a corresponding VS specific change needed somewhere too?

PR #93359, opened 2023-10-11
[release/8.0][wasm] Update workload description to include target framework

radical:
Issue: https://github.com/dotnet/runtime/issues/93314

## Customer Impact

Disambiguates the workload description for the 8.0 workload, so instead of `.NET WebAssembly build tools` the user will see `.NET WebAssembly build tools for net8.0`. And that's the same for 6.0, and 7.0 wasm workloads.

## Testing

Manual.

## Risk

None. Just a change in description.

### DIFF
--- a/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Current.Manifest/WorkloadManifest.json.in
+++ b/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Current.Manifest/WorkloadManifest.json.in
@@ -15,7 +15,7 @@
       "platforms": [ "win-x64", "win-arm64", "linux-x64", "linux-arm64", "osx-x64", "osx-arm64"]
     },
     "wasm-experimental": {
-      "description": ".NET WebAssembly experimental tooling",
+      "description": ".NET WebAssembly experimental tooling for net8.0",
       "packs": [
         "Microsoft.NET.Runtime.WebAssembly.Templates",
         "Microsoft.NETCore.App.Runtime.Mono.multithread.browser-wasm",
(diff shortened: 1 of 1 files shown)


radical:
@marcpopMSFT do I need to change the corresponding localized descriptions?

steveisok (approved):

marcpopMSFT:
> @marcpopMSFT do I need to change the corresponding localized descriptions?

@radical what's your process for localizing these? I think yes but I haven't actually tested out a localized VS install (on the CLI, we will display the localized description in dotnet workload search)

radical:
> > @marcpopMSFT do I need to change the corresponding localized descriptions?
> 
> @radical what's your process for localizing these? I think yes but I haven't actually tested out a localized VS install (on the CLI, we will display the localized description in dotnet workload search)

@lewing @steveisok ^

carlossanlop:
@radical can you please fill out the PR template in the description and send an email to Tactics requesting approval?

radical:
> > @marcpopMSFT do I need to change the corresponding localized descriptions?
> 
> @radical what's your process for localizing these? I think yes but I haven't actually tested out a localized VS install (on the CLI, we will display the localized description in dotnet workload search)

IIUC, we don't really have a process for that, or seem to update them really. I guess, it is only being done on VS side then.

radical:
@carlossanlop this is ready.

marcpopMSFT:
I doubt the loc is done on the VS side as it probably uses the value from here. @joeloff do you know for sure how loc versions of VS work for workload descriptions/names?

joeloff:
> I doubt the loc is done on the VS side as it probably uses the value from here. @joeloff do you know for sure how loc versions of VS work for workload descriptions/names?

There is no localization on the VS side because the component is built in the runtime repo and the tooling used to generate the setup authoring is unaware of the localization strings in the workload manifest package. It is a known issue. Fixing the strings in runtime will update the localization experience for CLI based installations only

carlossanlop:
> @carlossanlop this is ready.

Let's merge this!

Issue #82336, opened 2023-02-18
JIT: reconsider use of BBF_DONT_REMOVE on EH blocks


BruceForstall:
try/handler begin/last blocks are marked with BBF_DONT_REMOVE. Do we still need them to be? Can we check the block EH index instead when necessary to determine the EH status of a block? Can we generalize flow updates?

PR #93367, opened 2023-10-11
Remove `BBF_TRY_BEG`

SingleAccretion:
The flag duplicates information already present in the EH table.

Also delete some (not all) remnants of verification from `impVerifyEHBlock`.


SingleAccretion:
Failures look like #93404.

[No diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=436331&view=ms.vss-build-web.run-extensions-tab), but a minor TP win, @dotnet/jit-contrib .

src/coreclr/jit/jiteh.cpp
```diff
@@ -3499,12 +3498,6 @@ void Compiler::fgVerifyHandlerTab()
 #endif // FEATURE_EH_FUNCLETS
         }
 
-        // Only the first block of 'try' regions should have BBF_TRY_BEG set.
-        if (!blockTryBegSet[block->bbNum])
```
BruceForstall, line 3503: The only use of `blockTryBegSet` currently is in this check. You could remove it entirely.

src/coreclr/jit/fgopt.cpp
```diff
@@ -6400,10 +6400,9 @@ bool Compiler::fgUpdateFlowGraph(bool doTailDuplication, bool isPhase)
                 goto REPEAT;
             }
 
-            /* Remove unreachable or empty blocks - do not consider blocks marked BBF_DONT_REMOVE or genReturnBB block
-             * These include first and last block of a TRY, exception handlers and RANGE_CHECK_FAIL THROW blocks */
-
-            if ((block->bbFlags & BBF_DONT_REMOVE) == BBF_DONT_REMOVE || block == genReturnBB)
```
BruceForstall, line 6406: Is the `genReturnBB` check just unnecessary? (obsolete?)

BruceForstall (commented):
That's a nice cleanup. A couple notes.

As a follow-up: do you have any thoughts on https://github.com/dotnet/runtime/issues/82336?

src/coreclr/jit/fgopt.cpp
SingleAccretion: Unnecessary; the block is [already always marked](https://github.com/dotnet/runtime/blob/946c5245aea149d9ece53fd0debc18328c29083b/src/coreclr/jit/flowgraph.cpp#L2357-L2359) with `BBF_DONT_REMOVE`.

SingleAccretion:
> As a follow-up: do you have any thoughts on #82336?

Great question :). I am looking at the EH-related flowgraph things and in particular the ability to remove dead protected regions and handlers.

src/coreclr/jit/jiteh.cpp
SingleAccretion: Deleted.

BruceForstall:
@SingleAccretion Can you fix the merge conflict?

fwiw, all the test failures are "known"

SingleAccretion:
@BruceForstall conflict fixed.

BruceForstall (approved):

PR #93368, opened 2023-10-11
Enable IlasmRoundTrip tests for merged tests

TIHan:
Added a simple python script that gets generated in the merged test projects that handles roundtripping.
This also adds a very simple tool to check assemblies if they are managed and/or debug assemblies.

### DIFF
--- a/src/coreclr/tools/AssemblyChecker/AssemblyChecker.csproj
+++ b/src/coreclr/tools/AssemblyChecker/AssemblyChecker.csproj
@@ -1,0 +1,58 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <AssemblyName>AssemblyChecker</AssemblyName>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>$(NetCoreAppToolCurrent)</TargetFramework>
+    <PlatformTarget>AnyCPU</PlatformTarget>
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+    <AppendTargetFrameworkToOutputPath Condition="'$(BuildingInsideVisualStudio)' == 'true'">true</AppendTargetFrameworkToOutputPath>
+    <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
+    <EnableDefaultEmbeddedResourceItems>false</EnableDefaultEmbeddedResourceItems>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <OutputPath>$(RuntimeBinDir)\AssemblyChecker</OutputPath>
+    <RunAnalyzers>false</RunAnalyzers>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- Massage the OutputRID into an JITTools package RID that we can download -->
+    <_jitToolsRidPlatformIndex>$(OutputRID.LastIndexOf('-'))</_jitToolsRidPlatformIndex>
+    <JITToolsRidWithoutPlatform>$(OutputRID.Substring(0, $(_jitToolsRidPlatformIndex)))</JITToolsRidWithoutPlatform>
+    <JITToolsRidPlatform>$(OutputRID.Substring($(_jitToolsRidPlatformIndex)).TrimStart('-'))</JITToolsRidPlatform>
+
+    <!-- If it's not win/osx/linux-musl, it's a non-portable Linux. Treat as Linux. -->
+    <JITToolsRidWithoutPlatform Condition="'$(JITToolsRidWithoutPlatform)' != 'win' and '$(JITToolsRidWithoutPlatform)' != 'osx' and '$(JITToolsRidWithoutPlatform)' != 'linux-musl'">linux</JITToolsRidWithoutPlatform>
+
+    <!-- There are no x86 packages, so use x64 -->
+    <JITToolsRidPlatform Condition="'$(JITToolsRidPlatform)' == 'x86'">x64</JITToolsRidPlatform>
+    <JITToolsRidPlatform Condition="'$(JITToolsRidPlatform)' == 'arm' and '$(JITToolsRidWithoutPlatform)' == 'win'">arm64</JITToolsRidPlatform>
+    <JITToolsRid Condition="'$(JITToolsRid)' == ''">$(JITToolsRidWithoutPlatform)-$(JITToolsRidPlatform)</JITToolsRid>
+
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'linux-arm64'">$(runtimelinuxarm64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'linux-x64'">$(runtimelinuxx64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'linux-musl-arm64'">$(runtimelinuxmuslarm64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'linux-musl-x64'">$(runtimelinuxmuslx64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'win-arm64'">$(runtimewinarm64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'win-x64'">$(runtimewinx64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'osx-arm64'">$(runtimeosxarm64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'osx-x64'">$(runtimeosxx64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="$(MicrosoftCodeAnalysisCSharpVersion)" />
+    <PackageDownload Condition="'$(DotNetBuildFromSource)' != 'true'" Include="runtime.$(JITToolsRid).Microsoft.NETCore.Runtime.JIT.Tools">
+      <Version>[$(JITToolsVersion)]</Version>
+    </PackageDownload>
+  </ItemGroup>
+
+  <Target Name="CopyJitTools" AfterTargets="Build">
+    <ItemGroup>
+      <JitToolsFiles Include="$(NuGetPackageRoot)\runtime.$(JITToolsRid).microsoft.netcore.runtime.jit.tools\$(JITToolsVersion)\runtimes\**\*.*" />
+    </ItemGroup>
+    <Copy SourceFiles="@(JitToolsFiles)"
+          SkipUnchangedFiles="true"
+          OverwriteReadOnlyFiles="true"
+          DestinationFolder="$(OutputPath)\runtimes\%(RecursiveDir)" />
+  </Target>
+
+</Project>
--- a/src/coreclr/tools/AssemblyChecker/AssemblyChecker.csproj
+++ b/src/coreclr/tools/AssemblyChecker/AssemblyChecker.csproj
@@ -13,46 +13,4 @@
     <OutputPath>$(RuntimeBinDir)\AssemblyChecker</OutputPath>
     <RunAnalyzers>false</RunAnalyzers>
   </PropertyGroup>
-
-  <PropertyGroup>
-    <!-- Massage the OutputRID into an JITTools package RID that we can download -->
-    <_jitToolsRidPlatformIndex>$(OutputRID.LastIndexOf('-'))</_jitToolsRidPlatformIndex>
-    <JITToolsRidWithoutPlatform>$(OutputRID.Substring(0, $(_jitToolsRidPlatformIndex)))</JITToolsRidWithoutPlatform>
-    <JITToolsRidPlatform>$(OutputRID.Substring($(_jitToolsRidPlatformIndex)).TrimStart('-'))</JITToolsRidPlatform>
-
-    <!-- If it's not win/osx/linux-musl, it's a non-portable Linux. Treat as Linux. -->
-    <JITToolsRidWithoutPlatform Condition="'$(JITToolsRidWithoutPlatform)' != 'win' and '$(JITToolsRidWithoutPlatform)' != 'osx' and '$(JITToolsRidWithoutPlatform)' != 'linux-musl'">linux</JITToolsRidWithoutPlatform>
-
-    <!-- There are no x86 packages, so use x64 -->
-    <JITToolsRidPlatform Condition="'$(JITToolsRidPlatform)' == 'x86'">x64</JITToolsRidPlatform>
-    <JITToolsRidPlatform Condition="'$(JITToolsRidPlatform)' == 'arm' and '$(JITToolsRidWithoutPlatform)' == 'win'">arm64</JITToolsRidPlatform>
-    <JITToolsRid Condition="'$(JITToolsRid)' == ''">$(JITToolsRidWithoutPlatform)-$(JITToolsRidPlatform)</JITToolsRid>
-
-    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'linux-arm64'">$(runtimelinuxarm64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
-    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'linux-x64'">$(runtimelinuxx64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
-    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'linux-musl-arm64'">$(runtimelinuxmuslarm64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
-    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'linux-musl-x64'">$(runtimelinuxmuslx64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
-    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'win-arm64'">$(runtimewinarm64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
-    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'win-x64'">$(runtimewinx64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
-    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'osx-arm64'">$(runtimeosxarm64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
-    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'osx-x64'">$(runtimeosxx64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
-  </PropertyGroup>
-
-  <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="$(MicrosoftCodeAnalysisCSharpVersion)" />
-    <PackageDownload Condition="'$(DotNetBuildFromSource)' != 'true'" Include="runtime.$(JITToolsRid).Microsoft.NETCore.Runtime.JIT.Tools">
-      <Version>[$(JITToolsVersion)]</Version>
-    </PackageDownload>
-  </ItemGroup>
-
-  <Target Name="CopyJitTools" AfterTargets="Build">
-    <ItemGroup>
-      <JitToolsFiles Include="$(NuGetPackageRoot)\runtime.$(JITToolsRid).microsoft.netcore.runtime.jit.tools\$(JITToolsVersion)\runtimes\**\*.*" />
-    </ItemGroup>
-    <Copy SourceFiles="@(JitToolsFiles)"
-          SkipUnchangedFiles="true"
-          OverwriteReadOnlyFiles="true"
-          DestinationFolder="$(OutputPath)\runtimes\%(RecursiveDir)" />
-  </Target>
-
 </Project>
--- a/src/coreclr/tools/AssemblyChecker/Program.cs
+++ b/src/coreclr/tools/AssemblyChecker/Program.cs
@@ -1,6 +1,10 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+// This is a simple console application that is designed to answer True or False
+// questions about whether a give file is a managed assembly or not.
+// You can also ask whether or not the assembly is debuggable.
+
 using System.Text;
 using System.Diagnostics;
 using System.Reflection.Metadata;
@@ -79,12 +83,8 @@ namespace AssemblyChecker
         {
             if (args.Length == 0)
             {
-                throw new ArgumentException("Expected assembly file-path.");
-            }
-
-            if (args.Length > 2)
-            {
-                throw new ArgumentException("Too many arguments.");
+                Console.Error.WriteLine("Expected assembly file-path.");
+                return 2;
             }
 
             if (args.Length == 1)
@@ -106,11 +106,13 @@ namespace AssemblyChecker
                 }
                 else
                 {
-                    throw new ArgumentException("Invalid option.");
+                    Console.Error.WriteLine("Invalid option.");
+                    return 2;
                 }
             }
 
-            return -1;
+            Console.Error.WriteLine("Too many arguments.");
+            return 2;
         }
     }
 }
--- a/src/coreclr/tools/AssemblyChecker/Program.cs
+++ b/src/coreclr/tools/AssemblyChecker/Program.cs
@@ -131,11 +131,12 @@ namespace AssemblyChecker
                             else
                                 return 1;
                         }
-                }
-                else
-                {
-                    Console.Error.WriteLine("Invalid option.");
-                    return 2;
+
+                    default:
+                        {
+                            Console.Error.WriteLine("Invalid option.");
+                            return 2;
+                        }
                 }
             }
 
--- a/src/coreclr/tools/AssemblyChecker/Program.cs
+++ b/src/coreclr/tools/AssemblyChecker/Program.cs
@@ -81,6 +81,21 @@ namespace AssemblyChecker
             return attrs.Any(x => HasDebuggableAttribute(reader, reader.GetCustomAttribute(x).Constructor));
         }
 
+        static bool IsExe(string path)
+        {
+            using var fs = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
+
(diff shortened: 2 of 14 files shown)


TIHan:
I got the roundtrips working for windows, but re-enabling the roundtrip tests, we have some failures: https://dev.azure.com/dnceng-public/public/_build/results?buildId=437210&view=ms.vss-test-web.build-test-results-tab&runId=9719900&resultId=120758&paneView=debug

I'll consider disabling these tests in roundtrips so we can move this PR forward. I'll see how bad the failures are.

src/tests/Common/CLRTest.Jit.targets
```diff
@@ -365,6 +372,139 @@ set DOTNET_JitPath=%CORE_ROOT%\superpmi-shim-collector.dll
...
+
+if "arm" in platform.machine():
+  # TODO: Fix these roundtrip assemblies from failing on ARM64 when executing.
+  assembliesToIgnore = [
+    "value_numbering_checked_arithmetic_with_constants_ro.dll",
+    "ldc_mul_ovf_i4.dll",
+    "overflow02_mul.dll",
+    "b25815.dll",
+    "b71869.dll"
+  ]
```
markples, line 459: Does IlasmRoundTripIncompatible still work with this?  This work shouldn't be blocked by failing tests - just disable them - but the mechanism needs to be accessible like IlasmRoundTripIncompatible rather than changing this script.

src/tests/Common/CLRTest.Jit.targets
```diff
@@ -365,6 +372,139 @@ set DOTNET_JitPath=%CORE_ROOT%\superpmi-shim-collector.dll
...
+      <_IlasmRoundTripScriptText><![CDATA[
+import os
+import shutil
+import subprocess
+import sys
+import glob
+import struct
+import platform
+
+def is_managed_assembly(file):
```
markples, line 395: This is useful functionality of course, but it should probably be a standalone utility. You can avoid having to deal with file format details if you use https://learn.microsoft.com/en-us/dotnet/standard/assembly/identify.

src/tests/Common/CLRTest.Jit.targets
```diff
@@ -365,6 +372,139 @@ set DOTNET_JitPath=%CORE_ROOT%\superpmi-shim-collector.dll
...
+    "ldc_mul_ovf_i4.dll",
+    "overflow02_mul.dll",
+    "b25815.dll",
+    "b71869.dll"
+  ]
+else:
+  assembliesToIgnore = []
+
+for inputAssemblyName in glob.glob("*.dll"):
+  if (not inputAssemblyName.endswith("$(AssemblyName).dll")) and is_managed_assembly(inputAssemblyName):
```
markples, line 464: Why is the `not` clause needed?  Isn't that the main dll for the test?

In a previous version, you had the list of files from the build system.  That seems simpler - what went wrong there?

src/tests/Common/CLRTest.Jit.targets
```diff
@@ -365,6 +372,139 @@ set DOTNET_JitPath=%CORE_ROOT%\superpmi-shim-collector.dll
...
+    file_size = os.path.getsize(file)
+
+    if file_size < 64:
+        return False
+
+    with open(file, "rb") as f:
+        f.seek(0x3c)
+
+        # PE Header starts @ 0x3C (60). Its a 4 byte header.
+        peHeaderPointer = struct.unpack('I', f.read(4))[0]
```
markples, line 405: I think you need to specify byte order for all of these.  I think that's an indication that the C# link in my other comment is a better way to go.

src/tests/Common/CLRTest.Jit.targets
```diff
@@ -365,6 +372,139 @@ set DOTNET_JitPath=%CORE_ROOT%\superpmi-shim-collector.dll
...
+  if (not inputAssemblyName.endswith("$(AssemblyName).dll")) and is_managed_assembly(inputAssemblyName):
+    if inputAssemblyName not in assembliesToIgnore:
+      print("ILASM RoundTrip for " + inputAssemblyName)
+
+      ilasmSwitches = ""
+
+      # HACK: The 'poison.csproj' is the only project that needs to be built with the debug flag.
+      #       We do this here because we cannot easily determine whether or not it was actually built with the flag or not.
+      if inputAssemblyName.endswith("poison.dll"):
+        ilasmSwitches = "-DEBUG"
```
markples, line 473: See `_IlasmSwitches`

src/tests/Common/CLRTest.Jit.targets
TIHan: It doesn't and there isn't an easy way to know.

src/tests/Common/CLRTest.Jit.targets
TIHan: It should be, but we can deal with it later.

src/tests/Common/CLRTest.Jit.targets
TIHan: Yes, it is the main dll for the test. We still do the old ilasmroundtrip for 'exe' assemblies, and I do not want it to roundtrip a second time.

src/tests/Common/CLRTest.Jit.targets
```diff
@@ -365,6 +372,139 @@ set DOTNET_JitPath=%CORE_ROOT%\superpmi-shim-collector.dll
...
+
+      disassemblyName = os.path.join("IL-RT", os.path.splitext(os.path.basename(inputAssemblyName))[0] + ".il")
+      targetAssemblyName = os.path.join("IL-RT", os.path.basename(inputAssemblyName))
+
+      shutil.copyfile(inputAssemblyName, targetAssemblyName)
+
+      proc = subprocess.Popen(os.environ["CORE_ROOT"] + "/ildasm /raweh /unicode /out=" + disassemblyName + " " + targetAssemblyName)
```
markples, line 480: f-strings would improve this and the ilasm line

`subprocess.Popen(f'{os.environ["CORE_ROOT"]}/ildasm /raweh /unicode /out={disassemblyName} {targetAssemblyName}')`

src/tests/Common/CLRTest.Jit.targets
TIHan: After looking at CI, these tests were actually failing in non-ilasm roundtrip runs, so I suspect I will remove this.

src/tests/Common/CLRTest.Jit.targets
TIHan: `_IlasmSwitches` will not work here unfortunately. I tried this. The problem is that `_IlasmSwitches` refers to the merged test assembly and not `poison.csproj` which specifies the `DebugType`.

src/tests/Common/CLRTest.Jit.targets
markples: Wait, isn't this just called from the main .cmd/.sh scripts?  IlasmRoundTripIncompatible would stop the execution up front and never get here.

src/tests/Common/CLRTest.Jit.targets
markples: Delete the old ilasmroundtrip code so that we don't have two copies of it.

TIHan: This will be hard because I don't know if the assembly is a library or exe when we do roundtripping.

TIHan: Maybe we don't care about the existing tests that use the old roundtripping?

src/tests/Common/CLRTest.Jit.targets
TIHan: Right, but that only works if the merged test project is `IlasmRoundTripIncompatible`, not the individual test itself unfortunately.

markples:
A general comment - while the existing build system already writes the .cmd/.sh scripts for each test, I don't think it's a great pattern.  It conflates script logic and data in the build system output (and also raises security questions when we are writing code and then executing it).  Instead, just put the script in src\tests or src\tests\common, and add arguments to make it work for each test.  Then you can just execute the script on its own for testing.  (This applies to both the entire script and is_managed_assembly.)

src/tests/Common/CLRTest.Jit.targets
markples: This seems like a big problem.  There are lots of other tests that specify DebugType.

This is getting beyond my msbuild experience, but those properties are all around.  We may need to persist them.

src/tests/Common/CLRTest.Jit.targets
markples: Short-term fix for that is that RequiresProcessIsolation should be set.

src/tests/Common/CLRTest.Jit.targets
markples: I'm not sure what you mean.  Most test project have a single binary and use old roundtripping, so we definitely care about running them.  If you mean specifically whether this PR's code should handle those binaries, we don't want to separate ways to run roundtripping.  That will confuse the debugging process and is difficult to maintain.

src/tests/Common/CLRTest.Jit.targets
TIHan: I checked all the tests with `IlasmRoundTripIncomptabiel` and they all have either have OutputType=Exe or RequiresProcessIsolation=true, so we don't have to worry about it for now.

src/tests/Common/CLRTest.Jit.targets
TIHan: This should be in the right bye order already; I've already checked the values that come out of it.

src/tests/Common/CLRTest.Jit.targets
TIHan: That's true. Only thing I can think of is detecting if the assembly was marked as optimized or not and go based on that.

src/tests/Common/CLRTest.Jit.targets
markples: Yes, that was done intentionally.  They wouldn't have been skipped before either.  So you can keep using this instead of `assembliesToIgnore` if any are broken and for any future regressions that need disabling.

src/tests/Common/CLRTest.Jit.targets
markples: Per https://wiki.osdev.org/PE#Overview, fields in PE files are little-endian.  Per https://docs.python.org/3/library/struct.html, `'I'` is the same as `'@I'`, which is "native" byte order.  You are relying on the machine being little-endian (which in practice today is accurate).  It looks like the size is also machine dependent.  If you switch to `'<I'`, then you get fixed little-endian, 4 byte size.

src/tests/Common/CLRTest.Jit.targets
markples: Trying to reconstruct everything seems like a losing battle.  We "just" need to figure out how to get the information from msbuild to test-time.

src/tests/Common/CLRTest.Jit.targets
TIHan: I removed the python code and just wrote a C# tool to do the checks. We needed to check to see if the assembly was built in DEBUG anyway so we can use the proper flags for ilasm.

src/tests/Common/CLRTest.Jit.targets
TIHan: I bit the bullet and just wrote a tool to do the checks instead of the python code.

src/tests/Common/CLRTest.Jit.targets
TIHan: It only makes it a little bit harder to maintain, but yea I would prefer if we only had one way. I just do not want to spend more time on this than I already have.

TIHan:
@dotnet/jit-contrib @BruceForstall @markples this is ready

src/coreclr/tools/AssemblyChecker/Program.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            var attrs = reader.GetAssemblyDefinition().GetCustomAttributes();
+            return attrs.Any(x => HasDebuggableAttribute(reader, reader.GetCustomAttribute(x).Constructor));
+        }
+
+        static int Main(string[] args)
+        {
+            if (args.Length == 0)
+            {
+                throw new ArgumentException("Expected assembly file-path.");
```
BruceForstall, line 82: Should this throw, or print an error message and return a non-zero exit code? (same below)

src/coreclr/tools/AssemblyChecker/Program.cs
```diff
@@ -0,0 +1,116 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
```
BruceForstall, line 3: I'd like to see a header comment here that explains what this tool is, and how it is used.

src/coreclr/tools/AssemblyChecker/AssemblyChecker.csproj
```diff
@@ -0,0 +1,58 @@
...
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'linux-musl-arm64'">$(runtimelinuxmuslarm64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'linux-musl-x64'">$(runtimelinuxmuslx64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'win-arm64'">$(runtimewinarm64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'win-x64'">$(runtimewinx64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'osx-arm64'">$(runtimeosxarm64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+    <JITToolsVersion Condition="'$(JITToolsVersion)' == '' and '$(JITToolsRid)' == 'osx-x64'">$(runtimeosxx64MicrosoftNETCoreRuntimeJITToolsVersion)</JITToolsVersion>
+  </PropertyGroup>
```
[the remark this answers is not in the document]
BruceForstall: Do we really need to specify every rid in this file? That doesn't seem scalable (e.g., when new rids are added)

src/coreclr/tools/AssemblyChecker/Program.cs
TIHan: I guess it should not throw. It should just print an error and return a non-zero exit code.

src/coreclr/tools/AssemblyChecker/Program.cs
TIHan: Sure, I can add it.

src/coreclr/tools/AssemblyChecker/AssemblyChecker.csproj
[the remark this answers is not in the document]
TIHan: Yea none of this is necessary. I copied a csproj from SuperFileCheck and I kept everything in it when I didn't need to.

TIHan:
Running new pipeline https://dev.azure.com/dnceng-public/public/_build/results?buildId=441538&view=results

TIHan:
The pipeline passed. I double checked the outputs and it seems that `python` is not installed on the linux or OSX helix machines.

markples:
> The pipeline passed. I double checked the outputs and it seems that `python` is not installed on the linux or OSX helix machines.

It's there.  You can see later in the script calling it as `python3`.  run.cmd/sh have code that find it.  However, if you depend on that, then the tests won't be executable directly (that is, by calling `<test>.cmd` rather than having `run.cmd` call it).

With `AssemblyChecker` extracted, I'm not sure you're getting much value out of python anymore anyway.  You could probably unify in the .cmd/.sh generation (just a loop + call to AssemblyChecker) and avoid the python dependency now, and that would be a different way to get rid of the duplicate logic.

src/tests/Common/CLRTest.Jit.targets
```diff
@@ -63,6 +63,15 @@ export RunningIlasmRoundTrip=
 if [ -z "$DoLink" -a ! -z "$RunningIlasmRoundTrip" ];
 then
   mkdir IL-RT
+  if [ "$(IsMergedTestRunnerAssembly)" = "true" ]
+  then
+    python3 $(AssemblyName)_ilasmroundtrip.py
+    if [ $ERRORLEVEL -ne 0 ]
```
markples, line 69: The lab script didn't fail because the value is in `$?`, not `$ERRORLEVEL`.  Other parts of the script do `ERRORLEVEL=$?` because they want to save the value for another use.

src/tests/Common/CLRTest.Jit.targets
TIHan: It's hard to get it working because CI is so slow to verify.

I'll do the work to have only one impl.

TIHan:
@dotnet/jit-contrib This is ready again, see results: https://dev.azure.com/dnceng-public/public/_build/results?buildId=444219&view=results

src/coreclr/tools/AssemblyChecker/Program.cs
markples: If you put this in a `void PrintUsage()` method (still near the top of the file) and call it before all of the places where the tool returns 2, then you get the documentation at the top and command line help.

For "You can also ask...", add the argument so that the user doesn't have to look at the source code to figure out how.

src/tests/Common/CLRTest.Jit.targets
```diff
@@ -55,60 +55,22 @@ export RunningIlasmRoundTrip=
 ]]>
       </IlasmRoundTripBashScript>
 
-      <IlasmRoundTripBashScript Condition="'$(CLRTestKind)' == 'BuildAndRun'"><![CDATA[
+      <IlasmRoundTripBashScript Condition="'$(CLRTestKind)' == 'BuildAndRun' And '$(IlasmRoundTripIncompatible)' != 'true'"><![CDATA[
 # IlasmRoundTrip Script
-# Disable Ilasm round-tripping for Linker tests.
-# Todo: Ilasm round-trip on linked binaries.
```
markples, line 61: Why are you removing these comments?  Line 61 is still doing something with `DoLink`.

src/tests/Common/CLRTest.Jit.targets
```diff
@@ -55,60 +55,22 @@ export RunningIlasmRoundTrip=
 ]]>
       </IlasmRoundTripBashScript>
 
-      <IlasmRoundTripBashScript Condition="'$(CLRTestKind)' == 'BuildAndRun'"><![CDATA[
+      <IlasmRoundTripBashScript Condition="'$(CLRTestKind)' == 'BuildAndRun' And '$(IlasmRoundTripIncompatible)' != 'true'"><![CDATA[
```
markples, line 58: It is surprising that you would need this since it wasn't needed before.  The script should exit at the beginning if the test is incompatible.  (Leaving the logic in allows for a quick manual script modification in order to run such a test when trying to get it to work rather than rebuilding, though that is a fairly niche scenario.)

src/tests/Common/CLRTest.Jit.targets
TIHan: Didn't think they were needed. I'll add them back.

src/tests/Common/CLRTest.Jit.targets
TIHan: This was needed to fix a few tests that were marked 'incompatible' but still would run the '.cmd' or '.sh' files with RunningIlasmroundtrip=1. It doesn't hurt to have this.

markples: (This comment applies to several places.)

src/tests/Common/CLRTest.Jit.targets
```diff
@@ -55,60 +55,22 @@ export RunningIlasmRoundTrip=
...
-  if [ $ERRORLEVEL -ne 0 ]
-  then
+    echo ILASM ROUND-TRIP - FAILED $ERRORLEVEL
     exit 1
   fi
 fi
 ]]>
       </IlasmRoundTripBashScript>
-      <BashIlrtTestLaunchCmds Condition="'$(CLRTestKind)' == 'BuildAndRun'"><![CDATA[
+      <BashIlrtTestLaunchCmds Condition="'$(CLRTestKind)' == 'BuildAndRun' And '$(IlasmRoundTripIncompatible)' != 'true'"><![CDATA[
```
markples, line 73: This comment is more about the original IL-RT change (and any change shouldn't be part of this PR).

The ilasm roundtrip scripts already had a somewhat peculiar structure where they execute the original binary -and- the roundtripped binary.  Presumably this was to detect the case where the input binary was already broken, though any normal test run would already catch that.  The downside was that roundtrip testing had the cost of executing the tests twice.  However, now that the original binary is overwritten, I think it is just running the roundtripped binary twice.  On top of that, I think it will run the merged test wrapper script twice (for all tests) and then individual RequiresProcessIsolation tests twice within each cmd/sh invocation, so they might be called four times.

src/tests/JIT/Methodical/fp/exgen/10w5d_cs_do.csproj
```diff
@@ -3,6 +3,7 @@
     <!-- Needed for JIT/superpmi/superpmicollect.csproj which depends on the
          individual test .cmd file -->
     <RequiresProcessIsolation>true</RequiresProcessIsolation>
+    <IlasmRoundTripIncompatible>true</IlasmRoundTripIncompatible>
```
markples, line 6: Convention is to include reason(s) in the RequiresProcessIsolation comment, for which IlasmRoundTripIncompatible is a new one.

Are these due to ilasm/ildasm bugs that should have issues?  Or something else?

src/tests/Common/CLRTest.Jit.targets
TIHan: It's only new because of how the merged tests are structured. The actual tests here don't cause a problem.

markples: What does that mean?  What is the behavior with `IlasmRoundTripIncompatible` that is new with merged tests?  How does the script get to this check if the earlier part of the script exits?

TIHan: I don't remember since it was in the middle of the week. Give me a bit to reproduce what the issue was.

src/tests/Common/CLRTest.Jit.targets
TIHan: I would be surprised if it ran the tests twice. I think how the tests get run is that it finds all `.cmd`/`.sh` files and runs them.

src/tests/Common/CLRTest.Jit.targets
TIHan: Ok, I remember now.

The projects `10w3d_cs_do.csproj`, `Bytemark.csproj`, and `CscBench.csproj` are referenced by `superpmicollect.csproj`.

When they are referenced, the binaries for `10w3d_cs_do.csproj`, `Bytemark.csproj`, and `CscBench.csproj` are included in the output for `superpmicollect` in their own respective folders:
image

In those folders, they include a `.cmd`/`.sh`:
image

Is this necessarily a problem? No, but if we run the `.cmd` with roundtripping, it will not be able to find the corresponding `roundtrip.py` file because it's not emitted in this folder. The `.py` file is only emitted in the original location:
image

I could certainly try to figure out how to either not emit the `.cmd` in the subfolder location for `superpmicollect` or copy the `roundtrip.py` file over. But in the interest of time, it was easier to just disable these 3 tests.

markples: This makes sense, though I think it is about marking those 3 as incompatible.  I don't see how it connects to needing to change the generated script.

It looks like superpmicollect.csproj is just copying every target output (@(_SpmiTest), set by the OutputItemType in the ProjectReference.  An eventual fix might just be to then remove the .cmd/.sh items from that list.

src/tests/Common/CLRTest.Jit.targets
markples: I checked the logs.  For example, in [this log](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-93368-merge-13a9ee7cff6f474f9b/Regression_2/1/console.9dbafa52.log?helixlogtype=result), the line

```
"C:\h\w\BADE09EA\p\corerun.exe" -p "System.Reflection.Metadata.MetadataUpdater.IsSupported=false" -p "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization=true"  Runtime_91839.dll
```

appears 4 times in two pairs.  It doesn't matter how the .cmd are found.  The wrapper .cmd itself runs the whole set of tests twice, and the individual test .cmd runs its test twice.

src/tests/Common/CLRTest.Jit.targets
TIHan: I didn't see that, but that seems easy to do.

TIHan: Unfortunately, it's not as simple as I thought. There is no way to determine if the tests are built for ilasmroundtrip. We only know this when we run the actual test.

src/tests/Common/CLRTest.Jit.targets
TIHan: Why do they run multiple times? Is this intended behavior?

markples: My first comment has what I know.  It looks like it was originally intended, but it's not useful now that the two runs are the same.  I don't think it's worth trying to recreate the original behavior.  Instead I think you can just delete the execution step in the ilasm code and rely on the "normal" execution.

TIHan: Ah I understand now. You are saying we could just delete the extra execution step since we already overwrite it the original DLL.

TIHan: Yea, this would be a good change to do.

TIHan:
@markples this is ready again. Results: https://dev.azure.com/dnceng-public/public/_build/results?buildId=449897&view=results

Definitely removing the execution test made it go a lot faster.

src/coreclr/tools/AssemblyChecker/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+            }
+
+            return peReader.PEHeaders.IsExe;
+        }
+
+        static int Main(string[] args)
+        {
+            if (args.Length == 0)
+            {
+                Console.Error.WriteLine("Expected assembly file-path.");
```
markples, line 88: I recommend WriteLine(HelpText) here - no need to make people figure out -h (or --help or -? since tools are different)

src/coreclr/tools/AssemblyChecker/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+                    default:
+                        {
+                            Console.Error.WriteLine("Invalid option.");
+                            return 2;
+                        }
+                }
+            }
+
+            Console.Error.WriteLine("Too many arguments.");
+            return 2;
```
markples, line 136: HelpText on these too

src/coreclr/tools/AssemblyChecker/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+                        {
+                            return debuggableAttribute.IsJITOptimizerDisabled;
+                        }
+                        return false;
+                    });
+            }
+            else
+            {
+                return false;
+            }
```
markples, line 66: I believe the length check is superfluous here (made me wonder if empty is a special case compared to having elements none of which satisfy the `Any`.

I think the whole thing is the same as `attribs.OfType<DebuggableAttribute>.Any(attr => attr.IsJITOptimizerDisabled)`

src/coreclr/tools/AssemblyChecker/Program.cs
```diff
@@ -0,0 +1,139 @@
...
+                Console.WriteLine(HelpText);
+                return 0;
+            }
+
+            if (args.Length == 1)
+            {
+                if (IsAssembly(args[0]))
+                    return 0;
+                else
+                    return 1;
```
markples, line 104: This is purely style - you can fully choose - but I find these multiline things kind of distracting over `return IsAssembly(args[0]) ? 0 : 1` or return Convert.ToInt32(!IsAssembly(args[0])).

markples (approved):

markples:
I'm happy seeing some of that old code go away.  Thanks for the changes.

TIHan:
Failures are unrelated:
> packages/microsoft.dotnet.helix.sdk/9.0.0-beta.23528.1/tools/Microsoft.DotNet.Helix.Sdk.MultiQueue.targets(89,5): error : (NETCORE_ENGINEERING_TELEMETRY=Test) Work item System.Net.Quic.Functional.Tests in job a618718d-2811-4c62-894a-3062e12a0589 has failed.
Failure log: https://helix.dot.net/api/2019-06-17/jobs/a618718d-2811-4c62-894a-3062e12a0589/workitems/System.Net.Quic.Functional.Tests/console

> .packages\microsoft.dotnet.helix.sdk\9.0.0-beta.23528.1\tools\azure-pipelines\AzurePipelines.MultiQueue.targets(42,5): error : (NETCORE_ENGINEERING_TELEMETRY=Test) Test System.Net.Http.Functional.Tests.SocketsHttpHandler_HttpClientHandler_ConnectionPooling_Test.Http2_SmallConnectionTimeout_SubsequentRequestUsesDifferentConnection has failed. Check the Test tab or this console log: https://helix.dot.net/api/2019-06-17/jobs/497709c4-3df8-43c9-b12b-3832861e0976/workitems/System.Net.Http.Functional.Tests/console

merging

PR #93369, opened 2023-10-11
Vector512 Support for Enumerable<int>.Min/Max

Spacefish:
Use the new Vector512 type introduced in NET 8, when
- enough elements
- hardware support is present

to calculate Enumerable<int>.Min() and Enumerable<int>.Max()

### DIFF
--- a/src/libraries/System.Linq/src/System/Linq/Max.cs
+++ b/src/libraries/System.Linq/src/System/Linq/Max.cs
@@ -18,9 +18,7 @@ namespace System.Linq
             public static bool Compare(T left, T right) => left > right;
             public static Vector128<T> Compare(Vector128<T> left, Vector128<T> right) => Vector128.Max(left, right);
             public static Vector256<T> Compare(Vector256<T> left, Vector256<T> right) => Vector256.Max(left, right);
-#if NET8_0_OR_GREATER
             public static Vector512<T> Compare(Vector512<T> left, Vector512<T> right) => Vector512.Max(left, right);
-#endif
         }
 
         public static int? Max(this IEnumerable<int?> source) => MaxInteger(source);
--- a/src/libraries/System.Linq/src/System/Linq/Max.cs
+++ b/src/libraries/System.Linq/src/System/Linq/Max.cs
@@ -18,6 +18,9 @@ namespace System.Linq
             public static bool Compare(T left, T right) => left > right;
             public static Vector128<T> Compare(Vector128<T> left, Vector128<T> right) => Vector128.Max(left, right);
             public static Vector256<T> Compare(Vector256<T> left, Vector256<T> right) => Vector256.Max(left, right);
+#if NET8_0_OR_GREATER
+            public static Vector512<T> Compare(Vector512<T> left, Vector512<T> right) => Vector512.Max(left, right);
+#endif
         }
 
         public static int? Max(this IEnumerable<int?> source) => MaxInteger(source);
--- a/src/libraries/System.Linq/src/System/Linq/MaxMin.cs
+++ b/src/libraries/System.Linq/src/System/Linq/MaxMin.cs
@@ -122,7 +122,7 @@ namespace System.Linq
                         }
                     }
                 }
-#endif                
+#endif
             }
             else
             {
--- a/src/libraries/System.Linq/src/System/Linq/MaxMin.cs
+++ b/src/libraries/System.Linq/src/System/Linq/MaxMin.cs
@@ -16,9 +16,7 @@ namespace System.Linq
             public static abstract bool Compare(T left, T right);
             public static abstract Vector128<T> Compare(Vector128<T> left, Vector128<T> right);
             public static abstract Vector256<T> Compare(Vector256<T> left, Vector256<T> right);
-#if NET8_0_OR_GREATER
             public static abstract Vector512<T> Compare(Vector512<T> left, Vector512<T> right);
-#endif
         }
 
         private static T MinMaxInteger<T, TMinMax>(this IEnumerable<T> source)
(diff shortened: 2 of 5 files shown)


src/libraries/System.Linq/src/System/Linq/Max.cs
```diff
@@ -18,6 +18,9 @@ public static partial class Enumerable
             public static bool Compare(T left, T right) => left > right;
             public static Vector128<T> Compare(Vector128<T> left, Vector128<T> right) => Vector128.Max(left, right);
             public static Vector256<T> Compare(Vector256<T> left, Vector256<T> right) => Vector256.Max(left, right);
+#if NET8_0_OR_GREATER
```
MihaZupan, line 21: You shouldn't need the `#if`s since `System.Linq` is only using the latest TFMs.

Spacefish: Oh yes, you are right.. Removed them. Didn´t knew this, but it makes sense.
Fixed

src/libraries/System.Linq/src/System/Linq/MaxMin.cs
```diff
@@ -90,6 +91,30 @@ private interface IMinMaxCalc<T> where T : struct, IBinaryInteger<T>
...
+                    best = TMinMax.Compare(best, Vector512.LoadUnsafe(ref lastVectorStart));
+
+                    value = best[0];
+                    for (int i = 1; i < Vector512<T>.Count; i++)
+                    {
+                        if (TMinMax.Compare(best[i], value))
+                        {
+                            value = best[i];
+                        }
+                    }
```
gfoidl, line 116: Not for this PR, but I think we should revisit the conditions when the vectorized code-paths are entered and how the reduction to the scalar value is performed.

For e.g. `int` it's 16 iterations of that loop, and when (by accident) the input span is 16 ints long, then the whole vectorized min/max is pointless.

Maybe special case some common types like `int`, where (at least with Sse4.1) a vectorized Min version exists, so the Vec512 could be split into 4x Vec128 and continue vectorized (min + shuffling around). Or isn't that it worth for Linq and / or not worth the trouble for the specialization for some types?

Spacefish: I agree.. Probably the vectorized code path should only be entered if the number of elements is two or 4 time the vector length. But the additional checks would introduce additional overhead as well.

Regarding the Min/Max over the vector in the last code part: This could probably be optimized as well via vectorization for the `Vector512` and `Vector256` cases, by loading the Vector512 `best` Vector into two Vector256 doing the Min/Max there via SIMD and same with Vector256 -> Vector128..

Not sure how large the overhead on the actual target platforms is to do this, as we would switch between AVX512 to AVX256 (probably cheap) and then to SSE 4.2... If i remember correctly switching from SSE to AVX is quite expensive on some older Intel platforms.

gfoidl: > switching from SSE to AVX is quite expensive on some older Intel platforms

As you said "on older" ones, and `vzeroupper` can help here.

Maybe it's best to just keep the code as is (in your PR) now, and later overhaul the codebase once `ISimdVector` et. al. has proper helpers for reduction, etc. I remember there was a discussion in one issue / PR about such things, but can't find it ATM.

eiriktsarpalis (commented):
Would it be possible to increase the maximum length for Min and Max test data so that the Vector512 code path is better tested?

https://github.com/stephentoub/runtime/blob/710657575ffbc7de90c345239dc92ce52a9a2558/src/libraries/System.Linq/tests/MaxTests.cs#L14

https://github.com/stephentoub/runtime/blob/710657575ffbc7de90c345239dc92ce52a9a2558/src/libraries/System.Linq/tests/MinTests.cs#L14

Spacefish:
@eiriktsarpalis i increased them to 65, so we are at least 1 element wider even for byte than the Vector512.

src/libraries/System.Linq/tests/MaxTests.cs
```diff
@@ -11,13 +11,16 @@ public class MaxTests : EnumerableTests
...
-            for (int length = 2; length < 33; length++)
+            for (int length = 2; length < 65; length++)
             {
                 yield return new object[] { Shuffler.Shuffle(Enumerable.Range(length, length).Select(i => (byte)i)), (byte)(length + length - 1) };
                 yield return new object[] { Shuffler.Shuffle(Enumerable.Range(length, length).Select(i => (byte)i).ToArray()), (byte)(length + length - 1) };
 
-                yield return new object[] { Shuffler.Shuffle(Enumerable.Range(length, length).Select(i => (sbyte)i)), (sbyte)(length + length - 1) };
-                yield return new object[] { Shuffler.Shuffle(Enumerable.Range(length, length).Select(i => (sbyte)i).ToArray()), (sbyte)(length + length - 1) };
+                // Unit Tests does +T.One so we should generate data up to one value below sbyte.MaxValue
+                if ((length + length) < sbyte.MaxValue) {
```
eiriktsarpalis, line 20: This predates your own changes, but curious why this uses `length + length` instead of `2 * length`.

eiriktsarpalis (approved):

src/libraries/System.Linq/tests/MaxTests.cs
Spacefish: I used (length + length) as this is how it´s done during test data generation, so it uses the same style to check against this. It´s easier to spot like that / mentally link it to the generated test data i figured.

BruceForstall:
cc @dotnet/avx512-contrib

PR #93371, opened 2023-10-11
JIT: defer creating throw helper blocks until simple lower

AndyAyersMS:
Defer creating throw helper blocks until just before the late flow graph optimizations. This removes some of the flow graph updates from morph, with the aim of making it easier to show that an RPO traversal in morph is safe.

Unfortunately, it's hard to get these blocks in the exact same places they were before, so there are quite a few diffs. They are generally improvements. I also saw some cases where the early creation of throw helpers in try regions inhibited loop recognition; this no longer happens, and may enable loop cloning (and hence size increses).

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=441230&view=ms.vss-build-web.run-extensions-tab)

This also lays the groundwork for removing simple lowering.


AndyAyersMS:
cc @dotnet/jit-contrib 

Code size improvements here are somewhat surprising.... highlighting the need for a global layout algorithm, as there are multiple considerations to determine optimal placements of blocks.

AndyAyersMS:
SPMI replay errors look like #93527

EgorBo:
Nice wins! I presume it should fix various cases where we have unreachable throw helper blocks, e,g. I've just came up with:
```cs
void Test(bool cond)
{
    int b = 10;
    if (cond)
    {
        _ = checked((byte)b);
    }
}
```
Emits:
```asm
G_M12030_IG01:  ;; offset=0x0000
       sub      rsp, 40
						;; size=4 bbWeight=1 PerfScore 0.25
G_M12030_IG02:  ;; offset=0x0004
       add      rsp, 40
       ret      
						;; size=5 bbWeight=1 PerfScore 1.25
G_M12030_IG03:  ;; offset=0x0009
       call     CORINFO_HELP_OVERFLOW
       int3     
						;; size=6 bbWeight=0 PerfScore 0.00

; Total bytes of code 15
```

EgorBo (approved):

AndyAyersMS:
> presume it should fix various cases where we have unreachable throw helper blocks

Actually, no... let me look into this as a follow-up.

AndyAyersMS:
> > presume it should fix various cases where we have unreachable throw helper blocks
> 
> Actually, no... let me look into this as a follow-up.

Unfortunately, this seems tricky. We don't know if we really need a throw helper until after codegen, and by then it's not so easy to get rid of things. I suppose I can at least see how often a helper ends up being unused, if it turns out this is a common thing, we can find a way to fix it.

jakobbotsch:
It looks like this PR had a ton of misses, so the diffs may be even greater (depending on the reason for the misses I suppose).

AndyAyersMS:
> It looks like this PR had a ton of misses, so the diffs may be even greater (depending on the reason for the misses I suppose).

It could be a bad sign, like the JIT is asking for throw helpers now that it didn't ask for before. Maybe once we have new collections we can do a trial revert and see how that looks.

EgorBo:
Yes, something is off

image

AndyAyersMS:
PMI diffs show some new unnecessary throw helper blocks, eg

image

So let me dig into this and see how we can avoid adding such things.

PR #93377, opened 2023-10-12
Explicitly represent BBJ_EHFINALLYRET successors

### LINKED ISSUE #84278: Flow graph successors representation for BBJ_EHFINALLYRET
The `try` of a `try / finally` clause can contain a set of `leave` IL specifying to where the code should continue after the `finally` is invoked. The JIT transforms this into IR like:

```
// try
 ... code ...
BBJ_CALLFINALLY / bbJumpDest = the finally funclet
BBJ_ALWAYS / bbJumpDest = the "continuation" address specified by the `leave`
...
// finally
 ... code ...
BBJ_EHFINALLYRET
```

Note that there can be multiple `leave` IL that require invoking the same finally, and thus multiple `BBJ_CALLFINALLY` targeting the same finally. Each of the corresponding `BBJ_ALWAYS` (continuation) targets can be different, however.

Note also that if the finally is known to never return (due to an unconditional `throw` or infinite loop), the `BBJ_CALLFINALLY` will not have a `BBJ_ALWAYS` and the finally will not have a `BBJ_EHFINALLYRET`.

The "flowgraph successors" of the `BBJ_EHFINALLYRET` block are all the `BBJ_ALWAYS` blocks corresponding to the `BBJ_CALLFINALLY` blocks that target the finally.

Currently, the JIT does not directly represent the BasicBlock successors of a `BBJ_EHFINALLYRET` block. Finding those successors, as is done with the versions of `GetSucc/NumSucc` that take a `Compiler*`, is very expensive.

**Can we directly represent the `BBJ_EHFINALLYRET` successors in the IR? E.g., as an array of `BasicBlock*`, the same way `BBJ_SWITCH` successors are implemented?**

This would make forward `GetSucc/NumSucc` flowgraph traversals much cheaper.

Note that the IR is asymmetric here: the `BBJ_ALWAYS` pair of `BBJ_CALLFINALLY` has the `BBJ_EHFINALLYRET` in its predecessor list.

@AndyAyersMS @dotnet/jit-contrib Comments?

BruceForstall:
Currently, BBJ_EHFINALLYRET blocks have no explicit successors in the IR. To implement successor iteration, a very expensive process is followed to (1) find the region of blocks where a BBJ_CALLFINALLY block calling the `finally` might be found, (2) search the region for such blocks, and (3) return as a successor all the BBJ_ALWAYS blocks in the corresponding BBJ_CALLFINALLY/BBJ_ALWAYS pair.

Change the IR to explicitly represent and maintain this list of successors for BBJ_EHFINALLYRET blocks. The representation is a simple array of `BasicBlock*`, similar to how BBJ_SWITCH block targets are represented.

Fixes #84278

Notes:
1. The BBJ_EHFINALLYRET successors are computed in `impFixPredLists()`. There are various dumpers that run before this, so we need to tolerate incomplete successor information in some places.
2. `ehGetCallFinallyBlockRange()` is still used by some code. I changed the semantics to return a `[first..last]` range inclusive of `last` instead of the previous `[beginning..end)` range exclusive of `end`. This makes it easier to use with our BasicBlock iterators.

### DIFF
--- a/src/coreclr/jit/fgbasic.cpp
+++ b/src/coreclr/jit/fgbasic.cpp
@@ -611,50 +611,43 @@ void Compiler::fgReplaceEhfSuccessor(BasicBlock* block, BasicBlock* oldSucc, Bas
     assert(block->KindIs(BBJ_EHFINALLYRET));
     assert(fgPredsComputed);
 
-    BBehfDesc* const ehfDesc   = block->GetJumpEhf();
-    unsigned         succCount = ehfDesc->bbeCount;
-    BasicBlock**     succTab   = ehfDesc->bbeSuccs;
-
-    unsigned i = 0;
+    BBehfDesc* const   ehfDesc   = block->GetJumpEhf();
+    const unsigned     succCount = ehfDesc->bbeCount;
+    BasicBlock** const succTab   = ehfDesc->bbeSuccs;
 
     // Walk the successor table looking for blocks to update the preds for
-    while (i < succCount)
+    bool      found   = false;
+    FlowEdge* newEdge = nullptr;
+    for (unsigned i = 0; i < succCount; i++)
     {
-        if (succTab[i] == oldSucc) // We will update when succTab[i] matches
+        if (succTab[i] != oldSucc)
+        {
+            continue;
+        }
+
+        found = true;
+
+        // Change the succTab entry to branch to the new location
+        //
+        succTab[i] = newSucc;
+
+        if (newEdge == nullptr)
         {
             // Remove the old edge [block => oldSucc]
             //
             fgRemoveAllRefPreds(oldSucc, block);
 
-            // Change the succTab entry to branch to the new location
-            //
-            succTab[i] = newSucc;
-
             // Create the new edge [block => newSucc]
             //
-            FlowEdge* const newEdge = fgAddRefPred(newSucc, block);
-
-            // Now set the correct value of newEdge's DupCount
-            // and replace any other successors in succTab[] that go to oldSucc.
-            //
-            i++;
-            while (i < succCount)
-            {
-                if (succTab[i] == oldSucc)
-                {
-                    // We also must update this entry in the succTab
-                    //
-                    succTab[i] = newSucc;
-                    newSucc->bbRefs++;
-                    newEdge->incrementDupCount();
-                }
-                i++; // Check the next entry in succTab[]
-            }
-            return;
+            newEdge = fgAddRefPred(newSucc, block);
         }
-        i++; // Check the next entry in succTab[] for a match
+        else
+        {
+            newSucc->bbRefs++;
+            newEdge->incrementDupCount();
+        }
     }
-    noway_assert(!"Did not find oldTarget in succTab[]");
+    noway_assert(found && "Did not find oldSucc in succTab[]");
 }
 
 //------------------------------------------------------------------------
@@ -697,6 +690,7 @@ void Compiler::fgRemoveEhfSuccessor(BasicBlock* block, BasicBlock* succ)
             --succCount;
 
             found = true;
+
 #ifdef DEBUG
             // We only expect to see a successor once in the table.
             for (; i < succCount; i++)
@@ -704,6 +698,8 @@ void Compiler::fgRemoveEhfSuccessor(BasicBlock* block, BasicBlock* succ)
                 assert(succTab[i] != succ);
             }
 #endif // DEBUG
+
+            break;
         }
     }
     assert(found);
--- a/src/coreclr/jit/fgbasic.cpp
+++ b/src/coreclr/jit/fgbasic.cpp
@@ -382,6 +382,55 @@ void Compiler::fgRemoveReturnBlock(BasicBlock* block)
     }
 }
 
+//------------------------------------------------------------------------
+// fgConvertBBToThrowBB: Change a given block to a throw block.
+//
+// Arguments:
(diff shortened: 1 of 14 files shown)


BruceForstall:
[No diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=435455&view=ms.vss-build-web.run-extensions-tab), generally a slight TP improvement, in some cases significant.

src/coreclr/jit/fgbasic.cpp
```diff
@@ -516,6 +562,155 @@ void Compiler::fgReplaceSwitchJumpTarget(BasicBlock* blockSwitch, BasicBlock* ne
...
+
+            // Change the succTab entry to branch to the new location
+            //
+            succTab[i] = newSucc;
+
+            // Create the new edge [block => newSucc]
+            //
+            FlowEdge* const newEdge = fgAddRefPred(newSucc, block);
+
+            // Now set the correct value of newEdge's DupCount
```
AndyAyersMS, line 637: Seems like this could be a bit simpler if you remember that you'd already created a flow edge, something like

```c++
FlowEdge* edge = nullptr;
for (i = 0; i < succCount; i++)
{
    if (succTab[i] != oldSucc) continue;

    succTab[i] = newSucc;

    if (flowEdge == nullptr)
    {
       flowEdge = fgAddRefPred(newSucc, block);
    }
    else
    {
       flowEdge->incrementDupCount();
    }
}
```

src/coreclr/jit/fgbasic.cpp
```diff
@@ -516,6 +562,155 @@ void Compiler::fgReplaceSwitchJumpTarget(BasicBlock* blockSwitch, BasicBlock* ne
...
+            --succCount;
+
+            found = true;
+#ifdef DEBUG
+            // We only expect to see a successor once in the table.
+            for (; i < succCount; i++)
+            {
+                assert(succTab[i] != succ);
+            }
+#endif // DEBUG
```
AndyAyersMS, line 706: Break out of the loop here if not debug?

src/coreclr/jit/fgbasic.cpp
```diff
@@ -516,6 +562,155 @@ void Compiler::fgReplaceSwitchJumpTarget(BasicBlock* blockSwitch, BasicBlock* ne
...
+    unsigned         succCount = ehfDesc->bbeCount;
+    BasicBlock**     succTab   = ehfDesc->bbeSuccs;
+    bool             found     = false;
+
+    // Walk the successor table looking for the specified successor block.
+    for (unsigned i = 0; i < succCount; i++)
+    {
+        if (succTab[i] == succ)
+        {
+            // If it's not the last one, move everything after in the table down one slot.
```
AndyAyersMS, line 690: Maybe `assert(!found)` here instead of the loop below?

src/coreclr/jit/fgdiagnostic.cpp
```diff
@@ -2727,18 +2759,13 @@ bool BBPredsChecker::CheckEHFinallyRet(BasicBlock* blockPred, BasicBlock* block)
...
-    for (BasicBlock* bcall = begBlk; bcall != endBlk; bcall = bcall->Next())
+    for (BasicBlock* const bcall : comp->Blocks(firstBlock, lastBlock))
     {
-        if (!bcall->KindIs(BBJ_CALLFINALLY) || !bcall->HasJumpTo(finBeg))
-        {
-            continue;
-        }
-
-        if (bcall->NextIs(block))
+        if (bcall->KindIs(BBJ_CALLFINALLY) && bcall->HasJumpTo(finBeg) && bcall->NextIs(block))
```
AndyAyersMS, line 2768: From what I can tell this doesn't actually check that `block` is in the ehf successor table for `blockPred` -- should we check that too?

AndyAyersMS (commented):
Looks good overall. Left a few suggestions for possibly simplifying the code.

src/coreclr/jit/fgbasic.cpp
BruceForstall: What I meant to do was have a `break;` after the debug loop, since I don't want to keep iterating in the non-debug case.

src/coreclr/jit/fgbasic.cpp
BruceForstall: I added the `break` but I don't like having the loop iterations of the main loop being debug dependent.

src/coreclr/jit/fgdiagnostic.cpp
BruceForstall: Yes, will do.

BruceForstall:
@AndyAyersMS I addressed your suggestions.

AndyAyersMS (approved):

BruceForstall:
The superpmi-diffs failure is out of disk space.
The test failures are Quic (known) and 4 win-arm64 failures. Oddly, the exact same 4 failures were reported by a release/7.0 PR today: https://github.com/dotnet/runtime/issues/93424. I can't repro the failures locally either with my change, without it, or with a fresh build of HEAD of runtime/main.

BruceForstall:
Looks like @AndyAyersMS had a PR with the same win-arm64 failures: https://github.com/dotnet/runtime/pull/93371

BruceForstall:
> Oddly, the exact same 4 failures were reported by a release/7.0 PR

Maybe the build machines picked up a new MSVC arm64 compiler that has a bug? It does look like the compiler changed recently.

Issue #93354, opened 2023-10-11
Incorrect value in FirstDayOfWeek test


ilonatommy:
## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=432866
Build error leg or test failing: System.Globalization.Tests.DateTimeFormatInfoFirstDayOfWeek.FirstDayOfWeek
Pull request: https://github.com/dotnet/runtime/pull/93224

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json
{
  "ErrorMessage": "Expected: Monday",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

```
info: Expected: Monday
info: Actual:   Sunday
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=432866
**Error message validated:** `Expected: Monday`
**Result validation: :white_check_mark:** Known issue matched with the provided build.
**Validation performed at:** 10/11/2023 5:08:28 PM UTC

### Report

|Build|Definition|Test|Pull Request|
|---|---|---|---|
|[441921](https://dev.azure.com/dnceng-public/public/_build/results?buildId=441921)|dotnet/runtime|[normal-System.Globalization.Hybrid.WASM.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=441921&view=ms.vss-test-web.build-test-results-tab&runId=9851498&resultId=182357)|dotnet/runtime#93368|
|[440108](https://dev.azure.com/dnceng-public/public/_build/results?buildId=440108)|dotnet/runtime|[normal-System.Globalization.Hybrid.WASM.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=440108&view=ms.vss-test-web.build-test-results-tab&runId=9805516&resultId=172615)|dotnet/runtime#93409|
[...]

PR #93380, opened 2023-10-12
[browser][HybridGlobalization] Improve logging on FirstDayOfWeek failures

ilonatommy:
I cannot reproduce the [failure](https://github.com/dotnet/runtime/issues/93354) from CI. We are testing hundreds of locales in the same test and logging only:
```
info: Expected: Monday
info: Actual:   Sunday
```
on failure. Xunit's Assert.Equal does not have a `message` argument overload, so it was changed to `Assert.True` that has it. Now we will get e.g. `Failed for culture: pl-PL. Expected: Monday, Actual: Sunday`.
In order to change only hybrid-focused test cases the non-hybrid cases were moved to a separate test.


ilonatommy:
It turns out the failure is not on HybridGlobalization-specific test but on the general one. However, this fix will be helpful for the future anyway.

src/libraries/System.Globalization/tests/DateTimeFormatInfo/DateTimeFormatInfoFirstDayOfWeek.cs
```diff
@@ -10,209 +10,215 @@ public class DateTimeFormatInfoFirstDayOfWeek
...
+            yield return new object[] { "zh-HK", DayOfWeek.Sunday };
+            yield return new object[] { "zh-TW", DayOfWeek.Sunday };
         }
 
         [Theory]
         [MemberData(nameof(FirstDayOfWeek_Get_TestData))]
-        public void FirstDayOfWeek(DateTimeFormatInfo format, DayOfWeek expected)
+        public void FirstDayOfWeek(DateTimeFormatInfo format, DayOfWeek expected, string cultureName)
+        {
+            Assert.True(expected == format.FirstDayOfWeek, $"Failed for culture: {cultureName}. Expected: {expected}, Actual: {format.FirstDayOfWeek}");
```
maraf, line 213: nit: Can we call `FirstDayOfWeek` from `FirstDayOfWeekHybridGlobalization` (share the assert "logic" and message)?

maraf (approved):

src/libraries/System.Globalization/tests/DateTimeFormatInfo/DateTimeFormatInfoFirstDayOfWeek.cs
ilonatommy: good idea

PR #93405, opened 2023-10-12
Attribute HttpListenerRequest.UserAgent as nullable

### LINKED ISSUE #93033: HttpListenerRequest.UserAgent should be attributed as nullable
### Description

[HttpListenerRequest.UserAgent](https://github.com/dotnet/runtime/blob/a166bb71dea362129a25822852c23dbea1d802ed/src/libraries/System.Net.HttpListener/src/System/Net/HttpListenerRequest.cs#L205C77-L205C77) has a return type of `string` but it returns null if the request does not have a `User-Agent` header. Other public properties, such as [ContentType](https://github.com/dotnet/runtime/blob/a166bb71dea362129a25822852c23dbea1d802ed/src/libraries/System.Net.HttpListener/src/System/Net/HttpListenerRequest.cs#L109) and [UrlReferrer](https://github.com/dotnet/runtime/blob/a166bb71dea362129a25822852c23dbea1d802ed/src/libraries/System.Net.HttpListener/src/System/Net/HttpListenerRequest.cs#L211) are correctly attributed.

UserAgent uses the null-forgiving operator when reading the value:

```csharp
public string UserAgent => Headers[HttpKnownHeaderNames.UserAgent]!;
```

The same file also checks it for null:

```csharp
if (UserAgent != null && CultureInfo.InvariantCulture.CompareInfo.IsPrefix(UserAgent, "UP"))
```

Can UserAgent return `string?` instead of `string`? It matches the behavior both internally (above) and externally (requests missing `User-Agent` header).

It's especially confusing because other properties _are_ attributed.

### Reproduction Steps

Create an HTTP listener and send it a request without a `User-Agent` header. The `UserAgent` property will be null even though the return type is not attributed as such.

### Expected behavior

`UserAgent` property is attributed as nullable because it can easily be null.

### Actual behavior

`UserAgent` property is not attributed as nullable.

### Regression?

N/A. I think it was overlooked during #41189

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

TheMaximum:
Changed the datatype for HttpListenerRequest.UserAgent to be a nullable string to match internally and externally expected behaviour.

Fix #93033

### DIFF
--- a/src/libraries/System.Net.HttpListener/ref/System.Net.HttpListener.cs
+++ b/src/libraries/System.Net.HttpListener/ref/System.Net.HttpListener.cs
@@ -103,7 +103,7 @@ namespace System.Net
         public System.Net.TransportContext TransportContext { get { throw null; } }
         public System.Uri? Url { get { throw null; } }
         public System.Uri? UrlReferrer { get { throw null; } }
-        public string UserAgent { get { throw null; } }
+        public string? UserAgent { get { throw null; } }
         public string UserHostAddress { get { throw null; } }
         public string UserHostName { get { throw null; } }
         public string[]? UserLanguages { get { throw null; } }
--- a/src/libraries/System.Net.HttpListener/src/System/Net/HttpListenerRequest.cs
+++ b/src/libraries/System.Net.HttpListener/src/System/Net/HttpListenerRequest.cs
@@ -202,7 +202,7 @@ namespace System.Net
 
         private string RequestScheme => IsSecureConnection ? UriScheme.Https : UriScheme.Http;
 
-        public string UserAgent => Headers[HttpKnownHeaderNames.UserAgent]!;
+        public string? UserAgent => Headers[HttpKnownHeaderNames.UserAgent];
 
         public string UserHostAddress => LocalEndPoint!.ToString();
 


TheMaximum:
@dotnet-policy-service agree

ManickaP (approved):
LGTM, thank you for the contribution!
Dev inner loop failure is unrelated, this can be merged.

wfurt:
is this breaking change since it is touching `ref` assembly? I don't think the risk is high but I suspect there may be impact for somebody. Any thought @karelz?

ManickaP:
This isn't breaking change as far as I know. cc @stephentoub

stephentoub:
> This isn't breaking change as far as I know. cc @stephentoub

Changing a property that was returning `string` to one that returns `string?` can introduce new warnings, e.g. anyone with NRT enabled and was doing:
```C#
string ua = request.UserAgent;
```
will now get a new warning where they didn't previously, and if they have warnings-as-errors enabled, that will break the build.

So technically it is a breaking change. In the past we've typically just had a single catch-all breaking change notification about such nullability annotation changes. @jeffhandley, is there a process in place for aggregating all of these?

jeffhandley:
With as few of these types of breaking changes as we've had recently, we should just start documenting each of them individually. I recall @gewarren mentioning the aggegate approach also introduced challenges, but I don't recall the specifics.

I'm tagging this issue as needing a breaking change doc issue created. @ManickaP, it would be best for you to file the issue per the steps that the automated comment will provide. [Here's a good reference example](https://github.com/dotnet/docs/issues/33225) for the issue.

ManickaP:
https://github.com/dotnet/docs/issues/37900

gewarren:
We did aggregate (most of) the nullability changes for .NET 6: https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/nullable-ref-type-annotation-changes. I think it makes sense to aggregate them unless they fall under different areas.

Issue #93399, opened 2023-10-12
[trimmer] per-RID linked corlib assembly contains foreign platform intrinsics classes


grendello:
This happens on xamarin-android/main (.NET8), but I suppose it will apply to other platforms as well.

I noticed that `System.Private.CoreLib.dll` output by the linker for arm64 RID contains X86 and WASM intrinsics types:

```
549: System.Runtime.Intrinsics.Wasm.PackedSimd (flist=2253, mlist=6116, flags=0x100081, extends=0xa0)
554: System.Runtime.Intrinsics.X86.X86Base (flist=2253, mlist=6151, flags=0x100081, extends=0xa0)
555: System.Runtime.Intrinsics.X86.Sse (flist=2253, mlist=6151, flags=0x100081, extends=0x8a8)
556: System.Runtime.Intrinsics.X86.Sse2 (flist=2253, mlist=6151, flags=0x100081, extends=0x8ac)
557: System.Runtime.Intrinsics.X86.Sse3 (flist=2253, mlist=6154, flags=0x100081, extends=0x8b0)
558: System.Runtime.Intrinsics.X86.Sse41 (flist=2253, mlist=6154, flags=0x100081, extends=0x8bc)
559: System.Runtime.Intrinsics.X86.Ssse3 (flist=2253, mlist=6155, flags=0x100081, extends=0x8b4)
```

It appears some of the x86 classes are removed, but not all. For comparison, this is a list of types from the same assembly but linked for `x84`:

```
558: System.Runtime.Intrinsics.X86.X86Base (flist=2255, mlist=6330, flags=0x100081, extends=0xa0)
559: System.Runtime.Intrinsics.X86.X86Base/X64 (flist=2255, mlist=6333, flags=0x100082, extends=0xa0)
560: System.Runtime.Intrinsics.X86.Avx (flist=2255, mlist=6336, flags=0x100081, extends=0x90c)
561: System.Runtime.Intrinsics.X86.Avx2 (flist=2255, mlist=6338, flags=0x100081, extends=0x8c0)
562: System.Runtime.Intrinsics.X86.Avx512BW (flist=2255, mlist=6344, flags=0x100081, extends=0x8cc)
563: System.Runtime.Intrinsics.X86.Avx512F (flist=2255, mlist=6346, flags=0x100081, extends=0x8c4)
564: System.Runtime.Intrinsics.X86.Bmi1 (flist=2255, mlist=6347, flags=0x100081, extends=0x8b8)
[...]

PR #37615, opened 2020-06-08
Add ILLink.Substitutions.xml files for System.Private.CoreLib.

eerhardt:
Ported the settings from https://github.com/mono/mono/blob/eaa32d13659f0a6b6b5e62ddb49af68b1f9efb6c/sdks/wasm/src/linker-subs.xml and split them out as appropriate to reduce duplication across the different platform builds.

Contributes to #31785

The only API proposed in #31785 that I didn't cover was `System.Runtime.InteropService.OSPlatform` since there isn't a good way of supporting that in the linker, AFAIK.

For a "Hello, World" wasm app, `System.Private.CoreLib.dll` sizes go from:

| Build          | Size     |
|----------------|----------|
| master         | 1,534 KB |
| PR             | 1,475 KB |
| PR + Invariant Mode | 1,456 KB |

PR #93407, opened 2023-10-12
Add ILLink substitution for wasm intrinsics

akoeplinger:
Noticed those were missing while looking at https://github.com/dotnet/runtime/issues/93399
Also deleted an empty file that was checked in by mistake.

Fixed an issue where we weren't recursively calling IsSupported for PackedSimd which is the pattern we've been using (the JIT/AOT compiler is supposed to use the intrinsic instead)

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Wasm/PackedSimd.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Wasm/PackedSimd.cs
@@ -14,7 +14,7 @@ namespace System.Runtime.Intrinsics.Wasm
     [CLSCompliant(false)]
     public abstract class PackedSimd
     {
-        public static bool IsSupported { [Intrinsic] get { return false; } }
+        public static bool IsSupported { [Intrinsic] get { return IsSupported; } }
 
         // Constructing SIMD Values
 
--- a/src/libraries/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.NoWasmIntrinsics.xml
+++ b/src/libraries/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.NoWasmIntrinsics.xml
@@ -1,0 +1,7 @@
+<linker>
+  <assembly fullname="System.Private.CoreLib">
+    <type fullname="System.Runtime.Intrinsics.Wasm.PackedSimd">
+      <method signature="System.Boolean get_IsSupported()" body="stub" value="false" />
+    </type>
+  </assembly>
+</linker>
--- a/src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
+++ b/src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
@@ -56,6 +56,7 @@
     <ILLinkSubstitutionsXmls Include="$(ILLinkSharedDirectory)ILLink.Substitutions.64bit.xml" Condition="'$(Is64Bit)' == 'true'" />
     <ILLinkSubstitutionsXmls Include="$(ILLinkSharedDirectory)ILLink.Substitutions.NoArmIntrinsics.xml" Condition="'$(SupportsArmIntrinsics)' != 'true'" />
     <ILLinkSubstitutionsXmls Include="$(ILLinkSharedDirectory)ILLink.Substitutions.NoX86Intrinsics.xml" Condition="'$(SupportsX86Intrinsics)' != 'true'" />
+    <ILLinkSubstitutionsXmls Include="$(ILLinkSharedDirectory)ILLink.Substitutions.NoWasmIntrinsics.xml" Condition="'$(SupportsWasmIntrinsics)' != 'true'" />
     <ILLinkSubstitutionsXmls Include="$(ILLinkSharedDirectory)ILLink.Substitutions.OSX.xml" Condition="'$(IsOSXLike)' == 'true'" />
     <ILLinkSubstitutionsXmls Include="$(ILLinkSharedDirectory)ILLink.Substitutions.Windows.xml" Condition="'$(TargetsWindows)' == 'true'" />
     <ILLinkSubstitutionsXmls Include="$(ILLinkSharedDirectory)ILLink.Substitutions.Browser.xml" Condition="'$(TargetsBrowser)' == 'true'" />


akoeplinger:
@tannergooding looks like your recent PR broke the build :)

lewing (approved):

MichalStrehovsky:
I wonder - do these do anything, or are we just cargo culting them since they were introduced in #37615?

The way we build CoreLib source code ensures that `PackedSimd.IsSupported` is hardcoded to return false on non-WASM (we build the PackedSimd.PlatformNotSupported.cs file):

https://github.com/dotnet/runtime/blob/851c274cc50544bf11a59e1011dbc594ca7010b0/src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Wasm/PackedSimd.PlatformNotSupported.cs#L14

In the WASM CoreLib "supported" version of the file is supposed to recursively calls itself like these do for all the other platforms but it also returns false (I think this is a bug):

https://github.com/dotnet/runtime/blob/851c274cc50544bf11a59e1011dbc594ca7010b0/src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Wasm/PackedSimd.cs#L17

ILLink should be able to inline the false value even without these substitutions (@vitek-karas @sbomer to double check me on this).

The reason why we do these as a recursive call on platforms where IsSupported could be true at runtime are twofold:
* Prevent ILLinker inlining the wrong value, like it probably does for PackedSimd on WASM right now
* Make sure reflection-invoking IsSupported does the right thing.

Sample of a correct IsSupported implementation here:

https://github.com/dotnet/runtime/blob/851c274cc50544bf11a59e1011dbc594ca7010b0/src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Arm/AdvSimd.cs#L18

Hardcoded to false on platforms that are not ARM:

https://github.com/dotnet/runtime/blob/851c274cc50544bf11a59e1011dbc594ca7010b0/src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Arm/AdvSimd.PlatformNotSupported.cs#L22

akoeplinger:
> In the WASM CoreLib "supported" version of the file is supposed to recursively calls itself like these do for all the other platforms but it also returns false (I think this is a bug):

I agree, I pushed a fix for that, let's see if it breaks anything 😄 

> ILLink should be able to inline the false value even without these substitutions 

I don't see it doing that in practice though, if I remove the substitution then it doesn't inline the false value.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Wasm/PackedSimd.cs
```diff
@@ -14,7 +14,7 @@ namespace System.Runtime.Intrinsics.Wasm
     [CLSCompliant(false)]
     public abstract class PackedSimd
     {
-        public static bool IsSupported { [Intrinsic] get { return false; } }
+        public static bool IsSupported { [Intrinsic] get { return IsSupported; } }
```
vitek-karas, line 17: Could you please add a comment why it's done this way?
I was looking at this yesterday and it made no sense... :-)

Would be nice to have comments in the other places as well, at least eventually.

vitek-karas:
> ILLink should be able to inline the false value even without these substitutions (@vitek-karas @sbomer to double check me on this).

Yes - it should be able to do that. But it's definitely worth a test or validation (it depends on the callsite, if the condition around it is too complex we might give up on it).

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Wasm/PackedSimd.cs
akoeplinger: There are quite a lot of intrinsics, I'll do that in a follow-up PR.

MichalStrehovsky:
> I don't see it doing that in practice though, if I remove the substitution then it doesn't inline the false value from the property.

Looking at the source code, looks like ILLinker aborts this if there's IntrinsicAttribute on the method :(

akoeplinger:
Is there a reason why we'd have `[Intrinsic]` on the non-supported case? maybe we should remove that?

MichalStrehovsky:
> Is there a reason why we'd have `[Intrinsic]` on the non-supported case? maybe we should remove that?

Yeah, that would probably help. I don't think it serves a purpose.

jkotas:
> > Is there a reason why we'd have [Intrinsic] on the non-supported case? maybe we should remove that?

> Yeah, that would probably help. I don't think it serves a purpose.

This was introduced in https://github.com/dotnet/coreclr/pull/23751. It does serve purpose. It allows RyuJIT to get rid of the unreachable code sooner. If you delete `[Intrinsic]` on all unsupported cases, I expect that we will see perf regressions.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Wasm/PackedSimd.cs
jkotas: `[Intrinsic]` means that the code you see may not be what is actually executed. The self-recursive implementation is used a lot with `[Intrinsic]` methods.

tannergooding:
> It does serve purpose. 

What Jan said, but noting that it also provides a minor TP increase as the JIT doesn't have to process the IL, method body, etc. We can at import skip all the normal steps and just directly generate a node that returns constant `false` or which throws a `PlatformNotSupportedException`. This additionally allows dead code elimination and other low cost optimizations sooner.

vitek-karas:
Looking at this in more detail - trimmer should really not assume a value of an Intrinsic method - since it can't know what the runtime will do. So I think the fact that it doesn't do constant propagation in this case is by design.

Which means we will need the substitution XML to tell it to do it anyway in these cases.

MichalStrehovsky:
> > It does serve purpose.
> 
> What Jan said, but noting that it also provides a minor TP increase as the JIT doesn't have to process the IL, method body, etc. We can at import skip all the normal steps and just directly generate a node that returns constant `false` or which throws a `PlatformNotSupportedException`. This additionally allows dead code elimination and other low cost optimizations sooner.

I don't see RyuJIT recognizing `Wasm.PackedSimd`. I don't think there's a single metric (JIT compile throughput, size on disk of the XML and `IntrinsicAttribute`, maintenance cost to maintain the XML, etc.) where keeping `Wasm.PackedSimd.IsSupported` marked `Intrinsic` wouldn't be causing a regression.

jkotas:
> I don't see RyuJIT recognizing Wasm.PackedSimd

RyuJIT is recognizing all `IsSupported` intrinsic methods under `System.Runtime.Intrinsics` namespace in CoreLib.

Compile `Console.WriteLine(System.Runtime.Intrinsics.Wasm.PackedSimd.IsSupported)` and run it with `DOTNET_JitDump`. You should see the IsSupported intrinsic recognized:

```
Importing BB01 (PC=000) of 'My:Main()'
    [ 0]   0 (0x000) call 0A000005
In Compiler::impImportCall: opcode is call, kind=0, callRetType is ubyte, structSize is 0
Named Intrinsic System.Runtime.Intrinsics.Wasm.PackedSimd.get_IsSupported: Unsupported - return false
```

PR #93409, opened 2023-10-12
Use the improved vectorization algorithm for binary and ternary TensorPrimitives operations

tannergooding:
This is a continuation of https://github.com/dotnet/runtime/pull/93296


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -13,6 +13,20 @@ namespace System.Numerics.Tensors
...
+        ///
+        ///     This can be beneficial when working with large amounts of memory where the writes would otherwise
+        ///     cause large amounts of repeated updates and evictions. The hardware optimization manuals recommend
+        ///     the threshold to be roughly half the size of the last level of on-die cache -- that is, if you have approximately
+        ///     4MB of L3 cache per core, you'd want this to be approx. 1-2MB, depending on if hyperthreading was enabled.
+        ///
+        ///     However, actually computing the amount of L3 cache per core can be tricky or error prone. Native memcpy
+        ///     algorithms use a constant threshold that is typically around 256KB and we match that here for simplicity. This
+        ///     threshold accounts for most processors in the last 10-15 years that had approx. 1MB L3 per core and support
+        ///     hyperthreading, giving a per core last level cache of approx. 512KB.
```
stephentoub, line 28: Love the comment. Thanks.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2138,684 +2149,5399 @@ static void Vectorized512Small(ref float xRef, ref float dRef, nuint remainder)
...
+
+                                vector1.Store(dPtr + (uint)(Vector512<float>.Count * 0));
+                                vector2.Store(dPtr + (uint)(Vector512<float>.Count * 1));
+                                vector3.Store(dPtr + (uint)(Vector512<float>.Count * 2));
+                                vector4.Store(dPtr + (uint)(Vector512<float>.Count * 3));
+
+                                // We load, process, and store the next four vectors
+
+                                vector1 = TTernaryOperator.Invoke(Vector512.Load(xPtr + (uint)(Vector512<float>.Count * 4)),
+
```
stephentoub, line 7452: Is this better than just letting it be the default case?

stephentoub (approved):

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: Talked over teams and this was meant to be on one of the bits of code that did `case 15: case 14: case 13: ...` all together...

My response was
> Given what the JIT generates right now, maybe...
> that is, since it inserts the cmp; ja right now anyways, it might be better for those to be default
>
> but once that's fixed and its not generating any cmp; ja since it knows everything is definitively in range
> it'd probably be worse, since it would force the branch to exist rather than being part of the regular address lookup table

PR #93255, opened 2023-10-09
Rename `server.socket.address` to `network.peer.address`

antonfirsov:
Address #92956:

Keep .NET metrics in sync with OTel by implementing a naming change from https://github.com/open-telemetry/semantic-conventions/pull/342.  An 8.0 backport is planned to avoid future breaking changes.

PR #93414, opened 2023-10-12
[release/8.0] Adjust System.Net metrics for OTel conformance and consistency

### LINKED ISSUE #92837: `dns.lookup.duration` should report `error.type`

### LINKED ISSUE #92917: Rename `dns.lookupS.duration` to `dns.lookup.duration`

### LINKED ISSUE #92956: [HTTP Metrics] Rename `server.socket.address` to `network.peer.address`
https://github.com/open-telemetry/semantic-conventions/pull/342 is introducing changes to the OTel standard that will effect the conformance of our `http.client.open_connections` and `http.client.connection.duration` metrics. The PR is expected to be merged very soon giving us an opportunity to adjust our metrics for 8.0 GA and avoid breaking changes in 9.0.

- We should rename `server.socket.address` to `network.peer.address` in both metrics
- Given that [it's recommended to set `network.peer.port` when `network.peer.address` is set](https://github.com/open-telemetry/semantic-conventions/blob/d714318c7a29be094cc0152b347b6c4c290a115c/docs/http/http-spans.md#http-client), we should introduce that attribute. The implementation should be trivial, since we are flowing down the peer's `IPEndPoint` to `HttpConnectionBase`.

https://github.com/dotnet/runtime/blob/3452a99cc11fef73ffc34d43c86410d451e33d39/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionBase.cs#L53-L59

### LINKED ISSUE #93302: [Metrics] `error.type` doesn't report `Exception.FullName`

antonfirsov:
Fixes #92956, fixes #92917, fixes #92837, fixes #93302

Backport of PRs #93255, #93254, #93322

## Customer Impact

OpenTelemetry metrics is a new feature in .NET 8.0.
There are minor semantic aspects which are either not conforming to the OpenTelemetry standard (#92956, #92917), inconsistent with ASP.NET (#93302) or highly undesirable (#92837). Fixing these problems in .NET 9.0 would be a breaking change against 8.0, while implementing them in 8.0 GA has a much lower negative impact on customers.

## Testing

Target test cases were added.

## Risk

Low. It is new feature in 8.0, therefore only early adopters of the new feature may be broken - we expect they will prefer conformance to standard.

### DIFF
--- a/src/libraries/System.Net.Http/src/System/Net/Http/Metrics/MetricsHandler.cs
+++ b/src/libraries/System.Net.Http/src/System/Net/Http/Metrics/MetricsHandler.cs
@@ -169,7 +169,7 @@ namespace System.Net.Http.Metrics
                 HttpRequestError.ConfigurationLimitExceeded => "configuration_limit_exceeded",
 
                 // Fall back to the exception type name in case of HttpRequestError.Unknown or when exception is not an HttpRequestException.
-                _ => exception.GetType().Name
+                _ => exception.GetType().FullName!
             };
             return true;
         }
--- a/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/Metrics/ConnectionMetrics.cs
+++ b/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/Metrics/ConnectionMetrics.cs
@@ -14,10 +14,10 @@ namespace System.Net.Http.Metrics
         private readonly object _schemeTag;
         private readonly object _hostTag;
         private readonly object? _portTag;
-        private readonly object? _socketAddressTag;
+        private readonly object? _peerAddressTag;
         private bool _currentlyIdle;
 
-        public ConnectionMetrics(SocketsHttpHandlerMetrics metrics, string protocolVersion, string scheme, string host, int? port, string? socketAddress)
+        public ConnectionMetrics(SocketsHttpHandlerMetrics metrics, string protocolVersion, string scheme, string host, int? port, string? peerAddress)
         {
             _metrics = metrics;
             _openConnectionsEnabled = _metrics.OpenConnections.Enabled;
@@ -25,7 +25,7 @@ namespace System.Net.Http.Metrics
             _schemeTag = scheme;
             _hostTag = host;
             _portTag = port;
-            _socketAddressTag = socketAddress;
+            _peerAddressTag = peerAddress;
         }
 
         // TagList is a huge struct, so we avoid storing it in a field to reduce the amount we allocate on the heap.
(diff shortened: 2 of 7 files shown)


MihaZupan (approved):

karelz (approved):

carlossanlop:

carlossanlop:
The CI failures are unrelated.

PR #93417, opened 2023-10-12
Add `repeatCount` in superpmi.exe to repeat the compilation of 1 or more methods

kunalspathak:
I wanted to investigate the TP of 1 particular method and why it takes longer to JIT. Using superpmi if I can compile it multiple times I can see the area that is slow. One option would be to specify the method context id in `mcl` file repeatedly, but we have a check that they should be incremental. So I added a `-repeatCount` flag that will compile a single context id or entire `mch` file specified number of times, which will help in gathering accurate profile information.

Edit: For now, I did not propagated to `superpmi.py` and can be done if people find this option useful.

### DIFF
--- a/src/coreclr/tools/superpmi/superpmi/superpmi.cpp
+++ b/src/coreclr/tools/superpmi/superpmi/superpmi.cpp
@@ -288,7 +288,7 @@ int __cdecl main(int argc, char* argv[])
     // The method context reader handles skipping any unrequested method contexts
     // Used in conjunction with an MCI file, it does a lot less work...
     MethodContextReader* reader =
-        new MethodContextReader(o.nameOfInputMethodContextFile, o.indexes, o.indexCount, o.hash, o.offset, o.increment);
+        new MethodContextReader(o.nameOfInputMethodContextFile, o.indexes, o.indexCount, o.repeatCount, o.hash, o.offset, o.increment);
     if (!reader->isValid())
     {
         return (int)SpmiResult::GeneralFailure;
@@ -327,6 +327,8 @@ int __cdecl main(int argc, char* argv[])
         }
     }
 
+    for (int iter = 0; iter < o.repeatCount; iter++)
+    {
     while (true)
     {
         MethodContextBuffer mcb = reader->GetNextMethodContext();
@@ -709,6 +711,8 @@ int __cdecl main(int argc, char* argv[])
 
         delete crl;
         delete mc;
+    }
+        reader->Reset();
     }
     delete reader;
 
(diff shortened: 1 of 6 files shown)


kunalspathak:
@dotnet/jit-contrib

src/coreclr/tools/superpmi/superpmi/superpmi.cpp
```diff
@@ -327,6 +327,8 @@ int __cdecl main(int argc, char* argv[])
         }
     }
 
+    for (int iter = 0; iter < o.repeatCount; iter++)
+    {
```
BruceForstall, line 331: It seems like we shouldn't need to change the MethodContextReader at all. Instead of the "repeatCount" loop here, couldn't it be just before the `CompileResult* crl = mc->cr;` line? (There might need to be some kind of saving/restoring of the originalCR).

kunalspathak: possibly, the problem with that though is that the percent completion report will be slower. E.g. if someone pass `-repeatcount 10000` for entire `mch`, the time after which you would see the console output would be after 10000 * 500 compilations.

BruceForstall: It seems wrong that the MethodContextReader should have any knowledge of the repeatCount.

How about have the repeatCount loop include the %complete logging, and displayed % complete would be `repeatIteration * reader->PercentComplete() / repeatCount`. For a repeatCount of 1, this would be no change.

Then, change it to output every 500 iterations whether that is due to loadedCount or repeatCount, e.g.,
```
if ((totalIterations++ % 500 == 0) && (loadedCount > 0))
```

BruceForstall: > How about have the repeatCount loop include the %complete logging, and displayed % complete would be `repeatIteration * reader->PercentComplete() / repeatCount`. For a repeatCount of 1, this would be no change.

My math here is wrong. It would need to be:

```
100.0 * (((reader->CurrentCount() - 1) * repeatCount + repeatIteration) / (repeatCount * reader->TotalCount())
```

for newly exposed MethodContextReader `curIndexPos` (as `CurrentCount()`) and `IndexCount` (as `TotalCount()`)

(this depends on `curIndexPos` returning `1` when you're processing the first method context. curIndexPos and IndexCount also depend on there being a .MCT file, which there isn't always, in which case we would have to return the pos/fileSize like `PercentComplete` does, which wouldn't be as accurate.)

kunalspathak: sorry, just seeing your comment. One of the other problem of moving loop to iterate over `jit->CompileMethod()` is that it will be a tight loop without showing the % progress unless we add another `LogVerbose` inside it. So for eg, if I pass `-c 59 -repeatCount 10000`, i am expected to see % completion after every 500 iterations or so. Unless I am misunderstanding your suggestion of where to add the logging.

kunalspathak:
Replaced by https://github.com/dotnet/runtime/pull/94503

PR #93418, opened 2023-10-12
Revert "Adjust the DNS lookup duration metric (#93254)"

lewing:
This reverts commit 946c5245aea149d9ece53fd0debc18328c29083b.

@antonfirsov looks like the build started failing after #93254 there were known failures but the build analysis was red with several new failures.  Reverting until this is fixed

### DIFF
--- a/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
+++ b/src/libraries/System.Net.NameResolution/src/System/Net/Dns.cs
@@ -23,13 +23,13 @@ namespace System.Net
             {
                 name = NameResolutionPal.GetHostName();
             }
-            catch (Exception ex) when (LogFailure(string.Empty, startingTimestamp, ex))
+            catch when (LogFailure(string.Empty, startingTimestamp))
             {
                 Debug.Fail("LogFailure should return false");
                 throw;
             }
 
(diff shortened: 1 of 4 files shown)


lewing:
It's possible the errors are in fact unrelated in which case feel free to close this, I'm looking into them as well

lewing:
https://github.com/dotnet/runtime/pull/93409/checks?check_run_id=17658454913

antonfirsov:
> https://github.com/dotnet/runtime/pull/93409/checks?check_run_id=17658454913

I cannot spot anything that could be related to #93254, but I see a lot of cases of `TcpReceiveSendGetsCanceledByDispose` failing. #93198 should fix that ideally, I see you already triggered a rerun, feel free to merge it when the CI complates. I wondering if there was any kernel update recently in the queues?

Also note that I did a run of `runtime` and `runtime-libraries-outerloop` under #93254 a couple of hours ago, everything suceeded the "usually failing" set of outerloop runs. I did not run the `runtime-extra-platforms` pipeline though.

This is a critical feature to be backported to `release/8.0` in #93414, I prefer not to revert it unless there is a definite proof that #93254 is to blame, closing for now.

Issue #93403, opened 2023-10-12
Options source gen produce diagnostics with LengthAttribute when applied on properties of interface type


tarekgh:
When using something like the following:

```C#
    [System.ComponentModel.DataAnnotations.Length(1, int.MaxValue)]
    public IList<Foo> Property { get; set; }
```

The options source gen will produce diagnostics 

```
error SYSLIB1217: The validation attribute LengthAttribute should only be applied to properties of type string, array, or ICollection. Using it with the type IList could lead to runtime failures. 
```

This is a regression from the changes in https://github.com/dotnet/runtime/pull/93088.

PR #93426, opened 2023-10-12
Fix Options Source Gen with Length attributes applied on properties of Interface type

tarekgh:
https://github.com/dotnet/runtime/issues/93403

This is fixing a regression from last changes https://github.com/dotnet/runtime/pull/93088. The problem was when we were looking for the `Count` property inside a type, if the type is interfacing the lookup was failing. The reason is `ITypeSymbol.GetMembers` doesn't report the members of the extended interfaces. The fix is just lookup the property in all interfaces implemented by this type.


ericstj (approved):

src/libraries/Microsoft.Extensions.Options/tests/SourceGeneration.Unit.Tests/Baselines/UsingInterfaceAsPropertyTypeForLengthAttributesTests.netcore.g.cs
```diff
@@ -0,0 +1,252 @@
+
+    // <auto-generated/>
+    #nullable enable
+    #pragma warning disable CS1591 // Compensate for https://github.com/dotnet/roslyn/issues/54103
+    namespace Test
+{
```
RussKie, line 6: [nit] mis-indented

src/libraries/Microsoft.Extensions.Options/tests/SourceGeneration.Unit.Tests/Baselines/UsingInterfaceAsPropertyTypeForLengthAttributesTests.netfx.g.cs
```diff
@@ -0,0 +1,177 @@
+
+    // <auto-generated/>
+    #nullable enable
+    #pragma warning disable CS1591 // Compensate for https://github.com/dotnet/roslyn/issues/54103
+    namespace Test
```
RussKie, line 5: ditto

src/libraries/Microsoft.Extensions.Options/tests/SourceGeneration.Unit.Tests/Baselines/UsingInterfaceAsPropertyTypeForLengthAttributesTests.netcore.g.cs
tarekgh: We can fix this in another PR. I want to limit the change here as we need to port this to 8.0. Thanks for the comment.

tarekgh:
failures are tracked by https://github.com/dotnet/runtime/issues/92423 which are unrelated. I am not sure why the build analysis didn't detect it automatically. CC @agocke

Issue #82335, opened 2023-02-18
JIT: support dead block elimination for EH handlers


BruceForstall:
`fgRemoveDeadBlocks()` removes unreachable code by starting from a set of initial blocks and walking control-flow paths until all reachable blocks are marked, then removing any unmarked blocks.

However, this will not remove any EH handlers (finally/fault/catch/filter/filter-handler) even if the associated `try` is not reachable.

Change the logic to only add the EH handler entries corresponding to a `try` to the reachable set once a reachable `try` body is found.

This will also require ensuring the related EH table entry is removed if necessary.

EgorBo:
Example:

```csharp
int Test() => 1;

void Foo()
{
    if (Test() != 1)
    {
        // unreachable
        try
        {
            Console.WriteLine();
        }
        catch
        {
        }
    }
}
```
it's expected for `Foo` to be lowered to no-op, but instead, it emits:
```asm
; Assembly listing for method Program:Foo():this (FullOpts)
G_M52879_IG01:
       push     rbp
       sub      rsp, 16
       lea      rbp, [rsp+10H]
       mov      qword ptr [rbp-10H], rsp
						;; size=14 bbWeight=1 PerfScore 2.75
G_M52879_IG02:
       add      rsp, 16
       pop      rbp
       ret      
						;; size=6 bbWeight=1 PerfScore 1.75
G_M52879_IG03:
       int3     
						;; size=1 bbWeight=0 PerfScore 0.00
G_M52879_IG04:
       push     rbp
       sub      rsp, 16
       mov      rbp, qword ptr [rcx]
       mov      qword ptr [rsp], rbp
       lea      rbp, [rbp+10H]
						;; size=16 bbWeight=0 PerfScore 0.00
G_M52879_IG05:
       lea      rax, G_M52879_IG02
						;; size=7 bbWeight=0 PerfScore 0.00
G_M52879_IG06:
       add      rsp, 16
       pop      rbp
       ret      
; Total bytes of code 50
```

saucecontrol:
I frequently use a pattern where I use exception filters with a JIT-time constant value.  These are unmanaged callbacks, which SEH unwinds properly on Windows without the catch, so I want to catch only on Linux.  The filter makes the catch unreachable on Windows, but the EH block stays (actually I get 2 -- one for the filtered case and one without)

```cs
static readonly bool capture = !OperatingSystem.IsWindows();

static int foo(int x)
{
    int res = 0;
    try {
[...]

PR #93428, opened 2023-10-12
Improve EH dead code elimination

BruceForstall:
Update `fgRemoveDeadBlocks()` to only treat EH handlers as reachable if their corresponding `try` block entry is reachable.

This causes more dead code to be eliminated in handlers that are not reachable.

There is a little more work to do to completely clean these up:
1. If a `try` is not reachable, we should get rid of its entire EH table entry. Currently, its block is generated as `int3` (on xarch).
2. Even though the handler is not reachable, we still generate the prolog followed by `int3` for the body.

Contributes to #82335

### DIFF
--- a/src/coreclr/jit/block.cpp
+++ b/src/coreclr/jit/block.cpp
@@ -637,7 +637,7 @@ void BasicBlock::dspJumpKind()
             break;
 
         case BBJ_EHFILTERRET:
-            printf(" (fltret)");
+            printf(" -> " FMT_BB " (fltret)", bbJumpDest->bbNum);
             break;
 
         case BBJ_EHCATCHRET:
--- a/src/coreclr/jit/clrjit.natvis
+++ b/src/coreclr/jit/clrjit.natvis
@@ -21,7 +21,7 @@ Documentation for VS debugger format specifiers: https://docs.microsoft.com/en-u
   </Type>
 
   <Type Name="BasicBlock">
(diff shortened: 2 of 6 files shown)


BruceForstall:
@AndyAyersMS PTAL
cc @dotnet/jit-contrib

kunalspathak (approved):
LGTM

AndyAyersMS:
> 1. If a `try` is not reachable, we should get rid of its entire EH table entry.

We encode EH region numbers into the AddCodeDesc entries (starting in morph) so if we remove things from the EH table we'll need to fix these up too. Or perhaps rework all this so these entries hang off the EH table somehow and migrate organically along with their related regions.

BruceForstall:
> We encode EH region numbers into the AddCodeDsc entries

In particular, using `bbThrowIndex()`, to ensure `fgAddCodeRef` only shares throw blocks in the same EH region.

AndyAyersMS (approved):
Interesting that we don't see much of a TP win here.

I wonder if we should try doing this earlier?

BruceForstall:
> Interesting that we don't see much of a TP win here.

Would you expect a TP win because of less IR? It seems like unreachable handlers should be rare, especially after the early EH opts. This still doesn't remove an unreachable `try` or the unreachable handler (it just turns them into `int3`), or the "dead" EH region.

> I wonder if we should try doing this earlier?

Seems like it needs to be late enough for all the normal opts creating dead flow to run.

BruceForstall:
Test failure is https://github.com/dotnet/runtime/issues/93506

Issue #91826, opened 2023-09-08
Update .NET Linux Build for .NET 9


richlander:
Update (2024-07): We decided to keep on targeting Ubuntu 16.04 based on other feedback. Doing so hasn't had any significant drawback and maintains significant compatibility.

We should update the Linux build to use Azure Linux 3.0 and target Ubuntu 18.04 for .NET 9. This plan builds on our [successful .NET 8 Linux build project](https://github.com/dotnet/runtime/issues/83428). This is for the Microsoft build. Other builders of .NET 9 can and will make different choices.

- Minimum supported OSes: https://github.com/dotnet/core/blob/main/release-notes/9.0/supported-os.md#linux
- Required libc versions: https://github.com/dotnet/core/blob/main/release-notes/9.0/supported-os.md#linux-compatibility

For .NET 8, we targeted Ubuntu 16.04, in order to provide broad support for older Linux distributions. We considered targeting Ubuntu 18.04 in that timeframe, but that would have broken compatibility with Amazon Linux 2. Upon the request of Amazon, we chose to support Ubuntu 16.04 to avoid that compatibility break. NET 8 will be the last release to provide `glibc` compatibility for Amazon Linux 2, such that we can move forward to targeting Ubuntu 18.04. This means that the Linux support matrix for .NET 8 and 9 will be significantly different.

Ubuntu `glibc` versions:

- Ubuntu 22.04: 2.35
- Ubuntu 20.04: 2.31
- **Ubuntu 18.04: 2.27**
- Ubuntu 16.04: 2.23

Here are some other distributions (bolded distro versions are compatible):

- **AlmaLinux 8: 2.28**
- **AlmaLinux 9: 2.34**
- **Amazon Linux 2023: 2.34**
- Amazon Linux 2: 2.26
- CentOS 7: 2.17
- **CentOS Stream 9: 2.34**
- **Debian 11: 2.31**
- **Debian 12: 2.36**
- **Arch Linux 20230723.0.166908: 2.37**
- **Fedora 37: 2.36**
- **Fedora 38: 2.37**

am11:
<kbd> The older the libc, the broader the compatibility. </kdb>

IIUC, this update is desired not because we are actually using anything specific from modern glibc in code, but the reason for update is infrastructure related. In that case:

[...]

PR #93437, opened 2023-10-13
(chore) set min version of cmake in native CmakeLists.txt to surpass Cmake CMP0000 policy warning

### LINKED ISSUE #93436: [NativeAOT] set StaticICULinking results in a CmakeLists.txt that violates policy CMP0000 of Cmake (affects CMake version >= 3.9.0)
### Intro:
when compiling with `NativeAOT` and `StaticICULinking` flags, the resulting `CmakeLists.txt` does not respect [policy CMP0000 of Cmake ](https://cmake.org/cmake/help/latest/policy/CMP0000.html) and produces a warning : 
```
 CMake Warning (dev) in CMakeLists.txt:
    No cmake_minimum_required command is present.  A line of code such as
  
      cmake_minimum_required(VERSION 3.22)
  
    should be added at the top of the file. 
```

### steps to produce
1- create a new dotnet console project
2- add ` <StaticICULinking>true</StaticICULinking>
        <PublishAot>true</PublishAot>` to `.csproj` file 
3- add `Microsoft.Dotnet.ILCompiler 8.0.0-rc.2.23479.6` nuget
4- run ` dotnet publish -r linux-x64 -c Release  --self-contained`

### Expected Behavior 
no warning should be emitted by Cmake 

### Actual Behavior 
Cmake emits warning about policy CMP0000

### Other Areas Effected ?
perhaps
`System.IO.Compression.Native`,
`System.Security.Cryptography.Native`,
`System.Security.Cryptography.Native.Android`,
`System.Security.Cryptography.Native.Apple`
...

### Solution
myabe setting the min version of cmake to the version that reflects
`https://github.com/dotnet/runtime/blob/main/src/coreclr/CMakeLists.txt` 
or
`https://github.com/dotnet/runtime/blob/main/src/native/libs/CMakeLists.txt`

### Tested System
1- `dotnet 8.0.100-rc.2.23502.2`
2- `cmake version 3.22.1` , This affects CMake version >= 3.9.0 
3- `Linux neo-Intel 6.2.0-34-generic #34~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Sep  7 13:12:03 UTC 2 x86_64 
     x86_64 x86_64 GNU/Linux`
4- `Microsoft.Dotnet.ILCompiler 8.0.0-rc.2.23479.6` nuget

Neo-vortex:
This closes https://github.com/dotnet/runtime/issues/93436

Setting the min version for `cmake` reflecting `https://github.com/dotnet/runtime/blob/main/src/native/libs/CMakeLists.txt`


Neo-vortex:
@dotnet-policy-service agree

ViktorHofer:
cc @jkoritzinsky as you dealt with minimum cmake versions in our repo in the past

src/native/libs/System.Globalization.Native/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+cmake_minimum_required(VERSION 3.20)
```
jkotas, line 1:
```suggestion
# Required for StaticICULinking
cmake_minimum_required(VERSION 3.20)
```

src/native/libs/System.IO.Compression.Native/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+cmake_minimum_required(VERSION 3.20)
+
```
jkotas, line 2:
```suggestion
```
We need this only for the libraries that we support static linking for. The fewer places with cmake version, the better.

src/native/libs/System.IO.Ports.Native/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+cmake_minimum_required(VERSION 3.20)
+
```
jkotas, line 2:
```suggestion
```

src/native/libs/System.Security.Cryptography.Native/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+cmake_minimum_required(VERSION 3.20)
```
jkotas, line 1:
```suggestion
# Required for StaticOpenSslLinking
cmake_minimum_required(VERSION 3.20)
```

src/native/libs/System.Native/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+cmake_minimum_required(VERSION 3.20)
+
```
jkotas, line 2:
```suggestion
```

src/native/libs/System.Net.Security.Native/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+cmake_minimum_required(VERSION 3.20)
+
```
jkotas, line 2:
```suggestion
```

src/native/libs/System.Security.Cryptography.Native.Android/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+cmake_minimum_required(VERSION 3.20)
+
```
jkotas, line 2:
```suggestion
```

src/native/libs/System.Security.Cryptography.Native.Apple/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+cmake_minimum_required(VERSION 3.20)
+
```
jkotas, line 2:
```suggestion
```

jkotas (approved):
Thank you

src/native/libs/System.Globalization.Native/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+# Required for StaticICULinking
+cmake_minimum_required(VERSION 3.20)
```
jkoritzinsky, line 2:
```suggestion
cmake_minimum_required(VERSION 3.16...3.20)
```

Can we change the minimum version to 3.16 to be consistent with other CMake scripts we ship to external customers?

src/native/libs/System.Security.Cryptography.Native/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+# Required for StaticOpenSslLinking
+cmake_minimum_required(VERSION 3.20)
```
jkoritzinsky, line 2:
```suggestion
cmake_minimum_required(VERSION 3.16...3.20)
```

src/native/libs/System.Globalization.Native/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+# Required for StaticICULinking
+cmake_minimum_required(VERSION 3.16...3.20)
```
am11, line 2: What happens when we have > 3.20 installed? Do we want to restrict on max value or can it just be set to 3.16 minimum?

Neo-vortex: you are right.
limiting the max version is probably not what we want to do.
we can set `cmake_minimum_required(VERSION 3.16)`
fixed in the latest commit

jkotas:
How did you come up with 3.16?

It would be better for this version to match the min cmake version used in the rest of the repo.

src/native/libs/System.Globalization.Native/CMakeLists.txt
jkotas: What are the other CMake scripts that we ship?

am11:
> How did you come up with 3.16?

@jkotas, please expand the comments to see the history. It was a comment from @jkoritzinsky to use the version range 3.16...3.20, on which I commented that maybe we don't need to restrict to the max value. I think some old distros are providing older versions, so it makes sense to lower the requirements of "shipping" cmake script, which are not using any new feature.

jkoritzinsky:
3.16 is the minimum required version for the CMake scripts we use to build Apple and Android apps on customer machines (see the CMake templates in AppleAppBuilder and AndroidAppBuilder).

src/native/libs/System.Globalization.Native/CMakeLists.txt
jkoritzinsky: This is not setting a max version. This is setting the version of CMake to set the policy values for. Setting it to 3.20 ensures that for the dotnet/runtime build, we use a consistent set of CMake policy configurations.

Neo-vortex:
@jkotas  
if @jkoritzinsky is right about the min cmake version, probably when can merge this PR as is.

jkotas:
I do not see a strong reason for keeping the Apple/Android builder min cmake version to be in sync with min cmake version for this Linux-specific opt-in option.

I see the following options:

- Use `3.20` to keep things simple. 
- Use `3.16..3.20`. As @jkoritzinsky pointed out, it is important to set 3.20 as the upper bound so that the dotnet/runtime build uses consistent set of policies. 
- Match cmake version that comes with the oldest Linux distro supported by .NET runtime. If it is going [`Ubuntu 18.04`](https://github.com/dotnet/runtime/issues/91826#issuecomment-1714378388) for .NET 9, it would make it `3.10..3.20`.

am11:
>  it is important to set 3.20 as the upper bound so that the dotnet/runtime build uses consistent set of policies.

We don't have any restriction on > 3.20 version. In fact, the _minimum_ version is 3.20.
https://github.com/dotnet/runtime/blob/765798c49d5d908e512af0931b18fed782b4cf00/src/coreclr/CMakeLists.txt#L1

jkoritzinsky:
> > it is important to set 3.20 as the upper bound so that the dotnet/runtime build uses consistent set of policies.
> 
> We don't have any restriction on > 3.20 version. In fact, the _minimum_ version is 3.20.
> 
> https://github.com/dotnet/runtime/blob/765798c49d5d908e512af0931b18fed782b4cf00/src/coreclr/CMakeLists.txt#L1

One aspect of setting the minimum required as 3.20 in that was is to set the CMake policy defaults to the 3.20 rules, no matter the version of CMake that’s actually running the build. Using the `3.10…3.20` syntax says “3.10 is the minimum required version for the features used. Use the newest version of CMake policies available until 3.20. Then use the 3.20 policy defaults”

Neo-vortex:
> > > it is important to set 3.20 as the upper bound so that the dotnet/runtime build uses consistent set of policies.
> > 
> > 
> > We don't have any restriction on > 3.20 version. In fact, the _minimum_ version is 3.20.
> > https://github.com/dotnet/runtime/blob/765798c49d5d908e512af0931b18fed782b4cf00/src/coreclr/CMakeLists.txt#L1
> 
> One aspect of setting the minimum required as 3.20 in that was is to set the CMake policy defaults to the 3.20 rules, no matter the version of CMake that’s actually running the build. Using the `3.10…3.20` syntax says “3.10 is the minimum required version for the features used. Use the newest version of CMake policies available until 3.20. Then use the 3.20 policy defaults”

so `3.10..3.20` is the the best practice here where :
1- builds fail on anything below `3.10`
2- anything above `3.20` would use `3.20` policy defaults

jkoritzinsky (approved):

PR #93445, opened 2023-10-13
Reduce differences between successor iterators

BruceForstall:
The NumSucc/GetSucc/Succs iterators have two variants. It is hard to understand the distinction. The goal of this change is to reduce the differences, so the only difference is the SWITCH case (iterating either all successors or only unique successors).

So:
1. BBJ_EHFILTERRET: always return the single successor, which is the first block of the filter handler.
2. BBJ_EHFINALLYRET: Move to standard successor iterator

### DIFF
--- a/src/coreclr/jit/importer.cpp
+++ b/src/coreclr/jit/importer.cpp
@@ -11471,16 +11471,14 @@ SPILLSTACK:
         // This will re-import all the successors of block (as well as each of their predecessors)
         impReimportSpillClique(block);
 
-        // For blocks that haven't been imported yet, we still need to mark them as pending import.
-        // Filter successor from BBJ_EHFILTERRET have already been handled, above.
-        if (!block->KindIs(BBJ_EHFILTERRET))
+        // We don't expect to see BBJ_EHFILTERRET here.
+        assert(!block->KindIs(BBJ_EHFILTERRET));
+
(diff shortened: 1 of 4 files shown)


BruceForstall:
@AndyAyersMS PTAL
cc @dotnet/jit-contrib

src/coreclr/jit/importer.cpp
```diff
@@ -11528,22 +11528,30 @@ void Compiler::impImportBlock(BasicBlock* block)
         impReimportSpillClique(block);
 
         // For blocks that haven't been imported yet, we still need to mark them as pending import.
-        for (BasicBlock* const succ : block->Succs())
+        // Filter successor from BBJ_EHFILTERRET have already been handled, above.
```
AndyAyersMS, line 11531: I couldn't figure out where the "above" was. Would be nice to avoid this special case.

Seems like we should never get into the spill clique case for filters since the exit stack will be emtpy?

BruceForstall: Ah, I can remove this from the spill clique case if you think it's unnecessary. (Just assert block is not a filterret?) I was just being defensive, since I hit an assert without this in the "normal" case.

BruceForstall:
@AndyAyersMS updated

AndyAyersMS (approved):

Issue #93444, opened 2023-10-13
Test failure: JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b71869/b71869/b71869.dll


BruceForstall:
## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=436667
Build error leg or test failing: JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b71869/b71869/b71869.dll
Pull request: https://github.com/dotnet/runtime/pull/93377

## Error Message

```
03:11:30.340 Running test: JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b71869/b71869/b71869.dll
Xunit.Sdk.EqualException: Assert.Equal() Failure
Expected: 100
Actual:   101
   at Xunit.Assert.Equal[T](T expected, T actual, IEqualityComparer`1 comparer) in /_/src/xunit.assert/Asserts/EqualityAsserts.cs:line 96
   at Program.<<Main>$>g__TestExecutor177|0_178(StreamWriter tempLogSw, StreamWriter statsCsvSw, <>c__DisplayClass0_0&)
03:11:31.175 Failed test: JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b71869/b71869/b71869.dll
```

```json
{
  "ErrorMessage": "Failed test: JIT/Regression/CLR-x86-JIT/V1-M12-Beta2/b71869/b71869/b71869.dll",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

### Report
|Build|Definition|Step Name|Console log|Pull Request|
|---|---|---|---|---|
|[437345](https://dev.azure.com/dnceng-public/public/_build/results?buildId=437345)|dotnet/runtime|coreclr  Pri0 Runtime Tests Run Composite linux x64 checked|[Log](https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_apis/build/builds/437345/logs/567)||
|[437310](https://dev.azure.com/dnceng-public/public/_build/results?buildId=437310)|dotnet/runtime|coreclr  Pri1 Runtime Tests Run R2R_Composite linux x64 checked|[Log](https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_apis/build/builds/437310/logs/1054)||
[...]

Issue #93424, opened 2023-10-12
[7.0] Multiple JIT failures in coreclr arm64 windows


carlossanlop:
Multiple JIT failures. One is an AV, the others are unexpected values.

- Affected 7.0 PR: https://github.com/dotnet/runtime/pull/93307
- Queue: `coreclr Pri0 Runtime Tests Run windows arm64 checked`
- Build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=436582&view=logs&j=bb94f37e-a09b-5b51-09b2-ab78e7aa5103&t=506cadc9-3cc4-5d96-6847-011f75a33382

<details><summary>JIT.1</summary>

- Log file: https://helixre8s23ayyeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-93307-merge-324267ca2c1d4455b5/JIT.1/1/console.2d72728a.log?helixlogtype=result
```
Microsoft.DotNet.XUnitConsoleRunner v2.5.0 (64-bit .NET 7.0.10)
  Discovering: JIT.CheckProjects.XUnitWrapper (method display = ClassAndMethod, method display options = None)
  Discovered:  JIT.CheckProjects.XUnitWrapper (found 1 test case)
  Starting:    JIT.CheckProjects.XUnitWrapper (parallel test collections = on, max threads = 2)
  Finished:    JIT.CheckProjects.XUnitWrapper
  Discovering: JIT.opt.XUnitWrapper (method display = ClassAndMethod, method display options = None)
  Discovered:  JIT.opt.XUnitWrapper (found 127 test cases)
  Starting:    JIT.opt.XUnitWrapper (parallel test collections = on, max threads = 2)
    JIT\opt\AssertionPropagation\ArrBoundElim\ArrBoundElim.cmd [FAIL]
      Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
         at Tests.MulOverflow()
         at Program.RunTestThrows(System.Action)
         at Program.Main(System.String[])
      
      Return code:      1
      Raw output file:      C:\h\w\9F6008E0\w\B1F0099D\uploads\Reports\JIT.opt\AssertionPropagation\ArrBoundElim\ArrBoundElim.output.txt
      Raw output:
      BEGIN EXECUTION
       "C:\h\w\9F6008E0\p\corerun.exe" -p "System.Reflection.Metadata.MetadataUpdater.IsSupported=false"  ArrBoundElim.dll 
[...]

PR #93451, opened 2023-10-13
Disable failing win-arm64 JIT tests

jakobbotsch:
Disable these while we investigate what happened.

Related: #93424, #93441, #93442, #93443, #93444


EgorBo (approved):

jakobbotsch:
@BruceForstall That's right, I've already root caused it in https://github.com/dotnet/runtime/issues/93442#issuecomment-1761247844

jakobbotsch:
We need to revert the compiler upgrade or find a workaround. We can probably work around it by writing the logic in that function in a different, more convoluted way.

PR #93456, opened 2023-10-13
[browser] Move PublishTrimmed=true from pack to SDK

maraf:
- PublishTrimmed must be set early enough so that `ProcessFrameworkReferences` knows it should include a reference to package `Microsoft.NET.ILLink.Tasks`
- SDK PR is https://github.com/dotnet/sdk/pull/36094
- It is not required to be backported
- WBT
  - Move RID to project template
  - Remove duplicated `testCommand` from error message


radical:
oops, I commented on the wrong PR. I was asking for wasmconsole/browser tests to be added in WBT.

maraf:
> and from `Local.Directory.Build.props`, so we are testing the correct thing?

Good idea, why is `RuntimeIdentifier` set there? Isn't it every csproj?

> oops, I commented on the wrong PR. I was asking for wasmconsole/browser tests to be added in WBT.

It's only browser template for now. Do you have some idea for a good test other than testing a specific dll that should get trimmed out?

radical:
> > and from `Local.Directory.Build.props`, so we are testing the correct thing?
> 
> Good idea, why is `RuntimeIdentifier` set there? Isn't it every csproj?

It might be a leftover from some older thing. Does everything still work if it is removed?

> > oops, I commented on the wrong PR. I was asking for wasmconsole/browser tests to be added in WBT.
> 
> It's only browser template for now. Do you have some idea for a good test other than testing a specific dll that should get trimmed out?

Um you could run a target after the linker target, and just check that the `linked/` directory has non-zero number of assemblies. Trying to depend on a certain assembly to trimmed out would likely be brittle.

lewing (approved):

lewing:
one it is green... :)

maraf:
> once it is green... :)

Not yet, but thank you

maraf:
> and from `Local.Directory.Build.props`, so we are testing the correct thing?

Interesting, the `InitProjectDir` copies over the Workloads and Local directory props and targets, but than `CreateWasmTemplateProject` overrides them https://github.com/dotnet/runtime/blob/main/src/mono/wasm/Wasm.Build.Tests/WasmTemplateTestBase.cs#L27-L30

lewing (approved):

PR #93457, opened 2023-10-13
Fix test build error when BuildAsStandalone is true

clamp03:
Build error of provider.ilproj test when BuildAsStandalone is true
`EXEC : error : No entry point declared for executable`

This is due to https://github.com/dotnet/runtime/pull/92029
Restore OutputType to Library.

### DIFF
--- a/src/tests/Regressions/coreclr/22021/provider.ilproj
+++ b/src/tests/Regressions/coreclr/22021/provider.ilproj
@@ -1,4 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.IL">
+  <PropertyGroup>
+    <OutputType>Library</OutputType>
+  </PropertyGroup>
   <ItemGroup>
     <Compile Include="provider.il" />
   </ItemGroup>


trylek (approved):
Thanks for fixing this, I'm however somewhat surprised why we need to explicitly specify the Library OutputType, I was under the impression that it's the default, why isn't the default logic kicking in here?

clamp03:
Actually, I don't know well about tests in dotnet/runtime. I am just guessing that `OutputType` set to `Exe` when `BuildAsStandalone` is set like https://github.com/dotnet/runtime/blob/139f45e56b85b7e643d7e4f81cb5cdf640cd9021/src/tests/Directory.Build.props#L199

clamp03:
@trylek @jkotas I think the failures are not caused by the PR. Could you merge if it doesn't have any issue?

clamp03:
CC @gbalykov  Our internal CI fails for days because of this issue. Thank you.

gbalykov (commented):
LGTM

gbalykov:
@clamp03 thank you!

Issue #89409, opened 2023-07-24
[wasm][debugger] Debugger tests getting randomly aborted on CI


radical:
[log](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-89395-merge-93f9180bd3f44ef8bb/chrome-DebuggerTests.EvaluateOnCallFrameTests/1/console.2085d9ea.log?helixlogtype=result):

```
...
Detected a container, disabling sandboxing for debugger tests.
The active test run was aborted. Reason: Test host process crashed
Results File: /root/helix/work/workitem/uploads/xharness-output/logs/testResults.trx

Passed!  - Failed:     0, Passed:    85, Skipped:     0, Total:    85, Duration: 9 m 18 s - DebuggerTestSuite.dll (net8.0)
Test Run Aborted.
```

lewing:
I bet we are taking longer than 100ms to close https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.CrossPlatEngine/Client/ProxyOperationManager.cs#L325-L327

See the discussion at https://github.com/microsoft/vstest/issues/2952

radical:
I haven't seen this issue since Jul 26. It would still be a good idea to change the timeout like @lewing suggested.

radical:
This hasn't been hit on `main` again, but it is being seen on 7.0 .

SamMonoRT:
hit again in https://github.com/dotnet/runtime/pull/91393

SamMonoRT:
also in https://github.com/dotnet/runtime/pull/91664

pavelsavara:
Hit again in #94299 [log](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-94299-merge-d71f9208eed34257b7/chrome-DebuggerTests.EvaluateOnCallFrameTests/1/console.08664a2b.log?helixlogtype=result)

pavelsavara:
> This hasn't been hit on `main` again, but it is being seen on 7.0 .

I see it on main [Log](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-94299-merge-34f57efe4a224c04b7/chrome-DebuggerTests.EvaluateOnCallFrameTests/1/console.8111fca8.log?helixlogtype=result)

radical:
[...]

PR #93466, opened 2023-10-13
[browser] refactor entropy

### LINKED ISSUE #93398: [wasm] MT browser-bench crash
I was trying to measure wasm/MT perf recently. The browser-bench is crashing for me, during the app start measurements, in this place:

```
AppStart, Reach managed count: 29, per call: 170.79310344827587ms, total: 4.953s
AppStart, Reach managed count: 31, per call: 178.83870967741936ms, total: 5.544s
dotnet.native.js:4738 Uncaught TypeError: Failed to execute 'getRandomValues' on 'Crypto': The provided ArrayBufferView value must not be shared.
    at Object.getBatchedRandomValues (dotnet.native.js:4738:11)
    at _dotnet_browser_entropy (dotnet.native.js:4749:17)
    at dotnet.native.wasm:0x2110c7
    at dotnet.native.wasm:0x210f5d
    at dotnet.native.wasm:0x21f50
    at dotnet.native.wasm:0x1ecb4
    at dotnet.native.wasm:0x12fdd
    at dotnet.native.wasm:0x200e3
    at dotnet.native.wasm:0xfa6b2
    at dotnet.native.wasm:0xc82b7
```

pavelsavara:
Fixes https://github.com/dotnet/runtime/issues/93398

- the real fix is addressing know issue 
    - "In some cases, `instanceof SharedArrayBuffer` returns false even though buffer is an SAB."
- I did it by refactoring away `pal_random.lib.js` and including the logic into the rest of the TypeScript code we have
    - because we already have existing solution for it


src/mono/wasm/runtime/crypto.ts
```diff
@@ -0,0 +1,27 @@
...
+    // When threading is enabled, Chrome doesn't want SharedArrayBuffer to be passed to crypto APIs
+    const needsCopy = isSharedArrayBuffer(memoryView.buffer);
+    const targetBuffer = needsCopy
+        ? new Uint8Array(bufferLength)
+        : targetView;
+
+    // fill the targetBuffer in batches of batchedQuotaMax
+    for (let i = 0; i < bufferLength; i += batchedQuotaMax) {
+        const targetBatch = new Uint8Array(targetBuffer, i, Math.min(bufferLength - i, batchedQuotaMax));
+        crypto.getRandomValues(targetBatch);
```
radekdoulik, line 21: the original implementation was detecting missing crypto in the browser. do we drop that intentionally? I think it makes sense, the coverage is >96% https://caniuse.com/cryptography

radekdoulik (approved):

src/mono/wasm/runtime/crypto.ts
pavelsavara: I added back the original logic. Thanks!

radekdoulik:
the debugger failure is https://github.com/dotnet/runtime/issues/89409

PR #93467, opened 2023-10-13
Remove CompilationProvider dependency from the STJ incremental generator.

eiriktsarpalis:
Removes dependency on `CompilationProvider` from the STJ incremental generator. This should theoretically reduce the number of times that the parser phase is being run in interactive settings, at the expense of liveness when edits are being made in the transitive type graph at locations beyond the `JsonSerializableAttribute` annotations.

FWIW I tested the changes locally in my IDE, but didn't detect any observable impact on liveness. For example, I tried forcing a warning diagnostic while editing a type in the transitive type graph and it appeared almost as soon as I had made the impacting change.

Creating this as a draft PR for now to spark a discussion as to how we want to design source generators of this kind.

cc @ericstj @jaredpar @dotnet/roslyn

### DIFF
--- a/src/libraries/System.Text.Json/gen/JsonSourceGenerator.Roslyn4.0.cs
+++ b/src/libraries/System.Text.Json/gen/JsonSourceGenerator.Roslyn4.0.cs
@@ -28,9 +28,6 @@ namespace System.Text.Json.SourceGeneration
 #if LAUNCH_DEBUGGER
             System.Diagnostics.Debugger.Launch();
 #endif
-            IncrementalValueProvider<KnownTypeSymbols> knownTypeSymbols = context.CompilationProvider
-                .Select((compilation, _) => new KnownTypeSymbols(compilation));
-
             IncrementalValuesProvider<(ContextGenerationSpec?, ImmutableEquatableArray<DiagnosticInfo>)> contextGenerationSpecs = context.SyntaxProvider
                 .ForAttributeWithMetadataName(
 #if !ROSLYN4_4_OR_GREATER
@@ -39,11 +36,11 @@ namespace System.Text.Json.SourceGeneration
                     Parser.JsonSerializableAttributeFullName,
                     (node, _) => node is ClassDeclarationSyntax,
                     (context, _) => (ContextClass: (ClassDeclarationSyntax)context.TargetNode, context.SemanticModel))
-                .Combine(knownTypeSymbols)
                 .Select(static (tuple, cancellationToken) =>
                 {
-                    Parser parser = new(tuple.Right);
-                    ContextGenerationSpec? contextGenerationSpec = parser.ParseContextGenerationSpec(tuple.Left.ContextClass, tuple.Left.SemanticModel, cancellationToken);
+                    KnownTypeSymbols knownTypes = new(tuple.SemanticModel.Compilation);
+                    Parser parser = new(knownTypes);
+                    ContextGenerationSpec? contextGenerationSpec = parser.ParseContextGenerationSpec(tuple.ContextClass, tuple.SemanticModel, cancellationToken);
                     ImmutableEquatableArray<DiagnosticInfo> diagnostics = parser.Diagnostics.ToImmutableEquatableArray();
                     return (contextGenerationSpec, diagnostics);
                 })


src/libraries/System.Text.Json/gen/JsonSourceGenerator.Roslyn4.0.cs
```diff
@@ -39,11 +36,11 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
                     Parser.JsonSerializableAttributeFullName,
                     (node, _) => node is ClassDeclarationSyntax,
                     (context, _) => (ContextClass: (ClassDeclarationSyntax)context.TargetNode, context.SemanticModel))
-                .Combine(knownTypeSymbols)
                 .Select(static (tuple, cancellationToken) =>
                 {
-                    Parser parser = new(tuple.Right);
-                    ContextGenerationSpec? contextGenerationSpec = parser.ParseContextGenerationSpec(tuple.Left.ContextClass, tuple.Left.SemanticModel, cancellationToken);
+                    KnownTypeSymbols knownTypes = new(tuple.SemanticModel.Compilation);
```
333fred, line 41: The reason that this has no impact on liveliness is because it's still running on every keystroke. The `transform` callback will be executed on every keystroke, because semantic information can change on any edit. Practically, I don't believe that this change will have a benefit; in fact, it may be a deoptimization since `KnownTypeSymbols` won't be shared between callbacks.

src/libraries/System.Text.Json/gen/JsonSourceGenerator.Roslyn4.0.cs
```diff
@@ -39,11 +36,11 @@ public void Initialize(IncrementalGeneratorInitializationContext context)
                     Parser.JsonSerializableAttributeFullName,
                     (node, _) => node is ClassDeclarationSyntax,
                     (context, _) => (ContextClass: (ClassDeclarationSyntax)context.TargetNode, context.SemanticModel))
-                .Combine(knownTypeSymbols)
                 .Select(static (tuple, cancellationToken) =>
                 {
-                    Parser parser = new(tuple.Right);
```
CyrusNajmabadi, line 45: note: based on what we saw with the options-generator, can you check this tuple *first* and not do *anything* in the case where the number of found nodes is 0?  that will avoid even doing things like loading symbols from teh compilation when it's entirely not necessary.  Tagging @tarekgh

tarekgh: @CyrusNajmabadi just FYI, I have merged the suggested change in options source generator in main and opened PR https://github.com/dotnet/runtime/pull/93481 to backport it to 8.0 release.

src/libraries/System.Text.Json/gen/JsonSourceGenerator.Roslyn4.0.cs
eiriktsarpalis: Is it not possible though for the transform callback to result in equal values under certain keystrokes? (in this case the `(ClassDeclarationSyntax, SemanticModel)` tuple). Or is it always the case that a keystroke will result in a new SyntaxNode/SemanticModel instance being created?

src/libraries/System.Text.Json/gen/JsonSourceGenerator.Roslyn4.0.cs
eiriktsarpalis: When it comes to symbol loading in the `KnownTypeSymbols` class it's all done lazily so an empty input set wouldn't force any.

eiriktsarpalis: Come to think of it, given that this incremental value is subscribed to attribute annotations, I would expect that the number of nodes should always be greater than zero.

src/libraries/System.Text.Json/gen/JsonSourceGenerator.Roslyn4.0.cs
333fred: New SyntaxNode, no. New SemanticModel, yes, that will be new every time.

eiriktsarpalis: Yeah, I suppose that makes sense. Ok, I think we can leave it as-is in that case.

CyrusNajmabadi: note: even if the number of hits is zero, you're still calling into this select every time.  and you'll have to ensrue that 100% of this code is lazy, or risk regressions.  Easiest is to do what @tarekgh did and literally just bail out as the very first thing if hte hit-count is zero.

Issue #82488, opened 2023-02-22
[Analyzer] Suggest more optimal SIMD patterns where applicable


tannergooding:
Writing SIMD code can be complex and the "optimal" pattern can vary per platform/architecture. This can get more complicated when newer ISAs are also available for use.

As such, we should provide a code fixer that identifies such patterns and suggests simple transformations to alternative methods that are likely to be better performing.

Category: [Performance](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/performance-warnings)
[Severity = suggestion](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-options#severity-level)

### Example

A simple example is given below. The exact scenarios supported/recognized will likely grow over time based on input from SIMD experts:
```diff
- return value * 2;
+ return value + value;
```

-or-
```diff
- return Sse.Shuffle(left, right, 0b11_11_10_10);
+ return Sse.UnpackHigh(left, right);
```

Some of these are patterns that the JIT could recognize and optimize implicitly, but that can be equally complex. Recognizing and suggesting the fixes in C# is often simpler/faster.

terrajobst:
Category: **Performance**

For code that replaces calls into intrinsics this seems fine, such as:

```diff
- Sse.Shuffle(value, value, 0b11_11_10_10);
+ Sse.UnpackHigh(value, value);
```

For more general code that doesn't involve calls to intrinsics we should consider recognizing those in the JIT instead, such as:

```diff
- value * 2;
+ value + value;
```

JimBobSquarePants:
Analyzers like this really cannot come soon enough. Even a cheat sheet would be useful just now.

huoyaoyuan:
The recommended patterns is likely to grow in the future.
Should these patterns use a same diagnostic ID, or separated ones? Does them require a category or range then?
Is there any available cheat sheet, or should us brainstorm for it?

tannergooding:
> Should these patterns use a same diagnostic ID, or separated ones? 

I think it really depends on the operation in question and how likely it is for users to want to separate a given pattern out from the rest.

> Is there any available cheat sheet, or should us brainstorm for it?

I do think https://github.com/dotnet/runtime/issues/82486 is the more prominent/important one to handle first, while its targeted around maintainability/portability, it also has a big impact on JIT codegen and factors into perf as well.

[...]

PR #93469, opened 2023-10-13
Vectorizes IndexOfMin/Max/Magnitude

michaelgsharp:
Adds a vectorized implementation of `IndexOfMin`, `IndexOfMax`, `IndexOfMaxMagnitude`, `IndexOfMinMagnitude`.


michaelgsharp:
I'm still finishing up vectorizing the NetStandard version so its not included here but I wanted to get it up for review. Should have the NetStandard done here shortly. When its done the ifdefs in `TensorPrimitives.cs` will go away, so don't worry about those.

Also, not sure if the `.sln` change should be there or not? Anyone know?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -3098,6 +3332,802 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
+                        curMax = result[i];
+                        curIn = maxIndex[i];
+                    }
+                    else if (result[i] > curMax)
+                    {
+                        curMax = result[i];
+                        curIn = maxIndex[i];
+                    }
+                }
+                return curIn;
```
stephentoub, line 3410: This will do the operation four times.  Could we instead compare result[0]/result[1] and result[2]/result[3], and then compare the result of those two operations to make it 3 instead of 4?

Same should be doable for Vector256, doing four-pairwise comparisons, then two more, and then a final one, giving you 7.  But you should also be able to achieve that just by delegating with GetLower and GetUpper to the Vector128 implementation and then just comparing the results of those.  Similarly for the Vector512.

Seems like you should also be able to factor out the body of this loop into an aggressively inlined helper that you use from all of the vector widths.

That's all unless Tanner has a good idea for how to vectorize this horizontal aggregation.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -3098,6 +3332,802 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
+                    }
+                    else if (current > curMax)
+                    {
+                        curMax = current;
+                        curIn = i;
+                    }
+                }
+
+                return curIn;
+            }
```
stephentoub, line 3390: If you keep a separate scalar function, you should be able to then call it from both the netcore and netstandard implementations.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -316,6 +316,10 @@ public static float Distance(ReadOnlySpan<float> x, ReadOnlySpan<float> y)
         /// </remarks>
         public static int IndexOfMax(ReadOnlySpan<float> x)
         {
+#if NET8_0_OR_GREATER
+            return IndexOfMinMaxCore<IndexOfMaxOperator>(x);
+#else
```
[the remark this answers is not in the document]
michaelgsharp: Thats correct. Should be done soon.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1048,7 +1047,7 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
             if (Vector512.IsHardwareAccelerated && x.Length >= Vector512<float>.Count)
             {
                 ref float xRef = ref MemoryMarshal.GetReference(x);
-
+                Vector512<int> minIndex = Vector512.Create(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16);
```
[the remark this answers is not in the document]
michaelgsharp: Shouldn't be there. Its removed now.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1213,6 +1212,197 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+        /// <remarks>
+        /// This is the same as <see cref="Aggregate{TLoad, TAggregate}(ReadOnlySpan{float})"/>,
+        /// except it early exits on NaN.
+        /// </remarks>
+        private static int IndexOfMinMaxCore<TIndexOfMinMax>(ReadOnlySpan<float> x) where TIndexOfMinMax : struct, IIndexOfOperator
+        {
+            if (x.IsEmpty)
+            {
+                return -1;
+            }
```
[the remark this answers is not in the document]
michaelgsharp: Moved into the shared method.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1213,6 +1212,197 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+            // otherwise returns the greater of the inputs.
+            // It treats +0 as greater than -0 as per the specification.
+
+#if NET8_0_OR_GREATER
+            if (Vector512.IsHardwareAccelerated && x.Length >= Vector512<float>.Count)
+            {
+                ref float xRef = ref MemoryMarshal.GetReference(x);
+                Vector512<int> maxIndex = Vector512.Create(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+                Vector512<int> curIndex = Vector512.Create(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+                Vector512<int> increment = Vector512.Create(16);
```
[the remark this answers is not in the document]
michaelgsharp: I guess my vote would be to use `.Count`? But I'm fine with whatever.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -3098,6 +3332,802 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
+                    Vector128.Equals(remainderMask.AsInt32(), Vector128<int>.Zero),
+                    maxIndex,
+                    Vector128.ConditionalSelect(Vector128.Equals(max, current).AsInt32(),
+                    Vector128.ConditionalSelect(IsNegative(max).AsInt32(), curIndex, maxIndex),
+                    Vector128.ConditionalSelect(Vector128.GreaterThan(max, current).AsInt32(), maxIndex, curIndex)));
+
+                max = Vector128.ConditionalSelect(
+                    Vector128.Equals(remainderMask, Vector128<float>.Zero),
+                    max,
+                    MaxOperator.Invoke(max, current));
```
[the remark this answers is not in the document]
michaelgsharp: It only matters for the last vector and we should only be calling into that method on the last vector. If its calling it before then than its a bug, though I wasn't seeing that behavior.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1213,6 +1212,197 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+            if (Vector512.IsHardwareAccelerated && x.Length >= Vector512<float>.Count)
+            {
+                ref float xRef = ref MemoryMarshal.GetReference(x);
+                Vector512<int> maxIndex = Vector512.Create(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+                Vector512<int> curIndex = Vector512.Create(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+                Vector512<int> increment = Vector512.Create(16);
+
+                // Load the first vector as the initial set of results, and bail immediately
+                // to scalar handling if it contains any NaNs (which don't compare equally to themselves).
+                Vector512<float> result = Vector512.LoadUnsafe(ref xRef, 0), current;
```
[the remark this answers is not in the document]
michaelgsharp: Yeah, that is correct. Any ideas for a good name?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1213,6 +1212,197 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+            // This matches the IEEE 754:2019 `maximum`/`minimum` functions.
+            // It propagates NaN inputs back to the caller and
+            // otherwise returns the greater of the inputs.
+            // It treats +0 as greater than -0 as per the specification.
+
+#if NET8_0_OR_GREATER
+            if (Vector512.IsHardwareAccelerated && x.Length >= Vector512<float>.Count)
+            {
+                ref float xRef = ref MemoryMarshal.GetReference(x);
+                Vector512<int> maxIndex = Vector512.Create(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
```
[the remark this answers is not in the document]
michaelgsharp: Correct, I thought I had renamed them all to result (and I'll rename them again when we figure out a better name) but obviously I missed some. Fixed now.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
michaelgsharp: I was trying to figure out how to do the comparison as you are suggesting, but since I need the value and the index I decided to leave it like this for now until I finished the NetStandard code. Then i'll come back and revisit it.

Since it only happens at the very end/1 time do you think it will end up making a big perf difference?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
michaelgsharp: Good idea. I'll do that.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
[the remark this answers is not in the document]
michaelgsharp: netstandard is now done so this code is now gone.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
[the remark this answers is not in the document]
michaelgsharp: Switched to use `.Count`

ericstj:
Failures here are test failures, the `MathFMaxMagnitude` uses `IndexOfMaxMagnitude` which is changed in this PR.  I suspect this change broke those tests.  Probably you don't have those tests locally as Stephen recently added them.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -3098,6 +3326,802 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
+            {
+                Vector128<float> maxMag = Vector128.Abs(max), currentMag = Vector128.Abs(current);
+
+                maxIndex = Vector128.ConditionalSelect(Vector128.Equals(max, current).AsInt32(),
+                    Vector128.ConditionalSelect(IsNegative(max).AsInt32(), curIndex, maxIndex),
+                    Vector128.ConditionalSelect(Vector128.GreaterThan(maxMag, currentMag).AsInt32(), maxIndex, curIndex));
+
+                max = Vector128.ConditionalSelect(Vector128.Equals(max, current),
+                        Vector128.ConditionalSelect(IsNegative(max), current, max),
+                        Vector128.Max(max, current));
```
ericstj, line 3606: Looks like this needs the same fix I made to netstandar version

ericstj: I just put up a fix for this as well.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1213,6 +1211,192 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                    TIndexOfMinMax.Invoke(ref result, current, ref resultIndex, curIndex, remainderMask);
+
+                }
+
+                // Aggregate the lanes in the vector to create the final scalar result.
+                return TIndexOfMinMax.Invoke(result, resultIndex);
+            }
+
+            // Scalar path used when either vectorization is not supported or the input is too small to vectorize.
+            {
```
ericstj, line 1395: nit: Were braces here intentional?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -3098,6 +3326,802 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
+            public static int Invoke(ReadOnlySpan<float> result)
+            {
+                float curMax = result[0];
+                int curIn = 0;
+                if (float.IsNaN(curMax))
+                {
+                    return curIn;
+                }
+
+                for (int i = 1; i < result.Length; i++)
```
ericstj, line 3360: I was expecting this to look like https://github.com/dotnet/runtime/pull/93469/files#diff-c9855f90de28628f094b952b640408bbbd586d0d9ac0faf2542b5ba1e289b01fL324 starting at 0 but it seems you changed it a bit.  Was there a reason for this?  Also I see your other implementations start at 0 / MinValue.
I have the same comment for all cases where you changed to start at 1.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -3098,6 +3326,802 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
+                {
+                    float current = result[i];
+                    if (float.IsNaN(current))
+                    {
+                        return i;
+                    }
+
+                    if (MathF.Abs(curMax) == MathF.Abs(current))
```
ericstj, line 3556: Was there a reason you didn't capture the value of Abs in a local like was done before?  https://github.com/dotnet/runtime/pull/93469/files#diff-c9855f90de28628f094b952b640408bbbd586d0d9ac0faf2542b5ba1e289b01fL381
I had the same comment for all instances of these.

src/libraries/System.Numerics.Tensors/src/ReferenceAssemblyExclusions.txt
```diff
@@ -0,0 +1,2 @@
+M:System.Numerics.Tensors.TensorPrimitives.ConvertToHalf(System.ReadOnlySpan{System.Single},System.Span{System.Half})
+M:System.Numerics.Tensors.TensorPrimitives.ConvertToSingle(System.ReadOnlySpan{System.Half},System.Span{System.Single})
```
tannergooding, line 2: Why are these excluded?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1213,6 +1211,192 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+            if (Vector512.IsHardwareAccelerated && x.Length >= Vector512<float>.Count)
+            {
+                ref float xRef = ref MemoryMarshal.GetReference(x);
+                Vector512<int> resultIndex = Vector512.Create(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+                Vector512<int> curIndex = Vector512.Create(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+                Vector512<int> increment = Vector512.Create(Vector512<float>.Count);
+
+                // Load the first vector as the initial set of results, and bail immediately
+                // to scalar handling if it contains any NaNs (which don't compare equally to themselves).
+                Vector512<float> result = Vector512.LoadUnsafe(ref xRef, 0), current;
```
tannergooding, line 1235: nit: Please separate this. We should have 2 explicit declarations:
```suggestion
                Vector512<float> result = Vector512.LoadUnsafe(ref xRef, 0);
                Vector512<float> current;
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1213,6 +1211,192 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+
+                int oneVectorFromEnd = x.Length - Vector512<float>.Count;
+                int i = Vector512<float>.Count;
+
+                // Aggregate additional vectors into the result as long as there's at least one full vector left to process.
+                while (i <= oneVectorFromEnd)
+                {
+                    // Load the next vector, and early exit on NaN.
+                    current = Vector512.LoadUnsafe(ref xRef, (uint)i);
+                    curIndex = Vector512.Add(curIndex, increment);
```
tannergooding, line 1249: Operators work and are often easier to read:
```suggestion
                    curIndex += increment;
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1213,6 +1211,192 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+            {
+                ref float xRef = ref MemoryMarshal.GetReference(x);
+                Vector512<int> resultIndex = Vector512.Create(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+                Vector512<int> curIndex = Vector512.Create(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+                Vector512<int> increment = Vector512.Create(Vector512<float>.Count);
+
+                // Load the first vector as the initial set of results, and bail immediately
+                // to scalar handling if it contains any NaNs (which don't compare equally to themselves).
+                Vector512<float> result = Vector512.LoadUnsafe(ref xRef, 0), current;
+                if (!Vector512.EqualsAll(result, result))
```
tannergooding, line 1236: Operators work and the comment helps clarify intent of the check:
```suggestion
                if (result != result) // check if any are NaN
```

tannergooding: Alternatively, this would probably be more efficient as some form of:
```csharp
Vector512<float> nanMask = Vector512.Equals(result, result);

if (nanMask != Vector512<float>.Zero)
{
    return i + IndexOfFirstMatch(nanMask);
}
```

tannergooding: Where `IndexOfFirstMatch` is functionally:
```csharp
static int IndexOfFirstMatch(Vector512<float> mask)
{
    return BitOperations.TrailingZeroCount(mask.ExtractMostSignificantBits);
}
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: For `Arm64` you want `AdvSimd.MaxAcross`.

For `x64` the "best" approach is to shuffle and compare `log2(count)` times. That is, you start with `x = [10, 20, 30, 40]`:
```
y = Shuffle(x, Vector128.Create(2, 3, 0, 1))   // y = [30, 40, 10, 20]
x = Max(x, y)                                  // x = [30, 40, 30, 40]
y = Shuffle(x, Vector128.Create(1, 0, 3, 2))   // y = [40, 30, 40, 30]
x = Max(x, y))                                 // x = [40, 40, 40, 40]
return x.ToScalar();
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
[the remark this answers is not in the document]
tannergooding: In the scalar implementation we simply call it:
```csharp
max: the currently tracked max value
result: the currently tracked index of that max value
current: the current values being checked
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -3098,6 +3326,802 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public static void Invoke(ref Vector128<float> max, Vector128<float> current, ref Vector128<int> maxIndex, Vector128<int> curIndex)
+            {
+                maxIndex = Vector128.ConditionalSelect(Vector128.Equals(max, current).AsInt32(),
+                    Vector128.ConditionalSelect(IsNegative(max).AsInt32(), curIndex, maxIndex),
+                    Vector128.ConditionalSelect(Vector128.GreaterThan(max, current).AsInt32(), maxIndex, curIndex));
+
+                max = Vector128.ConditionalSelect(Vector128.Equals(max, current),
+                        Vector128.ConditionalSelect(IsNegative(max), current, max),
+                        Vector128.Max(max, current));
```
tannergooding, line 3416: Is this resulting in good codegen? It's a lot of conditionals and some of it is handling cases that are expected to be exceptional.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: > For x64 the "best" approach is to shuffle and compare log2(count) times

@tannergooding, could you look at the other aggregations being performed for other operators, in particular the HorizontalAggregate helper:
https://github.com/dotnet/runtime/blob/bf10b617cbc6b7175b00b1a68599f2597c949710/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs#L2866-L2884
and help fix them if there's meaningful opportunity to do better? Thanks.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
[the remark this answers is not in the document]
michaelgsharp: I believe the names I have match what we have in the scalar

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
michaelgsharp: @tannergooding how did you know which numbers to use in the `Vector128.Create`? And how would you then be able to take that and apply it to the index?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
michaelgsharp: They were not. Leftovers that should have been cleaned up.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
michaelgsharp: I should change them all to start at 1 instead of zero, saves 1 itteration of the loop. But based on some comments earlier I will be changing how it does the final step and so will probably be removing the loop entirely. If I don't end up removing the loop I'll make them all start consistently.

src/libraries/System.Numerics.Tensors/src/ReferenceAssemblyExclusions.txt
michaelgsharp: They can't go in the ref because `System.Half` doens't exist in netstandard 2.0/framework.

tannergooding: If they aren’t in the ref, they aren’t available on net8.0 either.

they likely need to be conditioned in the ref so they exist for net8.0 and not for netstandard2.0

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
michaelgsharp: ALright, I have converted it to this pattern.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
michaelgsharp: How can I check if the codegen is good? Lets sync on that tomorrow.

src/libraries/System.Numerics.Tensors/src/ReferenceAssemblyExclusions.txt
stephentoub: They're already in a separate file that's already conditionally compiled in. I assume Michael is doing this so that when he runs the generator, it doesn't try to add them to the main file?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: The signature for shuffle is `public static Vector128<float> Shuffle (Vector128<float> vector, Vector128<int> indices);`, so the values passed to `Vector128.Create(...)` are simply the indice of `vector` that you want in the particular position.

For `Vector128` we simply want to ensure that every value and so the indices picked don't matter too much, as long as they eventually add up. I tend to start with simply swapping the upper/lower halves, and then swapping pairs for the second. Thus if the default indice order is `[0, 1, 2, 3]`, I choose `[2, 3, 0, 1]`, and then `[1, 0, 3, 2]`.

For `Vector256` and `Vector512` you basically want to do the upper/lower halves and then do the 128-bit algorithm. So for `MaxAcross` for `Vector512` is:
```
// Max the upper/lower halves of the Vector512
Vector256<float> val256 = Max(val512.GetLower(), val512.GetUpper());

// Max the upper/lower halves of the Vector256
Vector128<float> val128 = Max(val256.GetLower(), val256.GetUpper());

// The standard 128-bit algorithm given above
Vector128<float> tmp128 = Shuffle(val128, Vector128.Create(2, 3, 0, 1))
val128 = Max(val128, tmp128);
tmp128 = Shuffle(val128, Vector128.Create(1, 0, 3, 2));
val128 = Max(val128, tmp128);
return val128.ToScalar();
```

src/libraries/System.Numerics.Tensors/src/ReferenceAssemblyExclusions.txt
michaelgsharp: Thats correct, this prevents them from being added to the main file.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
michaelgsharp: Nope, no reason. The other places are now using different code and don't have this behavior. Would it be better to capture it in a local instead of doing it twice?

ericstj (approved):

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -316,39 +316,11 @@ public static float Distance(ReadOnlySpan<float> x, ReadOnlySpan<float> y)
...
-                    {
-                        result = i;
-                        max = current;
-                    }
-                }
+                return -1;
             }
-
-            return result;
+            return IndexOfMinMaxCore<IndexOfMaxOperator>(x);
```
stephentoub, line 323: Nit:
```suggestion

            return IndexOfMinMaxCore<IndexOfMaxOperator>(x);
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -367,43 +339,11 @@ public static int IndexOfMax(ReadOnlySpan<float> x)
...
-                        result = i;
-                        max = current;
-                        maxMag = currentMag;
-                    }
-                }
+                return -1;
             }
-
-            return result;
+            return IndexOfMinMaxCore<IndexOfMaxMagnitudeOperator>(x);
```
stephentoub, line 346: Nit:
```suggestion

            return IndexOfMinMaxCore<IndexOfMaxMagnitudeOperator>(x);
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -421,39 +361,11 @@ public static int IndexOfMaxMagnitude(ReadOnlySpan<float> x)
...
-                    {
-                        result = i;
-                        min = current;
-                    }
-                }
+                return -1;
             }
-
-            return result;
+            return IndexOfMinMaxCore<IndexOfMinOperator>(x);
```
stephentoub, line 368: Nit:
```suggestion

            return IndexOfMinMaxCore<IndexOfMinOperator>(x);
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -472,43 +384,11 @@ public static int IndexOfMin(ReadOnlySpan<float> x)
...
-                        result = i;
-                        min = current;
-                        minMag = currentMag;
-                    }
-                }
+                return -1;
             }
-
-            return result;
+            return IndexOfMinMaxCore<IndexOfMinMagnitudeOperator>(x);
```
stephentoub, line 391: Nit:
```suggestion

            return IndexOfMinMaxCore<IndexOfMinMagnitudeOperator>(x);
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
             }
         }
 
+        /// <remarks>
+        /// This is the same as <see cref="Aggregate{TLoad, TAggregate}(ReadOnlySpan{float})"/>,
```
stephentoub, line 1229: This comment looks wrong.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+            // It propagates NaN inputs back to the caller and
+            // otherwise returns the greater of the inputs.
+            // It treats +0 as greater than -0 as per the specification.
+
+#if NET8_0_OR_GREATER
+            if (Vector512.IsHardwareAccelerated && x.Length >= Vector512<float>.Count)
+            {
+                ref float xRef = ref MemoryMarshal.GetReference(x);
+                Vector512<int> resultIndex = Vector512.Create(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+                Vector512<int> curIndex = Vector512.Create(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
```
stephentoub, line 1244:
```suggestion
                Vector512<int> curIndex = resultIndex;
```
?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                    if (nanMask != Vector512<float>.Zero)
+                    {
+                        return curIndex[IndexOfFirstMatch(nanMask)];
+                    }
+
+                    TIndexOfMinMax.Invoke(ref result, current, ref resultIndex, curIndex);
+                }
+
+                // Aggregate the lanes in the vector to create the final scalar result.
+
```
stephentoub, line 1296: Nit:
```suggestion
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                // Load the first vector as the initial set of results, and bail immediately
+                // to scalar handling if it contains any NaNs (which don't compare equally to themselves).
+                Vector512<float> result = Vector512.LoadUnsafe(ref xRef);
+                Vector512<float> current;
+                Vector512<float> nanMask = ~Vector512.Equals(result, result);
+
+                if (nanMask != Vector512<float>.Zero)
+                {
+                    return IndexOfFirstMatch(nanMask);
+                }
```
stephentoub, line 1256: Nit:
```suggestion

                Vector512<float> nanMask = ~Vector512.Equals(result, result);
                if (nanMask != Vector512<float>.Zero)
                {
                    return IndexOfFirstMatch(nanMask);
                }
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+
+                return TIndexOfMinMax.Invoke(result, resultIndex);
+            }
+#endif
+
+            if (Vector256.IsHardwareAccelerated && x.Length >= Vector256<float>.Count)
+            {
+                ref float xRef = ref MemoryMarshal.GetReference(x);
+                Vector256<int> resultIndex = Vector256.Create(0, 1, 2, 3, 4, 5, 6, 7);
+                Vector256<int> curIndex = Vector256.Create(0, 1, 2, 3, 4, 5, 6, 7);
```
stephentoub, line 1305:
```suggestion
                Vector256<int> curIndex = resultIndex;
```
?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                Vector256<int> increment = Vector256.Create(Vector256<float>.Count);
+
+                // Load the first vector as the initial set of results, and bail immediately
+                // to scalar handling if it contains any NaNs (which don't compare equally to themselves).
+                Vector256<float> result = Vector256.LoadUnsafe(ref xRef);
+                Vector256<float> current;
+                Vector256<float> nanMask = ~Vector256.Equals(result, result);
+
+                if (nanMask != Vector256<float>.Zero)
```
stephentoub, line 1314: Nit:
```suggestion

                Vector256<float> nanMask = ~Vector256.Equals(result, result);
                if (nanMask != Vector256<float>.Zero)
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                // Aggregate additional vectors into the result as long as there's at least one full vector left to process.
+                while (i <= oneVectorFromEnd)
+                {
+                    // Load the next vector, and early exit on NaN.
+                    current = Vector256.LoadUnsafe(ref xRef, (uint)i);
+                    curIndex += increment;
+
+                    nanMask = ~Vector256.Equals(current, current);
+
+                    if (nanMask != Vector256<float>.Zero)
```
stephentoub, line 1331: Nit:
```suggestion
                    nanMask = ~Vector256.Equals(current, current);
                    if (nanMask != Vector256<float>.Zero)
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                    {
+                        return curIndex[IndexOfFirstMatch(nanMask)];
+                    }
+
+                    TIndexOfMinMax.Invoke(ref result, current, ref resultIndex, curIndex);
+
```
stephentoub, line 1355: Nit:
```suggestion
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                // If any elements remain, handle them in one final vector.
+                if (i != x.Length)
+                {
+                    current = Vector256.LoadUnsafe(ref xRef, (uint)(x.Length - Vector256<float>.Count));
+                    curIndex += Vector256.Create(x.Length - i);
+
+                    nanMask = ~Vector256.Equals(current, current);
+
+                    if (nanMask != Vector256<float>.Zero)
```
stephentoub, line 1349: Nit:
```suggestion
                    nanMask = ~Vector256.Equals(current, current);
                    if (nanMask != Vector256<float>.Zero)
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                // Aggregate additional vectors into the result as long as there's at least one full vector left to process.
+                while (i <= oneVectorFromEnd)
+                {
+                    // Load the next vector, and early exit on NaN.
+                    current = Vector512.LoadUnsafe(ref xRef, (uint)i);
+                    curIndex += increment;
+
+                    nanMask = ~Vector512.Equals(current, current);
+
+                    if (nanMask != Vector512<float>.Zero)
```
stephentoub, line 1270: Nit:
```suggestion
                    nanMask = ~Vector512.Equals(current, current);
                    if (nanMask != Vector512<float>.Zero)
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                Vector512<int> curIndex = Vector512.Create(0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15);
+                Vector512<int> increment = Vector512.Create(Vector512<float>.Count);
+
+                // Load the first vector as the initial set of results, and bail immediately
+                // to scalar handling if it contains any NaNs (which don't compare equally to themselves).
+                Vector512<float> result = Vector512.LoadUnsafe(ref xRef);
+                Vector512<float> current;
+                Vector512<float> nanMask = ~Vector512.Equals(result, result);
+
+                if (nanMask != Vector512<float>.Zero)
```
stephentoub, line 1253: Nit:
```suggestion

                Vector512<float> nanMask = ~Vector512.Equals(result, result);
                if (nanMask != Vector512<float>.Zero)
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+
+                // If any elements remain, handle them in one final vector.
+                if (i != x.Length)
+                {
+                    current = Vector512.LoadUnsafe(ref xRef, (uint)(x.Length - Vector512<float>.Count));
+                    curIndex += Vector512.Create(x.Length - i);
+
+                    nanMask = ~Vector512.Equals(current, current);
+
+                    if (nanMask != Vector512<float>.Zero)
```
stephentoub, line 1287: Nit:
```suggestion

                    nanMask = ~Vector512.Equals(current, current);
                    if (nanMask != Vector512<float>.Zero)
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                Vector128<int> curIndex = Vector128.Create(0, 1, 2, 3);
+                Vector128<int> increment = Vector128.Create(Vector128<float>.Count);
+
+                // Load the first vector as the initial set of results, and bail immediately
+                // to scalar handling if it contains any NaNs (which don't compare equally to themselves).
+                Vector128<float> result = Vector128.LoadUnsafe(ref xRef);
+                Vector128<float> current;
+                Vector128<float> nanMask = ~Vector128.Equals(result, result);
+
+                if (nanMask != Vector128<float>.Zero)
```
stephentoub, line 1375: Nit:
```suggestion

                Vector128<float> nanMask = ~Vector128.Equals(result, result);
                if (nanMask != Vector128<float>.Zero)
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                // Aggregate additional vectors into the result as long as there's at least one full vector left to process.
+                while (i <= oneVectorFromEnd)
+                {
+                    // Load the next vector, and early exit on NaN.
+                    current = Vector128.LoadUnsafe(ref xRef, (uint)i);
+                    curIndex += increment;
+
+                    nanMask = ~Vector128.Equals(current, current);
+
+                    if (nanMask != Vector128<float>.Zero)
```
stephentoub, line 1392: Nit:
```suggestion

                    nanMask = ~Vector128.Equals(current, current);
                    if (nanMask != Vector128<float>.Zero)
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+
+                // If any elements remain, handle them in one final vector.
+                if (i != x.Length)
+                {
+                    curIndex += Vector128.Create(x.Length - i);
+
+                    current = Vector128.LoadUnsafe(ref xRef, (uint)(x.Length - Vector128<float>.Count));
+                    nanMask = ~Vector128.Equals(current, current);
+
+                    if (nanMask != Vector128<float>.Zero)
```
stephentoub, line 1410: Nit:
```suggestion

                    nanMask = ~Vector128.Equals(current, current);
                    if (nanMask != Vector128<float>.Zero)
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                    {
+                        return curIndex[IndexOfFirstMatch(nanMask)];
+                    }
+
+                    TIndexOfMinMax.Invoke(ref result, current, ref resultIndex, curIndex);
+
```
stephentoub, line 1416: Nit:
```suggestion
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                    nanMask = ~Vector256.Equals(current, current);
+
+                    if (nanMask != Vector256<float>.Zero)
+                    {
+                        return i + IndexOfFirstMatch(nanMask);
+                    }
+
+                    TIndexOfMinMax.Invoke(ref result, current, ref resultIndex, curIndex);
+
```
stephentoub, line 1337: Nit:
```suggestion
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,219 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                    nanMask = ~Vector128.Equals(current, current);
+
+                    if (nanMask != Vector128<float>.Zero)
+                    {
+                        return i + IndexOfFirstMatch(nanMask);
+                    }
+
+                    TIndexOfMinMax.Invoke(ref result, current, ref resultIndex, curIndex);
+
```
stephentoub, line 1398: Nit:
```suggestion
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -7824,80 +8079,697 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
+            static abstract int Invoke(Vector128<float> result, Vector128<int> resultIndex);
+            static abstract void Invoke(ref Vector128<float> result, Vector128<float> current, ref Vector128<int> resultIndex, Vector128<int> curIndex);
+            static abstract int Invoke(Vector256<float> result, Vector256<int> resultIndex);
+            static abstract void Invoke(ref Vector256<float> result, Vector256<float> current, ref Vector256<int> resultIndex, Vector256<int> curIndex);
+#if NET8_0_OR_GREATER
+            static abstract int Invoke(Vector512<float> result, Vector512<int> resultIndex);
+            static abstract void Invoke(ref Vector512<float> result, Vector512<float> current, ref Vector512<int> resultIndex, Vector512<int> curIndex);
+#endif
+
+            static virtual float IdentityValue => throw new NotSupportedException();
```
stephentoub, line 8094: Where is this implemented or used?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -9231,6 +10103,62 @@ public static Vector512<float> Invoke(Vector512<float> x)
...
+            else return Vector128.ConditionalSelect(mask, left, right);
+        }
+
+        private static Vector128<int> ElementWiseSelect(Vector128<int> mask, Vector128<int> left, Vector128<int> right)
+        {
+            if (Sse41.IsSupported)
+            {
+                return Sse41.BlendVariable(left, right, ~mask);
+            }
+            else return Vector128.ConditionalSelect(mask, left, right);
```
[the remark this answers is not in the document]
stephentoub: And if BlendVariable is better here, are there other places in this file we're using ConditionalSelect we should instead be using BlendVariable?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
```diff
@@ -311,6 +311,93 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x, TMinMaxO
             return result;
         }
 
+        /// <remarks>
+        /// This is the same as <see cref="Aggregate{TTransformOperator, TAggregationOperator}(ReadOnlySpan{float}, TTransformOperator, TAggregationOperator)"/>
```
stephentoub, line 315: This comment looks wrong.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
```diff
@@ -311,6 +311,93 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x, TMinMaxO
...
+                    }
+
+                    // If any elements remain, handle them in one final vector.
+                    if (i != x.Length)
+                    {
+                        for(int j = 0; j < Vector<float>.Count; j++)
+                        {
+                            incrementArray[j] = x.Length - i;
+                        }
+                        curIndex = Vector.Add(curIndex, AsVector(ref MemoryMarshal.GetReference(incrementArray.AsSpan()), 0));
```
stephentoub, line 378:
```suggestion
                        curIndex = Vector.Add(curIndex, new Vector<int>(x.Length - i));
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
```diff
@@ -2349,6 +2450,346 @@ public Vector<float> Invoke(Vector<float> x, Vector<float> y)
             public Vector<float> Invoke(Vector<float> x, Vector<float> y) => x / y;
         }
 
+        private interface IIndexOfOperator
```
stephentoub, line 2453: Similar comments as in the netcore file

stephentoub (changes requested):

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
michaelgsharp: Yep. Must have been a copy paste error.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
michaelgsharp: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
michaelgsharp: It is, must have been a copy paste error. Removed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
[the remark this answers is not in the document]
tannergooding: `ConditionalSelect` does a `bitwise select`, and thus it is logically `(mask & left) | (~mask & right)`. This is the only way to support the operation downlevel (SSE-SSSE3 capable hardware) and it is possible as a single instruction on Arm64 or on x64 hardware with AVX-512 (using `vpternlog`).

`BlendVariable` on the other hand selects `left` or `right` based solely on the most significant bit of each element. So for a 32-bit `T` it is more akin to `ConditionalSelect((mask >> 32), left, right)` (assuming signed right-shift).

`ConditionalSelect` already emits `BlendVariable` when the JIT can determine that `mask` is a per-element mask (each element is known to be `AllBitsSet` or `Zero`). But there are cases, like here, where the mask is produced earlier or modified in a way that the JIT can no longer track, but where the end user knows that it's still a per-element mask. -- We have similar `*Unsafe` APIs in a couple other places, like `ShuffleUnsafe`, where we do the generally same operation but allow for nuance in edge case handling so that it can remain efficient.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -9231,6 +10103,62 @@ public static Vector512<float> Invoke(Vector512<float> x)
 #endif
         }
 
+        private static Vector128<float> ElementWiseSelect(Vector128<float> mask, Vector128<float> left, Vector128<float> right)
+        {
+            if (Sse41.IsSupported)
+            {
+                return Sse41.BlendVariable(left, right, ~mask);
+            }
+            else return Vector128.ConditionalSelect(mask, left, right);
```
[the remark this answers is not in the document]
tannergooding: > (I personally have a preference for the former.)

Same, at least for this case in particular 😄

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
michaelgsharp: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
```diff
@@ -311,6 +311,93 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x, TMinMaxO
...
+                for (int j = 0; j < Vector<float>.Count; j++)
+                {
+                    resultIndexArray[j] = j;
+                    curIndexArray[j] = j;
+                    incrementArray[j] = Vector<float>.Count;
+                }
+
+                Vector<int> resultIndex = new Vector<int>(resultIndexArray);
+                Vector<int> curIndex = new Vector<int>(curIndexArray);
+                Vector<int> increment = new Vector<int>(incrementArray);
```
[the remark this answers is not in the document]
michaelgsharp: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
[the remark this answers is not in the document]
stephentoub: Most if not all of the other places we're using ConditionalSelect in this file are operating on AllBitsSet/Zero.  Do we think in most/all of those cases the JIT can already see that and do the "right thing"? Or should we be switching some of those over as well?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -9231,6 +10103,62 @@ public static Vector512<float> Invoke(Vector512<float> x)
...
+        }
+
+        private static Vector128<int> ElementWiseSelect(Vector128<int> mask, Vector128<int> left, Vector128<int> right)
```
stephentoub, line 10115: Does this get inlined? If not, should we mark as aggressive inlining?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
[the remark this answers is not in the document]
tannergooding: At a glance they were doing the right thing. I'll do another pass to see if I spot any outliers.

Long term I think we'd look at making `ConditionalSelectUnsafe` exposed publicly and integrate a suggestion to use it (if the input isn't obviously a mask) alongside https://github.com/dotnet/runtime/issues/82488.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
michaelgsharp: Marked it as aggressive inlining.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
michaelgsharp: Nope. Removed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -7824,80 +8079,697 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
+
+            static virtual float IdentityValue => throw new NotSupportedException();
+        }
 
+        /// <summary>MathF.Max(x, y) (but NaNs may not be propagated)</summary>
+        private readonly struct IndexOfMaxOperator : IIndexOfOperator
+        {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
+            public static int Invoke(ReadOnlySpan<float> result)
```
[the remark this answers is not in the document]
michaelgsharp: Refactored it to match how Min/Max do it. Still not fully shared but much better.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -7824,80 +8079,697 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
             {
-                float xMag = MathF.Abs(x), yMag = MathF.Abs(y);
-                return
-                    xMag == yMag ?
-                        (IsNegative(x) ? y : x) :
-                        (xMag > yMag ? x : y);
+                Vector128<float> greaterThanMask = Vector128.GreaterThan(max, current);
+
+                Vector128<float> equalMask = Vector128.Equals(max, current);
+
```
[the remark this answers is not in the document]
michaelgsharp: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -7824,80 +8079,697 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
 
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            public static Vector256<float> Invoke(Vector256<float> x, Vector256<float> y)
+            public static void Invoke(ref Vector256<float> max, Vector256<float> current, ref Vector256<int> maxIndex, Vector256<int> curIndex)
             {
-                Vector256<float> xMag = Vector256.Abs(x), yMag = Vector256.Abs(y);
+                Vector256<float> greaterThanMask = Vector256.GreaterThan(max, current);
+
+                Vector256<float> equalMask = Vector256.Equals(max, current);
+
```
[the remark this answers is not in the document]
michaelgsharp: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -7824,80 +8079,697 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
+                Invoke(ref resultLower256, result.GetUpper(), ref indexLower256, maxIndex.GetUpper());
+
+                // Max the upper/lower halves of the Vector256
+                Vector128<float> resultLower128 = resultLower256.GetLower();
+                Vector128<int> indexLower128 = indexLower256.GetLower();
+
+                Invoke(ref resultLower128, resultLower256.GetUpper(), ref indexLower128, indexLower256.GetUpper());
+
+                return Invoke(resultLower128, indexLower128);
```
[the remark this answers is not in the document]
michaelgsharp: Yeah, Fixed and renamed variables.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -7824,80 +8079,697 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
+            }
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public static void Invoke(ref Vector512<float> max, Vector512<float> current, ref Vector512<int> maxIndex, Vector512<int> curIndex)
+            {
+                Vector512<float> greaterThanMask = Vector512.GreaterThan(max, current);
+
+                Vector512<float> equalMask = Vector512.Equals(max, current);
+
```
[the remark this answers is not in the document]
michaelgsharp: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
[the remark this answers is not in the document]
michaelgsharp: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
```diff
@@ -311,6 +311,94 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x, TMinMaxO
...
+            // It treats +0 as greater than -0 as per the specification.
+
+            int result;
+            int i;
+
+            if (Vector.IsHardwareAccelerated && Vector<int>.Count <= 8 && x.Length >= Vector<float>.Count)
+            {
+                ref float xRef = ref MemoryMarshal.GetReference(x);
+
+                Vector<int> resultIndex = new Vector<int>([0, 1, 2, 3, 4, 5, 6, 7], 0);
```
stephentoub, line 329: Two things:
1. I forgot there's no ReadOnlySpan-based constructor in the netstandard20 surface area of vector. That means this is still going to be allocating a new array on every use.
2. The `,0` is superflous.

Net result is we should cache this array, e.g.:
```C#
private static readonly int[] s_0through7 = [0, 1, 2, 3, 4, 5, 6, 7];
```
and then use it here:
```C#
Vector<int> resultIndex = new Vector<int>(s_0through7);
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -1227,6 +1225,227 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+        }
+#endif
+        private static int IndexOfFirstMatch(Vector256<float> mask)
+        {
+            return BitOperations.TrailingZeroCount(mask.ExtractMostSignificantBits());
+        }
+        private static int IndexOfFirstMatch(Vector128<float> mask)
+        {
+            return BitOperations.TrailingZeroCount(mask.ExtractMostSignificantBits());
+        }
```
stephentoub, line 1448: Nit: space between the methods, and ordered from 128 to 512 as we have elsewhere:
```suggestion
        private static int IndexOfFirstMatch(Vector128<float> mask)
        {
            return BitOperations.TrailingZeroCount(mask.ExtractMostSignificantBits());
        }
        
        private static int IndexOfFirstMatch(Vector256<float> mask)
        {
            return BitOperations.TrailingZeroCount(mask.ExtractMostSignificantBits());
        }
        
#if NET8_0_OR_GREATER
        private static int IndexOfFirstMatch(Vector512<float> mask)
        {
            return BitOperations.TrailingZeroCount(mask.ExtractMostSignificantBits());
        }
#endif
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -7824,6 +8087,514 @@ public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y)
...
+                return maxIndex.ToScalar();
+            }
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            public static void Invoke(ref Vector128<float> max, Vector128<float> current, ref Vector128<int> maxIndex, Vector128<int> curIndex)
+            {
+                Vector128<float> greaterThanMask = Vector128.GreaterThan(max, current);
+
+                Vector128<float> equalMask = Vector128.Equals(max, current);
+               if (equalMask.AsInt32() != Vector128<int>.Zero)
```
stephentoub, line 8127: Nit:
```suggestion
                if (equalMask.AsInt32() != Vector128<int>.Zero)
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
```diff
@@ -311,6 +311,94 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x, TMinMaxO
...
+
+                    return result;
+                }
+            }
+
+        // Scalar path used when either vectorization is not supported, the input is too small to vectorize,
+        // or a NaN is encountered.
+        Scalar:
+            float curResult = x[0];
+            int curIn = 0;
```
stephentoub, line 381: Shouldn't we be starting from i rather than from 0? Otherwise, if we bail out to here from finding a NaN in the last vector, aren't we going to reprocess the entire input?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
```diff
@@ -311,6 +311,94 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x, TMinMaxO
...
+            if (float.IsNaN(curResult))
+            {
+                return curIn;
+            }
+
+            for (i = 1; i < x.Length; i++)
+            {
+                float current = x[i];
+                //float currentAbs = MathF.Abs(current);
```
stephentoub, line 390: What is this?

stephentoub (commented):
Mostly LGTM, but two issues to be addressed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
michaelgsharp: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
michaelgsharp: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
michaelgsharp: Removed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netstandard.cs
michaelgsharp: Didn't have to rework intro logic (as long as our tests cover this case anyways). Just had to initialize `i`, change those `0` to `i`, and modify the forloop slightly.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
michaelgsharp: Fixed.

stephentoub (approved):

PR #93473, opened 2023-10-13
[browser] Trim globalization code

ilonatommy:
In the measurements a [sample WASM](https://github.com/dotnet/runtime/tree/e8b8a77b0c1364e90263bc565d0c0a5552f53dba/src/mono/sample/wasm/browser) app is used with added a few Globalization method calls.

Goal: 
- reduce size of Hybrid Globalization application (sample app: 6 408 kB)
- reduce size of non-Hybrid Globalization application (sample app: 8 428 kB)

Changes:
- correct a bug in `System.Globalization.Hybrid` feature switch substitution,
- globalization-connected task `_GetWasmGenerateAppBundleDependencies` that was setting MsBuild properties for feature switches used to be run **after** `IILink` task. From this reason, all unset MsBuild properties that were planned to be substituted when they are "false" were ignored. Initial feature switches setup was copied from https://github.com/dotnet/runtime/blob/e8b8a77b0c1364e90263bc565d0c0a5552f53dba/src/mono/nuget/Microsoft.NET.Sdk.WebAssembly.Pack/build/Microsoft.NET.Sdk.WebAssembly.Browser.targets#L65
- Blazor trims localizable resources by default, so wasm app should do the same. Because we expect certain exception message in IcuSharding WBT, build setting for that test was changed.
- Trimming tests used to set feature switches that do not have any effect on trimming when we add default values definitions in `WasmApp.targets`. They were changed to property switches.
- Trimming tests `DebuggerTypeProxyAttributeTests` and `DebuggerVisualizerAttributeTests` need to set `DebuggerSupport=true` in order not to have the debugger code trimmed:
https://github.com/dotnet/runtime/blob/04ede0dadfbf2f1082d3bd218cea6fe46bbbed4a/docs/workflow/trimming/feature-switches.md?plain=1#L11
The test used to expect the debugger code to be preserved even though it did not set the property to `true`.
Same thing for `EventSourcePropertyValueTest`, `EventSourceManifestTest`.

Results:
- HG: 6 408 kB -> 6 096 kB (-312 kB)
- non HG: 8 428 kB -> 8 360 kB (-68 kB)

Quoted sizes of `AppBundles` are not brotlied.

### DIFF
--- a/src/libraries/System.Diagnostics.Tracing/tests/TrimmingTests/System.Diagnostics.Tracing.TrimmingTests.proj
+++ b/src/libraries/System.Diagnostics.Tracing/tests/TrimmingTests/System.Diagnostics.Tracing.TrimmingTests.proj
@@ -2,12 +2,8 @@
   <Import Project="$([MSBuild]::GetPathOfFileAbove(Directory.Build.props))" />
 
   <ItemGroup>
-    <TestConsoleAppSourceFiles Include="EventSourcePropertyValueTest.cs" >
-      <EnabledProperties>EventSourceSupport</EnabledProperties>
-    </TestConsoleAppSourceFiles>
-    <TestConsoleAppSourceFiles Include="EventSourceManifestTest.cs" >
-      <EnabledProperties>EventSourceSupport</EnabledProperties>
-    </TestConsoleAppSourceFiles>
+    <TestConsoleAppSourceFiles Include="EventSourcePropertyValueTest.cs" EnabledProperties="EventSourceSupport" />
+    <TestConsoleAppSourceFiles Include="EventSourceManifestTest.cs" EnabledProperties="EventSourceSupport" />
   </ItemGroup>
 
   <Import Project="$([MSBuild]::GetPathOfFileAbove(Directory.Build.targets))" />
--- a/src/libraries/System.Diagnostics.Tracing/tests/TrimmingTests/System.Diagnostics.Tracing.TrimmingTests.proj
+++ b/src/libraries/System.Diagnostics.Tracing/tests/TrimmingTests/System.Diagnostics.Tracing.TrimmingTests.proj
@@ -2,8 +2,12 @@
   <Import Project="$([MSBuild]::GetPathOfFileAbove(Directory.Build.props))" />
 
   <ItemGroup>
-    <TestConsoleAppSourceFiles Include="EventSourcePropertyValueTest.cs" />
-    <TestConsoleAppSourceFiles Include="EventSourceManifestTest.cs" />
+    <TestConsoleAppSourceFiles Include="EventSourcePropertyValueTest.cs" >
+      <EnabledProperties>EventSourceSupport</EnabledProperties>
+    </TestConsoleAppSourceFiles>
+    <TestConsoleAppSourceFiles Include="EventSourceManifestTest.cs" >
+      <EnabledProperties>EventSourceSupport</EnabledProperties>
+    </TestConsoleAppSourceFiles>
   </ItemGroup>
 
   <Import Project="$([MSBuild]::GetPathOfFileAbove(Directory.Build.targets))" />
--- a/src/libraries/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.Browser.xml
+++ b/src/libraries/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.Browser.xml
@@ -1,9 +1,9 @@
 <linker>
   <assembly fullname="System.Private.CoreLib">
-    <type fullname="System.Globalization.GlobalizationMode/Settings">
+    <type fullname="System.Globalization.GlobalizationMode">
       <method signature="System.Boolean get_Hybrid()" body="stub" value="false" feature="System.Globalization.Hybrid" featurevalue="false" />
     </type>
-    <type fullname="System.Globalization.GlobalizationMode/Settings">
+    <type fullname="System.Globalization.GlobalizationMode">
       <method signature="System.Boolean get_Hybrid()" body="stub" value="true" feature="System.Globalization.Hybrid" featurevalue="true" />
     </type>
   </assembly>
--- a/src/libraries/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.Browser.xml
+++ b/src/libraries/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.Browser.xml
@@ -1,10 +1,10 @@
 <linker>
   <assembly fullname="System.Private.CoreLib">
-      <type fullname="System.Globalization.GlobalizationMode">
-      <method signature="System.Boolean get_Hybrid()" body="stub" value="true" feature="System.Globalization.Hybrid" featurevalue="true" />
-    </type>
     <type fullname="System.Globalization.GlobalizationMode/Settings">
       <method signature="System.Boolean get_Hybrid()" body="stub" value="false" feature="System.Globalization.Hybrid" featurevalue="false" />
+    </type>
+    <type fullname="System.Globalization.GlobalizationMode/Settings">
+      <method signature="System.Boolean get_Hybrid()" body="stub" value="true" feature="System.Globalization.Hybrid" featurevalue="true" />
     </type>
   </assembly>
 </linker>
--- a/src/libraries/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.Browser.xml
+++ b/src/libraries/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.Browser.xml
@@ -1,7 +1,10 @@
 <linker>
   <assembly fullname="System.Private.CoreLib">
-    <type fullname="System.Globalization.GlobalizationMode">
-      <method signature="System.Boolean get_Hybrid()" body="stub" value="true" feature="System.Globalization.Hybrid" featurevalue="false" />
+      <type fullname="System.Globalization.GlobalizationMode">
+      <method signature="System.Boolean get_Hybrid()" body="stub" value="true" feature="System.Globalization.Hybrid" featurevalue="true" />
+    </type>
+    <type fullname="System.Globalization.GlobalizationMode/Settings">
+      <method signature="System.Boolean get_Hybrid()" body="stub" value="false" feature="System.Globalization.Hybrid" featurevalue="false" />
     </type>
   </assembly>
 </linker>
--- a/src/mono/wasm/build/WasmApp.targets
+++ b/src/mono/wasm/build/WasmApp.targets
@@ -166,7 +166,6 @@
     <NullabilityInfoContextSupport Condition="'$(NullabilityInfoContextSupport)' == ''">false</NullabilityInfoContextSupport>
     <_AggressiveAttributeTrimming Condition="'$(_AggressiveAttributeTrimming)' == ''">true</_AggressiveAttributeTrimming>
     <DebuggerSupport Condition="'$(DebuggerSupport)' == '' and '$(Configuration)' != 'Debug'">false</DebuggerSupport>
-    <WasmFingerprintDotnetJs Condition="'$(WasmFingerprintDotnetJs)' == ''">false</WasmFingerprintDotnetJs>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/mono/wasm/build/WasmApp.targets
+++ b/src/mono/wasm/build/WasmApp.targets
@@ -155,22 +155,21 @@
 
     <WasmRuntimeAssetsLocation Condition="'$(WasmRuntimeAssetsLocation)' == ''">_framework</WasmRuntimeAssetsLocation>
 
-    <!-- Default values for build properties -->
+    <!-- Runtime feature defaults to trim unnecessary code -->
     <InvariantTimezone Condition="'$(InvariantTimezone)' == ''">false</InvariantTimezone>
+    <HybridGlobalization Condition="'$(HybridGlobalization)' == ''">false</HybridGlobalization>
+    <InvariantGlobalization Condition="'$(InvariantGlobalization)' == ''">false</InvariantGlobalization>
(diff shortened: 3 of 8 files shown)


src/libraries/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.Browser.xml
```diff
@@ -1,7 +1,10 @@
 <linker>
   <assembly fullname="System.Private.CoreLib">
-    <type fullname="System.Globalization.GlobalizationMode">
-      <method signature="System.Boolean get_Hybrid()" body="stub" value="true" feature="System.Globalization.Hybrid" featurevalue="false" />
+      <type fullname="System.Globalization.GlobalizationMode">
```
vitek-karas, line 3: Nit indentation

src/libraries/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.Browser.xml
```diff
@@ -1,7 +1,10 @@
...
+      <method signature="System.Boolean get_Hybrid()" body="stub" value="true" feature="System.Globalization.Hybrid" featurevalue="true" />
+    </type>
+    <type fullname="System.Globalization.GlobalizationMode/Settings">
+      <method signature="System.Boolean get_Hybrid()" body="stub" value="false" feature="System.Globalization.Hybrid" featurevalue="false" />
```
vitek-karas, line 7: This looks weird:
* We have two properties to allow for better trimming, but functionally they should be the same I assume. So why do they have different substitutions?
* I think trimmer should be able to perform constant propagation across the `GlobalizationMode.Hybrid` property, meaning it should only be necessary to substitute the `GlobalizationMode.Settings.Hybrid` property, would need to be tested though.

src/mono/wasm/build/WasmApp.targets
```diff
@@ -155,6 +155,21 @@
 
     <WasmRuntimeAssetsLocation Condition="'$(WasmRuntimeAssetsLocation)' == ''">_framework</WasmRuntimeAssetsLocation>
     <BlazorCacheBootResources Condition="'$(BlazorCacheBootResources)' == ''">true</BlazorCacheBootResources>
+
```
maraf, line 158: Blazor specific props can be omitted here
- BlazorCacheBootResources
- BlazorEnableTimeZoneSupport

src/libraries/System.Private.CoreLib/src/ILLink/ILLink.Substitutions.Browser.xml
ilonatommy: You are right, if we move both stubs to the Settings level then the result is the same.

ilonatommy: My mistake: both should be replaced on the `GlobalizationMode` level, not `GlobalizationMode.Settings` level. This is checked by `InvariantGlobalizationTrue` test and this kind of trimming allows us to remove the whole `Settings` class on Unix.

src/mono/wasm/build/WasmApp.targets
```diff
@@ -154,10 +154,23 @@
...
+    <RuntimeHostConfigurationOption Condition="'$(EventSourceSupport)' == ''" Include="System.Diagnostics.Tracing.EventSource.IsSupported" Value="false" />
+    <RuntimeHostConfigurationOption Condition="'$(UseSystemResourceKeys)' == ''" Include="System.Resources.UseSystemResourceKeys" Value="true" />
+    <RuntimeHostConfigurationOption Condition="'$(EnableUnsafeUTF7Encoding)' == ''" Include="System.Text.Encoding.EnableUnsafeUTF7Encoding" Value="false" />
+    <RuntimeHostConfigurationOption Condition="'$(HttpActivityPropagationSupport)' == ''" Include="System.Net.Http.EnableActivityPropagation" Value="false" />
+    <RuntimeHostConfigurationOption Condition="'$(NullabilityInfoContextSupport)' == ''" Include="System.Reflection.NullabilityInfoContext.IsSupported" Value="false" />
+    <RuntimeHostConfigurationOption Condition="'$(_AggressiveAttributeTrimming)' == ''" Include="System.AggressiveAttributeTrimming" Value="true" />
+    <RuntimeHostConfigurationOption Condition="'$(DebuggerSupport)' == ''and '$(Configuration)' != 'Debug'" Include="System.Diagnostics.Debugger.IsSupported" Value="false" />
```
vitek-karas, line 172: I'm curious about this. I don't know when this happens during the build... but ideally this would happen pretty early on. The reasoning is that other parts of build/publish may look at the values here (ILLink, AOT compilers do, but there might be other parts as well).

The other potential problem is if somebody sets the item group directly. This is something we're not good at in other parts of the SDK, but it is a real scenario. Sometimes people add the value to the `RuntimeHostConfigurationOption` directly, mainly because it's a documented way how to set properties for runtime. We've had bugs in the tools where they got confused by getting the same value specified multiple times. We don't have a good solution how to handle that case, unless everything uses the properties, then it works pretty well.

So, I guess my main question is - why don't we manipulate the properties instead?

For example, for console app trimming, the property values are set to their trimming defaults very early on, so that other parts of the build can look at them and potentially modify the default value. The general recommendation is to use the property values to handle the feature switches everywhere possible (avoids the item group problem mentioned above).

ilonatommy: > "Why don't we manipulate the properties instead?"

I was trying to fix the trimming tests. It "fixed" them but I did not see that I reverted the size changes by that as well, so it is an incorrect change. Why do we have a problem here:
let's take e.g. `InvariantGlobalizationTrue` where we expect the non-invariant parts to be trimmed. We set the invariant mode using feature switch:
https://github.com/dotnet/runtime/blob/0f4196f67a1981dc414d9845629b13e00ce3f4e0/src/libraries/System.Runtime/tests/TrimmingTests/System.Runtime.TrimmingTests.proj#L24
and we expect it will work. However, the flow is:
1) evaluate project's build properties (we do not set `InvariantGlobalization` property, only the switch like this, so it's undefined after this step)
```
  <ItemGroup>
    <RuntimeHostConfigurationOption Include="System.Globalization.Invariant" Value="true" Trim="true" />
  </ItemGroup>
```
2) evaluate runtime build's properties from `WasmApp.targets` (it was undefined, so we set it to `false`, regardless of the feature's value)
3) linker time, the property for `Invariant` is set to false, so we do not trim non-invariant code
4) `InvariantGlobalizationTrue` checks if non-invariant code is there and it throws.

It's only the test's problem, on a sample app it works fine (users normally set properties, not features switches). In order to fix this situation we would could change trimming tests logic to set properties instead of feature switches. But I do not understand why we preferred to use switches over properties in the tests, so I was trying not to touch the current state. I will update the PR to show what I mean exactly.

vitek-karas: If this is an infra only change... then I don't mind at all (I understand it's complicated).
I'm mostly worried if this is changing how a WASM app is built for users.

ilonatommy: Yes, the differences are only for the setup that was leaving properties undefined.
The following switches are passed to linker with different values with this PR:
- `System.AggressiveAttributeTrimming` is true by default now, so it will be a breaking change,
- `System.Globalization.Invariant`, `System.Globalization.Hybrid`, `System.Reflection.NullabilityInfoContext.IsSupported` are false.
Without this PR they were not passed at all.

vitek-karas: Thanks for the explanation

src/libraries/System.Diagnostics.Tracing/tests/TrimmingTests/System.Diagnostics.Tracing.TrimmingTests.proj
```diff
@@ -2,8 +2,12 @@
...
 
   <ItemGroup>
-    <TestConsoleAppSourceFiles Include="EventSourcePropertyValueTest.cs" />
-    <TestConsoleAppSourceFiles Include="EventSourceManifestTest.cs" />
+    <TestConsoleAppSourceFiles Include="EventSourcePropertyValueTest.cs" >
+      <EnabledProperties>EventSourceSupport</EnabledProperties>
+    </TestConsoleAppSourceFiles>
+    <TestConsoleAppSourceFiles Include="EventSourceManifestTest.cs" >
+      <EnabledProperties>EventSourceSupport</EnabledProperties>
+    </TestConsoleAppSourceFiles>
```
radical, line 10:
```suggestion
    <TestConsoleAppSourceFiles Include="EventSourcePropertyValueTest.cs" EnabledProperties="EventSourceSupport" />
    <TestConsoleAppSourceFiles Include="EventSourceManifestTest.cs" EnabledProperties="EventSourceSupport" />
```

src/mono/wasm/build/WasmApp.targets
```diff
@@ -154,7 +154,19 @@
     <WasmStripILAfterAOT Condition="'$(WasmStripILAfterAOT)' == ''">false</WasmStripILAfterAOT>
 
     <WasmRuntimeAssetsLocation Condition="'$(WasmRuntimeAssetsLocation)' == ''">_framework</WasmRuntimeAssetsLocation>
-    <BlazorCacheBootResources Condition="'$(BlazorCacheBootResources)' == ''">true</BlazorCacheBootResources>
+
+    <!-- Runtime feature defaults to trim unnecessary code -->
```
radical, line 158: We should move this to a separate file, and share that between WasmApp.targets, and *Browser.targets, so they don't go out of sync.
Later we'll figure out the two are going to come together.

maraf: We could move them to workload manifest, right?
Or is moving them to a common file and sharing the file easy enough? I would like to unify the bunding early in 9 cycle

radical: I don't think the manifest would be a good place for this. We should use a common file for now, and then see how the unification shapes up, which would give clarity on this.

src/mono/wasm/build/WasmApp.targets
```diff
@@ -154,7 +154,19 @@
...
+    <HybridGlobalization Condition="'$(HybridGlobalization)' == ''">false</HybridGlobalization>
+    <InvariantGlobalization Condition="'$(InvariantGlobalization)' == ''">false</InvariantGlobalization>
+    <EventSourceSupport Condition="'$(EventSourceSupport)' == ''">false</EventSourceSupport>
+    <UseSystemResourceKeys Condition="'$(UseSystemResourceKeys)' == ''">true</UseSystemResourceKeys>
+    <EnableUnsafeUTF7Encoding Condition="'$(EnableUnsafeUTF7Encoding)' == ''">false</EnableUnsafeUTF7Encoding>
+    <HttpActivityPropagationSupport Condition="'$(HttpActivityPropagationSupport)' == ''">false</HttpActivityPropagationSupport>
+    <NullabilityInfoContextSupport Condition="'$(NullabilityInfoContextSupport)' == ''">false</NullabilityInfoContextSupport>
+    <_AggressiveAttributeTrimming Condition="'$(_AggressiveAttributeTrimming)' == ''">true</_AggressiveAttributeTrimming>
+    <DebuggerSupport Condition="'$(DebuggerSupport)' == '' and '$(Configuration)' != 'Debug'">false</DebuggerSupport>
+    <WasmFingerprintDotnetJs Condition="'$(WasmFingerprintDotnetJs)' == ''">false</WasmFingerprintDotnetJs>
```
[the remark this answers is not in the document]
ilonatommy: You mean `WasmFingerprintDotnetJs` only, right? `NullabilityInfoContextSupport`, `_AggressiveAttributeTrimming` and `DebuggerSupport` are used to define feature switches and have influence on how we trim the app. Removing trimming and nullability setup has negative effect on the size. Removing debugger support setup does not influence the size but only today, in the future it might make a difference. For reference, feature switches that by default the linker gets with this PR:
```
--feature Microsoft.Extensions.DependencyInjection.VerifyOpenGenericServiceTrimmability true
--feature System.AggressiveAttributeTrimming true
--feature System.ComponentModel.TypeConverter.EnableUnsafeBinaryFormatterInDesigntimeLicenseContextSerialization false
--feature System.Diagnostics.Debugger.IsSupported false
--feature System.Diagnostics.Tracing.EventSource.IsSupported false
--feature System.Globalization.Invariant false
--feature System.Globalization.Hybrid false
--feature System.Net.Http.EnableActivityPropagation false
--feature System.Reflection.Metadata.MetadataUpdater.IsSupported false
--feature System.Reflection.NullabilityInfoContext.IsSupported false
--feature System.Resources.ResourceManager.AllowCustomResourceTypes false
--feature System.Resources.UseSystemResourceKeys true
--feature System.Runtime.InteropServices.BuiltInComInterop.IsSupported false
--feature System.Runtime.InteropServices.EnableConsumingManagedCodeFromNativeHosting false
--feature System.Runtime.InteropServices.EnableCppCLIHostActivation false
--feature System.Runtime.InteropServices.Marshalling.EnableGeneratedComInterfaceComImportInterop false
--feature System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization false
--feature System.StartupHookProvider.IsSupported false
--feature System.Text.Encoding.EnableUnsafeUTF7Encoding false
--feature System.Text.Json.JsonSerializer.IsReflectionEnabledByDefault true
--feature System.Threading.Thread.EnableAutoreleasePool false
```

src/mono/wasm/build/WasmApp.targets
maraf: Just curiosity, why isn't the workload manifest good place?
What alternative do you suggest? Should we link the file to SDK pack and workload include it in both places?

radical: > Just curiosity, why isn't the workload manifest good place?

The manifest is not just for wasm, and it doesn't really belong there. But have it there for now. We'll have some refactoring happening soon anyway.

maraf (approved):
Looks good to me 👍

PR #93475, opened 2023-10-13
[docs] Howto configure the wasm runtime in Blazor projects in net8.0

lambdageek:


docs/workflow/debugging/mono/wasm-debugging.md
```diff
@@ -257,3 +257,87 @@ C @ blazor.webassembly.js:1
...
+        Blazor.start({
+            configureRuntime: dotnet => {
+                dotnet.withEnvironmentVariable("MONO_LOG_LEVEL", "debug");
+                dotnet.withEnvironmentVariable("MONO_LOG_MASK", "all");
+            }
+        });
+    </script></body>
+```
+
+## Blazor
```
maraf, line 314: Is it worth having two snippets if they diff only in script name? What about just mentioning that?

maraf (approved):

docs/workflow/debugging/mono/wasm-debugging.md
lambdageek: That's not the only difference. the argument to `Blazor.start` is different.  the `blazor` project needs an extra `webAssembly:` dictionary

maraf: Thanks! I missed that

PR #93490, opened 2023-10-13
Support BBJ_EHFINALLYRET successors in profile synthesis

BruceForstall:


BruceForstall:
@AndyAyersMS PTAL
cc @dotnet/jit-contrib

AndyAyersMS:
This is ok for consistency's sake, but ultimately, we'll want to determine these likelihoods by looking at all the callfinallies. That is, we won't set these during the initial assignments as they won't affect the profiles we come up with; they can be set after the fac.

BruceForstall:
I was implementing it just to be consistent. If you think that's incorrect for the intended semantics, I can close this.

BruceForstall:
> This is ok for consistency's sake, but ultimately, we'll want to determine these likelihoods by looking at all the callfinallies. That is, we won't set these during the initial assignments as they won't affect the profiles we come up with; they can be set after the fac.

I guess I don't understand the algorithm used here. by "looking at the callfinallies", do you mean the synthetic count of the call finally continuation should be the same as that of the corresponding call finally? I guess that makes more sense than what this is doing, which is distributing all the calls to the finally among the various call finallies.

(Presumably we would assume that every finally returns and doesn't throw.)

AndyAyersMS:
Yeah, we would look at the callfinallies counts and match up those with the retfinally edges. EG if there were two callfinallies, say A with 90 counts and B with 10, then the edge from the retfinally to A's pair tail would have 90% likelihood.

But I don't believe we'll ever act on this information. If we were to make finally cloning be profile aware, we'd probably look at the weights of the incoming edges to the finally begin, instead of the outgoing edges from the finally end.

BruceForstall:
One thing that confused me is here:

https://github.com/dotnet/runtime/blob/69a0ce0357715ed9fce6819ccf6a5f9768d055db/src/coreclr/jit/fgprofilesynthesis.cpp#L146-L152

The comment says "Note we handle flow to the finally specially; this represents return from the finally" which I think is trying to implement what we described here in this issue that the continuation likelihood should match the callfinally likelihood. But the code is broken because a BBJ_ALWAYS pair of BBJ_CALLFINALLY shouldn't have a pred edge to the BBJ_CALLFINALLY (it will have a pred edge to the BBJ_EHFINALLYRET), so the code should crash. But it turns out this code never executes when BBJ_CALLFINALLY exists: it only runs (currently, under config) very very early, when we have LEAVE but haven't converted the LEAVE to the other EH flow. That makes the synthesized data suspect for anything related to EH flow, presumably.

AndyAyersMS:
Right we don't have call finallys yet, flow into finallys is handled specially (since they don't have preds yet), flow out of finallys is ignored (and leaves in trys that will become callfinallys just propagate flow to their targets, which will become the pair tails), and flow into and out of catches and filters is ignored (and we assume all such are cold).

I don't think anything is fundamentally broken here, but if we were to try and rerun synthesis later we'd want to handle these cases differently.

PR #93492, opened 2023-10-13
Enable Adding Disabled Tests Skip Reason to their XML Results Log From issues.targets

### LINKED ISSUE #91562: CoreCLR test infra: improve "Skip" messages for tests blocked in issues.targets
Assume a test under the merged system is blocked with an issue using the file:

```src/tests/issues.targets```

Today, the infrastructure ends up with the following snippet in the xml log:

<pre>
&lt;test name="baseservices\threading\regressions\2164\foreground-shutdown\foreground-shutdown.cmd" type="OutOfProcessTest" method="baseservices\threading\regressions\2164\foreground-shutdown\foreground-shutdown.cmd" time="0.000000" result="Skip">
    <reason>&lt;![CDATA[No Known Skip Reason]]>&lt;/reason>
&lt;/test>
</pre>

corresponding to the following section in the issues.targets file:

<pre>
        &lt;ExcludeList Include="$(XunitTestBinBase)/baseservices/threading/regressions/2164/foreground-shutdown/*">
            &lt;Issue>https://github.com/dotnet/runtime/issues/84006&lt;/Issue>
        &lt;/ExcludeList>
</pre>

I believe we should fix the merged wrapper generator support code and test
build scripts to somehow transport the reasoning string from the issues.targets
file to the output log to make it easier to understand the skip reasons.

Thanks

Tomas

/cc @dotnet/runtime-infrastructure @jkoritzinsky

### PR DESCRIPTION
**Pull request by ivdiazsa**
Feature that fixes issue #91562. This PR adds the necessary functionality to write the reason any given test was skipped in their results _xml_ file.

Currently, it is hard-coded to the message _"No Known Skip Reason"_. With this PR, the reason will be taken from the test's disable explanation in `src/tests/issues.targets`, in their respective `<Issue>` tag. If there is no such message, then we will default to today's message of _"No Known Skip Reason"_.

A good amount of code cleanup was done as part of this PR as well. Due to how Github handles file differences and spacing, especially with long lines, I highly recommend using the *Unified* view instead of the *Split* one for reviewing this one PR.

### DIFF
--- a/src/tests/Common/XUnitWrapperGenerator/ITestInfo.cs
+++ b/src/tests/Common/XUnitWrapperGenerator/ITestInfo.cs
@@ -169,6 +169,7 @@ public sealed class ConditionalTest : ITestInfo
 
         using (builder.NewBracesScope())
         {
+            builder.AppendLine("string reason = string.Empty;");
             builder.AppendLine(testReporterWrapper.GenerateSkippedTestReporting(_innerTest));
         }
         return builder;
@@ -448,13 +449,15 @@ public sealed class WrapperLibraryTestSummaryReporting : ITestReporterWrapper
 
             using (builder.NewBracesScope())
             {
-                builder.AppendLine($"{_summaryLocalIdentifier}.ReportStartingTest({test.TestNameExpression},"
+                builder.AppendLine($"{_summaryLocalIdentifier}.ReportStartingTest("
+                                 + $"{test.TestNameExpression},"
                                  + $" System.Console.Out);");
 
                 builder.AppendLine($"{_outputRecorderIdentifier}.ResetTestOutput();");
                 builder.Append(testExecutionExpression);
 
-                builder.AppendLine($"{_summaryLocalIdentifier}.ReportPassedTest({test.TestNameExpression},"
+                builder.AppendLine($"{_summaryLocalIdentifier}.ReportPassedTest("
+                                 + $"{test.TestNameExpression},"
                                  + $" \"{test.ContainingType}\","
                                  + $" @\"{test.Method}\","
                                  + $" stopwatch.Elapsed - testStart,"
@@ -468,7 +471,8 @@ public sealed class WrapperLibraryTestSummaryReporting : ITestReporterWrapper
 
             using (builder.NewBracesScope())
             {
-                builder.AppendLine($"{_summaryLocalIdentifier}.ReportFailedTest({test.TestNameExpression},"
+                builder.AppendLine($"{_summaryLocalIdentifier}.ReportFailedTest("
+                                 + $"{test.TestNameExpression},"
                                  + $" \"{test.ContainingType}\","
                                  + $" @\"{test.Method}\","
                                  + $" stopwatch.Elapsed - testStart,"
@@ -484,6 +488,8 @@ public sealed class WrapperLibraryTestSummaryReporting : ITestReporterWrapper
 
         using (builder.NewBracesScope())
         {
+            builder.AppendLine($"string reason = {_filterLocalIdentifier}"
+                             + $".GetTestExclusionReason({test.TestNameExpression});");
             builder.AppendLine(GenerateSkippedTestReporting(test));
         }
         return builder;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/tests/Common/XUnitWrapperLibrary/TestFilter.cs [hunk 1]
```diff
@@ -191,30 +210,83 @@ public bool ShouldRunTest(string fullyQualifiedName, string displayName, string[
...
+               : string.Empty;
+    }
+
+    // GH dotnet/runtime issue #91562: Some tests are purposefully not run for a number
+    // of reasons. They are specified in src/tests/issues.targets, along with a brief
+    // explanation on why they are skipped inside an '<Issue>' tag.
+    //
+    // When building any test or test subtree, if any of the tests built matches an entry
+    // in issues.targets, then the exclusions list file ($CORE_ROOT/TestExclusions.txt is
+    // the default) is updated by adding a new a comma-separated entry:
```
trylek, line 236: IIRC they are actually semicolon / CRLF-separated.

trylek: Hmm, maybe I just misunderstood this and the comma separation is a new addition of yours, I continue looking...

trylek (approved):
LGTM in general, I believe it's good to go it if works in the lab.

src/tests/Common/XUnitWrapperLibrary/TestFilter.cs
ivdiazsa: Oh no, this is the `TestExclusions.txt` file. It used to be just one test per line, and I modified it to also include the reason with a comma. As for line endings, we're now using the platform's one (i.e. Linux/MacOS LF and Windows CRLF).

ivdiazsa:
Ok there's a bug with with certain cases where we have no specified reason for skipping a test. Will get to work on a fix for it.

src/tests/Common/XUnitWrapperGenerator/ITestInfo.cs
```diff
@@ -72,7 +80,7 @@ public override int GetHashCode()
         hash = hash * 23 + (TestNameExpression?.GetHashCode() ?? 0);
         hash = hash * 23 + (Method?.GetHashCode() ?? 0);
         hash = hash * 23 + (ContainingType?.GetHashCode() ?? 0);
-        hash = hash * 23 + (ExecutionStatement?.GetHashCode() ?? 0);
+        hash = hash * 23 + (_executionStatement?.GetHashCode() ?? 0);
```
danmoseley, line 83: Can you use HashCode.Combine ?

src/tests/Common/XUnitWrapperLibrary/TestFilter.cs
```diff
@@ -191,30 +210,83 @@ public bool ShouldRunTest(string fullyQualifiedName, string displayName, string[
...
+    // When a test runner is executed (e.g. Methodical_d1.sh), it uses a compiler-generated
+    // source file to actually run the tests (e.g. FullRunner.g.cs - This is detailed in
+    // XUnitWrapperGenerator.cs). This generated source file is the one in charge of
+    // reading the test exclusions list file, and stores the comma-separated values into a
+    // table represented by the dictionary called _testExclusionTable in this file.
+
+    public static Dictionary<string, string> LoadTestExclusionTable()
     {
-        HashSet<string> output = new ();
+        Dictionary<string, string> output = new Dictionary<string, string>();
```
danmoseley, line 249: Explicit comparer?

src/tests/Common/XUnitWrapperGenerator/ITestInfo.cs
jkoritzinsky: This code is in a source generator which targets .NET Standard 2.0.

danmoseley: right, you'd have to have a package reference to the OOB. probably not worth it I guess.

jkoritzinsky: Referencing OOB dependencies for Roslyn components is very annoying to get right. I’d recommend against it unless necessary tbh.

ivdiazsa:
Failures are unrelated so merging this feature now.

Issue #76257, opened 2022-09-27
signal SIGFPE during garbage collecting 


grospelliergilles:
### Description

I have an application which use `coreclr` embedded in a C++ program.
In this application, I need to activate floating point exception to detected invalid mathematical operations (for example calling `log` with a negative argument). I use the C [feenableexcept](https://linux.die.net/man/3/feenableexcept) function to enable exceptions:

~~~{cpp]
#include <fenv.h>
feenableexcept(FE_DIVBYZERO|FE_INVALID)
~~~

After calling this method, I receive a SIGFPE when executing C# code when the garbage collector is running.

### Reproduction Steps

I can not provide a small case to reproduce the problem because my project is too big.

### Expected behavior

Not throwing signal SIGFPE 

### Actual behavior

The application sends a SIGFPE signal.
The debugger give me the following stack trace:

~~~{cpp}
WKS::gc_heap::dt_high_frag_p(WKS::gc_tuning_point, int, int) at /__w/1/s/src/coreclr/gc/gc.cpp:3174:101
WKS::gc_heap::generation_to_condemn(int, int*, int*, int) at /__w/1/s/src/coreclr/gc/gc.cpp:0:13
WKS::gc_heap::garbage_collect(int) at /__w/1/s/src/coreclr/gc/gc.cpp:22458:39
WKS::GCHeap::GarbageCollectGeneration(unsigned int, gc_reason) at /__w/1/s/src/coreclr/gc/gc.cpp:46902:13
WKS::GCHeap::GarbageCollect(int, bool, int) at /__w/1/s/src/coreclr/gc/gc.cpp:46080:52
Volatile<unsigned int>::LoadWithoutBarrier() const at /__w/1/s/src/coreclr/inc/volatile.h:383:31
  (inlined by) Thread::PreemptiveGCDisabled() at /__w/1/s/src/coreclr/vm/threads.h:1837:40
  (inlined by) void GCHolderBase::PopInternal<1>() at /__w/1/s/src/coreclr/vm/threads.h:0:0
  (inlined by) GCCoop::~GCCoop() at /__w/1/s/src/coreclr/vm/threads.h:5363:15
  (inlined by) GCInterface_Collect at /__w/1/s/src/coreclr/vm/comutilnative.cpp:894:5
~~~

and  the corresponding line in the source file `gc.cpp` is:

~~~
    float frag_ratio = (float)(dd_fragmentation (dynamic_data_of (max_generation))) / (float)generation_size (max_generation);
[...]

PR #93517, opened 2023-10-14
Fix potential 'Floating Point Exception' with speculative execution

grospelliergilles:
With the current version of runtime and version `8.0-rc2`, there is a 'Floating Point Exception' (signal `SIGFPE`) when floating point exception are enabled (via `feenableexcept()`).

It seems to be due to speculative execution of the division in the check.

This patch remove the check and make sure we do not divide by zero.

Alternatively, because the result of this function is only to check if the fragmentation is high enough, we can rewrite the part of the code to remove all divide operation.

### DIFF
--- a/src/coreclr/gc/gc.cpp
+++ b/src/coreclr/gc/gc.cpp
@@ -3416,9 +3416,9 @@ gc_heap::dt_high_frag_p (gc_tuning_point tp,
                     fragmentation_burden = (gen_size ? ((float)fr / (float)gen_size) : 0.0f);
                     ret = (fragmentation_burden > dd_v_fragmentation_burden_limit (dd));
                 }
-                dprintf (GTC_LOG, ("h%d: gen%d, frag is %zd, alloc effi: %d%%, unusable frag is %zd, ratio is %d",
+                dprintf (GTC_LOG, ("h%d: gen%d, frag is %zd, alloc effi: %zu%%, unusable frag is %zd, ratio is %d",
                     heap_number, gen_number, dd_fragmentation (dd),
-                    (int)(generation_allocator_efficiency_percent (generation_of (gen_number))),
+                    generation_allocator_efficiency_percent (generation_of (gen_number)),
                     fr, (int)(fragmentation_burden*100)));
             }
             break;
@@ -33143,9 +33143,9 @@ void gc_heap::plan_phase (int condemned_gen_number)
         if ((free_list_allocated + rejected_free_space) != 0)
             free_list_efficiency = (int)(((float) (free_list_allocated) / (float)(free_list_allocated + rejected_free_space)) * (float)100);
 
-        int running_free_list_efficiency = (int)(generation_allocator_efficiency_percent(older_gen));
+        size_t running_free_list_efficiency = generation_allocator_efficiency_percent(older_gen);
 
-        dprintf (1, ("gen%d free list alloc effi: %d%%, current effi: %d%%",
+        dprintf (1, ("gen%d free list alloc effi: %d%%, current effi: %zu%%",
                     older_gen->gen_num,
                     free_list_efficiency, running_free_list_efficiency));
 
--- a/src/coreclr/gc/gc.cpp
+++ b/src/coreclr/gc/gc.cpp
@@ -3418,7 +3418,7 @@ gc_heap::dt_high_frag_p (gc_tuning_point tp,
                 }
                 dprintf (GTC_LOG, ("h%d: gen%d, frag is %zd, alloc effi: %d%%, unusable frag is %zd, ratio is %d",
                     heap_number, gen_number, dd_fragmentation (dd),
-                    (int)(100*generation_allocator_efficiency (generation_of (gen_number))),
+                    (int)(generation_allocator_efficiency_percent (generation_of (gen_number))),
                     fr, (int)(fragmentation_burden*100)));
             }
             break;
@@ -22420,7 +22420,7 @@ void gc_heap::gc1()
             }
         }
 
-        get_gc_data_per_heap()->maxgen_size_info.running_free_list_efficiency = (uint32_t)(generation_allocator_efficiency (generation_of (max_generation)) * 100);
+        get_gc_data_per_heap()->maxgen_size_info.running_free_list_efficiency = (uint32_t)(generation_allocator_efficiency_percent (generation_of (max_generation)));
 
         free_list_info (max_generation, "after computing new dynamic data");
     }
@@ -33143,7 +33143,7 @@ void gc_heap::plan_phase (int condemned_gen_number)
         if ((free_list_allocated + rejected_free_space) != 0)
             free_list_efficiency = (int)(((float) (free_list_allocated) / (float)(free_list_allocated + rejected_free_space)) * (float)100);
 
-        int running_free_list_efficiency = (int)(generation_allocator_efficiency(older_gen)*100);
+        int running_free_list_efficiency = (int)(generation_allocator_efficiency_percent(older_gen));
 
         dprintf (1, ("gen%d free list alloc effi: %d%%, current effi: %d%%",
                     older_gen->gen_num,
--- a/src/coreclr/gc/gcpriv.h
+++ b/src/coreclr/gc/gcpriv.h
@@ -5104,25 +5104,30 @@ size_t& generation_allocated_since_last_pin (generation* inst)
 }
 #endif //FREE_USAGE_STATS
 
+// Return the percentage of efficiency (between 0 and 100) of the allocator.
 inline
-uint64_t generation_allocator_efficiency_percent (generation* inst)
+size_t generation_allocator_efficiency_percent (generation* inst)
 {
+    // Use integer division to prevent potential floating point exception.
+    // FPE may occur if we use floating point division because of speculative execution.
     uint64_t free_obj_space = generation_free_obj_space (inst);
     uint64_t free_list_allocated = generation_free_list_allocated (inst);
-    if (free_list_allocated==0)
+    if (free_list_allocated == 0)
       return 0;
-    return (100 * free_list_allocated) / (free_list_allocated + free_obj_space);
+    return (size_t)((100 * free_list_allocated) / (free_list_allocated + free_obj_space));
 }
 
 inline
-uint64_t generation_unusable_fragmentation (generation* inst)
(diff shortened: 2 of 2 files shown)


tannergooding:
Running with floating-point exceptions enabled is explicitly not supported, the behavior that occurs if it is enabled is undefined.

CC. @jkotas

grospelliergilles:
Thanks for your response.
I already had this kind of error in old versions of the runtime (#76257) and this was fixed.
In my application I do not have control on enabling floating point exception. And because it occurs in the garbage collector  I can not disable them in this part.
The fix is pretty simple and should not change the results. Do you think it could be merged ?

mangod9:
@Maoni0 as well.

Maoni0:
this is a different issue from the fix I made in https://github.com/dotnet/runtime/pull/76294/. if a speculative execution can result in SIGFPE (which seems quite dangerous to me), that means the fix I made would subject to the same problem. 

we should decide whether we want to support this speculative execution scenario at all - if we do, we should harden all the places this can happen instead of making one off fixes like this. 

I don't have a strong opinion on whether to support it or not, @jkotas, do you?

jkotas:
As I have said in https://github.com/dotnet/runtime/issues/76257#issuecomment-1260226375, .NET runtime and libraries are not compatible with enabled floating point exceptions. If you need floating point exceptions enabled for your C/C++ library, you need to disable around them around the calls to .NET runtime.

I do not think that one-off workarounds like the one proposed in this PR make sense.

grospelliergilles:
Thansk @jkotas for your response. It is difficult to disable FPE around this part of the code because it is called by the garbage collector and I do not know when it will be called. I will try to use `GC.RegisterForFullGCNotification()` to see if it works.

Alternatively, I think the code may be rewritten without using division because the result is only used in comparison. So instead of checking `a>b/c` we can rewrite as `a*c>b`. If you agree I can provide a patch to remove them if you want.

jkotas:
> So instead of checking a>b/c we can rewrite as a*c>b

`generation_allocator_efficiency` is called in number of places. I do not see how you can do this rewrite.

grospelliergilles:
I have removed floating point usage in the two functions which can throw FPE.
Using only integer division there is no longer potential FPE.

grospelliergilles:
> @grospelliergilles please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement
@dotnet-policy-service agree

grospelliergilles:
@dotnet-policy-service agree

src/coreclr/gc/gcpriv.h
```diff
@@ -5105,23 +5105,24 @@ size_t& generation_allocated_since_last_pin (generation* inst)
...
-    // It will not change the result if sum is not zero because in this
-    // case sum>1.0 and precision of float is limited to 7 digits
-
-    float sum = (float)(generation_free_list_allocated (inst) + generation_free_obj_space (inst));
-    return ((float) (generation_free_list_allocated (inst))) / (sum + 1.0e-10f);
+    size_t free_obj_space = generation_free_obj_space (inst);
+    size_t free_list_allocated = generation_free_list_allocated (inst);
+    if (free_list_allocated==0)
+      return 0;
+    return (100 * free_list_allocated) / (free_list_allocated + free_obj_space);
```
jkotas, line 5114: Can this hit integer overflow?

grospelliergilles: if `size_t` is 64bit, it can not overflow because `free_list_allocated` will never be greater than `2^60`.
I can use `uint64_t` if you prefer.

mangod9: @grospelliergilles, do you plan to update the PR with `uint64_t` and perhaps we can evaluate again.

grospelliergilles: Thanks for you reply @mangod9. Yes I will update my PR if you want. I will do that this week

mangod9:
@grospelliergilles, there appear to be build breaks on x86. @Maoni0 please review as well.

src/coreclr/gc/gcpriv.h
```diff
@@ -5105,20 +5105,24 @@ size_t& generation_allocated_since_last_pin (generation* inst)
...
-    return (size_t)(generation_free_obj_space (inst) +
-                    (1.0f-generation_allocator_efficiency(inst))*generation_free_list_space (inst));
+uint64_t generation_unusable_fragmentation (generation* inst)
+{
+    uint64_t free_obj_space = generation_free_obj_space (inst);
+    uint64_t free_list_allocated = generation_free_list_allocated (inst);
+    uint64_t free_list_space = generation_free_list_space (inst);
+    if (free_obj_space==0)
+      return 0;
+    return (free_obj_space + (free_obj_space * free_list_space) / (free_list_allocated + free_obj_space));
```
markples, line 5125: This could use a comment, either that it is based on rewriting of `(100 - efficiency)/100 * free_list_space` or `inefficiency * free_list_space`, being careful in both cases to minimize the impact of the integer division.

markples (commented):
I feel like there is some readability loss here.  Perhaps it can be addressed (see my specific remark on a comment that could help).  Also, I think it would help to limit the use of `uint64_t` to where the large values can occur.  `generation_allocator_efficiency_percent` can have an intermediate large value but returns a percentage.  `generation_unusable_fragmentation` also can have an intermediate large value but returns a size (in fact, this is the cause of the build break because the call site is appropriately expecting a `size_t`.

However, given this, I wonder if adding something like FLT_EPSILON to the denominator would be a clearer fix.  I thought about suggesting that `generation_unusable_fragmentation` continue to call `generation_allocator_efficiency_percent` with appropriate scaling, but I see that you were careful to avoid rounding errors due to integer truncation and I don't think we want to try to analyze if percent, per mille, etc., would be sufficient.

src/coreclr/gc/gcpriv.h
grospelliergilles: Thanks for your report. I will make the required change and add the information your want.

Maoni0: I would really rather to not have to have the presence of FLT_EPSILON in the GC code at all if I can avoid it :) it's just another thing that someone who's not familiar with FP has to care about. 

but I agree with @markples that these should continue to return size_t, instead of uint64_t. I would also add a comment above `generation_allocator_efficiency_percent` to explain why this is written this way (to accommodate this exception specific to FP which is not a common knowledge AFAIK) instead of the previous way which seems more natural.

src/coreclr/gc/gcpriv.h
```diff
@@ -5105,20 +5105,24 @@ size_t& generation_allocated_since_last_pin (generation* inst)
...
-size_t generation_unusable_fragmentation (generation* inst)
-{
-    return (size_t)(generation_free_obj_space (inst) +
-                    (1.0f-generation_allocator_efficiency(inst))*generation_free_list_space (inst));
+uint64_t generation_unusable_fragmentation (generation* inst)
+{
+    uint64_t free_obj_space = generation_free_obj_space (inst);
+    uint64_t free_list_allocated = generation_free_list_allocated (inst);
+    uint64_t free_list_space = generation_free_list_space (inst);
+    if (free_obj_space==0)
```
Maoni0, line 5123:
```suggestion
    if ((free_list_allocated + free_obj_space) == 0)
```

markples: I think the idea is that if free_obj_space is zero, then the computation will be zero regardless of the free_list_allocated value.  (unless one can be negative...)  It's subtle and perhaps not worth the microoptimization or worth a comment.  The same patterns occurs in the efficiency method.

grospelliergilles: Yes if `free_obj_space` is zero then the result is always zero. The operands are of type `size_t` and so can not be negative.

Maoni0: yes, I got the idea. but this kind of microoptimization is not worth it at all. I'd much rather have the code be easier to understand.

grospelliergilles: Thanks for your report. I will do as requested.

src/coreclr/gc/gc.cpp
```diff
@@ -3418,7 +3418,7 @@ gc_heap::dt_high_frag_p (gc_tuning_point tp,
                 }
                 dprintf (GTC_LOG, ("h%d: gen%d, frag is %zd, alloc effi: %d%%, unusable frag is %zd, ratio is %d",
                     heap_number, gen_number, dd_fragmentation (dd),
-                    (int)(100*generation_allocator_efficiency (generation_of (gen_number))),
+                    (int)(generation_allocator_efficiency_percent (generation_of (gen_number))),
```
jkotas, line 3421: All places that call `generation_allocator_efficiency_percent` cast the result to int or uint32_t. Should the result be int or uint32_t to avoid these casts?

grospelliergilles: Yes I can to that. I did not change it in previous commits because of flags in `printf` but I can do the change if you want. The result is always positive. Do you prefer `int` or `uint32_t` ?

jkotas: @Maoni0 Do you have a preference?

grospelliergilles:
> @grospelliergilles is this PR ready to review again?

I have fixed compile errors so I think it is OK.

mangod9:
@Maoni0 can you please take another look?

Maoni0 (approved):
LGTM

grospelliergilles:
Thank you very much @mangod9 for accepting my PR.

PR #93521, opened 2023-10-14
JIT: Optimize constant range tests

EgorBo:
```cs
void Test(char c)
{
    if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z'))
        Console.WriteLine("a-zA-Z");

// also: "X < CNS1 || X > CNS2", etc
}
```
Current codegen:
```asm
; Assembly listing for method Prog:Test(ushort):this (FullOpts)
       movzx    rcx, dx
       cmp      ecx, 97
       jl       SHORT G_M56518_IG04
       cmp      ecx, 122
       jle      SHORT G_M56518_IG05
G_M56518_IG04:
       cmp      ecx, 65
       jl       SHORT G_M56518_IG07
       cmp      ecx, 90
       jg       SHORT G_M56518_IG07
G_M56518_IG05:
       mov      rcx, 0x13A00208D40      ; 'a-zA-Z'
       tail.jmp [System.Console:WriteLine(System.String)]
G_M56518_IG07:
       ret      
; Total bytes of code 40
```
New codegen:
```asm
; Assembly listing for method Prog:Test(ushort):this (FullOpts)
       movzx    rcx, dx
       mov      eax, ecx
       sub      eax, 97
       cmp      eax, 25
       jbe      SHORT G_M56518_IG04
       sub      ecx, 65
       cmp      ecx, 25
       ja       SHORT G_M56518_IG06
G_M56518_IG04:
       mov      rcx, 0x1E200008D40      ; 'a-zA-Z'
G_M56518_IG05:
       tail.jmp [System.Console:WriteLine(System.String)]
G_M56518_IG06:
       ret      
; Total bytes of code 38
```

Not a big size win (sometimes even a size regression actually), but less branches (takes PGO into account).

Alternative patterns:

- `c < 'a' || c > 'z'`
- `c is >= 'a' and <= 'z' or >= 'A' and <= 'Z'`

### DIFF
--- a/src/coreclr/jit/optimizebools.cpp
+++ b/src/coreclr/jit/optimizebools.cpp
@@ -405,6 +405,7 @@ bool OptBoolsDsc::FindCompareChain(GenTree* condition, bool* isTestCondition)
     return false;
 }
 
+// Does block represent a conditional branch with a constant range test? E.g. "if (X > 10)"
 static bool IsConstantRangeTest(const BasicBlock* block, GenTree** lcl, GenTreeIntCon** cns, genTreeOps* cmpOp)
 {
     // NOTE: caller is expected to check that a block has multiple statements or not
@@ -412,6 +413,7 @@ static bool IsConstantRangeTest(const BasicBlock* block, GenTree** lcl, GenTreeI
     assert(block->KindIs(BBJ_COND) && rootNode->OperIs(GT_JTRUE));
     const GenTree* tree = rootNode->gtGetOp1();
 
+    // We're only interested in continuous range tests
     if (!tree->OperIs(GT_LE, GT_LT, GT_GT, GT_GE) || tree->IsUnsigned())
     {
         return false;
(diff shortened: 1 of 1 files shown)


EgorBo:
@AndyAyersMS @jakobbotsch @dotnet/jit-contrib PTAL, [diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=439167&view=ms.vss-build-web.run-extensions-tab) are not too big because we only save 2 bytes on average, but we get rid of a branch, so, hopefully, it will be noticeable. Outerloop/Jitstress/fuzzlyn pass. Can be separately extended to:

- `if (x >= 0 && x < arr.Length)` -> `if ((uint)x < arr.Length)` (if arr is not null)
- `if (x > 10 && x > 100)` -> `if (x > 100)` (although, for this one it's better to be done in RedundantBranchOpts)

Motivated by https://github.com/ricomariani/wrath-othello benchmark.

src/coreclr/jit/optimizebools.cpp
```diff
@@ -404,6 +405,290 @@ bool OptBoolsDsc::FindCompareChain(GenTree* condition, bool* isTestCondition)
     return false;
 }
 
+// Given two ranges, return true if they intersect and form a closed range.
```
AndyAyersMS, line 408: can you give these (and the ones below) the normal header comments?

AndyAyersMS (approved):
Nice.

I think it would be a bit more challenging to fit this sort of thing into RBO because it modifies the upper branch and not the lower one. But still worth doing someday.

EgorBo:
Failure is https://github.com/dotnet/runtime/issues/93527

EgorBo:
> I assume it isn't worth the effort to detect+apply the bit fiddling to map lower to upper case (masking 1 bit) and then check a single range, like in the new ascii helpers? That would also remove the branch, but at another level.

I am not sure I understood, can you explaind it via code? 🙂

PR #93523, opened 2023-10-14
Workaround for #93442

jkotas:


src/coreclr/jit/utils.h
```diff
@@ -1080,6 +1084,10 @@ bool MulOverflows(T x, T y, bool unsignedMul)
     }
 }
 
+#if defined(_MSC_VER) && defined(HOST_ARM64) // Workaround for https://github.com/dotnet/runtime/issues/93442
+#pragma optimize("", on)
+#endif
+
```
jakobbotsch, line 1090: `ClrSafeInt::operator*` and `ClrSafeInt::multiply` probably also will need to have optimizations disabled. Presumably otherwise MSVC will generate a call to the optimized versions of `operator*`/`multiply` that have the same bug.

jakobbotsch (approved):

jkotas:
Opened https://github.com/dotnet/runtime/issues/93526 to remove the workaround once the compiler is fixed.

PR #93524, opened 2023-10-15
Allows CustomAttribute table to be unsorted in EnC delta

tmat:
Fix https://github.com/dotnet/roslyn/pull/70350 of https://github.com/dotnet/roslyn/issues/70357 introduces a workaround to Roslyn, which is to use Reflection to avoid sorting of Custom Attribute table in EnC delta.

This change allows us to remove the workaround when Roslyn updates to the latest S.R.M. -- tracked by https://github.com/dotnet/roslyn/issues/70389


jkoritzinsky (changes requested):
Can you also update the "is sorted" bitflag to accurately represent if the CustomAttrivute table is sorted?

jkoritzinsky (approved):
LGTM!

PR #93531, opened 2023-10-15
Add GT_ARR_LEN to optNonNullAssertionProp_Ind

### LINKED ISSUE #84248: Redundant nullcheck in ARR_LEN/IND on top of non-null array

EgorBo:
Closes https://github.com/dotnet/runtime/issues/84248
We were not taking advantage of assert propagation for GT_ARR_LENGTH (and MD_LENGHT).

```cs
void Test(int[]? arr)
{
    if (arr != null)
    {
        var _ = arr.Length;
    }
}
```
```diff
; Assembly listing for method Prog:Test(int[]):this (FullOpts)
-      test     rdx, rdx
-      je       SHORT G_M2580_IG04
-      mov      eax, dword ptr [rdx+0x08]
-G_M2580_IG04:
       ret      
-; Total bytes of code 9
+; Total bytes of code 1
```

### DIFF
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -4454,14 +4454,14 @@ GenTree* Compiler::optNonNullAssertionProp_Call(ASSERT_VALARG_TP assertions, Gen
 //
 bool Compiler::optNonNullAssertionProp_Ind(ASSERT_VALARG_TP assertions, GenTree* indir)
 {
-    assert(indir->OperIsIndir() || indir->OperIs(GT_ARR_LENGTH));
+    assert(indir->OperIsIndir() || indir->OperIsArrLength());
 
     if (!(indir->gtFlags & GTF_EXCEPT))
     {
         return false;
     }
 
-    GenTree* addr = indir->OperIs(GT_ARR_LENGTH) ? indir->AsArrLen()->ArrRef() : indir->AsIndir()->Addr();
+    GenTree* addr = indir->OperIsArrLength() ? indir->AsArrCommon()->ArrRef() : indir->AsIndir()->Addr();
 #ifdef DEBUG
     bool           vnBased = false;
     AssertionIndex index   = NO_ASSERTION_INDEX;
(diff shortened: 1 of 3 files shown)


src/coreclr/jit/assertionprop.cpp
```diff
@@ -4465,7 +4465,7 @@ bool Compiler::optNonNullAssertionProp_Ind(ASSERT_VALARG_TP assertions, GenTree*
     bool           vnBased = false;
     AssertionIndex index   = NO_ASSERTION_INDEX;
 #endif
-    if (optAssertionIsNonNull(indir->AsIndir()->Addr(), assertions DEBUGARG(&vnBased) DEBUGARG(&index)))
+    if (optAssertionIsNonNull(indir->GetIndirOrArrMetaDataAddr(), assertions DEBUGARG(&vnBased) DEBUGARG(&index)))
```
SingleAccretion, line 4468: I will note a subtle property of this code: it is relying on array metadata only being used for arrays (`TYP_REF`, aka heap objects). Array metadata indirections contain an implicit offset addition before the dereference - if array metadata nodes were more "general-purpose" (applicable to random pointers), the code would needs to be made aware of it. However, `TYP_REF` is either null or a valid object, so this is not necessary.

jakobbotsch (approved):
LGTM. Diffs look a bit mixed, any idea why? Do we end up kicking out useful assertions?
Also seems somewhat costly TP wise.

EgorBo:
So for
```
void Test(int i, long[] items)
{
    items[i] = 0;
    items[i] = 0;
}
``` 
with this PR we regress by emitting two bound checks instead of one, it seems to be happening because the 2nd bound check now has a differen VN (same VN as the first one but without NullExc set), JITDUMP diff: https://www.diffchecker.com/CJGsabh8/

Quoting @jakobbotsch: "I also have a note about this issue in CSE somewhere, essentially there is an asymmetry in exception sets between uses/defs that CSE does not manage to make good use of"

So as the actual change seems to be harmless and useful, it unintentionally hits problems we want to fix first before landing it, will re-open it if we fix them.

PR #93535, opened 2023-10-16
Change several more arrays to be assembly data spans (w/ collection expressions)

stephentoub:


src/libraries/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/Semantics/ExpressionBinder.cs
```diff
@@ -1275,19 +1275,19 @@ private void AdjustCallArgumentsForParams(CType callingObjectType, CType type, M
             }
         }
 
-        private static readonly PredefinedType[] s_rgptIntOp =
-        {
+        private static ReadOnlySpan<PredefinedType> RgptIntOp() =>
```
En3Tho, line 1278: Any particular reason this is a method and not a property?

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/Runtime/General/Helpers.cs
```diff
@@ -113,7 +113,7 @@ public static string EscapeTypeNameIdentifier(this string identifier)
             // We're avoiding calling into MemoryExtensions here as it has paths that lead to reflection,
```
jkotas, line 113: This comment says that this code is intentionally avoiding `MemoryExtensions` (ie `AsSpan`). If this change passes CI, it is no longer applicable and it should be deleted.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/Runtime/General/Helpers.cs
```diff
@@ -113,7 +113,7 @@ public static string EscapeTypeNameIdentifier(this string identifier)
...
             // and that would lead to an infinite loop given that this is the implementation of reflection.
 #pragma warning disable CA1870 // Use a cached 'SearchValues' instance
-            if (identifier != null && identifier.IndexOfAny(s_charsToEscape) != -1)
+            if (identifier != null && identifier.AsSpan().IndexOfAny(CharsToEscape) >= 0)
```
jkotas, line 116:
```suggestion
            if (identifier.AsSpan().Contains(CharsToEscape))
```

src/libraries/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/Semantics/ExpressionBinder.cs
stephentoub: It could be a property.

En3Tho: Just asking because of consistency

stephentoub:
@jkotas, could you take another look?

jkotas (approved):
LGTM

jeffhandley (approved):
Looks good to me

PR #93539, opened 2023-10-16
reuse existing resource file

adamsitnik:
A follow up of https://github.com/dotnet/runtime/pull/92185#discussion_r1334336418


src/libraries/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests.csproj
```diff
@@ -11,8 +12,6 @@
     <DefineConstants Condition="'$(TargetPlatformIdentifier)' == 'windows'">$(DefineConstants);TargetsWindows</DefineConstants>
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="$(CoreLibSharedDir)System\PasteArguments.cs"
```
stephentoub, line 14: These are being removed because they're actually unused?

src/libraries/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests.csproj
```diff
@@ -1,6 +1,7 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <StringResourcesPath>$(MSBuildProjectDirectory)\..\src\Resources\Strings.resx</StringResourcesPath>
```
stephentoub, line 4: What are these resources needed for in the test project?

adamsitnik: Because in #92185 we have added a new test that uses pinvokes to verify the results:

https://github.com/dotnet/runtime/blob/018efc57d557b5234e168b5c16e6ea1153de6db2/src/libraries/System.Diagnostics.Process/tests/ProcessTests.Unix.cs#L858

and the easiest way to get it working was to include these 3 files:

https://github.com/dotnet/runtime/blob/018efc57d557b5234e168b5c16e6ea1153de6db2/src/libraries/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests.csproj#L65-L70

and at least one of them uses the resource that is missing in the test project:

https://github.com/dotnet/runtime/blob/018efc57d557b5234e168b5c16e6ea1153de6db2/src/libraries/Common/src/Interop/OSX/Interop.libproc.cs#L371

I've merged a PR which introduced a copy of the resource:

https://github.com/dotnet/runtime/blob/018efc57d557b5234e168b5c16e6ea1153de6db2/src/libraries/System.Diagnostics.Process/tests/Resources/Strings.resx#L126-L128

while I believe that better solution would be to simply include the src resource in the test project.

src/libraries/System.Diagnostics.Process/tests/System.Diagnostics.Process.Tests.csproj
adamsitnik: It seems to be some kind of leftover after refactoring.

This internal type has dedicated tests defined here:

https://github.com/dotnet/runtime/blob/018efc57d557b5234e168b5c16e6ea1153de6db2/src/libraries/Common/tests/Tests/System/PasteArgumentsTests.cs#L10

and included in:

https://github.com/dotnet/runtime/blob/018efc57d557b5234e168b5c16e6ea1153de6db2/src/libraries/Common/tests/Common.Tests.csproj#L80

jozkee (approved):
Thanks, Adam.

PR #93540, opened 2023-10-16
Fix analysis of interface methods on generic types

MichalStrehovsky:
Fixes an issue observed in #92881. The dependency analysis within the compiler was incorrectly considering `Equals(object x, object y)` to be the implementation of `IEqualityComparer<T>.Equals(T, T)`. When we generate the interface dispatch table, we'd use the correct algorithm (that looks at uninstantiated types) and fail the compilation. The fix is to use the same algorithm during dependency analysis.

Looks like this has been broken ever since interface support was added to CoreRT in 2016 and the bug survived additions of default and static interface methods: https://github.com/dotnet/corert/pull/626.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
@@ -437,10 +437,14 @@ namespace ILCompiler.DependencyAnalysis
                 // Add conditional dependencies for interface methods the type implements. For example, if the type T implements
                 // interface IFoo which has a method M1, add a dependency on T.M1 dependent on IFoo.M1 being called, since it's
                 // possible for any IFoo object to actually be an instance of T.
+                DefType defTypeDefinition = (DefType)defType.GetTypeDefinition();
                 DefType[] defTypeRuntimeInterfaces = defType.RuntimeInterfaces;
+                DefType[] defTypeDefinitionRuntimeInterfaces = defTypeDefinition.RuntimeInterfaces;
+                Debug.Assert(defTypeDefinitionRuntimeInterfaces.Length == defTypeRuntimeInterfaces.Length);
                 for (int interfaceIndex = 0; interfaceIndex < defTypeRuntimeInterfaces.Length; interfaceIndex++)
                 {
                     DefType interfaceType = defTypeRuntimeInterfaces[interfaceIndex];
+                    DefType interfaceDefinitionType = defTypeDefinitionRuntimeInterfaces[interfaceIndex];
 
                     Debug.Assert(interfaceType.IsInterface);
 
@@ -457,11 +461,22 @@ namespace ILCompiler.DependencyAnalysis
                         if (!isStaticInterfaceMethod && !needsDependenciesForInstanceInterfaceMethodImpls)
                             continue;
 
+                        MethodDesc interfaceMethodDefinition = interfaceMethod;
+                        if (interfaceType != interfaceDefinitionType)
+                            interfaceMethodDefinition = factory.TypeSystemContext.GetMethodForInstantiatedType(interfaceMethodDefinition.GetTypicalMethodDefinition(), (InstantiatedType)interfaceDefinitionType);
+
                         MethodDesc implMethod = isStaticInterfaceMethod ?
-                            defType.ResolveInterfaceMethodToStaticVirtualMethodOnType(interfaceMethod) :
-                            defType.ResolveInterfaceMethodToVirtualMethodOnType(interfaceMethod);
+                            defTypeDefinition.ResolveInterfaceMethodToStaticVirtualMethodOnType(interfaceMethodDefinition) :
+                            defTypeDefinition.ResolveInterfaceMethodToVirtualMethodOnType(interfaceMethodDefinition);
                         if (implMethod != null)
                         {
+                            TypeDesc implType = defType;
+                            while (!implType.HasSameTypeDefinition(implMethod.OwningType))
+                                implType = implType.BaseType;
+
+                            if (!implType.IsTypeDefinition)
+                                implMethod = factory.TypeSystemContext.GetMethodForInstantiatedType(implMethod.GetTypicalMethodDefinition(), (InstantiatedType)implType);
+
                             if (isStaticInterfaceMethod)
                             {
                                 Debug.Assert(!implMethod.IsVirtual);
@@ -500,12 +515,7 @@ namespace ILCompiler.DependencyAnalysis
                             // Is the implementation provided by a default interface method?
                             // If so, add a dependency on the entrypoint directly since nobody else is going to do that
                             // (interface types have an empty vtable, modulo their generic dictionary).
-                            TypeDesc interfaceOnDefinition = defType.GetTypeDefinition().RuntimeInterfaces[interfaceIndex];
-                            MethodDesc interfaceMethodDefinition = interfaceMethod;
-                            if (!interfaceType.IsTypeDefinition)
-                                interfaceMethodDefinition = factory.TypeSystemContext.GetMethodForInstantiatedType(interfaceMethod.GetTypicalMethodDefinition(), (InstantiatedType)interfaceOnDefinition);
-
-                            var resolution = defType.GetTypeDefinition().ResolveInterfaceMethodToDefaultImplementationOnType(interfaceMethodDefinition, out implMethod);
+                            var resolution = defTypeDefinition.ResolveInterfaceMethodToDefaultImplementationOnType(interfaceMethodDefinition, out implMethod);
                             if (resolution == DefaultInterfaceMethodResolution.DefaultImplementation)
                             {
                                 DefType providingInterfaceDefinitionType = (DefType)implMethod.OwningType;
--- a/src/tests/Loader/classloader/StaticVirtualMethods/NegativeTestCases/MethodBodyOnUnrelatedType.ilproj
+++ b/src/tests/Loader/classloader/StaticVirtualMethods/NegativeTestCases/MethodBodyOnUnrelatedType.ilproj
@@ -5,6 +5,9 @@
 
     <!-- Crossgen2 currently doesn't support this negative check - that should be fine as runtime behavior is undefined in the presence of invalid IL. -->
     <CrossGenTest>false</CrossGenTest>
+
+    <!-- Testing TypeLoad/MissingMethod exceptions in situations that are expensive to detect -->
+    <NativeAotIncompatible>true</NativeAotIncompatible>
   </PropertyGroup>
   <PropertyGroup>
     <DebugType>Full</DebugType>
--- a/src/tests/nativeaot/SmokeTests/UnitTests/Interfaces.cs
+++ b/src/tests/nativeaot/SmokeTests/UnitTests/Interfaces.cs
@@ -40,6 +40,7 @@ public class Interfaces
         TestDefaultInterfaceVariance.Run();
         TestVariantInterfaceOptimizations.Run();
         TestSharedInterfaceMethods.Run();
+        TestGenericAnalysis.Run();
         TestCovariantReturns.Run();
         TestDynamicInterfaceCastable.Run();
         TestStaticInterfaceMethodsAnalysis.Run();
@@ -649,6 +650,54 @@ public class Interfaces
             IFace<object> o = new Yadda() { InnerValue = "SomeString" };
             string r3 = o.GrabValue("Hello there");
             if (r3 != "'SomeString' over 'System.Object' with 'Hello there'")
+                throw new Exception();
+        }
+    }
+
+    class TestGenericAnalysis
+    {
+        interface IInterface
+        {
+            string Method(object p);
+        }
+
+        interface IInterface<T>
+        {
+            string Method(T p);
+        }
+
+        class C1<T> : IInterface, IInterface<T>
+        {
+            public string Method(object p) => "Method(object)";
+            public string Method(T p) => "Method(T)";
+        }
+
+        class C2<T> : IInterface, IInterface<T>
+        {
+            public string Method(object p) => "Method(object)";
+            public string Method(T p) => "Method(T)";
+        }
+
+        class C3<T> : IInterface, IInterface<T>
+        {
+            public string Method(object p) => "Method(object)";
+            public string Method(T p) => "Method(T)";
+        }
+
+        static IInterface s_c1 = new C1<object>();
+        static IInterface<object> s_c2 = new C2<object>();
+        static IInterface<object> s_c3a = new C3<object>();
+        static IInterface s_c3b = new C3<object>();
+
+        public static void Run()
+        {
+            if (s_c1.Method(null) != "Method(object)")
+                throw new Exception();
+            if (s_c2.Method(null) != "Method(T)")
+                throw new Exception();
+            if (s_c3a.Method(null) != "Method(T)")
+                throw new Exception();
+            if (s_c3b.Method(null) != "Method(object)")
                 throw new Exception();
         }
     }


src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -457,11 +461,22 @@ public sealed override IEnumerable<CombinedDependencyListEntry> GetConditionalSt
...
                         MethodDesc implMethod = isStaticInterfaceMethod ?
-                            defType.ResolveInterfaceMethodToStaticVirtualMethodOnType(interfaceMethod) :
-                            defType.ResolveInterfaceMethodToVirtualMethodOnType(interfaceMethod);
+                            defTypeDefinition.ResolveInterfaceMethodToStaticVirtualMethodOnType(interfaceMethodDefinition) :
+                            defTypeDefinition.ResolveInterfaceMethodToVirtualMethodOnType(interfaceMethodDefinition);
                         if (implMethod != null)
                         {
+                            TypeDesc implType = defType;
+                            while (!implType.HasSameTypeDefinition(implMethod.OwningType))
+                                implType = implType.BaseType;
```
MichalStrehovsky, line 475: The test I'm disabling is hitting a nullref in this block because we walk past system.object without finding a suitable base (the test is testing invalid IL). We could do something when we hit null here, but I'm not sympathetic to that idea.  We already disable the test in R2R testing for similar reasons.

The "proper" fix would be to walk the potential vtable as part of EEType validation in the EETypeNode constructor (so that we throw there and can invalidate whatever method body is causing this) but this is expensive.

MichalStrehovsky: > but this is expensive.

By that I mean compiler throughput, not amount of work.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -457,11 +461,22 @@ public sealed override IEnumerable<CombinedDependencyListEntry> GetConditionalSt
                         if (!isStaticInterfaceMethod && !needsDependenciesForInstanceInterfaceMethodImpls)
                             continue;
 
+                        MethodDesc interfaceMethodDefinition = interfaceMethod;
```
agocke, line 464: When we say GVMs aren't tracked here, does GVM mean that the method itself doesn't have generic parameters, or that the containing type also doesn't have generic parameters (or both)?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -457,11 +461,22 @@ public sealed override IEnumerable<CombinedDependencyListEntry> GetConditionalSt
...
+                        if (interfaceType != interfaceDefinitionType)
+                            interfaceMethodDefinition = factory.TypeSystemContext.GetMethodForInstantiatedType(interfaceMethodDefinition.GetTypicalMethodDefinition(), (InstantiatedType)interfaceDefinitionType);
```
agocke, line 466: Not sure what `GetTypicalMethodDefinition` does. I think we get here when we have something like `I1<T>` and `I1<string>`, where `interfaceType = I1<string>`. And if `I1<T>` looks like

```
interface I1<T>
{
    void M(T t);
}
```

then `interfaceMethod = I1<string>.M(string)`?

Then does this statement get `I1<T>.M(T t)`?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -457,11 +461,22 @@ public sealed override IEnumerable<CombinedDependencyListEntry> GetConditionalSt
...
+
                         MethodDesc implMethod = isStaticInterfaceMethod ?
-                            defType.ResolveInterfaceMethodToStaticVirtualMethodOnType(interfaceMethod) :
-                            defType.ResolveInterfaceMethodToVirtualMethodOnType(interfaceMethod);
+                            defTypeDefinition.ResolveInterfaceMethodToStaticVirtualMethodOnType(interfaceMethodDefinition) :
+                            defTypeDefinition.ResolveInterfaceMethodToVirtualMethodOnType(interfaceMethodDefinition);
```
agocke, line 470: So this maps something like `C<T>` to `I1<T>.M(T)`?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -457,11 +461,22 @@ public sealed override IEnumerable<CombinedDependencyListEntry> GetConditionalSt
...
                         if (implMethod != null)
                         {
+                            TypeDesc implType = defType;
+                            while (!implType.HasSameTypeDefinition(implMethod.OwningType))
+                                implType = implType.BaseType;
+
+                            if (!implType.IsTypeDefinition)
+                                implMethod = factory.TypeSystemContext.GetMethodForInstantiatedType(implMethod.GetTypicalMethodDefinition(), (InstantiatedType)implType);
```
agocke, line 478: And very much not sure what we're doing here.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
MichalStrehovsky: In the ILCompiler codebase, we exclusively refer to GVMs as methods that have generic parameters themselves.

Methods on generic types (that don't have their own generic method parameters) are not called GVMs here.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
MichalStrehovsky: It gets the method definition on the uninstantiated version of the owning type:

`I1<!0>.Method`.GetTypicalMethodDefinition == `I1<T>.Method`
`I1<object>.Method`.GetTypicalMethodDefinition == `I1<T>.Method`
`I1<List<!0>>.Method`.GetTypicalMethodDefinition == `I1<T>.Method`

And also (but it might not be important here):

`I1<!0>.Method<!1>`.GetTypicalMethodDefinition == `I1<T>.Method<U>`
`I1<object>.Method<!1>`.GetTypicalMethodDefinition == `I1<T>.Method<U>`
`I1<List<!0>>.Method<object>`.GetTypicalMethodDefinition == `I1<T>.Method<U>`

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
MichalStrehovsky: It would map to `I1<!0>.M(!0)` (the owning interface is instantiated over the relevant signature variable based on the interface list).

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
MichalStrehovsky: Given:

```csharp
interface IFoo<T>
{
    void M(T t);
}

class Foo<T, U> : IFoo<U>
{
    public void M(U t) { }
}
```

And computing the conditional dependencies for `Foo<object, object>`:

The above logic just resolved that `IFoo<!1>.M` is implemented by `Foo<T, U>.M(U)`. We now need to find this method on `Foo<object, object>`. This grabs the typical method definition of `Foo<T, U>.M(U)` (the same thing we have right now) and finds it on `Foo<object, object>`, returning `Foo<object, object>(object)`

This code is here mostly to cover the more interesting case of:

```csharp
class Base<U>
{
    public void M(U t) { }
}

class Foo<T, U> : Base<U>, IFoo<U>
{
}
```

Where the result of the resolution would be `Base<!1>.M` and we want `Base<object>.M`.

ViktorHofer (approved):

MichalStrehovsky:
I'm merging to unblock Viktor but @agocke @sbomer let's continue the review and I'll submit a subsequent PR if there's anything that needs changing.

hez2010:
Will this be backported to 8.0?

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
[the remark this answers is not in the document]
MichalStrehovsky: Sounds like a better name. I've opened #93645 to address it here and elsewhere where the same name is used.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -437,10 +437,14 @@ public sealed override IEnumerable<CombinedDependencyListEntry> GetConditionalSt
...
                 // interface IFoo which has a method M1, add a dependency on T.M1 dependent on IFoo.M1 being called, since it's
                 // possible for any IFoo object to actually be an instance of T.
+                DefType defTypeDefinition = (DefType)defType.GetTypeDefinition();
                 DefType[] defTypeRuntimeInterfaces = defType.RuntimeInterfaces;
+                DefType[] defTypeDefinitionRuntimeInterfaces = defTypeDefinition.RuntimeInterfaces;
+                Debug.Assert(defTypeDefinitionRuntimeInterfaces.Length == defTypeRuntimeInterfaces.Length);
                 for (int interfaceIndex = 0; interfaceIndex < defTypeRuntimeInterfaces.Length; interfaceIndex++)
                 {
                     DefType interfaceType = defTypeRuntimeInterfaces[interfaceIndex];
+                    DefType interfaceDefinitionType = defTypeDefinitionRuntimeInterfaces[interfaceIndex];
```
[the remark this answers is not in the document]
MichalStrehovsky: Sounds like a better name. I've opened #93645 to address it here and elsewhere where the same name is used.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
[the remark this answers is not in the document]
MichalStrehovsky: I think walking the methods on both forms would be equivalent. But we need both `interfaceMethod` (in the form of `IFoo<!0>.M()`) and `interfaceMethodDefinition` (in the form of `IFoo<object>.M()`). We have to choose some type to iterate methods on and we're choosing this one...

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
```diff
@@ -457,11 +461,22 @@ public sealed override IEnumerable<CombinedDependencyListEntry> GetConditionalSt
...
+
                         MethodDesc implMethod = isStaticInterfaceMethod ?
-                            defType.ResolveInterfaceMethodToStaticVirtualMethodOnType(interfaceMethod) :
-                            defType.ResolveInterfaceMethodToVirtualMethodOnType(interfaceMethod);
+                            defTypeDefinition.ResolveInterfaceMethodToStaticVirtualMethodOnType(interfaceMethodDefinition) :
+                            defTypeDefinition.ResolveInterfaceMethodToVirtualMethodOnType(interfaceMethodDefinition);
                         if (implMethod != null)
                         {
+                            TypeDesc implType = defType;
+                            while (!implType.HasSameTypeDefinition(implMethod.OwningType))
```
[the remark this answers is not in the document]
MichalStrehovsky: So in theory we shouldn't need to. We'd just call `implMethod.InstantiateSignature(defType.Instantiation, Instantiation.Empty)`.

This however doesn't work due to a quirk. I've been thinking about doing something about it but also I don't really want to perturb things too much for the sake of making things prettier.

The problem is this:

The type system has a representation of a generic type definition and "an open type instantiation". One is `Foo<T, U>`. The other is `Foo<!0, !1>`. We need to be able to distinguish between these because of stuff like `class Foo<T, U> { void Do(T m) { typeof(Foo<>); typeof(Foo<T, U>); }` - these typeofs need to produce a different thing. The instantiation in the first case is over generic parameters (`GenericParameterDesc`) in the former case, and signature variables in the latter case

Now here's the quirk. When iterating methods on these types, we get `Do(T m)` and `Do(!0 m)` respectively. Calling `InstantiateSignature` on `Do(T m)` doesn't do anything. When one thinks about it, it really better shouldn't do anything on generic parameters because `InstantiateSignature` is also used to make things like specialized method bodies and that would coalesce the `typeof(Foo<>)` and `typeof(Foo<T, U>)` cases into the same thing.

Since we do the resolution on `Foo<T, U>`, we'll typically get `Foo<T, U>.Method` back. But sometimes also `Base<!0>.Method`. `InstantiateSignature` would work on the latter, but not on the former. So we have a bunch of code written like this that can handle both cases correctly.

MichalStrehovsky:
> Will this be backported to 8.0?

I wasn't planning to since this is also broken on .NET 7 and very likely also on .NET Native and nobody ran into this.

agocke:
I'm fine with backporting if we run into a customer scenario. As Michal said, this doesn't seem super common, so maybe that won't be necessary.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/EETypeNode.cs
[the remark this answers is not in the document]
agocke: > class Foo<T, U> { void Do(T m) { typeof(Foo<>); typeof(Foo<T, U>); }

FWIW Roslyn has a similar distinction, but maybe it's subtly different. From Roslyn's perspective, `Foo<>` is a completely unbound generic, similar to a definition. I believe you'll get the same symbol from that as you would calling `OriginalDefinition` on any given instantiation of `Foo<T, U>`.

`typeof(Foo<T, U>)`, on the other hand, is in no way open or unbound in that instance. It's fully instantiated using the type variables `T` and `U`. In other words, when looking at the definition of the class `class Foo<T, U>`, if you look at `T` and `U` you will see that they are definition symbols -- they define `T` and `U`. If you look at `T` and `U` in `typeof(Foo<T, U>)`, however, you'll see that they are _type arguments_ -- they are references to the definitions and would be no different than if you said `typeof(Foo<int, string>)`.

I'm not sure if that helps anything. I think the Roslyn equivalent of something similar to the above would be calling `Construct` on an already constructed type symbol. Roslyn produces an exception if you try to do this. So maybe that's the same as NAOT. Roslyn would demand that you pass in the original definition before calling `Construct`.

MrJul:
What a timing!

Here's a stripped down repro of a real world problem encountered a few hours ago in an application that uses CommunityToolkit.Mvvm, when publishing as NativeAOT:

```csharp
using System.Windows.Input;
using CommunityToolkit.Mvvm.Input;

internal static class Program
{
    {
        var command = (ICommand)Activator.CreateInstance(Type.GetType(args[0]));
        command.Execute(null);
        new RelayCommand<object?>(_ => { });
    }
}
```

(the `Activator.CreateInstance` isn't real code, it's simply to get an `ICommand` whose real type is not known by ILC)

<details>
  <summary>Version without any package dependency</summary>
  
```csharp
internal static class Program
{
    {
        var command = (ICommand)Activator.CreateInstance(Type.GetType(args[0]));
        command.Execute(null);
        new RelayCommand<object?>(_ => { });
    }
}

public interface ICommand
{
    void Execute(object? parameter);
}

public interface IRelayCommand<in T> : ICommand
{
    void Execute(T? parameter);
}

public sealed class RelayCommand<T> : IRelayCommand<T>
{
    public RelayCommand(Action<T> execute) { }
    public void Execute(T? parameter) { }
    public void Execute(object? parameter) { }
}
```
  
</details>

Publish with `/p:NativeAot=true`. It fails on 7.0 and 8.0-rc2 with an `OverflowException` in `InterfaceDispatchMapNode.EmitDispatchMap` since a matching method isn't found anymore in `RelayCommand<__Canon>` to implement `ICommand.Execute(object)`: `GetVirtualMethodSlot` returns -1.

To summarize, using a `RelayCommand<object>` when there's an `ICommand.Execute` call elsewhere in the program (a very common case in UI frameworks) will prevent an application from being compiled with NativeAOT. In my opinion, this should be backported to 8.0.

MichalStrehovsky:
Note that to actually hit this, the app needs to _only_ have `RelayCommand<object>` and no other `RelayCommand` instantiation (i.e. adding `if (string.Empty.Length > 0) new RelayCommand<string>(null);` anywhere in the above repro will work around the bug). I'd still question how common that is - people have been using native AOT with Avalonia long before native AOT was called native AOT and we've not heard about it. Is there something that changed on Avalonia side that would make this more common now than ever?

The truth is that the failure mode is rather obscure so maybe we'll need to service this just because of that, but I would definitely be interested in knowing how you were able to root cause the compiler crash to the specific problematic interface/class.

hez2010:
@MichalStrehovsky Avalonia people don't use CommunityToolkit.Mvvm, but WinUI and MAUI people do. Once CsWinRT completes its support for NativeAOT, people will start to encounter this issue more commonly IMO.

MrJul:
> Note that to actually hit this, the app needs to _only_ have `RelayCommand<object>` and no other `RelayCommand` instantiation

Oh, that's the point I was missing, thanks! I couldn't quite believe that it didn't happen more commonly in random apps. Indeed, I expect the issue to be way rarer with this condition added.

>Is there something that changed on Avalonia side that would make this more common now than ever?

Nothing has changed in Avalonia in this regard. This was reported on a new application using Avalonia + CommunityToolkit.Mvvm, explaining the single generic instantiation of the `RelayCommand<T>`. I don't have any stats to back that up, but CommunityToolkit seems to be gaining traction over ReactiveUI in the Avalonia world.

> I would definitely be interested in knowing how you were able to root cause the compiler crash to the specific problematic interface/class.

Knowing it was related to `RelayCommand<T>` (since removing it made the bug disappear) and a related interface (from ilc's stack trace), I started to strip the code down as much as possible. I compiled ilc 8.0-rc2 to be able to debug the exception, where it was quite clear through debugging that `Execute(object)` wasn't present in `RelayCommand<__Canon>`.

The next step was trial and error to understand which extra code was triggering the bug. I assumed a second rooted `ICommand` implementation was needed (it wasn't) and that usages of the `RelayCommand<T>` were made via the `ICommand` interface. Simple cases without Avalonia obviously didn't trigger the bug because ilc is quite good at devirtualizing and removing all unneeded members when it can infer the actual type.

agocke:
Sorry, I haven't been fully following the details here. Has someone managed to repro this bug using external code? Either CommunityToolkit or Avalonia?

Because if so, I think that would meet the bar for backporting, mostly because as Michal said, it's difficult to figure out what the problem is.

MrJul:
> Has someone managed to repro this bug using external code? Either CommunityToolkit or Avalonia?

Yes, here's some basic reproduction using Avalonia + CommunityToolkit.Mvvm: https://github.com/MrJul/IlcCrashRepro
No convoluted code here, it's basically a hello world with a single button.

`dotnet publish` it with .NET 8.0-rc2 and observe the IL compiler crashing.

agocke:
OK, this seems worth backporting to me.

PR #93546, opened 2023-10-16
Renamed stoppingToken to cancellationToken.

Sixaxys317:
Renamed  stoppingToken to cancellationToken where relevant.


Sixaxys317:
@dotnet-policy-service agree

stephentoub:
Thanks, but changing parameter names in exported surface area is a breaking change.

Sixaxys317:
Thanks for quick response!

Aha, I didn't know that. It was worth a shot to fix the inconsistency in the naming of the parameter.

PR #93547, opened 2023-10-16
[mono] Optimizing arm64 trampolines

jandupej:
Two occurrences are addressed here:

(1) Sequential SIMD loads/stores try to use paired opcodes, e.g. in `generic_trampoline_jit`
```asm
A0 47 80 3D        str    q0, [x29, #0x110]  ;
A1 4B 80 3D        str    q1, [x29, #0x120]  ; Merge with above into one stp
A2 4F 80 3D        str    q2, [x29, #0x130]
A3 53 80 3D        str    q3, [x29, #0x140]
A4 57 80 3D        str    q4, [x29, #0x150]
A5 5B 80 3D        str    q5, [x29, #0x160]
A6 5F 80 3D        str    q6, [x29, #0x170]
A7 63 80 3D        str    q7, [x29, #0x180]
```

(2) Arithmetics with large immediate values takes advantage of the full immediate range (but not shifted imm), e.g. in `generic_trampoline_jit`:
```asm
31 02 04 91        add    x17, x17, #0x100
31 02 04 91        add    x17, x17, #0x100
31 82 00 91        add    x17, x17, #0x20   ; These three should fuse into add x17,x17,#0x220 
```
According to https://developer.arm.com/documentation/dui0801/c/A64-General-Instructions/ADD--immediate-?lang=en the immediate range is 0..4095

### DIFF
--- a/src/mono/mono/arch/arm64/arm64-codegen.h
+++ b/src/mono/mono/arch/arm64/arm64-codegen.h
@@ -480,18 +480,19 @@ arm_encode_imm7 (int imm, int size)
 #define arm_neon_ldrq_lit(p, rd, target) arm_emit ((p), 0b00011100000000000000000000000000 | (0b10 << 30) | (arm_get_disp19 ((p), (target)) << 5) | (rd))
 #define arm_neon_ldrq_lit_fixup(p, target) *((guint32*)p) = (*((guint32*)p) & 0xff00001f) | (arm_get_disp19 ((p), (target)) << 5)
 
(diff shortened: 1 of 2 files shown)


vargaz (approved):

vargaz:
Note that this code is designed to be run like once per callsite, so optimizing it is not going to achieve much. The only case where its called more often is delegate invokes.

jandupej:
The CI failures appear unrelated. Merging.

PR #93558, opened 2023-10-16
[hot reload][debugger] Resolve field types before use

### LINKED ISSUE #93036: [mono][hotreload] Add instance field
If I add a field and do not assign anything to it the field->type is NULL.

Adding a test case in this draft PR:
https://github.com/dotnet/runtime/pull/93035/files

lambdageek:
Fixes #93036 - if a field is added in a hot reload delta, but it isn't used anywhere, the type might not be resolved yet


lambdageek:
in reflection, we already resolve field types in the icalls, so no change needed there

thaystg (approved):

PR #93560, opened 2023-10-16
[browser] Disable HybridGlobalization failure

ilonatommy:
Scenarios: browser (ChromeDriver 118.0.5993.0) and normal (v8-11.8.172) fail on Linux and Windows (but Windows does not preserve logs from Helix). Failures is not deterministic even though theoretically we should get the same result each time. Disabling till I reproduce and investigate.


pavelsavara (approved):

radical:
> but Windows does not preserve logs from Helix

where?

PR #93563, opened 2023-10-16
Fix binding types with optional string parameters

### LINKED ISSUE #93159: Cannot use ConfigurationBinder source generator with a type that has a string ctor parameter with default value
### Description

When binding to a type with a constructor that has a default string parameter, the generator emits code that doesn't compile.

```
C:\scratch\configStringCtorParam\Microsoft.Extensions.Configuration.Binder.SourceGeneration\Microsoft.Extensions.Config
uration.Binder.SourceGeneration.ConfigurationBindingGenerator\BindingExtensions.g.cs(77,27): error CS0103: The name 's'
 does not exist in the current context [C:\scratch\configStringCtorParam\configStringCtorParam.csproj]
```

The code generated looks like this:
```c#
 public static SD InitializeSD(IConfiguration configuration, BinderOptions? binderOptions)
        {
            if (configuration["S"] is string value1)
            {
                string s = value1;
            }

            return new SD(s);
        }
```

Bug is likely in the EmitBindImplForMember method but I haven't tracked it down.  I suspect there's a disagreement between the case where the member is string parseable but also has a default value.  This same repro works when removing the default value.

### Reproduction Steps

```C#
var x = c.Get<SD>();
Console.WriteLine(x);

public class SD
{
    public SD(string s = "default value") { S = s; }
    public string S { get;}
    public override string ToString() => S;
}
```

### Expected behavior

Compiles successfully.

### Actual behavior

Compilation failure.

### Regression?

No

### Known Workarounds

Remove default value from ctor param.

### Configuration

_No response_

### Other information

_No response_

ericstj:
Fixes https://github.com/dotnet/runtime/issues/93159

Previously the generator would skip emitting the default in the case the member was ParsableFromStringSpec and  AssignFromSectionValue.  It should have only done this for the ErrorOnFailedBinding.

Most of this PR is an added test to track the code-gen for default parameters.


src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.TestClasses.cs
```diff
@@ -129,7 +129,7 @@ public class ClassWhereParametersHaveDefaultValue
             public StringComparison? NSC { get; }
             public char? NC { get; }
 
-            public ClassWhereParametersHaveDefaultValue(string? name, string address, 
+            public ClassWhereParametersHaveDefaultValue(string? name = "John Doe", string address = "1 Microsoft Way",
```
tarekgh, line 132: >ClassWhereParametersHaveDefaultValue [](http://example.com/codeflow?start=19&length=36)

do we have any test using [Primary constructor](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/tutorials/primary-constructors) with default values?

tarekgh (approved):
LGTM. I added a minor question.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/Common/ConfigurationBinderTests.TestClasses.cs
ericstj: We could add a test for that.  From [metadata I think it's the same](https://sharplab.io/#v2:EYLgtghglgdgNAExAagD4AEBMBGAsAKAPQGYACLUgYQBsIBnOgdQAsBTAJ1YAUJ2IxWAFw50AEhABurACKsAZhACu1QQDUI1RawAU6bAAYA/KRj9WpALykARACkA9sxilp91tbjkDpCAgScGSxtsUgBZKABjdns6ezlBUkYIAE8PAlIMzIzYBIgAc3MrABZMTzlqewgEuSCSgDp9OU8Ee0VganMEIOI67CLsAFYATmbWCKhIalIwbt7+4cwANgHiAYAOIYB2IeJMYjWixaWi4mJQzwBlQXZYPMp7MAAHXihY5zoIoKubmDuH55ubzqAHl2AhYBpPBFmLxSJ8rAByACOCLg6SymRyxlMBVqpVI5UqgmxNWKmAaTVILTaHWxXSsPT6gxGVLGEw02JmDLmzKWK3WWx2ewOR0WJzOl2ut3uTxeb2xHy+Ut+MoBr3sMBBYIh1ChMPY2PhpGRCIAlAQAN7ojIkLxGUgAOTMpAtpAKggA3KQAL5BUwCD3W8hkPT6UgAQT8AToLrdQi9vqsvn8rAYgfwWVtOQjuNd7oTQXyrHTmbIhKqpAAYrH8z6gnIS5lbdT2uZpDX43WrAhGzayAg2ZMwh3PV3pr3g6RvtL/nKNVPKCOC1YPhPbdDYYu853E3C12QsY7w7m46Pdzji0HbeXiY7q9uz36G1f+61W8YHe2H8uTD2X6zxkmD9QiXMcYDAfcp2VP5ZUBDUPwuLdTx/GBV3/DcDUdJDa3PCJ029AgCCAA=), but it's possible the roslyn representation seen by source gen is different.

eiriktsarpalis: From my experience primary constructors are represented the same way in the symbols API, but it can't hurt to add a few tests.

eiriktsarpalis (approved):

ericstj:
Added a test here.  I think I'd like to combine this with https://github.com/dotnet/runtime/pull/94070 since both impact the source generator's ability to handle default parameters.

ericstj:
Need to update this as the baselines are now stale with https://github.com/dotnet/runtime/pull/94267 merged.

Issue #93497, opened 2023-10-14
[API Proposal]: Add protected factory method for creating Label in ILGenerator, abstract out LocalBuilder


buyaa-n:
### Background and motivation

For adding [`AssemblyBuilder.Save()` equivalent in .NET Core](https://github.com/dotnet/runtime/issues/62956) we have [abstracted](https://github.com/dotnet/runtime/issues/78542) most of the System.Refleciton.Emit types and adding new implementations for them. This abstraction did not cover `Label` and `LocalBuilder` types and its constructors are internal, so I could not create an instance from the new `ILGenerator` implementation.

### API Proposal
- For `LocalBuilder` we need to unseal the `LocalBuilder` class and make it abstract, make the `internal` constructor `protected`, also add a public getter that exposes the `method` the local belongs to.

  ```diff
  namespace System.Reflection.Emit;

  -public sealed class LocalBuilder : System.Reflection.LocalVariableInfo
  +public abstract class LocalBuilder : System.Reflection.LocalVariableInfo
  {
  -   internal LocalBuilder(int localIndex, Type localType, MethodInfo method, bool isPinned) { }
  +   protected LocalBuilder() { }
  +   public abstract MethodInfo LocalMethod { get; }
  }
  ```

  Abstract `LocalMethod` property can be used for validation. Parent `LocalVariableInfo` has public virtual properties like `LocalType`, `LocalIndex`, `IsPinned`.

- For `Label` struct we need a protected factory method in the abstract `ILGenerator` type so that we can create an instance from derived types. 

  ```diff
  namespace System.Reflection.Emit;

  public abstract class ILGenerator
  {
      protected ILGenerator() { }
      ...
      public abstract Label DefineLabel();
  +   protected static Label CreateLabel(int id) { throw null; }
  }

  public readonly partial struct Label : System.IEquatable<System.Reflection.Emit.Label>
  {
      ...
  +   public int Id { get; }
  }
  ```

### API Usage

```csharp
internal sealed class ILGeneratorImpl : ILGenerator
{
    public override Label DefineLabel()
    {
        LabelHandle metadataLabel = _il.DefineLabel();
        Label emitLabel = CreateLabel(metadataLabel.Id);
        _labelTable.Add(emitLabel, metadataLabel);
        return emitLabel;
    }
}
```

Related to https://github.com/dotnet/runtime/issues/92975

CC @steveharter

teo-tsirpanis:
[...]

Issue #91757, opened 2023-09-07
CI error: System.Net.Quic.QuicException: The connection timed out from inactivity


steveharter:
## Build Information
Build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=640353
Build error leg or test failing: System.Net.Http.Functional.Tests.SocketsHttpHandler_HttpClientHandler_MaxResponseHeadersLength_Http3.LargeSingleHeader_ThrowsException

## Error Message

Multiple `System.Net.Quic.QuicException : The connection timed out from inactivity.` in System.Net.Quic tests on "smaller" platforms, e.g. Alpine3.15 arm 32

```
 System.Net.Quic.QuicException : The connection timed out from inactivity.
      Stack Trace:
        /_/src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnection.cs(547,0): at System.Net.Quic.QuicConnection.HandleConnectionEvent(QUIC_CONNECTION_EVENT& connectionEvent)
        /_/src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnection.cs(581,0): at System.Net.Quic.QuicConnection.NativeCallback(QUIC_HANDLE* connection, Void* context, QUIC_CONNECTION_EVENT* connectionEvent)
        --- End of stack trace from previous location ---
        /_/src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/ValueTaskSource.cs(176,0): at System.Net.Quic.ValueTaskSource.System.Threading.Tasks.Sources.IValueTaskSource.GetResult(Int16 token)
        /_/src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnection.cs(324,0): at System.Net.Quic.QuicConnection.FinishConnectAsync(QuicClientConnectionOptions options, CancellationToken cancellationToken)
        /_/src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnection.cs(72,0): at System.Net.Quic.QuicConnection.<ConnectAsync>g__StartConnectAsync|2_0(QuicClientConnectionOptions options, CancellationToken cancellationToken)
        /_/src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnection.cs(77,0): at System.Net.Quic.QuicConnection.<ConnectAsync>g__StartConnectAsync|2_0(QuicClientConnectionOptions options, CancellationToken cancellationToken)
[...]

PR #93565, opened 2023-10-16
Add Label support in new ILGenerator

buyaa-n:
Add implementation for `DefineLabel()`, `(OpCode opcode, Label label)`, `Emit(OpCode opcode, Label[] labels)`, `MarkLabel(Label loc)` and corresponding tests

Contributes to https://github.com/dotnet/runtime/issues/92975 and https://github.com/dotnet/runtime/issues/93497


jkotas (approved):

src/libraries/System.Reflection.Emit.ILGeneration/tests/Label/LabelId.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            TypeBuilder type = Helpers.DynamicType(TypeAttributes.NotPublic);
+            MethodBuilder method = type.DefineMethod("Method", MethodAttributes.Public);
+            ILGenerator ilGenerator = method.GetILGenerator();
+            for (int i = 0; i < 100; i++)
+            {
+                Label label = ilGenerator.DefineLabel();
+                Assert.Equal(i, label.GetHashCode());
+            }
+        }
```
AaronRobinsonMSFT, line 31: These seems like unnecessary validation on behavior that the system should be agnostic too. For example, no one should be taking a dependency on `GetHashCode()` being the ID of the label, that seems like a bad assumption or policy to define. The default initialization portion does seems reasonable, but unsure if providing a lot of value.

buyaa-n: Sorry my bad, the tests copied from [LabelGetHashCode](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Reflection.Emit.ILGeneration/tests/Label/LabelGetHashCode.cs) test and forgot to update the `Assert.Equal(i, label.GetHashCode());` row
```suggestion
        [Fact]
        public void LabelId_DefaultConstuctor_ReturnsZero()
        {
            Label label1 = new Label();
            Label label2 = new Label();

            Assert.Equal(0, label1.Id);
            Assert.Equal(label2.Id, label1.Id);
        }

        [Fact]
        public void LabelId_CreatedByILGenerator_ReturnsId()
        {
            TypeBuilder type = Helpers.DynamicType(TypeAttributes.NotPublic);
            MethodBuilder method = type.DefineMethod("Method", MethodAttributes.Public);
            ILGenerator ilGenerator = method.GetILGenerator();
            for (int i = 0; i < 100; i++)
            {
                Label label = ilGenerator.DefineLabel();
                Assert.Equal(i, label.Id);
            }
        }
```

AaronRobinsonMSFT (approved):

buyaa-n:
The failure is [known ](https://github.com/dotnet/runtime/issues/91757)

PR #93575, opened 2023-10-16
Update test build group boundaries to today test distribution

trylek:
Test build group boundaries are several years old; since then, the JIT/HardwareIntrinsics tests have arrived and substantially changed the test distribution. Moreover, as by now the vast majority (over 95%) of tests use the new-style merged wrappers, it makes no sense to establish build group boundaries in the middle of merged wrappers because that only means that some test projects need to be traversed twice, once in the group they belong to according to the partitioning and one more time as a dependency of their merged test wrapper.

Thanks

Tomas

### DIFF
--- a/src/tests/Common/dirs.proj
+++ b/src/tests/Common/dirs.proj
@@ -65,11 +65,11 @@
         <GroupNumber>1</GroupNumber>
       </_GroupStartsWith> -->
 
-      <_GroupStartsWith Include="$(TestRoot)JIT\Methodical\Boxing\callconv\_relinstance_il.ilproj">
+      <_GroupStartsWith Include="$(TestRoot)JIT\HardwareIntrinsics\HardwareIntrinsics_General_r.csproj">
         <GroupNumber>2</GroupNumber>
       </_GroupStartsWith>
 
-      <_GroupStartsWith Include="$(TestRoot)JIT\Regression\CLR-x86-JIT\V1-M12-Beta2\b35351\b35351.ilproj">
+      <_GroupStartsWith Include="$(TestRoot)JIT\HardwareIntrinsics\HardwareIntrinsics_X86_Avx_r.csproj">
         <GroupNumber>3</GroupNumber>
       </_GroupStartsWith>
     </ItemGroup>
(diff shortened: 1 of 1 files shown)


markples:
I'm trying to follow how the `_GroupsStartsWith` / `__TestGroupToBuild` / `AllProject` logic works.  Are those ordinal comparisons used to do the partitioning?  If so, should these `_GroupStartsWith` values to be path prefixes instead of projects to avoid splitting a directory in half?  (e.g., `$(TestRoot)JIT\jit64` instead of `$(TestRoot)JIT\jit64\jit64_1.csproj`).  If this is the case, I'm not sure how to do HardwareIntrinsics, which gets more than one group.  Perhaps the merged csproj should be dropped into the subdirectories?

trylek:
@markples - Thanks for your comments, I think these use case sensitive ordinal comparisons per

https://github.com/dotnet/runtime/blob/ccc475ed5646abe71ef596702d06e78e34981736/src/tests/Common/dirs.proj#L134

(case insensitive would be actually more appropriate according to my opinion) and I guess it should be fine to specify folders instead of individual tests, I was somewhat worried whether we're not attempting resolution of the paths somewhere that may affect their normalization or whatnot but perhaps it would be best to just try and see.

For the HW intrinsics, the problem is that they account for 50% of Pri1 tests total (in fact they're not Pri1/Pri0 split so that in Pri0 mode they utterly trounce the remaining tests). Having said that, they're super fast to execute so the number of tests probably isn't too problematic but they still take a bit to build. But then, this change is really kind of just an update of the boundaries / cleanup, the actual lab perf isn't affected greatly because even if the tests get built twice, MSBuild notices that pretty early and skips most of the chores during the second build attempt.

markples:
I was specifically worried about (and should have written the first time) having something like

```
B\A\A.csproj
B\B.csproj # merged group
B\C\C.csproj
```

where the merged group csproj is "in the middle" of the directory.  (Case sensitivity is an interesting wrinkle on this too.)

jkoritzinsky (approved):

trylek:
I have adjusted the group boundaries and in most cases I removed the project names and I only left in the directories which is of course easier to work with in normal cases. The sad bit is that this doesn't work well for the HardwareIntrinsics tests that constitute about 1/2 of our Pri1 test portfolio and about 80% of the Pri0 set; so for Pri1 I have kept the project names as the tests take much longer to build than other "normal" test cases; I'll review the build times once I'm able to pass innerloop and outerloop testing and I'll see whether I might be able to further tweak them to simplify the boundaries.

trylek:
@jkoritzinsky, @markples - I found out that the crashes in Mono interpreter legs were caused by the distribution still being pretty uneven. I have played with it a little more - I changed the group filtering to use case insensitive comparison, I added logging of test counts in each group and that let me achieve a much better distribution that for Pri0 tests on my local machine reduced managed component build time to about one half. If the interpreter leg starts passing with this change, I'll also run the outerloop run and then I'll ask you to take another look at the change.

trylek:
@markples - Jeremy is apparently on vacation, all tests are passing now, could you please take another look when you have a chance to see whether you find this change reasonable now? I have removed all explicit references to test projects and I verified in the test logs that the test distribution is now more or less even both in Pri0 and Pri1 case.

markples (approved):

PR #93576, opened 2023-10-16
Handle MultiRegNode in gtGetRegMask()

### LINKED ISSUE #93527: SuperPMI replay failure: (unsigned)reg < ArrLen(regMasks)
## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=438916
Build error leg or test failing: Build / SuperPMI replay windows x64 checked / Send job to Helix (Windows)
Pull request: https://github.com/dotnet/runtime/pull/93429

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json
{
  "ErrorMessage": "(unsigned)reg < ArrLen(regMasks)",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

Failed in 3 different PRs, log: https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-92844-merge-5ef593d10a7842f881/linux-arm64-2/1/console.17223591.log?helixlogtype=result

cc @dotnet/jit-contrib 

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=438916
**Error message validated:** `(unsigned)reg < ArrLen(regMasks)`
**Result validation: :x:** Known issue did not match with the provided build.
**Validation performed at:** 10/15/2023 3:53:49 PM UTC

### Report

|---|---|---|---|
|[439272](https://dev.azure.com/dnceng-public/public/_build/results?buildId=439272)|dotnet/runtime|[HardwareIntrinsics_Arm_ro.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=439272&view=ms.vss-test-web.build-test-results-tab&runId=9777234&resultId=118772)||
|[439144](https://dev.azure.com/dnceng-public/public/_build/results?buildId=439144)|dotnet/runtime|[HardwareIntrinsics_Arm_ro.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=439144&view=ms.vss-test-web.build-test-results-tab&runId=9773234&resultId=118775)||
#### Summary
|---|---|---|
|0|2|2|

kunalspathak:
Adding multi-def support exposed the existing shortcoming in `gtGetRegMask` where we were not handling if the multi-reg intrinsic or node is source of `GT_COPY` or `GT_RELOAD`. We will only get the `regMask` if the `multiRegIndex` of source holds a valid register. One of the things I don't like about this is we have to find `Compiler*` to this method to get the register count.

Fixes: #93527

### DIFF
--- a/src/coreclr/jit/gentree.cpp
+++ b/src/coreclr/jit/gentree.cpp
@@ -1060,17 +1060,17 @@ bool GenTree::NeedsConsecutiveRegisters() const
 // Return Value:
 //    Reg Mask of GenTree node.
 //
-regMaskTP GenTree::gtGetContainedRegMask()
+regMaskTP GenTree::gtGetContainedRegMask(Compiler* comp)
 {
     if (!isContained())
     {
-        return isUsedFromReg() ? gtGetRegMask() : RBM_NONE;
+        return isUsedFromReg() ? gtGetRegMask(comp) : RBM_NONE;
     }
 
     regMaskTP mask = 0;
     for (GenTree* operand : Operands())
     {
-        mask |= operand->gtGetContainedRegMask();
+        mask |= operand->gtGetContainedRegMask(comp);
     }
     return mask;
 }
@@ -1084,7 +1084,7 @@ regMaskTP GenTree::gtGetContainedRegMask()
 // Return Value:
 //    Reg Mask of GenTree node.
 //
-regMaskTP GenTree::gtGetRegMask() const
+regMaskTP GenTree::gtGetRegMask(Compiler* comp) const
 {
     regMaskTP resultMask;
 
@@ -1113,6 +1113,27 @@ regMaskTP GenTree::gtGetRegMask() const
             }
         }
     }
+#ifdef TARGET_ARM64
+    else if (IsCopyOrReloadOfMultiRegNode())
+    {
+        // A multi-reg copy or reload, will have valid regs for only those
+        // positions that need to be copied or reloaded.  Hence we need
+        // to consider only those registers for computing reg mask.
+
+        const GenTreeCopyOrReload* copyOrReload = AsCopyOrReload();
+        const unsigned             regCount     = GetMultiRegCount(comp);
+
+        resultMask = RBM_NONE;
+        for (unsigned i = 0; i < regCount; i++)
+        {
+            regNumber reg = copyOrReload->GetRegNumByIdx(i);
+            if (reg != REG_NA)
+            {
+                resultMask |= genRegMask(reg);
+            }
+        }
+    }
+#endif
 #if FEATURE_ARG_SPLIT
     else if (compFeatureArgSplit() && OperIsPutArgSplit())
     {
--- a/src/coreclr/jit/gentree.cpp
+++ b/src/coreclr/jit/gentree.cpp
@@ -799,30 +799,43 @@ bool GenTree::gtHasReg(Compiler* comp) const
 {
     bool hasReg = false;
 
-    if (IsMultiRegCall() || IsMultiRegHWIntrinsic())
+    if (IsMultiRegNode())
     {
-        const unsigned regCount = GetMultiRegCount(comp);
+        // A Multi-reg copy or reload node is said to have regs,
+        // if it has valid regs in any of the positions.
 
-        // A Multi-reg call node is said to have regs, if it has
-        // reg assigned to each of its result registers.
-        for (unsigned i = 0; i < regCount; ++i)
+        if (IsMultiRegCall())
         {
-            hasReg = (GetRegByIndex(i) != REG_NA);
-            if (!hasReg)
+            const GenTreeCall* call     = AsCall();
+            const unsigned     regCount = call->GetReturnTypeDesc()->GetReturnRegCount();
+
+            for (unsigned i = 0; i < regCount; ++i)
             {
-                break;
+                hasReg = (call->GetRegNumByIdx(i) != REG_NA);
+                if (hasReg)
+                {
+                    break;
+                }
             }
         }
-    }
-    else if (IsCopyOrReload())
-    {
-        const GenTreeCopyOrReload* copyOrReload = AsCopyOrReload();
-        const GenTree*             op1          = copyOrReload->gtGetOp1();
-        if (IsMultiRegCall() || IsMultiRegHWIntrinsic())
+        else if (IsMultiRegLclVar())
         {
-            const unsigned regCount = op1->GetMultiRegCount(comp);
-            // A Multi-reg copy or reload node is said to have regs,
-            // if it has valid regs in any of the positions.
+            const GenTreeLclVar* lclVar   = AsLclVar();
+            const unsigned       regCount = lclVar->GetFieldCount(comp);
+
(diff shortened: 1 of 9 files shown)


kunalspathak:
@dotnet/jit-contrib

src/coreclr/jit/gentree.cpp
```diff
@@ -1113,6 +1113,27 @@ regMaskTP GenTree::gtGetRegMask() const
...
+        for (unsigned i = 0; i < regCount; i++)
+        {
+            regNumber reg = copyOrReload->GetRegNumByIdx(i);
+            if (reg != REG_NA)
+            {
+                resultMask |= genRegMask(reg);
+            }
+        }
+    }
+#endif
```
jakobbotsch, line 1136: It looks like this case could be combined with the previous case to make the TP cost smaller. Also, shouldn't this be handled for all platforms?

kunalspathak: > It looks like this case could be combined with the previous case to make the TP cost smaller.

I would rather keep it separate, the way it is for other places in codebase. Also, for TP, we will still need to do the check for `IsCopyOrReloadOfMultiRegNode()` and another check again when getting `regCount`.

> Also, shouldn't this be handled for all platforms?

Possibly, but I haven't seen cases where we hit this for other platforms, so would rather defer removing the `#ifdef`.

jakobbotsch: > I would rather keep it separate, the way it is for other places in codebase. Also, for TP, we will still need to do the check for IsCopyOrReloadOfMultiRegNode() and another check again when getting regCount.

Why can't the previous case just be completely removed? It seems like this one handles it. If not, you can at least combine it to be
```cpp
if (IsCopyOrReload())
{
  GenTree* op1 = gtGetOp1();
  if (op1->IsMultiRegCall())
  {
    // previous case
  }
  else if (op1->IsMultiRegNode())
  {
    // current case
  }
}
```

The MinOpts TP costs right now look significant.

> Possibly, but I haven't seen cases where we hit this for other platforms, so would rather defer removing the #ifdef.

It seems like a bad idea to wait until we hit the bug on other platforms to fix it.

kunalspathak: > If not, you can at least combine it to be

I think I misunderstood your previous comment. Yes, I can definitely combine it with `IsCopyOrReload()`. But thinking a bit more about this, I think this should be handled at other code paths as well where we currently just check for `IsCopyOrReloadOfMultiRegCall()` and `IsMultiRegLclVar()` and should also have a case for `IsMultiRegNode()`.

kunalspathak: There are 3 places where we need to handle `multiRegIdx` and fixed as part of this PR:
1. `gtConsumeReg(m)`: This API already took `multiRegIndex` and we should fetch the corresponding `regNum` from the tree. If it is `REG_NA`, we should pass it to gcinfo.
2. `gtProduceReg()`: I handled two scenarios here: If the tree is multi-reg hwintrinsic node or if it is a `GT_COPYRELOAD` whose operand is multi-reg hwintrinsic node. 
3. `gtHasReg()`: Same as `gtProduceReg()`.

kunalspathak: > 1. `gtConsumeReg(m)`: This API already took `multiRegIndex` and we should fetch the corresponding `regNum` from the tree. If it is not `REG_NA`, we should pass it to gcinfo.

By fixing this, the main issue was resolved. Below are just the safety measures to make sure we handle it correctly at other places.

> 2\. `gtProduceReg()`: I handled two scenarios here: If the tree is multi-reg hwintrinsic node or if it is a `GT_COPYRELOAD` whose operand is multi-reg hwintrinsic node.

The only multi reg intrinsic we have in xarch is DivRem, so yes, we were missing updating the gcinfo when we produced in the registers for this intrinsic.

> 3\. `gtHasReg()`: Same as `gtProduceReg()`.

We do not have multi-reg intrinsic where we will partially assign registers to them. So this can be simply checked using `GetRegNum() != REG_NA` and doesn't need special handling. This still need to be handled for `CopyOrReload` node, so added that path.

kunalspathak:
@jakobbotsch - can you take a look?

src/coreclr/jit/gentree.cpp
```diff
@@ -848,6 +848,28 @@ bool GenTree::gtHasReg(Compiler* comp) const
...
+                break;
+            }
+        }
+    }
+    // else if (IsMultiRegHWIntrinsic())
+    //{
+    //    // For intrinsic that returns multiple values, we always assign
+    //    // them registers to all or none and hence checking if the reg
+    //    // is assigned at 0th index using GetRegNum() is sufficient.
+    //}
```
jakobbotsch, line 872: Delete?

jakobbotsch (commented):
The code looks ok to me, but the TP code seems high, especially in MinOpts where we don't expect to have multireg nodes (at least allocated into registers).

There are very few places we actually can produce/consume multireg nodes -- I wonder if some of these core methods like `genProduceReg` should be split into `genProduceReg` and `genProduceMultiReg`, with `genProduceMultiReg` called directly from the places we can produce multiple registers. That would probably help with TP (maybe even turn this into an improvement overall). I would also be fine with leaving this as a follow-up .NET 9 work item.

src/coreclr/jit/gentree.cpp
kunalspathak: That was intentional for future reference.

kunalspathak:
> but the TP code seems high

I agree with you. I think the immediate fix for the original issue is just https://github.com/dotnet/runtime/pull/93576/files#diff-426fb8beee6e663b3fb0f11330af0ce0ce21ff22a8ea819f39d12a95d9d6cffcR1456-R1460. To unblock superpmi replay clean runs, I am thinking of just merging this change and thinking more about handling the cases for `gtHasReg()` and `genProduceReg()` separately. Thoughts?

jakobbotsch:
> I agree with you. The immediate fix for the original issue is just https://github.com/dotnet/runtime/pull/93576/files#diff-426fb8beee6e663b3fb0f11330af0ce0ce21ff22a8ea819f39d12a95d9d6cffcR1456-R1460. To unblock superpmi replay clean runs, I am thinking of just merging this change and thinking more about handling the cases for `gtHasReg()` and `genProduceReg()` separately. Thoughts?

Sounds good to me.

jakobbotsch (approved):

kunalspathak:
With some minor adjustment of how we get multireg in https://github.com/dotnet/runtime/pull/93576/files#diff-426fb8beee6e663b3fb0f11330af0ce0ce21ff22a8ea819f39d12a95d9d6cffcR2244 and https://github.com/dotnet/runtime/pull/93576/files#diff-fba867eda1c745875748370c59da2195c3823b85a17d199444068d93f91c749bR839 gets little TP gain.

image

Issue #8897, opened 2017-09-07
[Arm64/Unix] glibc memmove and System.Buffer:Memmove CopyThreshold


sdmaclea:
See dotnet/coreclr#13793. When glibc arm64 memmove optimization are present in most distros, revisit optimizing System.Buffer:Memmove CopyThreshold based on performance testing

PR #93579, opened 2023-10-16
[Arm64/Unix] Use native memcpy/memmove for ARM64 as glibc 2.36 added optimizations

Spacefish:
This commit: https://github.com/bminor/glibc/commit/9f298bfe1f183804bb54b54ff9071afc0494906c added an optimized memcpy implementation for ARM64 which uses the SVE extensions to glibc 2.36 and newer.

It is now used in some Linux Distributions like Debian 12 (bookworm), Ubuntu 23.04 or the upcoming Ubuntu 24.04  LTS in April 2024.

see https://github.com/dotnet/runtime/issues/8897

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Buffer.Unix.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Buffer.Unix.cs
@@ -5,12 +5,10 @@ namespace System
 {
     public static partial class Buffer
     {
-#if TARGET_ARM64 || TARGET_LOONGARCH64
+#if TARGET_LOONGARCH64
         // Managed code is currently faster than glibc unoptimized memmove
-        // TODO-ARM64-UNIX-OPT revisit when glibc optimized memmove is in Linux distros
-        // https://github.com/dotnet/runtime/issues/8897
         private static nuint MemmoveNativeThreshold => nuint.MaxValue;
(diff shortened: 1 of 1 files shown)


EgorBo:
512 bytes threshold looks pretty low, does SVE give any advantage over the current implementation? Currently, we align data to 16 bytes boundary and then do copy using 64 bytes blocks (using two `stp` SIMD stores), considering that there is no hardware that implements SVE with >128 bit vectors under the hood?

Pros:
* We migth potentially benefit from 256-bit (or wider) SVE stores in future (Although, hopefully, by that time JIT will be able to use SVE too)

Cons:
* We've seen regressions from the native memcpy compared to current implementation (e.g. on Ubuntu 22.04)

MichalPetryka:
> It is now used in some Linux Distributions like Debian 12 (bookworm), Ubuntu 23.04 or the upcoming Ubuntu 24.04 LTS in April 2024.

Is relying on something present only on such new distros a good idea?

Spacefish:
> Currently, we align data to 16 bytes boundary and then do copy using 64 bytes blocks (using two `stp` SIMD stores) + one 64-byte block for the trailing elements.

Oh i didn´t know that. Then my MR probably doesn´t make much sense, as it would lead to performance regressions on most distros out there currently.

PR #93591, opened 2023-10-17
Fix NativeAOT test runs

MichalStrehovsky:
These got broken when readytorun tests got merged. We still execute this target:

https://github.com/dotnet/runtime/blob/2d47f845f895a361d3a56ee442345ddd8fa297e8/src/tests/readytorun/tests/mainv1.csproj#L159-L168

even though the test doesn't actually build. The copy task obviously fails. Not sure if there's a better way than what I'm proposing here.

Cc @dotnet/ilc-contrib


trylek (approved):
LGTM, thanks for fixing this!

agocke (approved):
This is the best way I know of to stop the build itself, not just the test run

PR #93600, opened 2023-10-17
Add a test for determinism

### LINKED ISSUE #73931: Determinism for NativeAOT
Deterministic binaries should be supported for NativeAOT and we should have tests.

In short, the feature is that identical inputs to the exact same compiler toolchain should produce bit-for-bit identical results.

MichalStrehovsky:
Resolves #73931. This checks that the result of parallel build is the same as the result of single-threaded build.

I'd like to add some more code to this so that there's more junk to compile but don't have a good idea of what to add. `XmlSerializer`? Suggestions welcome.

Cc @dotnet/ilc-contrib


src/tests/nativeaot/SmokeTests/Determinism/Determinism.cs
```diff
@@ -0,0 +1,22 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.IO;
+
+using var baseline = File.OpenRead("baseline.object");
+using var compare = File.OpenRead("compare.object");
```
SingleAccretion, line 8: Invoke these via reflection? Downstream, I have seen determinism issues (RyuJit-induced, since then fixed) pop up in reflection internals.

src/tests/nativeaot/SmokeTests/Determinism/Determinism.cs
```diff
@@ -0,0 +1,22 @@
...
+
+Console.WriteLine($"Baseline size: {baseline.Length}");
+Console.WriteLine($"Compare size: {compare.Length}");
+
+if (baseline.Length != compare.Length)
+    throw new Exception("Different sizes");
+
+for (int i = 0; i < baseline.Length; i++)
```
jkotas, line 16:
```suggestion
long length = baseline.Length;
for (int i = 0; i < length; i++)
```

jkotas:
> more junk to compile

Roslyn?

MichalStrehovsky:
> > more junk to compile
> 
> Roslyn?

The unfortunate part about doing that would be that we then compile Roslyn single-threaded. We want more stuff, but not _too much_ stuff.

MichalStrehovsky:
I made this call into the other test that does a bunch of random things and exercises things like static virtuals, generic virtual methods, etc.

jkotas:
Looks like it found a bug:
```
   nativeaot\SmokeTests\Determinism\Determinism\Determinism.cmd [FAIL]
      Unhandled exception. System.Exception: Different sizes
         at Program.<Main>$(String[] args) + 0x32e
```

MichalStrehovsky:
> Looks like it found a bug:
> 
> ```
>    nativeaot\SmokeTests\Determinism\Determinism\Determinism.cmd [FAIL]
>       Unhandled exception. System.Exception: Different sizes
>          at Program.<Main>$(String[] args) + 0x32e
> ```

And of course the diff is in the `.debug$S` section because any other section would be too easy to root cause 😭

MichalStrehovsky:
> > Looks like it found a bug:
> > ```
> >    nativeaot\SmokeTests\Determinism\Determinism\Determinism.cmd [FAIL]
> >       Unhandled exception. System.Exception: Different sizes
> >          at Program.<Main>$(String[] args) + 0x32e
> > ```
> 
> And of course the diff is in the `.debug$S` section because any other section would be too easy to root cause 😭

I was wrong. The debug info part is just a symptom of the problem.

The problem is that codegen wasn't deterministic. Filed #93843. @SingleAccretion was right that the reflection stack is problematic.

jkotas (approved):
LGTM. Thanks!

PR #93606, opened 2023-10-17
[mono][jit] Partially implemented logical immediates on arm64

jandupej:
This implements a subset of https://developer.arm.com/documentation/ddi0596/2020-12/Base-Instructions/AND--immediate---Bitwise-AND--immediate-- (incl. `ands`, `eor`, `orr`). If the immediate value can be expressed as `(1<<n) - 1`, the new codegen is used. Although this covers only a few of the cases supported by the instruction, the instances where we are trying to mask away some of the upper bits seem prevalent (in our codebase) and are handled here.

E.g. clamping the shift amount in long shifts:
```asm
0000000000000018        mov     x30, #0x3f
000000000000001c        and     w1, w0, w30
```

becomes
```asm
0000000000000018        and     w1, w0, #0x3f
```

EDIT: The recognized bit pattern is now expanded to `0*1+0*`. A plain-ish HelloWorld build revealed almost 400 instances of this pattern.

### DIFF
--- a/src/mono/mono/mini/mini-arm64.c
+++ b/src/mono/mono/mini/mini-arm64.c
@@ -26,6 +26,7 @@
 #include <mono/utils/mono-mmap.h>
 #include <mono/utils/mono-memory-model.h>
 #include <mono/utils/mono-hwcap.h>
+#include <mono/utils/mono-bitutils.h>
 #include <mono/metadata/abi-details.h>
 #include <mono/metadata/tokentype.h>
 #include <mono/metadata/marshal-shared.h>
@@ -33,10 +34,6 @@
 
 #include "interp/interp.h"
 
-#ifdef _MSC_VER
-// for _BitScanFroward
-#include <intrin.h>
-#endif
 
 // The following defines are here to support the inclusion of simd-arm64.h
 #define EXPAND(x) x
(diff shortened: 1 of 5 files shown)


src/mono/mono/mini/mini-arm64.c
```diff
@@ -620,62 +625,131 @@ emit_subx_sp_imm (guint8 *code, int imm)
...
+
+static int 
+num_trailing_zeros (guint64 imm)
+{
+#ifdef _MSC_VER
+	long index = 0;
+	if ( _BitScanForward( &index, value ) )
+		return index;
+	else
+		return 32;
```
lambdageek, line 637: 64?

src/mono/mono/mini/mini-arm64.c
```diff
@@ -620,62 +625,131 @@ emit_subx_sp_imm (guint8 *code, int imm)
...
+#else
+	return __builtin_ctz(imm);
```
lambdageek, line 639: Should be `ctzl`
Ctz is for ints

src/mono/mono/mini/mini-arm64.c
```diff
@@ -620,62 +625,131 @@ emit_subx_sp_imm (guint8 *code, int imm)
 	return code;
 }
 
+
+static int 
+num_trailing_zeros (guint64 imm)
```
lambdageek, line 630: Can we put this in a header in mono/utils (or in minipal) so we can use it in more places

lambdageek (commented):
ctz is for 32-bit inputs

vargaz (approved):

jandupej:
CI failures are not related. Merging.

PR #93610, opened 2023-10-17
Consider atomics throwing

### LINKED ISSUE #44087: [RyuJIT] Unnecessary nullcheck in Exchange_long benchmark
[dotnet/performance/Perf.Interlocked.cs#L39](https://github.com/dotnet/performance/blob/8aed638c9ee65c034fe0cca4ea2bdc3a68d2a6b5/src/benchmarks/micro/libraries/System.Threading/Perf.Interlocked.cs#L39):
```csharp
private long _long = 0;

public long Exchange_long() => Interlocked.Exchange(ref _long, 1);
```
Codegen for `Exchange_long`:
```asm
       cmp      dword ptr [rcx], ecx ;; <--- nullcheck
       lea      rax, bword ptr [rcx+8]
       mov      edx, 1
       xchg     qword ptr [rax], rdx
       mov      rax, rdx
       ret
```
this code never throws NRE.
Is it `#define CONSERVATIVE_NULL_CHECK_BYREF_CREATION 1`'s fault?

/cc @dotnet/jit-contrib @sandreenko 

UPD: Same in MONO:
```
   0:   50                      push   %rax
   1:   48 85 ff                test   %rdi,%rdi  ;; <----
   4:   74 0b                   je     11 <xchange_long___+0x11>
   6:   b8 01 00 00 00          mov    $0x1,%eax
   b:   48 87 47 10             xchg   %rax,0x10(%rdi)
   f:   59                      pop    %rcx
  10:   c3                      retq
  11:   48 b8 60 e9 74 54 13    movabs $0x56135474e960,%rax
  18:   56 00 00
  1b:   bf 32 01 00 00          mov    $0x132,%edi
  20:   ff 10                   callq  *(%rax)
```

category:cq
theme:optimization
skill-level:beginner
cost:small
impact:small

### LINKED ISSUE #92784: Atomics are missing from `OperExceptions`

SingleAccretion:
Continuation of #93295.

Fixes #92784.
Fixes #44087.

A few diffs are expected due to the early propagation change.

### DIFF
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -2295,27 +2295,27 @@ void Compiler::optAssertionGen(GenTree* tree)
             }
             break;
 
-        case GT_BLK:
         case GT_IND:
+            // Dynamic block copy sources could be zero-sized and so should not generate assertions.
+            if (tree->TypeIs(TYP_STRUCT))
+            {
+                break;
+            }
+            FALLTHROUGH;
+
+        case GT_XAND:
+        case GT_XORR:
+        case GT_XADD:
+        case GT_XCHG:
+        case GT_CMPXCHG:
+        case GT_BLK:
         case GT_STOREIND:
         case GT_STORE_BLK:
-            // Dynamic block copy sources should not generate non-null assertions; we detect them via NO_CSE.
-            if (tree->CanCSE())
-            {
-                assertionInfo = optCreateAssertion(tree->AsIndir()->Addr(), nullptr, OAK_NOT_EQUAL);
-            }
-            break;
-
+        case GT_NULLCHECK:
         case GT_ARR_LENGTH:
         case GT_MDARR_LENGTH:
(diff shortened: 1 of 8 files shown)


SingleAccretion:
x86 SPMI looks to have timed out; ARM64 SPMI looks to be #93527.

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=442311&view=results), @dotnet/jit-contrib.

jakobbotsch (approved):
LGTM, thanks!

Issue #93581, opened 2023-10-17
Interop.cgroups.FindCGroupsVersion broken for cgroups v1


janvorli:
While working on some cgroups related changes, I've found that the `Interop.cgroups.FindCGroupsVersion` was unable to detect cgroups v1. I've investigated the issue and the problem is caused by the fact that the textual result of `new DriveInfo(SysFsCgroupFileSystemPath).DriveFormat` returns `udev` instead of `tmpfs`. The reason is that we get the format string from the `UnixFileSystemTypes` enum based on the filesystem "magic" constant and the `udev` and `tmpfs` (and few others) share the same value. 
The `DriveInfo` calls `Interop.Sys.GetFormatInfoForMountPoint` that converts the filesystem numeric type (the magic constant) using this:
https://github.com/dotnet/runtime/blob/ccc475ed5646abe71ef596702d06e78e34981736/src/libraries/Common/src/Interop/Unix/System.Native/Interop.MountPoints.FormatInfo.cs#L62-L63

The `Interop.cgroups.FindCGroupsVersion` uses this switch to detect cgroups version:
https://github.com/dotnet/runtime/blob/ccc475ed5646abe71ef596702d06e78e34981736/src/libraries/Common/src/Interop/Linux/cgroups/Interop.cgroups.cs#L115-L120

Since the name returned by the `Enum.ToString` is "udev", it fails to detect the cgroups 1.

janvorli:
Fixed by #93611

PR #93611, opened 2023-10-17
Enable cgroups hierarchical memory limits support

janvorli:
Cgroups can limit memory hierarchically. That means that ancestor cgroups can apply limits that are lower than the current cgroup's memory limit. .NET runtime is not honoring these hierarchical limits with either cgroups v1 or cgroups v2.
This change fixes it. The implementation is different for cgroups V1 and V2 because cgroups v1 provide aggregated hierarchical limit value in the memory.stat, but in cgroups v2, runtime needs to walk the cgroup hierarchy and find the lowest limit in it.


src/libraries/Common/src/Interop/Linux/cgroups/Interop.cgroups.cs
```diff
@@ -116,6 +209,7 @@ private static CGroupVersion FindCGroupVersion()
                 {
                     "cgroup2fs" => CGroupVersion.CGroup2,
                     "tmpfs" => CGroupVersion.CGroup1,
+                    "udev" => CGroupVersion.CGroup1,
```
janvorli, line 212: This is a hotfix for a bug in cgroup v1 handling (https://github.com/dotnet/runtime/issues/93581)

src/coreclr/pal/src/misc/cgroup.cpp
```diff
@@ -40,6 +40,8 @@ SET_DEFAULT_DEBUG_CHANNEL(MISC);
 #define CGROUP_MEMORY_STAT_FILENAME "/memory.stat"
 #define CGROUP1_MEMORY_USAGE_FILENAME "/memory.usage_in_bytes"
 #define CGROUP2_MEMORY_USAGE_FILENAME "/memory.current"
+#define CGROUP1_MEMORY_USE_HIERARCHY_FILENAME "/memory.use_hierarchy"
```
jkotas, line 43: Is the cgroup memory limit implementation in the Win32 PAL used for anything?

src/libraries/Common/src/Interop/Linux/cgroups/Interop.cgroups.cs
jkotas: Is this fix reliable? From #93581:

> The reason is that we get the format string from the UnixFileSystemTypes enum based on the filesystem "magic" constant and the udev and tmpfs (and few others) share the same value.

It means that the `DriveInfo.DriveFormat` is going to return one of the strings at random. It can start returning yet another string that happens to match the magic value tomorrow.

I think there are two problems here:
-  `DriveInfo.DriveFormat` should be more deterministic
- We probably should not be using `DriveInfo.DriveFormat` in the first place. It is unnecessarily complex API for what we need to do here. Can we call the relevant System.Native shim API directly here?

janvorli: I have actually put this change in originally just to enable me to test that it works correctly in cgroups v1 modulo this bug. I was going to leave the real fix for a separate PR and actually probably remove this hotfix from this PR. I agree that calling the relevant System.Native shim API directly here sounds like the right fix.

src/coreclr/pal/src/misc/cgroup.cpp
janvorli: That's a good point, I haven't realized we've stopped using that once we've moved embedded GC to its own PAL.

src/libraries/Common/src/Interop/Linux/cgroups/Interop.cgroups.cs
```diff
@@ -29,26 +29,119 @@ internal enum CGroupVersion { None, CGroup1, CGroup2 };
...
+                        if (line.StartsWith(fieldName))
+                        {
+                            bool foundFieldValue = ulong.TryParse(line.AsSpan(fieldName.Length), out val);
+                            return foundFieldValue;
+                        }
+                    }
+                }
+                catch (Exception e)
+                {
+                    Debug.Fail($"Failed to read \"{path}/memory.stat\": {e}");
```
jkotas, line 80: Would it be better to let the exception to propagate instead of failing silently?

jkotas:
Build breaks. Lgtm otherwise.

src/libraries/Common/src/Interop/Linux/cgroups/Interop.cgroups.cs
janvorli: When this exception occurs, it means that the cgroups are somehow badly broken on the device. It is hard to imagine how that would happen unless the kernel is broken. I don't think we want to cause user applications to break just because of that. The c++ code also handles failures to read a value as if it was not present / set.

jkotas: If the kernel is broken, all bets are off.

I think that the most likely cause of the exception thrown here is our own bug that this catch sweeps under the rug.

jkotas (approved):

src/libraries/Common/src/Interop/Linux/cgroups/Interop.cgroups.cs
jkotas: The best practice is to avoid defensive catches like this, and allow the bug to be found and fixed.

janvorli: We would also need to update the Process.MinWorkingSet doc - the list of exceptions it can throw.

src/libraries/System.Runtime.InteropServices.RuntimeInformation/tests/DescriptionNameTests.cs
```diff
@@ -51,11 +51,14 @@ public void DumpRuntimeInformationToConsole()
 
             Console.WriteLine($"### CURRENT DIRECTORY: {Environment.CurrentDirectory}");
 
-            Console.WriteLine($"### CGROUPS VERSION: {Interop.cgroups.s_cgroupVersion}");
-            string cgroupsLocation = Interop.cgroups.s_cgroupMemoryLimitPath;
-            if (cgroupsLocation != null)
+            if (osd.Contains("Linux"))
```
jkotas, line 54:
```suggestion
            if (OperatingSystem.IsLinux())
```
`OSDescription` is a free-form string. It is not meant to be used for checks like this.

src/libraries/Common/src/Interop/Linux/cgroups/Interop.cgroups.cs
jkotas: ok, it can stay as is

jkotas (approved):

src/libraries/System.Runtime.InteropServices.RuntimeInformation/tests/DescriptionNameTests.cs
janvorli: Thanks, I've mimicked what was used below in this file. I'm happy to use the other form.

lewing:
This broke the outloop palsuite tests because of a removed symbol

https://dev.azure.com/dnceng-public/public/_build/results?buildId=444318&view=logs&jobId=efb0e688-765b-56f1-1b94-fb914f4a08b7&j=efb0e688-765b-56f1-1b94-fb914f4a08b7&t=0e66e276-1ec2-597c-a7ec-8bd834919a09

```
 /__w/1/s/src/coreclr/pal/tests/palsuite/miscellaneous/CGroup/test1/test.cpp:37:24: error: use of undeclared identifier 'PAL_GetRestrictedPhysicalMemoryLimit'
      size_t mem_limit = PAL_GetRestrictedPhysicalMemoryLimit();
                         ^
```

janvorli:
Ah, I am sorry for that, Iet me fix it right away.

janvorli:
PR with the fix is here: https://github.com/dotnet/runtime/pull/93757

PR #93619, opened 2023-10-17
Add tests for Task.WhenAll

manandre:
Follow the same pattern as used for `Task.WhenAny` in https://github.com/dotnet/runtime/blob/c657cb0265a36b1f24bcd4d9b61d249ce3b68e58/src/libraries/System.Threading.Tasks/tests/MethodCoverage.cs#L181-L279

### DIFF
--- a/src/libraries/System.Runtime/tests/System.Threading.Tasks.Tests/MethodCoverage.cs
+++ b/src/libraries/System.Runtime/tests/System.Threading.Tasks.Tests/MethodCoverage.cs
@@ -444,7 +444,7 @@ namespace TaskCoverage
                 Assert.False(twa.IsCompleted);
                 t2.SetResult(2);
                 Assert.True(twa.IsCompletedSuccessfully);
-                Assert.Equal([1, 2], twa.Result);
+                Assert.Equal(new int[] { 1, 2 }, twa.Result);
             }
 
             // Non-generic, second completes first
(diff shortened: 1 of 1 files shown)


ericstj:
I merged in the rename from https://github.com/dotnet/runtime/pull/94644 to resolve conflicts.  @stephentoub can you have a look at these tests?

src/libraries/System.Runtime/tests/System.Threading.Tasks.Tests/MethodCoverage.cs
```diff
@@ -352,5 +352,581 @@ public static void FromAsync()
...
+
+                Task twa = Task.WhenAll((Task)t1.Task, (Task)t2.Task);
+                Assert.False(twa.IsCompleted);
+                t1.SetException(e1);
+                Assert.False(twa.IsCompleted);
+                t2.SetCanceled(ct2);
+                Assert.True(twa.IsCompleted);
+                Assert.True(twa.IsFaulted);
+                Assert.Equal(e1, twa.Exception?.InnerException);
+                Assert.False(twa.IsCanceled);
```
stephentoub, line 880: Some of these checks aren't really tests of WhenAll; a Task can't be in multiple states at the same time, so it's a general constraint of Task that IsCanceled would never be true if IsFaulted is true.

stephentoub (approved):
Looks good. Thanks for improving the test suite.

src/libraries/System.Runtime/tests/System.Threading.Tasks.Tests/MethodCoverage.cs
manandre: Useless assertions are removed.

PR #93629, opened 2023-10-17
[wasm] Use locally built ILLink.Tasks with WBT

radical:


src/mono/wasm/Wasm.Build.Tests/data/Workloads.Directory.Build.targets
```diff
@@ -3,5 +3,12 @@
     <Message Text="** MicrosoftNetCoreAppRuntimePackDir : '@(ResolvedRuntimePack -> '%(PackageDirectory)')'" Importance="High" Condition="@(ResolvedRuntimePack->Count()) > 0" />
   </Target>
 
+  <Target Name="_OverrideILLinkPack" BeforeTargets="ProcessFrameworkReferences">
+    <ItemGroup>
+      <KnownILLinkPack Update="Microsoft.NET.ILLink.Tasks" Condition="%(KnownILLinkPack.TargetFramework) == 'net8.0'" ILLinkPackVersion="$(_RuntimePackInWorkloadVersionCurrent)" />
+    </ItemGroup>
+  </Target>
```
ViktorHofer, line 10: cc @sbomer. We already have logic to consume the live illink pack: https://github.com/dotnet/runtime/blob/main/eng/liveILLink.targets

Why do we need this duplication?

radical: This is for Wasm.Build.Tests which run in a special way on helix:

- Use a sdk with `wasm-tools` workload installed
- The installed workload packs are from locally built nugets (thus the sdk will have binaries from artifacts)
- then each test creates a project, and builds it using that sdk
- So, we can't directly depend on `artifacts/`, or `liveILLink.targets`

The problem being solved here is that the workload itself does not contain the version for ILLink.tasks, the sdk does. So, for these tests you end up with sdk which uses the linker tasks from the nuget version specified in the sdk itself, and not the built nugets.
And to solve that we build the nuget for illink.tasks (with a version like `9.0.0-ci`) and then edit the version for `KnownILLinkPack`, allowing rest of the build to run normally like it would for a user project.

PR #93635, opened 2023-10-17
Implement the ComVariant struct, the ComVariantMarshaller type, and generator integration

### LINKED ISSUE #89543: [API Proposal]: Marshaller for OLE `VARIANT`
### Background and motivation

The `VARIANT` type is a discriminated union data structure that is common in COM interop. The built-in system has various APIs off of `Marshal` that help with marshalling of `VARIANT` (for example, [`Marshal.GetObjectForNativeVariant()`](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.marshal.getobjectfornativevariant)). There is also support in the built-in COM Interop for marshalling as a `VARIANT` (for example, [`[MarshalAs(UnmanagedType.Struct)] object a`](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.unmanagedtype)). This proposal is for add a marshaller for consumers that helps bridge the gap. Precise support what this type is permitted to hold will initially be limited to primitives and `IUnknown` types. However this can change but will largely be driven by user need/feedback.

The primary goal would be to support the WinForms effort to become AOT compatible.

Support for this would require usage of [`MarshalUsingAttribute`](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.marshalling.marshalusingattribute) as opposed to existing usages with [`MarshalAsAttribute`](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.marshalasattribute) in the example above. This decision is being made since current usages of the `MarshalAs` pattern are too permissive and we are attempting to narrow when usage is supported.

### API Proposal (Updated 21-09-2023)

```csharp
namespace System.Runtime.InteropServices.Marshalling;

[StructLayout(LayoutKind.Explicit)]
public struct OleVariant : IDisposable
{
    // private fields to match the layout of VARIANT

    public OleVariant();

    public void Dispose();

    // Determine the VarEnum for the variant based on T.
    // Respects the System.Runtime.InteropServices.*Wrapper types for specifying the variant type.
    // We will not allow System.Object as T (we will throw an InvalidOperationException).
    // If a System.Runtime.InteropServices.*Wrapper type that corresponds to a COM interface is specified,
    // the following behavior occurs:
    // - If Marshal.IsComObject() returns true, call Marshal.GetIUnknown/IDispatchForObject
    // - Otherwise, use ComInterfaceMarshaller<T> to get the interface pointer.
    // Types without a corresponding VarEnum type are not allowed.
    public static OleVariant Create<T>([DisallowNull] T value);

    // Explicitly specify the VarEnum for the variant and the raw value.
    // Throws if T is larger than the VARIANT's data union.
    // Throws for VT_DECIMAL.
    // Throws if T does not match the size of the data union member that vt corresponds to.
    public static OleVariant CreateRaw<T>(VarEnum vt, T rawValue) where T : unmanaged;

    // Create VT_NULL variant
    public static OleVariant Null { get; };

    // Get a .NET object of the specified type that represents the underlying VARIANT value.
    // For COM object-based VARIANTs, uses ComInterfaceMarshaller to create the managed object.
    // If passed one of the System.Runtime.InteropServices.*Wrapper types, the wrapper type will be used
    // if it matches the variant type. Using a wrapper type is not required.
    // If the requested type does not match the underlying type, throws an exception.
    public T As<T>();

    // For cases where the built-in Create/As methods are not sufficient:
    // Expose accessors for the underlying fields of the VARIANT struct
    // but in a limited fashion.
    // This will allow consumers such as WinForms to support VARIANT types
    // that we do not plan to support immediately (like SAFEARRAY).
    public VarEnum VarType { get; }

    // Returns a reference to the data union within the variant.
    // Throws if T is larger than the VARIANT's data union.
    [UnscopedRef]
    public ref T GetRawDataRef<T>() where T : unmanaged;
}

// Let's start with a marshaller for object<->OleVariant as this will cover the majority of cases.
// We can add a generic marshaller for T<->OleVariant later if needed (e.g. for an IDispatch-based generator)
[CustomMarshaller(typeof(object?), MarshalMode.Default, typeof(OleVariantMarshaller))]
// To ensure correct behavior, we will need to update the generator to use the "ref" shape for both of the following cases.
// Is this behavior we want to generalize in an opt-in way? Or should we just special-case this case for now?
[CustomMarshaller(typeof(object?), MarshalMode.UnmanagedToManagedIn, typeof(OleVariantMarshaller.RefPropogate))]
[CustomMarshaller(typeof(object?), MarshalMode.UnmanagedToManagedRef, typeof(OleVariantMarshaller.RefPropogate))]
public static class OleVariantMarshaller
{
    public static OleVariant ConvertToUnmanaged(object? managed);
    public static object? ConvertToManaged(OleVariant unmanaged);
    public static void Free(OleVariant unmanaged);

    public struct RefPropogate
    {
        public void FromUnmanaged(OleVariant unmanaged);
        public void FromManaged(object? managed);
        public OleVariant ToUnmanaged();
        public object? ToManaged();
        public void Free();
    }
}
```

### API Usage

```csharp
OleVariant variant = OleVariant.Create(1.0m);
OleVariant variant2 = OleVariant.Create(new ErrorWrapper(0));


if (variant2.VarType == VarEnum.VT_ERROR)
{
    variant2.GetRawDataRef<int>() = 1;
}

[LibraryImport("Foo")]
public static partial Bar([MarshalUsing(typeof(OleVariantMarshaller))] object x);
```

### Alternative Designs

_No response_

### Risks

Minimal. The primary risk is narrowly supporting `VARIANT` in ways that limit future efforts. The WinForms team has rewritten a majority of the `VARIANT` support for their purposes so we can learn from them and mitigate much of the risk. Ideally the Interop team will simply take ownership of the WinForms code so the entire .NET ecosystem can benefit.

jkoritzinsky:
Implement the OleVariant type and the OleVariantMarshaller type as proposed in #89543

Behavior changes from the proposal:

`OleVariant` does not support source-generated or built-in COM interop types in its Create method.

`OleVaraintMarshaller` does not support built-in COM interop types.

`OleVariantMarshaller` will only support propagating back results through a byref VARIANT in `ref` scenarios, not `in` scenarios.

The interop source generators will provide an info-level diagnostic when a parameter using the `OleVariantMarshaller` (whether through MarshalAs or directly) is declared as an `in` parameter in an unmanaged-to-managed scenario to notify the user of the possibly unintuitive behavior/breaking change from built-in interop.

Additionally, this PR replaces the (previously internal) `System.Runtime.InteropServices.Variant` type and changes the places that used it (COM Events and `dynamic` support for COM objects, NativeAOT VARIANT marshalling) to use the new `OleVariant` type instead.

Fixes #89543


jkotas:
Is there a good reason why we have named this `OleVariant` and not `ComVariant`? 

`OLE` prefix does not sound right to me. [OLE](https://en.wikipedia.org/wiki/Object_Linking_and_Embedding) is a proprietary tech that originated in Office for embedding documents into each other. `VARIANT` might have been part of it many years ago, but it was later carved out together with `IDispatch` and friends. I do not see a single mention of `OLE` in the [current `VARIANT` docs](https://learn.microsoft.com/en-us/windows/win32/api/oaidl/ns-oaidl-variant).

jkoritzinsky:
We chose OLE as the online search results generally showed equal or better results for OLE, the Microsoft standard that mention VARIANT are the MS-OLE Automation standards, MS-OAUT, and the native DLL that implements most of the VARIANT support on Windows is the OLEAUT DLL, not ComBase. Additionally, other libraries (such as Delphi and Perl) use the name OleVariant or OLE::Variant as well.

jkotas:
> the Microsoft standard that mention VARIANT are the MS-OLE Automation standards,

OLE Automation is the old obsolete name. I believe that the official name has been just "Automation" for many years now.

jkotas:
> Additionally, other languages (such as Delphi and Perl) use the name OleVariant 

Yes, it is expected since these languages introduced the type when the tech still used the now obsolete name.

I guess the question is whether we should stick with obsolete names when naming new APIs; or whether we should prefer current names. I think it should be the later. `OperatingSystem.IsMacOS` is an example where we have used the current name; even though many places in .NET use the obsolete name (OSX).

jkoritzinsky:
We also use the term OLE Variant throughout the runtime's source code. COM Variant in CoreCLR's code actually refers to a different type (the internal System.Variant type that's used by the runtime to convert between OLE VARIANTs and .NET objects).

Also, the Microsoft docs still refer to OLE Automation as OLE Automation: https://learn.microsoft.com/en-us/openspecs/windows_protocols/MS-OAUT/bbb05720-f724-45c7-8d17-f83c3d1a3961

jkotas:
> We also use the term OLE Variant throughout the runtime's source code.

Yes, this code is from the late 1990 when the tech used the now obsolete name, and nobody bothered to rename it. We have number of obsolete names like that.

> Also, the Microsoft docs still refer to OLE Automation as OLE Automation: https://learn.microsoft.com/en-us/openspecs/windows_protocols/MS-OAUT/bbb05720-f724-45c7-8d17-f83c3d1a3961

These docs were created for compliance with [consent decree](https://en.wikipedia.org/wiki/United_States_v._Microsoft_Corp.). They are frozen in time.

The live docs at https://learn.microsoft.com/en-us/windows/win32/api/oaidl/ns-oaidl-variant are better up-to-date reference.

AaronRobinsonMSFT:
> Is there a good reason why we have named this OleVariant and not ComVariant?

I'm fine with `ComVariant`. It would align with the `ComWrappers` name et al.

jkoritzinsky:
> The live docs at [learn.microsoft.com/en-us/windows/win32/api/oaidl/ns-oaidl-variant](https://learn.microsoft.com/en-us/windows/win32/api/oaidl/ns-oaidl-variant) are better up-to-date reference.


The Automation page that is the parent of this page refers to the technology as OLE Automation (to distinguish it from UI Automation).

In any case, I'll email the API Review Board to get the name change approved.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+    /// A type that represents an OLE VARIANT in managed code.
+    /// </summary>
+    [StructLayout(LayoutKind.Explicit)]
+    public struct OleVariant : IDisposable
+    {
+        private const VarEnum VT_VERSIONED_STREAM = (VarEnum)73;
+#if DEBUG
+        static unsafe OleVariant()
+        {
+            // Variant size is the size of 4 pointers (16 bytes) on a 32-bit processor,
```
AaronRobinsonMSFT, line 19: Reference to header file or doc? Reference what we are matching here.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+
+namespace System.Runtime.InteropServices.Marshalling
+{
+    /// <summary>
+    /// A type that represents an OLE VARIANT in managed code.
+    /// </summary>
+    [StructLayout(LayoutKind.Explicit)]
+    public struct OleVariant : IDisposable
+    {
+        private const VarEnum VT_VERSIONED_STREAM = (VarEnum)73;
```
AaronRobinsonMSFT, line 15: Reference?

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+            public IntPtr _record;
+            public IntPtr _recordInfo;
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
+        private struct Blob
+        {
+            public int _size;
+            public IntPtr _data;
+        }
```
AaronRobinsonMSFT, line 65: Is there a reason we can't use `void*` here? I think we prefer that when expressing these things.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+            public IntPtr _stream;
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
+        private unsafe struct ClipboardData
+        {
+            public uint _size;
+            public int _format;
+            public IntPtr _data;
+        }
```
AaronRobinsonMSFT, line 89: Marked `unsafe`, but not using `void*`.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+
+        /// <summary>
+        /// Release resources owned by this <see cref="OleVariant"/> instance.
+        /// </summary>
+        public unsafe void Dispose()
+        {
+#if TARGET_WINDOWS
+            fixed (void* pThis = &this)
+            {
+                Interop.Ole32.PropVariantClear((nint)pThis);
```
AaronRobinsonMSFT, line 129: Document why we are using `PropVariantClear()` rather than `VariantClear()`.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+        /// <summary>
+        /// Create an <see cref="OleVariant"/> instance from the specified value.
+        /// </summary>
+        /// <typeparam name="T">The type of the specified value.</typeparam>
+        /// <param name="value">The value to wrap in an <see cref="OleVariant"/>.</param>
+        /// <returns>An <see cref="OleVariant"/> that contains the provided value.</returns>
+        /// <exception cref="ArgumentException">When <typeparamref name="T"/> does not directly correspond to a <see cref="VarEnum"/> variant type.</exception>
+        public static OleVariant Create<T>([DisallowNull] T value)
+        {
+            Unsafe.SkipInit(out OleVariant variant);
```
AaronRobinsonMSFT, line 243: Is there a reason we want to declare this this way? It isn't obvious and is borderline for something we should be doing.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+            {
+                throw new ArgumentException("VT_DECIMAL is not supported in CreateRaw. Use the Create method.", nameof(vt));
+            }
+            if (vt == VarEnum.VT_VARIANT)
+            {
+                throw new ArgumentException("VT_VARIANT is not supported in variants.", nameof(vt));
+            }
+            if (vt.HasFlag(VarEnum.VT_ARRAY) && !OperatingSystem.IsWindows())
+            {
+                throw new PlatformNotSupportedException("OleVariants containing SAFEARRAYs are not supported on this platform");
```
AaronRobinsonMSFT, line 375: Localize error strings.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+            }
+
+            Unsafe.SkipInit(out OleVariant value);
```
AaronRobinsonMSFT, line 378: Again with the clever declaration.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+                (VarEnum.VT_I8 or VarEnum.VT_UI8 or VarEnum.VT_R8 or VarEnum.VT_DATE, 8) => rawValue,
+                (VarEnum.VT_INT or VarEnum.VT_UINT or VarEnum.VT_UNKNOWN or VarEnum.VT_DISPATCH or VarEnum.VT_LPSTR or VarEnum.VT_BSTR or VarEnum.VT_LPWSTR or VarEnum.VT_SAFEARRAY
+                    or VarEnum.VT_CLSID or VarEnum.VT_STREAM or VarEnum.VT_STREAMED_OBJECT or VarEnum.VT_STORAGE or VarEnum.VT_STORED_OBJECT or VarEnum.VT_CF or VT_VERSIONED_STREAM, _) when sizeof(T) == nint.Size => rawValue,
+                (VarEnum.VT_CY or VarEnum.VT_FILETIME, 8) => rawValue,
+                (VarEnum.VT_RECORD, _) when sizeof(T) == sizeof(Record) => rawValue,
+                _ when vt.HasFlag(VarEnum.VT_BYREF) && sizeof(T) == nint.Size => rawValue,
+                _ when vt.HasFlag(VarEnum.VT_VECTOR) && sizeof(T) == sizeof(Vector<byte>) => rawValue,
+                _ when vt.HasFlag(VarEnum.VT_ARRAY) && sizeof(T) == nint.Size => rawValue,
+                (VarEnum.VT_BLOB or VarEnum.VT_BLOB_OBJECT, _) when sizeof(T) == sizeof(Blob) => rawValue,
+                _ => throw new ArgumentException("Size of T should be the same size as the value specified by vt")
```
AaronRobinsonMSFT, line 394: Localize message.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+
+            Unsafe.SkipInit(out OleVariant value);
+            value.VarType = vt;
+            value.GetRawDataRef<T>() = (vt, sizeof(T)) switch
+            {
+                (VarEnum.VT_I1 or VarEnum.VT_UI1, 1) => rawValue,
+                (VarEnum.VT_I2 or VarEnum.VT_UI2 or VarEnum.VT_BOOL, 2) => rawValue,
+                (VarEnum.VT_ERROR or VarEnum.VT_HRESULT or VarEnum.VT_I4 or VarEnum.VT_UI4 or VarEnum.VT_R4 or VarEnum.VT_INT or VarEnum.VT_UINT, 4) => rawValue,
+                (VarEnum.VT_I8 or VarEnum.VT_UI8 or VarEnum.VT_R8 or VarEnum.VT_DATE, 8) => rawValue,
+                (VarEnum.VT_INT or VarEnum.VT_UINT or VarEnum.VT_UNKNOWN or VarEnum.VT_DISPATCH or VarEnum.VT_LPSTR or VarEnum.VT_BSTR or VarEnum.VT_LPWSTR or VarEnum.VT_SAFEARRAY
```
AaronRobinsonMSFT, line 386: `VarEnum.VT_INT or VarEnum.VT_UINT` are we trying to handle the `IntPtr` case here? If so, that warrants a comment and a specific test.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+        /// <summary>
+        /// A <see cref="OleVariant"/> instance that represents a null value with <see cref="VarEnum.VT_NULL"/> type.
+        /// </summary>
+        public static OleVariant Null { get; } = new() { VarType = VarEnum.VT_NULL };
+
+        private readonly void ThrowIfNotVarType(params VarEnum[] requiredType)
+        {
+            if (Array.IndexOf(requiredType, VarType) == -1)
+            {
+                throw new InvalidOperationException($"Variant type {VarType} cannot be cast to any of the supported variant types: [{string.Join(", ", requiredType)}]");
```
AaronRobinsonMSFT, line 409: Localize.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+        /// Create a managed value based on the value in the <see cref="OleVariant"/> instance.
+        /// </summary>
+        /// <typeparam name="T">The managed type to create an instance of.</typeparam>
+        /// <returns>The managed value contained in this variant.</returns>
+        /// <exception cref="ArgumentException">When <typeparamref name="T"/> does not match the <see cref="VarType"/> of the <see cref="OleVariant"/>.</exception>
+        public readonly T As<T>()
+        {
+            if (VarType == VarEnum.VT_EMPTY)
+            {
+                return default!;
```
AaronRobinsonMSFT, line 424: Why is the `!` needed?

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+            {
+                ThrowIfNotVarType(VarEnum.VT_I8);
+                return (T)(object)_typeUnion._unionTypes._i8;
+            }
+            else if (typeof(T) == typeof(ulong))
+            {
+                ThrowIfNotVarType(VarEnum.VT_UI8);
+                return (T)(object)_typeUnion._unionTypes._ui8;
+            }
+            throw new ArgumentException("Unsupported type", nameof(T));
```
AaronRobinsonMSFT, line 529: Localize.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
```diff
@@ -0,0 +1,560 @@
...
+        /// <returns>A reference to the storage location within this <see cref="OleVariant"/>.</returns>
+        /// <exception cref="ArgumentException"><typeparamref name="T"/> is <see cref="decimal"/> or larger than the storage space in an <see cref="OleVariant"/>.</exception>
+        [UnscopedRef]
+        public unsafe ref T GetRawDataRef<T>()
+            where T : unmanaged
+        {
+            ArgumentOutOfRangeException.ThrowIfGreaterThan(Unsafe.SizeOf<T>(), sizeof(UnionTypes), nameof(T));
+            if (typeof(T) == typeof(decimal))
+            {
+                throw new ArgumentException("VT_DECIMAL is not supported in GetRawDataRef. Use the As method.", nameof(T));
```
AaronRobinsonMSFT, line 555: Localize.

src/libraries/System.Runtime.InteropServices/gen/Microsoft.Interop.SourceGeneration/TypeNames.cs
```diff
@@ -311,5 +311,7 @@ public static string MarshalEx(InteropGenerationOptions options)
         public const string CallConvSuppressGCTransitionName = "System.Runtime.CompilerServices.CallConvSuppressGCTransition";
         public const string CallConvMemberFunctionName = "System.Runtime.CompilerServices.CallConvMemberFunction";
         public const string Nint = "nint";
+
```
AaronRobinsonMSFT, line 314:
```suggestion
```

src/libraries/System.Runtime.InteropServices/ref/System.Runtime.InteropServices.cs
```diff
@@ -460,6 +460,23 @@ public partial interface IUnmanagedVirtualMethodTableProvider
...
     }
+    [System.Runtime.InteropServices.Marshalling.CustomMarshallerAttribute(typeof(object), System.Runtime.InteropServices.Marshalling.MarshalMode.Default, typeof(System.Runtime.InteropServices.Marshalling.OleVariantMarshaller))]
+    [System.Runtime.InteropServices.Marshalling.CustomMarshallerAttribute(typeof(object), System.Runtime.InteropServices.Marshalling.MarshalMode.UnmanagedToManagedRef, typeof(System.Runtime.InteropServices.Marshalling.OleVariantMarshaller.RefPropogate))]
+    public static partial class OleVariantMarshaller
+    {
+        public static System.Runtime.InteropServices.Marshalling.OleVariant ConvertToUnmanaged(object? managed) { throw null; }
+        public static object? ConvertToManaged(System.Runtime.InteropServices.Marshalling.OleVariant unmanaged) { throw null; }
+        public static void Free(System.Runtime.InteropServices.Marshalling.OleVariant unmanaged) { }
+
+        public struct RefPropogate
```
AaronRobinsonMSFT, line 471:
```suggestion
        public struct RefPropagate
```

src/libraries/System.Runtime.InteropServices/src/System/Runtime/InteropServices/Marshalling/OleVariantMarshaller.cs
```diff
@@ -0,0 +1,332 @@
...
+                    return OleVariant.Null;
+            }
+#pragma warning restore CS0618 // Type or member is obsolete
+
+            if (TryCreateOleVariantForInterfaceWrapper(managed, out OleVariant variant))
+            {
+                return variant;
+            }
+
+            throw new ArgumentException("Type of managed object is not supported for marshalling as OleVariant.", nameof(managed));
```
AaronRobinsonMSFT, line 78: Localize

src/libraries/System.Runtime.InteropServices/src/System/Runtime/InteropServices/Marshalling/OleVariantMarshaller.cs
```diff
@@ -0,0 +1,332 @@
...
+                    return unmanaged.As<int>();
+                case VarEnum.VT_CY:
+                    return unmanaged.As<CurrencyWrapper>().WrappedObject;
+                case VarEnum.VT_UNKNOWN:
+                case VarEnum.VT_DISPATCH:
+                    return StrategyBasedComWrappers.DefaultMarshallingInstance.GetOrCreateObjectForComInstance(unmanaged.GetRawDataRef<nint>(), CreateObjectFlags.Unwrap);
+                case VarEnum.VT_BYREF | VarEnum.VT_VARIANT:
+                    return ConvertToManaged(*(OleVariant*)unmanaged.GetRawDataRef<nint>());
+                case VarEnum.VT_BYREF | VarEnum.VT_I1:
+                    return *(sbyte*)unmanaged.GetRawDataRef<nint>();
```
AaronRobinsonMSFT, line 157: If `unmanaged` is a value type, that is passed into this function on the stack, how can we return a `ref` to one of its fields?

src/libraries/System.Runtime.InteropServices/src/System/Runtime/InteropServices/Marshalling/OleVariantMarshaller.cs
```diff
@@ -0,0 +1,332 @@
...
+                case VarEnum.VT_BYREF | VarEnum.VT_DATE:
+                    return DateTime.FromOADate(*(double*)unmanaged.GetRawDataRef<nint>());
+                case VarEnum.VT_BYREF | VarEnum.VT_ERROR:
+                    return *(int*)unmanaged.GetRawDataRef<nint>();
+                case VarEnum.VT_BYREF | VarEnum.VT_CY:
+                    return decimal.FromOACurrency(*(long*)unmanaged.GetRawDataRef<nint>());
+                case VarEnum.VT_BYREF | VarEnum.VT_UNKNOWN:
+                    return StrategyBasedComWrappers.DefaultMarshallingInstance.GetOrCreateObjectForComInstance(*(nint*)unmanaged.GetRawDataRef<nint>(), CreateObjectFlags.Unwrap);
+                default:
+                    throw new ArgumentException("Type of unmanaged variant is not supported for marshalling to a managed object.", nameof(unmanaged));
```
AaronRobinsonMSFT, line 191: Localize

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
jkoritzinsky: We'd end up having to cast it to `IntPtr` anyway in its only usage, so it doesn't feel worthwhile to switch.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
jkoritzinsky: I did it this way because I felt there's no reason for us to zero-initialize the whole variant type if we're going to set the specific fields that we want to set anyway.

The C# style rules/Roslyn rules we have in the repo suggest that we use the "declare a variable in the out parameter location" pattern if that's the first usage. I can change it if you prefer, but I personally much prefer this declaration pattern in general. I can understand it being confusing in this case though.

Looking through the repo, I see usages of this same pattern in Decimal.DecCalc.cs and throughout much of the Numerics/Intrinsics space. I don't see any usages of a different pattern when in this scenario (where the variable is being declared and not passed in as a byref parameter).

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
jkoritzinsky: This was a typo. I didn't mean to leave these options in this case. Looks like I'm missing some negative tests that I should add.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
jkoritzinsky: The return type is non-nullable, but `default` can be `null`. I'll update the API to return a nullable value.

AaronRobinsonMSFT: Because `T` isn't constrained, right?

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
AaronRobinsonMSFT: Okay.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/OleVariant.cs
jkoritzinsky: Yes. I changed the return type to `T?` (which is basically "T" but if it's a reference type it can be `null`) to be more accurate.

src/coreclr/System.Private.CoreLib/src/System/Variant.cs
```diff
@@ -34,7 +34,7 @@ internal struct Variant
         // What are the consequences of making this an enum?
         ///////////////////////////////////////////////////////////////////////
         // If you update this, update the corresponding stuff in OAVariantLib.cs,
-        // COMOAVariant.cpp (2 tables, forwards and reverse), and perhaps OleVariant.h
+        // COMOAVariant.cpp (2 tables, forwards and reverse), and perhaps ComVariant.h
```
AaronRobinsonMSFT, line 37: This should be reverted.

src/libraries/Common/src/System/Runtime/InteropServices/BuiltInVariantExtensions.cs
```diff
@@ -0,0 +1,178 @@
...
+        /// <returns></returns>
+        public static object? ToObject(this ref ComVariant variant)
+        {
+            // Check the simple case upfront
+            if (variant.VarType == VarEnum.VT_EMPTY)
+            {
+                return null;
+            }
+
+            switch (variant.VarType)
```
AaronRobinsonMSFT, line 141: Isn't this one of those places where the `return variant.VarType switch ...` logic would be a win?

src/libraries/System.Runtime.InteropServices/src/System/Runtime/InteropServices/Marshalling/ComVariantMarshaller.cs
```diff
@@ -0,0 +1,332 @@
...
+                case float f:
+                    return ComVariant.Create(f);
+                case double d:
+                    return ComVariant.Create(d);
+                case decimal d:
+                    return ComVariant.Create(d);
+                case bool b:
+                    return ComVariant.Create(b);
+                case char c:
+                    return ComVariant.Create((ushort)c);
```
AaronRobinsonMSFT, line 57: Is this supported in some way already? What is the user intent here?

src/libraries/System.Runtime.InteropServices/tests/System.Runtime.InteropServices.UnitTests/System/Runtime/InteropServices/ComVariantTests.cs
```diff
@@ -0,0 +1,493 @@
...
+            using ComVariant variant = ComVariant.CreateRaw(VarEnum.VT_LPWSTR, Marshal.StringToCoTaskMemUni(str));
+            Assert.Equal(VarEnum.VT_LPWSTR, variant.VarType);
+            Assert.Throws<InvalidOperationException>(variant.As<string>);
+            Assert.Equal(str, Marshal.PtrToStringUni(variant.GetRawDataRef<IntPtr>()));
+        }
+
+        [Fact]
+        public void FileTime_Raw()
+        {
+            long fileTime = 1039348523;
```
AaronRobinsonMSFT, line 293: Does this have a meaning of any sort? If so, please comment what is represents.

src/libraries/System.Runtime.InteropServices/src/System/Runtime/InteropServices/Marshalling/ComVariantMarshaller.cs
jkoritzinsky: The built-in VARIANT marshalling always treats `char` values as `VT_UI2`.

As this is a little unintuitive, I can remove this for V1 and we can add it based on user feedback.

src/libraries/System.Runtime.InteropServices/tests/System.Runtime.InteropServices.UnitTests/System/Runtime/InteropServices/ComVariantTests.cs
```diff
@@ -0,0 +1,493 @@
...
+            Assert.Throws<ArgumentException>("T", () => variant.As<Blob>());
+            Assert.Equal(blob, variant.GetRawDataRef<Blob>());
+        }
+
+        [StructLayout(LayoutKind.Sequential)]
+        private unsafe struct ClipboardData
+        {
+            public uint _size;
+            public int _format;
+            public IntPtr _data;
```
AaronRobinsonMSFT, line 382: I believe this is supposed to be an offset into the `ClipboardData` instance. Similar to the C style:

```c
struct ClipboardData
{
    uint32_t size;
    int32_t format;
    uint8_t data[];
};
```

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/ComVariant.cs
```diff
@@ -0,0 +1,566 @@
...
+            int variantSize = sizeof(ComVariant);
+            if (IntPtr.Size == 4)
+            {
+                Debug.Assert(variantSize == (4 * IntPtr.Size));
+            }
+            else
+            {
+                Debug.Assert(IntPtr.Size == 8);
+                Debug.Assert(variantSize == (3 * IntPtr.Size));
+            }
```
AaronRobinsonMSFT, line 32: So many bugs will be avoided. +1

AaronRobinsonMSFT (approved):

src/libraries/System.Runtime.InteropServices/tests/System.Runtime.InteropServices.UnitTests/System/Runtime/InteropServices/ComVariantTests.cs
jkoritzinsky: The definition in the Windows headers is the following:

```c
typedef struct tagCLIPDATA
    {
    ULONG cbSize;
    LONG ulClipFmt;
    /* [size_is] */ BYTE *pClipData;
    } 	CLIPDATA;
```

AaronRobinsonMSFT: Okay, looks good then.

src/libraries/Common/src/System/Runtime/InteropServices/BuiltInVariantExtensions.cs
```diff
@@ -146,7 +144,7 @@ public static unsafe void CopyFromIndirect(this ref ComVariant variant, object v
                 VarEnum.VT_UI8 => variant.As<ulong>(),
                 VarEnum.VT_INT => variant.As<int>(),
                 VarEnum.VT_UINT => variant.As<uint>(),
-                VarEnum.VT_BOOL => variant.As<short>() != -1,
+                VarEnum.VT_BOOL => variant.As<short>() != ComVariant.VARIANT_FALSE,
```
AaronRobinsonMSFT, line 147: Always compare to `VARIANT_TRUE`.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/ComVariant.cs
```diff
@@ -489,7 +492,7 @@ public static unsafe ComVariant CreateRaw<T>(VarEnum vt, T rawValue)
             {
                 // bool values in OLE VARIANTs are VARIANT_BOOL values.
                 ThrowIfNotVarType(VarEnum.VT_BOOL);
-                return (T)(object)(_typeUnion._unionTypes._bool != -1);
+                return (T)(object)(_typeUnion._unionTypes._bool != VARIANT_FALSE);
```
AaronRobinsonMSFT, line 495: `VARIANT_TRUE`

src/libraries/System.Runtime.InteropServices/src/System/Runtime/InteropServices/Marshalling/ComVariantMarshaller.cs
```diff
@@ -174,7 +177,7 @@ private static unsafe bool TryCreateOleVariantForInterfaceWrapper(object managed
                 case VarEnum.VT_BYREF | VarEnum.VT_DECIMAL:
                     return *(decimal*)unmanaged.GetRawDataRef<nint>();
                 case VarEnum.VT_BYREF | VarEnum.VT_BOOL:
-                    return *(short*)unmanaged.GetRawDataRef<nint>() != -1;
+                    return *(short*)unmanaged.GetRawDataRef<nint>() != VARIANT_FALSE;
```
AaronRobinsonMSFT, line 180: `VARIANT_TRUE`

src/libraries/Common/src/System/Runtime/InteropServices/BuiltInVariantExtensions.cs
jkotas: Is it the recommended pattern for `VARIANT_BOOL`?

The recommended pattern for `BOOL` is to compare against FALSE and native code static analysis tools will complain when it is compared against `TRUE`. I would expect that the recommended pattern for `VARIANT_BOOL` is going to be the same.

jkoritzinsky: I'm seeing more guidance to compare against VARIANT_FALSE instead of VARIANT_TRUE as that better catches "incorrect" cases from misbehaving APIs (as `false` and `VARIANT_FALSE` are the same value, whereas `VARIANT_TRUE` and `true` are different bit patterns)

AaronRobinsonMSFT: Historically on the VS team, the comparrison has always been against `VARIANT_TRUE` because that is the only "true" value. Non-zero values, other than `-1` are not true, but errors. This was the impetus for the statement. However, @jkoritzinsky pointed out that the runtime current compares against `VARIANT_FALSE`, which I find odd but if those are the semantics we want to accept then so be it.

jkoritzinsky:
Failure is https://github.com/dotnet/runtime/issues/93999. I'll look into moving the tests into a different assembly at least for now to unblock merging this PR.

jkoritzinsky:
Blocked on #94059

PR #93636, opened 2023-10-17
JIT: make bbID available in release, use it for pred list ordering

AndyAyersMS:
Avoid needing to reorder pred lists (in most cases) by relying on `bbID` for a persistent stable ordering.


AndyAyersMS:
@kunalspathak FYI.

Locally this didn't help much with TP; let's see what CI says.

kunalspathak:
> @kunalspathak FYI.
> 
> Locally this didn't help much with TP; let's see what CI says.

Seems like a nice TP win except that Arm64 shows more regressions. I glanced at few of the diffs and seems to be coming out of resolution moves, so perhaps happening because the block reordering, like we see in https://github.com/dotnet/runtime/pull/93371.

PR #93647, opened 2023-10-18
Fix: unmanaged memory leak on parsing ill-formed certificates

sstronin:
There is a memory leak - unmanaged! - because ParsePkcs12 is called from ctors and in case of an exception is thrown allocated resources won't be released inside the call nor by GC.


sstronin:
I've tried to do it the simplest way. The problem is in base class, true, but it also doesn't fully implement Dispose pattern so there will be more changes to separate managed/unmanaged cleanup - and without that it looks unsafe in finalizer.

martincostello:
> doesn't fully implement Dispose pattern

So shouldn't the fix be to make it so it does?

bartonjs (dismissed):

sstronin:
I'll rework it to proper Dispose pattern

bartonjs:
> I'll rework it to proper Dispose pattern

I think the better approach is to just remove the constructor argument and move it to a load function, that way Dispose will automatically fire and things just work.

```C#
using (Pkcs12Reader reader = new WhateverKind())
{
    reader.Load(data);
    ...
}
```

sstronin:
That could be done although it'll break class contract. But I think proper Dispose - with finalizer - should be implemented anyway since there is an unmanaged resource and the class is not sealed.

bartonjs:
This was addressed in #98331 by following the strategy outlined in https://github.com/dotnet/runtime/pull/93647#issuecomment-1906981179

PR #93648, opened 2023-10-18
[browser] streamline Task/Promise marshalling - atttemp 2

pavelsavara:
Second attempt to https://github.com/dotnet/runtime/pull/93010

This streamlines passing Promise/Task from two cross-boundary calls per promise argument to one.
- none when it's passed
- one when it's resolved, rejected or garbage collected

## Changes
- drop `create_task_callback`
- simplify `complete_task`
- simplify `mono_wasm_marshal_promise` and rename it to `mono_wasm_resolve_or_reject_promise`
- introduced `GCVHandle`
    - it's like `GCHandle` in that it keeps proxy of C# object alive while JS owns it.
    - but it's allocated ahead of time on JS side
- introduced `JSVHandle`
    - it's like `JSHandle` in that it keeps proxy of JS object alive while C# owns it.
    - but it's allocated ahead of time on C# side
- improved `GetTaskResultDynamic` to handle `void` Task
- introduced internal `JSObject.DisposeLocal`
    - we can dispose both sides during resolve/reject call without extra dispose call

## Interop "stack frame"
- fixed `IntPtr` and handles to be `I32` rather than `U32`
- changed `MarshalerType` from `I32` to `U8` size
- added `MarshalerType.TaskResolved` and `MarshalerType.TaskRejected`
    - this allows us to pass better metadata in the same `JSMarshalerArgument` slot
- moved `JSMarshalerArgument.ElementType` to different offset
    - to not clash with handles
- `JSMarshalerArgument.ElementType` now contains type of the promise's result
    - this is necessary for marshaling with dynamic type
- changed `JSBindingType` layout accordingly. 
    - In future this could also help with nested generic types.
- bumped schema version to `2`

## Other
- better type for `WeakRefInternal<T>`
- interop unit test no longer `ForceDisposeProxies` in the middle of the program because it can't balance C# side of handles


src/libraries/System.Runtime.InteropServices.JavaScript/src/System/Runtime/InteropServices/JavaScript/JSFunctionBinding.cs
```diff
@@ -42,13 +42,28 @@ internal struct JSBindingHeader
         internal struct JSBindingType
         {
             internal MarshalerType Type;
+            internal MarshalerType __ReservedB1;
```
maraf, line 45: nit: Maybe add comment that these are not used yet etc

maraf (approved):

PR #93654, opened 2023-10-18
Remove RequiresProcessIsolation from some loader tests

MichalStrehovsky:
I had a look at these tests and none of them do weird stuff.

I can continue chipping off these in the Loader area if it's helpful.

One question I have:

I tried to remove this on:

```
src/tests/Loader/classloader/StaticVirtualMethods/GenericContext/GenericContextTest/GenericContextTest.ilproj
```

And this failed to build with:

```
C:\git\runtime\src\tests\Loader\XUnitWrapperGenerator\XUnitWrapperGenerator.XUnitWrapperGenerator\FullRunner.g.cs(439,7
3): error CS0012: The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assem
bly 'System.Runtime, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'. [C:\git\runtime\src\tests\Loader\Loader.cs
proj] [C:\git\runtime\src\tests\build.proj]
```

Is there something special for tests that have a ProjectReference (I can't think of anything else that would be special here).

### DIFF
--- a/src/tests/Loader/classloader/Casting/Functionpointer.csproj
+++ b/src/tests/Loader/classloader/Casting/Functionpointer.csproj
@@ -1,7 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <!-- Needed for mechanical merging of all remaining tests, this particular project may not actually need process isolation -->
-    <RequiresProcessIsolation>true</RequiresProcessIsolation>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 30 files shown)


trylek (approved):
Looks great, thanks! My main motivation was that at this stage removing the remaining generated wrappers is likely to provide more lab perf improvement than auditing the individual 1500 or so remaining tests so I decided to prioritize that. In general further cleanup in this area is certainly desirable.

MichalStrehovsky:
> My main motivation was that at this stage removing the remaining generated wrappers is likely to provide more lab perf improvement than auditing the individual 1500 or so remaining tests so I decided to prioritize that.

Yup, that makes total sense. I'm motivated by nativeaot testing where we run Loader and Interop tests on PRs and there's a lot more extra value if we can reduce the test count.

PR #93655, opened 2023-10-18
[RISC-V] Disable FastTailCall in case of split arg

t-mustafin:
Disable TCO in case if *callee* split arg fixes System.Security.Cryptography.Tests SIGSEGV/Release and assert/Checked.
Also disable TCO in case if *caller* split arg as for ARM32: #66282.

Part of #84834
cc @clamp03 @gbalykov @tomeksowi @Sirntar @viewizard @ashaurtaev.

### DIFF
--- a/src/coreclr/jit/lclvars.cpp
+++ b/src/coreclr/jit/lclvars.cpp
@@ -259,9 +259,9 @@ void Compiler::lvaInitTypeRef()
     LclVarDsc*              varDsc    = varDscInfo.varDsc;
     CORINFO_ARG_LIST_HANDLE localsSig = info.compMethodInfo->locals.args;
 
-#ifdef TARGET_ARM
+#if defined(TARGET_ARM) || defined(TARGET_RISCV64)
     compHasSplitParam = varDscInfo.hasSplitParam;
-#endif
+#endif // TARGET_ARM || TARGET_RISCV64
 
     for (unsigned i = 0; i < info.compMethodInfo->locals.numArgs;
          i++, varNum++, varDsc++, localsSig = info.compCompHnd->getArgNext(localsSig))
@@ -1043,6 +1043,9 @@ void Compiler::lvaInitUserArgs(InitVarDscInfo* varDscInfo, unsigned skipArgs, un
                         varDscInfo->setAllRegArgUsed(argRegTypeInStruct1);
 #if FEATURE_FASTTAILCALL
                         varDscInfo->stackArgSize += TARGET_POINTER_SIZE;
+#endif
+#ifdef TARGET_RISCV64
+                        varDscInfo->hasSplitParam = true;
 #endif
                     }
                 }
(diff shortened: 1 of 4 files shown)


gbalykov (approved):

t-mustafin:
@shushanhf Loongarch64 has the same SplitArg behaviour as Riscv64. Does Loongarch64 also needs this disable or Loongarch64 has proper TCO handling on SplitArg case?

ashaurtaev (approved):

shushanhf:
> @shushanhf Loongarch64 has the same SplitArg behaviour as Riscv64. Does Loongarch64 also needs this disable or Loongarch64 has proper TCO handling on SplitArg case?

Thanks!
We test the `System.Security.Cryptography.Tests` on runtime8.0-rc2, but is OK for LA64.
Do you test it on the main9.0 or the runtime8.0-rc2 ?

t-mustafin:
> Do you test it on the main9.0 or the runtime8.0-rc2 ?

We test on main9.0. Thanks for information.
Problem occurs with Tier-1 compilation executed for TCO-compatible method in case of callee has splitArg:
```
System.Security.Cryptography.SP800108HmacCounterKdf:DeriveBytesCore(System.ReadOnlySpan`1[ubyte],System.Security.Cryptography.HashAlgorithmName,System.ReadOnlySpan`1[ubyte],System.ReadOnlySpan`1[ubyte],System.Span`1[ubyte]) (MethodHash=b612d031)
```

src/coreclr/jit/morph.cpp
```diff
@@ -5501,7 +5501,13 @@ bool Compiler::fgCanFastTailCall(GenTreeCall* callee, const char** failReason)
             reportFastTailCallDecision("Argument splitting in callee is not supported on ARM32");
             return false;
         }
-#endif // TARGET_ARM
+#elif defined(TARGET_RISCV64)
```
tomeksowi, line 5504: nit: could be folded:
```c++
#if defined(TARGET_ARM) || defined(TARGET_RISCV64)
        if (arg.AbiInfo.IsSplit())
        {
            reportFastTailCallDecision("Argument splitting in callee is not supported on " TARGET_READABLE_NAME);
            return false;
        }
```

tomeksowi (approved):

src/coreclr/jit/morph.cpp
t-mustafin: Thank you!

clamp03 (approved):

src/coreclr/jit/lclvars.cpp
```diff
@@ -261,7 +261,9 @@ void Compiler::lvaInitTypeRef()
 
 #ifdef TARGET_ARM
     compHasSplitParam = varDscInfo.hasSplitParam;
-#endif
+#elif defined(TARGET_RISCV64)
+    compHasSplitParam = varDsc->lvIsSplit;
```
jakobbotsch, line 265: This only checks if the last parameter is split. That doesn't seem correct. You should add a case in the main loop over parameters in `lvaInitUserArgs` instead, next to the ARM32 code that sets `InitVarDscInfo::hasSplitParam` (and make that field available on RISCV64 as well).

jakobbotsch: In fact, it checks whether the first non-parameter local is split, which seemingly should always be false..? `varDsc` might also not be a valid pointer here.

t-mustafin: Thanks!

jakobbotsch (approved):

jakobbotsch:
Can you run jit-format?

gbalykov:
@jkotas can you please add @t-mustafin to dotnet org?

Issue #93447, opened 2023-10-13
[NativeAOT] Enable running libraries and functional tests on CI for maccatalyst and ios/tvos simulators 


ivanpovazan:
We currently do not run libraries and functional tests on CI with NativeAOT targeting maccatalyst, iossimulators and tvossimulators.
In order to improve test coverage we should enable testing on these platforms as well.

kotlarmilos:
Depends on https://github.com/dotnet/runtime/issues/91871.

kotlarmilos:
Related to https://github.com/dotnet/runtime/issues/81075.

PR #93658, opened 2023-10-18
[NativeAOT] Support library mode with NativeAOT on iOS-like platforms

### LINKED ISSUE #88737: NativeAOT library mode on iOS-like platforms

ivanpovazan:
This PR enables support for building libraries for iOS-like platforms with NativeAOT.

## Changes

- Extended NativeAOT build integration targets to support building libraries for iOS-like platforms
- Adapted `AppleAppBuilder` to support bundling in library mode for both MonoAOT and NativeAOT
- Adapted `AppleAppBuilder` to support bundling in library mode when targeting iOS-like devices
- Adapted `src/tests/FunctionalTests/iOS/Simulator/LibraryMode` for testing the library mode on `iossimulators-arm64/x64` with NativeAOT - currently this only works **locally** please see comments in the project file for more info
    - https://github.com/dotnet/runtime/issues/93447)
- Added `src/tests/FunctionalTests/iOS/Device/LibraryMode` for testing the library mode on `ios-arm64` for both MonoAOT and NativeAOT

## Additional notes

To test **locally** one can execute the following from the repo root directory:
- Test `MonoAOT` library mode on `ios-arm64`:
    ```
    ./dotnet.sh build /t:Test -c Debug /p:TargetOS=ios /p:TargetArchitecture=arm64 src/tests/FunctionalTests/iOS/Device/LibraryMode/iOS.Device.LibraryMode.Test.csproj /p:RuntimeFlavor=Mono
    ```
    
- Test `NativeAOT` library mode on `ios-arm64`:
    ```
    ./dotnet.sh build /t:Test -c Debug /p:TargetOS=ios /p:TargetArchitecture=arm64 src/tests/FunctionalTests/iOS/Device/LibraryMode/iOS.Device.LibraryMode.Test.csproj /p:UseNativeAOTRuntime=true /p:RunAOTCompilation=false /p:RuntimeFlavor=CoreCLR
    ```
    
## Documentation

- [x] https://github.com/dotnet/runtime/issues/96742

--- 
Fixes https://github.com/dotnet/runtime/issues/88737

### DIFF
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
@@ -58,17 +58,6 @@ The .NET Foundation licenses this file to you under the MIT license.
     </PropertyGroup>
 
     <PropertyGroup Condition="'$(_IsiOSLikePlatform)' == 'true'">
-      <Xcrun Condition="'$(Xcrun)' == ''">xcrun</Xcrun>
-      <_WhereXcrun>0</_WhereXcrun>
-    </PropertyGroup>
-
-    <Exec Command="command -v &quot;$(Xcrun)&quot;" IgnoreExitCode="true" StandardOutputImportance="Low" Condition="'$(_IsiOSLikePlatform)' == 'true'">
-      <Output TaskParameter="ExitCode" PropertyName="_WhereXcrun" />
-    </Exec>
-    <Error Condition="'$(_WhereXcrun)' != '0' and '$(_IsiOSLikePlatform)' == 'true'"
-      Text="'$(Xcrun)' not found in PATH. Make sure '$(Xcrun)' is available in PATH." />
-
-    <PropertyGroup Condition="'$(_IsiOSLikePlatform)' == 'true'">
       <AppleMinOSVersion Condition="'$(AppleMinOSVersion)' == '' and '$(_targetOS)' == 'maccatalyst' and '$(_targetArchitecture)' == 'x64'">13.5</AppleMinOSVersion>
       <AppleMinOSVersion Condition="'$(AppleMinOSVersion)' == '' and '$(_targetOS)' == 'maccatalyst'">14.2</AppleMinOSVersion>
       <AppleMinOSVersion Condition="'$(AppleMinOSVersion)' == '' and ($(_targetOS.StartsWith('ios')) or $(_targetOS.StartsWith('tvos')))">11.0</AppleMinOSVersion>
@@ -92,6 +81,17 @@ The .NET Foundation licenses this file to you under the MIT license.
 
     <Error Condition="'$(_IsiOSLikePlatform)' == 'true' and ('$(_AppleSdkName)' == '' or '$(CrossCompileArch)' == '' or '$(_AppleTripleOS)' == ''  or '$(AppleMinOSVersion)' == '' or '$(_AppleTripleAbi)' == '')"
       Text="One of the required Apple SDK properties is empty and was not properly resolved: _AppleSdkName = '$(_AppleSdkName)' CrossCompileArch = '$(CrossCompileArch)' _AppleTripleOS = '$(_AppleTripleOS)' AppleMinOSVersion = '$(AppleMinOSVersion)' _AppleTripleAbi = '$(_AppleTripleAbi)'" />
+
+    <PropertyGroup Condition="'$(_IsiOSLikePlatform)' == 'true'">
+      <Xcrun Condition="'$(Xcrun)' == ''">xcrun</Xcrun>
+      <_WhereXcrun>0</_WhereXcrun>
+    </PropertyGroup>
+
+    <Exec Command="command -v &quot;$(Xcrun)&quot;" IgnoreExitCode="true" StandardOutputImportance="Low" Condition="'$(_IsiOSLikePlatform)' == 'true'">
+      <Output TaskParameter="ExitCode" PropertyName="_WhereXcrun" />
+    </Exec>
+    <Error Condition="'$(_WhereXcrun)' != '0' and '$(_IsiOSLikePlatform)' == 'true'"
+      Text="'$(Xcrun)' not found in PATH. Make sure '$(Xcrun)' is available in PATH." />
 
     <Exec Command="&quot;$(Xcrun)&quot; --sdk $(_AppleSdkName) --show-sdk-path" IgnoreExitCode="true" StandardOutputImportance="Low" Condition="'$(SysRoot)' == '' and '$(_IsiOSLikePlatform)' == 'true'" ConsoleToMsBuild="true">
       <Output TaskParameter="ConsoleOutput" PropertyName="SysRoot" />
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
@@ -96,7 +96,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <Error Condition="'$(_iOSLibraryMode)' == 'true' and ('$(_AppleSdkName)' == '' or '$(_AppleTripleArch)' == '' or '$(_AppleTripleOS)' == ''  or '$(AppleMinOSVersion)' == '' or '$(_AppleTripleAbi)' == '')"
       Text="One of the required Apple SDK properties is empty and was not properly resolved: _AppleSdkName = '$(_AppleSdkName)' _AppleTripleArch = '$(_AppleTripleArch)' _AppleTripleOS = '$(_AppleTripleOS)' AppleMinOSVersion = '$(AppleMinOSVersion)' _AppleTripleAbi = '$(_AppleTripleAbi)'" />
 
-    <Exec Command="&quot;$(Xcrun)&quot; --sdk $(_AppleSdkName) --show-sdk-path" IgnoreExitCode="true" StandardOutputImportance="Low" Condition="'$(_iOSLibraryMode)' == 'true'" ConsoleToMsBuild="true">
+    <Exec Command="&quot;$(Xcrun)&quot; --sdk $(_AppleSdkName) --show-sdk-path" IgnoreExitCode="true" StandardOutputImportance="Low" Condition="'$(SysRoot)' == '' and '$(_iOSLibraryMode)' == 'true'" ConsoleToMsBuild="true">
       <Output TaskParameter="ConsoleOutput" PropertyName="SysRoot" />
     </Exec>
     <Error Condition="!Exists('$(SysRoot)') and '$(_iOSLibraryMode)' == 'true'"
@@ -175,8 +175,8 @@ The .NET Foundation licenses this file to you under the MIT license.
       <LinkerArg Include="-gz=zlib" Condition="'$(CompressSymbols)' != 'false'" />
       <LinkerArg Include="-fuse-ld=$(LinkerFlavor)" Condition="'$(LinkerFlavor)' != ''" />
       <LinkerArg Include="@(NativeLibrary)" />
-      <LinkerArg Include="--sysroot=$(SysRoot)" Condition="'$(SysRoot)' != '' and '$(_iOSLibraryMode)' != 'true'" />
-      <LinkerArg Include="-isysroot $(SysRoot)" Condition="'$(SysRoot)' != '' and '$(_iOSLibraryMode)' == 'true'" />
+      <LinkerArg Include="--sysroot=&quot;$(SysRoot)&quot;" Condition="'$(SysRoot)' != '' and '$(_IsApplePlatform)' != 'true'" />
+      <LinkerArg Include="-isysroot &quot;$(SysRoot)&quot;" Condition="'$(SysRoot)' != '' and '$(_IsApplePlatform)' == 'true'" />
       <LinkerArg Include="--target=$(TargetTriple)" Condition="('$(_IsApplePlatform)' != 'true' or '$(_iOSLibraryMode)' == 'true') and '$(TargetTriple)' != ''" />
       <LinkerArg Include="-arch $(CrossCompileArch)" Condition="'$(_IsApplePlatform)' == 'true' and '$(CrossCompileArch)' != ''" />
       <LinkerArg Include="-g" Condition="$(NativeDebugSymbols) == 'true'" />
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
@@ -33,7 +33,7 @@ The .NET Foundation licenses this file to you under the MIT license.
       <FullRuntimeName Condition="'$(ServerGarbageCollection)' == 'true'">libRuntime.ServerGC</FullRuntimeName>
 
       <CrossCompileRid />
-      <CrossCompileRid Condition="$(_hostOS) != $(_originalTargetOS) or !$(RuntimeIdentifier.EndsWith('-$(_hostArchitecture)'))">$(RuntimeIdentifier)</CrossCompileRid>
+      <CrossCompileRid Condition="'$(_hostOS)' != '$(_originalTargetOS)' or '$(_hostArchitecture)' != '$(_targetArchitecture)'">$(RuntimeIdentifier)</CrossCompileRid>
 
       <CrossCompileArch />
       <CrossCompileArch Condition="$(CrossCompileRid.EndsWith('-x64'))">x86_64</CrossCompileArch>
@@ -79,9 +79,6 @@ The .NET Foundation licenses this file to you under the MIT license.
       <_AppleSdkName Condition="'$(_targetOS)' == 'tvossimulator'">appletvsimulator</_AppleSdkName>
       <_AppleSdkName Condition="'$(_targetOS)' == 'maccatalyst'">macosx</_AppleSdkName>
 
-      <_AppleTripleArch Condition="'$(_targetArchitecture)' == 'x64'">x86_64</_AppleTripleArch>
-      <_AppleTripleArch Condition="'$(_targetArchitecture)' == 'arm64'">arm64</_AppleTripleArch>
-
       <_AppleTripleOS Condition="'$(_targetOS)' == 'maccatalyst' or $(_targetOS.StartsWith('ios'))">ios</_AppleTripleOS>
       <_AppleTripleOS Condition="$(_targetOS.StartsWith('tvos'))">tvos</_AppleTripleOS>
 
@@ -89,12 +86,12 @@ The .NET Foundation licenses this file to you under the MIT license.
       <_AppleTripleAbi Condition="'$(_targetOS)' == 'maccatalyst'">macabi</_AppleTripleAbi>
       <_AppleTripleAbi Condition="$(_targetOS.EndsWith('simulator'))">simulator</_AppleTripleAbi>
 
-      <TargetTriple>$(_AppleTripleArch)-apple-$(_AppleTripleOS)$(AppleMinOSVersion)-$(_AppleTripleAbi)</TargetTriple>
+      <TargetTriple>$(CrossCompileArch)-apple-$(_AppleTripleOS)$(AppleMinOSVersion)-$(_AppleTripleAbi)</TargetTriple>
       <SharedLibraryInstallName Condition="'$(SharedLibraryInstallName)' == '' and '$(NativeLib)' == 'Shared'">@rpath/$(TargetName)$(NativeBinaryExt)</SharedLibraryInstallName>
     </PropertyGroup>
 
-    <Error Condition="'$(_IsiOSLikePlatform)' == 'true' and ('$(_AppleSdkName)' == '' or '$(_AppleTripleArch)' == '' or '$(_AppleTripleOS)' == ''  or '$(AppleMinOSVersion)' == '' or '$(_AppleTripleAbi)' == '')"
-      Text="One of the required Apple SDK properties is empty and was not properly resolved: _AppleSdkName = '$(_AppleSdkName)' _AppleTripleArch = '$(_AppleTripleArch)' _AppleTripleOS = '$(_AppleTripleOS)' AppleMinOSVersion = '$(AppleMinOSVersion)' _AppleTripleAbi = '$(_AppleTripleAbi)'" />
+    <Error Condition="'$(_IsiOSLikePlatform)' == 'true' and ('$(_AppleSdkName)' == '' or '$(CrossCompileArch)' == '' or '$(_AppleTripleOS)' == ''  or '$(AppleMinOSVersion)' == '' or '$(_AppleTripleAbi)' == '')"
+      Text="One of the required Apple SDK properties is empty and was not properly resolved: _AppleSdkName = '$(_AppleSdkName)' CrossCompileArch = '$(CrossCompileArch)' _AppleTripleOS = '$(_AppleTripleOS)' AppleMinOSVersion = '$(AppleMinOSVersion)' _AppleTripleAbi = '$(_AppleTripleAbi)'" />
 
     <Exec Command="&quot;$(Xcrun)&quot; --sdk $(_AppleSdkName) --show-sdk-path" IgnoreExitCode="true" StandardOutputImportance="Low" Condition="'$(SysRoot)' == '' and '$(_IsiOSLikePlatform)' == 'true'" ConsoleToMsBuild="true">
       <Output TaskParameter="ConsoleOutput" PropertyName="SysRoot" />
--- a/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
+++ b/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
@@ -57,6 +57,51 @@ The .NET Foundation licenses this file to you under the MIT license.
       <StandaloneGCSupportName Condition="'$(IlcStandaloneGCSupport)' == 'true'">libstandalonegc-enabled</StandaloneGCSupportName>
     </PropertyGroup>
 
+    <PropertyGroup Condition="'$(_iOSLibraryMode)' == 'true'">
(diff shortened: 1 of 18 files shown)


src/tasks/AppleAppBuilder/Xcode.cs
```diff
@@ -330,8 +340,30 @@ public void CreateXcodeProject(string projectName, string cmakeDirectoryPath)
 
         if (isLibraryMode)
         {
-            string dylibName = $"lib{projectName}.dylib";
-            cmakeLists = cmakeLists.Replace("%DYLIB_PATH%", $"{Path.Combine(binDir, dylibName)}");
+            string libraryPath;
+            // TODO: unify MonoAOT and NativeAOT library paths
+            // Current differences:
+            // - NativeAOT produces {ProjectName}.dylib, while MonoAOT produces lib{ProjectName}.dylib
+            // - NativeAOT places the library in the 'workspace' location ie 'publish' folder, while MonoAOT places it in 'binDir' ie 'AppBundle'
```
rolfbjarne, line 347: A .dylib isn't usable for the end developer, you should probably produce a .framework instead.

ivanpovazan:
Failures are related:
1. Since I introduced `VerifyOSSpecificTooling` target, it seems I broke some dependency with tooling in the build
2. Building on Helix for the library mode tests seem to be broken

src/tasks/AppleAppBuilder/Xcode.cs
ivanpovazan: Thank you, this meant to be an "initial" support for the library mode. We will definitely consider going with `.framework` approach.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
```diff
@@ -173,7 +209,9 @@ The .NET Foundation licenses this file to you under the MIT license.
       <LinkerArg Include="@(NativeFramework->'-framework %(Identity)')" Condition="'$(_IsApplePlatform)' == 'true'" />
       <LinkerArg Include="-Wl,--eh-frame-hdr" Condition="'$(_IsApplePlatform)' != 'true'" />
     </ItemGroup>
+  </Target>
 
+  <Target Name="VerifyOSSpecificTooling">
```
MichalStrehovsky, line 214: Why do we need to split this? More places than just tests.singlefile.targets refer to the existing target (you can search github). I would rather not break them if I can help it.

ivanpovazan: I made the split for two reasons:
1. To avoid moving the code that probes OS-specific tooling to the beginning of the `SetupOSSpecificProps`
    - This is required because we need `xcrun` to be verified before populating all the Apple-specific properties introduced in this PR
2. To separate the logic in `SetupOSSpecificProps` as it is: 
    - Setting up various properties and items, and 
    - Verifying the existence of OS-specific tools that some of these properties and items depend on

I could remove the newly added target `VerifyOSSpecificTooling` and reorganize `SetupOSSpecificProps` to make all this work together, but I think it is cleaner to have a separate target. What do you think?

MichalStrehovsky: > To avoid moving the code that probes OS-specific tooling to the beginning

I might be missing something, but can we just take lines 284-294 and paste them before line 60? The only "problematic" property would be `_CommandProbe` because it's not defined there, but we can just replace `$(_CommandProbe)` with `command -v`. We already don't use `$(_CommandProbe)` with Apple specific stuff because cross compilation from non-Apple to Apple platform is an academic scenario.

> I think it is cleaner to have a separate target

`VerifyOSSpecificTooling` would still be doing stuff like compiling System.Globalization.Native for StaticICULinking and `SetupOSSpecificProps` would still be checking for a valid SDK with xcrun. It doesn't feel like it moves the needle. Many 3rd party repos depend on this target and this is not a good reason to break them.

ivanpovazan: I see, thank you for the suggestion and clarification! 
I will remove the `VerifyOSSpecificTooling` and adjust the existing target accordingly.

global.json
```diff
@@ -9,7 +9,7 @@
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "9.0.0-beta.23559.3",
-    "Microsoft.DotNet.Helix.Sdk": "9.0.0-beta.23559.3",
+    "Microsoft.DotNet.Helix.Sdk": "9.0.0-beta.23563.1",
```
ivanpovazan, line 12: This will be reverted, I bumped the version only to test the required fix for deep signing introduced in: https://github.com/dotnet/arcade/pull/14164

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
```diff
@@ -130,13 +175,15 @@ The .NET Foundation licenses this file to you under the MIT license.
       <LinkerArg Include="-gz=zlib" Condition="'$(CompressSymbols)' != 'false'" />
       <LinkerArg Include="-fuse-ld=$(LinkerFlavor)" Condition="'$(LinkerFlavor)' != ''" />
       <LinkerArg Include="@(NativeLibrary)" />
-      <LinkerArg Include="--sysroot=$(SysRoot)" Condition="'$(SysRoot)' != ''" />
-      <LinkerArg Include="--target=$(TargetTriple)" Condition="'$(_IsApplePlatform)' != 'true' and '$(TargetTriple)' != ''" />
+      <LinkerArg Include="--sysroot=$(SysRoot)" Condition="'$(SysRoot)' != '' and '$(_iOSLibraryMode)' != 'true'" />
+      <LinkerArg Include="-isysroot $(SysRoot)" Condition="'$(SysRoot)' != '' and '$(_iOSLibraryMode)' == 'true'" />
```
ivanpovazan, line 179: @jkotas regarding our discussion in https://github.com/dotnet/runtime/pull/94999#discussion_r1420504285 is there anything you would like me to address in this PR?

jkotas:
I am wondering why we need so much extra stuff for building iOS-like libraries. Is it because all this extra stuff comes from Xamarin iOS when building executables?

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
jkotas: I do not think it makes sense for `SysRoot` to mean SDK root (isysroot) when building iOS-like libraries and a classic SysRoot for all other Apple targets.

I think overriding SDK root is an advanced scenario for all Apple targets. If we want `SysRoot` on Apple platforms to mean SDK root, this condition should be `_IsApplePlatform`.

(Alternatively, if we want to keep `SysRoot` as a classic SysRoot on Apple platforms, we should invent a property with a different name for the SDK root. Again, it should be applicable to all Apple platforms.)

ivanpovazan:
> I am wondering why we need so much extra stuff for building iOS-like libraries. Is it because all this extra stuff comes from Xamarin iOS when building executables?

Yes, when building executables, apple SDK, target triple and all apple-specific build parameters, are resolved by Xamarin SDK and tooling that creates an application bundle. 
We have our internal `AppleAppBuilder` https://github.com/dotnet/runtime/tree/main/src/tasks/AppleAppBuilder having a minimal set of such capabilities, which we use for CI and local testing.

jkotas:
Is `_iOSLibraryMode` the right name for this then? Based on your description, it is a mode without dependency on Xamarin SDK or Apple builder. We happen to use for building dynamic libraries in this PR, but I expect that it can be potentially used for building executables too.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
akoeplinger, line 179: btw. these paths should be quoted since they could in theory have spaces in them

ivanpovazan: Thank you, address in the latest commit.

ivanpovazan: Thank you, address in the latest commit.

ivanpovazan:
> Is `_iOSLibraryMode` the right name for this then? Based on your description, it is a mode without dependency on Xamarin SDK or Apple builder. We happen to use for building dynamic libraries in this PR, but I expect that it can be potentially used for building executables too.

Could you please explain how would building an iOS executable look like in such case and how we would bundle it?

steveisok:
> > Is `_iOSLibraryMode` the right name for this then? Based on your description, it is a mode without dependency on Xamarin SDK or Apple builder. We happen to use for building dynamic libraries in this PR, but I expect that it can be potentially used for building executables too.
> 
> Could you please explain how would building an iOS executable look like in such case and how we would bundle it?

My opinion, we could do lightweight apps in theory, but given what the xam ios repo has, I can't think of any value doing it here. Maybe to align architecturally what comes out of runtime vs sdk's, but that's highly subjective.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
```diff
@@ -57,6 +57,51 @@ The .NET Foundation licenses this file to you under the MIT license.
       <StandaloneGCSupportName Condition="'$(IlcStandaloneGCSupport)' == 'true'">libstandalonegc-enabled</StandaloneGCSupportName>
     </PropertyGroup>
 
+    <PropertyGroup Condition="'$(_iOSLibraryMode)' == 'true'">
```
steveisok, line 60: Should this be in a stand alone task?

ivanpovazan: > Should this be in a stand alone task? At least in some form?

Do you mean to group these in a separate MSBuild target? 
- We could, but then we would have to escape setting up the following as they would occur after the new target https://github.com/dotnet/runtime/blob/b0e7ee7c8d165e49b99113e23e08d2d1c17f88a2/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets#L43-L47

- Additionally, this new target (lets call it `SetupAppleSpecificProps`) should be executed before `SetupOSSpecificProps` and I already had some attempts to make that work properly see: discussion above https://github.com/dotnet/runtime/pull/93658#discussion_r1364994941 
Basically the problem is that we already have some external targets hooking in before `SetupOSSpecificProps`, so if it happens that these external targets are also trying to set properties before `SetupAppleSpecificProps` there is no guarantee that the external target will be scheduled by MSBuild before this newly introduced one.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
```diff
@@ -30,6 +30,7 @@ The .NET Foundation licenses this file to you under the MIT license.
     <RunILLink>false</RunILLink>
     <_IsiOSLikePlatform Condition="'$(_targetOS)' == 'maccatalyst' or $(_targetOS.StartsWith('ios')) or $(_targetOS.StartsWith('tvos'))">true</_IsiOSLikePlatform>
     <_IsApplePlatform Condition="'$(_targetOS)' == 'osx' or '$(_IsiOSLikePlatform)' == 'true'">true</_IsApplePlatform>
+    <_iOSLibraryMode Condition="'$(_IsiOSLikePlatform)' == 'true' and !$(TargetFramework.Contains('$(_targetOS)')) and '$(NativeLib)' != '' and '$(CustomNativeMain)' != 'true'">true</_iOSLibraryMode>
```
am11, line 33: Library Mode term sounds too special in this context. Can this condition be simplified to `'$(_IsiOSLikePlatform)' == 'true' && '$(NativeLib)' == 'Shared'` and inline it at call-site?  `static` and `static-pie` libs are a bit different.

steveisok: Is there a better umbrella name to refer to when we want to consume a .net library from native?

am11: With `<OutputType>Exe` and `<PublishAot>true` in csproj, dotnet-publish generates a native executable. With `<OutputType>Shared` and `<PublishAot>true` we get  a shared library. We can additionally set `<NativeLib>Static` to get an static lib.

For platform specialization, we have targetXX and isXX type properties in place. Unless "Library Mod"e in `nativeaot/BuildIntegration` context is bringing something very unique (which, going by the first bullet in top comment, doesn't seem to be the case), it's better to use the existing / nearby convention.

am11: With `<OutputType>Exe` and `<PublishAot>true` in csproj, dotnet-publish generates a native executable. With `<OutputType>Library` and `<PublishAot>true` we get  a shared library. We can additionally set `<NativeLib>Static` to get a static lib.

For platform specialization, we have targetXX and isXX type properties in place. Unless "Library Mod"e in `nativeaot/BuildIntegration` context is bringing something very unique (which, going by the first bullet in top comment, doesn't seem to be the case), it's better to use the existing / nearby convention.

MichalStrehovsky: I agree with what @jkotas and @am11 are saying here. `_iOSLibraryMode` is introducing a new concept and the question is if a new concept is avoidable.

The way we typically do checks here is that we spell out if something is platform specific or library mode specific. Like so:

https://github.com/dotnet/runtime/blob/6f4f268d0b46fa5dbcee46b7220919defb148127/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets#L384

https://github.com/dotnet/runtime/blob/6f4f268d0b46fa5dbcee46b7220919defb148127/src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets#L69

(Plenty of other examples. We don't introduce a new property because the condition is still concise and readable.)

The conditions when `_iOSLibraryMode` is true are really difficult to parse for me. I wouldn't actually be able to quickly tell when this is true. For example, why do we need to condition on `!$(TargetFramework.Contains('$(_targetOS)'))` and `CustomNativeMain` and why those decide whether we pass `--target` to the linker (try to mentally inline the definition of `_iOSLibraryMode` into places that use the property and whether such definition really makes sense there).

In general - in the places that _use_ the new `_iOSLibraryMode` property, the question is why we're not keying off of `_IsiOSLikePlatform` and `NativeLib` (if necessary - I don't think it's necessary to key off the latter in many of the places that use `_iOSLibraryMode` right now - I think that goes back to Jan's hypothetical "what if we're building an exe" scenario - we don't need to have that scenario but it actually makes the conditions simpler to reason about because it makes it clear what part are due to platform differences and what parts are due to output type differences).

jkotas:
> Yes, when building executables, apple SDK, target triple and all apple-specific build parameters, are resolved by Xamarin SDK and tooling that creates an application bundle.
> We have our internal AppleAppBuilder https://github.com/dotnet/runtime/tree/main/src/tasks/AppleAppBuilder having a minimal set of such capabilities, which we use for CI and local testing.

Right. This PR is copying some logic that exists in the Xamarin SDK or AppleAppBuilder to native AOT build targets.

> Could you please explain how would building an iOS executable look like in such case and how we would bundle it?

We would copy more logic that exists in the Xamarin SDK or AppleAppBuilder to native AOT build targets. I can imagine that this can be useful for app types (games?) that do not want to use Objective C interop provided by Xamarin SDK. I am not suggesting that this PR implements. I am just suggesting that the changes in this PR work in this direction instead of introducing a new very specialized concept.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
ivanpovazan: Thank you all for you feedback and I understand the concerns.

The specific conditions (like `!$(TargetFramework.Contains('$(_targetOS)'))`) were introduced to avoid looking up the SDK and resolving relevant properties twice (when the build actually pulls in Xamarin SDK). 

I will try to simplify this.

ivanpovazan:
> We would copy more logic that exists in the Xamarin SDK or AppleAppBuilder to native AOT build targets. I can imagine that this can be useful for app types (games?) that do not want to use Objective C interop provided by Xamarin SDK. I am not suggesting that this PR implements it. I am just suggesting that the changes in this PR work in this direction instead of introducing a new very specialized concept.

Thank you for the explanation. 

Would it make sense to invest in refactoring `AppleAppBuilder` and productize it instead (not as part of this PR)? Duplicating the logic especially inlined in MSBuild targets is error prone and would be cleaner to implement in a custom build task (which AppleAppBuilder already is).

/cc: @steveisok

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
MichalStrehovsky: How expensive is running the commands? At the end of the day we're going to run an aot compiler so if it's not costly in relative terms to aot compilation I would not try to avoid it too hard.

I now understand that this is trying to approximate the exact conditions under which xamarin sdk or appbuilder would take over but that's also what makes it feel a bit like mystery code.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
```diff
@@ -57,6 +57,51 @@ The .NET Foundation licenses this file to you under the MIT license.
...
+      <AppleMinOSVersion Condition="'$(AppleMinOSVersion)' == '' and ($(_targetOS.StartsWith('ios')) or $(_targetOS.StartsWith('tvos')))">11.0</AppleMinOSVersion>
+
+      <_AppleSdkName Condition="'$(_targetOS)' == 'ios'">iphoneos</_AppleSdkName>
+      <_AppleSdkName Condition="'$(_targetOS)' == 'iossimulator'">iphonesimulator</_AppleSdkName>
+      <_AppleSdkName Condition="'$(_targetOS)' == 'tvos'">appletvos</_AppleSdkName>
+      <_AppleSdkName Condition="'$(_targetOS)' == 'tvossimulator'">appletvsimulator</_AppleSdkName>
+      <_AppleSdkName Condition="'$(_targetOS)' == 'maccatalyst'">macosx</_AppleSdkName>
+
+      <_AppleTripleArch Condition="'$(_targetArchitecture)' == 'x64'">x86_64</_AppleTripleArch>
+      <_AppleTripleArch Condition="'$(_targetArchitecture)' == 'arm64'">arm64</_AppleTripleArch>
```
am11, line 83: Are these different than `CrossCompileArch` resolved in the previous property group? If not, lets use the existing properties.

slightly related: line 36 can be simplified as:

```diff
-       <CrossCompileRid Condition="$(_hostOS) != $(_originalTargetOS) or !$(RuntimeIdentifier.EndsWith('-$(_hostArchitecture)'))">$(RuntimeIdentifier)</CrossCompileRid>
+       <CrossCompileRid Condition="'$(_hostOS)' != '$(_originalTargetOS)' or '$(_hostArchitecture)' != '$(_targetArchitecture)'">$(RuntimeIdentifier)</CrossCompileRid>
```

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.targets
ivanpovazan: I have simplified the logic as suggested. Please give it another look.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
ivanpovazan: It was an oversight, you are right, included in my last commit. Thank you!

src/tests/FunctionalTests/iOS/Device/LibraryMode/iOS.Device.LibraryMode.Test.csproj
```diff
@@ -0,0 +1,55 @@
...
+  <!-- TODO: This should be moved to a common build integration targets to support local builds when https://github.com/dotnet/runtime/issues/93447 is addressed -->
+  <Target Name="_PrepareSetupForLocalBuild" BeforeTargets="SetupProperties" Condition="'$(UseNativeAOTRuntime)' == 'true' and $(BuildTestsOn) == 'local'">
+    <PropertyGroup>
+      <IlcToolsPath>$(CoreCLRCrossILCompilerDir)</IlcToolsPath>
+      <IlcSdkPath>$(CoreCLRAotSdkDir)</IlcSdkPath>
+      <IlcFrameworkPath>$(LibrariesAllBinArtifactsPath)</IlcFrameworkPath>
+      <IlcFrameworkNativePath>$(LibrariesAllBinArtifactsPath)</IlcFrameworkNativePath>
+    </PropertyGroup>
+    <ItemGroup>
+      <DirectPInvoke Include="__Internal" KeepDuplicates="false" />
```
am11, line 51: nit: since .NET 7, `System.Runtime.InteropServices.NativeLibrary.GetMainProgramHandle()` returns the self pointer in mono and coreclr alike (which we can use in native library resolver, followed by `DllImport("AnyNameOfChoiceUsedInResolver")`).

am11 (approved):
Two optional nits, otherwise looks great! 👍

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
```diff
@@ -57,6 +57,48 @@ The .NET Foundation licenses this file to you under the MIT license.
...
+    <PropertyGroup Condition="'$(_IsiOSLikePlatform)' == 'true'">
+      <Xcrun Condition="'$(Xcrun)' == ''">xcrun</Xcrun>
+      <_WhereXcrun>0</_WhereXcrun>
+    </PropertyGroup>
+
+    <Exec Command="command -v &quot;$(Xcrun)&quot;" IgnoreExitCode="true" StandardOutputImportance="Low" Condition="'$(_IsiOSLikePlatform)' == 'true'">
+      <Output TaskParameter="ExitCode" PropertyName="_WhereXcrun" />
+    </Exec>
+    <Error Condition="'$(_WhereXcrun)' != '0' and '$(_IsiOSLikePlatform)' == 'true'"
+      Text="'$(Xcrun)' not found in PATH. Make sure '$(Xcrun)' is available in PATH." />
```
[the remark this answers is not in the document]
ivanpovazan: Thank you, I will look into this asap.

src/tests/FunctionalTests/iOS/Device/LibraryMode/iOS.Device.LibraryMode.Test.csproj
MichalPetryka: > nit: since .NET 7, `System.Runtime.InteropServices.NativeLibrary.GetMainProgramHandle()` returns the self pointer in mono and coreclr alike (which we can use in native library resolver, followed by `DllImport("AnyNameOfChoiceUsedInResolver")`).

Won't that behave like a normal PInvoke, not a direct one?

am11: Ah, if direct P/Invoke is optimized differently (without additional frames etc.) then yes. I don't know what DirectPInvoke entails. :)

(just [remembered](https://github.com/dotnet/runtime/commit/11ca24d39b5b0aa5c31ebe037cab8915693511da) that mono returns the same cached value with both `GetMainProgramHandle()` and `_Internal`)

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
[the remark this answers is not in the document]
ivanpovazan: I moved the validation block for better grouping, thank you.

src/tests/FunctionalTests/iOS/Device/LibraryMode/iOS.Device.LibraryMode.Test.csproj
ivanpovazan: Removed. This is not needed as the tests are not relying on direct P/Invokes. Thank you

src/tasks/AppleAppBuilder/Xcode.cs
ivanpovazan: I have opened https://github.com/dotnet/runtime/issues/96664 to track this.

ivanpovazan:
The CI failures are unrelated.

@steveisok @rolfbjarne @jkotas could you please take another look as we would like to get this in for `net9.0-p1`?

rolfbjarne (approved):
LGTM

akoeplinger (approved):

jkotas:
> @jkotas could you please take another look as we would like to get this in for net9.0-p1?

The changes in the shared native AOT files LGTM.

MichalStrehovsky (commented):
nativeaot/BuildIntegration looks good to me too!

PR #93661, opened 2023-10-18
Fix: OID's cache poisoning with random values

sstronin:
There is a code in NET and in OpenSsl that caches OIDs/names mapping. Both caches tend to grow uncontrollable when dealing with custom OIDs, but NET cache consumes x10 memory compared to OpenSsl cache. This fix disables NameToOid caching of unknown OIDs on NET level, improving memory consumption and eluminating speed degradation when an application process certificates with too many unknown OIDs.


src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/OidLookup.cs
```diff
@@ -77,7 +77,7 @@ internal static partial class OidLookup
 
             mappedOid = NativeFriendlyNameToOid(friendlyName, oidGroup, fallBackToAllGroups);
 
-            if (shouldUseCache)
+            if (shouldUseCache && mappedOid!=null)
```
danmoseley, line 80:
```suggestion
            if (shouldUseCache && mappedOid != null)
```

jozkee (approved):

PR #93662, opened 2023-10-18
ILLink: Allow interface methods to not have newslot

### LINKED ISSUE #93008: ILLink fails to resolve method's base interface method if the interface method doesn't have newslot
This is based of a repro from F#. The interface definition of a method in F# looks like this:
```
.method public hidebysig specialname abstract virtual 
	instance valuetype MyLib.OS get_OS () cil managed 
{
} // end of method IPlatformData::get_OS
```

Note that it doesn't have a `newslot` flag.

Similar interface definition in C# looks basically the same except that it has `newslot` as well.

When illink resolves methods and their base methods, it will not consider interface methods which don't have `newslot` on them (I don't know why).
https://github.com/dotnet/runtime/blob/64862501b238512db10eff9ed4b05ba94412974d/src/tools/illink/src/linker/Linker/TypeMapInfo.cs#L154

As a result, implementing the above interface in C# and trimming the result will create the implementation type without any of the interface method implementations.

vitek-karas:
Fixes https://github.com/dotnet/runtime/issues/93008

This comes from F#, where interface method defined in F# doesn't have `newslot` flag in its metadata. Trimmer gets confused and doesn't recognize such method as a "base" method for its implementation in the type which implements the interface. As a result, the implementation method is removed from the program because there doesn't seem to be a use for it anywhere.

The fix is to basically ignore the `newslot` flag for interface instance methods. The existing behavior must be kept for static interface methods though - as a redefinition of a method with the `new` C# keyword will produce an interface method definition without the `newslot` flag as well.

The change adds a new test and modifies the AOT/analyzer infra to run the test as well.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Trimming.Tests/TestCases/TestDatabase.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Trimming.Tests/TestCases/TestDatabase.cs
@@ -34,6 +34,11 @@ namespace Mono.Linker.Tests.TestCases
 			return TestNamesBySuiteName();
 		}
 
+        public static IEnumerable<object[]> Inheritance_Interaces()
+        {
+            return TestNamesBySuiteName("Inheritance.Interfaces");
+        }
+
         public static IEnumerable<object[]> Libraries()
         {
             return TestNamesBySuiteName();
--- a/src/coreclr/tools/aot/ILCompiler.Trimming.Tests/TestCases/TestSuites.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Trimming.Tests/TestCases/TestSuites.cs
@@ -38,6 +38,20 @@ namespace Mono.Linker.Tests.TestCases
 		}
 
         [Theory]
+        [MemberData(nameof(TestDatabase.Inheritance_Interaces), MemberType = typeof(TestDatabase))]
+        public void Inheritance_Interfaces(string t)
+        {
+            switch (t) {
+            case ".InterfaceWithoutNewSlot":
+                Run (t);
+                break;
+            default:
+                // Skip the rest for now
+                break;
+            }
+        }
+
+        [Theory]
         [MemberData(nameof(TestDatabase.Libraries), MemberType = typeof(TestDatabase))]
         public void Libraries(string t)
         {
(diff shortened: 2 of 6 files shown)


ivanpovazan (approved):
Thank you for looking into this!

MichalStrehovsky:
> The existing behavior must be kept for static interface methods though - as a redefinition of a method with the new C# keyword will produce an interface method definition without the newslot flag as well.

Is there a test that exercises this? I wasn't able to get C# to produce a static virtual method with a `newslot` flag set. Was this code doing anything for static virtuals?

vitek-karas:
I made this modification due to a failing test. Specifically this:
https://github.com/dotnet/runtime/blob/04ede0dadfbf2f1082d3bd218cea6fe46bbbed4a/src/tools/illink/test/Mono.Linker.Tests.Cases/Inheritance.Interfaces/StaticInterfaceMethods/StaticAbstractInterfaceMethods.cs#L391

We should remove the `IInheritsFromBase.UsedOnBaseOnlyConstrainedTypeImplicitImpl` because the call is made through the `IBase1` and the implementation has method-impl for both the method on `IBase1` as well as the "new" method on `IInheritsFromBase`. The "new" method doesn't get a `newslot` in IL, and so it was hitting the condition before the change.

vitek-karas:
I must admit that I don't know if IL allows for implicit implementation of a static virtual/abstract - that is done via method signature match and not relying on method-impl record. If it only allows this via method-impl records, then I agree we should just skip all static methods in this place (since the code after the if handles the method signature matching).

MichalStrehovsky:
Static virtual are only implemented through MethodImpl: https://github.com/dotnet/runtime/blob/main/docs/design/specs/Ecma-335-Augments.md#ii122-implementing-virtual-methods-on-interfaces

sbomer (approved):
Nice! LGTM, thank you for looking into this.

ivanpovazan:
@vitek-karas should we backport this to a servicing release?

charlesroddie:
> @vitek-karas should we backport this to a servicing release?

That would be very useful - effectively enabling F# support in dotnet8 ios nativeaot.

vitek-karas:
We discussed this couple of weeks ago, but then it was a weird time (right around the branches closing for RTM).
@agocke would you know which branch to backport this to for 8 servicing release?

ivanpovazan:
@vitek-karas it should be: `/backport to release/8.0-staging`

PR #93678, opened 2023-10-18
Replaced CMake's Repeated CONFIG Expressions With Their Condensed Counterparts

### LINKED ISSUE #93370: Use Improvements in CMake Generator Expressions to Simplify our Build System's ones
Now that we officially _CMake 3.20_ in our build system, we can simplify some of our generator expressions using some new features. For example, we can now simplify `$<OR:$<CONFIG:X>,$<CONFIG:Y>>:-flag>` to just `$<$<CONFIG:X,Y>:-flag>`.

Find the expressions like that in the build system and simplify them.

ivdiazsa:
Since 3.19, CMake now supports specifying multiple configurations with just one call to `$<CONFIG>`. Now, it supports multiple in one call using the following form:

Before 3.19:

`$<$<OR:$<CONFIG:X>,$<CONFIG:Y>>:-flag>`

From 3.19 and onwards:

`$<$<CONFIG:X,Y>:-flag>`

We now have a minimum required CMake version of 3.20, so this PR simplifies some code to use these new condensed generator expressions.


jkoritzinsky (approved):
LGTM!

ivdiazsa:
The failure is an unstable test and is unrelated to this PR. Merging now.

PR #93694, opened 2023-10-18
JIT: remove incorrect type deduction for an Unsafe.As case

### LINKED ISSUE #93650: Unsafe.As breaks callvirt after inlining in .NET 8 RC2
### Description

If a call to `Unsafe.As<A, B>()` and another call that uses the result (e.g., `ToString()`) are inlined by JIT, it results in bad codegen. If inlined manually, codegen is correct.

### Reproduction Steps

Bad codegen sample: https://godbolt.org/z/YEoYM7xY5

Minimal repro:
```cs
using System.Text;

internal static class Program
{
    static void Main()
    {
        var sb = new Holder();
        while (true)
        {
            var s = Bind(ref sb);
            if (s.Length != 0)
            {
                Console.WriteLine("StringBuilder.ToString() returned: " + s);
                return;
            }
        }
    }

    public static string Bind(ref Holder parameters) => GetString(parameters.GetBuilder());

    public static string GetString(StringBuilder sb) => sb.ToString();
}

internal struct Holder
{
    internal StringBuilder.AppendInterpolatedStringHandler _h;
    public Holder() => _h = new(0, 0, new());

    internal StringBuilder GetBuilder() => Unsafe.As<StringBuilder.AppendInterpolatedStringHandler, StringBuilder>(ref _h);
}
```

### Expected behavior

`StringBuilder:ToString()` is called (asm listing after manually inlining GetString):

```asm
; Assembly listing for method Program:Bind(byref):System.String (Tier1)
; Emitting BLENDED_CODE for X64 with AVX - Windows
; Tier1 code
; optimized code
; rsp based frame
; partially interruptible
; 0 inlinees with PGO data; 1 single block inlinees; 0 inlinees without PGO data

       sub      rsp, 40

G_M000_IG02:                ;; offset=0x0004
       mov      rcx, gword ptr [rcx]
       cmp      dword ptr [rcx], ecx
       call     [System.Text.StringBuilder:ToString():System.String:this]
       nop

G_M000_IG03:                ;; offset=0x0010
       add      rsp, 40
       ret
```

### Actual behavior

`Object:ToString()` is inlined instead:

```asm
; Assembly listing for method Program:Bind(byref):System.String (Tier1)
; Emitting BLENDED_CODE for X64 with AVX - Windows
; Tier1 code
; optimized code
; rsp based frame
; partially interruptible
; No PGO data
; 1 inlinees with PGO data; 3 single block inlinees; 0 inlinees without PGO data

       sub      rsp, 40

G_M000_IG02:                ;; offset=0x0004
       mov      rcx, gword ptr [rcx]
       cmp      byte  ptr [rcx], cl
       call     System.Object:GetType():System.Type:this
       mov      rcx, rax
       mov      edx, 1
       call     [System.RuntimeType:GetCachedName(int):System.String:this]
       nop

G_M000_IG03:                ;; offset=0x001D
       add      rsp, 40
       ret
```

### Regression?

This worked correctly in .NET 7

### Known Workarounds

_No response_

### Configuration

.NET 8.0.100-rc.2.23502.2, Windows x64

### Other information

_No response_

AndyAyersMS:
Thanks to @SingleAccretion for the fix.

The JIT was assuming that an indirectly accessed value type field had the type of the field, but that might not be the case if the field was accessed via `Unsafe.As`.

Fix this by limiting type deduction from these indirectly accessed fields to only ref type fields.

Closes #93650.

### DIFF
--- a/src/coreclr/jit/gentree.cpp
+++ b/src/coreclr/jit/gentree.cpp
@@ -18752,10 +18752,12 @@ CORINFO_CLASS_HANDLE Compiler::gtGetFieldClassHandle(CORINFO_FIELD_HANDLE fieldH
             {
                 JITDUMP("Field's current class not available\n");
             }
+
+            return fieldClass;
         }
     }
 
-    return fieldClass;
+    return NO_CLASS_HANDLE;
 }
 
 //------------------------------------------------------------------------
--- a/src/tests/JIT/Regression/JitBlue/Runtime_93650/Runtime_93650.cs
+++ b/src/tests/JIT/Regression/JitBlue/Runtime_93650/Runtime_93650.cs
@@ -1,0 +1,42 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Text;
+using Xunit;
+
+public struct Holder
+{
+    internal StringBuilder.AppendInterpolatedStringHandler _h;
+    public Holder() => _h = new(0, 0, new());
+
+    internal StringBuilder GetBuilder() => Unsafe.As<StringBuilder.AppendInterpolatedStringHandler, StringBuilder>(ref _h);
+}
+
+public static class Runtime_93650
+{
+    static int N = 1;
+
+    [Fact]
+    public static int Problem()
+    {
+        var sb = new Holder();
+        for (int i = 0; i < N; i++)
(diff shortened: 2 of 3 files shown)


AndyAyersMS:
@jakobbotsch PTAL
cc @dotnet/jit-contrib

jakobbotsch (approved):

AndyAyersMS:
No diffs or TP impact.

PR #93695, opened 2023-10-18
Use the improved vectorization algorithm for aggregate TensorPrimitives operations

tannergooding:
This is a continuation of https://github.com/dotnet/runtime/pull/93409


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -1172,7 +1198,7 @@ private static void ValidateInputOutputSpanNonOverlapping(ReadOnlySpan<float> in
             0x00000000, 0x00000000, 0x00000000, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF,
             0x00000000, 0x00000000, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF,
             0x00000000, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF,
-            0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF, 0xFFFFFFFF,
+            0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000, 0x00000000,
```
stephentoub, line 1201: Why did these change?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -797,248 +796,1699 @@ private static float CosineSimilarityCore(ReadOnlySpan<float> x, ReadOnlySpan<fl
...
+                    // length check, single jump, and then linear execution.
+
+                    result = Vectorized128Small(ref xRef, remainder);
+                }
+
+                return result;
+            }
+
+            // This is the software fallback when no acceleration is available
+            // It requires no branches to hit
```
stephentoub, line 871: Nit: missing periods
```suggestion
            // This is the software fallback when no acceleration is available.
            // It requires no branches to hit.
```

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -797,248 +796,1699 @@ private static float CosineSimilarityCore(ReadOnlySpan<float> x, ReadOnlySpan<fl
...
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            static float Vectorized256Small(ref float xRef, nuint remainder)
+            {
+                float result = TAggregationOperator.IdentityValue;
+
+                switch (remainder)
+                {
+                    case 7:
+                    case 6:
+                    case 5:
```
stephentoub, line 1269: Same question as I had on the other PR; I guess we decided it's not beneficial to change this to be `default:`?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -7760,10 +9220,18 @@ public static Vector512<float> Invoke(Vector512<float> x, Vector512<float> y)
...
 #endif
 
-            public static float Invoke(Vector128<float> x) => HorizontalAggregate<MultiplyOperator>(x);
-            public static float Invoke(Vector256<float> x) => HorizontalAggregate<MultiplyOperator>(x);
+            public static float Invoke(Vector128<float> x)
+            {
+                // We need to do log2(count) operations to compute the total sum
+
+                x *= Vector128.Shuffle(x, Vector128.Create(2, 3, 0, 1));
+                x *= Vector128.Shuffle(x, Vector128.Create(1, 0, 3, 2));
```
stephentoub, line 9228: This ends up doing two shuffles and two multiplications instead of three multiplications. I assume that tradeoff is a good one, especially if with the individual element loads to do the three multiplications?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -7760,10 +9220,18 @@ public static Vector512<float> Invoke(Vector512<float> x, Vector512<float> y)
             public static Vector512<float> Invoke(Vector512<float> x, Vector512<float> y) => x * y;
 #endif
 
-            public static float Invoke(Vector128<float> x) => HorizontalAggregate<MultiplyOperator>(x);
-            public static float Invoke(Vector256<float> x) => HorizontalAggregate<MultiplyOperator>(x);
```
stephentoub, line 7764: We can't just update HorizontalAggregate to do this in one place rather than adding all these shuffles? e.g.
```C#
private static float HorizontalAggregate<TAggregate>(Vector128<float> x) where TAggregate : struct, IBinaryOperator
{
    // We need to do log2(count) operations to compute the total sum
    x = TAggregate.Invoke(x, Vector128.Shuffle(x, Vector128.Create(2, 3, 0, 1)));
    x = TAggregate.Invoke(x, Vector128.Shuffle(x, Vector128.Create(1, 0, 3, 2));
    return x.ToScalar();
}
```

tannergooding: Yeah, we could probably share it like this. Will update

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: Right. And noting it's not just 3 multiplications but rather 3 extracts + 3 multiplications.

x64 doesn't have a "proper" floating-point extract operation. It has `extractps` but that moves the value into an integer register, so it doesn't get used. Instead, `x.GetElement(n)` works as a `nop` (0), `movshdup` (1), `unpckhps` (2), or `shufps` (3). All of these are 1 cycle operations, with 1 reciprocal throughput, and which execute on port 5. `mulps` is then a 4 cycle operation with a 0.5 reciprocal throughput, and which can execute on port 0 or 1.

So we're effectively trading off `(3 * 1) shuffle cycles + (3 * 4) multiply cycles = 15 cycles` for `(2 * 1) shuffle cycles + (2 * 4) multiply cycles = 10 cycles`

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: Right. Longer term, once the switch can take advantage of range propagation, we'll want the explicit entries so they get direct jump table entries rather than any comparison + branch as is required for a default handler.

-- It's possible that we could adjust the JIT to factor in `range prop` + `default` and figure out the whole range of entries `default` can apply to, but that's a bit more complex and likely not worth it.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
tannergooding: The table has 17, not 16, entries today.

No trailing elements means a mask of `0` is used, correspondingly `16` trailing elements should then never have bene hit (it should be handled as part of the block processing) and so this entry should've been technically dead before the PR.

For leading elements (those processed to become aligned), however, we do hit this last entry because the alignment expression is constructed in a way to account for `beg` always being used.

The last entry here for trailing elements is then similarly a repeat of the first entry to allow the scenario where `end` is always processed, but we want it to `nop` because the normal case handling will have already gotten it. Thus allowing us to avoid an extra branch.

stephentoub: Can you add a comment about it? Thanks.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
```diff
@@ -1149,6 +1149,32 @@ private static void ValidateInputOutputSpanNonOverlapping(ReadOnlySpan<float> in
             }
         }
 
+        /// <summary>Mask used to handle alignment elements before vectorized handling of the input.</summary>
+        /// <remarks>
+        /// Logically 16 rows of 16 uints. The Nth row should be used to handle N alignment elements at the
+        /// begging of the input, where elements in the vector after that will be zero'd.
```
gfoidl, line 1155:
```suggestion
        /// begin of the input, where elements in the vector after that will be zero'd.
```
?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: Fixed.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.cs
tannergooding: Should be beginning. Fixed.

stephentoub (approved):

Issue #93669, opened 2023-10-18
[wasm] Runtime tests failing - `_Vector2_3_4::Vector2_3_4Test.RunVector*Tests` because of missing support for native libraries


radical:
## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=442221
Build error leg or test failing: _Vector2_3_4::Vector2_3_4Test.RunVector2Tests()
Pull request: https://github.com/dotnet/runtime/pull/93646

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json
{
  "ErrorMessage": "",
  "ErrorPattern": "Vector2_3_4TestNativeBadPattern",
  "BuildRetry": false,
  "ExcludeConsoleLog": true
}
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=442221
**Error message validated:** `Vector2_3_4TestNativeBadPattern`
**Result validation: :x:** Known issue did not match with the provided build.
**Validation performed at:** 10/23/2023 11:02:03 AM UTC

### Report
|Build|Definition|Step Name|Console log|Pull Request|
|---|---|---|---|---|
|[447310](https://dev.azure.com/dnceng-public/public/_build/results?buildId=447310)|dotnet/runtime|coreclr  Pri1 Runtime Tests Run R2R linux x64 release|[Log](https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_apis/build/builds/447310/logs/1018)||
|[447311](https://dev.azure.com/dnceng-public/public/_build/results?buildId=447311)|dotnet/runtime|coreclr  Pri0 Runtime Tests Run Composite osx x64 checked|[Log](https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_apis/build/builds/447311/logs/508)||
[...]

PR #93712, opened 2023-10-19
[wasm] Disable failing runtime tests

radical:
`Interop/PInvoke/Vector2_3_4/` tests are failing as they depend on native test libraries. But
wasm build for runtime tests does not support that yet.

Issue: https://github.com/dotnet/runtime/issues/93669

### DIFF
--- a/src/tests/Interop/PInvoke/Vector2_3_4/Vector2_3_4.cs
+++ b/src/tests/Interop/PInvoke/Vector2_3_4/Vector2_3_4.cs
@@ -12,6 +12,7 @@ public class Vector2_3_4Test
     private const int NewIntValue = 18;
 
     [Fact]
+    [ActiveIssue("https://github.com/dotnet/runtime/issues/93669", TestPlatforms.Browser)]
     public static void RunVector2Tests()
     {
         Console.WriteLine($"Running {nameof(RunVector2Tests)}... ");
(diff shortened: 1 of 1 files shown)


radical:
@jkoritzinsky @trylek IIUC, excluding tests uses the paths specified in `issues.targets`. For a test like `Interop/PInvoke/Vector2_3_4`, a `BasicTestMethod` is used which uses the full method name to the filter's `ShouldRunTest`, and that won't match the path from `issues.targets`. If that is correct, then what is the recommended way of excluding such tests?
I see there is support for a dotnet-test like filter expression, but is that for manual runs?

There are existing exclusions for this like https://github.com/dotnet/runtime/blob/a40db0207101780e6b8d5ad261428d8d954f6df2/src/tests/issues.targets#L2630-L2632

.. but these do not work any more, AFAIU!

radical:
cc @lewing

jkoritzinsky (changes requested):
Please use the `PlatformSpecific` attribute or one of the other xunit/XunitExtensions attributes instead of issues.targets for merged tests like this.

jkoritzinsky (approved):
We should probably also disable these tests on the mobile platforms (all xharness platforms currently have the same infra limitations).

radical:
> We should probably also disable these tests on the mobile platforms (all xharness platforms currently have the same infra limitations).

@steveisok @SamMonoRT do you know if these tests fail for ios, and other platforms? Look for `Interop/PInvoke/Vector2_3_4/Vector2_3_4/**` in `src/tests/issues.targets`, it is disabled for a few combinations.
If these tests are not failing on those platforms, then we can simply remove the exclusion. If they are failing, then we can add to the `ActiveIssue` for these.

This can be done in a follow up PR.

Issue #69281, opened 2022-05-12
[API Proposal]: Support for Span<T> and ReadOnlySpan<T> in source-generated marshalling


jkoritzinsky:
### Background and motivation

One of the goals of the source-generated marshalling in .NET 7 is to support modern types at interop boundaries. In particular, there has been significant interest in using `Span<T>` and `ReadOnlySpan<T>` at interop boundaries. This API proposal includes a number of different Span-based marshallers. This proposal does not include defining any of these marshaller types as the "default" marshallers for span types; however, we can choose to define some of these as defaults if we so desire.

1. `SpanMarshaller<T>` and `ReadOnlySpanMarshaller<T>`. These types marshal the span values in the same style as arrays. However, since the `Span` types treat an empty span as identical to a null span, these types pass `null` to native code when an empty span is passed as the managed value.
2. `NeverNullSpanMarshaller<T>` and `NeverNullReadOnlySpanMarshaller<T>`. These types marshal the span values similarly to `SpanMarshaller<T>` and `ReadOnlySpanMarshaller<T>`, but when an empty or `null` span is passed as a value, the marshaller passes a non-`null` value to native code that can be dereferenced but should not be written to. This allows developers to opt-in to similar behavior as arrays (where we don't pass `null` when the input is an empty array, but instead pass a reference to where the zeroth element would live).

We also propose updating the source generator to recognize a static `GetPinnableReference` method of the same shape as an extension `GetPinnableReference` method on the marshaller type that takes the managed type. The marshallers will implement this pattern to provide a faster path for by-value P/Invoke scenarios without requiring them to be specified as the default marshallers for the types.

It seems that we want to make the `ReadOnlySpanMarshaller` and `SpanMarshaller` types the default marshallers for their respective types. If we want to do so, then we'll add `[NativeMarshalling]` attributes to the managed types pointing at these marshallers.

### API Proposal

```csharp
namespace System.Runtime.InteropServices.Marshalling
{
    [CustomTypeMarshaller(typeof(ReadOnlySpan<>), CustomTypeMarshallerKind.LinearCollection, Direction = CustomTypeMarshallerDirection.In, Features = CustomTypeMarshallerFeatures.UnmanagedResources | CustomTypeMarshallerFeatures.CallerAllocatedBuffer | CustomTypeMarshallerFeatures.TwoStageMarshalling, BufferSize = 0x200)]
    public unsafe ref struct ReadOnlySpanMarshaller<T>
    {
[...]

PR #47703, opened 2021-02-01
Improve S.DS.P tests, fix async API, and improve error messages

danmoseley:
Fixes https://github.com/dotnet/runtime/issues/46021
Fixes https://github.com/dotnet/runtime/issues/46005

Separated cleanly into commits.

1. Add the error message carried on the DirectoryResponse to the message in the DirectoryOperationException. See note below.
2. Tell server we want v3 protocol version before attempting to Bind. (We only support v3 anyway.) Without this change, the tests give 'protocol error' against slapd.
3. Fix the async API - the refactoring in #38161 removed an unused field, but had the side-effect of removing the place we start up the thread that services the async results list. Note there is already test coverage: this test now passes - we hadn't run these tests because they require server setup.
4. Improve test configuration system and replace rotted test instructions. It is now easy to test against two different OpenLDAP based servers using docker, as well as against an Active Directory DC. (I haven't investigated testing against an AD container - I set up a DC manually). This change also clarifies that the 'domain' in the configuration file refers to the root DN, which is not the same as the user domain for the AD user account.
5. Noted whether LDAP_OPT_SERVER_ERROR gives useful text, it does not, against any of the servers on either OS. As noted in #46021, there is a bug here because  this is a Windows-only LDAP option, and the other servers map the same value to LDAP_OPT_MATCHED_DN which is unrelated. I didn't fix that here; the fix would be to somehow determine the nature of the server. This text is not included in any exception messages, and I left it that way.
6. Fix the negative error codes noted in #46021. On servers based on openLDAP headers, some of the error codes are unexpectedly negative values, leading to numerous bug reports containing the exception message `The LDAP server returned an unknown error.`. Fortunately, Windows does not use negative codes; if the code is negative, pull it into the correct range. Now the error messages are meaningful.
7. Add a bunch of unit tests for various error cases to protect this fix and also validate other error codes. I used these to generate the messages below.
8. Where the error code is still unknown, which hopefully will no longer occur, include the error code in the exception message to help us diagnose it.
9. Note the ambiguity in LDAP_OPT mentioned above.

Ran all tests on Windows and all tests on Linux, each against each of the three servers, and all now pass. 

[...]

PR #93722, opened 2023-10-19
Improve System.DirectoryServices interop guidelines conformance

chrisdcmoore:
Hopefully puts a bit of a dent in https://github.com/dotnet/runtime/issues/51563 by moving a bunch of `LibraryImport`s out of `System.DirectoryServices` to the common Interop folder as per the [interop guidelines](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/interop-guidelines.md), assuming I'm interpreting them correctly.

Remarks and questions:
 - some of the library imports didn't have existing folders / code in the common interop folder, is there any additional work that needs doing because of this?
 - should `Kernel32.CompareString` be in `Interop.Globalization` or is it more correct moving it to its own file given that S.DS doesn't require the other methods in I.G?
 - should `LsaLogonProcessSafeHandle` live with the other safe handle classes under `Microsoft\Win32\SafeHandles` or is it OK for that to live under `Secur32` as an `Interop.*.cs` file?
 - except for changing method for library imports from public to internal whilst moving them, I haven't changed the visibility of other structs/enums/classes during the move, is that the correct thing to do here?

I don't have an AD setup to test against, but the non-AD-specific tests pass against an OpenDJ server.

This is my first time refactoring interop code so a critical review would be very welcome (or if this PR has fundamental flaws due to my general lack of experience contributing to the .NET codebase, please do just say so!)


chrisdcmoore:
@dotnet-policy-service agree

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsRoleGetPrimaryDomainInformation.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            DsRolePrimaryDomainInfoBasic = 1,
+            DsRoleUpgradeStatus = 2,
+            DsRoleOperationState = 3,
+            DsRolePrimaryDomainInfoBasicEx = 4
+        }
+
+        [LibraryImport(Libraries.Netapi32, StringMarshalling = StringMarshalling.Utf16)]
+        internal static partial int DsRoleGetPrimaryDomainInformation(
+            [MarshalAs(UnmanagedType.LPTStr)] string lpServer,
```
huoyaoyuan, line 21: The definition of this parameter is `LPWCSTR` which matches UTF16. Is `LPTStr` overridden in old code?

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsGetDcNext.cs
```diff
@@ -0,0 +1,18 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Netapi32
+    {
+        [LibraryImport(Libraries.Netapi32, EntryPoint = "DsGetDcNextW", StringMarshalling = StringMarshalling.Utf16)]
```
huoyaoyuan, line 11: Do we always specify `StringMarshalling` when string is not used?

src/libraries/Common/src/Interop/Windows/Kernel32/Interop.CompareString.cs
```diff
@@ -0,0 +1,20 @@
...
+    internal static partial class Kernel32
+    {
+        [LibraryImport(Libraries.Kernel32, EntryPoint = "CompareStringW", SetLastError = true, StringMarshalling = StringMarshalling.Utf16)]
+        internal static partial int CompareString(
+            uint locale,
+            uint dwCmpFlags,
+            IntPtr lpString1,
+            int cchCount1,
+            IntPtr lpString2,
+            int cchCount2);
```
huoyaoyuan, line 18: It's more efficient to use `char*` and get rid of `Marshal` at call site. Should this be included in this PR?

src/libraries/System.DirectoryServices/src/Interop/SafeNativeMethods.cs
```diff
@@ -1,34 +1,19 @@
...
             ERROR_MORE_DATA = 234,
             ERROR_SUCCESS = 0;
 
-        [LibraryImport(global::Interop.Libraries.Activeds, StringMarshalling = StringMarshalling.Utf16)]
-        public static unsafe partial int ADsGetLastError(out int error, char* errorBuffer, int errorBufferLength, char* nameBuffer, int nameBufferLength);
-
-        [LibraryImport(global::Interop.Libraries.Activeds, StringMarshalling = StringMarshalling.Utf16)]
-        public static partial int ADsSetLastError(int error, string? errorString, string? provider);
-
         public sealed class EnumVariant
```
huoyaoyuan, line 17: It's better to rename this file since it contains `EnumVariant` only now.

huoyaoyuan (commented):
The whole `Marshal` related interop in this library can be overhauled, but it's advanced topic and may not be handful for new contributors. Let's focus on keeping consistency with existing P/Invoke definitions.

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsRoleGetPrimaryDomainInformation.cs
chrisdcmoore: Except for removing the `EntryPoint` property, I didn't change anything here. The code generated for `LPTStr` vs `LPWStr` looks identical to me (both with and without the `StringMarshalling` property set). Should I just remove the `MarshalAs` attribute altogether here?

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsGetDcNext.cs
chrisdcmoore: From glancing at a few existing files, it would appear not - I'm happy to go through and remove the redundant ones touched in this PR? Likewise with any remaining redundant `EntryPoint` properties?

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsRoleGetPrimaryDomainInformation.cs
huoyaoyuan: Yes please. `LPTStr` looks wrong here. It seems to be overridden by `StringMarshalling.Utf16`.

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsGetDcNext.cs
huoyaoyuan: Keeping consistency with existing code under Interop folder should be preferred.

src/libraries/System.DirectoryServices/src/Interop/SafeNativeMethods.cs
chrisdcmoore: Should the `EnumVariant` class be moved out of the `SafeNativeMethods` class (and made internal?) too, or leave it as-is?

There are also a couple of consts alongside the `EnumVariant`:
 - three starting `FORMAT_MESSAGE_` appear to be unused, should I delete these?
 - two appear to be only used in combination with `ADsGetLastError` and `ADsSetLastError`, should I move these to an `Interop.ErrorCodes.cs` file under `Common\src\Interop\Windows\Activeds\` perhaps?

src/libraries/Common/src/Interop/Windows/Kernel32/Interop.CompareString.cs
chrisdcmoore: If I understand correctly, that would mean changing these `IntPtr`s to `char*` (and adding the unsafe modifier to the method sig, then changing the body of https://github.com/dotnet/runtime/blob/031204bd3464f4b57e2fc215072c0d6c50ceefa4/src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Utils.cs#L1843 to something like the code below - is that correct?

```c#
        internal static int Compare(string? s1, string? s2, uint compareFlags)
        {
            // This code block was specifically written for handling string comparison
            // involving null strings. The unmanaged API "Interop.Kernel32.CompareString"
            // does not handle null strings elegantly.
            //
            // This method handles comparison of the specified strings
            // if and only if either one of the two strings or both are null.
            if (s1 == null || s2 == null)
            {
                return string.Compare(s1, s2);
            }

            int result = 0;
            int cchCount1 = 0;
            int cchCount2 = 0;

            cchCount1 = s1.Length;
            cchCount2 = s2.Length;

            unsafe
            {
                fixed (char* p1 = s1)
                fixed (char* p2 = s2)
                {
                    result = Interop.Kernel32.CompareString(LCID, compareFlags, p1, cchCount1, p2, cchCount2);
                }
            }

            if (result == 0)
            {
                throw ExceptionHelper.GetExceptionFromErrorCode(Marshal.GetLastPInvokeError());
            }

            return (result - 2); // to give the semantics of <0, ==0, >0
        }
```

huoyaoyuan: The `LibraryImport` will do the string->char* conversion for you. Just use `string` in the interop method and omit the `fixed` statement.

src/libraries/System.DirectoryServices/src/Interop/SafeNativeMethods.cs
huoyaoyuan: I'm really not sure for this case, since it's COM interop rather than plain P/invoke.

src/libraries/Common/src/Interop/Windows/Kernel32/Interop.CompareString.cs
chrisdcmoore: Ah cool, thanks for the tip! In that case I think I've followed your recommendation in badbd6feb553924c89fcc3eaa47f0b849408046c (the `StringMarhsalling` attribute was already set).

src/libraries/System.DirectoryServices/src/Interop/SafeNativeMethods.cs
chrisdcmoore: Fair enough, I've left `EnumVariant` nested under `SafeNativeMethods` for now, but have renamed the file, and deleted/moved the consts across 162ce2b60c6e9674e06d141f49da34aec662d425, b100b983705c38ec584398f0c29424ff7e92eb46 and 5aca21963f8b9c1c5af295dbf4f658095f66c8d8 respectively.

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsGetDcNext.cs
chrisdcmoore: Roger that, done in f4dfe7f7271cea758098f065284cc55a8517f149 and f78219a77aaf13f1ebfae79af7d43d806b697daf

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsRoleGetPrimaryDomainInformation.cs
chrisdcmoore: Done in f4dfe7f7271cea758098f065284cc55a8517f149

chrisdcmoore:
Thanks for your guidance @huoyaoyuan - please feel free to unresolve and continue any conversations where I might have missed the mark so far. Have a good weekend!

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsRoleGetPrimaryDomainInformation.cs
danmoseley: Agreed we should match the Windows declarations. But @Jkotas when is LPWSTR different to LPTSTR in a managed app? UNICODE is effectively always defined. (Also, what practical effect of any on marshaling does the C for const in LPCWSTR have?)

danmoseley: Ah - when CharSet is not Unicode, as already mentioned.

src/libraries/Common/src/Interop/Windows/Activeds/Interop.ErrorCodes.cs
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+internal static partial class Interop
+{
+    internal static partial class Activeds
+    {
+        public const int ERROR_SUCCESS = 0;
+        public const int ERROR_MORE_DATA = 234;
+    }
```
huoyaoyuan, line 10: The error codes are already defined at https://github.com/dotnet/runtime/blob/2959e1acd09aec0f1e59db9cbe41f3701b0bd4dd/src/libraries/Common/src/Interop/Windows/Interop.Errors.cs

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsRoleGetPrimaryDomainInformation.cs
```diff
@@ -0,0 +1,31 @@
...
+        internal static partial int DsRoleGetPrimaryDomainInformation(
+            string lpServer,
+            DSROLE_PRIMARY_DOMAIN_INFO_LEVEL InfoLevel,
+            out IntPtr Buffer);
+
+        [LibraryImport(Libraries.Netapi32, StringMarshalling = StringMarshalling.Utf16)]
+        internal static partial int DsRoleGetPrimaryDomainInformation(
+            IntPtr lpServer,
+            DSROLE_PRIMARY_DOMAIN_INFO_LEVEL InfoLevel,
+            out IntPtr Buffer);
```
huoyaoyuan, line 29: The two overloads can be consolidated into one overload with `string?`. `IntPtr.Zero` at call site should be replaced to `null`.

src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Domain.cs
```diff
@@ -1265,7 +1265,7 @@ private ArrayList GetTrustsHelper(string? targetDomainName)
             {
                 try
                 {
-                    error = UnsafeNativeMethods.DsEnumerateDomainTrustsW(serverName, (int)DS_DOMAINTRUST_FLAG.DS_DOMAIN_IN_FOREST | (int)DS_DOMAINTRUST_FLAG.DS_DOMAIN_DIRECT_OUTBOUND | (int)DS_DOMAINTRUST_FLAG.DS_DOMAIN_DIRECT_INBOUND, out domains, out count);
+                    error = Interop.Netapi32.DsEnumerateDomainTrustsW(serverName, (int)DS_DOMAINTRUST_FLAG.DS_DOMAIN_IN_FOREST | (int)DS_DOMAINTRUST_FLAG.DS_DOMAIN_DIRECT_OUTBOUND | (int)DS_DOMAINTRUST_FLAG.DS_DOMAIN_DIRECT_INBOUND, out domains, out count);
```
huoyaoyuan, line 1268: The parameter types of these methods should be changed to corresponding enums so that casts will be unnecessary.

src/libraries/Common/src/Interop/Windows/Activeds/Interop.ErrorCodes.cs
chrisdcmoore: Ahah, thanks! Fixed in 3a754919e4cff17ea0e14062a1d1ebfd3647ce4b

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsRoleGetPrimaryDomainInformation.cs
chrisdcmoore: Nice, done in c7d261a55b2f9ea572a3c3541bc4f9cda4815ca6

src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Domain.cs
chrisdcmoore: Done in 245b06684255b885c433d8332b9ed6f2cb816dd8

This did lead me down a rabbit hole though, and left me with a few more questions:
 - should the `DS_DOMAINTRUST_FLAG` enum be `: uint` (rather than the default `int` base for enums) since the corresponding method signature [takes a `ULONG` for `Flags`](https://learn.microsoft.com/en-us/windows/win32/api/dsgetdc/nf-dsgetdc-dsenumeratedomaintrustsw)?
 - similarly, there are other places where we cast this enum to an `int` when comparing to `DS_DOMAIN_TRUSTS.Flags` on the managed side, populated by a call to `Interop.Netapi32.DsEnumerateDomainTrustsW` via `out domains` (e.g. https://github.com/dotnet/runtime/blob/b09e18e8b6092e35611c2567bd55e8909ea78d7d/src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Domain.cs#L1268) - should that `DS_DOMAIN_TRUSTS` class be using `uint`s for fields that correspond to `ULONG`s on [`DS_DOMAIN_TRUSTSW` on the unmanaged side](https://learn.microsoft.com/en-us/windows/win32/api/dsgetdc/ns-dsgetdc-ds_domain_trustsw) too? Should it also be renamed and moved to common interop too?
 - I feel like the `DS_DOMAIN_TRUST_FLAGS` enum should probably also be moved to common interop folder along with some of the others?

I also spent today setting up an Active Directory server and workstation VMs, and ran the tests from the workstation. There were three test failures which appear to be pre-existing to this PR. The test failures are all due to expectations around exception types (`COMExpcetion` vs `DirectoryServicesCOMException`):
 - the first is `TestInvalidSearchFilter` which I think is a badly written test because it intentionally introduces bad credentials (I think it shouldn't, and should instead be using just a bad search filter based on the name, and be expecting an `ArgumentException`?
 - the other two (`TestInvalidUsernameAndPassword` and `TestMissingUserAndPasswordInfo`)are intentionally using bad credentials, but the confusing thing is that if you give a valid username but an invalid password, then the tests pass (`DSCException`), but if you provide a bad username _and_ password_ then `ADsGetLastError` seems to give an error code of 0 and the library doesn't convert the `COMException` and so the test fails. Is this known behaviour?

Sorry if some of these questions have obvious answers - still getting used to interop code - thanks again for your time!

chrisdcmoore:
I _think_ the test failures are unrelated, looking at the errors, but a more familiar eye would be welcome on those.

src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Domain.cs
chrisdcmoore: I've pushed a few commits which try to express my intent with the above questions - nothing that can't be reverted though!

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DS_DOMAINTRUST_FLAG.cs
```diff
@@ -0,0 +1,22 @@
...
+        internal enum DS_DOMAINTRUST_FLAG : uint
+        {
+            DS_DOMAIN_IN_FOREST = 0x0001,
+            DS_DOMAIN_DIRECT_OUTBOUND = 0x0002,
+            DS_DOMAIN_TREE_ROOT = 0x0004,
+            DS_DOMAIN_PRIMARY = 0x0008,
+            DS_DOMAIN_NATIVE_MODE = 0x0010,
+            DS_DOMAIN_DIRECT_INBOUND = 0x0020
+        }
+
```
chrisdcmoore, line 20: nit: whitespace

src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Domain.cs
danmoseley: Those existing test failures are worth an issue if there isn't one already.

These tests were mostly written a long time ago, and in some cases need cleanup; only the ones that don't need a test server are run often. The others require a test server be set up manually. (In the parts of DS that work against Linux we can use docker  - you will have seen the test scripts I created in one of the DS test folders for that. But even for Linux this is done manually)

danmoseley: (I meant https://github.com/dotnet/runtime/blob/main/src/libraries/Common/tests/System/DirectoryServices/LDAP.Configuration.xml stuff)

huoyaoyuan: > * should the `DS_DOMAINTRUST_FLAG` enum be `: uint`

As always, consistency with existing interop enum definitions should be preferred.

> Should it also be renamed and moved to common interop too?

The enum should be shared with method using the same group of flags.

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.TRUSTED_DOMAIN_FULL_INFORMATION.cs
```diff
@@ -0,0 +1,18 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Advapi32
+    {
+        [StructLayout(LayoutKind.Sequential)]
+        internal sealed class TRUSTED_DOMAIN_FULL_INFORMATION
```
AaronRobinsonMSFT, line 11:
```suggestion
        internal struct TRUSTED_DOMAIN_FULL_INFORMATION
```

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.TRUSTED_POSIX_OFFSET_INFO.cs
```diff
@@ -0,0 +1,16 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Advapi32
+    {
+        [StructLayout(LayoutKind.Sequential)]
+        internal sealed class TRUSTED_POSIX_OFFSET_INFO
```
AaronRobinsonMSFT, line 11:
```suggestion
        internal struct TRUSTED_POSIX_OFFSET_INFO
```

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DS_DOMAIN_TRUSTS.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Netapi32
+    {
+        [StructLayout(LayoutKind.Sequential)]
+        internal sealed class DS_DOMAIN_TRUSTS
```
AaronRobinsonMSFT, line 12:
```suggestion
        internal struct DS_DOMAIN_TRUSTS
```

src/libraries/Common/src/Interop/Windows/Activeds/Interop.ADsGetLastError.cs
```diff
@@ -0,0 +1,13 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Activeds
+    {
+        [LibraryImport(Libraries.Activeds, StringMarshalling = StringMarshalling.Utf16)]
+        internal static unsafe partial int ADsGetLastError(out int error, char* errorBuffer, int errorBufferLength, char* nameBuffer, int nameBufferLength);
```
AaronRobinsonMSFT, line 11: What are the desired semantics here for `nameBuffer`?

src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/DirectoryContext.cs
```diff
@@ -555,7 +555,7 @@ internal static string GetLoggedOnDomain()
         {
             string? domainName = null;
 
-            NegotiateCallerNameRequest requestBuffer = default;
+            Interop.Secur32.NegotiateCallerNameRequest requestBuffer = default;
             int requestBufferLength = (int)Marshal.SizeOf(requestBuffer);
```
AaronRobinsonMSFT, line 559: Since this is a blittable type, you can use `sizeof()`.

src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Forest.cs
```diff
@@ -1116,8 +1116,8 @@ private TrustRelationshipInformationCollection GetTrustsHelper(string? targetFor
                     IntPtr addr = (IntPtr)0;
                     for (int i = 0; i < count; i++)
                     {
-                        addr = IntPtr.Add(domains, i * Marshal.SizeOf(typeof(DS_DOMAIN_TRUSTS)));
-                        DS_DOMAIN_TRUSTS unmanagedTrust = new DS_DOMAIN_TRUSTS();
+                        addr = IntPtr.Add(domains, i * Marshal.SizeOf(typeof(Interop.Netapi32.DS_DOMAIN_TRUSTS)));
```
AaronRobinsonMSFT, line 1119: Converting this to `struct`, will allow this to be `sizeof()`.

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DS_DOMAIN_TRUSTS.cs
chrisdcmoore: Done in 235ba378caccd739fd10653a24f739a5321edae8

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.TRUSTED_POSIX_OFFSET_INFO.cs
chrisdcmoore: Done in 25b4d8e237d7d3b0ae195d0ae7bc8bcd8c87dcce

src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/DirectoryContext.cs
chrisdcmoore: Done in 285bbfa58772b9dfaff8ae3c6aae9a4ec1bc5dc6

src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Forest.cs
chrisdcmoore: Done in c1fb32f4715626bfa4abf6e424fce87aa35c6875

src/libraries/Common/src/Interop/Windows/Activeds/Interop.ADsGetLastError.cs
chrisdcmoore: I don't rightly know - I lifted this code out as-is :\

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.TRUSTED_DOMAIN_FULL_INFORMATION.cs
chrisdcmoore: Done in bf9983a0650062e3e51c81d9487c57ee892ae874

Before I resolve this, though, is this also now blittable and so should I change the `Marshal.SizeOf` calls to `sizeof` e.g. at https://github.com/chrisdcmoore/runtime/blob/285bbfa58772b9dfaff8ae3c6aae9a4ec1bc5dc6/src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/TrustHelper.cs#L661

AaronRobinsonMSFT: This has nullable types so I don't think so. If they were marked as pointers then it would be okay. You can try switching to `sizeof()`, but my guess is it will fail.

src/libraries/Common/src/Interop/Windows/Activeds/Interop.ADsGetLastError.cs
AaronRobinsonMSFT: This could be changed a bit, but I don't think it is worth it for right now. Thanks!

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.TRUSTED_DOMAIN_FULL_INFORMATION.cs
chrisdcmoore: I'm actually not sure why that field is nullable given its definition on the unmanaged side: https://learn.microsoft.com/en-us/windows/win32/api/ntsecapi/ns-ntsecapi-trusted_domain_full_information

Am I missing something or should it be not nullable?

I'm also not entirely clear on the differing visibility modifiers on its members either...

chrisdcmoore: (I'm still learning when it comes to interop code so I'm sorry if these are stupid questions!)

src/libraries/Common/src/Interop/Windows/Activeds/Interop.ADsGetLastError.cs
chrisdcmoore: I'd be curious how if you have the time to explain or point me to a resource that explains it, even if it doesn't happen in this PR, for my own learning - no worries one way or the other!

AaronRobinsonMSFT: If you look at the use cases they are all passing in a caller allocated buffer, either through `stackalloc` or via an on stack `char`. The entire point of `LibraryImport` is to make interop consistent, safe and the lowest cost possible. This example could improve the safety dimension by accepting a `char[]` and the generated code will pin the array. It is the pattern I would normally recommend, but it does come with some cost (i.e., allocating an array). Another option here is one could leverage the [`Span<T>` marshaller](https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.marshalling.spanmarshaller-2.managedtounmanagedin) with `stackalloc` and remove the explicit pointer in the signature and in the caller code.

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.TRUSTED_DOMAIN_FULL_INFORMATION.cs
AaronRobinsonMSFT: > I'm sorry if these are stupid questions!

Definitely not. These questions are helpful and appreciated.

> I'm actually not sure why that field is nullable given its definition on the unmanaged side:

Hmmm. Agreed. This signature looks wrong. These don't seem to be nullable types. It looks like the type is incorrectly defined. I don't recall all the rules for `class`s marked as "Sequential" though so perhaps it is taking advantage of that. Given the documentation I think you should update it.

> I'm also not entirely clear on the differing visibility modifiers on its members either...

That is merely for C# and has nothing to do with the marshalling logic in this case.

/cc @jkoritzinsky @jtschuster

chrisdcmoore: Updated in ef237828a51d7efc0d09ea42dda53f1b8579da26 and now that they blittable, used sizeof in 142867362d4126f6c3346ae41896d0d82636f3c2.

Whilst doing some manual testing I've come across some pre-existing bugs. If I run the following code on a Windows machine joined to a domain ("MYLAB.LOCAL") which has a child domain ("SUB.MYLAB.LOCAL") then against `net48` the code works fine, but against `net8.0-windows` it throws an `ArgumentException`:

```c#
var domain = System.DirectoryServices.ActiveDirectory.Domain.GetCurrentDomain();
var isSelectiveAuthentication = domain.GetSelectiveAuthenticationStatus("sub.mylab.local");
```

I've tracked down the reason for this as calls to `Marshal.PtrToStructure(IntPtr, object)` where `object` is a struct rather than an instance of a class (`TRUSTED_DOMAIN_INFORMATION_EX`) - in full Framework code this is still a class, but has been turned into a struct at some point in the non-Framework code.

There are a few places in `TrustHelper.cs` where this mistake is present, for example:
https://github.com/dotnet/runtime/blob/654d5cd2ceb0693b3888e0a5a490a8f1a2ba356a/src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/TrustHelper.cs#L103

Would you like me to fix these bugs in this PR or open a separate issue to track (and PR to fix)? I'm conscious that the original intent of this PR was to lift and shift code into Common Interop only, with a view to making actual improvements, and fixes to the existing failing tests, in separate PRs to this so that they are each easy to review, but that is starting to no longer be the case.

src/libraries/Common/src/Interop/Windows/Activeds/Interop.ADsGetLastError.cs
chrisdcmoore: Understood, thanks for taking the time to explain that, and I didn't know `LibraryImport` code gen worked with Span and co. now!

Something like this, then?

```suggestion
        internal static partial int ADsGetLastError(out int error, Span<char> errorBuffer, int errorBufferLength, Span<char> nameBuffer, int nameBufferLength);
```

Then change the call sites to the follow:
```diff
diff --git a/src/libraries/System.DirectoryServices/src/System/DirectoryServices/DirectoryServicesCOMException.cs b/src/libraries/System.DirectoryServices/src/System/DirectoryServices/DirectoryServicesCOMException.cs
index 8e11602eb2a..516d13e6424 100644

@@ -53,14 +53,15 @@ internal static Exception CreateFormattedComException(int hr)
             return CreateFormattedComException(new COMException(errorMsg, hr));
         }

-        internal static unsafe Exception CreateFormattedComException(COMException e)
+        internal static Exception CreateFormattedComException(COMException e)
         {
             // get extended error information
             const int ErrorBufferLength = 256;
-            char* errorBuffer = stackalloc char[ErrorBufferLength];
-            char nameBuffer = '\0';
+            Span<char> errorBuffer = stackalloc char[ErrorBufferLength];
+
+            Span<char> nameBuffer = Span<char>.Empty;
             int error = 0;
-            Interop.Activeds.ADsGetLastError(out error, errorBuffer, ErrorBufferLength, &nameBuffer, 0);
+            Interop.Activeds.ADsGetLastError(out error, errorBuffer, ErrorBufferLength, nameBuffer, 0);

             if (error != 0)
                 return new DirectoryServicesCOMException(new string(errorBuffer), error, e);
```

```diff
diff --git a/src/libraries/System.DirectoryServices/src/System/DirectoryServices/SearchResultCollection.cs b/src/libraries/System.DirectoryServices/src/System/DirectoryServices/SearchResultCollection.cs
index 4b9a4884caa..8067d2e13cf 100644
--- a/src/libraries/System.DirectoryServices/src/System/DirectoryServices/SearchResultCollection.cs
+++ b/src/libraries/System.DirectoryServices/src/System/DirectoryServices/SearchResultCollection.cs
@@ -401,11 +401,12 @@ private void CleanLastError()
                 Interop.Activeds.ADsSetLastError(Interop.Errors.ERROR_SUCCESS, null, null);
             }

-            private unsafe int GetLastError(ref int errorCode)
+            private int GetLastError(ref int errorCode)
             {
-                char c1 = '\0', c2 = '\0';
+                Span<char> c1 = Span<char>.Empty;
+                Span<char> c2 = Span<char>.Empty;
                 errorCode = Interop.Errors.ERROR_SUCCESS;
-                return Interop.Activeds.ADsGetLastError(out errorCode, &c1, 0, &c2, 0);
+                return Interop.Activeds.ADsGetLastError(out errorCode, c1, 0, c2, 0);
             }
         }
     }
```

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.TRUSTED_DOMAIN_FULL_INFORMATION.cs
AaronRobinsonMSFT: > Would you like me to fix these bugs in this PR or open a separate issue to track (and PR to fix)?

Yes please. These changes should reflect updating usage as appropriate.

> I've tracked down the reason for this as calls to Marshal.PtrToStructure(IntPtr, object) where object is a struct rather than an instance of a class (TRUSTED_DOMAIN_INFORMATION_EX)

Since the `TRUSTED_DOMAIN_INFORMATION_EX` is now a `struct`, the conversion can be as simple as converting/casting a pointer to the desired type.

src/libraries/Common/src/Interop/Windows/Activeds/Interop.ADsGetLastError.cs
AaronRobinsonMSFT: The length can be referenced through the `Span<T>` instance instead of specifying a length in a disjoint manner. For example,

```csharp
c1, c1.Length
```

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.TRUSTED_DOMAIN_FULL_INFORMATION.cs
chrisdcmoore: To be absolutely clear, that is a struct on `main`, i.e. before I touched any of this code - do you still want that bug fixed in this PR (in the more optimal way that you've suggested vs what I suggested?)

huoyaoyuan: I suggest to treat this as advanced topic and leave for next pr.
I plan to replace the `Marshal` related interop in next pr.

src/libraries/Common/src/Interop/Windows/Activeds/Interop.ADsGetLastError.cs
chrisdcmoore: Done in 497bbb6ed478fb2bcdc47da75d885cb4a29b1ca8, thanks again for the guidance.

chrisdcmoore: Ah, this isn't quite right - it's causing `COMException: Invalid pointer` for empty spans - is there a way of having them passed as not null when they're empty?

chrisdcmoore: I could make the values I pass in `Span<char> arg = Array<char>.Empty;` rather than `Span<char>.Empty` to fix this I think (since "this marshaller will pass a non-null value for a zero-length span if the span was constructed with a non-null value.") , but it seems it would be better if that could be fixed on the interop method definition side?

chrisdcmoore: On reading the description of https://github.com/dotnet/runtime/issues/69281 - I think what I actually want here is an `ArrayMarshaller` so the interop method definition should accept `char[]` instead after all?

chrisdcmoore: Fixed in 042487f0cef9828dea36df4adcb1c94f0fbd11de and explained my reasoning for the choices I've made in the commit message, please let me know if I'm still missing the mark!

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.TRUSTED_DOMAIN_FULL_INFORMATION.cs
chrisdcmoore: I've elected to update the conversion code for structs that I've touched in 62d261f632610f188c8bbd15405648401f8ae3ad and to fix the pre-existing bugs in 1e6048b212a9150e878b2d88c2fac053e2c17135 - the latter did involve marking the containing methods (should I have used an unsafe block instead?).

It sounds like you have plans to overhaul a lot of this anyway @huoyaoyuan but I thought it best to change it here then at least the bug will be fixed until such a time.

src/libraries/System.DirectoryServices/src/System/DirectoryServices/DirectoryServicesCOMException.cs
```diff
@@ -56,17 +57,19 @@ internal static Exception CreateFormattedComException(int hr)
         internal static Exception CreateFormattedComException(COMException e)
         {
             // get extended error information
-            const int ErrorBufferLength = 256;
-            Span<char> errorBuffer = stackalloc char[ErrorBufferLength];
-
-            Span<char> nameBuffer = Span<char>.Empty;
+            char[] errorBuffer = ArrayPool<char>.Shared.Rent(256);
```
AaronRobinsonMSFT, line 60: There probably isn't a lot of value to using the `ArrayPool<char>` here. I would just allocate the array and avoid the management below.

src/libraries/Common/src/Interop/Windows/Activeds/Interop.ADsGetLastError.cs
AaronRobinsonMSFT: @jkoritzinsky @jtschuster Here is another occurrence of the empty `Span<T>` logic conflicting with `null` semantics. Just FYI.

src/libraries/System.DirectoryServices/src/System/DirectoryServices/DirectoryServicesCOMException.cs
chrisdcmoore: Changed in 50aa014cc1207fa27b7493cd7ec290e46c031b4f

src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Domain.cs
chrisdcmoore: After a bit more searching I came across [this comment](https://github.com/dotnet/runtime/pull/47703#issuecomment-778678410) from your previous investigation into these test failures which I think outlines the situation as it still currently stands:

> TestInvalidSearchFilter (which passes a bad password) expect a DirectoryServicesCOMException but are getting a COMException.

As I said above, I think this test should _not_ be passing a bad password, but rather valid credentials, and then be looking for a `ArgumentException` when using the invalid filter in a search.

> TestMissingUserAndPasswordInfo expects to get an exception when enumerating the directory, but it does not.

I think the reason it isn't, is because we both ran the test on an AD-connected machine under an AD user, so the bind is inheriting from the currently logged in user... Running it on a non-AD-connected machine against OpenJD (still with no creds) passes because then it has no "ambient" auth and really does have an exception thrown.

>  TestInvalidUserAndPassword expect a DirectoryServicesCOMException but are getting a COMException. this is because AdGetLastError is returning no error after the failed bind. I do not know why.

Yeah, I came to the same conclusion for this re. `ADsGetLastError` - when the username is correct but the password is wrong, I get no error from it. When the username is wrong, `ADsGetLastError` gives back what I expect. No idea on that one.

> TestComInterfaces fails because it expects the security descriptor of a new OU it creates to correspond to the user it's bound as but instead it's BUILTIN. This does not occur on .NET Framework. I do not know why.

At least on this branch, that test doesn't run against .NET Framework (possibly due to an `ItemGroup` condition change last year?), but I get the failure against both .NET Framework and non-Framework, and I think it's because it's not checking the OU it created, but the DirectoryEntry above it instead, so it should be a simple fix.

Shall I open this up as a new issue tomorrow, so that all the info from the linked comment by you and what I've put above are in one place?

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.ImpersonateAnonymousToken.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Advapi32
+    {
+        [LibraryImport(Libraries.Advapi32, SetLastError = true)]
+        internal static partial int ImpersonateAnonymousToken(IntPtr token);
```
AaronRobinsonMSFT, line 12:
```suggestion
        [return: MarshalAs(UnmanagedType.Bool)]
        internal static partial bool ImpersonateAnonymousToken(IntPtr token);
```

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.LsaCreateTrustedDomainEx.cs
```diff
@@ -0,0 +1,49 @@
...
+            public int IncomingAuthInfos;
+            public IntPtr IncomingAuthenticationInformation;
+            public IntPtr IncomingPreviousAuthenticationInformation;
+            public int OutgoingAuthInfos;
+            public IntPtr OutgoingAuthenticationInformation;
+            public IntPtr OutgoingPreviousAuthenticationInformation;
+        }
+
+        [LibraryImport(Libraries.Advapi32)]
+        internal static partial uint LsaCreateTrustedDomainEx(SafeLsaPolicyHandle handle, in TRUSTED_DOMAIN_INFORMATION_EX domainEx, in TRUSTED_DOMAIN_AUTH_INFORMATION authInfo, int classInfo, out IntPtr domainHandle);
```
AaronRobinsonMSFT, line 47:
```suggestion
        internal static partial uint LsaCreateTrustedDomainEx(SafeLsaPolicyHandle handle, in TRUSTED_DOMAIN_INFORMATION_EX domainEx, in TRUSTED_DOMAIN_AUTH_INFORMATION authInfo, int desiredAccess, out IntPtr domainHandle);
```

https://learn.microsoft.com/windows/win32/api/ntsecapi/nf-ntsecapi-lsacreatetrusteddomainex

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.LsaSetForestTrustInformation.cs
```diff
@@ -0,0 +1,15 @@
...
+using System;
+using System.Runtime.InteropServices;
+using Microsoft.Win32.SafeHandles;
+
+internal static partial class Interop
+{
+    internal static partial class Advapi32
+    {
+        [LibraryImport(Libraries.Advapi32)]
+        internal static partial uint LsaSetForestTrustInformation(SafeLsaPolicyHandle handle, in UNICODE_STRING target, IntPtr forestTrustInfo, int checkOnly, out IntPtr collisionInfo);
```
AaronRobinsonMSFT, line 13:
```suggestion
        internal static partial uint LsaSetForestTrustInformation(SafeLsaPolicyHandle handle, in UNICODE_STRING target, IntPtr forestTrustInfo, [MarshalAs(UnmanagedType.U1)] bool checkOnly, out IntPtr collisionInfo);
```
The `BOOLEAN` data type is a single byte - https://learn.microsoft.com/windows/win32/api/ntsecapi/nf-ntsecapi-lsasetforesttrustinformation

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.TRUSTED_DOMAIN_FULL_INFORMATION.cs
```diff
@@ -0,0 +1,18 @@
...
+internal static partial class Interop
+{
+    internal static partial class Advapi32
+    {
+        [StructLayout(LayoutKind.Sequential)]
+        internal struct TRUSTED_DOMAIN_FULL_INFORMATION
+        {
+            public TRUSTED_DOMAIN_INFORMATION_EX Information;
+            internal TRUSTED_POSIX_OFFSET_INFO PosixOffset;
+            public TRUSTED_DOMAIN_AUTH_INFORMATION AuthInformation;
```
AaronRobinsonMSFT, line 15:
```suggestion
            internal TRUSTED_DOMAIN_INFORMATION_EX Information;
            internal TRUSTED_POSIX_OFFSET_INFO PosixOffset;
            internal TRUSTED_DOMAIN_AUTH_INFORMATION AuthInformation;
```

Just to keep them consistent. The type itself is `internal` so this is a safe operation.

src/libraries/Common/src/Interop/Windows/Dnsapi/Interop.DnsRecordListFree.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Dnsapi
+    {
+        [LibraryImport(Libraries.Dnsapi)]
+        internal static partial void DnsRecordListFree(
+            IntPtr dnsResultList,
+            [MarshalAs(UnmanagedType.Bool)] bool dnsFreeType);
```
AaronRobinsonMSFT, line 14:
```suggestion
        [LibraryImport(Libraries.Dnsapi)]
        internal static partial void DnsRecordListFree(
            IntPtr dnsResultList,
            [MarshalAs(UnmanagedType.Bool)] bool dnsFreeType);
```

The above seems to be a bit odd. The `WinDNS.h` header defines this as follows:

```c
#if(_WIN32_WINNT >= 0x0501)
#define DnsRecordListFree(p,t)  DnsFree(p,DnsFreeRecordList)
#else
VOID
WINAPI
DnsRecordListFree(
    _Inout_opt_ PDNS_RECORD     pRecordList,
    _In_        DNS_FREE_TYPE   FreeType
    );
#endif /* _WIN32_WINNT >= 0x0501 */
```

The signature here is off and I _think_ this should just point to `DnsFree()` as we are long past Windows 2000.

Feel free to make this method `private` and keep the `DNS_FREE_TYPE` argument as an implementation detail as an `int`. You can then keep this signature as is, but implement it with `DnsFree()`.

src/libraries/Common/src/Interop/Windows/Interop.Errors.cs
```diff
@@ -62,24 +64,29 @@ internal static partial class Errors
...
+        internal const int ERROR_INVALID_DOMAINNAME = 0x4BC;
         internal const int ERROR_CANCELLED = 0x4C7;
         internal const int ERROR_NETWORK_UNREACHABLE = 0x4CF;
         internal const int ERROR_NON_ACCOUNT_SID = 0x4E9;
         internal const int ERROR_NOT_ALL_ASSIGNED = 0x514;
         internal const int ERROR_UNKNOWN_REVISION = 0x519;
+        internal const int ERROR_NO_SUCH_LOGON_SESSION = 0x520;
         internal const int ERROR_INVALID_OWNER = 0x51B;
         internal const int ERROR_INVALID_PRIMARY_GROUP = 0x51C;
+        internal const int ERROR_NO_LOGON_SERVERS = 0x51F;
```
AaronRobinsonMSFT, line 89:
```suggestion
        internal const int ERROR_INVALID_OWNER = 0x51B;
        internal const int ERROR_INVALID_PRIMARY_GROUP = 0x51C;
        internal const int ERROR_NO_LOGON_SERVERS = 0x51F;
        internal const int ERROR_NO_SUCH_LOGON_SESSION = 0x520;
```

src/libraries/Common/src/Interop/Windows/Kernel32/Interop.OpenThread_IntPtr.cs
```diff
@@ -0,0 +1,18 @@
...
+
+internal static partial class Interop
+{
+    internal static partial class Kernel32
+    {
+        private const uint STANDARD_RIGHTS_REQUIRED = 0x000F0000;
+        private const uint SYNCHRONIZE = 0x00100000;
+        internal const uint THREAD_ALL_ACCESS = STANDARD_RIGHTS_REQUIRED | SYNCHRONIZE | 0x3FF;
+
+        [LibraryImport(Libraries.Kernel32, EntryPoint = "OpenThread", SetLastError = true)]
```
AaronRobinsonMSFT, line 15:
```suggestion
        [LibraryImport(Libraries.Kernel32, SetLastError = true)]
```

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsGetDcNext.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Netapi32
+    {
+        [LibraryImport(Libraries.Netapi32, EntryPoint = "DsGetDcNextW")]
+        internal static partial int DsGetDcNext(
+            IntPtr getDcContextHandle,
+            ref IntPtr sockAddressCount,
```
AaronRobinsonMSFT, line 14:
```suggestion
            out IntPtr sockAddressCount,
```

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsRoleFreeMemory.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Netapi32
+    {
+        [LibraryImport(Libraries.Netapi32)]
+        internal static partial int DsRoleFreeMemory(IntPtr buffer);
```
AaronRobinsonMSFT, line 12:
```suggestion
        internal static partial void DsRoleFreeMemory(IntPtr buffer);
```

https://learn.microsoft.com/windows/win32/api/dsrole/nf-dsrole-dsrolefreememory

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.I_NetLogonControl2.cs
```diff
@@ -0,0 +1,14 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Netapi32
+    {
+        [LibraryImport(Libraries.Netapi32, EntryPoint = "I_NetLogonControl2", StringMarshalling = StringMarshalling.Utf16)]
```
AaronRobinsonMSFT, line 11:
```suggestion
        [LibraryImport(Libraries.Netapi32, StringMarshalling = StringMarshalling.Utf16)]
```

src/libraries/Common/src/Interop/Windows/NtDll/Interop.RtlInitUnicodeString.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class NtDll
+    {
+        [LibraryImport(Libraries.NtDll)]
+        internal static partial int RtlInitUnicodeString(out UNICODE_STRING result, IntPtr s);
```
AaronRobinsonMSFT, line 12:
```suggestion
        internal static partial void RtlInitUnicodeString(out UNICODE_STRING result, IntPtr s);
```

https://learn.microsoft.com/windows/win32/api/winternl/nf-winternl-rtlinitunicodestring

danmoseley: just curious, what would happen if 'int' was used here -- we would just receive garbage, presumably.

danmoseley: and speaking generally, do we have an options for detecting this kind of error? for example, comparing to the database that cswin32 runs off.

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.LsaSetForestTrustInformation.cs
danmoseley: fwiw the handy table lists this: https://learn.microsoft.com/en-us/dotnet/standard/native-interop/best-practices#common-windows-data-types

src/libraries/Common/src/Interop/Windows/NtDll/Interop.RtlInitUnicodeString.cs
AaronRobinsonMSFT: > just curious, what would happen if 'int' was used here -- we would just receive garbage, presumably.

Yep. For x86 platforms we get the last value of the `eax`/`rax` register.

> do we have an options for detecting this kind of error? for example, comparing to the database that cswin32 runs off.

Not at present. CsWin32 is the best approach for this. However, it is using a technically invalid mechanism for Roslyn source generators. I suppose that has been fixed now, but originally it was invalid.

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.ImpersonateAnonymousToken.cs
chrisdcmoore: 462826c917e

src/libraries/Common/src/Interop/Windows/Advapi32/Interop.LsaSetForestTrustInformation.cs
chrisdcmoore: 23ad205c7bd327a74318757457f3f0babed5e3eb

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsGetDcNext.cs
chrisdcmoore: Done in 80536c19b1c85a2a4606868670b9016688500468, however I noticed that the [documentation](https://learn.microsoft.com/en-us/windows/win32/api/dsgetdc/nf-dsgetdc-dsgetdcnextw) states that the caller should be calling `LocalFree` on the memory pointed to by `sockAddresses` but we don't seem to do that in https://github.com/dotnet/runtime/blob/920cd6e2b01d0fbede8882d015329e9b4d448c98/src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Locator.cs#L30

Is this something that should be fixed with a call to `Interop.Kernel32.LocalFree` (or `Marshal.FreeHGlobal`?) after each call (similar to how `dnsHostNamePtr` memory _is_ freed, if necessary, with `NetApiBufferFree`), or am I misunderstanding something?

chrisdcmoore: If it helps explain what I mean, it looks like the sock structures here are accumulating with each call to `DsGetDcNext` and remain past the method returning - if I put a call to `FreeHGlobal(sockAddressList)` in the `finally` block where we free the `dcDnsHostNamePtr` then the memory appears to get re-used across calls...

image

src/libraries/Common/src/Interop/Windows/Dnsapi/Interop.DnsRecordListFree.cs
chrisdcmoore: I was about to ask whether it would be preferable to just replace it wholesale (e.g. import `DnsFree`, define `enum DNS_FREE_TYPE : uint`, update call sites, etc....

But then the only two places `DnsRecordListFree` and `DnsQuery` are used are in `S.DS.AD.Locator.DnsQueryWrapper` which appears to be completely unused - should I just delete that private method, and purge the `Dnsapi` interop code altogether? It doesn't seem to be used in any other libraries...

```
Chris@DESKTOP-C8V4VRD MINGW64 ~/src/dotnet/runtime (system-directoryservices-interop-refactor)
$ grep -r --include '*.cs' DnsQueryWrapper src/libraries
src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Locator.cs:        private static Hashtable DnsQueryWrapper(string domainName, string siteName, long dcFlags)

Chris@DESKTOP-C8V4VRD MINGW64 ~/src/dotnet/runtime (system-directoryservices-interop-refactor)
$ grep -r --include '*.cs' DnsQuery src/libraries
src/libraries/Common/src/Interop/Windows/Dnsapi/Interop.DnsQuery.cs:        [LibraryImport(Libraries.Dnsapi, EntryPoint = "DnsQuery_W", StringMarshalling = StringMarshalling.Utf16)]
src/libraries/Common/src/Interop/Windows/Dnsapi/Interop.DnsQuery.cs:        internal static partial int DnsQuery(
src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Locator.cs:        private static Hashtable DnsQueryWrapper(string domainName, string siteName, long dcFlags)
src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Locator.cs:                options |= NativeMethods.DnsQueryBypassCache;
src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Locator.cs:            // Call DnsQuery
src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Locator.cs:            result = Interop.Dnsapi.DnsQuery(recordName, NativeMethods.DnsSrvData, options, IntPtr.Zero, out dnsResults, IntPtr.Zero);
src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/NativeMethods.cs:        internal const int DnsQueryBypassCache = 8;

Chris@DESKTOP-C8V4VRD MINGW64 ~/src/dotnet/runtime (system-directoryservices-interop-refactor)
$ grep -r --include '*.cs' DnsRecordListFree src/libraries
src/libraries/Common/src/Interop/Windows/Dnsapi/Interop.DnsRecordListFree.cs:        internal static partial void DnsRecordListFree(
src/libraries/System.DirectoryServices/src/System/DirectoryServices/ActiveDirectory/Locator.cs:                        Interop.Dnsapi.DnsRecordListFree(dnsResults, true);
```

AaronRobinsonMSFT: > should I just delete that private method, and purge the Dnsapi interop code altogether?

+1 Yes! Thank you for looking at that.

src/libraries/Common/src/Interop/Windows/Netapi32/Interop.DsGetDcNext.cs
AaronRobinsonMSFT: > however I noticed that the [documentation](https://learn.microsoft.com/en-us/windows/win32/api/dsgetdc/nf-dsgetdc-dsgetdcnextw) states that the caller should be calling LocalFree on the memory pointed to by sockAddresses but we don't seem to do that

Looks like you found a leak :) Nice job! Yes, this should be cleaned up prior to moving to the next iteration.

chrisdcmoore: Sweet, fixed in 4f9d43742b52d994edd4278f3490e3d5431d5310

src/libraries/Common/src/Interop/Windows/Dnsapi/Interop.DnsRecordListFree.cs
chrisdcmoore: Nice, done in 8808eed2df5301b04268fdf2cdde9a47fb9b5f74

AaronRobinsonMSFT: It looks like you can also delete `PartialDnsRecord` and `DnsRecord` in the `NativeMethods.cs` file, can you confirm that?

chrisdcmoore: Yes indeed, along with some Dnsapi-related consts in there too - done in b5c22d673c003682851be052ffa80cfdbfd5bc9e

AaronRobinsonMSFT (approved):
Thank you!

@buyaa-n This looks good to me from the interop side of things.

chrisdcmoore:
> Thank you!

Thank you, and thanks to all the other folks on this PR, for the guidance - I learnt a lot from it 😊

buyaa-n:
Thank you @chrisdcmoore @AaronRobinsonMSFT, the changes look good to me, this PR touches the part that owned by windows team, so requested review from them.

intothedarkness (approved):
the code change looks pretty good to me, the supplement brings a neat and tidy structure to the interop code of .NET SDS APIs. thanks a lot for the effort!

src/libraries/Common/src/Interop/Windows/Interop.Libraries.cs
```diff
@@ -12,7 +12,6 @@ internal static partial class Libraries
         internal const string Credui = "credui.dll";
         internal const string Crypt32 = "crypt32.dll";
         internal const string CryptUI = "cryptui.dll";
-        internal const string Dnsapi = "dnsapi.dll";
```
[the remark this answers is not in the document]
chrisdcmoore: Whilst discussing the Dnsapi code changes, I realised that the two places where we called methods from `dnsapi.dll` were all in dead code in S.DS, and then grepping over the rest of the runtime codebase showed this to be the only place we interop'd with `dnsapi.dll`: 
https://github.com/dotnet/runtime/pull/93722/files/50aa014cc1207fa27b7493cd7ec290e46c031b4f#diff-f30870c9aa28afc8698540506686362776b851bd5e4775f644b84779d7967461

buyaa-n:
Sorry for taking so long for the review @chrisdcmoore last time I heard windows team haven't done with the review, I will follow up again, meanwhile the PR got a conflict with main, could you resolve them?

kumarravik78c (approved):

chrisdcmoore:
> Sorry for taking so long for the review @chrisdcmoore last time I heard windows team haven't done with the review, I will follow up again, meanwhile the PR got a conflict with main, could you resolve them?

No worries @buyaa-n - totally get that folks are busy and it's been the holiday season, etc.

I've merged main in from upstream and resolved the merge conflicts - everything seemed to build fine locally, but we'll see what CI has to say about it!

If more conflicts arise, feel free to give me another nudge and I can resolve again to stop this branch getting too stale 👍

buyaa-n:
Thank you @chrisdcmoore, the CI on 2 legs are timed out, no other failure, merging.

PR #93734, opened 2023-10-19
[RISC-V] Test NFloat.IsNegative against its native type equivalent

tomeksowi:
Comparing it with the result of float.IsNegative failed for RISC-V64 because NFloat.IsNegative implicitly converts to NFloat, which upcasts to NativeType (double), which canonicalizes the NaN value (drops the sign and payload) on RISC-V. Comparing it to double.IsNegative ensures a similar platform-specific upcast is on the expected side.

I saw some other tests feature a similar condition.

This fixes System.Runtime.InteropServices.Tests.NFloatTests.IsNegative

Part of #84834
cc @wscho77 @HJLeee @JongHeonChoi @t-mustafin @gbalykov @clamp03 @Sirntar @yurai007

### DIFF
--- a/src/libraries/System.Runtime.InteropServices/tests/System.Runtime.InteropServices.UnitTests/System/Runtime/InteropServices/NFloatTests.cs
+++ b/src/libraries/System.Runtime.InteropServices/tests/System.Runtime.InteropServices.UnitTests/System/Runtime/InteropServices/NFloatTests.cs
@@ -793,7 +793,14 @@ namespace System.Runtime.InteropServices.Tests
         public void IsNegative(float value)
         {
             bool result = NFloat.IsNegative(value);
-            Assert.Equal(float.IsNegative(value), result);
+            if (Environment.Is64BitProcess)
+            {
+                Assert.Equal(double.IsNegative(value), result);
+            }
+            else
+            {
+                Assert.Equal(float.IsNegative(value), result);
+            }
         }
 
         [Theory]


gbalykov (approved):

AaronRobinsonMSFT:
@tannergooding Please advise on what you would like to see for this issue.

tannergooding (approved):
LGTM.

It's unfortunate that RISC-V works so differently from other platforms here and that the data is lost. It is, however, allowed by the IEEE 754 specification and is considered a "value preserving" optimization, so it's ultimately "fine".

src/libraries/System.Runtime.InteropServices/tests/System.Runtime.InteropServices.UnitTests/System/Runtime/InteropServices/NFloatTests.cs
```diff
@@ -793,7 +793,8 @@ public void IsNaN(float value)
         public void IsNegative(float value)
         {
             bool result = NFloat.IsNegative(value);
-            Assert.Equal(float.IsNegative(value), result);
+            bool expected = Environment.Is64BitProcess ? double.IsNegative(value) : float.IsNegative(value);
```
tannergooding, line 796: nit: It might be nice if this were setup similarly to other tests that have to do the same.

See `IsNormal` for an example.

clamp03 (approved):

src/libraries/System.Runtime.InteropServices/tests/System.Runtime.InteropServices.UnitTests/System/Runtime/InteropServices/NFloatTests.cs
tomeksowi: ok

tomeksowi:
@tannergooding anything to do with this PR or is it good to merge?

PR #93744, opened 2023-10-19
Don't use File.OpenWrite when trying to overwrite a file

akoeplinger:
File.OpenWrite will open an existing file which means if you write data that is shorter than the existing data the file will be corrupt. This is documented on https://learn.microsoft.com/en-us/dotnet/api/system.io.file.openwrite

> The OpenWrite method opens a file if one already exists for the file path, or creates a new file if one does not exist. For an existing file, it does not append the new text to the existing text. Instead, it overwrites the existing characters with the new characters. If you overwrite a longer string (such as "This is a test of the OpenWrite method") with a shorter string (such as "Second run"), the file will contain a mix of the strings ("Second runtest of the OpenWrite method").

The fix is to use the `FileMode.Create` option in cases where the intention is to create a new file or overwrite an existing one.

I noticed this recently when using the illink feature to dump a dependency analysis xml, the next time I ran the build after making a change that'd reduce dependencies the xml would get corrupted since it wrote less data than the existing file had.

There were a couple instances across the codebase, this fixes them.

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILAssemblyGeneratingMethodDebugInfoProvider.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILAssemblyGeneratingMethodDebugInfoProvider.cs
@@ -28,7 +28,7 @@ namespace ILCompiler
         {
             _wrappedProvider = wrappedProvider;
             _fileName = fileName;
-            _tw = new StreamWriter(File.OpenWrite(fileName));
+            _tw = new StreamWriter(new FileStream(fileName, FileMode.Create, FileAccess.Write, FileShare.None));
         }
 
         public override MethodDebugInformation GetDebugInfo(MethodIL methodIL)
--- a/src/tasks/MonoTargetsTasks/RuntimeConfigParser/RuntimeConfigParser.cs
+++ b/src/tasks/MonoTargetsTasks/RuntimeConfigParser/RuntimeConfigParser.cs
@@ -71,7 +71,7 @@ public class RuntimeConfigParserTask : Task
         ConvertDictionaryToBlob(configProperties, blobBuilder);
 
         Directory.CreateDirectory(Path.GetDirectoryName(OutputFile!)!);
-        using var stream = File.OpenWrite(OutputFile);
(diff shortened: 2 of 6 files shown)


radical (approved):

radical:
Good catch!

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILAssemblyGeneratingMethodDebugInfoProvider.cs
```diff
@@ -28,7 +28,7 @@ public ILAssemblyGeneratingMethodDebugInfoProvider(string fileName, DebugInforma
         {
             _wrappedProvider = wrappedProvider;
             _fileName = fileName;
-            _tw = new StreamWriter(File.OpenWrite(fileName));
+            _tw = new StreamWriter(new FileStream(fileName, FileMode.Create, FileAccess.Write, FileShare.None));
```
MichalStrehovsky, line 31: More compact alternative:

```suggestion
            _tw = new StreamWriter(File.Create(fileName));
```

MichalStrehovsky (approved):
Thank you!

vitek-karas (approved):

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/ILAssemblyGeneratingMethodDebugInfoProvider.cs
akoeplinger: Not that it matters here but File.Create() uses FileAccess.ReadWrite instead of FileAccess.Write so it is slightly different.
I'll keep this as is to avoid restarting CI but let me know if you feel strongly.

lewing:
we should probably backport the parts that apply to 8.0

MichalStrehovsky:
> we should probably backport the parts that apply to 8.0

The nativeaot part is related to producing a file used to debug the compiler. The code is under a non-public switch. The risk is pretty much none, but it's also not important for 8.0. This obviously only applies to that part of this pr.

akoeplinger:
Backported the relevant changes in https://github.com/dotnet/runtime/pull/93854

PR #93750, opened 2023-10-19
Update NotNullIfNotNull documentation

RikkiGibson:
Related to dotnet/roslyn#70414

Will have to figure out if this is the right doc change to make, and if this is the right place to make the change.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/NullableAttributes.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/NullableAttributes.cs
@@ -89,7 +89,7 @@ namespace System.Diagnostics.CodeAnalysis
     {
         /// <summary>Initializes the attribute with the associated parameter name.</summary>
         /// <param name="parameterName">
-        /// The associated parameter name.  The output will be non-null if the argument to the parameter specified is non-null.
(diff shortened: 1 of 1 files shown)


stephentoub (approved):

stephentoub:
Changes to these XML comments won't automatically flow to the docs. You'll need to also submit a PR here:
https://github.com/dotnet/dotnet-api-docs/blob/6cc8bf120d24105e31c32f9b5c1353cb0e9efe9f/xml/System.Diagnostics.CodeAnalysis/NotNullIfNotNullAttribute.xml#L76-L108

src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/NullableAttributes.cs
```diff
@@ -89,7 +89,7 @@ sealed class NotNullIfNotNullAttribute : Attribute
     {
         /// <summary>Initializes the attribute with the associated parameter name.</summary>
         /// <param name="parameterName">
-        /// The associated parameter name.  The output will be non-null if the argument to the parameter specified is non-null.
+        /// The associated parameter name.  The value will be non-null if the argument to the parameter specified is non-null.
```
joegoldman2, line 92:
```suggestion
        /// The associated parameter name. The value will be non-null if the argument to the parameter specified is non-null.
```

PR #93756, opened 2023-10-19
[mono] Test upgrading to ICU 72.3

steveisok:


lewing:
looks like they deprecated some apis 😬

src/mono/wasm/runtime/es6/dotnet.es6.lib.js
```diff
@@ -68,7 +68,7 @@ function setup(linkerSetup) {
 
 const DotnetSupportLib = {
     $DOTNET: { setup },
-    icudt68_dat: function () { throw new Error('dummy link symbol') },
+    icudt73_dat: function () { throw new Error('dummy link symbol') },
```
ilonatommy, line 71: Shouldn't it be `icudt72_dat`?
ToDo: Investigate this on the future attempt.

PR #93759, opened 2023-10-19
Optimize JsonContent fast path

pentp:
Remove a few `async` methods from the common fast path.

### DIFF
--- a/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/JsonContent.netcoreapp.cs
+++ b/src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/JsonContent.netcoreapp.cs
@@ -13,14 +13,9 @@ namespace System.Net.Http.Json
     {
         protected override void SerializeToStream(Stream stream, TransportContext? context, CancellationToken cancellationToken)
         {
-            Encoding? targetEncoding = JsonHelpers.GetEncoding(this);
-            if (targetEncoding != null && targetEncoding != Encoding.UTF8)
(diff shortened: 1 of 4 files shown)


src/libraries/System.Net.Http.Json/src/System/Net/Http/Json/JsonContent.netcoreapp.cs
```diff
@@ -10,9 +12,23 @@ namespace System.Net.Http.Json
     public sealed partial class JsonContent
     {
         protected override void SerializeToStream(Stream stream, TransportContext? context, CancellationToken cancellationToken)
-            => SerializeToStreamAsyncCore(stream, async: false, cancellationToken).GetAwaiter().GetResult();
+        {
+            Encoding? targetEncoding = JsonHelpers.GetEncoding(this);
```
stephentoub, line 16: The previous version kept all of the logic in the same method and just chose which to call based on the async argument. Now the logic is distributed across two different files, which makes it harder to keep in sync.  When the async argument was false, there wouldn't be any allocation, just a little bit of overhead associated with calling an async method that would complete synchronously, and only for a path that we don't expect to be used particularly frequently. I'd prefer if this part of the PR were reverted. The other ValueTask-related optimizations look good, though.

pentp: The only logic that's duplicated now is the encoding check - all the other logic was actually more complicated before because of the sync/async mix.

stephentoub (approved):

eiriktsarpalis (approved):
Thanks

PR #93761, opened 2023-10-20
[6.0] Disable STJ build for 6.0.25

carlossanlop:
We just did branding a couple days ago. The previous STJ fix should be available in 6.0.24.

We missed enabling MWC in the previous STJ change, so I'm enabling it now.

### DIFF
--- a/src/libraries/Microsoft.Windows.Compatibility/src/Microsoft.Windows.Compatibility.csproj
+++ b/src/libraries/Microsoft.Windows.Compatibility/src/Microsoft.Windows.Compatibility.csproj
@@ -5,8 +5,8 @@
     <!-- Reference the outputs for the dependency nodes calculation. -->
     <NoTargetsDoNotReferenceOutputAssemblies>false</NoTargetsDoNotReferenceOutputAssemblies>
     <IsPackable>true</IsPackable>
-    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
-    <ServicingVersion>6</ServicingVersion>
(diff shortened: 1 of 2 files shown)


src/libraries/Microsoft.Windows.Compatibility/src/Microsoft.Windows.Compatibility.csproj
```diff
@@ -5,8 +5,8 @@
     <!-- Reference the outputs for the dependency nodes calculation. -->
     <NoTargetsDoNotReferenceOutputAssemblies>false</NoTargetsDoNotReferenceOutputAssemblies>
     <IsPackable>true</IsPackable>
-    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
-    <ServicingVersion>6</ServicingVersion>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
ericstj, line 8: Why are we enabling this?

src/libraries/Microsoft.Windows.Compatibility/src/Microsoft.Windows.Compatibility.csproj
```diff
@@ -5,8 +5,8 @@
...
+    <ServicingVersion>7</ServicingVersion>
```
ericstj, line 9: Per discussion - STJ isn't actually referenced by this so it wasn't missed.
```suggestion
    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
    <ServicingVersion>6</ServicingVersion>
```

PR #93764, opened 2023-10-20
JIT: create recursive tail call scratch block early in morph

AndyAyersMS:
Create the scratch BB (if likely to be needed) before we get into morphing all the blocks, so that morph is not creating new blocks while morphing.

Also enhance detection of potential recursive tall calls to include cases created by GDV.

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -1901,17 +1901,16 @@ void Compiler::compInit(ArenaAllocator*       pAlloc,
         codeGen = nullptr;
     }
 
-    compJmpOpUsed                = false;
-    compLongUsed                 = false;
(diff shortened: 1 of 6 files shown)


AndyAyersMS:
cc @dotnet/jit-contrib 

Likely no diffs (local run hasn't finished yet). Seems like we might be able to craft a case that exposes the prior lack of `BBF_RECURSIVE_TAILCALL` (which was added in #79346) for GDV cases, but I haven't tried to do so.

jakobbotsch (approved):

AndyAyersMS:
Timeouts for x86 on both diff and replay are a bit concerning. Retrying diffs in CI and replay locally.

AndyAyersMS:
Local run was fine, so assuming these timeouts are related to some CI issue.

PR #93771, opened 2023-10-20
hosting windows service helper enhancement

### LINKED ISSUE #93613: Performance Issue with `IsWindowsService` due to Repeated Process Lookups
### Description

The `WindowsServiceHelpers.IsWindowsService` method in the `Microsoft.Extensions.Hosting.WindowsServices` package has a performance issue. It retrieves its parent process each time it is called, which takes approximately 35.93ms. While this may not seem significant, the time accumulates when the method is used multiple times, leading to noticeable delays.

To reproduce this issue, simply call the `WindowsServiceHelpers.IsWindowsService` method multiple times in a loop.

### Configuration

The issue was discovered on .NET 6.0, running on a Windows 11 Pro x64 (build 22621.1555) machine with the `Microsoft.Extensions.Hosting.WindowsServices` package version 6.0.2.
However, it is likely to be present on any system where this package is used. The machine had about 600 processes running at the time of testing. The test was done in release mode.

### Regression?

This issue has been present since the introduction of the `WindowsServiceHelpers.IsWindowsService` method. It is not a regression from a previous build or release.

### Data

Each call to `WindowsServiceHelpers.IsWindowsService` takes approximately 35.93ms due to the method retrieving its parent process each time. When used multiple times, these delays stack up. For example, calling the method 100 times would result in a delay of approximately 3.59 seconds.

### Analysis

The issue seems to stem from the method retrieving its parent process each time it is called. A potential solution could be to cache the parent process of the current process when the method is called for the first time, and then use this cached value in subsequent calls. This would prevent the need for repeated, time-consuming lookups.

WeihanLi:
fixes #93613

### DIFF
--- a/src/libraries/Microsoft.Extensions.Hosting.WindowsServices/src/WindowsServiceHelpers.cs
+++ b/src/libraries/Microsoft.Extensions.Hosting.WindowsServices/src/WindowsServiceHelpers.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Extensions.Hosting.WindowsServices
     /// </summary>
     public static class WindowsServiceHelpers
     {
-        private static bool? _isWindowService;
+        private static readonly bool _isWindowService = GetIsWindowsService();
 
         /// <summary>
         /// Check if the current process is hosted as a Windows Service.
@@ -21,8 +21,7 @@ namespace Microsoft.Extensions.Hosting.WindowsServices
         /// <see langword="true" /> if the current process is hosted as a Windows Service; otherwise, <see langword="false" />.
         /// </returns>
         [SupportedOSPlatformGuard("windows")]
-        public static bool IsWindowsService()
-            => _isWindowService ??= GetIsWindowsService();
+        public static bool IsWindowsService() => _isWindowService;
 
         private static bool GetIsWindowsService()
         {
--- a/src/libraries/Microsoft.Extensions.Hosting.WindowsServices/src/WindowsServiceHelpers.cs
+++ b/src/libraries/Microsoft.Extensions.Hosting.WindowsServices/src/WindowsServiceHelpers.cs
@@ -12,6 +12,8 @@ namespace Microsoft.Extensions.Hosting.WindowsServices
     /// </summary>
     public static class WindowsServiceHelpers
     {
+        private static bool? _isWindowService;
+
         /// <summary>
         /// Check if the current process is hosted as a Windows Service.
         /// </summary>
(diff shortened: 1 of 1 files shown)


src/libraries/Microsoft.Extensions.Hosting.WindowsServices/src/WindowsServiceHelpers.cs
```diff
@@ -12,6 +12,8 @@ namespace Microsoft.Extensions.Hosting.WindowsServices
     /// </summary>
     public static class WindowsServiceHelpers
     {
+        private static bool? _isWindowService;
```
huoyaoyuan, line 15: The field should be `readonly bool` so that JIT can turn it into constant in tiered code.

src/libraries/Microsoft.Extensions.Hosting.WindowsServices/src/WindowsServiceHelpers.cs
```diff
@@ -20,6 +22,9 @@ public static class WindowsServiceHelpers
         /// </returns>
         [SupportedOSPlatformGuard("windows")]
         public static bool IsWindowsService()
+            => _isWindowService ??= GetIsWindowsService();
```
huoyaoyuan, line 25: The initialization doesn't need to be "lazy" in this way. It can be in static field initializer and will run when this helper class is first touched.

WeihanLi: reasonable, thanks

WeihanLi: I use this way to follow the style in `SystemdHelper` 

https://github.com/dotnet/runtime/blob/111b089775820886b01607f4c7d0f6a2e7b81226/src/libraries/Microsoft.Extensions.Hosting.Systemd/src/SystemdHelpers.cs#L27-L28

Should I update the usage in `SystemdHelper`?

huoyaoyuan:
I wonder if we can convert `IsOSPlatform` to `OperatingSystem.IsWindows()` and pull the check out. Will this help for trimming?

buyaa-n:
> I wonder if we can convert `IsOSPlatform` to `OperatingSystem.IsWindows()` and pull the check out. Will this help for trimming?

Not sure how it affects trimming but, using `OperatingSystem.IsWindows()` on .NET 5+ build is preferable. @WeihanLi could you add an `#elif` case for that? Thank you!

```cs
            if (
#if NETFRAMEWORK
                Environment.OSVersion.Platform != PlatformID.Win32NT
#elif NET5_0_OR_GREATER
                !OperatingSystem.IsWindows()
#else
                !RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
#endif
                )
            {
                return false;
            }
```

buyaa-n (approved):
LGTM, thank you!

PR #93772, opened 2023-10-20
JIT: Add explicit block fallthrough successor

amanasifkhalid:
Next step for #93020. This change shouldn’t introduce any behavioral changes, as we don’t allow the fall-through successor to diverge from `bbNext` yet (that will be in a follow-up PR). Once we allow these to diverge (plus disallow BBJ_NONE until the block layout is set), we’ll theoretically no longer have to maintain an invariant of contiguous fall-through successors during block reordering, opening up more ordering possibilities.

Note that while BBJ_CALLFINALLY blocks can fall through, we don’t plan to allow their fall-through successors to diverge from `bbNext` yet. We will have to determine if there is any benefit in splitting up call-always pairs first.

### DIFF
--- a/src/coreclr/jit/block.cpp
+++ b/src/coreclr/jit/block.cpp
@@ -1278,7 +1278,7 @@ BasicBlock* BasicBlock::GetSucc(unsigned i, Compiler* comp)
             return bbJumpDest;
 
         case BBJ_NONE:
-            return bbNext;
+            return GetFallThroughSucc();
 
         case BBJ_COND:
             if (i == 0)
--- a/src/coreclr/jit/block.h
+++ b/src/coreclr/jit/block.h
@@ -681,9 +681,9 @@ public:
     {
         // TODO: Once we allow bbFallThroughSucc to diverge from bbNext, ensure we only set bbFallThroughSucc
         // to something other than bbNext when this block is a BBJ_COND.
-        // BBJ_CALLFINALLY can fall through too, but we don't allow its fallthrough successor to diverge from bbNext
-        // because we aren't interested in splitting up call-always pairs
-        assert(KindIs(BBJ_COND, BBJ_NONE));
+        // For BBJ_CALLFINALLY blocks part of call-always pairs, bbFallThroughSucc points to the BBJ_ALWAYS block.
+        // For now, call-always pairs are always contiguous, so bbFallThroughSucc cannot diverge from bbNext.
+        assert(KindIs(BBJ_COND));
         bbFallThroughSucc = fallThroughSucc;
         // For now, bbFallThroughSucc cannot diverge from bbNext
         assert(FallsIntoNext());
--- a/src/coreclr/jit/block.h
+++ b/src/coreclr/jit/block.h
@@ -537,6 +537,8 @@ private:
         BBehfDesc*  bbJumpEhf;  // BBJ_EHFINALLYRET descriptor
     };
 
+    BasicBlock* bbFallThroughSucc;
+
 public:
 #ifdef DEBUG
     // When creating a block with a jump, we require its jump kind and target be initialized simultaneously.
@@ -581,7 +583,8 @@ public:
 
     void SetNext(BasicBlock* next)
     {
-        bbNext = next;
+        bbNext            = next;
+        bbFallThroughSucc = bbNext;
         if (next)
         {
             next->bbPrev = this;
@@ -666,6 +669,38 @@ public:
         SetJumpKindAndTarget(jumpKind, jumpDest DEBUG_ARG(compiler));
     }
 
+    BasicBlock* GetFallThroughSucc() const
+    {
+        assert(bbFallsThrough());
+        // For now, bbFallThroughSucc cannot diverge from bbNext
+        assert(FallsIntoNext());
+        return bbFallThroughSucc;
+    }
+
+    void SetFallThroughSucc(BasicBlock* fallThroughSucc)
+    {
+        // TODO: Once we allow bbFallThroughSucc to diverge from bbNext, ensure we only set bbFallThroughSucc
+        // to something other than bbNext when this block is a BBJ_COND.
+        // BBJ_CALLFINALLY can fall through too, but we don't allow its fallthrough successor to diverge from bbNext
+        // because we aren't interested in splitting up call-always pairs
+        assert(KindIs(BBJ_COND, BBJ_NONE));
+        bbFallThroughSucc = fallThroughSucc;
+        // For now, bbFallThroughSucc cannot diverge from bbNext
+        assert(FallsIntoNext());
+    }
+
+    bool FallsInto(const BasicBlock* dest) const
+    {
+        assert(bbFallsThrough());
+        return (bbFallThroughSucc == dest);
+    }
+
+    bool FallsIntoNext() const
+    {
+        assert(bbFallsThrough());
+        return (bbFallThroughSucc == bbNext);
+    }
+
     bool HasJump() const
     {
         return (bbJumpDest != nullptr);
@@ -681,7 +716,7 @@ public:
     bool JumpsToNext() const
     {
         assert(HasJump());
-        return (bbJumpDest == bbNext);
+        return (bbJumpDest == bbFallThroughSucc);
     }
 
     BBswtDesc* GetJumpSwt() const
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -5297,7 +5297,7 @@ public:
             {
                 // Scenario where next block and conditional block, both point to the same block.
                 // In such case, intersect the assertions present on both the out edges of predBlock.
-                assert(predBlock->NextIs(block));
+                assert(predBlock->FallsInto(block));
                 BitVecOps::IntersectionD(apTraits, pAssertionOut, predBlock->bbAssertionOut);
 
                 if (VerboseDataflow())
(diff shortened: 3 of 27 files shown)


amanasifkhalid:
cc @dotnet/jit-contrib, @AndyAyersMS PTAL. superpmi-replay failure is #93527

src/coreclr/jit/block.h
```diff
@@ -666,6 +669,38 @@ struct BasicBlock : private LIR::Range
...
+        assert(FallsIntoNext());
+        return bbFallThroughSucc;
+    }
+
+    void SetFallThroughSucc(BasicBlock* fallThroughSucc)
+    {
+        // TODO: Once we allow bbFallThroughSucc to diverge from bbNext, ensure we only set bbFallThroughSucc
+        // to something other than bbNext when this block is a BBJ_COND.
+        // BBJ_CALLFINALLY can fall through too, but we don't allow its fallthrough successor to diverge from bbNext
+        // because we aren't interested in splitting up call-always pairs
```
SingleAccretion, line 685: This is a confusing comment. `BBJ_CALLFINALLY` never fall through, as far the flowgraph is concerned - they always "jump" to their corresponding finally handler. The handler then "jumps back" to the `BBJ_ALWAYS` part of the pair via `BBJ_EHFINALLYRET`.

(Physically this can be implemented as a call, so it looks a bit like `BBJ_CALLFINALLY` falls through to `BBJ_ALWAYS`, but that's a very low-level view)

src/coreclr/jit/block.h
```diff
@@ -666,6 +669,38 @@ struct BasicBlock : private LIR::Range
         SetJumpKindAndTarget(jumpKind, jumpDest DEBUG_ARG(compiler));
     }
 
+    BasicBlock* GetFallThroughSucc() const
```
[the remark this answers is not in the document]
amanasifkhalid: Yes, in the case of `BBJ_COND` blocks, this would be the destination of the false branch. I wanted this name to capture the idea that, in the case of `BBJ_COND` blocks, `bbFallThroughSucc` will be taken if `bbJumpDest` isn't, and that this fall-through successor isn't necessarily the next block (which means the control flow doesn't actually "fall through"). Do you have any suggestions for the name @SingleAccretion? I don't want it to sound specific to `BBJ_COND` since we might decide to allow call-finally pairs to be non-contiguous later, thus making `bbFallThroughSucc` relevant.

src/coreclr/jit/block.h
amanasifkhalid: I used the "fall through" language there based on the fact that `BasicBlock::bbFallsThrough` returns true for `BBJ_CALLFINALLY` blocks part of `BBJ_CALLFINALLY/BBJ_ALWAYS` pairs, though I see how describing these blocks as falling through isn't wholly accurate. I can clarify that comment; something like "For `BBJ_CALLFINALLY` blocks part of call-always pairs, `bbFallThroughSucc` points to the `BBJ_ALWAYS` block. For now, call-always pairs are always contiguous, so `bbFallThroughSucc` cannot diverge from `bbNext`."

amanasifkhalid:
[No asmdiffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=445685&view=ms.vss-build-web.run-extensions-tab), but some TP diffs from having to set `bbFallThroughSucc` alongside `bbNext`.

src/coreclr/jit/block.h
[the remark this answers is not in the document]
SingleAccretion: I would suggest `GetNormalJumpDest()` for symmetry with `GetJumpDest()`. It matches the expected codegen for `BBJ_COND`:
```
jcc <cond jump dest>
jmp <normal jump dest>
```

I wouldn't worry too much about the callfinally cases. Callfinally cases are inherently complex, one has to look up how they work specifically. Even if you make the `BBJ_ALWAYS` part of the pair non-contiguous with the callfinally, you'll still not have it as a "successor" in the flowgraph sense.

amanasifkhalid:
FYI: looks like `[skip ci]` in the commit message doesn't skip CI in dotnet/runtime...

src/coreclr/jit/block.h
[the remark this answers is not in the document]
amanasifkhalid: I see, thanks for the suggestion! I'll change it.

src/coreclr/jit/block.cpp
```diff
@@ -1278,7 +1278,7 @@ BasicBlock* BasicBlock::GetSucc(unsigned i, Compiler* comp)
             return bbJumpDest;
 
         case BBJ_NONE:
-            return bbNext;
+            return GetFallThroughSucc();
```
SingleAccretion, line 1281: Here and in a number of other places in this change, the pattern seems to be to change: `BBJ_NONE -> bbNext / BBJ_COND -> bbNext+bbJumpDest` to `BBJ_NONE -> GetFallThroughSucc() / BBJ_COND -> GetFallThroughSucc()+bbJumpDest`. The `BBJ_COND` part makes sense, but the `BBJ_NONE` changes confuse me.

If `GetFallThroughSucc()` is the normal successor of `BBJ_COND`, why touch `BBJ_NONE`?

Side note: I think you will also want to update `VisitAllSuccs/VisitRegularSuccs`.

amanasifkhalid: I'm not sure what @AndyAyersMS's plan is here, but a "lazy" approach to reordering the block layout might be moving a block without considering it might be a `BBJ_NONE` and its successor is its old `bbNext`, and just converting the `BBJ_NONE` to `BBJ_ALWAYS` after the reorder. If we do something like this, I think it would be nice to always use `GetFallThroughSucc` for `BBJ_NONE` blocks so we can assert the fall-through successor matches `bbNext`; if we mess something up during block reordering, hopefully this will help us catch it.

> Side note: I think you will also want to update `VisitAllSuccs/VisitRegularSuccs`.
Thanks for catching that. I'll fix those.

SingleAccretion: I feel there is some overloading of meanings for "the fall-through successor".

The change at large is needed because in the refactored flowgraph, the "false" successor of `BBJ_COND` (and possibly the `BBJ_ALWAYS` part of callfinally pairs) will be distinct from `bbNext`. So, overall, the imported blocks will change as follows:
1) `BBJ_NONE -> bbNext` -> `BBJ_ALWAYS -> bbJumpDest` (can be reordered arbitrarily later on).
2) `BBJ_COND -> bbNext, bbJumpDest` -> `BBJ_COND -> bbTheNewFieldBeingAdded, bbJumpDest` (same).

In other words, "fall through" will not exist as an explicit flowgraph concept like it does today (perhaps until after a very late layout stage, where `BBJ_NONE` may appear as an optional optimization).

In that light, I would expect changes to basically replace `case BBJ_COND: if (<false branch>) return bbNext;` with `case BBJ_COND: if (<false branch>) return bbTheNewFieldBeingAdded;`, and indeed they do in some places like switch recognition, but not in others.

AndyAyersMS: My initial plan was to disallow `BBJ_NONE` until we'd set block layout. I think we'll have to chip away at this as we have a lot of code that needs to work both before and after we do layout.

So a plausible starting point is to allow `BBJ_NONE` but treat it somewhat like `BBJ_ALWAYS`, meaning it has to explicitly name its flow successor, and find some way to start incrementally removing the early `BBJ_NONE` cases.

SingleAccretion: I wonder if simply not producing `BBJ_NONE` (anywhere), instead producing `BBJ_ALWAYS`, and adding a small peephole to codegen to not emit "branch to next block" would be a faster/easier way to go about this.

amanasifkhalid: > I wonder if simply not producing BBJ_NONE (anywhere), instead producing BBJ_ALWAYS, and adding a small peephole to codegen to not emit "branch to next block" would be a faster/easier way to go about this.

With this approach, maybe we can get rid of `bbFallThroughSucc` and instead add a new member specific to `BBJ_COND` for its false branch target.

hoyosjs:
> FYI: looks like `[skip ci]` in the commit message doesn't skip CI in dotnet/runtime... @hoyosjs

It's documented in https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/azure-repos-git?view=azure-devops&tabs=yaml#skipping-ci-for-individual-pushes and there's so much "caveat" that seems broken by design for PR. In particular:

> The pipelines specified by the target branch's build validation policy will run on the merge commit (the merged code between the source and target branches of the pull request), regardless if there exist pushed commits whose messages or descriptions contain [skip ci] (or any of its variants).

Looks like it has a long history of how broken it is for GH / AzDO PRs and it's claimed to be by design. :( https://github.com/Microsoft/azure-pipelines-agent/issues/858 https://github.com/microsoft/azure-pipelines-agent/issues/2944. If people think there's enough value, we can easily add something to the evaluate paths step to skip with the string (@agocke and @jkoritzinsky, thoughts on such check for PRs).

amanasifkhalid:
> Looks like it has a long history of how broken it is for GH / AzDO PRs and it's claimed to be by design.

I see, thank you for digging into this!

> If people think there's enough value, we can easily add something to the evaluate paths step to skip with the string

+1, I think that would be a great quality-of-life improvement, particularly for trivial style/comment updates to PRs that touch source code.

amanasifkhalid:
@AndyAyersMS should we put this change on hold for now, and I can try removing `BBJ_NONE` altogether? Jakob seems interested in that approach, too.

AndyAyersMS:
> I can try removing `BBJ_NONE` altogether?

Sure, give it a try and see what happens.

amanasifkhalid:
Closing in favor of #95773.

PR #93773, opened 2023-10-20
Add common parent for frozen objects

MichalStrehovsky:
Frozen strings and frozen arbitrary objects were separate classes but giving them a common ancestors makes handling these easier.

Cc @dotnet/ilc-contrib


jkotas (approved):
Thanks!

PR #93790, opened 2023-10-20
Add runtime base address to Native AOT crash info JSON

mikem8361:
# Customer Impact

The Native AOT Watson and SOS support needs to the module's base address containing the runtime for adequate performance. Add the module base address in the crash info as "runtime_base". Changed the "runtime" to the clearer "runtime_version" property.

# Testing

Verified the address in local testing. Modified the !crashinfo command in SOS to display this new base address.

# Risk

Very low. Native AOT unhandled exception crash info.


jkotas (approved):
Thanks!

PR #93805, opened 2023-10-20
[API Implementation]: CngProperty constructor with ReadOnlySpan<byte> overload

### LINKED ISSUE #69061: [API Proposal]: CngProperty constructor with ReadOnlySpan<byte> overload
### Background and motivation

There is an internal constructor on `CngProperty` that accepts the `value` as a `ReadOnlySpan<byte>`.

https://github.com/dotnet/runtime/blob/be20c44826bc55d201c406f4ac007d24ed7a9137/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/CngProperty.cs#L30

I propose making this public so that callers that have data as a `ReadOnlySpan<byte>` don't need to convert it to a `byte[]` first, and then get copied, again, since it keeps a defensive copy.

### API Proposal

```diff
namespace System.Security.Cryptography {
    public partial struct CngProperty {
-       internal CngProperty(string name, ReadOnlySpan<byte> value, CngPropertyOptions options);
+       public CngProperty(string name, ReadOnlySpan<byte> value, CngPropertyOptions options);
    }
}
```

### API Usage

```csharp
ReadOnlySpan<byte> value = "I need a nap"u8;
CngProperty test = new(
    "CustomLabel",
    value,
    CngPropertyOptions. CustomProperty);
```

### Alternative Designs

_No response_

### Risks

_No response_

deeprobin:
Fixes #69061.

## Implemented API

```cs
namespace System.Security.Cryptography {
    public partial struct CngProperty {
       public CngProperty(string name, ReadOnlySpan<byte> value, CngPropertyOptions options);
    }
}
```


src/libraries/System.Security.Cryptography/tests/CngPropertyTests.cs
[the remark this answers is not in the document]
deeprobin: Tests moved, see last commit

src/libraries/System.Security.Cryptography.Cng/tests/PropertyTests.cs
```diff
@@ -160,5 +160,33 @@ public static void NullValueRoundtrip()
             CngProperty property = new CngProperty("banana", null, CngPropertyOptions.CustomProperty);
             Assert.Null(property.GetValue());
         }
+
+        [Fact]
+        public void TestConstructorSpan()
```
vcsjones, line 165:
```suggestion
        public static void TestConstructorSpan()
```

src/libraries/System.Security.Cryptography.Cng/tests/PropertyTests.cs
```diff
@@ -160,5 +160,33 @@ public static void NullValueRoundtrip()
...
+            ReadOnlySpan<byte> value = new byte[12];
+            value[5] = 1;
+            value[6] = 2;
+            value[7] = 3;
+
+            _ = new CngProperty(name, value, CngPropertyOptions.CustomProperty);
+        }
+
+        [Fact]
+        public void TestConstructorSpan_NameNull()
```
vcsjones, line 177:
```suggestion
        public static void TestConstructorSpan_NameNull()
```

src/libraries/System.Security.Cryptography.Cng/tests/PropertyTests.cs
```diff
@@ -160,5 +160,33 @@ public static void NullValueRoundtrip()
...
         }
+
+        [Fact]
+        public static void TestConstructorSpan()
+        {
+            var name = "dotnet-test";
+            ReadOnlySpan<byte> value = new byte[12];
+            value[5] = 1;
+            value[6] = 2;
+            value[7] = 3;
```
vcsjones, line 171: This is a `ReadOnlySpan` so it can’t be written to.
```suggestion
            ReadOnlySpan<byte> value = [1, 2, 3, 4, 5];
```

src/libraries/System.Security.Cryptography.Cng/tests/PropertyTests.cs
```diff
@@ -160,5 +160,33 @@ public static void NullValueRoundtrip()
...
+
+            _ = new CngProperty(name, value, CngPropertyOptions.CustomProperty);
```
vcsjones, line 173: Consider calling `GetValue` on the new property and asserting its value.
```suggestion
            CngProperty property = new CngProperty(name, value, CngPropertyOptions.CustomProperty);
            AssertExtensions.SequenceEqual(value, property.GetValue());
```

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/CngProperty.cs
```diff
@@ -27,7 +27,14 @@ public CngProperty(string name, byte[]? value, CngPropertyOptions options)
...
 
-        internal CngProperty(string name, ReadOnlySpan<byte> value, CngPropertyOptions options)
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CngProperty"/> struct.
+        /// </summary>
+        /// <param name="name">The property name to initialize.</param>
+        /// <param name="value">The property value to initialize.</param>
+        /// <param name="options">A bitwise combination of the enumeration values that specify how the property is stored.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="name"/> is <see langref="null"/>.</exception>
+        public CngProperty(string name, ReadOnlySpan<byte> value, CngPropertyOptions options)
```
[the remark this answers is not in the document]
bartonjs: What I care about is that there's a test that says a CngProperty built on `(byte[])null` is the same as one built on `(ReadOnlySpan<byte>)(byte[])null`, and similarly for `Array.Empty<byte[]>()`.  If that equivalence is met, I don't particularly care what this line of code looks like :smile:.

If a CngProperty based on `null` is basically useless (CNG will never return one, and CNG will never accept one... it can only be used within application code for some random purpose) then my position probably changes to "the null span and the empty span should be the same".  But I want to start with the tests.  Particularly if null and empty are both valid inputs to CNG but mean different things (e.g. if null means "erase this property")

deeprobin:
Idea why this throws not on Windows, but on Linux & OSX?

.packages\microsoft.dotnet.helix.sdk\9.0.0-beta.23518.2\tools\azure-pipelines\AzurePipelines.MultiQueue.targets(42,5): error : (NETCORE_ENGINEERING_TELEMETRY=Test) Test System.Security.Cryptography.Cng.Tests.PropertyTests.TestConstructorSpan_NameNull has failed. Check the Test tab or this console log: https://helix.dot.net/api/2019-06-17/jobs/0487afd8-1be4-4ba4-999d-d1a551b33f7c/workitems/System.Security.Cryptography.Cng.Tests/console

src/libraries/System.Security.Cryptography.Cng/tests/PropertyTests.cs
```diff
@@ -160,5 +160,31 @@ public static void NullValueRoundtrip()
...
+            var value = new byte[12];
+            value[5] = 1;
+            value[6] = 2;
+            value[7] = 3;
+
+            Assert.Throws<ArgumentNullException>(() =>
+            {
+                ReadOnlySpan<byte> span = value;
+                _ = new CngProperty(name, span, CngPropertyOptions.CustomProperty);
+            });
```
bartonjs, line 187: This assertion should be validating the name of the parameter that triggered the ArgumentNullException.  IIRC, something like

```suggestion
            Assert.Throws<ArgumentNullException>(() =>
            {
                ReadOnlySpan<byte> span = value;
                _ = new CngProperty(name, span, CngPropertyOptions.CustomProperty);
            },
            "name");
```

But it's not clear why this line is expected to throw anyways.  The test says "NameNull", but name is `"dotnet-test"`, so this looks like it should be not throwing.

src/libraries/System.Security.Cryptography.Cng/tests/PropertyTests.cs
```diff
@@ -160,5 +160,31 @@ public static void NullValueRoundtrip()
...
+            ReadOnlySpan<byte> value = [1, 2, 3, 4, 5];
+
+            CngProperty property = new CngProperty(name, value, CngPropertyOptions.CustomProperty);
+            AssertExtensions.SequenceEqual(value, property.GetValue());
+        }
+
+        [Fact]
+        public static void TestConstructorSpan_NameNull()
+        {
+            var name = "dotnet-test";
```
bartonjs, line 177: The type name of a variable declaration has to always appear.  While string literals used to be obvious-typed (but still failed the var rule), with UTF-8 string literals it now requires reading the whole line and paying attention to see if something is string or ReadOnlySpan-byte... so now string literals are super-not-approved-for-var.

```suggestion
            string name = "dotnet-test";
```

src/libraries/System.Security.Cryptography.Cng/tests/PropertyTests.cs
deeprobin: oh yes... my mistake

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/CngProperty.cs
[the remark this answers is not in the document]
bartonjs: I still don't see tests around this behavior.

There should be a test showing that setting a property to `(byte[])null` (via the byte[] ctor) and `default(ReadOnlySpan<byte>)` (via the new ctor) do the same thing.

There should also be a test (same one? different one? surprise me) showing that `Array.Empty<byte>()` (old ctor) and `Array.Empty<byte>().AsSpan()` (new ctor) do the same thing.  They might even do the same thing as `null`.

What sort of things are the same?

* Does HasProperty give the same answer for the same value in both the old and new ctor when assigning a property that the key doesn't already have?
* Does HasProperty give the same answer for the same value in both the old and new ctor when overwriting an existing (non-empty) value?
* Does GetProperty give identical answers for the same value in both the old and new ctor?
* Anything else that feels relevant, though their might not actually be "anything else" since there's not a lot of API surface here.

PR #93809, opened 2023-10-21
Abstract out LocalBuilder, emit Locals with new ILGenerator

### LINKED ISSUE #93497: [API Proposal]: Add protected factory method for creating Label in ILGenerator, abstract out LocalBuilder

buyaa-n:
- Abstract out LocalBuilder 
- Added implementation for `DeclareLocal(Type localType, bool pinned)`, `(OpCode opcode, LocalBuilder local)`, and corresponding tests

Fixes https://github.com/dotnet/runtime/issues/93497


src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeLocalBuilder.cs
```diff
@@ -24,21 +24,11 @@ internal LocalBuilder(int localIndex, Type localType, MethodInfo methodBuilder,
...
-        {
-            return m_methodBuilder;
-        }
-        #endregion
-
         #region LocalVariableInfo Override
         public override bool IsPinned => m_isPinned;
         public override Type LocalType => m_localType;
         public override int LocalIndex => m_localIndex;
+        public override MethodInfo Method => m_methodBuilder;
```
AaronRobinsonMSFT, line 31: Unsure of the policy for `region` tags, but this likely shouldn't be in the `LocalVariableInfo` section. Should we create one for `LocalBuilder` or remove them entirely?

src/mono/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeLocalBuilder.Mono.cs
```diff
@@ -99,5 +99,8 @@ internal int EndOffset
         {
             get { return endOffset; }
         }
+
+        // Mono uses ILGenerator instance instead of MethodBuilder, could not get reference to the MethodBuilder the local belongs to.
+        public override MethodInfo Method => throw new NotImplementedException();
```
jkotas, line 104: This looks suspect.

Do we really need to expose `Method` as a public abstract property? Would it be enough to expose it as an internal method on LocalBuilderImpl?

buyaa-n: I thought it will be OK as mono is not using it anywhere ...
> Would it be enough to expose it as an internal method on LocalBuilderImpl?

Yes, it can be an internal method, though would need to cast from `LocalBuilder` into `LocalBuilderImpl`, thought adding a public API could be useful if someone build a new `ILGenerator` implementation ...

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -192,7 +209,31 @@ public override void Emit(OpCode opcode, string str)
...
+        {
+            ArgumentNullException.ThrowIfNull(local);
+
+            if (local.Method != _methodBuilder)
+            {
+                throw new ArgumentException(SR.Argument_UnmatchedMethodForLocal, nameof(local));
+            }
+
+            int tempVal = local.LocalIndex;
+            if (opcode.Name!.StartsWith("ldloca"))
```
AaronRobinsonMSFT, line 222: This statement has a `!`, but the others don't. Is this intended? If so can we add a comment as to the why here? I assume to keep the compiler quiet, but it is a weird thing to have in one but no the others.

src/mono/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeLocalBuilder.Mono.cs
jkotas: This is a new public API. It needs to work on all runtimes.

If somebody build a new ILGenerator, I think they can keep track of the method just like we do in our own ILGeneratorImpl.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/LocalBuilderImpl.cs
```diff
@@ -0,0 +1,32 @@
...
+            _localType = type;
+            _method = method;
+        }
+        #endregion
+
+        #region LocalVariableInfo Override
+        public override bool IsPinned => _isPinned;
+        public override Type LocalType => _localType;
+        public override int LocalIndex => _localIndex;
+        public override MethodInfo Method => _method;
```
AaronRobinsonMSFT, line 29: Should be from `LocalBuilder`.

src/mono/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeLocalBuilder.Mono.cs
AaronRobinsonMSFT, line 104: We should file an issue for this.

/cc @lambdageek

AaronRobinsonMSFT: @lambdageek for suggestions here.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
buyaa-n: Yes, it is for handling nullability warning, the null forgiving operator `!` not only shuts down the compiler, but it also tells the nullability analyzer no need to warn for `opcode.Name` further, this is by design, no need a comment.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/LocalBuilderImpl.cs
buyaa-n: Sorry did not understood, the abstract `LocalBuilder` doesn't have this field.

AaronRobinsonMSFT: It is defined in https://github.com/dotnet/runtime/pull/93809/files#diff-b17fa465e3ae3bd103760932136e6317b718ce3d30f551b3976e7ee3c9070d45?

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -38,7 +42,20 @@ internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
...
-        public override LocalBuilder DeclareLocal(Type localType, bool pinned) => throw new NotImplementedException();
+        public override LocalBuilder DeclareLocal(Type localType, bool pinned)
+        {
+            if (_methodBuilder is not MethodBuilderImpl methodBuilder)
+                throw new NotSupportedException();
+
+            ArgumentNullException.ThrowIfNull(localType);
+
+            LocalBuilder local = new LocalBuilderImpl(_localCount++, localType, methodBuilder, pinned);
+            _locals ??= new();
```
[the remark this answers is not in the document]
buyaa-n: I thought most methods will not have locals, but of course don't for sure, I can update

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
AaronRobinsonMSFT: Let me propose a different approach to make this clearer. Create a local that performs the `!` and then use the local throughout.

```csharp
string name = opcode.Name!;
if (name.StartsWith("ldloca"))
...
```

src/mono/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeLocalBuilder.Mono.cs
lambdageek: @buyaa-n

1. You probably know more at this point, but it doesnt' seem like it would be a problem to add a `Method` member to mono's `RuntimeILGenerator` and add an extra argument to all the `RuntimeILGenerator` constructors.

2. Actually I have a question about this API: Why is it a `MethodInfo` and not a `MethodBase` ? Constructors can have locals too.

src/libraries/System.Reflection.Emit.ILGeneration/ref/System.Reflection.Emit.ILGeneration.cs
```diff
@@ -65,12 +65,13 @@ public abstract class ILGenerator
...
     }
-    public sealed partial class LocalBuilder : System.Reflection.LocalVariableInfo
+    public abstract class LocalBuilder : System.Reflection.LocalVariableInfo
     {
-        internal LocalBuilder() { }
+        protected LocalBuilder() { }
         public override bool IsPinned { get { throw null; } }
         public override int LocalIndex { get { throw null; } }
         public override System.Type LocalType { get { throw null; } }
+        public abstract MethodInfo Method { get; }
```
lambdageek, line 74: @jkotas @AaronRobinsonMSFT @buyaa-n What if the local is in a constructor?  I would expect `MethodBase` here, not `MethodInfo`

AaronRobinsonMSFT: That is a good point. I was also interested in why this sort of property wasn't on the base class, `LocalVariableInfo`, but didn't raise it since API review had already occurred. I think the constructor case though does seem to warrant further consideration.

@buyaa-n or @steveharter Was the constructor case discussed in the API review? I didn't attend so perhaps I am missing context.

jkotas: I think we should drop the property from public surface and keep it as an internal detail in the managed RefEmit.

buyaa-n: > @buyaa-n or @steveharter Was the constructor case discussed in the API review? 

No, existing runtime implementation creates a MethodBuilder for constructors, i.e. they are kept as a method that is why it works, not sure how mono handle constructors

AaronRobinsonMSFT: I see. The issue here then becomes `ConstructorBuilder.GetILGenerator()` and the path to this API is now broken because one could not get `ConstructorInfo` when starting from a `ConstructorBuilder`.

> I think we should drop the property from public surface and keep it as an internal detail in the managed RefEmit.

I'm starting to agree.

jkotas: The method builder created for constructors is internal implementation detail. I do not think that this method builder is meant to leak out through the public surface. It should always show up wrapper by `ConstructorBuilder`.

AaronRobinsonMSFT: @jkotas I agree with your statement. In order to be concrete, are you using that statement as a way to further demonstrate that `public abstract MethodInfo Method { get; }` should not be part of the public API and instead an internal implementation detail?

jkotas: Yep. The current API shape is not right and the easiest way to fix that is to avoid exposing the API in the public surface.

buyaa-n: Added a comment for [removing the property](https://github.com/dotnet/runtime/issues/93497#issuecomment-1781514695) in the approved API, will remove it form this PR

src/mono/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeILGenerator.Mono.cs
```diff
@@ -619,10 +619,11 @@ public override void Emit(OpCode opcode, Label[] labels)
         public override void Emit(OpCode opcode, LocalBuilder local)
         {
             ArgumentNullException.ThrowIfNull(local);
-            if (local.ilgen != this)
+            RuntimeLocalBuilder localBuilder = (RuntimeLocalBuilder)local;
+            if (localBuilder.ilgen != this)
```
jkotas, line 623:
```suggestion
            if (local is not RuntimeLocalBuilder localBuilder || localBuilder.ilgen != this)
```
This will avoid invalid cast exception when somebody passes in LocalBuilder that is not one of ours.

jkotas: (Dtto for other places that cast to RuntimeLocalBuilder.)

src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeILGenerator.cs
```diff
@@ -819,8 +819,8 @@ public override void Emit(OpCode opcode, LocalBuilder local)
             ArgumentNullException.ThrowIfNull(local);
 
             // Puts the opcode onto the IL stream followed by the information for local variable local.
-            int tempVal = local.GetLocalIndex();
-            if (local.GetMethodBuilder() != m_methodBuilder)
+            int tempVal = local.LocalIndex;
+            if (((RuntimeLocalBuilder)local).GetMethodBuilder() != m_methodBuilder)
```
buyaa-n, line 823:
```suggestion
            if (local is RuntimeLocalBuilder localBuilder && localBuilder.GetMethodBuilder() != m_methodBuilder)
```

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -228,7 +241,34 @@ public override void Emit(OpCode opcode, Label[] labels)
             }
         }
 
-        public override void Emit(OpCode opcode, LocalBuilder local) => throw new NotImplementedException();
+        public override void Emit(OpCode opcode, LocalBuilder local)
+        {
+            ArgumentNullException.ThrowIfNull(local);
+
+            if (((LocalBuilderImpl)local).GetMethodBuilder() != _methodBuilder)
```
buyaa-n, line 248:
```suggestion
            if (local is LocalBuilderImpl localBuilder && localBuilder.GetMethodBuilder() != _methodBuilder)
```

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -228,7 +241,34 @@ public override void Emit(OpCode opcode, Label[] labels)
             }
         }
 
-        public override void Emit(OpCode opcode, LocalBuilder local) => throw new NotImplementedException();
+        public override void Emit(OpCode opcode, LocalBuilder local)
+        {
+            ArgumentNullException.ThrowIfNull(local);
+
+            if (local is LocalBuilderImpl localBuilder && localBuilder.GetMethodBuilder() != _methodBuilder)
```
buyaa-n, line 248:
```suggestion
            if (local is not LocalBuilderImpl localBuilder || localBuilder.GetMethodBuilder() != _methodBuilder)
```

src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeILGenerator.cs
```diff
@@ -819,8 +819,8 @@ public override void Emit(OpCode opcode, LocalBuilder local)
             ArgumentNullException.ThrowIfNull(local);
 
             // Puts the opcode onto the IL stream followed by the information for local variable local.
-            int tempVal = local.GetLocalIndex();
-            if (local.GetMethodBuilder() != m_methodBuilder)
+            int tempVal = local.LocalIndex;
+            if (local is RuntimeLocalBuilder localBuilder && localBuilder.GetMethodBuilder() != m_methodBuilder)
```
buyaa-n, line 823:
```suggestion
            if (local is not RuntimeLocalBuilder localBuilder || localBuilder.GetMethodBuilder() != m_methodBuilder)
```

jkotas (approved):
LGTM. Thank you!

AaronRobinsonMSFT (approved):

buyaa-n:
Thank you for the review! All failures are known.

PR #93816, opened 2023-10-21
JIT: stop creating gc poll blocks in morph

AndyAyersMS:
Instead, mark the block as needing a GC poll and add one later in the normal GC poll insertion phase. This only impacts x86.

This change also stops creating inline GC polls for these cases in unoptimized code.

### DIFF
--- a/src/coreclr/jit/block.cpp
+++ b/src/coreclr/jit/block.cpp
@@ -521,6 +521,10 @@ void BasicBlock::dspFlags()
     {
         printf("mdarr ");
     }
+    if (bbFlags & BBF_NEEDS_GCPOLL)
+    {
+        printf("gcpoll ");
+    }
 }
 
 /*****************************************************************************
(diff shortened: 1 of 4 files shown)


AndyAyersMS:
@jakobbotsch PTAL
cc @dotnet/jit-contrib

jakobbotsch (approved):

EgorBo:
```
Assert failure(PID 29 [0x0000001d], Thread: 217 [0x00d9]): Assertion failed 'doesMethodHaveRecursiveTailcall()' in 'System.Reflection.SignatureType:GetHashCode():int:this' during 'Morph - Global' (IL size 7; hash 0xe30863bc; Instrumented Tier1)

    File: /__w/1/s/src/coreclr/jit/morph.cpp Line: 7459
    Image: /root/helix/work/correlation/dotnet
```
can it be related?

jakobbotsch:
The assert was added in #93764. I've opened #93844 for it.

PR #93817, opened 2023-10-21
Expose BitConverter APIs for Int128 and UInt128

### LINKED ISSUE #80337: [API Proposal]: Add Int128 and UInt128 `BitConverter.GetBytes` support
### Background and motivation

Get the underlying bytes of 128 bit integers for use in binary serialization and others

### API Proposal

EDIT:
Signatures for the methods:
```cs
namespace System;

public static partial class BitConverter
{
    public static byte[] GetBytes(Int128 value);
    public static Int128 ToInt128(ReadOnlySpan<byte> value);
    public static Int128 ToInt128(byte[] value, int startIndex);
    public static bool TryWriteBytes(Span<byte> destination, Int128 value);

    public static byte[] GetBytes(UInt128 value);
    public static UInt128 ToUInt128(ReadOnlySpan<byte> value);
    public static UInt128 ToUInt128(byte[] value, int startIndex);
    public static bool TryWriteBytes(Span<byte> destination, UInt128 value);
}
```

### API Usage

```csharp
public void Example(Int128 value)
{
    var bytes = BitConverter.GetBytes(value);
    var newValue = BitConverter.ToInt128(bytes, 0);
}
```

### Alternative Designs

_No response_

### Risks

_No response_

tannergooding:
This resolves #80337


src/libraries/System.Private.CoreLib/src/System/BitConverter.cs
```diff
@@ -463,6 +519,44 @@ public static long ToInt64(ReadOnlySpan<byte> value)
...
+        /// <exception cref="ArgumentException">
+        /// <paramref name="startIndex"/> is greater than or equal to the length of <paramref name="value"/> minus 15,
+        /// and is less than or equal to the length of <paramref name="value"/> minus 1.
+        /// </exception>
+        /// <exception cref="ArgumentNullException"><paramref name="value"/> is null.</exception>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="startIndex"/> is less than zero or greater than the length of <paramref name="value"/> minus 1.</exception>
+        public static Int128 ToInt128(byte[] value, int startIndex)
+        {
+            if (value == null)
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.value);
```
danmoseley, line 537: ArgumentNullException.ThrowIfNull, etc?

tannergooding: Just following the existing convention in the rest of the file. Updating everything to use `ThrowIfNull` could/should be done in a separate PR so things stay consistent.

src/libraries/System.Private.CoreLib/src/System/BitConverter.cs
```diff
@@ -159,6 +159,33 @@ public static bool TryWriteBytes(Span<byte> destination, long value)
...
+        /// </summary>
+        /// <param name="destination">When this method returns, the bytes representing the converted 128-bit signed integer.</param>
+        /// <param name="value">The 128-bit signed integer to convert.</param>
+        /// <returns><see langword="true"/> if the conversion was successful; <see langword="false"/> otherwise.</returns>
+        public static bool TryWriteBytes(Span<byte> destination, Int128 value)
+        {
+            if (destination.Length < Int128.Size)
+                return false;
+
+            Unsafe.WriteUnaligned(ref MemoryMarshal.GetReference(destination), value);
```
stephentoub, line 185: Why use WriteUnaligned here and Unsafe.As in the array-based overload? Is the array-based one depending on the runtime guaranteeing alignment of arrays?

src/libraries/System.Private.CoreLib/src/System/BitConverter.cs
```diff
@@ -463,6 +519,44 @@ public static long ToInt64(ReadOnlySpan<byte> value)
...
+        /// <paramref name="startIndex"/> is greater than or equal to the length of <paramref name="value"/> minus 15,
+        /// and is less than or equal to the length of <paramref name="value"/> minus 1.
+        /// </exception>
+        /// <exception cref="ArgumentNullException"><paramref name="value"/> is null.</exception>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="startIndex"/> is less than zero or greater than the length of <paramref name="value"/> minus 1.</exception>
+        public static Int128 ToInt128(byte[] value, int startIndex)
+        {
+            if (value == null)
+                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.value);
+            if (unchecked((uint)startIndex) >= unchecked((uint)value.Length))
```
stephentoub, line 538: Why are these unchecked's needed?

src/libraries/System.Private.CoreLib/src/System/BitConverter.cs
tannergooding: Just copy/pasting the existing pattern used by all the other APIs.

It could be that it's relying on runtime alignment guarantees and it could also be that someone didn't consider alignment for arrays. We aren't super consistent about either throughout the BCL.

src/libraries/System.Private.CoreLib/src/System/BitConverter.cs
tannergooding: Copy/paste from the same pattern used everywhere in the file. They shouldn't be needed unless we're enabling checked arithmetic project wide somewhere. It's possible we were doing that at one point.

stephentoub (approved):

src/libraries/System.Private.CoreLib/src/System/BitConverter.cs
```diff
@@ -550,6 +644,46 @@ public static ulong ToUInt64(ReadOnlySpan<byte> value)
             return Unsafe.ReadUnaligned<ulong>(ref MemoryMarshal.GetReference(value));
         }
 
+        /// <summary>
+        /// Returns a 128-bit unsigned integer converted from four bytes at a specified position in a byte array.
```
rhuijben, line 648: 8 bytes for long, so this would be 16 bytes?

src/libraries/System.Private.CoreLib/src/System/BitConverter.cs
```diff
@@ -550,6 +644,46 @@ public static ulong ToUInt64(ReadOnlySpan<byte> value)
...
+        /// </summary>
+        /// <param name="value">An array of bytes.</param>
+        /// <param name="startIndex">The starting position within <paramref name="value"/>.</param>
+        /// <returns>A 128-bit unsigned integer formed by sixteen bytes beginning at <paramref name="startIndex"/>.</returns>
+        /// <exception cref="ArgumentException">
+        /// <paramref name="startIndex"/> is greater than or equal to the length of <paramref name="value"/> minus 15,
+        /// and is less than or equal to the length of <paramref name="value"/> minus 1.
+        /// </exception>
+        /// <exception cref="ArgumentNullException"><paramref name="value"/> is null.</exception>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="startIndex"/> is less than zero or greater than the length of <paramref name="value"/> minus 1.</exception>
```
rhuijben, line 658: Minus 16?

src/libraries/System.Private.CoreLib/src/System/BitConverter.cs
tannergooding: Yes. Thought I had fixed all these already.

It means the `ulong` one is also wrong and says `four` rather than `eight`

PRs are welcome!

PR #93818, opened 2023-10-21
Optimize "X == Y" to "(X ^ Y) == 0 for SIMD

EgorBo:
Matches native compilers now https://godbolt.org/z/afaE18saG

Quick example:
```cs
bool Test(Vector256<byte> v1, Vector256<byte> v2) => v1 == v2;
```
```diff
; Method Test
       vzeroupper 
       vmovups  ymm0, ymmword ptr [rdx]
-      vpcmpeqb k1, ymm0, ymmword ptr [r8]
-      kortestd k1, k1
-      setb     al
+      vpxor    ymm0, ymm0, ymmword ptr [r8]
+      vptest   ymm0, ymm0
+      sete     al
       movzx    rax, al
       vzeroupper 
       ret      
-; Total bytes of code: 28
+; Total bytes of code: 27
```

### DIFF
--- a/src/coreclr/jit/lowerxarch.cpp
+++ b/src/coreclr/jit/lowerxarch.cpp
@@ -1780,10 +1780,24 @@ GenTree* Lowering::LowerHWIntrinsicCmpOp(GenTreeHWIntrinsic* node, genTreeOps cm
     GenTree*     op2    = node->Op(2);
     GenCondition cmpCnd = (cmpOp == GT_EQ) ? GenCondition::EQ : GenCondition::NE;
 
-    if (!varTypeIsFloating(simdBaseType) && (simdSize != 64) && op2->IsVectorZero() &&
+    if (!varTypeIsFloating(simdBaseType) && (simdSize != 64) &&
         comp->compOpportunisticallyDependsOn(InstructionSet_SSE41) &&
         !op1->OperIsHWIntrinsic(NI_AVX512F_ConvertMaskToVector))
     {
+        if (!op2->IsVectorZero())
+        {
+            // Optimize "X == Y" to "(X ^ Y) == 0"
+            GenTree* zeroVec = comp->gtNewZeroConNode(simdType);
+            GenTree* xorVec  = comp->gtNewSimdBinOpNode(GT_XOR, simdType, op1, op2, simdBaseJitType, simdSize);
+            node->Op(1)      = xorVec;
+            node->Op(2)      = zeroVec;
+            BlockRange().InsertBefore(node, xorVec);
+            BlockRange().InsertBefore(node, zeroVec);
+
+            // We'll re-visit the comparison node again
+            return xorVec;
+        }
+
         // On SSE4.1 or higher we can optimize comparisons against zero to
         // just use PTEST. We can't support it for floating-point, however,
         // as it has both +0.0 and -0.0 where +0.0 == -0.0
--- a/src/coreclr/jit/lowerxarch.cpp
+++ b/src/coreclr/jit/lowerxarch.cpp
@@ -1784,9 +1784,11 @@ GenTree* Lowering::LowerHWIntrinsicCmpOp(GenTreeHWIntrinsic* node, genTreeOps cm
         comp->compOpportunisticallyDependsOn(InstructionSet_SSE41) &&
         !op1->OperIsHWIntrinsic(NI_AVX512F_ConvertMaskToVector))
     {
-        if (!op2->IsVectorZero())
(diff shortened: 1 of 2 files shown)


src/coreclr/jit/lowerxarch.cpp
```diff
@@ -1780,10 +1780,24 @@ GenTree* Lowering::LowerHWIntrinsicCmpOp(GenTreeHWIntrinsic* node, genTreeOps cm
...
     {
+        if (!op2->IsVectorZero())
+        {
+            // Optimize "X == Y" to "(X ^ Y) == 0"
+            GenTree* zeroVec = comp->gtNewZeroConNode(simdType);
+            GenTree* xorVec  = comp->gtNewSimdBinOpNode(GT_XOR, simdType, op1, op2, simdBaseJitType, simdSize);
+            node->Op(1)      = xorVec;
+            node->Op(2)      = zeroVec;
+            BlockRange().InsertBefore(node, xorVec);
+            BlockRange().InsertBefore(node, zeroVec);
```
jakobbotsch, line 1795:
```suggestion
            BlockRange().InsertBefore(node, xorVec, zeroVec);
```

tannergooding:
> Quick example:

I'm not sure this is "better". `vptest` is more limited when EVEX is enabled and our register allocator doesn't have the ability to handle that level of nuance today.

For example, `vptest` can't be used for anything that requires the EVEX encoding. This limits LSRA to only using `xmm0-xmm15` (cutting out `xmm16-xmm31`). Likewise, it prevents the ability to use features like embedded broadcast, can't be used with floating-point, or with other types of comparisons that aren't strict equality. I would expect the native compilers take this into account and will sometimes opt to use `kortest` instead when register pressure is high or if the total instruction sequence could benefit from an EVEX only feature.

Given that, I'd probably lean towards it being better to just keep things "as is" here and only do the `xor, ptest` trick for VEX where it is clearly faster.

-------

Also notably, here are the timings for the instructions involved here

| Instruction | Intel | AMD |
| ----- | ----- | ----- |
| kortest | 1 uop, 2 latency | 1 uop, 6 latency |
| vpcmp ( VEX) | 1 uop, 1 latency | 1 uop, 1 latency |
| vpcmp (EVEX) | 1 uop, 3 latency | 1 uop, 3 latency |
| vptest (XMM) | 2 uops, 4 latency | 2 uops, 7 latency |
| vptest (YMM) | 2 uops, 6 latency | 2 uops, 9 latency |
| vpxor | 1 uop, 1 latency | 1 uop, 1 latency |

Which shows that, at least theoretically, `vpcmp + kortest` is a better sequence in many, if not most, cases. That is, it shows either the same latency, but 1 less uop; or faster latency + less uops in most cases. The one exception is handling XMM on AMD, where it is 1 latency higher, but 1 less uop.

EgorBo:
> I'm not sure this is "better"

Once you switched vector equality to use kortest even for XMM/YMM it regressed my benchmarks on AMD, we don't have AVX512 hardware so it was never reflected in our perflab infra via regressions. As you can see from my godbolt link, native compilers also don't use it.

> Also notably, here are the timings for the instructions involved here

I think you forgot to include RThrougput here.

Also, it's smaller in terms of code size, hence, potentially better.

And finally, uiCA stats on these for Tiger Lake (pretty much same for others):

image

image

EgorBo:
Benchmark:

```cs
private static byte[] arr1 = new byte[1024];
private static byte[] arr2 = new byte[1024];

[Benchmark]
public bool VectorEquals()
{
    ref byte a1 = ref arr1[0];
    ref byte a2 = ref arr2[0];

    for (nuint i = 0; i < 1024; i+=16)
    {
        if (Vector128.LoadUnsafe(ref a1, i) != Vector128.LoadUnsafe(ref a2, i))
            return false;
    }
    return true;
}
```

| Method       | Job        | Mean     | Error    | StdDev   | Ratio |
|------------- |----------- |---------:|---------:|---------:|------:|
| VectorEquals | Main       | 20.26 ns | 1.239 ns | 0.068 ns |  1.00 |
| VectorEquals | PR         | 15.24 ns | 0.697 ns | 0.038 ns |  0.75 |

Ryzen 7950X, will find some modern intel to check there but I suspect the same results.

tannergooding:
> Ryzen 7950X, will find some modern intel to check there but I suspect the same results. But the difference is quite noticeable.

Can you also check for YMM?

Likewise, what happens in register heavy code where we end up needing to pick XMM16-XMM31 to avoid spilling?

EgorBo:
Ah diffs are empty if I disable it for non-AVX512 hw (presumably, most of our SPMI collections either don't have AVX512 or it's ignored because of throttling issues) so going to close for now. We need some better coverage fro AVX512 in SPMI

PR #93823, opened 2023-10-22
Proposal to add WasmImportAttribute to control wasm module names

yowl:
This PR proposes a new attribute to be used to control the Wasm import module and function names.  In Wasm when a function is imported it can be specified as coming from a named module using this WAT

```
(import "hellowasi" "reverse" (func $reverse (type 4)))
```

The first use of this would be in the experimental NativeAOT-LLVM backend, but the desire is to have something that works for Mono and NativeAOT-LLVM.  Something that we can use to proceed with Wasm Components and WIT binding via source gen for both Mono and NativeAOT-LLVM.

`DllImportAttribute` has a `Value` and `EntryPoint` which can be used to define the names, but to preserve existing semantics including static linking for `DllImport("*")` and `DirectPInvoke` we cannot simply say that we should always create a Wasm import if the `DllImport` `Value` is not `*`.  The presence of this new `WasmImport` will distinguish these use cases and the attribute has no properties.

Please see https://github.com/dotnet/runtimelab/issues/2414 for how we got here and in particular, https://github.com/dotnet/runtimelab/issues/2414#issuecomment-1772551025.

Other issues and PRs of note on this subject:

https://github.com/dotnet/runtimelab/issues/1390
https://github.com/dotnet/runtimelab/pull/1845
https://github.com/dotnet/runtimelab/pull/2410
https://github.com/dotnet/runtimelab/issues/2383

It has been noted (https://github.com/dotnet/runtimelab/issues/2414#issuecomment-1773024314) that `WasmImport` is a potentially confusing name as the `Import` suffix can infer it's use should be as an alternative to `DllImport` but this is not the case, so some other alternatives that I could think of:

`WasmDynamicLink`
`WasmImportLink`
`WasmModuleLink`
`LlvmWasmImport` (not good for Mono interpreter perhaps)
`WasmNamedModule`

Thanks for reading.

cc @AaronRobinsonMSFT @dotnet/nativeaot-llvm @pavelsavara


AaronRobinsonMSFT:
@yowl Typically we would start with an issue with the API proposal. I will create that quickly and point to this PR. However, this PR will need a few things.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/WasmImportAttribute.cs
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Runtime.InteropServices
```
AaronRobinsonMSFT, line 4: Reference assembly update is needed and API review. See https://github.com/dotnet/runtime/issues/93824.

kg:
Instead of a new attribute, could you add a new `CallingConvention.Wasm` or `CallingConvention.WasmImport` that is used to identify a specific DllImport as one that's a *wasm* import? So instead of
```[DllImport("foo_module", EntryPoint="bar"), WasmImport]```
you would have
```[DllImport("foo_module", EntryPoint="bar", CallingConvention=CallingConvention.Wasm)]```

**Comment by SingleAccretion** [comment]
@kg Calling conventions are orthogonal to symbol resolution. For example, `CallingConvention.WasmImport` would not apply to function pointers and indirect calls.

**Comment by lambdageek** [comment]
@pavelsavara

**Comment by pavelsavara** [comment]
Are there limits to what the wasm import could do ? 

Should we support full set of existing `DllImport` features from day 1 ?
(maybe it's easier to add functionality across releases than remove it later)

What would be behavior of `DllImport` for call conventions ? Should we explicitly limit it to particular call convention? 
(as compared to usual native function with a stack, registers, ... )
In particular we have 2 "stacks" one is the wasm import arguments and the other is the shadow stack of LLVM.
Is there some issue with passing structs on stack ?

Do we expect that the imported functions could `malloc`, `free`, `stackAlloc` or WASI `cabi_realloc` ?
(that probably also assumes we export those functions)

Should we support all string marshaling options ?

Are all unsupported scenarios compile time errors ? 

Is this act of opening new ABI ?

**Comment by SingleAccretion** [comment]
@pavelsavara the attribute will not affect the ABI or calling convention (which both default to C interop, with marshalling, shadow stacks, arbitrary reentrancy, etc).

**Comment by pavelsavara** [comment]
> @pavelsavara the attribute will not affect the ABI or calling convention (which both default to C interop, with marshalling, shadow stacks, arbitrary reentrancy, etc).

C calling conventions are OS/CPU/compiler dependent behavior. From whatever consumes this wasm import it's "undefined" details unless they share the exact same compiler (and maybe compiler flags). Are we hand-waving it ?

I think we have this problem with incompatible behavior of different versions of emscripten between versions of dotnet already. The existing use-case is when we link static library into .wasm file. Hope I remember it correctly. @lewing  ?

**Comment by pavelsavara** [comment]
Thanks!

File: src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/WasmImportLinkageAttribute.cs [hunk 2]
```diff
@@ -0,0 +1,17 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Runtime.InteropServices
+{
+    /// <summary>
+    /// Used to control WASM import module linkage for an associated P/Invoke.
+    /// </summary>
```
AaronRobinsonMSFT, line 8: @SingleAccretion or @yowl Would either of you be interested in providing some addtional documentation for this attribute? I still don't feel like I understand the nuances of the WASM scenarios to speak intelligently about it.

SingleAccretion:
```
Specifies that the P/Invoke marked with this attribute should be linked in as a WASM import.
```
Should be sufficient I think.

Are we allowed links in docs? If so, it could link directly to the spec as well:
```
https://webassembly.github.io/spec/core/syntax/modules.html#imports.
```

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/WasmImportLinkageAttribute.cs
```diff
@@ -0,0 +1,17 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Runtime.InteropServices
+{
+    /// <summary>
+    /// Used to control WASM import module linkage for an associated P/Invoke.
```
AaronRobinsonMSFT, line 7:
```suggestion
    /// Specifies that the P/Invoke marked with this attribute should be linked in as a WASM import.
```

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/WasmImportLinkageAttribute.cs
AaronRobinsonMSFT, line 8:
```suggestion
    /// </summary>
    /// <remarks>
    /// See https://webassembly.github.io/spec/core/syntax/modules.html#imports.
    /// </remarks>
```

AaronRobinsonMSFT (approved):

AaronRobinsonMSFT:
@dotnet/interop-contrib or @dotnet/nativeaot-llvm Can I get another sign off?

jkotas (approved):

PR #93830, opened 2023-10-22
JIT: Clean up GenTreeCallFlags, remove INLINE_DATA define

jakobbotsch:
* Introduce a debug-only `GenTreeCall::gtCallDebugFlags` field that contains flags we only need in debug
* Move `GTF_CALL_M_STRESS_TAILCALL`, `GTF_CALL_M_DEVIRTUALIZED`, `GTF_CALL_M_UNBOXED`, `GTF_CALL_M_GUARDED` into the new debug-only flags
* Delete `GTF_CALL_M_R2R_REL_INDIRECT`; we can get that information from `gtEntryPoint`
* Delete `GTF_CALL_M_EXP_TLS_ACCESS`; we can get this by just checking the helper num
* Gather all the tailcall flags in the same area of flags
* Delete `INLINE_DATA` define; it no longer compiles, and it would need extra work to with the new flags, which seemed unnecessary since it isn't being used
* Make tail call stress more clearly "debug-only" 

Overall this frees up 6 bits in `GenTreeCallFlags`.

No diffs are expected.

### DIFF
--- a/src/coreclr/jit/inline.cpp
+++ b/src/coreclr/jit/inline.cpp
@@ -339,12 +339,12 @@ InlineContext::InlineContext(InlineStrategy* strategy)
     , m_CodeSizeEstimate(0)
     , m_Ordinal(0)
     , m_Success(true)
-    , m_Devirtualized(false)
-    , m_Guarded(false)
-    , m_Unboxed(false)
 #if defined(DEBUG) || defined(INLINE_DATA)
     , m_Policy(nullptr)
     , m_TreeID(0)
+    , m_Devirtualized(false)
+    , m_Guarded(false)
+    , m_Unboxed(false)
 #endif // defined(DEBUG) || defined(INLINE_DATA)
 #ifdef DEBUG
     , m_ILInstsSet(nullptr)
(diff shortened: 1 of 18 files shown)


jakobbotsch:
Failures are known. [No diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=447156&view=ms.vss-build-web.run-extensions-tab) except for some TP ones.

src/coreclr/jit/inline.cpp
```diff
@@ -339,21 +339,21 @@ InlineContext::InlineContext(InlineStrategy* strategy)
...
+    , m_Policy(nullptr)
+    , m_TreeID(0)
     , m_Devirtualized(false)
     , m_Guarded(false)
     , m_Unboxed(false)
-#if defined(DEBUG) || defined(INLINE_DATA)
-    , m_Policy(nullptr)
-    , m_TreeID(0)
-#endif // defined(DEBUG) || defined(INLINE_DATA)
+#endif // defined(DEBUG)
```
AndyAyersMS, line 348: Fuse this with the bit just below?

AndyAyersMS (approved):

Issue #78214, opened 2022-11-11
Arithmetic overflow in code, generated by `GeneratedRegexAttribute`


MaceWindu:
### Description

Note that we use checked builds (which probably not used by runtime tests): `<CheckForOverflowUnderflow>True</CheckForOverflowUnderflow>`

### Reproduction Steps

```cs
using System.Text.RegularExpressions;

internal partial class Program
{
    [GeneratedRegex(@"^\s*GO\s*$\n*", RegexOptions.IgnoreCase | RegexOptions.Multiline | RegexOptions.Compiled)]
    private static partial Regex BuildRegex();

    private static void Main(string[] args)
    {
        BuildRegex().Split("X\r\n XX");
    }
}
```

### Expected behavior

No exceptions

### Actual behavior

```
System.OverflowException
  HResult=0x80131516
  Message=Arithmetic operation resulted in an overflow.
  Source=ConsoleApp41
  StackTrace:
   at System.Text.RegularExpressions.Generated.<RegexGenerator_g>F74B1AE921BCEFE4BA601AA541C7A23B1CA9711EA81E8FE504B5B6446748E035A__BuildRegex_0.RunnerFactory.Runner.TryFindNextPossibleStartingPosition(ReadOnlySpan`1 inputSpan) in RegexGenerator.g.cs:line 95
   at System.Text.RegularExpressions.Generated.<RegexGenerator_g>F74B1AE921BCEFE4BA601AA541C7A23B1CA9711EA81E8FE504B5B6446748E035A__BuildRegex_0.RunnerFactory.Runner.Scan(ReadOnlySpan`1 inputSpan) in RegexGenerator.g.cs:line 68
   at System.Text.RegularExpressions.Regex.ScanInternal(RegexRunnerMode mode, Boolean reuseMatchObject, String input, Int32 beginning, RegexRunner runner, ReadOnlySpan`1 span, Boolean returnNullIfReuseMatchObject)
   at System.Text.RegularExpressions.Regex.RunAllMatchesWithCallback[TState](String inputString, ReadOnlySpan`1 inputSpan, Int32 startat, TState& state, MatchCallback`1 callback, RegexRunnerMode mode, Boolean reuseMatchObject)
   at System.Text.RegularExpressions.Regex.RunAllMatchesWithCallback[TState](String input, Int32 startat, TState& state, MatchCallback`1 callback, RegexRunnerMode mode, Boolean reuseMatchObject)
[...]

PR #93834, opened 2023-10-22
Fold "X >= 0 && X < NN" to "X u< NN"

EgorBo:
Fold e.g. `X >= 0 && X < arr.Length` to `(uint)X < arr.Length`. Not too many diffs because:
1) We already use `uint` hack (I call it a **hack** because IMO it's not how most developers check ranges. Also, it's not "use checked context" [friendly](https://github.com/dotnet/runtime/issues/78214)) everywhere in BCL
2) For arrays we need to land https://github.com/dotnet/runtime/pull/93531 first (and that depends on a fix for a CSE issue)

```cs
static void Test(Span<int> span, int i)
{
    if (i >= 0 && i < span.Length)
        Console.WriteLine("In range!");
}
```
```diff
; Method InitblkPerformanceIssue.Program:Test(System.Span`1[int],int) (FullOpts)
       sub      rsp, 40
-      test     edx, edx
-      jl       SHORT G_M55087_IG04
       cmp      edx, dword ptr [rcx+0x08]
-      jge      SHORT G_M55087_IG04
+      jae      SHORT G_M55087_IG04
       mov      rcx, 0x1AC80209320      ; 'In range!'
       call     [System.Console:WriteLine(System.String)]
G_M55087_IG04:
       nop      
       add      rsp, 40
       ret      
-; Total bytes of code: 35
+; Total bytes of code: 31

```

### DIFF
--- a/src/coreclr/jit/optimizebools.cpp
+++ b/src/coreclr/jit/optimizebools.cpp
@@ -533,6 +533,96 @@ bool IsConstantRangeTest(GenTreeOp* tree, GenTree** varNode, GenTreeIntCon** cns
 }
 
 //------------------------------------------------------------------------------
+// FoldRangeTests: Given two compare nodes (cmp1 && cmp2) where cmp1 is X >= 0
+//    and cmp2 is X < NN (NeverNegative), try to fold the range test into a single
+//    X u< NN (unsigned) compare node.
+//
+// Arguments:
+//    compiler       - compiler instance
+//    cmp1           - first compare node
+//    cmp1IsReversed - true if cmp1 is in fact reversed
+//    cmp2           - second compare node
+//    cmp2IsReversed - true if cmp2 is in fact reversed
+//
+// Returns:
+//    true if cmp1 now represents the folded range check and cmp2 can be removed.
+//
+bool FoldNeverNegativeRangeTest(
+    Compiler* comp, GenTreeOp* cmp1, bool cmp1IsReversed, GenTreeOp* cmp2, bool cmp2IsReversed)
+{
+    GenTree*       var1Node;
+    GenTreeIntCon* cns1Node;
+    genTreeOps     cmp1Op;
+
+    // First cmp has to be "X >= 0" (or "0 <= X")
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
Would still prefer we sort these out earlier, but also like having some "defense in depth."

> 1. We already use `uint` hack (I call it a **hack** because IMO it's not how most developers check ranges. Also, it's not "use checked context" [friendly](https://github.com/dotnet/runtime/issues/78214)) everywhere in BCL

Once take this change, can we undo some of these?

AndyAyersMS (approved):

markples:
@EgorBo just a few notes on the test as I looked through it due to #94146.  Some are just ideas to make test authoring easier in the future and might not be worth retroactive changes

- The messages in the test exceptions have copy/paste errors.  I believe that TestEntryPoint could be split into 3 (or 4) different [Fact] methods and then the messages wouldn't be needed at all as the test names would suffice.
- I think `Assert.Throws` could be used instead of reimplementing it.
- Are Test_Span1 and Test_Array identical?  I don't understand the comment in Test_Array about arr being null since arr is a Span.

BruceForstall:
Please cc @dotnet/jit-contrib for JIT PRs.

EgorBo:
> Please cc @dotnet/jit-contrib for JIT PRs.

It wasn't merged by me, I planned to ask for review once I am back from vacation 🤷‍♂️ It's unfortunate that `return 100;` typo didn't make CI red.

@markples thanks, will apply the suggestions as part of some other PR

kunalspathak:
> It wasn't merged by me, I planned to ask for review once I am back from vacation

Sorry, i was going through the "approved PRs" list and merged this one.

>  It's unfortunate that return 100; typo in the test didn't make CI red.

Do we know why this happened and if there is a fix needed?

EgorBo:
> Do we know why this happened and if there is a fix needed?

https://github.com/dotnet/runtime/pull/94146#issuecomment-1786058264 it's not clear yet

Issue #93852, opened 2023-10-23
[mono][tests] Disable System.Reflection and System.Threading.Tasks.Parallel tests 


kotlarmilos:
## Description

This PR disables failing tests on Apple mobile platforms. The following library tests consistently fail on tvOS, reporting the tcp connection issue:
`System.Reflection.InvokeEmit.Tests`
`System.Reflection.InvokeInterpreted.Tests`

The `System.Threading.Tasks.Parallel` tests fail with exit code 4, which may indicate permissions issue, possibly related to tests attempting to access the disk. Additionally, the System.Security.Cryptography tests report timeout.

The `System.Reflection` tests fail locally with the SIGSEGV error.
```
* thread #12, name = '.NET TP Worker', stop reason = signal SIGSEGV
    frame #0: 0x0000004d10000010
    frame #1: 0x000000010d9589a4 System.Reflection.InvokeEmit.Tests`System_Reflection_Tests_FunctionPointerMethods_CallFcnPtr_FP___int(fn=268435472, value=42) at MethodInfoTests.cs:1104
    frame #2: 0x000000010c08eef8 System.Reflection.InvokeEmit.Tests`wrapper_runtime_invoke_object_runtime_invoke_dynamic_intptr_intptr_intptr_intptr + 296
    frame #3: 0x0000000104952f74 System.Reflection.InvokeEmit.Tests`mono_jit_runtime_invoke(method=<unavailable>, obj=<unavailable>, params=<unavailable>, exc=<unavailable>, error=0x000000016bfc7570) at mini-runtime.c:3634:3 [opt]
    frame #4: 0x00000001048efa78 System.Reflection.InvokeEmit.Tests`mono_runtime_invoke_checked [inlined] do_runtime_invoke(method=0x000000013c41faa8, obj=0x0000000000000000, params=0x000000016bfc74a0, exc=0x0000000000000000, error=0x000000016bfc7570) at object.c:2541:11 [opt]
    frame #5: 0x00000001048efa3c System.Reflection.InvokeEmit.Tests`mono_runtime_invoke_checked(method=0x000000013c41faa8, obj=0x0000000000000000, params=0x000000016bfc74a0, error=0x000000016bfc7570) at object.c:2757:9 [opt]
[...]

PR #93853, opened 2023-10-23
[mono][tests] Disable System.Reflection and System.Threading.Tasks.Parallel tests 

kotlarmilos:
## Description

This PR disables failing library tests on Apple mobile platforms. Additionally, it increases the timeout to run the `System.Security.Cryptography` tests. More details available at https://github.com/dotnet/runtime/issues/93852.


kotlarmilos:
@vargaz When conducting local tests, it has been noticed that the `MONO_TYPE_FNPTR` cases are missing when utilizing the mini backend for arm64. Please take a look and suggest if additional changes are needed.

steveisok:
I'll be. tvOS is all green!

steveisok (approved):

PR #93854, opened 2023-10-23
[release/8.0-staging] Don't use File.OpenWrite when trying to overwrite a file in illink/wasm build tasks

akoeplinger:
Backport of #93744

File.OpenWrite will open an existing file and not truncate existing data, this can lead to unexpected results if the new data is shorter than the existing data.

## Customer Impact

illink has a feature to dump a dependency analysis xml during the build and when you run another build after making a change that'd reduce dependencies the xml would get corrupted since it wrote less data than the existing file had.

The fix is to use the FileMode.Create option in cases where the intention is to create a new file or overwrite an existing one.

I noticed another instance of this in the Wasm build tasks so fixed it there as well.

## Testing

Verified an existing file gets overwritten as expected.

## Risk

Low


carlossanlop:
@akoeplinger we still have time to merge fixes into release/8.0, in case you want to retarget this PR to that branch. Otherwise, we'll ship this in servicing version 1.

akoeplinger:
Servicing version 1 is fine.

akoeplinger:
Approved by tactics via email.

steveisok (approved):

PR #93875, opened 2023-10-23
[mono] Fix deadlock in static constructor initialization

### LINKED ISSUE #93778: [mono] Deadlock on TypeInitializationLock in mono_runtime_class_init_full
We've seen a deadlock in roslyn when performing a source-build of rc2 natively on ppc64le (using the Mono runtime).  In a nutshell, it seems the logic behind this block code code in `mono_runtime_class_init_full` is not thread-safe:
```
                /* see if the thread doing the initialization is already blocked on this thread */
                gboolean is_blocked = TRUE;
                blocked = GUINT_TO_POINTER (MONO_NATIVE_THREAD_ID_TO_UINT (lock->initializing_tid));
                while ((pending_lock = (TypeInitializationLock*) g_hash_table_lookup (blocked_thread_hash, blocked))) {
                        if (mono_native_thread_id_equals (pending_lock->initializing_tid, tid)) {
                                if (!pending_lock->done) {
                                        mono_type_initialization_unlock ();
                                        goto return_true;
                                } else {
                                        /* the thread doing the initialization is blocked on this thread,
                                           but on a lock that has already been freed. It just hasn't got
                                           time to awake */
                                        is_blocked = FALSE;
                                        break;
                                }
                        }
                        blocked = GUINT_TO_POINTER (MONO_NATIVE_THREAD_ID_TO_UINT (pending_lock->initializing_tid));
                }
 ```

To trigger the race, it is necessary that two threads each are in `mono_runtime_class_init_full` twice.  This can happen since `mono_runtime_class_init_full` performs `mono_runtime_try_invoke`, which executes some unknown managed code, which in turn can trigger a recursive `mono_runtime_class_init_full` call.   If both threads try to initialize the same two classes X and Y, but in reverse order (i.e. thread A accesses Y in the .cctor of X while thread B accesses X in the .cctor of Y), we may end up in a deadlock.

To trigger the deadlock, it seems that a third class Z also has to be involved.  That can lead to erroneous logic in the code above: Assume thread A is working both on X and (recursively) Z, and after Z is done tries to work on Y.  But thread B has already started working on Y and is now recursively blocked on Z before then also requiring X.    Since thread A sees that thread B is blocked on Z, which it itself has already completed, the code above triggers:
```
                                        /* the thread doing the initialization is blocked on this thread,
                                           but on a lock that has already been freed. It just hasn't got
                                           time to awake */
```
However, while it is true that thread B will indeed get woken up from Z's lock, that does not mean it will successfully complete initialization of Y as the Y's .cctor also needs X - where thread B will get again blocked on thread A after all.

In the example we're seeing, those nested initializations happen on various instantiations of `Microsoft.CodeAnalysis.VisualBasic.Symbols.OverrideHidingHelper`, which has this .cctor:
```
        Shared Sub New()
            OverrideHidingHelper(Of MethodSymbol).s_runtimeSignatureComparer = MethodSignatureComparer.RuntimeMethodSignatureComparer
            OverrideHidingHelper(Of PropertySymbol).s_runtimeSignatureComparer = PropertySignatureComparer.RuntimePropertySignatureComparer
            OverrideHidingHelper(Of EventSymbol).s_runtimeSignatureComparer = EventSignatureComparer.RuntimeEventSignatureComparer
        End Sub
```

A detailed timeline for a possible deadlock is as follows.  (I cannot guarantee that this timeline is *exacty* what happened since I can only examine the deadlocked state.  But it is a possibility how we could have gotten there.)

```
Thread A                                             Thread B

  mono_runtime_class_init_full (MonoVTable X)          mono_runtime_class_init_full (MonoVTable Y)
   mono_type_initialization_lock
   lookup X in type_initialization_hash: n/a
   allocate TypeInitializationLock LX
   insert into type_initialization_hash: X -> LX
   mono_type_initialization_unlock
                                                         mono_type_initialization_lock
                                                         lookup Y in type_initialization_hash: n/a
                                                         allocate TypeInitializationLock LY
                                                         insert into type_initialization_hash: Y -> LY
                                                         mono_type_initialization_unlock

   mono_runtime_try_invoke (X .cctor)                    mono_runtime_try_invoke (Y .cctor)
     ... managed code ...                                   ... managed code ...
     mono_runtime_class_init_full (MonoVTable Z)            mono_runtime_class_init_full (MonoVTable Z)
       mono_type_initialization_lock
       lookup Z in type_initialization_hash: n/a
       allocate TypeInitializionLock LZ
       insert into type_initialization_hash: Z -> LZ
       mono_type_initialization_unlock
                                                              mono_type_initialization_lock
                                                              lookup Z in type_initialization_hash: LZ
                                                              LZ->initializing_tid is thread A
                                                              lookup A in blocked_thread_hash: n/a
                                                              insert into blocked_thread_hash: B -> LZ
                                                              mono_type_initialization_unlock
                                                              wait on LZ
       [... complete initialization of Z ...]
       LZ->done = TRUE;
       wakeup waiters on LZ
     mono_runtime_class_init_full (MonoVTable Z) returns
     ... managed code ...
     mono_runtime_class_init_full (MonoVTable Y)
       mono_type_initialization_lock
       lookup Y in type_initialization_hash: LY
       LY->initializing_tid is thread B
       lookup B in blocked_thread_hash: LZ
       LZ->initializing_tid is thread A
       LZ->done is TRUE
         /* the thread doing the initialization is blocked on this thread,
            but on a lock that has already been freed. It just hasn't got
            time to awake */
       mono_type_initialization_unlock
       wait on LY
                                                              wakeup from LZ
                                                              mono_type_initialization_lock
                                                              remove from blocked_thread_hash: B-> LZ
                                                              remove from type_initialization_hash: Z -> LZ
                                                              mono_type_initialization_unlock
                                                            mono_runtime_class_init_full (MonoVTable Z) returns
                                                            ... managed code ...
                                                            mono_runtime_class_init_full (MonoVTable X)
                                                              mono_type_initialization_lock
                                                              lookup X in type_initialization_hash: LX
                                                              LX->initializing_tid is thread A
                                                              lookup A in blocked_thread_hash: n/a
                                                              insert into blocked_thread_hash: B -> LX
                                                              mono_type_initialization_unlock
                                                              wait on LX
```

CC - @directhex @lambdageek @vargaz @akoeplinger
FYI - @giritrivedi @alhad-deshpande @janani66 @omajid @tmds

### PR DESCRIPTION
**Pull request by lambdageek**
If two threads (A and B) need to call the static constructors for 3 classes X, Y, Z in this order:

Thread A: X { Z, Y }
Thread B: Y { Z, X }

where the cctor for X in thread A invokes the cctor for Z and for Y, and the cctor for Y in thread B invokes the cctor for Z and X, then we can get into a situation where A and B both start the cctors for X and Y (so they will be in the type_initialization_hash for those two classes) and then both will try to init Z.  In that case it could be that A will be responsible for initializing Z and B will block.  Then A could finish initializing Z but B may not have woken up yet (and so it will be in blocked_thread_hash waiting for Z).  At that point A (who is at this point already need to init Y) may think that it can wait for B to finish initializing Y.  That is we get to
`mono_coop_cond_timedwait (&lock->cond, &lock->mutex, timeout_ms)` with "lock" being the lock for `Y`.  But in fact thread B will not be able to complete initializing Y because it will attempt to init X next - but meanwhile we did not indicate that thread A is blocked. As a result in thread A the timed wait will eventually timeout.  And in this case we need to go back to the top and now correctly detect that A is waiting for Y and B is waiting for X.  (At that point there is a cctor deadlock and ECMA rules allow one of the threads to return without calling the cctor)

The old code here used to do an infinite wait:

```c
  while (!lock->done)
    mono_coop_cond_wait (&lock->cond, &lock->mutex)
```

This cannot succeed because "lock" (in thread A it's the lock for Y) will not be signaled since B (who is supposed to init Y) will instead block on the cctor for X.

Fixes https://github.com/dotnet/runtime/issues/93778

### DIFF
--- a/src/mono/mono/metadata/object.c
+++ b/src/mono/mono/metadata/object.c
@@ -507,7 +507,7 @@ retry_top:
 		blocked = GUINT_TO_POINTER (MONO_NATIVE_THREAD_ID_TO_UINT (lock->initializing_tid));
 		while ((pending_lock = (TypeInitializationLock*) g_hash_table_lookup (blocked_thread_hash, blocked))) {
 			if (mono_native_thread_id_equals (pending_lock->initializing_tid, tid)) {
-				if (mono_trace_is_traced (G_LOG_LEVEL_DEBUG, MONO_TRACE_TYPE) && strstr (m_class_get_image(klass)->name, "HelloWorld") != NULL) {
+				if (mono_trace_is_traced (G_LOG_LEVEL_DEBUG, MONO_TRACE_TYPE)) {
 					char* type_name = mono_type_full_name (m_class_get_byval_arg (klass));
 					mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_TYPE, "Detected deadlock for class .cctor for %s from '%s'", type_name, m_class_get_image (klass)->name);
 					g_free (type_name);
@@ -517,8 +517,6 @@ retry_top:
 					mono_type_initialization_unlock ();
 					goto return_true;
 				} else {
-					mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_TYPE, "special case");
-
 					/* the thread doing the initialization is blocked on this thread,
 					   but on a lock that has already been freed. It just hasn't got
 					   time to awake */
--- a/src/mono/mono/metadata/object.c
+++ b/src/mono/mono/metadata/object.c
@@ -647,9 +647,9 @@ retry_top:
 				mono_type_initialization_unlock ();
 				goto retry_top;
 			} else if (wait_result == 0) {
-				/* success.. we were signaled that the other thread is done.  Proceed */
+				/* Success: we were signaled that the other thread is done.  Proceed */
 			} else {
-				g_error ("unexpected return value %d from mono_coop_cond_timedwait", wait_result);
+				g_assert_not_reached ();
 			}
 		}
 		mono_type_init_unlock (lock);
--- a/src/mono/mono/metadata/object.c
+++ b/src/mono/mono/metadata/object.c
@@ -457,6 +457,7 @@ mono_runtime_class_init_full (MonoVTable *vtable, MonoError *error)
 	 * on this cond var.
 	 */
 
+retry_top:
 	mono_type_initialization_lock ();
 	/* double check... */
 	if (vtable->initialized) {
@@ -506,10 +507,18 @@ mono_runtime_class_init_full (MonoVTable *vtable, MonoError *error)
 		blocked = GUINT_TO_POINTER (MONO_NATIVE_THREAD_ID_TO_UINT (lock->initializing_tid));
 		while ((pending_lock = (TypeInitializationLock*) g_hash_table_lookup (blocked_thread_hash, blocked))) {
 			if (mono_native_thread_id_equals (pending_lock->initializing_tid, tid)) {
+				if (mono_trace_is_traced (G_LOG_LEVEL_DEBUG, MONO_TRACE_TYPE) && strstr (m_class_get_image(klass)->name, "HelloWorld") != NULL) {
+					char* type_name = mono_type_full_name (m_class_get_byval_arg (klass));
+					mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_TYPE, "Detected deadlock for class .cctor for %s from '%s'", type_name, m_class_get_image (klass)->name);
+					g_free (type_name);
+				}
+
 				if (!pending_lock->done) {
 					mono_type_initialization_unlock ();
 					goto return_true;
 				} else {
+					mono_trace (G_LOG_LEVEL_DEBUG, MONO_TRACE_TYPE, "special case");
+
 					/* the thread doing the initialization is blocked on this thread,
 					   but on a lock that has already been freed. It just hasn't got
 					   time to awake */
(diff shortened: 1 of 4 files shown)


lambdageek:
I wonder if I need a test where a static cctor dependency takes a long time to unblock (longer than the timeout that I added here).  Just to verify that I didn't break the ability for one thread to wait for another to complete.

src/mono/mono/metadata/object.c
```diff
@@ -604,9 +611,49 @@ mono_runtime_class_init_full (MonoVTable *vtable, MonoError *error)
...
+				g_hash_table_remove (blocked_thread_hash, GUINT_TO_POINTER (tid));
+				gboolean deleted = unref_type_lock (lock);
+				if (deleted)
+					g_hash_table_remove (type_initialization_hash, vtable);
+				mono_type_initialization_unlock ();
+				goto retry_top;
+			} else if (wait_result == 0) {
+				/* success.. we were signaled that the other thread is done.  Proceed */
+			} else {
+				g_error ("unexpected return value %d from mono_coop_cond_timedwait", wait_result);
```
vargaz, line 652: Could this be EINTR or something?

vargaz (approved):

src/mono/mono/metadata/object.c
lateralusX: Looks like mono_coop_cond_timedwait will only return -1 and 0, so if this is hit, then something changed in mono_coop_cond_timedwait and mono_coop_cond_timedwait will already do:

if (G_UNLIKELY (res != 0 && res != ETIMEDOUT)) {
	g_print ("cond: %p mutex: %p\n", *(gpointer*)cond, *(gpointer*)mutex);
	g_error ("%s: pthread_cond_timedwait failed with \"%s\" (%d) %ld %ld %d", __func__, g_strerror (res), res, ts.tv_sec, ts.tv_nsec, timeout_ms);
}

so maybe this should just be a g_assert_not_reached instead?

lateralusX (approved):
LGTM! Implementation details looks reasonable and since PR just switched from a infinite wait to a timeout wait with a retry loop following the current patterns shouldn't introduce any new side effects.

PR #93884, opened 2023-10-23
[fix] Add missing record keyword

ngarside:


src/libraries/Common/src/System/CSharpHelpers.cs
```diff
@@ -60,6 +60,7 @@ internal abstract class CSharpHelpers
             "object",
             "params",
             "public",
+            "record",
```
vcsjones, line 63: This looks like it's a list of keywords. `record` is a [Contextual Keyword](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/#contextual-keywords), not a "full" keyword. It seems like all of the other contextual keywords are missing from this list.

Can you describe why you are making this change and what problem it is solving?

ngarside: Ah my bad, noticed it when building a source generator and thought it got missed, but didn't realise it wasn't a proper keyword. I'll close this PR.

PR #93886, opened 2023-10-23
Change OpenSslVersion to check for OpenSSL before attempting to determine openssl version

### LINKED ISSUE #93883: SafeEvpPKeyHandle abort on macOS
The `SafeEvpPKeyHandle` class will abort the process on macOS because of the static constructor to set `OpenSslVersion`:

https://github.com/dotnet/runtime/blob/111b089775820886b01607f4c7d0f6a2e7b81226/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SafeEvpPKeyHandle.OpenSsl.cs#L88

We can fix this to guard like the other methods and properties do, but do we want the property to throw (something we generally don't like properties to do, but it's better than an abort) or do we want it to return a "I don't know" sentinel like `0`.

vcsjones:
I decided to do a partial so the implementation remains unchanged on Unix. If someone took a dependency on this property being cheap, it stays so on Unix.

I don't like throwing an exception on a property read, but it's better than a process abort, and it seemed preferable to some sentinel "zero" return value.

Fixes #93883

### DIFF
--- a/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SafeEvpPKeyHandle.OpenSsl.macOS.cs
+++ b/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SafeEvpPKeyHandle.OpenSsl.macOS.cs
@@ -8,7 +8,7 @@ namespace System.Security.Cryptography
 {
     public sealed partial class SafeEvpPKeyHandle
     {
-        private static long _openSslVersion;
+        private static readonly long _openSslVersion = !Interop.OpenSslNoInit.OpenSslIsAvailable ? 0 : Interop.OpenSsl.OpenSslVersionNumber();
 
         /// <summary>
         /// The runtime version number for the loaded version of OpenSSL.
@@ -26,17 +26,11 @@ namespace System.Security.Cryptography
         {
             get
             {
-                long version = Volatile.Read(ref _openSslVersion);
+                long version = _openSslVersion;
 
                 if (version == 0)
                 {
-                    if (!Interop.OpenSslNoInit.OpenSslIsAvailable)
-                    {
-                        throw new PlatformNotSupportedException(SR.PlatformNotSupported_CryptographyOpenSSL);
-                    }
-
-                    version = Interop.OpenSsl.OpenSslVersionNumber();
-                    Volatile.Write(ref _openSslVersion, version);
+                    throw new PlatformNotSupportedException(SR.PlatformNotSupported_CryptographyOpenSSL);
                 }
 
                 return version;
--- a/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SafeEvpPKeyHandle.OpenSsl.macOS.cs
+++ b/src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SafeEvpPKeyHandle.OpenSsl.macOS.cs
@@ -1,0 +1,46 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.Versioning;
+using System.Threading;
+
+namespace System.Security.Cryptography
+{
+    public sealed partial class SafeEvpPKeyHandle
+    {
+        private static long _openSslVersion;
+
+        /// <summary>
+        /// The runtime version number for the loaded version of OpenSSL.
+        /// </summary>
+        /// <remarks>
+        /// For OpenSSL 1.1+ this is the result of <code>OpenSSL_version_num()</code>,
+        /// for OpenSSL 1.0.x this is the result of <code>SSLeay()</code>.
+        /// </remarks>
+        [UnsupportedOSPlatform("android")]
+        [UnsupportedOSPlatform("browser")]
+        [UnsupportedOSPlatform("ios")]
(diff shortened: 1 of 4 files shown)


src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SafeEvpPKeyHandle.OpenSsl.macOS.cs
```diff
@@ -0,0 +1,46 @@
...
+        [UnsupportedOSPlatform("android")]
+        [UnsupportedOSPlatform("browser")]
+        [UnsupportedOSPlatform("ios")]
+        [UnsupportedOSPlatform("tvos")]
+        [UnsupportedOSPlatform("windows")]
+        public static long OpenSslVersion
+        {
+            get
+            {
+                long version = Volatile.Read(ref _openSslVersion);
```
Jozkee, line 29: What would happen if you don't use Volatile operations?

Jozkee:
Why does it abort on macOS but not on Unix?

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SafeEvpPKeyHandle.OpenSsl.macOS.cs
vcsjones: Since the field is static, one thread could be reading the field and another could be writing to it. We don't want one thread reading a cached value if another thread updated it.

In theory if macOS supported 32-bit it could also result in torn reads, but since macOS doesn't, that's not a concern.

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SafeEvpPKeyHandle.OpenSsl.macOS.cs
```diff
@@ -0,0 +1,46 @@
...
+                    {
+                        throw new PlatformNotSupportedException(SR.PlatformNotSupported_CryptographyOpenSSL);
+                    }
+
+                    version = Interop.OpenSsl.OpenSslVersionNumber();
+                    Volatile.Write(ref _openSslVersion, version);
+                }
+
+                return version;
+            }
```
MichalPetryka, line 43:
```suggestion
        private static readonly long _openSslVersion = !Interop.OpenSslNoInit.OpenSslIsAvailable ? 0 : Interop.OpenSsl.OpenSslVersionNumber();

        /// <summary>
        /// The runtime version number for the loaded version of OpenSSL.
        /// </summary>
        /// <remarks>
        /// For OpenSSL 1.1+ this is the result of <code>OpenSSL_version_num()</code>,
        /// for OpenSSL 1.0.x this is the result of <code>SSLeay()</code>.
        /// </remarks>
        [UnsupportedOSPlatform("android")]
        [UnsupportedOSPlatform("browser")]
        [UnsupportedOSPlatform("ios")]
        [UnsupportedOSPlatform("tvos")]
        [UnsupportedOSPlatform("windows")]
        public static long OpenSslVersion
        {
            get
            {
                long version = _openSslVersion;

                if (version == 0)
                {
                    throw new PlatformNotSupportedException(SR.PlatformNotSupported_CryptographyOpenSSL);
                }

                return version;
            }
```
Wouldn't this be simpler and better for perf?

vcsjones:
> Why does it abort on macOS but not on Unix?

.NET on Unix does abort without libssl (OpenSSL). However on Unix, OpenSSL is a hard requirement. .NET is basically non-functional on Unix without OpenSSL.

On macOS, OpenSSL is a soft requirement. You don't need it to use .NET on macOS, but some functionality depends on it (like SafeEvpPKeyHandle). In those places, macOS needs to guard with `Interop.OpenSslNoInit.OpenSslIsAvailable`.

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SafeEvpPKeyHandle.OpenSsl.macOS.cs
vcsjones: Yeah that is probably a better way to do this.

src/libraries/System.Security.Cryptography/src/System/Security/Cryptography/SafeEvpPKeyHandle.OpenSsl.macOS.cs
Jozkee: > We don't want one thread reading a cached value if another thread updated it.

What would be a dangerous cached value? I'm think it would be if openSSL is updated in the middle on the process running, is that correct?

vcsjones: Resolving since we don't need to use the volatile since its back in a cctor now.

danmoseley: I guess I'm confused as well. A thread either sees 0 (and finds and writes the value) or the correct value. What does "reading while another is writing" mean if there is no tearing?

jozkee (approved):

PR #89330, opened 2023-07-21
Use AVX512 intrinsics for number conversion when possible

martinothamar:
This PR implements AVX512DQ intrinsics where possible for converting integers to floating points and vice versa.

Usecase: I had been working on vectorized PRNGs [in my Fast.PRNGs .NET library](https://github.com/martinothamar/Fast.PRNGs.NET) and [in my simd-rand Rust library](https://github.com/martinothamar/simd-rand). I was unable to match the performance in the Rust implementation partially due to uint64 -> double conversion being slower (when running on hardware with AVX512).

I did also search for related issues, where I found this: #85207. I think this fixes the first few bullet points there.

I benchmarked with disasm for a ulong -> double Vector256 conversion which seems to look good. I used a modified Vector256 version of this benchmark: https://github.com/dotnet/performance/blob/981a1a6730a5c54b3292428690ac9948e68c4823/src/benchmarks/micro/libraries/System.Runtime.Intrinsics/Perf_Vector128.cs#L36
My only modification was not using a `static readonly` field to convert in the benchmark method, rather initializing a member field in a `GlobalSetup` method using random values (it had some small effect on the generated code). I noticed an open issue relating to the benchmarking setup: https://github.com/dotnet/performance/issues/2555. I could contribute benchmarks for wider vectors, but when doing conversion in a loop I struggled and didn't manage to avoid some dead code elimination to mess up the generated code, so the benchmarks would look pretty similar to the current 128bit ones, aside from the static data. Is there a good way to try to prevent these types of compiler optimizations from benchmark code? Something like the [black_box](https://doc.rust-lang.org/stable/std/hint/fn.black_box.html) intrinsic in Rust, or inline asm tricks commonly used in C/C++

# Benchmarks

## .NET 8.0.0 (8.0.23.32907), X64 RyuJIT AVX2
```assembly
; System.Runtime.Intrinsics.Tests.Perf_Vector256.ConvertULongToDoubleBenchmark()
       vzeroupper
       vmovups   ymm0,[rdi+8]
       vpblendd  ymm1,ymm0,[7F9E757F1580],0AA
       vpsrlq    ymm0,ymm0,20
       vpxor     ymm0,ymm0,[7F9E757F15A0]
       vsubpd    ymm0,ymm0,[7F9E757F15C0]
       vaddpd    ymm0,ymm0,ymm1
       vmovups   [rsi],ymm0
       mov       rax,rsi
       vzeroupper
       ret
; Total bytes of code 54
```

```

[...]

PR #84384, opened 2023-04-05
Optimize scalar conversions with AVX512

khushal1996:
This PR optimize the following cases:

---------------------------------
| Case | Previous Code | Optimized Instruction |
|-------|----------|-------------|
| ulong -> double| vcvtsi2sd  | vcvtusi2sd |

```C#
public static double UIntToDouble(UInt64 val)
{
    return (double)val;
}
```

**Assembly before optimization**
```asm
G_M33997_IG01:              ;; offset=0000H
       C5F877               vzeroupper 
						;; size=3 bbWeight=1 PerfScore 1.00
G_M33997_IG02:              ;; offset=0003H
       62F17C0857C0         vxorps   xmm0, xmm0
       62F1FF082AC1         vcvtsi2sd  xmm0, rcx
       4885C9               test     rcx, rcx
       7D0A                 jge      SHORT G_M33997_IG03
       62F1FF08580502000000 vaddsd   xmm0, qword ptr [reloc @RWD00]
						;; size=27 bbWeight=1 PerfScore 12.58
G_M33997_IG03:              ;; offset=001EH
       C3                   ret
```

**Assembly after optimization**
```asm
G_M33997_IG01:              ;; offset=0000H
       C5F877               vzeroupper 
						;; size=3 bbWeight=1 PerfScore 1.00
G_M33997_IG02:              ;; offset=0003H
       62F1FF087BC1         vcvtusi2sd xmm0, rcx
						;; size=6 bbWeight=1 PerfScore 4.00
G_M33997_IG03:              ;; offset=0009H
       C3                   ret
```

| Case | Previous Code | Optimized Instruction |
|-------|----------|-------------|
| ulong -> float| ulong->double->float | vcvttusi2ss|

```C#
[MethodImplAttribute(MethodImplOptions.NoInlining)]
public static float ConvUlongToFloat(ulong val)
{
    return (float)val;
}
```

**Assembly before optimization**
```asm
G_M2883_IG01:  ;; offset=0000H
       push     rbp
       sub      rsp, 48
       vzeroupper 
       lea      rbp, [rsp+30H]
       xor      eax, eax
       mov      dword ptr [rbp-04H], eax
       mov      qword ptr [rbp+10H], rcx
						;; size=22 bbWeight=1 PerfScore 5.00
G_M2883_IG02:  ;; offset=0016H
       cmp      dword ptr [(reloc 0x7ff8bc0d2898)], 0
       je       SHORT G_M2883_IG04
[...]

Issue #61885, opened 2021-11-21
Update the runtime to have deterministic floating-point to integer conversions


tannergooding:
### Issue

.NET currently depends on the underlying platform to perform `floating-point to integer` conversions. For most scenarios, this behavior is correct and IEEE 754 compliant.

However, for cases of overflow or cases where the underlying platform does not directly support a given conversion, each platform can return different results.

This differing behavior can lead to downstream issues when dealing with such conversions and lead to hard to diagnose bugs.

### Proposal

We should follow in the footsteps of other languages and standardize our conversion behavior here to be consistent on all platforms.

In particular, we should standardize our behavior to be consistent with ARM64, [Rust](https://github.com/rust-lang/rust/issues/67058), and WASM and to saturate the conversions rather than some other behavior such as using a sentinel value (x86/x64). `NaN` will likewise be specially handled and overflow to `0`.

### What's Required

For ARM32 and ARM64, there will be effectively no impact and the codegen will remain identical to what it is today as the underlying conversions already perform saturation on overflow. The exception is for conversions to small integers where there is no direct underlying platform support and so manual clamping of the input to the appropriate range will be required before converting.

For x86 and x64, the platform only directly supports converting to int32 with additional support for converting to int64 on x64. Conversions to small integer types and unsigned types will require the same manual clamping as ARM32/ARM64. Additionally, since the underlying platform returns a sentinel value (`0x80000000` or `0x80000000_00000000`) we will need to have logic to handle this.

### Perf Impact

For the conversions that need additional handling, there is going to be additional cost and measurable perf impact.

For a Skylake processor simple conversions would go from ~7 cycles to approx. between 19 and 24 cycles, or a theoretical "worst case" taking ~3.4x more time.

The measured impact by Rust was much lower and was closer to a 0.8x regression in real world scenarios involving heavy floating-point to integer conversions (in particular an "RBG JPEG encoding" algorithm).

### Additional Considerations

We should additionally expose a set of unsafe APIs that perform the "platform specific" conversions. These would allow developers who understand the xplat differences and need the perf to still get the underlying behavior.

[...]

PR #93896, opened 2023-10-23
Use AVX512 intrinsics for numeric conversion in JIT

martinothamar:
Jit version of #89330

For `Vector128.ConvertToDouble`

### Before
```sh
G_M000_IG02:                ;; offset=0x0007
       call     [OthmarTesting.Testing:Src():long]
       vpbroadcastq  xmm0, rax
       vpblendd xmm1, xmm0, xmmword ptr [reloc @RWD00], 10
       vpsrlq   xmm0, xmm0, 32
       vpxorq   xmm0, xmm0, qword ptr [reloc @RWD16] {1to2}
       vsubpd   xmm0, xmm0, qword ptr [reloc @RWD24] {1to2}
       vaddpd   xmm0, xmm0, xmm1
       vmovups  xmmword ptr [rbx], xmm0
       mov      rax, rbx
```

### After
```sh
G_M000_IG02:                ;; offset=0x0007
       call     [OthmarTesting.Testing:Src():long]
       vpbroadcastq  xmm0, rax
       vcvtqq2pd xmm0, xmm0
       vmovups  xmmword ptr [rbx], xmm0
       mov      rax, rbx
```

### DIFF
--- a/src/coreclr/jit/hwintrinsicxarch.cpp
+++ b/src/coreclr/jit/hwintrinsicxarch.cpp
@@ -1405,6 +1405,30 @@ GenTree* Compiler::impSpecialIntrinsic(NamedIntrinsic        intrinsic,
 
         case NI_Vector128_ConvertToDouble:
         case NI_Vector256_ConvertToDouble:
+        {
+            assert(sig->numArgs == 1);
+            assert(varTypeIsLong(simdBaseType));
+
+            if (!compOpportunisticallyDependsOn(InstructionSet_AVX512DQ_VL))
+            {
+                break;
+            }
+
+            switch (simdSize)
+            {
+                case 16:
+                    intrinsic = NI_AVX512DQ_VL_ConvertToVector128Double;
+                    break;
+                case 32:
+                    intrinsic = NI_AVX512DQ_VL_ConvertToVector256Double;
+                    break;
+                default:
+                    unreached();
+            }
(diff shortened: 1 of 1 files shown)


martinothamar:
Should I implement these hardware intrinsics for `Vector<T>` (`simdashwintrinsic.cpp`) as well? I see that `NI_VectorT128_ConvertToDouble` for example is only accelerated under ARM64

adamsitnik:
@dotnet/avx512-contrib the previous PR (#89330) provided by @martinothamar was left with no feedback for over two months, please don't let this PR become stale too.

tannergooding:
We appreciate the PR, but this is not something we'll be able to take at this time.

There are longer discussions covering the reasons why in https://github.com/dotnet/runtime/pull/84384 and https://github.com/dotnet/runtime/issues/61885, but in essence we have a desire to ensure the default behavior is deterministic across platforms.

This entails us not simply executing the direct instruction, but also doing the work to ensure that overflow results in saturation, rather than returning some sentinel value and doing the equivalent work to ensure that scalars, constant folding, and other paths all have the same behavior. This work is significantly more involved and requires additional testing, general integration, etc.

As such, it is work that we will be doing ourselves when the time is appropriate (this will likely be in the next few months).

martinothamar:
Ahh I see that makes sense, thanks for the quick reply. I'll go ahead and close the other PR as well

PR #93899, opened 2023-10-24
[release/8.0][wasm] WBT: Log more info to help debug CI

radical:
`release/8.0` HEAD is broken with all the WBT tests failing with:
```
    Wasm.Build.NativeRebuild.Tests.FlagsChangeRebuildTests.ExtraEmccFlagsSetButNoRealChange [FAIL]
      System.TypeInitializationException : The type initializer for 'Wasm.Build.Tests.BuildTestBase' threw an exception.
      ---- System.Exception : Expected the multithreading runtime pack to be available when running on CI. IsRunningOnCI is true but IsWorkloadWithMultiThreadingForDefaultFramework is false.
```


carlossanlop:
Please retarget your PR to `release/8.0-staging`.

lewing:
I think the issue was we were picking up 2xx sdks for workload testing and they have workload sets enabled and are breaking wbts https://github.com/dotnet/runtime/pull/94097

PR #93901, opened 2023-10-24
Report StackOverflowException on NativeAOT on Linux

### LINKED ISSUE #82334: [NativeAOT] Stackoverflow reporting on Linux
### Repro
```csharp
Recursion(1);

void Recursion(int x)
{
    Recursion(x+1);
    Recursion(x+1);
}
```

### Actual result

`Segmentation fault`

### Expected result

`Process is terminating due to StackOverflowException`

(Reported by partner team.)

jtschuster:
Uses almost the same code as CoreCLR to allocate an alternate stack for handling SIGSEGV, and properly reports a StackOverflowException.

Fixes https://github.com/dotnet/runtime/issues/82334


src/coreclr/nativeaot/Runtime/unix/HardwareExceptions.cpp
```diff
@@ -598,26 +613,30 @@ bool InitializeHardwareExceptionHandling()
     {
         return false;
     }
+#ifdef TARGET_UNIX
+    Thread* curThread = ThreadStore::RawGetCurrentThread();
+    curThread->EnsureSignalAlternateStack();
```
jkotas, line 618: `InitializeHardwareExceptionHandling` runs once per process. It does not run for each thread.

Try to trigger the stackoverflow on a secondary thread. I do not think it is going to be reported correctly.

jtschuster: There's a call to EnsureSignalAlternateStack in the Thread::Construct method that should handle new threads. It reports Stack Overflow in a repro project that overflows in a new thread.

jkotas: Ah ok. Why do we need to call it here? Should the call in Thread::Construct be enough?

jtschuster: Good point, it looks like the Thread::Construct call is enough.

janvorli (approved):
LGTM modulo the nit comment

src/tests/nativeaot/SmokeTests/StackOverflowReporting/Program.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+public static class StackOverflowReporting
+{
+    public static int Main(string[] args)
```
jkotas, line 7: Do we have a test like this for regular CoreCLR? It would be better to depend on a general test for this. We should not need native-aot specific smoke tests for this.

janvorli:
One thing I wonder about - don't we also want to call `PalRaiseFailFastException` instead of `RhFailFast` (like we do on Windows for stack overflow in `RhpVectoredExceptionHandler`)?

src/tests/nativeaot/SmokeTests/StackOverflowReporting/Program.cs
janvorli: We have stack overflow test for coreclr. That test tests several kinds of stack overflow (main thread / secondary thread / large frame / small frame). I think it would need to be modified to work with nativeaot though, as it checks part of the error message.

jkotas:
> One thing I wonder about - don't we also want to call PalRaiseFailFastException instead of RhFailFast (like we do on Windows for stack overflow in RhpVectoredExceptionHandler)?

RhFailFast calls PalRaiseFailFastException with default arguments. I do not think calling PalRaiseFailFastException in this change would make any difference.

janvorli:
Ah, thank you, I've missed that.

src/tests/nativeaot/SmokeTests/StackOverflowReporting/Program.cs
agocke: We currently pass `nativeaot;Loader;Interop` in CI: https://github.com/dotnet/runtime/blob/f28a8a72eee05e0b909a4f2cb58d66bc0f7a01fe/eng/pipelines/runtime.yml#L300C77-L300C77

Ideally  we could add `exceptions` to that if it isn't too expensive. Otherwise, if this is in pri0 tests I think it should run in the runtime-nativeaot-outerloop.

jtschuster: Thank you, I'll modify that test to work with NativeAOT.

jtschuster: I modified the test and removed the exclusion for `baseservices/exceptions/stackoverflow` tests in nativeaot, so the CoreCLR test should run in the outerloop.

src/tests/baseservices/exceptions/stackoverflow/stackoverflowtester.cs
```diff
@@ -4,27 +4,45 @@
...
 
-        static bool TestStackOverflow(string testName, string testArgs, out List<string> stderrLines)
+        static Process GetNativeAotProcess(string testName, string testArgs)
         {
-            Console.WriteLine($"Running {testName} test({testArgs})");
-            List<string> lines = new List<string>();
-
             Process testProcess = new Process();
+            string executableExtension = TestLibrary.Utilities.IsWindows ? ".exe" : "";
+            testProcess.StartInfo.FileName = Path.Combine(s_currentPath, "native", $"{ThisProjectName}{executableExtension}");
```
MichalStrehovsky, line 22: I'd rather we don't hardcode the directory structure with `native` and everything if we can help it. Would this work?

```suggestion
            testProcess.StartInfo.FileName = Environment.ProcessPath;
```

src/tests/Directory.Build.targets
```diff
@@ -603,7 +603,7 @@
           DependsOnTargets="Build;LinkNativeIfBuildAndRun" />
 
   <Target Name="LinkNativeIfBuildAndRun"
-          Condition="'$(_WillCLRTestProjectBuild)' == 'true' and '$(CLRTestKind)' == 'BuildAndRun'"
+          Condition="'$(_WillCLRTestProjectBuild)' == 'true' and ('$(CLRTestKind)' == 'BuildAndRun' or '$(CLRTestKind)' == 'BuildOnly')"
```
MichalStrehovsky, line 606: Is this still needed?

janvorli:
@jtschuster I am not sure if merging of the tests into a single exe won't complicate the work on merging coreclr tests that @trylek works on. I wonder what the reason for this change was.

MichalStrehovsky:
> @jtschuster I am not sure if merging of the tests into a single exe won't complicate the work on merging coreclr tests that @trylek works on. I wonder what the reason for this change was.

Can you clarify how it would complicate things? I think we'd just mark this single exe project as RequiresProcessIsolation and that will be the extent of it.

The current setup of this test is weird - there's no apparent connection between the three projects (not even a `<ProjectReference... ReferenceOutputAssembly="false" />` ) and the test is just hoping to find the DLL at runtime by digging through the filesystem and hardcoding assumptions about how test build lays out things on disk. We'd need to hardcode even more assumptions about how test build is done for Native AOT (we cannot execute the DLL, we need to find the native artifact which is in yet another directory). I'm trying to get rid of the last bit of this build system setup hardcoding with https://github.com/dotnet/runtime/pull/93901#discussion_r1374100662. I think this unification is an improvement.

janvorli:
@MichalStrehovsky I was thinking that in the current state, the actual test could be merged and just the two helper exes would stay separate. But I am not opposed to this change at all, it was just a question to understand the reason. I don't have enough understanding of how we currently merge tests, I was not aware that we can use a single property to keep the test separate.

MichalStrehovsky:
> I was thinking that in the current state, the actual test could be merged and just the two helper exes would stay separate

Locating the helper exes would likely become a lot more involved since they'd no longer be at `${currentDirectory}/../${testname}/${testname}.dll`. The test was hardcoding assumptions about the test build that would no longer apply in a merged runner - it doesn't look like a straightforward conversion either.

jtschuster:
> I wonder what the reason for this change was.

When running the the NativeAOT test on Linux, the other executables didn't have execute permissions and it failed to launch the subprocess, so I thought the easiest solution would be to combine all the functionality into the test exe that has permissions to run.

src/tests/Directory.Build.targets
jtschuster: No, I'll revert that, thank you!

agocke:
FYI I already do something similar for the DWARF symbol test -- I run dwarfdump on the currently running exe and compare the results with a baseline. This doesn't seem far off from that.

trylek:
At the first glance I don't see anything in this change that would go against our runtime infrastructure efforts - we already have many precedents of tests that need to run out of process and in particular the vast majority of GC tests support command-line arguments for use in various local testing scenarios, the merged test system does respect that.

src/tests/baseservices/exceptions/stackoverflow/stackoverflowtester.cs
```diff
@@ -4,27 +4,45 @@
...
-        static bool TestStackOverflow(string testName, string testArgs, out List<string> stderrLines)
+        static Process GetNativeAotProcess(string testName, string testArgs)
         {
-            Console.WriteLine($"Running {testName} test({testArgs})");
-            List<string> lines = new List<string>();
-
             Process testProcess = new Process();
+            testProcess.StartInfo.FileName = Environment.ProcessPath;
+            testProcess.StartInfo.Arguments = testArgs;
+            testProcess.StartInfo.Environment.Add("DOTNET_DbgEnableMiniDump", "0");
```
janvorli, line 23: Can you please set this env variable for non-nativeaot case too?

src/tests/baseservices/exceptions/stackoverflow/stackoverflowtester.cs
```diff
@@ -4,27 +4,45 @@
...
+            testArgs = $"{testName} {testArgs}";
+            List<string> lines = new List<string>();
+
+            Process testProcess;
+            if (TestLibrary.Utilities.IsNativeAot)
+            {
+                testProcess = GetNativeAotProcess(testName, testArgs);
+            }
+            else {
+                testProcess  = new Process();
```
janvorli, line 39: A nit - it would be nice to create the process for both native aot and non-native aot inline or in separate function (or maybe just use a single function for the process creation) instead of doing it inline for one and in a function for the other case.

jtschuster: Moved them both to be inline.

src/coreclr/nativeaot/Runtime/thread.h
```diff
@@ -112,6 +112,9 @@ struct ThreadBuffer
 #ifdef FEATURE_GC_STRESS
     uint32_t                m_uRand;                                // current per-thread random number
 #endif // FEATURE_GC_STRESS
+#if defined(TARGET_UNIX) && !HAVE_MACH_EXCEPTIONS && !defined(HOST_TVOS)
+    void *                   m_alternateStack;                      // ptr to alternate signal stack
```
janvorli, line 116: A nit - I've just noticed that the indentation is off by one space
```suggestion
    void *                  m_alternateStack;                      // ptr to alternate signal stack
```

src/tests/baseservices/exceptions/stackoverflow/stackoverflowtester.cs
```diff
@@ -3,26 +3,69 @@
...
-
-            testProcess.StartInfo.FileName = Path.Combine(Environment.GetEnvironmentVariable("CORE_ROOT"), "corerun");
-            testProcess.StartInfo.Arguments = $"{Path.Combine(Directory.GetCurrentDirectory(), "..", testName, $"{testName}.dll")} {testArgs}";
+            // Always use whatever runner started this test, or the exe if we're running single file / NativeAOT
+            testProcess.StartInfo.FileName = Environment.ProcessPath;
+            // We want this assembly in CoreCLR and empty string for single file / NativeAOT
+            testProcess.StartInfo.Arguments = typeof(Program).Assembly.Location;
+            testProcess.StartInfo.Environment.Add(EnableMiniDumpEnvVar, "0");
+            // Set the environment so the subprocess will trigger a stack overflow
+            testProcess.StartInfo.Environment.Add(TestNameEnvVar, testName);
```
janvorli, line 64: Why don't we just pass the test name and args as command line arguments?

jtschuster: With the XUnit generator generating `Main`, there's not a way to parse the arguments and change what gets run, so I went with using environment variables and made the test methods instance methods instead of static, and then it can check the environment in the test object ctor.

jtschuster: It's hacky for sure, and I'm open to doing it a differently if there's a better way.

janvorli: Why do we want to use the mode when xunit generates the Main? I have thought that this is a test that would stay to be a separate executable, like many other similar ones.

jtschuster: Not exactly sure, it was done in https://github.com/dotnet/runtime/pull/91560. @trylek Is there any special reason for this test to be moved to the generator besides unifying the tests?

janvorli: You can add `<ReferenceXUnitWrapperGenerator>false</ReferenceXUnitWrapperGenerator>` and `<RequiresProcessIsolation>true</RequiresProcessIsolation>` to the `PropertyGroup` in the csproj and then change the `[Fact]` method to good old `Main` with arguments. Then you can pass the arguments as usual. 
@trylek has changed it the way he did because in the world of test with two separate test exes, it was not causing any troubles.

janvorli (approved):

src/tests/baseservices/exceptions/stackoverflow/stackoverflowtester.cs
```diff
@@ -3,26 +3,82 @@
...
+        [UnconditionalSuppressMessage("SingleFile", "IL3000", Justification = "We want an empty string for location if the test is running as single file")]
+        static bool TestStackOverflow(string testName, string testArgs, out string[] stderrLines, out bool checkStackFrame)
         {
-            Console.WriteLine($"Running {testName} test({testArgs})");
             List<string> lines = new List<string>();
 
-            Process testProcess = new Process();
+            string thisAssemblyPath = typeof(Program).Assembly.Location;
+            // TestLibrary.Utilities.IsNativeAot brings in ~4mb on Linux, so we'll approximate with this for now. https://github.com/dotnet/runtime/issues/94313
+            bool IsNativeAot = string.IsNullOrEmpty(thisAssemblyPath) && !RuntimeFeature.IsDynamicCodeSupported;
```
MichalStrehovsky, line 70: I think you _actually_ want to test for single file, not for Native AOT. If we ever add single file (non-native AOT) runs of the test tree, this code would need tweaking.

Might as well just test for single file.

https://github.com/dotnet/runtime/blob/9d3f324640e8083ddfaebeed91119cd69d43cd33/src/tests/Common/CoreCLRTestLibrary/Utilities.cs#L100

jtschuster:
I made some mistake when I pulled main back into this branch. Closing this PR and opened a new one.

Issue #90563, opened 2023-08-15
Lock contention in StackTrace/Exception.ToString()


jkotas:
Lock contention has been observed in a real-world service at the following callstacks during a failover period where a lot of exceptions are thrown:

![image](https://github.com/dotnet/runtime/assets/6668460/75dc3de2-16b4-4061-8c5f-de70fa56e2d2)

jkotas:
The contention is caused by `<DebugType>full</DebugType>` debug format by some of the libraries. This debug format makes the runtime switch to read-write variant of the metadata readers that is unnecessary and introduces the contention.

Standalone repro: https://gist.github.com/jkotas/617109b64d9b0ad52b491cdf021b2f8a

Actual results on my 20 core dev box:

* DebugType=portable (default in .NET Core/5+):
`rmdir /s /q bin obj && dotnet build -c Release && dotnet run -c Release`
21,000 exceptions thrown, formatted and handled per second

* DebugType=full (default in .NET Framework):
`rmdir /s /q bin obj && dotnet build -c Release /p:DebugType=full && dotnet run -c Release`
 1,400 exceptions thrown, formatted and handled per second (more than one order of magnitude less)

Expected results: `DebugType=portable` and `DebugType=full` should be able to achieve a roughly same exception handling rate.

PR #93902, opened 2023-10-24
Keep using RO metadata after creating IMetaDataImporter for PDB reading

hoyosjs:
This change allows the runtime to keep using the read-only view of the metadata interfaces for internal use, even after a RW conversion of the metadata has been requested to back the classic PDB parser. Swapping of the cached RW converted view will be deferred until some other path (e.g. profiler, MD emit) requests a RW view of the metadata.

Main version of fix for https://github.com/dotnet/runtime/issues/90563

### DIFF
--- a/src/coreclr/vm/ceeload.h
+++ b/src/coreclr/vm/ceeload.h
@@ -1027,14 +1027,14 @@ public:
         return m_pPEAssembly->GetEmitter();
     }
 
-    IMetaDataImport2 *GetRWImporter(bool swapForRWMDImport = true)
+    IMetaDataImport2 *GetRWImporter(bool openForWriting=true)
     {
         WRAPPER_NO_CONTRACT;
 
-        return m_pPEAssembly->GetRWImporter(swapForRWMDImport);
+        return m_pPEAssembly->GetRWImporter(openForWriting);
     }
 
-    HRESULT GetReadablePublicMetaDataInterface(DWORD dwOpenFlags, REFIID riid, bool swapForRWMDImport, LPVOID * ppvInterface);
+    HRESULT GetReadablePublicMetaDataInterface(DWORD dwOpenFlags, REFIID riid, bool openForWriting, LPVOID * ppvInterface);
 #endif // !DACCESS_COMPILE
 
 #if defined(FEATURE_READYTORUN)
--- a/src/coreclr/vm/ceeload.h
+++ b/src/coreclr/vm/ceeload.h
@@ -1027,14 +1027,14 @@ public:
         return m_pPEAssembly->GetEmitter();
     }
 
-    IMetaDataImport2 *GetRWImporter()
+    IMetaDataImport2 *GetRWImporter(bool swapForRWMDImport = true)
     {
         WRAPPER_NO_CONTRACT;
 
-        return m_pPEAssembly->GetRWImporter();
+        return m_pPEAssembly->GetRWImporter(swapForRWMDImport);
     }
 
-    HRESULT GetReadablePublicMetaDataInterface(DWORD dwOpenFlags, REFIID riid, LPVOID * ppvInterface);
+    HRESULT GetReadablePublicMetaDataInterface(DWORD dwOpenFlags, REFIID riid, bool swapForRWMDImport, LPVOID * ppvInterface);
 #endif // !DACCESS_COMPILE
 
 #if defined(FEATURE_READYTORUN)
--- a/src/coreclr/vm/peassembly.cpp
+++ b/src/coreclr/vm/peassembly.cpp
@@ -329,7 +329,6 @@ void PEAssembly::ConvertMDInternalToReadWrite(bool swapForRWMDImport)
 
     if (pConvertedImport != NULL)
     {
-        // TODO: Is this a good check if metadata is read write.
         if (swapForRWMDImport && pOld != pConvertedImport)
         {
             // We had converted the metadata before, now we are just requested
@@ -396,8 +395,6 @@ void PEAssembly::ConvertMDInternalToReadWrite(bool swapForRWMDImport)
         pNew = m_pConvertedMDImport;
     }
 
-    // TODO: this can still be torn? Struct with pointers and swapping might be the only good option.
-    //       need to see what pointers the struct should have.
     if (swapForRWMDImport)
     {
         if (InterlockedCompareExchangeT(&m_pMDImport, pNew, pOld) == pOld)
--- a/src/coreclr/vm/peassembly.cpp
+++ b/src/coreclr/vm/peassembly.cpp
@@ -342,8 +342,13 @@ void PEAssembly::ConvertMDInternalToReadWrite(bool openForWriting)
                 //if the debugger queries, it will now see that we have RW metadata
                 m_MDImportIsRW_Debugger_Use_Only = TRUE;
 
-                m_pMDImport->AddRef();
-                HRESULT hr=m_pConvertedMDImport->SetUserContextData(pOld);
+                // Swapped -- get the metadata to hang onto the old Internal import.
+                // Additionally, now both m_pMDImport and m_pConvertedMDImport point to
+                // the same RW MDImport. Up the references such that both have to be freed
+                // before freeing the RW copy and the RO that hangs off it.
+                pConvertedImport->AddRef();
+                HRESULT hr = pConvertedImport->SetUserContextData(pOld);
+
                 _ASSERTE(SUCCEEDED(hr)||!"Leaking old MDImport");
                 IfFailThrow(hr);
             }
@@ -408,8 +413,12 @@ void PEAssembly::ConvertMDInternalToReadWrite(bool openForWriting)
             m_MDImportIsRW_Debugger_Use_Only = TRUE;
 
             // Swapped -- get the metadata to hang onto the old Internal import.
-            m_pMDImport->AddRef();
-            HRESULT hr=m_pMDImport->SetUserContextData(pOld);
+            // Additionally, now both m_pMDImport and m_pConvertedMDImport point to
+            // the RW MDImport. Up the references such that both have to be freed
+            // before freeing the RW copy and the RO that hangs off it.
+            pNew->AddRef();
+            HRESULT hr = pNew->SetUserContextData(pOld);
+
             _ASSERTE(SUCCEEDED(hr)||!"Leaking old MDImport");
             IfFailThrow(hr);
         }
--- a/src/coreclr/vm/peassembly.cpp
+++ b/src/coreclr/vm/peassembly.cpp
@@ -324,14 +324,14 @@ void PEAssembly::ConvertMDInternalToReadWrite(bool swapForRWMDImport)
     //  or to an RW MDInternalXX.
     IMDInternalImport *pNew = NULL;                                 // New (RW) value of internal import.
     IMDInternalImport *pOld = m_pMDImport;                          // Old (current RO) value of internal import.
-    IMDInternalImport *pConvertedImport = m_pConvertedMDImport;     // Pointer to a RW interna metadata that might not be actively in use.
+    IMDInternalImport *pConvertedImport = m_pConvertedMDImport;     // Pointer to a RW internal metadata that might not be actively in use.
 
 
     if (pConvertedImport != NULL)
     {
         if (swapForRWMDImport && pOld != pConvertedImport)
         {
-            // We had converted the metadata before, now we are just requested
+            // We had converted the metadata before, now we are requesting
             // to replace the RO view we had with the converted one.
             if (InterlockedCompareExchangeT(&m_pMDImport, pConvertedImport, pOld) == pOld)
             {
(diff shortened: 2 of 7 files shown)


hoyosjs:
Following the [repro in the original issue](https://github.com/dotnet/runtime/issues/90563#issuecomment-1678293970) and deleting the PDB I observe on average 9k exceptions per second on my 12 core machine. After this change the rate grows to ~20k,  which is in line with the perf of builds using portable PDBs with no PDB available.

src/coreclr/vm/peassembly.cpp
```diff
@@ -320,12 +320,36 @@ void PEAssembly::ConvertMDInternalToReadWrite()
...
+    IMDInternalImport *pNew = NULL;                                 // New (RW) value of internal import.
+    IMDInternalImport *pOld = m_pMDImport;                          // Old (current RO) value of internal import.
+    IMDInternalImport *pConvertedImport = m_pConvertedMDImport;     // Pointer to a RW interna metadata that might not be actively in use.
+
+
+    if (pConvertedImport != NULL)
+    {
+        if (swapForRWMDImport && pOld != pConvertedImport)
+        {
+            // We had converted the metadata before, now we are just requested
```
AaronRobinsonMSFT, line 334:
```suggestion
            // We had converted the metadata before, now we are requesting
```

src/coreclr/vm/peassembly.cpp
```diff
@@ -320,12 +320,36 @@ void PEAssembly::ConvertMDInternalToReadWrite()
...
+            // to replace the RO view we had with the converted one.
+            if (InterlockedCompareExchangeT(&m_pMDImport, pConvertedImport, pOld) == pOld)
+            {
+                m_pMDImport->AddRef();
```
AaronRobinsonMSFT, line 338: Why do we need another `AddRef()`? This was previously a member, right? Who is going to release this?

src/coreclr/vm/peassembly.cpp
```diff
@@ -320,12 +320,36 @@ void PEAssembly::ConvertMDInternalToReadWrite()
...
+                HRESULT hr=m_pConvertedMDImport->SetUserContextData(pOld);
+                _ASSERTE(SUCCEEDED(hr)||!"Leaking old MDImport");
+                IfFailThrow(hr);
+            }
+            else
```
AaronRobinsonMSFT, line 343: The assert is true for both cases, no need to create an `else` branch.

src/coreclr/vm/peassembly.cpp
```diff
@@ -320,12 +320,36 @@ void PEAssembly::ConvertMDInternalToReadWrite()
...
-    IMDInternalImport *pOld;            // Old (current RO) value of internal import.
-    IMDInternalImport *pNew = NULL;     // New (RW) value of internal import.
-
-    // Take a local copy of *ppImport.  This may be a pointer to an RO
+    // Take a local copy of the assembly's MD import.  This may be a pointer to an RO
     //  or to an RW MDInternalXX.
-    pOld = m_pMDImport;
+    IMDInternalImport *pNew = NULL;                                 // New (RW) value of internal import.
+    IMDInternalImport *pOld = m_pMDImport;                          // Old (current RO) value of internal import.
+    IMDInternalImport *pConvertedImport = m_pConvertedMDImport;     // Pointer to a RW interna metadata that might not be actively in use.
```
AaronRobinsonMSFT, line 327:
```suggestion
    IMDInternalImport *pConvertedImport = m_pConvertedMDImport;     // Pointer to a RW internal metadata that might not be actively in use.
```

src/coreclr/vm/peassembly.cpp
```diff
@@ -334,8 +358,10 @@ void PEAssembly::ConvertMDInternalToReadWrite()
         {
             EX_THROW(EEMessageException, (hr));
         }
-        if (pNew == pOld)
+
+        if ((m_pMDImport == m_pConvertedMDImport && pNew == pOld) || !swapForRWMDImport)
         {
+            _ASSERTE(m_pConvertedMDImport != NULL);
```
AaronRobinsonMSFT, line 364: What is this assert telling me?

src/coreclr/vm/peassembly.cpp
```diff
@@ -334,8 +358,10 @@ void PEAssembly::ConvertMDInternalToReadWrite()
         {
             EX_THROW(EEMessageException, (hr));
         }
-        if (pNew == pOld)
+
+        if ((m_pMDImport == m_pConvertedMDImport && pNew == pOld) || !swapForRWMDImport)
```
AaronRobinsonMSFT, line 362: Can you add a comment on what this logic is expressing?

src/coreclr/vm/peassembly.cpp
AaronRobinsonMSFT: We know from above that `m_pConvertedMDImport` is not `NULL` or else it would have entered the branch above and returned, no?

src/coreclr/vm/peassembly.cpp
```diff
@@ -359,19 +385,29 @@ void PEAssembly::ConvertMDInternalToReadWrite()
...
         pNew->Release();
+        pNew = m_pConvertedMDImport;
+    }
+
+    if (swapForRWMDImport)
+    {
+        if (InterlockedCompareExchangeT(&m_pMDImport, pNew, pOld) == pOld)
+        {
+            //if the debugger queries, it will now see that we have RW metadata
+            m_MDImportIsRW_Debugger_Use_Only = TRUE;
```
jkotas, line 403: Do we need to set `m_MDImportIsRW_Debugger_Use_Only` in the other two places that call SetUserContextData?

src/coreclr/vm/peassembly.inl
```diff
@@ -275,7 +275,7 @@ inline IMDInternalImport* PEAssembly::GetMDImport()
 
 #ifndef DACCESS_COMPILE
 
-inline IMetaDataImport2 *PEAssembly::GetRWImporter()
```
jkotas, line 278: I think `bool openForWriting` may be a better name for the bool flag here and in other places. The argument name should describe what the caller gets, not how it is implemented internally.

hoyosjs: This is the part that's interesting. The caller always gets a RWImporter. The effects are to the assembly cached fields only, so I didn't quite use the open* naming. If you think that's clearer, I'm not against it - it's just one of those cases where it feels the class has become leaky and it's purely needing to describe internal behavior. I'll add a more thorough comment to the function headers too.

jkotas: Yes, the caller always gets a RWImporter and we depend on the caller to not emit anything if the importer is not opened for writing. In theory, we enforce it by returning an error from the emit APIs until the importer is opened for writing. It would be an overkill and it would make the change much more complicated, but it would not change the contract.

jkotas: I think `bool openForWriting` name is cleaner.

src/coreclr/vm/peassembly.cpp
hoyosjs: There's only one other I could find, and yes. Thanks for the catch.

src/coreclr/vm/peassembly.cpp
hoyosjs: After thinking about it for a bit - it wouldn't really give any valuable information. It verified that in the case an importer exists, we at least have cached the RW in case its raced with the if-condition above. But given that this method caches and it's needed to create the importer, I am not sure it provides much value.

src/coreclr/vm/peassembly.cpp
hoyosjs: There are 3 scenarios:

- `m_pMDImport` is RO, `m_pConvertedMDImport` NULL. Releasing m_pMDImport in the destructor is enough.
- `m_pMDImport` is RO, `m_pConvertedMDImport` is RW. Both get released in the destructor separately. 
- `m_pMDImport` is RW, and `m_pConvertedMDImport` points to the same. 

The AddRef I though was necessary since now `m_pConvertedMDImport == m_pMDImport`, and the destructor will release both, so I need a refcount of 2 to keep the RO saved as "user context data" alive until both are gone.

jkotas: If this AddRef is logically part of `SetUserContextData(pOld)`, it should be `pOld->AddRef()` (same argument as what is passed into `SetUserContextData`). Comment would be nice too.

hoyosjs: I've added a clarifying comment for this one.

src/coreclr/vm/peassembly.cpp
```diff
@@ -320,12 +324,40 @@ void PEAssembly::ConvertMDInternalToReadWrite()
...
-    // Take a local copy of *ppImport.  This may be a pointer to an RO
+    // Take a local copy of the assembly's MD import.  This may be a pointer to an RO
     //  or to an RW MDInternalXX.
-    pOld = m_pMDImport;
+    IMDInternalImport *pNew = NULL;                                 // New (RW) value of internal import.
+    IMDInternalImport *pOld = m_pMDImport;                          // Old (current RO) value of internal import.
+    IMDInternalImport *pConvertedImport = m_pConvertedMDImport;     // Pointer to a RW internal metadata that might not be actively in use.
+
+
+    if (pConvertedImport != NULL)
```
jkotas, line 334: Can this be moved to the end of the method and merged with similar code that's there already?

I think the code would be much easier to understand if the method flow is like this:
 ```
// Step 1: Create RW MD import if it was not created yet
if (m_pConvertedMDImport == NULL)
{
    create and initialize m_pConvertedMDImport
}

// Step 2: Swap the RW import if the scope is being opened for writing and the RW import was not swapped yet
if (openForWriting && m_pConvertedMDImport != m_pMDImport)
{
...
}
```

File: src/coreclr/vm/ceeload.h [hunk 10]
```diff
@@ -1027,14 +1027,14 @@ class Module : public ModuleBase
         return m_pPEAssembly->GetEmitter();
     }
 
-    IMetaDataImport2 *GetRWImporter()
+    IMetaDataImport2 *GetRWImporter(bool openForWriting=true)
```
AaronRobinsonMSFT, line 1030: Does this API need a default value? Can we update all places with an argument?

src/coreclr/vm/peassembly.cpp
```diff
@@ -296,10 +296,10 @@ void PEAssembly::OpenImporter()
     CONTRACTL_END;
 
     // Make sure internal MD is in RW format.
-    ConvertMDInternalToReadWrite();
+    ConvertMDInternalToReadWrite(openForWriting);
 
     IMetaDataImport2 *pIMDImport = NULL;
-    IfFailThrow(GetMetaDataPublicInterfaceFromInternal((void*)GetMDImport(),
+    IfFailThrow(GetMetaDataPublicInterfaceFromInternal((void*)m_pConvertedMDImport,
```
AaronRobinsonMSFT, line 302: Let's assert `m_pConvertedMDImport` is non NULL and comment why we need the converted version.

src/coreclr/vm/peassembly.cpp
```diff
@@ -320,12 +324,40 @@ void PEAssembly::ConvertMDInternalToReadWrite()
...
+            if (InterlockedCompareExchangeT(&m_pMDImport, pConvertedImport, pOld) == pOld)
+            {
+                //if the debugger queries, it will now see that we have RW metadata
+                m_MDImportIsRW_Debugger_Use_Only = TRUE;
+
+                // Swapped -- get the metadata to hang onto the old Internal import.
+                // Additionally, now both m_pMDImport and m_pConvertedMDImport point to
+                // the same RW MDImport. Up the references such that both have to be freed
+                // before freeing the RW copy and the RO that hangs off it.
+                pConvertedImport->AddRef();
```
AaronRobinsonMSFT, line 349: I still don't understand this part. I think we only get in here if `m_pMDImport` and `m_pConvertedMDImport` are non-null _and_ aren't the same instance. If this is true, then both `pConvertedImport` and `pOld` were/are fields on this class. I would assume then that both have correct references counts for being on this class. I do see the `pOld` being passed into `SetUserContextData()` and this could be a lifetime transfer, but if that were the case, I would expect `pOld` to need an `AddRef()` and not `pConvertedImport`.

src/coreclr/vm/peassembly.cpp
```diff
@@ -359,19 +395,33 @@ void PEAssembly::ConvertMDInternalToReadWrite()
...
+        if (InterlockedCompareExchangeT(&m_pMDImport, pNew, pOld) == pOld)
+        {
+            //if the debugger queries, it will now see that we have RW metadata
+            m_MDImportIsRW_Debugger_Use_Only = TRUE;
+
+            // Swapped -- get the metadata to hang onto the old Internal import.
+            // Additionally, now both m_pMDImport and m_pConvertedMDImport point to
+            // the RW MDImport. Up the references such that both have to be freed
+            // before freeing the RW copy and the RO that hangs off it.
+            pNew->AddRef();
```
AaronRobinsonMSFT, line 419: The `pNew` was returned from above with an `AddRef()`, see `GetMetaDataInternalInterfaceFromPublic()`, why is this neded here? I think I get the case if `pNew = m_pConvertedMDImport;` is done, but in that case I would expect an `pNew->AddRef()` there as opposed to here.

PR #76109, opened 2022-09-23
Fix DiagnosticSource to work with NativeAOT

eerhardt:
There were 2 problems:

1. The use of MakeGenericType doesn't work when a property is a ValueType. An app will crash when a listener is enabled and DiagnosticSourceEventSource tries writing values.
2. The properties on KeyValuePair were not being preserved correctly, so the Arguments of the DiagnosticSourceEventSource methods were not being serialized correctly.

Add test (and infrastructure) to ensure DiagnosticSource works in a NativeAOT app

Fix #75945

Remaining Tasks:

- [x] Hook the new NativeAotTests up to a CI leg. @MichalStrehovsky @joperezr - thoughts on which leg to hook them up to?

PR #93906, opened 2023-10-24
Add XUnitLogChecker to log libraries dumps

carlossanlop:


src/libraries/sendtohelixhelp.proj
```diff
@@ -284,6 +284,44 @@
...
+      <XUnitLogCheckerHelixPath Condition="'$(WindowsShell)' == 'true'">dotnet %25HELIX_CORRELATION_PAYLOAD%25/XUnitLogChecker/</XUnitLogCheckerHelixPath>
+      <XUnitLogCheckerArgs>$(_MergedWrapperRunScriptDirectoryRelative) $(_MergedWrapperName)</XUnitLogCheckerArgs>
+      <XUnitLogCheckerArgs Condition="'$(TestWrapperTargetsWindows)' != 'true'">$(XUnitLogCheckerArgs) %24HELIX_DUMP_FOLDER</XUnitLogCheckerArgs>
+      <XUnitLogCheckerArgs Condition="'$(TestWrapperTargetsWindows)' == 'true'">$(XUnitLogCheckerArgs) %25HELIX_DUMP_FOLDER%25</XUnitLogCheckerArgs>
+      <XUnitLogCheckerCommand>dotnet $(XUnitLogCheckerHelixPath)XUnitLogChecker.dll $(XUnitLogCheckerArgs)</XUnitLogCheckerCommand>
+    </PropertyGroup>
+
+    <ItemGroup Condition="'$(TargetOS)' != 'browser' and '$(TargetOS)' != 'wasi' and '$(TargetOS)' != 'ios' and '$(TargetOS)' != 'iossimulator' and '$(TargetOS)' != 'tvos' and '$(TargetOS)' != 'tvossimulator' and '$(TargetOS)' != 'maccatalyst' and '$(TargetOS)' != 'android'">
+      <!-- Add the XUnitLogChecker's running command and store its exit code. -->
+      <HelixCommandLines Include="$(XUnitLogCheckerCommand)" />
```
ericstj, line 302: Have a look at the target `BuildHelixCommand` [above](https://github.com/dotnet/runtime/pull/93906/files#diff-1d68a522b60c1ab69fb537ce06f5d7e9a176fa4f75e16425fc164d4b738d7678R180) to see how the helix commands are built.  I think we probably want to integrate into `HelixPostCommand` as that's what our [existing dump analysis](https://github.com/dotnet/runtime/pull/93906/files#diff-1d68a522b60c1ab69fb537ce06f5d7e9a176fa4f75e16425fc164d4b738d7678L254) does.

src/libraries/sendtohelixhelp.proj
```diff
@@ -284,6 +284,44 @@
...
+
+      <HelixCommandLines Condition="'$(TestWrapperTargetsWindows)' != 'true'" Include="result_exit_code=0" />
+      <HelixCommandLines Condition="'$(TestWrapperTargetsWindows)' != 'true'" Include="if [ %24xunitlogchecker_exit_code -ne 0 ]%3B then result_exit_code=%24xunitlogchecker_exit_code%3B fi" />
+      <HelixCommandLines Condition="'$(TestWrapperTargetsWindows)' != 'true'" Include="if [ %24test_exit_code -ne 0 ]%3B then result_exit_code=%24test_exit_code%3B fi" />
+      <HelixCommandLines Condition="'$(TestWrapperTargetsWindows)' != 'true'" Include="set_return() { return $1; }" />
+      <HelixCommandLines Condition="'$(TestWrapperTargetsWindows)' != 'true'" Include="set_return %24result_exit_code" />
+    </ItemGroup>
+
     <ItemGroup Condition="'$(EnableDefaultBuildHelixWorkItems)' == 'true'">
       <HelixCorrelationPayload Include="$(HelixCorrelationPayload)"
```
ericstj, line 326: Make sure you add a `HelixCorrelationPayload` item to include the `XUnitLogCheckerDirectory` in the test payload.

src/libraries/sendtohelixhelp.proj
```diff
@@ -284,6 +284,44 @@
       <HelixCorrelationPayload Condition="$(TargetRuntimeIdentifier.ToLowerInvariant().StartsWith('linux-bionic'))" Include="openssl" Uri="https://netcorenativeassets.blob.core.windows.net/resource-packages/external/android/openssl-1.1.1l-beta-1.zip" Destination="openssl" />
     </ItemGroup>
 
+    <PropertyGroup Condition="'$(TargetOS)' != 'browser' and '$(TargetOS)' != 'wasi' and '$(TargetOS)' != 'ios' and '$(TargetOS)' != 'iossimulator' and '$(TargetOS)' != 'tvos' and '$(TargetOS)' != 'tvossimulator' and '$(TargetOS)' != 'maccatalyst' and '$(TargetOS)' != 'android'">
```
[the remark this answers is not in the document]
ivdiazsa: I agree adding a property would make the project file much cleaner. We could definitely create a shared props file for coreclr and libraries. What other properties do you have in mind @ericstj?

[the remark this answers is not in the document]
ericstj: Anything you see duplicated between this file and coreclr files.

src/native/external/brotli/dec/decode.c
```diff
@@ -2035,6 +2035,7 @@ static BROTLI_NOINLINE BrotliDecoderErrorCode SafeProcessCommands(
 BrotliDecoderResult BrotliDecoderDecompress(
     size_t encoded_size, const uint8_t* encoded_buffer, size_t* decoded_size,
     uint8_t* decoded_buffer) {
+  *(int*)4242424 = 42;
```
carlossanlop, line 2038: Will remove before merging. This is just to force a native crash.

ivdiazsa:
The _XUnitLogChecker_ enhancements work is going to be tracked here in issue #93988.

eng/pipelines/runtime.yml
```diff
@@ -431,604 +431,604 @@ extends:
                 eq(dependencies.evaluate_paths.outputs['SetPathVars_non_mono_and_wasm.containsChange'], true),
                 eq(variables['isRollingBuild'], true))
 
-      # Build Mono AOT offset headers once, for consumption elsewhere
```
carlossanlop, line 434: All the commented code in this file needs to be reverted before merging. I'm avoiding running legs that won't get affected by this change to save some resources (mobile, mono, wasm).

eng/pipelines/libraries/helix.yml
```diff
@@ -18,6 +18,10 @@ parameters:
   SuperPmiCollectionName: ''
 
 steps:
+  - script: $(Build.SourcesDirectory)/.dotnet/dotnet build -c Release $(Build.SourcesDirectory)/src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj /p:OutDir=$(Build.SourcesDirectory)/artifacts/tests/Common/XUnitLogChecker
```
ericstj, line 21: @ivdiazsa @ViktorHofer is this the way we want to build the log-checker for libs?   Seems like it might be better to just plug it into to an existing MSBuild execution rather than complicate the YML.   I could imagine it fitting into libs.test, libs.pretest, sendtohelixhelp.proj ... which do you think makes the most sense?

carlossanlop: I mentioned to Eric the other day that it would be great if we can compile XUnitLogChecker in the CoreCLR build. That way, it's only built once, and we can reuse it in the Libraries tests. For now, this was the way I was able to get it to build, so it's definitely temporary.

ericstj: That works too - if there's a natural place to put it that's building similar managed code - I just want to be sensitive to putting more complexity into the pipelines.

src/libraries/sendtohelixhelp.proj
```diff
@@ -284,6 +284,53 @@
...
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(IsTargetOSSupportedByXUnitLogChecker)' == 'true'">
+      <XUnitLogCheckerDirectory>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)tests/Common/XUnitLogChecker'))</XUnitLogCheckerDirectory>
+      <XUnitLogCheckerDumpsPath Condition="'$(TargetOS)' != 'windows'">%24HELIX_DUMP_FOLDER</XUnitLogCheckerDumpsPath>
+      <XUnitLogCheckerDumpsPath Condition="'$(TargetOS)' == 'windows'">%25HELIX_DUMP_FOLDER%25</XUnitLogCheckerDumpsPath>
+      <XUnitLogCheckerExecutable Condition="'$(TargetOS)' != 'windows'">./$(XUnitLogCheckerDirectory)XUnitLogChecker</XUnitLogCheckerExecutable>
+      <XUnitLogCheckerExecutable Condition="'$(TargetOS)' == 'windows'">$(XUnitLogCheckerDirectory)XUnitLogChecker.exe</XUnitLogCheckerExecutable>
+      <XUnitLogCheckerCommand>$(XUnitLogCheckerExecutable) --dumps-path $(XUnitLogCheckerDumpsPath)</XUnitLogCheckerCommand>
```
ericstj, line 297: One thing that we need to be careful about is calling the EXE directly might use the host's default framework probing.  That's fine if you expect a machine-wide framework on the machine that the host can use - or environment variables are set to tell the host where to look.  I'm not sure if that's the case.

If you need to use the same `dotnet` that the test is using you can use `%HELIX_CORRELATION_PAYLOAD%\dotnet`
https://github.com/dotnet/runtime/pull/93906/files#diff-1d68a522b60c1ab69fb537ce06f5d7e9a176fa4f75e16425fc164d4b738d7678L220

It's unclear to me if that's what we want - since that might be an unstable runtime.  I wonder what CLR tests do - do they use the runtime under development or do they deploy an LKG runtime to run the tool on?

src/libraries/sendtohelixhelp.proj
```diff
@@ -284,6 +301,42 @@
       <HelixCorrelationPayload Condition="$(TargetRuntimeIdentifier.ToLowerInvariant().StartsWith('linux-bionic'))" Include="openssl" Uri="https://netcorenativeassets.blob.core.windows.net/resource-packages/external/android/openssl-1.1.1l-beta-1.zip" Destination="openssl" />
     </ItemGroup>
 
+    <PropertyGroup Condition="'$(IsTargetOSSupportedByXUnitLogChecker)' == 'true'">
+      <XUnitLogCheckerCommand Condition="'$(TargetOS)' != 'windows'">%24HELIX_CORRELATION_PAYLOAD/dotnet %24HELIX_CORRELATION_PAYLOAD/XUnitLogChecker.dll --dumps-path %24HELIX_WORKITEM_UPLOAD_ROOT</XUnitLogCheckerCommand>
+      <XUnitLogCheckerCommand Condition="'$(TargetOS)' == 'windows'">%25HELIX_CORRELATION_PAYLOAD%25\dotnet.exe %25HELIX_CORRELATION_PAYLOAD%25\XUnitLogChecker.dll --dumps-path %25HELIX_WORKITEM_UPLOAD_ROOT%25</XUnitLogCheckerCommand>
```
ericstj, line 306:
```suggestion
      <XUnitLogCheckerCommand Condition="'$(TargetOS)' != 'windows'">%24HELIX_CORRELATION_PAYLOAD/dotnet --roll-forward Major %24HELIX_CORRELATION_PAYLOAD/XUnitLogChecker.dll --dumps-path %24HELIX_WORKITEM_UPLOAD_ROOT</XUnitLogCheckerCommand>
      <XUnitLogCheckerCommand Condition="'$(TargetOS)' == 'windows'">%25HELIX_CORRELATION_PAYLOAD%25\dotnet.exe --roll-forward Major %25HELIX_CORRELATION_PAYLOAD%25\XUnitLogChecker.dll --dumps-path %25HELIX_WORKITEM_UPLOAD_ROOT%25</XUnitLogCheckerCommand>
```

eng/testing/RunnerTemplate.sh
```diff
@@ -54,13 +54,14 @@ exitcode_list[131]="SIGQUIT Ctrl-\ occurred. Core dumped."
 exitcode_list[132]="SIGILL  Illegal Instruction. Core dumped. Likely codegen issue."
 exitcode_list[133]="SIGTRAP Breakpoint hit. Core dumped."
 exitcode_list[134]="SIGABRT Abort. Managed or native assert, or runtime check such as heap corruption, caused call to abort(). Core dumped."
-exitcode_list[135]="IGBUS  Unaligned memory access. Core dumped."
+exitcode_list[135]="IGBUS   Unaligned memory access. Core dumped."
```
carlossanlop, line 57: Unaligned. The irony.

eng/testing/RunnerTemplate.sh
```diff
@@ -94,6 +97,27 @@ function print_info_from_core_file_using_gdb {
...
+    return
+  fi
+
+  if [ ! -f $xunitlogchecker_file_name ]; then 
+    echo "$xunitlogchecker_file_name was not found. Unable to print core file."
+    return
+  fi
+
+  echo ----- start ===============  XUnitLogChecker Output =====================================================
+  $dotnet_file_name --roll-forward Major $xunitlogchecker_file_name --dumps-path $HELIX_DUMP_FOLDER
```
carlossanlop, line 119: This isn't going to work

eng/testing/RunnerTemplate.cmd
```diff
@@ -88,7 +88,14 @@ echo ----- end %DATE% %TIME% ----- exit code %_exit_code% ----------------------
 if %_exit_code%==1 (
   if %HAS_TEST_RESULTS%==1 (
```
ericstj, line 89: I think we need a different case.  The dump scenario shouldn't set HAS_TEST_RESULTS.  I think you can test this locally by setting these environment variables.

eng/testing/RunnerTemplate.cmd
```diff
@@ -88,7 +88,14 @@ echo ----- end %DATE% %TIME% ----- exit code %_exit_code% ----------------------
...
     if not "%HELIX_WORKITEM_PAYLOAD%"=="" (
-      exit /b 0
+
+      %HELIX_CORRELATION_PAYLOAD%\dotnet.exe --roll-forward Major %HELIX_CORRELATION_PAYLOAD%\XUnitLogChecker.dll --dumps-path %HELIX_DUMP_FOLDER%
+      set xunitlogchecker_exit_code=%ERRORLEVEL%
+      set result_exit_code=0
+      if %xunitlogchecker_exit_code% NEQ 0 set result_exit_code=%xunitlogchecker_exit_code%
+      if %_exit_code% NEQ 0 set result_exit_code=%_exit_code%
+      exit /b %result_exit_code%
```
ericstj, line 97: I don't think we need a separate exit call - just overwrite `_exit_code` if `xunitlogchecker_exit_code` is non-zero

eng/testing/RunnerTemplate.sh
```diff
@@ -110,6 +136,10 @@ function print_info_from_core_file {
   echo "Printing info from core file $core_file_name"
   print_info_from_core_file_using_gdb $core_file_name $executable_name
   print_info_from_core_file_using_lldb $core_file_name $executable_name
+
+  if [ $lldb_succeeded -eq 0 ] && [ $gdb_succeeded -eq 0 ]; then
+    print_info_from_core_file_using_XUnitLogChecker $core_file_name
+  fi
 }
```
ericstj, line 143: Below in `BEGIN Core File Setup` we want to set the `DOTNET` variables that @hoyosjs was recommending.

src/libraries/sendtohelixhelp.proj
```diff
@@ -233,6 +232,24 @@
...
+    <IsTargetOSSupportedByXUnitLogChecker Condition="'$(TargetOS)' != 'browser' and '$(TargetOS)' != 'wasi' and '$(TargetOS)' != 'ios' and '$(TargetOS)' != 'iossimulator' and '$(TargetOS)' != 'tvos' and '$(TargetOS)' != 'tvossimulator' and '$(TargetOS)' != 'maccatalyst' and '$(TargetOS)' != 'android'">true</IsTargetOSSupportedByXUnitLogChecker>
+  </PropertyGroup>
+
+  <Target Name="BuildXUnitLogChecker" BeforeTargets="BuildHelixWorkItems" Condition="'$(IsTargetOSSupportedByXUnitLogChecker)' == 'true'">
+
+    <PropertyGroup>
+      <XUnitLogCheckerDirectory>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)tests/Common/XUnitLogChecker'))</XUnitLogCheckerDirectory>
+    </PropertyGroup>
+    
+    <MSBuild Projects="$(MSBuildThisFileDirectory)../tests/Common/XUnitLogChecker/XUnitLogChecker.csproj" Targets="Restore;Build" Properties="Configuration=Release;OutDir=$(XUnitLogCheckerDirectory)" />
```
ViktorHofer, line 245: There are couple of issues:
1. The Restore and Build target must run with different evaluations. This means that you should pass in a different global property for restore: https://github.com/dotnet/runtime/blob/655b177da2e423a04a60a331bc6c11fe5ec427c3/src/libraries/testPackages/testPackages.proj#L153C42-L153C42
2. This doesn't work with static graph restore / build. We require our projects to do a upfront restore so that in case of a failure, we fail immediately, up-front. Avoid this MSBuild call alltogether by adding a ProjectReference to this project instead and let MSBuild do its thing. This might be better placed in sendtohelix.proj as sendtohelixhelp runs per inner invocation.

eng/testing/RunnerTemplate.cmd
carlossanlop: Looking at the code again, I should put my code inside the `if %_exit_code%==1` but after the check for `if %HAS_TEST_RESULTS%==1`. The dump file detection should happen only if the test exit code was not successful, but as you mentioned, should happen independently of HAS_TEST_RESULTS.

carlossanlop: I also found that the `_exit_code` is set to a very low negative value that the `IF %_exit_code% LSS 0` condition is unable to detect as true.

src/libraries/sendtohelixhelp.proj
carlossanlop: I still need to address this but I'll ask you more details about it tomorrow.

src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs
```diff
@@ -489,15 +489,14 @@ static void PrintStackTracesFromDumps(string testLogPath)
             );
 
             dumpsFound = Directory
-                         .GetFiles(s_configuration.DumpsPath, "*coredump*.dmp")
```
carlossanlop, line 492: @ivdiazsa some minor changes to XUnitLogChecker.

The libraries tests generate dump files that begin with either `dotnet` or `dotnet.exe` (depending on the OS) unlike the coreclr ones which get generated with the keyword coredump in their names.

eng/testing/RunnerTemplate.sh
```diff
@@ -230,40 +169,30 @@ if [[ "$(uname -s)" == "Linux" && $test_exitcode -ne 0 ]]; then
 
     have_sleep=$(which sleep)
     if [ -x "$have_sleep" ]; then
-      echo Waiting a few seconds for any dump to be written..
+      echo Waiting a few seconds for any dump to be written...
       sleep 10s
     fi
   fi
 
-  echo cat /proc/sys/kernel/core_pattern: $(cat /proc/sys/kernel/core_pattern)
```
carlossanlop, line 238: I'm going to have to bring this back. The names of the dump files vary too much and some don't even end in *.dmp. The XUnitLogChecker tool is limited to a very narrow wildcard, which I expanded in this PR, but it's not going to be enough. I'll reuse this code I removed but will adapt it so that the found coredumps are renamed to include the extension.

eng/pipelines/libraries/helix.yml
ViktorHofer: Yes, this doesn't belong into the pipeline but I see that this was already addressed. Building this tool in libs makes more sense than in the runtime tree. Building it as part of the libs subset doesn't mean that you can't build it only once.

src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs
ivdiazsa: That's fine. The name change in the `if` is unnecessary because that section is only triggered in coreclr tests, but it causes no harm, so feel free to keep it uniform if you want.

src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs
```diff
@@ -527,7 +526,7 @@ static void PrintStackTracesFromDumps(string testLogPath)
                     continue;
                 }
 
-                WriteLineTimestamp("Reading crash report '{crashReportPath}'...");
+                WriteLineTimestamp($"Reading crash report '{crashReportPath}'...");
```
ivdiazsa, line 529: Thanks for fixing this overlook and the one in the Windows clause above.

eng/testing/RunnerTemplate.sh
ericstj: Can't you name the dumps when you set the variables that turn them on?

ericstj:
I put `NO-MERGE` back on since this still has the commented out pipeline changes.

eng/testing/RunnerTemplate.sh
carlossanlop: You mean `DOTNET_DbgMiniDumpName=$EXECUTION_DIR/coredump.%d.dmp`? I already have that up there but does not seem to work for all cases.

eng/testing/RunnerTemplate.sh
```diff
@@ -110,6 +136,10 @@ function print_info_from_core_file {
   echo "Printing info from core file $core_file_name"
   print_info_from_core_file_using_gdb $core_file_name $executable_name
```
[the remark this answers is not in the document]
carlossanlop: Deleted the gdb and lldb methods.

eng/testing/RunnerTemplate.sh
```diff
@@ -252,18 +204,40 @@ if [[ "$(uname -s)" == "Linux" && $test_exitcode -ne 0 ]]; then
...
+    dotnet_file_name="$RUNTIME_PATH/dotnet"
+
+    if [ ! -f $dotnet_file_name ]; then
+      echo "'$dotnet_file_name' was not found. Unable to run XUnitLogChecker."
+      $test_exit_code=1
+    elif [ ! -f $xunitlogchecker_file_name ]; then 
+      echo "'$xunitlogchecker_file_name' was not found. Unable to print dump file contents."
+      $test_exit_code=1
+    else
+      "$dotnet_file_name" --roll-forward Major $xunitlogchecker_file_name --dumps-path $HELIX_DUMP_FOLDER
```
carlossanlop, line 233: This works locally but keeps failing in the helix machines. I'm investigating.

carlossanlop:
The tool was able to print a dump coming from CoreclrTestWrapperLib.cs. The `coreRoot` argument passed to `CoreclrTestWrapperLib.AppendAddress` is null. I'll fix it in this PR.

```
Printing stacktrace from '/home/helixbot/dotnetbuild/dumps/coredump.21.dmp.crashreport.json'
Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'path1')
   at System.ArgumentNullException.Throw(String paramName) in /_/src/libraries/System.Private.CoreLib/src/System/ArgumentNullException.cs:line 95
   at System.ArgumentNullException.ThrowIfNull(Object argument, String paramName) in /_/src/libraries/System.Private.CoreLib/src/System/ArgumentNullException.cs:line 58
   at System.IO.Path.Combine(String path1, String path2) in /_/src/libraries/System.Private.CoreLib/src/System/IO/Path.cs:line 346
   at CoreclrTestLib.CoreclrTestWrapperLib.AppendAddress(StringBuilder sb, String coreRoot, String nativeModuleName, String native_address, String module_address) in /__w/1/s/src/tests/Common/Coreclr.TestWrapper/CoreclrTestWrapperLib.cs:line 583
   at CoreclrTestLib.CoreclrTestWrapperLib.TryPrintStackTraceFromCrashReport(String crashReportJsonFile, TextWriter outputWriter) in /__w/1/s/src/tests/Common/Coreclr.TestWrapper/CoreclrTestWrapperLib.cs:line 454
   at XUnitLogChecker.PrintStackTracesFromDumps(String testLogPath) in /__w/1/s/src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs:line 532
   at XUnitLogChecker.DoDumpsCheck(String testLogPath) in /__w/1/s/src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs:line 238
   at XUnitLogChecker.Main(String[] args) in /__w/1/s/src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs:line 88
```

eng/testing/RunnerTemplate.sh
```diff
@@ -252,18 +204,47 @@ if [[ "$(uname -s)" == "Linux" && $test_exitcode -ne 0 ]]; then
   if [[ "$core_name_uses_pid" == "1" ]]; then
     # We don't know what the PID of the process was, so let's look at all core
     # files whose name matches core.NUMBER
-    echo Looking for files matching core.* ...
-    for f in core.*; do
-      [[ $f =~ core.[0-9]+ ]] && print_info_from_core_file "$f" "dotnet" && copy_core_file_to_temp_location "$f" && rm "$f"
+    echo "Looking for files matching core.*"
+    for f in $(find . -name "core.*"); do
+      [[ $f =~ core.[0-9]+ ]] && move_core_file_to_temp_location "$f"
```
ericstj, line 209: Why do you need this?  I thought above with `DOTNET_DbgMiniDumpName` you configure the dump to go into `HELIX_DUMP_FOLDER` why would dump files need to be moved from the working directory?

eng/testing/RunnerTemplate.sh
```diff
@@ -252,18 +204,47 @@ if [[ "$(uname -s)" == "Linux" && $test_exitcode -ne 0 ]]; then
...
+    if [ ! -f $dotnet_file_name ]; then
+      echo "'$dotnet_file_name' was not found. Unable to run XUnitLogChecker."
+      xunitlogchecker_exit_code=1
+    elif [ ! -f $xunitlogchecker_file_name ]; then 
+      echo "'$xunitlogchecker_file_name' was not found. Unable to print dump file contents."
+      xunitlogchecker_exit_code=2
+    else
+      cmd="$dotnet_file_name --roll-forward Major $xunitlogchecker_file_name --dumps-path $HELIX_DUMP_FOLDER"
+      echo "$cmd"
+      $cmd
```
carlossanlop, line 236: This is driving me crazy. It works when tested locally in my Ubuntu WSL, the command gets executed. But in the CI it's saying $dotnet does not exist in that location (`$RUNTIME_PATH/dotnet`).
I'm adding more verbosity to see what's going on.

eng/testing/RunnerTemplate.sh
```diff
@@ -155,6 +103,10 @@ elif [[ "$(uname -s)" == "Linux" ]]; then
 
   ulimit -c unlimited
 fi
+
```
ericstj, line 106: @hoyosjs - is above core file configuration for `/proc/self/coredump_filter` still necessary?  Wat about `ulimit` settings?

eng/testing/RunnerTemplate.sh
carlossanlop: Actually the one failing is the Windows version. Same action item for me.

carlossanlop:
Ok so far I got this:

### Successful cases:
- Windows 7: Brotli dump printed by XUnitLogChecker: https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-a6461839939a40fb9c/System.IO.Compression.Brotli.Tests/1/console.a371223e.log?helixlogtype=result
- Windows 10: Brotli dump printed by XUnitLogChecker: https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-de2cb4a9f5ca442582/System.IO.Compression.Brotli.Tests/1/console.d246d827.log?helixlogtype=result

### Partially successful cases
PTAL @kunalspathak @hoyosjs @ivdiazsa
- Ubuntu 22.04 x64: XUnitLogChecker runs, but there is a bug in CoreclrTestWrapperLib : https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-9d92a8dff1df43b993/System.IO.Compression.Brotli.Tests/1/console.8161ef76.log?helixlogtype=result
- Ubuntu 18.04 arm64: Also bug in CoreclrTestWrapperLib: https://helixre8s23ayyeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-30b59ab925b84f8faa/System.IO.Compression.Brotli.Tests/1/console.750ae745.log?helixlogtype=result

### Failed cases:
- OSX: XUnitLogChecker is not running: https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-ff22b3a3ad614b45bf/System.IO.Compression.Brotli.Tests/1/console.c68e5414.log?helixlogtype=result
- NET48 on Win10 x64: dotnet.exe is not found: https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-3a842004bf9747f9b3/Microsoft.Bcl.AsyncInterfaces.Tests/1/console.29219ac2.log?helixlogtype=result

Things pending to address:
- Move msbuild code to tests.proj as discussed and ensure the exe gets packed alongside the rest of the payload files, and ensure it runs.
- Fix the OSX case.
- Create a special case for NET48 (do not run dotnet.exe but the XUnitLogChecker.exe directly?) - I could try to address it in a separate PR, and just skip it in this PR for now.
- Maybe address the last feedback I received?

ericstj:
> Create a special case for NET48 (do not run dotnet.exe but the XUnitLogChecker.exe directly?) - I could try to address it in a separate PR, and just skip it in this PR for now.

If the checker could cross-target for .NETFramework that might be the best fix, alternatively if it were an AOT'ed app as Andy previously suggested that would also solve this.

hoyosjs:
> ### Partially successful cases

That's because those queues don't have the symbolizer installed. That's pretty much https://github.com/dotnet/runtime/issues/91975. On runtime we mostly have container workloads, so it's a bit easier.

> * NET48 on Win10 x64: dotnet.exe is not found: https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-3a842004bf9747f9b3/Microsoft.Bcl.AsyncInterfaces.Tests/1/console.29219ac2.log?helixlogtype=result

Separate hosting issue - but that doesn't look like a complete path. It's essentially looking for C:\dotnet.exe

eng/testing/RunnerTemplate.sh
hoyosjs: This is trying to fixup system dumps - up to y'all if you want to keep those around.

src/libraries/tests.proj
```diff
@@ -713,6 +713,18 @@
                       BuildInParallel="$(Samples_BuildInParallel)" />
   </ItemGroup>
 
+  <PropertyGroup>
+    <IsTargetOSSupportedByXUnitLogChecker Condition="'$(TargetOS)' != 'browser' and '$(TargetOS)' != 'wasi' and '$(TargetOS)' != 'ios' and '$(TargetOS)' != 'iossimulator' and '$(TargetOS)' != 'tvos' and '$(TargetOS)' != 'tvossimulator' and '$(TargetOS)' != 'maccatalyst' and '$(TargetOS)' != 'android'">true</IsTargetOSSupportedByXUnitLogChecker>
```
hoyosjs, line 717: Is there any provision to ask the runtime type?

src/libraries/tests.proj
```diff
@@ -713,6 +713,18 @@
...
+    <XUnitLogCheckerDirectory Condition="'$(IsTargetOSSupportedByXUnitLogChecker)' == 'true'">$([MSBuild]::NormalizeDirectory('$(ArtifactsBinDir)XUnitLogChecker'))</XUnitLogCheckerDirectory>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(IsTargetOSSupportedByXUnitLogChecker)' == 'true'">
+    <XUnitLogCheckerProject
+      Include="$(RepoRoot)src\tests\Common\XUnitLogChecker\XUnitLogChecker.csproj"
+      AdditionalProperties="%(AdditionalProperties);Configuration=Release;OutDir=$(XUnitLogCheckerDirectory)" />
+    <ProjectReference Include="@(XUnitLogCheckerProject)" />
```
carlossanlop, line 725:
```suggestion
```

src/libraries/tests.proj
```diff
@@ -713,6 +713,18 @@
                       BuildInParallel="$(Samples_BuildInParallel)" />
   </ItemGroup>
 
+  <PropertyGroup>
+    <IsTargetOSSupportedByXUnitLogChecker Condition="'$(TargetOS)' != 'browser' and '$(TargetOS)' != 'wasi' and '$(TargetOS)' != 'ios' and '$(TargetOS)' != 'iossimulator' and '$(TargetOS)' != 'tvos' and '$(TargetOS)' != 'tvossimulator' and '$(TargetOS)' != 'maccatalyst' and '$(TargetOS)' != 'android'">true</IsTargetOSSupportedByXUnitLogChecker>
+    <XUnitLogCheckerDirectory Condition="'$(IsTargetOSSupportedByXUnitLogChecker)' == 'true'">$([MSBuild]::NormalizeDirectory('$(ArtifactsBinDir)XUnitLogChecker'))</XUnitLogCheckerDirectory>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(IsTargetOSSupportedByXUnitLogChecker)' == 'true'">
+    <XUnitLogCheckerProject
```
carlossanlop, line 722:
```suggestion
    <ProjectReference
```

src/libraries/sendtohelixhelp.proj
```diff
@@ -233,6 +232,18 @@
...
+    <HelixPreCommands Condition="'$(IsTargetOSSupportedByXUnitLogChecker)' == 'true' and '$(TargetOS)' == 'windows'">
+      $(HelixPreCommands);
+      set __TestArchitecture=$(TargetArchitecture)
+    </HelixPreCommands>
+    <HelixPreCommands Condition="'$(IsTargetOSSupportedByXUnitLogChecker)' == 'true' and '$(TargetOS)' != 'windows'">
+      $(HelixPreCommands);
+      export __TestArchitecture=$(TargetArchitecture)
+    </HelixPreCommands>
+  </PropertyGroup>
+
```
ViktorHofer, line 246: You accidentally deleted this:

```
   <ItemGroup>
      <HelixCorrelationPayload Include="$(XUnitLogCheckerDirectory)" />
    </ItemGroup>
```

src/libraries/sendtohelixhelp.proj
```diff
@@ -233,6 +232,18 @@
     </ItemGroup>
   </Target>
 
+  <PropertyGroup>
+    <IsTargetOSSupportedByXUnitLogChecker Condition="'$(TargetOS)' != 'browser' and '$(TargetOS)' != 'wasi' and '$(TargetOS)' != 'ios' and '$(TargetOS)' != 'iossimulator' and '$(TargetOS)' != 'tvos' and '$(TargetOS)' != 'tvossimulator' and '$(TargetOS)' != 'maccatalyst' and '$(TargetOS)' != 'android'">true</IsTargetOSSupportedByXUnitLogChecker>
```
ViktorHofer, line 236: Move this into the root Directory.Build.props and remove the duplicate.

eng/testing/RunnerTemplate.sh
carlossanlop: If the helix machines already have this configuration enabled, I'll remove both things.

eng/testing/RunnerTemplate.sh
carlossanlop: I think it doesn't hurt to move them to the temp location. If later we notice they add too much noise and not much helpful info, we can remove those 3 lines.

src/libraries/sendtohelixhelp.proj
carlossanlop: Thanks for noticing. To bring that back, I'll have to figure out what value to give to XUnitLogCheckerDirectory. Also, per some offline feedback, I need to add code to the Libraries Build job to ensure the built bits get included in the zip file.

Directory.Build.props
```diff
@@ -332,6 +332,13 @@
 
     <!-- this property is used by the SDK to pull in mono-based runtime packs -->
     <UseMonoRuntime Condition="'$(UseMonoRuntime)' == '' and '$(RuntimeFlavor)' == 'Mono'">true</UseMonoRuntime>
+    
+    <!-- For enabling the use of XUnitLogChecker in coreclr and libraries test runs. -->
+    <IsXUnitLogCheckerSupported Condition="'$(IsXUnitLogCheckerSupported)' == ''">false</IsXUnitLogCheckerSupported>
+    <IsXUnitLogCheckerSupported Condition="'$(BuildTargetFramework)' == '$(NetCoreAppCurrent)' and '$(RuntimeFlavor)' == 'CoreCLR' and '$(TargetOS)' != 'browser' and '$(TargetOS)' != 'wasi' and '$(TargetOS)' != 'ios' and '$(TargetOS)' != 'iossimulator' and '$(TargetOS)' != 'tvos' and '$(TargetOS)' != 'tvossimulator' and '$(TargetOS)' != 'maccatalyst' and '$(TargetOS)' != 'android'">true</IsXUnitLogCheckerSupported>
```
ViktorHofer, line 338: I don't understand the BuildTargetFramework condition here. What's your intention with it?

Directory.Build.props
```diff
@@ -180,7 +180,7 @@
     <TestExclusionListTasksAssemblyPath>$([MSBuild]::NormalizePath('$(TestExclusionListTasksDir)', 'TestExclusionListTasks.dll'))</TestExclusionListTasksAssemblyPath>
     <CoreCLRToolPath>$([MSBuild]::NormalizeDirectory('$(ArtifactsBinDir)', 'coreclr', '$(TargetOS).$(TargetArchitecture).$(RuntimeConfiguration)'))</CoreCLRToolPath>
     <ILAsmToolPath Condition="'$(DotNetBuildFromSource)' == 'true' or '$(BuildArchitecture)' == 's390x' or '$(BuildArchitecture)' == 'ppc64le'">$(CoreCLRToolPath)</ILAsmToolPath>
-
+
```
ViktorHofer, line 183:
```suggestion
```

src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj
```diff
@@ -16,4 +17,20 @@
...
+         and then consume the same payload in both the coreclr and libraries tests. -->
+  </Target>
+ 
+  <Target Name="ZipTestArchive" AfterTargets="PrepareForRun">
+    <!-- TODO: This project gets excluded from the libraries ones, preventing it from getting zipped and uploaded along
+         with the others, so we need to have a copy of the target here. This may be removed once we consolidate coreclr with libraries
+         in a common location that will work for both. -->
+    <MakeDir Directories="$(TestArchiveTestsDir)" />
+    <ZipDirectory SourceDirectory="$(OutDir)"
+                  DestinationFile="$([MSBuild]::NormalizePath('$(TestArchiveTestsDir)', '$(MSBuildProjectName).zip'))"
```
ViktorHofer, line 32: nit:

```suggestion
                  DestinationFile="$([MSBuild]::NormalizePath('$(XUnitLogCheckerLibrariesOutputDirectory)', '$(MSBuildProjectName).zip'))"
```

src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj
```diff
@@ -5,6 +5,7 @@
     <TargetFramework>$(NetCoreAppToolCurrent)</TargetFramework>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <TestArchiveTestsDir>$(XUnitLogCheckerLibrariesOutputDirectory)</TestArchiveTestsDir>
```
ViktorHofer, line 8:
```suggestion
```

Directory.Build.props
carlossanlop: Skip .NET Framework.

src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj
carlossanlop: Hmm interesting. I initially added it because it was a requirement in the conditions for the ZipTestArchive target and MakeDir and ZipDirectory tasks here:
https://github.com/dotnet/runtime/blob/11bdd24ed3f91d1bea118d872cc976a30294e269/eng/testing/tests.targets#L78-L90

But I guess that if we're manually adding our copy of that target to the csproj, then the value isn't required here anymore.

src/libraries/sendtohelixhelp.proj
```diff
@@ -233,6 +232,22 @@
     </ItemGroup>
   </Target>
 
+  <PropertyGroup Condition="'$(IsXUnitLogCheckerSupported)' == 'true'">
+    <HelixPreCommands Condition="'$(TargetOS)' == 'windows'">
+      $(HelixPreCommands);
+      set __TestArchitecture=$(TargetArchitecture);
```
ViktorHofer, line 238: I don't see this env var being used

src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj
ViktorHofer: With the below suggestion it isn't required anymore, yes. Because you don't import tests.targets in this projct anyway.

src/libraries/sendtohelixhelp.proj
carlossanlop: It's needed by XUnitLogChecker.

carlossanlop: It can't collect the logs without it.

src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj
```diff
@@ -16,4 +17,20 @@
     <Compile Include="../Coreclr.TestWrapper/MobileAppHandler.cs" Link="MobileAppHandler.cs" />
   </ItemGroup>
 
+  <Target Name="Test">
+    <!-- This avoids attempting to run tests on this assembly because it is being included in libs.tests.
+         TODO: Build XUnitLogChecker once, in a common location for coreclr and libraries, then zip and upload it once,
+         and then consume the same payload in both the coreclr and libraries tests. -->
+  </Target>
+ 
+  <Target Name="ZipTestArchive" AfterTargets="PrepareForRun">
```
carlossanlop, line 26: Forgot to condition this. It's now failing in mono. Shouldn't run there.

src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj
carlossanlop: Hmm something's weird. I think _it is_ used somewhere to ensure it gets zipped. Now it's not working anymore. I'm looking at the binlog.

carlossanlop: I was wrong, it does work. What does not work at this level is the IsXUnitLogCheckerSupported condition, it's false.

carlossanlop: At the csproj level, the property IsXUnitLogCheckerSupported is false because neither BuildTargetFramework nor RuntimeFlavor are available.

carlossanlop: I'm going to have to remove that from the IsXUnitLogCheckerSupported property and separately check for the framework version depending if it's the csproj or somewhere else.

src/libraries/sendtohelixhelp.proj
ViktorHofer: OK 👍

src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj
ViktorHofer: Yes, the BuildTargetFramework condition only really makes sense in sendtohelixhelp.proj. Avoid using it somewhere else.

src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj
ViktorHofer: A condition isn't required here as the project shouldn't built at all on Mono.

carlossanlop: Okay, yes. Since I solved the problem in the other [comment ](https://github.com/dotnet/runtime/pull/93906#discussion_r1388496293) (it was due to the property not existing there), it can be removed from here.

carlossanlop:
The CI is not running. I got this error:
```
This agent request is not running because you have reached the maximum number of requests that can run for parallelism type 'Microsoft-Hosted Public'. Current position in queue: 77. Current parallelism: 200. Max Parallelism for parallelism type 'Microsoft-Hosted Public': 200
```

Directory.Build.props
```diff
@@ -179,8 +179,7 @@
     <MonoTargetsTasksAssemblyPath>$([MSBuild]::NormalizePath('$(MonoTargetsTasksDir)', 'MonoTargetsTasks.dll'))</MonoTargetsTasksAssemblyPath>
     <TestExclusionListTasksAssemblyPath>$([MSBuild]::NormalizePath('$(TestExclusionListTasksDir)', 'TestExclusionListTasks.dll'))</TestExclusionListTasksAssemblyPath>
     <CoreCLRToolPath>$([MSBuild]::NormalizeDirectory('$(ArtifactsBinDir)', 'coreclr', '$(TargetOS).$(TargetArchitecture).$(RuntimeConfiguration)'))</CoreCLRToolPath>
-    <ILAsmToolPath Condition="'$(DotNetBuildFromSource)' == 'true' or '$(BuildArchitecture)' == 's390x' or '$(BuildArchitecture)' == 'ppc64le'">$(CoreCLRToolPath)</ILAsmToolPath>
-
+    <ILAsmToolPath Condition="'$(DotNetBuildFromSource)' == 'true' or '$(BuildArchitecture)' == 's390x' or '$(BuildArchitecture)' == 'ppc64le'">$(CoreCLRToolPath)</ILAsmToolPath>
```
ViktorHofer, line 182:
```suggestion
    <ILAsmToolPath Condition="'$(DotNetBuildFromSource)' == 'true' or '$(BuildArchitecture)' == 's390x' or '$(BuildArchitecture)' == 'ppc64le'">$(CoreCLRToolPath)</ILAsmToolPath>
```

Directory.Build.props
```diff
@@ -332,6 +331,11 @@
...
+    
+    <!-- For enabling the use of XUnitLogChecker in coreclr and libraries test runs. -->
+    <IsXUnitLogCheckerSupported Condition="'$(IsXUnitLogCheckerSupported)' == ''">false</IsXUnitLogCheckerSupported>
+    <IsXUnitLogCheckerSupported Condition="'$(RuntimeFlavor)' == 'CoreCLR' and '$(TestNativeAot)' != 'true' and '$(TestRunNamePrefixSuffix)' != 'NativeAOT_Release' and '$(TargetOS)' != 'browser' and '$(TargetOS)' != 'wasi' and '$(TargetOS)' != 'ios' and '$(TargetOS)' != 'iossimulator' and '$(TargetOS)' != 'tvos' and '$(TargetOS)' != 'tvossimulator' and '$(TargetOS)' != 'maccatalyst' and '$(TargetOS)' != 'android'">true</IsXUnitLogCheckerSupported>
+    <XUnitLogCheckerPrinted>'$(RuntimeFlavor)' == 'CoreCLR' and '$(TestNativeAot)' != 'true' and '$(TestRunNamePrefixSuffix)' != 'NativeAOT_Release' and '$(TargetOS)' != 'browser' and '$(TargetOS)' != 'wasi' and '$(TargetOS)' != 'ios' and '$(TargetOS)' != 'iossimulator' and '$(TargetOS)' != 'tvos' and '$(TargetOS)' != 'tvossimulator' and '$(TargetOS)' != 'maccatalyst' and '$(TargetOS)' != 'android'</XUnitLogCheckerPrinted>
```
ViktorHofer, line 338: What is this property used for?

Directory.Build.props
```diff
@@ -332,6 +331,11 @@
 
     <!-- this property is used by the SDK to pull in mono-based runtime packs -->
     <UseMonoRuntime Condition="'$(UseMonoRuntime)' == '' and '$(RuntimeFlavor)' == 'Mono'">true</UseMonoRuntime>
+    
+    <!-- For enabling the use of XUnitLogChecker in coreclr and libraries test runs. -->
+    <IsXUnitLogCheckerSupported Condition="'$(IsXUnitLogCheckerSupported)' == ''">false</IsXUnitLogCheckerSupported>
+    <IsXUnitLogCheckerSupported Condition="'$(RuntimeFlavor)' == 'CoreCLR' and '$(TestNativeAot)' != 'true' and '$(TestRunNamePrefixSuffix)' != 'NativeAOT_Release' and '$(TargetOS)' != 'browser' and '$(TargetOS)' != 'wasi' and '$(TargetOS)' != 'ios' and '$(TargetOS)' != 'iossimulator' and '$(TargetOS)' != 'tvos' and '$(TargetOS)' != 'tvossimulator' and '$(TargetOS)' != 'maccatalyst' and '$(TargetOS)' != 'android'">true</IsXUnitLogCheckerSupported>
```
ViktorHofer, line 337: Why do you need to condition on `TestRunNamePrefixSuffix`? NativeAOT is already excluded via the `TestNativeAot` property, isn't it?

eng/testing/RunnerTemplate.cmd
```diff
@@ -92,6 +92,45 @@ if %_exit_code%==1 (
...
 )
+
+if NOT DEFINED __IsXUnitLogCheckerSupported (
+  echo The '__IsXUnitLogCheckerSupported' env var is not set.
+  GOTO SKIP_XUNITLOGCHECKER
+)
+if NOT %__IsXUnitLogCheckerSupported%==1 (
+  echo XUnitLogChecker not supported for this test case. Skipping.
+  GOTO SKIP_XUNITLOGCHECKER
+)
```
ViktorHofer, line 103: Why do you need these different cases? Isn't one condition enough?

src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj
```diff
@@ -5,6 +5,7 @@
     <TargetFramework>$(NetCoreAppToolCurrent)</TargetFramework>
     <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <XUnitLogCheckerLibrariesHelixDirectory>$([MSBuild]::NormalizeDirectory('$(ArtifactsDir)', 'helix', 'tests', 'XUnitLogChecker'))</XUnitLogCheckerLibrariesHelixDirectory>
```
ViktorHofer, line 8: This property is used by both XUnitLogChecker and sendtohelixhelp and should therefore be centrally defined.

src/libraries/sendtohelixhelp.proj
```diff
@@ -233,6 +232,38 @@
...
+      set __IsXUnitLogCheckerSupported=$(_ShouldSetXUnitLogCheckerEnvVar)
+    </HelixPreCommands>
+    <HelixPreCommands Condition="'$(TargetOS)' != 'windows'">
+      $(HelixPreCommands);
+      export __IsXUnitLogCheckerSupported=$(_ShouldSetXUnitLogCheckerEnvVar)
+    </HelixPreCommands>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(BuildTargetFramework)' == '$(NetCoreAppCurrent)' and '$(IsXUnitLogCheckerSupported)' == 'true'">  
+    <HelixCorrelationPayload Include="$([MSBuild]::NormalizePath('$(ArtifactsDir)', 'helix', 'tests', 'XUnitLogChecker', 'XUnitLogChecker.zip'))" />
```
ViktorHofer, line 264: Use the centrally defined property here.

eng/testing/tests.targets
```diff
@@ -155,7 +155,7 @@ TEST_ARCH=$(_AndroidArchitecture)
     </WriteLinesToFile>
   </Target>
 
-  <Target Name="RunTests">
+  <Target Name="RunTests" Condition="'$(MSBuildProjectName)' != 'XUnitLogChecker'">
```
ViktorHofer, line 158: I'm confused. I thought we diagnosed that this file isn't imported for XUnitLogChecker?

carlossanlop: The empty "Test" target in the csproj didn't work. So I tried this instead hoping it would prevent running tests on it, but look https://dev.azure.com/dnceng-public/public/_build/results?buildId=465642&view=logs&j=fd49b2ea-3ee3-5298-d793-bdb9fd631e7e&t=26d44400-a1b8-5ed1-d719-0cc35b38251b&l=55

ViktorHofer: Move the empty "Test" target into a Directory.Build.targets file in the same directory as the project file. Make sure that the Directory.Build.targets file imports the parent Directory.Build.targets file.

Directory.Build.props
carlossanlop: For debugging, I'll remove it when I'm done.

src/tests/Common/XUnitLogChecker/XUnitLogChecker.csproj
carlossanlop: One's for the build output, another one is for the zipping output. I realized the helix one needed to be located under helix/tests to ensure the zip was uploaded.

eng/testing/RunnerTemplate.cmd
carlossanlop: I can remove the first case. I just noticed the "set" command for this env var is verbose in the logged file, so if it does not get set, we won't see it.

src/libraries/sendtohelixhelp.proj
```diff
@@ -233,13 +232,46 @@
     </ItemGroup>
   </Target>
 
+  <PropertyGroup>
+    <_ShouldSetXUnitLogCheckerEnvVar>0</_ShouldSetXUnitLogCheckerEnvVar>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(IsXUnitLogCheckerSupported)' == 'true'">
+    <_ShouldSetXUnitLogCheckerEnvVar Condition="$(TargetFrameworkIdentifier) == '.NETCoreApp'">1</_ShouldSetXUnitLogCheckerEnvVar>
```
ViktorHofer, line 240: TargetFrameworkIdentifier isn't defined in this project. Use BuildTargetFramework instead.

src/libraries/sendtohelixhelp.proj
```diff
@@ -233,13 +232,46 @@
...
+      set __TestArchitecture=$(TargetArchitecture)
+    </HelixPreCommands>
+    <HelixPreCommands Condition="'$(TargetOS)' != 'windows'">
+      $(HelixPreCommands);
+      export __TestArchitecture=$(TargetArchitecture)
+    </HelixPreCommands>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- The IsXUnitLogCheckerSupported property should always be set. -->
```
ViktorHofer, line 252: Hmm I thought I left a comment for this. Why do you need the env var to always be defined? That's not what we usually do.

carlossanlop: I want to skip the execution of XUnitLogChecker in the lines below in the script when it's value is set to zero.

src/libraries/sendtohelixhelp.proj
carlossanlop: Done: $(BuildTargetFramework) == '$(NetCoreAppCurrent)'

src/libraries/sendtohelixhelp.proj
ViktorHofer: Right, but why can't you skip execution when the env var isn't defined or 0?

carlossanlop: Okay yes, I can move them up here.

src/libraries/sendtohelixhelp.proj
```diff
@@ -233,14 +232,35 @@
...
+      set __IsXUnitLogCheckerSupported=$(_ShouldSetXUnitLogCheckerEnvVar)
+    </HelixPreCommands>
+    <HelixPreCommands Condition="'$(TargetOS)' != 'windows'">
+      $(HelixPreCommands);
+      export __TestArchitecture=$(TargetArchitecture)
+      export __IsXUnitLogCheckerSupported=$(_ShouldSetXUnitLogCheckerEnvVar)
+    </HelixPreCommands>
+  </PropertyGroup>
+
+  <ItemGroup Condition="'$(BuildTargetFramework)' == '$(NetCoreAppCurrent)' and '$(IsXUnitLogCheckerSupported)' == 'true'">
```
ViktorHofer, line 253: re-use the same condition here:

```suggestion
  <ItemGroup Condition="'$(_ShouldSetXUnitLogCheckerEnvVar)' == '1'">  
```

src/libraries/sendtohelixhelp.proj
```diff
@@ -233,14 +232,35 @@
     </ItemGroup>
   </Target>
 
+  <PropertyGroup>
+    <_ShouldSetXUnitLogCheckerEnvVar>0</_ShouldSetXUnitLogCheckerEnvVar>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(IsXUnitLogCheckerSupported)' == 'true'">
```
ViktorHofer, line 239: Why don't you move the `BuildTargetFramework` condition up into this PropertyGroup so that all these env vars are only set when necessary?

carlossanlop: Hmm good point. There's no .NET Framework in unix, so that should work.

carlossanlop:
I temporarily brought back the commit that causes the crash and ignores the unrelated CI legs: https://github.com/dotnet/runtime/pull/93906/commits/25f6a705d374c969a3fc912df24d2ec70c8e803c

I'll revert it once I confirm the crashes are getting logged properly.

@hoyosjs @ivdiazsa can I get a review meanwhile, please?

src/libraries/sendtohelixhelp.proj
```diff
@@ -233,13 +232,36 @@
     </ItemGroup>
   </Target>
 
+  <PropertyGroup>
+    <_ShouldSetXUnitLogCheckerEnvVar>0</_ShouldSetXUnitLogCheckerEnvVar>
+  </PropertyGroup>
+
```
carlossanlop, line 238:
```suggestion
```

carlossanlop:
Here are the remaining tasks for future PRs:
- https://github.com/dotnet/runtime/issues/91975 - [example 1](https://helixre8s23ayyeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-f123e25602cd4d2792/System.IO.Compression.Brotli.Tests/1/console.98306b61.log?helixlogtype=result), [example 2](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-c4de9359aec447c8ad/System.IO.Compression.Brotli.Tests/1/console.369431af.log?helixlogtype=result), [Example 3](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-d61992d1d3604904a5/System.IO.Compression.Brotli.Tests/1/console.d80b036e.log?helixlogtype=result)
- osx: No dumps found in /cores - [Example 1](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-5491c59ba3e94b3a88/System.IO.Compression.Brotli.Tests/1/console.24609a8d.log?helixlogtype=result), [Example 2](https://helixre107v0xdeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-d4160ad03abd454cab/System.IO.Compression.Brotli.Tests/1/console.5b69f91a.log?helixlogtype=result)
- Frames are sometimes not resolved in the printed dump contents - [Example 1](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-f0b140d21ce348bc89/System.IO.Compression.Brotli.Tests/1/console.7c296702.log?helixlogtype=result), [Example 2](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-f1656fd4a7794c27a8/System.IO.Compression.Brotli.Tests/1/console.01a1cf6f.log?helixlogtype=result), [Example 3](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-93906-merge-d61710e0f43545ab9d/System.IO.Compression.Brotli.Tests/1/console.795e54e9.log?helixlogtype=result)
- We're skipping NativeAOT because it cannot find dotnet.exe, need a different way of building and running XUnitLogChecker - [Code that skips it](https://github.com/dotnet/runtime/blob/69df4b6357d0e5fe6d167bd15a40d9936d1f765f/Directory.Build.props#L337)

ivdiazsa:
All the remaining observations I had have already been pointed out in the review. So, it looks good to me now.

carlossanlop:
@ViktorHofer I addressed your latest comment (remove unused envvar property defined outside of the condition). I only need a sign-off, and remove the temporary change that causes the crash and skips unrelated CIs, then I can merge if the CI is green in the final run.

Directory.Build.props
hoyosjs: Do you need both in that case? Also, runtime flavor should remove a lot of the TargetOS ones.

eng/testing/RunnerTemplate.sh
```diff
@@ -54,80 +54,60 @@ exitcode_list[131]="SIGQUIT Ctrl-\ occurred. Core dumped."
...
-function print_info_from_core_file_using_lldb {
+function move_core_file_to_temp_location {
   local core_file_name=$1
-  local executable_name=$2
-  local plugin_path_name="$RUNTIME_PATH/shared/Microsoft.NETCore.App/9.9.9/libsosplugin.so"
 
-  # check for existence of lldb on the path
-  hash lldb 2>/dev/null || { echo >&2 "lldb was not found. Unable to print core file."; return; }
+  # Append the dmp extension to ensure XUnitLogChecker finds it
+  local new_location=$HELIX_DUMP_FOLDER/core.$RANDOM.dmp
```
hoyosjs, line 68: This is not ideal - you're missing info here (the pid decorated name). Any reason not to append .dmp in case that's not the extension?

eng/testing/RunnerTemplate.sh
```diff
@@ -220,51 +190,69 @@ if [[ $test_exitcode -ne 0 ]]; then
...
+
+if [ -n "$HELIX_WORKITEM_PAYLOAD" ]; then
+  # For abrupt failures, in Helix, dump some of the kernel log, in case there is a hint
+  if [[ $test_exitcode -ne 1 ]]; then
+    dmesg | tail -50
+  fi
+
+  have_sleep=$(which sleep)
+  if [ -x "$have_sleep" ]; then
+    echo Waiting a few seconds for any dump to be written...
```
hoyosjs, line 231: This shouldn't be needed - the runtime itself writes the dump before the process exits.

eng/testing/RunnerTemplate.sh
```diff
@@ -220,51 +190,69 @@ if [[ $test_exitcode -ne 0 ]]; then
...
+if [[ -z "$__IsXUnitLogCheckerSupported" ]]; then
+  echo "The '__IsXUnitLogCheckerSupported' env var is not set."
+elif [[ "$__IsXUnitLogCheckerSupported" != "1" ]]; then
+  echo "XUnitLogChecker not supported for this test case. Skipping."
+else
+  echo ----- start ===============  XUnitLogChecker Output =====================================================
+  
+  invoke_xunitlogchecker "$HELIX_DUMP_FOLDER"
+
+  # osx stores giant dumps in a separate folder
```
hoyosjs, line 245: This is only the system cores - which I thought we disabled?

src/libraries/sendtohelixhelp.proj
```diff
@@ -233,13 +232,32 @@
...
+  <PropertyGroup Condition="'$(IsXUnitLogCheckerSupported)' == 'true' and '$(BuildTargetFramework)' == '$(NetCoreAppCurrent)'">
+    <_ShouldSetXUnitLogCheckerEnvVar>1</_ShouldSetXUnitLogCheckerEnvVar>
+    <HelixPreCommands Condition="'$(TargetOS)' == 'windows'">
+      $(HelixPreCommands);
+      set __TestArchitecture=$(TargetArchitecture)
+      set __IsXUnitLogCheckerSupported=$(_ShouldSetXUnitLogCheckerEnvVar)
+    </HelixPreCommands>
+    <HelixPreCommands Condition="'$(TargetOS)' != 'windows'">
+      $(HelixPreCommands);
+      export __TestArchitecture=$(TargetArchitecture)
```
hoyosjs, line 244: These don't need semicolons between the lines?

hoyosjs (approved):
LGTM - minor q's

src/libraries/sendtohelixhelp.proj
carlossanlop: I am following the same pattern as the other similar examples in this file.

eng/testing/RunnerTemplate.sh
carlossanlop: XUnitLogChecker only looks for files that end in *.dmp.

eng/testing/RunnerTemplate.sh
carlossanlop: This code was already here, just moved it to a different location which is why it looks like I added it.

eng/testing/RunnerTemplate.sh
carlossanlop: It seems that $HELIX_DUMP_FOLDER equals `/cores` anyway in OSX, which means this code is unnecessary, yes, because `invoke_xunitlogchecker` is called twice on the same folder. So yes, I can remove it.

eng/testing/RunnerTemplate.sh
```diff
@@ -220,51 +190,69 @@ if [[ $test_exitcode -ne 0 ]]; then
...
+  if [[ $system_name == "Darwin" ]]; then
+    invoke_xunitlogchecker "/cores"
+  fi
+
```
carlossanlop, line 249:
```suggestion
```

src/libraries/sendtohelixhelp.proj
```diff
@@ -233,13 +232,32 @@
     </ItemGroup>
   </Target>
 
+  <PropertyGroup Condition="'$(IsXUnitLogCheckerSupported)' == 'true' and '$(BuildTargetFramework)' == '$(NetCoreAppCurrent)'">
+    <_ShouldSetXUnitLogCheckerEnvVar>1</_ShouldSetXUnitLogCheckerEnvVar>
+    <HelixPreCommands Condition="'$(TargetOS)' == 'windows'">
+      $(HelixPreCommands);
+      set __TestArchitecture=$(TargetArchitecture)
```
carlossanlop, line 239:
```suggestion
      set __TestArchitecture=$(TargetArchitecture);
```

carlossanlop: @hoyosjs I actually missed a semicolon here if I'm going to follow the right pattern.

src/libraries/sendtohelixhelp.proj
carlossanlop: Ah, yes, missed semicolon
```suggestion
      export __TestArchitecture=$(TargetArchitecture);
```

eng/testing/RunnerTemplate.sh
carlossanlop: I can try to preserve the exact same name, add a $RANDOM (in case there is more than one `core` file) and then append *.dmp.

carlossanlop, line 68:
```suggestion
  local new_location=$HELIX_DUMP_FOLDER/$core_file_name.$RANDOM.dmp
```

eng/testing/RunnerTemplate.sh
```diff
@@ -220,51 +190,64 @@ if [[ $test_exitcode -ne 0 ]]; then
...
+  # For abrupt failures, in Helix, dump some of the kernel log, in case there is a hint
+  if [[ $test_exitcode -ne 1 ]]; then
+    dmesg | tail -50
+  fi
+
+  have_sleep=$(which sleep)
+  if [ -x "$have_sleep" ]; then
+    echo Waiting a few seconds for any dump to be written...
+    sleep 10s
   fi
```
carlossanlop, line 233:
```suggestion
```

eng/testing/RunnerTemplate.sh
hoyosjs: I mean - why do we need $RANDOM?

Directory.Build.props
carlossanlop: I noticed some NativeAot runs were not getting ignored unless I added both conditions. They are temporary anyway. I opened an issue so the NativeAOT owners can get it enabled properly.

eng/testing/RunnerTemplate.sh
carlossanlop: I opened this issue to track why the /cores dumps are not being read by XUnitLogChecker: https://github.com/dotnet/runtime/issues/94720

eng/testing/RunnerTemplate.sh
carlossanlop: I removed it anyway.

eng/testing/RunnerTemplate.sh
carlossanlop: Originally, I was moving the dumps from different folders into a central one. I was worried that there could be a name collision, so I preserved the $RANDOM. Since now we're only moving them from one place, and there can only be one file named `core`, the $RANDOM is not necessary anymore, so I'll remove it.

Directory.Build.props
MichalStrehovsky: Having to condition on `'$(TestRunNamePrefixSuffix)' != 'NativeAOT_Release'` is extremely suspicious. We only ever run native AOT testing with `TestNativeAot`. It's not native AOT testing if that property is not true. This will likely produce __obscure__ failure modes if we add new test pipelines or rename existing. It will be expensive to root cause for whoever isn't already aware of this hack. Cc @dotnet/ilc-contrib for awareness.

> They are temporary anyway. I opened an issue so the NativeAOT owners can get it enabled properly.

Unless you already talked to some "NativeAOT owners" and scheduled their time to fix this, this temporary workaround is likely to be in the product for a long time, like most temporary workarounds. I'm very much against this condition being checked in as it is, please investigate this. One thing off top of my head is that this might need to check for `RunNativeAotTestApps` instead, which gates special kinds of library tests that was added in #76109.

jkotas:
@carlossanlop This broke native AOT outer loop runs.

From https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_apis/build/builds/470488/logs/135 👍 

error : Correlation Payload 'D:\a\_work\1\s\artifacts\bin\XUnitLogChecker\' not found. [D:\a\_work\1\s\src\libraries\sendtohelixhelp.proj]

Could you please take a look asap?

carlossanlop:
@jkotas apologies. I'll revert it, and then manually run outerloop.

Directory.Build.props
carlossanlop: @MichalStrehovsky Since I had to revert the change from main due to NativeAOT tests being broken in outerloop, I can remove the TestRunNamePrefixSuffix property from the condition in the new PR.
I also opened https://github.com/dotnet/runtime/issues/94722 to track enabling them in the future.

MichalStrehovsky: Thanks! The issue might have some overlap with #94270.

carlossanlop: Feel free to mark mine as duplicate of yours if they're tracking the same.

Issue #72192, opened 2022-07-14
Failed to create CoreCLR, HRESULT: 0x80004005 with WSL1


josephmoresena:
### Description

I'm trying to build CoreCLR on Ubuntu using WSL1. The system displays the following output:
`
__DistroRid: linux-x64
Failed to create CoreCLR, HRESULT: 0x80004005
Build failed with exit code 137. Check errors above.
`

### Reproduction Steps

In runtime repo directory execute:
`./build.sh -subset clr`

### Expected behavior

Start CoreCLR compiling.

### Actual behavior

Build failed with exit code 137. Check errors above.

EgorBo:
You need to inspect the full log, e.g. make sure you installed all the needed dependencies from https://github.com/dotnet/runtime/blob/main/docs/workflow/requirements/linux-requirements.md
Also, check your cmake version

josephmoresena:
> You need to inspect the full log, e.g. make sure you installed all the needed dependencies from https://github.com/dotnet/runtime/blob/main/docs/workflow/requirements/linux-requirements.md Also, check your cmake version

This is my current configuration:
`
Reading package lists... Done
Building dependency tree
Reading state information... Done
Note, selecting 'python-is-python2' instead of 'python'
Note, selecting 'libunwind-dev' instead of 'libunwind8-dev'
cmake is already the newest version (3.16.3-1ubuntu1).
gettext is already the newest version (0.19.8.1-10build1).
liblttng-ust-dev is already the newest version (2.11.0-1).
libnuma-dev is already the newest version (2.0.12-1).
libunwind-dev is already the newest version (1.2.1-9build1).
libunwind8 is already the newest version (1.2.1-9build1).
clang-9 is already the newest version (1:9.0.1-12).
liblldb-6.0-dev is already the newest version (1:6.0.1-14).
lldb-6.0 is already the newest version (1:6.0.1-14).
llvm-9 is already the newest version (1:9.0.1-12).
ninja-build is already the newest version (1.10.0-1build1).
python-is-python2 is already the newest version (2.7.17-4).
build-essential is already the newest version (12.8ubuntu1.1).
curl is already the newest version (7.68.0-1ubuntu2.12).
git is already the newest version (1:2.25.1-1ubuntu3.5).
libicu-dev is already the newest version (66.1-2ubuntu2.1).
libkrb5-dev is already the newest version (1.17-6ubuntu4.1).
[...]

PR #93907, opened 2023-10-24
Haiku: Add Haiku support for CoreCLR/PAL

trungnt2910:
This contains the code required to build this repo's native components and get `paltests` to pass on Haiku.

Most implementation details related to this commit are documented [here](https://www.haiku-os.org/blog/trungnt2910/2023-05-14_gsoc_2023_dotnet_port_progress_1/).

Part of https://github.com/dotnet/runtime/issues/55803.

### DIFF
--- a/src/coreclr/pal/inc/rt/safecrt.h
+++ b/src/coreclr/pal/inc/rt/safecrt.h
@@ -112,15 +112,6 @@ typedef int errno_t; /* standard */
 /* error codes */
 #if !defined(_SECURECRT_ERRCODE_VALUES_DEFINED)
 #define _SECURECRT_ERRCODE_VALUES_DEFINED
-#if !defined(EINVAL)
-#define EINVAL          22
-#endif
-#if !defined(ERANGE)
-#define ERANGE          34
-#endif
-#if !defined(EILSEQ)
-#define EILSEQ          42
-#endif
 #if !defined(STRUNCATE)
 #define STRUNCATE       80
 #endif
--- a/src/coreclr/pal/src/include/pal/thread.hpp
+++ b/src/coreclr/pal/src/include/pal/thread.hpp
@@ -24,7 +24,6 @@ Abstract:
 #include "cs.hpp"
 
 #include <pthread.h>
-#include <sys/syscall.h>
 #if HAVE_MACH_EXCEPTIONS
 #include <mach/mach.h>
 #endif // HAVE_MACH_EXCEPTIONS
--- a/src/coreclr/pal/src/map/map.cpp
+++ b/src/coreclr/pal/src/map/map.cpp
@@ -108,6 +108,11 @@ MAPmmapAndRecord(
 #define MAP_ANON MAP_ANONYMOUS
 #endif
 
+/* Some platforms like Haiku does not provide MAP_FILE */
+#ifndef MAP_FILE
+#define MAP_FILE 0
+#endif
+
 void
 FileMappingCleanupRoutine(
     CPalThread *pThread,
--- a/src/coreclr/pal/src/map/map.cpp
+++ b/src/coreclr/pal/src/map/map.cpp
@@ -108,11 +108,6 @@ MAPmmapAndRecord(
 #define MAP_ANON MAP_ANONYMOUS
 #endif
 
-/* Some platforms like Haiku does not provide MAP_FILE */
-#ifndef MAP_FILE
-#define MAP_FILE 0
-#endif
-
 void
 FileMappingCleanupRoutine(
     CPalThread *pThread,
@@ -2115,7 +2110,7 @@ void * MAPMapPEFile(HANDLE hFile, off_t offset)
 #endif
     SIZE_T reserveSize = 0;
     bool forceOveralign = false;
-    int readWriteFlags = MAP_FILE|MAP_PRIVATE|MAP_FIXED;
+    int readWriteFlags = MAP_PRIVATE|MAP_FIXED;
     int readOnlyFlags = readWriteFlags;
 
     ENTRY("MAPMapPEFile (hFile=%p offset=%zx)\n", hFile, offset);
@@ -2322,7 +2317,7 @@ void * MAPMapPEFile(HANDLE hFile, off_t offset)
         // If PAL_MAP_READONLY_PE_HUGE_PAGE_AS_SHARED is set to 1. map the readonly sections as shared
         // which works well with the behavior of the hugetlbfs
         if (mapAsShared != NULL && (strcmp(mapAsShared, "1") == 0))
-            readOnlyFlags = MAP_FILE|MAP_SHARED|MAP_FIXED;
+            readOnlyFlags = MAP_SHARED|MAP_FIXED;
     }
 
     //we have now reserved memory (potentially we got rebased).  Walk the PE sections and map each part
--- a/src/native/minipal/debugger.c
+++ b/src/native/minipal/debugger.c
@@ -35,6 +35,8 @@
 #include <sys/proc.h>
 #include <sys/types.h>
 #include <sys/procfs.h>
+#elif defined(__HAIKU__)
+#include <OS.h>
 #endif
 
 bool minipal_is_native_debugger_present(void)
@@ -128,6 +130,14 @@ bool minipal_is_native_debugger_present(void)
     pid_t pid = getpid();
     getprocs64(&proc, sizeof(proc), NULL, 0, &pid, 1);
     return (proc.pi_flags & STRC) != 0; // SMPTRACE or SWTED might work too
+
+#elif defined(__HAIKU__)
+    team_info info;
+    if (get_team_info(B_CURRENT_TEAM, &info) == B_OK)
+    {
+        return info.debugger_nub_thread > 0;
+    }
+    return false;
 
 #else
     return false;
--- a/src/coreclr/ilasm/ilasmpch.h
+++ b/src/coreclr/ilasm/ilasmpch.h
@@ -25,4 +25,12 @@
 #include "mdfileformat.h"
 #include "stgpooli.h"
 
+#ifdef _EXPORT
+#undef _EXPORT
+#endif
+
+#ifdef _IMPORT
+#undef _IMPORT
+#endif
+
 #endif
--- a/src/coreclr/pal/inc/pal.h
+++ b/src/coreclr/pal/inc/pal.h
@@ -2644,6 +2644,8 @@ PALIMPORT BOOL PALAPI PAL_GetUnwindInfoSize(SIZE_T baseAddress, ULONG64 ehFrameH
 #define PAL_CS_NATIVE_DATA_SIZE 96
 #elif defined(__linux__) && defined(__riscv) && __riscv_xlen == 64
 #define PAL_CS_NATIVE_DATA_SIZE 96
+#elif defined(__HAIKU__) && defined(__x86_64__)
+#define PAL_CS_NATIVE_DATA_SIZE 56
 #else
 #error  PAL_CS_NATIVE_DATA_SIZE is not defined for this architecture
 #endif
@@ -3952,7 +3954,9 @@ unsigned int __cdecl _rotr(unsigned int value, int shift)
 PALIMPORT DLLEXPORT char * __cdecl PAL_getenv(const char *);
 PALIMPORT DLLEXPORT int __cdecl _putenv(const char *);
 
+#ifndef ERANGE
 #define ERANGE          34
+#endif
 
 /****************PAL Perf functions for PInvoke*********************/
 #if PAL_PERF
--- a/src/coreclr/pal/src/arch/amd64/signalhandlerhelper.cpp
+++ b/src/coreclr/pal/src/arch/amd64/signalhandlerhelper.cpp
@@ -8,7 +8,10 @@ SET_DEFAULT_DEBUG_CHANNEL(EXCEPT); // some headers have code with asserts, so do
 #include "pal/context.h"
 #include "pal/signal.hpp"
 #include "pal/utils.h"
+
+#if HAVE_SYS_UCONTEXT_H
 #include <sys/ucontext.h>
+#endif
 
 /*++
 Function :
--- a/src/coreclr/pal/src/config.h.in
+++ b/src/coreclr/pal/src/config.h.in
@@ -21,6 +21,8 @@
 #cmakedefine01 HAVE_PTHREAD_NP_H
 #cmakedefine01 HAVE_AUXV_HWCAP_H
 #cmakedefine01 HAVE_SYS_PTRACE_H
+#cmakedefine01 HAVE_SYS_UCONTEXT_H
+#cmakedefine01 HAVE_SYS_MOUNT_H
 #cmakedefine01 HAVE_UCONTEXT_H
 #cmakedefine01 HAVE_GETAUXVAL
 
--- a/src/coreclr/pal/src/configure.cmake
+++ b/src/coreclr/pal/src/configure.cmake
@@ -47,6 +47,8 @@ check_include_files(semaphore.h HAVE_SEMAPHORE_H)
 check_include_files(sys/prctl.h HAVE_PRCTL_H)
 check_include_files("sys/auxv.h;asm/hwcap.h" HAVE_AUXV_HWCAP_H)
 check_include_files("sys/ptrace.h" HAVE_SYS_PTRACE_H)
+check_include_files("sys/ucontext.h" HAVE_SYS_UCONTEXT_H)
+check_include_files("sys/mount.h" HAVE_SYS_MOUNT_H)
 check_include_files(ucontext.h HAVE_UCONTEXT_H)
 check_symbol_exists(getauxval sys/auxv.h HAVE_GETAUXVAL)
 
--- a/src/coreclr/pal/src/debug/debug.cpp
+++ b/src/coreclr/pal/src/debug/debug.cpp
@@ -45,7 +45,7 @@ SET_DEFAULT_DEBUG_CHANNEL(DEBUG); // some headers have code with asserts, so do 
 #include <unistd.h>
 #elif defined(HAVE_TTRACE) // HAVE_PROCFS_CTL
 #include <sys/ttrace.h>
-#else // defined(HAVE_TTRACE)
+#elif HAVE_SYS_PTRACE_H
 #include <sys/ptrace.h>
 #endif  // HAVE_PROCFS_CTL
 #if HAVE_VM_READ
@@ -635,7 +635,7 @@ Function:
   PAL_ReadProcessMemory
 
 Abstract
-  Reads process memory. 
+  Reads process memory.
 
 Parameter
   handle : from PAL_OpenProcessMemory
@@ -753,7 +753,7 @@ PAL_ProbeMemory(
 
     flags = fcntl(fds[0], F_GETFL, 0);
     fcntl(fds[0], F_SETFL, flags | O_NONBLOCK);
-    
+
     flags = fcntl(fds[1], F_GETFL, 0);
     fcntl(fds[1], F_SETFL, flags | O_NONBLOCK);
 
--- a/src/coreclr/pal/src/exception/seh.cpp
+++ b/src/coreclr/pal/src/exception/seh.cpp
@@ -305,7 +305,7 @@ PAL_ERROR SEHEnable(CPalThread *pthrCurrent)
 {
 #if HAVE_MACH_EXCEPTIONS
     return pthrCurrent->EnableMachExceptions();
-#elif defined(__linux__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__sun)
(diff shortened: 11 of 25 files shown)


src/native/minipal/errno.h
```diff
@@ -0,0 +1,48 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
jkotas, line 1: This file should not be needed.

CoreCLR PAL implementation should be able to use definition of these constants from Haiku system headers. I assume that all these constants are defined in Haiky system headers. Is that correct?

Components using the CoreCLR PAL should use definitions provided by the CoreCLR PAL header.

trungnt2910: I believe the discussion has been done here: https://github.com/dotnet/runtime/issues/55803#issuecomment-886096255, unless something major has changed since the .NET 6 - 7 times.

@jessicah's port (https://github.com/jessicah/dotnet-runtime) uses `#ifdef` for every `errno` wherever needed; this header gathers them all to one place.

jkotas: I agree that there may be some changes required to deal with the errno given how the CoreCLR PAL is structured (https://github.com/dotnet/runtime/issues/55803#issuecomment-886696103).

All changes related to errno should be under src/coreclr/pal. No changes should be needed outside this directory, except of deleting unnecessary code. I have left some comments on the @jessicah's commit: https://github.com/jessicah/dotnet-runtime/commit/7bd4f1d40b4d938b81eecbc77d80e145e005118b

trungnt2910: Done. I've removed the `minpal/errno.h` header.

However, unlike what you said about the definitions in `sstring.cpp` (a file _outside_ `src/coreclr/pal`, located in `src/coreclr/utilcode` instead), the definitions are still needed:

https://github.com/dotnet/runtime/blob/b09e18e8b6092e35611c2567bd55e8909ea78d7d/src/coreclr/utilcode/sstring.cpp#L1794-L1803

For these files, I've replaced the `#include <minipal/errno.h>` with `#include <pal.h>`, though I don't really know what's the correct solution.

For now, nothing seems to have broken. Builds are still fine, `paltests` still passing.

trungnt2910: Oops, seems like `pal.h` is not accessible from `sstring.cpp`:

> D:\a\_work\1\s\src\coreclr\utilcode\sstring.cpp(14): fatal error C1083: Cannot open include file: 'pal.h': No such file or directory

src/coreclr/pal/inc/rt/safecrt.h
```diff
@@ -131,15 +131,7 @@ typedef int errno_t; /* standard */
...
-#if !defined(EINVAL)
-#define EINVAL          22
-#endif
-#if !defined(ERANGE)
-#define ERANGE          34
-#endif
-#if !defined(EILSEQ)
-#define EILSEQ          42
-#endif
+#include <pal.h>
```
janvorli, line 134: This include should not be needed, the palrt.h includes pal.h and then it includes the safecrt.h, so pal.h should be already included here.

src/coreclr/utilcode/sstring.cpp
```diff
@@ -11,6 +11,7 @@
 #include "ex.h"
 #include "holder.h"
 
+#include <pal.h>
```
janvorli, line 14: I think that pal.h should have been included transitively from one of the previous includes, so this should not be needed.

src/native/libs/System.Native/pal_networkchange.c
```diff
@@ -175,8 +178,6 @@ Error SystemNative_ReadEvents(intptr_t sock, NetworkChangeEvent onNetworkChange)
 {
     (void)sock;
     (void)onNetworkChange;
-    // unreachable
-    abort();
-    return Error_SUCCESS;
+    return Error_EINVAL;
```
janvorli, line 181: Should we keep the old aborting code as a fallback and add separate #if for HAIKU instead? I am not familiar enough with the contract of this function though.

src/native/minipal/errno.h
janvorli: @trungnt2910 the pal.h is not available on Windows. You should never include it anywhere from the runtime code explicitly.  It gets automatically included when runtime includes any of the standard headers. They are redirected to palrt.h and that one pulls in pal.h

src/native/libs/System.Native/pal_networkchange.c
trungnt2910: I've added back the `abort()` logic, as an assertion that the function is unreachable.

This is true because on unsupported platforms, the higher-level C# libraries should already throw while trying to construct a network events socket.

Returning `Error_SUCCESS` does not make much sense though.

src/coreclr/utilcode/sstring.cpp
trungnt2910: Done. Let's see if it fixes/breaks anything.

src/coreclr/pal/inc/rt/safecrt.h
trungnt2910: Done.

src/coreclr/minipal/Unix/doublemapping.cpp
```diff
@@ -53,14 +53,24 @@ bool VMToOSInterface::CreateDoubleMemoryMapper(void** pHandle, size_t *pMaxExecu
 
 #ifdef TARGET_FREEBSD
     int fd = shm_open(SHM_ANON, O_RDWR | O_CREAT, S_IRWXU);
-#elif defined(TARGET_SUNOS) // has POSIX implementation
-    char name[24];
-    sprintf(name, "/shm-dotnet-%d", getpid());
-    name[sizeof(name) - 1] = '\0';
-    shm_unlink(name);
-    int fd = shm_open(name, O_RDWR | O_CREAT | O_EXCL | O_NOFOLLOW, 0600);
 #else // TARGET_FREEBSD
```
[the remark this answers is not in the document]
trungnt2910: Done. The final `#else` assigns the `fd` variable to `-1` instead of doing the POSIX fallback itself, in case any of the platform-specific functions above return an error.

This should solve #72192 from the .NET 7 era.

trungnt2910:
Strange, I'm not modifying any FreeBSD-related codepaths, what's wrong with the FreeBSD build?

jkotas:
FreeBSD build error is:
```
/__w/1/s/src/coreclr/pal/src/misc/sysinfo.cpp:215:57: error: use of undeclared identifier 'NUPML4E'
    lpSystemInfo->lpMaximumApplicationAddress = (PVOID) VM_MAXUSER_ADDRESS;
                                                        ^
/crossrootfs/x64/usr/include/machine/vmparam.h:186:35: note: expanded from macro 'VM_MAXUSER_ADDRESS'
#define VM_MAXUSER_ADDRESS      UVADDR(NUPML4E, 0, 0, 0)
                                       ^
```
I guess you have introduced by changing conditions around some of the includes.

trungnt2910:
This is suspicious:

```
-- Looking for include file sys/user.h
-- Looking for include file sys/user.h - not found
```

I wonder how I could get more detailed CMake logs for the CI runs.

trungnt2910:
When looking at other files, it seems like `sys/user.h` is only used on FreeBSD. Somehow `sysinfo.cpp` includes this header for all platforms.

This time, instead of checking for this header during configuration as in #86391, a FreeBSD `#ifdef` guard is added instead.

src/coreclr/minipal/Unix/doublemapping.cpp
```diff
@@ -53,15 +53,22 @@ bool VMToOSInterface::CreateDoubleMemoryMapper(void** pHandle, size_t *pMaxExecu
...
+    // POSIX fallback
+    if (fd == -1)
+    {
+        char name[24];
+        sprintf(name, "/shm-dotnet-%d", getpid());
+        name[sizeof(name) - 1] = '\0';
+        shm_unlink(name);
+        fd = shm_open(name, O_RDWR | O_CREAT | O_EXCL | O_NOFOLLOW, 0600);
+        shm_unlink(name);
+    }
```
huoyaoyuan, line 71: Shouldn't this block be in `#else`?

src/coreclr/gc/unix/numasupport.cpp
```diff
@@ -9,9 +9,12 @@
 #include <stdio.h>
 #include <dirent.h>
 #include <string.h>
-#include <sys/syscall.h>
 #include <minipal/utils.h>
 
+#ifdef TARGET_LINUX
+#include <sys/syscall.h>
+#endif
```
huoyaoyuan, line 16: Is this used by other Unix systems?

src/coreclr/pal/inc/pal_mstypes.h
```diff
@@ -574,8 +574,8 @@ typedef long ptrdiff_t;
 typedef unsigned int size_t;
 typedef int ptrdiff_t;
 #endif // !HOST_64BIT
-#endif // !PAL_STDCPP_COMPAT
 #define _SIZE_T_DEFINED
+#endif // !PAL_STDCPP_COMPAT
```
huoyaoyuan, line 578: What's the impact range of this?

src/coreclr/minipal/Unix/doublemapping.cpp
trungnt2910: No, it should not, as stated in https://github.com/dotnet/runtime/pull/93907#discussion_r1371114236.

Even for OSes with platform-specific ways to get a special memory file open, should their method fail, we can still always use the POSIX fallback.

src/coreclr/gc/unix/numasupport.cpp
trungnt2910: No, it is only used on Linux:

https://github.com/dotnet/runtime/blob/e5c631d28b985b773b28d0cb2346f49a92bd41dd/src/coreclr/gc/unix/numasupport.cpp#L52-L63

and 

https://github.com/dotnet/runtime/blob/e5c631d28b985b773b28d0cb2346f49a92bd41dd/src/coreclr/gc/unix/numasupport.cpp#L81-L83

src/coreclr/pal/inc/pal_mstypes.h
trungnt2910: What do you mean by "impact range"?

The line should affect all platforms. However, I believe that this is how things should be, since with `PAL_STDCPP_COMPAT` the header should not define things that potentially mess up system headers.

Some more comments about this change here: https://github.com/dotnet/runtime/issues/55803#issuecomment-1537114390

trungnt2910:
Anything known issues that I have to take extra attention of related to the Linux SSL implementation? Somehow I've been getting CI errors for this PR on Linux only; I don't really know what's happening.

mangod9:
Since this is a large change, would it be possible to split the PR info smaller component PRs for PAL, hosting, libraries etc. so relevant teams can review the changes? Thx

trungnt2910:
Not right now (I'm too busy). I'm still tracking it though, so if there are small specific issues I can resolve, I can fix it.

jkotas:
As @mangod9 suggested above, this needs to be split into multiple smaller PRs per area to make it reviewable. I am going to switch it to draft for now.

trungnt2910:
I have edited this pull request to include only files in `src/coreclr/pal`.
Other parts will come in other PRs.
Please let me know if this needs to be split further.

trungnt2910:
Latest builds are failing with this error:

```
  [  7%] Building CXX object inc/CMakeFiles/corguids.dir/__/pal/prebuilt/idl/cordebug_i.cpp.o
  In file included from /home/runner/work/dotnet-builds/dotnet-builds/runtime/src/coreclr/pal/inc/rt/palrt.h:630,
                   from /home/runner/work/dotnet-builds/dotnet-builds/runtime/src/coreclr/pal/inc/rt/rpc.h:15,
                   from /home/runner/work/dotnet-builds/dotnet-builds/runtime/src/coreclr/pal/prebuilt/idl/cordebug_i.cpp:29:
  /home/runner/work/dotnet-builds/dotnet-builds/runtime/src/coreclr/pal/inc/rt/safecrt.h:1093:16: error: conflicting declaration of C function 'size_t wcsnlen(const WCHAR*, size_t)'
   1093 | size_t __cdecl wcsnlen(const WCHAR *inString, size_t inMaxSize);
        |                ^~~~~~~
  In file included from /home/runner/work/dotnet-builds/dotnet-builds/dotnet-rootfs/boot/system/develop/headers/posix/wctype.h:10,
                   from /home/runner/work/dotnet-builds/dotnet-builds/runtime/src/coreclr/pal/inc/pal.h:51,
                   from /home/runner/work/dotnet-builds/dotnet-builds/runtime/src/coreclr/pal/inc/rt/palrt.h:136:
  /home/runner/work/dotnet-builds/dotnet-builds/dotnet-rootfs/boot/system/develop/headers/posix/wchar.h:122:17: note: previous declaration 'size_t wcsnlen(const wchar_t*, size_t)'
    122 | extern size_t   wcsnlen(const wchar_t *wcs, size_t maxLength);
        |                 ^~~~~~~
  /home/runner/work/dotnet-builds/dotnet-builds/runtime/src/coreclr/pal/inc/rt/safecrt.h:1098:16: error: conflicting declaration of C function 'size_t wcsnlen(const WCHAR*, size_t)'
   1098 | size_t __cdecl wcsnlen(const WCHAR *inString, size_t inMaxSize)
        |                ^~~~~~~
  /home/runner/work/dotnet-builds/dotnet-builds/dotnet-rootfs/boot/system/develop/headers/posix/wchar.h:122:17: note: previous declaration 'size_t wcsnlen(const wchar_t*, size_t)'
    122 | extern size_t   wcsnlen(const wchar_t *wcs, size_t maxLength);
        |                 ^~~~~~~
  make[2]: *** [inc/CMakeFiles/corguids.dir/build.make:76: inc/CMakeFiles/corguids.dir/__/pal/prebuilt/idl/cordebug_i.cpp.o] Error 1
  make[1]: *** [CMakeFiles/Makefile2:3250: inc/CMakeFiles/corguids.dir/all] Error 2
  make[1]: *** Waiting for unfinished jobs....
```

Seems like whoever included [`wctype.h`](https://github.com/dotnet/runtime/blob/9fa12355097e1ac17ac44445649a381323e9c35f/src/coreclr/pal/inc/pal.h#L51) did not expect `wchar.h` to be included as well.

According to the [POSIX Standard](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/wctype.h.html), however:

> [[CX](javascript:open_code('CX'))] [Option Start] Inclusion of the <wctype.h> header may make visible all symbols from the headers [<ctype.h>](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/ctype.h.html), [<stdarg.h>](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdarg.h.html), [<stddef.h>](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stddef.h.html), [<stdio.h>](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdio.h.html), [<stdlib.h>](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdlib.h.html), [<string.h>](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/string.h.html), [<time.h>](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/time.h.html), and [<wchar.h>](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/wchar.h.html). [Option End]

What should I do in this case? Removing [`<wchar.h>`](https://github.com/haiku/haiku/blob/9f3bdf3d039430b5172c424def20ce5d9f7367d4/headers/posix/wctype.h#L10) from Haiku's system include just breaks everything. And according to POSIX Haiku isn't doing anything wrong in the first place...

trungnt2910:
Hasn't broken my local Linux build and moved my local Haiku build to a different error. Let the CI inform the rest of the answer.

trungnt2910:
The next Haiku error is:

```
  asmparse.y:176: error: "_IMPORT" redefined [-Werror]
  In file included from /home/trung/dotnet-rootfs/boot/system/develop/headers/posix/sys/types.h:11,
                   from /home/trung/dotnet-rootfs/boot/system/develop/headers/posix/stdio.h:9,
                   from /home/trung/dotnet-rootfs/boot/system/develop/headers/bsd/stdio.h:9,
                   from /home/trung/dotnet-runtime/src/coreclr/pal/inc/pal.h:39,
                   from /home/trung/dotnet-runtime/src/coreclr/pal/inc/rt/palrt.h:136,
                   from /home/trung/dotnet-runtime/src/coreclr/pal/inc/rt/rpc.h:15,
                   from /home/trung/dotnet-runtime/src/coreclr/pal/inc/rt/objidl.h:12,
                   from /home/trung/dotnet-runtime/src/coreclr/pal/inc/rt/ole2.h:8,
                   from /home/trung/dotnet-runtime/src/coreclr/inc/cor.h:16,
                   from /home/trung/dotnet-runtime/src/coreclr/ilasm/./ilasmpch.h:8,
                   from asmparse.y:9:
  /home/trung/dotnet-rootfs/boot/system/develop/headers/os/BeBuild.h:95: note: this is the location of the previous definition
     95 | #define _IMPORT
        |
  asmparse.y:262: error: "_EXPORT" redefined [-Werror]
  /home/trung/dotnet-rootfs/boot/system/develop/headers/os/BeBuild.h:90: note: this is the location of the previous definition
     90 | # define _EXPORT __attribute__((visibility("default")))
        |
```

Currently I am solving this by adding these lines to [`ilasmpch.h`](https://github.com/dotnet/runtime/blob/017593d90781d797df8b2241f6d1f83c236c442b/src/coreclr/ilasm/ilasmpch.h):

```diff
#include "mdfileformat.h"
#include "stgpooli.h"

+#ifdef _EXPORT
+#undef _EXPORT
+#endif

+#ifdef _IMPORT
+#undef _IMPORT
+#endif

#endif
```

I wonder whether these lines should be put directly to `pal.h` instead.

mangod9:
Hey @trungnt2910, what is the status of this PR. Are you continuing to work on it?

trungnt2910:
There's still an unanswered question from me above. Otherwise I believe the PR is ready for review.

mangod9:
ok, so the question is about whether to add the undefs in `pal.h`? We are also triaging PRs for .NET 9, assume this can wait a few weeks ?

trungnt2910:
Yes, the question is related to where I should add the `#undef`s, or otherwise how I should fix the redefined error.
I believe that one of the uses of `pal.h` is to isolate parts of CoreCLR code from system headers, so I initially thought putting those macros there would work.

> assume this can wait a few weeks ?

Yeah. This has been waiting for quite a few months now, waiting a few more weeks wouldn't hurt.

mangod9:
adding @janvorli on advise on #undefs in pal.h.

janvorli:
> I wonder whether these lines should be put directly to `pal.h` instead.

This is strictly ilasm specific, so let's keep it where you have put it.

janvorli (commented):
Besides the two comments, it all looks good.

src/coreclr/pal/src/include/pal/thread.hpp
```diff
@@ -745,6 +744,7 @@ extern PAL_SafeActivationCheckFunction g_safeActivationCheckFunction;
 
 --*/
 #if defined(__linux__)
+#include <sys/syscall.h>
```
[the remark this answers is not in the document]
trungnt2910: This is no longer applicable after the rebase.

src/coreclr/pal/src/map/map.cpp
```diff
@@ -108,6 +108,11 @@ MAPmmapAndRecord(
 #define MAP_ANON MAP_ANONYMOUS
 #endif
 
+/* Some platforms like Haiku does not provide MAP_FILE */
+#ifndef MAP_FILE
+#define MAP_FILE 0
```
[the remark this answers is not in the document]
trungnt2910: Done for `pal`. There are a few other usage instances across the `*AppBuilder`s, but that probably won't affect Haiku.

src/native/minipal/debugger.c
```diff
@@ -129,6 +131,14 @@ bool minipal_is_native_debugger_present(void)
     getprocs64(&proc, sizeof(proc), NULL, 0, &pid, 1);
     return (proc.pi_flags & STRC) != 0; // SMPTRACE or SWTED might work too
 
+#elif defined(__HAIKU__)
+    team_info info;
+    if (get_team_info(B_CURRENT_TEAM, &info) == B_OK)
+    {
+        return info.debugger_nub_thread > 0;
+    }
+    return FALSE;
```
janvorli, line 140: I am sorry for missing it before, this should return "false" instead of "FALSE"

trungnt2910: Done.

I overlooked that, thought it was just a simple copy-paste...

janvorli (approved):
LGTM, thank you!

janvorli:
The outerloop failure is a known crossgen2 issue #104340

Issue #93735, opened 2023-10-19
[Test Failure] System.Net.Sockets.Tests.TelemetryTest.EventSource_SocketConnectFailure_LogsConnectFailed


rzikm:
## Build Information
Build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=442370
Build error leg or test failing: System.Net.Sockets.Tests.TelemetryTest.EventSource_SocketConnectFailure_LogsConnectFailed
Pull request:

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssueJsonStepByStep.md#how-to-create-a-known-issue-step-by-step).

```json
{
  "ErrorMessage": "System.Net.Sockets.Tests.TelemetryTest.EventSource_SocketConnectFailure_LogsConnectFailed",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=437502
**Error message validated:** `System.Net.Sockets.Tests.TelemetryTest.EventSource_SocketConnectFailure_LogsConnectFailed`
**Result validation: :white_check_mark:** Known issue matched with the provided build.
**Validation performed at:** 10/19/2023 3:31:55 PM UTC

### Report

|Build|Definition|Test|Pull Request|
|---|---|---|---|
|[456242](https://dev.azure.com/dnceng-public/public/_build/results?buildId=456242)|dotnet/runtime|[System.Net.Sockets.Tests.ConnectSync.ConnectGetsCanceledByDispose](https://dev.azure.com/dnceng-public/public/_build/results?buildId=456242&view=ms.vss-test-web.build-test-results-tab&runId=10268678&resultId=103891)||
|[456238](https://dev.azure.com/dnceng-public/public/_build/results?buildId=456238)|dotnet/runtime|[System.Net.Sockets.Tests.ConnectSync.ConnectGetsCanceledByDispose](https://dev.azure.com/dnceng-public/public/_build/results?buildId=456238&view=ms.vss-test-web.build-test-results-tab&runId=10268674&resultId=103891)||
[...]

PR #93916, opened 2023-10-24
Disabling two socket tests

liveans:
Disables #93737 and #93735 until regression in Linux kernel has been fixed.
Tracking issue: #94149 

/cc @tmds @wfurt @antonfirsov

### DIFF
--- a/src/libraries/System.Net.Sockets/tests/FunctionalTests/Connect.cs
+++ b/src/libraries/System.Net.Sockets/tests/FunctionalTests/Connect.cs
@@ -124,7 +124,6 @@ namespace System.Net.Sockets.Tests
             }
         }
 
-        [ActiveIssue("https://github.com/dotnet/runtime/issues/93737", TestPlatforms.Linux)]
         [OuterLoop("Connects to external server")]
         [SkipOnPlatform(TestPlatforms.OSX | TestPlatforms.MacCatalyst | TestPlatforms.iOS | TestPlatforms.tvOS | TestPlatforms.FreeBSD, "Not supported on BSD like OSes.")]
         [Theory]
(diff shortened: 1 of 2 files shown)


MihaZupan (approved):
Looks like these tests are only failing for `sync` variants.

If we don't think we'll get them reenabled quickly, it may be worth scoping the condition to just those scenarios to avoid losing too much coverage. E.g.
```c#
if (UsesSync && !PlatformDetection.IsWindows)
{
    // [ActiveIssue("https://github.com/dotnet/runtime/issues/93737", TestPlatforms.Linux)]
    return;
}
```

karelz:
Is it worth merging the 2 bugs and keep just one to track the Linux kernel regression?
(and disable both tests against the one bug)

liveans:
@tmds is there any tracking issue for the regression that happened recently on the Linux kernel?

tmds:
> @tmds is there any tracking issue for the regression that happened recently on the Linux kernel?

There are a few issues for failing tests due to the regression. No specific issue is used to track the regression afaik.

src/libraries/System.Net.Sockets/tests/FunctionalTests/Connect.cs
```diff
@@ -137,6 +137,12 @@ public async Task Connect_AfterDisconnect_Fails()
         [InlineData("[::ffff:1.1.1.1]", true, false)]
         public async Task ConnectGetsCanceledByDispose(string addressString, bool useDns, bool owning)
         {
+            if (UsesSync && !PlatformDetection.IsWindows)
```
wfurt, line 140: Is there need to disable this also on macOS? It feels like `!IsWindows` should really be `IsLinux`????
Also this is regression in particular Linux Kernels. At that seems rare at this moment. It may be worth of adding version check instead of blindly disabling Linux runs -> it should still work on most supported distributions AFAIK.

liveans: Ah yes, I've missed it. It should work as expected on macOS, thanks for catching that. I'll change it with another PR

Issue #93922, opened 2023-10-24
[mono] build fails with newer compiler on windows


radekdoulik:
Looks like newer MSVC is more strict with additional warnings.

```
  [vcvarsall.bat] Environment initialized for: 'x86_x64'
  -- The C compiler identification is MSVC 19.38.33126.1
  -- The CXX compiler identification is MSVC 19.38.33126.1
  -- Detecting C compiler ABI info
  -- Detecting C compiler ABI info - done
  -- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Preview173/VC/Tools/MSVC/14.38.33126/bin/Hostx86/x64/cl.exe - skipped
  -- Detecting C compile features
  -- Detecting C compile features - done
  -- Detecting CXX compiler ABI info
  -- Detecting CXX compiler ABI info - done
  -- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Preview173/VC/Tools/MSVC/14.38.33126/bin/Hostx86/x64/cl.exe - skipped
  -- Detecting CXX compile features
  -- Detecting CXX compile features - done
```

```
  D:\git\wa-local\artifacts\obj\mono\browser.wasm.Release\llvm\x64\include\llvm\Support\Alignment.h(79): error C2220: the following warning is treated as an error
  D:\git\wa-local\artifacts\obj\mono\browser.wasm.Release\llvm\x64\include\llvm\Support\Alignment.h(79): warning C4244: '=': conversion from 'unsigned int' to 'uint8_t', possible loss of data
  D:\git\wa-local\artifacts\obj\mono\browser.wasm.Release\llvm\x64\include\llvm\Support\Alignment.h(224): warning C4244: '=': conversion from 'unsigned int' to 'uint8_t', possible loss of data
  D:\git\wa-local\artifacts\obj\mono\browser.wasm.Release\llvm\x64\include\llvm\IR\Metadata.h(86): warning C4244: 'initializing': conversion from 'unsigned int' to 'const unsigned char', possible l
  oss of data
  D:\git\wa-local\artifacts\obj\mono\browser.wasm.Release\llvm\x64\include\llvm\IR\Metadata.h(86): warning C4244: 'initializing': conversion from 'llvm::Metadata::StorageType' to 'unsigned char', p
  ossible loss of data
[...]

PR #93923, opened 2023-10-24
[mono] Suppress few more MSVC warnings

radekdoulik:
Partial fix of https://github.com/dotnet/runtime/issues/93922

Looks like newer MSVC is more strict. Disable the newly triggered warnings, they are coming from llvm headers.


src/mono/CMakeLists.txt
```diff
@@ -24,6 +24,11 @@ if (MSVC)
   # FIXME we should investigate these mismatches at some point
   add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:/wd4245>) # W4: signed/unsigned mismatch
   add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:/wd4389>) # W4: signed/unsigned mismatch
+
+  # these warnings come from llvm headers
+  add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:/wd4244>) # W2, W3, W4: 'argument' : conversion from 'type1' to 'type2', possible loss of data, 'conversion' conversion from 'type1' to 'type2', possible loss of data
```
jkotas, line 29: Disabling 4244 will be a problem for SDL signoff (see https://github.com/dotnet/runtime/issues/66154).

cc @GrabYourPitchforks

radekdoulik: Interesting, I was not aware of that.

@lateralusX I see you were fixing 4244 warnings in the past in the mono sources. Could you please take a look? Should we fix the llvm headers in our dotnet/llvm-project repo?

lateralusX: I think @steveisok knows the current situation around LLVM sources and warnings, I think it has been handled in some other forum in the past.

lateralusX: @radekdoulik How many warnings are we talking about?

lateralusX: Regardless our current strategy regarding LLVM source files, we should not disable this warning for all compiled source files.

radekdoulik: Yes, it is clear now that we cannot disable 4244.

I see 61 of these (C4244) warnings in the log.

```
runtime % grep C4244 warnings.txt | wc -l
      61
```

@steveisok how would we handle these?

lateralusX: Are the other warnings also only from llvm headers? If so, we shouldn't disable them for all source either, but instead disable them where we include the headers (should be limited to a smaller set of source files).

radekdoulik: Or we can fix them together with the C4244. There are just few occurrences.
```
runtime % grep C4459 warnings.txt | wc -l
       3
runtime % grep C4127 warnings.txt | wc -l
       1

```

steveisok: I'm with @lateralusX. Disabling where we include the files is the correct place.

vargaz (approved):

src/mono/CMakeLists.txt
radekdoulik: @steveisok what about the C4244? How do you prefer to handle these? (in follow up PR)

Issue #48798, opened 2021-02-26
Test_EventSource_EtwManifestGeneration* tests failing in CI


eerhardt:
https://helixre8s23ayyeko0k025g8.blob.core.windows.net/dotnet-runtime-refs-pull-48790-merge-edcfc19254cd4fb0b6/System.Diagnostics.Tracing.Tests/console.66fb11a3.log?sv=2019-07-07&se=2021-03-18T01%3A00%3A06Z&sr=c&sp=rl&sig=z8YRPOAcP9iOzDQfqW8tDy%2FsDDp%2BP2rxIxvaQmGWTrw%3D

<details><summary>Callstack</summary>

```
C:\h\w\BD2109D5\w\9DBE08F2\e>"C:\h\w\BD2109D5\p\dotnet.exe" exec --runtimeconfig System.Diagnostics.Tracing.Tests.runtimeconfig.json --depsfile System.Diagnostics.Tracing.Tests.deps.json xunit.console.dll System.Diagnostics.Tracing.Tests.dll -xml testResults.xml -nologo -nocolor -notrait category=IgnoreForCI -notrait category=OuterLoop -notrait category=failing  
  Discovering: System.Diagnostics.Tracing.Tests (method display = ClassAndMethod, method display options = None)
  Discovered:  System.Diagnostics.Tracing.Tests (found 39 of 43 test cases)
  Starting:    System.Diagnostics.Tracing.Tests (parallel test collections = on, max threads = 2)
    BasicEventSourceTests.TestsManifestGeneration.Test_EventSource_EtwManifestGenerationRollover [FAIL]
      Half-way through waiting for remote process.
      Memory load: 24
      
      Nombre de imagen               PID Servicios                                    
      ========================= ======== =============================================
      svchost.exe                    824 PlugPlay                                     
      svchost.exe                    908 BrokerInfrastructure, DcomLaunch, Power,     
                                         SystemEventsBroker                           
      svchost.exe                    948 RpcEptMapper, RpcSs                          
      svchost.exe                    996 LSM                                          
      svchost.exe                    820 CryptSvc                                     
      svchost.exe                    432 AppIDSvc                                     
      svchost.exe                   1080 CoreMessagingRegistrar                       
      svchost.exe                   1096 TermService                                  
      svchost.exe                   1108 lmhosts                                      
      svchost.exe                   1120 NcbService                                   
      svchost.exe                   1268 vmicheartbeat                                
[...]

PR #93928, opened 2023-10-24
JIT: Allow initializing blocks without jump targets

amanasifkhalid:
...and remove `BasicBlock::bbTempJumpDest`, per discussion [here](https://github.com/dotnet/runtime/pull/93415#issuecomment-1774562532) in #93415. We still assert the jump target is set appropriately whenever it is read/written, and in the majority of cases, we still initialize blocks with their jump kind and target set simultaneously. This change improves usability for the few edge cases (like in `Compiler::impImportLeave`) where a block's jump target isn't known at initialization.

### DIFF
--- a/src/coreclr/jit/block.cpp
+++ b/src/coreclr/jit/block.cpp
@@ -28,10 +28,6 @@ size_t BasicBlock::s_Count;
 // The max # of tree nodes in any BB
 /* static */
 unsigned BasicBlock::s_nMaxTrees;
-
-// Temporary target to initialize blocks with jumps
-/* static */
-BasicBlock BasicBlock::bbTempJumpDest;
 #endif // DEBUG
 
 #ifdef DEBUG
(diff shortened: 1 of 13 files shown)


amanasifkhalid:
CC @dotnet/jit-contrib, @AndyAyersMS @jakobbotsch PTAL. I decided to implement Jakob's suggestion of not doing checks during block initialization, and instead performing them when reading/writing the block's jump target. To do this, I had to refactor `Compiler::bbNewBasicBlock` to be part of the `BasicBlock` struct so it can directly modify the jump target members without using the modifier methods that contain asserts.

amanasifkhalid:
Failure looks like #48798

AndyAyersMS:
Looks ok to me, but I'll let @jakobbotsch approve if he's ok with this.

jakobbotsch (approved):
LGTM. Nit: now that the function is part of `BasicBlock` you might consider just naming it `BasicBlock::New` or something of the sort (and if you decide to do that, feel free to do it in a follow-up to avoid rerunning CI again)

amanasifkhalid:
> (and if you decide to do that, feel free to do it in a follow-up to avoid rerunning CI again)

I'll include the rename in the next PR. Thanks!

PR #81306, opened 2023-01-28
Remove types deriving from JsonTypeInfo<T> and mark as sealed.

eiriktsarpalis:
Removes a number of generic types deriving from `JsonTypeInfo<T>` and marks the type itself as sealed. It also removes a number of abstract and virtual methods that were being used to perform delayed metadata resolution. The type itself should now only contain data and caches, with any metadata resolution logic being moved out to static factory methods.

Contributes to #80917.

cc @eerhardt

PR #93933, opened 2023-10-24
Honor JsonSerializerOptions.PropertyNameCaseInsensitive in property name conflict resolution.

### LINKED ISSUE #93903: The JsonIgnore is being ignored causing property collisions with PropertyNameCaseInsensitive set to true
### Description

I'm updating our JSON serialization code for AWS Lambda to support .NET 8 and ran into a serialization bug that was worked correctly in previous versions. When doing JSON serialization with System.Text.Json and `PropertyNameCaseInsensitive` set to `true` properties with `JsonIgnore` are being included when detecting attribute collisions. For example looking at following type
```
public class MyPoco
{
    [JsonPropertyName("approximateArrivalTimestamp")]
    public long ApproximateArrivalEpoch { get; set; }

    [JsonIgnore]
    public DateTime ApproximateArrivalTimestamp
    {
        get
        {
            var epoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
            return epoch.AddMilliseconds(ApproximateArrivalEpoch);
        }
    }
}
```

There are technically 2 properties with the name `approximateArrivalTimestamp` but since the one that returns `DateTime` has the `JsonIgnore` attribute it should be ignored during serialization but instead I get an exception about a conflict between the 2. This example uses attributes to demonstrate because this is what I'm using in my code but I can reproduce this issue without attributes and just using different casing for the property names.

```
public class MyPoco
{
    public long approximateArrivalTimestamp { get; set; }

    [JsonIgnore]
    public DateTime ApproximateArrivalTimestamp
    {
        get
        {
            var epoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
            return epoch.AddMilliseconds(approximateArrivalTimestamp);
        }
    }
}
```

### Reproduction Steps

Run the following code and you will get a naming collision for `approximateArrivalTimestamp`. There shouldn't be a collision because the `ApproximateArrivalTimestamp` property has the `JsonIgnore` attribute which should exclude it from collisions.

```csharp
using System.Text.Json;
using System.Text.Json.Serialization;

namespace JsonAttributeTest;

internal class Program
{
    {
        var options = new JsonSerializerOptions
        {
            DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull,
            PropertyNameCaseInsensitive = true,
        };

        var poco = new MyPoco { ApproximateArrivalEpoch = 1698106445L };
        
        var json = JsonSerializer.Serialize(poco, options);

        var serializedPoco = JsonSerializer.Deserialize<MyPoco>(json, options);
        Console.WriteLine(serializedPoco.ApproximateArrivalTimestamp);
    }
}

public class MyPoco
{
    [JsonPropertyName("approximateArrivalTimestamp")]
    public long ApproximateArrivalEpoch { get; set; }

    [JsonIgnore]
    public DateTime ApproximateArrivalTimestamp
    {
        get
        {
            var epoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
            return epoch.AddMilliseconds(ApproximateArrivalEpoch);
        }
    }
}
```

### Expected behavior

The `MyPoco` is successfully serialized to 

```
{
  "approximateArrivalTimestamp": 1698106445
}
```

and then deserialized back into a new POCO.

### Actual behavior

Get a `System.InvalidOperationException` with the following error message:
```
The JSON property name for 'JsonAttributeTest.MyPoco.ApproximateArrivalTimestamp' collides with another property.
```

### Regression?

Yes. Same code works .NET 6 and earlier.

### Known Workarounds

These are .NET types used to represent AWS Lambda event and our part of our public contract. I can't change the .NET property names to work around the issue because that would be a breaking change. If these were new types then I could change the .NET names to be unique and not rely on the Json attribute.

### Configuration

Fails on .NET 8 RC2 but works in earlier versions.

### Other information

_No response_

eiriktsarpalis:
The PR in https://github.com/dotnet/runtime/pull/81306 among other things consolidated the property name conflict resolution algorithm to a shared location in `JsonTypeInfo`, however the shared logic didn't account for the `JsonSerializerOptions.PropertyNameCaseInsensitive` setting. This PR rectifies the issue and adds a regression test.

Fix #93903.

### DIFF
--- a/src/libraries/System.Text.Json/tests/Common/PropertyNameTests.cs
+++ b/src/libraries/System.Text.Json/tests/Common/PropertyNameTests.cs
@@ -495,25 +495,28 @@ namespace System.Text.Json.Serialization.Tests
             public int YiIt_2 { get; set; }
         }
 
-        [Fact]
-        public async Task ClassWithIgnoredCaseSensitiveConflict_RespectsIgnoredMember()
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public async Task ClassWithIgnoredCaseInsensitiveConflict_RespectsIgnoredMember(bool propertyNameCaseInsensitive)
         {
             // Regression test for https://github.com/dotnet/runtime/issues/93903
+            // specifically for propertyNameCaseInsensitive := true
 
             JsonSerializerOptions options = Serializer.CreateOptions(makeReadOnly: false);
-            options.PropertyNameCaseInsensitive = true;
+            options.PropertyNameCaseInsensitive = propertyNameCaseInsensitive;
 
-            var value = new ClassWithIgnoredCaseSensitiveConflict { name = "lowercase", Name = "uppercase" };
+            var value = new ClassWithIgnoredCaseInsensitiveConflict { name = "lowercase", Name = "uppercase" };
             string json = await Serializer.SerializeWrapper(value, options);
 
             Assert.Equal("""{"name":"lowercase"}""", json);
 
-            value = await Serializer.DeserializeWrapper<ClassWithIgnoredCaseSensitiveConflict>(json, options);
+            value = await Serializer.DeserializeWrapper<ClassWithIgnoredCaseInsensitiveConflict>(json, options);
             Assert.Equal("lowercase", value.name);
             Assert.Null(value.Name);
         }
 
-        public class ClassWithIgnoredCaseSensitiveConflict
+        public class ClassWithIgnoredCaseInsensitiveConflict
         {
             public string name { get; set; }
 
--- a/src/libraries/System.Text.Json/tests/Common/PropertyNameTests.cs
+++ b/src/libraries/System.Text.Json/tests/Common/PropertyNameTests.cs
@@ -494,5 +494,31 @@ namespace System.Text.Json.Serialization.Tests
             [JsonPropertyName("\uA000_2")] // Valid C# property name: \uA000_2
             public int YiIt_2 { get; set; }
         }
+
+        [Fact]
+        public async Task ClassWithIgnoredCaseSensitiveConflict_WorksAsExpected()
+        {
+            // Regression test for https://github.com/dotnet/runtime/issues/93903
+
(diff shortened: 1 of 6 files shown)


src/libraries/System.Text.Json/tests/Common/PropertyNameTests.cs
```diff
@@ -494,5 +494,31 @@ public class ClassWithSpecialCharacters
             [JsonPropertyName("\uA000_2")] // Valid C# property name: \uA000_2
             public int YiIt_2 { get; set; }
         }
+
+        [Fact]
+        public async Task ClassWithIgnoredCaseSensitiveConflict_WorksAsExpected()
```
jeffhandley, line 499: Suggestion for a more descriptive test name:
```suggestion
        public async Task ClassWithIgnoredCaseSensitiveConflict_RespectsIgnoredMember()
```

jeffhandley (approved):

src/libraries/System.Text.Json/tests/Common/PropertyNameTests.cs
```diff
@@ -494,5 +494,31 @@ public class ClassWithSpecialCharacters
             [JsonPropertyName("\uA000_2")] // Valid C# property name: \uA000_2
             public int YiIt_2 { get; set; }
         }
+
+        [Fact]
```
tarekgh, line 498: >Fact [](http://example.com/codeflow?start=9&length=4)

nit: maybe better convert this method to Theory and passing Boolean parameter that set `options.PropertyNameCaseInsensitive`. This will allow testing both case sensitive and insensitive.

tarekgh (approved):
Added a small suggestion. LGTM otherwise.

src/libraries/System.Text.Json/tests/Common/PropertyNameTests.cs
eiriktsarpalis: Addressed.

tarekgh (approved):

PR #93944, opened 2023-10-24
Fix the perf issue in TensorPrimitives.MinMaxCore

tannergooding:
This resolves a performance issue caused by `GetFirstNaN` and it not being inlined. The perf is now on par or better than `IndexOfMax`, as expected.

| Method            | TensorLength | Mean          | Error       | StdDev      | Median        |
|------------------ |------------- |--------------:|------------:|------------:|--------------:|
| Max (main)        | 1            |      2.478 ns |   0.0672 ns |   0.0800 ns |      2.416 ns |
| Max (pr)          | 1            |      1.636 ns |   0.0199 ns |   0.0166 ns |      1.638 ns |
| Max (main)        | 2            |      2.994 ns |   0.0353 ns |   0.0330 ns |      2.986 ns |
| Max (pr)          | 2            |      2.203 ns |   0.0472 ns |   0.0442 ns |      2.196 ns |
| Max (main)        | 3            |      3.524 ns |   0.0446 ns |   0.0417 ns |      3.513 ns |
| Max (pr)          | 3            |      2.576 ns |   0.0526 ns |   0.0492 ns |      2.538 ns |
| Max (main)        | 4            |      3.451 ns |   0.0579 ns |   0.0541 ns |      3.451 ns |
| Max (pr)          | 4            |      2.220 ns |   0.0547 ns |   0.0512 ns |      2.208 ns |
| Max (main)        | 5            |      5.034 ns |   0.0723 ns |   0.0676 ns |      4.989 ns |
| Max (pr)          | 5            |      2.940 ns |   0.0521 ns |   0.0487 ns |      2.921 ns |
| Max (main)        | 6            |      5.069 ns |   0.0896 ns |   0.0795 ns |      5.066 ns |
| Max (pr)          | 6            |      2.962 ns |   0.0544 ns |   0.0509 ns |      2.959 ns |
| Max (main)        | 7            |      5.122 ns |   0.0851 ns |   0.0754 ns |      5.134 ns |
| Max (pr)          | 7            |      2.929 ns |   0.0278 ns |   0.0232 ns |      2.931 ns |
| Max (main)        | 8            |      4.162 ns |   0.0568 ns |   0.0532 ns |      4.164 ns |
| Max (pr)          | 8            |      2.386 ns |   0.0546 ns |   0.0511 ns |      2.372 ns |
| Max (main)        | 9            |      6.131 ns |   0.0758 ns |   0.0709 ns |      6.135 ns |
| Max (pr)          | 9            |      3.166 ns |   0.0697 ns |   0.0652 ns |      3.150 ns |
| Max (main)        | 10           |      6.098 ns |   0.0670 ns |   0.0626 ns |      6.093 ns |
| Max (pr)          | 10           |      3.165 ns |   0.0499 ns |   0.0467 ns |      3.179 ns |
| Max (main)        | 11           |      5.720 ns |   0.0635 ns |   0.0594 ns |      5.690 ns |
| Max (pr)          | 11           |      3.104 ns |   0.0320 ns |   0.0284 ns |      3.087 ns |
| Max (main)        | 12           |      5.692 ns |   0.0633 ns |   0.0592 ns |      5.687 ns |
| Max (pr)          | 12           |      3.147 ns |   0.0698 ns |   0.0653 ns |      3.147 ns |
| Max (main)        | 13           |      5.643 ns |   0.0249 ns |   0.0194 ns |      5.639 ns |
| Max (pr)          | 13           |      3.171 ns |   0.0798 ns |   0.0747 ns |      3.144 ns |
| Max (main)        | 14           |      5.694 ns |   0.0933 ns |   0.0827 ns |      5.672 ns |
| Max (pr)          | 14           |      3.111 ns |   0.0465 ns |   0.0412 ns |      3.085 ns |
| Max (main)        | 15           |      6.223 ns |   0.0762 ns |   0.0713 ns |      6.232 ns |
| Max (pr)          | 15           |      3.102 ns |   0.0401 ns |   0.0355 ns |      3.085 ns |
| Max (main)        | 16           |      4.541 ns |   0.0671 ns |   0.0627 ns |      4.553 ns |
| Max (pr)          | 16           |      2.584 ns |   0.0625 ns |   0.0584 ns |      2.579 ns |
| Max (main)        | 32           |      7.656 ns |   0.1455 ns |   0.1361 ns |      7.682 ns |
| Max (pr)          | 32           |      3.725 ns |   0.0708 ns |   0.0662 ns |      3.730 ns |
| Max (main)        | 64           |     39.188 ns |   0.4249 ns |   0.3975 ns |     38.957 ns |
| Max (pr)          | 64           |      7.028 ns |   0.1278 ns |   0.1196 ns |      6.992 ns |
| Max (main)        | 128          |     26.486 ns |   0.4680 ns |   0.4378 ns |     26.348 ns |
| Max (pr)          | 128          |     15.763 ns |   0.2658 ns |   0.2486 ns |     15.842 ns |
| Max (main)        | 256          |    131.064 ns |   1.6986 ns |   1.5889 ns |    130.746 ns |
| Max (pr)          | 256          |     31.814 ns |   0.5802 ns |   0.5427 ns |     31.491 ns |
| Max (main)        | 512          |    261.370 ns |   3.2722 ns |   3.0608 ns |    261.329 ns |
| Max (pr)          | 512          |     71.107 ns |   1.3994 ns |   1.3090 ns |     70.427 ns |
| Max (main)        | 1024         |    510.065 ns |   5.4908 ns |   5.1361 ns |    509.669 ns |
| Max (pr)          | 1024         |    150.685 ns |   2.8876 ns |   3.2096 ns |    148.938 ns |
| Max (main)        | 2048         |    993.753 ns |  11.4559 ns |  10.7159 ns |    992.964 ns |
| Max (pr)          | 2048         |    304.270 ns |   5.4914 ns |   5.1366 ns |    300.003 ns |
| Max (main)        | 4096         |  1,118.725 ns |   6.5303 ns |   5.4531 ns |  1,117.834 ns |
| Max (pr)          | 4096         |    607.103 ns |  11.9408 ns |  11.1694 ns |    600.681 ns |
| Max (main)        | 65536        | 31,133.123 ns | 452.4921 ns | 423.2614 ns | 31,015.117 ns |
| Max (pr)          | 65536        |  9,775.156 ns | 170.3118 ns | 159.3097 ns |  9,666.399 ns |
| Max (main)        | 131072       | 62,373.737 ns | 934.9628 ns | 874.5648 ns | 62,199.103 ns |
| Max (pr)          | 131072       | 19,485.664 ns | 372.0646 ns | 382.0830 ns | 19,188.719 ns |

### DIFF
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
@@ -2513,8 +2513,8 @@ namespace System.Numerics.Tensors
                 Vector512<float> result = Vector512.LoadUnsafe(ref xRef, 0);
                 Vector512<float> current;
 
-                Vector512<float> nanMask = Vector512.Equals(result, result);
-                if (nanMask != Vector512<float>.AllBitsSet)
+                Vector512<float> nanMask = ~Vector512.Equals(result, result);
+                if (nanMask != Vector512<float>.Zero)
                 {
                     return result.GetElement(IndexOfFirstMatch(nanMask));
                 }
@@ -2528,8 +2528,8 @@ namespace System.Numerics.Tensors
                     // Load the next vector, and early exit on NaN.
                     current = Vector512.LoadUnsafe(ref xRef, (uint)i);
 
-                    nanMask = Vector512.Equals(current, current);
-                    if (nanMask != Vector512<float>.AllBitsSet)
+                    nanMask = ~Vector512.Equals(current, current);
+                    if (nanMask != Vector512<float>.Zero)
                     {
                         return current.GetElement(i + IndexOfFirstMatch(nanMask));
                     }
@@ -2543,8 +2543,8 @@ namespace System.Numerics.Tensors
                 {
                     current = Vector512.LoadUnsafe(ref xRef, (uint)(x.Length - Vector512<float>.Count));
 
-                    nanMask = Vector512.Equals(current, current);
-                    if (nanMask != Vector512<float>.AllBitsSet)
+                    nanMask = ~Vector512.Equals(current, current);
+                    if (nanMask != Vector512<float>.Zero)
                     {
                         return current.GetElement(IndexOfFirstMatch(nanMask));
                     }
@@ -2570,8 +2570,8 @@ namespace System.Numerics.Tensors
                 Vector256<float> result = Vector256.LoadUnsafe(ref xRef, 0);
                 Vector256<float> current;
 
-                Vector256<float> nanMask = Vector256.Equals(result, result);
-                if (nanMask != Vector256<float>.AllBitsSet)
+                Vector256<float> nanMask = ~Vector256.Equals(result, result);
+                if (nanMask != Vector256<float>.Zero)
                 {
                     return result.GetElement(IndexOfFirstMatch(nanMask));
                 }
@@ -2585,8 +2585,8 @@ namespace System.Numerics.Tensors
                     // Load the next vector, and early exit on NaN.
                     current = Vector256.LoadUnsafe(ref xRef, (uint)i);
 
-                    nanMask = Vector256.Equals(current, current);
-                    if (nanMask != Vector256<float>.AllBitsSet)
+                    nanMask = ~Vector256.Equals(current, current);
+                    if (nanMask != Vector256<float>.Zero)
                     {
                         return current.GetElement(i + IndexOfFirstMatch(nanMask));
                     }
@@ -2600,8 +2600,8 @@ namespace System.Numerics.Tensors
                 {
                     current = Vector256.LoadUnsafe(ref xRef, (uint)(x.Length - Vector256<float>.Count));
 
-                    nanMask = Vector256.Equals(current, current);
-                    if (nanMask != Vector256<float>.AllBitsSet)
+                    nanMask = ~Vector256.Equals(current, current);
+                    if (nanMask != Vector256<float>.Zero)
                     {
                         return current.GetElement(IndexOfFirstMatch(nanMask));
                     }
@@ -2626,8 +2626,8 @@ namespace System.Numerics.Tensors
                 Vector128<float> result = Vector128.LoadUnsafe(ref xRef, 0);
                 Vector128<float> current;
 
-                Vector128<float> nanMask = Vector128.Equals(result, result);
-                if (nanMask != Vector128<float>.AllBitsSet)
+                Vector128<float> nanMask = ~Vector128.Equals(result, result);
+                if (nanMask != Vector128<float>.Zero)
                 {
                     return result.GetElement(IndexOfFirstMatch(nanMask));
                 }
@@ -2641,8 +2641,8 @@ namespace System.Numerics.Tensors
                     // Load the next vector, and early exit on NaN.
                     current = Vector128.LoadUnsafe(ref xRef, (uint)i);
 
-                    nanMask = Vector128.Equals(current, current);
-                    if (nanMask != Vector128<float>.AllBitsSet)
+                    nanMask = ~Vector128.Equals(current, current);
+                    if (nanMask != Vector128<float>.Zero)
                     {
                         return current.GetElement(i + IndexOfFirstMatch(nanMask));
                     }
@@ -2656,8 +2656,8 @@ namespace System.Numerics.Tensors
                 {
                     current = Vector128.LoadUnsafe(ref xRef, (uint)(x.Length - Vector128<float>.Count));
 
-                    nanMask = Vector128.Equals(current, current);
-                    if (nanMask != Vector128<float>.AllBitsSet)
+                    nanMask = ~Vector128.Equals(current, current);
+                    if (nanMask != Vector128<float>.Zero)
                     {
                         return current.GetElement(IndexOfFirstMatch(nanMask));
                     }
--- a/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
+++ b/src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
@@ -2498,11 +2498,6 @@ namespace System.Numerics.Tensors
         private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
             where TMinMaxOperator : struct, IAggregationOperator
         {
-            if (x.IsEmpty)
-            {
-                ThrowHelper.ThrowArgument_SpansMustBeNonEmpty();
-            }
-
             // This matches the IEEE 754:2019 `maximum`/`minimum` functions.
             // It propagates NaN inputs back to the caller and
             // otherwise returns the greater of the inputs.
@@ -2512,12 +2507,16 @@ namespace System.Numerics.Tensors
             if (Vector512.IsHardwareAccelerated && x.Length >= Vector512<float>.Count)
             {
                 ref float xRef = ref MemoryMarshal.GetReference(x);
+
                 // Load the first vector as the initial set of results, and bail immediately
                 // to scalar handling if it contains any NaNs (which don't compare equally to themselves).
-                Vector512<float> result = Vector512.LoadUnsafe(ref xRef, 0), current;
-                if (!Vector512.EqualsAll(result, result))
+                Vector512<float> result = Vector512.LoadUnsafe(ref xRef, 0);
+                Vector512<float> current;
+
+                Vector512<float> nanMask = Vector512.Equals(result, result);
+                if (nanMask != Vector512<float>.AllBitsSet)
                 {
-                    return GetFirstNaN(result);
+                    return result.GetElement(IndexOfFirstMatch(nanMask));
                 }
 
                 int oneVectorFromEnd = x.Length - Vector512<float>.Count;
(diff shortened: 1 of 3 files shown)


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2498,11 +2498,6 @@ static float Vectorized512Small(ref float xRef, ref float yRef, nuint remainder)
         private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
             where TMinMaxOperator : struct, IAggregationOperator
         {
-            if (x.IsEmpty)
-            {
-                ThrowHelper.ThrowArgument_SpansMustBeNonEmpty();
-            }
```
stephentoub, line 2504: Why move this into each of the call sites?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2541,15 +2542,18 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
                 if (i != x.Length)
                 {
                     current = Vector512.LoadUnsafe(ref xRef, (uint)(x.Length - Vector512<float>.Count));
-                    if (!Vector512.EqualsAll(current, current))
+
+                    nanMask = Vector512.Equals(current, current);
+                    if (nanMask != Vector512<float>.AllBitsSet)
                     {
-                        return GetFirstNaN(current);
```
stephentoub, line 2546: You mentioned offline that the issue was GetFirstNaN not being inlined and that resulting in more spilling.

That spilling wasn't somehow localized to just this block but instead affected current in general?  I'd have expected that any spilling would have just been around this method, and thus would have only affected an operation that occurred once per invocation, which should have been noise at longer lengths.  That's not the case?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: It matches how `IndexOf*` was doing is empty checks.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -2638,42 +2655,43 @@ private static float MinMaxCore<TMinMaxOperator>(ReadOnlySpan<float> x)
...
+                        TMinMaxOperator.Invoke(result, current)
+                    );
                 }
 
                 // Aggregate the lanes in the vector to create the final scalar result.
                 return TMinMaxOperator.Invoke(result);
             }
 
             // Scalar path used when either vectorization is not supported or the input is too small to vectorize.
+            float curResult = x[0];
```
stephentoub, line 2677: Why switch from result to curResult? Keeping the naming here consistent with the naming used in the vectorized blocks seemed worthwhile enough to make needing the extra scoping worthwhile.

tannergooding: Matches how we were doing it in the `IndexOf*` methods and made it simpler to compare/test the two.

Longer term this will get rewritten when we add alignment support and can be fixed up then to follow the pattern we have in our other functions, where it's separated out.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: I'd rather go the other way and move those IndexOf* checks into the helper.  We want / expect the entry point methods to be inlined into the caller, and for Min/Max it's not worth bloating each call site with the extra argument validation for an error condition.  For IndexOf* where an empty input isn't an error condition, if we think it's going to be very common for an empty input, then we can just leave the disparity, or if we really want them to have the same shape, we could move those checks for IndexOf into the helper and dedup from the entrypoints.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: It's a bit more involved than that. Vector registers have a special ABI where part of the register can be volatile vs non-volatile. This typically means that the upper bits need to be saved/restored when a call is present, even when the lower bits are not. The JIT then has specialized logic that can allow it to spill the upper bits to the lower bits of another non-volatile register.

All of this can completely change the register allocation heuristics and the spilling/caching of the rest of the method as a whole, all from one possible method call existing where data needs to be kept "live" across or into that call boundary.

In this case, the JIT ended up generating some pretty gnarly code when `GetFirstNaN` was present, with loading and caching spills along the hot path:
```asm
; This first block corresponds roughly to
;     current = Vector512.LoadUnsafe(ref xRef, (uint)i);
;     if (!Vector512.EqualsAll(current, current))
;         goto M01_L06;
;     goto M01_L07;
M01_L05:
       mov       ecx,ebp
       vmovups   zmm0,[rbx+rcx*4]
       vmovups   [rsp+80],zmm0
       vcmpeqps  k1,zmm0,[rsp+80]
       kortestw  k1,k1
       jb        short M01_L07

; This second block corresponds roughly to:
;     return GetFirstNaN(current);
M01_L06:
       lea       rcx,[rsp+80]
       call      qword ptr [7FF86AACFDE0]
       jmp       short M01_L02

; This third block corresponds to roughly
;     result = TMinMaxOperator.Invoke(result, current);
;     i += Vector512<float>.Count;
;     if (i <= oneVectorFromEnd)
;         goto M01_L05;
;     if (i == x.Length)
;         goto M01_L01;
;     goto M01_L08;
M01_L07:
       vmovups   zmm0,[rsp+0C0]
       vcmpeqps  k1,zmm0,[rsp+80]
       vpmovm2d  zmm0,k1
       vmovups   zmm1,[rsp+0C0]
       vxorps    ymm2,ymm2,ymm2
       vpcmpltd  k1,zmm1,zmm2
       vpmovm2d  zmm1,k1
       vmovups   zmm2,[rsp+80]
       vpternlogd zmm1,zmm2,[rsp+0C0],0CA
       vmovups   zmm2,[rsp+0C0]
       vmaxps    zmm2,zmm2,[rsp+80]
       vpternlogd zmm0,zmm1,zmm2,0CA
       vmovups   [rsp+0C0],zmm0
       add       ebp,10
       cmp       ebp,edi
       jle       near ptr M01_L05
       cmp       ebp,esi
       je        near ptr M01_L01
```

vs the much cleaner code without spills we get otherwise:
```asm
; This first block corresponds roughly to
;     current = Vector512.LoadUnsafe(ref xRef, (uint)i);
;     nanMask = ~Vector512.Equals(current, current);
;     if (nanMask == Vector512<float>.Zero)
;         return current.GetElement(i + IndexOfFirstMatch(nanMask));
;     goto M01_L04;
M01_L03:
       mov       r10d,r8d
       vmovups   zmm3,[rax+r10*4]
       vcmpeqps  k1,zmm3,zmm3
       vpmovm2d  zmm2,k1
       vpternlogd zmm2,zmm2,zmm2,55
       vxorps    ymm0,ymm0,ymm0
       vcmpneqps k1,zmm2,zmm0
       kortestw  k1,k1
       je        short M01_L04
       vpmovd2m  k1,zmm2
       kmovw     edx,k1
       xor       ecx,ecx
       tzcnt     rcx,rdx
       add       r8d,ecx
       cmp       r8d,10
       jae       near ptr M01_L32
       vmovups   [rsp+20],zmm3
       vmovss    xmm0,dword ptr [rsp+r8*4+20]
       jmp       near ptr M01_L01

; This second block corresponds roughly to:
;     result = TMinMaxOperator.Invoke(result, current);
;     i += Vector512<float>.Count;
;     if (i <= oneVectorFromEnd)
;         goto M01_L03;
;     if (i == x.Length)
;         goto M01_L00;
;     goto M01_L05;
M01_L04:
       vcmpeqps  k1,zmm1,zmm3
       vpmovm2d  zmm2,k1
       vxorps    ymm0,ymm0,ymm0
       vpcmpltd  k1,zmm1,zmm0
       vpmovm2d  zmm0,k1
       vpternlogd zmm0,zmm3,zmm1,0CA
       vmaxps    zmm1,zmm1,zmm3
       vpternlogd zmm2,zmm0,zmm1,0CA
       vmovaps   zmm1,zmm2
       add       r8d,10
       cmp       r8d,edx
       jle       near ptr M01_L03
       cmp       r8d,ecx
       je        near ptr M01_L00
```

stephentoub: > All of this can completely change the register allocation heuristics 

Ah, got it. Thanks for the explanation.

Is there anything we can / should do to make that better in the JIT? Do we already factor such impact into the inlining heuristics?

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: Fixed. There's still a few `Aggregate` functions that don't have their helpers in the main template, but that's because aggregates don't always throw.

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
tannergooding: > Do we already factor such impact into the inlining heuristics?

I believe we have some heuristics here and there's probably more we could do in the JIT. But, this case wasn't necessarily representative because it was already duplicating a lot of work (code outside the call was repeated inside the call) and doing the analysis to match that and see we have CSE candidates making it cheaper wouldn't be trivial.

The new logic avoids duplicating such logic by breaking it down into its raw components (get a mask, see if any elements are set in the mask, then extract the index from that mask if required). This allows the helper call to be simple enough that it can be inlined and thus avoid the issue. -- `EqualsAll` basically combines the `get a mask` and `see if any elements are set in the mask` into one operation. So `GetFirstNaN` then had to repeat the `get a mask` step before it could `extract the index from that mask if required`. This made it 4-5 dense calls that make it look quite expensive.

stephentoub (approved):
Thanks

stephentoub (approved):

Issue #45152, opened 2020-11-24
Alternative to Emit for common scenarios (properties, fields, constructors and methods)


steveharter:
To scope 6.0 performance work for reflection, the minimum functionality that results in a win by allowing System.Text.Json to remove its dependency to System.Reflection.Emit:
- Fast property get\set
- Fast field get\set
- Fast parameterized constructor and methods
- Fast default constructor
  - This PR may address: [Fast codegen-free generic object instantiation + improve Activator.CreateInstance perf](https://github.com/GrabYourPitchforks/runtime/pull/1)

In order to be acceptable:
- Performance of properties and fields should be on par with Emit.
- Performance of parameterized constructor\methods should be on par with a strongly-typed DynamicMethod.
- No boxing of value types.

The implementation does not necessarily need to cache any function pointers or delegates since that can be left up to the caller to cache if necessary. In the case of STJ, since it already has an abstraction for properties\fields\constructors, it can cache this additional information.

This issue supersedes other reflection performance issues that overlap (at least for 6.0):
  - [Unsafe.OffsetOf to get field offsets in types](https://github.com/dotnet/runtime/issues/43285)
  - [Proposal: RuntimeHelpers.GetRawData](https://github.com/dotnet/runtime/issues/28001)
  - [Proposal for high-performance codegen-less Reflection factory APIs](https://github.com/dotnet/runtime/issues/23716)
  - [Proposal: Add new GetValue method to System.Reflection](https://github.com/dotnet/runtime/issues/19484)
  - [Faster MethodInfo.Invoke](https://github.com/dotnet/runtime/issues/7560)

davidfowl:
This isn't just for JSON, what we've we doing should work for any framework that uses reflection today I assume. I'm hoping to see a deep analysis here of the scenarios that Would use these APIs 

cc @pranavkm @SteveSandersonMS

iSazonov:
PowerShell could be benefit from this - it is based on reflection, emit and dynamic.

xoofx:
As mentioned in some of the previous issues, and to add a note on the requirements as there is a reference to Reflection.Emit that might make this not clear: the solution should not involve any managed objects from `System.Reflection` but go through only `RuntimeXXXHandle` that can work directly with `ldtoken`&co.

jkotas:
[...]

PR #93946, opened 2023-10-24
Add support for UnsafeAccessorKind.FieldOffset (#45152)

xoofx:
Draft PR to add support for `UnsafeAccessorKind.FieldOffset`.

This PR is for now just a proof of concept of the suggestion of implementing field offset computation via `UnsafeAccessorKind.FieldOffset` as suggested [here](https://github.com/dotnet/runtime/issues/45152#issuecomment-1777734470)

### DIFF
--- a/src/tests/baseservices/compilerservices/UnsafeAccessors/UnsafeAccessorsTests.cs
+++ b/src/tests/baseservices/compilerservices/UnsafeAccessors/UnsafeAccessorsTests.cs
@@ -18,6 +18,7 @@ static unsafe class UnsafeAccessorsTests
     {
         public const string StaticFieldName = nameof(_F);
         public const string FieldName = nameof(_f);
+        public const string FieldSecondName = nameof(_i);
         public const string StaticMethodName = nameof(_M);
         public const string MethodName = nameof(_m);
         public const string StaticMethodVoidName = nameof(_MVV);
@@ -30,6 +31,7 @@ static unsafe class UnsafeAccessorsTests
 
         private static string _F = PrivateStatic;
         private string _f;
+        private int _i;
 
         public string Value => _f;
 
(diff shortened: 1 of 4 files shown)


AaronRobinsonMSFT:
@xoofx Do you want to file an API proposal for this? That would be the next step.

src/tests/baseservices/compilerservices/UnsafeAccessors/UnsafeAccessorsTests.cs
```diff
@@ -215,6 +217,18 @@ public static void Verify_AccessFieldClass()
         extern static ref string GetPrivateField(UserDataClass d);
     }
 
+    [Fact]
+    public static void Verify_AccessFieldOffsetClass()
+    {
+        Console.WriteLine($"Running {nameof(Verify_AccessFieldOffsetClass)}");
+
+        // Offset = method table pointer + first string pointer
+        Assert.Equal(IntPtr.Size + IntPtr.Size, GetPrivateFieldOffset());
```
hamarb123, line 226: This should just be `IntPtr.Size` I think, since you cannot legitimately obtain a ref before the first field iirc.

hamarb123 (changes requested):

src/tests/baseservices/compilerservices/UnsafeAccessors/UnsafeAccessorsTests.cs
```diff
@@ -215,6 +217,18 @@ public static void Verify_AccessFieldClass()
...
+
+        [UnsafeAccessor(UnsafeAccessorKind.FieldOffset, Name=UserDataClass.FieldSecondName)]
+        extern static nint GetPrivateFieldOffset(UserDataClass? d = null);
```
hamarb123, line 229: Not sure what others' opinions on this are, but I like to be explicit with these sorts of APIs - I'd like it if we could also specify a second parameter which is the type of the field - it would first check for fields with same type (including modifiers), and then same type (ignoring modifiers) (same as other checks for the `UnsafeAccessor`s iirc).

PR #93949, opened 2023-10-24
Optimize cancellation handling in Hosting

pentp:
Avoid creating unnecessary `CancellationTokenSource` links and use `ConfigureAwaitOptions.SuppressThrowing` in a few places.

### DIFF
--- a/src/libraries/Microsoft.Extensions.Hosting/src/Internal/ApplicationLifetime.cs
+++ b/src/libraries/Microsoft.Extensions.Hosting/src/Internal/ApplicationLifetime.cs
@@ -63,7 +63,7 @@ namespace Microsoft.Extensions.Hosting.Internal
             {
                 try
                 {
-                    ExecuteHandlers(_stoppingSource);
+                    _stoppingSource.Cancel();
                 }
                 catch (Exception ex)
                 {
@@ -81,7 +81,7 @@ namespace Microsoft.Extensions.Hosting.Internal
         {
             try
             {
-                ExecuteHandlers(_startedSource);
+                _startedSource.Cancel();
             }
             catch (Exception ex)
             {
@@ -98,7 +98,7 @@ namespace Microsoft.Extensions.Hosting.Internal
         {
             try
             {
-                ExecuteHandlers(_stoppedSource);
+                _stoppedSource.Cancel();
             }
             catch (Exception ex)
             {
@@ -106,18 +106,6 @@ namespace Microsoft.Extensions.Hosting.Internal
                                          "An error occurred stopping the application",
                                          ex);
             }
-        }
-
-        private static void ExecuteHandlers(CancellationTokenSource cancel)
-        {
-            // Noop if this is already cancelled
-            if (cancel.IsCancellationRequested)
-            {
-                return;
-            }
-
-            // Run the cancellation token callbacks
-            cancel.Cancel(throwOnFirstException: false);
         }
     }
 }
--- a/src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs
+++ b/src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs
@@ -74,26 +74,16 @@ namespace Microsoft.Extensions.Hosting.Internal
         {
             _logger.Starting();
 
-            CancellationTokenSource? cts = null;
-            CancellationTokenSource linkedCts;
-            if (_options.StartupTimeout != Timeout.InfiniteTimeSpan)
+            using (var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, _applicationLifetime.ApplicationStopping))
             {
-                cts = new CancellationTokenSource(_options.StartupTimeout);
-                linkedCts = CancellationTokenSource.CreateLinkedTokenSource(cts.Token, cancellationToken, _applicationLifetime.ApplicationStopping);
-            }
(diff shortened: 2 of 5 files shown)


steveharter:
Was this PR created because of something you noticed and measured and\or does it have some background to it not captured here?

src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs
```diff
@@ -377,7 +360,14 @@ public async Task StopAsync(CancellationToken cancellationToken = default)
                     }
                     catch (Exception ex)
                     {
-                        exceptions.AddRange(groupedTasks.Exception?.InnerExceptions ?? new[] { ex }.AsEnumerable());
+                        if (groupedTasks.IsFaulted)
```
steveharter, line 363: Why was this changed?

src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs
```diff
@@ -363,7 +346,7 @@ public async Task StopAsync(CancellationToken cancellationToken = default)
                     {
                         // The task encountered an await; add it to a list to run concurrently.
                         tasks ??= new();
-                        tasks.Add(Task.Run(() => task, token));
+                        tasks.Add(task);
```
steveharter, line 349: Reference: https://github.com/dotnet/runtime/issues/93970

pentp: This `Task.Run` is basically a no-op - the `token` applies only to the running of the lambda, so there's theoretically a way for the cancellation to apply, but it's close to zero and doesn't do what the original author probably thought (it doesn't apply to the returned task).

src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs
pentp: To reduce allocations.

src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs
```diff
@@ -421,34 +411,30 @@ public async Task StopAsync(CancellationToken cancellationToken = default)
...
 
-            static async ValueTask DisposeAsync(object o)
+            static ValueTask DisposeAsync(object o)
             {
                 switch (o)
                 {
                     case IAsyncDisposable asyncDisposable:
-                        await asyncDisposable.DisposeAsync().ConfigureAwait(false);
-                        break;
+                        return asyncDisposable.DisposeAsync();
```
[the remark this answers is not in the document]
pentp: There's no semantic change here, `DisposeAsync` callers continue to `await` with `ConfigureAwait(false)`.

src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs
steveharter: Sure; I was thinking of the new LOC for checking `IsFaulted` - when will it be `false`?

src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs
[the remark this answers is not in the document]
steveharter: There is an unnecessary `await` + `ConfigureAwait(false)` for the `IAsyncDisposable` case but I think that may have been to prevent awaiting a default task for the `IDisposable` case?

src/libraries/Microsoft.Extensions.Hosting/src/Internal/ApplicationLifetime.cs
```diff
@@ -107,17 +107,5 @@ public void NotifyStopped()
                                          ex);
             }
         }
-
-        private static void ExecuteHandlers(CancellationTokenSource cancel)
-        {
-            // Noop if this is already cancelled
-            if (cancel.IsCancellationRequested)
```
steveharter, line 114: What is the reason to no longer include this "noop" case?

src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs
steveharter: It believe it is a no-op now. I believe it used to support running the sync portion of each task (before the first await) in parallel before.

src/libraries/Microsoft.Extensions.Hosting/src/Internal/Host.cs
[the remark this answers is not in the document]
pentp: `default` for `ValueTask` is the same as `Task.CompletedTask` for `Task`

src/libraries/Microsoft.Extensions.Hosting/src/Internal/ApplicationLifetime.cs
pentp: `CancellationTokenSource` has the exact same check already, so calling it multiple times has no effect.

steveharter: OK, I verified this is correct for both Core and .NET Framework.

steveharter (approved):
LGTM; thanks

PR #93953, opened 2023-10-25
Optimize Interlocked.CompareExchange use in Task

pentp:
Always use `Interlocked.CompareExchange` result when possible instead of re-reading the value.
Use `m_stateObject` for storing `WhenAllPromise` failed/canceled tasks.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs
@@ -3473,15 +3473,16 @@ namespace System.Threading.Tasks
             }
 
             // Not a single; it must be a list.
-            List<object?> continuations = (List<object?>)continuationObject;
+            List<object?> list = (List<object?>)continuationObject;
 
             //
             // Begin processing of continuation list
             //
 
             // Wait for any concurrent adds or removes to be retired
-            lock (continuations) { }
-            int continuationCount = continuations.Count;
+            Monitor.Enter(list);
+            Monitor.Exit(list);
+            Span<object?> continuations = CollectionsMarshal.AsSpan(list);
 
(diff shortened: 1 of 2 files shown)


src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs
```diff
@@ -4521,55 +4522,67 @@ private bool AddTaskContinuationComplex(object tc, bool addBeforeOthers)
             Debug.Assert(tc != null, "Expected non-null tc object in AddTaskContinuationComplex");
 
             object? oldValue = m_continuationObject;
+            if (oldValue == s_taskCompletionSentinel)
+                goto alreadyCompleted;
```
stephentoub, line 4526: goto doesn't seem warranted for these cases. Why not just return false?

src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs
```diff
@@ -4521,55 +4522,67 @@ private bool AddTaskContinuationComplex(object tc, bool addBeforeOthers)
             Debug.Assert(tc != null, "Expected non-null tc object in AddTaskContinuationComplex");
 
             object? oldValue = m_continuationObject;
+            if (oldValue == s_taskCompletionSentinel)
```
stephentoub, line 4525: As long as we're changing this method, let's add an assert that codifies the comment above "// THIS METHOD ASSUMES THAT m_continuationObject IS NOT NULL.".

stephentoub (approved):

stephentoub (approved):
Thanks!

tannergooding:
@pentp, just wanted to make sure you had seen Stephen's comments. Are you planning on fixing those here in this PR or would you prefer to push up a second PR after this one is merged since they're relatively minor?

PR #93957, opened 2023-10-25
The OptionsCache class is the only implementation of the IOptionsMonitorCache interface

786744873:
I see that OptionsCache is also used elsewhere in the source code. For code standardization and unification, it is recommended to use the interface instantiation mode.

image


src/libraries/Microsoft.Extensions.Options/src/OptionsFactory.cs
```diff
@@ -94,10 +94,9 @@ public TOptions Create(string name)
         /// <summary>
         /// Creates a new instance of options type.
         /// </summary>
-        /// <param name="name">The name of the <typeparamref name="TOptions"/> instance to create.</param>
         /// <returns>The created <typeparamref name="TOptions"/> instance.</returns>
         /// <exception cref="MissingMethodException">The <typeparamref name="TOptions"/> does not have a public parameterless constructor or <typeparamref name="TOptions"/> is <see langword="abstract"/>.</exception>
-        protected virtual TOptions CreateInstance(string name)
```
huoyaoyuan, line 100: As #92549, changing this is not acceptable.

src/libraries/Microsoft.Extensions.Options/src/StartupValidatorOptions.cs
```diff
@@ -9,6 +9,6 @@ namespace Microsoft.Extensions.Options
     internal sealed class StartupValidatorOptions
     {
         // Maps each pair of a) options type and b) options name to a method that forces its evaluation, e.g. IOptionsMonitor<TOptions>.Get(name)
-        public Dictionary<(Type, string), Action> _validators { get; } = new Dictionary<(Type, string), Action>();
+        public Dictionary<(Type, string), Action> _validators { get; } = new();
```
huoyaoyuan, line 12: This is a style-only change and should not be mixed with other changes. We generally do not accept style-only change case by case without corresponding analyzer rule.

src/libraries/Microsoft.Extensions.Options/src/OptionsManager.cs
```diff
@@ -16,7 +16,7 @@ public class OptionsManager<[DynamicallyAccessedMembers(Options.DynamicallyAcces
         where TOptions : class
     {
         private readonly IOptionsFactory<TOptions> _factory;
-        private readonly OptionsCache<TOptions> _cache = new OptionsCache<TOptions>(); // Note: this is a private cache
+        private readonly IOptionsMonitorCache<TOptions> _cache = new OptionsCache<TOptions>(); // Note: this is a private cache
```
huoyaoyuan, line 19: As the analyzer message says, the class is preferred to improve performance.

huoyaoyuan (commented):
> For code standardization and unification, it is recommended to use the interface instantiation mode.

The difference here is that the field is created by the class itself, not passed by user via public constructor.
We have the rule to always use most specified type for non-public API to maximize the performance.

src/libraries/Microsoft.Extensions.Options/src/OptionsFactory.cs
huoyaoyuan: This was brought from your existing `main` branch. It's not suggested to use `main` branch for contribution, open a new branch instead.

tarekgh:
Thank you, @786744873, for your pull request. We have reservations about accepting these changes due to the lack of clear benefits. Furthermore, @huoyaoyuan's comment is valid, highlighting a public surface change that is unacceptable and performance concerns. As a result, we regret to inform you that we are closing this pull request.

PR #93962, opened 2023-10-25
Fix double delims at the end of dirnames in tpa string when using corerun

sirntar:
### A short story
In `System.Runtime.Extensions.Tests` method `System.Tests.AppDomainTests.GetAssemblies` tested loading an extra assembly to *tpa*. In result new assembly should be loaded minimum twice. So corect *tpa* will be for example:
```bash
# [...]
/runtime/artifacts/tests/corefx/coreroot/System.Net.Http.dll
# [...]
/runtime/artifacts/tests/corefx/coreroot/System.Net.Http.dll
```
but instead we got:
```bash
# [...]
/runtime/artifacts/tests/corefx/coreroot//System.Net.Http.dll
# [...]
/runtime/artifacts/tests/corefx/coreroot/System.Net.Http.dll
```
And technically those are correct paths, but in `System.Tests.AppDomainTests.GetAssemblies` method they are compared as normal strings:
```C#
a.Location == someType.Assembly.Location
```

### Solution explanation
`core_root` and `core_library` are paths, that must be ended with `/`:
```c++
// fragment of corerun.cpp file
assert(dir.back() == pal::dir_delim); // pal::dir_delim = W("/")
string_t tmp = pal::build_file_list(dir, ext, [&](const char_t* file)
```
 but `pal::build_file_list` is also adding `/` after each dirname before It will include filename to stringstream. So skipping one of `/` will solve the problem. 

--------
cc @wscho77 @HJLeee @JongHeonChoi @t-mustafin @gbalykov @clamp03 @tomeksowi @yurai007

### DIFF
--- a/src/coreclr/hosts/corerun/corerun.hpp
+++ b/src/coreclr/hosts/corerun/corerun.hpp
@@ -552,9 +552,7 @@ namespace pal
             case DT_LNK:
             case DT_UNKNOWN:
                 {
-                    string_t full_filename;
-                    full_filename.append(directory);
-                    full_filename.append(1, pal::dir_delim);
+                    string_t full_filename{directory};
                     full_filename.append(entry->d_name);
                     if (!does_file_exist(full_filename.c_str()))
                         continue;
--- a/src/coreclr/hosts/corerun/corerun.hpp
+++ b/src/coreclr/hosts/corerun/corerun.hpp
@@ -572,7 +572,7 @@ namespace pal
             // Make sure if we have an assembly with multiple extensions present,
             // we insert only one version of it.
             if (should_add(entry->d_name))
-                file_list << directory << dir_delim << entry->d_name << env_path_delim;
+                file_list << directory << entry->d_name << env_path_delim;
         }
 
         closedir(dir);


gbalykov (approved):

src/coreclr/hosts/corerun/corerun.hpp
```diff
@@ -572,7 +572,7 @@ namespace pal
             // Make sure if we have an assembly with multiple extensions present,
             // we insert only one version of it.
             if (should_add(entry->d_name))
-                file_list << directory << dir_delim << entry->d_name << env_path_delim;
+                file_list << directory << entry->d_name << env_path_delim;
```
tomeksowi, line 575: What about DT_UNKNOWN case in line 557?

Sirntar: That doesn't matter, because DT_UNKNOWN and DT_LNK are only used to check whether file exist and path with double `//` is correct in Unix, plus it's not affecting anything outside `corerun`. 

```C++
case DT_LNK:
case DT_UNKNOWN:
    {
        string_t full_filename;
        full_filename.append(directory);
        full_filename.append(1, pal::dir_delim);
        full_filename.append(entry->d_name);
        if (!does_file_exist(full_filename.c_str()))
            continue;
    }
    break;
```

But yes, `full_filename.append(1, pal::dir_delim);` can also be removed, because it's not needed.

tomeksowi (approved):

src/coreclr/hosts/corerun/corerun.hpp
Sirntar: I've decided to remove it.

wscho77 (approved):

elinor-fung (approved):
The change looks good.

Quick question about how this was found in tests:
This repo runs libraries tests, like `System.Runtime.Extensions.Tests` in the description, using the product host (`dotnet`) rather than the test host for runtime tests (`corerun`). Is the bring-up for RISC-V using `corerun` for both runtime and libraries tests?

sirntar:
We haven't pushed those changes to the upstream, because  we thought you are not interested in implementing `dotnet` functionality inside `corerun`.

Issue #93912, opened 2023-10-24
Vector128 intrinsic causing crashes on ARM platform.


maciejaszyk:
### Description

We've noticed that our tests crash on the ARM platform (both Linux & MacOS) in release mode. After investigation, I identified one of our optimizations that use intrinsics is causing it. For testing purposes, I forced our code to use a scalar version and tests are running without any crashes.

 I also tried to reproduce it on Windows (x64) by disabling AVX/AVX2/removing Vector256 from the code below but it works just fine.  I suspect this can be the codegen issue.

Please notice the fact that doesn't appear in Debug mode.

Code that causing the issue:
https://github.com/ravendb/ravendb/blob/bb639435eb6b6148e6d2de03add1240a119f5828/src/Voron/Data/Lookups/Lookup.cs#L791-L800

### Reproduction Steps

Run `Tryouts` project in Release mode from a branch on ARM machine: https://github.com/maciejaszyk/ravendb/tree/ArmRepro

Should happen almost immediately.

### Expected behavior

No crashing

### Actual behavior

I've seen exit code 139.

### Configuration

.NET SDK:
 Version:   7.0.203
 Commit:    5b005c19f5

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  13.3
 OS Platform: Darwin
 RID:         osx.13-arm64
 Base Path:   /usr/local/share/dotnet/sdk/7.0.203/

Host:
  Version:      7.0.12
  Architecture: arm64
  Commit:       4a824ef37c

.NET SDKs installed:
  6.0.402 [/usr/local/share/dotnet/sdk]
  7.0.203 [/usr/local/share/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.10 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.5 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.12 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.10 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.5 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]
[...]

Issue #93047, opened 2023-10-05
dotnet 7 causes Segmentation fault on arm64 platforms


gnattu:
### Description

When I'm trying to test [Jellyfin](https://github.com/jellyfin/jellyfin) with recent dotnet 7 upgrade, I noticed that the library scan would cause segfault on all arm64 platforms I've been tested on(Linux and macOS). When doing a library scan with lots of video files, which I have 80 in my test folder and it reliably caused the segfault).

I've tested x86-64 on Windows and it runs fine, which led me to believe this is an arm64 specific issue.

I also noticed that if you try to attach a dotnet debugger to the process, the segfault would gone.

What I've tried to workaround this so far:
- Use `COMPlus_GCName=libclrgc.so`, still segfaults
- Use server gc, still segfaults

I also tried to get a coredump for the process, but it is too big to upload here. If that is needed please let me know.

### Reproduction Steps

I'm sorry I cannot come up with a simpler reproduction step because it is hard for me to extract a code pattern that would cause the segfault. So I will tell you how to reproduce it with jellyfin:

- Initial account creation may not be triggered for the master branch, so I would suggest you to install a release version of Jellyfin 10.8.11, start that first, then complete the account creation step in the web UI but don't add any library folders yet.

```shell
git clone https://github.com/jellyfin/jellyfin.git
cd jellyfin
dotnet publish --config Release
cd /Jellyfin.Server/bin/Release/net7.0
# if you are on Linux, you should mv jellyfin jellyfin-server
wget https://repo.jellyfin.org/releases/server/portable/versions/unstable/web/20231005.2/jellyfin-web_20231005.2_portable.tar.gz
tar xvf jellyfin-web_20231005.2_portable.tar.gz
mv jellyfin-web_20231005.2 jellyfin-web
# change the data-dir to the path pointed by the one created by the release version of Jellyfin 10.8.11
./jellyfin-server --datadir ~/jellyfin-data
```

After that, go to localhost:8096, log in with the account created by the release version, and add a media library that points to the folder containing a lot of video files. Mine are all h264 encoded in mp4 or flv containers. (You can find the option at the menu button at the top left corner. Click on Dashboard, then click on Libraries.)

You will see a segfault near the end of the Library scan.

### Expected behavior

The Library scan should success.

### Actual behavior

dotnet runtime segfaults.

### Configuration

`mcr.microsoft.com/dotnet/sdk:7.0-bookworm-slim` docker image on Linux arm64
[...]

PR #93982, opened 2023-10-25
Arm64: Do not contain LEA if offset or scale is not zero

kunalspathak:
Do not mark `LEA` as contained if the `offset < 0`.
Details: https://github.com/dotnet/runtime/issues/93912#issuecomment-1778544956

Fixes: https://github.com/dotnet/runtime/issues/93912
Fixes: https://github.com/dotnet/runtime/issues/93047


kunalspathak:
https://github.com/dotnet/runtime/pull/93983 will be backported to release/7.0

PR #93987, opened 2023-10-25
[mono] Simplify of OffsetToStringData as it's done on coreclr

thaystg:


vargaz:
Looks ok, but why is this useful ?

vargaz (approved):

src/mono/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.Mono.cs
```diff
@@ -32,8 +32,20 @@ public static void InitializeArray(Array array, RuntimeFieldHandle fldHandle)
...
-            get => OffsetToStringData;
+            [NonVersionable]
+            get =>
+                // Number of bytes from the address pointed to by a reference to
+                // a String to the first 16-bit character in the String.  Skip
+                // over the MethodTable pointer, & String
+                // length.  Of course, the String reference points to the memory
+                // after the sync block, so don't count that.
+                // This property allows C#'s fixed statement to work on Strings.
+                // On 64 bit platforms, this should be 12 (8+4) and on 32 bit 8 (4+4).
```
vargaz, line 43: Is this comment still correct  with 12/8 bytes ?

thaystg: fixed the comment, thanks!

src/mono/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.Mono.cs
```diff
@@ -32,8 +33,13 @@ public static void InitializeArray(Array array, RuntimeFieldHandle fldHandle)
         [Obsolete("OffsetToStringData has been deprecated. Use string.GetPinnableReference() instead.")]
         public static int OffsetToStringData
         {
-            [Intrinsic]
-            get => OffsetToStringData;
+            [NonVersionable]
```
marek-safar, line 36: This attribute has no use on Mono

thaystg: Thanks

src/mono/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.Mono.cs
```diff
@@ -32,8 +32,12 @@ public static void InitializeArray(Array array, RuntimeFieldHandle fldHandle)
         [Obsolete("OffsetToStringData has been deprecated. Use string.GetPinnableReference() instead.")]
         public static int OffsetToStringData
         {
-            [Intrinsic]
-            get => OffsetToStringData;
+            get =>
+#if TARGET_64BIT
+                20;
```
lambdageek, line 37: Can we define an internal constant in String.Mono.cs instead and then refer to it by name here. That way the value is closer to the rest of the string implementation.

lambdageek (approved):
Ok, but I would prefer if the magic number was in the string class

lambdageek (approved):

thaystg:
@lambdageek it looks like this is causing performance issues, should we revert?

lambdageek:
@thaystg It looks like the interpreter regressed. Maybe put back the interpreter intrinsic? I guess for some reason it's not inlining that getter method and the constant?

or maybe add `[MethodImpl(MethodImplOptions.AggressiveInlining)]` to it?

PR #93989, opened 2023-10-25
Test macOS runtime test crash dump collection

jkoritzinsky:


jkoritzinsky:
Looks like `createdump` is working again on macOS in our jobs:

> [createdump] Dump successfully written in 1191ms

PR #93991, opened 2023-10-25
Remove xunit.assert from S.C.I rd.xml

agocke:
Native AOT tests are failing because the rd.xml for preserving types includes references to things that don't exist anymore after we started using the XUnit AOT-compatible fork. Given that the fork is meant to be AOT-compatible, we shouldn't need the rd.xml anymore anyway.


jkotas (approved):

eiriktsarpalis:
@agocke for my education, could you point to the AOT-compatible fork? Does it support generic theory methods as well?

agocke:
@eiriktsarpalis It's only a fork of XUnit.Assert right now: https://github.com/dotnet/arcade/tree/main/src/Microsoft.DotNet.XUnitAssert

Sorry to get your hopes up.

This PR might not work, actually. There's more than just `xunit.assert` stuff in the XML. So I might have to add that back and just remove the xunit.assert lines.

agocke:
This looks like it solves the Assert problems. The timeouts are unrelated.

PR #93994, opened 2023-10-25
Implement PriorityQueue.Remove

### LINKED ISSUE #93925: [API Proposal]: Add a `Remove` method to `PriorityQueue`
### Background and motivation

When we shipped the `PriorityQueue` class back in .NET 6 we made the explicit decision to not add support for efficient, $\mathcal O(\log n)$ priority updates. This was due to priority updates not being supported without sacrificing overall performance and lack of widespread use in .NET codebases. We have a [separate issue](https://github.com/dotnet/runtime/issues/44871) that tracks adding a separate heap type that does support priority updates.

Even though the `PriorityQueue` type itself cannot be retrofitted to support efficient priority updates, it can be made to support $\mathcal O(n)$ priority updates. This can be useful in educational contexts or coding interviews, where the actual asymptotic performance is not mission critical and can be overlooked. Put differently, users have asked for this facility so that they can implement Dijkstra's algorithm in leetCode using C#. 

In the case of the Java, users can encode priority updates in the `PriorityQueue` class by piggybacking on its [built-in remove method](https://docs.oracle.com/javase/8/docs/api/java/util/PriorityQueue.html#remove-java.lang.Object-).

### API Proposal

```csharp

public partial class PriorityQueue<TElement, TPriority>
{
    // Scans the heap for elements equal to the provided value and removes the first occurrence.
    public bool Remove(TElement element, out TPriority priority, IEqualityComparer<TElement>? equalityComparer = null);
}
```

### API Usage

It's possible to use the above API to encode a $\mathcal O(n)$ priority update as follows:
```C#
public static void EnqueueOrUpdate<TElement, TPriority>(this PriorityQueue<TElement, TPriority> queue, TElement element, TPriority priority)
{
    queue.Remove(element, out _);
    queue.Enqueue(element, priority);
}
```

### Alternative Designs

_No response_

### Risks

It could end up being used in production code. I expect this risk would be mitigated since we're not explicitly adding an "update" API, instead users need to invoke a `Remove` API that will be explicitly documented as a linear-time scan operation.

eiriktsarpalis:
Fix #93925.

### DIFF
--- a/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
+++ b/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
@@ -546,8 +546,6 @@ namespace System.Collections.Generic
                 {
                     MoveDownCustomComparer(lastNode, index);
                 }
-
-                nodes[newSize] = default;
             }
 
             nodes[newSize] = default;
--- a/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
+++ b/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
@@ -503,7 +503,7 @@ namespace System.Collections.Generic
         }
 
         /// <summary>
-        /// Removes the first occurrence the equals the specified parameter.
+        /// Removes the first occurrence that equals the specified parameter.
         /// </summary>
         /// <param name="element">The element to look for.</param>
         /// <param name="removedElement">The actual element that got removed from the queue.</param>
--- a/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
+++ b/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
@@ -512,7 +512,7 @@ namespace System.Collections.Generic
         /// <returns><see langword="true"/> if matching entry was found and removed, <see langword="false"/> otherwise.</returns>
         /// <remarks>
         /// The method performs a linear-time scan of every element in the heap, removing the first value found to match the <paramref name="element"/> parameter.
-        /// In case of duplicate entries, what entry does get removed is non-deterministic, does not take priority into account and reflects the structure of the heap at the time of removal.
+        /// In case of duplicate entries, what entry does get removed is non-deterministic and does not take priority into account.
         /// </remarks>
         public bool Remove(
             TElement element,
--- a/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
+++ b/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
@@ -503,6 +503,58 @@ namespace System.Collections.Generic
         }
 
         /// <summary>
+        /// Removes the first occurrence the equals the specified parameter.
+        /// </summary>
+        /// <param name="element">The element to look for.</param>
+        /// <param name="removedElement">The actual element that got removed from the queue.</param>
+        /// <param name="priority">The priority value associated with the removed element.</param>
+        /// <param name="equalityComparer">The equality comparer governing element equality.</param>
+        /// <returns><see langword="true"/> if matching entry was found and removed, <see langword="false"/> otherwise.</returns>
+        /// <remarks>
+        /// The method performs a linear-time scan of every element in the heap, removing the first value found to match the <paramref name="element"/> parameter.
+        /// In case of duplicate entries, what entry does get removed is non-deterministic, does not take priority into account and reflects the structure of the heap at the time of removal.
+        /// </remarks>
+        public bool Remove(
+            TElement element,
+            [MaybeNullWhen(false)] out TElement removedElement,
+            [MaybeNullWhen(false)] out TPriority priority,
+            IEqualityComparer<TElement>? equalityComparer = null)
+        {
+            int index = FindIndex(element, equalityComparer);
+            if (index == -1)
+            {
+                removedElement = default;
+                priority = default;
+                return false;
+            }
+
+            (TElement Element, TPriority Priority)[] nodes = _nodes;
+            (removedElement, priority) = nodes[index];
+            int newSize = --_size;
+
+            if (index < newSize)
+            {
+                // We're removing an element from the middle of the heap.
+                // Pop the last element in the collection and sift downward from the removed index.
+                (TElement Element, TPriority Priority) lastNode = nodes[newSize];
+
+                if (_comparer == null)
+                {
+                    MoveDownDefaultComparer(lastNode, index);
+                }
+                else
+                {
+                    MoveDownCustomComparer(lastNode, index);
+                }
+
+                nodes[newSize] = default;
+            }
+
+            nodes[newSize] = default;
+            return true;
+        }
+
+        /// <summary>
         ///  Removes all items from the <see cref="PriorityQueue{TElement, TPriority}"/>.
         /// </summary>
         public void Clear()
@@ -807,6 +859,42 @@ namespace System.Collections.Generic
             }
 
             nodes[nodeIndex] = node;
+        }
+
+        /// <summary>
+        /// Scans the heap for the first index containing an element equal to the specified parameter.
+        /// </summary>
+        private int FindIndex(TElement element, IEqualityComparer<TElement>? equalityComparer)
+        {
+            equalityComparer ??= EqualityComparer<TElement>.Default;
+            (TElement Element, TPriority Priority)[] nodes = _nodes;
+            int size = _size;
+
+            // Currently the JIT doesn't optimize direct EqualityComparer<T>.Default.Equals
+            // calls for reference types, so we want to cache the comparer instance instead.
+            // TODO https://github.com/dotnet/runtime/issues/10050: Update if this changes in the future.
+            if (typeof(TElement).IsValueType && equalityComparer == EqualityComparer<TElement>.Default)
+            {
+                for (int i = 0; i < size; i++)
+                {
+                    if (EqualityComparer<TElement>.Default.Equals(element, nodes[i].Element))
+                    {
+                        return i;
+                    }
+                }
+            }
+            else
+            {
+                for (int i = 0; i < size; i++)
+                {
+                    if (equalityComparer.Equals(element, nodes[i].Element))
+                    {
+                        return i;
+                    }
+                }
+            }
+
+            return -1;
         }
 
         /// <summary>
--- a/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
+++ b/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
@@ -505,7 +505,7 @@ namespace System.Collections.Generic
         /// <summary>
         /// Removes the first occurrence that equals the specified parameter.
         /// </summary>
-        /// <param name="element">The element to look for.</param>
+        /// <param name="element">The element to try to remove.</param>
         /// <param name="removedElement">The actual element that got removed from the queue.</param>
         /// <param name="priority">The priority value associated with the removed element.</param>
         /// <param name="equalityComparer">The equality comparer governing element equality.</param>
--- a/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
+++ b/src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
@@ -513,6 +513,8 @@ namespace System.Collections.Generic
         /// <remarks>
         /// The method performs a linear-time scan of every element in the heap, removing the first value found to match the <paramref name="element"/> parameter.
         /// In case of duplicate entries, what entry does get removed is non-deterministic and does not take priority into account.
+        ///
+        /// If no <paramref name="equalityComparer"/> is specified, <see cref="EqualityComparer{TElement}.Default"/> will be used instead.
         /// </remarks>
         public bool Remove(
             TElement element,
@@ -549,6 +551,7 @@ namespace System.Collections.Generic
             }
 
             nodes[newSize] = default;
+            _version++;
             return true;
         }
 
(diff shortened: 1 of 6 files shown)


src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
```diff
@@ -502,6 +502,58 @@ public void EnqueueRange(IEnumerable<TElement> elements, TPriority priority)
...
+                if (_comparer == null)
+                {
+                    MoveDownDefaultComparer(lastNode, index);
+                }
+                else
+                {
+                    MoveDownCustomComparer(lastNode, index);
+                }
+
+                nodes[newSize] = default;
```
eiriktsarpalis, line 550:
```suggestion

                nodes[newSize] = default;
```

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
```diff
@@ -502,6 +502,56 @@ public void EnqueueRange(IEnumerable<TElement> elements, TPriority priority)
...
+        /// Removes the first occurrence the equals the specified parameter.
+        /// </summary>
+        /// <param name="element">The element to look for.</param>
+        /// <param name="removedElement">The actual element that got removed from the queue.</param>
+        /// <param name="priority">The priority value associated with the removed element.</param>
+        /// <param name="equalityComparer">The equality comparer governing element equality.</param>
+        /// <returns><see langword="true"/> if matching entry was found and removed, <see langword="false"/> otherwise.</returns>
+        /// <remarks>
+        /// The method performs a linear-time scan of every element in the heap, removing the first value found to match the <paramref name="element"/> parameter.
+        /// In case of duplicate entries, what entry does get removed is non-deterministic, does not take priority into account and reflects the structure of the heap at the time of removal.
```
danmoseley, line 515:
```suggestion
        /// In case of duplicate entries, what entry does get removed is non-deterministic and does not take priority into account.
```
Might mentioning heap structure imply that given the same sequence of construction, what is removed might actually be deterministic? We might change the implementation.

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
```diff
@@ -502,6 +502,56 @@ public void EnqueueRange(IEnumerable<TElement> elements, TPriority priority)
             }
         }
 
+        /// <summary>
+        /// Removes the first occurrence the equals the specified parameter.
```
tarekgh, line 506: >the [](http://example.com/codeflow?start=41&length=3)

that?

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
```diff
@@ -502,6 +502,56 @@ public void EnqueueRange(IEnumerable<TElement> elements, TPriority priority)
...
+        /// </summary>
+        /// <param name="element">The element to look for.</param>
```
tarekgh, line 508: >look for [](http://example.com/codeflow?start=49&length=8)

try to remove

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
```diff
@@ -502,6 +502,56 @@ public void EnqueueRange(IEnumerable<TElement> elements, TPriority priority)
...
+        /// <param name="removedElement">The actual element that got removed from the queue.</param>
+        /// <param name="priority">The priority value associated with the removed element.</param>
+        /// <param name="equalityComparer">The equality comparer governing element equality.</param>
+        /// <returns><see langword="true"/> if matching entry was found and removed, <see langword="false"/> otherwise.</returns>
+        /// <remarks>
+        /// The method performs a linear-time scan of every element in the heap, removing the first value found to match the <paramref name="element"/> parameter.
+        /// In case of duplicate entries, what entry does get removed is non-deterministic, does not take priority into account and reflects the structure of the heap at the time of removal.
+        /// </remarks>
```
tarekgh, line 516: ></remarks> [](http://example.com/codeflow?start=12&length=10)

Can mention the behavior when passing null equalityComparer in the remarks?

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
```diff
@@ -502,6 +502,56 @@ public void EnqueueRange(IEnumerable<TElement> elements, TPriority priority)
...
+                    MoveDownDefaultComparer(lastNode, index);
+                }
+                else
+                {
+                    MoveDownCustomComparer(lastNode, index);
+                }
+            }
+
+            nodes[newSize] = default;
+            return true;
```
tarekgh, line 552: >return true; [](http://example.com/codeflow?start=12&length=12)

should we change _version?

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
```diff
@@ -809,6 +859,42 @@ private void MoveDownCustomComparer((TElement Element, TPriority Priority) node,
...
+            int size = _size;
+
+            // Currently the JIT doesn't optimize direct EqualityComparer<T>.Default.Equals
+            // calls for reference types, so we want to cache the comparer instance instead.
+            // TODO https://github.com/dotnet/runtime/issues/10050: Update if this changes in the future.
+            if (typeof(TElement).IsValueType && equalityComparer == EqualityComparer<TElement>.Default)
+            {
+                for (int i = 0; i < size; i++)
+                {
+                    if (EqualityComparer<TElement>.Default.Equals(element, nodes[i].Element))
```
tarekgh, line 878: >EqualityComparer<TElement>.Default [](http://example.com/codeflow?start=24&length=34)

does using equalityComparer here will prevent the JIT to optimize it? even for value types?

src/libraries/System.Collections/tests/Generic/PriorityQueue/PriorityQueue.Generic.Tests.cs
```diff
@@ -246,6 +246,35 @@ public void PriorityQueue_DequeueEnqueue(int count)
...
+        {
+            bool result;
+            TElement removedElement;
+            TPriority removedPriority;
+
+            PriorityQueue<TElement, TPriority> queue = CreatePriorityQueue(count, count, out List<(TElement element, TPriority priority)> generatedItems);
+
+            for (int i = count - 1; i >= 0; i--)
+            {
+                (TElement element, TPriority priority) = generatedItems[i];
```
tarekgh, line 261: >(TElement element, TPriority priority) = generatedItems[i]; [](http://example.com/codeflow?start=16&length=59)

I think this will be better to make it remove from random index instead of always last one. i.e. use `Random.Next(i)` instead of `i`.

src/libraries/System.Collections/tests/Generic/PriorityQueue/PriorityQueue.Tests.cs
```diff
@@ -167,6 +167,41 @@ public void PriorityQueue_Generic_EnqueueRange_Null()
             Assert.Equal("not null", queue.Dequeue());
         }
 
+        [Fact]
+        public void PriorityQueue_Generic_Remove_MatchingElement()
```
tarekgh, line 171: will be good to add a test inserting same element more than once but with different priority then remove and validate the first match is always removed.

tarekgh (approved):
Added some questions and suggestions. LGTM otherwise.

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
eiriktsarpalis: Good point, I'll remove that bit.

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
eiriktsarpalis, line 506:
```suggestion
        /// Removes the first occurrence that equals the specified parameter.
```

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
```diff
@@ -502,6 +502,56 @@ public void EnqueueRange(IEnumerable<TElement> elements, TPriority priority)
             }
         }
 
+        /// <summary>
+        /// Removes the first occurrence that equals the specified parameter.
+        /// </summary>
+        /// <param name="element">The element to look for.</param>
```
eiriktsarpalis, line 508:
```suggestion
        /// <param name="element">The element to try to remove.</param>
```

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
eiriktsarpalis: The JIT will devirtualize `EqualityComparer<TElement>.Default.Equals` calls but only when `TElement` is a value type. For reference types it's better to reference the equality comparer from a local.

src/libraries/System.Collections/tests/Generic/PriorityQueue/PriorityQueue.Generic.Tests.cs
eiriktsarpalis: We generally want to avoid nondeterminism in unit tests. I could try using a `Random` instance with a seed, however the `CreatePriorityQueue` already uses pseudorandomness when populating the heap.

Another thing to note is that the index on `generatedItems` is not a predictor of where the element is located within the heap itself, since it is structured based on the priority values.

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
```diff
@@ -809,6 +859,42 @@ private void MoveDownCustomComparer((TElement Element, TPriority Priority) node,
...
         }
 
+        /// <summary>
+        /// Scans the heap for the first index containing an element equal to the specified parameter.
+        /// </summary>
+        private int FindIndex(TElement element, IEqualityComparer<TElement>? equalityComparer)
+        {
+            equalityComparer ??= EqualityComparer<TElement>.Default;
+            (TElement Element, TPriority Priority)[] nodes = _nodes;
+            int size = _size;
```
stephentoub, line 869: If we did:
```C#
ReadOnlySpan<(TElement Element, TPriority Priority)> nodes = _nodes.AsSpan(0, _size);
...
for (int i = 0; i < nodes.Length; i++)
{
    ... nodes[i]...
}
```
we'd avoid some bounds checking.

src/libraries/System.Collections/tests/Generic/PriorityQueue/PriorityQueue.Tests.cs
eiriktsarpalis: There's no guarantee of what element will get removed in case of duplicates, but you're right that there should some test exercising that particular scenario.

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
stephentoub, line 878: How important is the performance of Remove?  If it's really important, we could consider subsequently changing the data storage of PQ, to be two arrays (priorities and elements) rather than one that combines them. Then here this could use IndexOf and get the SIMD benefits where applicable. (Did we try such splitting in the original prototyping/implementation of PQ? In theory it could have benefits even for normal enqueue/dequeue operations, as the priorities we'd be sifting through would be more tightly packed.)

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
```diff
@@ -502,6 +502,56 @@ public void EnqueueRange(IEnumerable<TElement> elements, TPriority priority)
...
+        /// In case of duplicate entries, what entry does get removed is non-deterministic and does not take priority into account.
+        /// </remarks>
+        public bool Remove(
+            TElement element,
+            [MaybeNullWhen(false)] out TElement removedElement,
+            [MaybeNullWhen(false)] out TPriority priority,
+            IEqualityComparer<TElement>? equalityComparer = null)
+        {
+            int index = FindIndex(element, equalityComparer);
+            if (index == -1)
```
stephentoub, line 524:
```suggestion
            if (index < 0)
```

src/libraries/System.Collections/tests/Generic/PriorityQueue/PriorityQueue.Tests.cs
```diff
@@ -207,6 +242,16 @@ public void PriorityQueue_EmptyCollection_Peek_ShouldReturnFalse()
...
+        [Fact]
+        public void PriorityQueue_EmptyCollection_Remove_ShouldReturnFalse()
+        {
+            var queue = new PriorityQueue<string, string>();
+
+            Assert.False(queue.Remove(element: "element", out string removedElement, out string removedPriority));
+            Assert.Null(removedElement);
+            Assert.Null(removedPriority);
+        }
+
```
stephentoub, line 254: Since the primary reported purpose of adding this method is to enable non-production scenarios involving implementing Djikstra's algorithm, it'd be good to have a test that does so.

src/libraries/System.Collections/src/System/Collections/Generic/PriorityQueue.cs
eiriktsarpalis: The original prototype did use separate arrays, IIRC we ended up fusing into one to optimize the regular heap operations and enumerator. I don't think the performance of `Remove` is important enough to override the original concerns.

src/libraries/System.Collections/tests/Generic/PriorityQueue/PriorityQueue.Tests.cs
eiriktsarpalis: How does this look? https://github.com/dotnet/runtime/pull/93994/commits/5d2f0bfe44a2db15643033643677d466c87cd3b0

Issue #93246, opened 2023-10-09
JIT: use reverse post-order (RPO) traversal for morph


AndyAyersMS:
Morph, like many JIT phases, visits all the blocks in a method by following the `bbNext` chain. There's a missed opportunity here to cheaply propagate some information from block to block.

For "forward" phases like morph it is often preferable to visit the blocks in reverse post-order (RPO). An RPO ensures that for most blocks all the predecessors of the block have been visited before the block itself.

Currently value numbering implements an RPO visit. It's also possible to create an RPO using `fgDfsReversePostorder`.

The initial goal of this work is to modify morph to rely on RPO, and then to enable a simple form of global assertion prop for Morph (aka "cross-block local assertion prop") that can push facts forward across block boundaries. https://github.com/dotnet/runtime/pull/86822 has a prototype implementation. The main remaining challenges there are to make the RPO efficient and to properly handle cases where control flow is altered.

- [ ] (Q4 23) Implement RPO for Morph, and enable cross-block assertion prop
  - [ ] Revise local assertion prop to stop clearing the assertion table and track live assertions via bit vectors
  - [ ] Defer some of the control flow expansion (`fgCreateGCPoll`) from morph
  - [ ] Come up with a strategy to ensure that the remaining control flow changes in morph are "safe" in that they do not invalidate assumptions made earlier about the flow into a block
  - [ ] Implement the assertion merging algorithm
  - [ ] Revise morph to visit in RPO
  - [ ] Enable cross-block assertion prop
  - [ ] (stretch) avoid visiting unreachable blocks, and drop them from the RPO (perhaps delete the eagerly)
  - [ ] (stretch) refine RPO on the fly to avoid visiting blocks that become unreachable because of cross-block assertion prop
  - [ ] (stretch) do `QMARK` expansion before morph, remove bespoke handling in morph for `QMARK` assertions (see https://github.com/dotnet/runtime/pull/86778)
- [ ] (stretch) Implement an efficient, reusable RPO traversal; look for other phases that can benefit from RPO
- [ ] (stretch) Rework value numbering to use the reusable traversal

There will be extra TP cost from the RPO and from the assertion changes; the hope is that these are paid back by improved optimizations and that this entire change can be zero-cost.

Note for "backwards" phases post-order provides similar benefits (all successors likely visited before a block is visited).

cc @dotnet/jit-contrib

AndyAyersMS:
[...]

PR #93997, opened 2023-10-25
JIT: move return merging earlier

AndyAyersMS:
Instead of merging returns to the common return block in morph, do all the merging in `fgAddInternal` (where we already did some merging). This removes a case where morph would add a control flow edge in a way that might disrupt an ongoing RPO.

Earlier merging also opens up the possibility of tail merging some of the copies into the canonical return local, and possibly even some of the computations that feed the copies.

Modify the flow alterations done by morph. Previously if a tail call was expressed via a call to a `CORINFO_TAILCALL_HELPER`, morph would change the block kind to `BBJ_RETURN` and then merge the return, changing the block kind to `BBJ_ALWAYS`. Since merging now happens before moprh, morph needs to leave the block kind alone.

Generalize the post-tail-call sanity check in morph to recognize one new case that can come up.

Contributes to #93246.


AndyAyersMS:
cc @dotnet/jit-contrib 

A modest number of diffs, small net decrease in code size.

src/coreclr/jit/flowgraph.cpp
```diff
@@ -2367,8 +2367,13 @@ class MergedReturns
                 assert(searchLimit < maxReturns);
                 mergedReturnBlock = CreateReturnBB(searchLimit);
                 comp->genReturnBB = mergedReturnBlock;
-                // Downstream code expects the `genReturnBB` to always remain
-                // once created, so that it can redirect flow edges to it.
+
+                // In principle we should be able to remove the proctection
+                // we add here, and remove the genReturnBB if it becomes unreachable,
+                // since we now introduce all the flow to it during this phase.
```
SingleAccretion, line 2373: There's also some somewhat random code in lowering and codegen that references this block, so the deleted comment is still accurate.

AndyAyersMS: I think all those can be replaced by checks for the "right" BBJ_RETURN (the one and only without `BBF_HAS_JMP`), though I didn't want to push things quite that far.

AndyAyersMS:
I'm puzzed by some the CI runs... eg the libraries x64 linxu failure:
```
Assert failure(PID 71575 [0x00011797], Thread: 72229 [0x11a25]): Assertion failed 'retExpr->gtOper == GT_RETURN'
    File: /__w/1/s/src/coreclr/jit/morph.cpp Line: 15300
```
In my sources this assert is on line 15305.

BruceForstall:
> In my sources this assert is on line 15305.

Because locally you don't have https://github.com/dotnet/runtime/pull/93610?

AndyAyersMS:
> > In my sources this assert is on line 15305.
> 
> Because locally you don't have #93610?

Ah, could be. Thanks.

AndyAyersMS:
libraries jitstress failure: physical promotion is adding read-backs after a tail call
```
Replacing in BB04 [01B..028) -> BB06 (always), preds={BB03} succs={BB06}

STMT00003 ( 0x01B[E-] ... 0x027 )
               [000031] DAC-G------                         *  STORE_LCL_VAR struct<Microsoft.CodeAnalysis.CSharp.Symbols.TypeWithState, 16> V04 tmp2         
               [000019] --C-G------                         \--*  CALL      struct Microsoft.CodeAnalysis.CSharp.Symbols.TypeWithState:Create(Microsoft.CodeAnalysis.CSharp.Symbols.TypeSymbol,ubyte):Microsoft.CodeAnalysis.CSharp.Symbols.TypeWithState
               [000017] ----------- arg0                       +--*  LCL_VAR   ref    V05 tmp3          (last use)
               [000018] ----------- arg1                       \--*  CNS_INT   int    0
Processing block operation [000031] that involves replacements
Fields of [000031] in range [000..016) need to be read back: cannot decompose store
  V07 (V04.[000..008)) marked
  V08 (V04.[008..009)) marked
Reading back replacement V04.[000..008) -> V07 near the end of BB04:
STMT00008 ( ??? ... ??? )
               [000038] DA---------                         *  STORE_LCL_VAR ref    V07 tmp5         
               [000037] -----------                         \--*  LCL_FLD   ref    V04 tmp2         [+0]
Reading back replacement V04.[008..009) -> V08 near the end of BB04:
STMT00009 ( ??? ... ??? )
               [000040] DA---------                         *  STORE_LCL_VAR int    V08 tmp6         
               [000039] -----------                         \--*  LCL_FLD   ubyte  V04 tmp2         [+8]
```
and this makes the post-tail call sanity checker unhappy.
```
Assert failure(PID 10101 [0x00002775], Thread: 10115 [0x2783]): Assertion failed 'srcLclNum == callResultLclNumber' in 'Microsoft.CodeAnalysis.CSharp.NullableWalker:<VisitArgumentOutboundAssignmentsAndPostConditions>g__applyPostConditionsUnconditionally|242_3(Microsoft.CodeAnalysis.CSharp.Symbols.TypeWithState,int):Microsoft.CodeAnalysis.CSharp.Symbols.TypeWithState' during 'Morph - Global' (IL size 42; hash 0x2bb65d02; Tier1)

    File: /home/andy/repos/runtime0/src/coreclr/jit/morph.cpp Line: 15343
    Image: /home/andy/repos/runtime0/artifacts/bin/testhost/net9.0-linux-Release-x64/dotnet
```
One of the flaws of merging returns during `fgAddInternal` is that blocks with tail call candidates become attractive nuisances... they are not yet definite return points, so code after the call is potentially not dead, but if they become tail calls, we have to prove we don't lose anything by not executing that code.

We could avoid this by doing the tail call conversion earlier or the return merging later. The former seems difficult. The latter loses the potential benefits we might get from tail merges of the return value copies. So neither looks great but if necessary we can defer merging until just before morph.

Or we could try to avoid adding code to these spots...  In this case the read backs could be avoided since along this path to the return the struct fields already have the right values... but along other paths they may not, so the writebacks in the return block would have to be hoisted into a pred for non-tail-call paths:
```
STMT00005 ( ??? ... ??? )
               [000029] -A---------                         *  RETURN    struct
               [000048] -A---------                         \--*  COMMA     struct
               [000047] UA---------                            +--*  STORE_LCL_FLD ref    V04 tmp2         [+0]
               [000046] -----------                            |  \--*  LCL_VAR   ref    V07 tmp5         
               [000051] -A---------                            \--*  COMMA     struct
               [000050] UA---------                               +--*  STORE_LCL_FLD ubyte  V04 tmp2         [+8]
               [000049] -----------                               |  \--*  LCL_VAR   int    V08 tmp6         
               [000028] -------N---                               \--*  LCL_VAR   struct<Microsoft.CodeAnalysis.CSharp.Symbols.TypeWithState, 16> V04 tmp2          (last use)
```
not sure how feasible this would be. @jakobbotsch any thoughts?

Even if we fix this it might be an uphill battle to keep other phases in the JIT from trying to add code after the calls. Right now it does not seem to be widespread problem. If we went this route we could beef up the diagnostic checker since we know the problematic blocks are those that end with a tail call and are marked with `BBF_RETURN_MERGED`; it would need to allow the same sort of post-call code that `fgCheckStmtAfterTailCall` is ok with (incidentally that checker and `fgValidateIRForTailCall` seem like they have a lot of overlap, and perhaps the former can be removed?).

jakobbotsch:
How many tailcalls do we lose if we just discard the cases that we cannot prove to be safe? We have `fgValidateIRForTailCall` that should be a sound and complete checker before this change. I do not think that `fgCheckStmtAfterTailCall` is a sound checker today -- I actually didn't realize we had that. I would be fine with just discarding the cases we cannot prove to be safe with `fgValidateIRForTailCall`. Presumably this is going to be a super rare case after this PR, and seems like long-term it makes things more robust.

AndyAyersMS:
> How many tailcalls do we lose if we just discard the cases that we cannot prove to be safe?

Let me try this and see what the diff is.

AndyAyersMS:
Going to abandon this and just teach morph RPO to tolerate new edges being added to `genReturnBB`.

PR #94038, opened 2023-10-26
[iOS][non-icu] Implement missing Locale and Casing functions

mkhamoyan:
Implement below functions

- GetDefaultLocaleNameNative
- IsPredefinedLocaleNative

Run all tests under System.Globalization.Tests and adjust them for hybrid mode

cc @SamMonoRT


src/libraries/Common/tests/TestUtilities/System/PlatformDetection.cs
```diff
@@ -365,12 +365,16 @@ public static string GetDistroVersionString()
...
+        public static bool IsHybridGlobalization => m_isHybrid.Value;
         public static bool IsHybridGlobalizationOnBrowser => m_isHybrid.Value && IsBrowser;
         public static bool IsHybridGlobalizationOnOSX => m_isHybrid.Value && (IsOSX || IsMacCatalyst || IsiOS || IstvOS);
         public static bool IsNotHybridGlobalizationOnBrowser => !IsHybridGlobalizationOnBrowser;
         public static bool IsNotInvariantGlobalization => !IsInvariantGlobalization;
+        public static bool IsNotHybridGlobalization => !IsHybridGlobalization;
+        public static bool IsNotHybridGlobalizationOnOSX => !IsHybridGlobalizationOnOSX;
         public static bool IsIcuGlobalization => ICUVersion > new Version(0, 0, 0, 0);
         public static bool IsIcuGlobalizationAndNotHybridOnBrowser => IsIcuGlobalization && IsNotHybridGlobalizationOnBrowser;
+        public static bool IsIcuGlobalizationAndNotHybrid => IsIcuGlobalization && IsNotHybridGlobalizationOnBrowser && IsNotHybridGlobalizationOnOSX;
```
mdh1418, line 377:
```suggestion
        public static bool IsIcuGlobalizationAndNotHybrid => IsIcuGlobalization && IsNotHybridGlobalization;
```
Or does is this bool supposed to be true for `IsIcuGlobalization && IsHybridGlobalization && !(IsBrowser || IsOSX || IsMacCatalyst || IsiOS || IstvOS)`

src/libraries/System.Globalization/tests/CompareInfo/CompareInfoTestsBase.cs
```diff
@@ -45,5 +45,6 @@ public class CompareInfoTestsBase
                                                               s_invariantCompare.Compare("\u3060", "\uFF80\uFF9E", CompareOptions.IgnoreKanaType | CompareOptions.IgnoreWidth | CompareOptions.IgnoreCase) == 0;
 
         protected static bool IsNotWindowsKanaRegressedVersionAndNotHybridGlobalizationOnWasm() => !PlatformDetection.IsHybridGlobalizationOnBrowser && IsNotWindowsKanaRegressedVersion();
+        protected static bool IsNotWindowsKanaRegressedVersionAndNotHybridGlobalization() => IsNotWindowsKanaRegressedVersionAndNotHybridGlobalizationOnWasm() && !PlatformDetection.IsHybridGlobalizationOnOSX;
```
mdh1418, line 48:
```suggestion
        protected static bool IsNotWindowsKanaRegressedVersionAndNotHybridGlobalization() => IsNotWindowsKanaRegressedVersion() && PlatformDetection.IsNotHybridGlobalization();
```
Similar to the comment in PlatformDetection.cs

Since we have `IsNotHybridGlobalizationOnBrowser`, maybe we can update the `!PlatformDetection.IsHybridGlobalizationOnBrowser` to be `PlatformDetection.IsNotHybridGlobalizationOnBrowser`.

Also it would be nice to be consistent with the `Browser` vs `Wasm`. Would there be HybridGlobalization on Wasi?

src/libraries/System.Private.CoreLib/src/System/Globalization/CultureData.Icu.cs
```diff
@@ -287,6 +306,12 @@ private unsafe string IcuGetTimeFormatString(bool shortFormat)
         internal static bool IcuIsEnsurePredefinedLocaleName(string name)
         {
             Debug.Assert(!GlobalizationMode.UseNls);
+#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS
+            if (GlobalizationMode.Hybrid)
+            {
+                return Interop.Globalization.IsPredefinedLocaleNative(name);
+            }
```
mdh1418, line 313: We should maintain consistency
```
#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS
    if (Hybrid)
    {
        HybridImplementation
    }
    else
    {
        NotHybridImplementation
    }
#else
    NotHybridImplementation
#endif
```

Or we could make all of them (in a separate PR to address all instances and not bloat this PR)
```
#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS
    if (Hybrid)
    {
        HybridImplementation
    }
#endif
    NotHybridImplementation
```

src/libraries/System.Private.CoreLib/src/System/Globalization/CultureInfo.Unix.cs
```diff
@@ -13,7 +13,22 @@ internal static CultureInfo GetUserDefaultCulture()
                 return CultureInfo.InvariantCulture;
 
             CultureInfo cultureInfo;
-            if (CultureData.GetDefaultLocaleName(out string? localeName))
+            string? localeName;
+            bool result = false;
+#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS
+            if (GlobalizationMode.Hybrid)
```
mdh1418, line 19: Should we bump the hybrid logic to `GetDefaultLocaleName` in `CultureData.Icu.cs` to be more consistent with calling the `Interop.Globalization.HybridImplementation` `Interop.Globalization.NonHybridImplementation` in the same area.

src/native/libs/System.Globalization.Native/pal_locale.m
```diff
@@ -734,4 +734,49 @@ int32_t GlobalizationNative_GetLocalesNative(UChar* value, int32_t length)
...
+
+// GlobalizationNative_IsPredefinedLocaleNative returns TRUE if localeName exists in availableLocaleIdentifiers.
+// Otherwise it returns FALSE;
+
+int32_t GlobalizationNative_IsPredefinedLocaleNative(const char* localeName)
+{
+    @autoreleasepool
+    {
+        NSString *localeIdentifier = [NSString stringWithFormat:@"%s", localeName];
+        NSString *desiredLocaleIdentifier = [localeIdentifier stringByReplacingOccurrencesOfString:@"-" withString:@"_"];
```
mdh1418, line 771: Why do we replace all `-` with `_`? `NSLocale availableLocaleIdentifiers` returns a `Locale ID` which [may include `-` for script designators](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html#//apple_ref/doc/uid/10000171i-CH15-SW9) right?

mdh1418 (commented):
Are all of the tests that we are disabling for HybridGlobalizationOnOSX expected to not work?

src/native/libs/System.Globalization.Native/pal_casing.m
```diff
@@ -149,4 +149,25 @@ int32_t GlobalizationNative_ChangeCaseInvariantNative(const uint16_t* lpSrc, int
     }
 }
 
+void GlobalizationNative_InitOrdinalCasingPageNative(int32_t pageNumber, UChar* pTarget)
```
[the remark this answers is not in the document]
mkhamoyan: Now it does not get trimmed in hybrid mode. But once we will not load ICU at all `pal_casing.c` will also not be loaded.

src/libraries/Common/tests/TestUtilities/System/PlatformDetection.cs
mkhamoyan: `IsNotHybridGlobalization` is good. Now hybrid mode is only for Browser and iOS platforms.

src/libraries/System.Globalization/tests/CompareInfo/CompareInfoTestsBase.cs
mkhamoyan: In the future maybe there will be `HybridGlobalization` on `WASI`, but now all the implementation in `WASM` is done using `browser API` and there is no `browser API` in `WASI`.

src/native/libs/System.Globalization.Native/pal_locale.m
mdh1418: Would that mean that `[NSLocale availableLocaleIdentifiers]` doesnt contain `Local ID`s with `_` on your system? Could it be different based on the OS version or on other devices?

Which `desiredLocaleIdentifier` contained a `-` that we needed to cast to `_` to match the `availableLocales`?

Given that the docs specify there is a possibility of including `-`, because we only cast `desiredLocaleIdentifier` to only have `_`, it seems like there is a chance to miss a match.

src/libraries/System.Private.CoreLib/src/System/Globalization/CultureData.Icu.cs
mkhamoyan: First code block is used when it is not possible to return in `HybridImplementation`

#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS
    if (Hybrid)
    {
        HybridImplementation
    }
    else
    {
        NotHybridImplementation
    }
#else
    NotHybridImplementation
#endif

Second code block is used when we can return in `HybridImplementation` part, otherwise `NotHybridImplementation` code will overwrite values
```
#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS
    if (Hybrid)
    {
        HybridImplementation
    }
#endif
    NotHybridImplementation
```
For consistency we can use everywhere first way (second way will not be possible for all cases).

mkhamoyan:
> Are all of the tests that we are disabling for HybridGlobalizationOnOSX expected to not work?

Yes. These are cases which either were not possible (yet) to implement in hybrid mode, or behavioural changes after implementation

src/native/libs/System.Globalization.Native/pal_locale.m
mkhamoyan: Here are examples from `availableLocaleIdentifiers` in my system 
 avialableLocales 
I can see 949 Local IDs and none of them has `-`.
You are right this can be my system specific result. Will modify to not miss a match.

src/libraries/System.Private.CoreLib/src/System/Globalization/CultureData.Icu.cs
AustinWise: Another possibility is:


```
#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS
    if (Hybrid)
    {
        HybridImplementation
    }
    else
#endif
    {
        NotHybridImplementation
    }
```

This avoids duplicating the `NotHybridImplementation` logic.

src/native/libs/System.Globalization.Native/pal_casing.m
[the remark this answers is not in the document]
mdh1418: > But once we will not load ICU at all pal_casing.c will also not be loaded.

What would this look like? Would it get trimmed out or would it be a file exclusion?
Since this and `GlobalizationNative_InitOrdinalCasingPage` have the exact same logic`, it would be nice if we found a way to not duplicate this code.

src/libraries/System.Globalization/tests/CompareInfo/CompareInfoTestsBase.cs
mdh1418: Right, I would expect `IsNotWindowsKanaRegressedVersionAndNotHybridGlobalizationOnWasm` in the future to be `!PlatformDetection.IsHybridGlobalizationOnWasm && IsNotWindowsKanaRegressedVersion()` (what I meant by consistent)

Where maybe `PlatformDetection.IsHybridGlobalizationOnWasm = PlatformDetection.IsHybridGlobalizationOnBrowser || PlatformDetection.IsHybridGlobalizationOnWasi`.

So to that pattern `IsNotWindowsKanaRegressedVersionAndNotHybridGlobalizationOnWasm` -> `IsNotWindowsKanaRegressedVersionAndNotHybridGlobalizationOnBrowser` here would be more consistent but thats just a nit and can be addressed when `HybridGlobalization` is extended to `Wasi`.

mdh1418:
> These are cases which either were not possible (yet) to implement in hybrid mode, or behavioural changes after implementation

Can we document the behavior changes/impossible to implement, and if it falls under the "not yet possible", can we have an issue tracking that?

src/native/libs/System.Globalization.Native/pal_locale.m
mkhamoyan: Looks like that [apple doc](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LanguageandLocaleIDs/LanguageandLocaleIDs.html#//apple_ref/doc/uid/10000171i-CH15-SW9) and what I get in `availableLocaleIdentifiers` are not same. E.g.  `zh-Hant_HK` in `availableLocaleIdentifiers` is `zh_Hant_HK`, so if I change the implementation to have `localeIdentifier.ReplaceLastOccurenceOf('-', '_')` for the test case `zh-Hans-HK` it will return false but in reality it is in the `availableLocaleIdentifiers` but with the format `[language designator]_[script designator]_[region designator]` not like mentioned in the doc.

mkhamoyan:
> > These are cases which either were not possible (yet) to implement in hybrid mode, or behavioural changes after implementation
> 
> Can we document the behavior changes/impossible to implement, and if it falls under the "not yet possible", can we have an issue tracking that?

Behaviour changes and things that are not supported are documented here https://github.com/dotnet/runtime/blob/main/docs/design/features/globalization-hybrid-mode.md . 

Opened a ticket for the functionalities to investigate again if there is a possibility to implement https://github.com/dotnet/runtime/issues/94216

src/native/libs/System.Globalization.Native/pal_casing.m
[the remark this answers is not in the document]
mkhamoyan: From here https://github.com/dotnet/runtime/blob/f2e4ddf20936b37445bbbb2d139e8080ca00f16c/src/mono/mono/mini/CMakeLists.txt#L61 we will filter out for Apple platforms `pal_casing.c`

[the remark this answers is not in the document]
mkhamoyan: Our goal is to make hybrid globalization default mode for Apple platforms and in that case we will not need most of the  ICU related functions (we will not need most of the `pal_"name".c`). I guess at the end to not have duplicates I can move all these kind of functions in some common place like `pal_common.c`

src/native/libs/System.Globalization.Native/pal_locale.m
mkhamoyan: From the code I found out that Apple calls to ICU for that information check `CFLocaleCopyAvailableLocaleIdentifiers` here https://opensource.apple.com/source/CF/CF-1153.18/CFLocale.c.auto.html. `uloc_getAvailable` calls `_canonicalize` and in that function looks like all parts in `LocalID` are appended by `'_'` https://github.com/dotnet/icu/blob/dotnet/main/icu/icu4c/source/common/uloc.cpp#L1537

mdh1418 (approved):
After ensuring `GlobalizationNative_IsPredefinedLocaleNative` behavior is the same on all apple platforms for all `localeName`, looks good to me! Thanks!

src/native/libs/System.Globalization.Native/pal_casing.m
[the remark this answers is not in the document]
mkhamoyan: No this was first function. Moved it to pal_common.c .

mkhamoyan:
Failures are not related.

Issue #91538, opened 2023-09-04
[MT][browser] `System.Transactions.Tests.AsyncTransactionScopeTests` error: Thread failed to start


ilonatommy:
Failing after enabling all MT library tests (https://github.com/dotnet/runtime/pull/91536).
```
  [ERROR] FATAL UNHANDLED EXCEPTION: System.Threading.ThreadStartException: Thread failed to start.
    ---> System.ExecutionEngineException: mono_thread_platform_create_thread() failed
    --- End of inner exception stack trace ---
    at System.Threading.Thread.ThrowThreadStartException(Exception ex)
    at System.Threading.Thread.StartCore()
    at System.Threading.Thread.Start(Boolean captureContext, Boolean internalThread)
    at System.Threading.Thread.UnsafeStart()
    at System.Threading.WebWorkerEventLoop.StartExitable(Thread thread, Boolean captureContext)
    at System.Threading.PortableThreadPool.WorkerThread.CreateWorkerThread()
    at System.Threading.PortableThreadPool.WorkerThread.MaybeAddWorkingWorker(PortableThreadPool threadPoolInstance)
    at System.Threading.PortableThreadPool.AdjustMaxWorkersActive()
    at System.Threading.PortableThreadPool.NotifyWorkItemProgress(Object threadLocalCompletionCountObject, Int32 currentTimeMs)
    at System.Threading.PortableThreadPool.NotifyWorkItemComplete(Object threadLocalCompletionCountObject, Int32 currentTimeMs)
    at System.Threading.ThreadPoolWorkQueue.Dispatch()
    at System.Threading.PortableThreadPool.WorkerThread.WorkLoopSemaphoreSuccess(LowLevelLifoAsyncWaitSemaphore semaphore, Object stateObject)
    at System.Threading.LowLevelLifoAsyncWaitSemaphore.InternalAsyncWaitSuccess(LowLevelLifoAsyncWaitSemaphore self, WaitEntry internalWaitEntry)
    at System.Threading.LowLevelLifoAsyncWaitSemaphore.SuccessCallback(IntPtr lifoSemaphore, IntPtr userData)
```
Test got blocked.

pavelsavara:
Possibly we are leaking threads with external event loop (like thread pool threads)
We should make sure `emscripten_thread_cleanup` is called.

Also for `JSWebWorker` we need to make sure that the thread is not re-used.
See `PThread.returnWorkerToPool`

ilonatommy:
[...]

PR #94056, opened 2023-10-26
[browser][MT] unify pthread pool size to 32

pavelsavara:
- bump default thread pool size to 32 and unify to that default for all tests
 - see which existing active issues are caused by the same reason
 - keep ActiveIssue tracking it, because we know that it's too large default
     - https://github.com/dotnet/runtime/issues/91538


src/libraries/Microsoft.Extensions.HostFactoryResolver/tests/HostFactoryResolverTests.cs
```diff
@@ -279,7 +284,7 @@ public void ApplicationNameSetFromArgument()
 
         [ConditionalFact(typeof(PlatformDetection), nameof(PlatformDetection.IsThreadingSupported))]
         [DynamicDependency(DynamicallyAccessedMemberTypes.All, typeof(NoSpecialEntryPointPattern.Program))]
-        [ActiveIssue("https://github.com/dotnet/runtime/issues/91538", typeof(PlatformDetection), nameof(PlatformDetection.IsWasmThreadingSupported))]
+        [ActiveIssue("https://github.com/dotnet/runtime/issues/73420", typeof(PlatformDetection), nameof(PlatformDetection.IsNativeAot))]
```
ilonatommy, line 287: This is a closed issue, is it a typo or should we reopen it?

src/libraries/System.Private.CoreLib/src/System/Threading/PortableThreadPool.cs
```diff
@@ -17,7 +17,9 @@ internal sealed partial class PortableThreadPool
 
         private const short MaxPossibleThreadCount = short.MaxValue;
 
-#if TARGET_64BIT
+#if TARGET_BROWSER
+        private const short DefaultMaxWorkerThreadCount = 16;
```
ilonatommy, line 21: not 32?

ilonatommy (commented):
I checked and all removed issues are relevant

src/libraries/Microsoft.Extensions.HostFactoryResolver/tests/HostFactoryResolverTests.cs
pavelsavara: bad merge, sorry

src/libraries/System.Private.CoreLib/src/System/Threading/PortableThreadPool.cs
pavelsavara: there is 
- pthread pool (emscripten level)
- managed thread pool (portion of pthread pool )
- threads allocated by user code (will use the rest of the available pthreads, but even that could be not enough)

eng/pipelines/extra-platforms/runtime-extra-platforms-wasm.yml
```diff
@@ -112,14 +112,14 @@ jobs:
...
+      extraBuildArgs: /p:MonoWasmBuildVariant=multithread /p:WasmEnableThreads=true /p:AotHostArchitecture=x64 /p:AotHostOS=$(_hostedOS)
       isExtraPlatformsBuild: ${{ parameters.isExtraPlatformsBuild }}
       isWasmOnlyBuild: ${{ parameters.isWasmOnlyBuild }}
       # Always run for runtime-wasm because tests are not run in runtime
       alwaysRun: ${{ parameters.isWasmOnlyBuild }}
 
-       # NOTE - Since threading is experimental, we don't want to block mainline work
-      shouldContinueOnError: true
+      # NOTE - Since threading is experimental, we don't want to block mainline work
+      # TODO put back shouldContinueOnError: true
```
pavelsavara, line 122: we should put this back before merge

PR #94064, opened 2023-10-26
[wasm] Error out for AOT with no il trimming

radical:
Without trimming AOT builds would be very slow, and unusable.


vargaz (approved):

vargaz:
Some other cases:
- Debug+aot is not very usable either, aot doesn't support managed debugging.
- it might be possible to use profiled aot without il linking.
-

ilonatommy (approved):

PR #94066, opened 2023-10-26
[mono][debugger] Improve debugger features

thaystg:
- Creating new messages
- Fixing some messages or adding extra information.
- Fixing when jit_end is called while running a invoke_method


src/mono/mono/component/debugger-agent.c
```diff
@@ -7940,6 +7948,17 @@ domain_commands (int command, guint8 *p, guint8 *end, Buffer *buf)
...
+	case MDBGPROT_CMD_APPDOMAIN_GET_TYPE: {
+		MonoClass *klass;
+		domain = decode_domainid (p, &p, end, NULL, &err);
+		MonoTypeEnum type = decode_int (p, &p, end);
+		klass = decode_typeid (p, &p, end, NULL, &err);
+		int rank = decode_int (p, &p, end);
+		if (type == MONO_TYPE_SZARRAY || type == MONO_TYPE_ARRAY)
+			klass = mono_class_create_array (klass, rank);
+		buffer_add_typeid (buf, domain, klass);
+		break;
```
lambdageek, line 7960: This is a weird command.  It always sends a "type" enum, a base type and a rank, but if "type" is not array it just sends base the basetype (and ignores the rank) and otherwise it creates an array?  Does it really have to behave like this?

(Also will you need to support array bounds later?)

src/mono/mono/component/debugger-agent.c
```diff
@@ -8282,6 +8301,14 @@ field_commands (int command, guint8 *p, guint8 *end, Buffer *buf)
 		}
 		break;
 	}
+	case MDBGPROT_CMD_FIELD_GET_TOKEN_AND_TYPE: {
+		MonoClassField *f = decode_fieldid (p, &p, end, &domain, &err);
+		buffer_add_int (buf, mono_class_get_field_token (f));
```
lambdageek, line 8306: I think you need something like this for hot reload:

```c
			if (G_UNLIKELY (!f->type)) {
				ERROR_DECL(field_error);
				mono_field_resolve_type (f, field_error);
				mono_error_cleanup (field_error);
				if (!f->type)
					/* ERROR */;
			}
```

src/mono/mono/component/debugger-agent.c
```diff
@@ -9002,6 +8994,33 @@ type_commands_internal (int command, MonoClass *klass, MonoDomain *domain, guint
...
 		break;
 	}
+	case MDBGPROT_CMD_TYPE_GET_FIELD_RVA: 
+	{
+		gpointer iter = NULL;
+		int field_token = decode_int (p, &p, end);
+		while ((f = mono_class_get_fields_internal (klass, &iter))) {
+			if (mono_class_get_field_token (f) == field_token)
+			{
+				if (f->type->attrs & FIELD_ATTRIBUTE_HAS_FIELD_RVA)
```
lambdageek, line 9004: Also here:

```c
			if (G_UNLIKELY (!f->type)) {
				ERROR_DECL(field_error);
				mono_field_resolve_type (f, field_error);
				mono_error_cleanup (field_error);
				if (!f->type)
					continue;
			}
```

src/mono/mono/component/debugger-agent.c
```diff
@@ -9002,6 +8994,33 @@ type_commands_internal (int command, MonoClass *klass, MonoDomain *domain, guint
...
+					int swizzle = 1;
+					int align;
+				#if G_BYTE_ORDER != G_LITTLE_ENDIAN
+					swizzle = mono_type_size (type, &align);
+				#endif
+
+					int dummy;
+					int count = mono_type_size (f->type, &dummy)/mono_type_size (f->type, &align);
+					const char* arr = mono_field_get_rva (f, swizzle);
+					m_dbgprot_buffer_add_byte_array (buf, (uint8_t *)arr, count);
```
lambdageek, line 9015: Can we make a helper function in `object-internals.h`/`object.c` and share it with `ves_icall_System_Runtime_CompilerServices_RuntimeHelpers_GetSpanDataFrom`

lambdageek (approved):
Looks good. 

1. A couple of places that look at `MonoClassField:type` need hot reload protection.
2. It would be better if we didn't copy/paste the code from `ves_icall_System_Runtime_CompilerServices_RuntimeHelpers_GetSpanDataFrom`
3. The new `MDBGPROT_CMD_APPDOMAIN_GET_TYPE` command is kind of weird - I wonder if you can specialize it only for arrays (so it errors out if `type` is not `MONO_TYPE_SZARRAY||MONO_TYPE_ARRAY`).  Or is this supposed to emulate some CoreCLR behavior (in that case, please add a comment that it has to do it in this weird way)

src/mono/mono/component/debugger-agent.c
```diff
@@ -7940,6 +7950,42 @@ domain_commands (int command, guint8 *p, guint8 *end, Buffer *buf)
 		buffer_add_objid (buf, o);
 		break;
 	}
+	case MDBGPROT_CMD_APPDOMAIN_GET_TYPE: {
```
lambdageek, line 7953: it it only works on arrays? why is it called "get type" and not something like "create array type"

src/mono/mono/component/debugger-agent.c
```diff
@@ -8282,6 +8328,21 @@ field_commands (int command, guint8 *p, guint8 *end, Buffer *buf)
...
 		break;
 	}
+	case MDBGPROT_CMD_FIELD_GET_TOKEN_AND_TYPE: {
+		MonoClassField *f = decode_fieldid (p, &p, end, &domain, &err);
+		if (G_UNLIKELY (!f->type)) {
+			ERROR_DECL(field_error);
+			mono_field_resolve_type (f, field_error);
+			mono_error_cleanup (field_error);
+			if (!f->type)
+				/* ERROR */;
```
lambdageek, line 8338: This should actually error out - otherwise you'll dereference a null pointer on line 8341

```suggestion
				return ERR_INVALID_ARGUMENT;
```

"invalid argument" is not right.  do we have a  "bad object" or "bad class" error code?

lambdageek (approved):
lgtm, except one error case that should return an error code.
also I think "get type" is not a good name for a command that creates an array type (but it's ok if you want to leave it)

src/mono/mono/component/debugger-agent.c
thaystg: It works also for pointers, but I'm not sure how to implement it yet, that is why I didn't implement.

PR #94068, opened 2023-10-26
Added the LOHThreshold public key to the GC Config.

### LINKED ISSUE #93963: GC.GetConfigurationVariables() not showing 'System.GC.LOHThreshold'
### Description

I am investigating a performance degradation after upgrading from .NET 6 to .NET 7. I suspected it could be due to [.NET 7 breaking change: environment variable precedence](https://learn.microsoft.com/en-us/dotnet/core/compatibility/aspnet-core/7.0/environment-variable-precedence). That's why I want to log the GC settings used.

I haven't figured out the reason of the performance degradation, I am still investigating this. But I already wanted to share the problem I encountered with `GC.GetConfigurationVariables()`.

*EDIT: still don't understand the performance problems I observe with .NET 7. However the same code runs faster in .NET 8 RC2, even when comparing with .NET 6. Therefore I will just wait until .NET 8 is officially released. I still think `GC.GetConfigurationVariables()` needs improvement, to list all GC settings (e.g. `gcLOHThreshold`) even if users didn't override them.*

### Reproduction Steps

I use the following code to log all GC settings:
```csharp
IReadOnlyDictionary<string, object> gcConfig = GC.GetConfigurationVariables();
logger.LogInformation("{GC.GetConfigurationVariables}", JsonConvert.SerializeObject(gcConfig));
```

### Expected behavior

I expect it also logs `gcLOHThreshold`, especially because I have set the following two environment variables:

```yaml
env:
  - name: DOTNET_GCHeapHardLimitPercent
    value: '0x5F'
  - name: DOTNET_GCLOHThreshold
    value: '0x7A120'
```

### Actual behavior

I get the following log with .NET 8 RC2:
```json
{
    "serverGC": true,
    "concurrentGC": true,
    "retainVM": false,
    "noAffinitize": false,
    "gcCpuGroup": false,
    "gcLargePages": false,
    "heapCount": 4,
    "maxHeapCount": 0,
    "gcHeapAffinitizeMask": 0,
    "gcHeapAffinitizeRanges": "",
    "gcHighMemPercent": 0,
    "gcHeapHardLimit": 8160437862,
    "gcHeapHardLimitPercent": 95,
    "gcHeapHardLimitSOH": 0,
    "gcHeapHardLimitLOH": 0,
    "gcHeapHardLimitPOH": 0,
    "gcHeapHardLimitSOHPercent": 0,
    "gcHeapHardLimitLOHPercent": 0,
    "gcHeapHardLimitPOHPercent": 0,
    "gcConserveMem": 0,
    "gcName": "",
    "gcDynamicAdaptationMode": 0
}
```

### Regression?

Not a regression.

In .NET 7 it doesn't correctly log `gcHeapHardLimitPercent`. I found #84198 and #84201 which fixed the problem for `gcHeapHardLimitPercent` in .NET 8 RC 2, but not for `gcLOHThresshold`.

### Known Workarounds

Not that I know of.

### Configuration

On Windows 10 Enterprise x64.

```powershell
> dotnet --version
8.0.100-rc.2.23502.2
```

### Other information

_No response_

mrsharm:
Fixes #93963. The GC Configuration API only displays the values for those configurations with public keys defined i.e., an assigned property for the runtime configuration json file. This PR adds the LOHThreshold public key that is mentioned in the documentation: https://learn.microsoft.com/en-us/dotnet/core/runtime-config/garbage-collector#large-object-heap-threshold  but presumably forgotten to be added to the gcconfig.

### Program

```csharp
    {
        var results = GC.GetConfigurationVariables();
        foreach (var result in results)
        {
            Console.WriteLine("{ " + $"{result.Key} : {result.Value?.ToString()}" + " }");
        }
    }
```

### Configuration

```json
{
  "runtimeOptions": {
    "tfm": "net7.0",
    "includedFrameworks": [
      {
        "name": "Microsoft.NETCore.App",
        "version": "7.0.12"
      }
    ],
    "configProperties": {
      "System.Reflection.Metadata.MetadataUpdater.IsSupported": false,
      __"System.GC.LOHThreshold": 850000__
    }
  }
}
```

### Confirmed Result

```
{ ServerGC : False }
{ ConcurrentGC : True }
{ RetainVM : False }
{ NoAffinitize : False }
{ GCCpuGroup : False }
{ GCLargePages : False }
**{ LOHThreshold : 850000 }**
{ HeapCount : 1 }
{ MaxHeapCount : 0 }
{ GCHeapAffinitizeMask : 0 }
{ GCHeapAffinitizeRanges :  }
{ GCHighMemPercent : 0 }
{ GCHeapHardLimit : 0 }
{ GCHeapHardLimitPercent : 0 }
{ GCHeapHardLimitSOH : 0 }
{ GCHeapHardLimitLOH : 0 }
{ GCHeapHardLimitPOH : 0 }
{ GCHeapHardLimitSOHPercent : 0 }
{ GCHeapHardLimitLOHPercent : 0 }
{ GCHeapHardLimitPOHPercent : 0 }
{ GCConserveMem : 0 }
{ GCName : clrgcexp.dll }
{ GCDynamicAdaptationMode : 0 }
```

Aso, tested with an environment variable > 85,000 KB. 

Minor clean up: changed the 0 in the SpinCount config registration to a NULL for consistency.

### DIFF
--- a/src/coreclr/gc/gcconfig.h
+++ b/src/coreclr/gc/gcconfig.h
@@ -138,7 +138,7 @@ public:
     INT_CONFIG   (GCConserveMem,             "GCConserveMemory",          "System.GC.ConserveMemory",          0,                  "Specifies how hard GC should try to conserve memory - values 0-9")                       \
     INT_CONFIG   (GCWriteBarrier,            "GCWriteBarrier",            NULL,                                0,                  "Specifies whether GC should use more precise but slower write barrier")                  \
     STRING_CONFIG(GCName,                    "GCName",                    "System.GC.Name",                                        "Specifies the path of the standalone GC implementation.")                                \
-    INT_CONFIG   (GCSpinCountUnit,           "GCSpinCountUnit",           NULL,                                0,               "Specifies the spin count unit used by the GC.")                                             \
+    INT_CONFIG   (GCSpinCountUnit,           "GCSpinCountUnit",           NULL,                                0,                  "Specifies the spin count unit used by the GC.")                                          \
     INT_CONFIG   (GCDynamicAdaptationMode,   "GCDynamicAdaptationMode",   "System.GC.DynamicAdaptationMode",   0,                  "Enable the GC to dynamically adapt to application sizes.")
 // This class is responsible for retreiving configuration information
 // for how the GC should operate.
--- a/src/coreclr/gc/gcconfig.h
+++ b/src/coreclr/gc/gcconfig.h
@@ -79,7 +79,7 @@ public:
     BOOL_CONFIG  (GCLargePages,              "GCLargePages",              "System.GC.LargePages",              false,              "Enables using Large Pages in the GC")                                                     \
     INT_CONFIG   (HeapVerifyLevel,           "HeapVerify",                NULL,                                HEAPVERIFY_NONE,    "When set verifies the integrity of the managed heap on entry and exit of each GC")       \
     INT_CONFIG   (LOHCompactionMode,         "GCLOHCompact",              NULL,                                0,                  "Specifies the LOH compaction mode")                                                      \
-    INT_CONFIG   (LOHThreshold,              "GCLOHThreshold",            NULL,                                LARGE_OBJECT_SIZE,  "Specifies the size that will make objects go on LOH")                                     \
+    INT_CONFIG   (LOHThreshold,              "GCLOHThreshold",            "System.GC.LOHThreshold",            LARGE_OBJECT_SIZE,  "Specifies the size that will make objects go on LOH")                                    \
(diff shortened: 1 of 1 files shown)


mangod9:
are there other such cases which need to be looked at?

mrsharm:
> are there other such cases which need to be looked at?

Just went through the configs in gcconfig.h along with https://learn.microsoft.com/en-us/dotnet/core/runtime-config/garbage-collector to confirm that we aren't missing any more configs that have public keys that are not registered in the file. We should be good here.

Maoni0 (approved):

mangod9 (commented):
Is the GCSpinCountUnit change unrelated to the fix?

mrsharm:
> Is the GCSpinCountUnit change unrelated to the fix?
just used this opportunity to clean up - unrelated but thought it's unobtrusive enough :)

mangod9 (approved):

PR #94072, opened 2023-10-27
Fix comment typo in object.h

ericmutta:
The following sentence:

> Object is the **respesentation** of an...

Shoud read:

> Object is the **representation** of an...


jkotas (approved):
Thanks

PR #94073, opened 2023-10-27
Fix a refleak in the IEnumerator marshalling tests.

jkoritzinsky:
This leak is what was making this test non-unloadable.


src/tests/Interop/PInvoke/IEnumerator/IEnumeratorNative.cpp
```diff
@@ -98,6 +98,13 @@ extern "C" DLL_EXPORT HRESULT STDMETHODCALLTYPE VerifyIntegerEnumeration(IDispat
         return hr;
     }
 
+    hr = VariantClear(&result);
```
AaronRobinsonMSFT, line 101: This should be immediately after the `QueryInterface()`. The reason being the lifetime of the `VARIANT` is transferred in the QI, so it can be released at that point, regardless of success or failure of that QI.

AaronRobinsonMSFT (approved):

PR #94082, opened 2023-10-27
Avoid handing out duplicate `CORINFO_MODULE_STRUCT_*` handles

### LINKED ISSUE #93843: Non-deterministic codegen in S_P_CoreLib_System_Reflection_Runtime_TypeInfos_RuntimeTypeInfo__GetPropertyImpl
In #93600 I'm adding a test to ensure the output of the native aot compiler is deterministic, but it found an issue in codegen of GetPropertyImpl:

```diff
diff --git a/baseline.txt b/compare.txt
index faad015..81a66b5 100644
--- a/baseline.txt
+++ b/compare.txt
@@ -79,7 +79,7 @@ S_P_CoreLib_System_Reflection_Runtime_TypeInfos_RuntimeTypeInfo__GetPropertyImpl
                     00
   00000000000ADE75: 4C 8B 4A 08        mov         r9,qword ptr [rdx+8]
   00000000000ADE79: 48 85 DB           test        rbx,rbx
-  00000000000ADE7C: 0F 84 85 02 00 00  je          00000000000AE107
+  00000000000ADE7C: 0F 84 78 02 00 00  je          00000000000AE0FA
   00000000000ADE82: 48 89 5C 24 20     mov         qword ptr [rsp+20h],rbx
   00000000000ADE87: 89 6C 24 28        mov         dword ptr [rsp+28h],ebp
   00000000000ADE8B: 33 D2              xor         edx,edx
@@ -215,7 +215,7 @@ S_P_CoreLib_System_Reflection_Runtime_TypeInfos_RuntimeTypeInfo__GetPropertyImpl
   00000000000AE047: 48 8B C3           mov         rax,rbx
   00000000000AE04A: E9 91 00 00 00     jmp         00000000000AE0E0
   00000000000AE04F: 4D 85 FF           test        r15,r15
-  00000000000AE052: 0F 84 BC 00 00 00  je          00000000000AE114
+  00000000000AE052: 0F 84 AF 00 00 00  je          00000000000AE107
   00000000000AE058: F7 C5 00 00 01 00  test        ebp,10000h
   00000000000AE05E: 74 3E              je          00000000000AE09E
   00000000000AE060: 48 8D 4C 24 60     lea         rcx,[rsp+60h]
@@ -271,23 +271,16 @@ S_P_CoreLib_System_Reflection_Runtime_TypeInfos_RuntimeTypeInfo__GetPropertyImpl
                     00
   00000000000AE101: E8 6A EC F7 FF     call        S_P_CoreLib_System_ArgumentNullException__Throw
   00000000000AE106: CC                 int         3
-  00000000000AE107: 48 8D 0D 00 00 00  lea         rcx,[__Str_name]
-                    00
-  00000000000AE10E: E8 5D EC F7 FF     call        S_P_CoreLib_System_ArgumentNullException__Throw
-  00000000000AE113: CC                 int         3
-  00000000000AE114: 48 8B CB           mov         rcx,rbx
-  00000000000AE117: E8 04 15 F9 FF     call        S_P_CoreLib_System_ThrowHelper__GetAmbiguousMatchException
-  00000000000AE11C: 48 8B C8           mov         rcx,rax
-  00000000000AE11F: E8 00 00 00 00     call        RhpThrowEx
-  00000000000AE124: CC                 int         3
-  00000000000AE125: 90                 nop
-  00000000000AE126: 90                 nop
-  00000000000AE127: 90                 nop
-  00000000000AE128: 90                 nop
-  00000000000AE129: 90                 nop
-  00000000000AE12A: 90                 nop
-  00000000000AE12B: 90                 nop
-  00000000000AE12C: 90                 nop
-  00000000000AE12D: 90                 nop
-  00000000000AE12E: 90                 nop
-  00000000000AE12F: 90                 nop
+  00000000000AE107: 48 8B CB           mov         rcx,rbx
+  00000000000AE10A: E8 11 15 F9 FF     call        S_P_CoreLib_System_ThrowHelper__GetAmbiguousMatchException
+  00000000000AE10F: 48 8B C8           mov         rcx,rax
+  00000000000AE112: E8 00 00 00 00     call        RhpThrowEx
+  00000000000AE117: CC                 int         3
+  00000000000AE118: 90                 nop
+  00000000000AE119: 90                 nop
+  00000000000AE11A: 90                 nop
+  00000000000AE11B: 90                 nop
+  00000000000AE11C: 90                 nop
+  00000000000AE11D: 90                 nop
+  00000000000AE11E: 90                 nop
+  00000000000AE11F: 90                 nop
```

In one run, two `ArgumentNullException.Throw("name")` calls got coalesced into a single throwing code. In the other run, there are two identical blocks to throw called from two separate places.

MichalStrehovsky:
RyuJIT depends on never seeing two different `CORINFO_MODULE_STRUCT` for the same thing.

Fixes #93843.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
+++ b/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
@@ -707,9 +707,15 @@ namespace Internal.JitInterface
             _instantiationToJitVisibleInstantiation = null;
 
             _pgoResults.Clear();
+
+            // We need to clear out this cache because the next compilation could actually come up
+            // with a different MethodIL for the same MethodDesc. This happens when we need to replace
+            // a MethodIL with a throw helper.
+            _methodILScopeToHandle.Clear();
         }
 
         private Dictionary<object, IntPtr> _objectToHandle = new Dictionary<object, IntPtr>(new JitObjectComparer());
+        private Dictionary<MethodDesc, IntPtr> _methodILScopeToHandle = new Dictionary<MethodDesc, IntPtr>(new JitObjectComparer());
         private List<object> _handleToObject = new List<object>();
 
         private const int handleMultiplier = 8;
@@ -720,6 +726,13 @@ namespace Internal.JitInterface
 #endif
 
         private IntPtr ObjectToHandle(object obj)
+        {
+            // MethodILScopes need to go through ObjectToHandle(MethodILScope methodIL).
+            Debug.Assert(obj is not MethodILScope);
+            return ObjectToHandleUnchecked(obj);
+        }
+
+        private IntPtr ObjectToHandleUnchecked(object obj)
         {
             // SuperPMI relies on the handle returned from this function being stable for the lifetime of the crossgen2 process
             // If handle deletion is implemented, please update SuperPMI
@@ -752,9 +765,18 @@ namespace Internal.JitInterface
         private FieldDesc HandleToObject(CORINFO_FIELD_STRUCT_* field) => (FieldDesc)HandleToObject((void*)field);
         private CORINFO_FIELD_STRUCT_* ObjectToHandle(FieldDesc field) => (CORINFO_FIELD_STRUCT_*)ObjectToHandle((object)field);
         private MethodILScope HandleToObject(CORINFO_MODULE_STRUCT_* module) => (MethodIL)HandleToObject((void*)module);
-        private CORINFO_MODULE_STRUCT_* ObjectToHandle(MethodILScope methodIL) => (CORINFO_MODULE_STRUCT_*)ObjectToHandle((object)methodIL);
         private MethodSignature HandleToObject(MethodSignatureInfo* method) => (MethodSignature)HandleToObject((void*)method);
         private MethodSignatureInfo* ObjectToHandle(MethodSignature method) => (MethodSignatureInfo*)ObjectToHandle((object)method);
+
+        private CORINFO_MODULE_STRUCT_* ObjectToHandle(MethodILScope methodIL)
+        {
+            // RyuJIT requires CORINFO_MODULE_STRUCT to be unique. MethodILScope might not be unique
+            // due to ILProvider cache purging. See https://github.com/dotnet/runtime/issues/93843.
+            MethodDesc owningMethod = methodIL.OwningMethod;
+            if (!_methodILScopeToHandle.TryGetValue(owningMethod, out IntPtr handle))
+                _methodILScopeToHandle[owningMethod] = handle = ObjectToHandleUnchecked((object)methodIL);
+            return (CORINFO_MODULE_STRUCT_*)handle;
+        }
 
         private bool Get_CORINFO_METHOD_INFO(MethodDesc method, MethodIL methodIL, CORINFO_METHOD_INFO* methodInfo)
         {


src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
```diff
@@ -707,9 +707,15 @@ private void CompileMethodCleanup()
...
             _pgoResults.Clear();
+
+            // We need to clear out this cache because the next compilation could actually come up
+            // with a different MethodIL for the same MethodDesc. This happens when we need to replace
+            // a MethodIL with a throw helper.
+            _methodILScopeToHandle.Clear();
         }
 
         private Dictionary<object, IntPtr> _objectToHandle = new Dictionary<object, IntPtr>(new JitObjectComparer());
+        private Dictionary<MethodDesc, IntPtr> _methodILScopeToHandle = new Dictionary<MethodDesc, IntPtr>(new JitObjectComparer());
```
vitek-karas, line 718: Just for my education - this is a per-method cache, as in, we're going to clear it after each method compilation. And JIT doesn't maintain any state between methods either, right?

vitek-karas (approved):

src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
MichalStrehovsky: Yes, it's per method and there's not supposed to be extra state on the JIT side either.

jkotas:
> RyuJIT depends on never seeing two different CORINFO_MODULE_STRUCT for the same thing.

Where is this dependency exactly?

MichalStrehovsky:
> > RyuJIT depends on never seeing two different CORINFO_MODULE_STRUCT for the same thing.
> 
> Where is this dependency exactly?

Per https://github.com/dotnet/runtime/issues/93843#issuecomment-1775697925 it's in GenTreeStrCon equality. I think it's this:

https://github.com/dotnet/runtime/blob/2dc7c31b6b1ef1ebe5c3914aa4a03b8348a12873/src/coreclr/jit/gentree.cpp#L2704-L2710

jkotas:
```

**Comment by jkotas** [comment]
```
     if ((op1->AsStrCon()->gtSconCPX == op2->AsStrCon()->gtSconCPX) && 
         (op1->AsStrCon()->gtScpHnd == op2->AsStrCon()->gtScpHnd)) 
```
This condition is not a precise check to validate whether two string literals are going produce same string object. It is going to miss on string literals from different modules with JIT, and it is going to miss on string literals from different methods with AOT/R2R. It may be better to abstract this comparison via JIT/EE interface, so that it can be more precise.

In general, JIT depending on module handle equality seems to be an anti-pattern.

SingleAccretion:
> It may be better to abstract this comparison via JIT/EE interface, so that it can be more precise.

That was something I considered. It has the following drawbacks:

1) `GenTree::Equals` becomes more expensive. It feels questionable for "simple tree equality" to depend on (potentially expensive) Jit-EE calls.
2) VN-based optimizations such as CSE operate on the lowered form of `CNS_STR`, so it won't move the needle for them.

(As I said in the issue) I don't know what the "right" fix here is. This change seems like a practical option.

jkotas:
I agree that this change is a practical option. It will work ok as long as nothing depends on the comparison being precise.

MichalStrehovsky:
> I agree that this change is a practical option. It will work ok as long as nothing depends on the comparison being precise.

@dotnet/jit-contrib anyone else has an opinion about this fix? We're working around RyuJIT making assumptions about `CORINFO_MODULE_STRUCT*` handles. The workaround is not great and the RyuJIT assumptions have other problems enumerated by Jan above. Would a JIT-side fix make more sense?

If not, I'll merge this, but it is quite fragile. A more robust fix on the driver side would be to never purge the ILProvider cache and eat the memory overhead (essentially turn a cache into a memory leak). That one is obviously not great either.

jakobbotsch:
This fix looks good to me. I agree with @SingleAccretion about the concerns for `GenTree:: Compare`. This function is meant to be a fast syntactic equality check, it is not meant to be a precise or semantic equality check.

PR #94085, opened 2023-10-27
Format usings to match code style

akoeplinger:
This was done by running `dotnet format style --diagnostics IDE0055` against all the source projects in the repo to bring the usings into compliance with the code style rules: sorted alphabetically, `System.*` usings first


ViktorHofer:
Do we have validation for that during the build now as well?

src/coreclr/nativeaot/System.Private.CoreLib/src/System/EETypePtr.cs
```diff
@@ -10,17 +10,15 @@
...
-using System.Runtime;
 using System.Diagnostics;
-using System.Runtime.InteropServices;
+using System.Runtime;
 using System.Runtime.CompilerServices;
-
+using System.Runtime.InteropServices;
 using Internal.Runtime.CompilerServices;
-
+using CorElementType = System.Reflection.CorElementType;
```
MichalStrehovsky, line 18: This is deleting newlines between `using X` and `using Foo = X.Foo`. The repo style disallows separating these with newlines?

We also separate System and Internal namespace with newlines. Is the repo style disallowing that too?

akoeplinger: Right now it does, yeah. There is an option to separate using groups with a newline: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/dotnet-formatting-options#dotnet_separate_import_directive_groups

I can set that for the NativeAOT sources paths, we don't seem to do that in other parts of the repo.

akoeplinger: @MichalStrehovsky should look better now

ViktorHofer:
@akoeplinger I assume we can't just enable a rule that is validated by an existing analyzer / compiler? We need dotnet format to perform the validation?

akoeplinger:
Yeah I tried forcing it during the build https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview?tabs=net-7#enable-on-build but that doesn't seem to apply to the sort usings option...

ViktorHofer:
> but that doesn't seem to apply to the sort usings option...

Is that by design or a bug? Asking as honestly speaking, I'm struggling to see the value of this change without build protection enabled.

.editorconfig
```diff
@@ -154,6 +154,9 @@ csharp_space_between_square_brackets = false
 # License header
 file_header_template = Licensed to the .NET Foundation under one or more agreements.\nThe .NET Foundation licenses this file to you under the MIT license.
 
+[src/coreclr/nativeaot/**/*.cs]
+dotnet_separate_import_directive_groups = true
```
MichalStrehovsky, line 158: I think the repo preference is to put this in a local `.editorconfig`. So this would go to `src/coreclr/nativeao/.editorconfig`

MichalStrehovsky:
> > but that doesn't seem to apply to the sort usings option...
> 
> Is that by design or a bug? Asking as honestly speaking, I'm struggling to see the value of this change without build protection enabled.

Yeah - the thing with these PR's is that they disrupt ongoing PRs and repo history (although repo history for usings is not very interesting). If we don't enforce it, this cleanup is going to happen again down the line. This PR already conflicted with one of the PRs that merged in the past ~1 working day. Didn't check but I'm pretty sure it will conflict with #93440 that is a large PR. I think we should try to minimize disrupting other PRs. Cleanups like this should ideally be only done once.

.editorconfig
akoeplinger: done

akoeplinger:
Yep I agree that this large cleanup should be done infrequently. I'm happy to wait with this PR until large conflicting PRs are merged.

Regarding build protection, we essentially have three options:
1) Enforce it during build time, even locally. When this was discussed in the past there was pushback because it makes iterating painful if you need to constantly please the style analyzer or remember to pass an option to turn it off.
2) Enforce it on CI/PR. This has the downside of only noticing that something is wrong when you push to GitHub, though I've seen other communities where a bot is pushing formatting commits directly into your PR (I'm not a fan of that).
3) Periodically run formatting in a CI job and send an update PR e.g. weekly. This minimizes disruption and also makes sure that we don't accumulate a large diff like if we do it manually once every few years.

I prefer the last option, to me it strikes a good balance between keeping the repo clean and not annoying devs too much.

ViktorHofer:
cc @stephentoub as you drove many of the code ruleset changes

danmoseley:
Cc @jaredpar  as usual for the warning enforcement controversy 😀

jaredpar:
Recently put my thoughts on this [here](https://github.com/dotnet/runtime/issues/78414#issuecomment-1783408373). Happy to elaborate if there are questions.

danmoseley:
My suggestion is to close this PR as it's too disruptive and figure out the right way to do this with appropriate enforcement and maybe incrementally?

danmoseley:
Sometimes better for stuff like this to get consensus in an issue first?

jkotas:
This change LGTM once the build is green. It moves us closer to a state where we can consider enabling enforcement for this particular rule. There may be a few files where the violation may get reintroduced in the meantime and that's fine. 

> Enforce it on CI/PR. 

This option would be my personal preference. I like Roslyn model where the formatting is enforced by a specific leg only. We use the Roslyn model for enforcing formatting under src\clr\jit in this repo.

> Enforce it during build time, even locally. 

FWIW, I have `RunAnalyzers=false` set globally in my environment. I find the analyzers to be more disruptive than helpful most of the time. I would be nice to keep this global-disable option around.

jaredpar:
> Enforce it on CI/PR.

Looked at the .editorconfig in your repo and noticed that the [aot .editorconfig](https://github.com/dotnet/runtime/blob/main/src/coreclr/tools/aot/.editorconfig#L5) has IDE005 (remove unused usings) set to error. 

Wanted to caution that this particular rule on a repo of this complexity and frequency of committing can lead to some headaches. You can easily get into a situation where two PRs are independently green but when merged together turn red. Either because one PR removed a `using` the other depended on, another added a `using` that conflicted with the other, etc ... 
 Roslyn tried to enforce this rule for a time but eventually decided to stop as it caused too much friction.

akoeplinger:
One thing that makes enforcing locally difficult is that due to the way we build the System.Private.CoreLib project we include/exclude source files depending on the OS+arch so you'll need to run format across the whole matrix. For this PR I did this by setting `TargetOS` and `TargetArchitecture` env vars which make it into `dotnet format` because msbuild reads properties from the environment.

jkotas (approved):

PR #94089, opened 2023-10-27
Safib

KoltIP:
Changes related to type casting errors, as well as cosmetic fixes and memory leak fixes


lambdageek:
@KoltIP thanks for the contribution. It looks like your PR is trying to merge changes from an older branch of dotnet/runtime into the main branch.  You should instead checkout the `main` branch and make changes there.  Thanks!

PR #94094, opened 2023-10-27
Add Separate Mode to XUnitLogChecker Where Only Dumps are Analyzed

### LINKED ISSUE #93988: Allow XUnitLogChecker to Work Without Paths
As part of the general infrastructure improvements, the _XUnitLogChecker_ tool previously added to _coreclr_ as part of the *Merge-on-Red project*, is now needed on the libraries' tests side. However, it needs some overhauling due to the differences between the coreclr and libraries respective infrastructures. This issue is for tracking this work.

ivdiazsa:
In issue #93988, we described a new requirement for the XUnitLogChecker, which is to be able to only run the dumps checking stage without looking into test logs. This requirement came to be from the Merge-on-Green initiative to improve our testing infrastructure in general. While CoreCLR testing uses the whole log checker, libraries testing will only need to use the dumps analyzer. This PR overhauls as necessary to add this possibility.

Additionally, as part of this effort, this PR also adds actual command-line flags to the XUnitLogChecker, rather than relying on argument positioning. This will certainly make it more robust and easier to use in other places in the future.

### DIFF
--- a/src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs
+++ b/src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs
@@ -30,41 +30,116 @@ public class XUnitLogChecker
         public TagCategory Category { get; init; }
     }
 
+    private struct LogCheckerConfigParameters
+    {
+        public LogCheckerConfigParameters()
+        {
+            ResultsPath = string.Empty;
+            TestWrapperName = string.Empty;
+            DumpsPath = string.Empty;
+        }
+
+        public string ResultsPath { get; set; }
+        public string TestWrapperName { get; set; }
+        public string DumpsPath { get; set; }
+
+        public bool HasResultsPath() => !string.IsNullOrEmpty(ResultsPath);
+        public bool HasTestWrapper() => !string.IsNullOrEmpty(TestWrapperName);
+        public bool HasDumpsPath()   => !string.IsNullOrEmpty(DumpsPath);
+    }
+
     private enum TagCategory { OPENING, CLOSING }
 
     private const int SUCCESS = 0;
     private const int MISSING_ARGS = -1;
     private const int FAILURE = -2;
 
+    private static LogCheckerConfigParameters s_configuration;
+
     static int Main(string[] args)
     {
-        if (args.Length < 2)
+        s_configuration = new LogCheckerConfigParameters();
+
+        // We start with a 'FAILURE' exit code because if something goes wrong
+        // later on and the log checker fails, we want to know it.
+        int exitCode = FAILURE;
+
+        ParseCommandLineArgs(args);
+
+        // In CoreCLR tests, we record each test's results in an XML file, which
+        // is stored in the same directory as said test's script. So, the
+        // XUnitLogChecker looks there for said results log to do its job of
+        // fixing it if necessary. Afterwards, it checks if there are any dumps
+        // to read and display. This is the log checker's full run.
+        //
+        // In the Libraries tests however, the test logs are stored and created
+        // in a different way, currently outside the scope of the XUnitLogChecker,
+        // and therefore only the dump checking functionality is required in
+        // those cases, so we skip directly to that step.
+
+        if (s_configuration.HasResultsPath() && s_configuration.HasTestWrapper())
         {
-            WriteLineTimestamp("The path to the log file and"
-                               + " the name of the wrapper are required for an"
-                               + " accurate check and fixing.");
-            return MISSING_ARGS;
+            WriteLineTimestamp("The full run will be done.");
+            exitCode = DoFullRun();
+        }
+        else
+        {
+            WriteLineTimestamp("Only dumps checking will be done.");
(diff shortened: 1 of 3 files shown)


ivdiazsa:
😀 1

ivdiazsa:
+:100:

ivdiazsa:
:100:

src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs
```diff
@@ -30,41 +30,116 @@ public TagResult(string matchValue, TagCategory matchCategory)
...
+                    s_configuration.ResultsPath = nextValue;
+                    break;
+
+                case "--test-wrapper":
+                    s_configuration.TestWrapperName = nextValue;
+                    break;
+
+                case "--dumps-path":
+                    s_configuration.DumpsPath = nextValue;
+                    break;
```
carlossanlop, line 117: For the 3 cases, if `nextValue` is `string.Empty`, I think we should throw an exception indicating that "a value for <case name> was not provided". This is especially necessary since `DoFullRun`'s first few lines assume the TestWrapperName` and `ResultsPath` have a valid non-empty value.

src/tests/Common/helixpublishwitharcade.proj
```diff
@@ -412,9 +412,13 @@
       <XUnitLogCheckerHelixPath Condition="'$(TestWrapperTargetsWindows)' == 'true'">%25HELIX_CORRELATION_PAYLOAD%25/</XUnitLogCheckerHelixPath>
       <XUnitLogCheckerHelixPath>$(XUnitLogCheckerHelixPath)XUnitLogChecker/</XUnitLogCheckerHelixPath>
 
-      <XUnitLogCheckerArgs>$(_MergedWrapperRunScriptDirectoryRelative) $(_MergedWrapperName)</XUnitLogCheckerArgs>
+      <XUnitLogCheckerArgs>--results-path $(_MergedWrapperRunScriptDirectoryRelative)</XUnitLogCheckerArgs>
+      <XUnitLogCheckerArgs>$(XUnitLogCheckerArgs) --test-wrapper $(_MergedWrapperName)</XUnitLogCheckerArgs>
+      <XUnitLogCheckerArgs>$(XUnitLogCheckerArgs) --dumps-path</XUnitLogCheckerArgs>
```
carlossanlop, line 417: This is being passed without a value. This line could be removed, and we can save a loop when analyzing the arguments.

carlossanlop (commented):
Left a couple comments for you to consider.

src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs
ivdiazsa: We are already checking for the arguments to not be empty where needed so this is not necessary.

src/tests/Common/helixpublishwitharcade.proj
ivdiazsa: What is being passed without value? If you're talking about `--dumps-path`, the value is passed on the next lines. It is an environment variable so we have to distinguish between Linux and Windows.

carlossanlop: Ah I see it now. Thanks for pointing it out.

jkoritzinsky (approved):

src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs
carlossanlop: If a value isn't passed, the forloop is going to misbehave.

`XUnitLogChecker.exe --results-path --test-wrapper --dumps-path`

src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs
```diff
@@ -379,27 +459,46 @@ static TagResult[] GetOrderedTagMatches(Match[] openingTags, Match[] closingTags
...
+            // We know from the XUnitWrapperGenerator that the top element
+            // is the 'assembly' tag we're looking for.
+            var testRunDateTime = DateTime.ParseExact
+            (
+                fixedLogTree.Attribute("run-date-time").Value,
+                "yyyy-MM-dd HH:mm:ss",
+                System.Globalization.CultureInfo.InvariantCulture
+            );
+
+            dumpsFound = Directory
```
trylek, line 491: Should we report and / or delete the obsolete dumps or do we know that the OS or Helix infra deals with that somehow?

trylek (commented):
LGTM, thank you!

src/tests/Common/XUnitLogChecker/XUnitLogChecker.cs
ivdiazsa: This is not meant to be a tool used for general consumption. This is specifically for our test infrastructure with specific calls that should be made correctly in the first place (i.e. that case you describe should never happen)

Issue #93900, opened 2023-10-24
[release/8.0][wasm] WBT completely broken on 8.0 branch


radical:
Broken build - [20231023.109](https://dev.azure.com/dnceng-public/public/_build/results?buildId=448681&view=results), and [log](https://helixre107v0xd1eu3ibi6ka.blob.core.windows.net/dotnet-runtime-refs-heads-release-80-68eb4e1c553c43f3a4/Workloads-NoWebcil-Wasm.Build.NativeRebuild.Tests.FlagsChangeRebuildTests/1/console.80772c30.log?helixlogtype=result).

```
    Wasm.Build.NativeRebuild.Tests.FlagsChangeRebuildTests.ExtraEmccFlagsSetButNoRealChange [FAIL]
      System.TypeInitializationException : The type initializer for 'Wasm.Build.Tests.BuildTestBase' threw an exception.
      ---- System.Exception : Expected the multithreading runtime pack to be available when running on CI. IsRunningOnCI is true but IsWorkloadWithMultiThreadingForDefaultFramework is false.
```

[Last passing build](https://dev.azure.com/dnceng-public/public/_build/results?buildId=448014&view=results). Diff since then: e1da68dc...d8d69276

### Known Issue Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json
{
  "ErrorMessage": "Expected the multithreading runtime pack to be available when running on CI",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

### Report

|Build|Definition|Test|Pull Request|
|---|---|---|---|
|[468826](https://dev.azure.com/dnceng-public/public/_build/results?buildId=468826)|dotnet/runtime|[Wasm.Build.Tests.TestAppScenarios.LazyLoadingTests.LoadLazyAssemblyBeforeItIsNeeded](https://dev.azure.com/dnceng-public/public/_build/results?buildId=468826&view=ms.vss-test-web.build-test-results-tab&runId=10665564&resultId=100100)|dotnet/runtime#94516|
[...]

PR #94097, opened 2023-10-27
[release/8.0-staging][wasm] WBT Stop taking latest sdk in release/8.0

lewing:
It looks like wasm build tests have started picking up 8.0.2xx sdks after being unpinned.  This pins the WBT sdk to a recent 8.0.1xx sdk which should be new enough for testing and not contain the workload set changes.  We could probably set the channel correctly for 1xx and unpin but that can happen later.

We will need to fix wbt for the workload set changes but that definitely doesn't need to happen before 8.0.0.

Fixes https://github.com/dotnet/runtime/issues/93900

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -258,7 +258,7 @@
     <MicrosoftExtensionsLoggingVersion>3.1.7</MicrosoftExtensionsLoggingVersion>
     <MicrosoftSymbolStoreVersion>1.0.406601</MicrosoftSymbolStoreVersion>
     <!-- installer version, for testing workloads -->
-    <MicrosoftDotnetSdkInternalVersion>8.0.100-rtm.23478.7</MicrosoftDotnetSdkInternalVersion>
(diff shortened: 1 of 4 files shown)


radical (approved):

lewing:
@carlossanlop infra only, fixes the wasm failures in release/8.0*

carlossanlop:
Is this blocking release/8.0? If yes, then we should merge, but @mmitche needs to confirm.

Otherwise, can you please target 8.0-staging? Because as of right now, the branch is closed.

lewing:
It is just fixing tests, we can just target staging and ignore the failures if we need to take anything more into release/8.0

carlossanlop:
Okay thanks, please retarget: edit the PR title, switch the branch in the dropdown. Note that _sometimes_ git changes the branch in a bad way and ends up bringing in thinks we don't want, so please double check.

lewing:
I'll retarget once CI finishes then I'd suggest fast merging there

lewing:
This is now all the `release/8.0` to `release/8.0-staging` flow in #93874

lewing:
removed the other commits

lewing:
fast merging to unblock staging

PR #94100, opened 2023-10-27
Enable SA1136: Enum values should be on separate lines #94098

### LINKED ISSUE #94098: SA1136: Enum values should be on separate lines

artl93:
Fixes #94098, enabling another style rule.

Includes a couple places where the rule does not make sense, but otherwise, this includes all the code clean-up as well. 

Style updates were done by hand.

### DIFF
--- a/src/libraries/System.Diagnostics.Tracing/tests/BasicEventSourceTest/TestsWrite.cs
+++ b/src/libraries/System.Diagnostics.Tracing/tests/BasicEventSourceTest/TestsWrite.cs
@@ -13,7 +13,6 @@ using System.Diagnostics;
 
 namespace BasicEventSourceTests
 {
-#pragma warning disable SA1136 // Enum values should be on separate lines
     internal enum Color { Red, Blue, Green };
     internal enum ColorUInt32 : uint { Red, Blue, Green };
     internal enum ColorByte : byte { Red, Blue, Green };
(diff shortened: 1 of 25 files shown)


src/libraries/System.Diagnostics.Tracing/tests/BasicEventSourceTest/TestsWrite.cs
```diff
@@ -13,6 +13,7 @@
 
 namespace BasicEventSourceTests
 {
+#pragma warning disable SA1136 // Enum values should be on separate lines
```
danmoseley, line 16: Or you could enable the rule for product only not tests. We do that with some rules.

danmoseley (approved):
Source changes look good, I defer to others on whether it would be info or warning and whether should include tests

danmoseley:
Btw it's good to mention whether cleanup was done by hand or with a fixer, as the latter hopefully needs less careful eyeballing...

danmoseley (approved):

stephentoub (approved):
LGTM

danmoseley:
BTW, we often put a comma on the last enum entry, to reduce diffs when adding to the enum. I didn't suggest that because it didn't seem that important. But maybe if there was a fixer for this, it should do that.

Issue #93095, opened 2023-10-06
Arm64: Add SVE/SVE2 support in .NET 9


kunalspathak:
Until Armv8, NEON architecture enabled users to write vectorized code using SIMD instructions. The vector length for NEON instructions remains fixed at 128 bits. To take into account High Performance Computing applications, newer versions of Arm like v9 has developed Scalable Vector Extension (SVE). SVE is a new programming model that contains SIMD instructions operating on flexible vector length, ranging from 128 bits to 2048 bits. SVE2 extends SVE to enable domains like Computer vision, multimedia, etc. More details about SVE can be found on [Arm's website](https://developer.arm.com/documentation/102340/0100/Introducing-SVE2). In .NET 9, we want to start the foundational work in .NET libraries and RyuJIT to add support for SVE2 feature.

- [ ] **Design**: The instructions present in SVE2 needs to be exposed through .NET libraries to our customers. However, we want to make sure that the customers don't have to rewrite their code to consume SVE2 functionality and the experience of using SVE2 features should be seamless. Currently, .NET has [`Vector<T>`](https://learn.microsoft.com/en-us/dotnet/api/system.numerics.vector-1?view=net-7.0) that represents flexible vector length depending on the underlying hardware and the idea is to expose SVE2 "flexible vector length" functionality using `Vector<T>`.  We need to think about the common instructions and validate if they can be represented by APIs that just takes `Vector<T>` as parameter. Additionally, SVE2 introduces "predicate register" to mark vector lanes active/inactive. We do not want to expose this concept as well to our consumer through .NET libraries because of reason mentioned above. Hence, for every API, need to come up with a pseudo code of how the API should be implemented internally in JIT such that the "predicate register" concept is created and consumed inside JIT implicitly. There is a good discussion that happened in the past about the API proposal in https://github.com/dotnet/runtime/issues/88140.
- [ ] **Implement APIs in `System.Runtime.Intrinsic.Arm.SVE2`**:  Once design is finalized, need to add all the APIs in a new `SVE2` class under `System.Runtime.Intrinsic.Arm`. They need to be plumbed through the JIT (by transforming tree nodes to represent "predicate register" concept, if needed). all the way to generating code. https://github.com/dotnet/runtime/issues/99957
- [ ] If possible, automatically generate the test template for various APIs
- [ ] Update Antigen to pick up the new SVE APIs
- [ ] SVE hardware availability in CI
[...]

PR #94104, opened 2023-10-27
Arm64: Update certain encodings of instructions to make them size agnostic

kunalspathak:
While working on the tool to parse the [Arm64 encoding xmls](https://developer.arm.com/downloads/-/exploration-tools), I realized that there were some typos in the comments of our instruction table. Also, for certain instructions, I have updated the hex codes to be more generalized specially around size field. During encoding, we overwrite them anyhow and making them general would match with the encodings the tool parses from xml file.


kunalspathak:
@dotnet/jit-contrib @BruceForstall PTAL

BruceForstall (approved):

PR #94109, opened 2023-10-27
Move many interop tests to run in-proc and move many to built in the merged runner itself

jkoritzinsky:
Reference the native assets from the merged runner assembly itself so we can run more tests in-proc. Move many of the in-proc tests to build in-assembly (any test that isn't testing an assembly-wide flag is eligible for this treatment).

In the process of this. move more tests to use the XUnit attributes on the test methods directly, and fix some Xunit analyzer warnings in the GenericsTest test suite.

I've left the files that are in the merged runner in their existing places in the file tree and left the native builds as-is to limit diffs. In the future, we can follow-up this work by moving the sources around (and possibly combining the native libraries) to a way that we can better understand in this new model.


trylek (approved):
Awesome cleanup and improvement, thanks Jeremy!

AaronRobinsonMSFT (approved):

jkoritzinsky:
Failures are all #95712

AaronRobinsonMSFT (approved):

trylek (approved):
Looks great to me, thanks for pulling this off!

PR #94112, opened 2023-10-27
make Http2_RequestFailsWithAppropriateHttpProtocolException test more deterministic

### LINKED ISSUE #92647: [Test Failure] Http2_IncorrectServerPreface_RequestFailsWithAppropriateHttpProtocolException
## Build Information
Build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=417421
Build error leg or test failing: System.Net.Http.Functional.Tests.SocketsHttpHandlerTest_Http2.Http2_IncorrectServerPreface_RequestFailsWithAppropriateHttpProtocolException

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssueJsonStepByStep.md#how-to-create-a-known-issue-step-by-step).

```json
{
  "ErrorMessage": "System.Net.Http.Functional.Tests.SocketsHttpHandlerTest_Http2.Http2_IncorrectServerPreface_RequestFailsWithAppropriateHttpProtocolException [FAIL]",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=417421
**Error message validated:** `System.Net.Http.Functional.Tests.SocketsHttpHandlerTest_Http2.Http2_IncorrectServerPreface_RequestFailsWithAppropriateHttpProtocolException [FAIL]`
**Result validation: :white_check_mark:** Known issue matched with the provided build.
**Validation performed at:** 9/26/2023 3:12:06 PM UTC

### Report

|---|---|---|---|
|[447249](https://dev.azure.com/dnceng-public/public/_build/results?buildId=447249)|dotnet/runtime|[System.Net.Http.Functional.Tests.SocketsHttpHandlerTest_Http2.Http2_IncorrectServerPreface_RequestFailsWithAppropriateHttpProtocolException](https://dev.azure.com/dnceng-public/public/_build/results?buildId=447249&view=ms.vss-test-web.build-test-results-tab&runId=10011604&resultId=209349)|dotnet/runtime#86391|
#### Summary
|---|---|---|
|0|0|1|

wfurt:
When I look at the normal passing run I see something like

passingTest

when it fails, the exchange looks like 
failTest

So it seems like if the GOAWAY is sent before the client even start making request the test fails with  ` Couldn't find HttpProtocolException with matching error code in exception: System.Net.Http.HttpRequestException: An HTTP/2 connection could not be established because the server did not complete the HTTP/2 handshake.`

This may (or may not) be caused by the connection pool refactor. Since the requests are now detached, the connection can be closed by the 'GOAWAY` and the request does not fail with protocol error. 

To minimize risk, I only changed the test to wait for beginning of the request to arrive so when `HttpClient` process the response it would produce deterministic result. I was running tests in loop with this change and I did not see failure. It was failing for me before frequently with less than 20-50 iterations. 
 
fixes #92647

### DIFF
--- a/src/libraries/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.Http2.cs
+++ b/src/libraries/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.Http2.cs
@@ -139,7 +139,8 @@ namespace System.Net.Http.Functional.Tests
 
                 Http2LoopbackConnection connection = await server.AcceptConnectionAsync(timeout: null);
                 _ = await connection.ReadSettingsAsync();
-
+                // Wait until client starts sending request
+                Frame frame = await connection.ReadFrameAsync(TestHelper.PassingTestTimeout);
                 GoAwayFrame goAwayFrame = new GoAwayFrame(lastStreamId: 0, (int)ProtocolErrors.HTTP_1_1_REQUIRED, additionalDebugData: Array.Empty<byte>(), streamId: 0);
                 await connection.WriteFrameAsync(goAwayFrame);
 
@@ -319,6 +320,8 @@ namespace System.Net.Http.Functional.Tests
 
                 Http2LoopbackConnection connection = await server.AcceptConnectionAsync();
                 await connection.ReadSettingsAsync();
+                // Wait until client starts sending request
+                Frame frame = await connection.ReadFrameAsync(TestHelper.PassingTestTimeout);
                 await connection.SendGoAway(0, ProtocolErrors.INTERNAL_ERROR);
 
                 await AssertProtocolErrorAsync(sendTask, ProtocolErrors.INTERNAL_ERROR);
--- a/src/libraries/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.Http2.cs
+++ b/src/libraries/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.Http2.cs
@@ -140,7 +140,7 @@ namespace System.Net.Http.Functional.Tests
                 Http2LoopbackConnection connection = await server.AcceptConnectionAsync(timeout: null);
                 _ = await connection.ReadSettingsAsync();
                 // Wait until client starts sending request
-                Frame frame = await connection.ReadFrameAsync(TestHelper.PassingTestTimeout);
+                _ = await connection.ReadFrameAsync(TestHelper.PassingTestTimeout);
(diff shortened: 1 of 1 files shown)


src/libraries/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.Http2.cs
```diff
@@ -139,7 +139,8 @@ public async Task Http2_StreamResetByServerBeforePrefix_RequestFailsWithGoawayPr
 
                 Http2LoopbackConnection connection = await server.AcceptConnectionAsync(timeout: null);
                 _ = await connection.ReadSettingsAsync();
-
+                // Wait until client starts sending request
+                Frame frame = await connection.ReadFrameAsync(TestHelper.PassingTestTimeout);
```
antonfirsov, line 143: Nit: we don't use the frame returned.

```suggestion
                _ = await connection.ReadFrameAsync(TestHelper.PassingTestTimeout);
```

src/libraries/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.Http2.cs
```diff
@@ -319,6 +320,8 @@ public async Task Http2_IncorrectServerPreface_RequestFailsWithAppropriateHttpPr
 
                 Http2LoopbackConnection connection = await server.AcceptConnectionAsync();
                 await connection.ReadSettingsAsync();
+                // Wait until client starts sending request
+                Frame frame = await connection.ReadFrameAsync(TestHelper.PassingTestTimeout);
```
antonfirsov, line 324:
```suggestion
                _ = await connection.ReadFrameAsync(TestHelper.PassingTestTimeout);
```

antonfirsov (approved):
LGTM

src/libraries/System.Net.Http/tests/FunctionalTests/HttpClientHandlerTest.Http2.cs
wfurt: updated. I originally played with the frame type but I did not use it at the end.

wfurt:
enterprised-linux failures are unrelated. looks like infrastructure problem.

PR #94116, opened 2023-10-27
Emit/Call member references in new ILGenerator

buyaa-n:
Add implementation for  `Emit(OpCode opcode, ConstructorInfo con)`, `Emit(OpCode opcode, FieldInfo field)`, `Emit(OpCode opcode, MethodInfo meth)`, `Emit(OpCode opcode, Type cls)` and `EmitCall(OpCode opcode, MethodInfo methodInfo, Type[]? optionalParameterTypes)`


src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -210,7 +220,32 @@ public override void Emit(OpCode opcode, string str)
             _il.Token(tempVal);
         }
 
-        public override void Emit(OpCode opcode, ConstructorInfo con) => throw new NotImplementedException();
+        public override void Emit(OpCode opcode, ConstructorInfo con)
+        {
+            ArgumentNullException.ThrowIfNull(con);
+
+            if (!(opcode.Equals(OpCodes.Call) || opcode.Equals(OpCodes.Callvirt) || opcode.Equals(OpCodes.Newobj)))
```
buyaa-n, line 227: Runtime implementation Asserts this instead of throwing:
https://github.com/dotnet/runtime/blob/5f0fe49eb2920d3a9ed59e7459c1b2bea503fa90/src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeILGenerator.cs#L680-L694
I think this should throw instead, same as in the `EmitCall(OpCode opcode, MethodInfo methodInfo, Type[]? optionalParameterTypes)`:
https://github.com/dotnet/runtime/blob/5f0fe49eb2920d3a9ed59e7459c1b2bea503fa90/src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeILGenerator.cs#L603-L604

src/libraries/System.Reflection.Emit/src/Resources/Strings.resx
```diff
@@ -195,4 +195,10 @@
   <data name="Argument_MustBeSwitchOpCode" xml:space="preserve">
     <value>Only 'OpCode.Switch' can be used.</value>
   </data>
+  <data name="Argument_NotMethodCallOpcode" xml:space="preserve">
+    <value>The specified opcode cannot be passed to EmitCall.</value>
+  </data>
+  <data name="InvalidOperation_NotAVarArgCallingConvention" xml:space="preserve">
```
steveharter, line 201: Does not appear to be used.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -68,6 +71,13 @@ private void UpdateStackSize(OpCode opCode)
             _maxStackSize = Math.Max(_maxStackSize, _currentStack);
         }
 
+        private void UpdateStackSize(OpCode opCode, int stackchange)
```
steveharter, line 74: nit (if at all): "stackchange" -> "stackChange" (here and elsewhere)

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -210,7 +220,32 @@ public override void Emit(OpCode opcode, string str)
...
+            ArgumentNullException.ThrowIfNull(con);
+
+            if (!(opcode.Equals(OpCodes.Call) || opcode.Equals(OpCodes.Callvirt) || opcode.Equals(OpCodes.Newobj)))
+            {
+                throw new ArgumentException(SR.Argument_NotMethodCallOpcode, nameof(opcode));
+            }
+
+            int stackchange = 0;
+            // Push the return value if there is one.
+            stackchange++;
```
steveharter, line 234: Are we just assuming there is a return value then?

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -210,7 +220,32 @@ public override void Emit(OpCode opcode, string str)
...
+        {
+            ArgumentNullException.ThrowIfNull(con);
+
+            if (!(opcode.Equals(OpCodes.Call) || opcode.Equals(OpCodes.Callvirt) || opcode.Equals(OpCodes.Newobj)))
+            {
+                throw new ArgumentException(SR.Argument_NotMethodCallOpcode, nameof(opcode));
+            }
+
+            int stackchange = 0;
+            // Push the return value if there is one.
```
steveharter, line 233: Comment seems misplaced \ unnecessary since there is no code to check if there is a return value.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ModuleBuilderImpl.cs
```diff
@@ -180,30 +183,59 @@ internal void AppendMetadata(MethodBodyStreamEncoder methodBodyEncoder)
...
                     StandaloneSignatureHandle signature = il.Locals.Count == 0 ? default :
                         _metadataBuilder.AddStandaloneSignature(_metadataBuilder.GetOrAddBlob(MetadataSignatureHelper.LocalSignatureEncoder(il.Locals, this)));
                     offset = AddMethodBody(method, il, signature, methodBodyEncoder);
                 }
 
-                MethodDefinitionHandle methodHandle = AddMethodDefinition(method, method.GetMethodSignatureBlob(), offset, _nextParameterRowId);
-                WriteCustomAttributes(method._customAttributes, methodHandle);
-                _nextMethodDefRowId++;
+                MethodDefinitionHandle handle = AddMethodDefinition(method, method.GetMethodSignatureBlob(), offset, _nextParameterRowId);
+                Debug.Assert(method._handle == handle);
```
steveharter, line 229: Why was this Assert added? Can it be added to AddMethodDefinition() instead?

buyaa-n: > Why was this Assert added? 

To make sure the method handle populated with AddMethodDefinition is in sync with prepopulated one.

> Can it be added to AddMethodDefinition() instead?

It could, but `AddMethodDefinition` is a one liner that just calls `MetadataBuilder.AddMethodDefinition`, if you are OK with, don't really want to add another row.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
buyaa-n: Yes, as it is a constructor

steveharter (approved):

Issue #93132, opened 2023-10-06
ConfigurationBinder source generator emits SYSLIB1100 when it shouldn't apply


ericstj:
### Description

I have a public type with an internal constructor in the same assembly as my bind call. 

a. I wouldn't expect the source generator to need to construct the object at all since it's a Bind call.
b. The constructor is internal, which makes it visible to the source generator.

### Reproduction Steps

Repro project: [internalConstructor.zip](https://github.com/dotnet/runtime/files/12833309/internalConstructor.zip)

```c#
using Microsoft.Extensions.Configuration;

var c = new ConfigurationBuilder()
    .AddInMemoryCollection(new Dictionary<string,string?>()
    {
        ["Value"] = "Hello",
        ["Value2"] = "Hello world"
    })
    .Build();


MyType x = new();
c.Bind(x);
Console.WriteLine(x.Value);
Console.WriteLine(x.Value2);


public class MyType
{
    internal MyType(){ }
    public string? Value { get; set; }
    public string? Value2 { get; set; }
}
```

### Expected behavior

Compiles successfully without warning.

### Actual behavior

> C:\scratch\internalConstructor\Program.cs(13,1): warning SYSLIB1100: Cannot create instance of type 'MyType' because it
 is missing a public instance constructor. (https://learn.microsoft.com/dotnet/fundamentals/syslib-diagnostics/syslib11
00) [C:\scratch\internalConstructor\internalConstructor.csproj]

### Regression?

No

### Known Workarounds

Suppress the warning

ericstj:
I see a lot of folks hitting `SYSLIB1100` and `SYSLIB1101` for non-root level types - these behave similar to a root level type that's being passed to `Bind` -- the binder doesn't always *need* to create a type so the fact that it cannot may not be a worthwhile diagnostic to raise to the user.

This problem is exacerbated by lack of the ability to granularly suppress these warnings - or suppress at all in source due to https://github.com/dotnet/runtime/issues/92509.

I wonder if we should split these diagnostics into two - one set where it's clear that we can't do what the user wanted and another where it's more informational (and not a warning).  @tarekgh @eiriktsarpalis @eerhardt

ericstj:
What I was discussing above is even broader than the original issue.  The internal constructor problem is a plain-old-bug but one that will may be moot if we take the broader look at this behavior.

My point is that binding to types which are not constructable is actually a very common scenario and doesn't warrant raising a warning.

ericstj:
[...]

PR #94120, opened 2023-10-27
Exclude delegates from ConfigurationBinder source generator

### LINKED ISSUE #94117: Exclude properties of delegate types in configuration source generator
Presently, the configuration source generator generates binding code for properties employing `delegate`, `Func`, or `Action` types. However, there is no means to bind configuration to such types. Consequently, the generator generates an abundance of code for all types associated with these delegates, many of which will not be utilized and may lead to code that does not compile. Here's an example:

https://github.com/dotnet/extensions/blob/14943b99572078f1bcf2c51e9507cf2631133fae/src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/FailureEventMetricsOptions.cs#L11

This issue is tracking to fix the source generator to skip generating any code for such properties or types.

ericstj:
Fixes #94117

### DIFF
--- a/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
+++ b/src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
@@ -201,6 +201,10 @@ namespace Microsoft.Extensions.Configuration.Binder.SourceGeneration
                 {
                     spec = new ConfigurationSectionSpec(type);
                 }
+                else if (type.TypeKind is TypeKind.Delegate)
+                {
+                    spec = CreateUnsupportedTypeSpec(typeParseInfo, NotSupportedReason.DelegateNotSupported);
+                }
                 else if (type is INamedTypeSymbol)
                 {
                     spec = CreateObjectSpec(typeParseInfo);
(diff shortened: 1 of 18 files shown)


davidfowl:
How does JSON avoid this problem? Does it has any logic for stopping the scan?

ericstj:
Json allows folks to specify `JsonIgnore` on members you don't want it to mess with.  We don't have that *yet* but I discussed it here: https://github.com/dotnet/runtime/issues/93132
I'll double check if there's any default handling for wacky types in JSON.  Moving this to draft anyhow as I'm noticing I might have missed some places I need to block.

ericstj:
This didn't fix the case where the Func was in an array, collection, etc.  The generator was throwing an IOE there.  Need to take a closer look at this code and see if there's a natural place to make type exclusions or if we'll need to add one.

ericstj:
JSON has a default set of types, including delegates that it ignores: https://github.com/dotnet/runtime/blob/9461f4b5bd036005f37e78516c48b13396f76750/src/libraries/System.Text.Json/gen/JsonSourceGenerator.Parser.cs#L1687C26-L1687C43

davidfowl:
> JSON has a default set of types, including delegates that it ignores: https://github.com/dotnet/runtime/blob/9461f4b5bd036005f37e78516c48b13396f76750/src/libraries/System.Text.Json/gen/JsonSourceGenerator.Parser.cs#L1687C26-L1687C43

This is what we have to do, not explode.

PR #94121, opened 2023-10-27
Remove the JIT/CheckProjects tool

### LINKED ISSUE #76087: JIT/CheckProjects tool isn't used correctly in CI
I've noticed that the `src/tests/JIT/CheckProjects/CheckProjects.csproj` test doesn't actually work in CI, and hasn't even worked locally since repo consolidation in .NET 5.

This test is meant to validate that the project files in the `src/tests` tree meet some basic invariants established by the JIT team. Mainly that the `_d`, `_do`, `_r`, and `_ro` suffixed project files set the right `DebugType` and `Optimize` flags to be run in their specific configuration no matter the build configuration of the test tree.

The main problem with this tool is that we currently only run it on Helix, where we don't have a clone of the repo. Additionally, it hasn't been updated since repo consolidation and it went unnoticed since the tool silently passes when the file layout is not what it expects.

I think we should do one of the following actions to solve these problems:

1. Delete the test. If we don't feel that we need or want this validation, then we don't need to have the test, and it's just adding a test that doesn't give us anything in our test tree.
2. Move the test to `src/tests/Common` and add a job in PRs like the formatting jobs that only triggers on changes to the test tree to validate the invariants the tool expects.
3. Move the test to https://github.com/dotnet/jitutils and consume it in a PR job (like step 2, but the JIT team would own the tool in their own repo, like the `jitformat` tool).

cc: @dotnet/jit-contrib

jkoritzinsky:
Fixes #76087


AndyAyersMS (approved):
It was useful, once, long ago.

PR #94122, opened 2023-10-27
Merge CodeGenBringUpTests projects

jkoritzinsky:
Change CodeGenBringUpTests to use merged test runners that include all of the sources directly. This removes ~640 csproj files from our test build (about 7-8% of all csproj files in src/tests)


markples:
I'd love to see what the benefits are (overall build, the "do-nothing" incremental build that has to look at all the csproj, whatever else).

This is an interesting place to start - easy of course given the structure of the tests (no dedup necessary, for example) - but some of the merged group difficulties are the most relevant here for retargeting.  In https://github.com/dotnet/runtime/blob/main/docs/design/coreclr/jit/porting-ryujit.md, we give instructions to use BuildAsStandalone in order to execute each test (meaning every test, not just one for debugging) separately.  How does one do that now?

Similarly, the "crashes take out all tests" issue may have extra relevance here since local runs could be tracking progress against all tests with some of them crashing.

jkoritzinsky:
I'll try to pull data about benefits from this change.

The primary goal of this work would be to reduce the number of csproj files so we can get rid of the nested MSBuild invocations in the future. This is the third largest subfolder in terms of number of csproj files (all folders under src/tests and folders under src/tests/JIT).

During a port, I'd recommend disabling the tests that don't work with an ActiveIssue when we port to a new architecture and enable tests as they are fixed (same policy as today, but with ActiveIssue instead of issues.targets). For local testing, there's also a `TestFilter` property that can be used to filter tests at build time that can be used with BuildAsStandalone to get back to the old style of tests.

I agree that the crash issue will become more relevant as we go this route where we can't break-out tests into separate processes easily.

markples:
What is your recommendation for review strategy?  I feel good about the _ro, etc., settings being consolidated in 4 files.  It would  be easy for there to previously be typos that this removes.  I'm less sure about any other details.  For example, I did notice that ArrayExc was not marked priority==1 but is now OuterLoop.  I suspect that is actually better, but I'm not sure how to check such things.

Do you have before/after test counts?  ArrayExc is interesting again in that it only had 2 of the configurations before (oversight?), so that will actually be a +2.

jkoritzinsky:
I don't have the test counts. @trylek can you send me the tooling you use to validate the test counts? I likely won't get to it before I'm OOF, but I'll take a look when I get back and validate the information.

markples:
> The primary goal of this work would be to reduce the number of csproj files so we can get rid of the nested MSBuild invocations in the future.

Thanks for pointing out the simplification that is (sort of) independent from the performance improvements.  +1

> During a port, I'd recommend disabling the tests that don't work with an ActiveIssue ...
> For local testing, there's also a `TestFilter` property that can be used to filter tests at build time ...

Perhaps someone from @dotnet/jit-contrib can respond further on this.  My impression is that this is pretty heavyweight for local dev work.  ActiveIssue is great for locking in wins in CI, but I think a lot of this is "implement add, see that 4/500 tests work, implement sub, see that 8/500 work, do a bunch more before thinking about merging".  TestFilter (is there documentation somewhere on how to use this?  I never ended up learning how myself.) seems like it is aimed for "debug these N tests, for small N, possibly one" - which I think does satisfy the "how do I debug test failure X when X is in a large merged source group" problem (great!).

trylek (approved):
To me this looks awesome but please listen to Mark's feedback as he's the JIT expert, not me.

trylek:
Mark's comments actually nicely tie in with my recent chat with Maoni who pointed out that we should update our various docs to make devs aware of the various trade-offs w.r.t. merged tests, ways of authoring new tests, conditions for using the process isolation property and such, considering that by now more than 95% tests have been converted I can hardly deny this is long overdue. I'll try to put something together in the course of next week but your, Mark's and DavidWr's input will be of paramount importance as the vast majority of tests running under the new system have been authored by the four of us.

src/tests/JIT/CodeGenBringUpTests/CodeGenBringUpTests_do.csproj
```diff
@@ -4,6 +4,7 @@
     <Optimize>False</Optimize>
```
BruceForstall, line 4: Shouldn't this be `<Optimize>True</Optimize>`?

BruceForstall (commented):
So after that, we end up with 4 built test assemblies, each of which runs about 160 tests?

How do you run exactly one test (e.g., the previous `Add1_d.dll`)?

If I `set BuildAsStandalone=true` and build, do I still get each .cs file compiled to a separate binary file?

What is `TestFilter`?

markples:
> So after that, we end up with 4 built test assemblies, each of which runs about 160 tests?

Yes.  (paraphrasing Jeremy's reply above - This may have some short-term benefit in reducing build time, but reducing the project count is itself a goal as many special features / workarounds are dealing with that.  Other folders may be more difficult.  For example, how are tests with different test priorities combined?  However, it may be that the need to only build pri-0 tests for speed is simply due to the number of projects, not the test code, in which case they can all be built and then the wrappers/execution logic can run subsets.)

> How do you run exactly one test (e.g., the previous `Add1_d.dll`)?
> What is `TestFilter`?

You can specify the parameter TestFilter at msbuild time to filter a group down to one test, or if you have a full merged group you can specify the test on the command line.

> If I `set BuildAsStandalone=true` and build, do I still get each .cs file compiled to a separate binary file?

This is the tricky one.  I don't know if it's possible/reasonable to have msbuild generate N projects (really N C# compilations, but I think that has to mean projects), and if so, it's going to deviate even more from the normal build since N projects is exactly what needs to be avoided.

I think the key feature that is needed is the ability to recover from test crashes - the same issue that is already present in merged groups but becomes even more relevant with merged projects.  I could see the merged wrappers invoking themselves recursively as a technique for this (not for one test at a time since that defeats the purpose, but for tests X-Y with the abilty to then run X+k+1 to Y when test X+k crashes) as well as some of the other cases that require process isolation today (like environment variables), but I think Jeremy has a different plan tied into the merged-on-red scripts.

The question would then be what other scenarios require separate binary files for everything.

src/tests/JIT/CodeGenBringUpTests/CodeGenBringUpTests_do.csproj
jkoritzinsky: Yes, you're right.
```suggestion
    <Optimize>True</Optimize>
```

jkoritzinsky:
> > So after that, we end up with 4 built test assemblies, each of which runs about 160 tests?
> 
> Yes. (paraphrasing Jeremy's reply above - This may have some short-term benefit in reducing build time, but reducing the project count is itself a goal as many special features / workarounds are dealing with that. Other folders may be more difficult. For example, how are tests with different test priorities combined? However, it may be that the need to only build pri-0 tests for speed is simply due to the number of projects, not the test code, in which case they can all be built and then the wrappers/execution logic can run subsets.)

My gut feeling is that we can compile all of the tests, pri-0 and pri-1, in all cases once we have less project files. I think Roslyn scales significantly better than MSBuild in this case.

> > How do you run exactly one test (e.g., the previous `Add1_d.dll`)?
> > What is `TestFilter`?
> 
> You can specify the parameter TestFilter at msbuild time to filter a group down to one test, or if you have a full merged group you can specify the test on the command line.

:+1:

> > If I `set BuildAsStandalone=true` and build, do I still get each .cs file compiled to a separate binary file?
> 
> This is the tricky one. I don't know if it's possible/reasonable to have msbuild generate N projects (really N C# compilations, but I think that has to mean projects), and if so, it's going to deviate even more from the normal build since N projects is exactly what needs to be avoided.

Yeah we won't be able to build every test into a separate binary file after merging the projects together. I agree that we should solve the underlying problem (as Mark elucidated) before merging tests that are more likely to regress.

> I think the key feature that is needed is the ability to recover from test crashes - the same issue that is already present in merged groups but becomes even more relevant with merged projects. I could see the merged wrappers invoking themselves recursively as a technique for this (not for one test at a time since that defeats the purpose, but for tests X-Y with the abilty to then run X+k+1 to Y when test X+k crashes) as well as some of the other cases that require process isolation today (like environment variables), but I think Jeremy has a different plan tied into the merged-on-red scripts.
> 
> The question would then be what other scenarios require separate binary files for everything.

Yes, we had an idea to enhance the watcher to have some sort of communication with the merged runner or have some additional features in the log checker that can integrate with the merged runner so we can detect when the test runner crashed and re-launch it to run the remaining tests. We haven't scheduled that work yet.

BruceForstall:
> Yeah we won't be able to build every test into a separate binary file after merging the projects together.

The goal of the CodeGenBringUpTest has been to have the absolute minimum amount of code in each file. Maybe that's not useful after platform bring-up is complete. But if this change is merged and the amount of code increases, it reduces the utility of these for bring-up scenarios. I can imagine if we have such a scenario again, we'd essentially need to undo this change during the bring-up phase.

jkoritzinsky:
> > Yeah we won't be able to build every test into a separate binary file after merging the projects together.
> 
> The goal of the CodeGenBringUpTest has been to have the absolute minimum amount of code in each file. Maybe that's not useful after platform bring-up is complete. But if this change is merged and the amount of code increases, it reduces the utility of these for bring-up scenarios. I can imagine if we have such a scenario again, we'd essentially need to undo this change during the bring-up phase.

I think the solution in the bringup phase would be to build the BringUpTest projects as BuildAsStandalone=true and use ActiveIssue for the tests that haven't been implemented yet. That's one of the use cases I had in mind for the standalone mode: truly minimal "just run my test code".

jkoritzinsky:
Although, even the fact that there's calls there might be an issue I guess.

jkoritzinsky:
Closing as it doesn't make sense to merge these tests if we'd have to unmerge them when they're useful.

PR #94123, opened 2023-10-27
[ILLink analyzer] Add branch analysis

### LINKED ISSUE #2715: RequiresDynamicCode analyzer should understand RuntimeFeature.IsDynamicCodeSupported
Even though M1 below calls a method annotated `RequiresDynamicCode`, it does so [in a code path](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.runtimefeature.isdynamiccodesupported?view=net-6.0) that is not reachable at runtime when dynamic code is not supported. We should not generate a warning.

```csharp
static void M1()
{
    if (RuntimeFeature.IsDynamicCodeSupported)
    {
        M2();
    }
}

[RequiresDynamicCode("...")]
static void M2()
{
}
```

Cc @tlakollo

sbomer:
This adds an understanding of `RuntimeFeature.IsDynamicCodeSupported` to the ILLink Roslyn analyzer. This is done by maintaining an additional state, `FeatureContext`, alongside the tracking done for local variables.

The `FeatureContext` maintains a set of features known to be enabled at a given point. The lattice meet operator is set intersection, so that we get features enabled along all paths, and the top value is the set of all features. At the beginning of the analysis, each basic block starts out in the "top" state, but the entry point gets initialized to the empty set of features. Feature checks (calls to `RuntimeFeature.IsDynamicCodeSupported`) modify the state in the outgoing branch to track the fact that a given feature is enabled.

When a branch operation is seen at the end of a basic block, the analysis examines it to see if it is a supported feature check. If so, it returns an abstract representation of the checked condition (which can be a boolean expression involving various features): `FeatureCheckValue`. Because this may be assumed to be true or false (depending on the outgoing branch), it has to track included and excluded features, unlike `FeatureContext`.

The analysis then separates the output state for the basic block into two, one for the conditional successor, and one for the fall-through successor. It applies the `FeatureCheckValue` to each state separately, once assuming that the check is true, and once assuming that it's false, depending on the branch semantics, possibly modifying the `FeatureContext` in each branch.

To support this, the dataflow state tracking is now done per branch, instead of per basic block. Previously we tracked the state at the end of each block; now we track the state at the input to each edge in the control-flow graph.

The supported feature checks are hard-coded in the analyzer (this change isn't introducing any kind of attribute-based model to replace the feature xml files). For now the only supported check is `RuntimeFeature.IsDynamicCodeSupported`, but it should be easy to add new feature checks. This change includes testing of feature checks for `RequiresUnreferencedCodeAttribute` and `RequiresAssemblyFilesAttribute`, by including code in the analyzer that looks for specific feature guards in the test namespace. Happy to change to another approach if we don't like this.

There are still some pieces of the Requires analyzer logic (generic instantiations and dynamic objects, for example) that need to be moved over to the dataflow analysis, so that the feature checks can act as guards for all of the related patterns. Until that is done, feature checks won't silence those particular warnings. I'll continue working on moving the remaining logic over, but I don't think it needs to block this change.

Fixes https://github.com/dotnet/linker/issues/2715


src/tools/illink/src/ILLink.Shared/DataFlow/ForwardDataFlowAnalysis.cs
```diff
@@ -155,14 +169,61 @@ public TState Get (TBlock block)
...
+			Action<IControlFlowGraph<TBlock, TRegion>.ControlFlowBranch, TValue> updateState
+		)
+		{
+			TConditionValue? conditionValue = transfer.Transfer (block, state);
+
+			if (cfg.GetConditionalSuccessor (block) is IControlFlowGraph<TBlock, TRegion>.ControlFlowBranch conditionalBranch) {
+				// Duplicate the current state so that it's not shared with fall-through state.
+				TValue conditionalCurrentState = lattice.Meet (lattice.Top, state.Current);
+
+				if (conditionValue != null) {
```
vitek-karas, line 192: Vast majority of cases will have conditionValue==null (or should, feature checks should be very rare). Why do we need to duplicate the state for conditions which are not about feature checks?

src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/FeatureSwitchDataFlow.cs
```diff
@@ -0,0 +1,995 @@
...
+		class FeatureCheckBooleanExpressions
+		{
+			[ExpectedWarning ("IL2026", nameof (RequiresUnreferencedCode), ProducedBy = Tool.Trimmer | Tool.NativeAot)]
+			[ExpectedWarning ("IL3050", nameof (RequiresDynamicCode), ProducedBy = Tool.NativeAot)]
+			static void And ()
+			{
+				if (RuntimeFeature.IsDynamicCodeSupported && TestFeatures.IsUnreferencedCodeSupported) {
+					RequiresDynamicCode ();
+					RequiresUnreferencedCode ();
+				}
```
vitek-karas, line 263: How come NativeAOT produces 3050 here? I would expect the branch to be eliminated. I must be missing something here. Is it because of compound condition?

src/tools/illink/src/ILLink.Shared/DataFlow/ITransfer.cs
```diff
@@ -23,16 +23,29 @@ namespace ILLink.Shared.DataFlow
...
 		// that won't be reflected in the caller).
-		void Transfer (TOperation operation, TState state);
+		// The optional return value can be used to model branch conditions.
+		TConditionValue? Transfer (TOperation operation, TState state);
+
+		// Applies a condition value to a dataflow state. This is called in the
+		// outgoing branch of a basic block which returns a TConditionValue
+		// to model the state in the conditional branch where the condition is
+		// assumed to be true.
+		void ApplyCondition (TConditionValue condition, ref TValue value);
```
vitek-karas, line 49: "Applies a condition value to a dataflow state." this is confusing. It applies the condition to a "value" (per the interface terminology).

Aside - the "value" here is super confusing, because in the usage we actually call that thing a "state" as well - but that's not something we should fix with this PR.

src/tools/illink/src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -205,6 +207,24 @@ internal static ValueSet<TValue> Meet (ValueSet<TValue> left, ValueSet<TValue> r
 			return new ValueSet<TValue> (values);
 		}
 
+		internal static ValueSet<TValue> Intersection (ValueSet<TValue> left, ValueSet<TValue> right)
+		{
+			if (left._values == null)
+				return left;
```
vitek-karas, line 213: Might be worth adding a comment that we don't need DeepCopy, since the value is empty. (It's confusing that Union has to call DeepCopy everythwere, and Intersetion only sometimes)
Or alternatively we migth return `Empty` - for better readability.

src/tools/illink/src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -205,6 +207,24 @@ internal static ValueSet<TValue> Meet (ValueSet<TValue> left, ValueSet<TValue> r
...
+				return right;
+
+			if (left._values is not EnumerableValues)
+				return right.Contains ((TValue) left._values) ? left.DeepCopy () : default;
+
+			if (right._values is not EnumerableValues)
+				return left.Contains ((TValue) right._values) ? right.DeepCopy () : default;
+
+			var values = new EnumerableValues (left.DeepCopy ());
+			values.IntersectWith (right.DeepCopy ());
```
vitek-karas, line 224: Do we need the DeepCopy here? I would expect the IntersectWith to not add the value fromt he parameter to the hashset - since it already has all the values it needs (unlike Union).

src/tools/illink/src/ILLink.Shared/DataFlow/IControlFlowGraph.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.Contracts;
```
vitek-karas, line 8: This can be removed

src/tools/illink/src/ILLink.RoslynAnalyzer/DataFlow/ControlFlowGraphProxy.cs
```diff
@@ -51,25 +55,40 @@ public override string ToString ()
...
 
+		public static ControlFlowBranch? CreateProxyBranch (Microsoft.CodeAnalysis.FlowAnalysis.ControlFlowBranch? branch)
+		{
+			if (branch == null)
+				return null;
+
+			var finallyRegions = ImmutableArray.CreateBuilder<RegionProxy> ();
+			foreach (var region in branch.FinallyRegions) {
+				if (region == null)
+					throw new InvalidOperationException ();
```
vitek-karas, line 66: Do we really want to crash the analyzer here?

src/tools/illink/src/ILLink.Shared/DataFlow/IControlFlowGraph.cs
```diff
@@ -19,23 +20,79 @@ public enum RegionKind
...
 
-		public readonly struct Predecessor
+		// Represents an edge in the control flow graph, from a source to a destination basic block.
+		public readonly struct ControlFlowBranch : IEquatable<ControlFlowBranch>
 		{
-			public readonly TBlock Block;
+			public readonly TBlock Source;
+			private readonly TBlock? Destination;
+
 			public readonly ImmutableArray<TRegion> FinallyRegions;
```
vitek-karas, line 51: It is weird to think about "Finally reagions on an edge in the CFG" - in reality these are the finally reagions of the source block. I assume we don't model that in this class directly because we can't create the proxies here, but otherwise this should be basically `=> Source.FinallyRegions`.
Following that - I would rename this to `SourceFinallyRegions` to make it clearer.

Looking at this more - another option would be to add a `GetSourceFinallyRegions(ControlFlowBranch branch)` into the `IControlFlowGraph` interface. We model all of the other "accessors" there as well, so why do we have to make this one special. There's no state associated with the branch for the finally regions anyway, is there?

src/tools/illink/src/ILLink.Shared/DataFlow/IControlFlowGraph.cs
```diff
@@ -19,23 +20,79 @@ public enum RegionKind
...
+		public readonly struct Successor
+		{
+			public readonly TBlock Block;
+			public readonly bool IsConditional;
+			public Successor (TBlock block, bool isConditional)
 			{
-				(Block, FinallyRegions) = (block, finallyRegions);
+				(Block, IsConditional) = (block, isConditional);
 			}
 		}
```
vitek-karas, line 97: This is not used anywhere I think... ?

vitek-karas (commented):
Didn't review all of it - but enough comments to submit the review so that I don't lose them :-)

src/tools/illink/src/ILLink.RoslynAnalyzer/DataFlow/LocalDataFlowVisitor.cs
```diff
@@ -660,12 +697,89 @@ public override TValue VisitFlowAnonymousFunction (IFlowAnonymousFunctionOperati
...
+				//     DoesNotReturn();
+				// UseFeature();
+				//
+				// The feature context at 'UseFeature()' is computed as the meet of the if branch output
+				// (all features enabled), and the fall-through output of the block before the check (feature is enabled)
+				// so that we see the feature enabled at 'UseFeature()'.
+				var current = state.Current;
+				current.Context = LocalContextLattice.ContextLattice.Top;
+				state.Current = current;
+				return value;
```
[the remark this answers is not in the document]
vitek-karas: Interesting - but I guess makes sense since Roslyn can't reason about multi-threading or reflection really.

src/tools/illink/src/ILLink.RoslynAnalyzer/RequiresUnreferencedCodeAnalyzer.cs
```diff
@@ -60,6 +61,10 @@ public sealed class RequiresUnreferencedCodeAnalyzer : RequiresAnalyzerBase
 
 		private protected override string RequiresAttributeName => RequiresUnreferencedCodeAttribute;
 
+		public const string UnreferencedCode = nameof (UnreferencedCode);
+
+		internal override string FeatureName => UnreferencedCode;
```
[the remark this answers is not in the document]
vitek-karas: Right - my bad.

src/tools/illink/src/ILLink.Shared/DataFlow/ITransfer.cs
vitek-karas: I'm not ready for that naming discussion :-)

src/tools/illink/src/ILLink.Shared/DataFlow/IControlFlowGraph.cs
vitek-karas: Thanks for the explanation - might be worth a comment - I know it's "the same way as Roslyn", but Roslyn doesn't have a doc about it, so leaving something in our codebase should help.

src/tools/illink/src/ILLink.RoslynAnalyzer/DataFlow/LocalStateAndContextLattice.cs
```diff
@@ -0,0 +1,56 @@
...
+		}
+
+		public bool Equals (LocalStateAndContext<TValue, TContext> other) =>
+			LocalState.Equals (other.LocalState) && Context.Equals (other.Context);
+
+		public override bool Equals (object? obj) => obj is LocalStateAndContext<TValue, TContext> other && Equals (other);
+		public override int GetHashCode () => HashUtils.Combine (LocalState, Context);
+	}
+
+	public readonly struct LocalContextLattice<TValue, TContext, TValueLattice, TContextLattice> : ILattice<LocalStateAndContext<TValue, TContext>>
```
vitek-karas, line 31: Should be probably called `LocalStateAndContextLattice`

src/tools/illink/src/ILLink.RoslynAnalyzer/TrimAnalysis/TrimDataFlowAnalysis.cs
```diff
@@ -13,33 +13,47 @@
...
 
-		public TrimDataFlowAnalysis (OperationBlockAnalysisContext context, IOperation operationBlock)
-			: base (context, operationBlock)
+		DataFlowAnalyzerContext _dataFlowAnalyzerContext;
+
+		public TrimDataFlowAnalysis (
+			OperationBlockAnalysisContext context,
+			DataFlowAnalyzerContext dataFlowAnalyzerContext,
+			IOperation operationBlock)
+			: base (context, operationBlock, initialContext: FeatureContext.None)
```
vitek-karas, line 40: Probably worth the comment about the initial context state being "none".

vitek-karas (commented):
Reviewed all of product code, still have tests to do.

src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/FeatureCheckDataFlow.cs
```diff
@@ -0,0 +1,999 @@
...
+			GuardedPatterns.Test ();
+			ExceptionalDataFlow.Test ();
+		}
+
+		class CallFeatureUnguarded
+		{
+			[ExpectedWarning ("IL3050", nameof (RequiresDynamicCode), ProducedBy = Tool.Analyzer | Tool.NativeAot)]
+			static void Unguarded ()
+			{
+				RequiresDynamicCode ();
```
vitek-karas, line 34: For completeness I would add RUC and RAF to this and maybe the if test below. As a way to validate that the feature check for RDC doesn't accidentally also turn off RUC/RAF warnings.
Not needed for all the combinations, just a couple.

src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/FeatureCheckDataFlow.cs
```diff
@@ -0,0 +1,999 @@
...
+				GuardedDoesNotReturnIfTrue ();
+				GuardedDoesNotReturnIfFalse ();
+				GuardedDoesNotReturn ();
+				GuardedDoesNotReturnIfFalseCtor ();
+			}
+
+			static void GuardedIf ()
+			{
+				if (RuntimeFeature.IsDynamicCodeSupported)
+					RequiresDynamicCode ();
```
vitek-karas, line 167: Same as above - for a couple of these add RUC/RAF method calls to check that the feature check for RDC has no effect on those.

src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/FeatureCheckDataFlow.cs
```diff
@@ -0,0 +1,999 @@
...
+			[ExpectedWarning ("IL2026", nameof (RequiresUnreferencedCode), ProducedBy = Tool.Trimmer | Tool.NativeAot)]
+			[ExpectedWarning ("IL3050", nameof (RequiresDynamicCode), ProducedBy = Tool.NativeAot)]
+			static void And ()
+			{
+				if (RuntimeFeature.IsDynamicCodeSupported && TestFeatures.IsUnreferencedCodeSupported) {
+					RequiresDynamicCode ();
+					RequiresUnreferencedCode ();
+				}
+			}
+
```
vitek-karas, line 266: To get consistent behavior for trimmer we should pass the right feature switch so that IsDynamicCodeSupported is substituted to false. The NativeAOT already does this in the test infra for all tests: https://github.com/dotnet/runtime/blob/df6fdefa27068126794b253d4d822706221a92db/src/coreclr/tools/aot/ILCompiler.Trimming.Tests/TestCasesRunner/TrimmingArgumentBuilder.cs#L45C69-L45C69
This test case should also do it for the trimmer.

Similarly to emulate the equivalent behavior in AOT we should add a substitutions XML for the AOT tests which will set the TestFeature properties to constants (since AOT can't do constant prop from return values).

We should note the behavior differences, but for these tests we should try to make them behave as close to each other as possible since the goal of the test is to validate all the patterns which are correctly recognized by each tool - specifying feature switches and substitutions is not part of that I think.

src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/FeatureCheckDataFlow.cs
```diff
@@ -0,0 +1,999 @@
...
+				FalseNotEquals ();
+				IsTrue ();
+				IsFalse ();
+				IsNotTrue ();
+				IsNotFalse ();
+				Contradiction ();
+			}
+		}
+
+		class SupportedFeatureChecks
```
vitek-karas, line 440: Should this be called TestFeatureChecks (since it only validates the test-only stuff)

vitek-karas (approved):

vitek-karas (commented):
Some comments on the tests, but this looks good. Thanks a lot!

src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/FeatureCheckDataFlowTestSubstitutions.xml
```diff
@@ -0,0 +1,8 @@
+<linker>
+  <assembly fullname="test, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null">
+    <type fullname="ILLink.RoslynAnalyzer.TestFeatures">
```
vitek-karas, line 3: ILLink should allow specifying a substitution for CoreLib in the main app's assembly - if I remember correctly there are only two assemblies which are allowed to specify substitutions for "other" assembly - CoreLib and the main app's assembly (and command line substitutions obviously).
So for ILLink we could either add a substitutions for `IsDynamicCodeSupported` here by targeting CoreLib in the XML, or we could pass it as a command line parameter and separate xml file.

src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/FeatureCheckDataFlow.cs
```diff
@@ -8,11 +8,17 @@
...
+using Mono.Linker.Tests.Cases.Expectations.Metadata;
 
 namespace Mono.Linker.Tests.Cases.DataFlow
 {
 	[SkipKeptItemsValidation]
 	[ExpectedNoWarnings]
+	// Note: the XML must be passed as an embedded resource named ILLink.Substitutions.xml,
+	// not as a separate substitution file, for it to work with NativeAot.
+	// Related: https://github.com/dotnet/runtime/issues/88647
+    [SetupCompileResource ("FeatureCheckDataFlowTestSubstitutions.xml", "ILLink.Substitutions.xml")]
```
vitek-karas, line 20: Nit: Indentation

src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/FeatureCheckDataFlowTestSubstitutions.xml
vitek-karas: OK - makes sense... I'm not asking to change anything, this looks good.

src/tools/illink/src/ILLink.RoslynAnalyzer/RequiresAssemblyFilesAnalyzer.cs
```diff
@@ -58,6 +61,21 @@ internal override bool IsAnalyzerEnabled (AnalyzerOptions options)
...
 
+		internal override bool IsRequiresCheck (Compilation compilation, IPropertySymbol propertySymbol)
+		{
+			// "IsAssemblyFilesSupported" is treated as a requires check for testing purposes only, and
+			// is not officially-supported product behavior.
+			var runtimeFeaturesType = compilation.GetTypeByMetadataName ("ILLink.RoslynAnalyzer.TestFeatures");
+			if (runtimeFeaturesType == null)
+				return false;
+
+			var isDynamicCodeSupportedProperty = runtimeFeaturesType.GetMembers ("IsAssemblyFilesSupported").OfType<IPropertySymbol> ().FirstOrDefault ();
```
eerhardt, line 72: Did we add code/tests for IsDynamicCodeCompiled? See https://github.com/dotnet/linker/issues/2715#issuecomment-1212577966

eerhardt:
Can we also remove these suppressions now?

https://github.com/dotnet/runtime/blob/e61f45ac1d8cfccbf40c77bfcbb1c26a73ccd7b8/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializerOptionsUpdateHandler.cs#L28-L33

https://github.com/dotnet/runtime/blob/e61f45ac1d8cfccbf40c77bfcbb1c26a73ccd7b8/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceProvider.cs#L252-L266

https://github.com/dotnet/runtime/blob/e61f45ac1d8cfccbf40c77bfcbb1c26a73ccd7b8/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs#L86C1-L87

PR #94130, opened 2023-10-28
JIT: update tail call IR validity checks

AndyAyersMS:
Remove `fgCheckStmtAfterTailCall` as it did less thorough and less correct checking. We will rely on `fgValidateIRForTailCall` to sanity check that there is no unexpcted IR after a tail call.

Contributes to #93246.

### DIFF
--- a/src/coreclr/jit/morph.cpp
+++ b/src/coreclr/jit/morph.cpp
@@ -5894,12 +5894,6 @@ GenTree* Compiler::fgMorphPotentialTailCall(GenTreeCall* call)
         }
     }
 
-    if (!fgValidateIRForTailCall(call))
-    {
-        failTailCall("Unexpected statements after the tail call");
-        return nullptr;
-    }
-
     const char* failReason      = nullptr;
     bool        canFastTailCall = fgCanFastTailCall(call, &failReason);
 
@@ -6087,6 +6081,8 @@ GenTree* Compiler::fgMorphPotentialTailCall(GenTreeCall* call)
         call->gtArgs.ResetFinalArgsAndABIInfo();
 #endif
     }
+
+    fgValidateIRForTailCall(call);
 
     // If this block has a flow successor, make suitable updates.
     //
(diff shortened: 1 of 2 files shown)


AndyAyersMS:
FYI @dotnet/jit-contrib 

No diffs expected.

See  https://github.com/dotnet/runtime/pull/93997#issuecomment-1783492576 for context.

src/coreclr/jit/morph.cpp
```diff
@@ -6452,24 +6483,39 @@ void Compiler::fgValidateIRForTailCall(GenTreeCall* call)
         }
     };
 
+    JITDUMP("Validating IR for tail call candidate [%06u] in " FMT_STMT "\n", dspTreeID(call), compCurStmt->GetID());
+
     TailCallIRValidatorVisitor visitor(this, call);
-    for (Statement* stmt = compCurStmt; stmt != nullptr; stmt = stmt->GetNextStmt())
+    for (Statement* stmt = compCurStmt; visitor.IsValid() && (stmt != nullptr); stmt = stmt->GetNextStmt())
     {
         visitor.WalkTree(stmt->GetRootNodePointer(), nullptr);
```
jakobbotsch, line 6491: Nit: I think it would be a bit cleaner to return `WALK_ABORT` on validation failure in the visitor (looks like at least `GT_RETURN` needs to be fixed) and have early exits for when `WalkTree` returns that.

jakobbotsch (approved):

jakobbotsch:
Hmm, I just realized this is problematic for explicit tailcalls... We cannot be allowed to invalidate those. Is it possible to have #93997 avoid merging tailcall cases instead?

AndyAyersMS:
> Hmm, I just realized this is problematic for explicit tailcalls... We cannot be allowed to invalidate those. Is it possible to have #93997 avoid merging (explicit) tailcall cases instead?

Yes, we can do something like that. I am currently thinking:
* don't merge any tail call cases in `fgAddInternal`
* enhance tail merge to handle returns specially (think you pointed this out this a while back)
* when we get to morph, if there are any tail call cases, mark the `genReturnBB` as out of order for RPO, since it may gain preds during morph. So we'll lose any morph assertion prop into `genReturnBB` but that should be ok.

I can adapt this PR to go back to assertion checking like it was doing before, and drop `fgCheckStmtAfterTailCal`.

jakobbotsch:
That sounds good to me.

AndyAyersMS:
Revised per above.

jakobbotsch (approved):

AndyAyersMS:
Failures are known.

Issue #93890, opened 2023-10-23
New TimeSpan.From overloads which take integers.


tannergooding:
### Summary

`TimeSpan` exposes several `From*` methods that allow users to create a timespan using a `double`. However, `double` is a binary based floating-point format and thus has natural imprecision that can introduce error. For example, `TimeSpan.FromSeconds(101.832)` is not `101 seconds, 832 milliseconds`, but rather is `101 seconds, 831.9999999999936335370875895023345947265625 milliseconds`.

This has, over time, led to repeated user confusion and bugs in the API surface that users have had to rationalize and deal with. It is also one of the less efficient ways to represent this data and makes it problematic for users wanting or expecting a particular behavior

As such, it is proposed that new overloads be provided, allowing users to pass in integers so they can get the desired and intended behavior. It may be worth considering whether or not the `double` overloads should be "deprecated" (either via obsoletion or some analyzer) to help direct users towards success.

### API Proposal

```csharp
namespace System;

public partial struct TimeSpan
{
    public static TimeSpan FromDays(int days, int hours = 0, int minutes = 0, int seconds = 0, int milliseconds = 0, int microseconds = 0);
    public static TimeSpan FromHours(int hours, int minutes = 0, int seconds = 0, int milliseconds = 0, int microseconds = 0);
    public static TimeSpan FromMinutes(int minutes, int seconds = 0, int milliseconds = 0, int microseconds = 0);
    public static TimeSpan FromSeconds(int seconds, int milliseconds = 0, int microseconds = 0);
    public static TimeSpan FromMilliseconds(int milliseconds, int microseconds = 0);
    public static TimeSpan FromMicroseconds(int microseconds);
}
```

### Additional Considerations

It may be desirable to not use default parameters. However, this provides a benefit that a user can do something like `FromDays(5, seconds: 30)` if that were appropriate for their use case.

After these APIs are exposed, calling `From(x)` where `x` was an integer may produce a different result in various edge cases.

APIs where the user specifies a `uint`, `long`, or `ulong` would still end up calling the `double` overload due to the implicit conversion that exists in C#. Thus, an analyzer or deprecation of the `double` overloads may be desirable to help users achieve success.

Exposing overloads that take `decimal` could also be desirable and would avoid the issues around precision loss, but would be much less performant than breaking it apart like this.

tannergooding:
[...]

PR #94143, opened 2023-10-28
Implements new `int`and `long` overloads for `TimeSpan`

tcortega:
I went ahead and took the liberty to add an `public static System.TimeSpan FromMilliseconds(long milliseconds);` that was not detailed in the proposal, but felt like it does not make sense to not add it.


tcortega:
@dotnet-policy-service agree

tcortega:
Resolves #93890

danmoseley:
> I went ahead and took the liberty to add an public static System.TimeSpan FromMilliseconds(long milliseconds); that was not detailed in the proposal, but felt like it does not make sense to not add it

It will need API review if it wasn't approved... that may be quick.

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
```diff
@@ -271,6 +271,22 @@ public static TimeSpan FromDays(double value)
             return Interval(value, TicksPerDay);
         }
 
+        public static TimeSpan FromDays(int days)
+        {
+            return new TimeSpan((long)days * TicksPerDay);
+        }
+
+        public static TimeSpan FromDays(int days, int hours = 0, long minutes = 0, long seconds = 0, long milliseconds = 0, long microseconds = 0)
```
tarekgh, line 279: > public static TimeSpan FromDays(int days, int hours = 0, long minutes = 0, long seconds = 0, long milliseconds = 0, long microseconds = 0) [](http://example.com/codeflow?start=7&length=139)

Please look at the code of the constructor `public TimeSpan(int days, int hours, int minutes, int seconds, int milliseconds, int microseconds)`. It has some logic to ensure no overflow by checking the total max microseconds. I suggest you refactor this code into its own method and then call it from here and from the constructor.

tarekgh: wait, we are using long and not int for the parameters. I think we need to implement this to ensure there is no overflow will occur.

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
```diff
@@ -307,6 +323,21 @@ public static TimeSpan FromHours(double value)
             return Interval(value, TicksPerHour);
         }
 
+        public static TimeSpan FromHours(int hours)
+        {
+            return new TimeSpan((long)hours * TicksPerHour);
+        }
+
+        public static TimeSpan FromHours(int hours, long minutes = 0, long seconds = 0, long milliseconds = 0, long microseconds = 0)
```
tarekgh, line 331: >TimeSpan [](http://example.com/codeflow?start=22&length=8)

all introduced method need to ensure no overflow and and not exceeding the max allowed value.

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
```diff
@@ -271,6 +271,22 @@ public static TimeSpan FromDays(double value)
             return Interval(value, TicksPerDay);
         }
 
+        public static TimeSpan FromDays(int days)
```
tarekgh, line 274: >public [](http://example.com/codeflow?start=8&length=6)

All APIs need to have the Xml doc elements.

tarekgh:
> I went ahead and took the liberty to add an public static System.TimeSpan FromMilliseconds(long milliseconds); that was not detailed in the proposal, but felt like it does not make sense to not add it.

It doesn't make sense to have `FromMilliseconds(long milliseconds);`. The proposal already has `public static TimeSpan FromMilliseconds(long milliseconds, long microseconds = 0);`

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
tcortega: @tarekgh I'm struggling to see the benefit of manually checking for overflows given that `long.MaxValue` represents the upper limit for TimeSpan's length. Any computation exceeding this limit inherently triggers an `OverflowException`. Implementing explicit checks seems to introduce additional CPU overhead, thus potentially reducing performance.

Here are some benchmarks:
```
| Method                            | Input               | Mean          | Error      | StdDev     | Gen0   | Allocated |
|---------------------------------- |-------------------- |--------------:|-----------:|-----------:|-------:|----------:|
| DefaultOverflowException          | 9223372036854775807 |     0.0192 ns |  0.0177 ns |  0.0165 ns |      - |         - |
| ManualOverflowChecksThenException | 9223372036854775807 | 3,188.9095 ns | 42.1336 ns | 39.4118 ns | 0.0229 |     240 B |
```

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
tcortega: Waiting for your reply in the review up there

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
tcortega: Haha I originally wrote it with the XML Docs but noticed that the older methods do not have any documentation and removed them. No problem.

tcortega:
> > I went ahead and took the liberty to add an public static System.TimeSpan FromMilliseconds(long milliseconds); that was not detailed in the proposal, but felt like it does not make sense to not add it.
> 
> It doesn't make sense to have `FromMilliseconds(long milliseconds);`. The proposal already has `public static TimeSpan FromMilliseconds(long milliseconds, long microseconds = 0);`

The main benefit appears to be the conservation of CPU cycles. Otherwise, why have both `TimeSpan FromDays(int days)` and an overloaded `TimeSpan FromDays(int days, int hours = 0, ...)` if the latter could suffice? It seems the oversight here is the lack of a `TimeSpan FromMilliseconds(long milliseconds)` method.

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
tarekgh: > Any computation exceeding this limit inherently triggers an OverflowException

No. this will trigger it if the operation is `checked` which we don't enable by default. https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
tarekgh: replied. We can try to optimize the checking as much as we can but have to ensure we produce the correct behavior.

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
tcortega: > > Any computation exceeding this limit inherently triggers an OverflowException
> 
> No. this will trigger it if the operation is `checked` which we don't enable by default. https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked

I see, and what are the implications of having the `checked` keyword being used explicitly in there?

tarekgh:
> The main benefit appears to be the conservation of CPU cycles. Otherwise, why have both TimeSpan FromDays(int days) and an overloaded TimeSpan FromDays(int days, int hours = 0, ...) if the latter could suffice? It seems the oversight here is the lack of a TimeSpan FromMilliseconds(long milliseconds) method.

Agree, in the approved overloads, the second parameter shouldn't be defaulted on. @bartonjs @tannergooding do you have thoughts about that?

tcortega:
> > The main benefit appears to be the conservation of CPU cycles. Otherwise, why have both TimeSpan FromDays(int days) and an overloaded TimeSpan FromDays(int days, int hours = 0, ...) if the latter could suffice? It seems the oversight here is the lack of a TimeSpan FromMilliseconds(long milliseconds) method.
> 
> Agree, in the approved overloads, the second parameter shouldn't be defaulted on. @bartonjs @tannergooding do you have thoughts about that?

Btw, if the purpose of these methods is to save CPU cycles, wouldn't this imply that we need to implement overflow checks repeatedly? Without these checks, we'd still incur the CPU cost for computations like `0 * N + N...`, regardless of their simplicity.

tarekgh:
> Btw, if the purpose of these methods is to save CPU cycles

I don't think this is the main goal. The main goal is to provide API produce accurate results no rounding or truncation as the overloads that take double. Still be good to optimize as much as we can as a secondary goal.

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
tarekgh: enable `checked` still have some perf cost too and we'll need to decide if we need to catch the thrown overflow exception and convert it to something like ThrowHelper.ThrowArgumentOutOfRange_TimeSpanTooLong(). @bartonjs do you know if we have any design guidelines around using `checked` in the APIs? I did some search but couldn't find any.

tcortega:
> > Btw, if the purpose of these methods is to save CPU cycles
> 
> I don't think this is the main goal. The main goal is to provide API produce accurate results no rounding or truncation as the overloads that take double. Still be good to optimize as much as we can as a secondary goal.

I understand that delivering precise results is the primary objective. However, my point is that the rationale for having distinct methods like `FromDays` and others would be to enhance their performance; without such an improvement, their separate existence wouldn't be justifiable.

tarekgh:
> without such an improvement, their separate existence wouldn't be justifiable.

To be honest, I have a different perspective. The "From" methods we offer allow you to create a time span without needing to provide all the parameters that constructors require. This is why we are utilizing the default parameters. I am not saying we should not care about performance. I am trying to say we should first ensure the precision and user experience of the APIs usage first and then try to make the APIs performant without scarifying the main goal.

tcortega:
> > without such an improvement, their separate existence wouldn't be justifiable.
> 
> To be honest, I have a different perspective. The "From" methods we offer allow you to create a time span without needing to provide all the parameters that constructors require. This is why we are utilizing the default parameters. I am not saying we should not care about performance. I am trying to say we should first ensure the precision and user experience of the APIs usage first and then try to make the APIs performant without scarifying the main goal.

But from the user's perspective, what would be the difference if we removed the `TimeSpan FromDays(int days)`?

bartonjs:
> Otherwise, why have both TimeSpan FromDays(int days) and an overloaded TimeSpan FromDays(int days, int hours = 0, ...) if the latter could suffice? 

Usability studies have shown that junior/novice developers get confused with the IntelliSense presentation of methods with many default parameters.  That resulted in this guideline in [Framework Design Guidelines, 3rd edition](DO provide a simple overload, with no default parameters, for any
method with two or more defaulted parameters.) (specifically in 5.1.1 (Member Overloading), p143): `DO provide a simple overload, with no default parameters, for any method with two or more defaulted parameters.`.  `FromDays` had more than 1.  `FromMilliseconds` only had 1.

bartonjs:
There's also room for improved performance (from various reasons) from the shorter overloads, but the reason they got added in API Review was because of the usability guideline, not a performance goal.

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
danmoseley: Yes we are slowly adding missing XML docs through these libraries. Historically docs were not generated from sources (unlike most .NET libraries out there - perhaps because many were written before the XML generates useful docs). As we copy the official docs onto the API (aided by tools) we switch one library after another to being the "source of truth" for docs and eventually they all will be. There is an ongoing project for this in case you are interested in contributing to it at some point.

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
danmoseley: @tannergooding  do you see any trick to efficiently detect overflow for this particular calculation? Something better than doing each one in turn and checking whether the sign changed after each multiplication, or doing it all in an int128.

tarekgh:
@bartonjs 

> Usability studies have shown that junior/novice developers get confused with the IntelliSense presentation of methods with many default parameters.

Should we add `public static TimeSpan FromMilliseconds(long milliseconds);` to the proposal then? 

Also, in the approved APIs, should all overloads that taking optional parameters has the second argument be not defaulted? like 

```C#
    public static TimeSpan FromDays(int days);
    public static TimeSpan FromDays(int days, int hours = 0, long minutes = 0, long seconds = 0, long milliseconds = 0, long  microseconds = 0);
```

the hours parameter shouldn't be defaulted. or do we have this if someone uses named arguments?

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
tarekgh: That is what I was trying to point at in my comment https://github.com/dotnet/runtime/pull/94143#discussion_r1375532302

bartonjs:
> Should we add public static TimeSpan FromMilliseconds(long milliseconds); to the https://github.com/dotnet/runtime/issues/93890#issuecomment-1777761585 then?

No, because the current method for it does not have two or more defaulted parameters.  It has one mandatory parameter and one optional/defaulted parameter.

> Also, in the approved APIs, should all overloads that taking optional parameters has the second argument be not defaulted? like ... the hours parameter shouldn't be defaulted. or do we have this if someone uses named arguments?

The way that it got approved is correct.  The number of hours in FromDays should, indeed, be defaulted so that someone can call `FromDays(192, minutes: 3)`, if they like.

tarekgh:
@tcortega let's continue with the original approved proposal. Could you please update the code according to the discussions here? Thanks!

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
tannergooding: It really depends on how complicated we want to make it...

The simplest implementation is probably this:
```csharp
public static TimeSpan FromHours(int hours, long minutes = 0, long seconds = 0, long milliseconds = 0, long microseconds = 0)
{
    return FromHours(hours)
         + FromMinutes(minutes)
         + FromSeconds(seconds)
         + FromMilliseconds(milliseconds)
         + FromMicroseconds(microseconds);
}
```

This is no worse than the naive implementation a user might do and for all constant inputs, the JIT can produce a constant. It currently doesn't because the implementation of `operator +` is "bad" in several ways, but it does when that gets fixed. -- The static operator defers to an instance method and directly throws. It should be the inverse where the instance method defers to the static operator and a throw helper is used instead.

It's not, strictly speaking, the most efficient implementation for non-constant inputs. But that's probably fine at the end of the day and helps what is probably the overall more common scenario.

tannergooding: Likewise, if we didn't require the `OverflowException` to exactly contain the message `"TimeSpan overflowed because the duration is too long"`, we could simply make `add` do `return new TimeSpan(checked(xTicks + yTicks))`. There would be some JIT work to still allow constant folding when `x op y` can be statically determined to not overflow still, but that seems like something generally worthwhile to do so that using `checked(...)` doesn't completely pessimize codegen and can be preferred over manual checks.

tannergooding: alternatively we could also update the JIT to recognize the `if ((xTicks >> 63 == yTicks >> 63) && (xTicks >> 63 != result >> 63))` pattern and have it emit `jo` (jump if overflow) rather than the two comparisons and jumps it currently does. This would also allow the same manual checks to be more efficient.

both of those two JIT improvements seem like relatively worthwhile to help improve throughput for our own code when we need to handle such cases.

tcortega: > alternatively we could also update the JIT to recognize the `if ((xTicks >> 63 == yTicks >> 63) && (xTicks >> 63 != result >> 63))` pattern and have it emit `jo` (jump if overflow) rather than the two comparisons and jumps it currently does. This would also allow the same manual checks to be more efficient.
> 
> both of those two JIT improvements seem like relatively worthwhile to help improve throughput for our own code when we need to handle such cases.

Benchmark tests were conducted for several proposed solutions and others that came to mind. The code snippets can be found here:
https://gist.github.com/tcortega/f518f899b3b80cb62bdc3b5e8514f3bc

The outcome of the benchmarks, specifically in scenarios where an exception is always thrown, is shown in milliseconds:
```
| Method            | Platform | Mean      | Error     | StdDev    | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|------------------ |--------- |----------:|----------:|----------:|------:|--------:|-------:|----------:|------------:|
| ManualChecks      | X64      |  4.577 us | 2.3261 us | 0.1275 us |  1.00 |    0.00 | 0.0610 |     576 B |        1.00 |
| Int128Checks      | X64      |  4.583 us | 1.1302 us | 0.0619 us |  1.00 |    0.04 | 0.0610 |     576 B |        1.00 |
| SmartInt128Checks | X64      |  5.176 us | 0.0715 us | 0.0039 us |  1.13 |    0.03 | 0.0610 |     576 B |        1.00 |
| CheckedKeyword    | X64      | 10.475 us | 2.2673 us | 0.1243 us |  2.29 |    0.09 | 0.0610 |     584 B |        1.01 |
|                   |          |           |           |           |       |         |        |           |             |
| ManualChecks      | X86      |  5.874 us | 0.1566 us | 0.0086 us |  1.00 |    0.00 | 0.0534 |     292 B |        1.00 |
| Int128Checks      | X86      |  6.027 us | 0.3815 us | 0.0209 us |  1.03 |    0.00 | 0.0534 |     292 B |        1.00 |
| SmartInt128Checks | X86      |  6.271 us | 0.6882 us | 0.0377 us |  1.07 |    0.01 | 0.0534 |     292 B |        1.00 |
| CheckedKeyword    | X86      | 15.735 us | 8.1722 us | 0.4479 us |  2.68 |    0.07 | 0.0305 |     300 B |        1.03 |
```
Results for scenarios without exceptions, simulating expected typical user inputs for the API:
```
| Method            | Platform | Mean      | Error     | StdDev    | Ratio | RatioSD | Allocated | Alloc Ratio |
|------------------ |--------- |----------:|----------:|----------:|------:|--------:|----------:|------------:|
| ManualChecks      | X64      |  4.239 ns | 4.3787 ns | 0.2400 ns |  1.00 |    0.00 |         - |          NA |
| Int128Checks      | X64      |  5.655 ns | 1.0012 ns | 0.0549 ns |  1.34 |    0.08 |         - |          NA |
| SmartInt128Checks | X64      | 10.133 ns | 0.4728 ns | 0.0259 ns |  2.40 |    0.14 |         - |          NA |
| CheckedKeyword    | X64      |  2.056 ns | 0.4019 ns | 0.0220 ns |  0.49 |    0.03 |         - |          NA |
|                   |          |           |           |           |       |         |           |             |
| ManualChecks      | X86      | 16.395 ns | 1.6665 ns | 0.0913 ns |  1.00 |    0.00 |         - |          NA |
| Int128Checks      | X86      | 52.030 ns | 7.5380 ns | 0.4132 ns |  3.17 |    0.04 |         - |          NA |
| SmartInt128Checks | X86      | 42.805 ns | 4.0426 ns | 0.2216 ns |  2.61 |    0.02 |         - |          NA |
| CheckedKeyword    | X86      | 19.274 ns | 2.3722 ns | 0.1300 ns |  1.18 |    0.01 |         - |          NA |
```
Based on these results, the ManualChecks approach seems promising. Despite being slightly less legible, it demonstrates superior performance. The potential for future JIT optimizations strengthens the case for this method.

Any thoughts?

tcortega: The CheckedKeyword results also look decent. Considering exceptions should be rare, the extra 10 microseconds and 8 bytes might not be a big deal.

tannergooding: I’d just go with the simple approach I detailed above as it lends the most capability for the JIT to optimize it.

We can revisit and optimize further if required later. Some of the easiest optimizations will be low hanging fruit such as improving `operator +`

tannergooding: That is, I'd go with doing this:
```csharp
public static TimeSpan FromHours(int hours, long minutes = 0, long seconds = 0, long milliseconds = 0, long microseconds = 0)
{
    return FromHours(hours)
         + FromMinutes(minutes)
         + FromSeconds(seconds)
         + FromMilliseconds(milliseconds)
         + FromMicroseconds(microseconds);
}
```

The other patterns will end up with less capability for the JIT to optimize, especially for constant inputs, and that is overall less desirable.

tannergooding: I opened https://github.com/dotnet/runtime/pull/94226 to fix up some of the cases like `operator +` so that the relevant inlining can occur.

tcortega: > I opened #94226 to fix up some of the cases like `operator +` so that the relevant inlining can occur.

Great, I already have the solution you wanted implemented, now I just have to write a few unit tests for it, although I'm not really certain what to test in specific. I guess I should test overflows and correct TimeSpans?

tarekgh:
@tcortega are you willing to finish this PR. @stefannikolei is interested in helping too. Thanks!

tarekgh:
@stefannikolei please create a new PR based on the discussions in this PR. Thanks @tcortega for letting us know.

PR #94146, opened 2023-10-29
Fix test returning 100 unconditionally

gbalykov:
Xunit error XUW1002: Tests should not unconditionally return 100. Convert to a void return.

Related change: https://github.com/dotnet/runtime/pull/93834

cc @EgorBo @clamp03 @t-mustafin

### DIFF
--- a/src/tests/JIT/opt/RedundantBranch/RedundantBranchUnsigned2.cs
+++ b/src/tests/JIT/opt/RedundantBranch/RedundantBranchUnsigned2.cs
@@ -9,7 +9,7 @@ using Xunit;
 public class RedundantBranchUnsigned2
 {
     [Fact]
-    public static int TestEntryPoint()
+    public static void TestEntryPoint()
     {
         int[] arr1 = new int[2];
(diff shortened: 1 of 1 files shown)


gbalykov:
@EgorBo please take a look, without this change clr tests build on main fails

EgorBo (approved):
Thanks!

jakobbotsch:
@markples Is it expected that CI doesn't fail for something like this?

markples:
> Is it expected that CI doesn't fail for something like this?

I would have expected it to fail, and I think I've seen them in CI before, so I'm not sure what is different here.  My guess is that it has to two with the two-pass (anycpu vs per-target) nature of the build in CI vs local builds.  I don't see anything obvious (to me anyway) in the history for src\tests\Directory.Build.targets or in XUnitWrapperGenerator.  It might also be related to XUnitWrapperGenerator being (ab)used to do some checks in addition to generating the wrappers; Jeremy suggested splitting that out.

@trylek @jkoritzinsky @ivdiazsa any ideas here?

PR #81251, opened 2023-01-27
Reduce the size of HttpRequestMessage and HttpRequestHeaders

MihaZupan:
Saves 8 bytes on `HttpRequestMessage` and 16 on `HttpRequestHeaders`.
This represents about ~1% of allocated bytes for a simple YARP request.

PR #94151, opened 2023-10-29
Re-use status flag inside Ping

Vannevelj:
Similar idea as https://github.com/dotnet/runtime/pull/81251 and a couple of other examples described in [the blog post](https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-8/#http). Instead of storing these flags in separate booleans, re-use the existing mechanism that stores them as separate values for the existing integer. 

I expect this to result in a smaller `Ping` object, presumably 16 bytes due to padding, but I'm not confident about that.

I created a small benchmark in dotnet/performance (locally):

```cs
[BenchmarkCategory(Categories.Libraries, Categories.NoWASM)]
public class PingTests
{
    [Benchmark]
    public void Ping() => new Ping();
}
```

Which gives me the following results (top is `main`, bottom is PR):

| Method | Job        | Toolchain                                                                                                        | Mean     | Error    | StdDev   | Median   | Min      | Max      | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|------- |----------- |----------------------------------------------------------------------------------------------------------------- |---------:|---------:|---------:|---------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
| Ping   | Job-EIVJMB | \dotnet\runtime\artifacts\bin\testhost\net9.0-windows-Release-x64\shared\Microsoft.NETCore.App\9.0.0\CoreRun.exe | 57.24 ns | 1.697 ns | 1.954 ns | 57.16 ns | 54.48 ns | 60.11 ns |  1.04 |    0.03 | 0.0218 |     184 B |        1.00 |
| Ping   | Job-ATHUQO | \runtime\artifacts\bin\testhost\net9.0-windows-Release-x64\shared\Microsoft.NETCore.App\9.0.0\CoreRun.exe        | 55.29 ns | 1.128 ns | 1.298 ns | 55.28 ns | 53.62 ns | 58.42 ns |  1.00 |    0.00 | 0.0218 |     184 B |        1.00 |

A small improvement in time but no change in allocation so I'd be interested to hear takes on why I'm not seeing the expected reduction in allocation.

### DIFF
--- a/src/libraries/System.Net.Ping/src/System/Net/NetworkInformation/Ping.PingUtility.cs
+++ b/src/libraries/System.Net.Ping/src/System/Net/NetworkInformation/Ping.PingUtility.cs
@@ -92,7 +92,7 @@ namespace System.Net.NetworkInformation
                 {
                     pingProcess.Kill();
                 }
-                if (_status == Cancelled)
+                if (_status == Canceled)
                 {
                     throw;
                 }
--- a/src/libraries/System.Net.Ping/src/System/Net/NetworkInformation/Ping.PingUtility.cs
+++ b/src/libraries/System.Net.Ping/src/System/Net/NetworkInformation/Ping.PingUtility.cs
@@ -92,7 +92,7 @@ namespace System.Net.NetworkInformation
                 {
                     pingProcess.Kill();
                 }
-                if (_canceled)
+                if (_status == Cancelled)
                 {
                     throw;
                 }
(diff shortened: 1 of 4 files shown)


Vannevelj:
I've done some digging into this and understand object layouts better now. I don't believe this change will yield any significant results so I'll close the PR. I've written up my newfound knowledge at https://vannevel.net/posts/exploring-object-layouts

danmoseley:
Nice write up @Vannevelj. Perhaps you'd like to look at a different help wanted issue?

PR #94159, opened 2023-10-30
reduce handle table lock contention

Maoni0:
when a handle is created it's using the home heap GC set for this thread but this home heap is for GC heap allocation. a thread could be allocating a lot of handles but hardly any allocation so multiple threads could contend on the same handle table which does not handle this situation well (it causes long suspension). also when there are way fewer heaps than number of procs, this means we are only allocating on way fewer handle tables.
 
this change moves the balancing of the handle allocations into the handle table code itself. I've repurposed the alloc_count field to include the handle table index and switch to a different table after the thread has allocated 15 handles. this uses all handle tables.

### DIFF
--- a/src/coreclr/gc/objecthandle.h
+++ b/src/coreclr/gc/objecthandle.h
@@ -71,6 +71,10 @@ uint32_t     CompareExchangeVariableHandleType(OBJECTHANDLE handle, uint32_t old
  */
 
 int GetCurrentThreadHomeHeapNumber();
+gc_alloc_context* GetCurrentThreadAllocContext();
+
+// helper for getting the total number of processors
+int getNumberOfSlots();
 
 /*
  * Table maintenance routines
--- a/src/coreclr/gc/objecthandle.h
+++ b/src/coreclr/gc/objecthandle.h
@@ -73,7 +73,7 @@ uint32_t     CompareExchangeVariableHandleType(OBJECTHANDLE handle, uint32_t old
 int GetCurrentThreadHomeHeapNumber();
 gc_alloc_context* GetCurrentThreadAllocContext();
 
-// helper for getting the total number of processors
+// helper for getting the total number of handle table slots we can allocate in
 int getNumberOfSlots();
 
 /*
--- a/src/coreclr/gc/gc.cpp
+++ b/src/coreclr/gc/gc.cpp
@@ -19072,9 +19072,9 @@ allocation_state gc_heap::try_allocate_more_space (alloc_context* acontext, size
 #ifdef MULTIPLE_HEAPS
 void gc_heap::balance_heaps (alloc_context* acontext)
 {
-    if (acontext->alloc_count < 4)
+    if (acontext->get_alloc_count() < 4)
     {
-        if (acontext->alloc_count == 0)
+        if (acontext->get_alloc_count() == 0)
         {
             int home_hp_num = heap_select::select_heap (acontext);
             acontext->set_home_heap (GCHeap::GetHeap (home_hp_num));
@@ -19123,12 +19123,12 @@ void gc_heap::balance_heaps (alloc_context* acontext)
 #endif //HEAP_BALANCE_INSTRUMENTATION
                 set_home_heap = TRUE;
             }
-            else if ((acontext->alloc_count & 15) == 0)
+            else if ((acontext->get_alloc_count() & 15) == 0)
                 set_home_heap = TRUE;
         }
         else
         {
-            if ((acontext->alloc_count & 3) == 0)
+            if ((acontext->get_alloc_count() & 3) == 0)
                 set_home_heap = TRUE;
         }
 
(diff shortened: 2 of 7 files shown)


src/coreclr/gc/objecthandle.h
```diff
@@ -71,6 +71,10 @@ uint32_t     CompareExchangeVariableHandleType(OBJECTHANDLE handle, uint32_t old
  */
 
 int GetCurrentThreadHomeHeapNumber();
+gc_alloc_context* GetCurrentThreadAllocContext();
+
+// helper for getting the total number of processors
```
markples, line 76: just a documentation nit here - is it intended that this will always be number of processors, or could it change in the future?  (in other words, what's the abstraction here - slots or processors - if the former, is the comment over-specifying the behavior, or if the latter, should the function be renamed?)

markples (approved):

src/coreclr/gc/objecthandle.h
Maoni0: good point - I will revise this to say it's the # of table slots (which for server GC happens to be the # of processors) we allocate in

BrennanConroy:
I noticed two impacts on E2E benchmark latency numbers from this change. Verified by using a runtime right before this change and one right after this change.

Intel Linux 28 core Connection Close benchmark: `.74ms` latency -> `.51ms` latency 👍 
<details>
<summary>crank commands</summary>

Before this PR:
`crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/plaintext.benchmarks.yml --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/build/azure.profile.yml --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/steadystate.profile.yml --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/build/ci.profile.yml --scenario connectionclose --profile intel-lin-app --profile intel-load2-load --variable connections=32 --application.framework net9.0 --application.collectDependencies true --application.options.collectCounters true --application.aspNetCoreVersion 9.0.0-alpha.1.23553.5 --application.runtimeVersion 9.0.0-alpha.1.23553.7 --application.sdkVersion 9.0.100-alpha.1.23553.1`

After this PR:
`crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/plaintext.benchmarks.yml --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/build/azure.profile.yml --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/steadystate.profile.yml --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/build/ci.profile.yml --scenario connectionclose --profile intel-lin-app --profile intel-load2-load --variable connections=32 --application.framework net9.0 --application.collectDependencies true --application.options.collectCounters true --application.aspNetCoreVersion 9.0.0-alpha.1.23553.5 --application.runtimeVersion 9.0.0-alpha.1.23553.8 --application.sdkVersion 9.0.100-alpha.1.23553.1`

</details>

Intel Windows 28 core JsonHttpSys: `.5ms` latency -> `.6ms` latency 👎 
<details>
<summary>crank commands</summary>

Before this PR:
`crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/json.benchmarks.yml --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/build/azure.profile.yml --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/steadystate.profile.yml --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/build/ci.profile.yml --scenario json --profile intel-win-app --profile intel-load2-load --variable server=HttpSys --application.options.requiredOperatingSystem windows --application.framework net9.0 --application.collectDependencies true --application.options.collectCounters true --application.aspNetCoreVersion 9.0.0-alpha.1.23553.4 --application.runtimeVersion 9.0.0-alpha.1.23553.7 --application.sdkVersion 9.0.100-alpha.1.23553.1`

After this PR:
`crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/json.benchmarks.yml --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/build/azure.profile.yml --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/steadystate.profile.yml --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/build/ci.profile.yml --scenario json --profile intel-win-app --profile intel-load2-load --variable server=HttpSys --application.options.requiredOperatingSystem windows --application.framework net9.0 --application.collectDependencies true --application.options.collectCounters true --application.aspNetCoreVersion 9.0.0-alpha.1.23553.4 --application.runtimeVersion 9.0.0-alpha.1.23553.8 --application.sdkVersion 9.0.100-alpha.1.23553.1`

mangod9:
Does the JsonHttpSys latency regression show up as reduced throughput as well? I have seen the latency numbers have high variance?

BrennanConroy:
Surprisingly the RPS didn't have a big hit, just eyeballing it, it does look like maybe the average RPS is down a little bit.
The latency numbers don't have a high variance. It is quite obvious that the latency is permanently affected.
latency

Maoni0:
RPS should regress exactly the same amount as avg latency, shouldn't it? 

this change is doing more work for handle creation - this is unavoidable if we want to balance the handle creation. it's surprising to see the avg latency regressed this much though. I'll take a look.

PR #94160, opened 2023-10-30
Safib changes

KoltIP:
The proposed changes concern the 12th iteration, created for the v6.0.21 runtime tag. Problems were found using Svace


MichalStrehovsky:
Thanks but we don't accept pull requests against the .NET 6 version. Fixes get backported from `main` if they meet the bar.

Please open a pull request with changes against the `main` branch, as per https://github.com/dotnet/runtime/pull/94089#issuecomment-1783041080.

PR #94166, opened 2023-10-30
[wasm] Fix AOT publish in paths with space on Windows

ilonatommy:
Connected to investigation: https://github.com/dotnet/runtime/pull/92399.

This is only fix for Windows, so we could not change the `TmpPath` for other platforms yet.
1) When `WasmDedup=true` we had `error : Can not open image C:\repos\wbt` because the path `C:\repos\wbt artifacts` was not put into quotations. This was a problem for all platforms and is fixed everywhere.
2) `_AOTObjectFile` items should not only be separated by space but also put into quotations, otherwise a spaced path will be treated as 2 separate items.
3) Tests that are using the path to the wbt test project were edited.

### DIFF
--- a/src/mono/wasm/Wasm.Build.Tests/Common/BuildEnvironment.cs
+++ b/src/mono/wasm/Wasm.Build.Tests/Common/BuildEnvironment.cs
@@ -32,7 +32,10 @@ namespace Wasm.Build.Tests
         public static readonly string           RelativeTestAssetsPath = @"..\testassets\";
         public static readonly string           TestAssetsPath = Path.Combine(AppContext.BaseDirectory, "testassets");
         public static readonly string           TestDataPath = Path.Combine(AppContext.BaseDirectory, "data");
-        public static readonly string           TmpPath = Path.Combine(AppContext.BaseDirectory, "wbt");
+        // ActiveIssue for Linux/OSx: https://github.com/dotnet/runtime/issues/92335
+        public static readonly string           TmpPath = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
+                                                    Path.Combine(AppContext.BaseDirectory, "wbt artifacts") :
+                                                    Path.Combine(AppContext.BaseDirectory, "wbt");
 
         public static readonly string           DefaultRuntimeIdentifier =
 #if TARGET_WASI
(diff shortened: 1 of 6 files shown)


maraf (approved):
Looks good to me 👍

src/mono/wasm/Wasm.Build.Tests/Common/BuildEnvironment.cs
```diff
@@ -32,7 +32,10 @@ public class BuildEnvironment
         public static readonly string           RelativeTestAssetsPath = @"..\testassets\";
         public static readonly string           TestAssetsPath = Path.Combine(AppContext.BaseDirectory, "testassets");
         public static readonly string           TestDataPath = Path.Combine(AppContext.BaseDirectory, "data");
-        public static readonly string           TmpPath = Path.Combine(AppContext.BaseDirectory, "wbt");
+        // ActiveIssue for Linux/OSx: https://github.com/dotnet/runtime/issues/92335
+        public static readonly string           TmpPath = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
+                                                    Path.Combine(AppContext.BaseDirectory, "wbt artifacts") :
```
radekdoulik, line 37: why do we use this path only on windows? does it currently break on the other platforms?

radekdoulik (approved):

src/tasks/AotCompilerTask/MonoAOTCompiler.cs
```diff
@@ -951,7 +951,7 @@ private PrecompileArguments GetPrecompileArgumentsFor(ITaskItem assemblyItem, st
         if (isDedup)
         {
             foreach (var aItem in _assembliesToCompile!)
-                processArgs.Add(aItem.ItemSpec);
+                processArgs.Add('"' + aItem.ItemSpec + '"');
```
radical, line 954: nit:
```suggestion
                processArgs.Add($"'{aItem.ItemSpec}'");
```

ilonatommy:
> What is the remaining issue on linux/macos?

That while a sample repro for them work, when the fix is applied to the repo, it fails. Please, see https://github.com/dotnet/runtime/pull/92399#issuecomment-1782955184 and the investigation PR for reference.

src/mono/wasm/Wasm.Build.Tests/Common/BuildEnvironment.cs
ilonatommy: Yes, I am still trying to find a solution for OSX, Linux.

radical (approved):

PR #94180, opened 2023-10-30
[browser][MT] use portable timer

pavelsavara:
`TimerQueue` ST implementation relies on `setTimeout` (which deadlocks in MT)
- we are moving `TimerQueue` from MT+ST to ST only. 
- After this change, MT will use portable timer (same as any Unix), which will create dedicated thread.


kg:
This looks like timers will only ever run on the main thread, is that a problem? I don't recall whether timers work that way on other platforms

kg (approved):

src/mono/System.Private.CoreLib/System.Private.CoreLib.csproj
```diff
@@ -288,6 +285,7 @@
       <Compile Include="$(BclSourcesRoot)\System\Threading\ThreadPoolBoundHandle.Browser.Mono.cs" />
       <Compile Include="$(BclSourcesRoot)\System\Threading\PreAllocatedOverlapped.Browser.Mono.cs" />
       <Compile Include="$(BclSourcesRoot)\System\Threading\ThreadPool.Browser.Mono.cs" />
+      <Compile Include="$(BclSourcesRoot)\System\Threading\TimerQueue.Browser.Mono.cs" />
```
ilonatommy, line 288: Why are we adding it when `FeatureWasmThreads=false`? Aren't we making changes for MT-only and non-MT should stay as it was?

ilonatommy (approved):

pavelsavara:
> This looks like timers will only ever run on the main thread, is that a problem? I don't recall whether timers work that way on other platforms

We discussed this elsewhere, portable timer in MT will create dedicated thread. ST will stay using `setTimeout`

PR #94185, opened 2023-10-30
JIT: make global morph its own phase

AndyAyersMS:
Contributes to #93246.

### DIFF
--- a/src/coreclr/jit/morph.cpp
+++ b/src/coreclr/jit/morph.cpp
@@ -13619,13 +13619,12 @@ void Compiler::fgMorphStmtBlockOps(BasicBlock* block, Statement* stmt)
     }
 }
 
-/*****************************************************************************
- *
- *  Morph the statements of the given block.
- *  This function should be called just once for a block. Use fgMorphBlockStmt()
- *  for reentrant calls.
- */
-
+//------------------------------------------------------------------------
+// fgMorphStmts: Morph all statements in a block
+//
+// Arguments:
+//    block - block in question
+//
 void Compiler::fgMorphStmts(BasicBlock* block)
 {
     fgRemoveRestOfBlock = false;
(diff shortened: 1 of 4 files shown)


AndyAyersMS:
PTAL @dotnet/jit-contrib 

Once RPO is in place, the block traversal order will change, so split out the per-block action as its own method.

No diffs. Has bugged me forever that there was no default IR dump after morph. Now there is.

AndyAyersMS:
No [diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=455655&view=ms.vss-build-web.run-extensions-tab), errors known.

src/coreclr/jit/morph.cpp
```diff
@@ -13862,53 +13890,16 @@ void Compiler::fgMorphBlocks()
...
-
-    BasicBlock* block = fgFirstBB;
-    noway_assert(block);
-
-    do
+    // Morph all blocks.
+    //
+    // Note morph can add blocks downstream from the current block,
+    // and alter (but not null out) the current block's bbNext;
+    // this iterator ensures they all get visited.
```
BruceForstall, line 13897: Doesn't the iterator load the bbNext before the iteration of the loop? If the callee can adjust the blocks, maybe we shouldn't use the for-based iterator here?

BruceForstall (approved):

src/coreclr/jit/morph.cpp
AndyAyersMS: It doesn't seem to cache anything...? Also that comment is a bit stale; with other changes I've made recently morph won't add blocks or (I suspect) alter `bbNext` anymore.

BruceForstall: Maybe it only caches the end-point of the iteration. But still, any change to the block list seems risky to depend on what the C++ compiler is going to do.

AndyAyersMS: For the `BasicBlockRangeList` form sure... but this is going via `BasicBlockSimpleList` which does't cache an endpoint.

AndyAyersMS: @BruceForstall still ok with merging? Or would you rather I change the iteration back?

BruceForstall: Yes, it's fine.

PR #94198, opened 2023-10-31
Changes related to type casting errors, as well as cosmetic fixes

KoltIP:


src/libraries/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/CSharpIsEventBinder.cs
```diff
@@ -22,7 +22,17 @@ public Expr DispatchPayload(RuntimeBinder runtimeBinder, ArgumentObject[] argume
...
+        {
+            var value = arguments[0].Value as Type;
+            if (arguments[0].Info.IsStaticType)
+            {
+                if (value == null)
+                    throw new Exception("invalid cast arguments[0].Value to Type");
+                SymbolTable.PopulateSymbolTableWithName(Name, null, value);
+            }
+            else
+                SymbolTable.PopulateSymbolTableWithName(Name, null, arguments[0].Type);
```
huoyaoyuan, line 34: Can you demonstrate in which condition the invalid cast branch will be hit?
If an invalid cast exception should be thrown, changing `as` to `(Type)arguments[0].Value` should be better.

src/libraries/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/Semantics/Operators.cs
```diff
@@ -2684,6 +2684,9 @@ private static AggregateType GetEnumBinOpType(ExpressionKind ek, CType argType1,
             AggregateType type1 = argType1 as AggregateType;
             AggregateType type2 = argType2 as AggregateType;
 
+            if (type1 == null)
+                throw new Exception("invalid cast argType1 as AggregateType");
+
```
huoyaoyuan, line 2689: Same here, when will this be hit?

src/libraries/System.Data.Odbc/src/Common/System/Data/Common/DBConnectionString.cs
```diff
@@ -308,12 +308,7 @@ private static void ValidateCombinedSet(DBConnectionString? componentSet, DBConn
                         // Combined - Component == null
                         Debug.Assert(!combinedSet._restrictionValues.Except(componentSet._restrictionValues).Any(), "Combined set allows values not allowed by component set");
                     }
-                    else if (combinedSet._behavior == KeyRestrictionBehavior.PreventUsage)
-                    {
-                        // Component==Allow, Combined==PreventUsage
-                        // Preventions override allows, so there is nothing to check here
-                    }
-                    else
+                    else if (combinedSet._behavior != KeyRestrictionBehavior.PreventUsage)
```
huoyaoyuan, line 311: Different types of changes shouldn't be mixed in one pull request, especially for different areas.

And also, the comment for no-op is valuable and should be preserved in some way.

src/libraries/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/CSharpIsEventBinder.cs
KoltIP: We can't analyze the entire chain of calls, but this code refers to a variable resulting from a conversion without first checking whether the conversion was successful

src/libraries/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/Semantics/Operators.cs
KoltIP: We can't analyze the entire chain of calls, but this code refers to a variable resulting from a conversion without first checking whether the conversion was successful

src/libraries/System.Data.Odbc/src/Common/System/Data/Common/DBConnectionString.cs
KoltIP: We believe that the empty condition body is misleading and can be replaced with a clearer option.

KoltIP: We believe that the empty condition body is misleading and can be replaced with a clearer option.

huoyaoyuan: It should not be misleading with comment describing why no action should be taken.

src/libraries/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/CSharpIsEventBinder.cs
huoyaoyuan: There can be conventions for states of `ArgumentObject`. The better approach should be expressing the convention more in code.

huoyaoyuan: In addition, such check can be covered by nullable reference type analysis, since cast result of `as` is nullable.

Currently NRT warnings are disabled for `Microsoft.CSharp`. /cc @stephentoub Should we take effort to enable it?

stephentoub:
What is this PR trying to address?  There's no PR description.

Per https://github.com/dotnet/runtime/tree/f2e4ddf20936b37445bbbb2d139e8080ca00f16c/src/libraries/Microsoft.CSharp:
> We are not accepting feature contributions to Microsoft.CSharp. The library is effectively archived. The library and supporting language features are mature and no longer evolving, and the risk of code change likely exceeds the benefit. We will consider changes that address significant bugs or regressions, or changes that are necessary to continue shipping the binaries. Other changes will be rejected.

danmoseley:
I'm closing based on unanswered question above.

PR #94204, opened 2023-10-31
[browser][MT] unify pthread pool size to 40

pavelsavara:
- bump default thread pool size to 40 and unify to that default for all tests
 - limit managed thread pool size to 10
 - see which existing active issues are caused by the same reason
 - keep ActiveIssue tracking it, because we know that it's too large default
     - https://github.com/dotnet/runtime/issues/91538 
     - until we have solution to allocate threads dynamically


eng/pipelines/extra-platforms/runtime-extra-platforms-wasm.yml
```diff
@@ -112,14 +112,14 @@ jobs:
...
+      extraBuildArgs: /p:MonoWasmBuildVariant=multithread /p:WasmEnableThreads=true /p:AotHostArchitecture=x64 /p:AotHostOS=$(_hostedOS)
       isExtraPlatformsBuild: ${{ parameters.isExtraPlatformsBuild }}
       isWasmOnlyBuild: ${{ parameters.isWasmOnlyBuild }}
       # Always run for runtime-wasm because tests are not run in runtime
       alwaysRun: ${{ parameters.isWasmOnlyBuild }}
 
-       # NOTE - Since threading is experimental, we don't want to block mainline work
-      shouldContinueOnError: true
+      # NOTE - Since threading is experimental, we don't want to block mainline work
+      # TODO put back shouldContinueOnError: true
```
pavelsavara, line 122: todo

src/mono/wasm/debugger/tests/debugger-test/debugger-test.csproj
```diff
@@ -8,7 +8,6 @@
     <WasmGenerateAppBundle>true</WasmGenerateAppBundle>
     <OutputType>library</OutputType>
     <WasmEmitSymbolMap>true</WasmEmitSymbolMap>
-    <_WasmPThreadPoolSize Condition="'$(MonoWasmBuildVariant)' == 'multithread'">10</_WasmPThreadPoolSize>
```
ilonatommy, line 11: This caused failures in MT debugger tests., we should not merge without blocking them.

ilonatommy (approved):

kg:
Why 40 if the managed limit is 10? 40 is a lot

pavelsavara:
> Why 40 if the managed limit is 10? 40 is a lot

- Because there are tests which also allocate non-thread-pool managed threads.
- Possibly something is still leaking threads. 
- I know that managed portable  thread pool would release pthread only after some delay.
- I also know that mono will only release attached pthread on next attempt to allocate another.
But there could be browser loop run that emscripten need to have before it could re-use the released pthread.

But I don't claim this is a solution, this is a way how to see more unit tests running.

src/mono/wasm/debugger/tests/debugger-test/debugger-test.csproj
pavelsavara: My experience is that they are broken/fragile.
I set the `shouldContinueOnError: true` for `DebuggerTests_MultiThreaded` so that it would not fail the build, it would rather only become orange in AzDo.

cc @thaystg

pavelsavara:
MT lib tests pass with this change
image

MT debugger is silenced
image

All other CI issues are known issues.

PR #94208, opened 2023-10-31
Tests build fix with BuildAsStandalone

gbalykov:
Without this patch build of these tests fails with "No entry point declared for executable" with BuildAsStandalone.
Before #91560 they were built as OutputType=Library and these are actually not launched during testing, but are used as libs.

cc @trylek @t-mustafin @clamp03


gbalykov:
@trylek please take a look, without this change clr tests build with BuildAsStandalone on main fails

trylek (approved):
Great, thanks for fixing!

clamp03 (approved):

gbalykov:
@trylek can this be merged?

gbalykov:
@trylek Thanks!

PR #94211, opened 2023-10-31
Additional QuicConnectionOptions

### LINKED ISSUE #70090: [API Proposal]: Quic connection keep alive
### Background and motivation

An application protocol such as HTTP/3 needs to enable the transport keep alive to defer the Quic idle timeout (https://quicwg.org/base-drafts/draft-ietf-quic-http.html#name-idle-connections). It should also be able to do so only if there's pending requests or server pushes.

From looking at the msquic implementation, the keep alive interval can be updated on an active Quic connection. So this could be set through a new `QuicConnection` property.

Such a feature will also probably needed to implement keep alive in `System.Net.Http.SocketsHttpHandler.Http3Connection`.

### API Proposal

```csharp
namespace System.Net.Quic;

public sealed class QuicConnection : IDisposable
{
    // <summary>The keep alive interval to specify how often to send PING frames.</summary>
    public TimeSpan KeepAliveInterval { get; set; }
}
```

### API Usage

```csharp
// Enable keep alive
connection.KeepAliveInterval = TimeSpan.FromSeconds(15);

// Disable keep alive
connection.KeepAliveInterval = TimeSpan.MaxValue;
```

### Alternative Designs

_No response_

### Risks

_No response_

### LINKED ISSUE #72984: [API Proposal]: Additional QuicConnectionOptions
### Background and motivation

msquic has a lot of connection options and S.N.Q exposes just a few

List of msquic settings - <https://github.com/microsoft/msquic/blob/main/docs/Settings.md>

Settings I think should be added:

* ~~Disconnect Timeout - Importance discovered as part of <https://github.com/dotnet/runtime/issues/71927>. People may want to be more or less aggressive with closing unresponsive streams~~
  * ncl: we decided against exposing, as this is too MsQuic specific, see discussion in <https://github.com/dotnet/runtime/issues/72984#issuecomment-1749160603>
* Keep Alive Interval - Kestrel HTTP/2 layer has a similar option. Should expose it for feature parity.
* Flow Control Window - Kestrel HTTP/2 layer has a similar option. Should expose it for feature parity. Useful when sending large amounts of data over connection with latency. Allows you to trade potentially higher memory usage to avoid back pressure throttling the stream/connection
* Stream Receive Window - Kestrel HTTP/2 layer has a similar option. Same reason as connection flow control window.
  * ncl: QUIC exposes limits for individual stream types, so we will expose separate properties for those
* ~~DatagramReceiveEnabled - Maybe. I think this will be needed for WebTransport datagram support. Does QuicStream have a story for supporting datagram in the future?~~
  * ncl: This needs separate proposal because it is more complicated: <https://github.com/dotnet/runtime/issues/53533>
* Handshake Idle Timeout - Kestrel has an option to supply a handshake timeout. Would be nice to flow that through to QUIC.

I don't think any of these are necessary for .NET 7.

### API Proposal (edited by @rzikm)

See also discussion about naming below the proposal, we are open for suggestions for better names. Default values were copied from current MsQuic behavior.

```diff
namespace System.Net.Quic
{
    public abstract class QuicConnectionOptions
    {
        /// <summary>
        /// The interval at which keep alive packets are sent on the connection.
        /// Default <see cref="TimeSpan.Zero"/>, or <see cref="Timeout.InfiniteTimeSpan"> means no keep alive.
        /// </summary>
+       public TimeSpan KeepAliveInterval { get; set; } = TimeSpan.Zero;

        /// <summary>
        /// The initial flow-control window size for the connection.
        /// </summary>
+       public int InitialConnectionReceiveWindowSize { get; set; } = 16 * 1024 * 1024;

        /// <summary>
        /// The initial flow-control window size for locally initiated bidirectional streams.
        /// </summary>
+       public int InitialLocallyInitiatedBidirectionalStreamReceiveWindowSize { get; set; } = 64 * 1024;

        /// <summary>
        /// The initial flow-control window size for remotely initiated bidirectional streams.
        /// </summary>
+       public int InitialRemotelyInitiatedBidirectionalStreamReceiveWindowSize { get; set; } = 64 * 1024;

        /// <summary>
        /// The initial flow-control window size for (remotely initiated) unidirectional streams.
        /// </summary>
+       public int InitialUnidirectionalStreamReceiveWindowSize { get; set; } = 64 * 1024;

        /// <summary>
        /// The upper bound on time when the handshake must complete. If the handshake does not
        /// complete in this time, the connection is aborted.
        /// </summary>
+       public TimeSpan HandshakeTimeout { get; set; } = TimeSpan.FromSeconds(10);
    }
}
```

Properties `InitialConnectionReceiveWindowSize`, `InitialLocallyInitiatedBidirectionalStreamReceiveWindowSize`, `InitialRemotelyInitiatedBidirectionalStreamReceiveWindowSize`, and `InitialUnidirectionalStreamReceiveWindowSize` have rather long names but we did come up with a better alternative. They map directly to QUIC transport parameters `initial_max_data`, `initial_max_stream_data_bidi_local`, `initial_max_stream_data_bidi_remote`, and `initial_max_stream_data_uni` respectively. These transport parameters limit the amount of data the peer can send on the connection as a whole and on streams of individual types to prevent the peer from overwhelming the receiver with data. Thus, the `Initial*ReceiveWindowSize` part of the name seems appropriate. The rest of the name identifies the type of stream, where "Locally initiated bidirectional stream" and similar are terms used in the QUIC RFC and we should use them to avoid confusion.

For comparison, MsQuic uses names such as `StreamRecvWindowBidiLocalDefault`.
And will expand the receive window if appropriate.

<details>
<summary>Parameter definition from RFC</summary>

#### initial_max_data (0x04)

The initial maximum data parameter is an integer value that contains the
initial value for the maximum amount of data that can be sent on the
connection. This is equivalent to sending a MAX_DATA ([Section
19.9](https://datatracker.ietf.org/doc/html/rfc9000#frame-max-data)) for the
connection immediately after completing the handshake.

#### initial_max_stream_data_bidi_local (0x05)

This parameter is an integer value specifying the initial flow control limit
for locally initiated bidirectional streams. This limit applies to newly
created bidirectional streams opened by the endpoint that sends the
transport parameter. In client transport parameters, this applies to streams
with an identifier with the least significant two bits set to 0x00; in
server transport parameters, this applies to streams with the least
significant two bits set to 0x01.

#### initial_max_stream_data_bidi_remote (0x06)

This parameter is an integer value specifying the initial flow control limit
for peer-initiated bidirectional streams. This limit applies to newly
created bidirectional streams opened by the endpoint that receives the
transport parameter. In client transport parameters, this applies to streams
with an identifier with the least significant two bits set to 0x01; in
server transport parameters, this applies to streams with the least
significant two bits set to 0x00.

#### initial_max_stream_data_uni (0x07)

This parameter is an integer value specifying the initial flow control limit
for unidirectional streams. This limit applies to newly created
unidirectional streams opened by the endpoint that receives the transport
parameter. In client transport parameters, this applies to streams with an
identifier with the least significant two bits set to 0x03; in server
transport parameters, this applies to streams with the least significant two
bits set to 0x02.

</details>

### API Usage

```csharp
var quicListenerOptions = new QuicListenerOptions
{
    ApplicationProtocols = _tlsConnectionCallbackOptions.ApplicationProtocols,
    ListenEndPoint = listenEndPoint,
    ListenBacklog = options.Backlog,
    ConnectionOptionsCallback = async (connection, helloInfo, cancellationToken) =>
    {
        var connectionOptions = new QuicServerConnectionOptions
        {
            // ... set various properties for server connection ...
        };
        return connectionOptions;
    }
};
```

### Alternative Designs

To avoid long property names. We can introduce a wrapper type for the initial
connection window sizes. However, the wrapper type must be a class, otherwise
`options.InitialReceiveWindowSizes.ConnectionTotal = x` has no effect. We also
don't see another use for the type.

```diff
namespace System.Net.Quic
{
+   public class QuicReceiveWindowSizes
    {
        /// <summary>
        /// The flow-control window size for the entire connection.
        /// </summary>
+       public int ConnectionTotal { get; set; } = 16 * 1024 * 1024;

        /// <summary>
        /// The flow-control window size for locally initiated bidirectional streams.
        /// </summary>
+       public int LocallyInitiatedBidirectionalStreams { get; set; } = 64 * 1024;

        /// <summary>
        /// The flow-control window size for remotely initiated bidirectional streams.
        /// </summary>
+       public int RemotelyInitiatedBidirectionalStreams { get; set; } = 64 * 1024;

        /// <summary>
        /// The flow-control window size for (remotely initiated) unidirectional streams.
        /// </summary>
+       public int UnidirectionalStreams { get; set; } = 64 * 1024;

    }

    public abstract class QuicConnectionOptions
    {
        // unchanged
+       public TimeSpan KeepAliveInterval { get; set; } = TimeSpan.Zero;
+       public TimeSpan HandshakeTimeout { get; set; } = TimeSpan.FromSeconds(10);

        /// <summary>
        /// The initial connection window sizes.
        /// Default <see cref="TimeSpan.Zero"/>, or <see cref="Timeout.InfiniteTimeSpan"> means no keep alive.
        /// </summary>
        // note: actuall implementation can create the default instance lazily to avoid alloc if user only calls set
+       public QuicReceiveWindowSizes InitialReceiveWindowSizes { get; set; } = new QuicReceiveWindowSizes();
    }
}
```

### Risks

The options above are commonly exposed by other QUIC implementations and none of them are too specific as to vendor lock System.Net.Quic to MsQuic only.

rzikm:
Closes #72984.
Closes #70090.

Adds:
- KeepAliveInterval - sending ping frames to prevent the connection from idling out. (we may want to add the same property on QuicConnection itself so that users can turn pings on/off according to their needs)
- flow-control limits (max_data) - stream-level limits will have effect only with newer MsQuic having https://github.com/microsoft/msquic/pull/3948. 2.4+ should be it
- HandshakeTimeout - kills the connection if the handshake is not established in the given amount of time
  -  ServerOptionsCallback still has 10s timeout as before, once the callback returns, the timeout value from the options is applied. 
  The client times out as well.
  - Timeouts lead to peer receiving UserCanceled TLS Alert which is weird but is not against the spec, although getting CONNECTION_CLOSE with APPLICATION_ERROR would be more expected.
 
Other improvements:
- if the client terminates the connection during the handshake, CancellationToken passed to ServerOptionsCallback fires as well.
- Fix property names in ArgumentExceptions thrown from connection options validation (previously it put the message instead of the parameter name)

### DIFF
--- a/src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicConfiguration.cs
+++ b/src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicConfiguration.cs
@@ -117,14 +117,48 @@ internal static class MsQuicConfiguration
 #pragma warning restore SYSLIB0040
 
         QUIC_SETTINGS settings = default(QUIC_SETTINGS);
+
         settings.IsSet.PeerUnidiStreamCount = 1;
         settings.PeerUnidiStreamCount = (ushort)options.MaxInboundUnidirectionalStreams;
+
         settings.IsSet.PeerBidiStreamCount = 1;
         settings.PeerBidiStreamCount = (ushort)options.MaxInboundBidirectionalStreams;
+
         if (options.IdleTimeout != TimeSpan.Zero)
         {
             settings.IsSet.IdleTimeoutMs = 1;
             settings.IdleTimeoutMs = options.IdleTimeout != Timeout.InfiniteTimeSpan ? (ulong)options.IdleTimeout.TotalMilliseconds : 0;
+        }
+
+        if (options.KeepAliveInterval != TimeSpan.Zero)
+        {
+            settings.IsSet.KeepAliveIntervalMs = 1;
+            settings.KeepAliveIntervalMs = (uint)options.KeepAliveInterval.TotalMilliseconds;
+        }
+
+        if (options.InitialConnectionWindowSize > 0)
+        {
+            settings.IsSet.ConnFlowControlWindow = 1;
+            settings.ConnFlowControlWindow = (uint)options.InitialConnectionWindowSize;
+        }
+
+        // todo: these values need all to be powers of 2
+        if (options.InitialLocallyInitiatedBidirectionalStreamReceiveWindowSize > 0)
+        {
+            settings.IsSet.StreamRecvWindowBidiLocalDefault = 1;
+            settings.StreamRecvWindowBidiLocalDefault = (uint)options.InitialLocallyInitiatedBidirectionalStreamReceiveWindowSize;
+        }
+
+        if (options.InitialRemotelyInitiatedBidirectionalStreamReceiveWindowSize > 0)
+        {
+            settings.IsSet.StreamRecvWindowBidiRemoteDefault = 1;
+            settings.StreamRecvWindowBidiRemoteDefault = (uint)options.InitialRemotelyInitiatedBidirectionalStreamReceiveWindowSize;
+        }
+
+        if (options.InitialUnidirectionalStreamReceiveWindowSize > 0)
+        {
+            settings.IsSet.StreamRecvWindowUnidiDefault = 1;
+            settings.StreamRecvWindowUnidiDefault = (uint)options.InitialUnidirectionalStreamReceiveWindowSize;
         }
 
         QUIC_HANDLE* handle;
--- a/src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicConfiguration.cs
+++ b/src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicConfiguration.cs
@@ -132,27 +132,33 @@ internal static class MsQuicConfiguration
                 : 0; // 0 disables the timeout
         }
 
-        settings.IsSet.KeepAliveIntervalMs = 1;
-        settings.KeepAliveIntervalMs = options.KeepAliveInterval != Timeout.InfiniteTimeSpan
-            ? (uint)options.KeepAliveInterval.TotalMilliseconds
-            : 0; // 0 disables the keepalive
+        if (options.KeepAliveInterval != TimeSpan.Zero)
+        {
+            settings.IsSet.KeepAliveIntervalMs = 1;
+            settings.KeepAliveIntervalMs = options.KeepAliveInterval != Timeout.InfiniteTimeSpan
+                ? (uint)options.KeepAliveInterval.TotalMilliseconds
+                : 0; // 0 disables the keepalive
+        }
 
         settings.IsSet.ConnFlowControlWindow = 1;
-        settings.ConnFlowControlWindow = (uint)options.InitialReceiveWindowSizes.Connection;
+        settings.ConnFlowControlWindow = (uint)(options._initialRecieveWindowSizes?.Connection ?? QuicDefaults.DefaultConnectionMaxData);
 
         settings.IsSet.StreamRecvWindowBidiLocalDefault = 1;
-        settings.StreamRecvWindowBidiLocalDefault = (uint)options.InitialReceiveWindowSizes.LocallyInitiatedBidirectionalStream;
+        settings.StreamRecvWindowBidiLocalDefault = (uint)(options._initialRecieveWindowSizes?.LocallyInitiatedBidirectionalStream ?? QuicDefaults.DefaultStreamMaxData);
 
         settings.IsSet.StreamRecvWindowBidiRemoteDefault = 1;
-        settings.StreamRecvWindowBidiRemoteDefault = (uint)options.InitialReceiveWindowSizes.RemotelyInitiatedBidirectionalStream;
+        settings.StreamRecvWindowBidiRemoteDefault = (uint)(options._initialRecieveWindowSizes?.RemotelyInitiatedBidirectionalStream ?? QuicDefaults.DefaultStreamMaxData);
 
         settings.IsSet.StreamRecvWindowUnidiDefault = 1;
-        settings.StreamRecvWindowUnidiDefault = (uint)options.InitialReceiveWindowSizes.UnidirectionalStream;
+        settings.StreamRecvWindowUnidiDefault = (uint)(options._initialRecieveWindowSizes?.UnidirectionalStream ?? QuicDefaults.DefaultStreamMaxData);
 
-        settings.IsSet.HandshakeIdleTimeoutMs = 1;
-        settings.HandshakeIdleTimeoutMs = options.HandshakeTimeout != Timeout.InfiniteTimeSpan
-                ? (ulong)options.HandshakeTimeout.TotalMilliseconds
-                : 0; // 0 disables the timeout
+        if (options.HandshakeTimeout != TimeSpan.Zero)
+        {
+            settings.IsSet.HandshakeIdleTimeoutMs = 1;
+            settings.HandshakeIdleTimeoutMs = options.HandshakeTimeout != Timeout.InfiniteTimeSpan
+                    ? (ulong)options.HandshakeTimeout.TotalMilliseconds
+                    : 0; // 0 disables the timeout
+        }
 
         QUIC_HANDLE* handle;
 
--- a/src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicConfiguration.cs
+++ b/src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicConfiguration.cs
@@ -124,17 +124,18 @@ internal static class MsQuicConfiguration
         settings.IsSet.PeerBidiStreamCount = 1;
         settings.PeerBidiStreamCount = (ushort)options.MaxInboundBidirectionalStreams;
 
-        if (options.IdleTimeout != Timeout.InfiniteTimeSpan)
+        if (options.IdleTimeout != TimeSpan.Zero)
         {
             settings.IsSet.IdleTimeoutMs = 1;
-            settings.IdleTimeoutMs = (uint)options.IdleTimeout.TotalMilliseconds;
+            settings.IdleTimeoutMs = options.IdleTimeout != Timeout.InfiniteTimeSpan
+                ? (ulong)options.IdleTimeout.TotalMilliseconds
+                : 0; // 0 disables the timeout
         }
 
-        if (options.KeepAliveInterval != Timeout.InfiniteTimeSpan)
-        {
-            settings.IsSet.KeepAliveIntervalMs = 1;
-            settings.KeepAliveIntervalMs = (uint)options.KeepAliveInterval.TotalMilliseconds;
-        }
+        settings.IsSet.KeepAliveIntervalMs = 1;
+        settings.KeepAliveIntervalMs = options.KeepAliveInterval != Timeout.InfiniteTimeSpan
+            ? (uint)options.KeepAliveInterval.TotalMilliseconds
+            : 0; // 0 disables the keepalive
 
         settings.IsSet.ConnFlowControlWindow = 1;
         settings.ConnFlowControlWindow = (uint)options.InitialReceiveWindowSizes.Connection;
@@ -147,6 +148,11 @@ internal static class MsQuicConfiguration
 
         settings.IsSet.StreamRecvWindowUnidiDefault = 1;
         settings.StreamRecvWindowUnidiDefault = (uint)options.InitialReceiveWindowSizes.UnidirectionalStream;
+
+        settings.IsSet.HandshakeIdleTimeoutMs = 1;
+        settings.HandshakeIdleTimeoutMs = options.HandshakeTimeout != Timeout.InfiniteTimeSpan
+                ? (ulong)options.HandshakeTimeout.TotalMilliseconds
+                : 0; // 0 disables the timeout
 
         QUIC_HANDLE* handle;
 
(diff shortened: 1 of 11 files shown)


src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnection.cs
```diff
@@ -66,9 +66,21 @@ public static ValueTask<QuicConnection> ConnectAsync(QuicClientConnectionOptions
...
+            linkedCts.CancelAfter(options.HandshakeTimeout);
+
             try
             {
-                await connection.FinishConnectAsync(options, cancellationToken).ConfigureAwait(false);
+                await connection.FinishConnectAsync(options, linkedCts.Token).ConfigureAwait(false);
+            }
+            catch (OperationCanceledException) when (!cancellationToken.IsCancellationRequested)
+            {
+                // connect timeout elapsed, since handshake is not done yet, application error code
```
ManickaP, line 79: NIT: Handshake timeout, no? There's no connect timeout...

src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnectionOptions.cs
```diff
@@ -2,10 +2,53 @@
...
 
+/// <summary>
+/// Collection of receive window sizes for <see cref="QuicConnection"/> as a whole and individual <see cref="QuicStream"/> types.
+/// </summary>
+public sealed class QuicReceiveWindowSizes
+{
+    /// <summary>
+    /// The initial flow-control window size for the connection.
+    /// </summary>
+    public int Connection { get; set; } = 16 * 1024 * 1024;
```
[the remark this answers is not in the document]
ManickaP: Also we discussed whether these properties should be rather `long` as spec says variable int (up to 62 bits) and MsQuic allows uint (up to 4GB). But it seems like ridiculous range... WDYT?

ManickaP (commented):
I know it's a draft, but I might not have time to review this properly later.

src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnectionOptions.cs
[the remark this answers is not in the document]
rzikm: the setting says "I am willing to receive and buffer this much data at any given moment", it implies you have this large buffer ready for receiving stuff. I don't think values above `uint` range make sense for this. And furthermore, these are initial values, implementations are able to grow/shrink the window as necessary.

src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicConfiguration.cs
```diff
@@ -117,16 +117,43 @@ private static unsafe MsQuicSafeHandle Create(QuicConnectionOptions options, QUI
...
         {
             settings.IsSet.IdleTimeoutMs = 1;
-            settings.IdleTimeoutMs = options.IdleTimeout != Timeout.InfiniteTimeSpan ? (ulong)options.IdleTimeout.TotalMilliseconds : 0;
+            settings.IdleTimeoutMs = options.IdleTimeout != Timeout.InfiniteTimeSpan
+                ? (ulong)options.IdleTimeout.TotalMilliseconds
+                : 0; // 0 disables the timeout
         }
 
+        settings.IsSet.KeepAliveIntervalMs = 1;
+        settings.KeepAliveIntervalMs = options.KeepAliveInterval != Timeout.InfiniteTimeSpan
```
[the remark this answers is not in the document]
rzikm: We are defaulting to `InfinitetTimeSpan` in the settings, so that would come into effect only if user explicitly assigned `Zero`.

src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnectionOptions.cs
```diff
@@ -2,10 +2,53 @@
...
+    internal void Validate(string argumentName)
+    {
+        ValidatePowerOf2(argumentName, Connection);
+        ValidatePowerOf2(argumentName, LocallyInitiatedBidirectionalStream);
+        ValidatePowerOf2(argumentName, RemotelyInitiatedBidirectionalStream);
+        ValidatePowerOf2(argumentName, UnidirectionalStream);
+
+        static void ValidatePowerOf2(string argumentName, int value, [CallerArgumentExpression(nameof(value))] string? propertyName = null)
+        {
+            if (value <= 0 || ((value - 1) & value) != 0)
```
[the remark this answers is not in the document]
rzikm: The user has no way of increasing flow-control limits in our API, and MsQuic will not raise them on its own from 0. So with 0 the streams would be data-blocked from the start and would never progress.

src/libraries/System.Net.Quic/src/System/Net/Quic/Internal/MsQuicConfiguration.cs
```diff
@@ -117,16 +117,43 @@ private static unsafe MsQuicSafeHandle Create(QuicConnectionOptions options, QUI
...
+            ? (uint)options.KeepAliveInterval.TotalMilliseconds
+            : 0; // 0 disables the keepalive
+
+        settings.IsSet.ConnFlowControlWindow = 1;
+        settings.ConnFlowControlWindow = (uint)options.InitialReceiveWindowSizes.Connection;
+
+        settings.IsSet.StreamRecvWindowBidiLocalDefault = 1;
```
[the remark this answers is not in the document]
rzikm: Correct, the MsQuic version where this will be released should be 2.4. I was thinking whether we need to raise the minimum version of the library. AFAIK, it should work with older versions, the settings would just be ignored.

src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnectionOptions.cs
```diff
@@ -54,31 +97,58 @@ internal QuicConnectionOptions()
...
     public long DefaultCloseErrorCode { get; set; } = -1;
 
+    private QuicReceiveWindowSizes? _initialRecieveWindowSizes;
+
+    /// <summary>
+    /// The initial receive window sizes for the connection and individual stream types.
+    /// </summary>
+    public QuicReceiveWindowSizes InitialReceiveWindowSizes
+    {
+        get => _initialRecieveWindowSizes ??= new QuicReceiveWindowSizes();
```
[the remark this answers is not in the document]
rzikm: I agree, but I don't think it complicates the code that much. I can change it if you feel strongly about this.

src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnection.cs
```diff
@@ -93,6 +105,9 @@ static async ValueTask<QuicConnection> StartConnectAsync(QuicClientConnectionOpt
     private readonly ValueTaskSource _connectedTcs = new ValueTaskSource();
     private readonly ValueTaskSource _shutdownTcs = new ValueTaskSource();
 
+    private readonly CancellationTokenSource _shutdownCancellationTokenSource = new CancellationTokenSource();
```
CarnaViire, line 108: I think we might need to dispose `_shutdownCancellationTokenSource`?
Also, the naming confuses me... It implies the cancellation of the connection's shutdown, but it's not, it's for cancelling the handshake in the Listener, right? (and it also means it's not needed on the client side?)

rzikm: It is used only for listener to cancel the connection options callback if peer closes the connection before the handshake. I did not find a way to put the CTS outside QuicConnection.

I will try to come up with a better name

src/libraries/System.Net.Quic/tests/FunctionalTests/QuicListenerTests.cs
```diff
@@ -156,13 +158,60 @@ public async Task AcceptConnectionAsync_SlowOptionsCallback_TimesOut(bool useCan
             };
             await using QuicListener listener = await CreateQuicListener(listenerOptions);
 
-            ValueTask<QuicConnection> connectTask = CreateQuicConnection(listener.LocalEndPoint);
-            Exception exception = await AssertThrowsQuicExceptionAsync(QuicError.ConnectionTimeout, async () => await listener.AcceptConnectionAsync());
-            Assert.Equal(SR.Format(SR.net_quic_handshake_timeout, QuicDefaults.HandshakeTimeout), exception.Message);
+            QuicClientConnectionOptions clientOptions = CreateQuicClientOptions(listener.LocalEndPoint);
+            clientOptions.HandshakeTimeout = clientShorterTimeout ? TimeSpan.FromSeconds(2) : TimeSpan.FromSeconds(20);
```
ManickaP, line 162: I'm pretty sure this will eventually pop-up in CI report. Especially if you're asserting specific types of exception that should happen only in certain order of events depending on this timing.

But we can wait until we see it.

src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnection.cs
ManickaP: What do we do in SslStream case? Do we also react and cancel user callback when the client kills handshake in the meantime?

ManickaP (commented):
In general looks good. I think the last contentious thing is the `_shutdownCancellationTokenSource`.

src/libraries/System.Net.Quic/src/System/Net/Quic/QuicConnection.cs
rzikm: SslStream has only synchronous callback

rzikm: With sslstream we don't check for client-side activity while invoking the callback. We would have to always have a pending read on the inner stream and check for errors there. SslStream passes in the same token as to AuthenticateAsClientAsync.

rzikm: Renamed to `ConnectionShutdownToken` and added a comment

ManickaP (approved):
:shipit: 
Thank you!

ManickaP:
BTW, I assume this also fixes #70090

CarnaViire (approved):
LGTM, thanks

PR #94214, opened 2023-10-31
[browser][MT] enable MT PLINQ

pavelsavara:
Use normal threads in MT build of PLINQ rather than 
- `SinglePartitionMode => false;` when `FEATURE_WASM_THREADS`
- enabled more unit tests, which are no longer failing. There are still few which deadlock or have trouble.


ilonatommy (approved):
Looks logical

src/libraries/System.Threading.Tasks.Parallel/src/System/Threading/Tasks/Parallel.cs
```diff
@@ -238,8 +238,10 @@ public static void Invoke(ParallelOptions parallelOptions, params Action[] actio
             {
                 // If we've gotten this far, it's time to process the actions.
 
+#if !FEATURE_WASM_THREADS
                 // Web browsers need special treatment that is implemented in TaskReplicator
                 if (OperatingSystem.IsBrowser() ||
+#endif
```
campersau, line 244: This looks wrong. There is no `if (` when `FEATURE_WASM_THREADS` is enabled.

src/libraries/System.Threading.Tasks.Parallel/tests/System.Threading.Tasks.Parallel.Tests.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
   <PropertyGroup>
     <StartupObject />
+    <_WasmPThreadPoolSize Condition="'$(MonoWasmBuildVariant)' == 'multithread'">32</_WasmPThreadPoolSize>
```
kg, line 8: A value this high may cause reliability issues

pavelsavara: this is temporary and not production code

pavelsavara: If we have deputy design, we could let emscripten to allocate pthreads on the fly. 
That said these unit tests are nasty about creating many threads.

pavelsavara:
image

src/libraries/System.Threading.Tasks.Parallel/tests/System.Threading.Tasks.Parallel.Tests.csproj
```diff
@@ -6,6 +6,9 @@
   <PropertyGroup>
     <StartupObject />
   </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetOS)' == 'browser'">
+    <WasmXHarnessMonoArgs>--setenv=XHARNESS_LOG_TEST_START=true --no-memory-snapshot</WasmXHarnessMonoArgs>
```
kg, line 10: maybe add a comment explaining why this is here so we know whether we can remove it later

kg (approved):

src/libraries/System.Threading.Tasks.Parallel/tests/System.Threading.Tasks.Parallel.Tests.csproj
pavelsavara: I remove it, it helps you to see which test was running last and deadlocked.

PR #94226, opened 2023-10-31
Do some cleanup on TimeSpan to ensure better inlining and clearer semantics

tannergooding:
As per the title, this does some cleanup of TimeSpan to more consistently use throw helpers, to more consistently use named constants for important values, and to ensure that the implementation lives in the operators so less inlining is needed in the common case.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
@@ -566,7 +566,7 @@ namespace System
         {
             long result = t1._ticks + t2._ticks;
 
-            if (((t1._ticks >> 63) != (t2._ticks >> 63)) && ((t1._ticks >> 63) != (result >> 63)))
+            if (((t1._ticks >> 63) == (t2._ticks >> 63)) && ((t1._ticks >> 63) != (result >> 63)))
             {
                 // Overflow if signs of operands was identical and result's sign was opposite.
                 // >> 63 gives the sign bit (either 64 1's or 64 0's).
--- a/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
@@ -56,31 +56,59 @@ namespace System
         /// <remarks>
         /// The value of this constant is 10 thousand; that is, 10,000.
         /// </remarks>
-        public const long TicksPerMillisecond = TicksPerMicrosecond * 1000;
+        public const long TicksPerMillisecond = TicksPerMicrosecond * 1000;                         //          10,000
+        public const long TicksPerSecond = TicksPerMillisecond * 1000;                              //      10,000,000
+        public const long TicksPerMinute = TicksPerSecond * 60;                                     //     600,000,000
+        public const long TicksPerHour = TicksPerMinute * 60;                                       //  36,000,000,000
+        public const long TicksPerDay = TicksPerHour * 24;                                          // 864,000,000,000
 
-        public const long TicksPerSecond = TicksPerMillisecond * 1000;   // 10,000,000
+        internal const long MicrosecondsPerMillisecond = TicksPerMillisecond / TicksPerMicrosecond; //           1,000
+        internal const long MicrosecondsPerSecond = TicksPerSecond / TicksPerMicrosecond;           //       1,000,000
+        internal const long MicrosecondsPerMinute = TicksPerMinute / TicksPerMicrosecond;           //      60,000,000
+        internal const long MicrosecondsPerHour = TicksPerHour / TicksPerMicrosecond;               //   3,600,000,000
+        internal const long MicrosecondsPerDay = TicksPerDay / TicksPerMicrosecond;                 //  86,400,000,000
 
-        public const long TicksPerMinute = TicksPerSecond * 60;         // 600,000,000
+        internal const long MillisecondsPerSecond = TicksPerSecond / TicksPerMillisecond;           //           1,000
+        internal const long MillisecondsPerMinute = TicksPerMinute / TicksPerMillisecond;           //          60,000
+        internal const long MillisecondsPerHour = TicksPerHour / TicksPerMillisecond;               //       3,600,000
+        internal const long MillisecondsPerDay = TicksPerDay / TicksPerMillisecond;                 //      86,400,000
 
-        public const long TicksPerHour = TicksPerMinute * 60;        // 36,000,000,000
+        internal const long SecondsPerMinute = TicksPerMinute / TicksPerSecond;                     //              60
+        internal const long SecondsPerHour = TicksPerHour / TicksPerSecond;                         //           3,600
+        internal const long SecondsPerDay = TicksPerDay / TicksPerSecond;                           //          86,400
 
-        public const long TicksPerDay = TicksPerHour * 24;          // 864,000,000,000
+        internal const long MinutesPerHour = TicksPerHour / TicksPerMinute;                         //              60
+        internal const long MinutesPerDay = TicksPerDay / TicksPerMinute;                           //           1,440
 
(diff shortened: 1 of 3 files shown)


src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
```diff
@@ -141,19 +169,27 @@ public TimeSpan(int days, int hours, int minutes, int seconds)
         /// </exception>
         public TimeSpan(int days, int hours, int minutes, int seconds, int milliseconds, int microseconds)
         {
-            long totalMicroseconds = (((long)days * 3600 * 24 + (long)hours * 3600 + (long)minutes * 60 + seconds) * 1000 + milliseconds) * 1000 + microseconds;
-            if (totalMicroseconds > MaxMicroSeconds || totalMicroseconds < MinMicroSeconds)
+            long totalMicroseconds = (days * MicrosecondsPerDay)
```
tarekgh, line 172: >totalMicroseconds [](http://example.com/codeflow?start=17&length=17)

do we need to fix the possible overflow here?

tannergooding: Potentially, but I tried to keep this to non-behavioral changes.

I'd prefer to put a bug fix in its own PR where it can be isolated out and have relevant tests added.

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
```diff
@@ -545,50 +522,65 @@ public static bool TryParseExact([NotNullWhen(true)] string? input, [NotNullWhen
...
         }
 
-        public static TimeSpan operator -(TimeSpan t1, TimeSpan t2) => t1.Subtract(t2);
+        public static TimeSpan operator -(TimeSpan t1, TimeSpan t2)
+        {
+            long result = t1._ticks - t2._ticks;
+
+            if (((t1._ticks >> 63) != (t2._ticks >> 63)) && ((t1._ticks >> 63) != (result >> 63)))
+            {
+                // Overflow if signs of operands was identical and result's sign was opposite.
```
tannergooding, line 556:
```suggestion
                // Overflow if signs of operands was different and result's sign was opposite.
```

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
```diff
@@ -347,149 +349,126 @@ public static TimeSpan FromMilliseconds(double value)
...
         }
 
         // See System.Globalization.TimeSpanParse and System.Globalization.TimeSpanFormat
         #region ParseAndFormat
-        private static void ValidateStyles(TimeSpanStyles style, string parameterName)
+        private static void ValidateStyles(TimeSpanStyles style)
         {
-            if (style != TimeSpanStyles.None && style != TimeSpanStyles.AssumeNegative)
-                throw new ArgumentException(SR.Argument_InvalidTimeSpanStyles, parameterName);
+            if ((style != TimeSpanStyles.None) && (style != TimeSpanStyles.AssumeNegative))
```
tarekgh, line 390: >if ((style != TimeSpanStyles.None) && (style != TimeSpanStyles.AssumeNegative)) [](http://example.com/codeflow?start=12&length=79)

nit:

if (style is not TimeSpanStyles.None and not TimeSpanStyles.AssumeNegative)

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
```diff
@@ -347,149 +349,126 @@ public static TimeSpan FromMilliseconds(double value)
...
         }
-        public static bool TryParse(ReadOnlySpan<char> input, IFormatProvider? formatProvider, out TimeSpan result)
-        {
-            return TimeSpanParse.TryParse(input, formatProvider, out result);
-        }
+        public static bool TryParse(ReadOnlySpan<char> input, IFormatProvider? formatProvider, out TimeSpan result) => TimeSpanParse.TryParse(input, formatProvider, out result);
         public static bool TryParseExact([NotNullWhen(true)] string? input, [NotNullWhen(true), StringSyntax(StringSyntaxAttribute.TimeSpanFormat)] string? format, IFormatProvider? formatProvider, out TimeSpan result)
         {
-            if (input == null || format == null)
+            if ((input is null) || (format is null))
```
tarekgh, line 471: >if ((input is null) || (format is null)) [](http://example.com/codeflow?start=12&length=40)

if (input is null || format is null)

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
```diff
@@ -545,50 +522,65 @@ public static bool TryParseExact([NotNullWhen(true)] string? input, [NotNullWhen
...
+            }
             return new TimeSpan(-t._ticks);
         }
 
-        public static TimeSpan operator -(TimeSpan t1, TimeSpan t2) => t1.Subtract(t2);
+        public static TimeSpan operator -(TimeSpan t1, TimeSpan t2)
+        {
+            long result = t1._ticks - t2._ticks;
+
+            if (((t1._ticks >> 63) != (t2._ticks >> 63)) && ((t1._ticks >> 63) != (result >> 63)))
```
tarekgh, line 554: >(t1._ticks >> 63 [](http://example.com/codeflow?start=17&length=16)

does the jit optimize this calculation as it is used ywice in the statement? or should we store it and then use it?

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
```diff
@@ -545,50 +522,65 @@ public static bool TryParseExact([NotNullWhen(true)] string? input, [NotNullWhen
...
+        {
+            long result = t1._ticks + t2._ticks;
+
+            if (((t1._ticks >> 63) == (t2._ticks >> 63)) && ((t1._ticks >> 63) != (result >> 63)))
+            {
+                // Overflow if signs of operands was identical and result's sign was opposite.
+                // >> 63 gives the sign bit (either 64 1's or 64 0's).
+                ThrowHelper.ThrowOverflowException_TimeSpanTooLong();
+            }
+            return new TimeSpan(result);
```
tarekgh, line 575: can this refactored to its own method as it is used twice here?

tannergooding: It's a different check each time

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
tannergooding: It does. We can manually cache it and simplify things, however.

tarekgh (approved):
LGTM. thanks for cleaning this up!

src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs
```diff
@@ -545,50 +526,67 @@ public static bool TryParseExact([NotNullWhen(true)] string? input, [NotNullWhen
...
 
-        public static TimeSpan operator +(TimeSpan t1, TimeSpan t2) => t1.Add(t2);
+        public static TimeSpan operator +(TimeSpan t1, TimeSpan t2)
+        {
+            long result = t1._ticks + t2._ticks;
+            long t1Sign = t1._ticks >> 63;
+
+            if ((t1Sign == (t2._ticks >> 63)) && (t1Sign != (result >> 63)))
+            {
+                // Overflow if signs of operands was identical and result's sign was opposite.
```
danmoseley, line 577: just curious, you could have used >>> instead of >> here, right? then it would only be the low order bit holding the sign. does it make any difference to perf?

tannergooding: Signed and unsigned shift have the same cost in x64 and Arm64 hardware, so no real difference

PR #94234, opened 2023-10-31
Annotate projects that aren't XML source of truth

ViktorHofer:
with the `UseCompilerGeneratedDocXmlFile` switch.


src/libraries/System.Threading.RateLimiting/src/System.Threading.RateLimiting.csproj
```diff
@@ -41,12 +42,13 @@ System.Threading.RateLimiting.RateLimitLease</PackageDescription>
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
```
carlossanlop, line 44: Is there any advantage on switching their locations here?
Also, if they are ItemGroups with the exact same condition, wouldn't it be ok to group both ItemGroups into a single one?

carlossanlop (approved):

src/libraries/System.Threading.RateLimiting/src/System.Threading.RateLimiting.csproj
ViktorHofer: Our convention is properties first, then compile and other items and last reference items. Keeping the reference items in a separate item group makes it easier to read a project's dependencies.

PR #94236, opened 2023-10-31
[ILLink analyzer] Move dynamic object logic into dataflow analysis

sbomer:
Moves analysis of dynamic object invocations and related operations from the attribute analyzer into the dataflow analyzer. This lowers dynamic object operations into calls to methods on `Microsoft.CSharp.RuntimeBinder.Binder`, which have `RequiresUnreferencedCode` annotations, so that the analyzer produces the same warnings as ILLink.

Adds handling of previously unimplemented operations on dynamic objects.  Fixes https://github.com/dotnet/runtime/issues/94057

### DIFF
--- a/src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/DynamicObjects.cs
+++ b/src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/DynamicObjects.cs
@@ -14,6 +14,9 @@ namespace Mono.Linker.Tests.Cases.DataFlow
 	[Reference ("Microsoft.CSharp.dll")]
 	public class DynamicObjects
 	{
+		// Note on discrepancies between analyzer and NativeAot:
+		// Analyzer doesn't produce RequiresDynamicCode warnings for dynamic invocations.
+		// Tracked by https://github.com/dotnet/runtime/issues/94427.
 		public static void Main ()
 		{
 			InvocationOnDynamicType.Test ();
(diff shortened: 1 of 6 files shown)


src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/DynamicObjects.cs
```diff
@@ -134,7 +126,6 @@ class DynamicInRequiresUnreferencedCodeClass
 			[RequiresUnreferencedCode("message")]
 			class ClassWithRequires
 			{
-				// Analyzer hole: https://github.com/dotnet/runtime/issues/94057
 				[ExpectedWarning ("IL3050", ProducedBy = Tool.NativeAot)]
 				public static void MethodWithDynamicArg (dynamic arg)
```
vitek-karas, line 130: Should we have a new issue describing the need to fix this for RDC as well (eventually)?
or maybe I'm missing something - I would expect analyzer to behave the same as NativeAOT - or if it doesn't some comment or issue link describing the difference and "why".

src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/DynamicObjects.cs
```diff
@@ -151,7 +142,6 @@ public static void Test ()
 
 		class InvocationOnDynamicTypeInMethodWithRUCDoesNotWarnTwoTimes ()
 		{
-			// Analyzer hole: https://github.com/dotnet/runtime/issues/94057
 			[RequiresUnreferencedCode ("We should only see the warning related to this annotation, and none about the dynamic type.")]
 			[RequiresDynamicCode ("We should only see the warning related to this annotation, and none about the dynamic type.")]
```
vitek-karas, line 146: Same comment as above.

vitek-karas (approved):
Other than some test comment this looks good. Thanks!

Issue #92911, opened 2023-10-02
Support dsym bundles in symbol packages/indexing


agocke:
We currently use .dwarf files, which are unsupported.

BruceForstall:
Some notes on this issue:

Apple's supported symbol file format is the `.dSYM` bundle. A `.dSYM` bundle is a directory tree in a specific format that contains various files, including the DWARF format symbols for the executable, as well as various meta-data. In the macOS Finder it appears to be a single file (but can be explored using the "Show Package Contents" command). Confusingly, much documentation uses the term ".dSYM file", which is used essentially interchangeably with ".dSYM bundle". Note, however, that the `.dSYM` is always a bundle (directory tree), never a single file.

The .NET build process currently creates a single file for each executable file's symbols, with the file extension `.dwarf`. This is not the standard way that symbol files are created and used on macOS. There is some concern that this might not be supported long-term. There may also be developer experience improvements to using `.dSYM` format symbols, such as automatic Spotlight integration for symbol finding.

Creating a .dwarf file is done by running the `dsymutil --flat` command on an executable. Simply removing the `--flat` argument during the build process allows creating `.dSYM` bundles. However, there are lots of parts of the .NET build and packaging process that need to be updated to deal with the fact that symbols are now directory trees and not just single files.

It is best to consider a `.dSYM` bundle as opaque: do not make assumptions about the structure and contents of the bundle. There may be reasons to break this rule. For instance, one file inside the `.dSYM` bundle is the DWARF symbol file. We might want to read that file directly in order to index the symbols. However, even that might be better done by addressing the `.dSYM` bundle as a whole, e.g., using the following commands to get the build UUID for the contained symbols:
```
% dwarfdump --uuid createdump.dSYM 
UUID: D5BFA2CC-C73B-3E6F-8B2C-E96246EE2EAD (arm64) createdump.dSYM/Contents/Resources/DWARF/createdump
```

A description of some of the challenges in converting the .NET build to using `.dSYM` bundles follows.

## MSBuild files versus directories

The .NET build uses msbuild for most activities. MSBuild has a strong bias towards files and not directories, at least when it comes to [item glob expressions](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-items?view=vs-2022#use-wildcards-to-specify-items). For example, using:
```
<Files Include="*.pdb;*.dSYM"/>
```
Will include all the `*.pdb` files in the directory, but will ignore any `*.dSYM` bundles.

Using:
```
[...]

PR #94237, opened 2023-10-31
Build Mac symbols in .dSYM bundle, not single .dwarf file

BruceForstall:


BruceForstall:
Related: https://github.com/dotnet/runtime/issues/92911

BruceForstall:
@hoyosjs @agocke Here's a first attempt at trying to build .dSYM bundles instead of .dwarf files in our build.

BruceForstall:
cc @ericstj Since you've done some work in these areas.

eng/native/naming.props
```diff
@@ -18,7 +18,8 @@
         <LibPrefix>lib</LibPrefix>
         <LibSuffix>.dylib</LibSuffix>
         <StaticLibSuffix>.a</StaticLibSuffix>
-        <SymbolsSuffix>.dwarf</SymbolsSuffix>
+        <!-- <SymbolsSuffix>.dwarf</SymbolsSuffix> -->
+        <SymbolsSuffix>.dSYM</SymbolsSuffix>
```
hoyosjs, line 22: This might quickly become a leaky abstraction. For example, we also are missing updating `GetSymbolPackageFiles` in https://github.com/dotnet/runtime/blob/main/src/installer/pkg/projects/Directory.Build.targets (the excludes and all that)

PR #93704, opened 2023-10-18
JIT: remove simple lowering

AndyAyersMS:
Move the cast handling to rationalize, and the array length handling to lower.

Issue #93321, opened 2023-10-11
Test failure: AggregateException: Not all processes in process tree could be terminated.


v-wenyuxu:
**Failed in:** [runtime-coreclr jitstress 20231009.2](https://dev.azure.com/dnceng-public/public/_build/results?buildId=432736&view=ms.vss-test-web.build-test-results-tab&runId=9585790&resultId=118037&paneView=debug)

**Failed tests:**
```
coreclr osx x64 Checked jitstress2 @ OSX.1200.Amd64.Open
    - baseservices/exceptions/simple/ParallelCrashMainThread/ParallelCrashMainThread.sh
    - baseservices/exceptions/simple/ParallelCrash/ParallelCrash.sh
    - baseservices/exceptions/unhandled/unhandled/unhandled.sh
```

**Error message:**
```
Test Infrastructure Failure: System.AggregateException: Not all processes in process tree could be terminated. (Operation not permitted)
 ---> System.ComponentModel.Win32Exception (1): Operation not permitted
   --- End of inner exception stack trace ---
   at System.Diagnostics.Process.Kill(Boolean entireProcessTree)
   at CoreclrTestLib.CoreclrTestWrapperLib.CollectCrashDumpWithCreateDump(Process process, String crashDumpPath, StreamWriter outputWriter)
   at CoreclrTestLib.CoreclrTestWrapperLib.CollectCrashDump(Process process, String crashDumpPath, StreamWriter outputWriter)
   at CoreclrTestLib.CoreclrTestWrapperLib.RunTest(String executable, String outputFile, String errorFile, String category, String testBinaryBase, String outputDir)
   at baseservices_exceptions._simple_ParallelCraMainThread_ParallelCraMainThread_._simple_ParallelCrashMainThread_ParallelCrashMainThread_sh()
Expected: True
Actual:   False
```

**Stack trace:**
```
   at baseservices_exceptions._simple_ParallelCraMainThread_ParallelCraMainThread_._simple_ParallelCrashMainThread_ParallelCrashMainThread_sh()
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```
<html>
<body>
<div><br/><br/>

Queued | Pipeline | Pipeline_Configuration | OS | Arch | Test | Outcome | Properties | BuildReason
[...]

PR #86839, opened 2023-05-27
JIT: Clean up BB successor iteration

jakobbotsch:
* Remove old code
* Use visitor with an array for `AllSuccessorEnumerator` that is still used by SSA

PR #94247, opened 2023-10-31
JIT: morph blocks in RPO

AndyAyersMS:
When optimizing, process blocks in RPO. Disallow creation of new blocks and new flow edges (the latter with certain preapproved exceptions).

Morph does not yet take advantage of the RPO to enable more optimization.

Contributes to #93246.


AndyAyersMS:
@jakobbotsch PTAL
cc @dotnet/jit-contrib 

Not yet a general purpose utility -- will have to wait for a second "customer" to figure out how to best refactor this.

A bit higher TP impact than I'd like, around 0.25%. Some of this was paid for in advance by https://github.com/dotnet/runtime/pull/93704, which saved about 0.15%. I will keep looking for ways to pare this down.

Some diffs from places where morph is order sensitive. Mainly seems to be `bbNum` changes since we now do an up-front renumber and/or changes in when a local is marked as address exposed. For example `fgMorphPotentialTailCall` checks all lcl vars looking for exposed locals, but the set of exposed locals can change as we morph, so depending on ordering we may do more or fewer tail calls.

AndyAyersMS:
Re the order sensitivity (at least for address exposure): one could perhaps argue that any locals introduced during morph shouldn't count... that might reduce or remove this source of diffs all together.

src/coreclr/jit/morph.cpp
```diff
@@ -13892,13 +13891,70 @@ PhaseStatus Compiler::fgMorphBlocks()
...
+            fgMorphBlock(block);
+        }
+    }
+    else
+    {
+        // We are optimizing. Process in RPO.
+        //
+        fgRenumberBlocks();
+        EnsureBasicBlockEpoch();
+        fgComputeEnterBlocksSet();
```
jakobbotsch, line 13913: How much of the cost comes from having to do the above setup calls before even computing the order?
When I've tried to compute the order before it wasn't clear to me why I would need to renumber blocks to do that.

jakobbotsch (approved):

jakobbotsch:
> Re the order sensitivity (at least for address exposure): one could perhaps argue that any locals introduced during morph shouldn't count... that might reduce or remove this source of diffs all together.

Another source of these is that block morphing depends on whether structs are marked as DNER, and morph can mark structs as DNER in some cases.

I think it's fine (and preferable) to take this churn – it at least makes it more well defined and depending on a logical order when this may happen, so probably reduces spurious diffs in the long run.

src/coreclr/jit/morph.cpp
AndyAyersMS: I can grab a new profile, but the old one is probably pretty accurate: see https://github.com/dotnet/runtime/pull/86822#issuecomment-1626319133

Renumbering likely isn't needed and not renumbering should save a bit of TP. So let me try that.

Also looking back at that early PR reminded me that I should take a look at if SSA's topological sort, it might be more efficient.

jakobbotsch: It's also possible we can make `DfsBlockEntry` cheaper in the same way as #86839 (even though `GetSucc` looks cheap, it would probably still help a bit).

It also confuses me a bit that we have multiple notions of start nodes with `fgEnterBlks` and `fgDomFindStartNodes`. I assume the nodes found exclusively by `fgDomFindStartNodes` are unreachable (?) so ideally we would get rid of them earlier, but even if not, isn't it optimizable by using `bbRefs`? If not, then I think using `VisitSuccs` would be more efficient.

Anyway, don't feel the need to address it in this PR -- but might be a couple of avenues of future cleanup/improvements.

AndyAyersMS: Seems like for now at least renumbering is needed. Let me add some notes to #93246 about making all this more efficient. I'll come back to that after I get the assertion prop part working.

src/coreclr/jit/morph.cpp
```diff
@@ -13892,13 +13891,70 @@ PhaseStatus Compiler::fgMorphBlocks()
...
+        //
+        if (genReturnBB != nullptr)
+        {
+            genReturnBB->bbFlags |= BBF_CAN_ADD_PRED;
+        }
+        if (fgFirstBBisScratch())
+        {
+            fgFirstBB->Next()->bbFlags |= BBF_CAN_ADD_PRED;
+        }
+
```
BruceForstall, line 13936: It's unfortunate we need these special cases (and the new block flag), but I understand you tried to avoid it and this is a compromise.

src/coreclr/jit/morph.cpp
```diff
@@ -13892,13 +13891,70 @@ PhaseStatus Compiler::fgMorphBlocks()
...
+        unsigned const bbNumMax = fgBBNumMax;
```
BruceForstall, line 13937: Do you capture `fgBBNumMax` because the number of blocks can change during morph? But you've disabled block creation. Do you want an assert after the `for` loop to assert no new blocks were created?

AndyAyersMS:
Jit stress failure looks like #93321, probably unrelated.

src/coreclr/jit/morph.cpp
AndyAyersMS: Yeah this is a vestige of my earlier versions where the number of blocks could change, before I removed all the bits that create new blocks.

You think this should have `assert(fgBBNumMax == bbNumMax`)?

If so, sure. I can add that in my next PR.

BruceForstall: > You think this should have `assert(fgBBNumMax == bbNumMax`)?

Right. That would clarify the assumption that the set of blocks doesn't change across this loop.

PR #94248, opened 2023-10-31
Add some sanity checking for SPMI JIT options arguments

BruceForstall:
Check that the argument passed doesn't include a leading `DOTNET_`.

Update the help to specify the allowed `key#value` format.

Currently using `DOTNET_` causes a warning, but arguably superpmi.py should just pass it through and superpmi.exe should just strip the prefix, so it would "just work".


BruceForstall:
@jakobbotsch PTAL
cc @dotnet/jit-contrib

BruceForstall:
I was confused for a while when something I tried didn't cause any spmi replay errors, and it's because I was passing `-jitoption DOTNET_xxx=yyy`.

jakobbotsch (approved):

PR #94249, opened 2023-10-31
we need to skip the SIP region in the last step of handling the alloc region

Maoni0:
in `process_remaining_regions`, if we haven't gone through the while loop to handle the regions with pinned plugs, we can end up with `current_region` being an SIP region. then we mistakenly call `decide_on_demotion_pin_surv` which can change its planned generation. I'm filtering this out before we call `decide_on_demotion_pin_surv` because the pinned survival in an SIP region is always 0 so we wouldn't want `decide_on_demotion_pin_surv` to treat it as "to be empty". 

thanks very much, @kevingosse, for reporting this!


teo-tsirpanis:
I've been wondering, what is SIP?

Maoni0:
SIP stands for Swept In Plan - those are special regions that are swept during a compacting GC because the survival rate is very high on them. they are disabled by default as they might increase your heap size noticeably.

teo-tsirpanis:
Thanks!

cshung (approved):
LGTM

PR #94258, opened 2023-11-01
Make some Span methods NonVersionable

MichalPetryka:
Marks some potentially-hot span methods as NonVersionable to improve R2R codegen, motivated by bad codegen from RVA statics there.

cc @jakobbotsch


jkotas:
Can share before/after for the code improvement?

MichalPetryka:
> Could you please share before/after for the code improvement?

Can't check the exact change here right now, [but I assume it should be the same as `--inputbubble`.](https://godbolt.org/z/qb5jnaTzc)

jkotas:
`NonVersionable` does not work the same way as inputbubble. `NonVersionable` only works for leaf methods currently. I suspect that this change does nothing to improve R2R code quality.

src/libraries/System.Private.CoreLib/src/System/Span.cs
```diff
@@ -38,6 +38,7 @@ namespace System
         /// <param name="array">The target array.</param>
         /// <remarks>Returns default when <paramref name="array"/> is null.</remarks>
         /// <exception cref="ArrayTypeMismatchException">Thrown when <paramref name="array"/> is covariant and array's type is not exactly T[].</exception>
+        [NonVersionable]
```
huoyaoyuan, line 41: Hypothetically, would we ever change handling of array covariance? The constructor doesn't look trivial enough.

JulieLeeMSFT:
@MichalPetryka, please follow up on the comments from reviewers.

MichalPetryka:
Ah I forgot about this, I'll test the changes out and address feedback in a few weeks since I have some hardware issues currently.

src/libraries/System.Private.CoreLib/src/System/Span.cs
MichalPetryka: Changing it would require a R2R version bump anyway due to the indexer being an intrinsic.

MichalPetryka:
> Actually, some of the annotated constructors are leaf methods. It should work for those, but it would be a good idea to double check it.

I've gotten around to checking it, seems like all of the APIs are still not inlined even after the change, is there some other way to force the JIT to do so without `--inputbubble`?

MichalPetryka:
Seems like having a `call` blocks `NonVersionable`?
https://github.com/dotnet/runtime/blob/b94a82d63580775b2adb0ccac478719b8cd7049b/src/coreclr/tools/aot/ILCompiler.ReadyToRun/Compiler/ReadyToRunCompilationModuleGroupBase.cs#L669

jkotas:
> Seems like having a call blocks NonVersionable?

Yes, the call can refer to internal implementation details that we are not able to encode without introducing new references that go outside the versioning bubble.

We may be able to allow this for calls that get inlined without references to any internal details. Alternatively, we can make the internal implementation details that the non-versionable methods refer to be part of the public surface. It would require rethinking how this whole thing works.

Issue #94105, opened 2023-10-27
ConfigurationBinder source generator generates code with the wrong nullable annotations


eerhardt:
### Description

When attempting to use the ConfigurationBinder source generator in dotnet/extensions (see https://github.com/dotnet/extensions/pull/4625), I'm getting a nullable warning that is causing the generated code to not compile (I would need to disable nullable validation for the whole project if I wanted to ignore the nullable warning, which isn't acceptable).

I'm not sure what the general pattern is here that causes this issue, so I logged the issue for the scenario specifically. I think it might need to include nested generics where one of the nested generic types is nullable.

### Reproduction Steps

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <EnableConfigurationBindingGenerator>true</EnableConfigurationBindingGenerator>
    <NoWarn>$(NoWarn);SYSLIB1100;SYSLIB1101</NoWarn>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="8.0.0-rtm.23511.16" />
    <PackageReference Include="Microsoft.Extensions.Http.Resilience" Version="8.0.0-rc.2.23510.2" />
  </ItemGroup>

</Project>
```

```csharp
// needed to work around https://github.com/dotnet/runtime/issues/94065
global using Polly;
global using Polly.Hedging;
global using Polly.RateLimiting;
global using Polly.Timeout;

using Microsoft.Extensions.Configuration;
using Microsoft.Extensions.Http.Resilience;

var c = new ConfigurationBuilder().Build();
c.Get<HttpStandardHedgingResilienceOptions>();
```

### Expected behavior

Should be without warnings

### Actual behavior

Get a nullable warning from generated code:

```
[...]

PR #94267, opened 2023-11-01
Use fully qualified names when referencing user types.

### LINKED ISSUE #93498: ConfigurationBinder source generator encounters namespace conflicts in using statements
### Description

When the first part of a namespace of a bound type exists under the Microsoft namespace an error will occur
```c#
c.Get<Foo.Bar.BType>();

namespace Microsoft.Foo
{
    internal class AType {}
}

namespace Foo.Bar
{
    internal class BType {}
}
```

> C:\scratch\namespaceConflict\Microsoft.Extensions.Configuration.Binder.SourceGeneration\Microsoft.Extensions.Configuration.Binder.SourceGeneration.ConfigurationBindingGenerator\BindingExtensions.g.cs(22,15): error CS0234: The type or namespace name 'Bar' does not exist in the namespace 'Microsoft.Foo' (are you missing an assembly reference?) [C:\scratch\namespaceConflict\namespaceConflict.csproj]

This is happening because the generator places using statements in a namespace `Microsoft.Extensions.Configuration.Binder.SourceGeneration` and this will cause the compiler to first search in containing namespaces at every level before considering the global namespace for usings placed in that namespace.

Hit this when porting eShopOnContainers.

### Reproduction Steps

Build 
[namespaceConflict.zip](https://github.com/dotnet/runtime/files/12900998/namespaceConflict.zip)

### Expected behavior

Builds without issue

### Actual behavior

Compilation error.

### Regression?

No

### Known Workarounds

Alias the reference that has the conflicting namespace, change namespace names, etc.

### Configuration

_No response_

### Other information

_No response_

eiriktsarpalis:
This also disables nullability warnings (using annotations-only like with STJ) so this should mitigate the issue as reported in #94105.

Fix #93498.


src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Specs/TypeIndex.cs
```diff
@@ -109,14 +109,31 @@ public static string GetParseMethodName(ParsableFromStringSpec type)
         {
             Debug.Assert(type.StringParsableTypeKind is not StringParsableTypeKind.AssignFromSectionValue);
 
-            string displayString = type.DisplayString;
+            if (type.StringParsableTypeKind is StringParsableTypeKind.ByteArray)
```
eiriktsarpalis, line 112: I had to refactor this method somewhat to make it account for using fully qualified names.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -548,19 +547,19 @@ private ObjectSpec CreateObjectSpec(TypeParseInfo typeParseInfo)
                     if (!hasPublicParameterlessCtor && hasMultipleParameterizedCtors)
                     {
                         initDiagDescriptor = DiagnosticDescriptors.MultipleParameterizedConstructors;
-                        initExceptionMessage = string.Format(Emitter.ExceptionMessages.MultipleParameterizedConstructors, typeName);
+                        initExceptionMessage = string.Format(Emitter.ExceptionMessages.MultipleParameterizedConstructors, typeSymbol.GetFullName());
```
tarekgh, line 550: >typeSymbol.GetFullName() [](http://example.com/codeflow?start=122&length=24)

Looks this can be called multiple times. do we need to cache it better?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
```diff
@@ -437,7 +436,7 @@ void EmitBindImplForMember(MemberSpec member)
                         _writer.WriteLine($$"""
                             {{condition}}
                             {
-                                throw new {{Identifier.InvalidOperationException}}("{{string.Format(ExceptionMessages.ParameterHasNoMatchingConfig, type.Name, member.Name)}}");
+                                throw new {{Identifier.InvalidOperationException}}("{{string.Format(ExceptionMessages.ParameterHasNoMatchingConfig, type.FullName, member.Name)}}");
```
tarekgh, line 439: >FullName [](http://example.com/codeflow?start=153&length=8)

why didn't use `type.TypeRef.FullyQualifiedName` here?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
```diff
@@ -1110,24 +1109,24 @@ private bool EmitObjectInit(ComplexTypeSpec type, string memberAccessExpr, Initi
...
 
                         if (collectionWithCtorInitType.InstantiationConcreteType is not CollectionInstantiationConcreteType.Self)
                         {
-                            castExpr = $"({collectionWithCtorInitType.DisplayString})";
+                            castExpr = $"({collectionWithCtorInitType.TypeRef.FullyQualifiedName})";
                         }
 
-                        effectiveDisplayString = _typeIndex.GetInstantiationTypeDisplayString(collectionWithCtorInitType);
-                        initExpr = $"{castExpr}new {effectiveDisplayString}()";
+                        typeFQN = TypeIndex.GetInstantiationTypeDisplayString(collectionWithCtorInitType);
```
tarekgh, line 1128: >TypeIndex [](http://example.com/codeflow?start=34&length=9)

why didn't use `_typeIndex` here?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
```diff
@@ -1213,7 +1212,7 @@ private void EmitCollectionCastIfRequired(CollectionWithCtorInitSpec type, out s
                     return;
                 }
 
-                string castTypeDisplayString = _typeIndex.GetPopulationCastTypeDisplayString(type);
+                string castTypeDisplayString = TypeIndex.GetPopulationCastTypeDisplayString(type);
```
tarekgh, line 1215: >TypeIndex [](http://example.com/codeflow?start=47&length=9)

ditto

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
tarekgh: I see now you have added FullName property but why we didn't use this in the line 357?

tarekgh:
>             private readonly SortedSet<string> _namespaces = new()

do we still need collect the system namespaces? 

---
Refers to: src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Specs/BindingHelperInfo.cs:34 in 86a648b. [](commit_id = 86a648bf56e76e1c0fd785be482ef8f4469c40b3, deletion_comment = False)

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Specs/TypeIndex.cs
```diff
@@ -109,14 +109,31 @@ public static string GetParseMethodName(ParsableFromStringSpec type)
...
-                ? "ParseByteArray"
-                // MinimalDisplayString.Length is certainly > 2.
-                : $"Parse{(char.ToUpper(displayString[0]) + displayString.Substring(1)).Replace(".", "")}";
+            Debug.Assert(displayString.Length > 0);
+            if (char.IsLower(displayString[0]))
+            {
+                displayString = char.ToUpperInvariant(displayString[0]) + displayString.Substring(1);
+            }
+
+            if (displayString.Contains("."))
```
tarekgh, line 131: >"." [](http://example.com/codeflow?start=39&length=3)

`'.'`

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.cs
```diff
@@ -176,6 +176,33 @@ public class MyClass { }
...
+                }
+
+                namespace Foo.Bar
+                {
+                    internal class BType {}
+                }
+                """;
+
+            ConfigBindingGenRunResult result = await RunGeneratorAndUpdateCompilation(source);
+            Assert.NotNull(result.GeneratedSource);
```
tarekgh, line 202: >Assert.NotNull(result.GeneratedSource); [](http://example.com/codeflow?start=12&length=39)

would make sense to check the global names of both types in the generated source? I mean searching the text with these names? I am not feeling strongly about. just wanted to say it :-)

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Helpers.cs
```diff
@@ -138,7 +138,11 @@ private static async Task VerifyThatSourceIsGenerated(string testSourceCode)
                 string source = string.Join(Environment.NewLine, lines).TrimEnd(Environment.NewLine.ToCharArray()) + Environment.NewLine;
                 path = Path.Combine($"{repoRootDir}\\src\\libraries\\Microsoft.Extensions.Configuration.Binder\\tests\\SourceGenerationTests\\", path);
 
-                await File.WriteAllTextAsync(path, source).ConfigureAwait(false);
+#if NETCOREAPP
+                await File.WriteAllTextAsync(path, source);
```
tarekgh, line 142: >File.WriteAllTextAsync(path, source); [](http://example.com/codeflow?start=22&length=37)

I think you still need `.ConfigureAwait(false);`

tarekgh (approved):
LGTM. I added a few questions.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
eiriktsarpalis: The string is only used when we emit diagnostics. Given that diagnostics tend to be exceptional, I intentionally changed it so that we don't have to create the string every time.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
eiriktsarpalis: For runtime exceptions I'm using the "FullName" since there are tests checking for compatibility with `System.Type.ToString()` being thrown by the reflection binder.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Emitter/CoreBindingHelpers.cs
eiriktsarpalis: Because the implementation changed to use the FQN, the method doesn't require any indirection and was converted to `static`.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Helpers.cs
eiriktsarpalis: It's test code only run by xunit, so not really necessary.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.cs
eiriktsarpalis: I would consider this an implementation detail, from a testing perspective what matters is that the particular example compiles. I figured that the baselines should provide sufficient evidence as to what code does get generated.

eiriktsarpalis:
I'm guessing the baseline files require a manual backport.

PR #94275, opened 2023-11-01
Small cleanup for Lock in Mono

kouvel:
- Used the project property instead of disabling the compiler warning. Probably a bit cleaner, and would enable using Lock in other libraries to try it out.
- Undid an unnecessary change from https://github.com/dotnet/runtime/pull/87672

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
+++ b/src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
@@ -4,6 +4,9 @@
     <SharedGUID>c5ed3c1d-b572-46f1-8f96-522a85ce1179</SharedGUID>
     <StringResourcesName Condition="'$(StringResourcesName)' == ''">System.Private.CoreLib.Strings</StringResourcesName>
     <GenerateResourcesSubstitutions>true</GenerateResourcesSubstitutions>
+
+    <!-- CA2252: Opt in to preview features before using them (for System.Threading.Lock) -->
+    <EnablePreviewFeatures>true</EnablePreviewFeatures>
   </PropertyGroup>
   <PropertyGroup Label="Configuration">
     <Import_RootNamespace />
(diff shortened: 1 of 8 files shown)


src/mono/System.Private.CoreLib/System.Private.CoreLib.csproj
```diff
@@ -13,6 +13,9 @@
     <Platforms>x64;x86;arm;armv6;arm64;riscv64;s390x;wasm;ppc64le</Platforms>
 
     <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
+
+    <!-- CA2252: Opt in to preview features before using them (Lock) -->
```
jkotas, line 17: Should it rather be in the shared `src\libraries\System.Private.CoreLib\src\System.Private.CoreLib.Shared.projitems`?

kouvel: That would work better, updated

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
```diff
@@ -4,6 +4,9 @@
     <SharedGUID>c5ed3c1d-b572-46f1-8f96-522a85ce1179</SharedGUID>
     <StringResourcesName Condition="'$(StringResourcesName)' == ''">System.Private.CoreLib.Strings</StringResourcesName>
     <GenerateResourcesSubstitutions>true</GenerateResourcesSubstitutions>
+
+    <!-- CA2252: Opt in to preview features before using them (for System.Threading.Lock) -->
```
jkotas, line 8: You can also delete CA2252 pragmas from under src/libraries/System.Private.CoreLib

jkotas: E.g. here: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/NativeRuntimeEventSource.Threading.cs#L98

jkotas (approved):

PR #94285, opened 2023-11-02
Arm64: SVE/SVE2 encodings

kunalspathak:
## Summary

Add SVE/SVE2 encodings for all the possible instructions.

## Stats:
- 606 new instructions
- 1123 new encodings of newly added instructions
- 378 new instruction formats
- 103 different groupings of newly added instruction formats.

## Design

I have added 2 new files `emitfmtsarm64_sve.h` and `instrsarm64_sve.h` for SVE encodings. They are separate from existing `emitfmtsarm64.h` and `instrsarm64.h` because had we added SVE encodings in existing files that contains general/NEON instructions, it would impact JIT throughput heavily specially around the code to retrieve hex code for non-sve instructions. The instruction table preparation happens using existing machinery in a separate method `emitSveInsCode()`. The codegen, depending on the intrinsic, will call `emitSveInsCode()` and thus without affecting the code path of non-sve instructions. 

I had to readjust the offsets of `instrDesc` to include the new size of `_idIns` and `_idInsFmt`. Here are the offsets before/after:

Before:
```

clrjit_universal_arm64_x64!emitter::instrDesc
   +0x000 _idIns           : Pos 0, 10 Bits
   +0x000 _idInsFmt        : Pos 10, 8 Bits
   +0x000 _idGCref         : Pos 18, 2 Bits
   +0x000 _idReg1          : Pos 20, 6 Bits
   +0x000 _idReg2          : Pos 26, 6 Bits
   +0x004 _idSmallDsc      : Pos 0, 1 Bit
   +0x004 _idLargeCns      : Pos 1, 1 Bit
   +0x004 _idLargeDsp      : Pos 2, 1 Bit
   +0x004 _idLargeCall     : Pos 3, 1 Bit
   +0x004 _idBound         : Pos 4, 1 Bit
   +0x004 _idCallAddr      : Pos 5, 1 Bit
   +0x004 _idNoGC          : Pos 6, 1 Bit
   +0x004 _idOpSize        : Pos 7, 3 Bits
   +0x004 _idInsOpt        : Pos 10, 6 Bits
   +0x004 _idLclVar        : Pos 16, 1 Bit
   +0x004 _idLclVarPair    : Pos 17, 1 Bit
   +0x004 _idCnsReloc      : Pos 18, 1 Bit
   +0x004 _idDspReloc      : Pos 19, 1 Bit
   +0x004 _idSmallCns      : Pos 20, 12 Bits
   +0x008 _idAddrUnion     : emitter::instrDesc::idAddrUnion
```

After:
```
clrjit_universal_arm64_x64!emitter::instrDesc
   +0x000 _idIns           : Pos 0, 11 Bits
   +0x000 _idInsFmt        : Pos 11, 10 Bits
   +0x000 _idOpSize        : Pos 21, 3 Bits
   +0x000 _idInsOpt        : Pos 24, 6 Bits
   +0x000 _idGCref         : Pos 30, 2 Bits
   +0x004 _idReg1          : Pos 0, 6 Bits
   +0x004 _idReg2          : Pos 6, 6 Bits
   +0x004 _idSmallDsc      : Pos 12, 1 Bit
   +0x004 _idLargeCns      : Pos 13, 1 Bit
   +0x004 _idLargeDsp      : Pos 14, 1 Bit
   +0x004 _idLargeCall     : Pos 15, 1 Bit
   +0x004 _idBound         : Pos 16, 1 Bit
   +0x004 _idCallAddr      : Pos 17, 1 Bit
   +0x004 _idNoGC          : Pos 18, 1 Bit
   +0x004 _idLclVar        : Pos 19, 1 Bit
   +0x004 _idLclVarPair    : Pos 20, 1 Bit
   +0x004 _idCnsReloc      : Pos 21, 1 Bit
   +0x004 _idDspReloc      : Pos 22, 1 Bit
   +0x004 _idSmallCns      : Pos 23, 9 Bits
   +0x008 _idAddrUnion     : emitter::instrDesc::idAddrUnion
```

## Nomenclature
- All the instruction's id has a prefix of `sve_` to differentiate them from regular instructions, e.g. `INS_sve_mov` vs. `INS_mov`. 
- Likewise, instruction formats have prefix of `SVE_` to differentiate from existing instruction formats. e.g. `IF_SVE_AB_3A` vs. `IF_DI_3A`. 
- Currently, various instruction formats are grouped based on the common logic various instructions having them share during encoding and in various other methods. I do not have idea at this point how many of newly added instruction formats will share logic between various instructions, hence I have created as many unique as I could, and in future we could eliminate/combine many of them. 
- The naming convention for instruction formats is as follows: `SVE_WW_XY`, where `WW` is unique code like `AA`, `AB`, etc., `X` represents number of registers the instruction operates on, `Y` represents unique code `A`, `B` depending on if same instruction has different formats.

## Arm64 Encoding generation tool
As mentioned in https://github.com/dotnet/runtime/issues/93095, it is impossible to hand write each encoding by looking at the manual. Hence, I wrote a small tool that parses the manual available in [xml format on Arm site](https://developer.arm.com/downloads/-/exploration-tools) and generate C++ code that is included in `emitfmtsarm64_sve.h` and `instrsarm64_sve.h`. The plan is to open source the tool and check it in https://github.com/dotnet/jitutils/.


kunalspathak:
Need to figure out the cause of TP regression. Most likely it could be the difference in code generated for existing `emitInsCode()` because of increase in number of entries in `enum` (?). Understandable to see more regression on MinOpts than FullOpts because more instructions are emitted for minopts.

image

kunalspathak:
@dotnet/jit-contrib 
cc: @TamarChristinaArm , @a74nh , @TIHan @amanasifkhalid

kunalspathak:
> Need to figure out the cause of TP regression. Most likely it could be the difference in code generated for existing `emitInsCode()` because of increase in number of entries in `enum` (?). Understandable to see more regression on MinOpts than FullOpts because more instructions are emitted for minopts.
> 
> image

On @BruceForstall 's suggestion, I just made a PR https://github.com/dotnet/runtime/pull/94310 that changes the offsets of `instrDesc`, the way they are changed in this PR and see the TP impact and it looks exactly the same as one we see here. So clearly it is coming because of reduction of smallCns bit size.

[TP diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=458904&view=ms.vss-build-web.run-extensions-tab ):

image

a74nh:
> On @BruceForstall 's suggestion, I just made a PR #94310 that changes the offsets of `instrDesc`, the way they are changed in this PR and see the TP impact and it looks exactly the same as one we see here. So clearly it is coming because of reduction of smallCns bit size.

Am I right in thinking the overall size of a `instrDesc` has gone from 32bits to 64bits?

In https://github.com/dotnet/runtime/pull/94310, instead of increasing the variables, what happens if you instead add padding? Eg:
```
instruction _idIns : 10;
instruction _pad_idIns : 1;
```
etc for all the other fields.
That would confirm if the extra cost is due to the overall size of `instrDesc` or if it is the increased field sizes.

-----

Thinking ahead, SME is going to add a similar slowdown.

What about if SVE instruction enum started again from 0. The same with all the instruction formats. Treat it as a separate architecture.
You could have a single bit in `instrDesc` to indicate if it is Arm64 or Arm64Sve (and, then maybe increase to 2bits when doing SME). Or override instrDesc, creating an instrDescSve, and then emitNewInstr() can take in an enum which decides which instrDesc to allocate.
You'd have to have different `emitIns_R_R_R()` etc functions for sve. But, that's ok because it's keeping the switch statements shorter.

kunalspathak:
> Am I right in thinking the overall size of a `instrDesc` has gone from 32bits to 64bits?

It was and is 16 bytes. `instrDesc` has 2 chunks of 8 bytes each. The first 8 bytes contains various bit fields and the last 8 bytes represents the `idAddrUnion` . I increased the size of `idIns` and `idInsFmt` that falls in the first chunk, but still keeping the overall size of that chunk 8 bytes. To offset that, we have to reduce the size of bits needed for `idSmallCns`. As seen in https://github.com/dotnet/runtime/pull/94285/files#diff-125fb0b9396b0e85ee82c6d592cdf1750778a7cff38ad382f7c2966ef588dce1R891, what that means is, earlier, we could create `instrDesc` if a constant in the instruction is 12-bits wide. Now, that space has shrunk to 9-bits. A constant beyond that size, we will have to use `instrDescCns` which is 24 bytes big. That causes TP regression because there are extra code paths involved to process `instrDescCns`. It does increase the JIT memory size for constants in the range of 10-bits thru 12-bits. Alternatives are costly, see below.

> In #94310, instead of increasing the variables, what happens if you instead add padding? Eg:
> 
> ```
> instruction _idIns : 10;
> instruction _pad_idIns : 1;
> ```
> 
> etc for all the other fields. That would confirm if the extra cost is due to the overall size of `instrDesc` or if it is the increased field sizes.
> 
> Thinking ahead, SME is going to add a similar slowdown.
> 
> What about if SVE instruction enum started again from 0. The same with all the instruction formats. Treat it as a separate architecture. You could have a single bit in `instrDesc` to indicate if it is Arm64 or Arm64Sve (and, then maybe increase to 2bits when doing SME). Or override instrDesc, creating an instrDescSve, and then emitNewInstr() can take in an enum which decides which instrDesc to allocate. 

`emitNewInstr()` is a hot method and adding a single check will impact TP heavily. @TIHan tried doing that a while back and found out it is costly. If not in `emitNewInstrDesc`, I was thinking of having a `union` or something of `idIns` and `idFmt`, but even for that we need a bit to tell if it is SVE or not and then code *somewhere* and possibly in a hot-method if the `idIns` represents sve or not. With the existing model, it is ok to take the minimal TP cost, but overall it doesn't impact any non-sve code. I am hoping the same once we start adding more cases in common methods like `emitOutputInstr()`, etc.

> You'd have to have different `emitIns_R_R_R()` etc functions for sve. But, that's ok because it's keeping the switch statements shorter.

That will be lot of methods though.

src/coreclr/jit/emitarm64.cpp
```diff
@@ -2294,6 +2306,1470 @@ emitter::code_t emitter::emitInsCode(instruction ins, insFormat fmt)
     return code;
 }
 
+/*****************************************************************************
+ *
+ *  Returns the specific encoding of the given CPU instruction and format
+ */
+
+emitter::code_t emitter::emitSveInsCode(instruction ins, insFormat fmt)
```
BruceForstall, line 2314: nit: `emitInsCodeSve` instead?

src/coreclr/jit/emitfmtsarm64_sve.h
```diff
@@ -0,0 +1,514 @@
...
+IF_DEF(SVE_BB_2A,   IS_NONE, NONE) // SVE_BB_2A  ...........nnnnn .....iiiiiiddddd  -- sve_int_arith_vl
+IF_DEF(SVE_BC_1A,   IS_NONE, NONE) // SVE_BC_1A  ................ .....iiiiiiddddd  -- sve_int_read_vl_a
+IF_DEF(SVE_BD_3A,   IS_NONE, NONE) // SVE_BD_3A  ........xx.mmmmm ......nnnnnddddd  -- sve_int_mul_b
+IF_DEF(SVE_BD_3B,   IS_NONE, NONE) // SVE_BD_3B  ...........mmmmm ......nnnnnddddd  -- sve_int_mul_b
+IF_DEF(SVE_BE_3A,   IS_NONE, NONE) // SVE_BE_3A  ........xx.mmmmm ......nnnnnddddd  -- sve_int_sqdmulh
+IF_DEF(SVE_BF_2A,   IS_NONE, NONE) // SVE_BF_2A  ........xx.xxiii ......nnnnnddddd  -- sve_int_bin_cons_shift_b
+IF_DEF(SVE_BG_3A,   IS_NONE, NONE) // SVE_BG_3A  ........xx.mmmmm ......nnnnnddddd  -- sve_int_bin_cons_shift_a
+IF_DEF(SVE_BH_3A,   IS_NONE, NONE) // SVE_BH_3A  .........x.mmmmm ....hhnnnnnddddd  -- sve_int_bin_cons_misc_0_a
+IF_DEF(SVE_BH_3B,   IS_NONE, NONE) // SVE_BH_3B  ...........mmmmm ....hhnnnnnddddd  -- sve_int_bin_cons_misc_0_a
+IF_DEF(SVE_BH_3B_A, IS_NONE, NONE) // SVE_BH_3B_A  ...........mmmmm ....hhnnnnnddddd  --
```
BruceForstall, line 170: Why do some forms, like this one, have an extra discriminator (here, `_A`) at the end? This looks identical in pattern to the preceding one.

src/coreclr/jit/emitfmtsarm64_sve.h
```diff
@@ -0,0 +1,514 @@
...
+
+/*****************************************************************************
+*           SVE_WW_XY
+*           WW      -- code
+*           X       -- register count
+*           Y       -- unique id
+
+*****************************************************************************/
+
+IF_DEF(SVE_AA_3A,   IS_NONE, NONE) // SVE_AA_3A  ........xx...... ...gggmmmmmddddd  -- sve_int_bin_pred_log
```
BruceForstall, line 132: What is the comment at the end? Does that come from the XML? Is it one example instruction from the XML? How do we read the name?

src/coreclr/jit/emitfmtsarm64_sve.h
```diff
@@ -0,0 +1,514 @@
...
+IF_DEF(SVE_2BQ, IS_NONE, NONE) // Instruction has 2  possible encoding types, type BQ
+IF_DEF(SVE_2BR, IS_NONE, NONE) // Instruction has 2  possible encoding types, type BR
+IF_DEF(SVE_2BS, IS_NONE, NONE) // Instruction has 2  possible encoding types, type BS
+
+/*****************************************************************************
+*           SVE_WW_XY
+*           WW      -- code
+*           X       -- register count
+*           Y       -- unique id
+
```
BruceForstall, line 129: Missing leading `*`?
```suggestion
*
```

src/coreclr/jit/emitfmtsarm64_sve.h
```diff
@@ -0,0 +1,514 @@
...
+*****************************************************************************/
```
BruceForstall, line 130: Can you document what the letters are in the example instructions? e.g., `ggg - predicate register specification`, `mmmmm` - register in `Xm` position, etc.

BruceForstall (approved):
Looks really good to me. You can follow up with my comments/questions in a follow-up if desired.

src/coreclr/jit/CMakeLists.txt
```diff
@@ -398,8 +398,10 @@ set( JIT_I386_HEADERS ${JIT_AMD64_HEADERS} )
 set( JIT_ARM64_HEADERS
     emitarm64.h
     emitfmtsarm64.h
+    emitfmtsarm64_sve.h
```
a74nh, line 401: I don't see any other files with _ in the name. Should this be  emitfmtsarm64sve.h ?

src/coreclr/jit/emit.h
```diff
@@ -722,10 +728,13 @@ class emitter
...
                                   // doesn't cross a byte boundary.
-#elif defined(TARGET_ARM64) || defined(TARGET_LOONGARCH64) || defined(TARGET_RISCV64)
+#elif defined(TARGET_ARM64)
+        opSize  _idOpSize : 3; // operand size: 0=1 , 1=2 , 2=4 , 3=8, 4=16
+        insOpts _idInsOpt : 6; // options for instructions
+#elif defined(TARGET_LOONGARCH64) || defined(TARGET_RISCV64)
 /* _idOpSize defined below. */
 #else
-        opSize _idOpSize : 2; // operand size: 0=1 , 1=2 , 2=4 , 3=8
+        opSize    _idOpSize : 2; // operand size: 0=1 , 1=2 , 2=4 , 3=8
```
a74nh, line 737: I assume this space is so that it matches the generic version on line 722/728. But, the Arm64 versions on 732 and 733 don't match.

src/coreclr/jit/instrsarm64_sve.h
```diff
@@ -0,0 +1,2873 @@
...
+    // TRN1    <Zd>.Q, <Zn>.Q, <Zm>.Q                                                    SVE_BR_3B           00000101101mmmmm 000110nnnnnddddd     05A0 1800   
+    // TRN1    <Pd>.<T>, <Pn>.<T>, <Pm>.<T>                                              SVE_CI_3A           00000101xx10MMMM 0101000NNNN0DDDD     0520 5000   
+
+INST3(trn2,              "trn2",                  0,                       IF_SVE_3D,               0x05207400,      0x05A01C00,      0x05205400       )
+    // TRN2    <Zd>.<T>, <Zn>.<T>, <Zm>.<T>                                              SVE_BR_3A           00000101xx1mmmmm 011101nnnnnddddd     0520 7400   
+    // TRN2    <Zd>.Q, <Zn>.Q, <Zm>.Q                                                    SVE_BR_3B           00000101101mmmmm 000111nnnnnddddd     05A0 1C00   
+    // TRN2    <Pd>.<T>, <Pn>.<T>, <Pm>.<T>                                              SVE_CI_3A           00000101xx10MMMM 0101010NNNN0DDDD     0520 5400   
+
+INST3(uzp1,              "uzp1",                  0,                       IF_SVE_3D,               0x05206800,      0x05A00800,      0x05204800       )
```
a74nh, line 622: Comparing this against the UZP1 I manually added https://github.com/dotnet/runtime/pull/93067

I'm a little confused by `00000101xx1mmmmm` I can't see what bit pattern that refers to.

src/coreclr/jit/emitarm64.cpp
```diff
@@ -2294,6 +2306,1470 @@ emitter::code_t emitter::emitInsCode(instruction ins, insFormat fmt)
...
+            break;
+    }
+
+    assert(encoding_found);
+
+    switch (index)
+    {
+        case 0:
+            assert(ins < ArrLen(insCodes1));
+            code = insCodes1[ins];
```
a74nh, line 3716: To avoid this switch, there could be a NxM array insCodes which each insCodes[0] is a pointer to insCodes1,  insCodes[1] is a pointer to  insCodes2 etc

Then instead of the switch you can access via insCodes[index][ins].

src/coreclr/jit/emitfmtsarm64_sve.h
```diff
@@ -0,0 +1,514 @@
...
+IF_DEF(SVE_EK_3A,   IS_NONE, NONE) // SVE_EK_3A  ........xx.mmmmm ....rrnnnnnddddd  -- sve_intx_cmla
+IF_DEF(SVE_EL_3A,   IS_NONE, NONE) // SVE_EL_3A  ........xx.mmmmm ......nnnnnddddd  -- sve_intx_mlal_long
+IF_DEF(SVE_EM_3A,   IS_NONE, NONE) // SVE_EM_3A  ........xx.mmmmm ......nnnnnddddd  -- sve_intx_qrdmlah
+IF_DEF(SVE_EN_3A,   IS_NONE, NONE) // SVE_EN_3A  ........xx.mmmmm ......nnnnnddddd  -- sve_intx_qdmlalbt
+IF_DEF(SVE_EO_3A,   IS_NONE, NONE) // SVE_EO_3A  ........xx.mmmmm ......nnnnnddddd  -- sve_intx_qdmlal_long
+IF_DEF(SVE_EP_3A,   IS_NONE, NONE) // SVE_EP_3A  ........xx...... ...gggmmmmmddddd  -- sve_intx_pred_arith_binary
+IF_DEF(SVE_EQ_3A,   IS_NONE, NONE) // SVE_EQ_3A  ........xx...... ...gggnnnnnddddd  -- sve_intx_accumulate_long_pairs
+IF_DEF(SVE_ER_3A,   IS_NONE, NONE) // SVE_ER_3A  ........xx...... ...gggmmmmmddddd  -- sve_intx_arith_binary_pairs
+IF_DEF(SVE_ES_3A,   IS_NONE, NONE) // SVE_ES_3A  ........xx...... ...gggnnnnnddddd  -- sve_intx_pred_arith_unary
+IF_DEF(SVE_ET_3A,   IS_NONE, NONE) // SVE_ET_3A  ........xx...... ...gggmmmmmddddd  -- sve_intx_pred_arith_binary_sat
```
a74nh, line 283: SVE_ET_3A and SVE_EU_3A look identical?

src/coreclr/jit/emitfmtsarm64_sve.h
kunalspathak: Yes, they share the same "instruction format" as in ` ...........mmmmm ...gggnnnnnttttt`, but the actual opcode i.e. 0 and 1 are different for them. You can see them in `instrsarm64_sve.h`:

```
    // ST1D    {<Zt>.D }, <Pg>, [<Xn|SP>, <Xm>, LSL #3]                                  SVE_JD_4C           11100101111mmmmm 010gggnnnnnttttt     E5E0 4000   
    // ST1D    {<Zt>.Q }, <Pg>, [<Xn|SP>, <Xm>, LSL #3]                                  SVE_JD_4C_A         11100101110mmmmm 010gggnnnnnttttt     E5C0 4000   

```

It is possible to combine them to be something like ` ..........Xmmmmm ...gggnnnnnttttt`, where `X` in this pattern is the differing bit, but the tool won't know if logically they can be combined or not, so I kept it simple and gave it a new name. The reason for appending `_A` or `_B` at the end is so we can having unique format names with the instruction. We cannot have duplicate format names for any instruction because while searching, we won't know which of the formats we wanted to get hex code for.

src/coreclr/jit/emitfmtsarm64_sve.h
kunalspathak: Thanks for pointing it out. I was using the code names and now, using the description. Hope they serve better purpose.

src/coreclr/jit/emitfmtsarm64_sve.h
kunalspathak: I don't see such a description for existing file, but added few entries here.

src/coreclr/jit/emitfmtsarm64_sve.h
kunalspathak: Yes, they could be and so are other. They are given unique names based on "format pattern + SVE group class", so there might be format names that represent same format, but different group (SVE_ET_3A = `SVE2 saturating add/subtract` vs. SVE_EU_3A  = `SVE2 saturating/rounding bitwise shift left (predicated)`). They can (will) be hand combined once we start implementing the encoding logic.

src/coreclr/jit/instrsarm64_sve.h
kunalspathak: The changes [here](https://github.com/dotnet/runtime/pull/94285/files#diff-5e27623edecb939f3991300912bd8e69ca0746ad1cf4fc0dc3bbb5c761f2e029R623-R625) matches exactly to what you had in your PR [here](https://github.com/dotnet/runtime/pull/93067/files#diff-534bddc341b6c03fb9519adcc7a1d25c6d4fc2f66605cc1bae741f27a34e0212R1685-R1686).

The extra encoding `00000101101mmmmm 000010nnnnnddddd     05A0 0800 ` present is for `Even(quadwords)`:

image

src/coreclr/jit/emitarm64.cpp
kunalspathak: Yes, I think there are multiple ways to optimize this (as we spoke offline). I will probably have them as a separate change.

src/coreclr/jit/instrsarm64_sve.h
kunalspathak: Let me know if that answers your question.

src/coreclr/jit/emit.h
kunalspathak: May be I am not getting the line numbers correctly so not following your question. But these 2 fields were moved higher up to compact the data within byte boundary. If I don't do this, `_idSmallCns` was overflowing to `16th` byte.

src/coreclr/jit/emitfmtsarm64_sve.h
BruceForstall: I presume, if the tool can't automatically merge them, we would want a developer coming in afterwards to manually merge them and add the `X` (for example) bit? At least in this case?

src/coreclr/jit/instrsarm64_sve.h
a74nh: Got it, missed that one.

src/coreclr/jit/emit.h
a74nh: This was just formatting nit. I meant the literal space characters between "opSize" and "_idOpSize"

src/coreclr/jit/emitfmtsarm64_sve.h
kunalspathak: That's correct. So for an instruction, if there are 2 formats `SVE_AA` and `SVE_AB` that are worth merging into `SVE_AA`, we do not have to delete the entry of `SVE_AB` because it can be still used for other instructions. Once there are no usage of `SVE_AB`, we can delete its entry from format table.

Issue #91704, opened 2023-09-06
Unify RuntimeType with the shared RuntimeType


MichalStrehovsky:
Optimistically placing this in 9.0 milestone.

We should unify the reflection stack with the stack shared with Mono/CoreCLR-JIT.

Things we can enable once that's done:

- [ ] Roll back #91660. Instead of doing the extra analysis, we should no longer need to generate any metadata for unconstructed MethodTables (basically roll back EETypeNode/ConstructedEETypeNode/etc. to how it looked before #85810 that triggered the bugfix)
- [ ] Preinitialization of System.RuntimeType instances and related RyuJIT features that can directly reference System.RuntimeType addresses.
- [ ] Allocate lazily created System.RuntimeType instances on frozen heap

MichalPetryka:
Wouldn't this require NativeAOT to get rid od the RuntimeType subclasses for this?

MichalStrehovsky:
> Wouldn't this require NativeAOT to get rid of the RuntimeType subclasses for this?

Yes, that's the intention.

jkotas:
> Preinitialization of System.RuntimeType instances and related RyuJIT features that can directly reference System.RuntimeType addresses.

@MichalStrehovsky Feel free to pick this up and enable this optimization in ilc. It should be pretty straightforward now that RuntimeType is prepared for it. I do not think that I am going to have time to do it anytime soon.

EgorBo:
I can help with the JIT side once RuntimeType become frozen, ideally it's a matter of implementing some, currently no-op, JIT-VM apis and might need some JIT changes as well for relocable constants

MichalStrehovsky:
> > Preinitialization of System.RuntimeType instances and related RyuJIT features that can directly reference System.RuntimeType addresses.
> 
> @MichalStrehovsky Feel free to pick this up and enable this optimization in ilc. It should be pretty straightforward now that RuntimeType is prepared for it. I do not think that I am going to have time to do it anytime soon.

Looking into it!

Thanks for the help offer Egor!

MichalStrehovsky:
This has been resolved.

PR #94287, opened 2023-11-02
Restore EETypeNode to the state before we deleted reflection blocking

MichalStrehovsky:
Contributes to #91704.

When we deleted reflection blocking in #85810 we had to update `EETypeNode`/`ConstructedEETypeNode` to ensure any `MethodTable` gets metadata (constructed or not). This was needed because of how reflection was structured - we couldn't create a `RuntimeType` without knowing about the metadata. After the refactor in #93440, metadata is no longer a prerequisite to constructing a `RuntimeType`.

The compiler can go back to optimizing away the metadata. This affects things like `typeof(Foo) == bar.GetType()`. No metadata is needed for this (and we do optimized the `Foo` MethodTable to unconstructed one) but we still had to generate metadata for reflection stack sake.

Besides the rollback of EEType to the previous shape, this also has a bugfix for #91988 that was found later - interface types used in cast/dispatch should be considered constructed.

I'm seeing 0.1 - 0.7% size saving. Considering interfaces constructed doesn't result in pretty much any regressions and kind of makes sense (they don't have vtables and their interface list is not much cost).

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Compilation.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/Compilation.cs
@@ -301,9 +301,16 @@ namespace ILCompiler
                 case ReadyToRunHelperId.TypeHandleForCasting:
                     {
                         var type = (TypeDesc)targetOfLookup;
+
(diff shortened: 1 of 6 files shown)


jkotas:
Asserts in checked build

jkotas (approved):
Thanks!

Issue #73650, opened 2022-08-09
Use CLOCK_MONOTONIC for mutex waiting on iOS/macOS


akoeplinger:
See https://github.com/dotnet/runtime/issues/58737 and related PRs.

`CLOCK_MONOTONIC` is available on ios 10+, but not `pthread_condattr_setclock`. We disable the code on Apple platforms using `BROKEN_CLOCK_SOURCE` in https://github.com/dotnet/runtime/blob/4211dc40d06195682d4b0fc9a220bdb75732d720/src/mono/mono/utils/mono-os-mutex.h#L39-L41

We should look into using `pthread_cond_timedwait` with a `CLOCK_MONOTONIC` argument or alternatively enable `clock_gettime` (supported by newer macOS/iOS releases) on Apple platforms.

SamMonoRT:
Nice to have, but not blocking release. Moving to 9.0.0

ivanpovazan:
Nice to have, but not blocking release. Moving to 10.0.0

PR #94298, opened 2023-11-02
Enable aligned_alloc on macOS

### LINKED ISSUE #54296: Update `SystemNative_AlignedAlloc` to properly check `aligned_alloc` availability for Apple targets
https://github.com/dotnet/runtime/pull/54006 introduced new `NativeMemory` APIs including one that wraps `aligned_alloc` where available.

On Apple targets this is only available on 10.15 or later but we currently target 10.13 and as such have a `#if HAVE_ALIGNED_ALLOC && !defined(__APPLE__)` check.

Ideally we would update this to correctly check for the availability of `aligned_alloc` (either via CMake magic or some existing Apple specific define) so we can light up on iOS, OSX, and Apple Silicon.

jkotas:
Fixes #54296

### DIFF
--- a/src/native/libs/configure.cmake
+++ b/src/native/libs/configure.cmake
@@ -581,15 +581,15 @@ if(CLR_CMAKE_TARGET_IOS)
     # Manually set results from check_c_source_runs() since it's not possible to actually run it during CMake configure checking
     unset(HAVE_SHM_OPEN_THAT_WORKS_WELL_ENOUGH_WITH_MMAP)
     unset(HAVE_ALIGNED_ALLOC)   # only exists on iOS 13+
-    set(HAVE_CLOCK_MONOTONIC)
-    set(HAVE_CLOCK_REALTIME)
+    set(HAVE_CLOCK_MONOTONIC 1)
+    set(HAVE_CLOCK_REALTIME 1)
     unset(HAVE_FORK) # exists but blocked by kernel
 elseif(CLR_CMAKE_TARGET_MACCATALYST)
     # Manually set results from check_c_source_runs() since it's not possible to actually run it during CMake configure checking
     unset(HAVE_SHM_OPEN_THAT_WORKS_WELL_ENOUGH_WITH_MMAP)
     unset(HAVE_ALIGNED_ALLOC)   # only exists on iOS 13+
-    set(HAVE_CLOCK_MONOTONIC)
-    set(HAVE_CLOCK_REALTIME)
+    set(HAVE_CLOCK_MONOTONIC 1)
+    set(HAVE_CLOCK_REALTIME 1)
     unset(HAVE_FORK) # exists but blocked by kernel
 elseif(CLR_CMAKE_TARGET_TVOS)
     # Manually set results from check_c_source_runs() since it's not possible to actually run it during CMake configure checking
--- a/src/native/libs/configure.cmake
+++ b/src/native/libs/configure.cmake
@@ -595,8 +595,8 @@ elseif(CLR_CMAKE_TARGET_TVOS)
     # Manually set results from check_c_source_runs() since it's not possible to actually run it during CMake configure checking
     unset(HAVE_SHM_OPEN_THAT_WORKS_WELL_ENOUGH_WITH_MMAP)
     unset(HAVE_ALIGNED_ALLOC)   # only exists on iOS 13+
-    set(HAVE_CLOCK_MONOTONIC)
-    set(HAVE_CLOCK_REALTIME)
+    set(HAVE_CLOCK_MONOTONIC 1)
+    set(HAVE_CLOCK_REALTIME 1)
     unset(HAVE_FORK) # exists but blocked by kernel
 elseif(CLR_CMAKE_TARGET_ANDROID)
     # Manually set results from check_c_source_runs() since it's not possible to actually run it during CMake configure checking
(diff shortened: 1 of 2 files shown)


src/native/libs/configure.cmake
```diff
@@ -581,23 +581,22 @@ if(CLR_CMAKE_TARGET_IOS)
     # Manually set results from check_c_source_runs() since it's not possible to actually run it during CMake configure checking
     unset(HAVE_SHM_OPEN_THAT_WORKS_WELL_ENOUGH_WITH_MMAP)
     unset(HAVE_ALIGNED_ALLOC)   # only exists on iOS 13+
```
jkotas, line 583: @kotlarmilos What is our current minimum support iOS version? Can this block be deleted?

src/native/libs/configure.cmake
```diff
@@ -608,14 +607,10 @@ elseif(CLR_CMAKE_TARGET_ANDROID)
 elseif(CLR_CMAKE_TARGET_BROWSER OR CLR_CMAKE_TARGET_WASI)
     set(HAVE_FORK 0)
 else()
-    if(CLR_CMAKE_TARGET_OSX)
-        unset(HAVE_ALIGNED_ALLOC) # only exists on OSX 10.15+
```
jkotas, line 612: Our minimum supported macOS version is 10.15 so this block is no longer needed.

src/native/libs/configure.cmake
tannergooding: Do we know why we support such an old version of iOS? Apple starts considering them vintage after they've no longer been on sale for 5 years and obsolete after 7 years: https://support.apple.com/en-us/HT201624

The iPhone 6 is the newest vintage phone supported and it supports iOS 12.5.7 as its latest.
While the iPhone 7 is the oldest non-vintage phone, and that shipped with iOS 10 but currently runs iOS 15.8

For macOS, 10.15 hasn't had a security update in 15 months (https://support.apple.com/en-us/102769). While 11 still got one less than 2 months ago (https://support.apple.com/en-us/HT211896#macos11710)

kotlarmilos: I believe it depends on Xamarin/Maui versioning. @rolfbjarne What is the versioning policy for Xamarin?

steveisok: We had discussed bumping the min version to 12 in .net 9. Unless @akoeplinger found some kind of holdup?

jkotas: I have opened https://github.com/dotnet/runtime/issues/94345 on this.

jkotas:
@kotlarmilos @ivanpovazan Could you please review and approve?

src/native/libs/configure.cmake
```diff
@@ -581,23 +581,22 @@ if(CLR_CMAKE_TARGET_IOS)
     # Manually set results from check_c_source_runs() since it's not possible to actually run it during CMake configure checking
     unset(HAVE_SHM_OPEN_THAT_WORKS_WELL_ENOUGH_WITH_MMAP)
     unset(HAVE_ALIGNED_ALLOC)   # only exists on iOS 13+
-    unset(HAVE_CLOCK_MONOTONIC) # only exists on iOS 10+
-    unset(HAVE_CLOCK_REALTIME)  # only exists on iOS 10+
+    set(HAVE_CLOCK_MONOTONIC)
```
ivanpovazan, line 584: @akoeplinger could this change cause issues related to: https://github.com/dotnet/runtime/issues/73650

src/native/libs/configure.cmake
```diff
@@ -581,23 +581,22 @@ if(CLR_CMAKE_TARGET_IOS)
...
+    set(HAVE_CLOCK_REALTIME)
     unset(HAVE_FORK) # exists but blocked by kernel
 elseif(CLR_CMAKE_TARGET_TVOS)
     # Manually set results from check_c_source_runs() since it's not possible to actually run it during CMake configure checking
     unset(HAVE_SHM_OPEN_THAT_WORKS_WELL_ENOUGH_WITH_MMAP)
     unset(HAVE_ALIGNED_ALLOC)   # only exists on iOS 13+
-    unset(HAVE_CLOCK_MONOTONIC) # only exists on iOS 10+
-    unset(HAVE_CLOCK_REALTIME)  # only exists on iOS 10+
+    set(HAVE_CLOCK_MONOTONIC)
+    set(HAVE_CLOCK_REALTIME)
```
am11, line 599: Similar to Android block below, I think these should be:

```suggestion
    set(HAVE_CLOCK_MONOTONIC 1)
    set(HAVE_CLOCK_REALTIME 1)
```

(since we are using `#if HAVE_CLOCK_MONOTONIC` etc. and not `#ifdef HAVE_CLOCK_MONOTONIC`)

src/native/libs/configure.cmake
```diff
@@ -581,23 +581,22 @@ if(CLR_CMAKE_TARGET_IOS)
...
```
jkotas, line 585:
```suggestion
    set(HAVE_CLOCK_MONOTONIC 1)
    set(HAVE_CLOCK_REALTIME 1)
```

src/native/libs/configure.cmake
```diff
@@ -581,23 +581,22 @@ if(CLR_CMAKE_TARGET_IOS)
...
     unset(HAVE_FORK) # exists but blocked by kernel
 elseif(CLR_CMAKE_TARGET_MACCATALYST)
     # Manually set results from check_c_source_runs() since it's not possible to actually run it during CMake configure checking
-    # TODO: test to see if these all actually hold true on Mac Catalyst
     unset(HAVE_SHM_OPEN_THAT_WORKS_WELL_ENOUGH_WITH_MMAP)
     unset(HAVE_ALIGNED_ALLOC)   # only exists on iOS 13+
-    unset(HAVE_CLOCK_MONOTONIC) # only exists on iOS 10+
-    unset(HAVE_CLOCK_REALTIME)  # only exists on iOS 10+
+    set(HAVE_CLOCK_MONOTONIC)
+    set(HAVE_CLOCK_REALTIME)
```
jkotas, line 592:
```suggestion
    set(HAVE_CLOCK_MONOTONIC 1)
    set(HAVE_CLOCK_REALTIME 1)
```

kotlarmilos (approved):

PR #90771, opened 2023-08-17
[wasm][debugger] Trying to fix MT debugging experience.

thaystg:
I'm doing this workaround to avoid memory allocation while debugging a MT wasm app, it looks like there is some failure while updating the memory in all the workers while some of them are paused.
This will fix the CI and also improve the customer experience.

PR #94299, opened 2023-11-02
[browser][MT] fix stale memory on suspended thread

pavelsavara:
Alternative to https://github.com/dotnet/runtime/pull/90771

Problem: when WebWorker is suspended in the browser, the other running threads could `grow` the linear memory in the meantime. After the thread is un-suspended it may to try to de-reference pointer which is beyond it's known view. This is likely V8 bug.

We believe that described problem is cause of MT problems with debugger tests, example [log](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-93920-merge-f5727b81f7c54bb190/chrome-DebuggerTests.PointerTests/1/console.06520613.log?helixlogtype=result)

### DIFF
--- a/src/mono/wasm/runtime/memory.ts
+++ b/src/mono/wasm/runtime/memory.ts
@@ -408,3 +408,27 @@ export function isSharedArrayBuffer(buffer: any): buffer is SharedArrayBuffer {
     // See also https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag
     return sharedArrayBufferDefined && buffer[Symbol.toStringTag] === "SharedArrayBuffer";
 }
+
+/*
+Problem: When WebWorker is suspended in the browser, the other running threads could `grow` the linear memory in the meantime. 
+After the thread is un-suspended C code may to try to de-reference pointer which is beyond it's known view. 
+This is likely V8 bug. We don't have direct evidence, just failed debugger unit tests with MT runtime.
+*/
+export function forceThreadMemoryViewRefresh() {
+    // this condition should be eliminated by rollup on non-threading builds and it would become empty method.
+    if (!MonoWasmThreads) return;
+
+    const wasmMemory = Module.getMemory();
+
+    /*
+    Normally when wasm memory grows in v8, this size change is broadcast to other web workers via an 'interrupt', which works by setting a thread-local flag that needs to be checked. 
+    It's possible that at this point in execution the flag has not been checked yet (because this worker was suspended by the debugger in an unknown location), 
+    which means the size change has not taken effect in this worker. 
+    wasmMemory.grow's implementation in v8 checks to see whether other workers have already grown the buffer, 
+    and will update the current worker's knowledge of the buffer's size. 
+    After that we should be able to safely updateMemoryViews and get a correctly sized view. 
+    This only works because their implementation does not skip doing work even when you ask to grow by 0 pages.
+    */
+    wasmMemory.grow(0);
+    runtimeHelpers.updateMemoryViews();
+}
--- a/eng/pipelines/extra-platforms/runtime-extra-platforms-wasm.yml
+++ b/eng/pipelines/extra-platforms/runtime-extra-platforms-wasm.yml
@@ -262,7 +262,6 @@ jobs:
       isExtraPlatformsBuild: ${{ parameters.isExtraPlatformsBuild }}
       isWasmOnlyBuild: ${{ parameters.isWasmOnlyBuild }}
       runOnlyOnWasmOnlyPipelines: true
-      shouldContinueOnError: true
 
   # Disable for now
   #- template: /eng/pipelines/coreclr/perf-wasm-jobs.yml
--- a/src/mono/wasm/debugger/DebuggerTestSuite/AssignmentTests.cs
+++ b/src/mono/wasm/debugger/DebuggerTestSuite/AssignmentTests.cs
@@ -50,7 +50,6 @@ namespace DebuggerTests
 
         [ConditionalTheory(nameof(RunningOnChrome))]
         [MemberData("GetTestData")]
-        [ActiveIssue("https://github.com/dotnet/runtime/issues/86496", typeof(DebuggerTests), nameof(DebuggerTests.WasmMultiThreaded))]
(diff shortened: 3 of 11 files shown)


pavelsavara:
This helps
image

src/mono/wasm/runtime/memory.ts
```diff
@@ -408,3 +408,11 @@ export function isSharedArrayBuffer(buffer: any): buffer is SharedArrayBuffer {
     // See also https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag
     return sharedArrayBufferDefined && buffer[Symbol.toStringTag] === "SharedArrayBuffer";
 }
+
+export function forceThreadMemoryViewRefresh() {
+    // this condition should be eliminated by rollup on non-threading builds
+    if (!MonoWasmThreads) return false;
```
ilonatommy, line 414: Why false when we never check the returned value?
```suggestion
    if (!MonoWasmThreads) return;
```

src/mono/wasm/runtime/memory.ts
```diff
@@ -408,3 +408,11 @@ export function isSharedArrayBuffer(buffer: any): buffer is SharedArrayBuffer {
...
+    const wasmMemory = Module.getMemory();
+    wasmMemory.grow(0);
+    runtimeHelpers.updateMemoryViews();
```
ilonatommy, line 417: Without updating the effect does not get propagated? What is the role of update exactly?

ilonatommy (approved):

src/mono/wasm/runtime/memory.ts
ilonatommy: Ok, I see, it's how they do it, the re-allocate the heaps: https://github.com/emscripten-core/emscripten/blob/a2f3b925c967b9a8091b5b52fa8cbb649cfe248e/src/preamble.js#L146

pavelsavara: `wasmMemory.grow` always invalidates existing views

src/mono/wasm/runtime/memory.ts
```diff
@@ -408,3 +408,17 @@ export function isSharedArrayBuffer(buffer: any): buffer is SharedArrayBuffer {
...
+Problem: When WebWorker is suspended in the browser, the other running threads could `grow` the linear memory in the meantime. 
+After the thread is un-suspended C code may to try to de-reference pointer which is beyond it's known view. 
+This is likely V8 bug. We don't have direct evidence, just failed debugger unit tests with MT runtime.
+*/
+export function forceThreadMemoryViewRefresh() {
+    // this condition should be eliminated by rollup on non-threading builds and it would become empty method.
+    if (!MonoWasmThreads) return;
+
+    const wasmMemory = Module.getMemory();
+    wasmMemory.grow(0);
```
kg, line 422: Please add a comment explaining the significance of grow(0) and why we're doing it, i.e.

"Normally when wasm memory grows in v8, this size change is broadcast to other web workers via an 'interrupt', which works by setting a thread-local flag that needs to be checked. It's possible that at this point in execution the flag has not been checked yet (because this worker was suspended by the debugger in an unknown location), which means the size change has not taken effect in this worker. wasmMemory.grow's implementation in v8 checks to see whether other workers have already grown the buffer, and will update the current worker's knowledge of the buffer's size. After that we should be able to safely updateMemoryViews and get a correctly sized view. This only works because their implementation does not skip doing work even when you ask to grow by 0 pages."

pavelsavara:
@thaystg I will wait for your feedback/approval

pavelsavara:
Hmm, so now all errors are `"TypeError: Failed to fetch\n    at Object.L [as fetch_like] (http://localhost:9400/_framework/dotnet.js:3:4721)\n    at http://localhost:9400/_framework/dotnet.js:3:12002\n    at http://localhost:9400/_framework/dotnet.js:3:12018"` [Log](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-94299-merge-c1b09c0202b6489e91/chrome-DebuggerTests.AsyncTests/1/console.0ca4bc40.log?helixlogtype=result)


@thaystg does that mean my fix didn't help ? Or it means there is another issue ?

Is debugger making the app to download something ?

Does suspending thread while fetching break the fetch ?

thaystg:
There are a lot of core dumps generated: chrome is crashing, not sure if it's related to your fix.

image

thaystg:
I also remember to have already tried it and it didn't help. But for me it's okay to try again. :)

pavelsavara:
> I think there are a lot of core dumps generated: chrome is crashing, not sure if it's related to your fix.

Where is "there" ?

More details would help.

> I also remember to have already tried it and it didn't help. But for me it's okay to try again. :)

Could you please elaborate on what you tried ?

thaystg:
> > I think there are a lot of core dumps generated: chrome is crashing, not sure if it's related to your fix.
> 
> Where is "there" ?
> 
> More details would help.
> 
> > I also remember to have already tried it and it didn't help. But for me it's okay to try again. :)
> 
> Could you please elaborate on what you tried ?

Look at artifacts tab here: https://dev.azure.com/dnceng-public/public/_build/results?buildId=459235&view=ms.vss-test-web.build-test-results-tab&runId=10361536&resultId=100252&paneView=dotnet-dnceng.dnceng-build-release-tasks.helix-test-information-tab

image

thaystg:
> Could you please elaborate on what you tried ?

Unfortunately I don't remember details of what I tested, I tried a lot of things before the workaround in that PR. 
If I can help you in anyway by testing your solution or anything else, please let me know.

pavelsavara:
image

pavelsavara:
I think that the fix for memory views is good. Solid 10/10 
image


There are problems with https://github.com/dotnet/runtime/issues/89409 but that's unrelated issue
```
The active test run was aborted. Reason: Test host process crashed
Results File: /root/helix/work/workitem/uploads/xharness-output/logs/testResults.trx

Passed!  - Failed:     0, Passed:    80, Skipped:     0, Total:    80, Duration: 9 m - DebuggerTestSuite.dll (net8.0)
Test Run Aborted.
```

thaystg:
@pavelsavara thanks a lot and congrats for this fix!!!

PR #94308, opened 2023-11-02
Remove [EditorBrowsable(Never)] from [InlineArray], add XML docs

### LINKED ISSUE #94283: [InlineArray] should not have [EditorBrowsableState(Never)]
### Description

The `[InlineArray]` type is marked with `[EditorBrowsable(EditorBrowsableState.Never)]`:

https://github.com/dotnet/runtime/blob/655b177da2e423a04a60a331bc6c11fe5ec427c3/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/InlineArrayAttribute.cs#L11-L13

This seems... Wrong? There's no built-in C# syntax for inline arrays for now, so the attribute has to be used explicitly in user code. The fact it's hidden makes this clunkier than it needs to be. The attribute is already buried in a namespace that's explicitly documented as "unsafe" and only for advanced users, so this feels unnecessary.

Shouldn't we just remove this?
And if so, I take it the ship has sailed for .NET 8 GA at this point?

Sergio0694:
### Overview

This PR includes two changes to `[InlineArray]`:
- Removes `[EditorBrowsable(EditorBrowsableState.Never)]`
- Adds some XML remarks, as well as all missing XML docs

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/InlineArrayAttribute.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/InlineArrayAttribute.cs
@@ -6,14 +6,37 @@ namespace System.Runtime.CompilerServices
     /// <summary>
     /// Indicates that the instance's storage is sequentially replicated "length" times.
     /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This attribute can be used to annotate a <see langword="struct"/> type with a single field.
+    /// The runtime will replicate that field as many times as specified in the actual type layout.
+    /// </para>
+    /// <para>
+    /// Here's an example of how an inline array type with 8 <see cref="float"/> values can be declared:
+    /// <code lang="csharp">
+    /// [InlineArray(8)]
+    /// struct Float8InlineArray
+    /// {
+    ///     private float _value;
+    /// }
+    /// </code>
+    /// </para>
+    /// <para>
+    /// The compiler will also automatically generate an indexer property, and support casting the type
+    /// to <see cref="Span{T}"/> and <see cref="ReadOnlySpan{T}"/> for safe enumeration of all elements.
+    /// </para>
+    /// </remarks>
     [AttributeUsage(AttributeTargets.Struct, AllowMultiple = false)]
     public sealed class InlineArrayAttribute : Attribute
     {
+        /// <summary>Creates a new <see cref="InlineArrayAttribute"/> instance with the specified parameters.</summary>
+        /// <param name="length">The number of sequential fields to replicate in the inline array type.</param>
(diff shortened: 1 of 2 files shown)


src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/InlineArrayAttribute.cs
```diff
@@ -1,22 +1,42 @@
...
 namespace System.Runtime.CompilerServices
 {
     /// <summary>
     /// Indicates that the instance's storage is sequentially replicated "length" times.
     /// </summary>
-    [EditorBrowsable(EditorBrowsableState.Never)]
+    /// <remarks>
+    /// <para>
+    /// This attribute can be used to annotate a <see langword="struct"/> type with a single field.
+    /// The runtime will replicate that field as many times as specified in the actual type layout.
```
stephentoub, line 12:
```suggestion
    /// The runtime will replicate that field in the actual type layout as many times as is specified.
```

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/InlineArrayAttribute.cs
```diff
@@ -1,22 +1,42 @@
...
+    /// struct Float8InlineArray
+    /// {
+    ///     private float _value;
+    /// }
+    /// </code>
+    /// </para>
+    /// <para>
+    /// The compiler will also automatically generate an indexer property, and support casting the type
+    /// to <see cref="Span{T}"/> and <see cref="ReadOnlySpan{T}"/> for safe enumeration of all elements.
+    /// </para>
```
stephentoub, line 27: Let's only talk about runtime behaviors here. There are multiple compilers, in different languages, and they won't all do this.
```suggestion
```

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/InlineArrayAttribute.cs
```diff
@@ -1,22 +1,42 @@
...
+    /// </remarks>
     [AttributeUsage(AttributeTargets.Struct, AllowMultiple = false)]
     public sealed class InlineArrayAttribute : Attribute
     {
+        /// <summary>Creates a new <see cref="InlineArrayAttribute"/> instance with the specified parameters.</summary>
```
stephentoub, line 32:
```suggestion
        /// <summary>Creates a new <see cref="InlineArrayAttribute"/> instance with the specified length.</summary>
```

stephentoub (approved):

stephentoub:
Thanks.

PR #94320, opened 2023-11-02
Re-enable S.T.Json with source generation tests

MichalStrehovsky:
These should not be disabled.

### DIFF
--- a/src/libraries/tests.proj
+++ b/src/libraries/tests.proj
@@ -338,8 +338,6 @@
     <ProjectExclusions Include="$(MSBuildThisFileDirectory)Microsoft.Extensions.Logging.Abstractions\tests\Microsoft.Extensions.Logging.Generators.Tests\Microsoft.Extensions.Logging.Generators.Roslyn4.0.Tests.csproj" />
     <ProjectExclusions Include="$(MSBuildThisFileDirectory)System.Runtime.InteropServices\tests\LibraryImportGenerator.UnitTests\LibraryImportGenerator.Unit.Tests.csproj" />
     <ProjectExclusions Include="$(MSBuildThisFileDirectory)System.Runtime.InteropServices\tests\ComInterfaceGenerator.Unit.Tests\ComInterfaceGenerator.Unit.Tests.csproj" />
-    <ProjectExclusions Include="$(MSBuildThisFileDirectory)System.Text.Json\tests\System.Text.Json.SourceGeneration.Tests\System.Text.Json.SourceGeneration.Roslyn3.11.Tests.csproj" />
-    <ProjectExclusions Include="$(MSBuildThisFileDirectory)System.Text.Json\tests\System.Text.Json.SourceGeneration.Tests\System.Text.Json.SourceGeneration.Roslyn4.4.Tests.csproj" />
     <ProjectExclusions Include="$(MSBuildThisFileDirectory)System.Text.Json\tests\System.Text.Json.SourceGeneration.Unit.Tests\System.Text.Json.SourceGeneration.Roslyn3.11.Unit.Tests.csproj" />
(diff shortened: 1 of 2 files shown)


MichalStrehovsky:
@dotnet/ilc-contrib could someone have a look?

MichalStrehovsky:
@LakshanF I'm also undoing your test change from #93245 because it doesn't work.

```
  Starting:    GC.Stress.XUnitWrapper (parallel test collections = on, max threads = 2)
    GC/Stress/Framework/ReliabilityFramework/Tests/concurrentspin2.sh [FAIL]
      chmod: cannot access '/datadisks/disk1/work/AEAB08C4/w/9E3408BF/e/GC/Stress/Framework/ReliabilityFramework/Tests//native/concurrentspin2': No such file or directory
      /datadisks/disk1/work/AEAB08C4/p/nativeaottest.sh: line 15: /datadisks/disk1/work/AEAB08C4/w/9E3408BF/e/GC/Stress/Framework/ReliabilityFramework/Tests//native/concurrentspin2: No such file or directory
      
      Return code:      1
      Raw output file:      /datadisks/disk1/work/AEAB08C4/w/9E3408BF/uploads/Reports/GC.Stress/Framework/ReliabilityFramework/Tests/concurrentspin2.output.txt
      Raw output:
```

These tests are all marked https://github.com/dotnet/runtime/blob/aea45f7aaa8f73cb2a585d8b14ce66634d58cb68/src/tests/GC/Stress/Tests/Directory.Build.targets#L3 so they don't build unless we're in special mode.

If you're making native AOT specific test changes, please trigger the outerloop leg.

src/libraries/tests.proj
```diff
@@ -338,8 +338,6 @@
     <ProjectExclusions Include="$(MSBuildThisFileDirectory)Microsoft.Extensions.Logging.Abstractions\tests\Microsoft.Extensions.Logging.Generators.Tests\Microsoft.Extensions.Logging.Generators.Roslyn4.0.Tests.csproj" />
     <ProjectExclusions Include="$(MSBuildThisFileDirectory)System.Runtime.InteropServices\tests\LibraryImportGenerator.UnitTests\LibraryImportGenerator.Unit.Tests.csproj" />
     <ProjectExclusions Include="$(MSBuildThisFileDirectory)System.Runtime.InteropServices\tests\ComInterfaceGenerator.Unit.Tests\ComInterfaceGenerator.Unit.Tests.csproj" />
-    <ProjectExclusions Include="$(MSBuildThisFileDirectory)System.Text.Json\tests\System.Text.Json.SourceGeneration.Tests\System.Text.Json.SourceGeneration.Roslyn3.11.Tests.csproj" />
```
jkotas, line 341: These tests kept crashing for me in my RuntimeType PR with symptoms that suggested fatal OOMs. Example: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_apis/build/builds/456077/logs/69

I have disabled them thinking that they use Roslyn, but it is not actually the case.

jkotas (approved):
Thank you for fixing this!

src/libraries/tests.proj
MichalStrehovsky: They passed the outerloop test on the first commit of this PR. We get rather random OOMs in libs tests in general from time to time on Linux only. To fix it, we'd probably either need to drop parallelism or use the workstation GC, like we do for coreclr pri0 tests. I discussed it with Andy and the plan is to do something once it starts looking too severe.

Yep, these are the tests that the Json team fixed up for Native AOT in .NET 8 timeframe. The naming scheme always confuses me too (is this testing output of the source generator, or the source generator itself?).

PR #94324, opened 2023-11-02
Make System.Diagnostics.Tracing.Tests run sequentially

### LINKED ISSUE #88027: Test failure BasicEventSourceTests.TestsManifestGeneration.Test_EventSource_EtwManifestGenerationRollover

### LINKED ISSUE #91304: Test failure BasicEventSourceTests.TestsManifestGeneration.Test_EventSource_EtwManifestGenerationRollover
**Failed in:** [runtime-coreclr libraries-jitstress 20230829.1](https://dev.azure.com/dnceng-public/public/_build/results?buildId=388927&view=ms.vss-test-web.build-test-results-tab&runId=8336128&resultId=171432&paneView=debug)

**Failed tests:**
```
net8.0-windows-Release-x86-CoreCLR_checked-zapdisable-Windows.10.Amd64.Open
    - BasicEventSourceTests.TestsManifestGeneration.Test_EventSource_EtwManifestGenerationRollover
```

**Error message:**
```
Microsoft.DotNet.RemoteExecutor.RemoteExecutionException : Half-way through waiting for remote process.
Memory load: 15

Image Name                     PID Services                                    
========================= ======== ============================================
svchost.exe                    752 BrokerInfrastructure, DcomLaunch, LSM,      
                                   PlugPlay, Power, SystemEventsBroker         
svchost.exe                    816 RpcEptMapper, RpcSs                         
svchost.exe                    964 Appinfo, CertPropSvc, DsmSvc, gpsvc,        
                                   iphlpsvc, lfsvc, ProfSvc, sacsvr, Schedule, 
                                   SENS, SessionEnv, ShellHWDetection, Themes, 
                                   UserManager, Winmgmt, WpnService, wuauserv  
svchost.exe                    972 TermService                                 
svchost.exe                    332 Dhcp, EventLog, lmhosts, TimeBrokerSvc,     
                                   vmictimesync                                
svchost.exe                    740 BFE, CoreMessagingRegistrar, DPS, MpsSvc,   
                                   pla                                         
svchost.exe                    860 NcbService, PcaSvc, TrkWks, UALSVC,         
                                   UmRdpService, vmickvpexchange,              
                                   vmicshutdown, WdiSystemHost                 
svchost.exe                   1156 CDPSvc, EventSystem, FontCache, netprofm,   
                                   nsi, W32Time, WinHttpAutoProxySvc           
svchost.exe                   1176 vmicheartbeat                               
svchost.exe                   1428 CryptSvc, Dnscache, LanmanWorkstation,      
                                   NlaSvc, WinRM                               
svchost.exe                   1436 Wcmsvc                                      
svchost.exe                   2060 AppHostSvc                                  
svchost.exe                   2112 LanmanServer                                
svchost.exe                   2176 StateRepository, tiledatamodelsvc           
svchost.exe                   2200 DiagTrack                                   
svchost.exe                   2216 W3SVC, WAS                                  
svchost.exe                   3520 CDPUserSvc_494e1, OneSyncSvc_494e1          
svchost.exe                   3944 AppReadiness                                
Timed out at 8/29/2023 8:11:30 AM after 60000ms waiting for remote process.
Wrote mini dump to: C:\h\w\A8C7096F\w\AA86098F\uploads\5820.unh4zltf.kzk.dmp
	Process ID: 5820
	Handle: 1128
	Name: dotnet
	MainModule: C:\h\w\A8C7096F\p\dotnet.exe
	StartTime: 8/29/2023 8:10:29 AM
	TotalProcessorTime: 00:00:06.9375000
	Threads:
		Thread #1 (OS 0x9D4)   [MTA]
			[HelperMethodFrame] (System.Threading.Thread.SleepInternal)
			System.Threading.Thread.Sleep(Int32)
			BasicEventSourceTests.TestsManifestGeneration.<Test_EventSource_EtwManifestGenerationRollover>b__3_0()
			[HelperMethodFrame_PROTECTOBJ] (System.RuntimeMethodHandle.InvokeMethod)
			System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(System.Object, IntPtr*)
			System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(System.Object, System.Reflection.BindingFlags)
			System.Reflection.RuntimeMethodInfo.Invoke(System.Object, System.Reflection.BindingFlags, System.Reflection.Binder, System.Object[], System.Globalization.CultureInfo)
			Microsoft.DotNet.RemoteExecutor.Program.Main(System.String[])
			[InlinedCallFrame]
		Thread #2 (OS 0x9D0) [Finalizer] [Background] 
			[DebuggerU2MCatchHandlerFrame]

```

**Stack trace:**
```
   at Microsoft.DotNet.RemoteExecutor.RemoteInvokeHandle.Dispose(Boolean disposing) in /_/src/Microsoft.DotNet.RemoteExecutor/src/RemoteInvokeHandle.cs:line 225
   at BasicEventSourceTests.TestsManifestGeneration.Test_EventSource_EtwManifestGenerationRollover() in /_/src/libraries/System.Diagnostics.Tracing/tests/BasicEventSourceTest/TestsManifestGeneration.Etw.cs:line 81
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args) in /_/src/coreclr/System.Private.CoreLib/src/System/Reflection/MethodBaseInvoker.CoreCLR.cs:line 36
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr) in /_/src/libraries/System.Private.CoreLib/src/System/Reflection/MethodBaseInvoker.cs:line 57
```

### LINKED ISSUE #91769: Test failure BasicEventSourceTests.TestsWrite.Test_Write_T_ETW
**Failed in:** [runtime-coreclr libraries-jitstress 20230907.1](https://dev.azure.com/dnceng-public/public/_build/results?buildId=398021&view=ms.vss-test-web.build-test-results-tab&runId=8590802&resultId=137551&paneView=debug)

**Failed tests:**
```
net9.0-windows-Release-arm64-CoreCLR_checked-jitminopts-Windows.11.Arm64.Open
    - BasicEventSourceTests.TestsWrite.Test_Write_T_ETW
```

**Error message:**
```
System.Runtime.InteropServices.COMException : The instance name passed was not recognized as valid by a WMI data provider. (0x80071069)
```

**Stack trace:**
```
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHR(Int32 errorCode) in /_/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs:line 849
   at Microsoft.Diagnostics.Tracing.Session.TraceEventSession.Flush()
   at BasicEventSourceTests.EtwListener.Dispose() in /_/src/libraries/System.Diagnostics.Tracing/tests/BasicEventSourceTest/Harness/EtwListener.cs:line 97
   at BasicEventSourceTests.TestsWrite.Test_Write_T_ETW() in /_/src/libraries/System.Diagnostics.Tracing/tests/BasicEventSourceTest/TestsWrite.Etw.cs:line 27
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args) in /_/src/coreclr/System.Private.CoreLib/src/System/Reflection/MethodBaseInvoker.CoreCLR.cs:line 36
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr) in /_/src/libraries/System.Private.CoreLib/src/System/Reflection/MethodBaseInvoker.cs:line 57
```

davmason:
Fixes #91769
Fixes #91304
Fixes #88027

Apparently the `build.cmd -test` method of running tests needs a different way of disabling parallelization

### DIFF
--- a/src/libraries/System.Diagnostics.Tracing/tests/System.Diagnostics.Tracing.Tests.csproj
+++ b/src/libraries/System.Diagnostics.Tracing/tests/System.Diagnostics.Tracing.Tests.csproj
@@ -4,6 +4,7 @@
     <TargetFrameworks>$(NetCoreAppCurrent)-windows;$(NetCoreAppCurrent)</TargetFrameworks>
     <TestRuntime>true</TestRuntime>
     <IncludeRemoteExecutor>true</IncludeRemoteExecutor>
+    <TestDisableParallelization>true</TestDisableParallelization>
   </PropertyGroup>
   <ItemGroup Condition="'$(TargetsAppleMobile)' == 'true' or '$(TargetOS)' == 'android'">
     <RuntimeComponents Include="diagnostics_tracing" />


hoyosjs (approved):

danmoseley:
@davmason  what does this property do - pass some flag to xunit that tells it to ignore the attribute? What is that?

This is not the only test library that has those no parallelism annotations.

davmason:
@danmoseley I have no idea why the attributes are ignored, but through reverse engineering I found that this property sets maxthreads to 1 here https://github.com/dotnet/runtime/blob/d170785a1d6e64039c39efa2f0966455ec1402b5/eng/testing/xunit/xunit.console.targets#L18

From my testing it seemed like if you ran `dotnet test` in the folder it would respect the no parallelization attributes, but if you ran with `build.cmd/sh -test` like the CI legs do you have to set that property to disable parallelization

danmoseley:
@Bradwilson any idea why the attribute is being ignored? Context: https://github.com/dotnet/runtime/issues/91769#issuecomment-1754324947

bradwilson:
#### First: Is the console runner display broken?

The display from the console runner is based on an "outside" view of what's going on; that is, it only displays what the runner has requested the test framework do (based on defaults + configuration files + command line switches). The test framework does not report back out what it's actually doing. We provided no mechanism to do that in v2: we only support outside-in communication, not inside-out communication, regarding parallelism. So the assembly-level attribute does not get reflected into the display on the console runner.

That means this:

```
Starting:    System.Diagnostics.Tracing.Tests (parallel test collections = on, max threads = 4) <<<<<<<<<<<<<
```

is expected behavior.

The best way to get what you want is to remove the assembly-level attribute and add/update `xunit.runner.json` with [`parallelizeTestCollections`](https://xunit.net/docs/configuration-files#parallelizeTestCollections) set to `false`. (You can also pass `-parallel assemblies` or `-parallel none` to the console runner via command line switch, but this is not recommended in your situation, since you want to permanently disable parallelization in the test assembly and the VSTest adapter used from within Test Explorer does not provide a way to specify command line switches like this.)

#### Second: Is the attribute broken?

Through experimentation, I have found that the `CollectionBehavior.CollectionPerClass` and `CollectionBehavior.CollectionPerAssembly` is performing as I expected. 3 test classes, 1 test each with `Task.Delay(3000)` and I will see 3 seconds vs. 9 seconds of execution due to the differing default collection behavior. Of course, this only changes the _default_ collection assignment behavior. If you have any custom test collections defined, then any tests in those collections will continue to run in parallel against the default assembly-wide collection.

Unexpectedly, using `DisableTestParallelization = true` did not actually appear to disable test parallelization (everything took 3 seconds instead of 9), which means that this particular flag was broken at some point. The code break appears to have happened sometime between 2.0.0 and 2.1.0, which I can verify by running the code with the older runners:

image

So, congratulations! You've uncovered a bug that's more than 8 years old. 😄 The fact that this has been broken for that long tells me that people are overwhelmingly using the configuration file approach and not the attribute approach to disable test parallelization.

#### Recommendation

As such, I _**strongly**_ recommend you use `xunit.runner.json` to specify parallelism being disabled, at least until I can fix the issue with `[CollectionBehavior(DisableTestParallelization = true)]`. Using the configuration file is the only 100% replacement for the attribute, and I can verify that the configuration file approach does work (and has the added benefit of showing up correctly in the console runner display).

bradwilson:
Actually, ignore that. I tripped myself by having a configuration file that had `parallelTestCollections` set to `true`. _sigh_ I suppose I should be in bed at 1am and not trying to repro bugs. 😂 

I now believe you can ignore the second half of what I just wrote. `[CollectionBehavior(DisableTestParallelization = true)]` works as expected, and is not bugged. I am going to strikeout all the incorrect text above, and whatever the issue is that you're experiencing, if it's because of parallelism being enabled when you don't expect it, perhaps that's because you've enabled it in your `xunit.runner.json` file?

bradwilson:
image

image

danmoseley:
thanks @bradwilson . I realize that for some reason I had assumed that the attribute would win over any xunit.runner.json or command line flag, but apparently it's the reverse
https://github.com/xunit/xunit/blob/d300c8850978195b6e78d3cba17b78bb8a4e68f4/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestAssemblyRunner.cs#L93-L101

So the command line or xunit.runner.json must be overriding it. (Curiously, I'm not sure how our xunit.runner.json would have parallelizeTestCollections: true. The various layers around xunit in this repo and all the various ways it can be invoked, including a custom dummy xunit with fewer dependencies, make it hard to see, and I won't look further (eg actually run build.sh -test..))

One question though. Is there any functional difference between disabling parallelism and running with a limit of 1 thread? I guess this PR is doing the latter.
https://github.com/xunit/xunit/blob/d300c8850978195b6e78d3cba17b78bb8a4e68f4/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestAssemblyRunner.cs#L191-L194

danmoseley:
@davmason there are 18 test libraries in the tree that have `CollectionBehavior(DisableTestParallelization = true` but only the tracing tests have `<TestDisableParallelization>true</TestDisableParallelization>`.  I wonder whether they're running in parallel.

bradwilson:
> I realize that for some reason I had assumed that the attribute would win over any xunit.runner.json or command line flag, but apparently it's the reverse

Yes, it's definitely the reverse, and purposefully. We presume that things you put in source define the default behavior, and you can use interactive mechanisms like command line switches to alter that default behavior. The reverse wouldn't make much sense, as you wouldn't be able to override anything in source w/o changing the source itself.

> Is there any functional difference between disabling parallelism and running with a limit of 1 thread?

Yes. (Everything I'm about to describe is a summary of [`XunitTestAssemblyRunner.RunTestsCollectionsAsync`](https://github.com/xunit/xunit/blob/d300c8850978195b6e78d3cba17b78bb8a4e68f4/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestAssemblyRunner.cs#L187-L258) and the [base class method that it overrides from `TestAssemblyRunner`](https://github.com/xunit/xunit/blob/d300c8850978195b6e78d3cba17b78bb8a4e68f4/src/xunit.execution/Sdk/Frameworks/Runners/TestAssemblyRunner.cs#L237-L249).)

If parallelism is globally disabled, we do not create the `MaxConcurrencySyncContext` and thus the number of maximum parallel threads is irrelevant (and we do not create our own custom thread pool as it's the sync context that owns the thread pool). We run the test collections sequentially, and inside the test collection we always run tests sequentially as test collection is our parallelization boundary.

If parallelism is globally enabled, we do create the `MaxConcurrencySyncContext` with the requested number of threads in the pool. We then determine which test collections are "parallelizable" vs. "non-parallelizable". Test collections are parallelizable by default, but can be overridden with [`[CollectionDefinition(DisableParallelization = true)]`](https://github.com/xunit/xunit/blob/d300c8850978195b6e78d3cba17b78bb8a4e68f4/src/xunit.core/CollectionDefinitionAttribute.cs#L23).

For every parallelizable test collection, start the first test in the collection. These are all started in parallel. As each test in the collection finishes, we start the next test for that collection, until all the tests in the collection have run. Once all the tests from all the parallelizable test collections have run, then we sequentially run all the non-parallelizable test collections in a manner identical to the "parallelism globally disabled" behavior.

We will effectively _start_ running `n` tests, where `n` is the number of parallelizable test collections. If your maximum thread count is 1, then at any given time, only 1 of them will actually be running, but the rest will be ready to run when the thread becomes available. So if a test in this scenario hits an `await` against a Task that is not already complete, that test will wait for it's Task to complete and the one thread will become available for work, so a second test will now begin to run. When that test completes (or itself awaits an incomplete Task), then it also goes back into the waiting pool and another test runs.

(There is an important aside here: we rely on you not escaping the sync context to ensure parallelism. That means any call to `.ConfigureAwait(false)` inside the test method is an error condition, which is why we recently created an [analyzer which highlights this](https://xunit.net/xunit.analyzers/rules/xUnit1030). Similarly, if you use any other parallelism mechanism other than awaiting Tasks, such as spawning your own threads, then those mechanisms are things we don't know about and do not control, and as such any code running there is outside of our max thread count system.)

Generally speaking, people more often have a subset of tests that they want to prevent running in parallel with one another, typically because of a shared resource like a database. They use test collections and collection fixtures to coordinate that. Disabling parallelism globally is a mechanism that typically only makes sense if you have a globally shared resource and _**every test**_ in your test assembly needs access to that globally shared resource. This is a pretty uncommon scenario in my experience, but not unheard of.

danmoseley:
Super helpful, thanks.

There's some begged questions about our own infra here, but this issue will be helpful if those are actually encountered.

I'm guessing that these tracing tests don't run problematic code in awaits but I can't look at the moment.

bradwilson:
FWIW, the `.ConfigureAwait(false)` only applies to code in the test method itself. Any other code is irrelevant since that's the only code that needs to return to our sync context to ensure maximum # of running tests. However, any code under test that uses `.ConfigureAwait(false)` will continue to run on a normal thread pool thread, which means it can run in parallel with the test(s). This may or may not be an issue, though in practice it typically isn't.

Globally disabling parallelism doesn't have any effect here; it's just that the one test which may be awaiting an incomplete task will be sitting and waiting for the result, which may end up being calculated on a different thread. Even in that scenario, though, any non-parallelizable test collection (including globally disabling parallelism) means the test will just sit there and wait, while no other tests are running, rather than giving up a worker thread in the pool for another test.

bradwilson:
I'm also pushing up an update into the next release of the v2 core framework so that max parallel thread counts are only shown when test collection parallelization is enabled.

So previously:

```
Starting: assembly1 (parallel test collections = on, max threads = 24)
Starting: assembly2 (parallel test collections = off, max threads = 24)
```

Upcoming:

```
Starting: assembly1 (parallel test collections = on [24 threads])
Starting: assembly2 (parallel test collections = off)
```

This should add clarify about thread counts being tied into parallelism as well as a bit of brevity.

danmoseley:
that's helpful. what about this case ` [InlineData(false, null, null, null, "[Imp] =>   Starting:    testAssembly")]` .. I guess it means "defaults" -- would it make sense to say what those defaults are here? whoever reads the log probably cares more about whether the tests potentially ran in parallel than how that was achieved.

bradwilson:
That test is specifically about what people see without diagnostic messages enabled (that's the first `false`). You only see the extra information when you enable diagnostic messages.

PR #94330, opened 2023-11-03
Fix the type of handle constants replacing TYP_REF expressions

BruceForstall:
`optVNConstantPropOnTree` will propagate a constant handle value into a tree node currently typed as TYP_REF, where the new constant node is TYP_I_IMPL. This leads to trees like:

```
*  STORE_LCL_VAR ref   (AX) V08 tmp6
\--*  CNS_INT(h) long   $181
```

This seems like an undesired type mismatch. In another change, code was asserting because of this.

Change the type of the substituted constant node to be TYP_REF if necessary, to match the type of the node being replaced.

This change causes diffs: the constant is GC reported.

### DIFF
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -2649,6 +2649,13 @@ GenTree* Compiler::optVNConstantPropOnTree(BasicBlock* block, GenTree* tree)
                 if (!opts.compReloc)
                 {
                     conValTree = gtNewIconHandleNode(value, vnStore->GetHandleFlags(vnCns));
+
+                    // The generated constant node will have TYP_I_IMPL. However, we might be replacing a TYP_REF
+                    // node. If so, make it a TYP_REF.
+                    if (tree->TypeIs(TYP_REF))
+                    {
+                        conValTree->gtType = TYP_REF;
+                    }
                 }
             }
             else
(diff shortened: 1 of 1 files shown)


BruceForstall:
@jakobbotsch @EgorBo @AndyAyersMS Is this a reasonable change? Or does the IR need to allow TYP_I_IMPL CNS_INT nodes anywhere a TYP_REF node can be?

jakobbotsch:
> @jakobbotsch @EgorBo @AndyAyersMS Is this a reasonable change? Or does the IR need to allow TYP_I_IMPL CNS_INT nodes anywhere a TYP_REF node can be?

I'd more say we should disallow this kind of propagation entirely... What is the context around what happens here? Are we giving two `TYP_REF` and `TYP_I_IMPL` trees the same VN?

EgorBo:
I'd recommend running some jitstress, perhaps, some trees might not like that their operands suddenly changed TYP_LONG to TYP_REF here. I am not sure it's a correctness issue as long as we propagate GTF_ICON flag (we do here), but if disabling such propagation completely doesn't result in massive diffs - I am fine too.

BruceForstall:
> What is the context around what happens here? Are we giving two `TYP_REF` and `TYP_I_IMPL` trees the same VN?

Right. We assign a local a constant, then constant prop:
```
*  STORE_LCL_VAR ref    V09 tmp7         d:3 $VN.Void
\--*  COMMA     ref    $181
   +--*  STORE_LCL_VAR ref    V15 cse0         d:1 $VN.Void
   |  \--*  CNS_INT(h) ref    $181
   \--*  LCL_VAR   ref    V15 cse0         u:1 $181

*  STORE_LCL_VAR ref   (AX) V08 tmp6          $VN.Void
\--*  LCL_VAR   ref    V09 tmp7         u:3 (last use) $181

=>

*  STORE_LCL_VAR ref   (AX) V08 tmp6          $VN.Void
\--*  CNS_INT(h) long   $181
```

Looks perfectly reasonable, except for the type mismatch. Handle constants in VN are "type-less", always considered TYP_I_IMPL (makes sense).

> Presumably, there is nothing wrong that we don't report those TYP_I_IMPL to GC - they're either nongc objects or pinned handles?

I think these are frozen objects?

PR #94335, opened 2023-11-03
feat: add new IMemoryCache extensions

### LINKED ISSUE #92101: [API Proposal]: Provide a configurable overload of MemoryCacheEntryOptions for the commonly used method GetOrCreate in MemoryCacheExtension
### Background and motivation

In the MemoryCache extension class, The commonly used GetOrCreateAsync() should have an overloaded method for setting the MemoryCacheEntryOptions property

When we use the caching function. It is usually necessary to set MemoryCacheEntryOptions to determine the cache timeout and other configuration properties. If GetOrCreate does not provide an overload of valid MemoryCacheEntryOptions, this method will become unavailable in the scenario where MemoryCacheEntryOptions need to be configured. Users can only go back and select Get and call the Set method to configure.

### API Proposal

```csharp

public class MemoryCacheExtensions
{

        /// <summary>
        /// Gets the value associated with this key if it exists, or generates a new entry using the provided key and a value from the given factory if the key is not found.
        /// </summary>
        /// <typeparam name="TItem">The type of the object to get.</typeparam>
        /// <param name="cache">The <see cref="IMemoryCache"/> instance this method extends.</param>
        /// <param name="key">The key of the entry to look for or create.</param>
        /// <param name="factory">The factory that creates the value associated with this key if the key does not exist in the cache.</param>
        /// <param name="options">The existing <see cref="MemoryCacheEntryOptions"/> instance to apply to the new entry.</param>
        /// <returns>The value associated with this key.</returns>
        public static TItem? GetOrCreate<TItem>(this IMemoryCache cache, object key, Func<ICacheEntry, TItem> factory, MemoryCacheEntryOptions? options)
        {
            if (!cache.TryGetValue(key, out object? result))
            {
                using ICacheEntry entry = cache.CreateEntry(key);
                if (options != null) {
                    entry.SetOptions(options);
                }
                result = factory(entry);
                entry.Value = result;
            }

            return (TItem?)result;
        }

        /// <summary>
        /// Asynchronously gets the value associated with this key if it exists, or generates a new entry using the provided key and a value from the given factory if the key is not found.
        /// </summary>
        /// <typeparam name="TItem">The type of the object to get.</typeparam>
        /// <param name="cache">The <see cref="IMemoryCache"/> instance this method extends.</param>
        /// <param name="key">The key of the entry to look for or create.</param>
        /// <param name="factory">The factory task that creates the value associated with this key if the key does not exist in the cache.</param>
        /// <param name="options">The existing <see cref="MemoryCacheEntryOptions"/> instance to apply to the new entry.</param>
        /// <returns>The task object representing the asynchronous operation.</returns>
        public static async Task<TItem?> GetOrCreateAsync<TItem>(this IMemoryCache cache, object key, Func<ICacheEntry, Task<TItem>> factory, MemoryCacheEntryOptions? options)
        {
            if (!cache.TryGetValue(key, out object? result))
            {
                using ICacheEntry entry = cache.CreateEntry(key);
                if (options != null) {
                    entry.SetOptions(options);
                }
                result = await factory(entry).ConfigureAwait(false);
                entry.Value = result;
            }

            return (TItem?)result;
        }


}
```

### API Usage

```csharp
// GetOrCreateAsync And Set MemoryCacheEntryOptions

  var key = _cache.GetOrCreate("yj-key", c => "key-value", new MemoryCacheEntryOptions()
  {
      AbsoluteExpirationRelativeToNow = TimeSpan.FromDays(1)
  });

  var key = await _cache.GetOrCreateAsync("yj-key", c => "key-value", new MemoryCacheEntryOptions()
  {
      AbsoluteExpirationRelativeToNow = TimeSpan.FromDays(1)
  });
```

### Alternative Designs

You can change the existing GetOrCreate method, add parameters, and set default values to avoid destructive updates, but relatively speaking, overloading a new method is more suitable

### Risks

The code generated by the change is all existing code within the extension method, and no new logical code has been generated, resulting in a low risk factor.

But we still need to consider whether the execution order of SetOptions within the method in the above example code will have an impact on the functionality.

WeihanLi:
fixes #92101

### DIFF
--- a/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs
+++ b/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs
@@ -233,7 +233,7 @@ namespace Microsoft.Extensions.Caching.Memory
         /// <param name="cache">The <see cref="IMemoryCache"/> instance this method extends.</param>
         /// <param name="key">The key of the entry to look for or create.</param>
         /// <param name="factory">The factory task that creates the value associated with this key if the key does not exist in the cache.</param>
-        /// <param name="createOptions">The <see cref="MemoryCacheEntryOptions"/> needs to be applied to the created <see cref="ICacheEntry"/> if not null.</param>
+        /// <param name="createOptions">The options to be applied to the <see cref="ICacheEntry"/> if the key does not exist in the cache.</param>
         /// <returns>The task object representing the asynchronous operation.</returns>
         public static async Task<TItem?> GetOrCreateAsync<TItem>(this IMemoryCache cache, object key, Func<ICacheEntry, Task<TItem>> factory, MemoryCacheEntryOptions? createOptions)
         {
--- a/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs
+++ b/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs
@@ -185,7 +185,7 @@ namespace Microsoft.Extensions.Caching.Memory
         /// <param name="cache">The <see cref="IMemoryCache"/> instance this method extends.</param>
         /// <param name="key">The key of the entry to look for or create.</param>
         /// <param name="factory">The factory that creates the value associated with this key if the key does not exist in the cache.</param>
-        /// <param name="createOptions">The <see cref="MemoryCacheEntryOptions"/> needs to be applied to the created <see cref="ICacheEntry"/> if not null.</param>
+        /// <param name="createOptions">The options to be applied to the <see cref="ICacheEntry"/> if the key does not exist in the cache.</param>
         /// <returns>The value associated with this key.</returns>
         public static TItem? GetOrCreate<TItem>(this IMemoryCache cache, object key, Func<ICacheEntry, TItem> factory, MemoryCacheEntryOptions? createOptions)
         {
--- a/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs
+++ b/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs
@@ -167,15 +167,7 @@ namespace Microsoft.Extensions.Caching.Memory
         /// <returns>The value associated with this key.</returns>
         public static TItem? GetOrCreate<TItem>(this IMemoryCache cache, object key, Func<ICacheEntry, TItem> factory)
         {
-            if (!cache.TryGetValue(key, out object? result))
-            {
-                using ICacheEntry entry = cache.CreateEntry(key);
-
-                result = factory(entry);
-                entry.Value = result;
-            }
-
-            return (TItem?)result;
+            return GetOrCreate(cache, key, factory, null);
         }
 
         /// <summary>
@@ -213,17 +205,9 @@ namespace Microsoft.Extensions.Caching.Memory
         /// <param name="key">The key of the entry to look for or create.</param>
         /// <param name="factory">The factory task that creates the value associated with this key if the key does not exist in the cache.</param>
         /// <returns>The task object representing the asynchronous operation.</returns>
-        public static async Task<TItem?> GetOrCreateAsync<TItem>(this IMemoryCache cache, object key, Func<ICacheEntry, Task<TItem>> factory)
+        public static Task<TItem?> GetOrCreateAsync<TItem>(this IMemoryCache cache, object key, Func<ICacheEntry, Task<TItem>> factory)
         {
-            if (!cache.TryGetValue(key, out object? result))
-            {
-                using ICacheEntry entry = cache.CreateEntry(key);
-
-                result = await factory(entry).ConfigureAwait(false);
-                entry.Value = result;
-            }
-
-            return (TItem?)result;
+            return GetOrCreateAsync<TItem>(cache, key, factory, null);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs
```diff
@@ -198,5 +225,32 @@ public static TItem Set<TItem>(this IMemoryCache cache, object key, TItem value,
...
+            if (!cache.TryGetValue(key, out object? result))
+            {
+                using ICacheEntry entry = cache.CreateEntry(key);
+
+                if (createOptions != null)
+                {
+                    entry.SetOptions(createOptions);
+                }
+
+                result = await factory(entry).ConfigureAwait(false);
```
saul, line 249: If `GetOrCreateAsync` is called again while `await factory(entry)` is still in progress, this method will return `null`. I think the await needs to be moved above `CreateEntry` to avoid this

WeihanLi:
```c#
await using var services = new ServiceCollection()
    .AddMemoryCache()
    .BuildServiceProvider();
var memoryCache = services.GetRequiredService<IMemoryCache>();

Task.Run(() =>
{
    memoryCache.GetOrCreate("test", entry =>
    {
        Thread.Sleep(4000);
        Console.WriteLine("completed");
        return "1213";
    });
});
Thread.Sleep(2000);
if (memoryCache.TryGetValue("test", out var value))
{
    Console.WriteLine($"cache get and value is {value}");
}
else
{
    Console.WriteLine("cache not exists");
}
```

tried with a small demo, got `cache not exists` output, the cache entry would not be committed to the memory cache entries when it created

src/libraries/Microsoft.Extensions.Caching.Memory/tests/MemoryCacheSetAndRemoveTests.cs
```diff
@@ -793,6 +793,40 @@ public async Task GetOrCreateAsyncFromCacheWithNullKeyThrows()
...
+            var cache = CreateCache();
+            var expiry = 1000;
+            var cacheEntryOptions = new MemoryCacheEntryOptions()
+            {
+                AbsoluteExpirationRelativeToNow = TimeSpan.FromMilliseconds(expiry)
+            };            
+            var value = cache.GetOrCreate<string>(cacheKey, _ => cacheKey, cacheEntryOptions);
+            Assert.Equal(cacheKey, value);
+            Assert.True(cache.TryGetValue(cacheKey, out _));
+            Thread.Sleep(expiry * 2);
```
Jozkee, line 810: I would rather not add more timeout tests because they tend to fail on slow machines. Plus a two seconds way is a bit large for CI, is there a way we can test without expiration?

src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs
```diff
@@ -178,6 +178,33 @@ public static TItem Set<TItem>(this IMemoryCache cache, object key, TItem value,
...
+        /// <summary>
+        /// Gets the value associated with this key if it exists, or generates a new entry using the provided key and a value from the given factory if the key is not found.
+        /// </summary>
+        /// <typeparam name="TItem">The type of the object to get.</typeparam>
+        /// <param name="cache">The <see cref="IMemoryCache"/> instance this method extends.</param>
+        /// <param name="key">The key of the entry to look for or create.</param>
+        /// <param name="factory">The factory that creates the value associated with this key if the key does not exist in the cache.</param>
+        /// <param name="createOptions">The <see cref="MemoryCacheEntryOptions"/> needs to be applied to the created <see cref="ICacheEntry"/> if not null.</param>
+        /// <returns>The value associated with this key.</returns>
+        public static TItem? GetOrCreate<TItem>(this IMemoryCache cache, object key, Func<ICacheEntry, TItem> factory, MemoryCacheEntryOptions? createOptions)
```
Jozkee, line 190: Can we now refactor the existing `GetOrCreate[Async]` to call these new methods?

src/libraries/Microsoft.Extensions.Caching.Memory/tests/MemoryCacheSetAndRemoveTests.cs
Jozkee:
```cs
[Fact]
public async Task GetOrCreateAsyncWithCacheEntryOptions()
{
    var cacheKey = "test";
    var cache = CreateCache();
    ManualResetEvent mre = new ManualResetEvent(false);
    
    var options = new MemoryCacheEntryOptions();
    options.PostEvictionCallbacks.Add(new PostEvictionCallbackRegistration()
    {
        EvictionCallback = (key, value, reason, state) =>
        {
            Assert.Equal(cacheKey, key);
            Assert.Equal(cacheKey, value);
            Assert.Equal(EvictionReason.Removed, reason);
            mre.Set();
        }
    });

    var value = await cache.GetOrCreateAsync<string>(cacheKey, _ => Task.FromResult(cacheKey), cacheEntryOptions);
    Assert.Equal(cacheKey, value);
    Assert.True(cache.TryGetValue(cacheKey, out _));
    
    cache.Remove(cacheKey);
    Assert.True(mre.WaitOne(TimeSpan.FromSeconds(30)));
    Assert.False(cache.TryGetValue(cacheKey, out _));
}
```

src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs
WeihanLi: Sure, make sense, thanks very much, updated

stephentoub (approved):
Thanks

adamsitnik (approved):
LGTM, thank you for your contribution @WeihanLi !

PR #94342, opened 2023-11-03
Allow RyuJIT to refer to frozen RuntimeType instances

MichalStrehovsky:
With #93440 it became possible to place `RuntimeType` instances in the frozen region. This adds support for generating frozen `RuntimeType` instances within the compiler. We give these to RyuJIT whenever it asks for them.

* New `FrozenObjectNode` descendant that represents a `RuntimeType` instance. We have two flavors - one wraps constructed and the other wraps necessary `MethodTable`s (we have a need for freezing both kinds to keep our ability to optimize `RuntimeTypes` used in comparisons only).
* We hand out references to these whenever RyuJIT needs them.
* At `MethodTable` emission time, we check whether a frozen `RuntimeType` for this `MethodTable` was generated. If so, we pre-populate `MethodTable`’s `WritableData` section with a reloc to the `RuntimeType`. Otherwise we use null as usual.
* At runtime for `GetTypeFromEEType`, if `WritableData` is non-null, we just return that. Otherwise we create a pinned `RuntimeType` instance and write it into `WritableData`. In the future, we should allocate this on a frozen heap.

Old codegen for `Console.WriteLine(typeof(Program))`:

```asm
sub         rsp,28h
lea         rcx,[repro_Program::`vftable' (07FF7D03154E0h)]
call        S_P_CoreLib_Internal_Runtime_CompilerHelpers_LdTokenHelpers__GetRuntimeType (07FF7D0253290h)
mov         rcx,rax
call        System_Console_System_Console__WriteLine_11 (07FF7D0262AC0h)
nop
add         rsp,28h
ret
```

New codegen:

```asm
sub         rsp,28h
lea         rcx,[__RuntimeType_repro_Program (07FF7A218EC50h)]
call        System_Console_System_Console__WriteLine_11 (07FF7A20F2680h)
nop
add         rsp,28h
ret
```

I’ll do cctor preinitialization in a subsequent PR to keep things short. Contributes to #91704.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Type.NativeAot.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Type.NativeAot.cs
@@ -4,6 +4,7 @@
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
+using System.Runtime;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
@@ -30,15 +31,8 @@ namespace System
             // unifier's hash table.
             if (MethodTable.SupportsWritableData)
             {
-                ref UnsafeGCHandle handle = ref Unsafe.AsRef<UnsafeGCHandle>(pMT->WritableData);
-                if (handle.IsAllocated)
-                {
-                    return Unsafe.As<RuntimeType>(handle.Target);
-                }
-                else
-                {
-                    return GetTypeFromMethodTableSlow(pMT, ref handle);
-                }
+                ref RuntimeType? type = ref Unsafe.AsRef<RuntimeType?>(pMT->WritableData);
+                return type ?? GetTypeFromMethodTableSlow(pMT, ref type);
             }
             else
             {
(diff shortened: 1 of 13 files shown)


src/coreclr/nativeaot/System.Private.CoreLib/src/System/Type.NativeAot.cs
```diff
@@ -30,15 +31,8 @@ internal static unsafe RuntimeType GetTypeFromMethodTable(MethodTable* pMT)
...
-                if (handle.IsAllocated)
-                {
-                    return Unsafe.As<RuntimeType>(handle.Target);
-                }
-                else
-                {
-                    return GetTypeFromMethodTableSlow(pMT, ref handle);
-                }
+                ref RuntimeType? type = ref Unsafe.AsRef<RuntimeType?>(pMT->WritableData);
+                return type ?? GetTypeFromMethodTableSlow(pMT, ref type);
```
jkotas, line 35:
```suggestion
                return Unsafe.AsRef<RuntimeType?>(pMT->WritableData) ?? GetTypeFromMethodTableSlow(pMT);
```
We should fetch the writetable data slow in the slow method to minimize size of the aggressively inlined code.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Type.NativeAot.cs
```diff
@@ -47,17 +41,29 @@ internal static unsafe RuntimeType GetTypeFromMethodTable(MethodTable* pMT)
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private static unsafe RuntimeType GetTypeFromMethodTableSlow(MethodTable* pMT, ref UnsafeGCHandle handle)
+        private static unsafe RuntimeType GetTypeFromMethodTableSlow(MethodTable* pMT, ref RuntimeType? runtimeTypeCache)
         {
-            UnsafeGCHandle tempHandle = UnsafeGCHandle.Alloc(new RuntimeType(pMT));
+            // TODO: instead of fragmenting the frozen object heap, we should have our own allocator
```
jkotas, line 46: I have added check-box for this to https://github.com/dotnet/runtime/issues/91704

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/NecessaryCanonicalEETypeNode.cs
```diff
@@ -25,6 +26,8 @@ protected override ISymbolNode GetBaseTypeNode(NodeFactory factory)
             return _type.BaseType != null ? factory.NecessaryTypeSymbol(_type.BaseType.NormalizeInstantiation()) : null;
         }
 
+        protected override FrozenRuntimeTypeNode GetFrozenRuntimeTypeNode(NodeFactory factory) => throw new NotSupportedException();
```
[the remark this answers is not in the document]
MichalStrehovsky: Instantiations over `__Canon` are not visible to reflection and we don't need to ever access their `WritableData`. I'm making it so that we don't reserve space in the BSS section and emit a useless pointer here. See the change in `EETypeNode::.ctor` and `EETypeNode::OutputWritableData`.

jkotas (approved):
Nice!

EgorBo:
Nice! is this expected to be pre-initializeable now? (I checked your branch and it is not)
```
public class MyClass
{
    public static readonly Type type = typeof(int);
}
```

MichalStrehovsky:
> Nice! is this expected to be pre-initializeable now? (I checked your branch and it is not)
> 
> ```cs
> public class MyClass
> {
>     public static readonly Type type = typeof(int);
> }
> ```

"I’ll do cctor preinitialization in a subsequent PR to keep things short." at the bottom of top-post.

PR #94347, opened 2023-11-03
Convert DIV/MOD to UDIV/UMOD using assertions

### LINKED ISSUE #94218: Signed division performed even when values are provably non-negative
Repro:
```C#
static int Result(int i)
{
    if (i < 4) return 0;
    return i / 4;
}
```
This results in:
```asm
; Method Program:<<Main>$>g__Result|0_0(int):int (FullOpts)

G_M000_IG02:                ;; offset=0x0000
       cmp      ecx, 4
       jge      SHORT G_M000_IG05

G_M000_IG03:                ;; offset=0x0005
       xor      eax, eax

G_M000_IG04:                ;; offset=0x0007
       ret      

G_M000_IG05:                ;; offset=0x0008
       mov      eax, ecx
       sar      eax, 31
       and      eax, 3
       add      eax, ecx
       sar      eax, 2

G_M000_IG06:                ;; offset=0x0015
       ret      
; Total bytes of code: 22
```
but as the JIT can prove that the `i` used in `i / 4` isn't negative, so presumably it should be able to emit the code as if:
```C#
static int Result(int i)
{
    if (i < 4) return 0;
    return (int)((uint)i / 4);
}
```
were instead used? And it results in:
```asm
; Method Program:<<Main>$>g__Result|0_0(int):int (FullOpts)

G_M000_IG02:                ;; offset=0x0000
       cmp      ecx, 4
       jge      SHORT G_M000_IG05

G_M000_IG03:                ;; offset=0x0005
       xor      eax, eax

G_M000_IG04:                ;; offset=0x0007
       ret      

G_M000_IG05:                ;; offset=0x0008
       mov      eax, ecx
       shr      eax, 2

G_M000_IG06:                ;; offset=0x000D
       ret      
; Total bytes of code: 14
```

EgorBo:
Closes https://github.com/dotnet/runtime/issues/94218

```cs
static int Test(int i)
{
    if (i < 4) return 0;
    return i / 4;
}
```

```diff
; Method Benchmarks:Test(int):int (FullOpts)
       cmp      ecx, 4
       jge      SHORT G_M33641_IG05
       xor      eax, eax
       ret      
G_M33641_IG05:
       mov      eax, ecx
-      sar      eax, 31
-      and      eax, 3
-      add      eax, ecx
-      sar      eax, 2
+      shr      eax, 2
       ret      
-; Total bytes of code: 22
+; Total bytes of code: 14
```

### DIFF
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -4794,7 +4794,8 @@ GenTree* Compiler::optAssertionProp(ASSERT_VALARG_TP assertions, GenTree* tree, 
             // For now, we're mainly interested in "X op CNS" pattern (where CNS > 0).
             const GenTree* op1 = tree->gtGetOp1();
             const GenTree* op2 = tree->gtGetOp2();
-            if (!optLocalAssertionProp && !BitVecOps::IsEmpty(apTraits, assertions) && op2->IsNeverNegative(this))
+            if (varTypeIsIntegral(tree) && !optLocalAssertionProp && !BitVecOps::IsEmpty(apTraits, assertions) &&
+                op2->IsNeverNegative(this))
             {
                 const ValueNum lclVN = vnStore->VNConservativeNormalValue(op1->gtVNPair);
                 for (AssertionIndex index = 1; index <= optAssertionCount; index++)
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -3551,6 +3551,33 @@ GenTree* Compiler::optAssertionProp_BlockStore(ASSERT_VALARG_TP assertions, GenT
     return nullptr;
 }
 
+GenTree* Compiler::optAssertionProp_ModDiv(ASSERT_VALARG_TP assertions, GenTreeOp* tree, Statement* stmt)
+{
+    if (optLocalAssertionProp || !varTypeIsIntegral(tree) || BitVecOps::IsEmpty(apTraits, assertions))
+    {
+        return nullptr;
+    }
+
+    // For now, we're mainly interested in "X op CNS" pattern (where CNS > 0).
+    // Technically, we can check assertions for both operands, but it's not clear if it's worth it.
+    if (!tree->gtGetOp2()->IsNeverNegative(this))
+    {
+        return nullptr;
+    }
+
+    const ValueNum dividendVN = vnStore->VNConservativeNormalValue(tree->gtGetOp1()->gtVNPair);
+    for (AssertionIndex index = 1; index <= optAssertionCount; index++)
+    {
+        AssertionDsc* curAssertion = optGetAssertion(index);
+        if (BitVecOps::IsMember(apTraits, assertions, index - 1) && curAssertion->IsNeverNegative(this, dividendVN))
+        {
+            tree->SetOper(tree->OperIs(GT_DIV) ? GT_UDIV : GT_UMOD, GenTree::PRESERVE_VN);
+            return optAssertionProp_Update(tree, tree, stmt);
+        }
+    }
+    return nullptr;
+}
+
 //------------------------------------------------------------------------
 // optAssertionProp_Return: Try and optimize a GT_RETURN via assertions.
 //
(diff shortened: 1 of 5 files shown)


EgorBo:
@AndyAyersMS @dotnet/jit-contrib PTAL, diffs [are not too big](https://dev.azure.com/dnceng-public/public/_build/results?buildId=459951&view=ms.vss-build-web.run-extensions-tab), I think I saw in some places in BCL we explicitly add `(uint)` cast for better CQ so this PR should allow to use that hack less often. Currently, it doesn't support `long` because we don't create assertions for those (I'll look separately if I can cheaply enable them) as these assertions mainly focus on bound checks.

src/coreclr/jit/assertionprop.cpp
```diff
@@ -3551,6 +3551,86 @@ GenTree* Compiler::optAssertionProp_BlockStore(ASSERT_VALARG_TP assertions, GenT
...
+
+    // For now, we're mainly interested in "X op CNS" pattern (where CNS > 0).
+    // Technically, we can check assertions for both operands, but it's not clear if it's worth it.
+    if (!tree->gtGetOp2()->IsNeverNegative(this))
+    {
+        return nullptr;
+    }
+
+    const ValueNum dividendVN = vnStore->VNConservativeNormalValue(tree->gtGetOp1()->gtVNPair);
+    for (AssertionIndex index = 1; index <= optAssertionCount; index++)
```
AndyAyersMS, line 3584: For the most part we're probably better of iterating through the bits of `assertions` here, as hopefully that iteration is cheap (using HW accelerated `bsf` or similar) and that most bits in the vector are *not* set.

I am tempted to convert most of the search loops like this in assertion prop over to bitvector driven loops.

src/coreclr/jit/assertionprop.cpp
```diff
@@ -3551,6 +3551,86 @@ GenTree* Compiler::optAssertionProp_BlockStore(ASSERT_VALARG_TP assertions, GenT
...
+    }
+
+    // For now, we're mainly interested in "X op CNS" pattern (where CNS > 0).
+    // Technically, we can check assertions for both operands, but it's not clear if it's worth it.
+    if (!tree->gtGetOp2()->IsNeverNegative(this))
+    {
+        return nullptr;
+    }
+
+    const ValueNum dividendVN = vnStore->VNConservativeNormalValue(tree->gtGetOp1()->gtVNPair);
```
AndyAyersMS, line 3583: Couldn't this use the liberal VN?

src/coreclr/jit/assertionprop.cpp
```diff
@@ -3551,6 +3551,86 @@ GenTree* Compiler::optAssertionProp_BlockStore(ASSERT_VALARG_TP assertions, GenT
...
+
+        // Root assertion has to be either:
+        // (X relop CNS) == 0
+        // (X relop CNS) != 0
+        if ((curAssertion->op2.kind != O2K_CONST_INT) || (curAssertion->op2.u1.iconVal != 0))
+        {
+            continue;
+        }
+
+        auto cmpOper = static_cast<genTreeOps>(info.cmpOper);
```
AndyAyersMS, line 3615: I think we only use `auto` for complex types like lambdas?

EgorBo: Removed, I thought that we have the same guideline as in C# for `var` 🙂

src/coreclr/jit/assertionprop.cpp
EgorBo: Yeah I think it can be liberal, changed.

src/coreclr/jit/assertionprop.cpp
EgorBo: Do we already have an example where we do that? I didn't bother because TP impact was zero 🙂 (because DIV/MOD aren't too common)

src/coreclr/jit/assertionprop.cpp
SingleAccretion: The rest of assertion propagation uses conservative VNs only; it needs a comment on why using liberal ones here is safe.

EgorBo: Let me check diffs this make (if any)..

jakobbotsch: I think this should use conservative VNs... otherwise we can end up producing an impossible result under data races.

EgorBo: Reverted to conservative, diffs didn't change

src/coreclr/jit/assertionprop.cpp
EgorBo: Replaced with BitVecOps::Iter assuming that what you meant

src/coreclr/jit/assertionprop.cpp
```diff
@@ -3551,6 +3551,89 @@ GenTree* Compiler::optAssertionProp_BlockStore(ASSERT_VALARG_TP assertions, GenT
...
+        return nullptr;
+    }
+
+    const ValueNum  dividendVN = vnStore->VNConservativeNormalValue(tree->gtGetOp1()->gtVNPair);
+    BitVecOps::Iter iter(apTraits, assertions);
+    unsigned        index = 0;
+    while (iter.NextElem(&index))
+    {
+        const AssertionIndex assertionIndex = GetAssertionIndex(index);
+        if (assertionIndex > optAssertionCount)
```
AndyAyersMS, line 3589: This should never happen anymore (it used to be possible in local prop, if the "client" passed in `apFull`).

AndyAyersMS (approved):

EgorBo:
Failure is https://github.com/dotnet/runtime/issues/86019

PR #94363, opened 2023-11-03
JIT: cross-block local assertion prop in morph

AndyAyersMS:
During global morph, allow assertions to propagate to a block from the block's predecessors. Handle special cases where we can't allow this to happen:
* block has preds that have not yet been morphed
* block has no preds
* block is specially flagged as one that might gain new preds during morph
* block is an EH handler entry

Contributes to #93246.

For now this is off by default.

### DIFF
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -541,63 +541,87 @@ void Compiler::optAssertionTraitsInit(AssertionIndex assertionCount)
 
 void Compiler::optAssertionInit(bool isLocalProp)
 {
-    // See if we should disable cross-block local prop
+    assert(NO_ASSERTION_INDEX == 0);
+    const unsigned maxTrackedLocals = (unsigned)JitConfig.JitMaxLocalsToTrack();
+
+    // We initialize differently for local prop / global prop
     //
-    if (optCrossBlockLocalAssertionProp)
+    if (isLocalProp)
     {
-#ifdef DEBUG
-        // Disable via config
+        optLocalAssertionProp = true;
+
+        // See if we should disable cross-block local prop
         //
-        if (JitConfig.JitDoCrossBlockLocalAssertionProp() == 0)
+        if (optCrossBlockLocalAssertionProp)
         {
-            JITDUMP("Disabling cross-block assertion prop by config setting\n");
-            optCrossBlockLocalAssertionProp = false;
-        }
+#ifdef DEBUG
+            // Disable via config
+            //
+            if (JitConfig.JitDoCrossBlockLocalAssertionProp() == 0)
+            {
+                JITDUMP("Disabling cross-block assertion prop by config setting\n");
+                optCrossBlockLocalAssertionProp = false;
+            }
 #endif
 
-        // Disable if too many locals
-        //
-        // The typical number of local assertions is roughly proportional
-        // to the number of locals. So when we have huge numbers of locals,
-        // just do within-block local assertion prop.
-        //
-        if (lvaCount > (unsigned)JitConfig.JitMaxLocalsToTrack())
-        {
-            JITDUMP("Disabling cross-block assertion prop: too many locals\n");
-            optCrossBlockLocalAssertionProp = false;
+            // Disable if too many locals
+            //
+            // The typical number of local assertions is roughly proportional
+            // to the number of locals. So when we have huge numbers of locals,
+            // just do within-block local assertion prop.
+            //
+            if (lvaCount > maxTrackedLocals)
+            {
+                JITDUMP("Disabling cross-block assertion prop: too many locals\n");
+                optCrossBlockLocalAssertionProp = false;
+            }
         }
+
+        if (optCrossBlockLocalAssertionProp)
+        {
+            // We know lvaCount is less than the tracked limit.
+            // Allow for roughly one assertion per local, up to the tracked limit.
+            // (empirical studies show about 0.6 asserions/local)
+            //
+            optMaxAssertionCount = (AssertionIndex)min(maxTrackedLocals, ((lvaCount / 64) + 1) * 64);
+        }
+        else
+        {
+            // The table will be reset for each block, so can be smaller.
+            //
+            optMaxAssertionCount = 64;
+        }
+
+        optAssertionDep =
+            new (this, CMK_AssertionProp) JitExpandArray<ASSERT_TP>(getAllocator(CMK_AssertionProp), max(1, lvaCount));
     }
-
-    // Use a function countFunc to determine a proper maximum assertion count for the
-    // method being compiled. The function is linear to the IL size for small and
-    // moderate methods. For large methods, considering throughput impact, we track no
-    // more than 64 assertions.
-    // Note this tracks at most only 256 assertions.
-    static const AssertionIndex countFunc[] = {64, 128, 256, 64};
-    static const unsigned       lowerBound  = 0;
-    static const unsigned       upperBound  = ArrLen(countFunc) - 1;
-    const unsigned              codeSize    = info.compILCodeSize / 512;
-    optMaxAssertionCount                    = countFunc[isLocalProp ? lowerBound : min(upperBound, codeSize)];
-
-    optLocalAssertionProp  = isLocalProp;
-    optAssertionTabPrivate = new (this, CMK_AssertionProp) AssertionDsc[optMaxAssertionCount];
-    assert(NO_ASSERTION_INDEX == 0);
-
-    if (!isLocalProp)
+    else
     {
+        // General assertion prop.
+        //
+        optLocalAssertionProp = false;
+
+        // Use a function countFunc to determine a proper maximum assertion count for the
+        // method being compiled. The function is linear to the IL size for small and
+        // moderate methods. For large methods, considering throughput impact, we track no
+        // more than 64 assertions.
+        // Note this tracks at most only 256 assertions.
+        //
+        static const AssertionIndex countFunc[] = {64, 128, 256, 64};
+        static const unsigned       upperBound  = ArrLen(countFunc) - 1;
+        const unsigned              codeSize    = info.compILCodeSize / 512;
+        optMaxAssertionCount                    = countFunc[min(upperBound, codeSize)];
+
         optValueNumToAsserts =
             new (getAllocator(CMK_AssertionProp)) ValueNumToAssertsMap(getAllocator(CMK_AssertionProp));
         optComplementaryAssertionMap = new (this, CMK_AssertionProp)
             AssertionIndex[optMaxAssertionCount + 1](); // zero-inited (NO_ASSERTION_INDEX)
     }
 
-    if (optAssertionDep == nullptr)
-    {
-        optAssertionDep =
-            new (this, CMK_AssertionProp) JitExpandArray<ASSERT_TP>(getAllocator(CMK_AssertionProp), max(1, lvaCount));
-    }
+    optAssertionTabPrivate = new (this, CMK_AssertionProp) AssertionDsc[optMaxAssertionCount];
 
     optAssertionTraitsInit(optMaxAssertionCount);
+
     optAssertionCount      = 0;
     optAssertionOverflow   = 0;
     optAssertionPropagated = false;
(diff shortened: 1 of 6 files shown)


AndyAyersMS:
@jakobbotsch @dotnet/jit-contrib here's the "rough cut"

This will have about a 1% TP impact (win x64) and big code size improvements and regressions. Seems like it will take some time to chip away at those.

Running this now to flush out any remaining bugs.

AndyAyersMS:
@amanasifkhalid FYI I'm seeing the same sort of cloning issue you hit in this PR... there are 4600 size regressions so it would be nice to filter out the ones where there was extra cloning.

```
[18:08:46] Top method regressions (percentages):
[18:08:46]          216 (67.50% of base) : 18807.dasm - Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpHeaders:ValidateHeaderValueCharacters(System.String,Microsoft.Extensions.Primitives.StringValues,System.Func`2[System.String,System.Text.Encoding]) (Tier1)
```

AndyAyersMS:
Latest commit upscales the bit vectors proportional to `lvaCount` until we hit the tracked local limit (so using similar sized bit vectors to what we use for liveness, since empirically I see about 0.6 local assertions / local, and want to over-size a bit to reduce likelihood of running out of table space).

This will incur a bit more TP cost; locally looks like around 1.3% or so. But as a benefit we see considerably larger overall code size savings; x64 windows should be around -1MB.

### Thoughts about how to get this to the point where it can be merged

**Throughput**: 

I think I can claw some of the TP back by 

(1) optimizing search loops in assertion prop. We should generally never search the entire table (save when adding an assertion) and instead walk the live assertion set, and for local prop we should intersect that set with the dep vector.

The bit vector operations aren't free so it might make sense to do this only when there are sufficient numbers of assertions, though that makes the code uglier. Perhaps it can all be hidden behind a suitably smart enumerator.

(2) stop morphing blocks that are unreachable or become unreachable because of local assertion prop.

Aside from removing all the morph related overhead, I suspect sometimes this dead block morphing creates issues for live blocks, eg either useless assertions that cause the table to fill faster, or else global actions (say DNER) that are hard to undo.

Currently the RPO strategy won't allow for removal of dead EH and may or may not make it easy to remove unreachable cyclic flow.

Both of these can be spun off as preliminary checkins, though the full benefits may not be seen without this change, as main has small bit vectors and isn't able to propagate constants nearly as aggressively.

At the end of the day though I expect there will still be a TP impact.

**Code Size** / **Code Quality**

With the massive numbers of diffs any sort of manual assessment is going to be a random sampling at best. I will try and look at some of the bigger regressions, possibly suppressing cloning to help remove that as a factor.

The more aggressive copy prop that this enables puts pressure on LSRA, as copies to temps provide natural split points. I am not sure we can find effective heuristics to counterbalance this.

I think it makes sense to check all this in initially as off-by-default, and enable runs in the perf lab, and use that data to both identify CQ regressions and to decide if the CQ improvements justify the additional TP costs.

AndyAyersMS:
[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=462165&view=ms.vss-build-web.run-extensions-tab) @ [bc59375](https://github.com/dotnet/runtime/pull/94363/commits/bc59375de461b44632c7afa636b8c8a25a756633).

Aggregate size savings look good, but still quite a lot of churn, and a fair number of large regressions.

AndyAyersMS:
I'm going to disable this by default for now and set up some lab runs for correctness. I will alsoset up runs of this in a fork in the perf lab, so I can get a sense of the perf impact of all this CQ churn.

In the meantime I'll continue to look at fixing the bad diffs (and bad perf, once I have data) and whittling away at TP, so once this is merged I'll open a draft PR to enable and iterate there.

src/tests/Common/testenvironment.proj
```diff
@@ -220,6 +221,7 @@
     <TestEnvironment Include="jitobjectstackallocation" JitObjectStackAllocation="1" TieredCompilation="0" />
     <TestEnvironment Include="jitphysicalpromotion_only" JitStressModeNames="STRESS_NO_OLD_PROMOTION" TieredCompilation="0" />
     <TestEnvironment Include="jitphysicalpromotion_full" JitStressModeNames="STRESS_PHYSICAL_PROMOTION_COST STRESS_NO_OLD_PROMOTION" TieredCompilation="0" />
+    <TestEnvironment Include="jitcrossblocklocalassertionprop" JitDoCrossBlockLocalAssertionProp="1" TieredCompilation="0" />
```
BruceForstall, line 224: `TieredCompilation=0` is the default in this file, so you don't need to specify it (looks like some other cases also over-specify it)

src/coreclr/jit/jitconfigvalues.h
```diff
@@ -653,6 +653,9 @@ CONFIG_INTEGER(JitEnableHeadTailMerge, W("JitEnableHeadTailMerge"), 1)
 // Enable physical promotion
 CONFIG_INTEGER(JitEnablePhysicalPromotion, W("JitEnablePhysicalPromotion"), 1)
 
+// Enable cross-block local assertion prop
+CONFIG_INTEGER(JitDoCrossBlockLocalAssertionProp, W("JitDoCrossBlockLocalAssertionProp"), 0)
```
BruceForstall, line 657: `Do` seems fine, but should we use `Enable` instead, to match other similar cases (like `JitEnablePhysicalPromotion`, above)?

src/coreclr/jit/morph.cpp
```diff
@@ -13822,7 +13900,8 @@ PhaseStatus Compiler::fgMorphBlocks()
 
     // Local assertion prop is enabled if we are optimized
     //
-    optLocalAssertionProp = opts.OptimizationEnabled();
+    optLocalAssertionProp           = opts.OptimizationEnabled();
+    optCrossBlockLocalAssertionProp = optLocalAssertionProp;
```
BruceForstall, line 13904:
```suggestion
    optCrossBlockLocalAssertionProp = optLocalAssertionProp && JitConfig.JitDoCrossBlockLocalAssertionProp();
```
?

src/coreclr/jit/morph.cpp
```diff
@@ -13780,10 +13780,80 @@ void Compiler::fgMorphBlock(BasicBlock* block)
...
+                }
+            }
+            else
+            {
+                JITDUMP(FMT_BB " ineligible for cross-block; clearing assertions in\n", block->bbNum);
+            }
+
+            if (!canUsePredAssertions)
+            {
+                apLocal = BitVecOps::MakeEmpty(apTraits);
```
BruceForstall, line 13854: Does this also need `optAssertionReset(0);`?

Can the `(!optCrossBlockLocalAssertionProp)` and the "failure" cases be combined (tail merged)?

src/coreclr/jit/morph.cpp
```diff
@@ -13780,10 +13780,80 @@ void Compiler::fgMorphBlock(BasicBlock* block)
...
+                    }
+
+                    // No, not available.
+                    //
+                    JITDUMP(FMT_BB " pred " FMT_BB " not processed; clearing assertions in\n", block->bbNum,
+                            pred->bbNum);
+                    canUsePredAssertions = false;
+                }
+
+                // If there were no preds, there are no asserions in.
```
BruceForstall, line 13834:
```suggestion
                // If there were no preds, there are no assertions in.
```

src/coreclr/jit/morph.cpp
```diff
@@ -13780,10 +13780,80 @@ void Compiler::fgMorphBlock(BasicBlock* block)
...
+                        }
+
+                        continue;
+                    }
+
+                    // No, not available.
+                    //
+                    JITDUMP(FMT_BB " pred " FMT_BB " not processed; clearing assertions in\n", block->bbNum,
+                            pred->bbNum);
+                    canUsePredAssertions = false;
```
BruceForstall, line 13831: add `break;`?

Or better, make the failure case first, with `break`, and eliminate the `continue`

src/coreclr/jit/morph.cpp
BruceForstall: I see you consult the config in `optAssertionInit`. Is there a reason to wait until then?

src/coreclr/jit/jitconfigvalues.h
AndyAyersMS: Sure... it was up with the other `Do` cases before, but now that I moved it to be available in release it no longer matches its neighbors.

src/coreclr/jit/morph.cpp
AndyAyersMS: I went back and forth as to whether all this setup should be assertion prop's business; I could push more of the logic there I suppose.

src/coreclr/jit/morph.cpp
AndyAyersMS: > Does this also need `optAssertionReset(0);`?

No. There are 3 modes of operation here:
1. classic: block starts with empty table via `Reset(0)` and empty BV
2. cross-block: block starts with current table and intersection of pred BVs (which may end up empty)
3. cross-block special case: block starts with current table and empty BV

Modes 2 and 3 can mix with each other, but not with mode 1.

We use mode 1 if the feature is disabled, or if the method has a lot of locals.

When the feature is enabled, and the method does not have a lot of locals, we use mode 2 (if `canUsePredAssertions`) or mode 3 (otherwise) depending on the various block properties.

> Can the `(!optCrossBlockLocalAssertionProp)` and the "failure" cases be combined (tail merged)?

Let me look at tweaking this.

BruceForstall (approved):

src/coreclr/jit/assertionprop.cpp
```diff
@@ -541,36 +541,83 @@ void Compiler::optAssertionTraitsInit(AssertionIndex assertionCount)
...
+        //
+        // The typical number of local assertions is roughly proportional
+        // to the number of locals. So when we have huge numbers of locals,
+        // just do within-block local assertion prop.
+        //
+        if (lvaCount > maxTrackedLocals)
+        {
+            JITDUMP("Disabling cross-block assertion prop: too many locals\n");
+            optCrossBlockLocalAssertionProp = false;
+        }
```
jakobbotsch, line 572: This seems like it might be very "spurious diff" inducing (add a local and suddenly a bunch of assertions stop being propagated). You might have mentioned it somewhere else, but do you have any plans to make this less "hard threshold-y"?

jakobbotsch: Also looking at things like `optAssertionDep` below it _kind of_ feels like we should be calling `lvaSortByRefCount` here, although I don't think ref counts are up to date here (in particular ones introduced by physical promotion -- might not be so hard to fix that).

jakobbotsch: I guess it's one of those cases where it would be great to have separated "single block" locals and "global" locals.

jakobbotsch (approved):
LGTM.

src/coreclr/jit/assertionprop.cpp
AndyAyersMS: > do you have any plans to make this less "hard threshold-y"?

I don't have any great ideas here, no. Some of this is the inherent cost of a blind forward scheme, we push facts along hoping they will matter. SSA would be helpful, we could just attach assertions to SSA lifetimes.

> utilize tracked indices

Some kind of prioritization scheme might work out nicely; presumably frequently referenced locals are more interesting candidates for assertions.

> I guess it's one of those cases where it would be great to have separated "single block" locals and "global" locals.

I had the same thought -- I [tried an experiment](https://github.com/dotnet/runtime/issues/93246#issuecomment-1800121754) to see how many assertions were killed in the same block they were born in, and it was quite small; most of them get killed at merges.

It might be possible to leverage RCS_EARLY to deduce if newly generated assertions are likely to be referenced later on; if not they could be purged before they "escape" the block. But we're also enabling cross-block copy prop so even getting that right might be tricky, as the ref counts can and will change during morph.

AndyAyersMS:
There's an unexpected diff on 32 bit platforms, I want to chase this down... not obvious to me what this is from the code changes. I can repro locally.

AndyAyersMS:
> There's an unexpected diff on 32 bit platforms, I want to chase this down... not obvious to me what this is from the code changes. I can repro locally.

Ah, this is fixing an issue I introduced in https://github.com/dotnet/runtime/pull/94322. In `optAssertionIsNonNullInternal` we were inadvertently removing an assertion from the dep vector if it was no longer live (via `IntersectionD`). If it later came back to life we might not be able to find it since it would not be re-added to the dep vector (adding only happens when an assertion is first created). So we might fail to kill it if then again went dead.

PR #94366, opened 2023-11-03
Emit exception blocks for the new ILGenerator

buyaa-n:
Adds implementation for `Label BeginExceptionBlock()`, `BeginCatchBlock(Type? exceptionType)`, `BeginExceptFilterBlock()`, `BeginFinallyBlock()`, `BeginFaultBlock()` and `EndExceptionBlock()` for the new ILGenerator implementation.

### DIFF
--- a/src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
+++ b/src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Reflection.Metadata;
 using System.Reflection.Metadata.Ecma335;
 using System.Runtime.InteropServices;
@@ -22,7 +23,6 @@ namespace System.Reflection.Emit
         private Dictionary<Label, LabelHandle> _labelTable = new(2);
         private List<KeyValuePair<MemberInfo, BlobWriter>> _memberReferences = new();
         private List<ExceptionBlock> _exceptionStack = new();
-        private int _currentExStackIndex = -1;
 
         internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
         {
@@ -43,12 +43,12 @@ namespace System.Reflection.Emit
 
         public override void BeginCatchBlock(Type? exceptionType)
         {
-            if (_currentExStackIndex <= -1)
+            if (_exceptionStack.Count < 1)
             {
                 throw new NotSupportedException(SR.Argument_NotInExceptionBlock);
             }
 
-            ExceptionBlock currentExBlock = _exceptionStack[_currentExStackIndex];
+            ExceptionBlock currentExBlock = _exceptionStack[_exceptionStack.Count - 1];
             if (currentExBlock.State == ExceptionState.Filter)
             {
                 // Filter block  should be followed by catch block with null exception type
@@ -87,12 +87,12 @@ namespace System.Reflection.Emit
 
         public override void BeginExceptFilterBlock()
         {
-            if (_currentExStackIndex <= -1)
+            if (_exceptionStack.Count < 1)
             {
                 throw new NotSupportedException(SR.Argument_NotInExceptionBlock);
             }
 
-            ExceptionBlock currentExBlock = _exceptionStack[_currentExStackIndex];
+            ExceptionBlock currentExBlock = _exceptionStack[_exceptionStack.Count - 1];
             Emit(OpCodes.Leave, currentExBlock.EndLabel);
             if (currentExBlock.State == ExceptionState.Try)
             {
@@ -121,18 +121,17 @@ namespace System.Reflection.Emit
             MarkLabel(currentExBlock.TryStart);
             currentExBlock.State = ExceptionState.Try;
             _exceptionStack.Add(currentExBlock);
-            _currentExStackIndex++;
             return currentExBlock.EndLabel;
         }
 
         public override void BeginFaultBlock()
         {
-            if (_currentExStackIndex <= -1)
+            if (_exceptionStack.Count < 1)
             {
                 throw new NotSupportedException(SR.Argument_NotInExceptionBlock);
             }
 
-            ExceptionBlock currentExBlock = _exceptionStack[_currentExStackIndex];
+            ExceptionBlock currentExBlock = _exceptionStack[_exceptionStack.Count - 1];
             Emit(OpCodes.Leave, currentExBlock.EndLabel);
             if (currentExBlock.State == ExceptionState.Try)
             {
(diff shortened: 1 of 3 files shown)


src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -38,11 +41,145 @@ internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
...
         public override int ILOffset => _il.Offset;
 
-        public override void BeginCatchBlock(Type? exceptionType) => throw new NotImplementedException();
-        public override void BeginExceptFilterBlock() => throw new NotImplementedException();
-        public override Label BeginExceptionBlock() => throw new NotImplementedException();
-        public override void BeginFaultBlock() => throw new NotImplementedException();
-        public override void BeginFinallyBlock() => throw new NotImplementedException();
+        public override void BeginCatchBlock(Type? exceptionType)
+        {
+            if (_currentExStackIndex <= -1)
```
AaronRobinsonMSFT, line 46: Is this expression important as is? As in, can this be changed to `_currentExStackIndex < 0`?

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -38,11 +41,145 @@ internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
...
+            {
+                throw new NotSupportedException(SR.Argument_NotInExceptionBlock);
+            }
+
+            ExceptionBlock currentExBlock = _exceptionStack[_currentExStackIndex];
+            if (currentExBlock.State == ExceptionState.Filter)
+            {
+                // Filter block  should be followed by catch block with null exception type
```
AaronRobinsonMSFT, line 54:
```suggestion
                // Filter block should be followed by catch block with null exception type
```

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
AaronRobinsonMSFT: Instead of comparing to a specific index, would it be more appropriate to compare against the length of `_exceptionStack`?

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -38,11 +41,145 @@ internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
...
+                    _labelTable[currentExBlock.HandleStart], _labelTable[currentExBlock.HandleEnd], module.GetTypeHandle(exceptionType));
+                MarkLabel(currentExBlock.HandleStart);
+            }
+
+            currentExBlock.State = ExceptionState.Catch;
+        }
+
+        public override void BeginExceptFilterBlock()
+        {
+            if (_currentExStackIndex <= -1)
```
AaronRobinsonMSFT, line 90: Same questions as `BeginCatchBlock()` above.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -38,11 +41,145 @@ internal ILGeneratorImpl(MethodBuilder methodBuilder, int size)
...
+            MarkLabel(currentExBlock.TryStart);
+            currentExBlock.State = ExceptionState.Try;
+            _exceptionStack.Add(currentExBlock);
+            _currentExStackIndex++;
+            return currentExBlock.EndLabel;
+        }
+
+        public override void BeginFaultBlock()
+        {
+            if (_currentExStackIndex <= -1)
```
AaronRobinsonMSFT, line 130: Same question.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -15,19 +14,23 @@ internal sealed class ILGeneratorImpl : ILGenerator
...
         private readonly InstructionEncoder _il;
+        private readonly ControlFlowBuilder _cfBuilder;
         private bool _hasDynamicStackAllocation;
         private int _maxStackSize;
         private int _currentStack;
         private List<LocalBuilder> _locals = new();
         private Dictionary<Label, LabelHandle> _labelTable = new(2);
         private List<KeyValuePair<MemberInfo, BlobWriter>> _memberReferences = new();
+        private List<ExceptionBlock> _exceptionStack = new();
+        private int _currentExStackIndex = -1;
```
AaronRobinsonMSFT, line 25: Should the `List<ExceptionBlock>` really be a `Stack<ExceptionBlock>`? It seems like that is precisely how it is being used.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
```diff
@@ -398,7 +535,40 @@ private static int GetStackChange(OpCode opcode, MethodInfo methodInfo, Type[]?
...
+            else if (state == ExceptionState.Finally || state == ExceptionState.Fault)
+            {
+                Emit(OpCodes.Endfinally);
+                MarkLabel(currentExBlock.HandleEnd);
+            }
+
+            MarkLabel(endLabel);
+            currentExBlock.State = ExceptionState.Done;
+            _exceptionStack.Remove(currentExBlock);
+            _currentExStackIndex--;
```
AaronRobinsonMSFT, line 569: During the removal, is there any state we should assert? Some internal state that we are tracking as opposed to what the user is expected to do.

AaronRobinsonMSFT (commented):
Is there a specific reference you are using?

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
buyaa-n: Good point, but seems I could not use `Stack<T>` from CoreLib

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
buyaa-n: > Instead of comparing to a specific index, would it be more appropriate to compare against the length of _exceptionStack?

Thought if will be clearer, than using `_exceptionStack.length`, though have no strong opinion.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
buyaa-n: Could assert the removed index is the last element in the list

buyaa-n: > Could assert the removed index is the last element in the list

Well now when using `_exceptionStack.Count` instead of tracking `_currentExStackIndex` separately there is no need to assert that, if that is what you meant.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ILGeneratorImpl.cs
buyaa-n: Ah, actually it seems you meant to compare the `_currentExStackIndex` with `_exceptionStack` length.

buyaa-n:
> Is there a specific reference you are using?

Existing RuntimeILGenerator, ECMA-335 spec, `Reflection.Emit` and `Reflection.MetadataBuilder` docs etc which ever I could find

jkotas (approved):
Thanks

PR #94369, opened 2023-11-04
update IServiceProvider.GetServices return value nullability

WeihanLi:
update `IServiceProvider.GetServices` return value nullability

update to `IEnumerable<object>` to align with the method return value type

https://github.com/dotnet/runtime/blob/b4823d05c38d7eb44de0db41ba9ba6c6a2df4e46/src/libraries/Microsoft.Extensions.DependencyInjection.Abstractions/src/ServiceProviderServiceExtensions.cs#L88-L95


stephentoub:
Is there an issue where this was discussed? What guarantees the yielded objects won't be null?

WeihanLi:
> Is there an issue where this was discussed?

No, I created an issue just now https://github.com/dotnet/runtime/issues/94374

PR #94370, opened 2023-11-04
`JsonSerializerOptions.Web` for `JsonSerializerOptions`

### LINKED ISSUE #92181: [API Proposal]: add `JsonSerializerOptions.Web` for `JsonSerializerOptions`
### Background and motivation

Propose a public `JsonSerializerOptions` for the web, it could help avoid some duplicate cache

For example, if we extend the `AsyncEnumerable` for legacy frameworks, we could use the shared `JsonSerializerOptions.Web` without creating a new `JsonSerializerOptions`, and the original cache in runtime/aspnetcore etc would also benifit from this

`new JsonSerializerOptions(JsonSerializerDefaults.Web)` usage from grep:

https://grep.app/search?q=JsonSerializerOptions%28JsonSerializerDefaults.Web%29&filter[lang][0]=C%23

This may also help with the new analyzer rule

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1869

### API Proposal

```diff
namespace System.Text.Json;

public sealed partial class JsonSerializerOptions
{
    public static JsonSerializerOptions Default { get; }
+   public static JsonSerializerOptions Web { get; }
}
```

### API Usage

```csharp
using var httpClient = new HttpClient();
httpClient.BaseAddress = new Uri("http://localhost:5297");
await using var responseStream = await httpClient.GetStreamAsync("api/Jobs");
// var stream = JsonSerializer.DeserializeAsyncEnumerable<Job>(responseStream, new JsonSerializerOptions(JsonSerializerDefaults.Web));
var stream = JsonSerializer.DeserializeAsyncEnumerable<Job>(responseStream, JsonSerializerOptions.Web);
```

### Alternative Designs

_No response_

### Risks

_No response_

I-SER-I:
Close #92181


src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializerOptions.cs
```diff
@@ -956,6 +981,20 @@ private static JsonSerializerOptions GetOrCreateDefaultOptionsInstance()
...
+        private static JsonSerializerOptions GetOrCreateWebOptionsInstance()
+        {
+            var options = new JsonSerializerOptions
+            {
+                _propertyNameCaseInsensitive = true,
+                _jsonPropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+                _numberHandling = JsonNumberHandling.AllowReadingFromString
+            };
+
+            return Interlocked.CompareExchange(ref s_webOptions, options, null) ?? options;
```
stephentoub, line 995:
```suggestion
            return Interlocked.CompareExchange(ref s_webOptions, options, null) ?? s_webOptions;
```
That way, the property always returns the same published instance.

I-SER-I:
May i change [all statement in dontent/runtime](https://grep.app/search?q=JsonSerializerOptions%28JsonSerializerDefaults.Web%29&filter[lang][0]=C%23&filter[repo][0]=dotnet/runtime) `new JsonSerializerOptions(JsonSerializerDefaults.Web)` to `JsonSerializerOptions.Web`?

src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializerOptions.cs
```diff
@@ -956,6 +981,20 @@ private static JsonSerializerOptions GetOrCreateDefaultOptionsInstance()
...
+        [RequiresUnreferencedCode(JsonSerializer.SerializationUnreferencedCodeMessage)]
+        [RequiresDynamicCode(JsonSerializer.SerializationRequiresDynamicCodeMessage)]
+        private static JsonSerializerOptions GetOrCreateWebOptionsInstance()
+        {
+            var options = new JsonSerializerOptions
+            {
+                _propertyNameCaseInsensitive = true,
+                _jsonPropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+                _numberHandling = JsonNumberHandling.AllowReadingFromString
+            };
```
stephentoub, line 993: Rather than open code the same set of options, can this just be `new JsonSerializerOptions(JsonSerializerDefaults.Web)`?

stephentoub:
> May i change [all statement in dontent/runtime](https://grep.app/search?q=JsonSerializerOptions%28JsonSerializerDefaults.Web%29&filter%5Blang%5D%5B0%5D=C%23&filter%5Brepo%5D%5B0%5D=dotnet/runtime) `new JsonSerializerOptions(JsonSerializerDefaults.Web)` to `JsonSerializerOptions.Web`?

Yes, where possible, but most of these cases would involve removing the static readonly field entirely and just changing call sites to use the new property directly.

src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializerOptions.cs
eiriktsarpalis: This is also exposing the singleton as a mutable instance. Refer to the `Default` factory on how to make it read-only.

src/libraries/System.Text.Json/tests/System.Text.Json.Tests/Serialization/OptionsTests.cs
```diff
@@ -1253,7 +1253,7 @@ public static void DefaultSerializerOptions_General()
         [Fact]
         public static void PredefinedSerializerOptions_Web()
         {
-            var options = new JsonSerializerOptions(JsonSerializerDefaults.Web);
+            var options = JsonSerializerOptions.Web;
```
eiriktsarpalis, line 1256: This test is checking the behavior of the constructor and shouldn't be changed. Consider adding a separate test for the singleton instead.

src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializerOptions.cs
```diff
@@ -956,6 +981,14 @@ private static JsonSerializerOptions GetOrCreateDefaultOptionsInstance()
             return Interlocked.CompareExchange(ref s_defaultOptions, options, null) ?? options;
         }
 
+        [RequiresUnreferencedCode(JsonSerializer.SerializationUnreferencedCodeMessage)]
+        [RequiresDynamicCode(JsonSerializer.SerializationRequiresDynamicCodeMessage)]
+        private static JsonSerializerOptions GetOrCreateWebOptionsInstance()
+        {
+            var options = new JsonSerializerOptions(JsonSerializerDefaults.Web) { _isReadOnly = true };
```
eiriktsarpalis, line 988: You also need to set the `TypeInfoResolver` like the above method, otherwise this won't work as expected.

src/libraries/System.Text.Json/tests/System.Text.Json.Tests/Serialization/OptionsTests.cs
```diff
@@ -1036,6 +1036,31 @@ public static void JsonSerializerOptions_Default_IsReadOnly()
             Assert.Same(resolver, optionsSingleton.TypeInfoResolver);
         }
 
+        [Fact]
+        public static void JsonSerializerOptions_Web_MatchesConstructorWithJsonSerializerDefaults()
```
eiriktsarpalis, line 1040: Please also add a test that attempts to serialize a value using the new singleton. I don't believe this would work with the current implementation.

src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializerOptions.cs
```diff
@@ -52,6 +52,31 @@ public static JsonSerializerOptions Default
...
+            [RequiresUnreferencedCode(JsonSerializer.SerializationUnreferencedCodeMessage)]
+            [RequiresDynamicCode(JsonSerializer.SerializationRequiresDynamicCodeMessage)]
+            get
+            {
+                if (s_webOptions is not JsonSerializerOptions options)
+                {
+                    options = GetOrCreateWebOptionsInstance();
+                }
+
+                return options;
```
eiriktsarpalis, line 74: This could be simplified:
```suggestion
                return s_webOptions is JsonSerializerOptions options : GetOrCreateWebOptionsInstance();
```

If you could also update the existing method to follow the same style that would be great 👍

eiriktsarpalis:
@I-SER-I would it be possible to address the feedback when possible? Thanks!

eiriktsarpalis (approved):

PR #94378, opened 2023-11-05
Clean up S.N.Tensors project in main

### LINKED ISSUE #93939: Remove net6.0;net7.0 from System.Numerics.Tensors
In https://github.com/dotnet/runtime/pull/92269 we re-enabled `net6.0;net7.0` to make it easier to backport changes from main to 8.0.  We should remove those before we ship 9.0 since these frameworks will be out of support by the time it ships. cc @ViktorHofer

stephentoub:
- Remove stale build targets
- Remove ifdefs for .NET 8
- Remove stale IVT (I'm not sure why this was needed in the first place)

Fixes https://github.com/dotnet/runtime/issues/93939


src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
```diff
@@ -54,7 +54,6 @@ public static void ConvertToHalf(ReadOnlySpan<float> source, Span<Half> destinat
             ref ushort destinationRef = ref Unsafe.As<Half, ushort>(ref MemoryMarshal.GetReference(destination));
             int i = 0, twoVectorsFromEnd;
 
-#if NET8_0_OR_GREATER
```
tannergooding, line 57: How much more are we able to "clean" this up in main?

Are we considering switching over to using `ISimdVector` and similar to reduce code duplication between the paths or are we only minimally differing so that backporting isn't problematic if still required?

tannergooding (approved):

src/libraries/System.Numerics.Tensors/src/System/Numerics/Tensors/TensorPrimitives.netcore.cs
stephentoub: > Are we considering switching over to using ISimdVector and similar to reduce code duplication between the paths or are we only minimally differing so that backporting isn't problematic if still required?

ISimdVector is currently internal to corelib.

src/libraries/System.Numerics.Tensors/src/System.Numerics.Tensors.csproj
```diff
@@ -2,14 +2,9 @@
...
   <PropertyGroup>
     <TargetFrameworks>$(NetCoreAppCurrent);$(NetCoreAppPrevious);$(NetCoreAppMinimum);netstandard2.0;$(NetFrameworkMinimum)</TargetFrameworks>
-    <TargetFrameworks Condition="'$(DotNetBuildFromSource)' != 'true'">$(TargetFrameworks);net7.0;net6.0</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <IsPackable>true</IsPackable>
     <PackageDescription>Provides support for operating over tensors.</PackageDescription>
-    <!-- Disabling baseline validation since this is a brand new package.
-         Once this package has shipped a stable version, the following line
-         should be removed in order to re-enable validation. -->
-    <DisablePackageBaselineValidation>true</DisablePackageBaselineValidation>
```
stephentoub, line 12: @ViktorHofer, how do I create an appropriate baseline in order to fix the validation errors?

ViktorHofer: From the build log:

> API breaking changes found. If those are intentional, the APICompat suppression file can be updated by rebuilding with '/p:ApiCompatGenerateSuppressionFile=true'

stephentoub: @ViktorHofer, which project?  I've tried building both the ref and the src with that, and it has no affect.

ViktorHofer: The packable project, the source project. This should do it:

`dotnet pack src\libraries\System.Numerics.Tensors\src /p:ApiCompatGenerateSuppressionFile=true`.

src/libraries/System.Numerics.Tensors/src/CompatibilitySuppressions.xml
```diff
@@ -0,0 +1,214 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
ViktorHofer, line 1: Please add the following lines to the suppression file (change the assembly names): https://github.com/dotnet/runtime/blob/6d76479447b67e119173fa4cd3acb3a2e1887cd3/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/CompatibilitySuppressions.xml#L4-L14

This will make sure that APICompat doesn't emit errors for TFMs that we intentionally dropped in the current package (net6.0 and net7.0). We added these lines to all our packable projects in main as we currently don't have mechanism to suppress those TFMs globally. That work is tracked via https://github.com/dotnet/sdk/issues/22901

You can then remove the following entries:
- left: `lib/net6.0 or `lib/net7.0` compared against
- right: `lib/netstandard2.0`

cc @ericstj as there is quite a lot of room for improvements

ViktorHofer (approved):

src/libraries/System.Numerics.Tensors/src/CompatibilitySuppressions.xml
ericstj: @ViktorHofer if you see room for improvements file issues for those - I'm not certain what in particular you were referring to here.

ViktorHofer: https://github.com/dotnet/sdk/issues/22901 which Santiago filed two years ago. Especially now, as we decided to suppress net6.0 and net7.0 compatibility errors in all packable projects.

PR #94387, opened 2023-11-05
Merge return true/false blocks

EgorBo:
I plan to unconditionally expand all `return <condition>` blocks to `JTRUE`, merge tails (this PR) and then fold to `return <condition>;` back if tails were not found.

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=463538&view=ms.vss-build-web.run-extensions-tab)

### DIFF
--- a/src/coreclr/jit/fgopt.cpp
+++ b/src/coreclr/jit/fgopt.cpp
@@ -6729,69 +6729,7 @@ PhaseStatus Compiler::fgHeadTailMerge(bool early)
     // If return value is true, retry.
     // May also add to retryBlocks.
     //
-    auto tailMerge = [&](BasicBlock* block) -> bool {
-
-        if (block->countOfInEdges() < 2)
-        {
-            // Nothing to merge here
-            return false;
-        }
-
-        predInfo.Reset();
-
-        // Find the subset of preds that reach along non-critical edges
-        // and populate predInfo.
-        //
-        for (BasicBlock* const predBlock : block->PredBlocks())
-        {
-            if (predBlock->GetUniqueSucc() != block)
-            {
-                continue;
-            }
-
-            if (!BasicBlock::sameEHRegion(block, predBlock))
-            {
-                continue;
-            }
-
-            Statement* lastStmt = predBlock->lastStmt();
-
-            // Block might be empty.
-            //
-            if (lastStmt == nullptr)
-            {
-                continue;
-            }
-
-            // Walk back past any GT_NOPs.
-            //
-            Statement* const firstStmt = predBlock->firstStmt();
-            while (lastStmt->GetRootNode()->OperIs(GT_NOP))
-            {
-                if (lastStmt == firstStmt)
-                {
-                    // predBlock is evidently all GT_NOP.
-                    //
-                    lastStmt = nullptr;
-                    break;
-                }
-
-                lastStmt = lastStmt->GetPrevStmt();
-            }
-
-            // Block might be effectively empty.
-            //
-            if (lastStmt == nullptr)
-            {
-                continue;
-            }
-
-            // We don't expect to see PHIs but watch for them anyways.
-            //
-            assert(!lastStmt->IsPhiDefnStmt());
-            predInfo.Emplace(predBlock, lastStmt);
-        }
-
+    auto tailMergePreds = [&](BasicBlock* commSucc) -> bool {
         // Are there enough preds to make it interesting?
         //
         if (predInfo.Height() < 2)
@@ -6842,9 +6780,9 @@ PhaseStatus Compiler::fgHeadTailMerge(bool early)
             // We have some number of preds that have identical last statements.
             // If all preds of block have a matching last stmt, move that statement to the start of block.
             //
-            if (matchedPredInfo.Height() == (int)block->countOfInEdges())
+            if ((commSucc != nullptr) && (matchedPredInfo.Height() == (int)commSucc->countOfInEdges()))
             {
-                JITDUMP("All preds of " FMT_BB " end with the same tree, moving\n", block->bbNum);
+                JITDUMP("All preds of " FMT_BB " end with the same tree, moving\n", commSucc->bbNum);
                 JITDUMPEXEC(gtDispStmt(matchedPredInfo.TopRef(0).m_stmt));
 
                 for (int j = 0; j < matchedPredInfo.Height(); j++)
@@ -6860,8 +6798,8 @@ PhaseStatus Compiler::fgHeadTailMerge(bool early)
                     //
                     if (j == 0)
                     {
-                        fgInsertStmtAtBeg(block, stmt);
-                        block->bbFlags |= predBlock->bbFlags & BBF_COPY_PROPAGATE;
+                        fgInsertStmtAtBeg(commSucc, stmt);
+                        commSucc->bbFlags |= predBlock->bbFlags & BBF_COPY_PROPAGATE;
                     }
 
                     madeChanges = true;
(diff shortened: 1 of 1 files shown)


src/coreclr/jit/fgopt.cpp
```diff
@@ -7030,10 +7030,51 @@ PhaseStatus Compiler::fgHeadTailMerge(bool early)
...
+    BasicBlock* returnTrueBb  = nullptr;
+    BasicBlock* returnFalseBb = nullptr;
+
     // Visit each block and try to merge first statements of successors.
     //
     for (BasicBlock* const block : Blocks())
     {
+        // Also, merge "return true/false" blocks regardless of where they are in the graph (but in the same EH region).
+        // Technically, we can generalize this algorithm to merge any "return CNS" blocks, but it seems to be profitable
+        // only for booleans (might be also profitable for 'return null;').
```
jakobbotsch, line 7042: > it seems to be profitable only for booleans 

Hmm, why is that?

The transformation is really a form of tail merging, so it seems odd that we do it during this loop that handles head merging. The tail merging loop above has a bunch of smart stuff in it to decide how to merge tails -- doesn't it help for this case too? What if you just build the normal `predInfo` for return blocks and then run the normal tail merge with it?

jakobbotsch: @AndyAyersMS was hinting last week that he might do this work (https://github.com/dotnet/runtime/pull/94130#issuecomment-1783854877), so he probably has some thoughts too.

EgorBo: > Hmm, why is that? 

Tried to allow it for `return null;` but only got a few diffs, I can try to generalize the algorithm to be `HashSet<BasicBlock>` + support multiple statements (split them) but I fear that will be a lot more code + TP impact with  small wins on top of this.

> The transformation is really a form of tail merging, so it seems odd that we do it during this loop that handles head merging.

I just needed some loop over blocks to check the diffs

jakobbotsch: > I can try to generalize the algorithm to be HashSet<BasicBlock> + support multiple statements (split them) but I fear that will be a lot more code

Shouldn't the logic be basically the same as what is already there for tail merging already? Ideally we generalize that logic to handle returns instead of adding another separate implementation for it. The current implementation relies on a successor block to have a predecessor list. I think to generalize it we just build the `predInfo` directly for all return blocks (as if they had a pseudo successor).

I suppose the thing with return blocks is that we aren't going to have natural preexisting jumps to redirect, so there are additional heuristics necessary to determine profitability in the general case. I can see that the existing logic already has a preference towards picking blocks such they [won't need to be split](https://github.com/dotnet/runtime/blob/e733539a38458c5c3a64688b55a702444e470700/src/coreclr/jit/fgopt.cpp#L6900). That seems like one of the primary things in the implementation of this PR, so maybe require it for the new `BBJ_RETURN` case if that's the necessary profitability heuristic?

What I'm really after is making sure that we unify things to not end up with two different code paths that try to do basically the same thing.

EgorBo: thanks! Will take a look!

EgorBo: My main motivation is to unconditionally expand `return <cond>;` early to `JTRUE ret true/false` so then other opts like "optimize bool" can only focus on JTRUE (sort of cannonical form for all branchless opts, etc) and the redundant branch opt can benefit from this too.

Then, this opt merges all `return true/false` tails and then some other phases merges
```
JTRUE
return false
return true
```
back to RETURN if no tails were merged.

AndyAyersMS: I think in the long run we should consider merging all returns early (save perhaps for tail call candidates) and keep things merged until quite late, since we don't know the actual epilog costs until after LSRA.

I also wonder if there is value in generalizing this sort of thing, say any time we assign the result of a relop to a local at the end of a non-branching block that targets a join (where the return successor is an implicit kind of join), we do this expansion and then let tail merge see if it can clean things up....

AndyAyersMS: That would also expose the predicates to RBO... it currently doesn't do much for predicates that don't feed branches.

EgorBo:
@AndyAyersMS @jakobbotsch @dotnet/jit-contrib PTAL, the current shape is what @jakobbotsch suggested (generalized), my initial one was focusing on `return <bool>` and turned out to be too conservative.

I've limitted it to single-statement RETURN blocks, but it seems like it makes sense to enable it for all, but that will require some careful analysis of TP/Size regressions + needs a small fix in morph for explicit tail calls (I or Jakob will try to land it separately).
There 3 source of regressions:
1) Loop clonning (esp in aspnet collection), example is https://www.diffchecker.com/doR6p4zH/ (checked via JitDump that it's indeed loop clonning)
2) In some cases we no longer emit branchless cmove, etc due to a different layout of blocks
3) In some cases we light up a frame in x64 leaf functions, my understanding that it's because of that BB count heuristic

but overall it's a clear size win: https://dev.azure.com/dnceng-public/public/_build/results?buildId=463538&view=results

I also expect to see some improvements when I expand `return <compare>` into jtrue separately

src/coreclr/jit/block.cpp
```diff
@@ -853,12 +853,30 @@ Statement* BasicBlock::firstStmt() const
...
 {
-    return (firstStmt() != nullptr) && (firstStmt() == lastStmt());
+    if (firstStmt() == nullptr)
+    {
+        // No statements at all
+        return false;
+    }
+
+    const Statement* stmt = firstStmt();
+    while (ignoreNopStatements && ((stmt->GetRootNode() == nullptr) || (stmt->GetRootNode()->OperIs(GT_NOP))))
```
jakobbotsch, line 871: Is a statement with a null node possible? Overall, do we actually find any NOPs in the return blocks?

src/coreclr/jit/fgopt.cpp
```diff
@@ -6729,69 +6729,7 @@ PhaseStatus Compiler::fgHeadTailMerge(bool early)
...
-                continue;
-            }
-
-            // We don't expect to see PHIs but watch for them anyways.
-            //
-            assert(!lastStmt->IsPhiDefnStmt());
-            predInfo.Emplace(predBlock, lastStmt);
-        }
-
+    auto tailMergePreds = [&](BasicBlock* commSucc) -> bool {
```
jakobbotsch, line 6732: It would be nice to convert these to functions instead of massive lambdas... but that's a no-diff refactoring to be done separately.

src/coreclr/jit/block.cpp
EgorBo: I wasn't sure in this, but if I remove the nullcheck it won't fail on CI so I presume it's not possible? I wish we could have a centralize "IR reference" where we could check what's legal and what's not in IR

EgorBo: > Is a statement with a null node possible? Overall, do we actually find any NOPs in the return blocks?

I'll check the diffs

SingleAccretion: It is not possible.

src/coreclr/jit/gentree.cpp
```diff
@@ -2519,6 +2519,12 @@ bool GenTreeCall::Equals(GenTreeCall* c1, GenTreeCall* c2)
         return false;
     }
 
+    // Make sure "explicit tail call" flags match.
+    if (c1->IsTailPrefixedCall() != c2->IsTailPrefixedCall())
```
jakobbotsch, line 2523: I think this should be reverted (and fixed in the future PR), or we should check all the tailcall flags.

AndyAyersMS:
> I've limitted it to single-statement RETURN blocks, but it seems like it makes sense to enable it for all, but that will require some careful analysis of TP/Size regressions (up to +0.4% TP impact) + needs a small fix in morph for explicit tail calls (I or Jakob will try to land it separately).

I'm surprised checkin more statements has such a TP impact; it should be fairly self-limiting as there can't be that many return blocks. Unless it's enabling a lot more merging?

AndyAyersMS (commented):
I don't have anything to add other that what's already been noted.

Cool that this was relatively refactoring simple; I recall working on this a while back but got bogged down for some reason.

src/coreclr/jit/block.cpp
EgorBo: Reverted these changes, there were no additional diffs from that

src/coreclr/jit/gentree.cpp
EgorBo: Reverted

EgorBo:
> I'm surprised checkin more statements has such a TP impact; 

it's possible that the TP regressions in that case come from extra loop clonning etc, I'll check separately, this PR will help to isolate those changes

jakobbotsch (approved):

PR #94389, opened 2023-11-05
[wasm] WasmAppHost: Fix passing message as paramName

carlreinke:
The [`ArgumentNullException`](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception.-ctor#system-argumentnullexception-ctor(system-string)) constructor with one parameter takes a parameter name, not a message.  ~~Since it throws for empty as well as `null`, I changed it to throw `ArgumentException` instead.~~

As per feedback, it now throws `CommandLineException`.

### DIFF
--- a/src/mono/wasm/host/CommonConfiguration.cs
+++ b/src/mono/wasm/host/CommonConfiguration.cs
@@ -125,18 +125,18 @@ internal sealed class CommonConfiguration
     public static void CheckPathOrInAppPath(string appPath, string? path, string argName)
     {
         if (string.IsNullOrEmpty(path))
-            throw new ArgumentNullException($"Missing value for {argName}");
+            throw new CommandLineException($"Missing value for {argName}");
 
         if (Path.IsPathRooted(path))
         {
             if (!File.Exists(path))
-                throw new ArgumentException($"Cannot find {argName}: {path}");
+                throw new CommandLineException($"Cannot find {argName}: {path}");
         }
         else
         {
             string fullPath = Path.Combine(appPath, path);
             if (!File.Exists(fullPath))
-                throw new ArgumentException($"Cannot find {argName} {path} in app directory {appPath}");
+                throw new CommandLineException($"Cannot find {argName} {path} in app directory {appPath}");
         }
     }
 }


src/mono/wasm/host/CommonConfiguration.cs
```diff
@@ -125,7 +125,7 @@ public ProxyOptions ToProxyOptions()
     public static void CheckPathOrInAppPath(string appPath, string? path, string argName)
     {
         if (string.IsNullOrEmpty(path))
-            throw new ArgumentNullException($"Missing value for {argName}");
+            throw new ArgumentException($"Missing value for {argName}");
```
stephentoub, line 128: Can this instead just be:
```C#
ArgumentException.ThrowIfNullOrEmpty(argName);
```
?

carlreinke: I think the message does provide value.  It's indicating that there is a property missing in "runtimeconfig.json".

stephentoub:
```C#
string argName = "runtimeconfig.json";

try
{
    throw new ArgumentException($"Missing value for {argName}");
}
catch (Exception e)
{
    Console.WriteLine(e.Message);
}

try
{
    ArgumentException.ThrowIfNullOrEmpty("", argName);
}
catch (Exception e)
{
    Console.WriteLine(e.Message);
}
```
outputs:
```
Missing value for runtimeconfig.json
The value cannot be an empty string. (Parameter 'runtimeconfig.json')
```
The latter is pretty clear about the problem.

I don't have strong feelings about it. Just doesn't seem worth the extra code.

pavelsavara:
cc @maraf @radical

src/mono/wasm/host/CommonConfiguration.cs
maraf: I would prefer using `CommandLineException` instead of `Argument**` for all occurences in this method. We have a catch for `CommandLineException` and show the message without stacktrace etc https://github.com/dotnet/runtime/blob/e733539a38458c5c3a64688b55a702444e470700/src/mono/wasm/host/Program.cs#L52-L56

maraf (commented):
@carlreinke Thank you for catching this one!

radical (approved):

Issue #94011, opened 2023-10-26
[API Proposal]: Arm64: FEAT_SVE: stores


a74nh:
```csharp
namespace System.Runtime.Intrinsics.Arm;

/// VectorT Summary
public abstract partial class Sve : AdvSimd /// Feature: FEAT_SVE  Category: stores
{

  /// T: float, double, sbyte, short, int, long, byte, ushort, uint, ulong
  public static unsafe void Store(Vector<T> mask, T* address, Vector<T> data); // ST1W or ST1D or ST1B or ST1H

  /// T: [short, sbyte], [int, short], [int, sbyte], [long, short], [long, int], [long, sbyte]
  /// T: [ushort, byte], [uint, ushort], [uint, byte], [ulong, ushort], [ulong, uint], [ulong, byte]
  public static unsafe void StoreNarrowing(Vector<T> mask, T2* address, Vector<T> data); // ST1B or ST1H or ST1W

  /// T: float, double, sbyte, short, int, long, byte, ushort, uint, ulong
  public static unsafe void StoreNonTemporal(Vector<T> mask, T* address, Vector<T> data); // STNT1W or STNT1D or STNT1B or STNT1H

  /// T: float, double, sbyte, short, int, long, byte, ushort, uint, ulong
  public static unsafe void Storex2(Vector<T> mask, T* address, (Vector<T> Value1, Vector<T> Value2)); // ST2W or ST2D or ST2B or ST2H

  /// T: float, double, sbyte, short, int, long, byte, ushort, uint, ulong
  public static unsafe void Storex3(Vector<T> mask, T* address, (Vector<T> Value1, Vector<T> Value2, Vector<T> Value3)); // ST3W or ST3D or ST3B or ST3H

  /// T: float, double, sbyte, short, int, long, byte, ushort, uint, ulong
  public static unsafe void Storex4(Vector<T> mask, T* address, (Vector<T> Value1, Vector<T> Value2, Vector<T> Value3, Vector<T> Value4)); // ST4W or ST4D or ST4B or ST4H

  /// total method signatures: 17
}
```

a74nh:
```csharp

/// Full API
public abstract class Sve : AdvSimd /// Feature: FEAT_SVE  Category: stores
{
    /// StoreVector : Non-truncating store

[...]

PR #94400, opened 2023-11-06
[LoongArch64] add Intrinsics' API for LoongArch64.

shushanhf:
We have finished the SIMD on the runtime6.0 and the tests passed.

I will push the SIMD for LoongArch64.

This is the first PR about the API's name.

The  [[API Proposal]: LoongArch64: add Intrinsics' API for LoongArch64](https://github.com/dotnet/runtime/issues/94445#top)
#94445

@tannergooding
Can you give me some advices ?
Thanks


shushanhf:
@tannergooding
Can you give me some advices ?

This is just the API's name, and first focus on the class name and the API name.

Later I will update this PR to amend some details.

Thanks

huoyaoyuan:
As a new architecture, it's more risky to expose public APIs comparing to mature architectures. I'd suggest keeping them internal, and focusing on cross-platform Vector128/256 intrinsics now.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA64Base.PlatformNotSupported.cs
```diff
@@ -0,0 +1,217 @@
...
+    /// <summary>
+    /// This class provides access to the LA64 base hardware instructions via intrinsics
+    /// </summary>
+    [CLSCompliant(false)]
+#if SYSTEM_PRIVATE_CORELIB
+    public
+#else
+    internal
+#endif
+    abstract class LA64Base
```
shushanhf, line 18: The name `LA64Base` is about some instructions within the LoongArch64's general ISA.

huoyaoyuan:
For API names, you can open API proposal like #94011. API definition without JIT implementation should be unwanted.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA_LASX.cs
```diff
@@ -0,0 +1,2269 @@
...
+using System.Runtime.CompilerServices;
+
+namespace System.Runtime.Intrinsics.LoongArch64
+{
+    /// <summary>
+    /// This class provides access to the LASX-256bits hardware instructions via intrinsics
+    /// </summary>
+    [Intrinsic]
+    [CLSCompliant(false)]
+    public abstract class LA_LASX : LA_LSX
```
shushanhf, line 13: The `LSX` is the abbreviation of Loongson SIMD Extension which is SIMD-128bits.
The `LASX` is the abbreviation of Loongson Advanced SIMD Extension which is SIMD-256bits.

shushanhf:
> For API names, you can open API proposal like #94011. API definition without JIT implementation should be unwanted.

If the API is OK for LoongArch64, I will push the JIT implementation.

shushanhf:
> As a new architecture, it's more risky to expose public APIs comparing to mature architectures. I'd suggest keeping them internal, and focusing on cross-platform Vector128/256 intrinsics now.

yes, the  Vector128/256  is independent of the CPU.

Now the API for architecture is the most important for LoongArch64, I want to confirm them for LoongArch64.
And then I will push all the Hardware-Intrinsics and SIMD.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA64Base.PlatformNotSupported.cs
```diff
@@ -0,0 +1,217 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+#pragma warning disable IDE0060 // unused parameters
+using System.Runtime.CompilerServices;
+
+namespace System.Runtime.Intrinsics.LoongArch64
```
tannergooding, line 7: I've marked this as `NO-MERGE` since we cannot take it until after an API review has occurred. See https://github.com/dotnet/runtime/blob/main/docs/project/api-review-process.md

We need an API proposal, following the standard template, created first. We'll have the discussion on relevant name changes and other bits there, then I can then champion that and take it to API review. Once approved, we can then implement the API surface.

Until then, LoongArch would be relegated to only supporting the existing cross platform API surface. For example, `Leading/TrailingZeroCount` can be supported by accelerating `int.Leading/TrailingZeroCount` and the same methods on the other primitive types.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA_LASX.cs
tannergooding: We likely would just call these `Lasx` and `Lsx`, to match the normal .NET naming convention and roughly what we've done for other platforms.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA_LASX.cs
```diff
@@ -0,0 +1,2269 @@
...
+        /// float32x8_t xvfmin_s_f32 (float32x8_t a, float32x8_t b)
+        ///   LASX: XVFMIN.S Xd.8S, Xj.8S, Xk.8S
+        /// </summary>
+        public static Vector256<float> Min(Vector256<float> left, Vector256<float> right) => Min(left, right);
+
+        /// <summary>
+        /// float64x4_t xvfmin_d_f64 (float64x4_t a, float64x4_t b)
+        ///   LASX: XVFMIN.D Xd.4D, Xj.4D, Xk.4D
+        /// </summary>
+        public static Vector256<double> Min(Vector256<double> left, Vector256<double> right) => Min(left, right);
```
tannergooding, line 929: What's the semantics around `NaN` and `-0` handling on LoongArch?

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA64Base.cs
```diff
@@ -0,0 +1,212 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.CompilerServices;
+
+namespace System.Runtime.Intrinsics.LoongArch64
```
tannergooding, line 6: Is there any `LoongArch32` that needs to be considered? If so, does it essentially have the same set of SIMD instructions, etc?

For Arm and x86, we have a single namespace encompassing both the 32-bit and 64-bit processors. The unique ISAs are then exposed as `Sse.X64.Method()` or `AdvSimd.Arm32.Method()` and `AdvSimd.Arm64.Method()`

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA64Base.PlatformNotSupported.cs
shushanhf: Thanks !
Reviewing the API for LoongArch64 based on a PR maybe more clear. So I pushed this PR.
I will create an API proposal for LoongArch64's API.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA64Base.cs
shushanhf: > Is there any `LoongArch32` that needs to be considered? If so, does it essentially have the same set of SIMD instructions, etc?

If just considering the current CoreCLR's implementation which only supports the LoongArch64, of course the SIMD extension is only  binding with the 64-bits CPU.
Here using `LoongArch64` directly just for there is no a LoongArch32's implementation CoreCLR to test the SIMD.

I will update it as `namespace System.Runtime.Intrinsics.LoongArch` liking Arm and x86.
Thanks.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA_LASX.cs
shushanhf: OK, I will update these later.
Thanks.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA_LASX.cs
```diff
@@ -0,0 +1,2269 @@
...
+        /// float64x4_t xvffint_d_lu_f64_u64 (uint64x4_t a)
+        ///   LASX: XVFFINT.D.LU Xd.4D, Xj.4D
+        /// </summary>
+        public static Vector256<double> ConvertToDouble(Vector256<ulong> value) => ConvertToDouble(value);
+
+        /// <summary>
+        /// bool xvsetnez_v_u8 (uint8x32_t value)
+        ///   LASX: XVSETNEZ.V cd, Xj.32B
+        /// </summary>
+        public static bool HasElementsNotZero(Vector256<byte> value) => HasElementsNotZero(value);
```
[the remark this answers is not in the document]
shushanhf: I will answer these together later.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch/LoongArchBase.cs
```diff
@@ -0,0 +1,198 @@
...
+using System.Runtime.CompilerServices;
+
+namespace System.Runtime.Intrinsics.LoongArch
+{
+    /// <summary>
+    /// This class provides access to the LoongArch base hardware instructions via intrinsics
+    /// </summary>
+    [Intrinsic]
+    [CLSCompliant(false)]
+    public abstract class LoongArchBase
```
shushanhf, line 13: * Rename this file as `LoongArchBase.cs`, is it OK? 
Or Just name this file as `LABase.cs`?
* Naming this class as `LoongArchBase`, is it OK ?

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch/LoongArchBase.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            /// <summary>
+            ///   LA64: CLO.W rd, rj
+            /// </summary>
+            public static int LeadingSignCount(int value) => LeadingSignCount(value);
+
+            /// <summary>
+            ///   LA64: CLO.W rd, rj
+            /// </summary>
+            public static int LeadingSignCount(uint value) => LeadingSignCount(value);
```
shushanhf, line 34: Is it needed to add two types API with `int value` and `uint value` ?

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch/LoongArchBase.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            /// <summary>
+            ///   LA64: BITREV.W rd, rj
+            /// </summary>
+            public static long ReverseElementBits(int value) => ReverseElementBits(value);
+
+            /// <summary>
+            ///   LA64: BITREV.W rd, rj
+            /// </summary>
+            public static ulong ReverseElementBits(uint value) => ReverseElementBits(value);
```
shushanhf, line 134: Is it needed to add the `int value` and `uint value` for the API `ReverseElementBits()` ?

shushanhf (commented):
@tannergooding

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch/LoongArchBase.cs
```diff
@@ -0,0 +1,198 @@
...
+
+            /// <summary>
+            ///   LA64: REVB.D rd, rj
+            /// </summary>
+            public static long ReverseElementBits(long value) => ReverseElementBits(value);
+
+            /// <summary>
+            ///   LA64: REVB.D rd, rj
+            /// </summary>
+            public static ulong ReverseElementBits(ulong value) => ReverseElementBits(value);
```
shushanhf, line 154: These are part of instructions liking the Arm64's `REV, REV16, REV32, REV64`, but the `ArmBase class` doesn't support these, Why?

Is it needed to add these for LoongArch64.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA_LASX.cs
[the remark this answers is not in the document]
shushanhf: > How does this instruction work at the hardware level?
> 
> `Xj.32B` is clearly the input register, but I'm not familiar with `cd` here. Is it a general purpose register, a flag register, something else?

The `cd` is a float flag register which indicating the floats comparing results.
There are 8 `cd` float flag registers.

Of course here I didn't expose the `cd` the API just for simple usage.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA_LASX.cs
shushanhf: The float operation is implemented within the IEEE-754-2008, here is MinNum(x,y).

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA_LASX.cs
```diff
@@ -0,0 +1,2269 @@
...
+        /// float64x4_t xvld_f64 (float64_t const * ptr)
+        ///   LASX: XVLD Xd.4D, Rj, si12
+        /// </summary>
+        public static unsafe Vector256<double> LoadVector256(double* address) => LoadVector256(address);
+
+        /// <summary>
+        /// float32x8_t xvfrecip_s_f32 (float32x8_t a)
+        ///   LASX: XVFRECIP.S Xd.8S Xj.8S
+        /// </summary>
+        public static Vector256<float> Reciprocal(Vector256<float> value) => Reciprocal(value);
```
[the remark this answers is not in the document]
shushanhf: The `Reciprocal` is implemented with the IEEE754-2008   division(1.0,x).

shushanhf:
Hi, @tannergooding 
I finished the first version of this PR, could you please review this PR?
Thanks

shushanhf:
> > could you please review this PR?
> 
> I can potentially give it a pass today or tomorrow, but its still blocked until API review can happen. That probably won't happen until the new year as API review typically doesn't happen in December when most people are on holiday/vacation.

OK, Thanks
I can wait more reviewers.

I will push other PRs that are independent of these APIs liking the SIMD's instructions within the emitter #95456

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/LoongArch64/LA_LASX.cs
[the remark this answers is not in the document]
shushanhf: Only the `FRECIPE` and `FRSQRTE` within the `LoongArchBase` class are extimate.
But the  `FRECIP` and `FRSQRT` are exact.

tannergooding:
I'm still waiting for response to the question asked on the API proposal:

> Where is the spec for the LA64 SIMD ISA?
>
> Based on https://loongson.github.io/LoongArch-Documentation/README-EN.html, it looks like it should be Volume 2, but there doesn't appear to be any version published for it yet and the backing GitHub repo looks to be archived now and I do not see a replacement.

tannergooding:
Thanks! This is still on my backlog but is lower priority than some other work due to the API review not having happened yet (and this PR being blocked until that can happen).

I'll try to set some time aside in the next week or two to go through the SIMD ISA guide and compare it to the proposed API surface so that it can get marked `ready-for-review`

shushanhf:
> Thanks! This is still on my backlog but is lower priority than some other work due to the API review not having happened yet (and this PR being blocked until that can happen).
> 
> I'll try to set some time aside in the next week or two to go through the SIMD ISA guide and compare it to the proposed API surface so that it can get marked `ready-for-review`

OK, Thanks very much.
I will do it after the Feb. 17 for the Chinese Spring Festival.

steveisok:
@tannergooding what is the status of this review?

tannergooding:
As per the above, new public API surface cannot be added without it first going through API Review.

Last checked, the LoongArch specs weren't available or were still in draft form, so going to API review wasn't possible.

If that's changed, then it'd be great to have links and references to the official docs so I can help drive this through to completion.

PR #94405, opened 2023-11-06
Allow preinitializing RuntimeType instances

MichalStrehovsky:
Unblock serializing `RuntimeType` instances from the static constructor interpreter.

With this we can now preinitialize `static readonly Type s_foo = typeof(Foo)` and RyuJIT will optimize reads of `s_foo` into loading a constant value.

Contributes to #91704.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/jit/importer.cpp
+++ b/src/coreclr/jit/importer.cpp
@@ -8989,6 +8989,7 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                         FALLTHROUGH;
                     case CORINFO_FIELD_STATIC_SHARED_STATIC_HELPER:
                     case CORINFO_FIELD_STATIC_ADDRESS:
+                    case CORINFO_FIELD_STATIC_RELOCATABLE:
                         // Replace static read-only fields with constant if possible
                         if ((aflags & CORINFO_ACCESS_GET) && (fieldInfo.fieldFlags & CORINFO_FLG_FIELD_FINAL))
                         {
@@ -9005,7 +9006,6 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                     case CORINFO_FIELD_STATIC_RVA_ADDRESS:
                     case CORINFO_FIELD_STATIC_GENERICS_STATIC_HELPER:
                     case CORINFO_FIELD_STATIC_READYTORUN_HELPER:
-                    case CORINFO_FIELD_STATIC_RELOCATABLE:
                         op1 = impImportStaticFieldAddress(&resolvedToken, (CORINFO_ACCESS_FLAGS)aflags, &fieldInfo,
                                                           lclTyp, &indirFlags);
                         break;
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
@@ -375,7 +375,7 @@ namespace ILCompiler
                                 if (value is ValueTypeValue)
                                     stack.PushFromLocation(field.FieldType, value);
                                 else if (value is ReferenceTypeValue referenceType)
-                                    stack.PushFromLocation(field.FieldType, referenceType.ToForeignInstance(baseInstructionCounter));
+                                    stack.PushFromLocation(field.FieldType, referenceType.ToForeignInstance(baseInstructionCounter, this));
                                 else
                                     return Status.Fail(methodIL.OwningMethod, opcode);
                             }
@@ -2388,7 +2388,7 @@ namespace ILCompiler
             }
         }
 
-        private sealed class RuntimeTypeValue : ReferenceTypeValue, IInternalModelingOnlyValue
+        private sealed class RuntimeTypeValue : ReferenceTypeValue
         {
             public TypeDesc TypeRepresented { get; }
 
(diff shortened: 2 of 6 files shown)


src/coreclr/jit/importer.cpp
```diff
@@ -8989,6 +8989,7 @@ void Compiler::impImportBlockCode(BasicBlock* block)
                         FALLTHROUGH;
                     case CORINFO_FIELD_STATIC_SHARED_STATIC_HELPER:
                     case CORINFO_FIELD_STATIC_ADDRESS:
+                    case CORINFO_FIELD_STATIC_RELOCATABLE:
```
MichalStrehovsky, line 8992: @EgorBo could you have a look at this? This affects a lot more than just RuntimeType and I might be missing why we weren't doing this before. (We might also want to do this for `STATIC_RVA_ADDRESS`.)

src/tests/nativeaot/SmokeTests/UnitTests/Interfaces.cs
```diff
@@ -689,6 +689,8 @@ class C3<T> : IInterface, IInterface<T>
         static IInterface<object> s_c3a = new C3<object>();
         static IInterface s_c3b = new C3<object>();
 
+        // Works around https://github.com/dotnet/runtime/pull/94342
+        [MethodImpl(MethodImplOptions.NoOptimization)]
```
MichalStrehovsky, line 693: This found a new bug. The `Run` method is essentially optimized to a throw. It's ridiculous how well this optimizes:

* We figured out the value assigned to the static is read only using whole program view (the static is not marked read only in source code)
* RyuJIT then figured out exact type information from the preinit data blob.
* RyuJIT then devirtualized the call (into the wrong method, but that's the bug)
* RyuJIT then folded the string constant comparison.

All that's left is the throw (and a couple garbage instructions I couldn't fully make sense of).

src/tests/nativeaot/SmokeTests/UnitTests/Interfaces.cs
```diff
@@ -689,6 +689,8 @@ class C3<T> : IInterface, IInterface<T>
         static IInterface<object> s_c3a = new C3<object>();
         static IInterface s_c3b = new C3<object>();
 
+        // Works around https://github.com/dotnet/runtime/pull/94342
```
MichalStrehovsky, line 692:
```suggestion
        // Works around https://github.com/dotnet/runtime/issues/94399
```

src/coreclr/jit/importer.cpp
EgorBo: @MichalStrehovsky wow, it's actually a good catch! It should improve some things, e.g.:
```cs
static readonly string Fld = "Hello";

static string Test() => Fld;
```
```diff
; Assembly listing for method Benchmarks:Test():System.String (FullOpts)
-      mov      rax, qword ptr [(reloc 0x400000000042c388)]
-      mov      rax, gword ptr [rax+0x08]
+      lea      rax, gword ptr [(reloc 0x400000000042c398)]      ; '"Hello"'
       ret      
; Total bytes of code 12
```

EgorBo: Basically, `static readonly` wasn't foldable for gc types for NAOT

MichalStrehovsky:
@dotnet/ilc-contrib could someone have a look?

jkotas:
System.Threading.Tasks tests are failing with a message that we have not seen before:
```
[FAIL] System.Threading.Tasks.Tests.TaskRunSyncTests.TaskRunSyncTest10
Task wasn't RunSynchronously with TaskScheduler specified
   at System.Threading.Tasks.Tests.TaskRunSyncTest.RealRun() + 0x507
```
Is it a pre-existing problem or a regression introduced by this change?

MichalPetryka:
> Is it a pre-existing problem or a regression introduced by this change?

I think I've seen this once on one of my PRs.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
```diff
@@ -2400,11 +2400,21 @@ public RuntimeTypeValue(TypeDesc type)
 
             public override bool GetRawData(NodeFactory factory, out object data)
             {
-                data = null;
-                return false;
+                data = factory.SerializedMaximallyConstructableRuntimeTypeObject(TypeRepresented);
```
jkotas, line 2403: We are calling SerializedMaximallyConstructableRuntimeTypeObject here without taking `_devirtualizationManager.CanConstructType` info into account. The other places where this is called takes `_devirtualizationManager.CanConstructType` into account.

Can this disparity cause problems?

MichalStrehovsky: It won't cause problems because we consistently ask for the same thing between scanning and codegen in the preinit interpreter.

We only go through the CanConstructType path when we need to downgrade `Type` used in comparisons. Whether downgrade is allowed is detected in the IL scanner when it sees the `==` pattern - scanner will report such Types as unconstructed. Then codegen will get the right form when it asks for a frozen version.

We'd need a lot more infrastructure to be able to do this downgrade for `Type` that was roundtripped through a static field (which is the path that's relevant here). When we're running the cctor interpreter, we don't know how the preinitialized static fields are going to be used: we'll only know if it's okay to downgrade once we know all the places that read from the field (they have to be in the form of the `==` pattern), including reflection access to the field. It's doable, but doesn't feel worth the trouble. Asking for the constructed version is the safe but not optimal option.

MichalStrehovsky:
> > Is it a pre-existing problem or a regression introduced by this change?
> 
> I think I've seen this once on one of my PRs.

This test sometimes fails, always with a failure mode that doesn't produce dumps. It never reproed for me locally.

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/TypePreinit.cs
jkotas: Ok, thanks for the explanation.

jkotas (approved):
Thank you!

Issue #94230, opened 2023-10-31
[Android] WebRequest causes SSL exception on Android API <= 23 on .NET 8


Redth:
### Description

There's SSL failures on .NET Android app code when using `WebRequest.Create` instead of `HttpClient` API's in .NET 8.0 when running on API levels 23 or older.  This previously worked in .NET 7.

### Reproduction Steps

Working:

```
using var client = new HttpClient();
var content = await client.GetStringAsync("https://webcode.me");
Console.WriteLine(content);
```

Failing:

```
WebRequest webRequest = WebRequest.Create("https://webcode.me");
webRequest.Method = "GET";
using (var webResponse = webRequest.GetResponse() as HttpWebResponse)
{
    using (var inStream = webResponse.GetResponseStream())
    {
        using (var reader = new StreamReader(inStream))
        {
            var responseString = reader.ReadToEnd();
            Console.WriteLine(responseString);
        }
    }
}
```

Run this code in a net8.0-android app on API Level 23 or older (so Android 6.0 or older).
You can validate it works by running on API Level 24 or newer (Android 7.0 or newer).

### Expected behavior

No exceptions

### Actual behavior

Exception (see stack trace below).

### Regression?

Yes, this appears to be a regression from .NET 7

### Other information

Stack trace for failure:

```
[DOTNET] System.Net.WebException: The SSL connection could not be established, see inner exception.
[DOTNET]  ---> System.Net.Http.HttpRequestException: The SSL connection could not be established, see inner exception.
[DOTNET]  ---> System.Security.Authentication.AuthenticationException: Authentication failed, see inner exception.
[DOTNET]  ---> Interop+AndroidCrypto+SslException: Exception of type 'Interop+AndroidCrypto+SslException' was thrown.
[DOTNET]    --- End of inner exception stack trace ---
[DOTNET]    at System.Net.Security.SslStream.<ForceAuthenticationAsync>d__150`1[[System.Net.Security.AsyncReadWriteAdapter, System.Net.Security, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a]].MoveNext()
[DOTNET]    at System.Net.Http.ConnectHelper.EstablishSslConnectionAsync(SslClientAuthenticationOptions sslOptions, HttpRequestMessage request, Boolean async, Stream stream, CancellationToken cancellationToken)
[DOTNET]    --- End of inner exception stack trace ---
[...]

PR #94408, opened 2023-11-06
[Android] Fix SslStream on Android API 21-23

simonrozsival:
The `SSLEngine.getHandshakeSession` java method isn't available on Android API 21-23 (we don't support older API levels). As a workaround, we can access the private field that contains the handshake session directly. Since this change only affects very old Android versions which won't receive any updates (not even security updates), I think accessing the private field is OK.

This PR will have to be backported to .NET 8 once it's merged. These changes should be reverted once we drop support for Android API 23.

Ref #94230


wfurt (approved):
LGTM. I feel that using privates for legacy is ok -> it is unlikely to change.

PR #94409, opened 2023-11-06
JIT: Remove dead GT_NOP code in LSRA

### LINKED ISSUE #93048: DEREF_OF_NULL in /src/coreclr/jit/lsrabuild.cpp
Pointer `GenTree::gtGetOp1(...)`, that can have __only NULL value__, is passed as 1st parameter in call to function `LinearScan::BuildUse`

https://github.com/dotnet/runtime/blob/e40b3abf1b41621d4298642a5fd300ebf7cccf6d/src/coreclr/jit/lsraxarch.cpp#L211-L216

where it __is dereferenced__ at

https://github.com/dotnet/runtime/blob/e40b3abf1b41621d4298642a5fd300ebf7cccf6d/src/coreclr/jit/lsrabuild.cpp#L2929-L2930

https://github.com/dotnet/runtime/blob/e40b3abf1b41621d4298642a5fd300ebf7cccf6d/src/coreclr/jit/lsrabuild.cpp#L2964-L2967

Found by Linux Verification Center ([portal.linuxtesting.ru](https://portal.linuxtesting.ru/)) with SVACE.

Author E. Bykhanova.

jakobbotsch:
Rationalization removes all GT_NOP nodes, and we do not seem to introduce any of these before LSRA (and especially not the typed "passthrough" GT_NOP nodes). This code is just wrong for the passthrough nodes, so just switch all the LSRA code to only handle potential void NOPs.

Fix #93048

### DIFF
--- a/src/coreclr/jit/lsraarm.cpp
+++ b/src/coreclr/jit/lsraarm.cpp
@@ -258,21 +258,9 @@ int LinearScan::BuildNode(GenTree* tree)
             break;
 
         case GT_NOP:
-            // A GT_NOP is either a passthrough (if it is void, or if it has
-            // a child), but must be considered to produce a dummy value if it
-            // has a type but no child
             srcCount = 0;
-            assert((tree->gtGetOp1() == nullptr) || tree->isContained());
-            if (tree->TypeGet() != TYP_VOID && tree->gtGetOp1() == nullptr)
-            {
-                assert(dstCount == 1);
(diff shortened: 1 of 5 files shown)


jakobbotsch:
cc @dotnet/jit-contrib PTAL @kunalspathak

kunalspathak (approved):
LGTM

PR #94412, opened 2023-11-06
Add option for R2R configured runs.

cincuranet:
Depends on https://github.com/dotnet/performance/pull/3467.
Supports https://github.com/dotnet/performance/issues/3353.


LoopedBard3 (approved):
Are there plans to also make updates to perf-job.yml and run-performance-job.yml so that we don't have naming collisions for these runs? I don't have strong feelings for it being done here vs another commit regardless.

cincuranet:
PR updated. Better to do it in one commit.

LoopedBard3 (approved):
Looks good to me.

PR #94413, opened 2023-11-06
JIT: Fix `EvalHWIntrinsicFunBinary` for ARM `MultiplyByScalar`

### LINKED ISSUE #93876: Assertion failed 'm_store->TypeGet() == m_src->TypeGet()' during 'Stress gtSplitTree'
```c#
// Found by Antigen

using System;
using System.Numerics;
public class TestClass
{
    public struct S1
    {
    }
    public struct S2
    {
    }
    static float s_float_12 = 0f;
    static S1 s_s1_40 = new S1();
    static S2 s_s2_41 = new S2();
    long long_49 = 0;
    Vector4 v4_78 = new Vector4(5.090909f, -2.1474836E+09f, 5.148148f, 0.054054055f);
    S1 s1_79 = new S1();
    S2 s2_80 = new S2();
    public S1 Method45(ref long p_long_1227, S1 p_s1_1228, out S2 p_s2_1229, ref S1 p_s1_1230, ref S2 p_s2_1231, S1 p_s1_1232, out Vector4 p_v4_1233, ref S2 p_s2_1234, ref S1 p_s1_1235)
    {
        unchecked
        {
            p_s2_1229 = s_s2_41;
            p_v4_1233 = Vector4.Multiply(s_float_12, Vector4.One);
            return p_s1_1235;
        }
    }
    public void Method0()
    {
        unchecked
        {
            S1 s1_2605 = new S1();
            S2 s2_2606 = new S2();
            s1_2605 = Method45(ref long_49, s_s1_40, out s2_80, ref s1_79, ref s2_2606, s1_2605, out v4_78, ref s_s2_41, ref s_s1_40);
            return;
        }
    }
    {
        new TestClass().Method0();
    }
}
/*
Got output diff:
--------- Baseline ---------  
Environment:
--------- Test ---------  
Environment:
set DOTNET_JitStress=1
set DOTNET_AltJitName=clrjit_universal_arm64_x64.dll
set DOTNET_AltJit=Method45
Assert failure(PID 42792 [0x0000a728], Thread: 1268 [0x04f4]): Assertion failed 'm_store->TypeGet() == m_src->TypeGet()' in 'TestClass:Method45(byref,TestClass+S1,byref,byref,byref,TestClass+S1,byref,byref,byref):TestClass+S1:this' during 'Stress gtSplitTree' (IL size 41; hash 0x5f755485; FullOpts)
    File: E:\git\runtime2\src\coreclr\jit\morphblock.cpp Line: 671
    Image: e:\git\runtime2\artifacts\tests\coreclr\windows.x64.Checked\tests\Core_Root\CoreRun.exe
*/
```

jakobbotsch:
`MultiplyByScalar` on ARM/ARM64 can have differently typed operands, so we cannot fold multiplications by one/zero without ensuring that we get the proper result type.

Fix #93876

### DIFF
--- a/src/coreclr/jit/valuenum.cpp
+++ b/src/coreclr/jit/valuenum.cpp
@@ -7826,7 +7826,7 @@ ValueNum ValueNumStore::EvalHWIntrinsicFunBinary(var_types      type,
                 }
                 break;
             }
-#else
+#endif
 
 #ifdef TARGET_ARM64
             case NI_AdvSimd_Multiply:
--- a/src/coreclr/jit/valuenum.cpp
+++ b/src/coreclr/jit/valuenum.cpp
@@ -7810,10 +7810,13 @@ ValueNum ValueNumStore::EvalHWIntrinsicFunBinary(var_types      type,
 
                 assert((TypeOfVN(arg0VN) == type) && (TypeOfVN(arg1VN) == TYP_SIMD8));
 
-                // Handle x * 1 => x, but only if the scalar RHS is 1.
-                if (arg1VN == VNOneForSimdType(TYP_SIMD8, baseType))
+                // Handle x * 1 => x, but only if the scalar RHS is <1, ...>.
+                if (IsVNConstant(arg1VN))
                 {
-                    return arg0VN;
+                    if (EvaluateSimdGetElement(this, TYP_SIMD8, baseType, arg1VN, 0) == VNOneForType(baseType))
+                    {
+                        return arg0VN;
+                    }
                 }
                 break;
             }
--- a/src/coreclr/jit/valuenum.cpp
+++ b/src/coreclr/jit/valuenum.cpp
@@ -7793,10 +7793,44 @@ ValueNum ValueNumStore::EvalHWIntrinsicFunBinary(var_types      type,
             }
 
 #ifdef TARGET_ARM64
-            case NI_AdvSimd_Multiply:
             case NI_AdvSimd_MultiplyByScalar:
-            case NI_AdvSimd_Arm64_Multiply:
             case NI_AdvSimd_Arm64_MultiplyByScalar:
+            {
+                if (!varTypeIsFloating(baseType))
+                {
+                    // Handle `x * 0 == 0` and `0 * x == 0`
+                    // Not safe for floating-point when x == -0.0, NaN, +Inf, -Inf
+                    ValueNum zeroVN = VNZeroForType(TypeOfVN(cnsVN));
+
+                    if (cnsVN == zeroVN)
+                    {
+                        return VNZeroForType(type);
+                    }
+                }
+
+                // Handle x * 1 => x, but only if the scalar RHS is 1.
+                ValueNum oneVN;
+                if (varTypeIsSIMD(TypeOfVN(arg1VN)))
+                {
+                    oneVN = VNOneForSimdType(TypeOfVN(arg1VN), baseType);
+                }
+                else
+                {
+                    oneVN = VNOneForType(baseType);
+                }
+
+                if (arg1VN == oneVN)
+                {
+                    assert(TypeOfVN(arg0VN) == type);
+                    return arg0VN;
+                }
+                break;
+            }
+#else
+
+#ifdef TARGET_ARM64
+            case NI_AdvSimd_Multiply:
+            case NI_AdvSimd_Arm64_Multiply:
 #else
             case NI_SSE_Multiply:
             case NI_SSE2_Multiply:
@@ -7819,7 +7853,7 @@ ValueNum ValueNumStore::EvalHWIntrinsicFunBinary(var_types      type,
 
                     if (cnsVN == zeroVN)
                     {
-                        return VNZeroForType(type);
+                        return zeroVN;
                     }
                 }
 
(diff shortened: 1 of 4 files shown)


src/tests/JIT/Regression/JitBlue/Runtime_93876/Runtime_93876.csproj
```diff
@@ -0,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <Optimize>True</Optimize>
+  </PropertyGroup>
+  <ItemGroup>
+    <CLRTestEnvironmentVariable Include="DOTNET_JitStressModeNames" Value="STRESS_MAKE_CSE STRESS_SPLIT_TREES_REMOVE_COMMAS" />
```
SingleAccretion, line 6: In other tests there are comments that you need process isolation to use `CLRTestEnvironmentVariable`, is that no longer true?

src/tests/JIT/Regression/JitBlue/Runtime_93876/Runtime_93876.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    [Fact]
+    public static void Problem()
+    {
+        Vector4 v = Mul(0, 1);
+        Assert.Equal(Vector4.One, v);
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static Vector4 Mul(float a, float b) => Vector4.Multiply(a + b, Vector4.One);
```
tannergooding, line 18: What is the type of `a + b` here that causes the issue?

tannergooding (approved):

src/tests/JIT/Regression/JitBlue/Runtime_93876/Runtime_93876.cs
jakobbotsch: Hmm, I think `MultiplyByScalar` shouldn't be handled by this path. E.g. `Vector64.Multiply(Vector64<float>.One, x)` will produce `<x, 0>` instead of `<x, x>` because of that. So this needs a bit more work.

jakobbotsch: I pushed a change that handles `MultiplyByScalar` separately, can you take another look?

src/coreclr/jit/valuenum.cpp
```diff
@@ -7793,10 +7793,44 @@ ValueNum ValueNumStore::EvalHWIntrinsicFunBinary(var_types      type,
...
+                // Handle x * 1 => x, but only if the scalar RHS is 1.
+                ValueNum oneVN;
+                if (varTypeIsSIMD(TypeOfVN(arg1VN)))
+                {
+                    oneVN = VNOneForSimdType(TypeOfVN(arg1VN), baseType);
+                }
+                else
+                {
+                    oneVN = VNOneForType(baseType);
+                }
```
tannergooding, line 7820: Can this actually be hit given that we expect `TYP_SIMD8` for the second operand?

jakobbotsch: Good question... let me add some strong asserts and simplify this a bit.

src/coreclr/jit/valuenum.cpp
```diff
@@ -7793,10 +7793,44 @@ ValueNum ValueNumStore::EvalHWIntrinsicFunBinary(var_types      type,
...
+
+                if (arg1VN == oneVN)
+                {
+                    assert(TypeOfVN(arg0VN) == type);
+                    return arg0VN;
+                }
```
tannergooding: This looks correct, but may be slightly pessimizing for `MultiplyByScalar`.

While we try to self-import/normalize `vector * scalar` as `MultiplyByScalar(vector, Vector64.CreateScalarUnsafe(scalar))` and `CreateScalarUnsafe` will itself try to create a broadcast for a constant `scalar` (for efficiency reasons), a user could still call `MultiplyByScalar` themselves among other factors.

So I think the more correct thing to do here is to expect that `arg1VN` is a vector (check and exit if it isn't), then to extract element 0 and check if that is equal to `VNOneForType(baseType)`. Since `Vector64.Create(1, 0)` will also do the same thing here.

--------------

Alternatively, maybe the better thing would be to make `MultiplyByScalar` a lowering optimization. We could import it as `Multiply(vector, Vector128.Create(scalar))` and have lowering look for this pattern and swap it around. That would allow more CSE opportunities and less complex handling here.

tannergooding: -- I don't have a particular preference and aren't even particularly concerned about the slight pessimization.

Just wanted to call it out as something to consider since this code is being touched.

jakobbotsch: Does it look better now?

src/coreclr/jit/valuenum.cpp
tannergooding: I think the same path down in the regular multiply handling is also "dead". It was likely intended to handle the `vector * scalar` case, but we don't actually ever produce a `GT_HWINTRINSIC(TYP_SIMD, TYP_SCALAR)` node anymore, afaict.

We appear to always normalize to `Broadcast` on x86/x64. We similarly normalize to `Broadcast` on `Arm64` except for when a `MultiplyByScalar` variant exists, in which case we always insert `CreateScalarUnsafe`.

I then don't see any overloads of the instructions that would take a proper scalar register, so I don't think we'd ever hit it in practice.

tannergooding (approved):

PR #93126, opened 2023-10-06
Replace some stackallocs with collection expressions

stephentoub:
These will result in using an InlineArray rather than a localloc

PR #94414, opened 2023-11-06
Console.Unix: for echoing, write back to the terminal instead of writing to Console.Out.

### LINKED ISSUE #22314: console redirection is broken
_From @Spongman on June 15, 2017 21:5_

````
using System;
class Test {
  public static void Main () {
    while(true)
      Console.WriteLine(Console.ReadLine());
  }
}
````

stdout redirection:
=====
test command: `dotnet run > test.txt`
expected behavior is to show typed text in the console/tty, and to write the typed lines out to `test.txt`

on windows: works fine.
on linux: writes stdin to the file ok, but the typed text is not echoed to tty.

stdin redirection:
=====
test command: `dotnet run < test.txt` (test.txt exists and contains some lines of text)
expected behavior is to write the contents of `test.txt` to the console/tty.

on both windows & linux: console clears and the CPU is pegged!!

version: 1.0.4
````
Microsoft .NET Core Shared Framework Host

  Version  : 1.1.0
  Build    : 928f77c4bc3f49d892459992fb6e1d5542cb5e86
````

_Copied from original issue: dotnet/coreclr#12308_

### PR DESCRIPTION
**Pull request by tmds**
On Unix .NET implements its own echo to make Console behave similar to Windows.

The echo implementation was writing to standard output, which has the unintended effect of writing the echo characters into a redirected standard output.

This changes to write the echo to the stdin terminal where they were read from.

Fixes https://github.com/dotnet/runtime/issues/22314.

@adamsitnik @stephentoub ptal.

### DIFF
--- a/src/libraries/System.Console/src/System/ConsolePal.Unix.cs
+++ b/src/libraries/System.Console/src/System/ConsolePal.Unix.cs
@@ -243,7 +243,7 @@ namespace System
                 if (!string.IsNullOrEmpty(cursorAddressFormat))
                 {
                     string ansiStr = TermInfo.ParameterizedStrings.Evaluate(cursorAddressFormat, top, left);
-                    WriteStringToTerminal(terminalHandle, ansiStr);
+                    WriteTerminalAnsiString(terminalHandle, ansiStr);
                 }
 
                 SetCachedCursorPosition(left, top);
@@ -1094,7 +1094,7 @@ namespace System
             Volatile.Write(ref s_invalidateCachedSettings, 1);
         }
 
-        internal static void WriteStringToTerminal(SafeFileHandle terminalHandle, string? value, bool mayChangeCursorPosition = true)
+        internal static void WriteTerminalAnsiString(SafeFileHandle terminalHandle, string? value, bool mayChangeCursorPosition = true)
         {
             if (string.IsNullOrEmpty(value))
                 return;
@@ -1120,7 +1120,7 @@ namespace System
         /// <summary>Writes a terminfo-based ANSI escape string to stdout.</summary>
         /// <param name="value">The string to write.</param>
         /// <param name="mayChangeCursorPosition">Writing this value may change the cursor position.</param>
-       internal static void WriteStdoutAnsiString(string? value, bool mayChangeCursorPosition = true)
-            => WriteStringToTerminal(Interop.Sys.FileDescriptors.STDOUT_FILENO, value, mayChangeCursorPosition);
+       private static void WriteStdoutAnsiString(string? value, bool mayChangeCursorPosition = true)
+            => WriteTerminalAnsiString(Interop.Sys.FileDescriptors.STDOUT_FILENO, value, mayChangeCursorPosition);
     }
 }
--- a/src/libraries/System.Console/src/System/ConsolePal.Unix.cs
+++ b/src/libraries/System.Console/src/System/ConsolePal.Unix.cs
@@ -37,6 +37,7 @@ namespace System
         private static int s_windowWidth;   // Cached WindowWidth, -1 when invalid.
         private static int s_windowHeight;  // Cached WindowHeight, invalid when s_windowWidth == -1.
         private static int s_invalidateCachedSettings = 1; // Tracks whether we should invalidate the cached settings.
+        private static SafeFileHandle? s_terminalHandle; // Tracks the handle used for writing to the terminal.
 
         /// <summary>Gets the lazily-initialized terminal information for the terminal.</summary>
         public static TerminalFormatStrings TerminalFormatStringsInstance { get { return s_terminalFormatStringsInstance.Value; } }
@@ -199,7 +200,7 @@ namespace System
                 if (!string.IsNullOrEmpty(titleFormat))
                 {
                     string ansiStr = TermInfo.ParameterizedStrings.Evaluate(titleFormat, value);
-                    WriteStdoutAnsiString(ansiStr, mayChangeCursorPosition: false);
+                    WriteTerminalAnsiString(ansiStr, mayChangeCursorPosition: false);
                 }
             }
         }
@@ -208,7 +209,7 @@ namespace System
         {
             if (!Console.IsOutputRedirected)
             {
-                WriteStdoutAnsiString(TerminalFormatStringsInstance.Bell, mayChangeCursorPosition: false);
+                WriteTerminalAnsiString(TerminalFormatStringsInstance.Bell, mayChangeCursorPosition: false);
             }
         }
 
@@ -216,7 +217,7 @@ namespace System
         {
             if (!Console.IsOutputRedirected)
             {
-                WriteStdoutAnsiString(TerminalFormatStringsInstance.Clear);
+                WriteTerminalAnsiString(TerminalFormatStringsInstance.Clear);
             }
         }
 
@@ -225,10 +226,10 @@ namespace System
             if (Console.IsOutputRedirected)
                 return;
 
-            SetCursorPosition(Interop.Sys.FileDescriptors.STDOUT_FILENO, left, top);
+            SetTerminalCursorPosition(left, top);
         }
 
-        internal static void SetCursorPosition(SafeFileHandle terminalHandle, int left, int top)
+        public static void SetTerminalCursorPosition(int left, int top)
         {
             lock (Console.Out)
             {
@@ -243,7 +244,7 @@ namespace System
                 if (!string.IsNullOrEmpty(cursorAddressFormat))
                 {
                     string ansiStr = TermInfo.ParameterizedStrings.Evaluate(cursorAddressFormat, top, left);
-                    WriteTerminalAnsiString(terminalHandle, ansiStr);
+                    WriteTerminalAnsiString(ansiStr);
                 }
 
                 SetCachedCursorPosition(left, top);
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by tmds** [comment]
I need to make some additional changes so the characters that are written back for echo use the proper encoding.

**Comment by tmds** [comment]
> I need to make some additional changes so the characters that are written back for echo use the proper encoding.

This is done. The PR is up for review.

File: src/libraries/System.Console/src/System/IO/StdInReader.cs [hunk 1]
```diff
@@ -363,5 +375,21 @@ private unsafe ConsoleKeyInfo ReadKey()
 
         /// <summary>Gets whether there's input waiting on stdin.</summary>
         internal static bool StdinReady => Interop.Sys.StdinReady();
+
+        private void Echo(char c)
+        {
+            ReadOnlySpan<char> chars = [ c ];
+            Span<byte> bytes = stackalloc byte[32]; // 32 bytes seems ample
+            int bytesWritten = _echoEncoder.GetBytes(chars, bytes, flush: false);
```
adamsitnik, line 383: ASCII should be very common, we should add an optimized code path for it. Sth like this:

```cs
Span<byte> bytes;
if (Ascii.IsValid(c))
{
    byte b = (byte)c;
    bytes = new Span<byte>(in b);
}
else
{
    bytes = stackalloc byte[32]; // 32 bytes seems ample
    int bytesWritten = _echoEncoder.GetBytes(chars, bytes, flush: false);
    if (bytesWritten == 0)
    {
        return;
    }
    bytes = bytes.Slice(0, bytesWritten);
}

lock (Console.Out) // synchronize with other writers
{
    ConsolePal.Write(StdInHandle, bytes);
}
```

adamsitnik (commented):
The fix is great! It's relatively simple for the bug that has been hunting us for years! 👍 

I added some comments for improving perf. It would be great to address them before merging.

Thank you @tmds !

src/libraries/System.Console/src/System/ConsolePal.Unix.cs
```diff
@@ -231,6 +225,11 @@ public static void SetCursorPosition(int left, int top)
             if (Console.IsOutputRedirected)
                 return;
 
+            SetCursorPosition(Interop.Sys.FileDescriptors.STDOUT_FILENO, left, top);
+        }
+
+        internal static void SetCursorPosition(SafeFileHandle terminalHandle, int left, int top)
```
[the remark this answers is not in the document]
tmds: I ran the manual tests and I saw no regression.

I will try some additional scenarios this week.

src/libraries/System.Console/src/System/IO/StdInReader.cs
tmds: > if (Ascii.IsValid(c)) byte b = (byte)c;

I am not an encoding expert. Is this valid for any encoding?

adamsitnik: > I am not an encoding expert. Is this valid for any encoding?

@GrabYourPitchforks @tarekgh Could you please confirm/deny?

src/libraries/System.Console/src/System/IO/StdInReader.cs
```diff
@@ -363,5 +375,21 @@ private unsafe ConsoleKeyInfo ReadKey()
 
         /// <summary>Gets whether there's input waiting on stdin.</summary>
         internal static bool StdinReady => Interop.Sys.StdinReady();
+
+        private void Echo(char c)
+        {
+            ReadOnlySpan<char> chars = [ c ];
```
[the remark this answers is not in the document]
gfoidl: Yes the Roslyn compiler in a recent enough version is smart enough.

It "probes" for runtime-type support and if supported emits a ROS to the assembly's static data segment.
On older runtimes / compilers the array-allocation will happen.

Sharplab is too old for this -- I got tricked into the same in https://github.com/dotnet/runtime/pull/93126#pullrequestreview-1678048414

src/libraries/System.Console/src/System/IO/StdInReader.cs
gfoidl: > if (Ascii.IsValid(c)) byte b = (byte)c;

That's valid.

`Ascii.IsValid` returns `true` if `c <= 127` only, thus the byte-cast is valid and doesn't truncate anything, etc.

> for any encoding?

Side note: If there's any encoding that differs from ASCII in the 0-7th LSB then .NET would have a bigger problem with lots of broken code.

src/libraries/System.Console/src/System/IO/StdInReader.cs
```diff
@@ -363,5 +375,21 @@ private unsafe ConsoleKeyInfo ReadKey()
 
         /// <summary>Gets whether there's input waiting on stdin.</summary>
         internal static bool StdinReady => Interop.Sys.StdinReady();
+
+        private void Echo(char c)
```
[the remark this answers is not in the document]
gfoidl: `EchoToTerminal`?

That implies the echoing-nature and "to terminal".

src/libraries/System.Console/src/System/IO/StdInReader.cs
tarekgh: Be careful with that. There are some double-bytes encodings which have leading and trailing bytes. Some encodings have trailing bytes in the ASCII range. Like:

https://github.com/dotnet/runtime/blob/6ffa54863c810d0a0fe588a107cb44415709355f/src/libraries/System.Text.Encoding.CodePages/src/System/Text/GB18030Encoding.cs#L23

Another cases, some encodings use escape sequence in the encoding which can have ASCII characters like:

https://github.com/dotnet/runtime/blob/6ffa54863c810d0a0fe588a107cb44415709355f/src/libraries/System.Text.Encoding.CodePages/src/System/Text/ISO2022Encoding.cs#L326

gfoidl: Thanks for these insights!

But `if (Ascii.IsValid(c)) byte b = (byte)c;` is valid when `c` is a `char` (UTF-16)?

tarekgh: Yes, if you are doing encoding, then the code is fine. I was just pointing out to be careful if there is any code later is decoding bytes and using similar optimization.

src/libraries/System.Console/src/System/ConsolePal.Unix.cs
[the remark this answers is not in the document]
tmds: I changed the PR so it prefers using stdout as a terminal handle to remain closer to the current behavior.
I also made the cursor/window size getting to work against the stdin handle.

src/libraries/System.Console/src/System/IO/StdInReader.cs
```diff
@@ -363,5 +373,26 @@ private unsafe ConsoleKeyInfo ReadKey()
...
+            }
+            else
+            {
+                var chars = new ReadOnlySpan<char>(in c);
+                bytesWritten = _echoEncoder.GetBytes(chars, bytes, flush: false);
+                if (bytesWritten == 0)
+                {
+                    return;
+                }
+            }
```
tmds, line 393: @adamsitnik does this look good to you? I wrote it this way to have the `bytes` `Span` is in the same scope as the `WriteToTerminal` call.

tmds:
I want to highlight a specific special case which will now throw. When standard output is redirected, and standard input is a read-only terminal handle, `ReadLine` will now throw.

```
Unhandled exception. System.UnauthorizedAccessException: Access to the path is denied.
 ---> System.IO.IOException: Bad file descriptor
   --- End of inner exception stack trace ---
   at System.ConsolePal.Write(SafeFileHandle fd, ReadOnlySpan`1 buffer, Boolean mayChangeCursorPosition) in /home/tmds/repos/runtime/src/libraries/System.Console/src/System/ConsolePal.Unix.cs:line 1006
   at System.ConsolePal.WriteToTerminal(ReadOnlySpan`1 buffer, Boolean mayChangeCursorPosition) in /home/tmds/repos/runtime/src/libraries/System.Console/src/System/ConsolePal.Unix.cs:line 957
   at System.IO.StdInReader.EchoToTerminal(Char c) in /home/tmds/repos/runtime/src/libraries/System.Console/src/System/IO/StdInReader.cs:line 395
   at System.IO.StdInReader.ReadLineCore(Boolean consumeKeys) in /home/tmds/repos/runtime/src/libraries/System.Console/src/System/IO/StdInReader.cs:line 181
   at System.IO.StdInReader.ReadLine() in /home/tmds/repos/runtime/src/libraries/System.Console/src/System/IO/StdInReader.cs:line 95
   at System.IO.SyncTextReader.ReadLine() in /home/tmds/repos/runtime/src/libraries/System.Console/src/System/IO/SyncTextReader.cs:line 77
   at System.Console.ReadLine() in /home/tmds/repos/runtime/src/libraries/System.Console/src/System/Console.cs:line 752
   at Program.<Main>$(String[] args) in /tmp/console/Program.cs:line 7
```

This seems appropriate as we don't have a way to echo the characters to.
Previously this didn't throw as the characters got echoed to the standard output.
If desired, we could make this work by no longer throwing in this specific case which would result in the user not seeing an echo.

src/libraries/System.Console/src/System/ConsolePal.Wasi.cs
```diff
@@ -254,17 +250,21 @@ private static unsafe int Read(SafeFileHandle fd, Span<byte> buffer)
             }
         }
 
+        internal static unsafe void WriteFromConsoleStream(SafeFileHandle fd, ReadOnlySpan<byte> buffer)
```
tmds, line 253: These other changes sync with the `.Unix` file.

src/libraries/System.Console/src/System/ConsolePal.Wasi.cs
```diff
@@ -18,24 +18,20 @@ namespace System
...
     internal static partial class ConsolePal
     {
-        // StdInReader is only used when input isn't redirected and we're working
-        // with an interactive terminal.  In that case, performance isn't critical
-        // and we can use a smaller buffer to minimize working set.
-        // there is no dup on WASI
         public static Stream OpenStandardInput()
         {
-            return new UnixConsoleStream(Interop.CheckIo(Interop.Sys.FileDescriptors.STDIN_FILENO), FileAccess.Read,
+            return new UnixConsoleStream(Interop.Sys.FileDescriptors.STDIN_FILENO, FileAccess.Read,
```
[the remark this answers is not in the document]
adamsitnik: I am not familiar with this particular overload, let me share it here think loud about that:

https://github.com/dotnet/runtime/blob/487d7f010b90a82178d55eea21e72dc6dedf9d5d/src/libraries/Common/src/Interop/Unix/Interop.IOErrors.cs#L86-L105

1. The xml comment does not describe what it does.
2. If provided handle is invalid, then it gets last error for the current thread and throws an exception. 

I don't think that we need such check here, as we have not performed any IO yet (the handle is created by providing a constant number: 0, 1 or 2, it's not a result of a sys-call), so there is no last error that makes sense. It can become invalid once we try to use it in a sys-call, but other layers are ready for that.

So I am fine with removing these checks (and the method itself if it's not used elsewhere) 👍

adamsitnik (approved):
LGTM, thank you very much @tmds!

adamsitnik:
> I want to highlight a specific special case which will now throw. 

I am OK with that. I've applied `needs-breaking-change-doc-created` to ensure that we document that.

PR #94417, opened 2023-11-06
Expose ChannelReader.ReadAllAsync from non-core builds as well

stephentoub:
I'm not sure why we didn't fix this before when we shipped Microsoft.Bcl.AsyncInterfaces, but with IAsyncEnumerable available downlevel, there's no need to hide this method away; it can be in all builds.  Doing so makes it easier for others to create their own channel implementations, as they don't _need_ to multitarget in order to override everything they might want to.

I've not changed any C# code, just moved it between files.


src/libraries/System.Threading.Channels/ref/System.Threading.Channels.csproj
```diff
@@ -14,5 +14,6 @@
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="$(SystemThreadingTasksExtensionsVersion)" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="$(MicrosoftBclAsyncInterfacesVersion)" />
```
ViktorHofer, line 17: There's no need to use a prebuilt here. Please ProjectReference the live project instead. Same in the other projects.

src/libraries/System.Threading.Channels/tests/System.Threading.Channels.Tests.csproj
```diff
@@ -6,19 +6,17 @@
   <ItemGroup>
     <Compile Include="BoundedChannelTests.cs" />
     <Compile Include="ChannelClosedExceptionTests.cs" />
+    <Compile Include="ChannelClosedExceptionTests.netcoreapp.cs"
+             Condition="'$(TargetFramework)' == '$(NetCoreAppCurrent)'" />
```
ViktorHofer, line 10: tiny nit: This is fine today as we only build our test assemblies against the latest .NET version but we usually avoid conditioning on a specific TFM unless it makes sense in that context.

```suggestion
             Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'" />
```

stephentoub:
Thanks, @ViktorHofer. Fixed.

ViktorHofer (approved):

PR #94433, opened 2023-11-06
Make Deflate/GZip/ZlibStream empty writes have the same impact as no writes

stephentoub:
If you create such a compression stream, don't write to it, and close it, it won't write out any bytes.  But today if you perform an empty write, it'll result in the stream thinking data was written, and it'll end up writing out an empty chunk, which will result in several bytes of output.  Logically it's strange that there's such a difference between an empty write and no write; this seeks to make them the same.

This changed back in 2015, in https://github.com/dotnet/corefx/pull/4917, I _think_ unintentionally.

@ericstj, this removes a test you added in https://github.com/dotnet/corefx/pull/41190 (the test relies on an empty write resulting in a non-empty compressed payload).  Do you have concerns with this change? Is there reason to believe folks depend on this?  In general with compression we make no guarantees version to version about generated bytes other than that they'll roundtrip, but this change obviously breaks that test you added, which will now loop forever because it assumes the number of bytes written will not be 0.

### DIFF
--- a/src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateZLib/DeflateStream.cs
+++ b/src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateZLib/DeflateStream.cs
@@ -507,8 +507,13 @@ namespace System.IO.Compression
             EnsureCompressionMode();
             EnsureNotDisposed();
 
-            Debug.Assert(_deflater != null);
+            if (buffer.IsEmpty)
+            {
+                return;
+            }
+
             // Write compressed the bytes we already passed to the deflater:
+            Debug.Assert(_deflater != null);
             WriteDeflaterOutput();
 
             unsafe
@@ -793,8 +798,9 @@ namespace System.IO.Compression
             EnsureNoActiveAsyncOperation();
             EnsureNotDisposed();
 
-            return cancellationToken.IsCancellationRequested ?
-                ValueTask.FromCanceled(cancellationToken) :
+            return
+                cancellationToken.IsCancellationRequested ? ValueTask.FromCanceled(cancellationToken) :
+                buffer.IsEmpty ? default :
                 Core(buffer, cancellationToken);
 
             async ValueTask Core(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken)
@@ -804,8 +810,8 @@ namespace System.IO.Compression
                 {
                     await WriteDeflaterOutputAsync(cancellationToken).ConfigureAwait(false);
 
-                    Debug.Assert(_deflater != null);
                     // Pass new bytes through deflater
+                    Debug.Assert(_deflater != null);
                     _deflater.SetInput(buffer);
 
                     await WriteDeflaterOutputAsync(cancellationToken).ConfigureAwait(false);
--- a/src/libraries/System.IO.Compression/tests/CompressionStreamUnitTests.Gzip.cs
+++ b/src/libraries/System.IO.Compression/tests/CompressionStreamUnitTests.Gzip.cs
@@ -55,6 +55,65 @@ namespace System.IO.Compression
         {
         }
 
+        [Fact]
+        public async Task ConcatenatedEmptyGzipStreams()
+        {
+            const int copyToBufferSizeRequested = 0x8000;
+
+            // we'll request a specific size buffer, but we cannot guarantee that's the size
+            // that will be used since CopyTo will rent from the array pool
+            // take advantage of this knowledge to find out what size it will actually use
+            var rentedBuffer = ArrayPool<byte>.Shared.Rent(copyToBufferSizeRequested);
+            int actualBufferSize = rentedBuffer.Length;
+            ArrayPool<byte>.Shared.Return(rentedBuffer);
+
+            // use 3 buffers-full so that we can prime the stream with the first buffer-full,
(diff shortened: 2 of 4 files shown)


ericstj:
The main reason I added that test was so that I could reliably produce cases where the `Inflate` would return 0 bytes when not finished - to test the fix made to copy-to that mis-interpreting that.  It would be nice if we still had a way to produce those empty chunks since it's a valuable test case.  Could we do that by just having the test write the raw byte sequence for an empty chunk instead?

I agree that someone might _notice_ that zero-byte writes produce nothing when they previously produced these empty chunks - but I agree the new behavior is more desirable.  Zips with empty files will change with this - for the better.  Agreed that https://github.com/dotnet/corefx/pull/4917 changed this unintentionally.

src/libraries/System.IO.Compression/tests/CompressionStreamUnitTests.Gzip.cs
```diff
@@ -55,64 +55,6 @@ private class DerivedMemoryStream : MemoryStream
...
-                {
-                    // write one byte in order to allow us to prime the inflater buffer
-                    gz.WriteByte(3);
-                }
-
-                while (compressedStream.Length < minCompressedSize)
-                {
-                    using (var gz = new GZipStream(compressedStream, CompressionLevel.NoCompression, leaveOpen: true))
-                    {
-                        gz.Write(Array.Empty<byte>());
```
ericstj, line 88: Could we preserve this test by writing raw bytes to `compressedStream`?

stephentoub: Done

carlossanlop:
>  In general with compression we make no guarantees version to version about generated bytes other than that they'll roundtrip

> I agree that someone might notice that zero-byte writes produce nothing when they previously produced these empty chunks - but I agree the new behavior is more desirable

Should we mark this as a breaking change?

src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateZLib/DeflateStream.cs
```diff
@@ -507,8 +507,13 @@ internal void WriteCore(ReadOnlySpan<byte> buffer)
             EnsureCompressionMode();
             EnsureNotDisposed();
 
-            Debug.Assert(_deflater != null);
+            if (buffer.IsEmpty)
+            {
+                return;
+            }
+
             // Write compressed the bytes we already passed to the deflater:
```
carlossanlop, line 515: I know it's not from your change, but shouldn't this message say this?:

```suggestion
            // Write the compressed bytes we already passed to the deflater:
```

src/libraries/System.IO.Compression/tests/CompressionStreamUnitTests.Gzip.cs
carlossanlop: Nice case.

carlossanlop (approved):

src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateZLib/DeflateStream.cs
```diff
@@ -793,8 +798,9 @@ internal ValueTask WriteAsyncMemory(ReadOnlyMemory<byte> buffer, CancellationTok
             EnsureNoActiveAsyncOperation();
             EnsureNotDisposed();
 
-            return cancellationToken.IsCancellationRequested ?
-                ValueTask.FromCanceled(cancellationToken) :
+            return
+                cancellationToken.IsCancellationRequested ? ValueTask.FromCanceled(cancellationToken) :
+                buffer.IsEmpty ? default :
```
ViktorHofer, line 803: This is a bit hard to read outside of an IDE. Consider adding parentheses around the second inline if else statement.

ViktorHofer (approved):

src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateZLib/DeflateStream.cs
stephentoub: Technically I think it's more correct as it currently is, even though it reads awkwardly.  We didn't pass compressed bytes: we passed uncompressed bytes, and we're then writing out the result, so we're writing, compressed, the bytes we already passed to the defaltor.

src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateZLib/DeflateStream.cs
stephentoub: I read this as:
```C#
if (cancellationToken.IsCancellationRequested) return ValueTask.FromCanceled(cancellationToken);
if (buffer.IsEmpty) return default;
return Core(buffer, cancellationToken);
```
Where are you recommending parens go?

src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateZLib/DeflateStream.cs
carlossanlop: Oh! I see it now. it's weird but now it makes sense.

ericstj:
@stephentoub did you create a breaking change doc for this?

PR #94438, opened 2023-11-06
Avoid generating code for unsupported types in config source gen

tarekgh:
https://github.com/dotnet/runtime/issues/94117


tarekgh:
CC @eerhardt

tarekgh:
Note, this is a candidate change to port to the 8.0 servicing.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -506,6 +506,42 @@ private TypeSpec CreateEnumerableSpec(TypeParseInfo typeParseInfo)
                 };
             }
 
+            private bool ConvertTo(ITypeSymbol source, ITypeSymbol dest)
```
ericstj, line 509: Nit
```suggestion
            private bool CanConvertTo(ITypeSymbol source, ITypeSymbol dest)
```

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -506,6 +506,42 @@ private TypeSpec CreateEnumerableSpec(TypeParseInfo typeParseInfo)
...
+                var conversion = _typeSymbols.Compilation.ClassifyConversion(source, dest);
+                return conversion.IsReference && conversion.IsImplicit;
+            }
+
+            private bool IsUnsupportedType(ITypeSymbol type)
+            {
+                if (SymbolEqualityComparer.Default.Equals(_typeSymbols.IntPtr, type)  ||
+                    SymbolEqualityComparer.Default.Equals(_typeSymbols.UIntPtr, type) ||
+                    SymbolEqualityComparer.Default.Equals(_typeSymbols.SerializationInfo, type) ||
+                    ConvertTo(type, _typeSymbols.MemberInfo) ||
```
ericstj, line 520:
```suggestion
                    CanConvertTo(type, _typeSymbols.MemberInfo) ||
```

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -506,6 +506,42 @@ private TypeSpec CreateEnumerableSpec(TypeParseInfo typeParseInfo)
...
+                    ConvertTo(type, _typeSymbols.Delegate))
```
ericstj, line 521:
```suggestion
                    CanConvertTo(type, _typeSymbols.Delegate))
```

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Parser/KnownTypeSymbols.cs
```diff
@@ -120,7 +125,11 @@ public KnownTypeSymbols(CSharpCompilation compilation)
             Enum = compilation.GetSpecialType(SpecialType.System_Enum);
             ArgumentNullException = compilation.GetBestTypeByMetadataName(typeof(ArgumentNullException));
 
-            String = compilation.GetBestTypeByMetadataName(typeof(string));
```
ericstj, line 123: Was this removal intentional?  If so shouldn't the property be removed as well?

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.cs
```diff
@@ -965,5 +965,150 @@ public abstract class AbstractType_CannotInit
...
+                }
+                """;
+
+            ConfigBindingGenRunResult result = await VerifyAgainstBaselineUsingFile(
+                "UnsupportedTypes.generated.txt",
+                source);
+
+            Assert.Empty(result.Diagnostics);
+            Assert.True(result.GeneratedSource.HasValue);
+            Assert.DoesNotContain(result.GeneratedSource.Value.SourceText.ToString(), "SkipProp");
```
ericstj, line 1111: Could we also assert no occurrence of the unsupported types, not the types mentioned in generic parameters for any omitted member?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -580,7 +616,7 @@ private ObjectSpec CreateObjectSpec(TypeParseInfo typeParseInfo)
                     ImmutableArray<ISymbol> members = current.GetMembers();
                     foreach (ISymbol member in members)
                     {
-                        if (member is IPropertySymbol { IsIndexer: false, IsImplicitlyDeclared: false } property)
+                        if (member is IPropertySymbol { IsIndexer: false, IsImplicitlyDeclared: false } property && !IsUnsupportedType(property.Type))
```
ericstj, line 619: Do we need to do anything different for constructor initialization? I suppose it's ok just to omit the members and then expect them to be missing from constructor- when present the error message might be confusing. Not sure if there's anything else rough about that scenario

src/libraries/Microsoft.Extensions.Configuration.Binder/Microsoft.Extensions.Configuration.Binder.sln
```diff
@@ -1,4 +1,8 @@
-﻿Microsoft Visual Studio Solution File, Format Version 12.00
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
eiriktsarpalis, line 2: My VS also keeps auto-updating the sln files in this repo. I'm conflicted about whether to include the changes in my PRs. @ViktorHofer what's the recommended course of action here?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -506,6 +506,42 @@ private TypeSpec CreateEnumerableSpec(TypeParseInfo typeParseInfo)
                 };
             }
 
+            private bool ConvertTo(ITypeSymbol source, ITypeSymbol dest)
+            {
+                var conversion = _typeSymbols.Compilation.ClassifyConversion(source, dest);
```
eiriktsarpalis, line 511: Nit: type name?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -506,6 +506,42 @@ private TypeSpec CreateEnumerableSpec(TypeParseInfo typeParseInfo)
...
+                return conversion.IsReference && conversion.IsImplicit;
```
eiriktsarpalis, line 512: Not super familiar with the `ClassifyConversion` APIs but am I right to guess that this just checks that `source` is a subtype of `dest`? If so, perhaps this could be named to something like "IsAssignableTo" since I would tend to to associate the term "conversion" with a broader spectrum of checks.

src/libraries/Microsoft.Extensions.Configuration.Binder/Microsoft.Extensions.Configuration.Binder.sln
ViktorHofer: Please revert any sln header updates. We asked the VS team to stop auto-updating the format version but weren't successfully yet.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
eiriktsarpalis: Could we extend this for all types in System.Reflection somehow? For example, `ParameterInfo` is not a subtype of `MemberInfo`.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -506,6 +506,42 @@ private TypeSpec CreateEnumerableSpec(TypeParseInfo typeParseInfo)
...
+                }
+
+                if (type is IArrayTypeSymbol arrayTypeSymbol)
+                {
+                    return arrayTypeSymbol.Rank > 1 || IsUnsupportedType(arrayTypeSymbol.ElementType);
+                }
+
+                if (type is INamedTypeSymbol namedTypeSymbol && namedTypeSymbol.IsGenericType)
+                {
+                    foreach (var typeArguments in namedTypeSymbol.TypeArguments)
```
eiriktsarpalis, line 533: Nit: type name

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -506,6 +506,42 @@ private TypeSpec CreateEnumerableSpec(TypeParseInfo typeParseInfo)
...
+                {
+                    return true;
+                }
+
+                if (type is IArrayTypeSymbol arrayTypeSymbol)
+                {
+                    return arrayTypeSymbol.Rank > 1 || IsUnsupportedType(arrayTypeSymbol.ElementType);
+                }
+
+                if (type is INamedTypeSymbol namedTypeSymbol && namedTypeSymbol.IsGenericType)
```
eiriktsarpalis, line 531: Not crucial, but I sometimes express this as:
```suggestion
                if (type is INamedTypeSymbol { IsGenericType: true }namedTypeSymbol)
```

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
eiriktsarpalis, line 533: I don't think walking the type arguments and failing if one of them is unsupported is the correct approach. For example, I could have something like:
```C#
public record MyPoco<TUnused>(int Value);
```
Then binding configuration for something like `MyPoco<Func<int>>` would be perfrectly valid from data contract perspective. I think we should only be rejecting types if one of their property types (or element types in the case of collections) is unsupported.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/Parser/KnownTypeSymbols.cs
eiriktsarpalis: Also, if we do end up keeping this it should be initialized using `SpecialType.System_String` -- resolving by metadata name can be expensive.

tarekgh: If you scroll up a little you will find it is already defined 
```C#
            String = compilation.GetSpecialType(SpecialType.System_String);
```

I have no idea why we defined it twice.

src/libraries/Microsoft.Extensions.Configuration.Binder/Microsoft.Extensions.Configuration.Binder.sln
tarekgh: Ok, I'll revert this file.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: Can you clarify more? are you asking to rename the arguments?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: I am fine renaming it to whatever. I was using the name we are using in the options source gen.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: We thought about that and preferred to keep it like that. The reason is we already have a real case doing that and we shouldn't generate code for, in the original issue I pointed at example https://github.com/dotnet/extensions/blob/14943b99572078f1bcf2c51e9507cf2631133fae/src/Libraries/Microsoft.Extensions.Resilience/Resilience/Internal/FailureEventMetricsOptions.cs#L11. Second, if we run into some issue in the future (as this is not a popular scenario) we can consider tweaking this if needed. For now, we should be a little bit aggressive.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.cs
tarekgh: Sometimes we generate code using such types (not for binding) here is examples:

```C#

public static int ParseInt(string value, Func<string?> getPath)

public static BinderOptions? GetBinderOptions(Action<BinderOptions>? configureOptions)

```

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: That is a good point. I'll try fix the constructor case. I cannot think of any other scenario that needs to be handled here.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: I am seeing `ParameterInfo` is the only missing one. I cann this. any idea why text json didn't do it as we were trying to do what text json doing?

eiriktsarpalis: It's an omission, it should be added there as well eventually (though probably `ParameterInfo` is not a super common scenario compared to `Type` and `MemberInfo`).

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
eiriktsarpalis: Replacing `var` with the explicit type name.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
eiriktsarpalis: That particular example though falls under the element type not being supported. You would still need to check for that independently because it may or may not appear in the generic arguments of the particular type, i.e.
```C#
internal sealed class FailureEventMetricsOptions
{
    // No type arguments visible in the property type.
    public MyCustomDictionary  Factories { get; } = [];
}

public class MyCustomDictionary : IDictionary<Dictionary<Type, Func<object, FailureResultContext>>>
{
}
```
Point being, type arguments are independent of what data the type contains.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: ah, got it.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: this is renamed to `IsAssignableTo`

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: this is renamed to `IsAssignableTo`. Also I added a support for `ParameterInfo` too.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: this is renamed to `IsAssignableTo`

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: I did the suggested constructor parameters check.

src/libraries/Microsoft.Extensions.Configuration.Binder/Microsoft.Extensions.Configuration.Binder.sln
tarekgh: Done the reverting.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: done.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: Done renaming the method to `IsAssignableTo`

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: This code is removed now.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: This code is removed.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: I applied your suggestion not checking the generic type parameters and instead I am covering the collection types.

tarekgh:
@ericstj @eiriktsarpalis I believe I addressed all feedback now.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -111,7 +112,7 @@ private void CreateTypeSpecs(CancellationToken cancellationToken)
                     if (!_createdTypeSpecs.ContainsKey(typeSymbol))
                     {
                         _createdTypeSpecs.Add(typeSymbol, CreateTypeSpec(typeParseInfo));
-                    }
+                                            }
```
eiriktsarpalis, line 115: Looks accidental?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -342,6 +343,11 @@ private TypeSpec CreateArraySpec(TypeParseInfo typeParseInfo)
                     return CreateUnsupportedTypeSpec(typeParseInfo, NotSupportedReason.MultiDimArraysNotSupported);
                 }
 
+                if (IsUnsupportedType(typeSymbol.ElementType))
```
eiriktsarpalis, line 346: Am I right to assume that array types are special cases because they follow a separate code path than the `CreateEnumerableSpec` method?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -506,6 +522,64 @@ private TypeSpec CreateEnumerableSpec(TypeParseInfo typeParseInfo)
...
+                        return IsUnsupportedType(elementType);
+                    }
+                }
+
+                return false;
+            }
+
+            private bool ConstructorParametersContainUnsupportedType(IMethodSymbol ctor)
+            {
+                foreach (IParameterSymbol parameter in ctor.Parameters)
```
eiriktsarpalis, line 572: Nit: could this use `Enumerable.Any`?

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
```diff
@@ -506,6 +522,64 @@ private TypeSpec CreateEnumerableSpec(TypeParseInfo typeParseInfo)
...
+                    SymbolEqualityComparer.Default.Equals(_typeSymbols.UIntPtr, type) ||
+                    SymbolEqualityComparer.Default.Equals(_typeSymbols.SerializationInfo, type) ||
+                    SymbolEqualityComparer.Default.Equals(_typeSymbols.ParameterInfo, type) ||
+                    IsAssignableTo(type, _typeSymbols.MemberInfo) ||
+                    IsAssignableTo(type, _typeSymbols.Delegate))
+                {
+                    return true;
+                }
+
+                if (type is IArrayTypeSymbol arrayTypeSymbol)
```
eiriktsarpalis, line 548: Given that you're already checking for key types and elements in the `Create*Spec` methods, I don't think repeating the checks here are strictly necessary.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: We talked offline. Yes this is special casing the array case.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.cs
```diff
@@ -965,5 +965,153 @@ public abstract class AbstractType_CannotInit
...
+                		ConfigurationBuilder configurationBuilder = new();
+                		IConfiguration configuration = configurationBuilder.Build();
+
+                        // var ws = config.GetSection("MySection");
+                        var w = new Options();
+                        configuration.Bind(w);
+                        configuration.Get<Options>(_ => { });
+
+                        // Should generate a diagnostics
+                        configuration.Get<List<IntPtr>>(_ => { });
```
eiriktsarpalis, line 993: Turning this into a theory would let us test for other root type as well:
```suggestion
                        configuration.Get<{{unsupportedRootType}}>(_ => { });
```
Note that this would require `$$"""` syntax at the start of your triple-quote string.

src/libraries/Microsoft.Extensions.Configuration.Binder/gen/ConfigurationBindingGenerator.Parser.cs
tarekgh: We talked offline. We are trying to avoid producing diagnostics when handling properties. For root types we still produce diagnostics.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.cs
```diff
@@ -965,5 +965,153 @@ public abstract class AbstractType_CannotInit
...
+
+                    public ConstructorInfo? SkipProp57 { get; set; }
+                    public List<ConstructorInfo?> SkipProp58 { get; set; }
+                    public List<Dictionary<string, ConstructorInfo?>> SkipProp59 { get; set; }
+
+                    public string[,]? SkipProp60 { get; set; }
+                    public List<string[,]?> SkipProp61 { get; set; }
+                    public List<Dictionary<string, string[,]?>> SkipProp62 { get; set; }
+
+                    public Func<string>?[] SkipProp63 { get; set; }
```
eiriktsarpalis, line 1089: Similarly, you could try reducing the skipped property to just one and fill in using the input from the theory.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.cs
tarekgh: It is a good idea. The drawback of this will make the test much slower as we compare the result against the baseline.

tarekgh: For root testing, I can split it to another test and don't have to compare to baseline as I want to check the diagnostics for this case.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.cs
tarekgh: The drawback of this will make the test much slower as we compare the result against the baseline.

tarekgh: I can pass the whole list as one shot through to the theory.

src/libraries/Microsoft.Extensions.Configuration.Binder/tests/SourceGenerationTests/GeneratorTests.Baselines.cs
```diff
@@ -965,5 +965,153 @@ public abstract class AbstractType_CannotInit
...
+                        configuration.Get<List<IntPtr>>(_ => { });
+                	}
+                }
+
+                public class Options
+                {
+                    public string? Name { get; set; }
+                    public int Age { get; set; }
+                    public List<string> List { get; set; }
+                    public string[] Array { get; set; }
```
eiriktsarpalis, line 1002: Could you also add:

1. A POCO with a phantom type that doesn't get skipped, i.e. for `record Record<T>(int x);` ensure `Record<Action>` works as expected.
2. a non-generic type that explicitly implements a collection type with an unsupported collection element, i.e. `class MyCollection : IDictionary<string, Action>`.

eiriktsarpalis (approved):
LGTM, thanks.

PR #94440, opened 2023-11-06
Add document listing supported feature checks

sbomer:
The ILLink roslyn analyzer, ILLink, and ILCompiler all support different forms of branch removal and constant propagation. This document is meant to describe the supported/recommended patterns that should work with all of these tools. Related to the discussion in https://github.com/dotnet/linker/issues/2715.


docs/design/tools/illink/feature-checks.md
```diff
@@ -0,0 +1,281 @@
...
+
+```csharp
+Debug.Assert (Feature.IsSupported);
+
+Feature.Run(); // OK in analyzer, warns in ILLLink/ILCompiler
+```
+
+However, ILLink and ILCompiler don't optimize away branches based on `DoesNotReturnIfAttribute`, because the attribute doesn't have runtime semantics, and an incorrectly-applied attribute could result in the guarded branch being reachable at runtime. Or, in the case of `Debug.Assert`, if the assert is used incorrectly, the guarded code may still be reachable at runtime in `Release` builds.
+
+Here are some more examples:
```
vitek-karas, line 245: I think the recommendation should be to not rely on these attributes to avoid trim/AOT warnings - we want to recommend patterns which work the same everywhere.

vitek-karas (approved):
I assume you checked we have tests for each of these cases and that they behave as described here :-)

Looks good, thanks for writing this down.

MichalStrehovsky (approved):
Thank you for writing it down.

Depending on who the target audience is, should we also say that yield/async/etc state machines also break all of this outside the analyzer?

PR #94444, opened 2023-11-07
Merge remaining JIT tests

trylek:


src/tests/JIT/JIT_others.csproj
```diff
@@ -0,0 +1,12 @@
...
+    <MergedWrapperProjectReference Include="*/**/*.??proj" />
+
+    <MergedWrapperProjectReference Remove="*/**/*_d.??proj" />
+    <MergedWrapperProjectReference Remove="*/**/*_do.??proj" />
+    <MergedWrapperProjectReference Remove="*/**/*_r.??proj" />
+    <MergedWrapperProjectReference Remove="*/**/*_ro.??proj" />
+  </ItemGroup>
+
+  <Import Project="JIT_merged.targets" />
+</Project>
```
markples, line 12: I'm a bit concerned that this one especially (but also JIT_r, etc.) will be slow due to processing all the projects multiple times.  However, it's quite possible that this is overreacting since it is "just" the project names.

src/tests/JIT/JIT_merged.targets
markples, line 1: I found this filename confusing, especially how "Import JIT_merged.targets" handles filtering out the tests explicitly in other merged groups.  Maybe something like "Import remove_stuff.targets" (maybe .targets isn't the right extension there) and "import mergedtestrunner"?

src/tests/JIT/Stress/ABI/pinvokes_d.csproj
```diff
@@ -1,14 +1,19 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <OutputType>Exe</OutputType>
+    <!-- Needed for GCStressIncompatible, JitOptimizationSensitive -->
+    <RequiresProcessIsolation>true</RequiresProcessIsolation>
+
+    <!-- Main supports command-line arguments -->
+    <ReferenceXUnitWrapperGenerator>false</ReferenceXUnitWrapperGenerator>
+    <CLRTestExecutionArguments>--pinvokes --num-calls 1000 --no-ctrlc-summary</CLRTestExecutionArguments>
```
markples, line 8: I found this kind of test frustrating to convert.  I'd like to have something like

```
[Theory]
[InlineData("--pinvokes ...")]
public static test(string[] args)

// somehow not interfering with ReqProcIso/BuildAsStandalone
```

OR have the test have symbolic inputs (`bool pinvokes`) and only use the parser for Main.  But that's a lot of work so probably too much (not just for this PR but in general).

src/tests/JIT/JIT_others.csproj
trylek: Hmm, I think there should be some caching happening on the OS side, so far I haven't observed any huge pauses in the build because of this but I'll double-check in the next run, thanks for the heads-up.

jkoritzinsky (approved):
The mechanical conversion looks good. I agree with Mark's comments, but otherwise LGTM.

src/tests/JIT/JIT_merged.targets
trylek: Fixed, thanks for pointing that out!

src/tests/JIT/JIT_others.csproj
trylek: According to my observations this operation takes a few seconds at most, far less than the actual build times, so for now I treat it as just a nice-to-have for a future round of infra improvements.

src/tests/JIT/Stress/ABI/pinvokes_d.csproj
trylek: I completely agree there are many tests that can be improved in a similar vein. As I mentioned before, I still believe that the most immediate value can be had by just finishing the test merging process and disposing of the legacy infrastructure but I'm definitely looking forward to further cleanups in our vast test portfolio.

Issue #48796, opened 2021-02-26
Introduce Native Varargs attribute to indicate arguments are to be marshalled as native varargs


AaronRobinsonMSFT:
## Background and Motivation

Native varargs are a complicated interop scenario to support. At present, native varargs are only supported on the Windows platform through the undocumented `__arglist` keyword. Supporting varargs naturally in a P/Invoke scenario would be difficult from the C# language. However, it is possible to compromise by permitting support for the call with a fully specified `DllImport` signature and a hint from the user.

User scenario: https://github.com/dotnet/runtime/issues/48752

## Proposed API

```diff
namespace System.Runtime.InteropServices
{
+    [AttributeUsage(AttributeTargets.Method)]
+    public class NativeVarargsAttribute : Attribute
+    {
+        public NativeVarargsAttribute() { VarargBeginIndex = 0; }
+
+        /// <summary>
+        /// Zero-based index of the first variable argument.
+        /// </summary>
+        public int VarargBeginIndex;
+    }
}
```

## Usage Examples

Consider the following native export with varargs:

```c
void Varargs(int n, ...);
```

The following P/Invoke declarations would enable users to call and properly forward the arguments in a supported multi-platform manner.

```csharp
[NativeVarargsAttribute(VarargBeginIndex = 1)]
[DllImport(@"NativeLibrary.dll", EntryPoint = "Varargs")]
static extern void Varargs0(int n);

[NativeVarargsAttribute(VarargBeginIndex = 1)]
[DllImport(@"NativeLibrary.dll", EntryPoint = "Varargs")]
static extern void Varargs1(int n, int a);

[NativeVarargsAttribute(VarargBeginIndex = 1)]
[DllImport(@"NativeLibrary.dll", EntryPoint = "Varargs")]
static extern void Varargs2(int n, int a, int b);
```

## Alternative designs

Encode the information in the [`CallingConvention`](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.callingconvention) enum. This approach does remove the overhead of attribute reading, but does miss the added data of knowing where the varargs start - at present doesn't appear to be needed. This approach also impacts existing metadata tooling (for example, ILDasm, ILAsm, and ILSpy). See https://github.com/dotnet/runtime/issues/48796#issuecomment-786355000.

```diff
public enum CallingConvention
{
+    VarArg = 6
}
```

## Current state

[...]

PR #94446, opened 2023-11-07
[wasm] p/invoke and interop improvements

kg:
* Refactor logging in the wasm build tasks to use a LogAdapter helper class that can either write to stdout/stderr or via the msbuild logging APIs
* Refactor ManagedToNativeGenerator so that it can run in either in-process or out-of-process modes (but it still runs in-process right now). This will enable us to pivot to out-of-process in order to pick up function pointer support later on.
* Add support for function pointer types via reflection (so that we can still build and run on net4x for visual studio). Right now the underlying libraries are too old for this support to work, but once the libraries are updated it will.
* Add basic support for structs as arguments
* Implement scalar struct behavior from the wasm C ABI (structs containing one scalar are passed by-value instead of by-reference) in various places in the build tasks, interpreter, and AOT compiler
* Remove a use of varargs pinvoke from SystemNative because it isn't compatible with the WASM C ABI (it's replaced with two non-varargs pinvokes) 
* Generate more accurate C signatures for pinvokes


kg:
cc @lewing

src/tasks/WasmAppBuilder/LogAdapter.cs
```diff
@@ -0,0 +1,71 @@
...
+            else
+                return s;
+        }
+
+        public void LogMessage (string s, params object[] o) {
+            Helper?.LogMessage(s, o);
+            Output?.WriteLine(AutoFormat(s, o));
+        }
+
+        public void LogMessage (MessageImportance mi, string s, params object[] o) {
```
radical, line 42: Should this also drop `Log` prefix like the other methods?

radical (dismissed):

src/tasks/WasmAppBuilder/LogAdapter.cs
```diff
@@ -0,0 +1,71 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.IO;
+using Microsoft.Build.Utilities;
+using Microsoft.Build.Framework;
+
+namespace WasmAppBuilder {
```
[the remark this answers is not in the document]
kg: I did this originally as requested and it broke the build.

src/tasks/WasmAppBuilder/LogAdapter.cs
```diff
@@ -0,0 +1,71 @@
...
+            Output?.WriteLine(AutoFormat(s, o));
+        }
+
+        public void LogMessage (MessageImportance mi, string s, params object[] o) {
+            Helper?.LogMessage(mi, s, o);
+            Output?.WriteLine(AutoFormat(s, o));
+        }
+
+        public void Info (string code, string message, params object[] args) {
+            // We use MessageImportance.High to ensure this appears in build output, since
```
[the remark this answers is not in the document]
kg: The warnings go nowhere otherwise, we need a way to pipe diagnostic information to the user, since we can't use Warning severity

src/tasks/WasmAppBuilder/ManagedToNativeGenerator.cs
```diff
@@ -104,6 +137,119 @@ private void ExecuteInternal()
...
+            envVars: null,
+            workingDir: null,
+            silent: false,
+            logStdErrAsMessage: true
+        );
+        if (exitCode != 0)
+            logAdapter.Error("WASM0066", $"WasmAppBuilder external process failed with exit code {exitCode}!");
+    }
+
+    internal readonly record struct ExecuteArguments (
```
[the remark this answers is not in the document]
kg: can structs be sealed? you can't inherit from them

src/tasks/WasmAppBuilder/ManagedToNativeGenerator.cs
```diff
@@ -66,33 +66,66 @@ public override bool Execute()
...
+                    return filePath;
+                }
+            }
+            catch
             {
-                Log.LogMessage(MessageImportance.Low, $"Loading {asmPath} to scan for pinvokes, and icalls");
-                Assembly asm = mlc.LoadFromAssemblyPath(asmPath);
-                pinvoke.ScanAssembly(asm);
-                icall.ScanAssembly(asm);
+                // If we can't read a directory for any reason just skip it
```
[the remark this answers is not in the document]
kg: This behavior is from the existing .NET code I was asked to duplicate, I don't know why it's there

src/tasks/WasmAppBuilder/ManagedToNativeGenerator.cs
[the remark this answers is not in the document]
radical: Oops!

src/tasks/WasmAppBuilder/LogAdapter.cs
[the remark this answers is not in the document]
radical: Not sure I understand. Is that when running out-of-proc?

src/tasks/WasmAppBuilder/LogAdapter.cs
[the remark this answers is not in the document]
radical: oh? We use that all over the place in tasks.

src/tasks/WasmAppBuilder/LogAdapter.cs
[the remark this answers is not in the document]
kg: We build with warnaserror in a bunch of places, so the only way we can communicate information out of these build tasks without breaking the build is as high-importance messages, afaik

src/tasks/WasmAppBuilder/PInvokeTableGenerator.cs
```diff
@@ -386,12 +419,94 @@ private bool HasAssemblyDisableRuntimeMarshallingAttribute(Assembly assembly)
...
+
+            // HACK: SkiaSharp has pinvokes that rely on this
+            if (HasAttribute(type, "System.Runtime.InteropServices.UnmanagedFunctionPointerAttribute"))
+                return true;
+
+            if (type.Name == "__NonBlittableTypeForAutomatedTests__")
+                return false;
+
+            if (!type.IsValueType) {
+                log.Info("WASM0060", "Type {0} is not blittable: Not a ValueType", type);
```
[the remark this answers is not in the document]
kg: Can you provide an example of named warning codes? I'm unfamiliar with them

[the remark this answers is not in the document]
radical: I just meant for example - `const string WarnNotABlittableType = "WASM0060"`

src/tasks/WasmAppBuilder/LogAdapter.cs
kg: I left it this way so existing code would keep working, I can go through and rename all the call sites if you want (it looks like there are about 30)

radical: oh then I'm fine with either.

radical:
There are a couple of failures which seem relevant, like:

[System.Runtime.Intrinsics.Tests aot](https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-94446-merge-53f513650366430c87/normal-System.Runtime.Intrinsics.Tests/1/console.1a2c8876.log?helixlogtype=result):
```
[19:37:04] info: Starting:    System.Runtime.Intrinsics.Tests.dll
[19:37:05] info: /root/helix/work/workitem/e/wasm_build/AppBundle/_framework/dotnet.native.js:196: RuntimeError: null function or function signature mismatch
[19:37:05] info:       throw toThrow;
[19:37:05] info:       ^
[19:37:05] info: RuntimeError: null function or function signature mismatch
[19:37:05] info:     at aot_instances_System_Runtime_Intrinsics_Tests_Vectors_Vector64Tests_TestGetOne_uint16 (wasm://wasm/091f39e2:wasm-function[34037]:0x714a46)
[19:37:05] info:     at System_Runtime_Intrinsics_Tests_System_Runtime_Intrinsics_Tests_Vectors_Vector64Tests_GetOneUInt16 (wasm://wasm/091f39e2:wasm-function[23971]:0x497f53)
[19:37:05] info:     at aot_instances_aot_wrapper_gsharedvt_out_sig_void_this_ (wasm://wasm/091f39e2:wasm-function[29908]:0x5fda13)
[19:37:05] info:     at aot_instances_wrapper_runtime_invoke_object_runtime_invoke_sig_void_intptr_intptr_object_intptr_intptr_intptr (wasm://wasm/091f39e2:wasm-function[37318]:0x7e6ee8)
[19:37:05] info:     at mono_llvmonly_runtime_invoke (wasm://wasm/091f39e2:wasm-function[54970]:0xc48da7)
...
```

kg:
Interesting, it was passing tests before. Something must've changed in the BCL

mthalman:
The source build error is:

```
.packages/microsoft.dotnet.arcade.sdk/9.0.0-beta.23607.2/tools/SourceBuild/AfterSourceBuild.proj(68,5): error : (NETCORE_ENGINEERING_TELEMETRY=AfterSourceBuild) 1 new pre-builts discovered! Detailed usage report can be found at /__w/1/s/artifacts/source-build/self/prebuilt-report/baseline-comparison.xml.
See https://aka.ms/dotnet/prebuilts for guidance on what pre-builts are and how to eliminate them.
Package IDs are:
System.Reflection.MetadataLoadContext.8.0.0
```

This will require adding System.Reflection.MetadataLoadContext.8.0.0 to [SBRP](https://github.com/dotnet/source-build-reference-packages?tab=readme-ov-file#adding-new-packages). Once that is merged, it will produce an SBRP package that will flow into runtime ([example](https://github.com/dotnet/runtime/pull/96382)). Once that is merged into runtime main, merge the latest main branch into this PR and it should resolve the prebuilt.

kg:
It looks like SBRP hasn't updated since Jan 5. Is there something I need to do to make the flow happen? Should I just manually do it in this PR? https://github.com/dotnet/runtime/pull/96989 happened, which I thought would have updated it to a more recent commit, but it didn't. @mthalman

kg:
@BrzVlad @kotlarmilos @vargaz Any of you feel comfortable reviewing the interp/type system changes here? They're nuanced enough I want to make sure someone with expertise looks them over.

src/mono/mono/mini/mini-llvm.c
```diff
@@ -4948,6 +4948,8 @@ process_call (EmitContext *ctx, MonoBasicBlock *bb, LLVMBuilderRef *builder_ref,
 		if (!addresses [call->inst.dreg])
 			addresses [call->inst.dreg] = build_alloca_address (ctx, sig->ret);
 		emit_store (builder, lcall, convert_full (ctx, addresses [call->inst.dreg]->value, pointer_type (LLVMTypeOf (lcall)), FALSE), is_volatile);
+		load_name = "wasm_vtype_as_scalar";
```
kotlarmilos, line 4951: Should this be WASM only?

kotlarmilos (commented):
LGTM!

src/mono/mono/mini/aot-runtime-wasm.c
```diff
@@ -14,6 +14,9 @@
 
 #ifdef HOST_WASM
 
+gboolean
+mini_wasm_is_scalar_vtype (MonoType *type, MonoType **etype);
```
vargaz, line 18: This is declared in mini-wasm.h.

src/mono/mono/mini/mini-llvm.c
```diff
@@ -6154,8 +6157,15 @@ process_bb (EmitContext *ctx, MonoBasicBlock *bb)
 					}
 					break;
 				case LLVMArgWasmVtypeAsScalar:
-					g_assert (addresses [ins->sreg1]);
-					retval = LLVMBuildLoad2 (builder, ret_type, build_ptr_cast (builder, addresses [ins->sreg1]->value, pointer_type (ret_type)), "");
+					if (!addresses [ins->sreg1]) {
+						/* SIMD value */
+						g_assert (lhs);
+						retval = LLVMBuildBitCast (builder, lhs, ret_type, "");
+						mono_llvm_dump_value (lhs);
```
vargaz, line 6164: Leftovers.

src/mono/mono/mini/interp/interp.c
```diff
@@ -1345,10 +1345,22 @@ typedef enum {
...
 	int ilen, flen;
-	PInvokeArgType ret_type;
+	MonoType *ret_mono_type;
+	PInvokeArgType ret_pinvoke_type;
 	PInvokeArgType *arg_types;
 } BuildArgsFromSigInfo;
 
+static MonoType *
+filter_type_for_args_from_sig (MonoType *type) {
+#if defined(HOST_WASM) || defined(TARGET_WASM)
```
vargaz, line 1355: HOST_WASM should be enough.

src/native/libs/System.Native/pal_string.h
```diff
@@ -15,3 +15,12 @@
...
 PALEXPORT int32_t SystemNative_SNPrintF(char* string, int32_t size, const char* format, ...);
+
+/**
+ * Two specialized overloads for use from Interop.Sys, because these two signatures are not equivalent
+ * on some architectures (like 64-bit WebAssembly)
+*/
+
+PALEXPORT int32_t SystemNative_SNPrintF_1S(char* string, int32_t size, const char* format, char* str);
+
+PALEXPORT int32_t SystemNative_SNPrintF_1I(char* string, int32_t size, const char* format, int arg);
```
lewing, line 26: @AaronRobinsonMSFT how did you handle this in runtimelab ?

AaronRobinsonMSFT: This is the correct way to handle this scenario. Native varargs at the interop boundary have narrow support in .NET and are exclusive to the Windows platform and typically only for the C++/CLI scenario. See https://github.com/dotnet/runtime/issues/48796 for broader support.

The approach taken here is to provide an unmanaged export that managed code can call safely which then forwards to a native varargs function. This ensures the unmanaged calling convention is handled correctly during the P/Invoke and correctly at the dispatch to the unmanaged varargs.

It is ugly and not what anyone really wants, but the cost of supporting native varargs is expensive to support. This comes up rarely relative to other interop scenarios so just hasn't been prioritized. The aforementioned issue mentions some of the costs and relevant parties that would need to respond to support.

lewing: Makes perfect sense, just wanted to verify.

lewing:
@kg please double check the conflict resolution with https://github.com/dotnet/runtime/pull/97643/files#r1483684742 in mind

kg:
> @kg please double check the conflict resolution with https://github.com/dotnet/runtime/pull/97643/files#r1483684742 in mind

Looks right to me, and tests should fail if it's wrong. Thanks for sorting out the merge

lewing (approved):

kg:
S.T.J failures are known, merging

eng/Versions.props
```diff
@@ -123,16 +123,16 @@
...
     <SystemComponentModelAnnotationsVersion>5.0.0</SystemComponentModelAnnotationsVersion>
     <SystemDataSqlClientVersion>4.8.6</SystemDataSqlClientVersion>
     <SystemDrawingCommonVersion>8.0.0</SystemDrawingCommonVersion>
     <SystemIOFileSystemAccessControlVersion>5.0.0</SystemIOFileSystemAccessControlVersion>
     <SystemMemoryVersion>4.5.5</SystemMemoryVersion>
     <SystemReflectionMetadataVersion>9.0.0-alpha.1.24072.1</SystemReflectionMetadataVersion>
     <!-- Keep toolset versions in sync with dotnet/msbuild and dotnet/sdk -->
-    <SystemReflectionMetadataToolsetVersion>7.0.0</SystemReflectionMetadataToolsetVersion>
-    <SystemReflectionMetadataLoadContextVersion>7.0.0</SystemReflectionMetadataLoadContextVersion>
+    <SystemReflectionMetadataToolsetVersion>8.0.0</SystemReflectionMetadataToolsetVersion>
```
am11, line 134: Based on the comment above, this toolset version should not be updated without coordination with msbuild and sdk repos.

@ViktorHofer, FYI, this is blocking the codeflow https://github.com/dotnet/sdk/pull/38665 (while previous change made it to the next phases without problems https://github.com/dotnet/sdk/pull/38553). We can either revert this or update https://github.com/dotnet/sdk/blob/b32df9afff77e8f1afbd417e7c70841867cb6cdc/src/Tasks/Microsoft.NET.Build.Tasks/Microsoft.NET.Build.Tasks.csproj#L75 to 8.0 (if we choose the latter, we can close https://github.com/dotnet/runtime/issues/96809).

kg: I coordinated with people from the msbuild team at multiple points while developing this along with the SBRP team, so I'm sorry that something fell through the cracks here. Is there a specific step we missed?

kg: I see now from looking through the replies to Larry's comment on this issue that you told us there. For some reason, GitHub did not send me a notification for your comment, so I didn't see it :( I guess in the future blocking issues like that need to be PR comments instead of replies.

lewing: My preference would definitely be to take the bump in the sdk. VS 17.9 is the where these changes will ship and it has taken the update.

cc @ViktorHofer @marcpopMSFT

am11: NP, https://github.com/dotnet/runtime/issues/96809 is tracking it, but it seems (almost) resolved by your PR. Once confirmed, we can just update [sdk's Microsoft.NET.Build.Tasks.csproj#L74-L76](https://github.com/dotnet/sdk/blob/b32df9afff77e8f1afbd417e7c70841867cb6cdc/src/Tasks/Microsoft.NET.Build.Tasks/Microsoft.NET.Build.Tasks.csproj#L74-L76) to 8.0 (now that all the related toolings are using 8.0; msbuild, HostModel and VS).

ViktorHofer: I **think** for this to work, sdk would need to upgrade its .NET Framework leg to a VS 17.9 Preview image. I'm not sure if such is already available. Worst case, we would need to revert this dependency bump here. @lewing can you please make sure that this dependency flow issue gets resolved: https://github.com/dotnet/sdk/pull/38665?

akoeplinger: Correct me if I'm wrong but the sdk _build_ should be fine, it'd just need the testing legs to use VS17.9 Preview right? That should already be the case: https://github.com/dotnet/sdk/blob/0962c1f89f5daf924a9fe876c80e80b0bde63b0d/.vsts-ci.yml#L108-L111

I'll try bumping the .csproj in the sdk PR and see what happens.

ViktorHofer: Correct. Unification at run time.

akoeplinger: Looks like it worked :)

PR #94449, opened 2023-11-07
Disallow inlining Main

MichalStrehovsky:
When we compile managed code, `Main` is not the actual spot where execution of managed code starts. Instead it's the `StartupCodeMain` method that the compiler generates. This method is responsible for initializing the managed environment, calling `Main` and tearing down the environment. If `Main` is short enough, sometimes it gets inlined into `StartupCodeMain` this has bad impact on diagnostics (don't see `Main` in stack traces, can't set breakpoints). Pretend it was marked `NoInlining`.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
+++ b/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
@@ -1129,15 +1129,6 @@ namespace Internal.JitInterface
             {
                 result |= CorInfoFlag.CORINFO_FLG_FORCEINLINE;
             }
-#if !READYTORUN
-            else if (method.OwningType is EcmaType ecmaOwningType
-                && ecmaOwningType.EcmaModule.EntryPoint == method)
-            {
-                // Mark entrypoint as NoInlining for better debugging. We don't want it to inline
-                // into startup code because breakpoints won't hit and stack traces won't have Main.
-                result |= CorInfoFlag.CORINFO_FLG_DONT_INLINE;
-            }
-#endif
 
             if (method.OwningType.IsDelegate && method.Name == "Invoke")
             {
--- a/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
+++ b/src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
@@ -1129,6 +1129,15 @@ namespace Internal.JitInterface
             {
                 result |= CorInfoFlag.CORINFO_FLG_FORCEINLINE;
             }
+#if !READYTORUN
+            else if (method.OwningType is EcmaType ecmaOwningType
+                && ecmaOwningType.EcmaModule.EntryPoint == method)
+            {
+                // Mark entrypoint as NoInlining for better debugging. We don't want it to inline
+                // into startup code because breakpoints won't hit and stack traces won't have Main.
+                result |= CorInfoFlag.CORINFO_FLG_DONT_INLINE;
+            }
+#endif
 
             if (method.OwningType.IsDelegate && method.Name == "Invoke")
             {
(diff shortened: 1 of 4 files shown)


EgorBo:
Yeah it was annoying that `JitDisasm=Main` didn't work for NAOT sometimes🙂

src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
```diff
@@ -1129,6 +1129,15 @@ private uint getMethodAttribsInternal(MethodDesc method)
             {
                 result |= CorInfoFlag.CORINFO_FLG_FORCEINLINE;
             }
+#if !READYTORUN
```
EgorBo, line 1132: Does it need this `#if !READYTORUN` ?

MichalStrehovsky: It doesn't _need_ it, but it's a deoptimization if someone actually calls `Main`. It is valid to call `Main` from user code in .NET (as opposed to e.g. C++, where the language spec disallows it). I don't insist on keeping the ifdefs though...

src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
```diff
@@ -1129,6 +1129,15 @@ private uint getMethodAttribsInternal(MethodDesc method)
...
+            else if (method.OwningType is EcmaType ecmaOwningType
+                && ecmaOwningType.EcmaModule.EntryPoint == method)
+            {
+                // Mark entrypoint as NoInlining for better debugging. We don't want it to inline
+                // into startup code because breakpoints won't hit and stack traces won't have Main.
```
jkotas, line 1137: If we want to the stacktraces to always have Main, we should also disable tailcalls from Main in `canTailCall`  JIT/EE method.

BTW: JIT does that too - https://github.com/dotnet/runtime/blob/021ecb8eb7998bf3288ee68b12e952f775828c10/src/coreclr/vm/jitinterface.cpp#L8229-L8237

src/coreclr/tools/Common/JitInterface/CorInfoImpl.cs
jkotas: We can mark `MainMethodWrapper` as NoOptimization instead. It would not required ifdef and it would have a bit better fidelity with JIT behavior - Main would stay inlineable.

https://github.com/dotnet/runtime/blob/021ecb8eb7998bf3288ee68b12e952f775828c10/src/coreclr/tools/aot/ILCompiler.Compiler/IL/Stubs/StartupCode/StartupCodeMainMethod.cs#L207

src/coreclr/tools/aot/ILCompiler.RyuJit/JitInterface/CorInfoImpl.RyuJit.cs
```diff
@@ -799,6 +800,14 @@ private bool canTailCall(CORINFO_METHOD_STRUCT_* callerHnd, CORINFO_METHOD_STRUC
             {
                 MethodDesc caller = HandleToObject(callerHnd);
 
+                if (caller.OwningType is EcmaType ecmaOwningType
+                    && ecmaOwningType.EcmaModule.EntryPoint == caller)
+                {
+                    // We don't want to tailcall the entrypoint for an application; It results in a rather
+                    // confusing debugging experience.
```
jkotas, line 807:
```suggestion
                    // Do not tailcall from the application entrypoint.
                    // We want Main to be visible in stack traces.
```
We are disabling tailcalls *from* the entrypoint. Also, could you please update the same comment in the VM to match?

jkotas (approved):
Thanks

src/coreclr/tools/aot/ILCompiler.Compiler/IL/Stubs/StartupCode/StartupCodeMainMethod.cs
```diff
@@ -268,6 +287,9 @@ public override MethodIL EmitIL()
                 if (Context.Target.IsWindows)
                     codeStream.MarkDebuggerStepInPoint();
 
+                // This would be tail call eligible but we don't do tail calls
+                // if the method is marked NoOptimization and we just did it above.
```
MichalStrehovsky, line 291:
```suggestion
                // if the method is marked NoInlining and we just did it above.
```

PR #94450, opened 2023-11-07
Do not release frozen segment if we failed extending it

MichalStrehovsky:
Maybe I'm missing something, but if we try to commit more of the reserved segment and this fails, it doesn't feel right to release the whole thing. The previously committed part of the segment is still in use.


EgorBo:
This was a suggestion from @jkotas:

image

I don't have a strong opinion here, presumably if Commit fails we're in a bad state anyway?

EgorBo:
> That comment seems to be for the call in the constructor of FrozenObjectSegment 

Ah, right, then sounds OK to have it, unless @jkotas has an opposite opinion

EgorBo (approved):

jkotas (approved):
Good catch

PR #94451, opened 2023-11-07
Use live M.Bcl.AsyncInterfaces dependency

ViktorHofer:
Microsoft.Bcl.AsyncInterfaces is still built in the main branch and shouldn't be brought in as a prebuilt by the TimeProvider project.


ericstj (approved):
LGTM.  I assume we need to service this as well?  Maybe just as infrastructure so that the package is set up to use latest just in case it needs to ship in the future.

ViktorHofer:
Yes, we should backport this into release/8.0.

ViktorHofer:
Here's the manual backport: https://github.com/dotnet/runtime/pull/94459

tarekgh (approved):

PR #94453, opened 2023-11-07
Fix Comvariant Failure on s390x

saitama951:
The CI Reported two test failure on S390x
```
System.Runtime.InteropServices.Tests.ComVariantMarshallerTests.Decimal_Marshals_To_DECIMAL [FAIL]
System.Runtime.InteropServices.Tests.ComVariantTests.Decimal [FAIL]
```
cc: @uweigand

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/ComVariant.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/ComVariant.cs
@@ -47,8 +47,13 @@ namespace System.Runtime.InteropServices.Marshalling
         [StructLayout(LayoutKind.Sequential)]
         private struct TypeUnion
         {
+#if BIGENDIAN
+            public ushort _wReserved1;
+            public ushort _vt;
+#else
             public ushort _vt;
             public ushort _wReserved1;
+#endif
             public ushort _wReserved2;
             public ushort _wReserved3;
 
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/ComVariant.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/ComVariant.cs
@@ -47,6 +47,7 @@ namespace System.Runtime.InteropServices.Marshalling
         [StructLayout(LayoutKind.Sequential)]
         private struct TypeUnion
         {
+            // The layout of _wReserved1 and _vt fields needs to match Decimal._flags
 #if BIGENDIAN
             public ushort _wReserved1;
             public ushort _vt;


src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/ComVariant.cs
```diff
@@ -47,10 +47,17 @@ static unsafe ComVariant()
         [StructLayout(LayoutKind.Sequential)]
         private struct TypeUnion
         {
-            public ushort _vt;
-            public ushort _wReserved1;
-            public ushort _wReserved2;
-            public ushort _wReserved3;
+            #if TARGET_S390X
```
huoyaoyuan, line 50: The condition should be endianness.

Is the unit test using inappropriate reinterpret-cast? Is variant ever defined on non-Windows?

saitama951: @huoyaoyuan 
This is the test case, 
```
decimal value = 42.0m;
ComVariant variant = ComVariantMarshaller.ConvertToUnmanaged(value);
Assert.Equal(value, ComVariantMarshaller.ConvertToManaged(variant));
ComVariantMarshaller.Free(variant);
```
The `ComVariantMarshaller.ConvertToManaged` inturn ends up here 
https://github.com/dotnet/runtime/blob/bd27e2743588b11a67acd9e6b622d9f16cb6c872/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/ComVariant.cs#L497-L507
where I guess could be an endian issue relating to structure members

AaronRobinsonMSFT: This doesn't look like the correct location to fix this. The endianness issues are specifically related to the decimal conversion and not to the general layout of the variant type. In this case, reordering the fields doesn't seem correct and we should instead look at where the decimal conversion is taking place.

AaronRobinsonMSFT: > Is the unit test using inappropriate reinterpret-cast? Is variant ever defined on non-Windows?

@huoyaoyuan In this case, yes. This type is used for the COM source generator and enables narrowly defined cross-platform support for COM interop.

saitama951: @AaronRobinsonMSFT In my opinion the struct defined here https://github.com/dotnet/runtime/blob/23695f9897ed5d0b4004db91a3f3835fc26b7a3c/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/ComVariant.cs#L39-L56 , where the first two bytes of the lower word of the flags field in decimal data type should be untouched which would be the higher word for big endian platforms. ref:-
https://github.com/dotnet/runtime/blob/dc9c9d5f9e23ff6aa0b99af8f6c410a0558fadaf/src/libraries/System.Private.CoreLib/src/System/Decimal.cs#L116-L123

and I did find a similar fix here where where if defs are used in the variant layout wrt to the endianness
https://github.com/dotnet/runtime/blob/23695f9897ed5d0b4004db91a3f3835fc26b7a3c/src/coreclr/pal/inc/rt/palrt.h#L433-L450

jkotas: The palrt.h version has just wReserved1 and vt swapped. This one has all wReserved1, wReserved2, wReserved3 and vt swapped. Which one is correct?

saitama951: @jkotas Since this fix as well passes on the s390x, probably it's better to follow the `palrt.h` version. I was pushing the necessary changes.

AaronRobinsonMSFT: > @jkotas Since this fix as well passes on the s390x,

If this is true, then we have some missing test cases. Swapping all four vs just swapping the first two should be profoundly difference. The use of a very small number like 42 to do all the testing is likely insufficient to actually validate all that much. We need to stop being clever with this 42 nonsense and actually perform validation based on the valid domain.

> probably it's better to follow the palrt.h version

I accept this even though I don't at the moment fully understand why this is correct in this case. I've filed https://github.com/dotnet/runtime/issues/94521 to fill out this matrix.

uweigand: > > @jkotas Since this fix as well passes on the s390x,
> 
> If this is true, then we have some missing test cases. Swapping all four vs just swapping the first two should be profoundly different. The use of a very small number like 42 to do all the testing is likely insufficient to actually validate all that much. We need to stop being clever with this 42 nonsense and actually perform validation based on the valid domain.
> 
> > probably it's better to follow the palrt.h version
> 
> I accept this even though I don't at the moment fully understand why this is correct. I've filed #94521 to fill out this matrix.

The in-memory layout of `Decimal` is:
```
        private readonly int _flags;
        private readonly uint _hi32;
        private readonly ulong _lo64;
```
with the stipulation that 
```
// Bits 0-15 (the lower word) of the flags field are
// unused and must be zero; bits 16-23 contain must contain a value between
// 0 and 28, indicating the power of 10 to divide the 96-bit integer part
// by to produce the Decimal value; bits 24-30 are unused and must be zero;
// and finally bit 31 indicates the sign of the Decimal value, 0 meaning
// positive and 1 meaning negative.
```

On a little-endian system, this means if you look at the first four bytes of a `Decimal`, you'll find zeros in byte 0 and 1, the scale factor in byte 2, and the sign-bit in byte 3.   As this is overlayed with the `TypeUnion`, this means the `_vt` field overlays bytes 0 and 1 which are guaranteed to be zero for the decimal type.  (The _wReserved1 field overlays bytes 2 and 3 of the decimal's flags fields, while _wReserved2/3 overlay the decimal's `_hi32` field.)

Now, on a big-endian system, the order is inverted, so we have the sign-bit in byte 0, the scale factor in byte 1, and zeros in byte 2 and 3.  To make the same trick as above work, we'll have to overlay the `TypeUnion._vt` field with bytes 2 and 3 now.  This is achieved by swapping `_vt` with `_wReserved1` - as is done by the `palrt.h` version and now also the most recent version of this PR.

I agree that the initial version of this PR was not correct - it placed `_vt` where `_wReserved3` is, overlapping bytes 6 and 7 of the decimal (which on a big-endian system are two least significant bytes of `_hi32`).   This may have worked for the particular test case as the value 42 is small enough that `_hi32` probably just happens to be zero for this value.  For larger decimal values the test should have failed again.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/ComVariant.cs
```diff
@@ -47,8 +47,13 @@ static unsafe ComVariant()
         [StructLayout(LayoutKind.Sequential)]
         private struct TypeUnion
         {
+#if BIGENDIAN
```
jkotas, line 50:
```suggestion
            // The layout of _wReserved1 and _vt fields needs to match Decimal._flags
#if BIGENDIAN
```

jkotas (approved):

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshalling/ComVariant.cs
AaronRobinsonMSFT: @uweigand Thanks. Yes, I spent some time last night and figured out the issue - it is as you've described. I've got some frustrations with how we've defined our variant here and Decimal in general, but I don't think we need to fix this as documenting this is likely sufficient for the small number of people this will impact.

AaronRobinsonMSFT (approved):

saitama951:
The failures seem to be unrelated. Can this be merged?

PR #94454, opened 2023-11-07
[RISC-V] Implement SOS related Debugger API code.

viewizard:
All changes in runtime we need for diagnostic's SOS commands that use ICorDebug (`clrstack -i`, `clrstack -i -a`, ...) to work.

CC @clamp03 @wscho77 @HJLeee @JongHeonChoi @t-mustafin @gbalykov


clamp03 (approved):

src/coreclr/inc/cordebug.idl
```diff
@@ -3988,14 +3988,14 @@ interface ICorDebugRegisterSet : IUnknown
         REGISTER_LOONGARCH64_F31,
 
         REGISTER_RISCV64_PC = 0,
-        REGISTER_RISCV64_RA,
         REGISTER_RISCV64_SP,
+        REGISTER_RISCV64_FP,
```
jkotas, line 3992: Why is this reordering needed?

viewizard: In case of register indexes, we have "fixed" indexes for PS, SP and FP:
https://github.com/dotnet/runtime/blob/173eb1fb769edd42c856a83ce972ba0a63b425ca/src/coreclr/pal/prebuilt/inc/cordebug.h#L9476-L9478
in the same time, code also count on this and don't use arch specific registers by unique names for PS, SP and FP. For example:
https://github.com/dotnet/runtime/blob/173eb1fb769edd42c856a83ce972ba0a63b425ca/src/coreclr/debug/di/rsthread.cpp#L6135-L6143

gbalykov (approved):

src/coreclr/inc/cordebug.idl
```diff
@@ -3988,14 +3988,14 @@ interface ICorDebugRegisterSet : IUnknown
...
+        REGISTER_RISCV64_RA,
         REGISTER_RISCV64_GP,
```
jkotas, line 3994: There is `REGISTER_RISCV64_X0` with TODO at the end of this list that does not appear to be used anywhere. Delete it while you are on it?

viewizard: I am still not sure about `REGISTER_RISCV64_X0`, plus, I see it used here for some purpose:
https://github.com/dotnet/runtime/blob/23695f9897ed5d0b4004db91a3f3835fc26b7a3c/src/coreclr/debug/inc/riscv64/primitives.h#L52

Note, current implementation aimed and tested only with diagnostics SOS command with simple tests, that is extremely limited and I belive not cover all cases of this code usage. So, I leave it as is for now.

src/coreclr/debug/inc/dbgipcevents.h
```diff
@@ -1171,6 +1171,41 @@ struct MSLAYOUT DebuggerREGDISPLAY
     SIZE_T  S7;
     SIZE_T  S8;
     SIZE_T  PC;
+#elif defined(TARGET_RISCV64)
+    #define DebuggerIPCE_FloatCount 32
+    SIZE_T  R0;
```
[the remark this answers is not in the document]
viewizard: Looks like I miss this one, when copy registers from register context. Sure, no reason store R0 for frames at stack trace.

src/coreclr/inc/cordebug.idl
jkotas: I think `REGISTER_RISCV64_X0` tries to be the R0 register that is not a real register, so it does not need to be in the enum.

viewizard: @clamp03 could you please clarify, is X0 here used in the R0 meaning?

viewizard: `ICorDebugInfo::RegNum` have `REGNUM_R0` https://github.com/dotnet/runtime/blob/0ad3723332c37f6423a725fbd4ad889b8e0f3d7c/src/coreclr/inc/cordebuginfo.h#L183
and `REGISTER_RISCV64_X0` used just as "padding" for proper mapping with `g_JITToCorDbgReg`
https://github.com/dotnet/runtime/blob/23695f9897ed5d0b4004db91a3f3835fc26b7a3c/src/coreclr/debug/inc/riscv64/primitives.h#L155-L165

I am not really strong in this code, do we really could remove R0 from `ICorDebugInfo::RegNum` ? Is not  `ICorDebugInfo::RegNum` used as indexes for "normal" (not debug related) register context?

jkotas: This padding is an internal implementation detail. It should not leak through the public ICorDebug surface.

Arm64 has similar wzr register and we do not have a constant for it in the ICorDebug register enum either.

viewizard: Thank you for explanation, I got the idea. Will prepare appropriate changes.

src/coreclr/debug/inc/riscv64/primitives.h
```diff
@@ -49,7 +49,7 @@ inline CORDB_ADDRESS GetPatchEndAddr(CORDB_ADDRESS patchAddr)
 
 constexpr CorDebugRegister g_JITToCorDbgReg[] =
 {
-    REGISTER_RISCV64_X0, // TODO-RISCV64-CQ: Add X0 to access DbgReg in correct order
+    (CorDebugRegister)(-1), // R0 is not using, but we need padding here for proper mapping with ICorDebugInfo::RegNum.
```
jkotas, line 52:
```suggestion
    (CorDebugRegister)(-1), // X0 is zero register that is not a real register. We need padding here for proper mapping with ICorDebugInfo::RegNum.
```

src/coreclr/debug/inc/riscv64/primitives.h
```diff
@@ -49,7 +49,7 @@ inline CORDB_ADDRESS GetPatchEndAddr(CORDB_ADDRESS patchAddr)
...
     REGISTER_RISCV64_RA,
```
jkotas, line 53: Also, could you please update the assert in `ConvertRegNumToCorDebugRegister` from `_ASSERTE(reg >= 0);` to `_ASSERTE(reg > 0);`.
`

jkotas (approved):
LGTM modulo feedback

src/coreclr/debug/inc/riscv64/primitives.h
viewizard: Sure, will do.

clamp03: Noti: Build fails on clang16
Error message
```
  /home/clamp/runtime/src/coreclr/debug/inc/riscv64/primitives.h:52:5: error: integer value -1 is outside the valid range of values [0, 255] for this enumeration type [-Wenum-constexpr-conversion]
      (CorDebugRegister)(-1), // X0 is zero register that is not a real register. We need padding here for proper mapping with ICorDebugInfo::RegNum.
      ^
```

viewizard: Thanks! This probably mean we should use something like `(CorDebugRegister)(255)` instead here. Initially, the sense of `-1` usage was "out of range for sure". Will fix this in next PR.

Issue #85911, opened 2023-05-08
[API Proposal]: update API signatures to leverage ref readonly parameters


Sergio0694:
### Background and motivation

> **Note**: supersedes https://github.com/dotnet/runtime/issues/73608.

This issue tracks updating the signatures of all methods that can benefit from either the new `ref readonly` parameters (see https://github.com/dotnet/csharplang/issues/6010), as well as the additional flexibility from the compiler allowing changes in ref-ness for method parameters without it being a breaking change (also see speclet at https://github.com/dotnet/csharplang/pull/7165). Needs Roslyn changes first.

### API Proposal

```diff
namespace System
{
    public static class Nullable
    {
-       public static ref readonly T GetValueRefOrDefaultRef<T>(in T? nullable) where T : struct;
+       public static ref readonly T GetValueRefOrDefaultRef<T>(ref readonly T? nullable) where T : struct;
    }

    public readonly ref struct ReadOnlySpan<T>
    {
-       public ReadOnlySpan(in T reference);
+       public ReadOnlySpan(ref readonly T reference);
    }
}

namespace System.Runtime.CompilerServices
{
    public static class Unsafe
    {
-       public static bool AreSame<T>(ref T left, ref T right);
+       public static bool AreSame<T>(ref readonly T left, ref readonly T right);
-       public static ref T AsRef<T>(in T source);
+       public static ref T AsRef<T>(ref readonly T source);
-       public static nint ByteOffset<T>(ref T origin, ref T target);
+       public static nint ByteOffset<T>(ref readonly T origin, ref readonly T target);
-       public static void Copy<T>(void* destination, ref T source);
+       public static void Copy<T>(void* destination, ref readonly T source);
-       public static void CopyBlock(ref byte destination, ref byte source, uint byteCount);
+       public static void CopyBlock(ref byte destination, ref readonly byte source, uint byteCount);
-       public static void CopyBlockUnaligned(ref byte destination, ref byte source, uint byteCount);
+       public static void CopyBlockUnaligned(ref byte destination, ref readonly byte source, uint byteCount);
-       public static bool IsAddressGreaterThan<T>(ref T left, ref T right);
+       public static bool IsAddressGreaterThan<T>(ref readonly T left, ref readonly T right);
-       public static bool IsAddressLessThan<T>(ref T left, ref T right);
+       public static bool IsAddressLessThan<T>(ref readonly T left, ref readonly T right);
-       public static bool IsNullRef<T>(ref T source);
[...]

Issue #94317, opened 2023-11-02
Use of `ref readonly` instead of `in` in libraries is a breaking change for F# callers


Tarmil:
### Description

With the switch from `in` to `ref readonly` in `Unsafe.AsRef` (#89736), since as I understand it F# doesn't support `ref readonly`, it becomes impossible to take an unsafe mutable reference to an immutable record field.

### Reproduction Steps

```fsharp
type MyRecord =
    { Value : int }

    member this.SetValue(v: int) =
        (Unsafe.AsRef<int> &this.Value) <- v
```

### Expected behavior

The code compiles successfully, and calling `SetValue()` mutates the field.

### Actual behavior

The code fails to compile:

```
Error FS0041 : No overloads match for method 'AsRef'.

Known type of argument: inref<'T>

Available overloads:
 - Unsafe.AsRef<'T>(source: byref<'T>) : byref<'T> // Argument 'source' doesn't match
 - Unsafe.AsRef<'T>(source: voidptr) : byref<'T> // Argument 'source' doesn't match
```

### Regression?

This is a regression from .NET 7.

### Configuration

Using .NET 8 RC2 and the associated F# compiler.

Sergio0694:
If F# doesn't support `ref readonly`, wouldn't just every single API using it be broken there, unless the language is updated?

Tarmil:
Yes, I considered posting to F# too, and I probably will. I do find it odd though that only C# was taken into consideration to determine whether this would be a breaking change.

vzarytovskii:
> If F# doesn't support `ref readonly`, wouldn't just every single API using it be broken there, unless the language is updated?
> 
> 
> 
> In other words: isn't this an issue that should be in the F# repo? The API is working as intended.

We don't explicitly support it, the fix would probably be an easy one, however I don't understand the change in *existing* API.

Isn't it considered breaking in C#?

Sergio0694:
No, it's explicitly not a breaking change in C# 12, as we introduced compatibility rules for `in`/`ref`/`ref readonly` for this.

The goal was explicitly to allow "fixing" the signatures of existing APIs such as this, without it being a breaking change.
Which is the case, at least for C# only.

Szer:
> No, it's explicitly not a breaking change in C# 12, as we introduced compatibility rules for `in`/`ref`/`ref readonly` for this.

BCL is not "for C#" though, but "for .NET" (written in C# yes)

[...]

PR #94470, opened 2023-11-07
Update Marshal.QueryInterface() argument modifier (take 2)

Sergio0694:
> [!NOTE]
> This reverts commit a60b66da4a1f1087ea5504599c960da8c1c7636e (from #91983)

Follow up from a conversation with @AaronRobinsonMSFT offline. Now that .NET 8 is done and we're early in the .NET 9 cycle (and most importantly, all tooling is already updated to correctly handle `ref readonly`), we can attempt again to change `Marshal.QueryInterface` to use `in` for the `Guid` parameter (as approved in #85911). I've also updated most `Marshal.QueryInterface` callsites through libraries and tests to leverage the new `in` parameter modifier.

F# also just got a fix to handle `ref readonly` method parameters: https://github.com/dotnet/fsharp/pull/16232.

cc. @tannergooding @vzarytovskii

### DIFF
--- a/src/libraries/System.Data.OleDb/src/OleDbComWrappers.cs
+++ b/src/libraries/System.Data.OleDb/src/OleDbComWrappers.cs
@@ -37,8 +37,7 @@ namespace System.Data.OleDb
         {
             Debug.Assert(flags == CreateObjectFlags.UniqueInstance);
 
-            Guid errorInfoIID = IID_IErrorInfo;
-            int hr = Marshal.QueryInterface(externalComObject, ref errorInfoIID, out IntPtr comObject);
+            int hr = Marshal.QueryInterface(externalComObject, IID_IErrorInfo, out IntPtr comObject);
             if (hr == S_OK)
             {
                 return new ErrorInfoWrapper(comObject);
--- a/src/libraries/System.Data.OleDb/src/OleDbComWrappers.cs
+++ b/src/libraries/System.Data.OleDb/src/OleDbComWrappers.cs
@@ -37,7 +37,10 @@ namespace System.Data.OleDb
         {
             Debug.Assert(flags == CreateObjectFlags.UniqueInstance);
 
-            int hr = Marshal.QueryInterface(externalComObject, IID_IErrorInfo, out IntPtr comObject);
+            Guid errorInfoIID = IID_IErrorInfo;
+#pragma warning disable CS9191 // The 'ref' modifier for argument 1 corresponding to 'in' parameter is equivalent to 'in'. Consider using 'in' instead.
+            int hr = Marshal.QueryInterface(externalComObject, ref errorInfoIID, out IntPtr comObject);
+#pragma warning restore CS9191
             if (hr == S_OK)
             {
                 return new ErrorInfoWrapper(comObject);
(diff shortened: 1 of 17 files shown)


AaronRobinsonMSFT:
We also need to loop with any interop source generator teams about this change. Specifically CsWinRT and CsWin32 - @manodasanW and @AArnott.

src/libraries/System.Data.OleDb/src/OleDbComWrappers.cs
```diff
@@ -37,8 +37,7 @@ protected override object CreateObject(IntPtr externalComObject, CreateObjectFla
         {
             Debug.Assert(flags == CreateObjectFlags.UniqueInstance);
 
-            Guid errorInfoIID = IID_IErrorInfo;
-            int hr = Marshal.QueryInterface(externalComObject, ref errorInfoIID, out IntPtr comObject);
+            int hr = Marshal.QueryInterface(externalComObject, IID_IErrorInfo, out IntPtr comObject);
```
tannergooding, line 40: Looks like this one is cross compiled for netstandard2.0, and so we need to continue using `ref` and to suppress the warning. It could be coupled with `AsRef` to avoid the copy downlevel (given we know it won't be mutated)

Sergio0694:
Alright so to recap:
- F# is fixed and support should drop in .NET 8.0.1 already (https://github.com/dotnet/runtime/issues/94317#issuecomment-1808892865)
- For CsWinRT (cc. @manodasanW) and CsWin32 (cc. AArnott) there's two approaches based on context:
  - For multi-targeted sources, they can just use `ref`, and suppress the warning on .NET 8 (CS9191). Even without suppressing it, it'll just be a warning though, and not an actual build error. Of course, suppressing it is recommended, but just saying, it should never fail to built outright in this scenario.
  - For source-generated files, they can either also just use `ref` and suppress the warning again, or detect the signature of the method, pass it down to the incremental model and use that to decide what modifier to emit. Either is doable, of course the latter is a tiny bit more involved.

@AaronRobinsonMSFT does this sound acceptable? And do we need any specific other action here before merging? I could also open tracking issues for this in the other repos if it helped, so it doesn't risk getting lost?

AaronRobinsonMSFT:
@Sergio0694 It looks like CsWin32 created some mitigation for addressing https://github.com/microsoft/CsWin32/issues/1014. Therefore, I think CsWin32 is good. Please confirm the fix for that issue makes sense. We should also have a tracking issue in CsWinRT and reference this issue.

Sergio0694:
Ooh that's awesome, I had missed that one in CsWin32! Yeah the fix seems completely fine for this change as well. Using `in` doesn't require an explicit modifier at the callsite, but just like with C# 11 and lower, explicitly using `in` is perfectly fine and doesn't result in any warning. So if they already have that split for .NET 8 and greater, changing `ref readonly` to `in` for them should not result in any difference with respect to compile errors/warnings.

I'll open a tracking issue in CsWinRT as well 👍

Sergio0694:
Opened a tracking issue for CsWinRT: https://github.com/microsoft/CsWinRT/issues/1388.

AaronRobinsonMSFT (approved):
I'm good with this. Thanks for trying again.

@tannergooding Any more thoughts on this?

tannergooding (approved):

tannergooding:
@AaronRobinsonMSFT, LGTM. Feel free to merge whenever you think its ready.

Issue #76047, opened 2022-09-23
Improve Vector128.ExtractMostSignificantBits for arm64


EgorBo:
_Per discussion with @tannergooding on Discord_

`Vector128.ExtractMostSignificantBits` is quite an important API that is typically used together with comparisons and `TrailingZeroCount/LeadingZeroCount` to detect positions of an element in a vector - typically used in various `IndexOf`-like algorithms, etc. Example:

```csharp
static void PrintPostion()
{
    Vector128<byte> src = Vector128.Create((byte)0, 0, 0, 42, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0);
    Vector128<byte> val = Vector128.Create((byte)42);

    // prints 3 as the index of 42 is "3" in src vector
    Console.WriteLine(FirstMatch(src, val)); 
}

static int FirstMatch(Vector128<byte> src, Vector128<byte> val)
{
    Vector128<byte> eq = Vector128.Equals(src, val);
    return BitOperations.TrailingZeroCount(eq.ExtractMostSignificantBits());
}
```
Codegen for `FirstMatch` on x64:
```asm
       vmovupd  xmm0, xmmword ptr [rcx]
       vpcmpeqb xmm0, xmm0, xmmword ptr [rdx]
       vpmovmskb eax, xmm0
       tzcnt    eax, eax
```
Codegen for `FirstMatch` on arm64:
```asm
            cmeq    v16.16b, v0.16b, v1.16b
            ldr     q17, [@RWD00]
            and     v16.16b, v16.16b, v17.16b
            ldr     q17, [@RWD16]
            ushl    v16.16b, v16.16b, v17.16b
            movi    v17.4s, #0
            ext     v17.16b, v16.16b, v17.16b, #8
            addv    b17, v17.8b
            umov    w0, v17.b[0]
            lsl     w0, w0, #8
            addv    b16, v16.8b
            umov    w1, v16.b[0]
            orr     w0, w0, w1
            rbit    w0, w0
            clz     w0, w0
```
Because arm64 doesn't have a direct equivalent of `movmsk`. However, this particular case can be optimized because we know that input of `ExtractMostSignificantBits` is a comparison's result with all elements being either zero or all-bits-set, in the best case we can perform this smart trick from [arm blog](https://community.arm.com/arm-community-blogs/b/infrastructure-solutions-blog/posts/porting-x86-vector-bitmask-optimizations-to-arm-neon) by @danlark1
```asm
[...]

PR #94472, opened 2023-11-07
Introduce VectorXXX.GetMatchCount and VectorXXX.IndexOfFirstMatch helper methods to ensure optimal codegen for ARM64 and unrolled fallback paths

neon-sunset:
This PR is work in progress: pending tests and updating existing SIMD code to use the helpers where applicable.

The code in methods is ordered in such a way to produce optimal codegen for both ARM64 and SSE2/4.2 when executed on modern CPUs (i.e. NativeAOT default target) because both benefit from code written with V256 vectors which use helpers like these where necessary to achieve full ILP utilization (both x86 and ARM64 *including LITTLE cores* are usually able to dispatch 2-4 SIMD operations per cycle).

This change would not be possible without new physical promotion but I'm really happy that it's a thing now and we can now trade blows with the likes of LLVM.


src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Vector128.cs
```diff
@@ -3032,6 +3032,37 @@ internal static void SetUpperUnsafe<T>(in this Vector128<T> vector, Vector64<T>
             Unsafe.AsRef(in vector._upper) = value;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        [CompExactlyDependsOn(typeof(AdvSimd.Arm64))]
+        internal static int CountMatches<T>(this Vector128<T> vector)
```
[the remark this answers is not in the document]
neon-sunset: > Given that these are perf critical APIs where the JIT can do a better thing for inputs in a particular shape, they should likely be JIT intrinsics so we get the best performance on ARm64.

Yes but that's not trivial and way above my skill level 😅 There is, however, a tracking issue for recognizing and optimizing movemask shape for ARM64: https://github.com/dotnet/runtime/issues/76047 But since there is more than one shape (with risk of recognition being fragile), I think this change is in the right direction. Also, writing movemask expecting the JIT to correctly rewrite it given set conditions for ARM64 may seem a bit counterintuitive.

> We also know we want such APIs, so we should really be taking these to API review so the general shape can be vetted first. For example, IndexOfMatch isn't sufficient and needs to be IndexOfFirstMatch and IndexOfLastMatch. Likewise CountMatches should likely be GetMatchCount() to be more inline with other similar Get*Count() names we have throughout the BCL.

Yes, my plan was to introduce internal helpers first, change existing SIMD paths to use them, and then create a proposal which can showcase profit/need in having these methods. Noted on naming - will fix.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Vector128.cs
```diff
@@ -3032,6 +3032,37 @@ internal static void SetUpperUnsafe<T>(in this Vector128<T> vector, Vector64<T>
...
+                return AdvSimd.Arm64
+                    .AddAcross(AdvSimd.PopCount(vector.AsByte()))
+                    .ToScalar() / (8 * Unsafe.SizeOf<T>());
+            }
+
+            return BitOperations.PopCount(vector.ExtractMostSignificantBits());
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        [CompExactlyDependsOn(typeof(AdvSimd.Arm64))]
```
EgorBo, line 3050: You don't need CompExactlyDependsOn here, because you already check IsSupported explicitly

neon-sunset: Ah, I see, thanks.

neon-sunset:
@tannergooding I've noticed there already exists `IndexOfLastMatch()` on `ISimdVector` so I submitted a proposal to add remaining methods: https://github.com/dotnet/runtime/issues/94639
While it is being reviewed, I'll continue with this PR - will move around the code if necessary in the future.

src/libraries/System.Private.CoreLib/src/System/SpanHelpers.T.cs
```diff
@@ -3828,7 +3825,7 @@ internal static int LastIndexOfAnyExceptInRangeUnsignedNumber<T>(ref T searchSpa
                     ref T oneVectorAwayFromEnd = ref Unsafe.Subtract(ref end, Vector128<T>.Count);
                     do
                     {
-                        count += BitOperations.PopCount(Vector128.Equals(Vector128.LoadUnsafe(ref current), targetVector).ExtractMostSignificantBits());
+                        count += Vector128.Equals(Vector128.LoadUnsafe(ref current), targetVector).GetMatchCount();
```
neon-sunset, line 3828: This should especially help ARM64. Previously, this was doing movemask emulation only to load the scalar value back into a SIMD register to do popcount.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Vector512.cs
```diff
@@ -2950,6 +2950,84 @@ public static Vector512<T> WithUpper<T>(this Vector512<T> vector, Vector256<T> v
...
+            Vector128<T> vec1 = vector._lower._upper;
+            Vector128<T> vec2 = vector._upper._lower;
+            Vector128<T> vec3 = vector._upper._upper;
+
+            int cnt0 = Vector128.GetMatchCount(vec0);
+            int cnt1 = Vector128.GetMatchCount(vec1);
+            int cnt2 = Vector128.GetMatchCount(vec2);
+            int cnt3 = Vector128.GetMatchCount(vec3);
+
+            return cnt0 + cnt1 + cnt2 + cnt3;
```
MichalPetryka, line 2981: The `Vector512` fallbacks should probably not have separate V128 versions, just V256 ones.

neon-sunset:
Test failures are related - will fix.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Vector512.cs
neon-sunset: Let me check if the 128b fallback codegen is good enough.

PR #94489, opened 2023-11-07
ElapsedEventArgs: public constructor + sealed

### LINKED ISSUE #31204: ElapsedEventArgs doesn't have a public constructor
System.Timers.ElapsedEventArgs doesn't have a public constructor, despite just being a data class containing one property.  This makes it unnecessarily awkward to write a mock for something that uses an ElapsedEventHandler, because you need to use a workaround just to create this one object.

It's impossible to solve this by subclassing ElapsedEventArgs, because the SignalTime property is read-only and therefore cannot be set by your constructor.

TheMaximum:
Changed the internal ElapsedEventArgs constructor to be public, sealed the class, and renamed constructor argument to match property name.

Fix #31204

### DIFF
--- a/src/libraries/System.ComponentModel.TypeConverter/src/System/Timers/ElapsedEventArgs.cs
+++ b/src/libraries/System.ComponentModel.TypeConverter/src/System/Timers/ElapsedEventArgs.cs
@@ -3,13 +3,22 @@
 
 namespace System.Timers
 {
-    public class ElapsedEventArgs : EventArgs
+    /// <summary>
+    /// Provides data for the <see cref='System.Timers.Timer.Elapsed'/> event.
+    /// </summary>
+    public sealed class ElapsedEventArgs : EventArgs
     {
-        internal ElapsedEventArgs(DateTime localTime)
+        /// <summary>
+        /// Initializes a new instance of the <see cref='System.Timers.ElapsedEventArgs'/> class.
+        /// </summary>
+        public ElapsedEventArgs(DateTime signalTime)
         {
-            SignalTime = localTime;
+            SignalTime = signalTime;
         }
 
+        /// <summary>
+        /// Gets the time when the timer elapsed.
+        /// </summary>
         public DateTime SignalTime { get; }
     }
 }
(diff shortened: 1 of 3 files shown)


src/libraries/System.ComponentModel.TypeConverter/src/System/Timers/ElapsedEventArgs.cs
```diff
@@ -3,13 +3,22 @@
...
+    /// <summary>
+    /// Provides data for the <see cref='System.Timers.Timer.Elapsed'/> event.
+    /// </summary>
+    public sealed class ElapsedEventArgs : EventArgs
     {
-        internal ElapsedEventArgs(DateTime localTime)
+        /// <summary>
+        /// Initializes a new instance of the <see cref='System.Timers.ElapsedEventArgs'/> class.
+        /// </summary>
+        public ElapsedEventArgs(DateTime signalTime)
```
vcsjones, line 14: The parameter `signalTime` should be documented with a `<param name="signalTime">blah</param>` doc comment.

TheMaximum: I'd be happy to add it to the comment. I left it out because when checking other EventArgs derived types in the solution, none seemed to have the constructor arguments in the comments (though I might have missed something there?).

vcsjones: > left it out because when checking other EventArgs derived types in the solution

The runtime repo itself is not an accurate representation of documentation. All of the other EventArgs are fully documented, they just aren’t documented with XML comments. They are documented in the documentation repository themselves.

https://github.com/dotnet/dotnet-api-docs/blob/6dc61adf39521293ae251486f03a02d62fdd4dfa/xml/System.ComponentModel/PropertyChangedEventArgs.xml#L110-L113

However, new APIs in the repository are fed in to the documentation now.

So they are documented, you just don’t see it in code always for historical purposes, but going forward it is always done in code

TheMaximum: Ah, I was unaware of that. I've added XML comment documentation for `signalTime` :)

buyaa-n (approved):
LGTM, thanks @TheMaximum @vcsjones!

PR #94497, opened 2023-11-07
JIT: stop computing dom start nodes for morph RPO

AndyAyersMS:
Remove the up-front computation of enter blocks and dom
start nodes from the DFS computations used for RPO.

Also lay the groundwork for removing unreachable blocks, by tracking
the postorder number of the last reachable block.

### DIFF
--- a/src/coreclr/jit/compiler.h
+++ b/src/coreclr/jit/compiler.h
@@ -5330,13 +5330,14 @@ protected:
     BasicBlock* fgIntersectDom(BasicBlock* a, BasicBlock* b); // Intersect two immediate dominator sets.
 
     void fgDfsReversePostorder();
+    unsigned fgDfsReversePostorderCore(BlockSet_ValArg_T startBlocks);
     void fgDfsReversePostorderHelper(BasicBlock* block,
                                      BlockSet&   visited,
                                      unsigned&   preorderIndex,
                                      unsigned&   reversePostorderIndex);
 
-    BlockSet_ValRet_T fgDomFindStartNodes(); // Computes which basic blocks don't have incoming edges in the flow graph.
-                                             // Returns this as a set.
+    BlockSet_ValRet_T fgDomFindStartBlocks(); // Computes which basic blocks don't have incoming edges in the flow graph.
+                                              // Returns this as a set.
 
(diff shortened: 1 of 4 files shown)


AndyAyersMS:
@jakobbotsch PTAL
cc @dotnet/jit-contrib 

No diffs, modest (~0.05%) TP win (enabling RPO cost about 0.25%, so this gets back a chunk of that).

jakobbotsch (approved):
It looks fine to me, but why not improve `fgDfsReversePostorder` for everyone? It confuses me a little that morph wants different behavior from everyone else.

AndyAyersMS:
> It looks fine to me, but why not improve `fgDfsReversePostorder` for everyone? It confuses me a little that morph wants different behavior from everyone else.

We use this for synthesis, reachability, and dominance. The first two probably can change over. 

For dominance it may make sense to try and pick additional start blocks that are clearly dom tree roots&mdash;but let me see what happens if we don't.

AndyAyersMS:
Yeah, that makes sense. Let me see what not finding the "optimal" start point for the unvisited blocks does for dominance... hopefully it does not matter.

It would also be nice if we could use bit vectors (or their complements) for worklists, then we wouldn't have to search the entire block list for unvisited blocks, but I don't think we're set up to do that quite yet. We'd also need a map from bbNum to Block*.

AndyAyersMS:
@jakobbotsch looks like indeed none of it was needed, and now this more or less erases the TP cost from the RPO.

Enter blocks is still used in a few places (reachability and dominance) so we now have a couple similar looking computations between that, the "inlined" version here in the DFS, and another one in the dead blocks code -- it might be nice to figure out how to have just one that can meet all needs. But I'll save that for another day.

AndyAyersMS:
Still something broken with a few arm tests

No [Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=464203&view=ms.vss-build-web.run-extensions-tab) otherwise, and a TP win.

CI compiler seems to have been updated.

jakobbotsch (approved):
LGTM. Seems like there's an arm32 failure.

Looks like there is a single diff in x86. Presumably that's because we may start the DFS from handler blocks in a different order than before.

PR #94499, opened 2023-11-08
Add RIDs for Tizen8.0

JongHeonChoi:


ViktorHofer:
Wait, that shouldn't be required anymore in .NET 8 and .NET 9 as we now have portable RID graph. cc @tmds @agocke

tmds:
Eventually the goal is to remove this (tracked by https://github.com/dotnet/runtime/issues/90000).

I don't understand why `Microsoft.NETCore.App.Runtime.sfxproj` opts into the older graph here:

https://github.com/dotnet/runtime/blob/173eb1fb769edd42c856a83ce972ba0a63b425ca/src/installer/pkg/sfx/Microsoft.NETCore.App/Microsoft.NETCore.App.Runtime.sfxproj#L15

but I assume that's a good reason to keep updating this.

For Linux, these updates happen automagically during source-build.

ViktorHofer:
> but I assume that's a good reason to keep updating this.

Are we sure we need these Tizen RID updates just for shared framework deps.json file? Would it make sense to require the portable RID graph (which already is the default) when interacting with Tizen RIDs? cc @jkotas @agocke 

From the other thread (cc @elinor-fung):

> We don't have a specific release set for removing the switch. I think .NET 9 is a bit too soon though - my thought had been having at least a release of warning that the switch would go away. But yes, once the switch is removed, the non-portable graph could go way.

IMO we should start adding a warning in .NET 9 and then remove the switch with .NET 10. Otherwise we will never be able to stop updating and removing the legacy RID graph.

tmds:
> Are we sure we need these Tizen RID updates just for shared framework deps.json file? Would it make sense to require the portable RID graph (which already is the default) when interacting with Tizen RIDs when targeting .NET 8 / .NET 9? cc @jkotas @agocke

If this property is just for packing the old graph. Then I fully agree: we don't need to add any more rids and consider the legacy graph unsupported for platforms not yet included.

ViktorHofer:
@vitek-karas mentioned [here](https://github.com/dotnet/runtime/pull/92211#issuecomment-1725868346), that the Microsoft.NETCore.App.deps.json includes the full (legacy) RID graph to make that opt-out switch work.

vitek-karas:
The idea was that we do need to keep the big RID graph in `Microsft.NETCore.App.deps.json` to support the legacy behavior, but that we would not keep updating that graph - it would stay at .NET 8 level.

ViktorHofer:
Based on that, I'm closing this PR.

PR #94503, opened 2023-11-08
Add `-repeatCount` in superpmi.exe to repeat method compilation

BruceForstall:
Passing `-repeatCount N` for some integer `N > 0` will compile each function `N` times. This is true for each function in the MCH file, by default, or each one specified by a `-c` argument or .mcl file.

This can be used to help investigate the JIT throughput for compilation of a particular function or set of functions.

This is based on #93417.

### DIFF
--- a/src/coreclr/tools/superpmi/superpmi/superpmi.cpp
+++ b/src/coreclr/tools/superpmi/superpmi/superpmi.cpp
@@ -327,8 +327,17 @@ int __cdecl main(int argc, char* argv[])
         }
     }
 
+    // This doesn't change in the work loop below, so compute it once.
+    double totalWork = reader->TotalWork() * o.repeatCount;
+
     while (true)
     {
+        // Read the progress before doing the method context read. This is because when we output progress below,
+        // we output a sum of this plus the per-repeatCount progress for the currently read MC.
+        double progressBeforeCurrentMC = reader->Progress() * o.repeatCount;
+
+        // Read the Method Context data from the file.
+
         MethodContextBuffer mcb = reader->GetNextMethodContext();
         if (mcb.Error())
         {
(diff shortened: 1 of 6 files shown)


BruceForstall:
This is based on the work @kunalspathak investigated in #93417. I submitted it as a separate PR because it ended up so different.

The most annoying part here was getting the progress report to make sense for the normal case as well as the `-c 1 -repeatCount 5000` case.

Percent complete should actually be a little more accurate for the case of specified indexes without TOC, and TOC without specified indexes.

BruceForstall:
@kunalspathak PTAL
cc @dotnet/jit-contrib

src/coreclr/tools/superpmi/superpmi/superpmi.cpp
```diff
@@ -380,334 +372,367 @@ int __cdecl main(int argc, char* argv[])
...
-            jit = JitInstance::InitJit(o.nameOfJit, o.breakOnAssert, &stInitJit, mc, o.forceJitOptions, o.jitOptions);
+        // Compile this method context as many times as we have been asked to (default is once).
+        for (int iter = 0; iter < o.repeatCount; iter++)
+        {
             if (jit == nullptr)
             {
-                // InitJit already printed a failure message
-                return (int)SpmiResult::JitFailedToInit;
-            }
+                SimpleTimer stInitJit;
```
BruceForstall, line 383: Note that most of the changes starting here are because everything in the `iter` loop was indented 4 spaces.

src/coreclr/tools/superpmi/superpmi/superpmi.cpp
```diff
@@ -380,334 +372,367 @@ int __cdecl main(int argc, char* argv[])
...
 
-            if ((res.Result == ReplayResult::Miss) || (res2.Result == ReplayResult::Miss))
-            {
-                missingCount++;
-            }
-
-            if (o.details != nullptr)
+            // Write a progress message if necessary.
+            const int progressInterval = 500;
+            if ((jittedCount > 0) && (jittedCount % progressInterval == 0))
```
BruceForstall, line 708: The progress output was moved to the end of the loop because it made all the progress output scenarios work better.

kunalspathak (approved):
> The most annoying part here was getting the progress report to make sense for the normal case as well as the -c 1 -repeatCount 5000 case.

I think thats where I spent my energy to get it right., but your approach is more reasonable. Thanks!

JulieLeeMSFT:
> Percent complete should actually be a little more accurate for the case of specified indexes without TOC, and TOC without specified indexes.

What is TOC? 
BTW, this feature will be useful to study contribution of specific methods during app startup.

kunalspathak:
> What is TOC?

Table of contents. We have a `.mct` file that basically contains index of various method contexts for easy reading the method contexts from `.mch` files.

PR #94512, opened 2023-11-08
[RISC-V] Add branch label offset to disasm

tomeksowi:
Label offset is easier to look for than raw instruction offset.

Sample disassembly with this PR:
```asm
  000FC863  blt            t6, zero, 16         ;; offset=0x00AC
  01000513  addi           a0, zero, 16
  02A42223  sw             a0, 36(fp)
  00C0006F  jal            zero, 12
                                                ;; size=48 bbWeight=1 PerfScore 0.00
G_M14667_IG07:  ;; offset=0x00AC
  00F00513  addi           a0, zero, 15
```
Part of #84834
cc @wscho77 @HJLeee @clamp03 @JongHeonChoi @t-mustafin @gbalykov @viewizard @ashaurtaev @sirntar @yurai007 @Bajtazar

### DIFF
--- a/src/coreclr/jit/emitriscv64.cpp
+++ b/src/coreclr/jit/emitriscv64.cpp
@@ -2901,7 +2901,7 @@ static const char* const RegNames[] =
 //    id   - The instrDesc of the code if needed.
 //
 // Note:
-//    The length of the instruction's name include aligned space is 13.
+//    The length of the instruction's name include aligned space is 15.
 //
 
 void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
@@ -3223,7 +3223,7 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
                     return;
             }
         }
-        case 0x63:
+        case 0x63: // BRANCH
         {
             unsigned int opcode2 = (code >> 12) & 0x7;
             const char*  rs1     = RegNames[(code >> 15) & 0x1f];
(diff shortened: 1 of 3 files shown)


src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3236,28 +3236,42 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
+            static const int MAX_LEN = 32;
+            int len = printf("           %s, %s, %d", rs1, rs2, offset);
+            if (len <= 0 || len > MAX_LEN)
+                return;
+
+            if (!emitComp->opts.disDiffable)
+            {
+                printf("%*s;; offset=0x%04X\n", MAX_LEN - len, "", emitCurCodeOffs(insAdr) + offset);
+            }
+            else
```
yurai007, line 3271: nit: Since we always end up with '\n' maybe we could get rid of else like: 
```
if (!emitComp->opts.disDiffable) {
    printf("%*s;; offset=0x%04X", MAX_LEN - len, "", emitCurCodeOffs(insAdr) + offset);
}
printf("\n");
```

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3236,28 +3236,38 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
+                    printf("bltu");
+                    break;
+                case 7:
+                    printf("bgeu");
+                    break;
                 default:
                     printf("RISCV64 illegal instruction: 0x%08X\n", code);
                     return;
             }
+            static const int MAX_LEN = 32;
```
Bajtazar, line 3261: nit: I would use `static constexpr int` instead

yurai007 (approved):
LGTM

src/coreclr/jit/emitriscv64.cpp
tomeksowi: for simple stuff it's the same thing really

Bajtazar: Ok

Bajtazar (approved):

JongHeonChoi (approved):

clamp03 (approved):

jakobbotsch (approved):
Seems better than before, but our supported backends display the target labels for branches, so it would be ideal if RISC-V/LA64 did the same.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3236,28 +3236,38 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
+
+            int len = printf("           %s, %s, %d", rs1, rs2, offset);
+            if (len <= 0 || len > MAX_LEN)
+                return;
```
gbalykov, line 3265: Do we need to print "\n" in this case too?

tomeksowi: It's a safeguard so it's unlikely to happen, but still. I'll include a fix in the next PR.

tomeksowi:
> Seems better than before, but our supported backends display the target labels for branches, so it would be ideal if RISC-V/LA64 did the same.

That was my initial intention but `instrDesc` passed to this function didn't have a target `insGroup` in `idAddr()` so I went for the next best thing for the time being. We will need to do it like ARM or x86 eventually.

PR #94515, opened 2023-11-08
Implement frozen object heap

MichalStrehovsky:
When allocating a RuntimeType instances, we were creating an object on the pinned object heap, creating a handle to it, and purposefully leaked the handle. The RuntimeTypes live forever. This fragments the pinned object heap. So instead of doing that, port frozen object heap from CoreCLR. This is a line-by-line port. Frozen object heap is a segmented bump memory allocator that interacts with the GC to tell it the boundaries of the segments.

cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/nativeaot/Runtime/MiscHelpers.cpp
+++ b/src/coreclr/nativeaot/Runtime/MiscHelpers.cpp
@@ -356,12 +356,17 @@ EXTERN_C NATIVEAOT_API int32_t __cdecl RhpGetCurrentThreadStackTrace(void* pOutp
     return RhpCalculateStackTraceWorker(pOutputBuffer, outputBufferLength, pAddressInCurrentFrame);
 }
 
-COOP_PINVOKE_HELPER(void*, RhpRegisterFrozenSegment, (void* pSegmentStart, size_t length))
+COOP_PINVOKE_HELPER(void*, RhRegisterFrozenSegment, (void * pSection, size_t allocSize, size_t commitSize, size_t reservedSize))
 {
-    return RedhawkGCInterface::RegisterFrozenSegment(pSegmentStart, length);
+    return RedhawkGCInterface::RegisterFrozenSegment(pSection, allocSize, commitSize, reservedSize);
 }
 
-COOP_PINVOKE_HELPER(void, RhpUnregisterFrozenSegment, (void* pSegmentHandle))
+COOP_PINVOKE_HELPER(void, RhUpdateFrozenSegment, (void* pSegmentHandle, uint8_t* allocated, uint8_t* committed))
+{
+    RedhawkGCInterface::UpdateFrozenSegment((GcSegmentHandle)pSegmentHandle, allocated, committed);
+}
+
+COOP_PINVOKE_HELPER(void, RhUnregisterFrozenSegment, (void* pSegmentHandle))
 {
     RedhawkGCInterface::UnregisterFrozenSegment((GcSegmentHandle)pSegmentHandle);
 }
--- a/src/coreclr/nativeaot/Runtime/MiscHelpers.cpp
+++ b/src/coreclr/nativeaot/Runtime/MiscHelpers.cpp
@@ -356,17 +356,17 @@ EXTERN_C NATIVEAOT_API int32_t __cdecl RhpGetCurrentThreadStackTrace(void* pOutp
     return RhpCalculateStackTraceWorker(pOutputBuffer, outputBufferLength, pAddressInCurrentFrame);
 }
 
-COOP_PINVOKE_HELPER(void*, RhRegisterFrozenSegment, (void * pSection, size_t allocSize, size_t commitSize, size_t reservedSize))
+EXTERN_C NATIVEAOT_API void* __cdecl RhRegisterFrozenSegment(void * pSection, size_t allocSize, size_t commitSize, size_t reservedSize)
 {
     return RedhawkGCInterface::RegisterFrozenSegment(pSection, allocSize, commitSize, reservedSize);
 }
 
-COOP_PINVOKE_HELPER(void, RhUpdateFrozenSegment, (void* pSegmentHandle, uint8_t* allocated, uint8_t* committed))
+EXTERN_C NATIVEAOT_API void __cdecl RhUpdateFrozenSegment(void* pSegmentHandle, uint8_t* allocated, uint8_t* committed)
 {
     RedhawkGCInterface::UpdateFrozenSegment((GcSegmentHandle)pSegmentHandle, allocated, committed);
 }
 
-COOP_PINVOKE_HELPER(void, RhUnregisterFrozenSegment, (void* pSegmentHandle))
+EXTERN_C NATIVEAOT_API void __cdecl RhUnregisterFrozenSegment(void* pSegmentHandle)
 {
     RedhawkGCInterface::UnregisterFrozenSegment((GcSegmentHandle)pSegmentHandle);
 }
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.Unix.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.Unix.cs
@@ -1,7 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using Microsoft.Win32.SafeHandles;
+using System.Diagnostics;
 
 namespace Internal.Runtime
 {
@@ -15,7 +15,7 @@ namespace Internal.Runtime
                 size,
                 Interop.Sys.MemoryMappedProtections.PROT_NONE,
                 Interop.Sys.MemoryMappedFlags.MAP_PRIVATE | Interop.Sys.MemoryMappedFlags.MAP_ANONYMOUS,
-                new SafeFileHandle(-1, false),
+                -1,
                 0);
 
         }
@@ -32,6 +32,7 @@ namespace Internal.Runtime
 
         static void ClrVirtualFree(void* pBase, nuint size)
         {
+            Debug.Assert(size != 0);
             Interop.Sys.MUnmap((nint)pBase, size);
         }
     }
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.Unix.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.Unix.cs
@@ -1,0 +1,38 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Microsoft.Win32.SafeHandles;
+
+namespace Internal.Runtime
+{
+    internal unsafe partial class FrozenObjectHeapManager
+    {
+        static void* ClrVirtualReserve(nuint size)
+        {
+            // The shim will return null for failure
+            return (void*)Interop.Sys.MMap(
+                0,
+                size,
+                Interop.Sys.MemoryMappedProtections.PROT_NONE,
+                Interop.Sys.MemoryMappedFlags.MAP_PRIVATE | Interop.Sys.MemoryMappedFlags.MAP_ANONYMOUS,
+                new SafeFileHandle(-1, false),
+                0);
+
+        }
+
+        static void* ClrVirtualCommit(void* pBase, nuint size)
+        {
+            int result = Interop.Sys.MProtect(
+                (nint)pBase,
+                size,
+                Interop.Sys.MemoryMappedProtections.PROT_READ | Interop.Sys.MemoryMappedProtections.PROT_WRITE);
+
+            return result == 0 ? pBase : null;
+        }
+
+        static void ClrVirtualFree(void* pBase, nuint size)
+        {
+            Interop.Sys.MUnmap((nint)pBase, size);
+        }
+    }
+}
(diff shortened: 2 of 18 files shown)


EgorBo:
Cool! Will it make sense to have a single managed impl and call that from CoreCLR runtime? 🙂

EgorBo:
Also, I presume we won't benefit in codegen from objects allocated dynamically on FOH in NativeAOT so, I guess, you're only improving fragmentation in GC slightly for RuntimeType objects, right?

src/coreclr/nativeaot/Runtime/MiscHelpers.cpp
```diff
@@ -356,12 +356,17 @@ EXTERN_C NATIVEAOT_API int32_t __cdecl RhpGetCurrentThreadStackTrace(void* pOutp
     return RhpCalculateStackTraceWorker(pOutputBuffer, outputBufferLength, pAddressInCurrentFrame);
 }
 
-COOP_PINVOKE_HELPER(void*, RhpRegisterFrozenSegment, (void* pSegmentStart, size_t length))
+COOP_PINVOKE_HELPER(void*, RhRegisterFrozenSegment, (void * pSection, size_t allocSize, size_t commitSize, size_t reservedSize))
```
jkotas, line 359: `RhRegisterFrozenSegment` is a regular PInvoke called in preeemptive mode. It should not be defined using `COOP_PINVOKE_HELPER` macro.

jkotas: (same for other frozen segment methods)

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.cs
```diff
@@ -0,0 +1,269 @@
...
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+using Debug = System.Diagnostics.Debug;
+
+namespace Internal.Runtime
+{
+    internal unsafe partial class FrozenObjectHeapManager
+    {
+        public static FrozenObjectHeapManager Instance = new FrozenObjectHeapManager();
```
jkotas, line 16:
```suggestion
        public static readonly FrozenObjectHeapManager Instance = new FrozenObjectHeapManager();
```

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.cs
```diff
@@ -0,0 +1,269 @@
...
+                    if (m_CurrentSegment != null)
+                    {
+                        // Double the reserved size to reduce the number of frozen segments in apps with lots of frozen objects
+                        // Use the same size in case if prevSegmentSize*2 operation overflows.
+                        nuint prevSegmentSize = m_CurrentSegment.m_Size;
+                        newSegmentSize = Math.Max(prevSegmentSize, prevSegmentSize * 2);
+                    }
+
+                    m_CurrentSegment = new FrozenObjectSegment(newSegmentSize);
+                    m_FrozenSegments.Add(m_CurrentSegment);
```
jkotas, line 83: If `m_FrozenSegments.Add` fails due to OOM, we will be in inconsistent state.

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.Unix.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        static void* ClrVirtualReserve(nuint size)
+        {
+            // The shim will return null for failure
+            return (void*)Interop.Sys.MMap(
+                0,
+                size,
+                Interop.Sys.MemoryMappedProtections.PROT_NONE,
+                Interop.Sys.MemoryMappedFlags.MAP_PRIVATE | Interop.Sys.MemoryMappedFlags.MAP_ANONYMOUS,
+                new SafeFileHandle(-1, false),
```
jkotas, line 18: Add MMap import that just takes IntPtr as the handle?

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.cs
```diff
@@ -0,0 +1,269 @@
...
+            // Pointer to the end of the current segment, ready to be used as a pMT for a new object
+            // meaning that "m_pCurrent - sizeof(ObjHeader)" is the actual start of the new object (header).
+            //
+            // m_pCurrent <= m_SizeCommitted
+            public byte* m_pCurrent;
+
+            // Last known value of m_pCurrent that GC is aware of.
+            //
+            // m_pCurrentRegistered <= m_pCurrent
+            byte* m_pCurrentRegistered;
```
jkotas, line 138:
```suggestion
            private byte* m_pCurrentRegistered;
```

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.cs
```diff
@@ -0,0 +1,269 @@
...
+
+            // Memory committed in the current segment
+            //
+            // m_SizeCommitted <= m_pStart + FOH_SIZE_RESERVED
+            public nuint m_SizeCommitted;
+
+            // Total memory reserved for the current segment
+            public nuint m_Size;
+
+            IntPtr m_SegmentHandle;
```
jkotas, line 148:
```suggestion
            private IntPtr m_SegmentHandle;
```

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.cs
```diff
@@ -0,0 +1,269 @@
...
+
+namespace Internal.Runtime
+{
+    internal unsafe partial class FrozenObjectHeapManager
+    {
+        public static FrozenObjectHeapManager Instance = new FrozenObjectHeapManager();
+
+        private readonly LowLevelLock m_Crst = new LowLevelLock();
+        private readonly LowLevelLock m_SegmentRegistrationCrst = new LowLevelLock();
+        private ArrayBuilder<FrozenObjectSegment> m_FrozenSegments;
```
jkotas, line 20: This is unused in the current implementation.

VSadov:
I wonder if this can be better solved if GC had an API to allocate **immortal objects**. 

Like in: [GC_ALLOC_IMMORTAL ](https://github.com/VSadov/Satori/blob/eecbf6535eadc8cda8a8f1df1d1fc98d77635ffe/src/coreclr/gc/gcinterface.h#L1054)flag

Immortal objects are just objects that GC allocates in a region that will never be swept or compacted. Otherwise it is a real managed object. It is in GC heap range and can be covered by cardtable/barriers, thus even allowing writing managed references (they will be rooted obviously).

This should not be confused with External objects - where someone formats random memory to have objects in it. Those cannot be truly "managed" - these are what NativeAOT had for a long time.  
But Immortal objects are managed objects in any possible way, except that they are "forever leaked"

Also unlike External objects, immortal allocations can easily have a public API attached to it, implemented on Mono, etc...

I do not think we should keep reimplementing something that GC could do better.
I can write up a more detailed proposal if there is interest.

EgorBo:
> I wonder if this can be better solved if GC had an API to allocate **immortal objects**.
> 
> Like in: [GC_ALLOC_IMMORTAL ](https://github.com/VSadov/Satori/blob/eecbf6535eadc8cda8a8f1df1d1fc98d77635ffe/src/coreclr/gc/gcinterface.h#L1054)flag
> 
> Immortal objects are just objects that GC allocates in a region that will never be swept or compacted. Otherwise it is a real managed object. It is in GC heap range and can be covered by cardtable/barriers and participate in marking/updating, thus even allowing writing managed references.
> 
> This should not be confused with External objects - where someone formats random memory to have objects in it. Those cannot be truly "managed" - these are what NativeAOT had for a long time. But Immortal objects are managed objects in any possible way, except that they are "forever leaked"
> 
> Also unlike External objects, immortal allocations can easily have a public API attached to it, implemented on Mono, etc...
> 
> I do not think we should keep reimplementing something that GC could do better. I can write up a more detailed proposal if there is interest.

What exactly that would solve? We'll have to mark-n-sweep that new kind of immortal heap during gen1 collect and won't be able to omit write barriers for better codegen, right?

> Also unlike External objects, immortal allocations can easily have a public API attached to it, implemented on Mono, etc...

I noticed some interest from community in immortal objects mainly to avoid spending time in GC in those, e.g. https://github.com/dotnet/runtime/discussions/94411 (and have more examples on social networks) and/or mmap huge graphs of objects from file on start - but that is not trivial for a public API, right.

VSadov:
> What exactly that would solve? 

A uniform way of doing this via well defined API. 
In cases where GC implementation is shared, the implementation sharing as an additional benefit.

> We'll have to mark-n-sweep that new kind of immortal heap during gen1 collect and won't be able to omit write barriers for better codegen, right?

There is no sweeping, since nothing dies.

Marking/cardtable/barriers stuff only needs to be involved if reference fields are allowed. That is not a necessary feature, just a possibility. 
It would be a pay-for-play though. Immortal objects would be Gen2 or higher. In ephemeral collections you'd only visit locations dirtied in the card table. If there were no writes that rooted Gen1/0 objects, you'd have no reason to look at the immortal segment.

> PS: as it turns out, new heaps are expensive to add in terms of diagnostics work 😕

This assumes that ad-hoc immortal allocators do not need diagnostic support or that adding a new heap is the only way to implement.

In the absence of references the "implementation" could as well be "put the CLR implementation behind GC interface". 
You would still benefit from sharing everything with NativeAOT, and with other users if this becomes public API.

jkotas:
We had a long discussion about whether management and reporting of frozen heaps should be owned by the GC or kept external to the GC. The preference of the GC team was to make theme external - see the ASCII art at the top of https://github.com/dotnet/diagnostics/issues/4156 .

I agree with you that letting GC manage frozen heaps would be a viable alternative that can be extended further to allow frozen objects with references to non-frozen ones, but it is not where we landed.

VSadov:
Oh, well. 

Just to be sure - the suggestion is not about External objects. Or what is "Non-GC heap" in the ASCII art. 
Things that were never allocated via GC interfaces would still live there. Supporting pre-allocated objects like NativeAOT strings and statics would need that. 

This is mostly about an API that allows to allocate a forever object at run time without fear of fragmenting anything. There is clearly a need for this as this is the second time we implement this functionality.

EgorBo:
> This is mostly about an API that allows to allocate a forever object at run time without fear of fragmenting anything. There is clearly a need for this as this is the second time we implement this functionality.

But doesn't it mean that we'll need both anyway and the only advantage we can have from immortal ones that we're less constrained on what we can allocate there? (although, we still can't put stuff from collectible assemblies etc) - I wonder what exactly we'd be able to allocate that we currently can not?

jkotas:
> This is mostly about an API that allows to allocate a forever object at run time without fear of fragmenting anything. 

I understand. FWIW, the heap design discussion was centered around the needs of regular CoreCLR that only allocates these at runtime.

> I wonder what exactly we'd be able to allocate there that we currently can not?

I think the primary point of the discussion is what makes most sense as the overall system architecture. We can do all the things discussed here today. It may not have the 100% ideal perf characteristics, but perf is something that can be always fine-tuned based on data if the system was architected correctly.

> A public API for it has a questionable UX I guess - only for arrays?

Yes, it is same set of considerations as why we have `GC.AllocateArray(....)` and we have not added `GC.AllocateObject` (yet).

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.cs
```diff
@@ -0,0 +1,269 @@
...
+using Debug = System.Diagnostics.Debug;
+
+namespace Internal.Runtime
+{
+    internal unsafe partial class FrozenObjectHeapManager
+    {
+        public static FrozenObjectHeapManager Instance = new FrozenObjectHeapManager();
+
+        private readonly LowLevelLock m_Crst = new LowLevelLock();
+        private readonly LowLevelLock m_SegmentRegistrationCrst = new LowLevelLock();
```
jkotas, line 19: This complicated locking scheme is not needed for managed implementation. The complicated locking scheme is needed for the object enumeration diagnostic APIs. It is unlikely we would be able to implement the object enumeration diagnostic API using this managed implementation.

I understand why you started by trying to match the C++ impl in CoreCLR, but I am not convinced that it is helps at the end. It makes things more complicated than they need to be.

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.cs
```diff
@@ -0,0 +1,211 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+using Debug = System.Diagnostics.Debug;
+
```
jkotas, line 10:
```suggestion
// Rewrite of src\coreclr\vm\frozenobjectheap.cpp in C#

```

jkotas (approved):

src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/FrozenObjectHeapManager.cs
```diff
@@ -0,0 +1,212 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+using Debug = System.Diagnostics.Debug;
+// Rewrite of src\coreclr\vm\frozenobjectheap.cpp in C#
```
jkotas, line 10:
```suggestion
using Debug = System.Diagnostics.Debug;

// Rewrite of src\coreclr\vm\frozenobjectheap.cpp in C#
```

MichalStrehovsky:
The x64 Alpine leg is failing with the same failure mode we saw in https://github.com/dotnet/runtime/pull/94405#issuecomment-1804255092. Filed an issue on that. It's unrelated to either of those PRs and must exist in baseline.

PR #94528, opened 2023-11-08
Combine rootDir with relative file names

### LINKED ISSUE #50648: InMemoryDirectoryInfo is unexpectedly dependent on working directory due to its use of GetFullPath.
From the `runtime-coreclr libraries-jitstress` pipeline:

https://dev.azure.com/dnceng/public/_build/results?buildId=1069450&view=ms.vss-test-web.build-test-results-tab

This test fails all Windows arm64 runs (including JITMinOpts=1), with:
```
COMPlus_JITMinOpts=1
COMPlus_TieredCompilation=0
```

```
D:\h\w\BAA10A61\w\A0300974\e>call RunTests.cmd --runtime-path D:\h\w\BAA10A61\p 
----- start Fri 04/02/2021  1:01:32.06 ===============  To repro directly: ===================================================== 
pushd D:\h\w\BAA10A61\w\A0300974\e\
"D:\h\w\BAA10A61\p\dotnet.exe" exec --runtimeconfig Microsoft.Extensions.FileSystemGlobbing.Tests.runtimeconfig.json --depsfile Microsoft.Extensions.FileSystemGlobbing.Tests.deps.json xunit.console.dll Microsoft.Extensions.FileSystemGlobbing.Tests.dll -xml testResults.xml -nologo -nocolor -notrait category=IgnoreForCI -notrait category=OuterLoop -notrait category=failing 
popd
===========================================================================================================

D:\h\w\BAA10A61\w\A0300974\e>"D:\h\w\BAA10A61\p\dotnet.exe" exec --runtimeconfig Microsoft.Extensions.FileSystemGlobbing.Tests.runtimeconfig.json --depsfile Microsoft.Extensions.FileSystemGlobbing.Tests.deps.json xunit.console.dll Microsoft.Extensions.FileSystemGlobbing.Tests.dll -xml testResults.xml -nologo -nocolor -notrait category=IgnoreForCI -notrait category=OuterLoop -notrait category=failing  
  Discovering: Microsoft.Extensions.FileSystemGlobbing.Tests (method display = ClassAndMethod, method display options = None)
  Discovered:  Microsoft.Extensions.FileSystemGlobbing.Tests (found 97 test cases)
  Starting:    Microsoft.Extensions.FileSystemGlobbing.Tests (parallel test collections = on, max threads = 8)
    Microsoft.Extensions.FileSystemGlobbing.Tests.FunctionalTests.VerifyInMemoryDirectoryInfo_IsNotEmpty [FAIL]
      Assert.Equal() Failure
      Expected: 1
      Actual:   0
      Stack Trace:
        /_/src/libraries/Microsoft.Extensions.FileSystemGlobbing/tests/FunctionalTests.cs(858,0): at Microsoft.Extensions.FileSystemGlobbing.Tests.FunctionalTests.VerifyInMemoryDirectoryInfo_IsNotEmpty()
  Finished:    Microsoft.Extensions.FileSystemGlobbing.Tests
=== TEST EXECUTION SUMMARY ===
   Microsoft.Extensions.FileSystemGlobbing.Tests  Total: 343, Errors: 0, Failed: 1, Skipped: 0, Time: 7.984s
----- end Fri 04/02/2021  1:01:42.74 ----- exit code 1 ----------------------------------------------------------
```

jozkee:
Fixes https://github.com/dotnet/runtime/issues/50648

Using `Path.GetFullPath` on relative file names causes that the in-memory `Matcher.Match` to treat files as relative to the CWD blocking scenarios that are using a rootDir in a different drive as the one discussed in #50648. We should combine the rootDir with the relative file names instead. This is also what's described in the [documentation](https://learn.microsoft.com/dotnet/api/microsoft.extensions.filesystemglobbing.inmemorydirectoryinfo.-ctor).

> files IEnumerable\<String\>
> Collection of file names. If relative paths rootDir will be prepended to the paths.

However, this is a (massive?) breaking change as there's already a lot of tests showing that filenames are relative to CWD and not to rootDir, and in the wild, there will be most likely use cases that depend in this bug (need to check).

### DIFF
--- a/src/libraries/Microsoft.Extensions.FileSystemGlobbing/src/InMemoryDirectoryInfo.cs
+++ b/src/libraries/Microsoft.Extensions.FileSystemGlobbing/src/InMemoryDirectoryInfo.cs
@@ -46,16 +46,24 @@ namespace Microsoft.Extensions.FileSystemGlobbing
             else
             {
                 var fileList = new List<string>(files.Count());
+                string normalizedRoot = Path.GetFullPath(rootDir.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar));
 
                 // normalize
                 foreach (string file in files)
                 {
-                    fileList.Add(Path.GetFullPath(file.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar)));
+                    if (Path.IsPathRooted(file))
+                    {
+                        fileList.Add(Path.GetFullPath(file.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar)));
+                    }
+                    else
+                    {
+                        fileList.Add(Path.Combine(normalizedRoot, file.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar)));
+                    }
                 }
 
                 _files = fileList;
 
-                FullName = Path.GetFullPath(rootDir.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar));
+                FullName = normalizedRoot;
             }
         }
 
--- a/src/libraries/Microsoft.Extensions.FileSystemGlobbing/tests/FunctionalTests.cs
+++ b/src/libraries/Microsoft.Extensions.FileSystemGlobbing/tests/FunctionalTests.cs
@@ -372,12 +372,11 @@ namespace Microsoft.Extensions.FileSystemGlobbing.Tests
         public void StemCorrectWithDifferentWildCards_WithInMemory()
         {
             var matcher = new Matcher();
-            matcher.AddInclude("sub/*.cs");
-            matcher.AddInclude("**/*.cs");
+            matcher.AddInclude("src/project/sub/*.cs");
+            matcher.AddInclude("src/project/**/*.cs");
 
             var files = GetFileList();
-            var directoryPath = "src/project";
-            var results = matcher.Match(directoryPath, files);
+            var results = matcher.Match("./", files);
 
             var actual = results.Files.Select(match => match.Stem);
             var expected = new string[]
@@ -425,11 +424,10 @@ namespace Microsoft.Extensions.FileSystemGlobbing.Tests
         public void MultipleSubDirsAfterFirstWildcardMatch_HasCorrectStem_WithInMemory()
         {
             var matcher = new Matcher();
-            matcher.AddInclude("compiler/**/*.cs");
+            matcher.AddInclude("src/project/compiler/**/*.cs");
 
             var files = GetFileList();
-            var directoryPath = "src/project";
-            var results = matcher.Match(directoryPath, files);
+            var results = matcher.Match("./", files);
 
             var actual = results.Files.Select(match => match.Stem);
             var expected = new string[]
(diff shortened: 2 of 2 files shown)


Jozkee:
I found one usage of the affected overload in our first party data. It is for smoke tests on [dotnet/installer](https://github.com/dotnet/installer/blob/5037a63d97149e2b4827b4ea1cf48f4a253a017f/src/SourceBuild/content/test/Microsoft.DotNet.SourceBuild.SmokeTests/SdkContentTests.cs#L160). This leads me to believe that most people use the overloads that don’t take a rootDir, those use the cwd as rootDir inplicitly.

I’ll try to visualize the breaking change:
```cs
string rootDir = "dir1";
string[] files = ["dir1/test.0", "dir1/subdir/test.1", "dir2/test.2"];

PatternMatchingResult result = new Matcher().AddInclude("**/*").Match(rootDir, files);
Console.WriteLine(string.Join(", ", result.Files.Select(x => x.Path)));

```
Because both args are treated as relative to the cwd.
This currently prints `test.0, subdir/test.1`.
With the breaking change this now would print `dir1/test.0, dir1/subdir/test.1, dir2/test.2`. It includes dir2 because now all is under `rootDir`.

In order to get a similar (but not same) output as prior you need to change the code to something like the following:
```diff
-string rootDir = "dir1";
+string rootDir = "root";
string[] files = ["dir1/test.0", "dir1/subdir/test.1", "dir2/test.2"];

-PatternMatchingResult result = new Matcher().AddInclude("**/*").Match(rootDir, files);
+PatternMatchingResult result = new Matcher().AddInclude("dir1/**/*").Match(rootDir, files);
Console.WriteLine(string.Join(", ", result.Files.Select(x => x.Path)));

// prints
// dir1/test.0
// dir1/subdir/test.1
```

@ericstj would be good to send this to the VMR in order to validate that the dotnet/installer test doesn't break with this change?

Jozkee:
Ping @adamsitnik @jeffhandley @ericstj

jeffhandley:
Good find, @Jozkee! Is there a way to update the calling code _before the change goes through_ that would allow the calling code to function as desired both now and after the breaking change goes through? If so, we could go ahead and update dotnet/installer so that as the breaking change goes through the tests will keep passing. If not, then we might want to consider disabling that test until the change flows through and then re-enabling it.

jeffhandley (approved):

Jozkee:
@jeffhandley I can verify it, just need to find out how to run the smoke tests in dotnet/installer.

ericstj:
> With the breaking change this now would print dir1/test.0, dir1/subdir/test.1, dir2/test.2. It includes dir2 because now all is under rootDir.

How is `dir2` under `rootDir`?  I thought `rootDir` was `dir1`.

> @ericstj would be good to send this to the VMR

Not necessary - it wouldn't be running these tests (nor is it accepting commits yet).  You'd just need to watch this change flowing up to installer then fix up the test when it hits it.

Jozkee:
> How is dir2 under rootDir?

Because now `rootDir` is prefixed to each path in `files`, except when a path is absolute. In a simplified manner:
```
foreach (string file in files)
{
    if (Path.IsPathRooted(file))
        fileList.Add(file);
    else
        fileList.Add(Path.Combine(rootDir, file));
}
```

ericstj:
I see - might be better to not reuse the name - since the use of `dir1` implies the same `dir1`

Jozkee:
Breaking change doc. created: https://github.com/dotnet/docs/issues/39189

ViktorHofer:
I think this change caused another unintended breaking change. Filed: https://github.com/dotnet/docs/issues/39189

PR #94529, opened 2023-11-08
Arm64/SVE: Add Uzp1

kunalspathak:
This is similar to #93067, except that most of the code added in this PR was generated by the tool, the same tool that generated encodings in https://github.com/dotnet/runtime/pull/94285. The only thing that I realized the tool does not generate is the entries in `emitInsMayWriteToGCReg()`. I could add it, but I think it is easier to do it manually.

I have also added method placeholders for predicate registes which will need to be fixed once we add predicate register support in LSRA. I think the next step for this PR would be to make sure the encodings are correctly generated by comparing it with cordistool or lldb. That won't be possible until we add the actual Z and predicate registers.

Another TODO would be to fix the PerfScore numbers, not sure where to check that information though.

### DIFF
--- a/src/coreclr/jit/codegenarm64.cpp
+++ b/src/coreclr/jit/codegenarm64.cpp
@@ -6703,11 +6703,11 @@ void CodeGen::genArm64EmitterUnitTests()
     theEmitter->emitIns_R_R_R(INS_rorv, EA_4BYTE, REG_R8, REG_R9, REG_R10);
 
     // TODO-SVE: Fix once we add predicate registers
-    theEmitter->emitIns_R_R_R(INS_uzp1, EA_8BYTE, REG_R0, REG_R1, REG_R2,
+    theEmitter->emitIns_R_R_R(INS_sve_uzp1, EA_8BYTE, REG_R0, REG_R1, REG_R2,
                               INS_OPTS_8B); // UZP1    <Zd>.<T>, <Zn>.<T>, <Zm>.<T>
-    theEmitter->emitIns_R_R_R(INS_uzp1, EA_8BYTE, REG_R0, REG_R1, REG_R2,
+    theEmitter->emitIns_R_R_R(INS_sve_uzp1, EA_8BYTE, REG_R0, REG_R1, REG_R2,
                               INS_OPTS_8B); // UZP1    <Zd>.Q, <Zn>.Q, <Zm>.Q
-    theEmitter->emitIns_R_R_R(INS_uzp1, EA_8BYTE, REG_R0, REG_R1, REG_R2,
+    theEmitter->emitIns_R_R_R(INS_sve_uzp1, EA_8BYTE, REG_R0, REG_R1, REG_R2,
                               INS_OPTS_8B); // UZP1    <Pd>.<T>, <Pn>.<T>, <Pm>.<T>
 
 #endif // ALL_ARM64_EMITTER_UNIT_TESTS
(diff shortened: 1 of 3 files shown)


kunalspathak:
@a74nh @TIHan @amanasifkhalid @BruceForstall

src/coreclr/jit/emitarm64.cpp
```diff
@@ -10571,6 +10619,47 @@ void emitter::emitIns_Call(EmitCallType          callType,
     return ureg << 10;
 }
 
+/*****************************************************************************
+ *
+ *  Returns an encoding for the specified register used in the 'Vd' position
```
BruceForstall, line 10624:
```suggestion
 *  Returns an encoding for the specified register used in the 'Pd' position
```

src/coreclr/jit/emitarm64.cpp
```diff
@@ -10571,6 +10619,47 @@ void emitter::emitIns_Call(EmitCallType          callType,
...
+    // TODO-SVE: Fix once we add predicate registers
+    assert(emitter::isPredicateRegister(reg));
+    emitter::code_t ureg = (emitter::code_t)reg - (emitter::code_t)REG_V0;
+    assert((ureg >= 0) && (ureg <= 15));
+    return ureg;
+}
+
+/*****************************************************************************
+ *
+ *  Returns an encoding for the specified register used in the 'Vn' position
```
BruceForstall, line 10638:
```suggestion
 *  Returns an encoding for the specified register used in the 'Pn' position
```

src/coreclr/jit/emitarm64.cpp
```diff
@@ -10571,6 +10619,47 @@ void emitter::emitIns_Call(EmitCallType          callType,
...
+    // TODO-SVE: Fix once we add predicate registers
+    assert(emitter::isPredicateRegister(reg));
+    emitter::code_t ureg = (emitter::code_t)reg - (emitter::code_t)REG_V0;
+    assert((ureg >= 0) && (ureg <= 15));
+    return ureg << 5;
+}
+
+/*****************************************************************************
+ *
+ *  Returns an encoding for the specified register used in the 'Vm' position
```
BruceForstall, line 10652:
```suggestion
 *  Returns an encoding for the specified register used in the 'Pm' position
```

BruceForstall (approved):

a74nh:
I have a WIP patch which adds one SVE API entry. I'll rebase it on top on this and switch it to use UZP1.

src/coreclr/jit/codegenarm64.cpp
```diff
@@ -6702,6 +6702,14 @@ void CodeGen::genArm64EmitterUnitTests()
     theEmitter->emitIns_R_R_R(INS_asrv, EA_4BYTE, REG_R8, REG_R9, REG_R10);
     theEmitter->emitIns_R_R_R(INS_rorv, EA_4BYTE, REG_R8, REG_R9, REG_R10);
 
+    // TODO-SVE: Fix once we add predicate registers
+    theEmitter->emitIns_R_R_R(INS_uzp1, EA_8BYTE, REG_R0, REG_R1, REG_R2,
```
a74nh, line 6706: This is emitting a neon `uzp1`, it should be `INS_sve_uzp1`

kunalspathak: good catch. i will address this and fix the tool.

a74nh: With the above change, you're going to have to add `case INS_sve_uzp1` into `emitIns_R_R_R()`

kunalspathak: yes and many such cases in `emitIns*` methods.

kunalspathak:
Replaced by https://github.com/dotnet/runtime/pull/95016 and https://github.com/dotnet/runtime/pull/94765

PR #94530, opened 2023-11-08
improve debuging HTTP/2 tests

wfurt:
I bump to two issue while working on  #94112
1) I was using the SSLKEYLOG environment so I can decrypt TLS and peek at the HTTP/2 exchange. 
However, often the sessions keys were not there and I failed to decrypt the transaction. It seems like when the handshake happens and the process exist immediately not everything is written. Adding `Flush` seems to fix this.
It may come with some perf cost but this is Debug only feature so it should not matter as much. 

2) The issue would happen only when running whole test suite. I wanted to instrument the test to spit out the IPEndPoints so I can find the connection among many in the capture file. So I added `ToString` that can giver me something like
`  Http2LoopbackConnection 127.0.0.1:42543 <-> 127.0.0.1:60302`

with that I can set filer in Wireshark as `tcp.port==42543` and get

failTest

### DIFF
--- a/src/libraries/Common/tests/System/Net/Http/GenericLoopbackServer.cs
+++ b/src/libraries/Common/tests/System/Net/Http/GenericLoopbackServer.cs
@@ -94,8 +94,8 @@ namespace System.Net.Test.Common
             CloseWebSocket();
         }
 
-        public EndPoint? LocalEndPoint => _socket.LocalEndPoint;
-        public EndPoint? RemoteEndPoint => _socket.RemoteEndPoint;
+        public EndPoint? LocalEndPoint => _socket?.LocalEndPoint;
+        public EndPoint? RemoteEndPoint => _socket?.RemoteEndPoint;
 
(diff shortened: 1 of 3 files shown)


src/libraries/Common/tests/System/Net/Http/GenericLoopbackServer.cs
```diff
@@ -94,6 +94,9 @@ public void Close()
             CloseWebSocket();
         }
 
+        public EndPoint? LocalEndPoint => _socket.LocalEndPoint;
```
ManickaP, line 97: `_socket` will be null on mobile platforms, won't it?
```suggestion
        public EndPoint? LocalEndPoint => _socket?.LocalEndPoint;
```

wfurt: updated.

ManickaP (approved):

PR #94437, opened 2023-11-06
[mono] [imt] Don't increment vt_slot for non-virtual generic methods 

lambdageek:
Interfaces can have static generic methods, for example.  They don't have a `vt_slot`.

When building an IMT slot, we need to collect all the interface methods that map to a particular IMT slot along with their
implementing vtable entries.  To do that, `vt_slot` starts at the interface offset of a particular interface and keeps incrementing as we iterate over the methods of the interface.  It is crtitical that `vt_slot` is accurate - otherwise we may dispatch the interface call to the wrong virtual method.

Fixes #93770 

---

Also remove the `extra_interfaces` argument from `build_imt_slots` - it was used only for appdomain support (transparent proxy objects implemented extra interfaces).

---

Also fixup some ifdef'd debugging code.

PR #94531, opened 2023-11-08
Fix test returning 100 unconditionally

gbalykov:
Xunit error XUW1002: Tests should not unconditionally return 100. Convert to a void return.

Related change: https://github.com/dotnet/runtime/pull/94437

cc @lambdageek @clamp03 @t-mustafin


gbalykov:
@lambdageek please take a look, without this change clr tests build on main fails

lambdageek (approved):

clamp03 (approved):

PR #94533, opened 2023-11-08
JIT: Print finally clone stats with JitTimeLogCsv

amanasifkhalid:
For each method, print the number of finally handlers considered for cloning and the number of finally handlers that were cloned when logging with `DOTNET_JitTimeLogCsv`.

CC @dotnet/jit-contrib, @EgorBo PTAL. Note that if `Compiler::fgCloneFinally` doesn't try to do any cloning, both metrics ("Finally Clone Candidates" and "Finally Blocks Cloned") will be zero. Would we ever be interested in how many finally handlers there are if we didn't do any cloning?

### DIFF
--- a/src/coreclr/jit/compiler.cpp
+++ b/src/coreclr/jit/compiler.cpp
@@ -1884,7 +1884,9 @@ void Compiler::compInit(ArenaAllocator*       pAlloc,
         // Initialize all the per-method statistics gathering data structures.
         //
 
-        optLoopsCloned = 0;
+        optFinallyCount  = 0;
+        optFinallyCloned = 0;
+        optLoopsCloned   = 0;
 
 #if LOOP_HOIST_STATS
         m_loopsConsidered             = 0;
@@ -8914,6 +8916,8 @@ void JitTimer::PrintCsvHeader()
             fprintf(s_csvFile, "\"IL Bytes\",");
             fprintf(s_csvFile, "\"Basic Blocks\",");
             fprintf(s_csvFile, "\"Min Opts\",");
+            fprintf(s_csvFile, "\"Finally Clone Candidates\",");
+            fprintf(s_csvFile, "\"Finally Blocks Cloned\",");
             fprintf(s_csvFile, "\"Loops\",");
             fprintf(s_csvFile, "\"Loops Cloned\",");
 #if FEATURE_LOOP_ALIGN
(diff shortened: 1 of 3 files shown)


EgorBo:
Does it print that in JitDisasm like we discussed internally?

EgorBo (approved):

EgorBo:
LGTM assuming CI failures aren't related

amanasifkhalid:
> Does it print that in JitDisasm like we discussed internally?

No, only if `DOTNET_JitTimeLogCsv` is set (that'll also print similar stats like number of loops, loops cloned, etc.). Would you prefer we print finally blocks cloned in `JitDisasm` instead?

BruceForstall:
> Would you prefer we print finally blocks cloned in `JitDisasm` instead?

IMO, we shouldn't.

We currently print one line at the end of disasm, like:
```
Total bytes of code 48, prolog size 14, PerfScore 21.30, instruction count 18, allocated bytes for code 48 (MethodHash=95edb912) for method Program+<>c__DisplayClass0_1:.ctor():this (MinOpts) 
```

We shouldn't add printing of additional various JIT metrics here, or even on separate lines.

Though perhaps we could add an additional option like `DOTNET_JitDisasmStats` or `DOTNET_JitDisasmInfo` that would print a bunch of different stuff about the generated code immediately after/before this, e.g., number of loops, number of cloned loops, number of CSEs, etc. These could contribute to asm diffs so maybe you could more easily see what happened in the generated code without "reverse engineering" the code itself?

amanasifkhalid:
I agree with Bruce in that we shouldn't pollute the `JitDisasm` output with this.

> Though perhaps we could add an additional option like DOTNET_JitDisasmStats or DOTNET_JitDisasmInfo that would print a bunch of different stuff about the generated code immediately after/before this, e.g., number of loops, number of cloned loops, number of CSEs, etc.

@EgorBo what do you think of this idea? Then maybe we could add a flag to `superpmi.py` to enable this when running `asmdiffs`.

EgorBo:
> Though perhaps we could add an additional option like DOTNET_JitDisasmStats or DOTNET_JitDisasmInfo that would print a bunch of different stuff about the generated code immediately after/before this, e.g., number of loops, number of cloned loops, number of CSEs, etc. These could contribute to asm diffs so maybe you could more easily see what happened in the generated code without "reverse engineering" the code itself?

Agree, it is unlikely any of this info is useful outside of SPMI diffs

amanasifkhalid:
Sounds good, I'll work on adding `DOTNET_JitDisasmStats`. Do we still want this PR's addition to `JitTimeLogCsv`, or should I abandon these changes?

amanasifkhalid:
Closing this in favor of the new JIT metrics

Issue #93860, opened 2023-10-23
[Hot Reload] metadata.c:1326, condition `GINT_TO_UINT32(idx) < table_info_get_rows (t)' not met


rolfbjarne:
_From @jeromelaban on Fri, 20 Oct 2023 03:53:31 GMT_

### Steps to Reproduce

1. Create a maui app (net8 rc2)
2. Add a `[CreateNewOnMetadataUpdate]` attribute on `MainPage`
3. Build the app
4. Run the app
5. Add a `ToString();` call in the ctor, or any other `MainPage` modification
6. Perform a C# hot reload

### Expected Behavior

The code is updated without crashing.

### Actual Behavior

Sometimes:
```
error: * Assertion at /Users/runner/work/1/s/src/mono/mono/metadata/metadata.c:1326, condition `GINT_TO_UINT32(idx) < table_info_get_rows (t)' not met
```

Sometimes:
```
=================================================================
	Native Crash Reporting
=================================================================
Got a SIGSEGV while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application.
=================================================================

=================================================================
	Native stacktrace:
=================================================================
	0x104aba1d5 - /Users/jay2/Library/Developer/CoreSimulator/Devices/F4A03A05-42EF-47DF-8C0B-0DB5B0E49D76/data/Containers/Bundle/Application/FD06AA1D-DB01-4BC1-B692-B59CC36FB231/MauiApp3.app/libmonosgen-2.0.dylib : mono_dump_native_crash_info
	0x104a58b6e - /Users/jay2/Library/Developer/CoreSimulator/Devices/F4A03A05-42EF-47DF-8C0B-0DB5B0E49D76/data/Containers/Bundle/Application/FD06AA1D-DB01-4BC1-B692-B59CC36FB231/MauiApp3.app/libmonosgen-2.0.dylib : mono_handle_native_crash
	0x1049abaff - /Users/jay2/Library/Developer/CoreSimulator/Devices/F4A03A05-42EF-47DF-8C0B-0DB5B0E49D76/data/Containers/Bundle/Application/FD06AA1D-DB01-4BC1-B692-B59CC36FB231/MauiApp3.app/libmonosgen-2.0.dylib : mono_sigsegv_signal_handler_debug
	0x7ff8377405ed - /usr/lib/system/libsystem_platform.dylib : _sigtramp
	0x0 - Unknown
[...]

PR #94540, opened 2023-11-08
[mono] catch the case of updated methods in mono_debug_lookup_source_location

lambdageek:
If a newly added method is in a stack trace, look it up in the EnC debug information, if available.

Related to #93860


lambdageek:
I think I also need to update `mono_debug_lookup_source_location_by_il`, and figure out how to make a test get in there

vargaz (approved):

lambdageek:
Actually I think `mono_debug_lookup_source_location_by_il` doesn't need an update (or at least it's not obvious to me that it is useful).

It's called from 3 places:

1. From `mono_get_frame_info` if `il_offset != -1` - but as far as I can tell `il_offset` is never set to anything else.  That branch is dead code.
2. From `mono_emit_jit_dump` but that's support for profiling, which isn't something we support with hot reload.

PR #94548, opened 2023-11-09
[RISC-V] Initial patch to fix RISCV64 interpreter

JongHeonChoi:
Initial patch of interpreter for RISCV64 at HelloWorld level.

cc @wscho77 @HJLeee @clamp03 @JongHeonChoi @t-mustafin @gbalykov @viewizard @ashaurtaev @sirntar @yurai007 @Bajtazar @tomeksowi

### DIFF
--- a/src/coreclr/vm/riscv64/stubs.cpp
+++ b/src/coreclr/vm/riscv64/stubs.cpp
@@ -1211,7 +1211,7 @@ void StubLinkerCPU::EmitEpilog()
     EmitAddImm(RegSp, RegSp, GetStackFrameSize());
 
     // jalr x0, 0(ra)
-    EmitJumpRegister(1);
+    EmitJumpRegister(RegRa);
 }
 
 // Instruction types as per RISC-V Spec, Chapter 24 RV32/64G Instruction Set Listings
--- a/src/coreclr/vm/riscv64/stubs.cpp
+++ b/src/coreclr/vm/riscv64/stubs.cpp
@@ -1104,6 +1104,116 @@ void StubLinkerCPU::EmitJumpRegister(IntReg regTarget)
     Emit32(0x00000067 | (regTarget << 15));
 }
 
+void StubLinkerCPU::EmitProlog(unsigned short cIntRegArgs, unsigned short cFpRegArgs, unsigned short cbStackSpace)
+{
+    _ASSERTE(!m_fProlog);
+
+    unsigned short numberOfEntriesOnStack  = 2 + cIntRegArgs + cFpRegArgs; // 2 for fp, ra
+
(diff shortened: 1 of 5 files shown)


src/coreclr/vm/riscv64/stubs.cpp
```diff
@@ -1104,6 +1104,116 @@ void StubLinkerCPU::EmitJumpRegister(IntReg regTarget)
...
+
+    // 2. Restore FP/RA
+    EmitLoad(RegFp, RegSp, m_cbStackSpace);
+    EmitLoad(RegRa, RegSp, m_cbStackSpace + sizeof(void*));
+
+    // 1. Restore SP
+    EmitAddImm(RegSp, RegSp, GetStackFrameSize());
+
+    // jalr x0, 0(ra)
+    EmitJumpRegister(1);
```
t-mustafin, line 1214: Please use `RegRa` instead of magic number `1` if it possible here.

JongHeonChoi: Fixed.

clamp03:
@jkotas We have updated CoreCLR interpreter for RISC-V because I think the interpreter is useful for debugging in our product issue where JITC on RISC-V does not work well. However, I find it has many problems. Cannot compile on x64 and arm64 (before this PR). Cannot run helloworld from the beginning on arm64/x64/riscv64. (It can only run the helloworld main function). So I just wonder, do you have a plan to complete the interpreter? Or will you keep the interpreter as it is now?

jkotas:
We do not have any plans with the CoreCLR interpreter. We know that it is pretty broken.

clamp03 (approved):

clamp03:
Thank you for the comment. Okay. we need to use it very carefully.
Could you review this PR?

jkotas (approved):

tomeksowi:
> We have updated CoreCLR interpreter for RISC-V because I think the interpreter is useful for debugging in our product issue where JITC on RISC-V does not work well

Could you enumerate the painspots with JITC? Perhaps it could be a starting point for a to-do list of improvements.

clamp03:
We fixed many bugs in coreclr and corefx unit tests on QEMU and Sifive2 devices.
Next year, we need to start testing, debugging and optimising tizenfx tests and big real applications on tizen riscv devices, which are very unstable environment in both hardware and software. Maybe all crashes and errors would be assigned to us because all crashes and errors occur when tests or applications are running. I am so afraid of this situation. :cry:  IMO we need to prepare tools (like createdump, debugging tools and logging tools), tizen risc-v CI and so on. Fixing the CoreCLR interpreter is one of them.

PR #94550, opened 2023-11-09
[JIT] Fixes use of msvcdis

TIHan:
We have this option in the JIT, called `LATE_DISASM`, that enables use of MSVC's disassembly tool to disassemble the raw bytecode. We should enable this for our work on SVE2 so we can verify that we are emitting the right instructions.

### DIFF
--- a/src/coreclr/jit/codegen.h
+++ b/src/coreclr/jit/codegen.h
@@ -14,7 +14,11 @@
 #include "regset.h"
 #include "jitgcinfo.h"
 
+#ifdef LATE_DISASM
+class CodeGen : public CodeGenInterface
+#else // LATE_DISASM
 class CodeGen final : public CodeGenInterface
+#endif // !LATE_DISASM
 {
     friend class emitter;
     friend class DisAssembler;
@@ -33,6 +37,11 @@ public:
     // move it to Lower
     virtual bool genCreateAddrMode(
         GenTree* addr, bool fold, bool* revPtr, GenTree** rv1Ptr, GenTree** rv2Ptr, unsigned* mulPtr, ssize_t* cnsPtr);
+
+#ifdef LATE_DISASM
+    virtual const char* siStackVarName(size_t offs, size_t size, unsigned reg, unsigned stkOffs);
+    virtual const char* siRegVarName(size_t offs, size_t size, unsigned reg);
+#endif // LATE_DISASM
 
 private:
 #if defined(TARGET_XARCH)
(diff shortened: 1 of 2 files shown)


TIHan:
@dotnet/jit-contrib @kunalspathak

src/coreclr/jit/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
 include_directories("./jitstd")
 include_directories("../inc")
+include_directories("C:/work/inc")
```
kunalspathak, line 5: delete?

src/coreclr/jit/CMakeLists.txt
```diff
@@ -548,6 +553,7 @@ else()
        bcrypt.lib
        crypt32.lib
        RuntimeObject.lib
+       C:/work/lib/msvcdis140.lib
```
kunalspathak, line 556: delete?

src/coreclr/jit/compiler.h
```diff
@@ -43,6 +43,7 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 #include "tinyarray.h"
 #include "valuenum.h"
 #include "namedintrinsiclist.h"
+
```
kunalspathak, line 46: revert change

src/coreclr/jit/disasm.h
```diff
@@ -45,7 +45,7 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 #elif defined(TARGET_ARM64)
 #include "disarm64.h"
 #else // TARGET*
-#error Unsupported or unset target architecture
```
kunalspathak, line 48: why is this removed?

kunalspathak (changes requested):

src/coreclr/jit/CMakeLists.txt
TIHan: Ugh, yea this should not be here.

BruceForstall (approved):

kunalspathak (approved):
LGTM

Issue #94553, opened 2023-11-09
[API Proposal]: Generic overloads of existing TensorPrimitives methods


stephentoub:
### Background and motivation

For .NET 8, we added the new TensorPrimitives type, with methods dedicated to handling `float`.  For post-.NET 8, we're planning to augment this in three ways (https://github.com/dotnet/runtime/issues/93286):
1. Generic versions of these methods in order to handle other numerical types beyond float
2. Additional generic methods for all the operations on the generic math interfaces that aren't currently on TensorPrimitives
3. Additional generic methods to have good coverage of further relevant operations ala BLAS / LAPACK

This issue covers (1).

### API Proposal

Exactly the same signatures as on TensorPrimitives in .NET 8, with an overload that takes a `T` instead of `float`.
```C#
namespace System.Numerics.Tensors;

public static class TensorPrimitives
{
    public static void Abs<T>(ReadOnlySpan<T> x, Span<T> destination) where T : INumberBase<T>;
    public static void AddMultiply<T>(ReadOnlySpan<T> x, ReadOnlySpan<T> y, ReadOnlySpan<T> multiplier, Span<T> destination) where T : IAdditionOperators<T, T, T>, IMultiplyOperators<T, T, T>;
    public static void AddMultiply<T>(ReadOnlySpan<T> x, ReadOnlySpan<T> y, T multiplier, Span<T> destination) where T : IAdditionOperators<T, T, T>, IMultiplyOperators<T, T, T>;
    public static void AddMultiply<T>(ReadOnlySpan<T> x, T y, ReadOnlySpan<T> multiplier, Span<T> destination) where T : IAdditionOperators<T, T, T>, IMultiplyOperators<T, T, T>;
    public static void Add<T>(ReadOnlySpan<T> x, ReadOnlySpan<T> y, Span<T> destination) where T : IAdditionOperators<T, T, T>, IAdditiveIdentity<T, T>;
    public static void Add<T>(ReadOnlySpan<T> x, T y, Span<T> destination)  where T : IAdditionOperators<T, T, T>, IAdditiveIdentity<T, T>;
    public static void Cosh<T>(ReadOnlySpan<T> x, Span<T> destination) where T : IHyperbolicFunctions<T>;
    public static T CosineSimilarity<T>(ReadOnlySpan<T> x, ReadOnlySpan<T> y) where T : IRootFunctions<T>;
[...]

Issue #93217, opened 2023-10-09
[API Proposal]: Vectorized span-processing helpers


stephentoub:
This is more of a tracking issue right now than a proper API proposal.

In `TensorPrimitives`, we now have ~40 routines, most of which are vectorized across `Vector128`, `Vector256`, `Vector512`, plus a scalar fallback.  Rather than duplicate all of the boilerplate across them, we employ several helper functions that are parameterized via generic method parameters constrained to specific interfaces, with a caller supplying the kernel of the processing via a struct that implements the relevant interfaces.  For example, `TensorPrimitives.Abs` (which computes the absolute value of each input element and stores the result into the corresponding destination slot) is implemented as:
```C#
        public static void Abs(ReadOnlySpan<float> x, Span<float> destination) =>
            InvokeSpanIntoSpan<AbsoluteOperator>(x, destination);
```
and `TensorPrimitives.Add` (which adds each pair of elements and stores the result for each into the corresponding destination slot) is implemented as:
```C#
        public static void Add(ReadOnlySpan<float> x, ReadOnlySpan<float> y, Span<float> destination) =>
            InvokeSpanSpanIntoSpan<AddOperator>(x, y, destination);
```
There are helpers for various shapes of these element-wise operations, as well as helpers for performing various forms of aggregations. For example, `TensorPrimitives.ProductOfSums` (which sums each element-wise pair from the two inputs and multiplies together all of those sums) is implemented as:
```C#
        public static float ProductOfSums(ReadOnlySpan<float> x, ReadOnlySpan<float> y)
        {
            ...
            return Aggregate<AddOperator, MultiplyOperator>(x, y);
        }
```
These "Operator"s each implement various interfaces for `IUnaryOperator`, `IBinaryOperator`, etc. For example, the `AddMultipleOperator` (which implements `ITernaryOperator`) is defined as:
```C#
        readonly struct AddMultiplyOperator : ITernaryOperator
        {
            public static float Invoke(float x, float y, float z) => (x + y) * z;
            public static Vector128<float> Invoke(Vector128<float> x, Vector128<float> y, Vector128<float> z) => (x + y) * z;
            public static Vector256<float> Invoke(Vector256<float> x, Vector256<float> y, Vector256<float> z) => (x + y) * z;
            public static Vector512<float> Invoke(Vector512<float> x, Vector512<float> y, Vector512<float> z) => (x + y) * z;
        }
```

[...]

PR #94555, opened 2023-11-09
Add generic overloads to TensorPrimitives

stephentoub:
This overhauls the implementation and tests to have a generic overload for each existing float-based overload.  I've avoided touching the core logic, but have augmented the structure in a few ways, e.g. only taking vectorized code paths when the type supports vectorization.  To keep the shared definitions of the float-based APIs, on .NET 9 they delegate to shims that are implemented on top of the generic variants.

The tests have all been made instance members, with an abstract base class containing most of the tests, and calling into abstract methods for the core operations and validation routines. Derived types then fill in this logic, letting us use all the tests for both the non-generic and generic overloads.  Generic tests are validating most of the primitive types that implement the required interfaces.

This does not yet handle the following (and I plan to leave that for us to address subsequently):
- Provide generic overloads for the IndexOfMin/Max{Magnitude} methods
- Vectorize the trig-related functions for Ts other than floats

@tannergooding, these don't currently work for byte/sbyte/ushort/short because the "_Small" code for handling inputs smaller than 256/512 bit vectors was written assuming the element type was 4 bytes. What would you recommend we do?  I don't think we'll want to extend the current scheme to add all the cases necessary to handle all the possibilities here. Also, for alignment, I'm not sure if what's here still works correctly... do we have tests that exercise unaligned inputs at various alignment positions?

Contributes to https://github.com/dotnet/runtime/issues/94553


tannergooding:
> do we have tests that exercise unaligned inputs at various alignment positions?

I believe we're testing misalignment, but not all possible (or at least all interesting) misalignments.

The logic to become aligned should already be general purpose and correct. The logic to compute the number of remaining elements should be general-purpose/correct. The switch that handles the remaining elements on the vectorized path should likewise be fine already.

The switch that handles `Small` (less than a vectors worth) is not general purpose and would need to be expanded. Many modern branch predictors assume a loop will execute at least 6-7 times, so it'd probably be fine to isolate the `Small` handling to `sizeof(T) is 4 or 8`.

stephentoub:
> The switch that handles Small (less than a vectors worth) is not general purpose and would need to be expanded. Many modern branch predictors assume a loop will execute at least 6-7 times, so it'd probably be fine to isolate the Small handling to sizeof(T) is 4 or 8.

So to make sure I understand correctly, you mean basically:
```C#
if (remaining >= Vector256<T>.Count)
{
    ... // large handling that exists today
}
else if (sizeof(T) is 4 or 8)
{
    ... // small handling that exists today
}
else
{
   ... // new scalar loop
}
```
?

tannergooding:
> So to make sure I understand correctly, you mean basically:

Yeah, basically.

stephentoub:
> > So to make sure I understand correctly, you mean basically:
> 
> Yeah, basically.

For now I've just skipped the vectorization for those sizes. This affects not just the "Small" path but also the remainder handling in the long paths.

stephentoub:
@radekdoulik, can you help me figure out why this is failing on mono?  I suspect the mono vectorization support has some subtle differences.  Is it possible mono's implementation of Vector128.Abs is throwing an exception when using signed integers and encountering the min value, e.g. -128 for an sbyte?

stephentoub:
@tannergooding, can you review this? Thanks.

jeffhandley (approved):
Thanks for implementing this, Stephen. I recommend merging without the follow-up review from @tannergooding. We can conduct a post-merge review in January and address any new feedback/input.

MichalPetryka:
> Thanks for implementing this, Stephen. I recommend merging without the follow-up review from @tannergooding. We can conduct a post-merge review in January and address any new feedback/input.

Aren't the three failures here related to the changes?

stephentoub:
> > Thanks for implementing this, Stephen. I recommend merging without the follow-up review from @tannergooding. We can conduct a post-merge review in January and address any new feedback/input.
> 
> Aren't the three failures here related to the changes?

Yes, I still need to address them; that's why I haven't merged it yet.

src/libraries/System.Numerics.Tensors/src/System.Numerics.Tensors.csproj
```diff
@@ -9,16 +9,19 @@
...
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
-    <Compile Include="System\Numerics\Tensors\TensorPrimitives.netcore.cs" />
+    <Compile Include="System\Numerics\Tensors\netcore\TensorPrimitives.float.netcore.cs" />
+    <Compile Include="System\Numerics\Tensors\netcore\TensorPrimitives.T.cs" />
+    <Compile Include="System\Numerics\Tensors\netcore\TensorPrimitives.netcore.cs" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' != '.NETCoreApp'">
-    <Compile Include="System\Numerics\Tensors\TensorPrimitives.netstandard.cs" />
```
tannergooding, line 21: nit: use `Single` not `float`, it matches the naming rule conventions we use everywhere else around the type name

stephentoub:
> > > Thanks for implementing this, Stephen. I recommend merging without the follow-up review from @tannergooding. We can conduct a post-merge review in January and address any new feedback/input.
> > 
> > 
> > Aren't the three failures here related to the changes?
> 
> Yes, I still need to address them; that's why I haven't merged it yet.

The failures are https://github.com/dotnet/runtime/issues/96443. I've disabled the offending tests on mono.

stephentoub:
@tannergooding, can you take another look? Thanks.

tannergooding (approved):

asmirnov82:
Hi @stephentoub,
 
Do you have any plans to make some parts of TensorPrimitives underlying implementation public? It can be interfaces (like `IUnaryOperator<T>`, `IBinaryOperator<T>` and etc), concreate operator implementations and methods for applying operators on span of different types.

Current generic overloads of existing TensorPrimitives API are great, however they don’t cover all cases. Quite often it’s needed to do math calculation on spans of different data type. Or calculations on spans and scalar value. This functionality for example is required inside the `Microsoft.Data.Analysis.DataFrame`.

Current TensorPrimitives API for operations on Span<long> and Span<int> forces to convert Span<int> to Span<long> first and it’s extra memory allocation and coping.

In case if underlying IBinaryOperator<T> and invoke functions are public vector widening can be used for this purpose.

Do you think this functionality can be beneficial for .Net 9.0?

tannergooding:
@asmirnov82 that's covered by https://github.com/dotnet/runtime/issues/93217 and the investigation into providing an `ISimdVector` type.

PR #94563, opened 2023-11-09
JIT: Sparse conditional propagation in value numbering

### LINKED ISSUE #94559: Pinning ROS with known length causes unnecessary code generation.
Pinning a ``ReadOnlySpan<>`` where the length is constant generates code that checks the length.

This managed code
```csharp
static void Main()
{
    var data = new byte[128];
    fixed (byte* ptr = (ReadOnlySpan<byte>)data)
    {
        DoSomething(ptr);
    }
}

static void DoSomething(byte* bob) { }
```
generates this assembly on x64 for pinning (from godbolt, using the latest runtime):
```x86asm
mov      edi, 128
xor      rcx, rcx
test     edi, edi
cmovne   rcx, rax
mov      bword ptr [rsp], rcx
```

The JIT is able to figure out that the length is constant (``mov edi, 128``) which makes the rest of the code unnecessary. ``test edi, edi`` will always clear ``ZF`` because ``edi`` is constant not zero which will cause ``cmovne rcx, rax`` to always execute which will overwrite rcx that just got zerod. This whole sequence could be converted to a ``mov bword ptr [rsp], rax`` when the length can be proven to be constant.

The codegen on .NET 7 uses a jump instead of ``cmovne`` but the ``test``ing of constant is still present, .NET 6 does things differently (and worse) so this issue probably only applies to .NET 7 and newer.

jakobbotsch:
Utilize the fact that VN is able to decide whether many non-trivial branches will be taken or not taken to prove basic blocks dynamically unreachable. Take this into account when building PHIs.

Fix #94559

There are some additional improvements to make:
1. I'm not sure if we see statically unreachable blocks, but if we do, we should try to handle them similarly. Currently it seems like we pessimize PHIs if these are present.
2. We should fully skip VN'ing the blocks that are proven unreachable, but downstream phases currently are not prepared to handle that. We should maybe eliminate the proven unreachable blocks as part of VN to make sure downstream phases see a consistent view. (This is also necessary to make VN correct on its own when it is utilizing liberal VNs to prove the reachability.)
3. ~~I couldn't figure out how to look up the predecessor corresponding to a memory PHI arg, so currently we do not do this for memory PHIs.~~ Adding the pred blocks for these and making use of them has almost no diffs, so doesn't seem worth it.

Codegen of the example from #94559 before:
```asm
       mov      rcx, 0xD1FFAB1E      ; ubyte[]
       mov      edx, 128
       call     CORINFO_HELP_NEWARR_1_VC
       add      rax, 16
       mov      ecx, 128
       xor      rdx, rdx
       test     ecx, ecx
       cmovne   rdx, rax
       mov      bword ptr [rsp+0x20], rdx
       mov      rcx, rdx
       call     [Program:DoSomething(ulong)]
       xor      eax, eax
       mov      bword ptr [rsp+0x20], rax
```

Codegen after:
```asm
       mov      rcx, 0xD1FFAB1E      ; ubyte[]
       mov      edx, 128
       call     CORINFO_HELP_NEWARR_1_VC
       add      rax, 16
       mov      rcx, rax
       mov      bword ptr [rsp+0x20], rcx
       call     [Program:DoSomething(ulong)]
       xor      eax, eax
       mov      bword ptr [rsp+0x20], rax
```

### DIFF
--- a/src/coreclr/jit/valuenum.cpp
+++ b/src/coreclr/jit/valuenum.cpp
@@ -9853,7 +9853,7 @@ struct ValueNumberState
         GenTree* lastTree = predBlock->lastStmt()->GetRootNode();
         assert(lastTree->OperIs(GT_JTRUE));
 
-        GenTree* cond     = lastTree->gtGetOp1();
+        GenTree* cond = lastTree->gtGetOp1();
         // TODO-Cleanup: Using liberal VNs here is a bit questionable as it add
         // a cross-phase dependency on RBO to definitely fold this branch away.
         ValueNum normalVN = m_comp->vnStore->VNNormalValue(cond->GetVN(VNK_Liberal));
--- a/src/coreclr/jit/valuenum.cpp
+++ b/src/coreclr/jit/valuenum.cpp
@@ -9854,8 +9854,9 @@ struct ValueNumberState
         assert(lastTree->OperIs(GT_JTRUE));
 
         GenTree* cond = lastTree->gtGetOp1();
-        // TODO-Cleanup: Using liberal VNs here is a bit questionable as it add
-        // a cross-phase dependency on RBO to definitely fold this branch away.
+        // TODO-Cleanup: Using liberal VNs here is a bit questionable as it
+        // adds a cross-phase dependency on RBO to definitely fold this branch
+        // away.
         ValueNum normalVN = m_comp->vnStore->VNNormalValue(cond->GetVN(VNK_Liberal));
         if (!m_comp->vnStore->IsVNConstant(normalVN))
         {
--- a/src/coreclr/jit/valuenum.cpp
+++ b/src/coreclr/jit/valuenum.cpp
@@ -9854,7 +9854,7 @@ struct ValueNumberState
         assert(lastTree->OperIs(GT_JTRUE));
 
         GenTree* cond     = lastTree->gtGetOp1();
-        ValueNum normalVN = m_comp->vnStore->VNNormalValue(cond->GetVN(VNK_Liberal));
+        ValueNum normalVN = m_comp->vnStore->VNNormalValue(cond->GetVN(VNK_Conservative));
         if (!m_comp->vnStore->IsVNConstant(normalVN))
         {
             return true;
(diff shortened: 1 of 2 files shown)


jakobbotsch:
It would also be possible to add some of RBO's smarts in here to make it even stronger, of course at a TP cost.

jakobbotsch:

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=465237&view=ms.vss-build-web.run-extensions-tab). Mostly dominated by tests, in the real world I think the PHI disambiguation that RBO does handles most cases already. However, doing it this way in VN is naturally more precise. I want to try moving some of RBO's smarts to VN (at least the simple dominator checks).

TP regressions weren't as bad as I expected, so I didn't bother to do more work to optimize the TP. I also ended up not doing the work to switch to a precomputed RPO. I still want to do that, but as a follow-up.

src/coreclr/jit/valuenum.cpp
```diff
@@ -10021,9 +10067,22 @@ void Compiler::fgValueNumberBlock(BasicBlock* blk)
...
+            {
+                JITDUMP("  Phi arg [%06u] refers to proven unreachable pred " FMT_BB "\n", dspTreeID(phiArg),
+                        phiArg->gtPredBB->bbNum);
+
+                if ((use.GetNext() != nullptr) || (phiVNP.GetLiberal() != ValueNumStore::NoVN))
+                {
+                    continue;
+                }
+
+                JITDUMP("  ..but it looks like all preds are unreachable, so we are using it anyway\n");
```
AndyAyersMS, line 10081: Maybe add a comment that since we're quite sure this block is unreachable it's ok to use a somewhat arbitrary VN for the PHI? Can add it later if you're going to tweak nearby code again soon (for rbo say).

AndyAyersMS (approved):
Changes LGTM.

If you did switch over to the new RPO you could identify statically unreachable blocks. But maybe these are rare. At any rate rather than tolerate them here we probably should get rid of them before (or just after) we build SSA; it has identified these blocks too.

src/coreclr/jit/valuenum.cpp
jakobbotsch: Will do that in the follow-up.

jakobbotsch:
> If you did switch over to the new RPO you could identify statically unreachable blocks.

Yeah, I'll make use of that in the follow-up.

Issue #94289, opened 2023-11-02
Improve and unify debug views of immutable dictionaries


eiriktsarpalis:
> 
> >Almost forgot, but would it be possible to include this change in immutable collections as well? If you prefer to do this in a future PR that's also fine.
>
>It's on my list. I'd prefer to do it in a separate PR just to close this phase and keep scope of this PR smaller.

_Originally posted by @arturek in https://github.com/dotnet/runtime/issues/92534#issuecomment-1789512880_

PR #94581, opened 2023-11-09
Fix ReadOnlyDictionary and improve DebuggerAttributes

arturek:
* Fix debugger type proxy attributes for Key and Value collections of ReadOnlyDictionary
* Separate methods to test validity of the DebuggerTypeProxy attribute
  * ValidateDebuggerTypeProxyProperties - to verify debugger type proxy in most cases
  * CreateDebuggerTypeProxyWithNullArgument - to verify that the constructor of the debugger type proxy throws when its argument is null.
* Simplify and make more uniform debugger type proxy tests

This is the first part of work on #94289


src/libraries/System.Collections.NonGeneric/tests/ArrayListTests.cs
```diff
@@ -84,17 +84,8 @@ public static void DebuggerAttribute()
...
-            {
-                DebuggerAttributes.ValidateDebuggerTypeProxyProperties(typeof(ArrayList), null);
-            }
-            catch (TargetInvocationException ex)
-            {
-                threwNull = ex.InnerException is ArgumentNullException;
-            }
-
-            Assert.True(threwNull);
+            TargetInvocationException tie = Assert.Throws<TargetInvocationException>(() => DebuggerAttributes.CreateDebuggerTypeProxyWithNullArgument(typeof(ArrayList)));
```
eiriktsarpalis, line 87: Just trying to undertand how non-generic collections are related to this change. Is it part of an infrastructural refactoring? Is ReadOnlyDictionary used under wraps somehow?

src/libraries/System.Private.CoreLib/src/System/Collections/ObjectModel/ReadOnlyDictionary.cs
```diff
@@ -247,7 +247,7 @@ public DictionaryEntry Entry
 
         IEnumerable<TValue> IReadOnlyDictionary<TKey, TValue>.Values => Values;
 
-        [DebuggerTypeProxy(typeof(ICollectionDebugView<>))]
```
eiriktsarpalis, line 250: Can `ICollectionDebugView<>` be removed after this change?

arturek:
Hi @eiriktsarpalis, I have updated the code to fix an issue with the tests that I somehow overlooked earlier. The only failing test seems unrelated. I hope the PR can be committed now, so I can complete another one for immutable collections.

src/libraries/Common/tests/System/Collections/DebugView.Tests.cs
```diff
@@ -139,6 +139,14 @@ private static IEnumerable<object[]> TestDebuggerAttributes_ListInputs()
             yield return new object[] { new Dictionary<float, double> { { 1.0f, 1.0 }, { 2.0f, 2.0 } }.Values };
             yield return new object[] { new SortedDictionary<Guid, string> { { Guid.NewGuid(), "One" }, { Guid.NewGuid(), "Two" } }.Keys };
             yield return new object[] { new SortedDictionary<long, Guid> { { 1L, Guid.NewGuid() }, { 2L, Guid.NewGuid() } }.Values };
+#if !NETFRAMEWORK
```
eiriktsarpalis, line 142: I find it surprising that we're running collections tests against .NET framework here. @stephentoub do you know if this is intentional or could we just remove it?

stephentoub: This file is used by the immutable collection tests which run on netfx because immutable collections have a specialized netstandard/netfx build.

eiriktsarpalis (approved):
Thanks!

PR #94582, opened 2023-11-09
Provide System.Threading.Tasks.Dataflow package readme

eNeRGy164:
Fill out details for the `System.Threading.Tasks.Dataflow` NuGet package readme.

See #92228

### DIFF
--- a/src/libraries/System.Threading.Tasks.Dataflow/src/PACKAGE.md
+++ b/src/libraries/System.Threading.Tasks.Dataflow/src/PACKAGE.md
@@ -5,15 +5,13 @@
 Provides dataflow components that are collectively referred to as the *TPL Dataflow Library*.
 This dataflow model promotes actor-based programming by providing in-process message passing for coarse-grained dataflow and pipelining tasks.
 
-Note: This package is part of the larger *Task Parallel library* and is designed to help increase the robustness of concurrency-enabled applications.
-
 ## Key Features
 
 <!-- The key features of this package -->
 
 * Foundation for message passing and parallelizing CPU-intensive and I/O-intensive applications that have high throughput and low latency.
 * Provides multiple block types for various dataflow operations (e.g., `BufferBlock`, `ActionBlock`, `TransformBlock`).
-* Blocks support linking to form *networks*, allowing for the creation of complex processing topologies.
+* Dataflow blocks support linking to form *networks*, allowing you to create complex processing topologies.
 
 ## How to Use
 
@@ -23,7 +21,11 @@ This sample demonstrates a dataflow pipeline that downloads the book "The Iliad 
 
 ```csharp
 using System.Net;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks.Dataflow;
+
+var nonLetterRegex = new Regex(@"\P{L}", RegexOptions.Compiled);
+var client = new HttpClient(new HttpClientHandler { AutomaticDecompression = DecompressionMethods.GZip });
 
 // Setup blocks
 
@@ -32,7 +34,7 @@ TransformBlock<string, string> downloadString = new TransformBlock<string, strin
 {
     Console.WriteLine("Downloading '{0}'...", uri);
 
-    return await new HttpClient(new HttpClientHandler { AutomaticDecompression = DecompressionMethods.GZip }).GetStringAsync(uri);
+    return await client.GetStringAsync(uri);
 });
 
 // Separates the specified text into an array of words.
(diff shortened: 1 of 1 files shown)


ViktorHofer:
@area-system-threading-tasks PTAL

src/libraries/System.Threading.Tasks.Dataflow/src/PACKAGE.md
```diff
@@ -2,43 +2,143 @@
 
 <!-- A description of the package and where one can find more documentation -->
 
+Provides dataflow components that are collectively referred to as the *TPL Dataflow Library*.
+This dataflow model promotes actor-based programming by providing in-process message passing for coarse-grained dataflow and pipelining tasks.
 
+Note: This package is part of the larger *Task Parallel library* and is designed to help increase the robustness of concurrency-enabled applications.
```
stephentoub, line 9: I'd leave this out.  The TPL nomenclature was useful when it was a standalone experiment, but once it was incorporated into .NET Framework 4.0, having a separate name for it hurts more than it helps, IMHO.

src/libraries/System.Threading.Tasks.Dataflow/src/PACKAGE.md
```diff
@@ -2,43 +2,143 @@
...
+using System.Threading.Tasks.Dataflow;
+
+// Setup blocks
+
+// Downloads the requested resource as a string.
+TransformBlock<string, string> downloadString = new TransformBlock<string, string>(async uri =>
+{
+    Console.WriteLine("Downloading '{0}'...", uri);
+
+    return await new HttpClient(new HttpClientHandler { AutomaticDecompression = DecompressionMethods.GZip }).GetStringAsync(uri);
```
stephentoub, line 35: We shouldn't have a sample that creates a new HttpClient{Handler} for every individual message.  The HttpClient should be cached into a static readonly field as a singleton used for every message.

src/libraries/System.Threading.Tasks.Dataflow/src/PACKAGE.md
```diff
@@ -2,43 +2,143 @@
...
+});
+
+// Separates the specified text into an array of words.
+TransformBlock<string, string[]> createWordList = new TransformBlock<string, string[]>(text =>
+{
+    Console.WriteLine("Creating word list...");
+
+    // Remove common punctuation by replacing all non-letter characters with a space character.
+    char[] tokens = text.Select(c => char.IsLetter(c) ? c : ' ').ToArray();
+    text = new string(tokens);
```
stephentoub, line 45: It'd be cleaner to show this as a regex, e.g.
```C#
private static readonly Regex s_nonLetterRegex = new(@"\P{L}", RegexOptions.Compiled);
...
text = s_letterRegex.Replace(" ");
```

src/libraries/System.Threading.Tasks.Dataflow/src/PACKAGE.md
```diff
@@ -2,43 +2,143 @@
...
+TransformManyBlock<string[], string> findReversedWords = new TransformManyBlock<string[], string>(words =>
+{
+    Console.WriteLine("Finding reversed words...");
+
+    var wordsSet = new HashSet<string>(words);
+
+    return from word in words.AsParallel()
+           let reverse = new string(word.Reverse().ToArray())
+           where word != reverse && wordsSet.Contains(reverse)
+           select word;
```
stephentoub, line 72: This should not use AsParallel

src/libraries/System.Threading.Tasks.Dataflow/src/PACKAGE.md
```diff
@@ -2,43 +2,143 @@
...
+});
+
+// Removes short words and duplicates.
+TransformBlock<string[], string[]> filterWordList = new TransformBlock<string[], string[]>(words =>
+{
+    Console.WriteLine("Filtering word list...");
+
+    return words
+       .Where(word => word.Length > 3)
+       .Distinct()
```
stephentoub, line 58: Why use Distinct when findReversedWords is itself going to use a HashSet? findReversedWords can just iterate through wordsSet rather than iterating through words.

src/libraries/System.Threading.Tasks.Dataflow/src/PACKAGE.md
eNeRGy164: `static readonly` is not available in a non `Program.Main` style console app, but I put the creation of the handler outside the task.

src/libraries/System.Threading.Tasks.Dataflow/src/PACKAGE.md
eNeRGy164: `static readonly` is not available in a non `Program.Main` style console app, but I put the creation of the regex outside the task.

eNeRGy164:
@stephentoub processed the remarks. FYI, the code came from <https://learn.microsoft.com/dotnet/standard/parallel-programming/walkthrough-creating-a-dataflow-pipeline>, so maybe that should be improved as well. :)

PR #94587, opened 2023-11-09
Do not Path.Join user directory if it's "C:"

### LINKED ISSUE #68503: (Error .Net Core 5) "Directory.GetDirectories("D:")" returns not existingDirectory
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Error-Net-Core-5-DirectoryGetDirect/1696966)._

---

```
String[] dirSSS = Directory.GetDirectories("D:");
dirSSS[0];             // Output:  "D:\ref" (without quotes) <-- no this directory on disk "D:"!
Environment.Version;   // "5.0.13" (Net Core)
Environment.OSVersion; // "Microsoft Windows NT 6.3.9600.0"
                       // VS (Russian):
                       //    Microsoft Visual Studio Professional&nbsp;2019
                       //    Версия 16.11.9
```
There is another theme with this error, but it closed:
https://developercommunity2.visualstudio.com/t/DirectoryGetDirectoriesdrive-return-a/175411?entry=suggestion&space=8&preview2=true

---
#### Feedback Bot on 3/21/2022, 00:32 AM: 

(private comment, text removed)

jozkee:
Fixes https://github.com/dotnet/runtime/issues/68503

On [netfx](https://referencesource.microsoft.com/#mscorlib/system/io/path.cs,16ed6da326ce4745) we were concatenating user directories ending with volume separator `:` instead of joining with a path separator `\`. That is not the case in .NET [core]. This change brings that back as it is incorrect to join paths like "C:" with a separator in between the enumerated entries as the meaning of "relative to drive's CWD" gets lost.


src/libraries/System.Private.CoreLib/src/System/IO/Enumeration/FileSystemEntry.Windows.cs
```diff
@@ -75,5 +75,18 @@ public FileSystemInfo ToFileSystemInfo()
         /// <returns>A string representing the full path.</returns>
         public string ToFullPath() =>
             Path.Join(Directory, FileName);
+
+        private static string Join(
+            ReadOnlySpan<char> originalRootDirectory,
+            ReadOnlySpan<char> relativePath,
+            ReadOnlySpan<char> fileName)
+        {
+            if (originalRootDirectory.Length == 2 && originalRootDirectory[1] == Path.VolumeSeparatorChar)
```
adamsitnik, line 84: Is it possible for a Windows drive to have more than one letter?

adamsitnik (commented):
Overall it LGTM, I just need to make sure it's always a single letter. Thank you for fixing this bug @Jozkee !

adamsitnik (approved):
@Jozkee thank you!

src/libraries/System.Private.CoreLib/src/System/IO/Enumeration/FileSystemEntry.Windows.cs
am11: When A-Z alphabets are all assigned, user either needs to switch to UNC paths via GUIDs `\\?\Volume{GUID}` to access the next device, or mount it as directory (via Disk management). That device would not show up in explorer by default.

Jozkee:
CI errors are https://github.com/dotnet/runtime/issues/95298.

Jozkee:
Breaking change doc. issue: https://github.com/dotnet/docs/issues/39186

PR #94588, opened 2023-11-09
[release/8.0-staging] Avoid generating code for unsupported types in config source gen

tarekgh:
Backport of https://github.com/dotnet/runtime/pull/94438 to release/8.0-staging

Fixes https://github.com/dotnet/runtime/issues/94117

## Customer Impact
This issue is reported when trying to enable the configuration binder source generator in extensions repo and referencing Azure SDK libraries. 
The configuration source generator generates binding code for properties employing delegate, Func, or Action types. However, there is no means to bind configuration to such types. Consequently, the generator generates an abundance of code for all types associated with these delegates, many of which will not be utilized and may lead to code that does not compile or generate warnings. 

## Testing

The regression test was executed successfully. Additional tests were incorporated to address scenarios involving types for which we aim to refrain from generating code.

## Risk

Moderate. We are specifically handling cases for the types for which we intend to prevent code binding.

Package versioning was done in https://github.com/dotnet/runtime/pull/94307


eiriktsarpalis (approved):

tarekgh:
Approved offline

PR #82031, opened 2023-02-13
Use cinc instead of csel when possible

SwapnilGaikwad:
Use cinc instead of csel when the true and false case are constants with a difference of 1.

e.g., `result = a > 42 ? 6: 5`

This avoids an addition `mov` instr for the larger value.

Related: #55364

PR #81880, opened 2023-02-09
JIT: Transform SELECT(relop, 1/0, 0/1) to relop

jakobbotsch:
Fix #81479

Initially I generalized this to handle any power of two via `LSH(cond, shift)`. Like GCC and Clang do here: https://godbolt.org/z/hxPf3Ka8e
However, it had worse performance than a cmov from my measurements (around 20% slower). It would use one less register in many situations though.

We could also do this directly in if-conversion, though then we wouldn't get potential future SELECTs created in different places.

Example:

```csharp
[MethodImpl(MethodImplOptions.NoInlining)]
public bool IfElseCondition(bool input)
{
    if (input)
    {
        return false;
    }
    else
    {
        return true;
    }
}
```
Before:
```asm
G_M21872_IG02:              ;; offset=0000H
       84D2                 test     dl, dl
       7403                 je       SHORT G_M21872_IG05
                                                ;; size=4 bbWeight=1 PerfScore 1.25
G_M21872_IG03:              ;; offset=0004H
       33C0                 xor      eax, eax
                                                ;; size=2 bbWeight=0.50 PerfScore 0.12
G_M21872_IG04:              ;; offset=0006H
       C3                   ret
                                                ;; size=1 bbWeight=0.50 PerfScore 0.50
G_M21872_IG05:              ;; offset=0007H
       B801000000           mov      eax, 1
                                                ;; size=5 bbWeight=0.50 PerfScore 0.12
G_M21872_IG06:              ;; offset=000CH
       C3                   ret
                                                ;; size=1 bbWeight=0.50 PerfScore 0.50
```

After #81267:
```asm
G_M21872_IG02:              ;; offset=0000H
       B801000000           mov      eax, 1
       33C9                 xor      ecx, ecx
       84D2                 test     dl, dl
       0F45C1               cmovne   eax, ecx
                                                ;; size=12 bbWeight=1 PerfScore 1.00
G_M21872_IG03:              ;; offset=000CH
       C3                   ret
```

After #81267 + this PR:
```asm
G_M21872_IG02:              ;; offset=0000H
       33C0                 xor      eax, eax
       84D2                 test     dl, dl
       0F94C0               sete     al
                                                ;; size=7 bbWeight=1 PerfScore 1.50
G_M21872_IG03:              ;; offset=0007H
       C3                   ret
```

PR #94592, opened 2023-11-10
JIT: SELECT(COND, 11, 10) -> ADD(COND, 10)

EgorBo:
```cs
int Test(int x) => x == 42 ? 1 : 2;
```
```diff
; Method Benchmarks:Test(int):int:this (FullOpts)
-      mov      eax, 1
-      mov      ecx, 2
-      cmp      edx, 42
-      cmovne   eax, ecx
+      xor      eax, eax
+      cmp      edx, 42
+      setne    al
+      inc      eax
       ret      
-; Total bytes of code: 17
+; Total bytes of code: 11
```

We can do the same for `SELECT(COND, X, X+1)` if this pattern is popular.

### DIFF
--- a/src/coreclr/jit/ifconversion.cpp
+++ b/src/coreclr/jit/ifconversion.cpp
@@ -713,6 +713,7 @@ bool OptIfConversionDsc::optIfConvert()
     //
     // SELECT(COND, 10, 11) -> ADD(COND, 10)
     //
+    // Conservatively give up on relops with NaNs (we won't be able to reverse them).
     if (selectTrueInput->IsCnsIntOrI() && selectFalseInput->IsCnsIntOrI() &&
         ((m_cond->gtFlags & GTF_RELOP_NAN_UN) == 0))
     {
@@ -726,7 +727,7 @@ bool OptIfConversionDsc::optIfConvert()
                 m_cond->ChangeOper(GenTree::ReverseRelop(m_cond->OperGet()));
                 cns = selectTrueInput;
             }
-            select = m_comp->gtNewOperNode(GT_ADD, m_cond->TypeGet(), m_cond, cns);
+            select = m_comp->gtNewOperNode(GT_ADD, selectType, m_cond, cns);
         }
     }
 
--- a/src/coreclr/jit/ifconversion.cpp
+++ b/src/coreclr/jit/ifconversion.cpp
@@ -727,7 +727,16 @@ bool OptIfConversionDsc::optIfConvert()
                 m_cond->ChangeOper(GenTree::ReverseRelop(m_cond->OperGet()));
                 cns = selectTrueInput;
             }
-            select = m_comp->gtNewOperNode(GT_ADD, selectType, m_cond, cns);
+
+            if (!cns->IsIntegralConst(0))
+            {
+                select = m_comp->gtNewOperNode(GT_ADD, selectType, m_cond, cns);
+            }
+            else
+            {
+                // ADD(COND, 0) -> COND
+                select = m_cond;
+            }
         }
     }
 
--- a/src/coreclr/jit/ifconversion.cpp
+++ b/src/coreclr/jit/ifconversion.cpp
@@ -711,7 +711,8 @@ bool OptIfConversionDsc::optIfConvert()
     // For SELECT(COND, CNS1, CNS2) check if we can fold it into ADD(COND, MIN(CNS1, CNS2)) if
     // the difference between CNS1 and CNS2 is exactly 1. E.g.:
     //
-    // SELECT(COND, 10, 11) -> ADD(COND, 10)
+    // SELECT(COND, 2, 1) -> ADD(COND, 1)
+    // SELECT(COND, 1, 2) -> ADD(REVERSED_COND, 1)
     //
     // Conservatively give up on relops with NaNs (we won't be able to reverse them).
     if (selectTrueInput->IsCnsIntOrI() && selectFalseInput->IsCnsIntOrI() &&
--- a/src/coreclr/jit/ifconversion.cpp
+++ b/src/coreclr/jit/ifconversion.cpp
@@ -713,26 +713,20 @@ bool OptIfConversionDsc::optIfConvert()
     //
     // SELECT(COND, 10, 11) -> ADD(COND, 10)
     //
-    if (selectTrueInput->IsCnsIntOrI() && selectFalseInput->IsCnsIntOrI() && m_cond->TypeIs(TYP_INT))
+    if (selectTrueInput->IsCnsIntOrI() && selectFalseInput->IsCnsIntOrI() &&
+        ((m_cond->gtFlags & GTF_RELOP_NAN_UN) == 0))
     {
         const ssize_t selectTrueVal  = selectTrueInput->AsIntConCommon()->IconValue();
         const ssize_t selectFalseVal = selectFalseInput->AsIntConCommon()->IconValue();
         if (abs(selectTrueVal - selectFalseVal) == 1)
         {
+            GenTree* cns = selectFalseInput;
             if (selectTrueVal < selectFalseVal)
             {
-                // if selectTrueVal is less than selectFalseVal then we need to reverse the condition
-                if (m_cond->OperIsCompare() && ((m_cond->gtFlags & GTF_RELOP_NAN_UN) == 0))
-                {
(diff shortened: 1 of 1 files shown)


jakobbotsch:
Can you show micro benchmarks? This makes the critical path longer. (Also, it undoes work done in #82031, so it should be merged with that instead if it is actually profitable.)

EgorBo:
> Can you show micro benchmarks? This makes the critical path longer. (Also, it undoes work done in #82031, so it should be merged with that instead if it is actually profitable.)

I was doing this for XARCH where it's a pure win (native compilers do the same), I didn't limit it on ARM64 because I wanted to see diffs

jakobbotsch:
> I was doing this for XARCH where it's a pure win (native compilers do the same)

I benchmarked some of the transformations that native compilers do in the area when I first enabled `cmov`, and I saw performance regressions on modern hardware.

jakobbotsch:
> I didn't limit it on ARM64 because I wanted to see [diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=465518&view=ms.vss-build-web.run-extensions-tab)

Regardless, if we end up with this transformation I think it should be done in `TryLowerCnsIntCselToCinc` (which should be renamed). It is already doing the similar checks. I'd just like avoid having the transformation in two separate places for the two architectures.

EgorBo:
> > I didn't limit it on ARM64 because I wanted to see [diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=465518&view=ms.vss-build-web.run-extensions-tab)
> 
> Regardless, if we end up with this transformation I think it should be done in `TryLowerCnsIntCselToCinc` (which should be renamed). It is already doing the similar checks. I'd just like avoid having the transformation in two separate places for the two architectures.

I don't mind moving it to lower, but now for `x == 42 ? 1 : 0` we emit  `SELECT(X == 42 ? 1 : 0)` while it is just `X == 42` with this PR

jakobbotsch:
> I don't mind moving it to lower, but now for `x == 42 ? 1 : 0` we emit `SELECT(X == 42 ? 1 : 0)` while it is just `X == 42` with this PR

Do you mean for TP? We already handle this pattern in lowering, since #81880. (Looks like that PR is the one where I measured one of the transformations that native compilers do that aren't beneficial today.)

EgorBo:
> I benchmarked some of the transformations that native compilers do in the area when I first enabled `cmov`, and I saw performance regressions on modern hardware.

```cs
[Arguments(42)]
[Arguments(43)]
[Benchmark]
public int Bench(int x) => x == 42 ? 100 : 101;
```
```
| Method | Toolchain                   | x  | Mean      | Error     |
|------- |---------------------------- |--- |----------:|----------:|
| Bench  | \Core_Root\corerun.exe      | 42 | 0.0020 ns | 0.0342 ns |
| Bench  | \Core_Root_base\corerun.exe | 42 | 0.1785 ns | 0.2613 ns |
|        |                             |    |           |           |
| Bench  | \Core_Root\corerun.exe      | 43 | 0.0027 ns | 0.0838 ns |
| Bench  | \Core_Root_base\corerun.exe | 43 | 0.1850 ns | 0.1008 ns |
```
Ryzen 7950X, I wanted to add a benchmark where I do this in a loop but I guess we don't emit SELECT in loops?

jakobbotsch:
> Ryzen 7950X, I wanted to add a benchmark where I do this in a loop but I guess we don't emit SELECT in loops?

You can set `DOTNET_JitStressModeName=STRESS_IF_CONVERSION_INNER_LOOPS` to allow that.

What are the results if the input argument is random?

jakobbotsch:
E.g. for
```csharp
public class Tests
{
    private int[] _values = Enumerable.Range(0, 100000).Select(_ => Random.Shared.Next(2)).ToArray();

    [Benchmark]
    public int Test()
    {
        int result = 0;
        foreach (int value in _values)
        {
            result += value == 0 ? 42 : 43;
        }

        return result;
    }
}
```
on my machine, with `DOTNET_JitStressModeNames=STRESS_IF_CONVERSION_INNER_LOOPS`, I get

| Method |        Job |                   Toolchain |     Mean |    Error |   StdDev | Ratio |
|------- |----------- |---------------------------- |---------:|---------:|---------:|------:|
|   Test | Job-YBZJSD |      \Core_Root\corerun.exe | 46.47 us | 0.103 us | 0.096 us |  1.10 |
|   Test | Job-EBKPFW | \Core_Root_base\corerun.exe | 42.42 us | 0.075 us | 0.063 us |  1.00 |

EgorBo:
> You can set `DOTNET_JitStressModeName=STRESS_IF_CONVERSION_INNER_LOOPS` to allow that.

It didn't help in my case :(

> What are the results if the input argument is random?

Seems like perf is the same as cmov in that case.

> Regardless, if we end up with this transformation I think it should be done in TryLowerCnsIntCselToCinc

It seems it makes it a lot more complicated because at Lower we end up lowering relop to VOID GT_CMP so I no longer can easily use that with ADD

EgorBo:
> E.g. for

Do you see GT_SELECT being generated there? I don't see it locally even with `DOTNET_JitStressModeName=STRESS_IF_CONVERSION_INNER_LOOPS`

jakobbotsch:
> > E.g. for
> 
> Do you see GT_SELECT being generated there? I don't see it locally even with `DOTNET_JitStressModeName=STRESS_IF_CONVERSION_INNER_LOOPS`

I think so, my codegen is
```asm
G_M29319_IG03:  ;; offset=0x0015
       mov      r10d, edx
       mov      r10d, dword ptr [rcx+4*r10+0x10]
       mov      r9d, 42
       mov      r11d, 43
       test     r10d, r10d
       cmovne   r9d, r11d
       add      eax, r9d
       inc      edx
       cmp      r8d, edx
       jg       SHORT G_M29319_IG03
```
(base)

and
```asm
G_M29319_IG03:  ;; offset=0x0015
       mov      r10d, edx
       mov      r10d, dword ptr [rcx+4*r10+0x10]
       test     r10d, r10d
       setne    r10b
       movzx    r10, r10b
       add      r10d, 42
       add      eax, r10d
       inc      edx
       cmp      r8d, edx
       jg       SHORT G_M29319_IG03
```
(diff)

Did you pass it through `--envvars`?

Otherwise, try this one:
```csharp
    public class Benchmarks
    {
        private bool[] _values = Enumerable.Range(0, 10000).Select(i => Random.Shared.Next() % 2 == 0).ToArray();

        [Benchmark]
        public int Bench1()
        {
            int result = 0;
            foreach (bool b in _values)
            {
                result += Sel(b);
            }

            return result;
        }

        private int Sel(bool b) => b ? 43 : 42;
    }
```

This is what the benchmark I did in #81880 looked like (just with some different constants in `Sel`). I did that measurement on my Intel CPU, and the results with this PR are:

BenchmarkDotNet=v0.13.4, OS=Windows 11 (10.0.22635.2700)
Intel Core i9-10885H CPU 2.40GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK=8.0.100-preview.6.23330.14
  [Host]     : .NET 8.0.0 (8.0.23.32907), X64 RyuJIT AVX2
  Job-CTIBZZ : .NET 9.0.0 (42.42.42.42424), X64 RyuJIT AVX2
  Job-QNTRDF : .NET 9.0.0 (42.42.42.42424), X64 RyuJIT AVX2

EnvironmentVariables=DOTNET_JitDisasm=Sel

| Method |        Job |                   Toolchain |     Mean |    Error |   StdDev | Ratio | RatioSD |
|------- |----------- |---------------------------- |---------:|---------:|---------:|------:|--------:|
| Bench1 | Job-CTIBZZ |      \Core_Root\corerun.exe | 26.17 us | 0.572 us | 1.678 us |  1.47 |    0.10 |
| Bench1 | Job-QNTRDF | \Core_Root_base\corerun.exe | 19.05 us | 0.380 us | 0.317 us |  1.00 |    0.00 |

jakobbotsch:
> Do you see GT_SELECT being generated there? I don't see it locally even with `DOTNET_JitStressModeName=STRESS_IF_CONVERSION_INNER_LOOPS`

I made a typo above, the env var name is `DOTNET_JitStressModeNames`, with an 's'

EgorBo:
Thanks! will not move forward with this then

PR #94597, opened 2023-11-10
JIT: Improve local assertion prop throughput

AndyAyersMS:
Leverage the "dep vectors" to avoid the search the assertion table during local assertion prop. Helps the current (small table) behavior some, helps the future cross-block (larger table) behavior more.

Similar tricks may be possible for global AP, though the set of assertions there is more varied.

Contributes to #93246.

### DIFF
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -1617,7 +1617,7 @@ bool Compiler::optAssertionVnInvolvesNan(AssertionDsc* assertion)
  *
  *  If it is already in the assertion table return the assertionIndex that
  *  we use to refer to this element.
- *  Otherwise add it to the assertion table ad return the assertionIndex that
+ *  Otherwise add it to the assertion table and return the assertionIndex that
  *  we use to refer to this element.
  *  If we need to add to the table and the table is full return the value zero
  */
@@ -2490,15 +2490,19 @@ AssertionIndex Compiler::optFindComplementary(AssertionIndex assertIndex)
 //
 AssertionIndex Compiler::optAssertionIsSubrange(GenTree* tree, IntegralRange range, ASSERT_VALARG_TP assertions)
 {
-    if ((!optLocalAssertionProp && BitVecOps::IsEmpty(apTraits, assertions)) || !optCanPropSubRange)
+    if (!optCanPropSubRange)
     {
+        // (don't early out in checked, verify above)
         return NO_ASSERTION_INDEX;
     }
 
-    for (AssertionIndex index = 1; index <= optAssertionCount; index++)
+    BitVecOps::Iter iter(apTraits, assertions);
+    unsigned        bvIndex = 0;
+    while (iter.NextElem(&bvIndex))
     {
-        AssertionDsc* curAssertion = optGetAssertion(index);
-        if (BitVecOps::IsMember(apTraits, assertions, index - 1) && curAssertion->CanPropSubRange())
+        AssertionIndex const index        = GetAssertionIndex(bvIndex);
+        AssertionDsc* const  curAssertion = optGetAssertion(index);
+        if (curAssertion->CanPropSubRange())
         {
             // For local assertion prop use comparison on locals, and use comparison on vns for global prop.
             bool isEqual = optLocalAssertionProp
@@ -2530,18 +2534,12 @@ AssertionIndex Compiler::optAssertionIsSubrange(GenTree* tree, IntegralRange ran
  */
 AssertionIndex Compiler::optAssertionIsSubtype(GenTree* tree, GenTree* methodTableArg, ASSERT_VALARG_TP assertions)
 {
-    if (BitVecOps::IsEmpty(apTraits, assertions))
+    BitVecOps::Iter iter(apTraits, assertions);
+    unsigned        bvIndex = 0;
+    while (iter.NextElem(&bvIndex))
     {
-        return NO_ASSERTION_INDEX;
-    }
-    for (AssertionIndex index = 1; index <= optAssertionCount; index++)
-    {
-        if (!BitVecOps::IsMember(apTraits, assertions, index - 1))
-        {
-            continue;
-        }
-
-        AssertionDsc* curAssertion = optGetAssertion(index);
+        AssertionIndex const index        = GetAssertionIndex(bvIndex);
+        AssertionDsc*        curAssertion = optGetAssertion(index);
         if (curAssertion->assertionKind != OAK_EQUAL ||
             (curAssertion->op1.kind != O1K_SUBTYPE && curAssertion->op1.kind != O1K_EXACT_TYPE))
         {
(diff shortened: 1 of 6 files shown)


AndyAyersMS:
Locally this looks really good for TP once cross-block is enabled. I picked up the fix from #94608 so let's see what the lab thinks.

AndyAyersMS:
Old cross-block TP

image

With this PR at [034dbda](https://github.com/dotnet/runtime/pull/94597/commits/034dbda1a65d13612ec2b94f80f5cec3bc033a83)

image

AndyAyersMS:
With [6a07b54](https://github.com/dotnet/runtime/pull/94597/commits/6a07b54984d6f5197e8a71c6ec7ddb1d137f2e6b) I am now seeing 0.13% on benchmarks.run; baseline is more like 1.57%.... so this may get us to a happy TP place.

Would be nice to enacapsulate this pattern a bit better, perhaps. Also maybe in global AP we can think about ways to do similar indexing, as there are still various whole-table searches going on there.

AndyAyersMS:
One odd thing is that after the force push / rebase I lost some of the improvements and gained some sizeable regressions. EG win x64 went from -1.1M to -800K.

I've verified this change is no diff vs baseline (either with or without cross-block enabled) so not sure what happened just yet. Will need to diff against some older build I guess.

AndyAyersMS:
@jakobbotsch PTAL
cc @dotnet/jit-contrib

AndyAyersMS:
[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=466591&view=ms.vss-build-web.run-extensions-tab)

No codegen diffs; some TP a bit faster, some slower -- the more dramatic TP impact here is with cross-block enabled (see above). It may be that our bit vector iterator has some unnecessary overhead, as it seems like walking through an N entry table (with at most 64 entries) should generally be more costly that iterating through a bit vector with at most N (and likely just a few) bits set.

src/coreclr/jit/assertionprop.cpp
```diff
@@ -1633,13 +1633,57 @@ AssertionIndex Compiler::optAddAssertion(AssertionDsc* newAssertion)
...
+        unsigned        lclNum = newAssertion->op1.lcl.lclNum;
+        BitVecOps::Iter iter(apTraits, GetAssertionDep(lclNum));
+        unsigned        bvIndex = 0;
+        while (iter.NextElem(&bvIndex))
         {
-            return index;
+            AssertionIndex const index        = GetAssertionIndex(bvIndex);
+            AssertionDsc* const  curAssertion = optGetAssertion(index);
+
+            if (curAssertion->Equals(newAssertion, !optLocalAssertionProp))
```
jakobbotsch, line 1652:
```suggestion
            if (curAssertion->Equals(newAssertion, /* vnBased */ false))
```

jakobbotsch: (Also in the other cases)

src/coreclr/jit/assertionprop.cpp
```diff
@@ -1633,13 +1633,57 @@ AssertionIndex Compiler::optAddAssertion(AssertionDsc* newAssertion)
...
+            {
+                AssertionIndex const index        = GetAssertionIndex(bvIndex);
+                AssertionDsc* const  curAssertion = optGetAssertion(index);
+
+                if (curAssertion->Equals(newAssertion, !optLocalAssertionProp))
+                {
+                    return index;
+                }
+            }
+        }
```
jakobbotsch, line 1673: Is this necessary? Shouldn't the previous case have found it if there is an equal assertion? Or will we only keep an assertion like `v1 = v2` in one of the bit vectors?

AndyAyersMS: No, it's kept for both (see just below) -- so yeah that second loop is not needed.

AndyAyersMS:
> One odd thing is that after the force push / rebase I lost some of the improvements and gained some sizeable regressions. EG win x64 went from -1.1M to -800K.
> 
> I've verified this change is no diff vs baseline (either with or without cross-block enabled) so not sure what happened just yet. Will need to diff against some older build I guess.

Aha, some OSR cases got pessimized with the DFS reachability check, as the original method entry and such are considered unreachable. Will push a fix.

AndyAyersMS:
There are around 100 or so methods that see diffs now with cross-block enabled (because of the now fixed reachability check); seems like a lot of them are compiled regex methods:
```
[14:24:05] Top method improvements (percentages):
[14:24:05]          -65 (-4.90% of base) : 304049.dasm - System.Text.RegularExpressions.CompiledRegexRunner:Regex32447_TryMatchAtCurrentPosition(System.Text.RegularExpressions.RegexRunner,System.ReadOnlySpan`1[ushort]):ubyte (FullOpts)
[14:24:05]          -73 (-3.87% of base) : 298116.dasm - System.Text.RegularExpressions.CompiledRegexRunner:Regex5110_TryMatchAtCurrentPosition(System.Text.RegularExpressions.RegexRunner,System.ReadOnlySpan`1[ushort]):ubyte (FullOpts)
[14:24:05]          -34 (-3.84% of base) : 298413.dasm - System.Text.RegularExpressions.CompiledRegexRunner:Regex5473_TryMatchAtCurrentPosition(System.Text.RegularExpressions.RegexRunner,System.ReadOnlySpan`1[ushort]):ubyte (FullOpts)
[14:24:05]          -10 (-1.35% of base) : 304046.dasm - System.Text.RegularExpressions.CompiledRegexRunner:Regex32443_TryMatchAtCurrentPosition(System.Text.RegularExpressions.RegexRunner,System.ReadOnlySpan`1[ushort]):ubyte (FullOpts)
[14:24:05]           -9 (-1.18% of base) : 130597.dasm - System.IO.Compression.ZipArchive:WriteFile():this (FullOpts)
...etc...
```

AndyAyersMS:
Local TP diff with cross-block enabled (based on 7c8fff93d1e). Seems almost too good to be true.

image

I'll prep another PR for enabling once this bit is merged, and we'll see what the lab says.

AndyAyersMS:
@jakobbotsch take another look when you can...

src/coreclr/jit/fgopt.cpp
```diff
@@ -788,6 +788,16 @@ unsigned Compiler::fgDfsReversePostorder()
     //
     fgDfsReversePostorderHelper(fgFirstBB, visited, preorderIndex, postorderIndex);
 
+    // For OSR, walk from the original method entry too.
+    //
+    if (opts.IsOSR() && (fgEntryBB != nullptr))
+    {
+        if (!BlockSetOps::IsMember(this, visited, fgEntryBB->bbNum))
+        {
+            fgDfsReversePostorderHelper(fgEntryBB, visited, preorderIndex, postorderIndex);
```
jakobbotsch, line 797: These aren't really reachable, right? (Except for maybe tailcall-to-loop opt). But we want them in the order to be able to propagate facts from them.

I wonder if SSA/VN could benefit similarly to seeing these.

jakobbotsch (approved):

src/coreclr/jit/fgopt.cpp
jakobbotsch: It seems like a sign that CQ for OSR methods would benefit from inserting the jump to the start block very late, e.g. right before lowering (probably with some additional DCE for the unreachable blocks). Of course at some cost of TP, so maybe it would need to be restricted to loop-based patchpoints while the current behavior is ideal for partial compilation.

AndyAyersMS: Added a note to#33658 to try this someday. One consideration for deferral is that we might need to alter or block some optimizations that might remove computations from the OSR side (hoisting, CSE).

AndyAyersMS: Also note currently `fgEntryBB` gets nulled out after morph.

PR #94601, opened 2023-11-10
[browser][MT] GC coop safepoints

pavelsavara:


lambdageek:
Ah, I was wrong. We don't set the default in cmake for threaded wasm, we set it at runtime in the driver:

https://github.com/dotnet/runtime/blob/c07cb4441d6270327472e88919f3d4eb6cc0450f/src/mono/wasm/runtime/driver.c#L436C54-L436C54

I think what I would prefer is to set it in mono.proj like you've done here, and also add something like this over here:

https://github.com/dotnet/runtime/blob/c07cb4441d6270327472e88919f3d4eb6cc0450f/src/mono/mono/utils/mono-threads-wasm.c#L95-L98

```c
#include <mono/utils/mono-threads-coop.h>
...

void
mono_threads_suspend_init (void)
{
#ifndef DISABLE_THREADS
	// wasm threading require full cooperative suspend
	g_assert (mono_threads_is_cooperative_suspension_enabled ());
#endif
}
```

lambdageek (approved):

PR #94608, opened 2023-11-10
JIT: fix cross block assertion prop

AndyAyersMS:
My last round of revisions to #94363 introduced a bug. Fix.


AndyAyersMS:
@jakobbotsch PTAL
cc @dotnet/jit-contrib

jakobbotsch (approved):

Issue #94513, opened 2023-11-08
Can't import Component Model functions in WASI apps


mhmd-azeez:
### Description

While you can import functions in WASI apps (see #90786), the process breaks down when the imported namespace contains `:` and `/`. This is problematic because these are valid characters and are used widely in [Wasm Component Model spec](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md). For example, `wasi:http/handler` is a valid import namespace:
```wat
(component
  (import "custom-hook" (func (param string) (result string)))
  (import "wasi:http/handler" (instance
    (export "request" (type $request (sub resource)))
    (export "response" (type $response (sub resource)))
    (export "handle" (func (param (own $request)) (result (own $response))))
  ))
  (import "url=<https://mycdn.com/my-component.wasm>" (component ...))
  (import "relative-url=<./other-component.wasm>,integrity=<sha256-X9ArH3k...>" (component ...))
  (import "locked-dep=<my-registry:sqlite@1.2.3>,integrity=<sha256-H8BRh8j...>" (component ...))
  (import "unlocked-dep=<my-registry:imagemagick@{>=1.0.0}>" (instance ...))
  (import "integrity=<sha256-Y3BsI4l...>" (component ...))
  ... impl
  (export "wasi:http/handler" (instance $http_handler_impl))
  (export "get-JSON" (func $get_json_impl))
)
```

### Reproduction Steps

csproj:
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <RuntimeIdentifier>wasi-wasm</RuntimeIdentifier>
    <OutputType>Exe</OutputType>
    <PublishTrimmed>true</PublishTrimmed>
    <WasmSingleFileBundle>true</WasmSingleFileBundle>
    <WasmBuildNative>true</WasmBuildNative>
    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
  </PropertyGroup>

  <ItemGroup>
    <NativeFileReference Include="$(MSBuildThisFileDirectory)extism.c" />
    <_WasmNativeFileForLinking Include="@(NativeFileReference)" />
  </ItemGroup>

</Project>
```

Program.cs:
```csharp
using System;
using System.Runtime.CompilerServices;
using System.Runtime.InteropServices;

public class Program
{
    [DllImport("extism:host/user")]
    public static extern void do_something();

    public static void Main(string[] args)
    {
[...]

Issue #86985, opened 2023-05-31
maybe create WASM module export from `[UnmanagedCallersOnly]`


lewing:
UnmanagedCallersOnly symbols could create exports in  the wasm module.  Particularly if they are not resolved statically in the rest of the build.

jkotas:
cc @yowl

lewing:
Support was added for this in https://github.com/dotnet/runtime/pull/94615 but there is an interpreter bug that complicates calling the export.

Issue #86984, opened 2023-05-31
create WASM module import from any `[DLLImport]` which doesn't match known static native symbols


lewing:
From @SteveSandersonMS [link](https://github.com/dotnet/runtime/issues/65895#issuecomment-1546766222)

I've been trying out the pinvoke generation and found it can't yet generate real WebAssembly imports. There are two main reasons:

It ignores pinvokes unless they correspond to an entry on _WasmNativeFileForLinking (even though when you want to generate an import, you won't be linking with any module for it)
When generating pinvoke-table.h, it never adds any import_module/import_name clang attributes, which means:
The module is always compiled as env (so it's impossible to reference most real-world libraries)
The symbol name is mangled
I was able to work around it via some MSBuild hackery, so can be fairly confident what would be the steps to solve this:

Change the pinvoke generator so that when a DllImport's module name does not correspond to an entry in _WasmNativeFileForLinking, it does still generate the pinvoke-table.h entry
... but those pinvoke-table.h entries must also be annotated with import_module/import_name
For example, given these two DllImports:

[LibraryImport("libSystem.Native")]
private static partial int Method1(int a);

[LibraryImport("mylib", EntryPoint = "this:should:not:be:mangled")]
private static partial int Method2(int a);
... the pinvoke-table.h definition should look like:

int Method1 (int);

__attribute__((import_module("mylib"), import_name("this:should:not:be:mangled"))) int Method2 (int);
The first one does not need any __attribute__ because libSystem.Native does match up with one of the linker inputs (i.e., _WasmNativeFileForLinking entries). The existing codegen is correct.

The second one does need the clang import annotations because it doesn't match a _WasmNativeFileForLinking entry, so it has to become a real wasm import, and it's important to retain the real module name (mylib) and the real import name (this:should:not:be:mangled) even though it's not a legal C function name.

From @lewing 
The goal of this issue is making sure that once the build completes the wasm module is able to have imports of `<module>.<method>` with unmangled names for missing symbols, not to solve wasi component linking more generally.

jkotas:
cc @yowl

jkotas:
> like numbers and string (with encoding dictated by the ABI).

I expect that managed strings are going require some marshalling. What does that marshaling look like?

> I think we should not try to implement WIT on [DllImport] nor marshaling of data structures, right ?

Yes, we should avoid any runtime provided marshalling. Instead, depend on source generators to provide the marshalling code.

silesmo:
[...]

PR #94615, opened 2023-11-10
[wasm] Implement initial support for WasmImportLinkage in the AppBuilders

lewing:
In response to https://github.com/dotnet/runtime/pull/93823 this is a rough partial fix for https://github.com/dotnet/runtime/issues/94513, https://github.com/dotnet/runtime/issues/86984, and https://github.com/dotnet/runtime/issues/86985
With the default settings this works in wasi-wasm but emscripten will complain if you don't enable unresolved symbols on browser-wasm

Using
```csharp
    [WasmImportLinkage]
    [DllImport("extism:host/user")]
    public static extern void do_something();
```

Will generate
```wat
(type $i64_i32_=>_i64 (func (param i64 i32) (result i64)))
 (type $i32_i64_i64_i32_=>_none (func (param i32 i64 i64 i32)))
 (import "wasi_snapshot_preview1" "sock_accept" (func $fimport$0 (param i32 i32 i32) (result i32)))
 (import "extism:host/user" "do_something" (func $fimport$1))
 (import "wasi_snapshot_preview1" "args_get" (func $fimport$2 (param i32 i32) (result i32)))
```

### DIFF
--- a/src/mono/sample/wasi/native/Program.cs
+++ b/src/mono/sample/wasi/native/Program.cs
@@ -1,0 +1,40 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.InteropServices;
+
+public unsafe class Test
+{
+    [UnmanagedCallersOnly(EntryPoint = "ManagedFunc")]
+    public static int MyExport(int number)
+    {
+        // called from MyImport aka UnmanagedFunc
+        Console.WriteLine($"MyExport({number}) -> 42");
+        return 42;
+    }
+
+    [DllImport("*", EntryPoint = "UnmanagedFunc")]
+    public static extern void MyImport(); // calls ManagedFunc aka MyExport
+
+    [DllImport("*")]
+    public static unsafe extern void ReferenceFuncPtr(delegate* unmanaged<int,int> funcPtr);
+
+    static bool AlwaysFalse(string [] args) => false;
+
+    public unsafe static int Main(string[] args)
+    {
+        Console.WriteLine($"main: {args.Length}");
+        MyImport();
+
+        if (AlwaysFalse(args))
+        {
+            // never called (would an error on wasm) and doesn't actually do anything
+            // but required for wasm_native_to_interp_ftndesc initialization
+            // the lookup happens before main when this is here
+            ReferenceFuncPtr(&MyExport);
+        }
+
+        return 0;
+    }
+}
--- a/src/mono/sample/wasi/native/Program.cs
+++ b/src/mono/sample/wasi/native/Program.cs
@@ -17,24 +17,15 @@ public unsafe class Test
     [DllImport("*", EntryPoint = "UnmanagedFunc")]
     public static extern void MyImport(); // calls ManagedFunc aka MyExport
 
-    [DllImport("*")]
-    public static unsafe extern void ReferenceFuncPtr(delegate* unmanaged<int,int> funcPtr);
-
-    static bool AlwaysFalse(string [] args) => false;
-
     public unsafe static int Main(string[] args)
     {
         Console.WriteLine($"main: {args.Length}");
-        MyImport();
-
-        if (AlwaysFalse(args))
-        {
-            // never called (would an error on wasm) and doesn't actually do anything
-            // but required for wasm_native_to_interp_ftndesc initialization
-            // the lookup happens before main when this is here
-            ReferenceFuncPtr(&MyExport);
+        // workaround to force the interpreter to initialize wasm_native_to_interp_ftndesc for MyExport
+        if (args.Length > 10000) {
+            ((IntPtr)(delegate* unmanaged<int,int>)&MyExport).ToString();
         }
 
+        MyImport();
         return 0;
     }
 }
--- a/src/mono/sample/wasi/native/Wasi.Native.Sample.csproj
+++ b/src/mono/sample/wasi/native/Wasi.Native.Sample.csproj
@@ -1,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>$(NetCoreAppCurrent)</TargetFramework>
+    <RuntimeIdentifier>wasi-wasm</RuntimeIdentifier>
+    
+    <TargetOs>wasi</TargetOs>
+    <WasmBuildNative>true</WasmBuildNative>
+    <WasmNativeStrip>false</WasmNativeStrip>
+    <IsBrowserWasmProject>false</IsBrowserWasmProject>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <NativeFileReference Include="local.c" />
+  </ItemGroup>
+  <Target Name="RunSample" DependsOnTargets="RunSampleWithWasmtime" />
+</Project>
--- a/src/mono/sample/wasi/native/Wasi.Native.Sample.csproj
+++ b/src/mono/sample/wasi/native/Wasi.Native.Sample.csproj
@@ -7,6 +7,7 @@
     <WasmBuildNative>true</WasmBuildNative>
     <WasmNativeStrip>false</WasmNativeStrip>
     <IsBrowserWasmProject>false</IsBrowserWasmProject>
+    <WasmSingleFileBundle>true</WasmSingleFileBundle>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/mono/sample/wasi/native/local.c
+++ b/src/mono/sample/wasi/native/local.c
@@ -9,18 +9,3 @@ void UnmanagedFunc()
     ret = ManagedFunc(123);
     printf("ManagedFunc returned %d\n", ret);
 }
-
-
-#ifdef PARSING_FUNCTION_POINTERS_IS_SUPPORTED
-/*
- warning WASM0001: Could not get pinvoke, or callbacks for method 'Test::ReferenceFuncPtr' because 'Parsing function pointe
-r types in signatures is not supported.'
-warning WASM0001: Skipping pinvoke 'Test::ReferenceFuncPtr' because 'Parsing function pointer types in signatures is not s
-upported.'
-*/
-typedef int (*ManagedFuncPtr)(int);
-
-void ReferenceFuncPtr(ManagedFuncPtr funcPtr) {
-    printf("ReferenceFuncPtr %p\n", funcPtr);
-}
-#endif
--- a/src/mono/sample/wasi/native/local.c
+++ b/src/mono/sample/wasi/native/local.c
@@ -1,0 +1,26 @@
+#include <stdio.h>
+
+int ManagedFunc(int number);
+
+void UnmanagedFunc()
+{
+    int ret = 0;
+    printf("UnmanagedFunc calling ManagedFunc\n");
+    ret = ManagedFunc(123);
+    printf("ManagedFunc returned %d\n", ret);
+}
+
+
+#ifdef PARSING_FUNCTION_POINTERS_IS_SUPPORTED
+/*
+ warning WASM0001: Could not get pinvoke, or callbacks for method 'Test::ReferenceFuncPtr' because 'Parsing function pointe
+r types in signatures is not supported.'
+warning WASM0001: Skipping pinvoke 'Test::ReferenceFuncPtr' because 'Parsing function pointer types in signatures is not s
+upported.'
+*/
+typedef int (*ManagedFuncPtr)(int);
+
+void ReferenceFuncPtr(ManagedFuncPtr funcPtr) {
+    printf("ReferenceFuncPtr %p\n", funcPtr);
+}
+#endif
--- a/src/tasks/WasmAppBuilder/PInvokeCollector.cs
+++ b/src/tasks/WasmAppBuilder/PInvokeCollector.cs
@@ -15,17 +15,20 @@ using Microsoft.Build.Tasks;
 internal sealed class PInvoke : IEquatable<PInvoke>
 #pragma warning restore CA1067
 {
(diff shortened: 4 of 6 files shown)


src/tasks/WasmAppBuilder/PInvokeTableGenerator.cs
```diff
@@ -137,7 +132,7 @@ private void EmitPInvokeTable(StreamWriter w, Dictionary<string, string> modules
                 Where(l => l.Module == module && !l.Skip).
                 OrderBy(l => l.EntryPoint).
                 GroupBy(d => d.EntryPoint).
-                Select(l => "{\"" + _fixupSymbolName(l.Key) + "\", " + _fixupSymbolName(l.Key) + "}, " +
+                Select(l => "{\"" + l.Key + "\", " + TransformEntryPoint(l.First()) + "}, " +
```
maraf, line 135: Can we use `l.Key` instead of `TransformEntryPoint(l.First())`? It seems to me that it's the same

src/tasks/WasmAppBuilder/PInvokeTableGenerator.cs
```diff
@@ -63,38 +63,33 @@ public IEnumerable<string> Generate(string[] pinvokeModules, string outputPath)
...
-        {
-            try
+        w.WriteLine("// GENERATED FILE, DO NOT MODIFY");
+        w.WriteLine();
+
+        foreach (var pinvoke in pinvokes) {
+            // Handle special modules, and add them to the list of modules
+            // otherwise, skip them and throw an exception at runtime if they
+            // are called.
+            if (pinvoke.WasmLinkage && !modules.ContainsKey(pinvoke.Module))
```
maraf, line 75: nit: Move `!modules.ContainsKey(pinvoke.Module)` to a separate if and don't repeat it (I wanted to submit a patch, but github says it's not supported here)

src/tasks/WasmAppBuilder/PInvokeTableGenerator.cs
lewing: no, the l.key is unmangled here (I'll merge an escaping fix for c literals shortly). TransformEntryPoint(l.First()) will get you the c symbol. They have to be able to be different.

lewing: Added calls to escape the literal string in case the one of the names was something like `mymod"tacobell`

lewing:
added export support using WasmImportLinkage despite the questionable naming

```csharp
 
        [WasmImportLinkage]
        [UnmanagedCallersOnly(EntryPoint = "display_meaning")]
        public static void DisplayMeaningExport(int meaning)
        {
            Console.WriteLine($"The meaning of life is {meaning}");
        }
```

```wat
 (export "display_meaning" (func $13484))
```

SingleAccretion:
> added export support using WasmImportLinkage

The attribute was not meant to enable exports. They are meant to be enabled by simply using `EntryPoint` on UCO-adorned methods.

just-ero:
I apologize upfront if this is already up for consideration or an entirely inappropriate suggestion;

I think I would personally prefer if it were just
```cs
[WasmImport("extism:host/user")]
public static extern void do_something();
```

lewing:
> > added export support using WasmImportLinkage
> 
> The attribute was not meant to enable exports. They are meant to be enabled by simply using `EntryPoint` on UCO-adorned methods.

I'm aware, this is explicitly, as mentioned in the description, a quick hack to test out the changes.

lewing:
This is useful as is and blocking @kg from other related work

src/tasks/WasmAppBuilder/PInvokeTableGenerator.cs
```diff
@@ -245,21 +232,40 @@ private static bool TryIsMethodGetParametersUnsupported(MethodInfo method, [NotN
...
+        var method = export.Method;
+        // EntryPoint wasn't specified generate a name for the entry point
+        return _fixupSymbolName($"wasm_native_to_interp_{method.DeclaringType!.Module!.Assembly!.GetName()!.Name!}_{method.DeclaringType.Name}_{method.Name}");
+    }
+
+    private string DelegateKey(PInvokeCallback export)
+    {
+        // FIXME: this is a hack, we need to encode this better
+        // and allow reflection in the interp case but either way
+        // it needs to match the key passed to get_native_to_interp
```
lewing, line 259: Need to open an issue about this.  It's ugly and complicating a fix for calling into the entry point without having to create a managed delegate to initialize the lookup table

lewing:
cc @silesmo

silesmo:
> cc @silesmo
This is great, thanks, @lewing!

Do you know when this will get merged?

src/mono/sample/wasi/native/local.c
```diff
@@ -0,0 +1,11 @@
+#include <stdio.h>
+
+int ManagedFunc(int number);
+
+void UnmanagedFunc()
```
pavelsavara, line 5: let's rename this to `MyImport` so that it's not confusing the reader of C# side

src/mono/sample/wasi/native/local.c
```diff
@@ -0,0 +1,11 @@
+#include <stdio.h>
+
+int ManagedFunc(int number);
```
pavelsavara, line 3: rename to `MyExport` ?

src/mono/sample/wasi/native/Wasi.Native.Sample.csproj
```diff
@@ -0,0 +1,17 @@
...
+    
+    <TargetOs>wasi</TargetOs>
+    <WasmBuildNative>true</WasmBuildNative>
+    <WasmNativeStrip>false</WasmNativeStrip>
+    <IsBrowserWasmProject>false</IsBrowserWasmProject>
+    <WasmSingleFileBundle>true</WasmSingleFileBundle>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <NativeFileReference Include="local.c" />
```
pavelsavara, line 14: Do I understand correctly that all symbols in this sample are statically linked inside dotnet.wasm ? If so, should we also have sample that makes them external ?

src/mono/sample/wasi/native/Program.cs
```diff
@@ -0,0 +1,31 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.InteropServices;
+
+public unsafe class Test
+{
+    [UnmanagedCallersOnly(EntryPoint = "ManagedFunc")]
+    public static int MyExport(int number)
```
kg, line 10: Agreed with Pavel that it might be good for the entrypoint and managed name to match, unless we are explicitly testing that you can make them not match and it will still work

src/tasks/WasmAppBuilder/PInvokeTableGenerator.cs
```diff
@@ -245,21 +232,40 @@ private static bool TryIsMethodGetParametersUnsupported(MethodInfo method, [NotN
...
+        var method = export.Method;
+        string module_symbol = method.DeclaringType!.Module!.Assembly!.GetName()!.Name!;
+        return $"\"{module_symbol}_{method.DeclaringType.Name}_{method.Name}\"".Replace('.', '_');
```
kg, line 262: is `.` the only thing we need to replace? can't `+` and `<` show up too, among others?

kg (approved):

src/tasks/WasmAppBuilder/PInvokeTableGenerator.cs
lewing: > is `.` the only thing we need to replace? can't `+` and `<` show up too, among others?

This is a C string literal, not a symbol

src/mono/sample/wasi/native/Wasi.Native.Sample.csproj
lewing: This probably need to happen in the tests rather than the sample, I'm pretty sure the smoke tests run the samples and the import wouldn't resolve.

src/mono/sample/wasi/native/local.c
lewing: I did the mismatched names on purpose to demonstrate what setting EntryPoint does in both cases

src/tasks/WasmAppBuilder/PInvokeCollector.cs
```diff
@@ -241,8 +245,22 @@ internal sealed class PInvokeCallback
...
         Method = method;
+        foreach (var attr in method.CustomAttributes)
+        {
+            if (attr.AttributeType.Name == "UnmanagedCallersOnlyAttribute")
+            {
+                foreach(var arg in attr.NamedArguments)
+                {
+                    if (arg.MemberName == "EntryPoint")
+                    {
+                        EntryPoint = arg.TypedValue.Value!.ToString();
```
radical, line 256: can we return at this point?

src/tasks/WasmAppBuilder/PInvokeTableGenerator.cs
```diff
@@ -63,39 +64,39 @@ public IEnumerable<string> Generate(string[] pinvokeModules, string outputPath)
...
     }
 
-    private static bool HasAttribute(MemberInfo element, params string[] attributeNames)
+    private void EmitPInvokeTable(StreamWriter w, Dictionary<string, string> modules, List<PInvoke> pinvokes)
     {
-        foreach (CustomAttributeData cattr in CustomAttributeData.GetCustomAttributes(element))
-        {
-            try
+
+        foreach (var pinvoke in pinvokes) {
```
radical, line 70:
```suggestion
        foreach (var pinvoke in pinvokes)
        {
```

src/tasks/WasmAppBuilder/PInvokeCollector.cs
```diff
@@ -15,17 +15,20 @@
...
         Method = method;
+        WasmLinkage = wasmLinkage;
     }
 
     public string EntryPoint;
     public string Module;
     public MethodInfo Method;
     public bool Skip;
+    public bool WasmLinkage;
+
```
radical, line 31:
```suggestion
```

RReverser:
Oh this is awesome. I just finished rewriting bindings in our project for .NET 8, looks like I'll have to rewrite them again, but this removes some of the hacks I had to do so it's a win.

How can one try this PR out? Will it be available via `dotnet workload update` for `wasi-experimental` or do I need to pull it explicitly somehow?

lewing:
The packages will need to flow through the darc update process (hours, sometimes longer) until they hit `dotnet/installer` then you should be able to follow the instructions on the front page to download the latest 9.0-alpha.1 build (make sure to add the nuget feeds too) and then the workload commands should work.

lewing:
> I apologize upfront if this is already up for consideration or an entirely inappropriate suggestion;
> 
> I think I would personally prefer if it were just
> 
> ```cs
> [WasmImport("extism:host/user")]
> public static extern void do_something();
> ```

The API discussion happened here https://github.com/dotnet/runtime/issues/93824

RReverser:
> then you should be able to follow the instructions on the front page to download the latest 9.0-alpha.1 build (make sure to add the nuget feeds too

Ah, in that case I'll have to wait I guess. I thought it might be possible to get `wasi-experimental` workload updates independently from .NET itself for production .NET 8 use too, not just via alpha channel.

lewing:
There is the possibility that we would consider servicing some of these fixes in net8.0

RReverser:
That would be amazing. Having to manually wrap imports/exports via custom C file is the biggest remaining hurdle for us after migrating from Wasi.Sdk to .NET 8 WASI support, and it's a shame that this fix juuuust didn't make the release cut.

As we're providing a sort of Wasm hosting, where we provide a project template and users do `dotnet publish` themselves, we can't ask all end users to install an alpha version of the SDK, so it's either hoping that improvements like this will be backported to .NET 8 channel, or waiting for next release version of the SDK again.

silesmo:
> There is the possibility that we could consider servicing some of these fixes in net8.0

That would be great!

lewing:
> Btw, one more issue we were seeing is that for `UnmanagedCallersOnly` .NET complains that custom structs, even with fixed layout, are not blittable and can't be used as return type or arguments of UCO entry points.
> 
> Structs have well-defined ABI in Wasm C ABI, so this shouldn't be a problem and structs with fixed layout should be allowed like they are on other platforms.
> 
> At the very least it would be useful to allow a struct with a single field, since Wasm C ABI says those have the same ABI as their field. This is useful for creating typechecked handles where a `struct` just has a single integer field.

@kg has been looking into this

lewing:
> > There is the possibility that we could consider servicing some of these fixes in net8.0
> 
> That would be great!

Just to set expectations the leading edge will always be `main` here and I can help you test against that but we may have flexibility in servicing targeted changes (like this one)

jkotas:
I think it is unlikely that addition of WasmImportLinkageAttribute would be approved for servicing. The default bar for servicing is bug fixes only. WasmImportLinkageAttribute is a new feature.

lewing:
> I think it is unlikely that addition of WasmImportLinkageAttribute would be approved for servicing. The default bar for servicing is bug fixes only. WasmImportLinkageAttribute is a new feature.

As written this change Supports any Attribute named WasmLinkageAttribute

silesmo:
Hello!
Am I missing something when trying to use this with the nightly dotnet9 alpha?
I have installed the alpha and then when I try to install the wasi-experimental workload I get the following:
image

lewing:
> Hello! Am I missing something when trying to use this with the nightly dotnet9 alpha? I have installed the alpha and then when I try to install the wasi-experimental workload I get the following: image

cc @steveisok @radical

lewing:
> There have been several large changes recently (specifically around the aot cross compiler), we will take a look
> 
> cc @steveisok @radical

https://github.com/dotnet/runtime/pull/95170

PR #94623, opened 2023-11-11
JIT: Use post order computed by SSA in VN

jakobbotsch:
VN tries hard to dynamically compute a reverse post-order to visit the flow graph in. However, SSA has already computed such an order, so simply pass this along to VN instead.

A few positive diffs are expected. As [Andy has pointed out recently](https://github.com/dotnet/runtime/issues/93246#issuecomment-1787432449) the "dynamic RPO" VN was doing does not necessarily result in an actual RPO, so using SSA's order is expected to be a better order than what VN ends up with today.

### DIFF
--- a/src/coreclr/jit/fgopt.cpp
+++ b/src/coreclr/jit/fgopt.cpp
@@ -769,7 +769,6 @@ bool Compiler::fgRemoveDeadBlocks()
 // Notes:
 //   Each block's `bbPreorderNum` and `bbPostorderNum` is set.
 //   The `fgBBReversePostorder` array is filled in with the `BasicBlock*` in reverse post-order.
-//   This algorithm only pays attention to the actual blocks. It ignores any imaginary entry block.
 //
 //   Unreachable blocks will have higher pre and post order numbers than reachable blocks.
 //   Hence they will appear at lower indices in the fgBBReversePostorder array.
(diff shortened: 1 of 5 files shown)


src/coreclr/jit/fgopt.cpp
```diff
@@ -769,7 +769,6 @@ bool Compiler::fgRemoveDeadBlocks()
 // Notes:
 //   Each block's `bbPreorderNum` and `bbPostorderNum` is set.
 //   The `fgBBReversePostorder` array is filled in with the `BasicBlock*` in reverse post-order.
-//   This algorithm only pays attention to the actual blocks. It ignores any imaginary entry block.
```
jakobbotsch, line 772: I wasn't totally sure what this was referring to... but today we do ensure all blocks are visited below, so it seemed odd.

AndyAyersMS: SSA creates a stack-allocated BB0 when it computes dominators, I suspect this is the "imaginary entry block."

jakobbotsch: Hmm, might me. Still, not sure it is too relevant for this computation.

jakobbotsch:

Some minor positive [diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=466545&view=ms.vss-build-web.run-extensions-tab), both in codegen and TP.

AndyAyersMS (approved):
Nice.

We should think about how to safeguard against flowgraph changes... doesn't have to be in this PR but maybe you could use the bracketing "is safe to add" stuff I used when doing something similar for morph.

jakobbotsch:
> We should think about how to safeguard against flowgraph changes... doesn't have to be in this PR but maybe you could use the bracketing "is safe to add" stuff I used when doing something similar for morph.

Good idea, will think about how to reuse that.

cincuranet:
Possibly related improvement: https://github.com/dotnet/perf-autofiling-issues/issues/24588.

PR #94625, opened 2023-11-11
Add feature switch attribute design doc

sbomer:
Opening this as a draft for early review. I wanted to get a few eyes on this before making an API proposal.

edit: moved to https://github.com/dotnet/designs/pull/305

### DIFF
--- a/docs/design/tools/illink/feature-switch-attributes.md
+++ b/docs/design/tools/illink/feature-switch-attributes.md
@@ -1,0 +1,482 @@
+# Attribute-based model for feature switches
+
+.NET has [feature switches](https://github.com/dotnet/designs/blob/main/accepted/2020/feature-switches.md) which can be set to turn on/off areas of functionality in our libraries, with optional support for removing unused features when trimming or native AOT compiling. What we describe overall as "feature switches" have many pieces which fit together to enable this:
+
+- MSBuild property
+- RuntimHostConfigurationOption MSBuild ItemGroup
+- `runtimeconfig.json` setting
+- AppContext feature setting
+- **ILLink.Substitutions.xml**
+- **static boolean property**
+- **Requries attributes**
+
+The bold pieces are the focus of this document. [Feature switches](https://github.com/dotnet/designs/blob/main/accepted/2020/feature-switches.md) describes how settings flow from the MSBuild property through the AppContext (for runtime feature checks) or `ILLink.Substitutions.xml` (for feature settings baked-in when trimming). This document aims to describe an attribute-based model to replace some of the functionality currently implemented via ILLink.Substitutions.xml, used for branch elimination in ILLink and ILCompiler to remove branches that call into `Requires`-annotated code when trimming.
+
+## Background
+
+### Terminology
+
+We'll use the following terms to describe specific bits of functionality related to feature switches:
+- Feature switch property: the IL property whose value indicates whether a feature is enabled/supported
+  - For example: `RuntimeFeature.IsDynamicCodeSupported`
+- Feature switch name: the string that identifies a feature in `RuntimeHostConfigurationOption` and AppContext
+  - For example: `"System.Runtime.CompilerServices.RuntimeFeature.IsDynamicCodeSupported"`
+- Feature attribute: an attribute associated with a feature, used to annotate code that is directly related to the feature.
+  - For example: `RequiresDynamicCodeAttribute`
+- Feature guard property: an IL property whose value _depends_ on a feature being enabled, but isn't necessarily _defined_ by the availability of that feature.
+  - For example: `RuntimeFeature.IsDynamicCodeCompiled` depends on `IsDynamicCodeSupported`. It should return `false` when `IsDynamicCodeSupported` returns `false`, but it may return `false` even if `IsDynamicCodeSupported` is `true`.
+
+We'll say that a "feature switch property" is also necessarily a "feature guard property" for its defining feature, but not all "feature guard properties" are "feature switch properties".
+
+A "feature switch property" may also be a "feature guard property" for a feature _other than_ the defining feature. For example, we introduced the feature switch property `StartupHookProvider.IsSupported`, which is defined by the feature switch named `"System.StartupHookProvider.IsSupported"`, but additionally serves as a guard for code in the implementation that has `RequiresUnreferencedCodeAttribute`. Startup hook support is disabled whenever we trim out unreferenced code, but may alse be disabled independently by the feature switch.
+
+Similarly, one could imagine introducing a feature switch for `IsDynamicCodeCompiled`. 
+
+
+`IsDynamicCodeSupported` is an example of a feature that has an attribute, a feature switch, and a separate feature guard (`IsDynamicCodeCompiled`), but not all features have all of these bits of functionality.
+
+### Warning behavior
+
+Typically, feature switches come with support (via XML substitutions) for treating feature properties and feature guards as constants when publishing with ILLink/ILCompiler. These tools will eliminate guarded branches. This is useful as a code size optimization, and also as a way to prevent producing warnings for features that have attributes designed to produce warnings at callsites:
+
+```csharp
+UseDynamicCode(); // warns
+
+if (RuntimeFeature.IsDynamicCodeSupported)
+  UseDynamicCode(); // OK, no warning
+
+if (RuntimeFeature.IsDynamicCodeCompiled)
+  UseDynamicCode(); // OK, no warning in ILCompiler, but warns in analyzer
+
+[RequiresDynamicCode("Uses dynamically generated code")]
+static void UseDynamicCode() { }
+```
+
+The ILLink Roslyn analyzer has built-in support for treating `IsDynamicCodeSupported` as a guard for `RequiresDynamicCodeAttribute`, but has no other built-in support.
+
+## Goals
+
+- Teach the ILLink Roslyn analyzer to treat `IsDynamicCodeCompiled` as a guard for `RequiresDynamicCodeAttribute`
+- Allow libraries to define their own feature guard properties
+
+  Libraries should be able to introduce their own properties that can act as guards for `RequiresDynamicCodeAttribute`, or for other features that might produce warnings in the analyzer
+
+- Define an attribute-based model for such feature guards
+
+- Take into account how this would interact with an attribute-based model for feature switches
+
+  We will explore what an attribtute-based model for feature switches would look like to ensure that it interacts well with a model for feature guards. It's possible that we would design both in conjunction if they are naturally related.
+
+### Non-goals
+
+- Support branch elimination in the analyzer for all feature switches
+
+  The most important use case for the analyzer is analyzing libraries. Libraries typically don't bake in constants for feature switches, so the analyzer needs to consider all branches. It should support feature guards for features that produce warnings, but doesn't need to consider feature settings passed in from the project file to treat some branches as dead code.
+
+- Teach the ILLink Roslyn Analyzer about the substitution XML
+
+  We don't want to teach the analyzer to read the substitution XML. The analyzer is the first interaction that users typically have with trimming and AOT warnings. This should not be burdened by the XML format. Even if we did teach the analyzer about the XML, it would not solve the problem because the analyzer must not globally assume that `IsDynamicCodeSupported` is false as ILCompiler does.
+
+- Define a model with the full richness of the supported OS platform attributes
+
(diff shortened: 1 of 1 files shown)


docs/design/tools/illink/feature-switch-attributes.md
```diff
@@ -0,0 +1,482 @@
+# Attribute-based model for feature switches
+
+.NET has [feature switches](https://github.com/dotnet/designs/blob/main/accepted/2020/feature-switches.md) which can be set to turn on/off areas of functionality in our libraries, with optional support for removing unused features when trimming or native AOT compiling. What we describe overall as "feature switches" have many pieces which fit together to enable this:
```
gfoidl, line 3:
```suggestion
.NET has [feature switches](https://github.com/dotnet/designs/blob/main/accepted/2020/feature-switch.md) which can be set to turn on/off areas of functionality in our libraries, with optional support for removing unused features when trimming or native AOT compiling. What we describe overall as "feature switches" have many pieces which fit together to enable this:
```

docs/design/tools/illink/feature-switch-attributes.md
```diff
@@ -0,0 +1,482 @@
...
+
+- MSBuild property
+- RuntimHostConfigurationOption MSBuild ItemGroup
+- `runtimeconfig.json` setting
+- AppContext feature setting
+- **ILLink.Substitutions.xml**
+- **static boolean property**
+- **Requries attributes**
+
+The bold pieces are the focus of this document. [Feature switches](https://github.com/dotnet/designs/blob/main/accepted/2020/feature-switches.md) describes how settings flow from the MSBuild property through the AppContext (for runtime feature checks) or `ILLink.Substitutions.xml` (for feature settings baked-in when trimming). This document aims to describe an attribute-based model to replace some of the functionality currently implemented via ILLink.Substitutions.xml, used for branch elimination in ILLink and ILCompiler to remove branches that call into `Requires`-annotated code when trimming.
```
gfoidl, line 13:
```suggestion
The bold pieces are the focus of this document. [Feature switches](https://github.com/dotnet/designs/blob/main/accepted/2020/feature-switch.md) describes how settings flow from the MSBuild property through the AppContext (for runtime feature checks) or `ILLink.Substitutions.xml` (for feature settings baked-in when trimming). This document aims to describe an attribute-based model to replace some of the functionality currently implemented via ILLink.Substitutions.xml, used for branch elimination in ILLink and ILCompiler to remove branches that call into `Requires`-annotated code when trimming.
```

docs/design/tools/illink/feature-switch-attributes.md
```diff
@@ -0,0 +1,482 @@
...
+```
+
+The attribute definitions to support this might look like this:
+```csharp
+class FeatureGuardAttribute<T> : Attribute
+  where T : Attribute, IFeatureAttribute<T> {}
+
+interface IFeatureAttribute<TSelf> where TSelf : Attribute {
+    static abstract string FeatureName { get; }
+}
```
MichalStrehovsky, line 343: Could this be a custom attribute instead of an interface?

1. For non-execution scenarios it's an annoyance having to resolve virtual calls. The interface method could be implemented on a base, or by a default interface method and the analysis tool needs to deal with all of that just to get a string out. Custom attribute has it in the blob.
2. Static virtuals and generic attributes will not compile on netstandard2.0. Might not be a concern for trimming/AOT, but could be a concern for some other feature.

docs/design/tools/illink/feature-switch-attributes.md
```diff
@@ -0,0 +1,482 @@
...
+- Teach the ILLink Roslyn analyzer to treat `IsDynamicCodeCompiled` as a guard for `RequiresDynamicCodeAttribute`
+- Allow libraries to define their own feature guard properties
+
+  Libraries should be able to introduce their own properties that can act as guards for `RequiresDynamicCodeAttribute`, or for other features that might produce warnings in the analyzer
+
+- Define an attribute-based model for such feature guards
+
+- Take into account how this would interact with an attribute-based model for feature switches
+
+  We will explore what an attribtute-based model for feature switches would look like to ensure that it interacts well with a model for feature guards. It's possible that we would design both in conjunction if they are naturally related.
```
MichalStrehovsky, line 68: I agree that even if we don't actually implement the switch model, it would be great to have the design for it fully fleshed out. These feel very related.

docs/design/tools/illink/feature-switch-attributes.md
```diff
@@ -0,0 +1,482 @@
...
+```csharp
+class FeatureGuardAttribute<T> : Attribute
+  where T : Attribute, IFeatureAttribute<T> {}
+
+interface IFeatureAttribute<TSelf> where TSelf : Attribute {
+    static abstract string FeatureName { get; }
+}
+```
+
+We could use this model even for feature switches similar to `StartupHookSupport`, which don't currently have a `StartupHoopSupportAttribute`. There's no need to actually annotate related APIs with the attribute if that is overkill for a small feature. In this case the attribute definition would just serve as metadata that allows us to define a feature switch in a uniform way.
```
vitek-karas, line 346: On the surface this doesn't sound right to me - declaring an attribute which will do nothing when applied to something. 
On the other hand - it "might" work if we introduce the attribute as internal - in that case nobody would be able to apply it to anything anyway, but it would still serve the purpose as the ID of the feature.
Or similarly make the constructor of the attribute private -> same effect.

I'm still not sure I like this, but it would work.

One thing I do like about this - we're not using strings to identify things, we use IL metadata - which is nice.

docs/design/tools/illink/feature-switch-attributes.md
vitek-karas: +1 on this - as much as I would love to be able to use all the cool new things, the inability to compile this for `netstandard2.0` is pretty bad.
Using an attribute would effectively provide the same information. We would have to enforce its presence via analysis though.

docs/design/tools/illink/feature-switch-attributes.md
```diff
@@ -0,0 +1,482 @@
...
+
+This is fundamentally the same idea outlined in https://github.com/dotnet/designs/pull/261. The hope is that this document provides the motivation for using a unified representation for the attributes that are related to feature switches and trim/AOT analysis.
```
vitek-karas, line 348: Would be good to have a short section comparing the two specifically and possibly discussing a way to unify them (or even if we need that?)
Also some reasoning why we choose not to use the terminology from that PR instead of "feature" (note that "feature switch" is not actually encoded in any public API or MSBuild property - yet, it's just a concept, changing the terminology would not break anything)

vitek-karas, line 348: I would also like to see some discussion of what possibilities there are to implement this incrementally. For example, can we introduce feature guards alone - would it be a breaking change to add the rest later on?

docs/design/tools/illink/feature-switch-attributes.md
vitek-karas: The point are libraries which multi-target (our recommended approach). If such a library introduces its own feature guard/switch, it would need to use this. Some authors don't like ifdefs, so they may want to backfill the attributes, which would then not work for netstandard2.0. It's probably not very common, but still worth consideration.

The robustness and ease of the tooling is a bigger argument for me, and since this is not going to be common, we don't need the API to be super clean and super easy to use.

I don't feel strongly about this. I do agree with Michal's concerns though.

MichalStrehovsky: The interface approach introduces other class of robustness issues - there's nothing to enforce that the implementation of the method is a string literal, so we'd either need another analyzer for this to ensure string literal, or live with the fact that it might not be a string literal and have a way to report that.

Other thing is that the Roslyn analyzer wouldn't have visibility into the method body if it's in a different compilation unit.

docs/design/tools/illink/feature-switch-attributes.md
```diff
@@ -313,39 +311,115 @@ However, any attribute-based model that we pick to represent features should be
...
+    public static bool IsDynamicCodeCompiled => // ...
 }
 ```
 
 The attribute definitions to support this might look like this:
 ```csharp
-class FeatureGuardAttribute<T> : Attribute
-  where T : Attribute, IFeatureAttribute<T> {}
+[AttributeUsage(AttributeTargets.Class, Inherited = false)]
+abstract class FeatureAttribute : Attribute { }
```
vitek-karas, line 340: Worth mentioning - this would effectively provide a way to define custom `Requires...` attributes. Something we've discussed several times in the past. We should note that it's not part of this proposal to actually make that work. This will lead to somewhat weird place where if I define my own feature using these attributes, I would have the guard and everything work correctly, but I still would not get any warnings calling the `Requires..` annotated methods outside of the checked scope.

docs/design/tools/illink/feature-switch-attributes.md
```diff
@@ -0,0 +1,679 @@
...
+    APIWhichRequiresDynamicCode(); // No warnings
+}
+```
+
+This will silence warnings from the analyzer without any indication that `RequiresDynamicCode`-annotated code might be reached at runtime. We should be extremely clear in our documentation that this is not the intended use.
+
+### Validating correctness of feature guards
+
+Better would be for the analyzer to validate that the implementation of a feature check includes a check for `IsDynamicCodeSupported`. This could be done using the analyzer infrastructure we have in place without much cost for simple cases:
+
```
vitek-karas, line 166: There might be a need to do this type of analysis even in ILLink/ILCompiler, at least eventually. Especially if we start relying on `FeatureGuard` attributes in those tools. I don't think it's valid to rely on the analyzer for validating correctness of an app (even if for just a small portion of it). The one advantage here would be that the `FeatureGuard` attribute is introduced along side the corresponding analyzer support, but still.

docs/design/tools/illink/feature-switch-attributes.md
```diff
@@ -0,0 +1,679 @@
...
+  ```csharp
+  class Feature {
+      [FeatureGuard(typeof(RuntimeFeature), nameof(RuntimeFeature.IsDynamicCodeSupported))]
+      public static bool IsSupported => RuntimeFeature.IsDynamicCodeSupported;
+  }
+  ```
+
+  The analyzer would need to hard-code the fact that `RuntimeFeature.IsDynamicCodeSupported` corresponds to `RequiresDynamicCodeAttribute`, unless we model this relationship via attributes that represent feature switches.
+
+  This would be sufficient for ILLink/ILCompiler to treat `IsSupported` as a constant based on the feature switch name, assuming it has existing knowledge of the fact that `RuntimeFeature.IsDynamicCodeSupported` is contrelled by the feature switch named `"System.Runtime.CompilerServices.RuntimeFeature.IsDynamicCodeSupported"`, either from a substitution XML or from a separate attribute that encodes this.
```
vitek-karas, line 142:
```suggestion
  This would be sufficient for ILLink/ILCompiler to treat `IsSupported` as a constant based on the feature switch name, assuming it has existing knowledge of the fact that `RuntimeFeature.IsDynamicCodeSupported` is controlled by the feature switch named `"System.Runtime.CompilerServices.RuntimeFeature.IsDynamicCodeSupported"`, either from a substitution XML or from a separate attribute that encodes this.
```

docs/design/tools/illink/feature-switch-attributes.md
```diff
@@ -0,0 +1,679 @@
...
+
+For example, if we had a separate feature switch for `IsDynamicCodeCompiled`, this would allow setting `IsDynamicCodeCompiled` to `true` even when `IsDynamicCodeSupported` is `false`. This is essentially what we did for features like `StartupHookSupport`: this feature switch can be set even in trimmed apps.
+
+We rely on trim warnings to alert the app author to the problem, and also default `StartupHookSupport` to `false` for trimmed apps. If we have an attribute-based model for feature guards, we may want to consider inferring these defaults from the `FeatureGuard` instead (so a guard for `RequiresUnreferencedCode` that is also a feature switch would be `false` by default whenever "unreferenced code" is unavailable).
+
+The proposal for now is not to infer any defaults and just take care to set appropriate defaults in the SDK. This means that custom feature guards that are also feature switches will need to do the same. For example, a library that defines a feature switch which guards `RequiresUnreferencedCode` will need to ship with MSBuild targets that disable the feature by default for trimmed apps.
```
vitek-karas, line 251: I think we should add a sample for this - this part is probably the most complex topic in this document. Would be great to have a sample over startup hook, or something similar. How you tie a guard to a different feature switch and how it's supposed to work.

The defaults discussions is actually only important for illink/ilcompiler. The analyzer shouldn't really care as it needs to analyze both branches anyway. Also the correctness of this is not a problem because the "true" branch were the warnings are suppressed would still produce warnings in the app (illink/ilcompiler) if the feature switch is enabled. We basically need to be in a place where analyzer doesn't produce warnings in any case, but trimmer will produce them depending on the feature switch value.

PR #94626, opened 2023-11-11
FabricBot: Remove project board automation

jeffhandley:
We will be migrating away from FabricBot to a replacement service. This PR removes the automation configured around GitHub projects because:

* The new service does not support automation of GitHub projects
* We should move from Classic GitHub Projects to GitHub's New Projects
* FabricBot did not previously support New GitHub Projects
* GitHub's built-in project automation features should be used instead of relying on proprietary automation

1. @tommcdon / @hoyosjs
    - There was 6.0.0-specific automation for the [.NET Core Diagnostics](https://github.com/orgs/dotnet/projects/9) project
    - This automation is still in effect, but since it's specific to the 6.0.0 milestone, I presume it's safe to delete
   - If you're still using that project board, I recommend looking into GitHub's new projects
3. @mconnew
    - There was automation configured around the [Tellurium (WCF, Serialization, Etc.)](https://github.com/orgs/dotnet/projects/121) project
    - This uses GitHub's new projects
    - As far as I know, FabricBot never supported automation for GitHub's new projects; I suspect this automation never worked
5. @agocke
    - There was automation around the [Untriaged · AppModel](https://github.com/orgs/dotnet/projects/135) project
    - This uses GitHub's new projects
    - As far as I know, FabricBot never supported automation for GitHub's new projects; I suspect this automation never worked

_Edit: I don't know how I accidentally closed this PR; that was unintentional and it's reopened._


hoyosjs:
Removing the diagnostic automation sgtm

agocke (approved):

PR #94627, opened 2023-11-11
Expose the useful constant values in TimeSpan

### LINKED ISSUE #94545: [API Proposal]: Expose the remaining const values in TimeSpan
***Edited By @tarekgh***

This proposal has been approved before https://github.com/dotnet/runtime/issues/94545#issuecomment-2166531840. There was some feedback from @tannergooding and @IDisposable. 

@tannergooding said:  the concern I raised is that there are scenarios this can cause problems since `days * TimeSpan.HoursPerDay` could overflow. For example, if `days > 89,478,485` then this will overflow to a negative `int`.
`HoursPerDay` happens to be the only case where it is "sensible" for it to be `int`, since both `MaxDays` (`10,675,199`) and `MaxHours` (`256,204,778`) are within `int.MaxValue`. That is also why the signatures are `FromDays(int days, int hours, long minutes, long milliseconds, long microseconds)`. So other values, like `MinutesPerDay` (`1440`), `MinutesPerHour` (`60`), or even the existing `MillisecondsPerTick` (`10,000`) are all expected to be used in contexts where `long` is the target type.
When considering TimeSpan, which is where these constants are defined, hours and days are int, all other units are long.

---

### Background and motivation

In the [`TimeSpan`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs) class there are a few [`public const` values](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs#L59C1-L63C119) that are generally applicable that are usefully exposed.

There are *also* a bunch more that are [`internal const`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs#L65C1-L81C77) but would be generally useful for other code and are not more implementation driven than the already exposed const values.

**These values should be exposed publicly on the TimeSpan class.**

Note: NOT suggesting we expose the truly [`internal const`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs#L83C1-L106C77) values that are related to `MinTicks` and `MaxTicks`. They could be included but are just not as naturally useful elsewhere.

### API Proposal

```csharp

public class TimeSpan: // elided details
{
        // more elision

        public const long MicrosecondsPerMillisecond = TicksPerMillisecond / TicksPerMicrosecond; //           1,000
        public const long MicrosecondsPerSecond = TicksPerSecond / TicksPerMicrosecond;           //       1,000,000
        public const long MicrosecondsPerMinute = TicksPerMinute / TicksPerMicrosecond;           //      60,000,000
        public const long MicrosecondsPerHour = TicksPerHour / TicksPerMicrosecond;               //   3,600,000,000
        public const long MicrosecondsPerDay = TicksPerDay / TicksPerMicrosecond;                 //  86,400,000,000

        public const long MillisecondsPerSecond = TicksPerSecond / TicksPerMillisecond;           //           1,000
        public const long MillisecondsPerMinute = TicksPerMinute / TicksPerMillisecond;           //          60,000
        public const long MillisecondsPerHour = TicksPerHour / TicksPerMillisecond;               //       3,600,000
        public const long MillisecondsPerDay = TicksPerDay / TicksPerMillisecond;                 //      86,400,000

        public const long SecondsPerMinute = TicksPerMinute / TicksPerSecond;                     //              60
        public const long SecondsPerHour = TicksPerHour / TicksPerSecond;                         //           3,600
        public const long SecondsPerDay = TicksPerDay / TicksPerSecond;                           //          86,400

        public const long MinutesPerHour = TicksPerHour / TicksPerMinute;                         //              60
        public const long MinutesPerDay = TicksPerDay / TicksPerMinute;                           //           1,440

        public const long HoursPerDay = TicksPerDay / TicksPerHour;                               //              24
}
```

### API Usage

```csharp

if (TimeSpan.MinutesPerHour * numHours) > TimeSpan.MinutesPerDay * 7)
    Console.WriteLine("That's really weeks");
```

### Alternative Designs

N/A

### Risks

Since these are compile-time constant values, there would be no runtime cost in either memory or execution overhead. Additionally they would centralize basic real-world facts to one logical location (TimeSpan).

IDisposable:
Make the useful constants that TimeSpan has internally declare public because they're rooted in hard reality and not changeable, and quite useful elsewhere.
Fixes #94545


danmoseley:
Thanks, but the original issue needs to have the api-approved label before any PR. I suggest closing until then.

IDisposable:
That's why it's `Draft` :)

IDisposable:
Not sure why you would close a Draft... nor do I see how to reopen. That's not nice.

danmoseley:
@IDisposable We have many active PR's which is why we ask to get approval on new API before starting a PR.

As far as I know you can reopen it - if you can't I'm not sure why 🙂

tannergooding:
Notably, as per our [CONTRIBUTING.md](https://github.com/dotnet/runtime/blob/main/CONTRIBUTING.md)

> Please do not:
>
> ...
> * **DON'T** add API additions without filing an issue and discussing with us first. See [API Review Process](https://github.com/dotnet/runtime/blob/main/docs/project/api-review-process.md).

Every single PR opened has a cost, even for draft PRs. This comes in the form of e-mail notifications sent to developers, cross linking to the issues, time spent by developers who are giving the PR a preliminary glance, actual cost in terms of computing hours required to run tests and other validation, which in turn has downstream environmental impact and other considerations, etc.

There is nothing in this PR that requires it to be opened before the associated proposal goes to API review and so it is functionally noise until that occurs and it can opened if it gets approved.

PR #94631, opened 2023-11-11
JIT: Improve forwardsub for small types

EgorBo:
[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=466832&view=ms.vss-build-web.run-extensions-tab)

### DIFF
--- a/src/coreclr/jit/forwardsub.cpp
+++ b/src/coreclr/jit/forwardsub.cpp
@@ -521,7 +521,7 @@ bool Compiler::fgForwardSubStatement(Statement* stmt)
     // Bail if sub node has mismatched types.
     // Might be able to tolerate these by retyping.
     //
-    if (defNode->TypeGet() != fwdSubNode->TypeGet())
+    if (genActualType(defNode->TypeGet()) != genActualType(fwdSubNode->TypeGet()))
     {
         JITDUMP(" mismatched types (store)\n");
         return false;


EgorBo:
@jakobbotsch @SingleAccretion thanks for the suggestion for a better fix! Is there anything else needed here then? This should fix some of the regressions I hit in https://github.com/dotnet/runtime/pull/94621, but I want to land it separately to make that one zero-diff (ideally).

Diffs https://dev.azure.com/dnceng-public/public/_build/results?buildId=466832&view=ms.vss-build-web.run-extensions-tab

cc @dotnet/jit-contrib

jakobbotsch (approved):
LGTM -- please fix the PR title and commit message

jakobbotsch:
Should probably run jitstress and Fuzzlyn too.

EgorBo:
@jakobbotsch have you seen Fuzzlyn failures like this one https://dev.azure.com/dnceng-public/public/_build/results?buildId=466925&view=ms.vss-build-web.run-extensions-tab ?

jakobbotsch:
> @jakobbotsch have you seen Fuzzlyn failures like this one https://dev.azure.com/dnceng-public/public/_build/results?buildId=466925&view=ms.vss-build-web.run-extensions-tab ?

No, I haven't seen that before. I checked and it is hitting an assert with and without this change, so it is not related to this PR.
```scala
Assert failure(PID 15296 [0x00003bc0], Thread: 13572 [0x3504]): Assertion failed 'codeSize <= MAX_PROLOG_SIZE_BYTES' in 'Program:.cctor()' during 'Emit code' (IL size 10191; hash 0x02c770ad; FullOpts)

    File: C:\dev\dotnet\runtime\src\coreclr\jit\unwindarmarch.cpp Line: 834
    Image: H:\dotnet\runtime\artifacts\tests\coreclr\windows.arm64.Checked\Tests\Core_Root_base\corerun.exe
```
(Looks like Fuzzlyn should be looking at inner exceptions when it sees `System.TypeInitializationException` to be able to extract that assertion failure from it.)

cincuranet:
Regression in `System.Collections.ContainsTrueComparer<Int32>.FrozenSet`: https://github.com/dotnet/perf-autofiling-issues/issues/24578.

PR #94634, opened 2023-11-11
JIT: Document why the EH successor set is seemingly too large

jakobbotsch:
I spent some time today figuring out why we have these additional (and seemingly unnecessary) successors when iterating EH successors, and found these two reasons.

### DIFF
--- a/src/coreclr/jit/compiler.hpp
+++ b/src/coreclr/jit/compiler.hpp
@@ -553,21 +553,24 @@ static BasicBlockVisit VisitEHSuccessors(Compiler* comp, BasicBlock* block, TFun
 //   Whether or not the visiting should proceed.
 //
 // Remarks:
-//   Because we make the conservative assumption that control flow can jump
-//   from a try block to its handler at any time, the immediate (regular
-//   control flow) predecessor(s) of the first block of a try block are also
-//   considered to have the first block of the handler as an EH successor.
+//   The successors visited here are not true successors in the control flow
+//   sense -- it is not possible for control to transfer directly from the
+//   predecessor of a try to the try's handler. However, considering these as
+//   successors (and correspondingly, expanding the set of predecessors for
+//   handlers in BlockPredsWithEH) simplifies the handling in other places of
+//   the JIT:
 //
-//   As an example: for liveness this makes variables that are "live-in" to the
-//   handler become "live-out" for these try-predecessor block, so that they
-//   become live-in to the try -- which we require.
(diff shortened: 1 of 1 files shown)


src/coreclr/jit/compiler.hpp
```diff
@@ -553,21 +553,24 @@ static BasicBlockVisit VisitEHSuccessors(Compiler* comp, BasicBlock* block, TFun
...
-//   within the block T, including being a part of T's live-in set. That means
-//   that if T is the beginning of a try, then any predecessor of T will
-//   naturally also have E's live-in set as part of its live-out set.
+//   * It simplifies PHI arguments in handlers. In particular, when control is
+//   transferred to a handler, the reaching defs of live-in locals may come
+//   from predecessors of the try block if an exception was thrown before a def
+//   in the try. Without considering these as preds, we would have odd phi args
+//   (referring to blocks that aren't considered to be predecessors), or we
+//   would need to insert extra PHI definitions in the try block that the
+//   handler could refer to.
```
jakobbotsch, line 573: To be completely honest I think this alternative way of modelling things (adding "degenerate" phis at the entry `try` block) would be slightly more elegant.

AndyAyersMS: Interesting. 

Along those lines we could require the first (real) statement of the try be some sort of fake def -- GT_ENTERTRY? It would mess up some other things like empty block detection and try removal, but I expect we can fix those things.

That in turn might make it easier to allow trys with arbitrary or multiple entry points, since (perhaps) we'd no longer have to guarantee that the lexical try entry dominated all the blocks in the try. Each entry would just need one of these special statements...

jakobbotsch: Actually I don't think we need the degenerate phis... it is clearly not an uncommon scenario for a reaching def to be from some other block than the pred. So what I call out to be "odd" here is not actually odd.

As it turns out we already process the try heads specially, so I don't think my second reason described here exists. And it seems simple to generalize it to handle your case, we just need to generalize the following check to not be restricted to the head try block:

https://github.com/dotnet/runtime/blob/2454fe6785570d4b7d75e366289bcda2f4c363e4/src/coreclr/jit/ssabuilder.cpp#L1247

So that only leaves the dominator tree computation, as far as I can tell... seems like it would be more reasonable to handle it specially there instead of overapproximating the EH successor/predecessor sets for everyone. I'll probably give this a try and see if I hit any other snags.

jakobbotsch:
Superseded by #94672

PR #94640, opened 2023-11-11
UInt128: Throw DivideByZeroException when divisor is 0

### LINKED ISSUE #94593: {U}Int128 divided by 0 throws wrong exception type
Repro:
```C#
Int128 a = Int128.MaxValue;
Int128 b = 0;
Int128 result = a / b;
```
or
```C#
UInt128 a = UInt128.MaxValue;
UInt128 b = 0;
UInt128 result = a / b;
```

If `a` is set to a small value, like 42, this results in the expected `DivideByZeroException`. But when `a` is set to a large value, like `MaxValue`, it instead throws an ArgumentOutOfRangeException.

```
Unhandled exception. System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
   at System.UInt128.<op_Division>g__DivideSlow|110_2(UInt128 quotient, UInt128 divisor)
   at System.UInt128.op_Division(UInt128 left, UInt128 right)
   at System.Int128.op_Division(Int128 left, Int128 right)
```

manandre:
Fixes #94593

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/UInt128.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/UInt128.cs
@@ -1091,15 +1091,18 @@ namespace System
         /// <inheritdoc cref="IDivisionOperators{TSelf, TOther, TResult}.op_Division(TSelf, TOther)" />
         public static UInt128 operator /(UInt128 left, UInt128 right)
         {
-            if ((right._lower == 0) && (right._upper == 0))
+            if (right._upper == 0)
             {
-                throw new DivideByZeroException();
-            }
+                if (right._lower == 0)
+                {
+                    ThrowHelper.ThrowDivideByZeroException();
+                }
 
-            if ((right._upper == 0) && (left._upper == 0))
-            {
-                // left and right are both uint64
-                return left._lower / right._lower;
+                if (left._upper == 0)
+                {
+                    // left and right are both uint64
+                    return left._lower / right._lower;
+                }
             }
 
             if (right >= left)
--- a/src/libraries/System.Private.CoreLib/src/System/UInt128.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/UInt128.cs
@@ -1091,6 +1091,11 @@ namespace System
         /// <inheritdoc cref="IDivisionOperators{TSelf, TOther, TResult}.op_Division(TSelf, TOther)" />
         public static UInt128 operator /(UInt128 left, UInt128 right)
         {
+            if ((right._lower == 0) && (right._upper == 0))
+            {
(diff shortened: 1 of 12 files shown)


src/libraries/System.Private.CoreLib/src/System/UInt128.cs
```diff
@@ -1091,6 +1091,11 @@ public static UInt128 Log2(UInt128 value)
         /// <inheritdoc cref="IDivisionOperators{TSelf, TOther, TResult}.op_Division(TSelf, TOther)" />
         public static UInt128 operator /(UInt128 left, UInt128 right)
         {
+            if ((right._lower == 0) && (right._upper == 0))
+            {
+                throw new DivideByZeroException();
```
danmoseley, line 1096: Should this extract the throw info a helper method?

src/libraries/System.Private.CoreLib/src/System/UInt128.cs
```diff
@@ -1091,6 +1091,11 @@ public static UInt128 Log2(UInt128 value)
...
+            }
+
             if ((right._upper == 0) && (left._upper == 0))
```
danmoseley, line 1099: Is it better or worse to combine the check of right._upper == 0 with the one above?

skyoxZ: If we don't combine them, I guess putting the DevideByZero check after the both-int64 check could bring better performance.

src/libraries/System.Private.CoreLib/src/System/UInt128.cs
manandre: Yes, this is a good idea.

src/libraries/System.Private.CoreLib/src/System/UInt128.cs
manandre: Checks combined.
However, as a general rule, I prefer to keep input validation first, before routing the call to the `UInt64` divide operator.
@danmoseley Is the check order important here?

danmoseley: Area owners should answer that.

tannergooding: We should prefer throwing exceptions for "invalid" inputs first.

tannergooding (approved):

tannergooding:
There is a conflict that needs resolving before this can be merged.

manandre:
@tannergooding Commits rebased. The conflict was only about some test files moved.
See https://github.com/dotnet/runtime/pull/94644

danmoseley:
thanks @manandre for the contribution!

PR #94643, opened 2023-11-12
Fix for a stress crash in ComAwareWeakReferenceNative::HasInteropInfo

### LINKED ISSUE #94579: ExecutionEngineException in WinForms app (Paint.NET) on .NET 8 RC2, which does not occur in .NET 7 (GCStress)
### Description

I've had two spurious EEE's pop out at me while I was in the debugger over the last week, working on the new version of my app (Paint.NET v5.1) that is targeting .NET 8. I was worried it was due to my new code, some of which is clipboard related (always a bug farm) ... 

... But I'm now able to reproduce this 100% with my latest public release (Paint.NET v5.0.11, .NET 7.0.12), when it's retargeted for .NET 8. It has no trouble with .NET 7.

I believe this is a new bug in .NET 8, it's not due to any changes I've made in my code.

### Reproduction Steps

You will need access to https://github.com/dotnet/paint.net/ -- just ask and I will add you (for MSFT employees that is)

You will want this branch: https://github.com/dotnet/paint.net/tree/net8_EEE_repro . This is branched from v5.0.11, the latest public release using .NET 7, and then I've done the minimal amount of work to get it running on .NET 8 rc2 (mostly some nullable fixes that are already in my newer v5.1 branch)

1. You must have long file paths enabled for Windows, which may require a reboot, due to some source generators that end up using super long paths https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry
2. Load up ./paintdotnet.sln in Visual Studio 2022 17.8.0 Preview 7.0.
3. Due to something I haven't quite yet figured out with respect to project dependencies, you will first have to manually rebuild the AppVersion project. It's at Solution Explorer -> Misc (the first node) -> AppVersion
4. Now do a full rebuild
5. Verify that the `DOTNET_GCStress=2` environment variable is set up in the Debug properties
6. Debug menu -> Start Debugging
7. You will eventually (~20-60 seconds) see an `ExecutionEngineException` pop up _before_ the app's main window is displayed

### Expected behavior

App starts up just fine, albeit very slowly due to `DOTNET_GCStress=2`

### Actual behavior

App is not able to initialize, it dies with an EEE before the main window is displayed. On my system it's usually somewhere in the constructor for the `FileMenu` class, but there's some random variance as to exactly where it dies.

It usually won't even show a callstack in the debugger, and when it does it will not load symbols for the relevant .NET framework/runtime DLLs:

image

### Regression?

Yes. This works fine if you change `./paintdotnet/paintdotnet.csproj` to target .NET 7. Just remove the `<TargetFramework>` element at the top and do a full rebuild (sometimes need to do that twice after changing TargetFramework, as the first time will error due to what I think is a VS or msbuild bug). It will then revert to .NET 7 based on the values in `./TargetFramework.props`

### Known Workarounds

_No response_

### Configuration

Visual Studio 2022 17.8.0 Preview 7.0
x64
.NET 8 RC2

Happens on both my desktop:
Ryzen 9 7950X
Windows 11 23H2 build 22635.2700

and on my Lenovo P16 laptop:
Intel Core i7-12800HX
Windows 11 23H2 build 22631.2506

### Other information

cc @EgorBo who was commenting on this a little over on Discord, and who already has access to the dotnet/paint.net repository. Also @MichalPetryka who recommended me to file the issue

VSadov:
In rare cases it is possible for a syncblock to be collected in the time after we detected its presence and before we get into `ComAwareWeakReferenceNative::HasInteropInfo`, so we should not assume the syncblock is still there and must null-check before looking inside it. 

Fixes: https://github.com/dotnet/runtime/issues/94579

### DIFF
--- a/src/coreclr/vm/weakreferencenative.cpp
+++ b/src/coreclr/vm/weakreferencenative.cpp
@@ -172,8 +172,7 @@ FCIMPL1(FC_BOOL_RET, ComAwareWeakReferenceNative::HasInteropInfo, Object* pObjec
     _ASSERTE(pObject != nullptr);
 
     SyncBlock* pSyncBlock = pObject->PassiveGetSyncBlock();
-    _ASSERTE(pSyncBlock != nullptr);
-    return pSyncBlock->GetInteropInfoNoCreate() != nullptr;
+    return pSyncBlock != nullptr && pSyncBlock->GetInteropInfoNoCreate() != nullptr;
 }
 FCIMPLEND
 


VSadov:
```
     System.Net.Quic.Tests.MsQuicCipherSuitesPolicyTests.SupportedCipher_Success [FAIL]
      System.Net.Quic.QuicException : The connection timed out from inactivity.
```

The failure on musl appears to be https://github.com/dotnet/runtime/issues/91757

jkotas (approved):

VSadov:
Thanks!!

PR #94648, opened 2023-11-12
fixed doc comment copy mistake

SleepWellPupper:
On the `char.IsAsciiLetterUpper(Char c)` documentation comment, there is a mistake in the `return` tag. It conflicts the `summary` tags contents.


danmoseley (approved):
thank you!

PR #94663, opened 2023-11-13
Reenable runtime-richnav pipeline

akoeplinger:
The pipeline was disabled because the code indexing failed earlier, but it works now: https://dev.azure.com/dnceng-public/public/_build/results?buildId=467649&view=results


ViktorHofer (approved):

ericstj:
Do we want to enable this for 8.0 as well?

akoeplinger:
We can if necessary, we didn't really have a need for it so far.

PR #94664, opened 2023-11-13
Implement straightforward ServicePoint(Manager) properties in HttpWebRequest

liveans:
I'm going to create an issue with a list of properties that are not implemented. 
I'm still working on the `AllowWriteStreamBuffering` property, I'll create a follow-up PR for this.

Implemented properties from ServicePoint:
- TcpKeepAlive
- ReceiveBufferSize
- ContinueTimeout
- Expect100Continue

### DIFF
--- a/src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
+++ b/src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
@@ -1183,6 +1183,8 @@ namespace System.Net
                     request.Headers.ConnectionClose = true;
                 }
 
+                request.Headers.ExpectContinue = _servicePoint?.Expect100Continue ?? true;
+
                 request.Version = ProtocolVersion;
 
                 _sendRequestTask = async ?
--- a/src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
+++ b/src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
@@ -1183,7 +1183,10 @@ namespace System.Net
                     request.Headers.ConnectionClose = true;
                 }
 
-                request.Headers.ExpectContinue = _servicePoint?.Expect100Continue ?? true;
+                if (_servicePoint?.Expect100Continue == true)
+                {
+                    request.Headers.ExpectContinue = true;
+                }
 
                 request.Version = ProtocolVersion;
 
--- a/src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
+++ b/src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
@@ -1679,10 +1679,8 @@ namespace System.Net
                                 socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveInterval, parameters.ServicePoint.KeepAlive.Value.Interval);
                             }
                         }
-                        else
-                        {
-                            socket.NoDelay = true;
-                        }
+
+                        socket.NoDelay = true;
 
                         if (parameters.Async)
                         {
--- a/src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
+++ b/src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
@@ -166,7 +166,8 @@ namespace System.Net
                     && ReferenceEquals(Proxy, DefaultWebProxy)
                     && ServerCertificateValidationCallback == null
                     && ClientCertificates == null
-                    && CookieContainer == null;
+                    && CookieContainer == null
+                    && ServicePoint == null;
             }
         }
 
@@ -1665,18 +1666,18 @@ namespace System.Net
 
                     try
                     {
-                        if (parameters.ServicePoint is not null)
+                        if (parameters.ServicePoint is { } servicePoint)
                         {
-                            if (parameters.ServicePoint.ReceiveBufferSize != -1)
+                            if (servicePoint.ReceiveBufferSize != -1)
                             {
-                                socket.ReceiveBufferSize = parameters.ServicePoint.ReceiveBufferSize;
+                                socket.ReceiveBufferSize = servicePoint.ReceiveBufferSize;
                             }
 
-                            if (parameters.ServicePoint.KeepAlive is not null)
+                            if (servicePoint.KeepAlive is { } keepAlive)
                             {
                                 socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
-                                socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveTime, parameters.ServicePoint.KeepAlive.Time);
-                                socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveInterval, parameters.ServicePoint.KeepAlive.Interval);
+                                socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveTime, keepAlive.Time);
+                                socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveInterval, keepAlive.Interval);
                             }
                         }
 
--- a/src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
+++ b/src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
@@ -1687,8 +1687,6 @@ namespace System.Net
                             socket.NoDelay = true;
                         }
 
-                        socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.ReuseUnicastPort, parameters.ReusePort);
-
                         if (parameters.Async)
                         {
                             await socket.ConnectAsync(context.DnsEndPoint, cancellationToken).ConfigureAwait(false);
--- a/src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
+++ b/src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
@@ -114,7 +114,6 @@ namespace System.Net
             public readonly X509CertificateCollection? ClientCertificates;
             public readonly CookieContainer? CookieContainer;
             public readonly ServicePoint? ServicePoint;
-            public readonly bool ReusePort;
             public readonly TimeSpan ContinueTimeout;
 
             public HttpClientParameters(HttpWebRequest webRequest, bool async)
@@ -137,7 +136,6 @@ namespace System.Net
                 ClientCertificates = webRequest._clientCertificates;
                 CookieContainer = webRequest._cookieContainer;
                 ServicePoint = webRequest._servicePoint;
-                ReusePort = ServicePointManager.ReusePort;
                 ContinueTimeout = TimeSpan.FromMilliseconds(webRequest.ContinueTimeout);
             }
 
@@ -153,7 +151,6 @@ namespace System.Net
                     && Timeout == requestParameters.Timeout
                     && SslProtocols == requestParameters.SslProtocols
                     && CheckCertificateRevocationList == requestParameters.CheckCertificateRevocationList
-                    && ReusePort == requestParameters.ReusePort
                     && ContinueTimeout == requestParameters.ContinueTimeout
                     && ReferenceEquals(Credentials, requestParameters.Credentials)
                     && ReferenceEquals(Proxy, requestParameters.Proxy)
@@ -1670,8 +1667,10 @@ namespace System.Net
                     {
                         if (parameters.ServicePoint is not null)
                         {
-                            socket.NoDelay = !parameters.ServicePoint.UseNagleAlgorithm;
-                            socket.ReceiveBufferSize = parameters.ServicePoint.ReceiveBufferSize;
+                            if (parameters.ServicePoint.ReceiveBufferSize != -1)
+                            {
+                                socket.ReceiveBufferSize = parameters.ServicePoint.ReceiveBufferSize;
+                            }
 
                             if (parameters.ServicePoint.KeepAlive is not null)
                             {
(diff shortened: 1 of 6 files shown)


antonfirsov (commented):
I assume it's impossible to implement functional tests for this. How do you test/validate the changes?

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1660,7 +1668,26 @@ private static HttpClient CreateHttpClient(HttpClientParameters parameters, Http
...
+                            }
+
+                            // TODO (aaksoy): Implement support for ServicePoint.BindIPEndPointDelegate.
+                        }
+                        else
+                        {
+                            socket.NoDelay = true;
+                        }
+
+                        socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.ReuseUnicastPort, parameters.ReusePort);
```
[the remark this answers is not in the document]
liveans: Ah, in that case, this is not needed.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1660,7 +1668,26 @@ private static HttpClient CreateHttpClient(HttpClientParameters parameters, Http
 
                     try
                     {
-                        socket.NoDelay = true;
+                        if (parameters.ServicePoint is not null)
+                        {
+                            socket.NoDelay = !parameters.ServicePoint.UseNagleAlgorithm;
```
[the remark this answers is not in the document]
liveans: ServicePoint originally used the `Nagle Algorithm`. The default value for the `UseNagleAlgorithm` is `true`.
So, if we set this property to `false` underlying Socket will use the `Nagle Algorithm` as well. It's mostly because of consistency with the .NET Framework

liveans:
> I assume it's impossible to implement functional tests for this. How do you test/validate the changes?

I'm testing these options manually, but I'll try to write some tests for this.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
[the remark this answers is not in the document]
wfurt: I feel the `socket.NoDelay = true` should be default. Nagle is harmful to HTTP IMHO.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1660,7 +1668,24 @@ private static HttpClient CreateHttpClient(HttpClientParameters parameters, Http
...
+                            socket.ReceiveBufferSize = parameters.ServicePoint.ReceiveBufferSize;
+
+                            if (parameters.ServicePoint.KeepAlive is not null)
+                            {
+                                socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
+                                socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveTime, parameters.ServicePoint.KeepAlive.Value.Time);
+                                socket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.TcpKeepAliveInterval, parameters.ServicePoint.KeepAlive.Value.Interval);
+                            }
+
+                            // TODO (aaksoy): Implement support for ServicePoint.BindIPEndPointDelegate.
```
wfurt, line 1683: I don't think we should check-in TODOs

wfurt (commented):
generally looks good to me. 
We should perhaps add at least some get/set tests so we exercise the new code path.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1660,7 +1668,24 @@ private static HttpClient CreateHttpClient(HttpClientParameters parameters, Http
 
                     try
                     {
-                        socket.NoDelay = true;
+                        if (parameters.ServicePoint is not null)
+                        {
+                            socket.NoDelay = !parameters.ServicePoint.UseNagleAlgorithm;
+                            socket.ReceiveBufferSize = parameters.ServicePoint.ReceiveBufferSize;
```
ManickaP, line 1674: The behavior of both UseNagleAlgorithm and ReceiveBufferSize is slightly different in framework than what you have here, see https://referencesource.microsoft.com/#System/net/System/Net/ServicePoint.cs,319:
```c#
            // make this configurable from the user:
            if (!UseNagleAlgorithm) {
                finalSocket.SetSocketOption(SocketOptionLevel.Tcp, SocketOptionName.NoDelay, 1);
            }
            if (ReceiveBufferSize != -1) {
                finalSocket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.ReceiveBuffer, ReceiveBufferSize);
            }
```

Also without this condition, `socket.ReceiveBufferSize` will throw if user didn't set it, as the default is -1.

ManickaP: But I do agree with Tomas about Nagle, so maybe we should just ignore `UseNagleAlgorithm` or make sure we use it's `true` value only if it was explicitly set by the user and passed in as an unchanged default.

BTW, did any of the customers complained specifically about UseNagleAlgorithm?

ManickaP (commented):
+1 for the tests as well, some coverage would be nice.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1660,7 +1668,24 @@ private static HttpClient CreateHttpClient(HttpClientParameters parameters, Http
 
                     try
                     {
-                        socket.NoDelay = true;
+                        if (parameters.ServicePoint is not null)
```
stephentoub, line 1671: In what situation will the ServicePoint be null?

src/libraries/System.Net.Requests/src/System/Net/ServicePoint/ServicePointManager.cs
```diff
@@ -46,6 +46,8 @@ private static void ValidateSecurityProtocol(SecurityProtocolType value)
             }
         }
 
+        internal static TcpKeepAlive? KeepAlive { get; private set; }
```
stephentoub, line 49: statics and nullables generally don't mix well, as there can be race conditions in non-atomically reading the bool for whether there's a value and the actual value.  I suggest making this TcpKeepAlive a reference type instead.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
liveans: If nobody explicitly accesses to `ServicePoint` property, in case underlying `ServicePoint` field will be null, but we can always pass the property itself and forces it to find necessary `ServicePoint`.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
liveans: No, there is no customer who is complaining about the `UseNagleAlgorithm`. I'll remove it with the next commit.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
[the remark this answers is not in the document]
liveans: I left it as-is in that case.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -115,6 +113,8 @@ private sealed class HttpClientParameters
             public readonly RemoteCertificateValidationCallback? ServerCertificateValidationCallback;
             public readonly X509CertificateCollection? ClientCertificates;
             public readonly CookieContainer? CookieContainer;
+            public readonly ServicePoint? ServicePoint;
```
MihaZupan, line 116: We should likely be excluding parameters with a ServicePoint from caching (update `AreParametersAcceptableForCaching`), or we might limit the global cache to a single host.

On a side note, I'm not a fan of how we're doing client caching here overall, it's trivial to fall off a cliff from 100% to 0% client reuse. If we're looking at making WebRequest less bad, this is something we could look at improving.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1660,6 +1665,21 @@ private static HttpClient CreateHttpClient(HttpClientParameters parameters, Http
 
                     try
                     {
+                        if (parameters.ServicePoint is not null)
```
MihaZupan, line 1668: Nit: You can do something like
```c#
if (parameters.ServicePoint is { } servicePoint)

if (servicePoint.KeepAlive is { } keepAlive)
```

to avoid repeating the same `parameters.ServicePoint.KeepAlive.` expression.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2076,6 +2076,56 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+        public async Task SendHttpPostRequest_WithContinueTimeout_And_BodyWillSentByClient(bool expect100Continue, int continueTimeout)
+        {
+            await LoopbackServer.CreateServerAsync(
+            async (server, uri) =>
+            {
+                Task connection = server.AcceptConnectionAsync(async (client) => 
+                {
+                    await client.ReadRequestHeaderAsync();
+                    // This should not time out, because we're expecting the body itself and we should get it after 1 sec.
+                    string data = await client.ReadLineAsync().WaitAsync(TimeSpan.FromSeconds(3));
```
MihaZupan, line 2115: We may want to set the timeout here higher to avoid random failures in CI when machines are busy and even simple tasks can take a while to complete

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2076,6 +2076,56 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+                request.ContinueTimeout = 30000;
+                Stream requestStream = await request.GetRequestStreamAsync();
+                requestStream.Write("aaaa\r\n\r\n"u8);
+                var _ = request.GetResponseAsync();
+                await connection;
+            }).WaitAsync(TimeSpan.FromSeconds(40));
+        }
+
+        [Theory]
+        [InlineData(true, 1000)]
```
MihaZupan, line 2104:
```suggestion
        [InlineData(true, 1)]
```

antonfirsov (commented):
Product code LGTM, a few suggestions for the tests.

antonfirsov (approved):

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1178,6 +1183,8 @@ private async Task<WebResponse> SendRequest(bool async)
                     request.Headers.ConnectionClose = true;
                 }
 
+                request.Headers.ExpectContinue = _servicePoint?.Expect100Continue ?? true;
```
MihaZupan, line 1186:
```suggestion
                if (_servicePoint?.Expect100Continue == true)
                {
                    request.Headers.ExpectContinue = true;
                }
```

This would flip the default for Expect100Continue back to `true`.
AFAIR it used to be true on Framework, but not on Core.

Given WebRequest is considered obsolete, I think we should avoid making breaking changes here unless there is a good reason for it.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2076,6 +2077,95 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+            await LoopbackServer.CreateClientAndServerAsync(
+                async (uri) =>
+                {
+                    HttpWebRequest request = WebRequest.CreateHttp(uri);
+                    request.Method = "POST";
+                    request.ServicePoint.Expect100Continue = true;
+                    request.ContinueTimeout = 30000;
+                    Stream requestStream = await request.GetRequestStreamAsync();
+                    requestStream.Write("aaaa\r\n\r\n"u8);
+                    var _ = request.GetResponseAsync();
```
MihaZupan, line 2092: Can we avoid the fire-and-forget task here, something like `await Assert.ThrowsAnyAsync<Exception>() => request.GetResponseAsync());`?
(same for other tests below)

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2076,6 +2077,95 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+                {
+                    await server.AcceptConnectionAsync(async (client) => 
+                    {
+                        await client.ReadRequestHeaderAsync();
+                        // This should not time out, because we're expecting the body itself and we should get it after 1 sec.
+                        string data = await client.ReadLineAsync().WaitAsync(TimeSpan.FromSeconds(10));
+                        Assert.StartsWith("aaaa", data);
+                    });
+                }
+            ).WaitAsync(TestHelper.PassingTestTimeout);
```
MihaZupan, line 2132:
```suggestion
                });
```

Nit: When using `CreateClientAndServerAsync`, there is no need to set the timeout manually. The helper does that for you already.

MihaZupan (approved):

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
wfurt: It would be nice to clean all Tasks before test is done. Either as @MihaZupan suggested, store and await the Task later to using `.ContinueWith` as we do in some places in HTTP.

wfurt (approved):
LGTM % comments.
Good improvement.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
liveans: It was false on implementation, but it wasn't according to the documentation. Since we're using this property now, it will be set to `true` all the time, but I don't have any strong argument against what you suggested, but if we keep it as `false`. We need to update the [doc](https://learn.microsoft.com/en-us/dotnet/api/system.net.servicepoint.expect100continue?view=net-8.0) as well.

MihaZupan: My preference would be with sticking to the current default `false` as it matches that of `SocketsHttpHandler`. Updating the docs also seems like goodness since it sounds like they were misleading for a while already

liveans:
Outerloop failures unrelated

PR #94675, opened 2023-11-13
Disable CodeQL if we're not in the dedicated pipeline for it

akoeplinger:
Otherwise it gets injected automatically and slows down the build.


ViktorHofer (approved):

akoeplinger:
This is not tested by public CI, merging.

PR #94609, opened 2023-11-10
[release/8.0-staging] Update alpine arm32 in release 8.0 CI

CarnaViire:
Contributes to https://github.com/dotnet/runtime/issues/91757

Alpine docker images were rebuilt [on Oct 26](https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/923), which updated MsQuic to latest -- and brought in a regression for Alpine Arm32.

This PR rolls back to the previous image version that was used before Oct 26. This should stop the noise for release/8.0 that is https://github.com/dotnet/runtime/issues/91757. For main, I expect we would also update from 3.15 which is EOL to 3.16, so this PR is only for 8.0.

@carlossanlop this is test-infra-only change to reduce noise in CI.

Tested by running the tests in CI, see https://github.com/dotnet/runtime/pull/94609#issuecomment-1806588324

cc @wfurt @ManickaP @hoyosjs 

---

~~cc @wfurt - I took a look and alpine arm32 is only used in "checked runtime libraries test" combination. And in that case only 3.15 was used, which is EOL, so I believe we can just bump it to 3.17?~~

~~I was thinking and I'm not a fan of reverting to August's docker tag anymore... because it does not contain 10B fix (msquic 2.2.3)~~

PR #94683, opened 2023-11-13
Update alpine arm32 in CI

CarnaViire:
Contributes to https://github.com/dotnet/runtime/issues/91757

(Temporarily) pin Alpine Arm32 image to the previous tag -- to remove the noise in CI caused by MsQuic regression on Alpine Arm32 -- same change as in https://github.com/dotnet/runtime/pull/94609

cc @wfurt @ManickaP


ManickaP (approved):
Thanks.

wfurt (approved):

PR #90123, opened 2023-08-07
Add `emitDispEmbBroadcastCount` to display the embedded broadcasted element count.

Ruihan-Yin:
# Description 

JitDisasm doesn't include embedded broadcast syntax in disassembly.
 Eg.
`vaddps   zmm0, zmm0, dword ptr [reloc @RWD00]`

This PR will add the missing tag with the broadcasted element count, so the new disassembly will look like:

`vaddps   zmm0, zmm0, dword ptr [reloc @RWD00] {1to16}`

PR #87373, opened 2023-06-10
Make idSmallCns signed to allow including -1

tannergooding:
This makes `idSmallCns` signed on all platforms to allow including -1. This also means that all 8-bit values (signed or unsigned) can currently be tracked as small constants across all platforms. Doing this covers an additional 2% of constants, bringing the total number of small constants up to 94.33%

The [below](https://github.com/dotnet/runtime/pull/87373#issuecomment-1586197920) comment gives a comparison of the `before` vs `after` for crossgen of `S.P.Corelib`. There are various bits of interesting information that can be extrapolated, but the clearest is that 92% of all constants fit into an `int8_t`. There are then 4.5% of values that will never fit into a small constant (they take more than 16 bits) and so there's only about 1.2% of values that could fit into a small constant if we were given between `[11, 16]` bits to use.

This also includes some cleanup to the `EMITTER_STATS` handling/printing to ensure that scenarios are all being tracked.

PR #94684, opened 2023-11-13
Enable EVEX feature: Embedded Rounding for Avx512F.Add()

Ruihan-Yin:
## Description

Enabled EVEX feature: embedded rounding in JIT backend, and exposed only 1 related API:

 `public static Vector512<double> Add(Vector512<double> left, Vector512<double> right, FloatRoundingMode mode);`

 to establish the complete compiling path starting from importation. 

## Details

To tag the related AST node/instruction is embedded rounding enabled, we introduced some new flags:

1. For AST node, we introduced `GTF_HW_ER_*`: As the APIs are designed in a way that `FloatRoundingMode` information is passed as an extra operand, we choose to let the AST node carry this operand until `lowering`, and during lowering, the node will be converted back to the normal version, and the extra operand will be converted to a flag on the node. In this way, we can simply reuse the existing emit paths in the embedded rounding enabled cases.
2. When reaching the emit stage, we will have the corresponding instruction according to the node information, and we followed the design in Embedded Broadcast, used `insOpts` to carry this extra information to inform emitter embedded rounding is on, so in the `instrDesc`, we made the following changes: 
    1. renamed `_idEvexbContext` and extend it to 2 bit as `_idIsEmbBroadcast` and `_idIsEmbRounding`
    2. introduced 2 extra bits to indicate the rounding mode: `_idEmbRoundingMode`

## Follow-up in this PR
This PR, at current stage, is intended to show and discuss the design with maintainers, we are open to adjust and improve the designs. 

And we will need to add some unit tests to better cover this feature and related APIs.

## Follow-up after this PR
As this PR only exposed 1 embedded rounding related API, we will make another follow-up PR to expose other related APIs altogether. Since embedded rounding is mostly compatible with arithmetic and casting intrinsics, some more emit path will be impacted, we expect to extend the same design to different emit path if needed.


Ruihan-Yin:
Failures are known, will apply the format patch later.

Hi @tannergooding, I think we can start to have some discussion on the design of embedded rounding.  Would you please take a look at this draft? Thanks!

The impact on the throughput is less than I expected considering the extra bits I introduced in the `instrDesc`, is it attributed to the optimization given in #87373?

src/coreclr/jit/emit.h
```diff
@@ -779,8 +779,10 @@ class emitter
         unsigned _idCallAddr : 1; // IL indirect calls: can make a direct call to iiaAddr
         unsigned _idNoGC : 1;     // Some helpers don't get recorded in GC tables
 #if defined(TARGET_XARCH)
-        unsigned _idEvexbContext : 1; // does EVEX.b need to be set.
-#endif                                //  TARGET_XARCH
+        unsigned _idIsEmbBroadcast : 1; // does EVEX.b need to be set for embedded broadcast.
+        unsigned _idIsEmbRounding : 1;  // does EVEX.b need to be set for embedded rounding.
```
tannergooding, line 783: Do we need both of these?

My understanding is that `EVEX.b` can be either embedded broadcast or embedded rounding, but not both. Which it refers to then depends on whether it is `register, register` or `load+op`

Which is to say:
```
vaddps zmm0, zmm0, zmm0, {rd-sae}                    ; valid, use round down (toward -inf)
vaddps zmm0, zmm0, dword ptr [rax] {1to16}           ; valid, use embedded broadcast of 1 to 16 elements

vaddps xmm0, xmm0, xmm0, {rd-sae}                    ; invalid, embedded rounding is only valid for zmm and scalar
vaddps zmm0, zmm0, dword ptr [rax] {1to16}, {rd-sae} ; invalid, cannot combine embedded rounding and embedded broadcast
```

Notably, .NET already doesn't support IEEE 754 exceptions (setting `MXCSR.*M` is undefined behavior) and does not support the rounding mode being globally changed from the default (setting MXCSR.RC is undefined behavior), so we can additionally optimize out `{sae}` and `{rne-sae}` and just emit the regular instruction, so I think we actually only need `1` additional bit

That is, given the restrictions .NET has around MXCSR, we could have the following:
```cpp
if (insDesc->idHasMem())
{
    // 0b00 - no embedded broadcast
    // 0b01 - embedded broadcast
    // 0b10 - invalid
    // 0b11 - invalid
}
else
{
    // 0b00 - no embedded rounding
    // 0b01 - {rd-sae}, invalid for EA_ATTR 16/32 only usable for ZMM and scalar on supported instructions
    // 0b10 - {ru-sae}, invalid for EA_ATTR 16/32 only usable for ZMM and scalar on supported instructions
    // 0b11 - {rz-sae}, invalid for EA_ATTR 16/32 only usable for ZMM and scalar on supported instructions
}
```

I expect at least one of the other bits might be "unused" for SIMD register to register scenarios, so we probably don't even need to reduce the smallCns space here. We would however have to identify and ensure it is actually unused and won't cause any other code paths to be accidentally taken.

src/coreclr/jit/emit.h
```diff
@@ -1547,11 +1547,34 @@ class emitter
...
+        void idSetEvexbContext(insOpts instOptions)
+        {
+            if (instOptions == INS_OPTS_EVEX_b)
+            {
+                _idEvexbContext = 1;
+            }
+            else if (instOptions == INS_OPTS_EVEX_er_rd)
+            {
+                _idEvexbContext = 1;
+            }
```
tannergooding, line 1560: Is it worth aliasing these options to remove a branch? That is, make `INS_OPTS_EVEX_b` and `INS_OPTS_EVEX_er_rd` be the "same"?

Which it is ultimately comes down to whether a memory operand is contained

src/coreclr/jit/emitxarch.cpp
```diff
@@ -1309,18 +1332,43 @@ emitter::code_t emitter::AddEvexPrefix(const instrDesc* id, code_t code, emitAtt
...
     else if (attr == EA_64BYTE)
     {
-        // Set L' bits to 11 in case of instructions that operate on 512-bits.
+        // Set L' bits to 10 in case of instructions that operate on 512-bits.
         code |= LPRIMEBIT_IN_BYTE_EVEX_PREFIX;
     }
 
     if (id->idIsEvexbContext())
     {
+
```
tannergooding, line 1346: nit: stray newline
```suggestion
```

src/coreclr/jit/emitxarch.cpp
```diff
@@ -1309,18 +1332,43 @@ emitter::code_t emitter::AddEvexPrefix(const instrDesc* id, code_t code, emitAtt
...
+                code |= LPRIMEBIT_IN_BYTE_EVEX_PREFIX;
+                code &= ~(LBIT_IN_BYTE_EVEX_PREFIX);
+            }
+            else
+            {
+                // {rz-sae}
+                code |= LPRIMEBIT_IN_BYTE_EVEX_PREFIX;
+                code |= LBIT_IN_BYTE_EVEX_PREFIX;
+            }
+        }
```
tannergooding, line 1371: Should we assert that the EVEX.B context is `1` in the `else` case? (Since anything else should have been invalid)

src/coreclr/jit/lowerxarch.cpp
```diff
@@ -1658,6 +1658,48 @@ GenTree* Lowering::LowerHWIntrinsic(GenTreeHWIntrinsic* node)
             LowerHWIntrinsicCC(node, NI_AVX_PTEST, GenCondition::UGT);
             break;
 
+        // This branch will take care of the embedded rounding compatible binary nodes.
+        case NI_AVX512F_Add:
```
tannergooding, line 1662: I think we may want to make this table driven via a new flag.

The rules are fairly static (works for scalars and for V512) and we have a large number of instructions it will end up applying towards.

src/tests/JIT/HardwareIntrinsics/X86_Avx512/Avx512F/EmbeddedRounding.Double.cs
```diff
@@ -0,0 +1,50 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
tannergooding, line 1: We likely want to make this a new test template if possible for the same reasons as above. We have a large number of instructions this applies to and we really want to be adding basic tests for most of them.

Actually testing that two arbitrary floats/doubles are added and round correctly is likely non-trivial, however. I don't think this test is validating that so much as its just showing it doesn't explode for such inputs.

Given its testing `ToEven`, it's not really testing the new code paths either. That is, its not validating anything around the embedded rounding actually being used.

src/coreclr/jit/emitxarch.cpp
Ruihan-Yin: Do you mean arrange the branches like:
```C#
if (id->idIsEvexbContext())
    {

        code |= EVEX_B_BIT;

        if (!id->idHasMem())
        {
            // embedded rounding case.
            unsigned roundingMode = id->idGetEvexbContext();
            if (roundingMode == 1)
            {
                // {rd-sae}
                code &= ~(LPRIMEBIT_IN_BYTE_EVEX_PREFIX);
                code |= LBIT_IN_BYTE_EVEX_PREFIX;
            }
            else if (roundingMode == 2)
            {
                // {ru-sae}
                code |= LPRIMEBIT_IN_BYTE_EVEX_PREFIX;
                code &= ~(LBIT_IN_BYTE_EVEX_PREFIX);
            }
            else if (roundingMode == 3)
            {
                // {rz-sae}
                code |= LPRIMEBIT_IN_BYTE_EVEX_PREFIX;
                code |= LBIT_IN_BYTE_EVEX_PREFIX;
            }
            else
            {
                // I am a bit unclear why we should assert Evex.b context is 1 here, 
                // other than the cases above, everything else should be invalid, 
                // so putting a `unreached()` might make more sense?
                // As we defined _idEvexbContext - 2 bits:
                // 00 - no embedded broadcast/rounding
                // 01 - embedded broadcast / round down
                // 10 - round up
                // 11 - round to zero
            }
        }
    }
```

src/tests/JIT/HardwareIntrinsics/X86_Avx512/Avx512F/EmbeddedRounding.Double.cs
Ruihan-Yin: Yes, this test was initially to put a draft here which just tests the compiling path instead of the results. I was having some difficulty to manually calculate the expected results under difference rounding mode settings.

src/coreclr/jit/emit.h
```diff
@@ -885,8 +885,8 @@ class emitter
 
         ////////////////////////////////////////////////////////////////////////
         // Small constant size (with/without prev offset, assuming host==target):
-        // x86:         11/15 bits
-        // amd64:       10/15 bits
+        // x86:         8/12 bits
+        // amd64:       7/12 bits
```
tannergooding, line 889: Shouldn't this be the following, since we're only using 1 extra bit now:
```suggestion
        // x86:         10/14 bits
        // amd64:       9/14 bits
```

src/coreclr/jit/emit.h
```diff
@@ -1547,11 +1547,30 @@ class emitter
         {
             return _idEvexbContext != 0;
         }
-        void idSetEvexbContext()
+
+        void idSetEvexbContext(insOpts instOptions)
```
tannergooding, line 1551: Do we want to preserve the `assert(_idEvexbContext == 0);` to ensure its not being set twice?

src/coreclr/jit/emitxarch.cpp
tannergooding: Basically what you did, but also covering the case where `id->idHasMem()` is true, so...
```cpp
        if (!id->idHasMem())
        {
            // embedded rounding case.
            unsigned roundingMode = id->idGetEvexbContext();
            if (roundingMode == 1)
            {
                // {rd-sae}
                code &= ~(LPRIMEBIT_IN_BYTE_EVEX_PREFIX);
                code |= LBIT_IN_BYTE_EVEX_PREFIX;
            }
            else if (roundingMode == 2)
            {
                // {ru-sae}
                code |= LPRIMEBIT_IN_BYTE_EVEX_PREFIX;
                code &= ~(LBIT_IN_BYTE_EVEX_PREFIX);
            }
            else if (roundingMode == 3)
            {
                // {rz-sae}
                code |= LPRIMEBIT_IN_BYTE_EVEX_PREFIX;
                code |= LBIT_IN_BYTE_EVEX_PREFIX;
            }
            else
            {
                unreached();
            }
        }
        else
        {
            assert(id->idGetEvexbContext() == 0);
        }
```

src/coreclr/jit/emitxarch.cpp
```diff
@@ -6847,10 +6898,10 @@ void emitter::emitIns_R_R_A(
     id->idIns(ins);
     id->idReg1(reg1);
     id->idReg2(reg2);
-    if (instOptions == INS_OPTS_EVEX_b)
+    if (instOptions == INS_OPTS_EVEX_eb_er_rn)
     {
         assert(UseEvexEncoding());
-        id->idSetEvexbContext();
+        id->idSetEvexbContext(instOptions);
     }
```
tannergooding, line 6905: Should we assert that none of the other options like `INS_OPTS_EVEX_er_rd` are set in the `else` case?

tannergooding: Same goes for other cases. Maybe we could have a general helper like `SetEvexBroadcastIfNeeded(id, instOptions)` and it can basically do:
```cpp
if (instOptions == INS_OPTS_EVEX_eb_er_rn)
{
    assert(UseEvexEncoding());
    id->idSetEvexbContext(instOptions);
}
else
{
    assert(instOptions == 0);
}
```

or similar...

src/coreclr/jit/emitxarch.cpp
```diff
@@ -6847,10 +6898,10 @@ void emitter::emitIns_R_R_A(
     id->idIns(ins);
     id->idReg1(reg1);
     id->idReg2(reg2);
-    if (instOptions == INS_OPTS_EVEX_b)
+    if (instOptions == INS_OPTS_EVEX_eb_er_rn)
```
tannergooding, line 6901: Does this need to be `(instOptions & SomeMask) == INS_OPTS_EVEX_eb_er_rn` (or some helper to extract the relevant bits, etc)

src/coreclr/jit/hwintrinsic.h
```diff
@@ -203,6 +203,9 @@ enum HWIntrinsicFlag : unsigned int
 
     // The intrinsic is an embedded broadcast compatiable intrinsic
     HW_Flag_EmbBroadcastCompatible = 0x8000000,
+
+    // The intrinsic is an embedded rounding compatiable intrinsic
```
tannergooding, line 207:
```suggestion
    // The intrinsic is an embedded rounding compatible intrinsic
```

src/coreclr/jit/emitxarch.cpp
tannergooding: -- Just thinking from the perspective that `options` generally implies more than one can be set, so it may be relevant to add masking now so it isn't an issue later

src/coreclr/jit/hwintrinsic.h
```diff
@@ -587,6 +590,27 @@ struct HWIntrinsicInfo
...
+
+    static size_t EmbRoundingArgPos(NamedIntrinsic id)
+    {
+        // This helper function returns the expected position,
+        // where the embedded rounding control argument should be.
+        assert(IsEmbRoundingCompatible(id));
+        switch (id)
+        {
+            case NI_AVX512F_Add:
+                return 3;
```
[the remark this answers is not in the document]
tannergooding: If it isn't, we should be able to effectively use `node->GetOperandCount() - 1` instead.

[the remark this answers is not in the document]
Ruihan-Yin: This acts mostly like a table which records the expected number of operands when embedded rounding is activated, as we need an indication to tell if this feature is enabled or it can be just lowered as usual. 

The reason I add this is that as now `Avx512F.Add()` can take 2 or 3 inputs, 1 extra input as `FloatRoundingMode`, on the table in `hwintrinsiclistxarch.h`, we put `-1` at `NumArg` to accommodate multiply choices on input numbers. So we don't have how many args are expected in most of the cases, and as we will be handling some type casting and scalar intrinsics, I suppose only checking if the type of the last argument is constant int might not enough.

 In our case, `node->GetOperandCount()` would dynamically return 2 or 3 based on the number of inputs, and we might need some information to tell the difference.

Ruihan-Yin:
Hi @tannergooding,

As I was trying to come up with a way to establish the unit test template for these embedded rounding APIs, C++ is actually able to manipulate the rounding mode setting globally using some std APIs (Please check the codes below), so I would suppose I can compute the expected outputs of a given inputs set offline, and transfer them into .net unit tests as some constant value tables.

I wonder if this would be a cheap solution to the unit test issue.

```C++
#include <cfenv>
#include <iostream>
#include <immintrin.h>  // Include Intel Intrinsics header
#pragma fenv_access(on)

int main() {
    // Save the current rounding mode
    auto originalRound = std::fegetround();

    // Initialize two arrays of double values
    double array1[8] = { 0.05 , 0.10, 0.15, 0.20, -0.25, -0.30, -0.35, -0.40 };
    double array2[8] = { 0.45 , 0.40, 0.35, 0.30, -0.25, -0.20, -0.15, -0.10 };

    __m512d vector1 = _mm512_loadu_pd(array1);
    __m512d vector2 = _mm512_loadu_pd(array2);

    // Set the rounding mode to desired value (e.g., FE_DOWNWARD)
    std::fesetround(FE_DOWNWARD);

    // Perform floating-point operations with the new rounding mode
    __m512d result = _mm512_add_pd(vector1, vector2);

    // Restore the original rounding mode when done
    std::fesetround(FE_TONEAREST);

    // Cast the result vector to __m512i
    __m512i resultInt = _mm512_castpd_si512(result);

    // Store the result as uint64_t
    alignas(64) uint64_t resultArray[8];
    _mm512_store_si512((__m512i*)resultArray, resultInt);

    // Print the result
    std::cout << "Round up results:" << std::endl;
    for (int i = 0; i < 8; ++i) {
        std::cout << std::hex <<resultArray[i] << std::endl;
    }
    std::cout << std::endl;

    return 0;
}
```

tannergooding:
I think that would be fine. Notably if you're getting correct results from C you could also just call `_mm512_add_round_pd`

The real goal here is to just minimally assert the correct instruction is emitted, for which the current easiest solution is to execute a simple test and validate the output.

In a more ideal world we might be able to easily assert that `Avx512.Add(x, y, RoundingMode...)` caused `vaddps zmm0, zmm0, zmm0, {...-sae}` to be emitted and that executing the code doesn't throw. -- @TIHan added some support for validating emitted assembly, but I don't know how usable it is in this scenario

Ruihan-Yin:
> In a more ideal world we might be able to easily assert that `Avx512.Add(x, y, RoundingMode...)` caused `vaddps zmm0, zmm0, zmm0, {...-sae}` to be emitted and that executing the code doesn't throw. -- @TIHan added some support for validating emitted assembly, but I don't know how usable it is in this scenario

Thanks @tannergooding!
Hi @TIHan, would you please point me to the PR Tanner was referring to? Essentially, we want a way to check if EVEX prefix is properly set.

BruceForstall:
@Ruihan-Yin Here is a page describing writing tests that check the generated disassembly: https://github.com/dotnet/runtime/blob/main/docs/workflow/testing/coreclr/disasm-checks.md

(fyi, most people will be out until the new year)

Ruihan-Yin:
> @Ruihan-Yin Here is a page describing writing tests that check the generated disassembly: https://github.com/dotnet/runtime/blob/main/docs/workflow/testing/coreclr/disasm-checks.md
> 
> (fyi, most people will be out until the new year)

Hi @BruceForstall, thanks for the notice, we are working internally on the new test template for all the new embedded rounding APIs, we would expect to have a separate PR including exposing the rest of the APIs and all the related unit tests. Hopefully, we can have an external PR shortly. And we can catch up after the new year. :)

tannergooding:
Looks like this is pending some formatting fixes and tests, is that correct?

Ruihan-Yin:
> Looks like this is pending some formatting fixes and tests, is that correct?

Rebased the branch, running the pipeline tests again. I'll push the generated format patch shortly.

For the tests, we are working internally on templates to cover all the embedded rounding APIs, I can push a commit with the template and cover all the `Add` unit tests, so we can review the test design in this PR, and I would expect to have a separate PR to expose the rest of the APIs and unit tests.

Will the plan sound good to you?

Ruihan-Yin:
Updates:
1. Added the unit test template for embedded rounding APIs.
2. Added a jump table fallback for non-constant rounding mode control byte.
3. removed hand-written unit tests.

The expected test results were calculated on i9-9980XE with a [C++ program](https://github.com/Ruihan-Yin/EmbRoundingAnswerGenerator).

tannergooding (approved):
CC. @dotnet/jit-contrib for secondary review

src/coreclr/jit/emitxarch.cpp
```diff
@@ -10657,6 +10708,37 @@ void emitter::emitDispEmbBroadcastCount(instrDesc* id)
...
+{
+    if (!id->idIsEvexbContext())
+    {
+        return;
+    }
+    assert(!id->idHasMem());
+    unsigned roundingMode = id->idGetEvexbContext();
+    if (roundingMode == 1)
+    {
+        printf(" {rd-sae}");
```
BruceForstall, line 10726: nit: in the Intel manual I have, section "15.6.4 Static Rounding Mode and Suppress All Exceptions", there are these examples:
```
vaddps zmm7 {k6}, zmm2, zmm4, {rd-sae}
vmulps zmm7 {k6}, zmm2,[rax], {rd-sae}
vaddps ymm7 {k6}, ymm2, ymm4,{rd-sae}
```

In each case there is a comma before the `{rd-sae}`.

It's not clear from the instruction set reference whether the comma is expected or not. So what is the standard? Comma or no comma?

src/coreclr/jit/gentree.h
```diff
@@ -555,6 +555,12 @@ enum GenTreeFlags : unsigned int
 
     GTF_MDARRLOWERBOUND_NONFAULTING = 0x20000000, // GT_MDARR_LOWER_BOUND -- An MD array lower bound operation that cannot fault. Same as GT_IND_NONFAULTING.
 
+    GTF_HW_ER_MASK              = 0x30000000, // Bits used by handle types below 
+    GTF_HW_ER_TOEVEN            = 0x00000000, // GT_HWINTRINSIC -- embedded rounding mode: ToEven (Default).
+    GTF_HW_ER_TONEGATIVEINFINITY = 0x10000000, // GT_HWINTRINSIC -- embedded rounding mode: ToNegativeInfinity.
+    GTF_HW_ER_TOPOSITIVEINFINITY = 0x20000000, // GT_HWINTRINSIC -- embedded rounding mode: ToPositiveInfinity.
+    GTF_HW_ER_TOZERO            = 0x30000000, // GT_HWINTRINSIC -- embedded rounding mode: ToZero.
```
BruceForstall, line 562: A few notes: 
1. please align the `=`
2. Add underscores between words (otherwise it's to easy to read "TOE...", "TONE...", "TOP..."
```suggestion
    GTF_HW_ER_MASK              = 0x30000000, // Bits used by handle types below 
    GTF_HW_ER_TO_EVEN            = 0x00000000, // GT_HWINTRINSIC -- embedded rounding mode: FloatRoundingMode = ToEven (Default) "{rn-sae}"
    GTF_HW_ER_TO_NEGATIVE_INFINITY = 0x10000000, // GT_HWINTRINSIC -- embedded rounding mode: FloatRoundingMode = ToNegativeInfinity "{rd-sae}"
    GTF_HW_ER_TO_POSITIVE_INFINITY = 0x20000000, // GT_HWINTRINSIC -- embedded rounding mode: FloatRoundingMode = ToPositiveInfinity "{ru-sae}"
    GTF_HW_ER_TO_ZERO            = 0x30000000, // GT_HWINTRINSIC -- embedded rounding mode: FloatRoundingMode = ToZero "{rz-sae}"
```

src/coreclr/jit/lowerxarch.cpp
```diff
@@ -1055,6 +1055,70 @@ GenTree* Lowering::LowerHWIntrinsic(GenTreeHWIntrinsic* node)
 
     NamedIntrinsic intrinsicId = node->GetHWIntrinsicId();
 
+    if (HWIntrinsicInfo::IsEmbRoundingCompatible(intrinsicId))
+    {
+
```
BruceForstall, line 1060: nit: unnecessary extra line
```suggestion
```

src/coreclr/jit/lowerxarch.cpp
```diff
@@ -1055,6 +1055,70 @@ GenTree* Lowering::LowerHWIntrinsic(GenTreeHWIntrinsic* node)
...
+                    embRoundingNode->SetUnusedValue();
+                }
+                BlockRange().Remove(node);
+                BlockRange().Remove(lastOp);
+                node = embRoundingNode;
+                if (mode != 0x08)
+                {
+                    // As embedded rounding can only work under register-to-register form, we can skip contain check at
+                    // this
+                    // point.
```
BruceForstall, line 1108:
```suggestion
                    // this point.
```

src/coreclr/jit/gentree.h
```diff
@@ -2224,6 +2230,41 @@ struct GenTree
...
     }
 
+#ifdef FEATURE_HW_INTRINSICS
+
+    void ClearEmbRoundingMode()
+    {
+        assert(gtOper == GT_HWINTRINSIC);
+        gtFlags &= ~GTF_HW_ER_MASK;
+    }
+    void SetEmbRoundingMode(uint8_t mode)
```
BruceForstall, line 2240: Can you please add function comments? In particular, it would be nice to know that `mode` is one of the values from System.Runtime.Intrinsics.X86.FloatRoundingMode.

src/coreclr/jit/gentree.h
```diff
@@ -2224,6 +2230,41 @@ struct GenTree
...
+                gtFlags |= GTF_HW_ER_TONEGATIVEINFINITY;
+                break;
+            case 0x0A:
+                gtFlags |= GTF_HW_ER_TOPOSITIVEINFINITY;
+                break;
+            case 0x0B:
+                gtFlags |= GTF_HW_ER_TOZERO;
+                break;
+            default:
+                break;
```
BruceForstall, line 2256: should this be:
```suggestion
            case 0:
                break;
            default:
                unreached();
```
?

BruceForstall (changes requested):
Some small nits/questions/requests

src/coreclr/jit/emit.h
```diff
@@ -774,13 +774,13 @@ class emitter
         unsigned _idCallAddr : 1; // IL indirect calls: can make a direct call to iiaAddr
         unsigned _idNoGC : 1;     // Some helpers don't get recorded in GC tables
 #if defined(TARGET_XARCH)
-        unsigned _idEvexbContext : 1; // does EVEX.b need to be set.
-#endif                                //  TARGET_XARCH
+        unsigned _idEvexbContext : 2 // Does Evex.b need to be set for embedded broadcast/embedded rounding.
```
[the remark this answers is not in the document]
Ruihan-Yin: Thanks for the comments! 

`_idEvexbContext` is used to differentiate several features: normal intrinsics, embedded broadcast, embedded rounding. In the normal or embedded broadcast cases, the semantic for `Evex.L'L` remains the same, intrinsic vector length, while in embedded rounding case, the semantic changes to indicate the rounding mode. `_idEvexbContext` will be used to inform emitter to specially handle the `Evex.L'L` in the embedded rounding scenario. I will properly document this part in the comments.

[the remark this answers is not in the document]
BruceForstall: A comment like you just wrote would be great.

src/coreclr/jit/gentree.h
Ruihan-Yin: I think it is fine to go with no comma.

we adopted the no-comma format in embedded broadcast (#90123), and also [MSVC](https://godbolt.org/z/dsjxvox9s) adopted the same format.

src/coreclr/jit/emitxarch.cpp
Ruihan-Yin: Sorry commented in the wrong place, please see [comment](https://github.com/dotnet/runtime/pull/94684#discussion_r1456669338)

src/coreclr/jit/gentree.h
Ruihan-Yin: The new test failures come from this change:
When we call the embedded rounding APIs in the reflection call, JIT will generate a [jump table](https://github.com/dotnet/runtime/blob/main/src/coreclr/jit/hwintrinsiccodegenxarch.cpp#L1054) during the emit stage, this jump table is essentially a switch case containing all the possible results under different entry values, say in this case 0\~11 (by the existing design, the jump table will iterate from 0 to the max value we set, say here 11), so even this function is supposed to accept 8\~11 only, there are case when it might take 0~7. 
Based on the given information, do we consider revert the changes, or do we want to modify the jump table generation logic a bit?

p.s. I made a mistake setting the upper bound to be 12, that will take a small fix.

BruceForstall:
@Ruihan-Yin Looks like there are some test failures:
```
21:53:02.094 Running test: _Avx512F_r::JIT.HardwareIntrinsics.X86._Avx512F.Program.AddDoubleToNegativeInfinity()
Beginning scenario: RunBasicScenario_UnsafeRead
Beginning scenario: RunBasicScenario_Load
Beginning scenario: RunBasicScenario_LoadAligned
Beginning scenario: RunReflectionScenario_UnsafeRead

Assert failure(PID 73252 [0x00011e24], Thread: 73252 [0x11e24]): Assertion failed 'unreached' in 'System.Runtime.Intrinsics.X86.Avx512F:Add(System.Runtime.Intrinsics.Vector512`1[double],System.Runtime.Intrinsics.Vector512`1[double],ubyte):System.Runtime.Intrinsics.Vector512`1[double]' during 'Generate code' (IL size 9; hash 0x01102930; FullOpts)

    File: /__w/1/s/src/coreclr/jit/gentree.h Line: 2261
    Image: /datadisks/disk1/work/AFB708FB/p/corerun
```

https://dev.azure.com/dnceng-public/public/_build/results?buildId=532443&view=ms.vss-test-web.build-test-results-tab
https://dev.azure.com/dnceng-public/public/_build/results?buildId=532448&view=ms.vss-test-web.build-test-results-tab

Ruihan-Yin:
Yes, please see my [comment](https://github.com/dotnet/runtime/pull/94684#discussion_r1458125034) here.

src/coreclr/jit/gentree.h
Ruihan-Yin: It looks okay to revert the changes, and let this function accept those unexpected values. 

For reference, by the current design, `Avx2.GatherMaskVector128` is also generating a jump table containing all the results under entry values from 0~8, while this intrinsic only expect 1,2,4,8. Seems, those unexpected value shall be blocked out on the language level, say we don't define values other than those in the System.Runtime.Intrinsics.X86.FloatRoundingMode Enum. 

And for the generated results with the unexpected values would be not "harmful" as they will be the results under the default rounding mode.

BruceForstall: I see, so the generic jump table code generates options for all values `[0..maxValue]` for some API-specific maximum value, but some APIs (like this one) have a much more specific, smaller set of expected values. Seems ok to leave your code as it was -- maybe document (comment) why the switch allows "unexpected" values.

BruceForstall (approved):

Ruihan-Yin:
Thanks for all the reviews and help!

PR #86778, opened 2023-05-25
JIT: expand qmarks earlier

AndyAyersMS:
Move qmark expansion to happen at the end of the importation phases. That way we don't have to model the messy semantics of qmarks in analysis and optimization phases.

For now the downstream bits that dealt with qmarks are left intact. They will be removed later.

PR #94690, opened 2023-11-14
JIT: consistently handle no return calls in qmarks

AndyAyersMS:
When we expand QMARKS, ensure that any block with a no-return call gets changed to BBJ_THROW.

This fixes a case I am seeing with cross-block local assertion prop, as the upper QMARK gets optimized away and so we don't check if the expansing has any noreturn calls.

It also happens in places with just within-block local assertion prop.

Contributes to #94363.

### DIFF
--- a/src/coreclr/jit/compiler.h
+++ b/src/coreclr/jit/compiler.h
@@ -4892,8 +4892,8 @@ public:
     Statement* fgNewStmtFromTree(GenTree* tree, const DebugInfo& di);
 
     GenTree* fgGetTopLevelQmark(GenTree* expr, GenTree** ppDst = nullptr);
-    void fgExpandQmarkForCastInstOf(BasicBlock* block, Statement* stmt);
-    void fgExpandQmarkStmt(BasicBlock* block, Statement* stmt);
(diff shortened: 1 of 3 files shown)


AndyAyersMS:
@dotnet/jit-contrib fyi
@jakobbotsch PTAL

A handful of diffs expected (a few more with cross-block enabled, but still not a lot).

Ideally we'd tail merge these, but the noreturn calls often end up with different arguments, and tail merge / throw helper merge aren't able to handle these cases (yet).

jakobbotsch (approved):
LGTM. Is the `fgExpandQmarkForCastInstOf` special case still necessary?

jakobbotsch:
I mean the entirety of `fgExpandQmarkForCastInstOf`... it was never clear to my why the normal qmark expansion wouldn't be sufficient, apart from this difference in handling of no-return calls.

AndyAyersMS:
Ah... I can look at this. Maybe as part of revisiting #86778.

AndyAyersMS:
Will revisit some of the above when I look at moving qmark expansion earlier.

PR #94696, opened 2023-11-14
Extract common CodeAnalysis rules into separate file

akoeplinger:
Only rules differing between CodeAnalysis.src.globalconfig and CodeAnalysis.test.globalconfig are left in the respective files.

### DIFF
--- a/docs/project/analyzers.md
+++ b/docs/project/analyzers.md
@@ -8,7 +8,7 @@ To add an analyzer package to the build:
     ```XML
     <PackageReference Include="SonarAnalyzer.CSharp" Version="8.50.0.58025" PrivateAssets="all" />
     ```
-3. After that point, all builds will employ all rules in that analyzer package that are enabled by default.  You can change the severity for rules by adding entries to the globalconfig files [CodeAnalysis.src.globalconfig](https://github.com/dotnet/runtime/blob/main/eng/CodeAnalysis.src.globalconfig) or [CodeAnalysis.test.globalconfig](https://github.com/dotnet/runtime/blob/main/eng/CodeAnalysis.test.globalconfig).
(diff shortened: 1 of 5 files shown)


ViktorHofer (approved):
Approving but it would be good to wait for @stephentoub to also ack the change.

artl93:
If I want to change a rule that applies one way to tests and another to sources, would I now need to make the change in three places? (I.e., remove global, add source, add test?)

stephentoub:
Can you speak more to the motivation for this change?

One of the more common changes we make here is to enable an analyzer that we previously didn't have enabled.  Before that would mean just changing one word in the src config.  Now it'll mean deleting some lines from the common, adding them back to the test config, and adding them back to the src config mutated.

When new analyzers come online, we also today typically add them into both the src and test file, into src as warning and into test as disabled.  With the new approach, we're still modifying those two files.

I'm not fundamentally against the change, I just don't understand what it buys us.

akoeplinger:
> If I want to change a rule that applies one way to tests and another to sources, would I now need to make the change in three places? (I.e., remove global, add source, add test?)

Yeah

> Can you speak more to the motivation for this change?

I noticed that a bunch of the settings have corresponding entries both in the CodeAnalysis file and the root .editorconfig and  according to the [precedence rules](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files#precedence) the one in .editorconfig wins.

This seemed a bit confusing to me so I thought if we first extract the common rules then we can later deduplicate/remove the rules which are already set by .editorconfig.

I don't feel strongly about this though so if you think this is not worth it then I'm happy to close the PR :)

akoeplinger:
Closing since it looks like there's not enough benefit to warrant this change.

PR #94701, opened 2023-11-14
JIT: Optimize data flow used in assertion prop/CSE

jakobbotsch:
The data flow used by assertion prop and CSE utilize "all-succs" which includes handler successors. However, in reality neither analysis needs full treatment of handlers; we do not CSE variables that are live-in to handlers, and assertion prop never kills assertions that would need to be propagated to handlers.

Additionally, the data flow framework used conflates end-of-block propagation of facts with reachability of the handler. If no facts changed at the ends of the preds of the handler, then the handler is not visited. This is despite the fact that the handler in the general case is also reachable with facts from the beginning of every enclosed basic block (or more generally with facts that were invalidated in the middle of enclosed basic blocks).

This ends up working out today only because of the fake successor edges we have from preds of the try to the handler, in addition to the restrictions on the data flows described above. Since I'm removing these fake edges, we need a more explicit solution. The change here has such a solution, by making it more explicit what exactly is needed in the data flow for CSE and AP here: they only need to consider regular successors, with the added catch that they need to consider reachability of handlers once we see the corresponding 'try' being reachable.

Prerequisite to #94672.

No diffs are expected.

### DIFF
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -5493,10 +5493,9 @@ public:
     //   try region.
     //
     //   It suffices to intersect with only the head 'try' block's assertions,
-    //   since that block dominates all other blocks in the try and we never
-    //   kill assertions in global AP. Note that if we did kill assertions we
-    //   would need to be more careful about our mid-block handling when in a
-    //   try region.
+    //   since that block dominates all other blocks in the try, and since
+    //   assertions are VN-based and can never become false.
+    //
     void MergeHandler(BasicBlock* block, BasicBlock* firstTryBlock, BasicBlock* lastTryBlock)
     {
         if (VerboseDataflow())
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -5488,8 +5488,15 @@ public:
     //   lastTryBlock  - the last block of the try for "block" handler;.
     //
     // Notes:
-    //   We can jump to the handler from any instruction in the try region.
-    //   It means we can propagate only assertions that are valid for the whole try region.
+    //   We can jump to the handler from any instruction in the try region. It
+    //   means we can propagate only assertions that are valid for the whole
+    //   try region.
+    //
+    //   It suffices to intersect with only the head 'try' block's assertions,
+    //   since that block dominates all other blocks in the try and we never
+    //   kill assertions in global AP. Note that if we did kill assertions we
+    //   would need to be more careful about our mid-block handling when in a
+    //   try region.
     void MergeHandler(BasicBlock* block, BasicBlock* firstTryBlock, BasicBlock* lastTryBlock)
     {
         if (VerboseDataflow())
@@ -5498,11 +5505,8 @@ public:
             Compiler::optDumpAssertionIndices("in -> ", block->bbAssertionIn, "; ");
             JITDUMP("firstTryBlock " FMT_BB " ", firstTryBlock->bbNum);
             Compiler::optDumpAssertionIndices("in -> ", firstTryBlock->bbAssertionIn, "; ");
-            JITDUMP("lastTryBlock " FMT_BB " ", lastTryBlock->bbNum);
-            Compiler::optDumpAssertionIndices("out -> ", lastTryBlock->bbAssertionOut, "\n");
         }
         BitVecOps::IntersectionD(apTraits, block->bbAssertionIn, firstTryBlock->bbAssertionIn);
-        BitVecOps::IntersectionD(apTraits, block->bbAssertionIn, lastTryBlock->bbAssertionOut);
     }
 
     // At the end of the merge store results of the dataflow equations, in a postmerge state.
--- a/src/coreclr/jit/dataflow.h
+++ b/src/coreclr/jit/dataflow.h
@@ -58,7 +58,7 @@ void DataFlow::ForwardAnalysis(TCallback& callback)
         }
         else
         {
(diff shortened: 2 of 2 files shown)


src/coreclr/jit/assertionprop.cpp
```diff
@@ -5498,11 +5505,8 @@ class AssertionPropFlowCallback
             Compiler::optDumpAssertionIndices("in -> ", block->bbAssertionIn, "; ");
             JITDUMP("firstTryBlock " FMT_BB " ", firstTryBlock->bbNum);
             Compiler::optDumpAssertionIndices("in -> ", firstTryBlock->bbAssertionIn, "; ");
-            JITDUMP("lastTryBlock " FMT_BB " ", lastTryBlock->bbNum);
-            Compiler::optDumpAssertionIndices("out -> ", lastTryBlock->bbAssertionOut, "\n");
         }
         BitVecOps::IntersectionD(apTraits, block->bbAssertionIn, firstTryBlock->bbAssertionIn);
-        BitVecOps::IntersectionD(apTraits, block->bbAssertionIn, lastTryBlock->bbAssertionOut);
```
jakobbotsch, line 5505: This was pretty meaningless -- the lexically last block of a try region shouldn't be sufficient to guarantee any form of correctness.

jakobbotsch:

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=469223&view=ms.vss-build-web.run-extensions-tab) -- only a few on x86, that I investigated above. Other than that just some nice throughput improvements.

src/coreclr/jit/assertionprop.cpp
```diff
@@ -5488,8 +5488,15 @@ class AssertionPropFlowCallback
...
-    //   We can jump to the handler from any instruction in the try region.
-    //   It means we can propagate only assertions that are valid for the whole try region.
+    //   We can jump to the handler from any instruction in the try region. It
+    //   means we can propagate only assertions that are valid for the whole
+    //   try region.
+    //
+    //   It suffices to intersect with only the head 'try' block's assertions,
+    //   since that block dominates all other blocks in the try and we never
+    //   kill assertions in global AP. Note that if we did kill assertions we
+    //   would need to be more careful about our mid-block handling when in a
```
SingleAccretion, line 5498: >    Note that if we did kill assertions...

In my opinion, it is confusing to talk about the possibility of "killing assertions" in global AP. How would that work? It seems impossible almost as if by definition of the algorithm.

src/coreclr/jit/dataflow.h
```diff
@@ -58,19 +58,47 @@ void DataFlow::ForwardAnalysis(TCallback& callback)
         }
         else
         {
-            FlowEdge* preds = m_pCompiler->BlockPredsWithEH(block);
-            for (FlowEdge* pred = preds; pred; pred = pred->getNextPredEdge())
+            for (FlowEdge* pred = block->bbPreds; pred; pred = pred->getNextPredEdge())
```
SingleAccretion, line 61: Nit: can this loop use the `PredEdges` iterator?

src/coreclr/jit/dataflow.h
```diff
@@ -58,19 +58,47 @@ void DataFlow::ForwardAnalysis(TCallback& callback)
...
+            // that block dominates all other blocks in the 'try'. That happens
+            // implicitly below.
+            //
+            block->VisitRegularSuccs(m_pCompiler, [&worklist](BasicBlock* succ) {
                 worklist.insert(worklist.end(), succ);
                 return BasicBlockVisit::Continue;
             });
         }
+
+        if (m_pCompiler->bbIsTryBeg(block))
```
SingleAccretion, line 86: Why do we not need to consider the exceptional successors of filters here?

(Not saying it should)

jakobbotsch: Do you mean the second pass EH successors?

src/coreclr/jit/assertionprop.cpp
jakobbotsch: Do you mean because we use SSA/VN? I mean if facts could become false in other ways than due to control flow joins. We would need to propagate that to the handlers.

src/coreclr/jit/dataflow.h
SingleAccretion: Yes.

src/coreclr/jit/assertionprop.cpp
SingleAccretion: > Do you mean because we use SSA/VN?

Right. VNs are "immutable", so if a a fact about a VN dominates a subgraph, it will always be true on the entirety of that subgraph. I understand the comment, I would just request to reword to not mention AP along the lines of "no clients need the concept of 'kill's".

src/coreclr/jit/dataflow.h
jakobbotsch: I think that because the filter and its enclosed handlers are all dominated by a same try entry there is no need to consider the filter as a pred of those handlers.

src/coreclr/jit/assertionprop.cpp
jakobbotsch: Sure, I can reword it tomorrow.

src/coreclr/jit/dataflow.h
jakobbotsch: Yeah, let me simplify this tomorrow.

src/coreclr/jit/assertionprop.cpp
jakobbotsch: Do you think the wording looks better now?

jakobbotsch:
Ping @AndyAyersMS... when you have time.

AndyAyersMS (approved):

PR #94710, opened 2023-11-14
[main] Fixed DateTimeOffset.Now calls on Android with specific conditions

### LINKED ISSUE #94709: DateTimeOffset.Now throws exception on Android with Arabic language & western hemisphere timezone
### Description

There's an issue with Android implementation of `DateTimeOffset.Now`, that requires specific preconditions.
Issue happens in [this file on line 63](https://github.com/dotnet/runtime/blob/v8.0.0/src/libraries/System.Private.CoreLib/src/System/DateTimeOffset.Android.cs#L63), where we call `Convert.ToInt32(localDateTimeOffset)` method without providing a specific `IFormatProvider` value.

Thus, it uses default culture provider, which is arabic. In arabic culture, negative numbers pattern has a minus sign at the end, i.e. number `-2000` which is negative two thousands should be written as `2000-` for arabic culture in order to be converted/parsed. 

However, in [monodroid.c file on line 256](https://github.com/dotnet/runtime/blob/v8.0.0/src/tasks/AndroidAppBuilder/Templates/monodroid.c#L256) it uses "%d" C syntax in order to store a negative integer, which is not culture-aware and always places negative sign at the beginning.

As a result, we're getting ThrowOverflowOrFormatException, if we will call `DateTimeOffset.Now` property at the beginning of life of our Android application.

### Reproduction Steps

Preconditions:
1. Set locale on your Android smartphone to Arabic
2. Set timezone on your Android smartphone to some negative value (somewhere in western hemisphere, e.g. Phoenix, US)

Steps to reproduce:
1. Write a simple NET7/NET8 Android application where you'll call DateTimeOffset.Now at the very beginning (e.g. in your SplashActivity OnCreate method).
2. Run your application

### Expected behavior

App will execute without crashes, DateTimeOffset.Now will return a proper result

### Actual behavior

App crashes with ThrowOverflowOrFormatException

### Regression?

As far as I know, Xamarin.Android didn't have such issue

### Known Workarounds

Call `_ = TimeZoneInfo.Local;` in order to preload AndroidTZData
However, it will cause app slowdown in start which was the very point of making android-specific implementation of DateTimeOffset.Now property.

Fix actually should be very easy, [on line 63 here](https://github.com/dotnet/runtime/blob/v8.0.0/src/libraries/System.Private.CoreLib/src/System/DateTimeOffset.Android.cs#L63), we have to change the code from
`int localDateTimeOffsetSeconds = Convert.ToInt32(localDateTimeOffset);`
to
`int localDateTimeOffsetSeconds = Convert.ToInt32(localDateTimeOffset, CultureInfo.InvariantCulture);`

### Configuration

Versions of .NET: 7.0, 8.0
OS: Android
OS language: Arabic
OS Timezone: Phoenix, US (GMT-07:00)

### Other information

_No response_

Saratsin:
Fixed issue when DateTimeOffset.Now throws exception on Android with Arabic language & western hemisphere timezone
Fixes #94709

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/DateTimeOffset.Android.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/DateTimeOffset.Android.cs
@@ -1,8 +1,8 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.Threading;
 using System.Globalization;
+using System.Threading;
 
 namespace System
 {
--- a/src/libraries/System.Private.CoreLib/src/System/DateTimeOffset.Android.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/DateTimeOffset.Android.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Threading;
+using System.Globalization;
 
 namespace System
 {
@@ -60,7 +61,7 @@ namespace System
                 }
 
                 // Fast path obtained offset incorporated into ToLocalTime(DateTime.UtcNow, true) logic
-                int localDateTimeOffsetSeconds = Convert.ToInt32(localDateTimeOffset);
+                int localDateTimeOffsetSeconds = Convert.ToInt32(localDateTimeOffset, CultureInfo.InvariantCulture);
                 TimeSpan offset = TimeSpan.FromSeconds(localDateTimeOffsetSeconds);
                 long localTicks = utcDateTime.Ticks + offset.Ticks;
                 if (localTicks < DateTime.MinTicks || localTicks > DateTime.MaxTicks)


Saratsin:
I think that the same PR should be done to net7.0 in order to fix it there, but can please someone tell me what branch should I use in order to such PR?

mdh1418 (approved):
Looks good to me! Thank you for the catch and contribution!

Typically for backporting to previous releases, we make a `/backport to <branch>` comment. This will signal the github-actions bot to create the relevant PR which we would need to fill in Impact, Testing, and Risk of the change.

In this case, if we want to backport to 7.0 and 8.0, I believe we would want to backport to `release/7.0-staging` and `release/8.0-staging`

steveisok (approved):
@Saratsin, I appreciate the contribution! I'll drive getting this fix backported.

PR #94732, opened 2023-11-14
Add ConstructorBuilder implementation, integrate TypeBuilder.GetXyz() static methods

buyaa-n:
Constributes to https://github.com/dotnet/runtime/issues/92975
- Add `ConstructorBuilder` implementation
- Add `DefineConstructorCore` and `DefineDefaultConstructorCore` protected members for `TypeBuilderImpl`
-  Integrate `TypeBuilder.GetConstructor(...)`, `TypeBuilder.GetMethod(...)` and `TypeBuilder.GetField(...)` static methods so that they could be used for the new `Reflection.Emit.TypeBuilderImpl`
- Fix/implement other APIs that needed for accessing/writing types and methods constructed from runtime type and `TypeBuilderImpl` type (instances of `*BuilderInstantiation` types, like `MethodBuilderInstantiation`)
- Populate type and its members tokens on `CreateTypeInfoCore()`. Refactor Assembly.Save IL emit logic to require referenced types to be created ("baked").


src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/TypeBuilderImpl.cs
```diff
@@ -79,9 +81,101 @@ protected override void AddInterfaceImplementationCore([DynamicallyAccessedMembe
...
+        }
+
+        protected override ConstructorBuilder DefineConstructorCore(MethodAttributes attributes, CallingConventions callingConvention, Type[]? parameterTypes, Type[][]? requiredCustomModifiers, Type[][]? optionalCustomModifiers)
+        {
+            if ((_attributes & TypeAttributes.Interface) == TypeAttributes.Interface && (attributes & MethodAttributes.Static) != MethodAttributes.Static)
+            {
+                throw new InvalidOperationException(SR.InvalidOperation_ConstructorNotAllowedOnInterface);
+            }
+
+            ThrowIfTheTypeCreated();
```
jkotas, line 118: Should this be called in other methods too, to match runtime builder?

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/TypeBuilderImpl.cs
```diff
@@ -79,9 +81,101 @@ protected override void AddInterfaceImplementationCore([DynamicallyAccessedMembe
...
+                ((_attributes & (TypeAttributes.Abstract | TypeAttributes.Sealed)) != (TypeAttributes.Abstract | TypeAttributes.Sealed)))
+            {
+                DefineDefaultConstructor(MethodAttributes.Public);
+            }
+
+            _isCreated = true;
+            return this;
+        }
+
+        private void ThrowIfTheTypeCreated()
```
jkotas, line 103:
```suggestion
        private void ThrowIfCreated()
```
It is the usual naming convention for these methods in the repo. (E.g. look for `ThrowIfDisposed` or `ThrowIfNull`)

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/TypeBuilderImpl.cs
```diff
@@ -79,9 +81,101 @@ protected override void AddInterfaceImplementationCore([DynamicallyAccessedMembe
...
+            }
+
+            // Create a public default constructor if this class has no constructor. Except the type is Interface, ValueType, Enum, or a static class.
+            // (TypeAttributes.Abstract | TypeAttributes.Sealed) determines if the type is static
+            if (_constructorDefinitions.Count == 0 && (_attributes & TypeAttributes.Interface) == 0 && !IsValueType &&
+                ((_attributes & (TypeAttributes.Abstract | TypeAttributes.Sealed)) != (TypeAttributes.Abstract | TypeAttributes.Sealed)))
+            {
+                DefineDefaultConstructor(MethodAttributes.Public);
+            }
+
```
jkotas, line 98: Do we also need to populate method and field tokens here? (It can wait for follow up PR.)

The runtime and .NET Framework impl let you access the tokens of baked fields and methods.

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
```diff
@@ -697,6 +697,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)System\Reflection\Emit\TypeBuilder.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)System\Reflection\Emit\TypeBuilderInstantiation.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)System\Reflection\Emit\TypeNameBuilder.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)System\Reflection\Emit\TypeBuilder.Static.cs" Condition="'$(FeatureNativeAot)' != 'true'" />
```
jkotas, line 700: Is there anything preventing us from using this implementation for native AOT as well? (Together with MethodOnTypeBuilderInstantiation.cs and FieldOnTypeBuilderInstantiation.cs.)

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/TypeBuilderImpl.cs
buyaa-n: Yep

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/TypeBuilderImpl.cs
buyaa-n: Hm, the tokens could not be populated reliably until save or all types are baked.

jkotas: Why not?

jkotas: It is about def tokens for members of the baked type only.

buyaa-n: Yah, I guess we can force the baked types order to make it work

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
buyaa-n: Actually it is references `TypeBuilderInstantiation` which references `MethodOnTypeBuilderInstantiation`, `FieldOnTypeBuilderInstantiation` and `ConstructorOnTypeBuilderInstantiation` that are excluded from AOT build, probably because they references `RuntimeMethodInfo`, `RuntimeFieldInfo` and `RuntimeConstructorInfo` that are not exist in AOT corelib

jkotas: As far as I can tell, the Runtime...Infos are only referenced by asserts. We should delete the problematic asserts. (I do not think that these asserts are valid - it should be possible to write a test that makes them to fail.)

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/TypeBuilderImpl.cs
buyaa-n: Done, now types need to be baked before save

src/libraries/System.Private.CoreLib/src/System.Private.CoreLib.Shared.projitems
buyaa-n: Done

src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/ConstructorOnTypeBuilderInstantiation.cs
```diff
@@ -57,7 +54,6 @@ public override int MetadataToken
                 }
                 else
                 {
```
jkotas, line 56: This whole method can be simplified to just `public override int MetadataToken => _ctor.MetadataToken;`

jkotas: (Dtto for other files.)

src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/SymbolType.cs
```diff
@@ -286,7 +286,7 @@ public override Type MakeArrayType(int rank)
         public override int GetArrayRank()
         {
             if (!IsArray)
-                throw new NotSupportedException(SR.NotSupported_SubclassOverride);
+                throw new ArgumentException(SR.Argument_HasToBeArrayClass);
```
steveharter, line 289: InvalidOperationException instead?

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/EnumBuilderImpl.cs
```diff
@@ -11,12 +11,12 @@ internal sealed class EnumBuilderImpl : EnumBuilder
...
         internal readonly TypeBuilderImpl _typeBuilder;
 
-        internal EnumBuilderImpl(string name, Type underlyingType, TypeAttributes visibility, ModuleBuilderImpl module, TypeDefinitionHandle typeHandle)
+        internal EnumBuilderImpl(string name, Type underlyingType, TypeAttributes visibility, ModuleBuilderImpl module)
         {
             if ((visibility & ~TypeAttributes.VisibilityMask) != 0)
                 throw new ArgumentException(SR.Argument_ShouldOnlySetVisibilityFlags, nameof(name));
 
-            _typeBuilder = new TypeBuilderImpl(name, visibility | TypeAttributes.Sealed, typeof(Enum), module, typeHandle, null, PackingSize.Unspecified, TypeBuilder.UnspecifiedTypeSize, null);
+            _typeBuilder = new TypeBuilderImpl(name, visibility | TypeAttributes.Sealed, typeof(Enum), module, null, PackingSize.Unspecified, TypeBuilder.UnspecifiedTypeSize, null);
```
steveharter, line 19: nit: use named arguments for the `null` cases.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/MethodBuilderImpl.cs
```diff
@@ -36,13 +36,24 @@ internal sealed class MethodBuilderImpl : MethodBuilder
             _returnType = returnType ?? _module.GetTypeFromCoreAssembly(CoreTypeId.Void);
             _name = name;
             _attributes = attributes;
+
+            if ((attributes & MethodAttributes.Static) == 0)
+            {
+                // turn on the has this calling convention
+                callingConventions |= CallingConventions.HasThis;
+            }
+            else if ((attributes & MethodAttributes.Virtual) != 0 && (attributes & MethodAttributes.Abstract) == 0)
```
steveharter, line 45:
```suggestion
            else if ((attributes & (MethodAttributes.Virtual | MethodAttributes.Abstract)) == MethodAttributes.Virtual)
```

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/MethodBuilderImpl.cs
```diff
@@ -221,13 +254,37 @@ protected override void SetImplementationFlagsCore(MethodImplAttributes attribut
...
+            {
+                return Array.Empty<ParameterInfo>();
+            }
+
+            _parameterBuilders ??= new ParameterBuilderImpl[_parameterTypes.Length + 1]; // parameter 0 reserved for return type
+            ParameterInfo[] parameters = new ParameterInfo[_parameterTypes.Length];
+
+            for (int i = 0; i < _parameterTypes.Length; i++)
+            {
+                if (_parameterBuilders[i + 1] == null)
```
steveharter, line 276: When is the element `null` in this case?

src/libraries/System.Reflection.Emit/tests/PersistableAssemblyBuilder/AssemblySaveConstructorBuilderTests.cs
```diff
@@ -0,0 +1,188 @@
...
+
+namespace System.Reflection.Emit.Tests
+{
+    [ConditionalClass(typeof(PlatformDetection), nameof(PlatformDetection.IsNotBrowser))]
+    public class AssemblySaveConstructorBuilderTests
+    {
+        [Fact]
+        public void DefineConstructorsTest()
+        {
+            AssemblyBuilder ab = AssemblySaveTools.PopulateAssemblyBuilderTypeBuilderAndSaveMethod(out TypeBuilder type, out MethodInfo _);
```
steveharter, line 17: `ab` is not referenced in this test.

src/libraries/System.Reflection.Emit/tests/PersistableAssemblyBuilder/AssemblySaveConstructorBuilderTests.cs
```diff
@@ -0,0 +1,188 @@
...
+            child.SetParent(type);
+            child.DefineDefaultConstructor(MethodAttributes.Family);
+            child.CreateType();
+
+            ConstructorInfo[] ctors = child.GetConstructors(BindingFlags.NonPublic | BindingFlags.Instance);
+            Assert.Equal(1, ctors.Length);
+            Assert.True(ctors[0].IsFamily);
+            Assert.Empty(child.GetConstructors(BindingFlags.Public | BindingFlags.Instance));
+        }
+
```
steveharter, line 48: nit: extra line

src/libraries/System.Reflection.Emit/tests/PersistableAssemblyBuilder/AssemblySaveConstructorBuilderTests.cs
```diff
@@ -0,0 +1,188 @@
...
+
+        [Fact]
+        public void DefineDefaultConstructor_GenericParentCreated_Works()
+        {
+            AssemblyBuilder ab = AssemblySaveTools.PopulateAssemblyBuilderTypeBuilderAndSaveMethod(out TypeBuilder type, out MethodInfo _);
+            type.DefineGenericParameters("T");
```
steveharter, line 54: Is there a way to get the "T" back and verify? e.g. `GetGenericMethodDefinition().GetGenericArguments()[0].Name`

steveharter (approved):
Some misc feedback but otherwise LGTM.

src/libraries/System.Private.CoreLib/src/System/Reflection/Emit/SymbolType.cs
buyaa-n: Followed exception type and message for similar cases
https://github.com/dotnet/runtime/blob/9f6ac8c5a23fde39dbcbd7f85dcd73a7da6c191c/src/coreclr/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeModuleBuilder.cs#L1142-L1145

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/MethodBuilderImpl.cs
buyaa-n: When the parameters created with `parameterTypes` array: https://github.com/dotnet/runtime/blob/ffd43b6d9258e03543dac7f7e5b9ff54e1e7a3b6/src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/MethodBuilderImpl.cs#L53-L60
but not explicitly defined with `DefineParameterCore(...)`

src/libraries/System.Reflection.Emit/tests/PersistableAssemblyBuilder/AssemblySaveConstructorBuilderTests.cs
buyaa-n: Added `Assert.Equal("T", type.GetGenericTypeDefinition().GetGenericArguments()[0].Name);` as this is generic type definition not a method

Issue #94680, opened 2023-11-13
JIT: Assertion failed '!(block->bbFlags & BBF_FINALLY_TARGET)' during 'Global local var liveness'


jakobbotsch:
```csharp
// Generated by Fuzzlyn v1.6 on 2023-11-12 01:56:05
// Run on Arm Linux
// Seed: 15322237346947874736
// Reduced from 37.4 KiB to 7.0 KiB in 00:05:48
// Hits JIT assert in Release:
// Assertion failed '!(block->bbFlags & BBF_FINALLY_TARGET)' in 'Program:M0()' during 'Global local var liveness' (IL size 1536; hash 0xaf50ff37; FullOpts)
// 
//     File: /__w/1/s/src/coreclr/jit/fgbasic.cpp Line: 5209
// 
using System.Runtime.CompilerServices;

public interface I0
{
}

public class C0 : I0
{
    public int F0;
    public int F1;
    public C0(int f0, int f1)
    {
        F0 = f0;
        F1 = f1;
    }
}

public class C1
{
    public sbyte F0;
    public int F1;
    public C0 F2;
    public C1(sbyte f0, int f1, C0 f2)
    {
        F1 = f1;
        F2 = f2;
    }
}

public struct S0 : I0
{
    public sbyte F0;
    public C0 F1;
    public uint F2;
    public byte F3;
    public int F4;
    public S0(sbyte f0, C0 f1, uint f2, byte f3, int f4, byte f5): this()
    {
    }
}

public class Program
{
    public static IRuntime s_rt;
    public static byte[] s_1;
    public static byte s_2;
    public static ushort s_3;
    public static bool s_5;
    public static ulong s_7;
    public static bool s_8;
    public static I0[] s_9;
    public static C0[] s_14;
    public static C1 s_15;
    public static uint s_16;
    public static S0 s_17;
    public static S0 s_18;
    public static S0 s_21;
    public static C1 s_23;
    public static long s_26;
    public static byte[, ] s_31;
    public static void Main()
    {
        CollectibleALC alc = new CollectibleALC();
        System.Reflection.Assembly asm = alc.LoadFromAssemblyPath(System.Reflection.Assembly.GetExecutingAssembly().Location);
        System.Reflection.MethodInfo mi = asm.GetType(typeof(Program).FullName).GetMethod(nameof(MainInner));
        System.Type runtimeTy = asm.GetType(typeof(Runtime).FullName);
[...]

PR #73383, opened 2022-08-04
Delete NGen flavor of JIT configs

jkotas:
We do not need them anymore since the JIT configs are specified on AOT tool command line now.

PR #94733, opened 2023-11-14
Allow spmi collection of altjit compiles

BruceForstall:
When collecting with the `--altjit` argument, interpose the spmi collection shim between the VM and the altjit compiler, not the "main" compiler.

### DIFF
--- a/src/coreclr/scripts/superpmi.py
+++ b/src/coreclr/scripts/superpmi.py
@@ -275,7 +275,7 @@ superpmi_common_parser.add_argument("--skip_cleanup", action="store_true", help=
 superpmi_common_parser.add_argument("--sequential", action="store_true", help="Run SuperPMI in sequential mode. Default is to run in parallel for faster runs.")
 superpmi_common_parser.add_argument("-spmi_log_file", help=spmi_log_file_help)
 superpmi_common_parser.add_argument("-jit_name", help="Specify the filename of the jit to use, e.g., 'clrjit_universal_arm64_x64.dll'. Default is clrjit.dll/libclrjit.so")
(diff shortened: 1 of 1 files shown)


BruceForstall:
@jakobbotsch fyi

I did this work to try and collect the assertion failure in the altjit run of https://github.com/dotnet/runtime/issues/94680. Seems to collect and replay fine, but in that case I couldn't get the replay to hit the assert. I must be missing something on the replay, maybe a missing variable, e.g., the replay isn't being treated as an altjit in the same way as the collection.

src/coreclr/scripts/superpmi.py
```diff
@@ -823,6 +823,10 @@ def __collect_mc_files__(self):
                 dotnet_env["ZapDisable"] = "1"
                 dotnet_env["ReadyToRun"] = "0"
 
+            if self.coreclr_args.altjit:
+                dotnet_env["AltJit"] = "*"
+                dotnet_env["AltJitNgen"] = "*"
```
jakobbotsch, line 828: Looks like we missed deleting this one in #73383.

jakobbotsch:
> @jakobbotsch fyi
> 
> I did this work to try and collect the assertion failure in the altjit run of #94680. Seems to collect and replay fine, but in that case I couldn't get the replay to hit the assert. I must be missing something on the replay, maybe a missing variable, e.g., the replay isn't being treated as an altjit in the same way as the collection.

Hmm, odd. Are the dumps otherwise equivalent?

I wonder if `DOTNET_AltJit` should be in the https://github.com/dotnet/runtime/blob/7292b7104ee6d6d6b7ebfcfb7fccb46639184660/src/coreclr/tools/superpmi/superpmi-shim-collector/jithost.cpp#L33-L89 list so that it is ignored... not sure how it affects the JIT.

Issue #94264, opened 2023-11-01
Runtime event for Monitor.Wait


verdie-g:
<div type='discussions-op-text'>

<sup>Originally posted by **verdie-g** August 17, 2023 https://github.com/dotnet/runtime/discussions/90740</sup>
I recently got a latency spikes issue in my application because the async API of a library was actually waiting on a ManualResetEventSlim.Wait that would cause thread starvation. Now I would like to collect events about any thread that starts waiting. I can collect the contention event, the task waits, but it seems like there is no events emitted in ManualResetEventSlim. Looking at the code [ManualResetEventSlim uses Monitor.Wait](https://github.com/dotnet/runtime/blob/45acd380b37c9ee883070a70a2ef2cb7eca77683/src/libraries/System.Private.CoreLib/src/System/Threading/ManualResetEventSlim.cs#L561) so I've tried looking for an event there but couldn't find any. Any idea what I could do to easily detect blocked threads?</div>

What do you think of having a new event for Monitor.Wait?

Related issue where I suggested a dedicated event for synchronous task waits: https://github.com/dotnet/runtime/issues/63016.

verdie-g:
@noahfalk what do you think about that?

In my experience, most latency issues come from thread pool threads being blocked. It can be by a lock, by sync-over-async or simply by plain synchronous operations (e.g. [HttpClient.Send](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.send)).

When I see the `monitor-lock-contention-count` event counter increasing, I know that I can start collecting `ContentionStart` to get stack traces about where that happened. 

When `monitor-lock-contention-count` is low but I still see spikes of `threadpool-thread-count` or `threadpool-queue-length` I can try dumping the thread stacks and hoping to find the culprit but it's very likely I won't find anything.

If `Monitor.Wait` was emitting an event I could quickly find who's responsible for blocking the thread.

noahfalk:
When you say add an event I assume you are talking about an EventSource event? 

Off the cuff this seems like a reasonable thing to have, though I haven't investigated any details about hooking into the monitor implementation. First few thoughts:
 - There is recursion/deadlock risk if someone hooked up an EventListener, but it seems no worse than what already exists on the ContentionStart event. I feel OK telling people that if they want to hook callbacks into low-level synchronization mechanisms then they need to be careful what they do inside of that callback.
[...]

PR #94737, opened 2023-11-14
Add EventSource event for WaitHandle waits

verdie-g:
Context: #94264


src/coreclr/vm/ClrEtwAll.man
```diff
@@ -1775,6 +1775,30 @@
                         </UserData>
                     </template>
 
+                    <template tid="WaitStart_V1">
```
verdie-g, line 1778: I've named the event just Wait. MonitorWait doesn't really fit because the name Monitor is not used where the event is emitted. ContentionWait could be confused with Contention.

src/coreclr/vm/ClrEtwAll.man
```diff
@@ -3696,6 +3720,15 @@
                            task="Contention"
                            symbol="ContentionLockCreated" message="$(string.RuntimePublisher.ContentionLockCreatedEventMessage)"/>
 
+                    <event value="301" version="1" level="win:Informational"  template="WaitStart_V1"
```
verdie-g, line 3723: It seems like event versions start at 0 with no suffix in the name. I've started this event at 1 and added a prefix, tell me if I need to change that

src/coreclr/vm/ClrEtwAll.man
```diff
@@ -1775,6 +1775,30 @@
                         </UserData>
                     </template>
 
+                    <template tid="WaitStart_V1">
+                        <data name="ClrInstanceID" inType="win:UInt16" />
+                        <data name="LockID" inType="win:Pointer" />
+                        <data name="AssociatedObjectID" inType="win:Pointer" />
```
verdie-g, line 1781: The start event includes the same properties as `ContentionStart`: `ClrInstanceID`, `LockID`, `AssociatedObjectId` but no `ContentionFlags` or `LockOwnerThreadID`

src/coreclr/vm/ClrEtwAll.man
```diff
@@ -1775,6 +1775,30 @@
...
+                        <UserData>
```
verdie-g, line 1782: It could be interesting to add a `ThreadID` property. It would also help to have a `IsThreadPoolThread` but I don't think there is a concept of thread pool at that level

src/coreclr/vm/ClrEtwAll.man
```diff
@@ -1775,6 +1775,30 @@
...
+                            <Contention xmlns="myNs">
```
verdie-g, line 1783: The new event is under the keyword `contention (0x400)`. I think it makes sense to collect it whenever the contention events are collected

src/coreclr/vm/ClrEtwAll.man
verdie-g: Is `ClrInstanceId` a legacy thing? It looks like it's not implemented on linux https://github.com/dotnet/runtime/blob/6f30b7500ea2c0d0fb2923ea1fc5ed3c4ce2f29c/src/coreclr/vm/threads.cpp#L1030-L1039

src/coreclr/vm/ClrEtwAll.man
ezsilmar: I think thread id is added automatically to every emitted event (the same way as timestamp)

src/coreclr/vm/ClrEtwAll.man
noahfalk: > Is ClrInstanceId a legacy thing?

I suspect it was added in CLR 4.0 when support for in-proc SxS was added. Its still part of all our event definitions largely out of convention or the possibility that we'd do SxS again.

src/coreclr/vm/ClrEtwAll.man
noahfalk: I'd suggest we call it MonitorWaitStart. 'Wait' alone feels overly broad. I wouldn't worry about SyncBlock not using the term 'Monitor' in the source. SyncBlock is the name of the internal implementation detail and Monitor is the abstraction that .NET devs see. I assume the goal here is to give an event whenever Monitor.Wait() occurs, regardless of how the runtime might change the implementation internally so naming it after the abstraction seems accurate to me.

I'd also suggest dropping the V1. The convention is starting at version 0 with no suffix. Higher numbered versions are used when creating future revisions to the event.

src/coreclr/vm/ClrEtwAll.man
noahfalk: I'd suggest adding a new keyword instead. The general rule of thumb is that if an event is likely to occur more than 1000 times/sec then we'd like to give it a unique keyword so that tracing tools can toggle it independently. Its a lot better to have users that want multiple events enable several keywords vs. have users that only want one kind of event forced to pay the performance overhead for noisy events they didn't need.

src/coreclr/vm/ClrEtwAll.man
noahfalk: Yeah 0 and no suffix is prefereable. version 1+ is used for making changes to pre-existing events.

src/coreclr/vm/syncblk.cpp
```diff
@@ -2875,6 +2884,16 @@ BOOL SyncBlock::Wait(INT32 timeOut)
     GCPROTECT_END();
     m_Monitor.DecrementTransientPrecious();
 
+    if (isContentionKeywordEnabled)
```
danmoseley, line 2887: Just curious (not my area) given this is presumably a hot path, is there any advantage in moving the body of the if block into its own function? Or maybe PGO does this for us.

kouvel: It's already a very slow path. I haven't seen PGO do much outlining, but it may well move those blocks out of the way, which would be somewhat similar.

src/coreclr/vm/ClrEtwAll.man
kouvel: PerfView has a way to enable/disable individual events under a keyword, not sure about dotnet-trace. Would it be beneficial (also for ease-of-use) to have one keyword for a variety of potentially spammy events? I'm not sure there are a lot of keywords available for unique things like this. There is already a very noisy event in the Threading keyword ([ThreadPoolWorkerThreadWait](https://learn.microsoft.com/en-us/dotnet/fundamentals/diagnostics/runtime-thread-events#threadpoolworkerthreadwait-event) that I've seen to affect perf characteristics of apps when enabled. Wonder if that also should be moved to a different keyword (or be diabled by default, and enabled via config, for compat reasons)?

verdie-g: I moved it to a new keyword `WaitHandle`

src/coreclr/inc/eventtracebase.h
```diff
@@ -1153,6 +1153,19 @@ namespace ETW
...
+
+    class WaitHandleLog
+    {
+    public:
+        typedef union _WaitHandleStructs
+        {
+            typedef  enum _WaitReason {
+                Unknown=0,
+                MonitorWait=1
+            } WaitReason;
```
kouvel, line 1165: Maybe this also be called WaitSource like elsewhere, or all of them could be called WaitReason

src/coreclr/vm/ClrEtwAll.man
```diff
@@ -1775,6 +1781,34 @@
                         </UserData>
                     </template>
 
+                    <template tid="WaitHandleWaitStart">
+                        <data name="WaitSource" inType="win:UInt8" map="WaitHandleWaitSourceMap" />
+                        <data name="ClrInstanceID" inType="win:UInt16" />
```
kouvel, line 1786: Since this is a new event, ClrInstanceID could be moved to be the first parameter so that the other relevant parameters are together.

src/coreclr/vm/ClrEtwAll.man
```diff
@@ -1775,6 +1781,34 @@
...
+                                <ClrInstanceID> %2 </ClrInstanceID>
+                                <AssociatedObjectID> %3 </AssociatedObjectID>
+                            </WaitHandle>
+                        </UserData>
+                    </template>
+
+                    <template tid="WaitHandleWaitStop">
+                        <data name="WaitSource" inType="win:UInt8" map="WaitHandleWaitSourceMap" />
+                        <data name="ClrInstanceID" inType="win:UInt16" />
+                        <data name="AssociatedObjectID" inType="win:Pointer" />
```
kouvel, line 1800: The associated object ID is probably not necessary for the Stop event, there would be a matching Stop event for each Start event even in cases involving reentry, which would probably be less common anyway.

src/coreclr/vm/syncblk.cpp
```diff
@@ -2866,6 +2855,8 @@ BOOL SyncBlock::Wait(INT32 timeOut)
     {
         GCX_PREEMP();
 
+        syncState.m_Object = obj;
```
jkotas, line 2858: It is not valid to work with object references in preemptive GC mode.

src/coreclr/vm/ClrEtwAll.man
```diff
@@ -4211,6 +4244,17 @@
                            keywords ="PerfTrackKeyword" opcode="ExecutionCheckpoint" task="ExecutionCheckpoint" symbol="ExecutionCheckpoint"
                            message="$(string.RuntimePublisher.ExecutionCheckpointEventMessage)"/>
 
+                    <!-- Wait Handle events -->
+                    <event value="301" version="0" level="win:Informational" template="WaitHandleWaitStart"
```
kouvel, line 4248: Maybe `level="win:Verbose"` would be more appropriate for these events, as they can be fairly noisy

src/coreclr/vm/syncblk.cpp
kouvel: Perhaps m_Object could just be an opaque void * pointer and it could be set above just below `obj = m_Monitor.GetOwningObject();`, with `syncState.m_Object = OBJECTREFToObject(obj)`.

It would still be best-effort kind of info, as the object could been moved before the event is actually raised.

verdie-g: Yes I'm wondering if this object ref will actually be exploitable since it can be moved anytime

src/coreclr/vm/ClrEtwAll.man
verdie-g: Ok, I was wondering how a user could match a start from a stop. Maybe that's not a use-case

src/coreclr/vm/syncblk.cpp
kouvel: I suspect there are events that track movement of objects, so it may be possible to track that and figure out which is the real object pointer from the info.

verdie-g:
It doesn't compile, the `FireEtwWaitHandleWaitStart` function is not generated. I'll continue tomorrow.

src/mono/mono/metadata/icall-decl.h
```diff
@@ -176,6 +176,8 @@ ICALL_EXPORT void ves_icall_System_Diagnostics_Tracing_NativeRuntimeEventSource_
 ICALL_EXPORT void ves_icall_System_Diagnostics_Tracing_NativeRuntimeEventSource_LogContentionLockCreated (intptr_t lock_id, intptr_t associated_object_id, uint16_t clr_instance_id);
 ICALL_EXPORT void ves_icall_System_Diagnostics_Tracing_NativeRuntimeEventSource_LogContentionStart (uint8_t contention_flags, uint16_t clr_instance_id, intptr_t lock_id, intptr_t associated_object_id, uint64_t lock_owner_thread_id);
 ICALL_EXPORT void ves_icall_System_Diagnostics_Tracing_NativeRuntimeEventSource_LogContentionStop (uint8_t contention_flags, uint16_t clr_instance_id, double duration_ns);
+ICALL_EXPORT void ves_icall_System_Diagnostics_Tracing_NativeRuntimeEventSource_LogWaitHandleWaitStart (uint16_t clr_instance_id, uint32_t wait_source, intptr_t associated_object_id);
```
vargaz, line 179: These 2 don't seem to match the managed definition.

verdie-g: The managed log method needs to have the CrlInstanceId as last parameter to have a default value. For consistency I have moved it as last for all methods.

src/mono/System.Private.CoreLib/src/System/Diagnostics/Tracing/NativeRuntimeEventSource.Threading.NativeSinks.Mono.cs
```diff
@@ -97,5 +97,17 @@ ushort ClrInstanceID
             IntPtr NativeOverlapped,
             IntPtr Overlapped,
             ushort ClrInstanceID);
+
+        [NonEvent]
+        private static extern void LogWaitHandleWaitStart(
```
lateralusX, line 102: [MethodImplAttribute(MethodImplOptions.InternalCall)] ?

src/mono/System.Private.CoreLib/src/System/Diagnostics/Tracing/NativeRuntimeEventSource.Threading.NativeSinks.Mono.cs
```diff
@@ -97,5 +97,17 @@ ushort ClrInstanceID
...
+            WaitHandleWaitSourceMap WaitSource,
+            IntPtr AssociatedObjectID,
+            ushort ClrInstanceID);
+
+        [NonEvent]
+        private static extern void LogWaitHandleWaitStop(
```
lateralusX, line 108: [MethodImplAttribute(MethodImplOptions.InternalCall)] ?

src/coreclr/vm/ClrEtwAll.man
lateralusX: All events are tagged with thread id, so there will be a sequence of start/stop pairs per thread, so a stop should be related to the previous start event on the same thread. It means that the code parsing the nettrace file needs to keep track of start events if there is an interest in the object id and wait source. So exclude object id on stop will make parsing a little more complicated in favor of saving 4-8 bytes per stop event.

src/coreclr/vm/ClrEtwAll.man
```diff
@@ -1775,6 +1787,32 @@
                         </UserData>
                     </template>
 
+                    <template tid="WaitHandleWaitStart">
+                        <data name="WaitSource" inType="win:UInt32" map="WaitHandleWaitSourceMap" />
```
lateralusX, line 1791: Do we need 4 bytes for the wait source? Shouldn't a UInt8 be enough?

verdie-g: I used that first but I saw other enum with greater sizes so I preferred to be safe here.
GCReasonMap -> 32 bits
GCSuspendEEReasonMap -> 16 bits
ThreadAdjustmentReasonMap -> 32 bits
TailCallTypeMap -> 32 bits
GCRootKindMap -> 32 bits

kouvel: Probably a byte would be enough, a byte would have been enough for `ThreadAdjustmentReasonMap` too, but it started off being 4 bytes so that can't be changed now.

src/coreclr/vm/ClrEtwAll.man
kouvel: I figured saving on the size would be preferrable, esp. as these events can be noisy. I believe the parsing in PerfView already keeps track of start/stop events to compute a duration.

src/coreclr/vm/ClrEtwAll.man
```diff
@@ -1775,6 +1787,32 @@
...
+                                <WaitSource> %1 </WaitSource>
+                                <AssociatedObjectID> %2 </AssociatedObjectID>
+                                <ClrInstanceID> %3 </ClrInstanceID>
+                            </WaitHandle>
+                        </UserData>
+                    </template>
+
+                    <template tid="WaitHandleWaitStop">
+                        <data name="WaitSource" inType="win:UInt32" map="WaitHandleWaitSourceMap" />
+                        <data name="DurationNs" inType="win:Double" />
```
kouvel, line 1805: I'm skeptical about whether the `DurationNs` is necessary. It was included with the `ContentionStop` event because lock contention durations can indeed be very short. Waits can also be short (~10 microseconds or less), but probably the more interesting cases will be the longer waits. PerfView already computes a duration for start/stop events based on the event timestamps. I think that would be good enough for starters, maybe we can omit the duration in the event until it's deemed necessary.

ezsilmar: I think as long as we have QPC timestamp in the event data we should be good without the duration. PerfView lacks ability to analyze durations: there's no histograms or even dumping a stacktrace of a particular event (that would be a super useful feature to diagnose long waits like this btw). When I use fields like this I'm usually running custom code with TraceEvent anyway. And QPC precision is pretty much as good as it gets.

verdie-g:
It says the checks have passed but they were actually not run 🤔 Maybe it didn't like my last rebase.

kouvel:
Maybe the bot hasn't scheduled them yet, if it doesn't start in some time you can try closing/reopening the PR.

src/coreclr/vm/ClrEtwAll.man
```diff
@@ -504,6 +512,10 @@
                         <map value="0x0" message="$(string.RuntimePublisher.Contention.ManagedMapMessage)"/>
                         <map value="0x1" message="$(string.RuntimePublisher.Contention.NativeMapMessage)"/>
                     </valueMap>
+                    <valueMap name="WaitHandleWaitSourceMap">
+                        <map value="0x0" message="$(string.RuntimePublisher.WaitHandle.UnknownMapMessage)"/>
+                        <map value="0x1" message="$(string.RuntimePublisher.WaitHandle.MonitorWaitMapMessage)"/>
```
kouvel, line 517:
```suggestion
                        <map value="0x0" message="$(string.RuntimePublisher.WaitHandleWaitSource.UnknownMapMessage)"/>
                        <map value="0x1" message="$(string.RuntimePublisher.WaitHandleWaitSource.MonitorWaitMapMessage)"/>
```

kouvel (commented):
Just a few mostly minor comments so far, still looking

verdie-g:
Thanks, I'll address these comments.

I'm currently struggling to run the tests
```
./build.sh -c Debug
./src/tests/build.sh -p:LibrariesConfiguration=Debug -test:tracing/runtimeeventsource/runtimeeventsource.csproj
bash /runtime/artifacts/tests/coreclr/linux.x64.Debug/tracing/runtimeeventsource/runtimeeventsource/runtimeeventsource.sh
```
fail with
```
SKIPPING EXECUTION BECAUSE DOTNET_TieredCompilation has not been disabled and this test is marked JitOptimizationSensitive
```
I can't find any relevant doc about that. Any idea?

kouvel:
The doc here may help: https://github.com/dotnet/runtime/blob/main/docs/workflow/building/coreclr/nativeaot.md. 
It has instructions on building and running the runtime and libraries tests under NativeAOT.

src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/NativeRuntimeEventSource.Threading.cs
```diff
@@ -504,5 +514,45 @@ public unsafe void ThreadPoolIOPack(NativeOverlapped* nativeOverlapped)
             data[4].Reserved = 0;
             WriteEventCore(59, 5, data);
         }
+
+        [UnconditionalSuppressMessage("ReflectionAnalysis", "IL2026:UnrecognizedReflectionPattern", Justification = "Parameters to this method are primitive and are trimmer safe")]
```
danmoseley, line 518: maybe this was already discussed but is `WaitHandleWaitSourceMap` "trimmer safe"?

lateralusX:
> Thanks, I'll address these comments.
> 
> I'm currently struggling to run the tests
> 
> ```
> ./build.sh -c Debug
> ./src/tests/build.sh -p:LibrariesConfiguration=Debug -test:tracing/runtimeeventsource/runtimeeventsource.csproj
> bash /runtime/artifacts/tests/coreclr/linux.x64.Debug/tracing/runtimeeventsource/runtimeeventsource/runtimeeventsource.sh
> ```
> 
> fail with
> 
> ```
> SKIPPING EXECUTION BECAUSE DOTNET_TieredCompilation has not been disabled and this test is marked JitOptimizationSensitive
> ```
> 
> I can't find any relevant doc about that. Any idea?

For runtime tests it needs to be disabled in the env you are running the test, export DOTNET_TieredCompilation=0.

src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/NativeRuntimeEventSource.Threading.cs
MichalStrehovsky: This should be fine. We're doing suppressions like this for enums elsewhere. It's only problematic when this involves recursive graph serialization.

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Condition.cs
```diff
@@ -107,6 +107,16 @@ public unsafe bool Wait(int millisecondsTimeout)
...
             AddWaiter(waiter);
 
+            bool isWaitHandleKeywordEnabled = NativeRuntimeEventSource.Log.IsEnabled(
+                EventLevel.Informational,
+                NativeRuntimeEventSource.Keywords.WaitHandleKeyword);
+            if (isWaitHandleKeywordEnabled)
+            {
+                NativeRuntimeEventSource.Log.WaitHandleWaitStart(
+                    NativeRuntimeEventSource.WaitHandleWaitSourceMap.MonitorWait,
+                    0);
```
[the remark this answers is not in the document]
kouvel: Actually nevermind, the associated object would not be available in the info collected in the profile and it seems like it would be difficult to find the associated object from a heap dump given a Condition object. It seems like it would be easy enough to pass in the associated object from Monitor, that would probably be better. The method's signature could be changed to `bool Wait(int millisecondsTimeout, object? associatedObject = null)` and a pointer to the `associatedObject ?? this` object could be passed in for the event.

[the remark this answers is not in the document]
verdie-g: Are you thinking about creating a new internal method `Condition.Wait`?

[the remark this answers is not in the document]
kouvel: Could just add a new parameter to the existing method with a default value of null.

src/coreclr/vm/threads.cpp
```diff
@@ -3178,6 +3178,15 @@ DWORD Thread::DoAppropriateWait(int countHandles, HANDLE *handles, BOOL waitAll,
     param.mode = mode;
     param.dwRet = (DWORD) -1;
 
+    bool isWaitHandleKeywordEnabled = ETW_TRACING_CATEGORY_ENABLED(MICROSOFT_WINDOWS_DOTNETRUNTIME_PROVIDER_DOTNET_Context, TRACE_LEVEL_VERBOSE, CLR_WAITHANDLE_KEYWORD);
+    if (isWaitHandleKeywordEnabled)
+    {
+        FireEtwWaitHandleWaitStart(
+            ETW::WaitHandleLog::WaitHandleStructs::MonitorWait,
```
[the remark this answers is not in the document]
verdie-g:
```cs
    if (isWaitHandleKeywordEnabled)
    {
        if (syncState)
        {
            FireEtwWaitHandleWaitStart(ETW::WaitHandleLog::WaitHandleStructs::MonitorWait, syncState->m_Object, GetClrInstanceId());
        }
        else
        {
            FireEtwWaitHandleWaitStart(ETW::WaitHandleLog::WaitHandleStructs::Unknown, NULL, GetClrInstanceId());
        }
    }
```
?

src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs
```diff
@@ -5152,7 +5156,7 @@ internal sealed class ManifestBuilder
         }
 
         /// <summary>
-        /// <term>Will NOT build a manifest!</term> If the intention is to build a manifest don�t use this constructor.
+        /// <term>Will NOT build a manifest!</term> If the intention is to build a manifest don�t use this constructor.
```
kouvel, line 5159: Revert?

src/libraries/System.Private.CoreLib/src/System/Threading/WaitHandle.cs
```diff
@@ -111,6 +112,16 @@ private bool WaitOneNoCheck(int millisecondsTimeout)
             SafeWaitHandle? waitHandle = _waitHandle;
             ObjectDisposedException.ThrowIf(waitHandle is null, this);
 
+            bool isWaitHandleKeywordEnabled = NativeRuntimeEventSource.Log.IsEnabled(
+                EventLevel.Informational,
+                NativeRuntimeEventSource.Keywords.WaitHandleKeyword);
+            if (isWaitHandleKeywordEnabled)
+            {
+                NativeRuntimeEventSource.Log.WaitHandleWaitStart(
```
kouvel, line 120: I think there are cases where duplicate events would be raised for the same wait. For instance in NativeAOT, Condition would raise an event, and another event would be raised here. In CoreCLR, this event would be raised and another in Thread::DoAppropriateWait. There's probably some tweaking to do. For NativeAOT/Mono, I think raising the event from the Wait*Core methods would work, along with passing the info down through internal methods. For CoreCLR, it just let the native side raise the event.

kouvel (commented):
Couple more comments

src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs
kouvel: Hmm I see now the previous one was also a special character.
```suggestion
        /// <term>Will NOT build a manifest!</term> If the intention is to build a manifest don't use this constructor.
```

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Condition.cs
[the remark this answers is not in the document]
verdie-g: It seems to be a public API, I'm not sure we want to expose that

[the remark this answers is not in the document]
kouvel: I'm not sure why Condition is a public class, it seems to be used only from CoreLib and is not in the public API surface ([link](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Runtime/ref/System.Runtime.cs#L15019C11-L15019C11), [link](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Threading/ref/System.Threading.cs)). I expect it would be ok to change the method signature.

src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs
verdie-g: Yeah not in the scope anyway :)

src/libraries/System.Private.CoreLib/src/System/Threading/WaitHandle.cs
verdie-g: Hm it's very messy in my head, I'll try drawing the different paths and get back with a proposal.

verdie-g: image

If that's accurate is it needed to emit an event in managed code? 🤔

kouvel: Thinking about it more, it seems Monitor.Wait is kind of a special case where it's implemented in native code in CoreCLR and Mono, and almost always would be a blocking wait. For all other WaitHandle.Wait* cases, it would probably make sense to raise the event from the managed implementations. Maybe it would make sense to special-case the raising of the event for Monitor.Wait (though in Mono it looks like it could also be raised in managed code before calling out to the native implementation). I'll try to draft up a change of what I'm thinking.

kouvel: May need to also check in CoreCLR other wait cases from the native side that lead to waiting on wait handles, to see if it would be useful to cover.

jkotas:
Add tests for the new event?

src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Condition.cs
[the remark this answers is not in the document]
jkotas: > I'm not sure why Condition is a public class

Fixed in #95380

src/libraries/System.Private.CoreLib/src/System/Threading/WaitHandle.cs
kouvel: Yea probably makes sense to raise the event in Thread::DoAppropriateWait or below in CoreCLR. There are a couple other tweaks I was thinking of, will try to get a draft out sometime soon.

verdie-g:
What is the expected testing strategy? I was looking for contention lock event tests but I didn't find any.

src/libraries/System.Private.CoreLib/src/System/Threading/WaitHandle.cs
verdie-g: I've removed all managed code from the PR

kouvel: It looks like with the latest changes it would only work in CoreCLR.

I made some changes in this commit: https://github.com/dotnet/runtime/commit/35c13cfe05578ad2fabfe6711dcdd2ae1be0dcec

That brings back the minimal code to raise the event from the managed side for NativeAOT/Mono, along with a few other changes. The commit description has additional details. I've tested it and the events appear to work in CoreCLR, NativeAOT, and Mono. Please take a look, if that looks ok to you, maybe you can merge it into your branch?

kouvel:
> @kouvel looks good to you?

I've responded in https://github.com/dotnet/runtime/pull/94737#discussion_r1417948621, please take a look.

src/libraries/System.Private.CoreLib/src/System/Threading/WaitHandle.cs
verdie-g: Doesn't NativeAOT use the native code where the event emitted from? The test is passing without disabling NativeAOT.

I'm a bit confused about the mono part though, where is the event emitted?

kouvel: For WaitHandle, NativeAOT and Mono on Windows [p/invoke into Windows APIs](https://github.com/dotnet/runtime/blob/1aae18a21ebb5f08a2a734cfe31ba4bd00b2ad7b/src/libraries/System.Private.CoreLib/src/System/Threading/WaitHandle.Windows.cs#L52-L61), and on Unixes use the [managed WaitSubsystem implementation](https://github.com/dotnet/runtime/blob/1aae18a21ebb5f08a2a734cfe31ba4bd00b2ad7b/src/libraries/System.Private.CoreLib/src/System/Threading/WaitHandle.Unix.cs#L11).

Monitor.Wait on Mono has its implementation in native code, the event is emitted here in my commit: https://github.com/dotnet/runtime/commit/35c13cfe05578ad2fabfe6711dcdd2ae1be0dcec#diff-847bb36cd91f623c5b9f55a1980656052b342e6d93da972a41ded1210b57d719R130

kouvel: > The test is passing without disabling NativeAOT.

The test currently doesn't run with NativeAOT because the RemoteExecutor is currently not supported for it.

src/coreclr/vm/nativeeventsource.cpp
```diff
@@ -189,4 +189,24 @@ extern "C" void QCALLTYPE LogContentionStop(uint8_t ContentionFlags, uint16_t Cl
...
+
+extern "C" void QCALLTYPE LogWaitHandleWaitStop(uint16_t ClrInstanceID)
+{
+    QCALL_CONTRACT;
+    BEGIN_QCALL;
+
+    FireEtwWaitHandleWaitStop(ClrInstanceID);
+
+    END_QCALL;
+}
```
kouvel, line 210: These are unnecessary since the event is not raised from the managed side in CoreCLR currently

src/coreclr/vm/nativeeventsource.h
```diff
@@ -31,6 +31,8 @@ extern "C" void QCALLTYPE LogThreadPoolIOPack(_In_z_ void* nativeOverlapped, _In
 extern "C" void QCALLTYPE LogContentionLockCreated(void* LockID, void* AssociatedObjectID, uint16_t ClrInstanceID);
 extern "C" void QCALLTYPE LogContentionStart(uint8_t ContentionFlags, uint16_t ClrInstanceID, void* LockID, void* AssociatedObjectID, uint64_t LockOwnerThreadID);
 extern "C" void QCALLTYPE LogContentionStop(uint8_t ContentionFlags, uint16_t ClrInstanceID, double DurationNs);
+extern "C" void QCALLTYPE LogWaitHandleWaitStart(uint8_t WaitSource, void* AssociatedObjectID, uint16_t ClrInstanceID);
+extern "C" void QCALLTYPE LogWaitHandleWaitStop(uint16_t ClrInstanceID);
```
kouvel, line 35: These are unnecessary since the event is not raised from the managed side in CoreCLR currently

src/coreclr/vm/qcallentrypoints.cpp
```diff
@@ -310,6 +310,8 @@ static const Entry s_QCall[] =
     DllImportEntry(LogContentionLockCreated)
     DllImportEntry(LogContentionStart)
     DllImportEntry(LogContentionStop)
+    DllImportEntry(LogWaitHandleWaitStart)
+    DllImportEntry(LogWaitHandleWaitStop)
```
kouvel, line 314: These are unnecessary since the event is not raised from the managed side in CoreCLR currently

src/libraries/System.Private.CoreLib/src/System/Threading/WaitHandle.cs
```diff
@@ -111,21 +116,74 @@ private bool WaitOneNoCheck(int millisecondsTimeout)
...
+                    NativeRuntimeEventSource.WaitHandleWaitSourceMap.Unknown,
+                    0);
+            }
+
```
kouvel, line 128: This can be deleted, it's left over from before and is replaced by the changes in my commit

src/libraries/System.Private.CoreLib/src/System/Threading/WaitHandle.cs
```diff
@@ -139,6 +197,11 @@ private bool WaitOneNoCheck(int millisecondsTimeout)
             {
                 if (success)
                     waitHandle.DangerousRelease();
+
+                if (isWaitHandleKeywordEnabled)
+                {
+                    NativeRuntimeEventSource.Log.WaitHandleWaitStop();
+                }
```
kouvel, line 204: This can be deleted, it's left over from before and is replaced by the changes in my commit

verdie-g: Oops sorry about, I'll fix all that now.

kouvel (approved):
LGTM, thanks!

PR #94738, opened 2023-11-14
[wasi] Run the mono runtime component selection logic for wasi

lewing:
This filters the runtime components correctly such that we don't link both the stub and the full library.


lewing:
in tree needs some work

lewing:
Added conditioning on the task to only use it when using the workload which is similar to what browser is doing.  Should resolve the test failures.

radical (approved):

PR #94741, opened 2023-11-15
JIT: enable complementary jtrue assertions for cross-block local ap

AndyAyersMS:
Now that we can propagate assertions across block boundaries we can generate assertions for true/false branch conditions and propagate them along the corresponding edges.

Contributes to #93246.

### DIFF
--- a/src/coreclr/jit/morph.cpp
+++ b/src/coreclr/jit/morph.cpp
@@ -12937,8 +12937,8 @@ void Compiler::fgAssertionGen(GenTree* tree)
     };
 
     // For BBJ_COND nodes, we have two assertion out BVs.
-    // apLocal will be stored on bbAssertionOut and be used for false successors.
-    // apLocalIfTrue will be stored on bbAssertionGen and be used for true successors.
+    // apLocal will be stored on bbAssertionOutIfFalse and be used for false successors.
+    // apLocalIfTrue will be stored on bbAssertionOutIfTrue and be used for true successors.
     //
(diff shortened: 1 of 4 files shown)


AndyAyersMS:
@jakobbotsch PTAL
cc @dotnet/jit-contrib 

This doesn't have any impact unless cross-block is enabled. When it is I see this sort of impact


It's possible that enabling "if true" copy prop based on `if (x == y)` and "if false" based on `if (x != y)` is not a great idea; I may play around with filtering these out of the `GT_JTRUE` gen se

image

and a small TP hit

image
ts.

src/coreclr/jit/morph.cpp
```diff
@@ -13833,17 +13894,35 @@ void Compiler::fgMorphBlock(BasicBlock* block, unsigned highestReachablePostorde
...
-                    // Yes, pred assertions are available. If this is the first pred, copy.
+                    // Yes, pred assertions are available.
+                    // If the pred is (a non-degenerate) BBJ_COND, fetch the appropriate out set.
+                    //
+                    ASSERT_TP assertionsOut = pred->bbAssertionOut;
+
+                    if (pred->KindIs(BBJ_COND) && (pred->NumSucc() == 2) && (block == pred->GetJumpDest()))
+                    {
+                        JITDUMP("Using `if true` assertions from pred " FMT_BB "\n", pred->bbNum);
+                        assertionsOut = pred->bbAssertionGen;
```
jakobbotsch, line 13905: Consider adding new members to the unions to give these sets more proper names.

jakobbotsch (approved):
LGTM, a minor note about naming.

PR #94742, opened 2023-11-15
Fixed setting ReadMode to Message for NamedPipeClientStream on Windows

### LINKED ISSUE #83072: Cannot set ReadMode after creating a NamedPipeClientStream instance
### Background and motivation

Cannot set `ReadMode` after creating a `NamedPipeClientStream` instance.  This is due to not exposing the `NamedPipeClientStream` constructor which accepts a `PipeAccessRights`.

### API Proposal

```csharp
namespace System.IO.Pipes

// existing type
public sealed partial class NamedPipeClientStream : PipeStream
{
    // new constructor
    [SupportedOSPlatform("windows")]
    public NamedPipeClientStream(String serverName, String pipeName, PipeAccessRights desiredAccessRights,
           PipeOptions options, TokenImpersonationLevel impersonationLevel, HandleInheritability inheritability)
}

// existing type moved from System.IO.Pipes.AccessControl assembly.
[System.FlagsAttribute]
public enum PipeAccessRights
{
    ReadData = 1,
    WriteData = 2,
    CreateNewInstance = 4,
    ReadExtendedAttributes = 8,
    WriteExtendedAttributes = 16,
    ReadAttributes = 128,
    WriteAttributes = 256,
    Write = 274,
    Delete = 65536,
    ReadPermissions = 131072,
    Read = 131209,
    ReadWrite = 131483,
    ChangePermissions = 262144,
    TakeOwnership = 524288,
    Synchronize = 1048576,
    FullControl = 2032031,
    AccessSystemSecurity = 16777216,
}

```

### API Usage

```csharp
var pipeOut = new NamedPipeServerStream("SomeNamedPipe",
                                        PipeDirection.Out,
                                        1,
                                        PipeTransmissionMode.Message);

var pipeIn = new NamedPipeClientStream(".", 
                             "SomeNamedPipe", 
                             PipeAccessRights.ReadData | PipeAccessRights.WriteAttributes, 
                             PipeOptions.None, 
                             System.Security.Principal.TokenImpersonationLevel.None, 
                             System.IO.HandleInheritability.None);

pipeIn.Connect();
pipeIn.ReadMode = PipeTransmissionMode.Message; 
```

### Alternative Designs

We could expose this with a `Create` factory method instead, as was done for `NamedPipeServerStream` 
https://github.com/dotnet/runtime/blob/dbb335c6ba14b38400b2d8c3a5876698021ec089/src/libraries/System.IO.Pipes.AccessControl/ref/System.IO.Pipes.AccessControl.cs#L9-L16

However that's not necessary and probably a misnomer since this new constructor isn't really about ACLs, but instead access rights on the handle.  Contrast that to those existing factory methods which take `PipeSecurity` which is an ACLs concept.  Splitting hairs, perhaps, since the API is related to access and is likely not supported on non-Windows, but I think we can do better here by exposing the exact missing API so we should.

### Risks

Moving types around carries with it some risk.  Luckily all the assemblies here are in the shared framework (since 6.0) and referenced as a set so it's less of a risk,

### Original issue

Consider this code:

```cs
var pipeOut = new NamedPipeServerStream("SomeNamedPipe",
                                        PipeDirection.Out,
                                        1,
                                        PipeTransmissionMode.Message);

var pipeIn = new NamedPipeClientStream(".",
                                       "SomeNamedPipe",
                                       PipeDirection.In);

pipeIn.Connect();
pipeIn.ReadMode = PipeTransmissionMode.Message;  // <-- exception thrown here
```
 
In .NET Framework, there used to be a constructor on the client side that accepted `PipeAccessRights`, which would solve the issue, and would allow the user to change `ReadMode` later:
```cs
var pipeIn = 
   new NamedPipeClientStream(".", 
                             "SomeNamedPipe", 
                             PipeAccessRights.ReadData | PipeAccessRights.WriteAttributes, 
                             PipeOptions.None, 
                             System.Security.Principal.TokenImpersonationLevel.None, 
                             System.IO.HandleInheritability.None);
```
Unfortunately, we haven't ported that constructor to .NET Core+.

The underlying problem is that, according to the documentation of the Windows method `CreateNamedPipe`, we need to pass the `FILE_WRITE_ATTRIBUTES` access right when constructing the pipe handle, to be able to modify it after creation. This is explained in the remarks: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createnamedpipea#remarks

The .NET Framework solution worked because it would set `FILE_WRITE_ATTRIBUTES` when using the constructor that takes a `PipeAccessRights` argument.

Unfortunately, the only workaround available in .NET Core+ is to:

- Manually construct the `SafePipeHandle`, making sure to pass `FILE_WRITE_ATTRIBUTES` to the P/Invoke, which is not straightforward.
- Create the client stream instance using the constructor that takes the pipe handle, while also making sure to pass `true` to the argument that indicates that the pipe should get connected.
- Set the `ReadMode`.

I wrote a working example of this workaround for .NET 6.0 and published it here: https://github.com/carlossanlop/experiments/tree/NamedPipeClientStream_ReadMode

But we should consider solving it in runtime.

Note that the workaround provided above [had to call an internal constructor from `PipeSecurity` using reflection](https://github.com/carlossanlop/experiments/blob/761691495d881085c4dc836b07b65976eda2584c/client/Workaround.cs#L102). It's another argumemt supporting the idea that we should provide the solution ourselves.

RogerSanders:
Fixes a longstanding issue inherited from the .NET framework which prevents ReadMode being set to Message on the client end of read only named pipes on Windows platforms.

Also added in new tests for Message mode on Windows to verify correct functionality.

Fix #83072

### DIFF
--- a/src/libraries/System.IO.Pipes/src/System/IO/Pipes/NamedPipeClientStream.Windows.cs
+++ b/src/libraries/System.IO.Pipes/src/System/IO/Pipes/NamedPipeClientStream.Windows.cs
@@ -34,6 +34,7 @@ namespace System.IO.Pipes
                 _pipeFlags |= (((int)_impersonationLevel - 1) << 16);
             }
 
+            bool addedWriteAttributesPermission = false;
             int access = 0;
             if ((PipeDirection.In & _direction) != 0)
             {
@@ -46,9 +47,20 @@ namespace System.IO.Pipes
             else
             {
                 access |= Interop.Kernel32.FileOperations.FILE_WRITE_ATTRIBUTES;
+                addedWriteAttributesPermission = true;
             }
 
             SafePipeHandle handle = CreateNamedPipeClient(_normalizedPipePath, ref secAttrs, _pipeFlags, access);
+            if (handle.IsInvalid && addedWriteAttributesPermission && (Marshal.GetLastPInvokeError() == Interop.Errors.ERROR_ACCESS_DENIED))
+            {
+                // We need the WriteAttributes permission in order to enable message mode, but for backwards
+                // compatibility and to support pipes that have been created where this permission isn't provided to
+                // the client, we try again and strip this permission if our first attempt to open the pipe with this
+                // fails.
+                access &= ~Interop.Kernel32.FileOperations.FILE_WRITE_ATTRIBUTES;
+
+                handle = CreateNamedPipeClient(_normalizedPipePath, ref secAttrs, _pipeFlags, access);
+            }
 
             if (handle.IsInvalid)
             {
--- a/src/libraries/System.IO.Pipes/src/System/IO/Pipes/NamedPipeClientStream.Windows.cs
+++ b/src/libraries/System.IO.Pipes/src/System/IO/Pipes/NamedPipeClientStream.Windows.cs
@@ -43,6 +43,10 @@ namespace System.IO.Pipes
             {
                 access |= Interop.Kernel32.GenericOperations.GENERIC_WRITE;
             }
+            else
+            {
+                access |= Interop.Kernel32.FileOperations.FILE_WRITE_ATTRIBUTES;
+            }
 
             SafePipeHandle handle = CreateNamedPipeClient(_normalizedPipePath, ref secAttrs, _pipeFlags, access);
 
--- a/src/libraries/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Specific.cs
+++ b/src/libraries/System.IO.Pipes/tests/NamedPipeTests/NamedPipeTest.Specific.cs
@@ -474,6 +474,62 @@ namespace System.IO.Pipes.Tests
 
         [Fact]
         [PlatformSpecific(TestPlatforms.Windows)] // Unix doesn't currently support message mode
+        public void Windows_SetReadModeTo__PipeTransmissionModeMessage()
+        {
+            string pipeName = PipeStreamConformanceTests.GetUniquePipeName();
+            using (var server = new NamedPipeServerStream(pipeName, PipeDirection.In, 1, PipeTransmissionMode.Message, PipeOptions.Asynchronous))
+            using (var client = new NamedPipeClientStream(".", pipeName, PipeDirection.Out))
+            {
+                Task clientConnect = client.ConnectAsync();
+                server.WaitForConnection();
+                clientConnect.Wait();
+
+                // Throws regardless of connection status for the pipe that is set to PipeDirection.In. This is because
+                // as per MSDN, "This mode gives the server the equivalent of GENERIC_READ access to the pipe", which
+                // isn't sufficient to call SetNamedPipeHandleState, as we need FILE_WRITE_ATTRIBUTES as well, or
+                // GENERIC_WRITE instead which includes this permission. Since the readmode can be set for the server
+                // side when creating the named pipe, this issue can be avoided by specifying the desired mode from the
+                // start. In the theoretical case of wanting to create a pipe in message mode, then downgrade the
+                // readmode to byte mode on the server however, this isn't currently possible in the Win32 API.
+                Assert.Throws<UnauthorizedAccessException>(() => server.ReadMode = PipeTransmissionMode.Byte);
+                Assert.Throws<UnauthorizedAccessException>(() => server.ReadMode = PipeTransmissionMode.Message);
+
+                // Can select either message or byte mode on client end of the pipe
+                client.ReadMode = PipeTransmissionMode.Byte;
+                client.ReadMode = PipeTransmissionMode.Message;
+            }
+
+            using (var server = new NamedPipeServerStream(pipeName, PipeDirection.Out, 1, PipeTransmissionMode.Message, PipeOptions.Asynchronous))
+            using (var client = new NamedPipeClientStream(".", pipeName, PipeDirection.In))
+            {
+                Task clientConnect = client.ConnectAsync();
+                server.WaitForConnection();
+                clientConnect.Wait();
+
+                // Can select either message or byte mode on either end of the pipe
+                server.ReadMode = PipeTransmissionMode.Byte;
+                client.ReadMode = PipeTransmissionMode.Byte;
+                server.ReadMode = PipeTransmissionMode.Message;
+                client.ReadMode = PipeTransmissionMode.Message;
+            }
+
+            using (var server = new NamedPipeServerStream(pipeName, PipeDirection.InOut, 1, PipeTransmissionMode.Message, PipeOptions.Asynchronous))
+            using (var client = new NamedPipeClientStream(".", pipeName, PipeDirection.InOut, PipeOptions.Asynchronous))
+            {
+                Task clientConnect = client.ConnectAsync();
+                server.WaitForConnection();
+                clientConnect.Wait();
+
+                // Can select either message or byte mode on either end of the pipe
+                server.ReadMode = PipeTransmissionMode.Byte;
+                client.ReadMode = PipeTransmissionMode.Byte;
+                server.ReadMode = PipeTransmissionMode.Message;
+                client.ReadMode = PipeTransmissionMode.Message;
+            }
+        }
+
+        [Fact]
+        [PlatformSpecific(TestPlatforms.Windows)] // Unix doesn't currently support message mode
         public void Windows_SetReadModeTo__PipeTransmissionModeByte()
         {
             string pipeName = PipeStreamConformanceTests.GetUniquePipeName();
@@ -484,9 +540,20 @@ namespace System.IO.Pipes.Tests
                 server.WaitForConnection();
                 clientConnect.Wait();
 
-                // Throws regardless of connection status for the pipe that is set to PipeDirection.In
-                Assert.Throws<UnauthorizedAccessException>(() => server.ReadMode = PipeTransmissionMode.Byte);
                 client.ReadMode = PipeTransmissionMode.Byte;
+
+                // Throws regardless of connection status for the pipe that is set to PipeDirection.In. This is because
+                // as per MSDN, "This mode gives the server the equivalent of GENERIC_READ access to the pipe", which
+                // isn't sufficient to call SetNamedPipeHandleState, as we need FILE_WRITE_ATTRIBUTES as well, or
+                // GENERIC_WRITE instead which includes this permission. Since the readmode can be set for the server
+                // side when creating the named pipe, this issue can be avoided by specifying the desired mode from the
+                // start. In the theoretical case of wanting to create a pipe in message mode, then downgrade the
+                // readmode to byte mode on the server however, this isn't currently possible in the Win32 API.
+                Assert.Throws<UnauthorizedAccessException>(() => server.ReadMode = PipeTransmissionMode.Byte);
+                Assert.Throws<UnauthorizedAccessException>(() => server.ReadMode = PipeTransmissionMode.Message);
+
+                // Can't change to message mode on the client for a byte pipe
+                Assert.Throws<System.IO.IOException>(() => client.ReadMode = PipeTransmissionMode.Message);
             }
 
             using (var server = new NamedPipeServerStream(pipeName, PipeDirection.Out, 1, PipeTransmissionMode.Byte, PipeOptions.Asynchronous))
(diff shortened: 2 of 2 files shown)


RogerSanders:
Adding some comments here to explain what's happening. I recommend reading this stackoverflow post:
https://stackoverflow.com/questions/32739224/c-sharp-unauthorizedaccessexception-when-enabling-messagemode-for-read-only-name

That question/answer was posted by me back in 2015 describing a defect and a workaround for an issue with access permissions when creating named pipes in the .NET Framework. A bug report was made on Microsoft Connect, and never actioned. The same defect was inherited when .NET Core came into being, however the workaround we could do in the .NET Framework days is no longer possible, as it is no longer to manually control access permissions to work around the issue. I'm contributing the proper fix here. Areas of reference from MSDN:
https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createnamedpipea

See in particular this excerpt:

> A client process connects to a named pipe by using the [CreateFile](https://learn.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-createfilea) or [CallNamedPipe](https://learn.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-callnamedpipea) function. The client side of a named pipe starts out in byte mode, even if the server side is in message mode. To avoid problems receiving data, set the client side to message mode as well. To change the mode of the pipe, the pipe client must open a read-only pipe with GENERIC_READ and FILE_WRITE_ATTRIBUTES access.

This change simply brings the implementation inline with the MSDN recommendations. Since pipe message mode is Windows-specific, this doesn't affect other platforms.

RogerSanders:
@dotnet-policy-service agree

RogerSanders:
It looks like there are more unit tests to adjust which only run under "privileged" mode, but I didn't use an elevated command prompt to test so I missed them. I'll prepare a fix.

RogerSanders:
Ok, so there's a backwards compatibility issue the failing test was correct to complain about. There's no issue if the named pipe the client is trying to connect to has been created with default access permissions, and it was either created under the same user account or an administrator/system account.  There's also no problem if the pipe has been created with correctly set manual permissions to support message mode pipes. By default however, other accounts on the system (IE, basic users) are only granted "Read" access to the pipe, without the "WriteAttributes" permission. Attempts to open pipes under these circumstances now fail with this change as I first submitted it. This could also occur if the pipe was created with a manual permissions set which disallowed WriteAttributes. It makes sense that any pipe a client was able to connect to prior to this change, it must also be able to connect to after it, so the original change is not suitable.

Fixing this now requires a little bit of a compromise. On the client end, we don't know whether the "WriteAttributes" permission has been granted to the current user until we try and open the pipe. The NamedPipeClientStream class also doesn't know if message mode will be requested by the caller prior to opening the pipe, as the API doesn't require the caller to provide that information. We also can't obtain this permission after opening the pipe to my knowledge, nor can we safely disconnect and reconnect with the extra permission if we discover after connecting that we do require it (IE, when the caller changes ReadMode).

The failure case introduced by the fix here is unusual, as it will ONLY occur specifically on Windows if a single-directional named pipe is created, where data flows from the server to the client, where the server and client are running under different user accounts, and the account running the client is running as an unprivileged user. That is a very narrow case.

To fix this, I propose we attempt to open the named pipe with the WriteAttributes permission as per this change by default. If that fails with a privilege issue, we drop the WriteAttributes permission and try again. This double-connect attempt would only occur internally, transparently to the caller, in the specific case where we detect this privilege issue. This means in the 99.9% case here the pipe will only be attempted to be opened once, and we will now also be able to support message mode read-only pipes from server to client, which are completely broken and not supported right now in non-framework .NET versions with no workaround.

I'll prepare another changeset with this proposed modification.

RogerSanders:
Paging @jeffhandley, as I can't see if you got pulled in correctly by the codeowners check.

Jozkee:
@RogerSanders thanks for working on this. I think it would be best to follow the advice on https://github.com/dotnet/runtime/issues/83072#issuecomment-1461040626 to avoid granting `FILE_WRITE_ATTRIBUTES` by default without existing consumers of the API noticing. Would you be interested in following the [steps](https://github.com/dotnet/runtime/blob/main/docs/project/api-review-process.md#steps) to make the API proposal for that?

cc @ericstj.

RogerSanders:
Well, options around API changes would need more discussion. I intentionally tried to avoid making any public API changes with this fix. There's a clear issue in the API as it currently stands, as there's broken functionality with the ReadMode property, and this fix was trying to address this internally. If you want to talk about API changes, there are tradeoffs with all those that I can imagine too.

We _could_ restore the ability to specify `PipeAccessRights` on a constructor to `NamedPipeClientStream` as proposed in #83072, however I'd point out that this would be non-sensical on non-Windows platforms, and would still be non-discoverable for the user. They'd still get errors trying to set ReadMode, with no clear indication why, and they'd have to search and find they had to use a special constructor with an arbitrary set of access rights to make it work. This would effectively leave the functionality "broken" but restore the workaround from the .NET Framework days. I wouldn't recommend this route.

There's another option. There's a constructor for `NamedPipeClientStream` which takes a `PipeOptions` flag enum. You could add a new flag like `AllowMessageMode` to this enum, which if set explicitly informs the constructor the caller intends to set `ReadMode` to `Message`, in which case the constructor can attempt to obtain the necessary WriteAttributes permission, failing in this case if it can't be obtained. If this was a new API with no existing usage, that would seem to be the right choice. This isn't a new API though. If you were to go with this option, no existing code in the world currently sets this flag, but message mode will still work for write-only or read/write pipes, which would also seem strange. On a new API with no existing code usage, you'd probably want to reject the attempt to set ReadMode to Message in all cases if this flag had never been passed into the constructor, which would bring the client side into line with how things work on the `NamedPipeServerStream` class.

So it seems there are only two viable options to fix this from my perspective:

1. Make the WriteAttributes access right be obtained internally where possible as an implementation detail (as per this change right now). Only downside - this access right will now be requested in some cases when it was available previously, but not taken or used previously.

2. Make the WriteAttributes access right be obtained explicitly when a new `AllowMessageMode` flag is set. Two downsides - the public API and documentation requires changes to reflect this, and there's an extra step to make Message mode work properly which isn't immediately apparent (passing in a special flag), which for purposes of backwards compatibility needs to work anyway if the flag hasn't been provided for write-only and read-write pipes.

Which approach would you like me to continue with? I could do another fix which implements option two, but would this one be accepted? It seems to me the third option of doing nothing and leaving the message functionality completely broken isn't viable either, so it's one or the other.

Jozkee:
> this would be non-sensical on non-Windows platforms

Just like [`PipeTransmisisonMode.Message`](https://github.com/dotnet/runtime/blob/867b3e53094f2960f0bcc8b10ea2744bae538e0b/src/libraries/System.IO.Pipes/src/System/IO/Pipes/PipeTransmissionMode.cs#L11-L12) currently is, so that argument doesn't hold IMO. What would `AllowMessageMode` do on non-Windows?

Jozkee:
> would still be non-discoverable for the user. They'd still get errors trying to set ReadMode, with no clear indication why, and they'd have to search and find they had to use a special constructor with an arbitrary set of access rights to make it work.

Do you think that improving the error message could address this? We could append something like "Pipes need `Write` or `WriteAttributes` access in order to specify the transmission mode".

RogerSanders:
> > this would be non-sensical on non-Windows platforms
> 
> Just like [`PipeTransmisisonMode.Message`](https://github.com/dotnet/runtime/blob/867b3e53094f2960f0bcc8b10ea2744bae538e0b/src/libraries/System.IO.Pipes/src/System/IO/Pipes/PipeTransmissionMode.cs#L11-L12) currently is, so that argument doesn't hold IMO. What would `AllowMessageMode` do on non-Windows?

Emphasis on _currently_. Message mode isn't Windows specific, it's a logical peice of functionality which enables a form of atomicity/packets in reads/writes to a pipe. The pipe access permissions like WriteAtrributes however are highly Windows specific. I think it would be a mistake to specifically rely on exposing those attributes to the caller in order to allow them to properly enable message mode pipes. What happens if Linux or another .NET supported platform adds message mode for named pipes in the future?

This isn't as hypothetical as you may think. The pipe2 method supports the O_DIRECT flag as of kernel version 3.4, going back 10 years. That is sufficient go allow the concept of message mode to be implemented on unnamed pipes. In kernel version 4.5 (2016), it became possible to enable O_DIRECT mode for named pipes via a call to sys_fcntl. Theoretically that is enough to impmement this concept. I could add a patch to implement message mode support on Linux today... but if the API changes as you propose, it would become pretty messy. To get reliable cross platform support, the caller would be required to use a Windows-specific constructor on that platform.

RogerSanders:
To be clear, if you tell me the only way forward here that will be accepted is to create a new constructor with the PipeAccessRights enum and throw a more meaningful error, I'm willing to do that. I'd like to see this issue fixed one way or the other. I'm just giving some feedback on the proposed API changes from a user perspective. Ultimately it's not my call, you let me know what you'd like to see.

Jozkee:
Well yes, I thought that Message mode was only available on Windows only based on the `[SupportedOSPlatform]`. Also, I just  saw an issue that we currently have no good ideas in how to implement Message mode in Unix https://github.com/dotnet/runtime/issues/24097.

If you have a POC for it, can you please create a [new issue](https://github.com/dotnet/runtime/issues/new/choose)? Either a blank issue or API suggestion, but it would be good to discuss it before you dive into making a full POC just to get validation/ideas from other experienced folks and if its good enough, we will incorporate it into the framework.

Jozkee:
Also consider:

Named pipes in Unix are implemented using Sockets, not fifos, see https://github.com/dotnet/corefx/pull/6833, this is due to limitations vs Windows, therefore using O_DIRECT may not be possible.

Anonymous pipes on Windows don't support Message mode, so currently there's no story of making both platforms symmetric here, however, it is fair to add support to the platform that supports it. That should be discussed in a separate issue.

Going back to my original point, if it is not possible to implement Message mode for named pipes in Unix, then it seems to me that adding the Windows-only API is the way to go since it already existed in Windows. We have prior-art of APIs that are supported only in one platform e.g: UnixFileMode. It is not a sin to do https://github.com/dotnet/runtime/issues/83072#issuecomment-1461040626.

RogerSanders:
> Also consider:
> 
> Named pipes in Unix are implemented using Sockets, not fifos, see [dotnet/corefx#6833](https://github.com/dotnet/corefx/pull/6833), this is due to limitations vs Windows, therefore using O_DIRECT may not be possible.
[#83072 (comment)]
That's a surprising choice. Wow, what a mess. Since pipes are one of the primary mechanisms of IPC, it seems like a poor choice for the .NET framework to make, as you can't expect both the client and the server to be using the same language/tech stack, which may make it impossible to connect to existing systems that are built on fifos. I see an issue raised around that very point:
(https://github.com/dotnet/docs/issues/36710)

The primary advantage of pipes/fifos is performance. Being little more than a thin layer over shared memory, nothing else can get close. I would have expected pipes on windows to be replicated with fifos on Unix, not sockets, they're the natural matched concept.

It's true you can't do a bi-directional fifo on Unix. You can do two unidirectional fifos however, with the "reverse connection" established after a connection request from the client. This is how we (my company) currently abstract pipes for Windows/Linux, we don't use InOut mode, we use a pair of pipes, one In one Out. If I'd been implementing this functionality in .NET core originally, I would have made InOut mode Windows-specific (which it is currently), throw on other platforms when attempting to use it, and stayed with fifos, which would have then given high performance and message mode support on Unix.

OK, I've got a more radical suggestion. Add two new flags to the API, one which says AllowMessageMode as I suggested, and another which is called.... I don't know, NoDomainSockets or ForceNativeFifo or something like that, which as it suggests, actually uses FIFOs on Unix, and can then implement message mode. If you attempt to create a pipe with InOut mode on Unix, throw an error. That fixes the problem I've described here, and also resolves the lack of fifo support on Unix, which honestly is a pretty big limitation for compatibility and integration with existing systems, as well as a major performance drawback.

Since .NET has already effectively done a language-specific implementation for domain sockets anyway on Unix, you could go a step further and implement message mode under Unix without fifos by doing some framing with a small header (IE, payload byte count) and encoding/decoding packets internally anyway - it's what we currently do for sending messages across fifos on Unix. That's walking even further into a language-specific implementation, but that horse already bolted when domain sockets were used instead of FIFOs. Maybe you need a pair of flags, ForceUnixDomainSocket and ForceUnixFifo, which if you don't specify, the implementation is an undefined, runtime specific protocol. That's basically the case now, it's just not documented that way.

RogerSanders:
Again to be clear, I have no problems opening a separate issue with an API proposal detailing this, and even a working implementation, I just figured I'd bounce a few ideas off you in this issue here and get some feedback to help craft that proposal.

Jozkee:
I think your idea is worth of discussion with a broader forum, I suggest, again, that you file a new issue describing AllowMessageMode and ForceNativeFifo, displaying the value they would provide (perf, compatibility with other platforms, etc). 

Since .NET 9 development is just starting, we have plenty of time to evaluate if that is an approach worth chasing; if not, we can fallback to move the `PipeAccessRights` and adding the respective `ctor`.

PR #94759, opened 2023-11-15
Fix declaration of DiagAddNewRegion

MichalStrehovsky:
`uint8_t` is the type that is used thorough the codebase here. `BYTE` happens to not be defined for some build configurations and produces a build break. I'm also removing the space because the prevailing code style in this file is no space between type and asterisk.


cshung (approved):
Thanks!

PR #94762, opened 2023-11-15
[RISC-V] Flush-to-zero behavior for float-to-int conversion

denis-paranichev:
RISC-V specification does not provide option which enables flushing to zero return value of float-to-int conversion instructions in the case when input value is NaN or result can't be represented in the target type.
Thus we need to enable this option to support deterministic behavior across platforms.

cc @gbalykov @t-mustafin @clamp03 @tomeksowi @brucehoult @tannergooding


src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -3374,8 +3374,16 @@ void CodeGen::genFloatToIntCast(GenTree* treeNode)
     }
 
     genConsumeOperands(treeNode->AsOp());
+    regNumber tmpReg = rsGetRsvdReg();
 
     GetEmitter()->emitIns_R_R(ins, EA_8BYTE, treeNode->GetRegNum(), op1->GetRegNum());
+
+    // This part emulates the "flush to zero" option because the RISC-V specification does not provide it.
+    GetEmitter()->emitIns_R_I_I(INS_csrrci, EA_8BYTE, tmpReg, 22, 1);
```
tomeksowi, line 3382: Would be good to have 22 and 1 as named constants, it's hard to discern the intention just looking at the code.

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -3374,8 +3374,16 @@ void CodeGen::genFloatToIntCast(GenTree* treeNode)
...
+    GetEmitter()->emitIns_R_R_I(INS_andi, EA_8BYTE, tmpReg, tmpReg, 22);
+    GetEmitter()->emitIns_R_R_I(INS_beq, EA_8BYTE, tmpReg, REG_R0, 8);
```
tomeksowi, line 3384: Would be much better to jump to a label rather than hardcoded offset.

Besides, why 8? It means 2 instructions, what other instruction do you intend to skip except addiw below?

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3360,6 +3396,48 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
+                        printf("csrrc           %s, %d, %s\n", rd, csrtype, rs1);
+                        return;
+                    default:
+                        printf("RISCV64 illegal instruction: 0x%08X\n", code);
+                        break;
+                }
+            }
+            else
+            {
+                int imm12 = (((int)code) >> 15);
```
tomeksowi, line 3423: I think it's imm5 and I believe it needs to be masked off.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3360,6 +3396,48 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
+                switch (opcode2)
+                {
+                    case 0x5: // CSRRW
```
tomeksowi, line 3426: Nit: These comments can be removed, it's clear what they mean looking at the printfs. But if you want to keep them, it's CSRRW**I**.

tomeksowi (changes requested):

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -3374,8 +3374,16 @@ void CodeGen::genFloatToIntCast(GenTree* treeNode)
     }
 
     genConsumeOperands(treeNode->AsOp());
+    regNumber tmpReg = rsGetRsvdReg();
```
tomeksowi, line 3377: Does it need really need to be t6 or could you let the compiler assign any temp reg?

src/coreclr/jit/codegenriscv64.cpp
gbalykov, line 3382: It seems that you reset `NX`(0) and `DZ`(4) bits here. `FCVT.L.S`, `FCVT.LU.S`, `FCVT.L.D`, `FCVT.LU.D` set only `NV`(5) and `NX`(0) according to spec. Why reset `DZ`?

gbalykov: Also would be good to have comment here, that you get `fflags` portion of `fcsr`

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -3374,8 +3374,16 @@ void CodeGen::genFloatToIntCast(GenTree* treeNode)
...
 
     GetEmitter()->emitIns_R_R(ins, EA_8BYTE, treeNode->GetRegNum(), op1->GetRegNum());
+
+    // This part emulates the "flush to zero" option because the RISC-V specification does not provide it.
+    GetEmitter()->emitIns_R_I_I(INS_csrrci, EA_8BYTE, tmpReg, 22, 1);
+    GetEmitter()->emitIns_R_R_I(INS_andi, EA_8BYTE, tmpReg, tmpReg, 22);
```
gbalykov, line 3383: Why do we check `UF`(1, underflow), `OF`(2, overflow) and `NV`(4, invalid operation) here bits?

tannergooding:
> which enables flushing to zero return value of float-to-int conversion instructions in the case when input value is NaN or result can't be represented in the target type.

@DenisParal, I think there may have been a slight miscommunication.

The behavior is not "any unrepresentable value becomes 0", but rather the behavior is `saturation` for `finite` and `infinite` values, and with `NaN` becoming `0` (as there is no other sensible behavior).

tomeksowi:
> the behavior is `saturation` for `finite` and `infinite` values, and with `NaN` becoming `0`

In that case, only the NaN behavior needs mending, this can be done branchless on RV, sth like:
```
fcvt dest, src       # convert
feq temp, src, src   # NaN check
neg temp, temp       # zero if NaN, all 1s otherwise
and dest, dest, temp # mask the result
```

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -700,6 +700,12 @@ void emitter::emitIns_R_R_I(
         code |= ((imm >> 5) & 0x3f) << 25;
         code |= ((imm >> 12) & 0x1) << 31;
     }
+    else if (ins == INS_csrrs || ins == INS_csrrw || ins == INS_csrrc)
```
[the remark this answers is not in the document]
DenisParal: These instructions have different restrictions: in this case rd and rs can be zero register.

gbalykov:
> fcvt dest, src       # convert
> feq temp, src, src   # NaN check
> neg temp, temp       # zero if NaN, all 1s otherwise
> and dest, dest, temp # mask the result

@tomeksowi why is `neg` needed? `feq` will set `temp` to 0 if `src` is NaN.

tomeksowi:
> @tomeksowi why is `neg` needed? `feq` will set `temp` to 0 if `src` is NaN.

But it's 1 if not NaN, we need all an 1s mask (-1).

DenisParal:
Fixed flash-to-zero behavior to react only on NaN input.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -715,6 +721,36 @@ void emitter::emitIns_R_R_I(
...
+void emitter::emitIns_R_I_I(
+    instruction ins, emitAttr attr, regNumber reg1, ssize_t imm1, ssize_t imm2, insOpts opt) /* = INS_OPTS_NONE */
+{
+    code_t code = emitInsCode(ins);
+
+    if (INS_csrrwi <= ins && ins <= INS_csrrci)
+    {
+        code |= reg1 << 7;
+        code |= imm1 << 15;
+        code |= imm2 << 20;
```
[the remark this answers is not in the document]
DenisParal: Done. Thank you for comment.

src/coreclr/jit/codegenriscv64.cpp
DenisParal: Added mask constant and comment for 1 (fflags)

src/coreclr/jit/emitriscv64.cpp
DenisParal: Done. Thank you for comment.

src/coreclr/jit/emitriscv64.cpp
DenisParal: Fixed

src/coreclr/jit/codegenriscv64.cpp
DenisParal: Added constant mask for 22 and comment for 1

src/coreclr/jit/codegenriscv64.cpp
DenisParal: Changed behavior to react only on NaN input.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3360,6 +3402,48 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
+                        break;
+                }
+            }
+            else
+            {
+                int imm5 = ((code >> 15) & 0x1f);
+                switch (opcode2)
+                {
+                    case 0x5: // CSRRWI
+                        printf("csrrwi           %s, %d, %s\n", rd, csrtype, imm5);
```
clamp03, line 3433: imm5 is integer type. Please update `%s` in format string to the proper one.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3360,6 +3402,48 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
+                        return;
+                    case 0x6: // CSRRSI
+                        printf("csrrsi           %s, %d, %s\n", rd, csrtype, imm5);
```
clamp03, line 3436: Update here.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3360,6 +3402,48 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
+                        return;
+                    case 0x7: // CSRRCI
+                        printf("csrrci           %s, %d, %s\n", rd, csrtype, imm5);
```
clamp03, line 3439: Update here too.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3360,6 +3402,48 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
+                        printf("csrrc           %s, %d, %s\n", rd, csrtype, rs1);
+                        return;
+                    default:
+                        printf("RISCV64 illegal instruction: 0x%08X\n", code);
+                        break;
+                }
+            }
+            else
+            {
+                int imm5 = ((code >> 15) & 0x1f);
```
clamp03, line 3429: Nit: this value is `the five-bit zero-extended immediate`. I think unsigned int is better.

clamp03 (changes requested):

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -3374,8 +3374,17 @@ void CodeGen::genFloatToIntCast(GenTree* treeNode)
     }
 
     genConsumeOperands(treeNode->AsOp());
+    regNumber tmpReg = rsGetRsvdReg();
 
     GetEmitter()->emitIns_R_R(ins, EA_8BYTE, treeNode->GetRegNum(), op1->GetRegNum());
+
+    // This part emulates the "flush to zero" option because the RISC-V specification does not provide it.
+    unsigned mask = 16; // NV = 1 | DZ = 0 | OF = 0 | UF = 0 | NX = 0
```
clamp03, line 3382: Nit: How about `0b10000`?

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -3374,8 +3374,17 @@ void CodeGen::genFloatToIntCast(GenTree* treeNode)
...
+    GetEmitter()->emitIns_R_I_I(INS_csrrci, EA_8BYTE, tmpReg, mask, 1/*fflags*/);
```
clamp03, line 3383: Could you share any document which NaN input set the invalid operation exception flag in `fcvt_*`? Actually, I cannot find. Thank you.

clamp03 (changes requested):

src/coreclr/jit/codegenriscv64.cpp
DenisParal: Yes, unprivilege specification - page 69 - "Single-Precision Floating-Point Conversion and Move
Instructions"
https://drive.google.com/file/d/1s0lZxUZaa7eV_O0_WsZzaurFLLww7ou5/view?usp=drive_link
"If the rounded result is not representable in the destination format, it
is clipped to the nearest value and the invalid flag is set."
Thus not only NaN input, but any input that can't be represented in target type sets the invalid flag.

DenisParal:
@tannergooding, I have tested `float.MaxValue` conversion to `int` on x64 and it seems like this input also causes flushing the result to zero. Is it expected that RISC-V has different result for such input?
Here is reproducer 
[sample.tar.gz](https://github.com/dotnet/runtime/files/13391507/sample.tar.gz)

tannergooding:
> I have tested float.MaxValue conversion to uint on x64 and it seems like this input also causes flushing the result to zero

As mentioned on the other threads that caused this PR, we are working on normalizing all platforms, including x64, to follow the standardized behavior. This work is tracked by https://github.com/dotnet/runtime/issues/61885

The work to correctly handle x64 is being handled in parallel and will end up matching the stated behavior long term:
* NaN becomes 0
* Saturation occurs for finite and infinite inputs that are out of range

src/coreclr/jit/emitriscv64.cpp
DenisParal: Done. Thank you for comment.

src/coreclr/jit/emitriscv64.cpp
DenisParal: Done.

src/coreclr/jit/emitriscv64.cpp
DenisParal: Done.

src/coreclr/jit/emitriscv64.cpp
DenisParal: Done.

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -3374,8 +3374,17 @@ void CodeGen::genFloatToIntCast(GenTree* treeNode)
...
+    GetEmitter()->emitIns_R_R_I(INS_andi, EA_8BYTE, tmpReg, tmpReg, mask);
+    GetEmitter()->emitIns_R_R_I(INS_beq, EA_8BYTE, tmpReg, REG_R0,  2 << 2);
+    GetEmitter()->emitIns_R_R_I(INS_addiw, EA_8BYTE, treeNode->GetRegNum(), REG_R0, 0);
```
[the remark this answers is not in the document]
DenisParal: The implementation has changed.

src/coreclr/jit/codegenriscv64.cpp
DenisParal: The implementation has changed.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -715,6 +721,36 @@ void emitter::emitIns_R_R_I(
     appendToCurIG(id);
 }
 
+/*****************************************************************************
+ *
+ *  Add an instruction referencing register and two constants.
+ */
+
+void emitter::emitIns_R_I_I(
+    instruction ins, emitAttr attr, regNumber reg1, ssize_t imm1, ssize_t imm2, insOpts opt) /* = INS_OPTS_NONE */
```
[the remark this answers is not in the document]
DenisParal: Yes, but this interface does not only describe csr instructions. I have added uint check for imm1.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3411,13 +3411,13 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
-                        printf("csrrw           %s, %d, %s\n", rd, csrtype, rs1);
+                        printf("csrrw           %s, %d, %d\n", rd, csrtype, rs1);
                         return;
                     case 0x2: // CSRRS
-                        printf("csrrs           %s, %d, %s\n", rd, csrtype, rs1);
+                        printf("csrrs           %s, %d, %d\n", rd, csrtype, rs1);
                         return;
                     case 0x3: // CSRRC
-                        printf("csrrc           %s, %d, %s\n", rd, csrtype, rs1);
+                        printf("csrrc           %s, %d, %d\n", rd, csrtype, rs1);
```
clamp03, line 3420: You don't need to update these `csrrw`, `csrrs` and `csrrc`. `rs1` is `const char*`.

clamp03:
Please update code formatting with `jit-format` https://github.com/dotnet/jitutils/blob/main/doc/formatting.md

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3360,6 +3402,48 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
+                        printf("csrrw           %s, %d, %d\n", rd, csrtype, rs1);
+                        return;
+                    case 0x2: // CSRRS
+                        printf("csrrs           %s, %d, %d\n", rd, csrtype, rs1);
+                        return;
+                    case 0x3: // CSRRC
+                        printf("csrrc           %s, %d, %d\n", rd, csrtype, rs1);
+                        return;
+                    default:
+                        printf("RISCV64 illegal instruction: 0x%08X\n", code);
```
clamp03, line 3423: In case of `ebreak`, it prints `RISCV64 illegal instruction: ...` Please update it to work with `ebreak` well.

DenisParal: Thank you, fixed.

src/coreclr/jit/emitriscv64.cpp
DenisParal: Fixed.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -3360,6 +3402,51 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
         }
         case 0x73:
         {
+            unsigned int opcode2 = (code >> 12) & 0x7;
+            if (0x1 <= opcode2 && opcode2 <= 0x7)
```
clamp03, line 3406: I think it means `opcode2 != 0`.

src/coreclr/jit/codegenriscv64.cpp
DenisParal: I have tested GenTree::ExtractTempReg, and I got assert `genCountBits(availableSet) >= 1` with the `RBM_ALLINT` mask. Thus it seems like there is no temp reg available in such context. Do you have any clue why would it happened?
Reproducer: [sample.tar.gz](https://github.com/dotnet/runtime/files/13425842/sample.tar.gz)

tomeksowi: AFAIK you need to allocate it with buildInternalIntRegisterDefForNode and buildInternalRegisterUses in lsrarisc64.cpp. See GT_CMPXCHG for analogy.

clamp03 (approved):

src/coreclr/jit/codegenriscv64.cpp
tannergooding: Right. We should prefer having LSRA indicate a temp register is required and then to extract that here, over forcing a particular to be used.

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -3374,8 +3374,20 @@ void CodeGen::genFloatToIntCast(GenTree* treeNode)
...
+
+    // This part emulates the "flush to zero" option because the RISC-V specification does not provide it.
+    instruction feq_ins = INS_feq_s;
+    if (srcType == TYP_DOUBLE)
+    {
+        feq_ins = INS_feq_d;
+    }
+    GetEmitter()->emitIns_R_R_R(feq_ins, EA_8BYTE, tmpReg, op1->GetRegNum(), op1->GetRegNum());
+    GetEmitter()->emitIns_R_R_R(INS_sub, EA_8BYTE, tmpReg, REG_R0, tmpReg);
+    GetEmitter()->emitIns_R_R_R(INS_and, EA_8BYTE, treeNode->GetRegNum(), treeNode->GetRegNum(), tmpReg);
```
tannergooding, line 3389: This likely needs a comment explaining what its doing.

`feq_s` compares `op1` to itself. I'm assuming this produces `0` for `false` and `1` for `true`. This means `0` for `NaN` and `1` for anything else.

For `sub`, i'm presuming that `REG_R0` (if so, an alias might really help readability/intent) is always zero and so we're doing `result = 0 - result`, thus producing `0` for `NaN` and `-1` (AllBitsSet) for `true`

`and` then produces `0` for `NaN` and preserves the original for anything else.

tannergooding (approved):
The logic, assuming I understood the RISC-V instructions behavior correctly; looks correct to me.

Someone from @dotnet/jit-contrib needs to review/sign-off on the other changes.

src/coreclr/jit/codegenriscv64.cpp
DenisParal: Done.

src/coreclr/jit/codegenriscv64.cpp
DenisParal: Fixed.

tomeksowi (approved):

DenisParal:
A lot of System.Tests.UriMethodTests are failed with last commit, I work on fixing it.

clamp03:
@DenisParal 
I found that `tmpReg` allocated to the same register of `treeNode->GetRegNum()` in your code.
I implemented code like the below. System.Tests.UriMethodTests passed.
Could you review the codes and tests coreclr tests and corefx tests?
Thank you.
```
diff --git a/src/coreclr/jit/lsrariscv64.cpp b/src/coreclr/jit/lsrariscv64.cpp
index 2cf45116227..70396764d30 100644
--- a/src/coreclr/jit/lsrariscv64.cpp
+++ b/src/coreclr/jit/lsrariscv64.cpp
@@ -326,7 +326,9 @@ int LinearScan::BuildNode(GenTree* tree)
 
         case GT_CAST:
             assert(dstCount == 1);
-            buildInternalIntRegisterDefForNode(tree);
             srcCount = BuildCast(tree->AsCast());
-            buildInternalRegisterUses();
             break;
 
         case GT_NEG:
@@ -1269,6 +1271,12 @@ int LinearScan::BuildCast(GenTreeCast* cast)
     int srcCount = BuildOperandUses(cast->CastOp());
     BuildDef(cast);
 
+    if (varTypeIsFloating(cast->gtOp1) && !varTypeIsFloating(cast->TypeGet()))
+    {
+        buildInternalIntRegisterDefForNode(cast);
+        buildInternalRegisterUses();
+    }
+
     return srcCount;
 }
```

```
diff --git a/src/coreclr/jit/codegenriscv64.cpp b/src/coreclr/jit/codegenriscv64.cpp
index c2dafce1fb0..d708d4aab5c 100644
--- a/src/coreclr/jit/codegenriscv64.cpp
+++ b/src/coreclr/jit/codegenriscv64.cpp
@@ -3379,9 +3379,26 @@ void CodeGen::genFloatToIntCast(GenTree* treeNode)
     }
 
     genConsumeOperands(treeNode->AsOp());
+    regNumber tmpReg = treeNode->GetSingleTempReg();
+    assert(tmpReg != treeNode->GetRegNum());
+    assert(tmpReg != op1->GetRegNum());
 }
```

DenisParal:
@clamp03, thank you for suggestion, it looks better than my solution. coreclr tests are passed thus remove draft status.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -700,6 +700,15 @@ void emitter::emitIns_R_R_I(
         code |= ((imm >> 5) & 0x3f) << 25;
         code |= ((imm >> 12) & 0x1) << 31;
     }
+    else if (ins == INS_csrrs || ins == INS_csrrw || ins == INS_csrrc)
+    {
+        assert(isGeneralRegisterOrR0(reg1));
+        assert(isGeneralRegisterOrR0(reg2));
+        assert(isValidSimm12(imm));
```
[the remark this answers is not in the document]
DenisParal: Fixed, thank you.

DenisParal:
@jakobbotsch could you please review this PR?

src/coreclr/jit/registerriscv64.h
```diff
@@ -50,6 +50,7 @@ REGDEF(T5,     30, 0x40000000, "t5"  )
 REGDEF(T6,     31, 0x80000000, "t6"  )
 
 REGALIAS(R8, FP)
+REGALIAS(ZERO, R0)
```
jakobbotsch, line 53: Is this a common/official alias?

DenisParal: It is official according to RISC-V specification.

jakobbotsch (approved):

Issue #94081, opened 2023-10-27
[Tracking] Implement Mono support for Swift calling convention


kotlarmilos:
# Description

The Swift programming language has a different ABI, runtime environment, and object model. This issue tracks the work for implementing Mono support for Swift calling convention according to the [design document](https://github.com/dotnet/designs/blob/main/proposed/swift-interop.md). 

According to the calling convention, the self context has dedicated registers, and it is always passed through them since it's heavily used. Methods calling other methods on the same object can share the self context. Errors are also handled through registers, so the caller needs to check for errors and throw if necessary. Implementing P/Invoke thunks should simplify registers juggling by using predefined set of registers in these scenarios.

Below is a list of tasks that needs to be implemented. Please note that the breakdown of the issues does not impose hard deadlines. If some of the functionality does not get upstreamed for a particular preview version, this issue will be updated accordingly.

## Tasks
 - [ ] Frozen blittable structs lowering for sequence of length 4 or less
 - [x] https://github.com/dotnet/runtime/pull/98429
 - [x] https://github.com/dotnet/runtime/pull/94764
 - [ ] Implement UnmanagedCallersOnly support for Swift calling convention
 - [ ] Measure performance of direct Swift interop

ShikiSuen:
Please don't forget the InputMethodKit framework support for this time.

The InputMethodKit C# Support was absent during the entire Xamarin era.

kotlarmilos:
The remaining tasks have been moved to the parent tracking issue.

PR #94764, opened 2023-11-15
[mono] Add runtime support for Swift calling convention on x64 and arm64

kotlarmilos:
## Description

This PR introduces basic support for the Swift calling convention in the mini and interpreter backends on x64 and arm64 architectures. When a C# function calls a Swift library, it is expected that the self context is stored in a dedicated register. Similarly, when a Swift function returns, the error is supposed to be placed in a specific register.

In the method prolog, the self pointer is transferred to a specific register, and in the method epilog, the error register is moved to the stack and returned as an out argument. LLVM support will be addressed in a subsequent PR since the Mono runtime currently doesn't compile native wrappers using LLVM on Apple platforms.

Contributes to https://github.com/dotnet/runtime/issues/94081 and https://github.com/dotnet/runtime/issues/64215.


src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/CallingConvention.cs
```diff
@@ -11,5 +11,6 @@ public enum CallingConvention
         StdCall = 3,
         ThisCall = 4,
         FastCall = 5,
+        SwiftCall = 6,
```
jkotas, line 14: This should be reverted. We are not adding new calling conventions to this legacy enum.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Swift/SwiftType.cs
```diff
@@ -0,0 +1,47 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Runtime.InteropServices.Swift
+{
+    /// <summary>
+    /// Indicates that the argument is the self context.
+    /// </summary>
+    public readonly struct SwiftSelf
```
jkotas, line 9: These should go through API review before this PR is merged.

jkotas (dismissed):

src/mono/mono/mini/interp/interp.c
```diff
@@ -1710,6 +1710,16 @@ ves_pinvoke_method (
...
 #ifdef MONO_ARCH_HAVE_INTERP_PINVOKE_TRAMP
+	if (sig->call_convention == MONO_CALL_SWIFTCALL) {
+		int arg_index;
+		gpointer data = mono_arch_get_swift_error (&ccontext, sig, call_info, &arg_index);
+
+		// Perform an indirect store at arg_index stack location
+		if (data != NULL) {
+			stackval *result = (stackval*) STACK_ADD_BYTES (frame.stack, get_arg_offset (frame.imethod, sig, arg_index));
+			*(gpointer*)result->data.p = *(gpointer*)data;
+		}
```
lambdageek, line 1721: this could be under `#ifdef MONO_ARCH_HAVE_SWIFTCALL`.  Also in the case where we don't have swift interop support, we should throw `NotImplementedException` or something if we see a call with a swift calling convention

src/native/public/mono/metadata/details/metadata-types.h
```diff
@@ -29,7 +29,7 @@ typedef enum {
 	MONO_CALL_THISCALL,
 	MONO_CALL_FASTCALL,
 	MONO_CALL_VARARG = 0x05,
-	/* unused, */
+	MONO_CALL_SWIFTCALL = 0x06,
```
lambdageek, line 32: 1. We should not extend this enum. it's supposed to mirror the 
metadata encoding.

2. This is the public Mono API, we shouldn't just add things here unless there's a need.

---

We should handle this the same way as "SuppressGCTransition": use a bit in `MonoMethodSignature` to note that it's a swift call -  see `MonoMethodSignature:suppress_gc_transition`.

Another idea is to make a `MonoExtCallConv` enum (not in this file, in `src/mono/metadata/metadata-internals.h`) and then do:

```c
typedef enum {
  MONO_EXT_CALLCONV_SUPPRESS_GC_TRANSITION = 0x01,
  MONO_EXT_CALLCONV_SWIFTCALL = 0x02,
  /// see MonoMethodSignature:ext_callconv - only 8 bits
} MonoExtCallConv;

....

typedef struct _MonoMethodSignature {
  ...
  uint8_t    ext_callconv; // see MonoExtCallConv 
} MonoMethodSignature;

static inline gboolean
mono_method_signature_has_ext_callconv (MonoMethodSignature *sig, MonoExtCallConv flags)
{
   return (sig->callconv_ext & flags) != 0;
}
```

src/mono/mono/metadata/tabledefs.h
```diff
@@ -251,6 +251,7 @@ enum {
 #define PINVOKE_ATTRIBUTE_CALL_CONV_STDCALL   0x0300
 #define PINVOKE_ATTRIBUTE_CALL_CONV_THISCALL  0x0400
 #define PINVOKE_ATTRIBUTE_CALL_CONV_FASTCALL  0x0500
+#define PINVOKE_ATTRIBUTE_CALL_CONV_SWIFTCALL  0x0600
```
lambdageek, line 254: these defines are supposed to match what's in ECMA-335 (or the augments). the swift call isn't going to be in there. so this is not needed.

src/mono/System.Private.CoreLib/src/System/Reflection/RuntimeMethodInfo.Mono.cs
```diff
@@ -69,6 +69,7 @@ internal enum PInvokeAttributes
         CallConvStdcall = 0x0300,
         CallConvThiscall = 0x0400,
         CallConvFastcall = 0x0500,
+        CallConvSwift = 0x0600,
```
jkotas, line 72:
```suggestion
```
?

src/tests/Interop/Swift/README.md
```diff
@@ -0,0 +1,96 @@
+# .NET Swift Interop
```
jkotas, line 1: https://github.com/dotnet/designs/blob/main/proposed/swift-interop.md is more complete. Should this doc be deleted?

src/tests/Interop/Swift/SelfContext.cs
```diff
@@ -0,0 +1,32 @@
+using System;
```
jkotas, line 1: .cs files should have license headers.

src/tests/Interop/CMakeLists.txt
```diff
@@ -103,4 +103,5 @@ endif(CLR_CMAKE_TARGET_UNIX)
 if(CLR_CMAKE_TARGET_APPLE)
     add_subdirectory(ObjectiveC/AutoReleaseTest)
     add_subdirectory(ObjectiveC/ObjectiveCMarshalAPI)
+    add_subdirectory(Swift)
```
AaronRobinsonMSFT, line 106: Please organize these tests in sub directories.

src/tests/Interop/Swift/Dummy.swift
```diff
@@ -0,0 +1 @@
+public func dummyFunc() { }
```
AaronRobinsonMSFT, line 1: This doesn't seem like a helpful description. It would be helpful to name these in a manner that describes what it is for or what it is, perhaps `voidVoidFunc` or `simpleFunc`?

src/tests/Interop/Swift/ErrorHandling.cs
```diff
@@ -0,0 +1,60 @@
+using System;
```
AaronRobinsonMSFT, line 1: License banner.

src/tests/Interop/Swift/ErrorHandling.swift
```diff
@@ -0,0 +1,47 @@
...
+public func getMyErrorMessage(from pointer: UnsafePointer<MyError>) -> UnsafePointer<UInt8>? {
+    let pointerValue = UInt(bitPattern: pointer)
+    var offsetValue: UInt
+#if arch(arm64)
+    offsetValue = 0x48
+#elseif arch(x86_64)
+    offsetValue = 0x20
+#else
+    fatalError("Unsupported architecture")
+#endif
```
AaronRobinsonMSFT, line 31: These numbers need comments.

src/tests/Interop/Swift/ErrorHandling.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    [UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvSwift) })]
+    [DllImport(SwiftLib, EntryPoint = "$s13ErrorHandling018conditionallyThrowA004willD0SiSb_tKF")]
+    public unsafe static extern nint conditionallyThrowError(bool willThrow, SwiftError* error);
+
+    [DllImport(SwiftLib, EntryPoint = "$s13ErrorHandling05getMyA7Message4fromSPys5UInt8VGSgSPyAA0dA0OG_tF")]
+    public static extern IntPtr GetErrorMessage(IntPtr handle);
+
+    [Fact]
+    public unsafe static int TestEntryPoint()
```
AaronRobinsonMSFT, line 23: This is an unhelpful test name. What "entry point"? Based on the test below, it seems this is for exceptions, but there are multiple cases being tested here.

src/tests/Interop/Swift/ErrorHandling.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        const string expectedErrorMessage = "Catch me if you can!";
+        SetErrorMessageForSwift(expectedErrorMessage);
+
+        SwiftError error;
+
+        // This will throw an error
+        conditionallyThrowError(true, &error);
+        if (error.Value == IntPtr.Zero) { 
+            return 101;
```
AaronRobinsonMSFT, line 33: Please don't return values. Instead throw an exception with a message.

src/tests/Interop/Swift/InvalidCallingConv.cs
```diff
@@ -0,0 +1,51 @@
...
+
+    [Fact]
+    public unsafe static int TestEntryPoint()
+    {
+        SwiftSelf self = new SwiftSelf(IntPtr.Zero);
+        SwiftError error = new SwiftError(IntPtr.Zero);
+
+        try {
+            DummyFunc1(self, self);
+            return 101;
```
AaronRobinsonMSFT, line 35: These should all be `Assert.Fail()` with a message.

src/tests/Interop/Swift/InvalidCallingConv.cs
```diff
@@ -0,0 +1,51 @@
...
+    public static extern nint DummyFunc4(SwiftError error1);
+
+    [Fact]
+    public unsafe static int TestEntryPoint()
+    {
+        SwiftSelf self = new SwiftSelf(IntPtr.Zero);
+        SwiftError error = new SwiftError(IntPtr.Zero);
+
+        try {
+            DummyFunc1(self, self);
```
AaronRobinsonMSFT, line 34: It would also be helpful to comment as to why. If this fails, what went wrong? These tests don't address that.

src/tests/Interop/Swift/SelfContext.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    [DllImport(SwiftLib, EntryPoint = "$s11SelfContext0A7LibraryC11getInstanceSvyFZ")]
+    public static extern IntPtr getInstance();
+
+    [UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvSwift) })]
+    [DllImport(SwiftLib, EntryPoint = "$s11SelfContext0A7LibraryC14getMagicNumberSiyF")]
+    public static extern nint getMagicNumber(SwiftSelf self);
+
+    [Fact]
+    public static int TestEntryPoint()
```
AaronRobinsonMSFT, line 19: What is this testing?

src/tests/Interop/Swift/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+The Swift programming language has a different ABI, runtime environment, and object model, making it challenging to call from the .NET runtime. Existing solutions, like [Binding Tools for Swift](https://github.com/xamarin/binding-tools-for-swift) and [BeyondNet](https://github.com/royalapplications/beyondnet).
+
+This project aims to explore the possibilities and limitations of direct P/Invoke interop with Swift. It should implement runtime mechanisms for handling Swift ABI differences. For a comprehensive .NET Swift interop, the Binding Tools for Swift contains valuable components that could be reused or built upon.
+
+We want to focus on the runtime support for direct P/Invoke interop with Swift, which Xamarin bindings will consume to support running Maui apps with third-party Swift libraries. Ideally, we want to avoid generating extra wrappers and attempt to directly call all kinds of Swift functions. While external tools can be helpful, they may introduce additional complexity, and we intend to integrate them into the .NET toolchain ecosystem.
+
+## Swift ABI in a nutshell
```
AaronRobinsonMSFT, line 9: This is great information. I think this should be somewhere similar to where we have our Objective-C support. For example https://github.com/dotnet/designs/blob/b70a81ac399e0737be73ff3acfb8a0591ff395e9/accepted/2021/objectivec-interop.md captured our Objective-C details. I think having something under https://github.com/dotnet/runtime/tree/main/docs/design

src/tests/Interop/Swift/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+The goal of this experiment is to explore the possibilities and limitations of direct P/Invoke interop with Swift. It should implement runtime mechanisms for handling Swift ABI differences. In the first iteration of the experiment, our focus will be on the self context and error handling calling conventions with blittable types targeting MacCatalyst. After that, we plan to expand support for non-blittable types using binding wrappers and to include other Apple targets.
+
+We can choose to first support Mono AOT either with or without LLVM according to the initial analysis.
+
+## Tasks
+
+Progress on the implementation can be tracked at https://github.com/dotnet/runtime/issues/93631.
+
+## Local build and testing
```
AaronRobinsonMSFT, line 77: This is a duplicate of data that should be under https://github.com/dotnet/runtime/tree/main/docs/workflow/testing

src/mono/mono/metadata/class-internals.h
```diff
@@ -978,6 +978,25 @@ mono_class_try_get_##shortname##_class (void)	\
 	return klass;	\
 }
 
+
+#define GENERATE_TRY_GET_CLASS_REF_WITH_CACHE(shortname,name_space,name) \
+MonoClass*	\
+mono_class_try_get_##shortname##_class (void)	\
+{	\
```
vargaz, line 985: Maybe name this ```mono_class_try_get_##shortname##_ptr_class```

src/mono/mono/mini/method-to-ir.c
```diff
@@ -11146,6 +11151,36 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
...
+							swift_self_args++;
+						}
+					}
+				}
+				if (swift_self_args > 1 || swift_error_args > 1) {
+					mono_error_set_invalid_program (swiftcall_error, g_strdup_printf ("Method signature contains multiple SwiftSelf/SwiftError arguments."), mono_method_full_name (method, TRUE));
+				}
+
+				if (!is_ok (swiftcall_error)) {
+                    emit_invalid_program_with_msg (cfg, swiftcall_error, method, cmethod);
```
vargaz, line 11179: Indentation.

src/mono/mono/mini/mini-arm64.h
```diff
@@ -271,6 +275,8 @@ typedef struct {
 	host_mgreg_t gregs [PARAM_REGS + 1];
 	/* Floating registers */
 	double fregs [FP_PARAM_REGS];
+	/* Context registers */
+	host_mgreg_t cregs [CTX_REGS];
```
vargaz, line 279: These can be stored in gregs, same as on amd64.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/CallingConvention.cs
kotlarmilos: Yes, reverted.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Swift/SwiftType.cs
kotlarmilos: Thanks. I will open a new PR for the API changes and revert this.

src/mono/mono/mini/mini.h
```diff
@@ -2559,6 +2559,8 @@ void mono_arch_set_native_call_context_ret      (CallContext *ccontext, gpointer
 gpointer mono_arch_get_native_call_context_args     (CallContext *ccontext, gpointer frame, MonoMethodSignature *sig, gpointer call_info);
 // After the pinvoke call is done, this moves return value from the ccontext to the InterpFrame.
 void mono_arch_get_native_call_context_ret      (CallContext *ccontext, gpointer frame, MonoMethodSignature *sig, gpointer call_info);
+// After the pinvoke call is done, this return an error context value from the ccontext.
+gpointer mono_arch_get_swift_error 				(CallContext *ccontext, MonoMethodSignature *sig, gpointer call_info, int *arg_index);
```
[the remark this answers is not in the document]
kotlarmilos: Sounds good.

src/mono/mono/mini/interp/interp.c
kotlarmilos: Yes, we should also extend the tests.

src/mono/mono/metadata/class-internals.h
```diff
@@ -978,6 +978,25 @@ mono_class_try_get_##shortname##_class (void)	\
 	return klass;	\
 }
 
+
+#define GENERATE_TRY_GET_CLASS_REF_WITH_CACHE(shortname,name_space,name) \
```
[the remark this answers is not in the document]
kotlarmilos: Yes, changed it to `GENERATE_TRY_GET_CLASS_PTR_WITH_CACHE` since it is a pointer.

src/mono/mono/mini/mini-arm64.h
kotlarmilos: Updated to use gregs.

src/mono/mono/metadata/class-internals.h
kotlarmilos: Updated.

src/mono/System.Private.CoreLib/src/System/Reflection/RuntimeMethodInfo.Mono.cs
kotlarmilos: Yes, not needed. Thank you!

src/native/public/mono/metadata/details/metadata-types.h
kotlarmilos: Thank you. Your proposal looks good to me. I implemented the `ext_callconv` as it offers extensibility for future flags.

src/tests/Interop/Swift/Dummy.swift
kotlarmilos: The `simpleFunc` sounds good as a placeholder for invalid scenarios.

src/tests/Interop/Swift/README.md
kotlarmilos: Yes, this is not needed anymore and will be removed.

src/tests/Interop/Swift/ErrorHandling.cs
kotlarmilos: You are right. I've added different test cases.

src/tests/Interop/Swift/ErrorHandling.cs
kotlarmilos: Thanks, updated.

src/tests/Interop/Swift/InvalidCallingConv.cs
kotlarmilos: Thanks, updated.

src/tests/Interop/Swift/InvalidCallingConv.cs
kotlarmilos: Added short message for each failure.

src/tests/Interop/Swift/SelfContext.cs
kotlarmilos: This test should evaluate the self context. If the self context is not present in the dedicated register, the invocation will result in a native error. I have modified the test case name to align with this scenario.

src/tests/Interop/Swift/README.md
kotlarmilos: Yes, this will be removed.

src/tests/Interop/Swift/README.md
kotlarmilos: Do you think this is needed at all? This seems to overlap with the design document and may be redundant.

AaronRobinsonMSFT: Nope. Removal and deference to the design doc make sense to me.

src/native/public/mono/metadata/details/metadata-types.h
```diff
@@ -31,7 +31,6 @@ typedef enum {
 	MONO_CALL_VARARG = 0x05,
 	/* unused, */
 	/* unused, */
```
AaronRobinsonMSFT, line 33:
```suggestion
	/* unused, */
	/* unused, */
```

AaronRobinsonMSFT: All changed to this file can just be reverted.

src/tests/Interop/Swift/SwiftErrorHandling/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+project(SwiftErrorHandling)
+include ("${CLR_INTEROP_TEST_ROOT}/Interop.cmake")
+
+set(SOURCE SwiftErrorHandling)
+
+add_custom_target(${SOURCE} ALL
+    COMMAND xcrun swiftc -emit-library ${CMAKE_CURRENT_SOURCE_DIR}/${SOURCE}.swift -o ${CMAKE_CURRENT_BINARY_DIR}/lib${SOURCE}.dylib
+    DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${SOURCE}.swift
+    COMMENT "Generating ${SOURCE} library"
+)
```
AaronRobinsonMSFT, line 10: Is this the appropriate way to do this? Seems like CMake has first class support for Swift - https://github.com/compnerd/swift-cmake-examples/tree/main. Is there a reason to create a custom target?

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
```diff
@@ -0,0 +1,83 @@
...
+        if (error.Value == IntPtr.Zero)
+        {
+            Assert.Fail("A Swift error was expected to be thrown.");
+        }
+
+        string errorMessage = GetErrorMessageFromSwift(error, expectedErrorMessage.Length);
+        if (errorMessage != expectedErrorMessage)
+        {
+            Assert.Fail("The error message retrieved from Swift does not match the expected message.");
+        }
```
AaronRobinsonMSFT, line 45: I would convert to using `Assert.True()`

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.swift
```diff
@@ -0,0 +1,50 @@
...
+public func getMyErrorMessage(from pointer: UnsafePointer<MyError>) -> UnsafePointer<UInt8>? {
+    let pointerValue = UInt(bitPattern: pointer)
+    var offsetValue: UInt
+#if arch(arm64)
+    offsetValue = 0x48
+#elseif arch(x86_64)
+    offsetValue = 0x20
+#else
+    fatalError("Unsupported architecture")
+#endif
```
AaronRobinsonMSFT, line 34: What are these constants? Please document.

src/tests/Interop/Swift/SwiftErrorHandling/CMakeLists.txt
matouskozak: CMake currently supports Swift with Ninja and Xcode generators. By default, we use `Unix Makefiles` generator (on Ubuntu and OSX), so we decided to follow the example from https://github.com/dotnet/runtime/blob/3b61ad9e3acb6302fd2fcef4947373475a0dcc29/src/native/libs/System.Security.Cryptography.Native.Apple/CMakeLists.txt#L54.

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
kotlarmilos: Good. I didn't know that `Assert.True` can be used with a message.

src/tests/Interop/Swift/README.md
kotlarmilos: Removed.

src/tests/Interop/Swift/README.md
kotlarmilos: Removed.

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.swift
kotlarmilos: Added a comment, please take a look. Additional inspection is needed to determine what is the content of the offset.

AaronRobinsonMSFT: This seems like a part of the Swift ABI, which I would assume would be documented somewhere if we are relying upon it, no? Is this API something the runtime would provide?

src/tests/Interop/Swift/SwiftInvalidCallConv/SwiftInvalidCallConv.cs
```diff
@@ -0,0 +1,82 @@
...
+    public static void TestFuncWithTwoSelfParameters()
+    {
+        SwiftSelf self = new SwiftSelf(IntPtr.Zero);
+
+        try
+        {
+            FuncWithTwoSelfParameters(self, self);
+            Assert.Fail("FuncWithTwoSelfParameters should have thrown InvalidProgramException. Invalid due to multiple SwiftSelf arguments.");
+        }
+        catch (InvalidProgramException e) { }
```
jkoritzinsky, line 40: Use `Assert.Throws` instead of a try-catch?
```suggestion
        Assert.Throws<InvalidProgramException>(() => FuncWithTwoSelfParameters(self, self));
```

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.swift
kotlarmilos: I don't believe this falls within the runtime scope. The error handling may vary depending on the binding tool implementation. One potential approach is to refactor the test to only verify its reference without inspecting the payload. I will review further the `swift_allocError` to figure out what is the exact layout of the payload.

@stephen-hawley Any thoughts on this?

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
```diff
@@ -0,0 +1,69 @@
...
+        const string expectedErrorMessage = "Catch me if you can!";
+        SetErrorMessageForSwift(expectedErrorMessage);
+
+        SwiftError error;
+
+        // This will not throw an error
+        int result = (int)conditionallyThrowError(false, &error);
+
+        Assert.True(error.Value != IntPtr.Zero, "No Swift error was expected to be thrown.");
+        Assert.True(result != 42, "The result from Swift does not match the expected value.");
```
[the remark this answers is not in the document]
kotlarmilos: I messed up with the condition, it should be fixed now.

src/tests/Interop/Swift/SwiftErrorHandling/CMakeLists.txt
AaronRobinsonMSFT: I see. I'd like to hear @jkoritzinsky's opinion on this approach.

jkoritzinsky: I think this is fine.

If we were to switch to using Ninja by default, I'd like to move to using the built-in support when using Ninja, but since that's not in any plans, I don't think it's worth the effort.

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.swift
jkoritzinsky: The signature for `swift_allocError` returns a `BoxPair`. Basically a pair of a pointer to the heap object for the error (`SwiftError*` in the runtime, `Error` in Swift) and a pointer to the opaque error object.

https://github.com/apple/swift/blob/98e65d015979c7b5a58a6ecf2d8598a6f7c85794/stdlib/public/runtime/ErrorObjectNative.cpp#L68

https://github.com/apple/swift/blob/98e65d015979c7b5a58a6ecf2d8598a6f7c85794/include/swift/Runtime/HeapObject.h#L95

From what I can tell error register contains the `SwiftError*` value. I believe this offset contains the size of the header, as the value of the error is tail-alllocated after the header:

https://github.com/apple/swift/blob/98e65d015979c7b5a58a6ecf2d8598a6f7c85794/stdlib/public/runtime/ErrorObject.h#L49

I think if we change this method to take an `Error` directly, we will be able to avoid using these offsets.

kotlarmilos: Thank you for reviewing this. You are right, if the method accepts an Error directly, there is no need for these offsets.

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
```diff
@@ -0,0 +1,69 @@
...
+    
+    private static void SetErrorMessageForSwift(string message)
+    {
+        var messageBytes = Encoding.UTF8.GetBytes(message);
+        SetErrorMessage(messageBytes, messageBytes.Length);
+    }
+
+    private static string GetErrorMessageFromSwift(SwiftError error, int length)
+    {
+        IntPtr pointer = GetErrorMessage(error.Value);
```
AaronRobinsonMSFT, line 64: Who is responsible for freeing the message memory?

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.swift
```diff
@@ -0,0 +1,36 @@
...
+        return 42
+    }
+}
+
+public func getMyErrorMessage(from error: Error) -> UnsafePointer<UInt8>? {
+    if let myError = error as? MyError {
+        switch myError {
+        case .runtimeError(let message):
+            let messageBytes: [UInt8] = Array(message.utf8)
+            let buffer = UnsafeMutableBufferPointer<UInt8>.allocate(capacity: messageBytes.count)
```
AaronRobinsonMSFT, line 30: Where is this memory freed?

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
```diff
@@ -64,6 +64,7 @@ private static string GetErrorMessageFromSwift(SwiftError error, int length)
         IntPtr pointer = GetErrorMessage(error.Value);
         byte[] byteArray = new byte[length];
         Marshal.Copy(pointer, byteArray, 0, length);
+        Marshal.FreeCoTaskMem(pointer);
```
AaronRobinsonMSFT, line 67: This is surprising to me. I would expect `NativeLibrary.Free()` instead. I don't think we want to be reusing `Marshal.FreeCoTaskMem()` in Swift interop scenarios. It might also be worth adding to our design documentation the memory manager we are expecting to use. Is it simply C's `malloc` / `free`? Do we need a new API for Swift runtime memory rules?

kotlarmilos: Oh, I've just realized that the `FreeCoTaskMem` is for COM interop.

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
```diff
@@ -0,0 +1,70 @@
...
+        var messageBytes = Encoding.UTF8.GetBytes(message);
+        SetErrorMessage(messageBytes, messageBytes.Length);
+    }
+
+    private static string GetErrorMessageFromSwift(SwiftError error, int length)
+    {
+        IntPtr pointer = GetErrorMessage(error.Value);
+        byte[] byteArray = new byte[length];
+        Marshal.Copy(pointer, byteArray, 0, length);
+        NativeLibrary.Free(pointer);
```
jkotas, line 67:
```suggestion
        NativeMemory.Free(pointer);
```
I think Aaron meant `NativeMemory.Free`. `NativeLibrary.Free` is for unloading libraries.

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
```diff
@@ -0,0 +1,70 @@
...
+        const string expectedErrorMessage = "Catch me if you can!";
+        SetErrorMessageForSwift(expectedErrorMessage);
+
+        SwiftError error;
+
+        // This will throw an error
+        conditionallyThrowError(true, &error);
+        Assert.True(error.Value != IntPtr.Zero, "A Swift error was expected to be thrown.");
+
+        string errorMessage = GetErrorMessageFromSwift(error, expectedErrorMessage.Length);
```
jkotas, line 37: Should `GetErrorMessageFromSwift` be responsible for computing the error message length? The length of the actual message does not have to be `expectedErrorMessage.Length` when things go wrong.

If the error message is zero-terminated C string, you can use `MemoryMarshal.CreateReadOnlySpanFromNullTerminated` or ever just `Marshal.PtrToStringUTF8`.

kotlarmilos: You are right. The `Marshal.PtrToStringUTF8` allocates a managed string and copies zero-terminated C string into it.

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
kotlarmilos: Please take a look at https://github.com/dotnet/designs/pull/307.

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
```diff
@@ -0,0 +1,69 @@
...
+        int result = (int)conditionallyThrowError(false, &error);
+
+        Assert.True(error.Value == IntPtr.Zero, "No Swift error was expected to be thrown.");
+        Assert.True(result == 42, "The result from Swift does not match the expected value.");
+    }
+    
+    private static void SetErrorMessageForSwift(string message)
+    {
+        var messageBytes = Encoding.UTF8.GetBytes(message + char.MinValue); // Add null terminator
+        SetErrorMessage(messageBytes, messageBytes.Length);
```
jkotas, line 59: `messageBytes.Length` is going to include null terminator. Is that expected? It is very unusual for the zero terminator to be included in the string length.

jkotas, line 59: It would be more efficient and simpler to let the P/Invoke marshalling take care of the conversion to zero-terminated string and then construct the string using the constructor that takes UTF8 zero-terminated string on the Swift side: https://developer.apple.com/documentation/swift/string/init(validatingutf8:)-208fn

kotlarmilos: There was a failure on `osx-x64` platform: `Expected: Catch me if you can!, Actual: Catch me if you can!�`

We believe that the null terminator is missing. We will try out your suggestion, to let the P/Invoke marshalling take care of the conversion to zero-terminated string and then construct the string using the Swift constructor.

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
```diff
@@ -0,0 +1,69 @@
...
+        int result = (int)conditionallyThrowError(false, &error);
+
+        Assert.True(error.Value == IntPtr.Zero, "No Swift error was expected to be thrown.");
+        Assert.True(result == 42, "The result from Swift does not match the expected value.");
+    }
+    
+    private static void SetErrorMessageForSwift(string message)
+    {
+        char[] buffer = message.ToCharArray();
+        SetErrorMessage(buffer, buffer.Length);
```
jkotas, line 59:
```suggestion
        SetErrorMessage(message, message.Length);
```
Unnecessary copy

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Runtime.InteropServices.Swift;
+using System.Text;
+using Xunit;
+
+public class ErrorHandlingTests
+{
+    private const string SwiftLib = "libSwiftErrorHandling.dylib";
+
+    [DllImport(SwiftLib, EntryPoint = "$s18SwiftErrorHandling05setMyB7Message7message6lengthySPys6UInt16VG_SitF")]
+    public static extern void SetErrorMessage(char[] message, int length);
```
jkotas, line 16:
```suggestion
    [DllImport(SwiftLib, EntryPoint = "$s18SwiftErrorHandling05setMyB7Message7message6lengthySPys6UInt16VG_SitF", CharSet = CharSet.Unicode)]
    public static extern void SetErrorMessage(string message, int length);
```

matouskozak: Thank you, that is much cleaner solution.

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
matouskozak: Is there a P/Invoke marshalling mechanism that could take care of the conversion to zero-terminated string when receiving data from Interop call? I ran into a problem that string data from Swift wasn't zero-terminated and `PtrToStringUni` expected zero-terminated string. I solved it by adding zero-terminator on the Swift side (https://github.com/dotnet/runtime/pull/94764/commits/97a74cc2843028a7679ec8e67193b8f61b32c7ba) but there might be a cleaner solution.

jkotas: There is no magic to help in this case. You would have to return the length of the string in an extra `out` length parameter. 

I think it is fine to return the zero terminated like you have done it.

jkotas: On the other hand, it would be nice for both get... and set... methods to be symmetric. set... passes the length in explicitly, so it would be nice for get... to return it explicitly as well.

matouskozak: That sounds good. Do you know if `out` parameters work with interop or is there another way of returning multiple values? If not, it would probably be "cleaner" to set the message without passing the length and do some additional work on swift side to read the string data properly.

jkotas: Yes, `out int length` parameter will work with interop. I 

>  there another way of returning multiple values

Another way to pass in or return multiple values is via a struct that has the pointer and the length as fields. 

I would go with `out int length` parameter in this case.

matouskozak: Thank you, the `out` parameter works as intended with interop (https://github.com/dotnet/runtime/pull/94764/commits/7103cfc19b5ef32120ddb74b9c03d68ad49da472).

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
```diff
@@ -0,0 +1,69 @@
...
+    
+    private static void SetErrorMessageForSwift(string message)
+    {
+        SetErrorMessage(message, message.Length);
+    }
+
+    private unsafe static string GetErrorMessageFromSwift(SwiftError error)
+    {
+        int messageLength;
+        IntPtr pointer = GetErrorMessage(error.Value, out messageLength);
```
jkotas, line 64:
```suggestion
        IntPtr pointer = GetErrorMessage(error.Value, out int messageLength);
```
Style nit

src/tests/Interop/Swift/SwiftErrorHandling/SwiftErrorHandling.cs
```diff
@@ -0,0 +1,69 @@
...
+        string errorMessage = Marshal.PtrToStringUni(pointer, messageLength);
```
jkotas, line 65: Note that `PtrToStringUni(IntPtr ptr, int len)` is going to throw ArgumentNullException when pointer is null. 

I think this behavior is fine for the test here. It may not be the right behavior if this code is copied to shipping library.

src/mono/mono/mini/aot-compiler.c
```diff
@@ -3764,6 +3764,7 @@ encode_signature (MonoAotCompile *acfg, MonoMethodSignature *sig, guint8 *buf, g
 	if (sig->pinvoke)
 		flags |= 0x80;
 	flags |= (sig->call_convention & 0x0F);
+	// TODO: Encode ext_callconv
```
vargaz, line 3768: If this is not implemented, it should assert that its not set.

src/mono/mono/mini/mini-amd64.c
```diff
@@ -53,6 +53,9 @@
 MONO_DISABLE_WARNING(4127) /* conditional expression is constant */
 
 static GENERATE_TRY_GET_CLASS_WITH_CACHE (math, "System", "Math")
+static GENERATE_TRY_GET_CLASS_WITH_CACHE (swift_error, "System.Runtime.InteropServices.Swift", "SwiftError")
```
vargaz, line 56: Might be a good idea to move these to mini.c so its shared with mini-arm64.c.

src/mono/mono/mini/tramp-amd64.c
```diff
@@ -1097,6 +1104,12 @@ mono_arch_get_interp_to_native_trampoline (MonoTrampInfo **info)
 	for (i = 0; i < FLOAT_PARAM_REGS; ++i)
 		amd64_sse_movsd_reg_membase (code, i, AMD64_R11, MONO_STRUCT_OFFSET (CallContext, fregs) + i * sizeof (double));
 
+	/* store the values of context registers onto the stack and set the context registers from CallContext */
+	for (i = 0; i < CTX_REGS; i++) {
```
vargaz, line 1108: Should this be done on all platforms, or only on apple platforms ?

src/mono/mono/mini/method-to-ir.c
```diff
@@ -11170,6 +11171,13 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 			const MonoJitICallId jit_icall_id = (MonoJitICallId)token;
 			MonoJitICallInfo * const jit_icall_info = mono_find_jit_icall_info (jit_icall_id);
 
+			if (mono_method_signature_has_ext_callconv (method->signature, MONO_EXT_CALLCONV_SWIFTCALL)) {
+#ifndef MONO_ARCH_HAVE_SWIFTCALL
```
vargaz, line 11175: The ifndef could be before the if.

src/mono/mono/mini/tramp-amd64.c
kotlarmilos: Only on Apple platforms targeting amd64 and arm64.

src/mono/mono/mini/mini-amd64.c
kotlarmilos: If moved to the `mini.c`, would it be still defined and usable in `mini-<arch>.c`? I assume it would require a declaration.

matouskozak: Could we move it to `mini.h` instead of `mini.c`? Then it should get included for both mini and interpreter, right?

src/mono/mono/mini/aot-compiler.c
```diff
@@ -3764,9 +3764,11 @@ encode_signature (MonoAotCompile *acfg, MonoMethodSignature *sig, guint8 *buf, g
 	if (sig->pinvoke)
 		flags |= 0x80;
 	flags |= (sig->call_convention & 0x0F);
-	// TODO: Encode ext_callconv
-
 	*p = flags;
+
+	++p;
+	*p = sig->ext_callconv & 0x0F;
```
vargaz, line 3770: Since this is rarely used, it would be better to only emit it if needed, like by setting a bit in 'flags' if its set.

kotlarmilos: The `flags` appear to be fully utilized, but we can add a new variable. In terms of performance, would it be faster to always encode it rather than inserting a single bit and then having a logic for conditional decoding within the runtime?

vargaz: You can make the flags into a guint32, and encode and decode them using encode_value () / decode_value ().
Switch explicit_this to be encoded as 0x80 and pinvoke as 0x40, since pinvoke is more common, then encode having the ext callconv as 0x100. encode_value () will still use 1 byte for values <= 127.

kotlarmilos: Thanks, added.

src/mono/mono/mini/mini-amd64.c
matouskozak: Moved to `marshal.h` so that it can be accessed byt both `marshal.c` and `mini-<arch>.c`.

src/mono/mono/mini/mini-arm64.c
```diff
@@ -1902,6 +1917,10 @@ arg_get_storage (CallContext *ccontext, ArgInfo *ainfo)
 		case ArgVtypeInIRegs:
 		case ArgInIReg:
 			return &ccontext->gregs [ainfo->reg];
+		case ArgSwiftSelf:
+			return &ccontext->gregs [PARAM_REGS + 1 + ARMREG_R20 - CTX_REGS_OFFSET];
```
BrzVlad, line 1921: I don't see the point of `ARMREG_R20 - CTX_REGS_OFFSET` subtraction, we are hardcoding R20 here. Seems confusing. Why not just store at `PARAM_REGS + 1` and at `PARAM_REGS + 1 + 1` below.

src/mono/mono/mini/mini-arm64.h
```diff
@@ -268,8 +276,8 @@ struct CallInfo {
 };
 
 typedef struct {
-	/* General registers + ARMREG_R8 for indirect returns */
-	host_mgreg_t gregs [PARAM_REGS + 1];
+	/* General registers + context registers + ARMREG_R8 for indirect returns */
```
BrzVlad, line 279: This comment is slightly misleading. Maybe it is better: 
`/* General registers + ARMREG_R8 for indirect returns + context registers  */`

src/mono/mono/mini/interp/interp.c
```diff
@@ -1761,6 +1761,18 @@ ves_pinvoke_method (
 	interp_pop_lmf (&ext);
 
 #ifdef MONO_ARCH_HAVE_INTERP_PINVOKE_TRAMP
+#ifdef MONO_ARCH_HAVE_SWIFTCALL
+	if (mono_method_signature_has_ext_callconv (sig, MONO_EXT_CALLCONV_SWIFTCALL)) {
+		int arg_index;
+		gpointer data = mono_arch_get_swift_error (&ccontext, sig, call_info, &arg_index);
+
+		// Perform an indirect store at arg_index stack location
+		if (data != NULL) {
```
BrzVlad, line 1770: Why would `data` ever be null here ? Shouldn't we assert instead ?

src/mono/mono/mini/mini-amd64.c
```diff
@@ -1077,6 +1076,21 @@ get_call_info (MonoMemPool *mp, MonoMethodSignature *sig)
 		default:
 			g_assert_not_reached ();
 		}
+
+#ifdef MONO_ARCH_HAVE_SWIFTCALL
+		if (mono_method_signature_has_ext_callconv (sig, MONO_EXT_CALLCONV_SWIFTCALL)) {
+			MonoClass *swift_error_ptr = mono_class_try_get_swift_error_ptr_class ();
+			MonoClass *swift_self = mono_class_try_get_swift_self_class ();
+			MonoClass *klass = mono_class_from_mono_type_internal (sig->params [i]);
+			if (klass) {
```
BrzVlad, line 1085: Aren't we overwritting the previously assigned arg storage ? So, this param would first be handled as a `MONO_TYPE_I` with `add_general`, potentially reserving stack space to pass it if all the param registers are consumed ? I'm not sure if this is correct.

src/mono/mono/mini/mini-arm64.c
kotlarmilos: The idea was to indicate the self and error registers on arm64. The use of `PARAM_REGS + 1` and `PARAM_REGS + 1 + 1` is not descriptive in terms of which registers are being used. I agree that your suggestion is simpler and not confusing.

src/mono/mono/mini/mini-amd64.c
kotlarmilos: I think you are correct. Currently, we treat SwiftSelf as ArgVtypeInIRegs and SwiftError as ArgInIReg. There might be cases where they should be passed via the stack. I see two possible options:
 - Add a flag to ArgInfo
 - Always pass them as stack parameters

BrzVlad: I thought the swift cconv states that they are always passed in regs. ?

kotlarmilos: After offline sync, we determined that the new argument types in ArgInfo lack information on arg storage (param regs or stack). The code is updated to directly use context registers during the call_emit function.

src/mono/mono/mini/interp/interp.c
kotlarmilos: If SwiftError arg is not passed, the data will be NULL. The code is updated to first check if the arg is found and then asserts.

src/mono/mono/mini/mini-arm64.c
kotlarmilos: Updated.

src/mono/mono/mini/mini-amd64.c
```diff
@@ -2259,6 +2305,28 @@ mono_arch_emit_call (MonoCompile *cfg, MonoCallInst *call)
...
+			MonoClass *swift_self = mono_class_try_get_swift_self_class ();
+			MonoClass *swift_error = mono_class_try_get_swift_error_class ();
+			MonoClass *klass = mono_class_from_mono_type_internal (sig->params [i]);
+			if (klass) {
+				if (klass == swift_self) {
+					guint32 size = mini_type_stack_size_full (m_class_get_byval_arg (klass), NULL, sig->pinvoke && !sig->marshalling_disabled);
+					g_assert (size == sizeof (target_mgreg_t));
+					ainfo->storage = ArgValuetypeInReg;
+					ainfo->pair_storage [0] = ArgInIReg;
+					ainfo->pair_regs [0] = GINT32_TO_UINT8 (AMD64_R13);
```
BrzVlad, line 2319: Why are these done here. I expect the ainfo to be correctly initialized in `get_call_info` ?

kotlarmilos: Yes, it can be moved to `get_call_info`. The similar support should be added for `UnmanagedCallersOnly` which needs to be implemented on the callee side, in the method prolog and epilog. I would rather revert this change and implement this in the same way. We can add two storage variants for Swift types, like ArgSwiftSelf and ArgSwiftSelfOnStack. What do you think?

BrzVlad: I dont see what `ArgSwiftSelf` is meant to symbolize, or the benefits of its special definition. `ArgInfo` contains information about how to pass an argument, in regs / stack. Swift args are passed also in regs or stack. `get_call_info` should detect if an argument is swift related, and resolve its allocation to `ArgInIReg` or something else, according to what the swift cconv requires. Then the calling code should just work naturally, without explicit intervention to handle swift cconv. (unless it has some problems with allocating args to those special regs)

src/mono/mono/mini/mini-amd64.c
```diff
@@ -1971,6 +2000,11 @@ mono_arch_allocate_vars (MonoCompile *cfg)
 
 				break;
 			}
+			case ArgSwiftError: {
+					ins->flags &= ~MONO_INST_IS_DEAD;
```
vargaz, line 2004: Why is this needed ?

kotlarmilos: The SwiftError is marked as a dead arg and could be skipped in the method prolog:
```c
if (ins->flags & MONO_INST_IS_DEAD && !MONO_CFG_PROFILE (cfg, ENTER_CONTEXT))
	/* Unused arguments */
	continue;
```
The `cfg->arch.swift_error_var->flags |= MONO_INST_IS_DEAD;` is used in the method epilog. The runtime should store R12 into the SwiftError only if the arg is passed. On x64, the `swift_error_var` is always allocated, so this iteration sets MONO_INST_IS_DEAD flag for swift_error_var which is then checked in the method epilog.

src/mono/mono/mini/mini-amd64.c
kotlarmilos: SwiftSelf is considered a standard argument for P/Invoke, where the register allocator always utilizes the dedicated context register. SwiftError is introduced as a new arg storage type ArgSwiftError and passed to the managed-to-native wrapper, but it is not passed to the native function. Here is an example for arm64:
```
managed_to_native_wrapper(swiftself, swifterror)
str &swifterror, [sp, #swift_error_var] // Store the address of swifterror on the stack
ldr R20, swiftself // Swiftself is always passed through the dedicated register
ldr R21, 0x0 // Set SwiftError register to 0
call native function
ldr IP0, [sp, #swift_error_var] // Load &swifterror
str R21, IP0 // Store R21 into swifterror
```

src/mono/mono/mini/mini-arm64.c
```diff
@@ -6240,6 +6329,16 @@ mono_arch_emit_epilog (MonoCompile *cfg)
 		break;
 	}
 
+#ifdef MONO_ARCH_HAVE_SWIFTCALL
+	if (mono_method_signature_has_ext_callconv (cfg->method->signature, MONO_EXT_CALLCONV_SWIFTCALL)) {
+		if (cfg->arch.swift_error_var) {
```
vargaz, line 6334: The ifdef and the if is not needed, its enough to check cfg->arch.swift_error_var ?
Also, shouldn't this be done after the native call to the swift function instead of in the epilog ?

src/mono/mono/mini/method-to-ir.c
```diff
@@ -7656,6 +7656,7 @@ mono_method_to_ir (MonoCompile *cfg, MonoMethod *method, MonoBasicBlock *start_b
 					goto calli_end;
 				}
 			}
+
```
vargaz, line 7659: Whitespace change.

src/mono/mono/mini/mini-amd64.c
```diff
@@ -1008,6 +1007,34 @@ get_call_info (MonoMemPool *mp, MonoMethodSignature *sig)
 		}
 
 		ptype = mini_get_underlying_type (sig->params [i]);
+
+#ifdef MONO_ARCH_HAVE_SWIFTCALL
+		if (mono_method_signature_has_ext_callconv (sig, MONO_EXT_CALLCONV_SWIFTCALL)) {
+			MonoClass *swift_self = mono_class_try_get_swift_self_class ();
+			MonoClass *swift_error = mono_class_try_get_swift_error_class ();
+			MonoClass *klass = mono_class_from_mono_type_internal (sig->params [i]);
+			if (klass) {
```
vargaz, line 1016: Not sure if (klass) is needed, it shouldn't be null.

src/mono/mono/mini/mini-amd64.c
```diff
@@ -1249,6 +1278,22 @@ mono_arch_set_native_call_context_args (CallContext *ccontext, gpointer frame, M
 		else
 			storage = arg_get_storage (ccontext, ainfo);
 
+#ifdef MONO_ARCH_HAVE_SWIFTCALL
+		if (mono_method_signature_has_ext_callconv (sig, MONO_EXT_CALLCONV_SWIFTCALL)) {
+			MonoClass *swift_self = mono_class_try_get_swift_self_class ();
+			MonoClass *swift_error = mono_class_try_get_swift_error_class ();
+			MonoClass *klass = mono_class_from_mono_type_internal (sig->params [i]);
+			if (klass) {
+				if (klass == swift_self) {
```
vargaz, line 1287: Ditto.

src/mono/mono/mini/mini-amd64.c
vargaz: The variable is marked as VOLATILE when its created, so it shouldn't end up dead.

src/mono/mono/mini/mini-amd64.c
```diff
@@ -8304,6 +8404,16 @@ mono_arch_emit_epilog (MonoCompile *cfg)
 	/* Mark the start of the epilog */
 	mono_emit_unwind_op_mark_loc (cfg, code, 0);
 
+#ifdef MONO_ARCH_HAVE_SWIFTCALL
+	if (mono_method_signature_has_ext_callconv (cfg->method->signature, MONO_EXT_CALLCONV_SWIFTCALL)) {
+		if (cfg->arch.swift_error_var && (cfg->arch.swift_error_var->flags & MONO_INST_IS_DEAD)) {
```
vargaz, line 8409: The ifdef and the if is not needed.

src/mono/mono/mini/mini-arm64.c
```diff
@@ -1857,6 +1857,29 @@ get_call_info (MonoMemPool *mp, MonoMethodSignature *sig)
 			add_param (cinfo, &cinfo->sig_cookie, mono_get_int_type (), FALSE);
 		}
 
+#ifdef MONO_ARCH_HAVE_SWIFTCALL
+		if (mono_method_signature_has_ext_callconv (sig, MONO_EXT_CALLCONV_SWIFTCALL)) {
+			MonoClass *swift_self = mono_class_try_get_swift_self_class ();
+			MonoClass *swift_error = mono_class_try_get_swift_error_class ();
+			MonoClass *klass = mono_class_from_mono_type_internal (sig->params [pindex]);
+			if (klass) {
+				if (klass == swift_self && sig->pinvoke) {
```
vargaz, line 1866: Ditto.

src/mono/mono/mini/mini-amd64.c
kotlarmilos: The `cfg->arch.swift_error_var` is marked as volatile, but (arg) ins is not.

The `cfg->arch.swift_error_var` utilizes the MONO_INST_IS_DEAD to indicate that the arg is present in the signature. Is it a good idea to introduce a new flag for that instead?

src/mono/mono/mini/mini-arm64.c
kotlarmilos: It is done in the wrapper epilog which is after the native call. Is there a better place to do it?

src/mono/mono/mini/mini-amd64.c
kotlarmilos: Right, not needed.

src/mono/mono/mini/mini-arm64.c
```diff
@@ -2810,6 +2873,21 @@ mono_arch_allocate_vars (MonoCompile *cfg)
 			ins->inst_left = vtaddr;
 			break;
 		}
+		case ArgSwiftError: {
+			ins = mono_compile_create_var (cfg, mono_get_int_type (), OP_LOCAL);
+			ins->flags |= MONO_INST_VOLATILE;
```
vargaz, line 2874: ins is already set above to cfg->args [i], there is no need to create a new one.

kotlarmilos: You are right, both x64 and arm64 can use the existing ins.

src/mono/mono/mini/mini-amd64.c
```diff
@@ -8404,15 +8400,11 @@ mono_arch_emit_epilog (MonoCompile *cfg)
...
 
-#ifdef MONO_ARCH_HAVE_SWIFTCALL
-	if (mono_method_signature_has_ext_callconv (cfg->method->signature, MONO_EXT_CALLCONV_SWIFTCALL)) {
-		if (cfg->arch.swift_error_var && (cfg->arch.swift_error_var->flags & MONO_INST_IS_DEAD)) {
-			MonoInst *ins = cfg->arch.swift_error_var;
-			amd64_mov_reg_membase (code, AMD64_R11, ins->inst_basereg, ins->inst_offset, sizeof (target_mgreg_t));
-			amd64_mov_membase_reg (code, AMD64_R11, 0, AMD64_R12, sizeof (target_mgreg_t));
-		}
+	if (cfg->arch.swift_error_var && (cfg->arch.swift_error_var->flags & MONO_INST_IS_DEAD)) {
+		MonoInst *ins = cfg->arch.swift_error_var;
```
vargaz, line 8404: Don't think the IS_DEAD part is needed, also if its needed, shouldn't it be !IS_DEAD ?

kotlarmilos: Not needed, updated to use the existing arg ins.

src/mono/mono/mini/mini-amd64.c
kotlarmilos: The ArgSwiftError ins is marked as dead before the `mono_arch_allocate_vars`.

kotlarmilos: I would avoid iterating over args in `mono_arch_create_vars`. Do you think it is possible?

vargaz: You can store the argument index of the swifterror argument in CallInfo, and the use that in create_vars () to index into cfg->args.

src/mono/mono/mini/mini-arm64.c
kotlarmilos: Thanks, moved to `emit_move_return_value`.

vargaz:
The mono JIT changes look ok to me.

src/mono/mono/mini/mini-arm64.c
```diff
@@ -1857,6 +1857,28 @@ get_call_info (MonoMemPool *mp, MonoMethodSignature *sig)
...
+			MonoClass *swift_self = mono_class_try_get_swift_self_class ();
+			MonoClass *swift_error = mono_class_try_get_swift_error_class ();
+			MonoClass *swift_error_ptr = mono_class_create_ptr (m_class_get_this_arg (swift_error));
+			MonoClass *klass = mono_class_from_mono_type_internal (sig->params [pindex]);
+			if (klass == swift_self && sig->pinvoke) {
+				cinfo->gr--;
+				add_param (cinfo, ainfo, sig->params [pindex], FALSE);
+				ainfo->reg = ARMREG_R20;
+				continue;
+			} else if (klass == swift_error || klass == swift_error_ptr) {
```
lambdageek, line 1871: same as x64 - why do we look for both `SwiftError` and `SwiftError*`

src/mono/mono/mini/mini.c
```diff
@@ -92,6 +92,10 @@ static gint64 discarded_jit_time;
 #define mono_jit_unlock() mono_os_mutex_unlock (&jit_mutex)
 static mono_mutex_t jit_mutex;
 
+// used for Swift interop
+GENERATE_TRY_GET_CLASS_WITH_CACHE (swift_error, "System.Runtime.InteropServices.Swift", "SwiftError")
+GENERATE_TRY_GET_CLASS_WITH_CACHE (swift_self, "System.Runtime.InteropServices.Swift", "SwiftSelf")
+
```
lambdageek, line 98: you declare these in `metadata/marshal.h`, and define them in `mini/mini.c` - I think the definitions should be in `metadata/`, too.

src/mono/mono/metadata/marshal.c
```diff
@@ -3555,133 +3574,130 @@ mono_marshal_get_native_wrapper (MonoMethod *method, gboolean check_exceptions,
...
-	csig = mono_metadata_signature_dup_full (get_method_image (method), sig);
-	mono_marshal_set_callconv_from_modopt (method, csig, FALSE);
+		if (csig->call_convention == MONO_CALL_DEFAULT) {
+			/* If the calling convention has not been set, check the UnmanagedCallConv attribute */
+			mono_marshal_set_callconv_from_unmanaged_callconv_attribute (method, csig, &skip_gc_trans);
+		}
 
-	mspecs = g_new0 (MonoMarshalSpec*, sig->param_count + 1);
-	mono_method_get_marshal_info (method, mspecs);
+		if (mono_method_signature_has_ext_callconv (csig, MONO_EXT_CALLCONV_SWIFTCALL)) {
```
lambdageek, line 3625: @kotlarmilos @vargaz Do we need to prevent these m2n swiftcall wrappers from being inlined?  I think there's logic in the mini-arm64 and mini-amd64 backends that depends on the `pinvoke` bit of the signature.

lambdageek (approved):
lgtm.  some minor cleanup nits.  and two questions:

1. I'm not sure why the backend looks for both `SwiftError` and `SwiftError*` arguments.
2. Should we prevent these pinvoke wrappers from being inlined?

src/mono/mono/metadata/marshal.c
```diff
@@ -3427,23 +3431,31 @@ mono_marshal_get_native_wrapper (MonoMethod *method, gboolean check_exceptions,
...
+	ERROR_DECL (emitted_error);
 
 	g_assert (method != NULL);
 	g_assertf (mono_method_signature_internal (method)->pinvoke, "%s flags:%X iflags:%X param_count:%X",
 		method->name, method->flags, method->iflags, mono_method_signature_internal (method)->param_count);
 
-	GHashTable **cache_ptr;
-
-	MonoType *string_type = m_class_get_byval_arg (mono_defaults.string_class);
+	if (MONO_CLASS_IS_IMPORT (method->klass)) {
```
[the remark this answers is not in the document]
kotlarmilos: Yes, I remove redundant code since it was challenging to add param checks.

src/mono/mono/metadata/marshal.c
```diff
@@ -3555,133 +3574,130 @@ mono_marshal_get_native_wrapper (MonoMethod *method, gboolean check_exceptions,
...
+					if (param_klass == swift_error && method->signature->params [i]->byref__ == 0) {
+						swift_error_args = swift_self_args = 0;
+						mono_error_set_generic_error (emitted_error, "System", "InvalidProgramException", "SwiftError argument must be passed by reference.");
+						break;
+					} else if (param_klass == swift_error || param_klass == swift_error_ptr) {
+						swift_error_args++;
+					} else if (param_klass == swift_self) {
+						swift_self_args++;
+					} else if (!m_class_is_enumtype (param_klass) && !m_class_is_primitive (param_klass) && m_class_get_this_arg (param_klass)->type != MONO_TYPE_FNPTR) {
+						swift_error_args = swift_self_args = 0;
```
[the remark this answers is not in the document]
kotlarmilos: Blittable struct types should be allowed.

src/mono/mono/mini/mini-amd64.c
```diff
@@ -1360,6 +1406,28 @@ mono_arch_get_native_call_context_ret (CallContext *ccontext, gpointer frame, Mo
...
+ * If found, it retrieves the value from `ccontext`.
+ */
+gpointer
+mono_arch_get_swift_error (CallContext *ccontext, MonoMethodSignature *sig, int *arg_index)
+{
+	MonoClass *swift_error = mono_class_try_get_swift_error_class ();
+	MonoClass *swift_error_ptr = mono_class_create_ptr (m_class_get_this_arg (swift_error));
+	for (guint i = 0; i < sig->param_count + sig->hasthis; i++) {
+		MonoClass *klass = mono_class_from_mono_type_internal (sig->params [i]);
+		if (klass && (klass == swift_error || klass == swift_error_ptr)) {
```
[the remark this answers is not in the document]
kotlarmilos: The idea is to support passing by reference using the `ref` keyword. However, it is not supported for function pointers, so it is necessary to support `SwiftError*` as well.

src/mono/mono/mini/mini.c
kotlarmilos: I suggest defining them in `marshal.c` then, and they should be accessible from `mini-<arch>.c`.

src/mono/mono/metadata/marshal.c
kotlarmilos: Is the inlining performed before or after the lowering process? Do the method prolog and epilog get inlined as well?

ArgSwiftError is handled differently in the method prolog, which could be affected when it is inlined. Also, there is a potential problem related to context registers clobbering when wrappers are inlined.

src/mono/mono/mini/mini-amd64.c
```diff
@@ -1008,6 +1009,34 @@ get_call_info (MonoMemPool *mp, MonoMethodSignature *sig)
...
+		if (mono_method_signature_has_ext_callconv (sig, MONO_EXT_CALLCONV_SWIFTCALL)) {
+			MonoClass *swift_self = mono_class_try_get_swift_self_class ();
+			MonoClass *swift_error = mono_class_try_get_swift_error_class ();
+			MonoClass *swift_error_ptr = mono_class_create_ptr (m_class_get_this_arg (swift_error));
+			MonoClass *klass = mono_class_from_mono_type_internal (sig->params [i]);
+			if (klass == swift_self && sig->pinvoke) {
+				guint32 old_gr = gr;
+				if (gr >= PARAM_REGS)
+					gr--;
+				add_valuetype (sig, ainfo, ptype, FALSE, &gr, &fr, &stack_size);
```
BrzVlad, line 1023: This looks way more confusing that on arm64 case. We should probably just initialize ainfo explicitly as arg in ireg.

src/mono/mono/mini/mini-amd64.c
```diff
@@ -1249,6 +1280,21 @@ mono_arch_set_native_call_context_args (CallContext *ccontext, gpointer frame, M
 		else
 			storage = arg_get_storage (ccontext, ainfo);
 
+#ifdef MONO_ARCH_HAVE_SWIFTCALL
+		if (mono_method_signature_has_ext_callconv (sig, MONO_EXT_CALLCONV_SWIFTCALL)) {
```
BrzVlad, line 1284: Do we even need all this code ? Looks like for swift self, we should have something like arg in ireg with the context reg specified in `ainfo->reg`. So arg get storage should already handle it. For swift error similarly the storage is returned, why additional comparison is required ? I think all this code could just be: `if (ainfo == ArgSwiftError) *(gpointer*)storage = 0;`

src/mono/mono/mini/mini-arm64.c
```diff
@@ -2010,6 +2034,21 @@ mono_arch_set_native_call_context_args (CallContext *ccontext, gpointer frame, M
 		else
 			storage = arg_get_storage (ccontext, ainfo);
 
+#ifdef MONO_ARCH_HAVE_SWIFTCALL
+		if (mono_method_signature_has_ext_callconv (sig, MONO_EXT_CALLCONV_SWIFTCALL)) {
```
BrzVlad, line 2038: As in amd64 case, I think all this code can be removed. Here we would need to make `arg_get_storage` correctly return the storage for context regs, since the CallContext doesn't contain all the registers as on amd64.

src/mono/mono/mini/mini-arm64.c
```diff
@@ -1857,6 +1857,28 @@ get_call_info (MonoMemPool *mp, MonoMethodSignature *sig)
...
+			MonoClass *swift_self = mono_class_try_get_swift_self_class ();
+			MonoClass *swift_error = mono_class_try_get_swift_error_class ();
+			MonoClass *swift_error_ptr = mono_class_create_ptr (m_class_get_this_arg (swift_error));
+			MonoClass *klass = mono_class_from_mono_type_internal (sig->params [pindex]);
+			if (klass == swift_self && sig->pinvoke) {
+				cinfo->gr--;
+				add_param (cinfo, ainfo, sig->params [pindex], FALSE);
```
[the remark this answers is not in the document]
kotlarmilos: The idea was to reuse code and avoid maintaining changes in multiple places. However, as you suggested, it could be as simple as calculating the size of the payload.

src/mono/mono/mini/mini-amd64.c
kotlarmilos: Ok, same as in arm64 case.

src/mono/mono/metadata/marshal.c
vargaz: These wrappers are not inlined right now.

src/mono/mono/mini/mini-amd64.c
kotlarmilos: You are right. The SwiftSelf arg is already set properly and should be handled without additional logic. Thanks for the heads up!

BrzVlad (approved):
LGTM overall

kotlarmilos:
@dotnet/jit-contrib Please take a look at the runtime tests before the merge.

jakobbotsch (approved):

amanasifkhalid (approved):

amanasifkhalid:
runtime tests LGTM, too

PR #94765, opened 2023-11-15
UnzipEven() and Count16BitElements() intrinsics for Arm64 SVE

a74nh:
Many of the files touched here are autogenerated via InstructionSetDesc.txt.


a74nh:
Currently includes all of @kunalspathak's https://github.com/dotnet/runtime/pull/94529. I don't think it's possible to directly base a pull request on top of another pull request. All of my code starts from https://github.com/dotnet/runtime/pull/94765/commits/ceccafa8876488eb117825369e3f039aa46ab6dc

Not sure if the Arm64_Sve parts are required or if it should only be Sve. I think we want the readytorun/crossgen changes too (but it should be disabled for them?)

When running the included test, it fails with the following error:
```
Assert failure(PID 2048593 [0x001f4251], Thread: 2048593 [0x1f4251]): Assertion failed 'insOptsNone(id->idInsOpt())' in 'JIT.HardwareIntrinsics.Arm._AdvSimd.SveTest__SveTest:unzip(System.Numerics.Vector`1[int],System.Numerics.Vector`1[int]):System.Numerics.Vector`1[int]:this' during 'Generate code' (IL size 8; hash 0x0bc0f608; Tier0)

    File: /home/alahay01/dotnet/runtime_sve/src/coreclr/jit/emitarm64.cpp Line: 948
    Image: /home/alahay01/dotnet/runtime_sve/artifacts/tests/coreclr/linux.arm64.Checked/Tests/Core_Root/corerun
```
This is because `id->idInsOpt()` is set to `INS_OPTS_4S`. That is set in `genHWIntrinsic()`. I think we need to add a new flag `HW_Category_SIMDVariable` to do special handling.

This PR already touches quite a few files, so I was thinking any further codegen changes would go in a separate PR.


@tannergooding @BruceForstall @TamarChristinaArm

a74nh:
`Count16BitElements()` currently generates to `MRS` because we're missing all the codegen parts for the `CNTH` and `MRS` is a fairly safe instruction to use for now.

I had to do some changes to allow for API functions that don't use any vector registers. Happy to pull this part out of the PR into a new one.

kunalspathak:
> Many of the files touched here are autogenerated via InstructionSetDesc.txt.

IMO, lets have a separate PR for these changes and that way, this PR will just be adding the `UnzipEven()` and `Count16BitElements()`.

a74nh:
> > Many of the files touched here are autogenerated via InstructionSetDesc.txt.
> 
> IMO, lets have a separate PR for these changes and that way, this PR will just be adding the `UnzipEven()` and `Count16BitElements()`.

Done - https://github.com/dotnet/runtime/pull/94791

src/coreclr/jit/emitarm64.cpp
```diff
@@ -7987,6 +8035,14 @@ void emitter::emitIns_R_R_R(
             fmt = IF_DV_3A;
             break;
 
+        case INS_sve_uzp1:
+            // TODO: This could also be SVE_BR_3B or SVE_CI_3A
```
BruceForstall, line 8039: We should standardize on `TODO-SVE:` for marking all SVE-related "to-do" work items.

BruceForstall (commented):
Hopefully we can start getting some of the "base" PRs merged soon so we'll only be looking at incremental diffs...

kunalspathak:
> Hopefully we can start getting some of the "base" PRs merged soon so we'll only be looking at incremental diffs...

I agree. I think we should definitely get in #94791 (for InstructionSet related changes). I am under the impression that in order to merge other example PRs like this one or https://github.com/dotnet/runtime/pull/94529, we need to have register support, so we can validate the encoding. On the contrary, we could start merging the instructions in (probably commented code or some different way) and once register allocation changes are done and we have a tool to validate encoding, we can enable those instructions. Thoughts?

BruceForstall:
> Thoughts?

I wonder how many regalloc changes we really need to be able to make progress, assuming we don't need to generate predicate registers. Presumably we'd still use Vn registers, and they're encoded just like Zn registers. (Maybe the predicate register encoding could always assume predicate register zero, say, or use a Vn register temporarily instead of a new Pg register.)

kunalspathak:
> I wonder how many regalloc changes we really need to be able to make progress, assuming we don't need to generate predicate registers. Presumably we'd still use Vn registers, and they're encoded just like Zn registers.

Yes, for now, the easier change to do is add `Zn` register so the code gets disassembled properly and then reuse one of the vector register for `P0` predicate register.

src/coreclr/inc/corinfoinstructionset.h
```diff
@@ -764,6 +774,10 @@ inline const char *InstructionSetToString(CORINFO_InstructionSet instructionSet)
             return "VectorT128";
         case InstructionSet_Rcpc2 :
             return "Rcpc2";
+        case InstructionSet_Sve :
```
TamarChristinaArm, line 777: Curious, why is there two instruction sets here? For the other ones it was because of AArch32, but there's no AArch32 SVE.

a74nh: AIUI, we need both to matching the existing structure. All the functions will be in Sve and Sve.Arm64 will be empty.

src/tests/JIT/HardwareIntrinsics/Arm/Sve/Program.cs
```diff
@@ -0,0 +1,81 @@
...
+            if (!test.Succeeded)
+            {
+                throw new Exception("One or more scenarios did not complete as expected.");
+            }
+        }
+    }
+
+    public sealed unsafe class SveTest__SveTest
+    {
+        public bool IsSupported => AdvSimd.IsSupported;
```
MichalPetryka, line 46:
```suggestion
        public bool IsSupported => Sve.IsSupported;
```
?

a74nh:
This has been replaced by 
https://github.com/dotnet/runtime/pull/94845 and https://github.com/dotnet/runtime/pull/95021

PR #94735, opened 2023-11-14
[RISC-V] Add DynamicHelpers to riscv64 stubs

ashaurtaev:
Added implementation of dynamic helpers to stubs for riscv64

Part of https://github.com/dotnet/runtime/issues/84834
cc @wscho77 @HJLeee @clamp03 @JongHeonChoi @t-mustafin @gbalykov @yurai007 @sirntar @tomeksowi @Bajtazar @brucehoult

PR #94766, opened 2023-11-15
[LoongArch64] Fix the overwritten of reg21 when `pLookup->sizeOffset >2047`

shushanhf:
Fix the overwritten of reg21 when `pLookup->sizeOffset >2047` within the loongarch64/stubs.cpp.

@gbalykov 
Thanks very much for your advice!
https://github.com/dotnet/runtime/pull/94735#discussion_r1393771885

### DIFF
--- a/src/coreclr/vm/loongarch64/stubs.cpp
+++ b/src/coreclr/vm/loongarch64/stubs.cpp
@@ -1851,10 +1851,10 @@ PCODE DynamicHelpers::CreateDictionaryLookupHelper(LoaderAllocator * pAllocator,
 
                 if (pLookup->sizeOffset > 2047)
                 {
-                    // pcaddi  $r21,0
-                    *(DWORD*)p = 0x18000015; p += 4;
-                    // ld.d  $t4,$r21, #dataOffset
-                    *(DWORD*)p = 0x28c002b0 | (dataOffset << 10); p += 4;
+                    // pcaddi  $t4,0
+                    *(DWORD*)p = 0x18000010; p += 4;
+                    // ld.d  $t4,$t4, #dataOffset
+                    *(DWORD*)p = 0x28c00210 | (dataOffset << 10); p += 4;
                     // ldx.d  $t5,$a0,$t4
                     *(DWORD*)p = 0x380c4091; p += 4;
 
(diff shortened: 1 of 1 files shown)


src/coreclr/vm/loongarch64/stubs.cpp
```diff
@@ -1851,10 +1851,10 @@ PCODE DynamicHelpers::CreateDictionaryLookupHelper(LoaderAllocator * pAllocator,
 
                 if (pLookup->sizeOffset > 2047)
                 {
-                    // pcaddi  $r21,0
-                    *(DWORD*)p = 0x18000015; p += 4;
-                    // ld.d  $t4,$r21, #dataOffset
-                    *(DWORD*)p = 0x28c002b0 | (dataOffset << 10); p += 4;
+                    // pcaddi  $t4,0
```
gbalykov, line 1854: Not exactly what I meant. Arm64 does pc-relative load if `if (pLookup->sizeOffset > 32760)` 
https://github.com/dotnet/runtime/blob/18cac43e79503cb2d9424dd19822e7ee8444db26/src/coreclr/vm/arm64/stubs.cpp#L2009-L2010
and if `if(pLookup->offsets[i] > 32760)`
https://github.com/dotnet/runtime/blob/18cac43e79503cb2d9424dd19822e7ee8444db26/src/coreclr/vm/arm64/stubs.cpp#L2039-L2040. 

For loongarch `dataOffset` is also relative to current PC, it means that you need to load `pc` (i.e. add `pcaddi`) for both `pLookup->sizeOffset > 2047` and `pLookup->offsets[i] > 2047` cases). And then remove `pcaddi $r21,0` in the beginning of this function. Or, alternatively, make `dataOffset` relative to the beginning of whole code block, then it will be the opposite with just one `pcaddi` in the beginning. The first approach is implemented right now in PR for risc-v, and the second one is what we'll change it to.

shushanhf: Thanks.
I though that the LA64 had implemented the second way.
It's time long I didn't look these whole carefully yesterday.
I will do it now.

shushanhf: @gbalykov  Thanks very much for your reviewing.
I have rewrite the `dataOffset` for LoongArch64.

gbalykov (approved):

ashaurtaev (approved):

shushanhf:
Hi, @jkotas 
Could you please review this PR?
Thanks !

jkotas (approved):

PR #94768, opened 2023-11-15
Use Binary Search Tree for Frozen Collections length buckets strategy

andrewjsaid:
## Problem Statement

The LengthBuckets strategy currently works by grouping strings by their length into buckets of maximum size 5.
During lookup we examine each item in the bucket sequentially to find the original string.

This approach means that the `LengthBucketsFrozenDictionary` performance can be dramatically different depending on
the position of the key within the bucket as the 5th element is compared last.

## Targetted Benchmark

Consider the following simple length buckets scenario which simply compares `FrozenDictionary` to `Dictionary`
with the string keys `"0"` to `"4"`. We can see from the results that the `FrozenDictionary` actually performs
worse than `Dictionary` already when looking up the 3rd item and slows down further form there. Also note
how the speed is incredibly linear - increasing by 2-3ns per bucket miss.

```csharp
public class Tests
{
    private Dictionary<string, string> _dictionary;
    private FrozenDictionary<string, string> _frozenDictionary;
    private string _key;

    [Params(0, 1, 2, 3, 4)]
    public int Item { get; set; }


    [GlobalSetup]
    public void Setup()
    {
        // Sequence of keys designed to be slower for BST as we go down the list.
        string[] keys = ["1", "0", "3", "2", "4"];

        _dictionary = keys.ToDictionary(k => k, k => k);
        _frozenDictionary = _dictionary.ToFrozenDictionary();

        if (_frozenDictionary.GetType().Name != "LengthBucketsFrozenDictionary`1")
        {
            throw new Exception($"Expected LengthBucketsFrozenDictionary`1 got {_frozenDictionary.GetType().Name}");
        }

        _key = keys[Item]; // NOTE: REFERENCE EQUALITY!
    }

    [Benchmark, IterationCount(5)]
    public string FrozenDictionary()
    {
        _frozenDictionary.TryGetValue(_key, out var result);
        return result;
    }

    [Benchmark, IterationCount(5)]
    public string Dictionary()
    {
        _dictionary.TryGetValue(_key, out var result);
        return result;
    }
}
```

.NET 8:

```
| Method           | Item | Mean      | Error     | StdDev    |
|----------------- |----- |----------:|----------:|----------:|
| FrozenDictionary | 0    |  2.589 ns | 0.0359 ns | 0.0093 ns |
| Dictionary       | 0    |  6.032 ns | 0.0628 ns | 0.0163 ns |
| FrozenDictionary | 1    |  4.636 ns | 0.1902 ns | 0.0494 ns |
| Dictionary       | 1    |  6.528 ns | 0.0707 ns | 0.0184 ns |
| FrozenDictionary | 2    |  7.022 ns | 0.3860 ns | 0.1002 ns |
| Dictionary       | 2    |  5.527 ns | 0.0511 ns | 0.0133 ns |
| FrozenDictionary | 3    |  9.142 ns | 0.2393 ns | 0.0621 ns |
| Dictionary       | 3    |  6.227 ns | 0.2128 ns | 0.0553 ns |
| FrozenDictionary | 4    | 10.046 ns | 0.2936 ns | 0.0763 ns |
| Dictionary       | 4    |  5.511 ns | 0.1445 ns | 0.0375 ns |
```

In this PR I propose to use a binary search within the buckets of the `LengthBucketFrozenDictionary`.
Originally on first implementation `string.CompareOrdinal` was always used however I was concerned that
this might result in a performance regression when the buckets were sparse, thus I changed the code
to keep track as to whether a node is an internal node or a leaf node.
This allowed me to always use `string.Equals` for leaf nodes.

Furthermore, due to the shape of the tree it made sense to extend the bucket size to 7 which gives a
maximum of 3 comparisons to reach any leaf node. This approach can also be made to work with a bucket
size of 3 (or any `2^n - 1` i.e. the number of nodes in a tree of depth `n - 1`).

The same benchmark after the change is shown below. Note that the first two buckets are comparable,
then the binary search tree starts to show it's strength by keeping up with the dictionary for these
small numbers.

THIS PR:

```
| Method           | Item | Mean     | Error     | StdDev    |
|----------------- |----- |---------:|----------:|----------:|
| FrozenDictionary | 0    | 2.385 ns | 0.0715 ns | 0.0186 ns |
| Dictionary       | 0    | 6.033 ns | 0.1481 ns | 0.0385 ns |
| FrozenDictionary | 1    | 3.471 ns | 0.0681 ns | 0.0105 ns |
| Dictionary       | 1    | 6.597 ns | 0.1984 ns | 0.0515 ns |
| FrozenDictionary | 2    | 3.771 ns | 0.1517 ns | 0.0394 ns |
| Dictionary       | 2    | 5.629 ns | 0.3129 ns | 0.0813 ns |
| FrozenDictionary | 3    | 5.301 ns | 0.4461 ns | 0.1158 ns |
| Dictionary       | 3    | 5.972 ns | 0.0713 ns | 0.0185 ns |
| FrozenDictionary | 4    | 5.160 ns | 0.0901 ns | 0.0139 ns |
| Dictionary       | 4    | 5.894 ns | 0.1598 ns | 0.0247 ns |
```

**However the benchmarks above do not tell the whole difference. As we will see the performance difference between `string.Equals` and `string.CompareOrdinal` starts to reverse the improvements.**

## Performance Repository MicroBenchmarks:

This benchmark compares `main` with `bst2` builds, representing original version and this PR respectively.
The story here is similar to the previous section.

```
|                             Method |            Toolchain | Count | ItemsPerBucket |           Mean |       Error |      StdDev |         Median |            Min |            Max | Ratio | RatioSD | Allocated | Alloc Ratio |
|----------------------------------- |--------------------- |------ |--------------- |---------------:|------------:|------------:|---------------:|---------------:|---------------:|------:|--------:|----------:|------------:|
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-bst2 |    10 |              1 |      19.236 ns |   0.2321 ns |   0.2171 ns |      19.269 ns |      18.972 ns |      19.575 ns |  1.11 |    0.02 |         - |          NA |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-main |    10 |              1 |      17.273 ns |   0.0991 ns |   0.0927 ns |      17.254 ns |      17.121 ns |      17.451 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                      |       |                |                |             |             |                |                |                |       |         |           |             |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-bst2 |    10 |              5 |      33.786 ns |   0.2214 ns |   0.1849 ns |      33.720 ns |      33.576 ns |      34.262 ns |  0.57 |    0.01 |         - |          NA |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-main |    10 |              5 |      58.930 ns |   0.7018 ns |   0.5860 ns |      58.959 ns |      57.934 ns |      59.917 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                      |       |                |                |             |             |                |                |                |       |         |           |             |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-bst2 |   100 |              1 |     168.341 ns |   1.4832 ns |   1.3873 ns |     168.565 ns |     165.949 ns |     170.577 ns |  1.07 |    0.02 |         - |          NA |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-main |   100 |              1 |     156.916 ns |   1.7621 ns |   1.6482 ns |     157.252 ns |     154.902 ns |     159.392 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                      |       |                |                |             |             |                |                |                |       |         |           |             |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-bst2 |   100 |              5 |     339.521 ns |   3.3500 ns |   2.9697 ns |     340.219 ns |     334.883 ns |     344.286 ns |  0.49 |    0.01 |         - |          NA |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-main |   100 |              5 |     686.032 ns |  15.5412 ns |  17.2740 ns |     692.513 ns |     657.596 ns |     710.328 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                      |       |                |                |             |             |                |                |                |       |         |           |             |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-bst2 |  1000 |              1 |   2,128.099 ns |   4.4982 ns |   4.2076 ns |   2,128.557 ns |   2,118.837 ns |   2,132.305 ns |  1.15 |    0.00 |         - |          NA |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-main |  1000 |              1 |   1,857.150 ns |   6.0052 ns |   5.3234 ns |   1,856.236 ns |   1,848.552 ns |   1,865.444 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                      |       |                |                |             |             |                |                |                |       |         |           |             |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-bst2 |  1000 |              5 |   3,643.782 ns |  11.4531 ns |  10.1529 ns |   3,644.262 ns |   3,630.004 ns |   3,661.826 ns |  0.57 |    0.02 |         - |          NA |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-main |  1000 |              5 |   6,367.205 ns | 224.4550 ns | 258.4827 ns |   6,213.334 ns |   6,095.559 ns |   6,766.709 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                      |       |                |                |             |             |                |                |                |       |         |           |             |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-bst2 | 10000 |              1 |  43,092.475 ns | 416.6385 ns | 369.3391 ns |  43,027.767 ns |  42,500.017 ns |  43,625.749 ns |  1.05 |    0.01 |         - |          NA |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-main | 10000 |              1 |  41,156.368 ns | 437.2025 ns | 408.9595 ns |  41,090.494 ns |  40,601.868 ns |  41,711.219 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                      |       |                |                |             |             |                |                |                |       |         |           |             |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-bst2 | 10000 |              5 |  69,793.566 ns | 815.9931 ns | 763.2804 ns |  69,533.960 ns |  68,840.985 ns |  70,885.122 ns |  0.66 |    0.01 |         - |          NA |
|  TryGetValue_True_FrozenDictionary | \dotnet-runtime-main | 10000 |              5 | 106,390.438 ns | 236.8056 ns | 184.8822 ns | 106,383.404 ns | 106,044.172 ns | 106,689.105 ns |  1.00 |    0.00 |         - |          NA |
```

Here are the same benchmarks but modified so that the keys do not have reference equality as they currently do in dotnet/performance repository.
These results tell a slightly different story. As the strings get longer then the optimization starts be less effective.
We will dig deeper into that in the next section.

```
|                             Method |             Toolchain | Count | ItemsPerBucket |             Mean |           Error |          StdDev |           Median |              Min |              Max | Ratio | RatioSD | Allocated | Alloc Ratio |
|----------------------------------- |---------------------- |------ |--------------- |-----------------:|----------------:|----------------:|-----------------:|-----------------:|-----------------:|------:|--------:|----------:|------------:|
|  TryGetValue_True_FrozenDictionary |   \dotnet-runtime-bst |    10 |              1 |        37.898 ns |       0.3379 ns |       0.2995 ns |        37.917 ns |        37.474 ns |        38.561 ns |  1.08 |    0.01 |         - |          NA |
|  TryGetValue_True_FrozenDictionary |  \dotnet-runtime-main |    10 |              1 |        35.079 ns |       0.2959 ns |       0.2768 ns |        35.058 ns |        34.629 ns |        35.745 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                       |       |                |                  |                 |                 |                  |                  |                  |       |         |           |             |
|  TryGetValue_True_FrozenDictionary |   \dotnet-runtime-bst |    10 |              5 |        53.955 ns |       0.2902 ns |       0.2423 ns |        53.974 ns |        53.345 ns |        54.276 ns |  0.76 |    0.01 |         - |          NA |
|  TryGetValue_True_FrozenDictionary |  \dotnet-runtime-main |    10 |              5 |        71.017 ns |       1.1520 ns |       1.0776 ns |        70.440 ns |        69.817 ns |        72.968 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                       |       |                |                  |                 |                 |                  |                  |                  |       |         |           |             |
|  TryGetValue_True_FrozenDictionary |   \dotnet-runtime-bst |   100 |              1 |       460.724 ns |       1.1635 ns |       0.9716 ns |       460.786 ns |       458.704 ns |       462.197 ns |  1.06 |    0.01 |         - |          NA |
|  TryGetValue_True_FrozenDictionary |  \dotnet-runtime-main |   100 |              1 |       433.490 ns |       2.6143 ns |       2.3175 ns |       433.055 ns |       430.037 ns |       437.868 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                       |       |                |                  |                 |                 |                  |                  |                  |       |         |           |             |
|  TryGetValue_True_FrozenDictionary |   \dotnet-runtime-bst |   100 |              5 |       639.204 ns |       4.7388 ns |       4.4326 ns |       640.327 ns |       630.924 ns |       645.519 ns |  0.66 |    0.01 |         - |          NA |
|  TryGetValue_True_FrozenDictionary |  \dotnet-runtime-main |   100 |              5 |       961.252 ns |       6.8251 ns |       6.0502 ns |       962.279 ns |       945.454 ns |       969.008 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                       |       |                |                  |                 |                 |                  |                  |                  |       |         |           |             |
|  TryGetValue_True_FrozenDictionary |   \dotnet-runtime-bst |  1000 |              1 |    20,265.188 ns |     140.1288 ns |     131.0765 ns |    20,256.145 ns |    19,958.841 ns |    20,499.288 ns |  1.06 |    0.01 |         - |          NA |
|  TryGetValue_True_FrozenDictionary |  \dotnet-runtime-main |  1000 |              1 |    19,055.889 ns |     197.4487 ns |     184.6937 ns |    19,022.599 ns |    18,800.854 ns |    19,359.878 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                       |       |                |                  |                 |                 |                  |                  |                  |       |         |           |             |
|  TryGetValue_True_FrozenDictionary |   \dotnet-runtime-bst |  1000 |              5 |     9,602.806 ns |      72.6225 ns |      60.6431 ns |     9,589.461 ns |     9,494.558 ns |     9,723.208 ns |  0.96 |    0.01 |         - |          NA |
|  TryGetValue_True_FrozenDictionary |  \dotnet-runtime-main |  1000 |              5 |     9,957.731 ns |      43.2599 ns |      38.3488 ns |     9,952.207 ns |     9,890.784 ns |    10,024.245 ns |  1.00 |    0.00 |         - |          NA |
|                                    |                       |       |                |                  |                 |                 |                  |                  |                  |       |         |           |             |
|  TryGetValue_True_FrozenDictionary |   \dotnet-runtime-bst | 10000 |              1 | 5,220,791.250 ns | 100,482.1981 ns |  93,991.1105 ns | 5,210,727.083 ns | 5,099,825.000 ns | 5,430,350.000 ns |  1.01 |    0.03 |      15 B |        1.00 |
|  TryGetValue_True_FrozenDictionary |  \dotnet-runtime-main | 10000 |              1 | 5,142,916.667 ns | 102,045.9059 ns | 113,423.7541 ns | 5,119,847.917 ns | 5,008,162.500 ns | 5,384,531.250 ns |  1.00 |    0.00 |      15 B |        1.00 |
|                                    |                       |       |                |                  |                 |                 |                  |                  |                  |       |         |           |             |
|  TryGetValue_True_FrozenDictionary |   \dotnet-runtime-bst | 10000 |              5 | 1,234,222.014 ns |  35,098.0528 ns |  39,011.3927 ns | 1,226,642.788 ns | 1,153,505.288 ns | 1,308,748.077 ns |  1.18 |    0.03 |       4 B |        1.33 |
|  TryGetValue_True_FrozenDictionary |  \dotnet-runtime-main | 10000 |              5 | 1,044,516.854 ns |  24,257.5349 ns |  27,935.0132 ns | 1,040,302.083 ns | 1,011,802.708 ns | 1,116,438.125 ns |  1.00 |    0.00 |       3 B |        1.00 |
```

## Benchmark: CompareOrdinal vs Equals

This benchmark is similar to the Targetted Benchmark earlier however the keys are much longer and have a long common prefix as shown below, and we don't use reference equality any more
```csharp
string[] keys = [
    new string('a', 1024) + "1",
    new string('a', 1024) + "0",
    new string('a', 1024) + "3",
    new string('a', 1024) + "2",
    new string('a', 1024) + "4",
];

// To better understand the equality calculation we avoid ReferenceEquals
_key = new StringBuilder().Append(keys[Item]).ToString(); // avoid reference equality for these tests
```

The idea here is to understand how longer strings are affected by `string.CompareOrdinal` vs `string.Equals`.

```
| Method           | Job        | Toolchain            | Item | Mean      | Error     | StdDev   | Ratio | RatioSD |
|----------------- |----------- |--------------------- |----- |----------:|----------:|---------:|------:|--------:|
| FrozenDictionary | Job-WQHAZE | \dotnet-runtime-bst  | 0    |  54.80 ns |  1.049 ns | 0.162 ns |  1.64 |    0.02 |
| FrozenDictionary | Job-EPVMQI | \dotnet-runtime-main | 0    |  33.40 ns |  1.201 ns | 0.312 ns |  1.00 |    0.00 |
|                  |            |                      |      |           |           |          |       |         |
| FrozenDictionary | Job-WQHAZE | \dotnet-runtime-bst  | 1    |  90.09 ns |  0.344 ns | 0.089 ns |  1.48 |    0.02 |
| FrozenDictionary | Job-EPVMQI | \dotnet-runtime-main | 1    |  60.85 ns |  3.599 ns | 0.935 ns |  1.00 |    0.00 |
|                  |            |                      |      |           |           |          |       |         |
| FrozenDictionary | Job-WQHAZE | \dotnet-runtime-bst  | 2    | 116.13 ns |  2.278 ns | 0.592 ns |  1.18 |    0.00 |
| FrozenDictionary | Job-EPVMQI | \dotnet-runtime-main | 2    |  98.05 ns |  1.173 ns | 0.305 ns |  1.00 |    0.00 |
|                  |            |                      |      |           |           |          |       |         |
| FrozenDictionary | Job-WQHAZE | \dotnet-runtime-bst  | 3    | 154.86 ns | 23.785 ns | 6.177 ns |  1.25 |    0.05 |
| FrozenDictionary | Job-EPVMQI | \dotnet-runtime-main | 3    | 123.52 ns |  7.894 ns | 2.050 ns |  1.00 |    0.00 |
|                  |            |                      |      |           |           |          |       |         |
| FrozenDictionary | Job-WQHAZE | \dotnet-runtime-bst  | 4    | 130.09 ns |  4.783 ns | 0.740 ns |  0.91 |    0.01 |
| FrozenDictionary | Job-EPVMQI | \dotnet-runtime-main | 4    | 142.14 ns |  2.632 ns | 0.684 ns |  1.00 |    0.00 |
```

Interestingly enough for longer keys the slower `string.CompareOrdinal` impedes performance too much to
make it up by fewer comparisons, such that it only wins out in the worst case.
This seems to also  hold for various key lengths which makes me think that this comes down to the
ratio of speeds of `string.CompareOrdinal` to `string.Equals`.

As we see from the next benchmark below `CompareOrdinal` performed on average 75% slower than `Equals`
which matches our observations above and means that as we approach larger shared string prefixes,
the effects of this binary search are dampened by `CompareOrdinal` vs `Equals`.
Is somehow optimizing `string.CompareOrdinal` something that can be looked into for .NET 9?
Is it worth it? Is it even possible to optimize it further?

```
| Method    | Length | Mean      | Error     | StdDev    | Ratio | RatioSD |
|---------- |------- |----------:|----------:|----------:|------:|--------:|
| Equals    | 1      |  2.286 ns | 0.0029 ns | 0.0007 ns |  1.00 |    0.00 |
| Compare   | 1      |  3.183 ns | 0.0110 ns | 0.0029 ns |  1.39 |    0.00 |
|           |        |           |           |           |       |         |
| Equals    | 5      |  2.099 ns | 0.0122 ns | 0.0032 ns |  1.00 |    0.00 |
| Compare   | 5      |  3.807 ns | 0.0342 ns | 0.0053 ns |  1.81 |    0.00 |
|           |        |           |           |           |       |         |
| Equals    | 10     |  1.882 ns | 0.0091 ns | 0.0024 ns |  1.00 |    0.00 |
| Compare   | 10     |  3.002 ns | 0.0224 ns | 0.0058 ns |  1.60 |    0.00 |
|           |        |           |           |           |       |         |
| Equals    | 50     |  2.903 ns | 0.0224 ns | 0.0035 ns |  1.00 |    0.00 |
| Compare   | 50     |  5.590 ns | 0.0811 ns | 0.0211 ns |  1.93 |    0.01 |
|           |        |           |           |           |       |         |
| Equals    | 100    |  5.040 ns | 0.0824 ns | 0.0214 ns |  1.00 |    0.00 |
| Compare   | 100    |  8.467 ns | 0.1393 ns | 0.0362 ns |  1.68 |    0.00 |
|           |        |           |           |           |       |         |
| Equals    | 500    | 13.383 ns | 1.4757 ns | 0.2284 ns |  1.00 |    0.00 |
| Compare   | 500    | 28.589 ns | 0.4152 ns | 0.1078 ns |  2.14 |    0.04 |
|           |        |           |           |           |       |         |
| Equals    | 1000   | 29.238 ns | 0.0687 ns | 0.0178 ns |  1.00 |    0.00 |
| Compare   | 1000   | 48.386 ns | 0.1223 ns | 0.0189 ns |  1.65 |    0.00 |
```

### Conclusion for Equals vs Compare
Unfortunately I was hoping for an unambiguous win but due to the trade-off
it all depends on usage patterns. The longer the shared prefix, the more time
spent in comparing strings, then the less effective this optimization will be.

One mitigation could be to introduce another branch which only does linear
search (i.e. code as-is) for strings greater than a certain threshold length. This should
be a single branch per invocation all strings are the same length.

## Other Concerns

Obviously the PR adds complexity to the code but that is easy to judge up-front.

There's also the concern that the bucket size is increased which means more memory being used.

Finally, I noticed that we are probably being over-eager in choosing the LengthBuckets strategy
in the first place. Choosing it for the keys `["a","b","c","d","e"]` is the wrong approach IMO
as the hash-based strategies are better in this case. Thus I intend to follow-up with some
experimentation to find a heuristic to avoid LengthBuckets in such cases. Having said that,
extending the bucket size from 5 to 7 means that LengthBuckets might be preferred
in even more cases where it might be suboptimal, so this follow-up work is highly recommended.


src/libraries/System.Collections.Immutable/src/System/Collections/Frozen/String/LengthBuckets.cs
```diff
@@ -43,9 +58,9 @@ internal static class LengthBuckets
             // we rent a single dimension array, where every bucket has five slots.
```
andrewjsaid, line 58: TODO If going forward. `Five` -> `Seven` and similar on next line.

src/libraries/System.Collections.Immutable/src/System/Collections/Frozen/String/LengthBuckets.cs
```diff
@@ -87,6 +102,70 @@ internal static class LengthBuckets
...
+#endif
+            IndexedKeyComparer indexedKeyComparer = new IndexedKeyComparer(
+                comparer == EqualityComparer<string>.Default
+                ? StringComparer.Ordinal
+                : (StringComparer)comparer, keys);
+            for (int bucketStartIndex = 0; bucketStartIndex < arraySize; bucketStartIndex += MaxPerLength)
+            {
+                // Within the bucket we order the elements to make binary search easier.
+                // The first element is the first to be compared and thus is the middle value.
+                // e.g. 0,1,2,3,4,5,6 becomes 3,1,0,2,5,4,6
```
Jozkee, line 118: If the first element should be in the middle, why the middle value in this sample is 2 and not 0? 
Shouldn't it be 5, 3, 1, 0, 2, 4, 6?

andrewjsaid: In this PR I used Pre-Order traversal of the binary tree. `3,1,0,2,5,4,6` is the pre-order traversal for the tree below.

1. Looking at the slice `[..]` aka `[0..7]` We compare with the first element (`3`) to choose which branch to visit next.
2. If our number is equal to `3` then success.
3. Otherwise
   - if our number is lower than `3` then repeat using slice `[1..4]` = `102` and compare with first element (`1`).
   - if our number is higher than `3` then repeat using slice `[4..7]` = `546` and compare with first element (`5`)
```
      3
    /   \
  1      5
 / \    / \
0   2  4   6
```

Using pre-order traversal, we always move forwards in the array and the root of the tree is always the first element.

Jozkee:
@andrewjsaid thanks for working on this, it is a good try given the comparison with the regular dictionary. 

> I noticed that we are probably being over-eager in choosing the LengthBuckets strategy in the first place.

Do you have perf results for this? It may be worth sharing them.

@adamsitnik a few questions:

* What's more common, to have more or less than half elements in the bucket?
* Any suggestions on how the string.Compare* penalty could be mitigated?

PS: TryGetValue_True_FrozenDictionary with no reference equality, 10000 lengh and 5 items per bucket is slower (1.18) than 1 item (1.01)? It seems like an outlier.

Jozkee:
@andrewjsaid I'm closing this PR as it is not obvious if it’s a desirable direction for the type. Hopefully once we get more data or one of the area owners can do a more in-depth analysis, we can re-open it. To clarify, we haven't taken a look due to focusing on items with higher priority.

andrewjsaid:
Thank you very much appreciate your and the team's time and understand that you have to prioritise.

PR #94782, opened 2023-11-15
Fix build issues with clang18

Raphtaliyah:
Clang 18 introduced some changes that prevent building CoreCLR:
* The warning for VLAs being a C++ extension is now on by default which causes errors with ``-Werror``. https://github.com/llvm/llvm-project/issues/62836
* ``stdarg.h`` now has a guard that depends on ``va_arg`` not being defined, currently the PAL internal doesn't undefined that macro (despite PAL defining it) which prevent ``stdarg.h`` from being included again and causes missing symbols (``va_start``, ``va_end``, ``va_arg``).

This PR fixes the warning by disabling it to match the behaviour of previous clang versions and undefines ``va_arg`` in ``palinternal.h``.

### DIFF
--- a/eng/native/configurecompiler.cmake
+++ b/eng/native/configurecompiler.cmake
@@ -564,6 +564,11 @@ if (CLR_CMAKE_HOST_UNIX)
     add_compile_options(-Wimplicit-fallthrough)
   endif()
 
+  # VLAs are non standard in C++, aren't available on Windows and
+  # are a warning by default since clang 18.
+  # For consistency, enable warnings for all compiler versions.
+  add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wvla>)
+
   #These seem to indicate real issues
   add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wno-invalid-offsetof>)
 
--- a/eng/native/configurecompiler.cmake
+++ b/eng/native/configurecompiler.cmake
@@ -606,6 +606,10 @@ if (CLR_CMAKE_HOST_UNIX)
     # other clang 16.0 suppressions
     add_compile_options(-Wno-single-bit-bitfield-constant-conversion)
     add_compile_options(-Wno-cast-function-type-strict)
+    
+    # clang 18.0 warns about VLAs being a C++ extension by default.
+    # https://github.com/llvm/llvm-project/issues/62836
+    add_compile_options(-Wno-vla-cxx-extension)
   else()
     add_compile_options(-Wno-uninitialized)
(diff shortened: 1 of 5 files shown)


Raphtaliyah:
@dotnet-policy-service agree

eng/native/configurecompiler.cmake
```diff
@@ -606,6 +606,10 @@ if (CLR_CMAKE_HOST_UNIX)
     # other clang 16.0 suppressions
     add_compile_options(-Wno-single-bit-bitfield-constant-conversion)
     add_compile_options(-Wno-cast-function-type-strict)
+    
+    # clang 18.0 warns about VLAs being a C++ extension by default.
```
jkotas, line 610: Where do we use VLAs?

Raphtaliyah: There are two instance while building for Linux:
https://github.com/dotnet/runtime/blob/590d664b637db8ee6a95063f7927143d6ff66455/src/coreclr/gc/unix/gcenv.unix.cpp#L635
and
https://github.com/dotnet/runtime/blob/590d664b637db8ee6a95063f7927143d6ff66455/src/coreclr/pal/src/file/path.cpp#L402

jkotas: If it is just these two, I think it is better to convert them to use alloca.

Raphtaliyah: That should be easy, I'll change them tomorrow! Should I also check if there are more in code that's only used for different platforms? Also if there are no notes in the coding guidelines it might be worth mentioning that VLAs shouldn't be used (and maybe enable warnings for them?)

jkotas: > That should be easy, I'll change them tomorrow!

Thanks!

> Should I also check if there are more in code that's only used for different platforms?

I do not expect that you will find any. VLAs are not supported on Windows that prohibits their use in majority of the code. And there is very little code in the repo that is non-Windows non-Linux.

> there are no notes in the coding guidelines it might be worth mentioning that VLAs shouldn't be used (and maybe enable warnings for them

VLAs are very niche. I do not think it is worth mentioning them in the coding guidelines. Enabling the warnings for them on older compilers sounds good if it is possible.

Raphtaliyah: Converted them to ``alloca`` and enabled warnings for unix hosts. If I understand it correctly only MSVC is supported for Windows which doesn't support VLAs (for C++), for Linux there is Clang and GCC, both of which support ``-Wvla``, this should cover every supported compiler/platform.

jkotas (approved):
Thank you!

PR #94783, opened 2023-11-15
Run CrossBlockLocalAssertionProp config

### LINKED ISSUE #3482: Experiment: Cross Block Local Assertion Prop
`DOTNET_JitEnableCrossBlockLocalAssertionProp=1`

cc @AndyAyersMS

cincuranet:
Fixes https://github.com/dotnet/performance/issues/3482.
Depends on https://github.com/dotnet/performance/pull/3495.

### DIFF
--- a/eng/pipelines/coreclr/perf-non-wasm-jobs.yml
+++ b/eng/pipelines/coreclr/perf-non-wasm-jobs.yml
@@ -236,6 +236,39 @@ jobs:
         logicalmachine: 'perftiger'
         r2rRunType: --nor2r
 
+  # run coreclr perftiger microbenchmarks CrossBlockLocalAssertionProp perf jobs
+  - template: /eng/pipelines/common/platform-matrix.yml
+    parameters:
+      jobTemplate: /eng/pipelines/coreclr/templates/perf-job.yml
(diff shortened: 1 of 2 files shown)


eng/pipelines/coreclr/perf-non-wasm-jobs.yml
```diff
@@ -236,6 +236,39 @@ jobs:
...
+      platforms:
+      - windows_x64
+      jobParameters:
+        testGroup: perf
+        liveLibrariesBuildConfig: Release
+        projectFile: microbenchmarks.proj
+        runKind: micro
+        runJobTemplate: /eng/pipelines/coreclr/templates/run-performance-job.yml
+        logicalmachine: 'perftiger'
+        r2rRunType: -ExperimentName crossblocklocalassertionprop
```
DrewScoggins, line 254: I assume that this will change to the experiment name parameter once the other side is checked in?

cincuranet: Damn copypasta. It should've been like that in the first place. Fixed now.

DrewScoggins (approved):
LGTM

PR #94791, opened 2023-11-15
Add empty SVE API for Arm64

a74nh:
Adds empty System/Runtime/Intrinsics/Arm/Sve.cs and hwintrinsiclistarm64sve.h.
Plus all connecting code.
Many of the files touched here are autogenerated via InstructionSetDesc.txt.


a74nh:
Extracted from https://github.com/dotnet/runtime/pull/94765
Currently includes all of @kunalspathak's https://github.com/dotnet/runtime/pull/94529

Not sure if the Arm64_Sve parts are required or if it should only be Sve. I think we want the readytorun/crossgen changes too (but it should be disabled for them?)

When run on a Arm64 Neoverse V1, the test program correctly outputs "SVE is supported."

a74nh:
@tannergooding @BruceForstall

a74nh:
Rebased off head as Kunal's work isn't dependant on this work.

src/coreclr/inc/corinfoinstructionset.h
```diff
@@ -31,14 +31,16 @@ enum CORINFO_InstructionSet
...
     InstructionSet_Rcpc2=15,
-    InstructionSet_ArmBase_Arm64=16,
-    InstructionSet_AdvSimd_Arm64=17,
-    InstructionSet_Aes_Arm64=18,
-    InstructionSet_Crc32_Arm64=19,
-    InstructionSet_Dp_Arm64=20,
-    InstructionSet_Rdm_Arm64=21,
-    InstructionSet_Sha1_Arm64=22,
-    InstructionSet_Sha256_Arm64=23,
+    InstructionSet_Sve=16,
```
kunalspathak, line 34: Any reason why we cannot add new entries at the bottom and not touch other entries?

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Arm/Sve.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Numerics;
+
+namespace System.Runtime.Intrinsics.Arm
+{
+    /// <summary>
+    /// This class provides access to the ARM SVE hardware instructions via intrinsics
+    /// </summary>
+    [Intrinsic]
+    [CLSCompliant(false)]
+    public abstract class Sve : AdvSimd
```
kunalspathak, line 16: what is an example of APIs that goes in `Sve` vs. `Sve.Arm64`? I thought we wanted something like `System.Runtime.Instrinsics.Arm.Sve` and `System.Runtime.Instrinsics.Arm.Sve2`?

src/coreclr/inc/corinfoinstructionset.h
```diff
@@ -282,6 +284,8 @@ struct CORINFO_InstructionSetFlags
             AddInstructionSet(InstructionSet_Sha1_Arm64);
         if (HasInstructionSet(InstructionSet_Sha256))
             AddInstructionSet(InstructionSet_Sha256_Arm64);
+        if (HasInstructionSet(InstructionSet_Sve))
```
kunalspathak, line 287: same here, want to understand difference between `InstructionSet_Sve` vs. `InstructionSet_Sve_Arm64` instead of `InstructionSet_Sve` and `InstructionSet_Sve2`.

src/tests/JIT/HardwareIntrinsics/Arm/Sve/Program.cs
```diff
@@ -0,0 +1,53 @@
...
+    public static partial class Program
+    {
+        [Fact]
+        public static void SveTest()
+        {
+            var test = new SveTest__SveTest();
+
+            if (test.IsSupported)
+            {
+                Console.WriteLine("SVE is supported.");
```
kunalspathak, line 31: did you verify on SVE hardware if the CPU flags are returned and set correctly?

src/coreclr/tools/Common/Compiler/HardwareIntrinsicHelpers.cs
```diff
@@ -273,6 +276,8 @@ public static int FromInstructionSet(InstructionSet instructionSet)
                     InstructionSet.ARM64_Atomics => Atomics,
                     InstructionSet.ARM64_Rcpc => Rcpc,
                     InstructionSet.ARM64_Rcpc2 => Rcpc2,
+                    InstructionSet.ARM64_Sve => Sve,
```
kunalspathak, line 279: As mentioned elsewhere, IMO, this should 
```c#
InstructionSet.Arm64_Sve => Sve,
InstructionSet.Arm64_Sve2 => Sve2,
```

kunalspathak (changes requested):
Early version looks good. Added some questions.

a74nh:
```
2023-11-15T18:21:52.2206448Z /__w/1/s/src/native/minipal/cpufeatures.c:359:17: error: use of undeclared identifier 'HWCAP_SVE'
```

What OS and version are the builds run on? Surprised this does not exist.

src/coreclr/inc/corinfoinstructionset.h
tannergooding: This file is autogenerated and the new entry is "at the bottom"

The generator basically generates:
* `Isa1`
* `Isa2`
* `...`
* `Isa1_But64Bit`
* `Isa2_But64Bit`
* `..._But64Bit`

src/coreclr/inc/corinfoinstructionset.h
tannergooding: It doesn't look like this PR is adding `Sve2`

src/coreclr/tools/Common/Compiler/HardwareIntrinsicHelpers.cs
tannergooding: This PR isn't adding Sve. These entries are handling `Sve` and the nested class `Sve.Arm64`, which is consistent with all other ISAs we've added so far.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Arm/Sve.cs
tannergooding: The general setup is that `Isa` exists to hold APIs that work on any arm platform

`Isa.Arm32` exists to hold APIs that only work on Arm32
`Isa.Arm64` exists to hold APIs that only work on Arm64 (we then have `Isa.X64` for xarch)

SVE is interesting because there are many APIs that "could" work on Arm32, but there are currently no plans (based on a prior discussion on one of the proposals) to expose that support there.

This means `Arm64` may be empty, but it must exist for usability reasons since `Sve` inherits from `AdvSimd` and otherwise `Sve.Arm64.IsSupported` would resolve to `AdvSimd.Arm64.IsSupported` and would lead users into a pit of failure, thinking it meant `Sve.IsSupported` would also be true

tannergooding (approved):
LGTM, but we can't merge until after API review has happened.

BruceForstall:
> What OS and version are the builds run on? Surprised this does not exist.

All .NET Linux builds are done inside a [CBL-Mariner](https://github.com/microsoft/CBL-Mariner) docker container. You can see the details if you look at the "Initializing containers" step of the build, e.g.: https://dev.azure.com/dnceng-public/public/_build/results?buildId=471209&view=logs&jobId=3f327aea-bfb2-5154-7b3c-2691dcdd6bb2&j=3f327aea-bfb2-5154-7b3c-2691dcdd6bb2&t=aa810107-5ca4-4707-8689-54adfbc0a8ea. In this case, it looks like the container is `mcr.microsoft.com/dotnet-buildtools/prereqs:cbl-mariner-2.0-cross-biarch-amd64-arm64`.

@sbomer is the expert on these: @sbomer or @a74nh do you know why Mariner might be missing the `HWCAP_SVE` declaration?

BruceForstall:
Looks like maybe SVE support should have shown up in Linux kernels from 5.15 in around 2017? https://www.kernel.org/doc/html/v5.15/arm64/sve.html

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Arm/Sve.cs
kunalspathak: ok, I didn't pay attention to `: AdvSimd` portion actually. I am wondering why `Sve` and `Sve.Arm64` do not inherit from `ArmBase` and `ArmBase.Arm64` respectively? I assume there will be a separate class for `Sve2` next to `Sve` class.

src/coreclr/inc/corinfoinstructionset.h
kunalspathak: sure, what I was trying to get is that since we have generated boiler plate code for encodings, etc. of both SVE and SVE2 instructions and we should define `SVE2` as well (instead of doing it  in future PR).

src/coreclr/tools/Common/Compiler/HardwareIntrinsicHelpers.cs
kunalspathak: yep, I was suggesting to do `Sve2` as well in this PR.

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Arm/Sve.cs
tannergooding: SVE inherits from AdvSimd because it requires AdvSimd support to exist and so it makes it overall more convenient for the user

AdvSimd then inherits from ArmBase itself

src/coreclr/inc/corinfoinstructionset.h
tannergooding: There is a ton of work to do and we should really focus on finishing SVE before we start on Sve2, IMO

We don’t want to be in a scenario where we leave Sve2 incomplete and half done

src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Arm/Sve.cs
kunalspathak: that makes sense.

src/coreclr/inc/corinfoinstructionset.h
a74nh: Alternative suggestion: We should focus first on the the instructions that give extra functionality that are not in NEON and will be useful for the existing SDK libraries. That may include some SVE2 instructions. Doing that would ensure some performance uplift for .NET9.

On my TODO list is to rewrite some of the SDK routines in SVE and see if I can find any performance improvements. Then use that to prioritise the SVE work.

a74nh:
> Looks like maybe SVE support should have shown up in Linux kernels from 5.15 in around 2017? https://www.kernel.org/doc/html/v5.15/arm64/sve.html
> 
> (also, https://docs.kernel.org/next/arm64/elf_hwcaps.html)

Mariner is used for WSL. It's possible this has been disabled due to Windows not yet supporting SVE.

I'll add the following:
```
#ifndef HWCAP_SVE
#define HWCAP_SVE	(1 << 22)
#endif
```
That's safe as the kernel is never going to return that value on Mariner.

It's also what we did in gdb and also one of the [utilities in the Linux kernel](https://github.com/torvalds/linux/blob/c42d9eeef8e5ba9292eda36fd8e3c11f35ee065c/tools/perf/arch/arm64/util/perf_regs.c#L15) is doing is too.

src/tests/JIT/HardwareIntrinsics/Arm/Sve/Program.cs
a74nh: Yes - both with and without:

Graviton 3:

❯ $CORE_ROOT/corerun /home/alahay01/dotnet/runtime_sve/artifacts/tests/coreclr/linux.arm64.Release/JIT/HardwareIntrinsics/Arm/Sve/Sve/Sve.dll
SVE is supported.

Altra:

❯ $CORE_ROOT/corerun /home/alahay01/dotnet/runtime_sve/artifacts/tests/coreclr/linux.arm64.Release/JIT/HardwareIntrinsics/Arm/Sve/Sve/Sve.dll
SVE is not supported.

a74nh:
> LGTM, but we can't merge until after API review has happened.

To get this merged, we only need to get one piece of the API reviewed?

src/coreclr/inc/corinfoinstructionset.h
tannergooding: It would definitely be helpful to know what that list is.

Having an incomplete ISA is generally something we try to avoid as it can lead to issues when exposing APIs later, missed functionality, etc.

kunalspathak:
With InstructionSet changes in a separate PR https://github.com/dotnet/runtime/pull/95021 , this PR just adds the empty `hwintrinsiclistarm64sve.h` then?

a74nh:
> With InstructionSet changes in a separate PR #95021 , this PR just adds the empty `hwintrinsiclistarm64sve.h` then?

Done!

a74nh:
After discussion with @kunalspathak, this PR is mostly pointless now. Closing.

PR #94795, opened 2023-11-15
Fix Vector512.CopyTo

### LINKED ISSUE #94743: Vector512 CopyTo<T>(this Vector512<T> vector, T[] destination, int startIndex) incorrect
### Description

The current implementation of Vector512.CopyTo<T> uses a byte offset into the array instead of an offset for the size of the given type.  This causes unexpected results for all types outside of byte.

### Reproduction Steps

```cs
var a = new float[17];
var toSet = Vector512.Create(2.0f);
toSet.CopyTo(a, 1);
for (int i = 0; i < a.Length; i++)
{
    Console.WriteLine($"{i} = {a[i]}");
}
```

### Expected behavior

0 = 0
1 = 2
2 = 2
3 = 2
4 = 2
5 = 2
6 = 2
7 = 2
8 = 2
9 = 2
10 = 2
11 = 2
12 = 2
13 = 2
14 = 2
15 = 2
16 = 2

### Actual behavior

0 = 0
1 = 9E-44
2 = 9E-44
3 = 9E-44
4 = 9E-44
5 = 9E-44
6 = 9E-44
7 = 9E-44
8 = 9E-44
9 = 9E-44
10 = 9E-44
11 = 9E-44
12 = 9E-44
13 = 9E-44
14 = 9E-44
15 = 9E-44
16 = 9E-44

### Regression?

_No response_

### Known Workarounds

```cs
var a = new float[17];
var toSet = Vector512.Create(2.0f);
Vector512.StoreUnsafe(toSet, ref a[1]);
for (int i = 0; i < a.Length; i++)
{
    Console.WriteLine($"{i} = {a[i]}");
}
```

### Configuration

.net 8.0.100
Windows 11
x64, AMD 7950X.

### Other information

_No response_

MichalPetryka:
While doing this, I've unified the logic to be the same for all the vector types.

@tannergooding does this make sense for backporting to .Net 8?

Fixes #94743.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Vector512.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Vector512.cs
@@ -442,8 +442,7 @@ namespace System.Runtime.Intrinsics
                 ThrowHelper.ThrowArgumentException_DestinationTooShort();
             }
 
-            ref T address = ref MemoryMarshal.GetArrayDataReference(destination);
-            Unsafe.WriteUnaligned(ref Unsafe.As<T, byte>(ref address), vector);
+            Unsafe.WriteUnaligned(ref Unsafe.As<T, byte>(ref destination[0]), vector);
         }
 
         /// <summary>Copies a <see cref="Vector512{T}" /> to a given array starting at the specified index.</summary>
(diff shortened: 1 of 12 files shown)


src/libraries/System.Private.CoreLib/src/System/Runtime/Intrinsics/Vector512.cs
```diff
@@ -469,8 +469,8 @@ public static void CopyTo<T>(this Vector512<T> vector, T[] destination, int star
                 ThrowHelper.ThrowArgumentException_DestinationTooShort();
             }
 
-            ref byte address = ref Unsafe.As<T, byte>(ref MemoryMarshal.GetArrayDataReference(destination));
-            Unsafe.WriteUnaligned(ref Unsafe.Add(ref address, startIndex), vector);
+            ref T address = ref MemoryMarshal.GetArrayDataReference(destination);
```
tannergooding, line 472: Rather than changing 7 other files, you should probably just make this match what `V256` was doing.

It is a smaller change, more consistent, and avoids introducing new locals.

A regression test likewise needs to be added

tannergooding (approved):

Issue #94119, opened 2023-10-27
Support TimeSpan type in the RangeAttribute emitted code by options validation source generator 


tarekgh:
The RangeAttribute employs a TypeConverter, which renders it incompatible with AOT compilation. The Options source generator identifies the usage of the RangeAttribute and generates strongly typed attribute code. The generator currently supports this only for a predefined list of basic types, as indicated here: [link](https://github.com/dotnet/runtime/blob/9461f4b5bd036005f37e78516c48b13396f76750/src/libraries/Microsoft.Extensions.Options/gen/OptionsSourceGenContext.cs#L45).

TimeSpan is not included in the supported list, despite its use in real application scenarios. This issue is being tracked to add support for TimeSpan and possibly other types like DateTimeOffset.

tarekgh:
CC @eerhardt @ericstj

Clockwork-Muse:
`TimeSpan` is going to be a problem due to the whole "the value 24+ is read as days instead of hours" thing.

tarekgh:
>  is going to be a problem due to the whole "the value 24+ is read as days instead of hours" thing.

Why do you think this is a problem? The behavior is well defined, and users have control on the values they can use in the range attribute.

Clockwork-Muse:
Principle of least surprise problems.  The behavior is well defined, sure, but that assumes that the people doing the configuration know that whatever application has been written in C#, and that the element type follows the given behavior.
Config makes it more likely to obscure this requirement...

tarekgh:
The change here is not going to add or introduce any behavior change. The `RangeAttribute` today can accept `TimeSpan` and have the defined behavior. The support we are talking about here is the source generator using `RangeAttribute`. Cannot have different behavior for source generator than the runtime.

eiriktsarpalis:
> The RangeAttribute employs a TypeConverter, which renders it incompatible with AOT compilation.

I believe this could be addressed in part using the API as originally proposed here: https://github.com/dotnet/runtime/issues/82526. API review at the time indicated that we should focus on shipping a new generic range type instead, but wondering if this scenario is sufficient to reconsider. cc @jeffhandley

tarekgh:
@eiriktsarpalis this will help except of users already using the constructor `public [RangeAttribute(Type type, string minimum, string maximum)` we need to handle this case in the source gen till we do other things. I can imagine we can run into other similar types too (like DateTimeOffset as another example).

eiriktsarpalis:
[...]

PR #94798, opened 2023-11-15
Support TimeSpan with RangeAttribute in Options validation source generator

tarekgh:
https://github.com/dotnet/runtime/issues/94119

### DIFF
--- a/src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
+++ b/src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
@@ -374,11 +374,82 @@ namespace Microsoft.Extensions.Options.Generators
 """);
         }
 
-        public void EmitRangeAttribute(string modifier, string prefix, string className, string suffix)
+        public void EmitRangeAttribute(string modifier, string prefix, string className, string suffix, bool emitTimeSpanSupport)
         {
             OutGeneratedCodeAttribute();
 
             string qualifiedClassName = $"{prefix}{suffix}_{className}";
+
+            string initializationString = emitTimeSpanSupport ?
+            """
+                                if (OperandType == typeof(global::System.TimeSpan))
+                                {
+                                    if (!global::System.TimeSpan.TryParse((string)Minimum, culture, out global::System.TimeSpan timeSpanMinimum) ||
+                                        !global::System.TimeSpan.TryParse((string)Maximum, culture, out global::System.TimeSpan timeSpanMaximum))
+                                    {
+                                        throw new global::System.InvalidOperationException(c_minMaxError);
+                                    }
+                                    Minimum = timeSpanMinimum;
+                                    Maximum = timeSpanMaximum;
+                                }
+                                else
+                                {
+                                    Minimum = ConvertValue(Minimum, culture) ?? throw new global::System.InvalidOperationException(c_minMaxError);
+                                    Maximum = ConvertValue(Maximum, culture) ?? throw new global::System.InvalidOperationException(c_minMaxError);
+                                }
+            """
+            :
+            """
+                                Minimum = ConvertValue(Minimum, culture) ?? throw new global::System.InvalidOperationException(c_minMaxError);
+                                Maximum = ConvertValue(Maximum, culture) ?? throw new global::System.InvalidOperationException(c_minMaxError);
+            """;
+
+            string convertValue = emitTimeSpanSupport ?
+            """
+                        if (OperandType == typeof(global::System.TimeSpan))
+                        {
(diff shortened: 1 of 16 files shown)


tarekgh:
This is a candidate to port to 8.0 servicing. 

CC @ericstj @eerhardt

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
```diff
@@ -374,12 +374,83 @@ public void EmitCompareAttribute(string modifier, string prefix, string classNam
...
         {
             OutGeneratedCodeAttribute();
 
             string qualifiedClassName = $"{prefix}{suffix}_{className}";
 
+            string initializationString = emitTimeSpanSupport ?
+            """
+                                if (OperandType == typeof(global::System.TimeSpan))
+                                {
+                                    if (!global::System.TimeSpan.TryParse((string)Minimum, culture, out global::System.TimeSpan timeSpanMinimum) ||
```
eiriktsarpalis, line 387: Could this parsing happen at compile time, or do we have to honor local culture settings?

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
```diff
@@ -374,12 +374,83 @@ public void EmitCompareAttribute(string modifier, string prefix, string classNam
 """);
         }
 
-        public void EmitRangeAttribute(string modifier, string prefix, string className, string suffix)
+        public void EmitRangeAttribute(string modifier, string prefix, string className, string suffix, bool emitTimeSpanSupport)
```
eiriktsarpalis, line 377: Why is this emitting timespans conditional out of curiosity?

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
```diff
@@ -414,19 +485,20 @@ public void EmitRangeAttribute(string modifier, string prefix, string className,
                 string.Format(global::System.Globalization.CultureInfo.CurrentCulture, GetValidationErrorMessage(), name, Minimum, Maximum);
         private bool NeedToConvertMinMax { get; }
         private bool Initialized { get; set; }
+        private const string c_minMaxError = "The minimum and maximum values must be set to valid values.";
```
eiriktsarpalis, line 488: I would recommend extracting all error messages to a standalone source file.

src/libraries/Microsoft.Extensions.Options/gen/Parser.cs
```diff
@@ -641,23 +652,25 @@ private void TrackRangeAttributeForSubstitution(AttributeData attribute, ITypeSy
...
-                typeSymbol = ((INamedTypeSymbol)typeSymbol).TypeArguments[0];
-            }
-
-            if (argumentSpecialType != SpecialType.None &&
-                OptionsSourceGenContext.IsConvertibleBasicType(typeSymbol) &&
-                (constructorParameters.Length != 3 || typeSymbol.SpecialType == argumentSpecialType)) // When type is provided as a parameter, it has to match the property type.
+            if (argumentType is not null)
             {
-                _optionsSourceGenContext.EnsureTrackingAttribute(attribute.AttributeClass!.Name, createValue: false, out _);
+                _optionsSourceGenContext.EnsureTrackingAttribute(attribute.AttributeClass!.Name, createValue: hasTimeSpanType, out _);
```
eiriktsarpalis, line 673: I guess this asnwers my previous question :-)

eiriktsarpalis (approved):
LGTM modulo one question I had.

eerhardt:
I tested this fix with the https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.Http.Resilience library (which uses the Options Validation source generator) and I'm able to successfully publish a Native AOT application using the library built with this fix with no AOT warnings.

Thanks, @tarekgh!

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
tarekgh: The culture is one reason. The other reason is if I parse it at compile time will need to pass the parsed values to the attribute constructors (possibly as objects) which will need adding a new constructor there and will need to add more emitting code to generate such constructor parameter. I don't think this is worth doing it in servicing release.

src/libraries/Microsoft.Extensions.Options/gen/Emitter.cs
tarekgh: We can do that separately to reduce the changes we are porting to servicing.

PR #94816, opened 2023-11-15
Enable dprintf logging for the whole GC

cshung:
The change lifts some function declaration, the definition of the `dprintf` macro, and the underlying file implementation to `gc.h` and `gccommon.cpp`. This will allow us to add `dprintf` to code other than just `gc.cpp`  in the gc code base. (e.g. `handletable.cpp`)

### DIFF
--- a/src/coreclr/gc/gc.h
+++ b/src/coreclr/gc/gc.h
@@ -341,4 +341,56 @@ inline bool IsServerHeap()
 #endif // FEATURE_SVR_GC
 }
 
+#ifndef MAX_LONGPATH
+#define MAX_LONGPATH 1024
+#endif // MAX_LONGPATH
+
+// #define TRACE_GC
+// #define SIMPLE_DPRINTF
+
+#ifdef TRACE_GC
+#define MIN_CUSTOM_LOG_LEVEL 7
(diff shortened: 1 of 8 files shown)


mrsharm (approved):
LGTM!

src/coreclr/gc/gc.h
```diff
@@ -341,4 +341,56 @@ inline bool IsServerHeap()
...
+// do post processing on it with tools\GCLogParser. If you need to add some
+// detail to help with investigation that's not 't processed by tooling
+// prefix it with TEMP so that line will be written to the results as is in
+// the result. I have some already logged with HEAP_BALANCE_TEMP_LOG.
+#define HEAP_BALANCE_LOG (MIN_CUSTOM_LOG_LEVEL + 10)
+#define HEAP_BALANCE_TEMP_LOG (MIN_CUSTOM_LOG_LEVEL + 11)
+
+#ifdef SIMPLE_DPRINTF
+
+HRESULT initialize_log_file();
```
Maoni0, line 375: >HRESULT [](http://example.com/codeflow?start=0&length=7)

could we make this a bool instead?

Maoni0 (approved):
:shipit:

PR #94825, opened 2023-11-16
Improve the hashtable for EEClassHash

davidwrighton:
The `EEClassHashTable` implementation failed to use a hash function which included enclosing types, and has a variety of other deficiencies where it does unnecessary work. Include the enclosing type in the hash function, and reduce the amount of code duplication in the hash table.

In addition, I'm taking this opportunity to clean up the implementation substantially, creating a single path for insertion, and lookup.


davidwrighton:
Before checkin, I need to validate that we actually have some tests for the TypeRef that points at a TypeDef scenario.

docs/design/specs/Ecma-335-Augments.md
```diff
@@ -381,6 +381,12 @@ The text should be deleted:
 
 > Furthermore, ~~the InterfaceImpl table is sorted using the Interface column as a secondary key, and~~ the GenericParam table is sorted using the Number column as a secondary key.
 
+In addition to the TypeDef table having a special ordering constraint, the ExportedTypes table ALSO has the same constraint.
+
+This line should be changed.
+
+> Finally, this TypeDef _and ExportedType_ ~~table has~~ _tables have_ a special ordering constraint: the definition of an enclosing class shall precede the definintion of all classes it encloses.
```
jkotas, line 388:
```suggestion
> Finally, this TypeDef _and ExportedType_ ~~table has~~ _tables have_ a special ordering constraint: the definition of an enclosing class shall precede the definition of all classes it encloses.
```

src/coreclr/vm/clsload.cpp
```diff
@@ -3708,6 +3484,7 @@ VOID ClassLoader::AddAvailableClassDontHaveLock(Module *pModule,
     AddAvailableClassHaveLock(
         pModule,
         classdef,
+        NULL,
```
AaronRobinsonMSFT, line 3487: Instead of `NULL`, you could declare `SArray<EEClassHashEntry_t *> classEntries();` above and it will have 0 length by construction. It avoids creating a `NULL` check semantic in this code path.

src/coreclr/inc/corhlprpriv.h
```diff
@@ -294,9 +294,9 @@ class CQuickMemoryBase
     }
 
     // Copy single byte string and hold it
-    const char * SetStringNoThrow(const char * pStr, SIZE_T len)
+    const char * SetString(const char * pStr, SIZE_T len)
     {
-        LPSTR buffer = (LPSTR) AllocNoThrow(len + 1);
+        LPSTR buffer = (LPSTR) AllocThrows(len + 1);
```
hughbe, line 299: sorry if this is wrong, but will `buffer` never be `NULL`, because if `AllocThrows` fails, it throws? So we can remove the `if` check below

davidwrighton: Nice catch. The if check below is superfluous.

src/coreclr/vm/clsload.cpp
davidwrighton: Sure. This reduces the complexity of the code a bit.

src/coreclr/vm/classhash.h
```diff
@@ -97,15 +91,32 @@ class EEClassHashTable : public DacEnumerableHashTable<EEClassHashTable, EEClass
...
+
+    static BOOL IsNested(ModuleBase *pModule, mdToken token, mdToken *mdEncloser);
+    static BOOL IsNested(const NameHandle* pName, mdToken *mdEncloser);
+
+    PTR_EEClassHashTable m_pCaseSensitiveTable;  // This will be a recursive pointer for the case sensitive table
 };
 
+inline DWORD GetHash(PTR_EEClassHashEntry eeclassHashEntry)
+{
+    // This is only safe, as the only allocator for the EEClassHashEntry type is the hash table
```
AaronRobinsonMSFT, line 118:
```suggestion
    // This is safe, as the only allocator for the EEClassHashEntry type is the hash table
```

Right?

src/coreclr/vm/classhash.cpp
```diff
@@ -223,17 +223,17 @@ VOID EEClassHashTable::ConstructKeyFromData(PTR_EEClassHashEntry pEntry, // IN
...
     CONTRACTL_END;
 
     {
 #ifdef _DEBUG_IMPL
-        _ASSERTE(!(m_bCaseInsensitive && FORBIDGC_LOADER_USE_ENABLED()));
+        _ASSERTE(!(IsCaseInsensitiveTable() && FORBIDGC_LOADER_USE_ENABLED()));
 #endif
 
-        // cqb - If m_bCaseInsensitive is true for the hash table, the bytes in Key will be allocated
+        // cqb - If IsCaseInsensitiveTable() is true for the hash table, the bytes in Key will be allocated
```
AaronRobinsonMSFT, line 236: What is cqb? I don't see that anywhere in here.

src/coreclr/vm/clsload.cpp
```diff
@@ -881,21 +675,12 @@ void ClassLoader::LazyPopulateCaseSensitiveHashTables()
         // (either images compiled with an old version of crossgen, or for case-insensitive type lookups in R2R modules)
         _ASSERT(pModule->IsReadyToRun());
 
-        EEClassHashTable * pNewClassHash = EEClassHashTable::Create(pModule, AVAILABLE_CLASSES_HASH_BUCKETS, FALSE /* bCaseInsensitive */, &amTracker);
+        EEClassHashTable * pNewClassHash = EEClassHashTable::Create(pModule, AVAILABLE_CLASSES_HASH_BUCKETS, NULL, &amTracker);
```
AaronRobinsonMSFT, line 678: Do you want this to be `PTR_EEClassHashTable`?

src/coreclr/vm/clsload.cpp
```diff
@@ -3865,11 +3587,13 @@ VOID ClassLoader::AddExportedTypeDontHaveLock(Module *pManifestModule,
     AddExportedTypeHaveLock(
         pManifestModule,
         cl,
+        NULL,
```
AaronRobinsonMSFT, line 3590: Remove `NULL` and use empty `SArray<EEClassHashEntry_t *>`?

AaronRobinsonMSFT (approved):
A few nits, but overall looks like a win.

AaronRobinsonMSFT (approved):
A few nits, but overall looks like a win.

src/coreclr/vm/clsload.cpp
davidwrighton: No, SetAvailableClassHash takes a EEClassHashTable* and that's what we need there.

src/coreclr/vm/classhash.cpp
davidwrighton: It isn't there at all. This comment is confusing as it refers to a state in the code which hasn't existed in decades. I've tried to fix up the comment to make some amount of sense, but in general everything to do with the case insensitive table is nonsense.

src/coreclr/vm/clsload.cpp
davidwrighton: Sure. I did it for the other case.

PR #94828, opened 2023-11-16
Enable SA1021 Negative sign should be preceded by a space

### LINKED ISSUE #94827: SA1021: Negative sign should be preceded by a space

artl93:
Enables SA1021 Negative sign should be preceded by a space
Fixes #94827

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TypeThreadStaticIndexNode.cs
+++ b/src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TypeThreadStaticIndexNode.cs
@@ -68,7 +68,7 @@ namespace ILCompiler.DependencyAnalysis
                     // an index in the containing storage.
                     // We use a negative index to indicate that. Any negative value would work.
                     // For the purpose of natvis we will encode the offset of the type storage within the block.
-                    typeTlsIndex = - (_inlinedThreadStatics.GetTypeStorageOffset(_type) + factory.Target.PointerSize);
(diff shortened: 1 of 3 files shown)


artl93:
For the record, this has surprisingly little value, but this is my "exercise to the reader".

src/coreclr/tools/aot/ILCompiler.Compiler/Compiler/DependencyAnalysis/TypeThreadStaticIndexNode.cs
```diff
@@ -68,7 +68,7 @@ protected override ObjectData GetDehydratableData(NodeFactory factory, bool relo
                     // an index in the containing storage.
                     // We use a negative index to indicate that. Any negative value would work.
                     // For the purpose of natvis we will encode the offset of the type storage within the block.
-                    typeTlsIndex = - (_inlinedThreadStatics.GetTypeStorageOffset(_type) + factory.Target.PointerSize);
+                    typeTlsIndex = (-1 * (_inlinedThreadStatics.GetTypeStorageOffset(_type) + factory.Target.PointerSize));
```
tannergooding, line 71: This seems strictly worse. I’d expect it to just be`-(…)` rather than multiplying by -1

(I’m not sure if we recognize and optimize that pattern either)

artl93: Yeah... I'm ready to chuck this PR entirely. This rule already appears to have little value on the balance.

PR #94838, opened 2023-11-16
Enable disabled HTTP3 ALPN check

rzikm:
Stumbled upon an old disabled piece of code.  I don't think we can hit this code path unless something goes very wrong elsewhere but it does not hurt to be careful.


ManickaP (approved):
I hope it will not bite us with someone still using h3-29 or something like that...

rzikm:
@ManickaP since we are not offering draft h3 versions ourselves, it can't be negotiated

https://github.com/dotnet/runtime/blob/910782efdd6ab1cb61f00c71fd9e4f2bff32ba0c/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionPool.cs#L319

rzikm:
Failures are known build errors.

PR #94839, opened 2023-11-16
Fix format.sh script and remove dotnet-format tool from dotnet-tools.json

akoeplinger:
We're no longer using the separate dotnet-format tool since it is integrated into the dotnet SDK now.

With the move to the SDK the options changed a bit so we now need to use the `whitespace` format command so we can continue using the `--folder` option: https://github.com/dotnet/format/issues/1385

To run not just whitespace but code style formatters as well we'd need a workspace context (i.e. pass the .csproj to dotnet format), but inferring that from just the changed file list is hard.

### DIFF
--- a/eng/formatting/format.sh
+++ b/eng/formatting/format.sh
@@ -17,7 +17,7 @@ fi
 
 if [ -n "$MANAGED_FILES" ]; then
     # Format all selected files
-    echo "$MANAGED_FILES" | cat | xargs | sed -e 's/ /,/g' | dotnet format --no-restore --include -
+    echo "$MANAGED_FILES" | cat | xargs | sed -e 's/ /,/g' | dotnet format whitespace --include - --folder
 
     # Add back the modified files to staging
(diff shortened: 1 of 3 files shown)


eng/formatting/format.sh
```diff
@@ -17,7 +17,7 @@ fi
 
 if [ -n "$MANAGED_FILES" ]; then
     # Format all selected files
-    echo "$MANAGED_FILES" | cat | xargs | sed -e 's/ /,/g' | dotnet format --no-restore --include -
+    echo "$MANAGED_FILES" | cat | xargs | sed -e 's/ /,/g' | dotnet format whitespace --include - --folder
```
ViktorHofer, line 20: Is this script invoked in CI or just a dev innerloop thing?

ViktorHofer (approved):

eng/formatting/format.sh
akoeplinger: Just a dev innerloop thing.

PR #94845, opened 2023-11-16
Count16BitElements() intrinsic for SVE

a74nh:
Extracted from https://github.com/dotnet/runtime/pull/94765

For now, this is based off both https://github.com/dotnet/runtime/pull/94529 and https://github.com/dotnet/runtime/pull/94791

All the new code starts from https://github.com/dotnet/runtime/pull/94845/commits/3cb75c5b2cdec488a15d551c31839333c837a19d

For now it emits an `MRS` instruction. This needs updating to use an SVE `CNTH`, however, the codegen parts need adding alongside. Will wait until I can use the tool.

`JitDump`ing the test program, I see the `MRS`


src/tests/JIT/HardwareIntrinsics/Arm/Sve/Program.cs
```diff
@@ -0,0 +1,66 @@
...
+using System;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.Arm;
+using System.Numerics;
+using Xunit;
+
+namespace JIT.HardwareIntrinsics.Arm._AdvSimd
```
jkotas, line 20: Should this be _Sve?

a74nh:
This needs rebasing. But still can't progress into until skeleton library files are committed.

PR #94852, opened 2023-11-16
JIT: [experiment] Build SSA after early liveness

jakobbotsch:
Given that we already compute liveness and thread the locals I was curious to see how expensive it would be to also build SSA. This does that, and then immediately throws away all SSA as part of forward sub.


jakobbotsch:
cc @dotnet/jit-contrib. [Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=472585&view=ms.vss-build-web.run-extensions-tab) show about 1-2% throughput impact.
image

image

Some detailed throughput details: https://gist.github.com/jakobbotsch/fe27fe5efba78870177c4d72cb6270b5

PR #94853, opened 2023-11-16
[RISC-V] Implement NativeWalker.

viewizard:
With this changes managed debugger (tested with netcoredbg) can steps through managed code.

CC @clamp03 @wscho77 @HJLeee @JongHeonChoi @t-mustafin @gbalykov @brucehoult

### DIFF
--- a/src/coreclr/debug/ee/riscv64/walker.cpp
+++ b/src/coreclr/debug/ee/riscv64/walker.cpp
@@ -37,6 +37,14 @@ inline uint64_t BitExtract(uint64_t value, unsigned int highbit, unsigned int lo
     return signExtend ? SignExtend(extractedValue, highbit - lowbit) : extractedValue;
 }
 
+uint64_t NativeWalker::GetReg(uint64_t reg)
+{
+    _ASSERTE(reg <= 31);
+    _ASSERTE(m_registers->pCurrentContext->R0 == 0);
+
+    return (&m_registers->pCurrentContext->R0)[reg];
+}
+
 void NativeWalker::Decode()
 {
     // Reset so that we do not provide bogus info
@@ -50,8 +58,6 @@ void NativeWalker::Decode()
        // Without registers decoding will work only for handful of instructions
        return;
     }
-
-    PT_CONTEXT context = m_registers->pCurrentContext;
 
     // Fetch first word of the current instruction. If the current instruction is a break instruction, we'll
     // need to check the patch table to get the correct instruction.
@@ -95,7 +101,7 @@ void NativeWalker::Decode()
         uint64_t Rs1 = BitExtract(opcode, 19, 15);
         uint64_t Rd = BitExtract(opcode, 11, 7);
 
-        m_nextIP = (BYTE*)((GetReg(context, Rs1) + imm) & ~1ull);
+        m_nextIP = (BYTE*)((GetReg(Rs1) + imm) & ~1ull);
         // The standard software calling convention uses X1 as the return address register and X5 as an alternate link register.
         if (Rd == 1 || Rd == 5)
         {
@@ -119,8 +125,8 @@ void NativeWalker::Decode()
     {
         uint64_t Rs1 = BitExtract(opcode, 19, 15);
         uint64_t Rs2 = BitExtract(opcode, 24, 20);
-        int64_t Rs1SValue = GetReg(context, Rs1);
-        int64_t Rs2SValue = GetReg(context, Rs2);
+        int64_t Rs1SValue = GetReg(Rs1);
+        int64_t Rs2SValue = GetReg(Rs2);
 
         if ((((opcode & 0x707f) == 0x63) && Rs1SValue == Rs2SValue) ||
             (((opcode & 0x707f) == 0x1063) && Rs1SValue != Rs2SValue) ||
@@ -143,8 +149,8 @@ void NativeWalker::Decode()
     {
         uint64_t Rs1 = BitExtract(opcode, 19, 15);
         uint64_t Rs2 = BitExtract(opcode, 24, 20);
-        uint64_t Rs1Value = GetReg(context, Rs1);
-        uint64_t Rs2Value = GetReg(context, Rs2);
+        uint64_t Rs1Value = GetReg(Rs1);
+        uint64_t Rs2Value = GetReg(Rs2);
 
         if ((((opcode & 0x707f) == 0x6063) && Rs1Value < Rs2Value) ||
             (((opcode & 0x707f) == 0x7063) && Rs1Value >= Rs2Value))
--- a/src/coreclr/debug/ee/riscv64/walker.cpp
+++ b/src/coreclr/debug/ee/riscv64/walker.cpp
@@ -14,6 +14,152 @@
 
 #ifdef TARGET_RISCV64
 
-// #error "TODO-RISCV64: missing implementation"
+inline uint64_t SignExtend(uint64_t value, unsigned int signbit)
+{
+    _ASSERTE(signbit < 64);
+
+    if (signbit == 63)
+      return value;
+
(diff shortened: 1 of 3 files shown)


src/coreclr/debug/ee/riscv64/walker.cpp
```diff
@@ -14,6 +14,152 @@
...
+        m_nextIP = m_ip + imm;
+        // The standard software calling convention uses X1 as the return address register and X5 as an alternate link register.
+        if (Rd == 1 || Rd == 5)
+        {
+            m_skipIP = m_ip + 4;
+            m_type = WALK_CALL;
+        }
+        else
+        {
+            m_skipIP = m_nextIP;
```
clamp03, line 87: m_skipIP is set to NULL when walk is WALK_BRANCH for BL & Conditional Branches in ARM64. As I know, skip is "the function you’re currently debugging will be skipped.". I think NULL is right for m_skipIP in this case. Please check. Thank you.

src/coreclr/debug/ee/riscv64/walker.cpp
```diff
@@ -14,6 +14,152 @@
...
+            (((opcode & 0x707f) == 0x5063) && Rs1SValue >= Rs2SValue))
+        {
+            // B-immediate encodes a signed offset in multiples of 2 bytes
+            //       12      | 11                               1 | 0
+            // inst[31]/sign | inst[7] | inst[30:25] | inst[11:8] | 0
+            uint64_t imm = SignExtend((BitExtract(opcode, 11, 8) << 1) | (BitExtract(opcode, 30, 25) << 5) |
+                                      (BitExtract(opcode, 7, 7) << 11) | (BitExtract(opcode, 31, 31) << 12), 12);
+
+            m_nextIP = m_ip + imm;
+            m_skipIP = m_nextIP;
```
clamp03, line 137: Please check m_skipIP.

src/coreclr/debug/ee/riscv64/walker.cpp
```diff
@@ -14,6 +14,152 @@
...
+            (((opcode & 0x707f) == 0x7063) && Rs1Value >= Rs2Value))
+        {
+            // B-immediate encodes a signed offset in multiples of 2 bytes
+            //       12      | 11                               1 | 0
+            // inst[31]/sign | inst[7] | inst[30:25] | inst[11:8] | 0
+            uint64_t imm = SignExtend((BitExtract(opcode, 11, 8) << 1) | (BitExtract(opcode, 30, 25) << 5) |
+                                      (BitExtract(opcode, 7, 7) << 11) | (BitExtract(opcode, 31, 31) << 12), 12);
+
+            m_nextIP = m_ip + imm;
+            m_skipIP = m_nextIP;
```
clamp03, line 159: Here too.

clamp03 (changes requested):

src/coreclr/debug/ee/riscv64/walker.cpp
viewizard: I have no idea why, but we have different logic here for arm64 (that have walker code aimed to care about 2 steppers implementation - HW and SW single step) and arm32 (that have only SW single step).
I use same logic as arm32, since it more close to riscv (SW single step only):
https://github.com/dotnet/runtime/blob/45978bd7191fcbb793dc12b162540022063551b6/src/coreclr/debug/ee/arm/walker.cpp#L94-L96
https://github.com/dotnet/runtime/blob/45978bd7191fcbb793dc12b162540022063551b6/src/coreclr/debug/ee/arm/walker.cpp#L115-L117
https://github.com/dotnet/runtime/blob/45978bd7191fcbb793dc12b162540022063551b6/src/coreclr/debug/ee/arm/walker.cpp#L132-L134
https://github.com/dotnet/runtime/blob/45978bd7191fcbb793dc12b162540022063551b6/src/coreclr/debug/ee/arm/walker.cpp#L176-L178
https://github.com/dotnet/runtime/blob/45978bd7191fcbb793dc12b162540022063551b6/src/coreclr/debug/ee/arm/walker.cpp#L277-L279

My point is - stay with same logic as arm32, any way caller look at walk type first and only after that care about variables it really need. Probably this is the point why arm32 and arm64 have different logic here with `m_skipIP`.

clamp03: @jkotas As I know m_skipIP is used for handling skip command in debugger. In debugger, it skips steps for selected functions and files. IMO, `m_skipIP` is not needed for WALK_BRANCH in arm32 as well as riscv64. I investigated `src/coreclr/debug/` and it looks like `m_skipIP` is not used for WALK_BRANCH. Could you please look into it? Thank you.

jkotas: @dotnet/dotnet-diag Could you please help to answer this question?

clamp03:
@dotnet/dotnet-diag @tommcdon Could you please review this PR?

clamp03:
@dotnet/dotnet-diag @tommcdon @jkotas Could you check this PR and answer my question?

src/coreclr/vm/riscv64/singlestepper.cpp
```diff
@@ -299,7 +299,7 @@ uint64_t RiscV64SingleStepper::GetReg(T_CONTEXT *pCtx, uint64_t reg)
 {
     _ASSERTE(reg <= 31);
 
-    return (&pCtx->R0)[reg];
+    return reg ? (&pCtx->R0)[reg] : 0; // X0 could be written/corrupted in register context, make sure it return 0.
```
clamp03, line 302: Could you explain more about this? Actually, I cannot understand why X0 could be written/corrupted and why you set hard code 0 rather than fixing corruption. cc @gbalykov

viewizard: Inside runtime we have storred registers context for frames. For example, during emulate stepping we could change some registers with JALR, but I check that register with index 0 must not change context. Looks like we also have other places too, that don't check that dest register X0 and write back some data.

gbalykov:
@viewizard please revert last commit with x0 handling logic

src/coreclr/debug/ee/riscv64/walker.cpp
tommcdon: > I investigated src/coreclr/debug/ and it looks like m_skipIP is not used for WALK_BRANCH

The `WALK_BRANCH` case in `TrapStep` falls through to `WALK_CALL`:
https://github.com/dotnet/runtime/blob/35db500cedca4e7061fe1e91390ad1df62bc0f3c/src/coreclr/debug/ee/controller.cpp#L5901-L5911

Which then will read m_skipIP here:
https://github.com/dotnet/runtime/blob/35db500cedca4e7061fe1e91390ad1df62bc0f3c/src/coreclr/debug/ee/controller.cpp#L5978-L5985

Hope this helps!

src/coreclr/debug/ee/riscv64/walker.cpp
tommcdon: m_skipIP looks reasonable to me

src/coreclr/debug/ee/riscv64/walker.cpp
tommcdon: m_skipIP looks reasonable to me

tommcdon (approved):
I did not do any validation however these changes seem correct to me.  @viewizard's comments look correct to me as well.  Inspecting the walker.cpp code changes it seems that the `WALK_BRANCH` code is setting `m_skipIP`.  For testing, if the netcoredbg debugger is available on this architecture, I suggest setting breakpoints on various conditionals and verifying that the stepper correctly simulates stepping through the patch table.  Complus logs can be used to verify the `WALK_BRANCH` code is actually being executed.

PR #94854, opened 2023-11-16
Fix JsonArray.Add and ReplaceWith regressions.

### LINKED ISSUE #94842: .NET 8.0 JsonArray.Add throws System.InvalidOperationException: "The element cannot be an object or array."
### Description

Adding a `JsonElement` via `JsonArray.Add` throws an exception with .NET 8. In .NET 7 it worked without throwing.

It is interesting that `JsonArray.Add` worked in .NET 7 but `JsonValue.Create` fails in both .NET 7/8 when passing in a `JsonElement` with `ValueKind.Object/Array`.

Related: https://github.com/dotnet/runtime/issues/70427
Regression in https://github.com/dotnet/runtime/commit/1db4357891752cc028620710fe924425235b8f89 ?

### Reproduction Steps

```c#
using System.Text.Json;
using System.Text.Json.Nodes;

var element = JsonDocument.Parse("""{"test":"hello world"}""");

var array = new JsonArray();
array.Add(element.RootElement); // .NET 8.0 throws here

Console.WriteLine(array.ToJsonString());
```

### Expected behavior

Works without exception.

### Actual behavior

Throws a System.InvalidOperationException: "The element cannot be an object or array."

```
   bei System.Text.Json.ThrowHelper.ThrowInvalidOperationException_NodeElementCannotBeObjectOrArray()
   bei System.Text.Json.Nodes.JsonValue.VerifyJsonElementIsNotArrayOrObject(JsonElement& element)
   bei System.Text.Json.Nodes.JsonValue.Create[T](T value, Nullable`1 options)
   bei System.Text.Json.Nodes.JsonArray.Add[T](T value)
```

### Regression?

Yes, it works with .NET 7.0

### Known Workarounds

For this particular case the `JsonDocument` could be added instead of the `RootElement`.
A general workaround would be https://github.com/dotnet/runtime/issues/70427#issuecomment-1150960366

### Configuration

.NET 8
Windows
x64

### Other information

_No response_

eiriktsarpalis:
Fix #94842. See https://github.com/dotnet/runtime/issues/94842#issuecomment-1814750290 for an analysis of the problem and proposed fix. Should be backported to .NET 8.

### DIFF
--- a/src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonArray.cs
+++ b/src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonArray.cs
@@ -175,13 +175,7 @@ namespace System.Text.Json.Nodes
         [RequiresDynamicCode(JsonValue.CreateDynamicCodeMessage)]
         public void Add<T>(T? value)
         {
-            JsonNode? nodeToAdd = value switch
-            {
-                null => null,
-                JsonNode node => node,
-                _ => JsonValue.Create(value, Options)
-            };
-
+            JsonNode? nodeToAdd = ConvertFromValue(value, Options);
             Add(nodeToAdd);
         }
 
--- a/src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonNode.cs
+++ b/src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonNode.cs
@@ -349,6 +349,11 @@ namespace System.Text.Json.Nodes
             Parent = parent;
         }
 
+        /// <summary>
+        /// Adaptation of the equivalent JsonValue.Create factory method extended
+        /// to support arbitrary <see cref="JsonElement"/> and <see cref="JsonNode"/> values.
+        /// TODO consider making public cf. https://github.com/dotnet/runtime/issues/70427
+        /// </summary>
         [RequiresUnreferencedCode(JsonSerializer.SerializationUnreferencedCodeMessage)]
         [RequiresDynamicCode(JsonSerializer.SerializationRequiresDynamicCodeMessage)]
         internal static JsonNode? ConvertFromValue<T>(T? value, JsonNodeOptions? options = null)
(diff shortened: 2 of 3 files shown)


src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonArray.cs
```diff
@@ -175,13 +175,7 @@ internal JsonArray(JsonElement element, JsonNodeOptions? options = null) : base(
...
         {
-            JsonNode? nodeToAdd = value switch
-            {
-                null => null,
-                JsonNode node => node,
-                _ => JsonValue.Create(value, Options)
-            };
-
+            JsonNode? nodeToAdd = ConvertFromValue(value, Options);
             Add(nodeToAdd);
```
tarekgh, line 179: > Add(nodeToAdd); [](http://example.com/codeflow?start=11&length=16)

If we have `nodeToAdd` is `null`. do we still need to call Add(...)?

src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonNode.cs
```diff
@@ -311,17 +312,18 @@ public static bool DeepEquals(JsonNode? node1, JsonNode? node2)
         [RequiresDynamicCode(JsonValue.CreateDynamicCodeMessage)]
         public void ReplaceWith<T>(T value)
         {
+            JsonNode? node;
             switch (_parent)
             {
                 case null:
                     return;
```
tarekgh, line 319: do you need to special case this?

tarekgh (approved):
I added minor questions, otherwise LGTM.

src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonArray.cs
eiriktsarpalis: Yes. `JsonNode` uses `null` to represent `null` JSON values so it needs to be added in that case as well.

src/libraries/System.Text.Json/src/System/Text/Json/Nodes/JsonNode.cs
eiriktsarpalis: Yes, it needs to call different methods depending on whether the parent type is an object or an array.

tarekgh: Sorry, I was not clear. I meant do you need `case null` in the switch?

eiriktsarpalis: Ah yes, that seems redundant.

PR #94861, opened 2023-11-16
Catch exception when subscribing to NetworkChange

### LINKED ISSUE #94794: NetworkInformationException: Permission denied (.NET 8 + Docker +GCloud)
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Identical code (below) that works fine in .NET 7, fails with the following error in .NET 8:

```
Unhandled exception. System.Net.NetworkInformation.NetworkInformationException (13): Permission denied
2023-11-16 02:04:53.524 JST
   at System.Net.NetworkInformation.NetworkChange.CreateSocket()
2023-11-16 02:04:53.524 JST
   at System.Net.NetworkInformation.NetworkChange.add_NetworkAddressChanged(NetworkAddressChangedEventHandler value)
2023-11-16 02:04:53.524 JST
   at System.Net.Http.HttpConnectionPoolManager.StartMonitoringNetworkChanges()
2023-11-16 02:04:53.524 JST
   at System.Net.Http.HttpConnectionPool.HandleAltSvc(IEnumerable`1 altSvcHeaderValues, Nullable`1 responseAge)
2023-11-16 02:04:53.524 JST
   at System.Net.Http.HttpConnectionPool.ProcessAltSvc(HttpResponseMessage response)
2023-11-16 02:04:53.524 JST
   at System.Net.Http.HttpConnectionPool.SendWithVersionDetectionAndRetryAsync(HttpRequestMessage request, Boolean async, Boolean doRequestAuth, CancellationToken cancellationToken)
2023-11-16 02:04:53.524 JST
   at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, Boolean async, CancellationToken cancellationToken)
2023-11-16 02:04:53.524 JST
   at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)
```

Trimmed and adjusted code (same result for ANY https call):
```cs
var httpClient = new HttpClient();
_ = await httpClient.GetAsync("https://www.google.com");
```

The peculiarity of this issue lies in the fact that the code executes fine on a local environment. But the moment the container is spun up in a google cloud run instance, it fails. 

Another weird thing is that raw TCP connections work fine, we have a database connection that goes directly over TCP, rather than traditional HTTP(s). 

I have spent hours debugging this issue, and sadly we have no other choice but to revert back to .NET 7 for the time being. 

### Expected Behavior

A Proper HTTPs connection

### Steps To Reproduce

_No response_

### Exceptions (if any)

NetworkInformationException 

### .NET Version

.NET 8

### Anything else?

_No response_

rzikm:
Fixes #94794.

We can simply swallow the exception and not monitor network changes. RFC allows us to.

RFC: https://tools.ietf.org/html/rfc7838#section-2.2

> When alternative services are used to send a client to the most
optimal server, a change in network configuration can result in
cached values becoming suboptimal.  Therefore, clients SHOULD remove
from cache all alternative services that lack the "persist" flag with
the value "1" **when they detect such a change, when information about
network state is available.**

### DIFF
--- a/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionPoolManager.cs
+++ b/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionPoolManager.cs
@@ -174,9 +174,29 @@ namespace System.Net.Http
                 return;
             }
 
-            using (ExecutionContext.SuppressFlow())
+            // RFC: https://tools.ietf.org/html/rfc7838#section-2.2
+            //    When alternative services are used to send a client to the most
+            //    optimal server, a change in network configuration can result in
+            //    cached values becoming suboptimal.  Therefore, clients SHOULD remove
+            //    from cache all alternative services that lack the "persist" flag with
+            //    the value "1" when they detect such a change, when information about
+            //    network state is available.
+            try
             {
-                NetworkChange.NetworkAddressChanged += networkChangedDelegate;
+                using (ExecutionContext.SuppressFlow())
+                {
+                    NetworkChange.NetworkAddressChanged += networkChangedDelegate;
+                }
+            }
+            catch (Exception e)
+            {
+                if (NetEventSource.Log.IsEnabled()) NetEventSource.Error(this, $"Exception when subscribing to NetworkChange.NetworkAddressChanged: {e}");
+
+                // We can't monitor network changes, so technically "information
+                // about network state is not available" and we can just keep
+                // all Alt-Svc entries until their expiration time.
+                //
+                // keep the _networkChangeCleanup field null so we don't try again needlessly
             }
         }
 
--- a/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionPoolManager.cs
+++ b/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionPoolManager.cs
@@ -188,7 +188,7 @@ namespace System.Net.Http
                     NetworkChange.NetworkAddressChanged += networkChangedDelegate;
                 }
             }
-            catch (Exception e)
+            catch (NetworkInformationException e)
             {
                 if (NetEventSource.Log.IsEnabled()) NetEventSource.Error(this, $"Exception when subscribing to NetworkChange.NetworkAddressChanged: {e}");
 
@@ -196,7 +196,7 @@ namespace System.Net.Http
                 // about network state is not available" and we can just keep
                 // all Alt-Svc entries until their expiration time.
                 //
-                // keep the _networkChangeCleanup field null so we don't try again needlessly
+                // keep the _networkChangeCleanup field assigned so we don't try again needlessly
             }
         }
 


src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionPoolManager.cs
```diff
@@ -174,9 +174,29 @@ public void StartMonitoringNetworkChanges()
...
+            catch (Exception e)
             {
-                NetworkChange.NetworkAddressChanged += networkChangedDelegate;
+                if (NetEventSource.Log.IsEnabled()) NetEventSource.Error(this, $"Exception when subscribing to NetworkChange.NetworkAddressChanged: {e}");
+
+                // We can't monitor network changes, so technically "information
+                // about network state is not available" and we can just keep
+                // all Alt-Svc entries until their expiration time.
+                //
+                // keep the _networkChangeCleanup field null so we don't try again needlessly
```
wfurt, line 199: Can the cleanup be unhappy if we fail to add it?

ManickaP (approved):
LGTM, thanks.

If we're gonna backport it, we should test it (including the cleanup code). Even if just manually with faking the exception for the sake of the test.

src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionPoolManager.cs
CarnaViire: Yeah, I think we might need to prevent both disposal and finalization of _networkChangeCleanup...

src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionPoolManager.cs
```diff
@@ -174,9 +174,29 @@ public void StartMonitoringNetworkChanges()
...
+            //    the value "1" when they detect such a change, when information about
+            //    network state is available.
+            try
+            {
+                using (ExecutionContext.SuppressFlow())
+                {
+                    NetworkChange.NetworkAddressChanged += networkChangedDelegate;
+                }
+            }
+            catch (Exception e)
```
danmoseley, line 191: Is there a limited set of exceptions we expect here? If so if not catching just those would it make sense to add an assert that it is only those types, to alert us in future to any unexpected?

src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/HttpConnectionPoolManager.cs
rzikm: I checked only the Unix code, but the unregister call should be safe (it does not try to allocate a new socket, nor does it rely on the handler being previously registered in the first place)

rzikm: I checked other platform implementations and all are robust against unregistering callbacks which was not registered before.

rzikm:
CI failures are unrelated

PR #94869, opened 2023-11-16
Update HttpClientBuilderExtensions.cs

IEvangelist:
This API had misleading information. The predicate doesn't let you evaluate the header's _value_, but instead the header _name_, which is then used to determine whether to redact the header value.


MihaZupan (approved):
Thank you

PR #94241, opened 2023-10-31
Fix class construction recursion issue in `Lock` on NativeAOT

kouvel:
- The `Monitor` type was being constructed due to the use of `Monitor.DebugBlockingScope`, added that to the initialization phase so that the relevant type is initialized before it's used on the wait path
- A better alternative may be to move `DebugBlockingScope` to be under `Lock`, it would probably make more sense for `Monitor` to refer to `Lock` as it already does, than the inverse. Based on the comments the thread-static field is apparently bound-to in debugging scenarios currently. Something to look into, for now this should unblock the CI.
- Fixes https://github.com/dotnet/runtime/issues/94227

PR #94873, opened 2023-11-16
[NativeAOT] Fix Lock's usage of NativeRuntimeEventSource.Log to account for recursive accesses during its own class construction

### LINKED ISSUE #94728: Test failure - System.NullReferenceException in System.Threading.Lock.TryInitializeStatics
```
Running assembly:System.Net.HttpListener.Tests, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51
Process terminated. Exception thrown from SocketAsyncEngine event loop: System.TypeInitializationException: A type initializer threw an exception. To determine which type, inspect the InnerException's StackTrace property.
 ---> System.TypeInitializationException: A type initializer threw an exception. To determine which type, inspect the InnerException's StackTrace property.
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Threading.Lock.TryInitializeStatics() + 0xf8
   at System.Threading.Lock.TryEnterSlow(Int32, Lock.ThreadId) + 0x16c
   at System.Threading.Lock.EnterAndGetCurrentThreadId() + 0x68
   at System.Runtime.CompilerServices.ClassConstructorRunner.Cctor.GetCctor(StaticClassConstructionContext*) + 0x38
   at System.Runtime.CompilerServices.ClassConstructorRunner.EnsureClassConstructorRun(StaticClassConstructionContext*) + 0x38
   at System.Runtime.CompilerServices.ClassConstructorRunner.CheckStaticClassConstructionReturnGCStaticBase(StaticClassConstructionContext*, Object) + 0x14
   at System.Diagnostics.Tracing.NativeRuntimeEventSource..cctor() + 0x58
   at System.Runtime.CompilerServices.ClassConstructorRunner.EnsureClassConstructorRun(StaticClassConstructionContext*) + 0xc8
   --- End of inner exception stack trace ---
   at System.Runtime.CompilerServices.ClassConstructorRunner.EnsureClassConstructorRun(StaticClassConstructionContext*) + 0x15c
   at System.Runtime.CompilerServices.ClassConstructorRunner.CheckStaticClassConstructionReturnGCStaticBase(StaticClassConstructionContext*, Object) + 0x14
   at System.Threading.PortableThreadPool..ctor() + 0x1b0
   at System.Threading.PortableThreadPool..cctor() + 0x68
   at System.Runtime.CompilerServices.ClassConstructorRunner.EnsureClassConstructorRun(StaticClassConstructionContext*) + 0xc8
   --- End of inner exception stack trace ---
   at System.Runtime.CompilerServices.ClassConstructorRunner.EnsureClassConstructorRun(StaticClassConstructionContext*) + 0x15c
   at System.Runtime.CompilerServices.ClassConstructorRunner.CheckStaticClassConstructionReturnGCStaticBase(StaticClassConstructionContext*, Object) + 0x14
   at System.Threading.ThreadPoolWorkQueue.Enqueue(Object, Boolean) + 0x140
   at System.Threading.ThreadPool.UnsafeQueueUserWorkItem(IThreadPoolWorkItem, Boolean) + 0x60
   at System.Net.Sockets.SocketAsyncEngine.EventLoop() + 0xc4
   at System.RuntimeExceptionHelpers.FailFast(String, Exception, String, RhFailFastReason, IntPtr, IntPtr) + 0x180
   at System.Net.Sockets.SocketAsyncEngine.EventLoop() + 0x160
   at System.Runtime.EH.DispatchEx(StackFrameIterator&, EH.ExInfo&) + 0x204
   at System.Runtime.EH.RhThrowEx(Object, EH.ExInfo&) + 0x44
   at System.Threading.Thread.StartThread(IntPtr) + 0x154
```

## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=468884
Build error leg or test failing: System.Net.HttpListener.Tests.WorkItemExecution
Pull request: https://github.com/dotnet/runtime/pull/94692

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json
{
  "ErrorMessage": "System.Threading.Lock.TryInitializeStatics",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=468884
**Error message validated:** `System.Threading.Lock.TryInitializeStatics`
**Result validation: :white_check_mark:** Known issue matched with the provided build.
**Validation performed at:** 11/14/2023 6:27:46 PM UTC

### Report

|---|---|---|---|
|[531213](https://dev.azure.com/dnceng-public/public/_build/results?buildId=531213)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=531213&view=ms.vss-test-web.build-test-results-tab&runId=12528970&resultId=100028)|dotnet/runtime#97062|
|[531121](https://dev.azure.com/dnceng-public/public/_build/results?buildId=531121)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=531121&view=ms.vss-test-web.build-test-results-tab&runId=12526014&resultId=100028)|dotnet/runtime#97075|
|[530911](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530911)|dotnet/runtime|[Microsoft.Extensions.FileProviders.Physical.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530911&view=ms.vss-test-web.build-test-results-tab&runId=12524066&resultId=179577)||
|[530962](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530962)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530962&view=ms.vss-test-web.build-test-results-tab&runId=12523172&resultId=100028)|dotnet/runtime#97114|
|[530971](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530971)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530971&view=ms.vss-test-web.build-test-results-tab&runId=12523246&resultId=100022)|dotnet/runtime#97080|
|[530810](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530810)|dotnet/runtime|[System.Formats.Tar.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530810&view=ms.vss-test-web.build-test-results-tab&runId=12522876&resultId=178988)|dotnet/runtime#97062|
|[530819](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530819)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530819&view=ms.vss-test-web.build-test-results-tab&runId=12518690&resultId=100028)|dotnet/runtime#97104|
|[530811](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530811)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530811&view=ms.vss-test-web.build-test-results-tab&runId=12518518&resultId=100028)|dotnet/runtime#97052|
|[530826](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530826)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530826&view=ms.vss-test-web.build-test-results-tab&runId=12518390&resultId=100028)||
|[530592](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530592)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530592&view=ms.vss-test-web.build-test-results-tab&runId=12512346&resultId=100028)|dotnet/runtime#96841|
|[530528](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530528)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530528&view=ms.vss-test-web.build-test-results-tab&runId=12510350&resultId=100028)|dotnet/runtime#90505|
|[530404](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530404)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530404&view=ms.vss-test-web.build-test-results-tab&runId=12506726&resultId=100028)|dotnet/runtime#96386|
|[530373](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530373)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530373&view=ms.vss-test-web.build-test-results-tab&runId=12505780&resultId=100028)|dotnet/runtime#97087|
|[530327](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530327)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530327&view=ms.vss-test-web.build-test-results-tab&runId=12504236&resultId=100028)|dotnet/runtime#96354|
|[529245](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529245)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529245&view=ms.vss-test-web.build-test-results-tab&runId=12502826&resultId=100028)|dotnet/runtime#90505|
|[530151](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530151)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530151&view=ms.vss-test-web.build-test-results-tab&runId=12499538&resultId=100028)|dotnet/runtime#97096|
|[530144](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530144)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530144&view=ms.vss-test-web.build-test-results-tab&runId=12499208&resultId=100028)|dotnet/runtime#96974|
|[2355601](https://dev.azure.com/dnceng/internal/_build/results?buildId=2355601)|dotnet-runtime|[Device Startup - iOS NativeAOT HelloWorld nollvm nosymbols HybridGlobalization.WorkItemExecution](https://dev.azure.com/dnceng//internal/_build/results?buildId=2355601&view=ms.vss-test-web.build-test-results-tab&runId=52889653&resultId=100005)||
|[530041](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530041)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=530041&view=ms.vss-test-web.build-test-results-tab&runId=12496480&resultId=100028)|dotnet/runtime#97087|
|[529935](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529935)|dotnet/runtime|[System.IO.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529935&view=ms.vss-test-web.build-test-results-tab&runId=12495838&resultId=172995)||
|[529924](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529924)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529924&view=ms.vss-test-web.build-test-results-tab&runId=12493712&resultId=100028)|dotnet/runtime#97079|
|[529691](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529691)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529691&view=ms.vss-test-web.build-test-results-tab&runId=12487154&resultId=100028)|dotnet/runtime#97068|
|[529669](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529669)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529669&view=ms.vss-test-web.build-test-results-tab&runId=12486094&resultId=100028)|dotnet/runtime#97058|
|[529552](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529552)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529552&view=ms.vss-test-web.build-test-results-tab&runId=12483072&resultId=100028)|dotnet/runtime#95980|
|[529539](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529539)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529539&view=ms.vss-test-web.build-test-results-tab&runId=12482698&resultId=100022)|dotnet/runtime#96995|
|[529444](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529444)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529444&view=ms.vss-test-web.build-test-results-tab&runId=12480746&resultId=100028)|dotnet/runtime#97058|
|[529413](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529413)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529413&view=ms.vss-test-web.build-test-results-tab&runId=12480402&resultId=100021)|dotnet/runtime#97055|
|[529298](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529298)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529298&view=ms.vss-test-web.build-test-results-tab&runId=12475944&resultId=100021)|dotnet/runtime#96890|
|[529173](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529173)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=529173&view=ms.vss-test-web.build-test-results-tab&runId=12472068&resultId=100028)|dotnet/runtime#96386|
|[528933](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528933)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528933&view=ms.vss-test-web.build-test-results-tab&runId=12464568&resultId=100028)|dotnet/runtime#94764|
|[528672](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528672)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528672&view=ms.vss-test-web.build-test-results-tab&runId=12456832&resultId=100028)|dotnet/runtime#96896|
|[528657](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528657)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528657&view=ms.vss-test-web.build-test-results-tab&runId=12456104&resultId=100028)|dotnet/runtime#96841|
|[528612](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528612)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528612&view=ms.vss-test-web.build-test-results-tab&runId=12454952&resultId=100028)|dotnet/runtime#96678|
|[528493](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528493)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528493&view=ms.vss-test-web.build-test-results-tab&runId=12451210&resultId=100028)|dotnet/runtime#96974|
|[528295](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528295)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528295&view=ms.vss-test-web.build-test-results-tab&runId=12447706&resultId=113602)||
|[528343](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528343)|dotnet/runtime|[System.IO.FileSystem.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528343&view=ms.vss-test-web.build-test-results-tab&runId=12446692&resultId=123863)|dotnet/runtime#96888|
|[528270](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528270)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528270&view=ms.vss-test-web.build-test-results-tab&runId=12444522&resultId=100019)|dotnet/runtime#94764|
|[528225](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528225)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528225&view=ms.vss-test-web.build-test-results-tab&runId=12443066&resultId=100026)|dotnet/runtime#97017|
|[528183](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528183)|dotnet/runtime|[System.IO.FileSystem.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528183&view=ms.vss-test-web.build-test-results-tab&runId=12442820&resultId=123863)|dotnet/runtime#96805|
|[528007](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528007)|dotnet/runtime|[JIT.CodeGenBringUpTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528007&view=ms.vss-test-web.build-test-results-tab&runId=12438620&resultId=113582)||
|[528035](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528035)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528035&view=ms.vss-test-web.build-test-results-tab&runId=12438678&resultId=100026)|dotnet/runtime#96982|
|[528027](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528027)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=528027&view=ms.vss-test-web.build-test-results-tab&runId=12438330&resultId=100026)|dotnet/runtime#96983|
|[527919](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527919)|dotnet/runtime|[JIT.CodeGenBringUpTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527919&view=ms.vss-test-web.build-test-results-tab&runId=12436934&resultId=113582)|dotnet/runtime#96982|
|[527994](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527994)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527994&view=ms.vss-test-web.build-test-results-tab&runId=12436668&resultId=100026)|dotnet/runtime#97005|
|[527889](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527889)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527889&view=ms.vss-test-web.build-test-results-tab&runId=12433234&resultId=100026)|dotnet/runtime#94764|
|[527849](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527849)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527849&view=ms.vss-test-web.build-test-results-tab&runId=12432396&resultId=100026)|dotnet/runtime#96995|
|[527816](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527816)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527816&view=ms.vss-test-web.build-test-results-tab&runId=12431408&resultId=100026)|dotnet/runtime#96201|
|[527706](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527706)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527706&view=ms.vss-test-web.build-test-results-tab&runId=12429120&resultId=100026)|dotnet/runtime#96992|
|[527644](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527644)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527644&view=ms.vss-test-web.build-test-results-tab&runId=12427360&resultId=100026)|dotnet/runtime#96991|
|[527576](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527576)|dotnet/runtime|[nativeaot/SmokeTests/Reflection/Reflection_FromUsage/Reflection_FromUsage.sh](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527576&view=ms.vss-test-web.build-test-results-tab&runId=12427794&resultId=100020)|dotnet/runtime#96678|
|[527456](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527456)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527456&view=ms.vss-test-web.build-test-results-tab&runId=12423902&resultId=100019)|dotnet/runtime#96676|
|[527428](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527428)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527428&view=ms.vss-test-web.build-test-results-tab&runId=12423106&resultId=100024)|dotnet/runtime#96985|
|[527361](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527361)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527361&view=ms.vss-test-web.build-test-results-tab&runId=12419276&resultId=100026)|dotnet/runtime#96969|
|[527355](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527355)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527355&view=ms.vss-test-web.build-test-results-tab&runId=12418806&resultId=100026)|dotnet/runtime#96983|
|[527210](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527210)|dotnet/runtime|[System.Web.HttpUtility.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527210&view=ms.vss-test-web.build-test-results-tab&runId=12417934&resultId=179603)||
|[527253](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527253)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527253&view=ms.vss-test-web.build-test-results-tab&runId=12416398&resultId=100026)|dotnet/runtime#96741|
|[527177](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527177)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527177&view=ms.vss-test-web.build-test-results-tab&runId=12414758&resultId=100026)|dotnet/runtime#96969|
|[527171](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527171)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527171&view=ms.vss-test-web.build-test-results-tab&runId=12414446&resultId=100028)|dotnet/runtime#96894|
|[527165](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527165)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=527165&view=ms.vss-test-web.build-test-results-tab&runId=12413902&resultId=100026)||
|[526886](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526886)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526886&view=ms.vss-test-web.build-test-results-tab&runId=12406328&resultId=113602)||
|[526876](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526876)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526876&view=ms.vss-test-web.build-test-results-tab&runId=12405670&resultId=100026)||
|[526833](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526833)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526833&view=ms.vss-test-web.build-test-results-tab&runId=12404664&resultId=100026)|dotnet/runtime#96927|
|[526822](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526822)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526822&view=ms.vss-test-web.build-test-results-tab&runId=12404376&resultId=100024)|dotnet/runtime#96960|
|[526766](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526766)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526766&view=ms.vss-test-web.build-test-results-tab&runId=12402672&resultId=100020)|dotnet/runtime#96896|
|[526623](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526623)|dotnet/runtime|[System.Threading.Tasks.Dataflow.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526623&view=ms.vss-test-web.build-test-results-tab&runId=12399454&resultId=179314)||
|[526521](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526521)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526521&view=ms.vss-test-web.build-test-results-tab&runId=12395520&resultId=100026)|dotnet/runtime#96926|
|[526484](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526484)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526484&view=ms.vss-test-web.build-test-results-tab&runId=12395524&resultId=100026)|dotnet/runtime#96794|
|[526474](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526474)|dotnet/runtime|[Microsoft.Extensions.Options.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526474&view=ms.vss-test-web.build-test-results-tab&runId=12395272&resultId=179550)||
|[526396](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526396)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526396&view=ms.vss-test-web.build-test-results-tab&runId=12390960&resultId=100026)|dotnet/runtime#96890|
|[526349](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526349)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526349&view=ms.vss-test-web.build-test-results-tab&runId=12390064&resultId=100026)|dotnet/runtime#96896|
|[526289](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526289)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526289&view=ms.vss-test-web.build-test-results-tab&runId=12387810&resultId=100026)|dotnet/runtime#96676|
|[526254](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526254)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526254&view=ms.vss-test-web.build-test-results-tab&runId=12386938&resultId=100026)|dotnet/runtime#95904|
|[526165](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526165)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526165&view=ms.vss-test-web.build-test-results-tab&runId=12383876&resultId=100026)|dotnet/runtime#96936|
|[526144](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526144)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526144&view=ms.vss-test-web.build-test-results-tab&runId=12383156&resultId=100026)|dotnet/runtime#96933|
|[526132](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526132)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=526132&view=ms.vss-test-web.build-test-results-tab&runId=12382908&resultId=100024)|dotnet/runtime#96931|
|[525957](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525957)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525957&view=ms.vss-test-web.build-test-results-tab&runId=12379384&resultId=100026)|dotnet/runtime#96927|
|[525923](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525923)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525923&view=ms.vss-test-web.build-test-results-tab&runId=12378856&resultId=100026)|dotnet/runtime#96926|
|[525906](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525906)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525906&view=ms.vss-test-web.build-test-results-tab&runId=12378432&resultId=100020)|dotnet/runtime#96877|
|[525842](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525842)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525842&view=ms.vss-test-web.build-test-results-tab&runId=12376492&resultId=100026)|dotnet/runtime#96924|
|[525667](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525667)|dotnet/runtime|[JIT.CodeGenBringUpTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525667&view=ms.vss-test-web.build-test-results-tab&runId=12375226&resultId=113582)||
|[525785](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525785)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525785&view=ms.vss-test-web.build-test-results-tab&runId=12374702&resultId=100020)|dotnet/runtime#96913|
|[525765](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525765)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525765&view=ms.vss-test-web.build-test-results-tab&runId=12374358&resultId=100020)|dotnet/runtime#96921|
|[525420](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525420)|dotnet/runtime|[Microsoft.Extensions.Configuration.Xml.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525420&view=ms.vss-test-web.build-test-results-tab&runId=12364836&resultId=179597)|dotnet/runtime#96906|
|[525477](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525477)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525477&view=ms.vss-test-web.build-test-results-tab&runId=12362268&resultId=100026)|dotnet/runtime#96805|
|[525402](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525402)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525402&view=ms.vss-test-web.build-test-results-tab&runId=12360260&resultId=100026)|dotnet/runtime#96902|
|[525362](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525362)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525362&view=ms.vss-test-web.build-test-results-tab&runId=12359240&resultId=100024)|dotnet/runtime#96880|
|[525344](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525344)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525344&view=ms.vss-test-web.build-test-results-tab&runId=12358820&resultId=100026)|dotnet/runtime#96741|
|[525261](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525261)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525261&view=ms.vss-test-web.build-test-results-tab&runId=12357284&resultId=100020)|dotnet/runtime#96678|
|[525140](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525140)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525140&view=ms.vss-test-web.build-test-results-tab&runId=12354682&resultId=100026)|dotnet/runtime#96676|
|[525105](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525105)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525105&view=ms.vss-test-web.build-test-results-tab&runId=12353660&resultId=100026)|dotnet/runtime#95904|
|[525086](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525086)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=525086&view=ms.vss-test-web.build-test-results-tab&runId=12353432&resultId=100026)|dotnet/runtime#96893|
|[524959](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524959)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524959&view=ms.vss-test-web.build-test-results-tab&runId=12351540&resultId=100020)|dotnet/runtime#96558|
|[524821](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524821)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524821&view=ms.vss-test-web.build-test-results-tab&runId=12350160&resultId=113602)||
|[524798](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524798)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524798&view=ms.vss-test-web.build-test-results-tab&runId=12347544&resultId=100026)||
|[524790](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524790)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524790&view=ms.vss-test-web.build-test-results-tab&runId=12347066&resultId=100026)|dotnet/runtime#95583|
|[524255](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524255)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524255&view=ms.vss-test-web.build-test-results-tab&runId=12338876&resultId=100026)|dotnet/runtime#96829|
|[524273](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524273)|dotnet/runtime|[System.Net.Primitives.Pal.Tests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524273&view=ms.vss-test-web.build-test-results-tab&runId=12336536&resultId=179601)||
|[524311](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524311)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524311&view=ms.vss-test-web.build-test-results-tab&runId=12332440&resultId=100026)|dotnet/runtime#96870|
|[524291](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524291)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524291&view=ms.vss-test-web.build-test-results-tab&runId=12332040&resultId=100020)|dotnet/runtime#96867|
|[524168](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524168)|dotnet/runtime|[nativeaot.SmokeTests.WorkItemExecution](https://dev.azure.com/dnceng-public/public/_build/results?buildId=524168&view=ms.vss-test-web.build-test-results-tab&runId=12327612&resultId=100026)|dotnet/runtime#96805|
Displaying 100 of 215 results
#### Summary
|---|---|---|
|23|112|215|

kouvel:
`NativeRuntimeEventSource.Log` can be null if it's being constructed in the same thread earlier in the stack. Added null checks.

Fixes https://github.com/dotnet/runtime/issues/94728

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Lock.NativeAot.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Lock.NativeAot.cs
@@ -173,7 +173,7 @@ namespace System.Threading
                 s_minSpinCount = DetermineMinSpinCount();
 
                 // Also initialize some types that are used later to prevent potential class construction cycles
-                NativeRuntimeEventSource.Log.IsEnabled();
+                _ = NativeRuntimeEventSource.Log;
             }
             catch
             {
--- a/src/libraries/System.Private.CoreLib/src/System/Threading/Lock.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Threading/Lock.cs
@@ -444,9 +444,9 @@ namespace System.Threading
 
         Wait:
             bool areContentionEventsEnabled =
-                NativeRuntimeEventSource.Log.IsEnabled(
+                NativeRuntimeEventSource.Log?.IsEnabled(
                     EventLevel.Informational,
-                    NativeRuntimeEventSource.Keywords.ContentionKeyword);
+                    NativeRuntimeEventSource.Keywords.ContentionKeyword) ?? false;
             AutoResetEvent waitEvent = _waitEvent ?? CreateWaitEvent(areContentionEventsEnabled);
             if (State.TryLockBeforeWait(this))
             {
@@ -463,7 +463,7 @@ namespace System.Threading
                 long waitStartTimeTicks = 0;
                 if (areContentionEventsEnabled)
                 {
-                    NativeRuntimeEventSource.Log.ContentionStart(this);
+                    NativeRuntimeEventSource.Log!.ContentionStart(this);
                     waitStartTimeTicks = Stopwatch.GetTimestamp();
                 }
 
@@ -535,7 +535,7 @@ namespace System.Threading
                     {
                         double waitDurationNs =
                             (Stopwatch.GetTimestamp() - waitStartTimeTicks) * 1_000_000_000.0 / Stopwatch.Frequency;
-                        NativeRuntimeEventSource.Log.ContentionStop(waitDurationNs);
+                        NativeRuntimeEventSource.Log!.ContentionStop(waitDurationNs);
                     }
 
                     return currentThreadId;
--- a/src/libraries/System.Private.CoreLib/src/System/Threading/WaitHandle.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Threading/WaitHandle.cs
@@ -132,6 +132,12 @@ namespace System.Threading
 #if !CORECLR // CoreCLR sends the wait events from the native side
                     bool sendWaitEvents =
                         millisecondsTimeout != 0 &&
+#if NATIVEAOT
+                        // A null check is necessary in NativeAOT due to the possibility of reentrance during class
+                        // construction, as this path can be reached through Lock. See
+                        // https://github.com/dotnet/runtime/issues/94728 for a call stack.
+                        NativeRuntimeEventSource.Log != null &&
+#endif
                         NativeRuntimeEventSource.Log.IsEnabled(
                             EventLevel.Verbose,
                             NativeRuntimeEventSource.Keywords.WaitHandleKeyword);


MichalStrehovsky:
With this and #94241, what is our confidence that `Lock` can still safely be used within the implementation of the native AOT runtime? Do we need to consider migrating these low level usages to `LowLevelLock`?

In #94515 I used `LowLevelLock` not because I hit some issue, but I just couldn't reason if it's safe to use `Lock` in the implementation of the C# `typeof` (or `object.GetType`).

VSadov:
> With this and https://github.com/dotnet/runtime/pull/94241, what is our confidence that Lock can still safely be used within the implementation of the native AOT runtime? Do we need to consider migrating these low level usages to LowLevelLock?

When we were stabilizing the old Lock the same question was asked. I think the same reasoning still applies - 
NativeAOT runtime is written in C#, uses off-the shelf components and calls everchanging set of various things. It would be impractical to require that nothing of that uses locks or must use a special kind of lock. On the other hand, the Lock is a very simple mechanism, with a very simple purpose and calls a very limited set of APIs. It is definitely possible to ensure that those APIs do not use Lock and thus guarantee that there is no reentrancy. 

Old lock did not have this problem, ... eventually. The implementation in #93879 does not seem to be affected. This is a fixable issue.

> In https://github.com/dotnet/runtime/pull/94515 I used LowLevelLock not because I hit some issue, but I just couldn't reason if it's safe to use Lock in the implementation of the C# typeof (or object.GetType).

The only legit use of LowLevelLock in NativeAOT is in implementation of Wait subsystem. Since Lock uses that, we can't use Lock there.
In most other places, including `typeof`, it should be fine to use Lock.

VSadov:
@kouvel - is there a good understanding of the repro? It worries me a bit that the bug reproduces exclusively on linux-arm64. That was also the case with the previous bug.
Is linux-arm64 the only configuration that enables logging, by any chance?  

Relying on linux-arm64 for stress might not be a good place to be. If there was a way to force this to repro on other platforms, especially on win-x64, it would be very valuable. I wonder what makes linux-arm64 special here.

MichalStrehovsky:
> The only legit use of LowLevelLock in NativeAOT is in implementation of Wait subsystem. Since Lock uses that, we can't use Lock there.
> In most other places, including typeof, it should be fine to use Lock.

Thanks, I sent out a PR to use that instead: #94991.

> Is linux-arm64 the only configuration that enables logging, by any chance?

It shouldn't be linux-arm64 specific, but we definitely enable EventSource for fewer tests. EventSource is disabled by default in the shipping configuration (some workloads like ASP.NET will enable it back). The tests mirror that. Only tests that test event source enable it back (search for `<EventSourceSupport>true</EventSourceSupport>` to see the tests that do).

kouvel:
> @kouvel - is there a good understanding of the repro? It worries me a bit that the bug reproduces exclusively on linux-arm64. That was also the case with the previous bug.

It's a timing issue that can lead to the stack trace in https://github.com/dotnet/runtime/issues/94728 or something like it. It's not specific to a platform or architecture, probably not easy to repro without inducing timings.

> With this and https://github.com/dotnet/runtime/pull/94241, what is our confidence that `Lock` can still safely be used within the implementation of the native AOT runtime?

There appears to be a circular dependency where Lock does some lazy initialization that involves class construction, which in turn uses Lock for synchronization. That makes the system a bit brittle to changes. For instance, new issues were uncovered when raising events through `NativeRuntimeEventSource`. Maybe one way to break that circular dependency would be for runtime uses of Lock to pass in a boolean that bypasses lazy initialization and uses defaults, along with not raising events. I'll see if we see more of this type of issue or cases that are not easily fixable from the Lock side, maybe something like that can be done.

VSadov:
`RuntimeEventSource` is eagerly initialized on module load because it has `[ModuleInitializer]`. That indirectly initializes `NativeRuntimeEventSource`. By the time we start running some meaningful code, `NativeRuntimeEventSource` should be initialized.  

Somehow on arm64 we start seeing contentions from sockets before module initialization is complete. Not sure if that was expected by whoever have put `[ModuleInitializer]` there.

VSadov:
I think the most reliable fix for this would be to stop calling static initialization from the Lock entirely, and make something else do the initialization instead.  
For that we need: 
1. make static initialization optional for the Lock functionality. It may result in losing a few events when timing is tight, but it does not seem to be a big deal or avoidable. 
(currently suggested change is half way there already - it allows losing events while the Log is not ready)  
1. make something else to perform initialization of statics. There are a few options here:
-- Use [ModuleInitializer] 
-- Use finalizer of some dummy object 
like in https://github.com/dotnet/runtime/blob/1c4bd555ecca1ed49cd7a259f23c35e6d83665ee/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Threading/Lock.NativeAot.cs#L130-L146

In either case there is a chance to have a small window of operating with uninitialized statics, but that is ok as long as the initialization is optional.  

The advantage of [ModuleInitializer] is being more explicit and upfront, but there could be concern that module init would run at program startup and the NativeRuntimeEventSource initialization could be a bit heavy (there appears to be some reflection use, type loading, attribute parsing,..). 
I am not sure about the cost of reading/parsing the config settings.

The Finalizer approach allows to wait until we are sure that we have a program that may have contentions to report.

MichalStrehovsky:
> Somehow on arm64 we start seeing contentions from sockets before module initialization is complete. Not sure if that was expected by whoever have put [ModuleInitializer] there.

Maybe an ordering issue with this: https://github.com/dotnet/runtime/blob/784537a9cdc97bc5f45f6606f3f9837a7f755236/src/libraries/Common/tests/System/Net/QuicLoad.cs#L12-L20

The code looks to be Linux-only so that would explain we don't see it on Windows. Cc @LakshanF. IIRC RuntimeEventSource initialization it was made a ModuleInitializer because running it as part of CoreLib initialization was too early and we don't have any other initialization phase before Main. Maybe we need one.

jkotas:
We do topological sort of ModuleInitializers. CoreLib module initializer should run first and the Quic module initializer should run second. Both CoreLib module initializer and the Quic module initializer start new threads. The problem seems to be triggered when the background work done by the Quic module initializer happens faster than the background work done by the Quic module initializer.

I think the root cause of the problem is use of managed `Lock` by the native AOT ClassConstructorRunner. The equivalent code in regular CoreCLR uses low-level lock and it is why it avoids the dependency cycles. Should we fix this issue by switching ClassConstructorRunner to use the low-level lock?

kouvel:
> Second related problem is that the managed locks use high-level wait that can be overridden using synchronization context.

Given that it wouldn't be feasible to use `Lock` in low-level cases, `Lock` could be simplified a bit by removing the lazy initialization and maybe other things.

Should I make the relevant changes to ClassConstructorRunner also?

jkotas:
> Should I make the relevant changes to ClassConstructorRunner 

I think so. And delete/simplify code that has been working around the Lock use in ClassConstructorRunner. Thank you!

VSadov:
> We do topological sort of ModuleInitializers. CoreLib module initializer should run first and the Quic module initializer should run second. Both CoreLib module initializer and the Quic module initializer start new threads. The problem seems to be triggered when the background work done by the Quic module initializer happens faster than the background work done by the Quic module initializer.

Is that a problem or something expected?
If they both start threads and complete in random order, why bother ordering them? Is it the order of their start or completion that matters?  
(I do not have a lot of opinion on this, but it kind of still feel suspicious that it affects only linux-arm64)

jkotas:
> >  The problem seems to be triggered when the background work done by the Quic module initializer happens faster than the background work done by the Quic module initializer.

> Is that a problem or something expected?

It is expected. 

The problem is that two threads running in parallel cause contention in constructor runner that leads to the crash.

VSadov:
> Second related problem is that the managed locks use high-level wait that can be overridden using synchronization context. Try running this:

It is a bit surprising why we decided to do this (perhaps something with COM/STA as usual), and keep maintaining. In normal use Lock has 2 levels of optimizations to avoid Waiting, and `lock{}` has 3 levels. Outside of deadlocks the feature use would be very nondeterministic. 
It might be late to argue that this feature is as useful as `Thread.Interrupt` and would not be very missed in the same way.

jkotas:
> How far will this go?

Yes, it needs to stop somewhere. The split between ordinary locks and managed locks is very arbitrary today. My point is that the best way to fix reentrancy issues specific to native AOT is by using same type of lock between runtimes in the current state of the world.

> I think it is doable and is a lot cheaper to just keep lock{}/Lock compatible with the runtime except for a set of functionality that Lock itself uses (and that set can be very small).

I would be open to the idea of making the new Lock use primitive waits. It would require investigation of what SynchronizationContext overrides are used for today and whether they would get broken by it.

VSadov:
> I would be open to the idea of making the new Lock use primitive waits. It would require investigation of what SynchronizationContext overrides are used for today and whether they would get broken by it.

Right. I was going to literally suggest "I'd go as far as use uninterruptible Wait in Lock", but decided to search for anything that could be possibly broken. (other than tests).

So far I found nothing interesting. SynchronizationContext is used, but mostly for Post/Send/Complete purposes. 

Noone seems to be overriding Wait in any nontrivial way. I found only 2 cases both just forward to nonalertable WaitForMultipleObjectsEx, ironically - to suppress the message pumping behavior of standard wait.

VSadov:
> Yes, it needs to stop somewhere. The split between ordinary locks and managed locks is very arbitrary today. My point is that the best way to fix reentrancy issues specific to native AOT is by using same type of lock between runtimes in the current state of the world.

I think the root cause for reentrancy is that Lock initializes EventSource as a part of its regular operations and that runs arbitrary code and that causes stress regressions. If that initialization happens on a side - in module initializer or in a finalizer, the problem will disappear. 

I do not think that CoreCLR is a good guidance on what lock to use. The reasons in CoreCLR could be different, often just because of different native/managed split or some appdomain/sql stuff that may not even be applicable anymore. 
In NativeAOT most locks can be managed. I did not see a convincing reason other than the issue above, and that has a simpler solution than making specialized locks.

The static initialization is actually a good example. 

LowLevelLock lacks functionality for that. It also has a fairly poor performance. That was never an issue as LowLevelLock was basically a "backstop" lock for cases that happen rarely or nearly always resulted in blocking wait. Perf improvements would not "move the needle".  
For static initializers we would need a recursive lock that also supports timeouts and preferably has the best performance that we can have. We may end up with basically a copy of Lock, just one that uses nonblocking Wait and does not do EventSource diagnostics. 

It would be a roundabout solution to something that could be solved inside the Lock itself.

Alternatively, Lock could have an internal ctor/flag to not initialize EventSource, but I think plugging into the diagnostics would be nice even for system locks.

jkotas:
> For static initializers we would need a recursive lock that also supports timeouts

I do not think we need that for static initializers. The static initializers are scheduled by "dead-lock aware lock". The regular (low-level) lock is needed to protect the dead-lock aware lock state. There should not be any recursion. The low-level lock for this purpose does not need to be fine-tuned, given other overheads involved in running the static constructors. 

We have the low-level lock exposed by the System.Native PAL. I think it would be hard to get rid of all uses of it. As long as we have the low-level lock exposed by the System.Native PAL, it should be fine to use it in S.P.CoreLib as needed.

VSadov:
> There should not be any recursion. The low-level lock for this purpose does not need to be fine-tuned, given other overheads involved in running the static constructors.

Given the fact that we run into contentions, and often enough to bring up the reentrancy issue, indicates that this lock can be a hot spot and thus a potential bottleneck if lock does not handle contentions well.

> As long as we have the low-level lock exposed by the System.Native PAL, it should be fine to use it in S.P.CoreLib as needed.

It is ok to use it. It works. It is just generally there are no reasons to use it vs. regular `lock{}`/`Lock`. 
Perhaps people get some sense of extra safety - "oh oh, this looks like low level code, I probably should use LowLevelLock".

There is an issue of "what if someone does things with SynchronizationContext.Wait", but to shield from that most locks would need to be LowLevelLock. That is probably why noone "does things with SynchronizationContext.Wait" as that would be very breaking.

kouvel:
What about locks taken on UI threads? I imagine they would still need to use a message-pumping wait for correctness and that could run user code and potentially lead to reentrancy.

kouvel:
Given that class construction runs user code, I'm figuring that the lock used for synchronizing it should have the same behavior. As for SynchronizationContext wait overrides and Thread.Interrupt, they are things that should probably be recommended against, and maybe even deprecated eventually. I'm not sure though what kind of breaks would be involved from these low-level kind of uses of interruptible/overridable/message-pumping waits.

kouvel:
Not sure though if a lock is actually held around running user code during class construction. If not, then it could very well be a simpler lock.

jkotas:
> Not sure though if a lock is actually held around running user code during class construction.

There is a lock held around running user static constructor code, but the attempt to run static constructor recursively needs to be short-circuited: https://github.com/dotnet/runtime/blob/57e3b0b75b9799f4b62e3a73fa81cc7c7fd671e0/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Runtime/CompilerServices/ClassConstructorRunner.cs#L145-L146 . It means that you need to be able to check whether the lock is taken by current thread, but it is easy to on the side using thread id if the lock does not support it.

I do not know why the native AOT ClassConstructorRunner uses timeouts. It looks both inefficient and complicated.

VSadov:
> What about locks taken on UI threads? I imagine they would still need to use a message-pumping wait for correctness and that could run user code and potentially lead to reentrancy. Should locks used in the runtime (in low-level situations) also use message-pumping waits?

With runtime is implemented mostly in managed code it a simpler model to see it as just extension of user's code - just another library. There are some exceptions for things too close to GC, but mostly it is code that the app could contain.

For the message pumping - I'd guess by default the same considerations equally apply to user code vs. managed runtime and sorting locks onto which may not need to pump might not make a lot of difference overall.

> As for SynchronizationContext wait overrides and Thread.Interrupt, they are things that should probably be recommended against, and maybe even deprecated eventually. 

I agree with this. 

Thread.Interrupt is actually not supported on NativeAOT, but on the long term TODO list. Might be a good time to just deprecate it. I wonder why it was not deprecated at the time Thread.Abort was.

jkotas:
> I wonder why it was not deprecated at the time Thread.Abort was.

It is popular in code ported from Java. According to https://apisof.net/catalog/4edbcb9f-8e93-65b1-23e2-49877d968c90, it is used by 12.4% of apps in our appcompat test suite.

jkotas:
> With runtime implemented mostly in managed code it a simpler model to see it as extension of user's code - just another library. There are some exceptions for things too close to GC, but mostly it is code that the app could contain.
> For the message pumping - I'd guess by default the same considerations equally apply to user code vs. managed runtime and sorting locks onto which may not need to pump might not make a lot of difference overall.

This is correct until things go wrong. Things go wrong with the recursive message pumping all the time. I think it is important that the lowest levels of the runtime are immune to this mess so that it manifest as an exception that the app developer can understand.

rickbrew:
> > As for SynchronizationContext wait overrides and Thread.Interrupt, they are things that should probably be recommended against, and maybe even deprecated eventually.
> 
> I agree with this.

I use a custom `SynchronizationContext` for the reason listed earlier -- to stop pumping during certain regions of code that are sensitive to reentrancy. Deprecating/removing that without a suitable replacement would be an adoption blocker for me. I don't use NativeAOT, however.

hamarb123:
> > As for SynchronizationContext wait overrides and Thread.Interrupt, they are things that should probably be recommended against, and maybe even deprecated eventually.
> 
> I agree with this.

I use `Thread.Interrupt`, removing it would break me. If the current `Thread.Interrupt` solution removed, it would be good to add a replacement for it, where you could `WaitInterrupt` explicitly or similar - this removes the issue of exceptions being thrown in random places, and would be nicer and more reliable to use overall.

> Thread.Interrupt is actually not supported on NativeAOT

If we remove the current `Thread.Interrupt` solution, hopefully we could design one that works on .NET 9 (with NativeAOT also).

kouvel:
> If the current `Thread.Interrupt` solution removed, it would be good to add a replacement for it, where you could `WaitInterrupt` explicitly or similar - this removes the issue of exceptions being thrown in random places, and would be nicer and more reliable to use overall.

Agreed, I feel `Thread.Interrupt` can be a bit disruptive, perhaps a better option would be to offer an interruptible wait explicitly so that it can be more clearly controlled which wait is intended to be interrupted.

> I use a custom `SynchronizationContext` for the reason listed earlier -- to stop pumping during certain regions of code that are sensitive to reentrancy. Deprecating/removing that without a suitable replacement would be an adoption blocker for me. I don't use NativeAOT, however.

Yep that's reasonable, maybe the wait overrides would be difficult to avoid, the actual wait may also be in a different library that is not in the user's control. The runtime may need to be careful about using such overridable waits in some cases.

jkotas:
> I feel Thread.Interrupt can be a bit disruptive,

Thread.Interrupt can be used safely only if you are certain that the code running on the other thread expects it.

> Thread.Interrupt is actually not supported on NativeAOT

If we find that Thread.Interrupt is needed in NativeAOT, I do not see any fundamental reason why it cannot be implemented in the current shape. It is "just work".

jkotas:
@kouvel Do you have a plan in mind for how to proceed on this one?

kouvel:
> @kouvel Do you have a plan in mind for how to proceed on this one?

Yes. I'm more in favor of reusing `Lock` for this purpose with a wait that does not allow reentrance. `LowLevelLock` (or some variant of it) would certainly be an option, but it's wasn't meant for this kind of purpose in its original or current state (which are the same). There might also be more uses for non-reentrant locks in other cases. `LowLevelLock` was created for specific purposes that are likely still valid. I'd like to see `LowLevelLock` go away eventually though, ideally `Lock` should be able to handle those cases just as well.

MichalStrehovsky:
@kouvel do you have any timelines for this? It is a heavy hitter in the CI (I see this failing every day) and we're about to snap .NET 9 Preview 1.

kouvel:
> @kouvel do you have any timelines for this? It is a heavy hitter in the CI (I personally see this failing every day and CI stats say we hit it 15 times a day on average) and we're about to snap .NET 9 Preview 1.

The proposed change in this PR would I believe fix the CI issue. The discussion in this PR has evolved into resolving a larger issue, which I'm currently looking into, hope to have something out soon. After thinking about it more, I may continue to employ the same technique for the CI issue, as using a finalizer doesn't guarantee timely initialization and I'd prefer to avoid going down the wait path before initialization is complete to avoid other potential reentrancy issues (and to have a way to fix a reentrancy issue on that path if one is introduced), though I'm still evaluating it.

jkotas (approved):
Thanks

jkotas:
> resolving a larger issue, which I'm currently looking into, hope to have something out soon. 

Could you please open an issue for this if there is not one already?

kouvel:
Rebased, and added a commit with one more null check in the wait path used by Lock. Opened https://github.com/dotnet/runtime/issues/97105 to track making some of the waits unalertable.

kouvel:
Looks like all remaining CI failures are known issues.

PR #94875, opened 2023-11-16
Don't put any modifiers after ref

### LINKED ISSUE #94840: LibraryImport generator can't generate correct syntax for ref struct
The current generated syntax is `internal ref unsafe partial struct`. However, current C# version requires `unsafe` to be in front of `ref`, e.g. `internal unsafe ref partial struct`.

The error message is also confusing now (type name required).

Related: https://github.com/dotnet/csharplang/blob/main/meetings/2018/LDM-2018-04-04.md#ordering-of-ref-and-partial-keywords

jtschuster:
Fixes https://github.com/dotnet/runtime/issues/94840.

We need to make sure we don't add the unsafe modifier after the `ref` keyword. We'll insert at the min index of `partial` and `ref` keyword.

### DIFF
--- a/src/libraries/System.Runtime.InteropServices/gen/Microsoft.Interop.SourceGeneration/SyntaxExtensions.cs
+++ b/src/libraries/System.Runtime.InteropServices/gen/Microsoft.Interop.SourceGeneration/SyntaxExtensions.cs
@@ -98,12 +98,22 @@ namespace Microsoft.Interop
         public static SyntaxTokenList AddToModifiers(this SyntaxTokenList modifiers, SyntaxKind modifierToAdd)
         {
             if (modifiers.IndexOf(modifierToAdd) >= 0)
+            {
                 return modifiers;
+            }
 
-            int idx = modifiers.IndexOf(SyntaxKind.PartialKeyword);
-            return idx >= 0
-                ? modifiers.Insert(idx, SyntaxFactory.Token(modifierToAdd))
-                : modifiers.Add(SyntaxFactory.Token(modifierToAdd));
+            int idxPartial = modifiers.IndexOf(SyntaxKind.PartialKeyword);
+            int idxRef = modifiers.IndexOf(SyntaxKind.RefKeyword);
+
+            int idxInsert = (idxPartial, idxRef) switch
+            {
+                (-1, -1) => modifiers.Count,
+                (-1, _) => idxRef,
+                (_, -1) => idxPartial,
+                (_, _) => Math.Min(idxPartial, idxRef)
+            };
+
+            return modifiers.Insert(idxInsert, SyntaxFactory.Token(modifierToAdd));
         }
 
         public static bool IsInPartialContext(this TypeDeclarationSyntax syntax, [NotNullWhen(false)] out SyntaxToken? nonPartialIdentifier)
--- a/src/libraries/System.Runtime.InteropServices/gen/Microsoft.Interop.SourceGeneration/SyntaxExtensions.cs
+++ b/src/libraries/System.Runtime.InteropServices/gen/Microsoft.Interop.SourceGeneration/SyntaxExtensions.cs
@@ -102,6 +102,7 @@ namespace Microsoft.Interop
                 return modifiers;
             }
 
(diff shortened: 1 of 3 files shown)


src/libraries/System.Runtime.InteropServices/tests/LibraryImportGenerator.UnitTests/Compiles.cs
```diff
@@ -439,8 +443,10 @@ public static IEnumerable<object[]> CustomCollections()
         }
 
         [Theory]
-        [MemberData(nameof(CodeSnippetsToCompile))]
-        [MemberData(nameof(CustomCollections))]
+        //[MemberData(nameof(CodeSnippetsToCompile))]
+        //[MemberData(nameof(CustomCollections))]
+        [MemberData(nameof(MyTests))]
+
```
AaronRobinsonMSFT, line 449:
```suggestion
        [MemberData(nameof(CodeSnippetsToCompile))]
        [MemberData(nameof(CustomCollections))]
```

jtschuster: Oops, forgot to change the tests back, thanks for pointing that out.

AaronRobinsonMSFT: I think this file might need to be reverted. I'm not terribly concerned with adding a new test here. My assumption on the test front would be that we should have observed a warning, no?

src/libraries/System.Runtime.InteropServices/tests/LibraryImportGenerator.UnitTests/CodeSnippets.cs
```diff
@@ -374,6 +374,54 @@ partial class Test
...
 
+        public static string LibraryImportInRefStruct = """
+            using System;
+            using System.Runtime.InteropServices;
+
+            public static partial class MyClass
+            {
+                public ref partial struct RSPublic
+                {
+                    [LibraryImport("DNE")]
```
AaronRobinsonMSFT, line 385: I'm usually okay with acrunyms like `"NDE"`, but it does tend to breed the question of what that means? Since this is in tests and the point is for them to be read, I think it generally appropriate to avoid the non-standard acronym. Since there is no context about that string I would prefer `"DoesNotExist"`.

src/libraries/System.Runtime.InteropServices/gen/Microsoft.Interop.SourceGeneration/SyntaxExtensions.cs
```diff
@@ -98,12 +98,22 @@ public static SyntaxTokenList StripAccessibilityModifiers(this SyntaxTokenList t
         public static SyntaxTokenList AddToModifiers(this SyntaxTokenList modifiers, SyntaxKind modifierToAdd)
         {
             if (modifiers.IndexOf(modifierToAdd) >= 0)
+            {
                 return modifiers;
+            }
+
+            int idxPartial = modifiers.IndexOf(SyntaxKind.PartialKeyword);
```
AaronRobinsonMSFT, line 105:
```suggestion
            // https://github.com/dotnet/csharplang/blob/main/meetings/2018/LDM-2018-04-04.md#ordering-of-ref-and-partial-keywords
            int idxPartial = modifiers.IndexOf(SyntaxKind.PartialKeyword);
```

src/libraries/System.Runtime.InteropServices/tests/LibraryImportGenerator.UnitTests/Compiles.cs
jtschuster: What do you mean we should have observed a warning?

jtschuster: It's an error from having a LibraryImport in a ref struct. I think it's just so uncommon put a LibraryImport in a ref struct before so we hadn't found the issue previously.

AaronRobinsonMSFT (approved):

PR #87841, opened 2023-06-20
Try adding cross-compilers for Musl

directhex:

PR #94876, opened 2023-11-16
We already have cross-compilers etc for musl and win-arm64. Actually use them

directhex:
This should get working wasi & wasm publishing on Alpine (i.e. lots of our Docker users)


lewing (approved):

directhex:
Seems the missing packages are missing due to an under-investigation yml issue

directhex:
I think I fixed the missing cross-compilers from #87841. They were building fine, but not ending up in packages.

directhex:
Official test build https://dev.azure.com/dnceng/internal/_build/results?buildId=2320816&view=results

PR #94879, opened 2023-11-16
Implements more advanced ILLink/ILCompiler analysis for Type.GetMember

### LINKED ISSUE #94745: Support `Type.GetMember` usage analysis properly with IL3050 warning (NAOT)
### Description

Using the `Type.GetMember` function gives warnings that are not necessary when AOT analysis is enabled.

### Reproduction Steps

Enable AOT compatible analysis: `<IsAotCompatible>true</IsAotCompatible>`.
Write code like so:
```csharp
type.GetMember("SomeMethod", MemberTypes.Method, BindingFlags.Public | BindingFlags.Static | BindingFlags.NonPublic);
```

### Expected behavior

All that should be necessary to appease the analyser is:
```csharp
[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods | DynamicallyAccessedMemberTypes.NonPublicMethods)]
```

### Actual behavior

Instead I get the following:
image

### Regression?

_No response_

### Known Workarounds

Ignore the warning.

### Configuration

.NET SDK 8.0.0
VS 17.8.0
Windows 10.0.19045 x64
I don't see why it would be specific to this configuration

### Other information

This was likely just looked over when implementing the analyser. I've only discovered it now since I'm only now adding AOT support.

hamarb123:
Fixes https://github.com/dotnet/runtime/issues/94745

Allows narrowing members down by name & by member type.


src/tools/illink/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -1216,6 +1278,23 @@ private IEnumerable<MultiValue> ProcessGetMethodByName (TypeProxy type, string m
...
+				yield return method;
+				foundAny = true;
+			}
+
+			// If there were no methods found the API will return null at runtime, so we should
+			// track the null as a return value as well.
+			// This also prevents warnings in such case, since if we don't set the return value it will be
+			// "unknown" and consumers may warn.
+			if (!foundAny)
+				yield return NullValue.Instance;
```
hamarb123, line 1295: Can a reviewer confirm if I want this section here, or not, or if I want it somewhere else? I don't know what it does (it's copied from the non WithPrefix version).

src/tools/illink/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -1454,5 +1592,15 @@ internal static bool BindingFlagsAreUnsupported (BindingFlags? bindingFlags)
...
+
+		private partial void MarkEventsOnTypeHierarchyWithPrefix(TypeProxy type, string name, bool nameIsPrefix, BindingFlags? bindingFlags);
+
+		private partial void MarkFieldsOnTypeHierarchyWithPrefix(TypeProxy type, string name, bool nameIsPrefix, BindingFlags? bindingFlags);
+
+		private partial void MarkPropertiesOnTypeHierarchyWithPrefix(TypeProxy type, string name, bool nameIsPrefix, BindingFlags? bindingFlags);
+
+		private partial IEnumerable<SystemReflectionMethodBaseValue> GetMethodsOnTypeHierarchyWithPrefix (TypeProxy type, string name, bool nameIsPrefix, BindingFlags? bindingFlags);
+
+		private partial IEnumerable<SystemTypeValue> GetNestedTypesOnTypeWithPrefix (TypeProxy type, string name, bool nameIsPrefix, BindingFlags? bindingFlags);
```
hamarb123, line 1604: Should I move all these `WithPrefix` versions next to the non `WithPrefix` versions in both files & make it so the non-`WithPrefix` version just calls the `WithPrefix` version?

src/tools/illink/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -1421,6 +1524,41 @@ internal static bool BindingFlagsAreUnsupported (BindingFlags? bindingFlags)
 			GetDynamicallyAccessedMemberTypesFromBindingFlagsForProperties (bindingFlags) |
 			GetDynamicallyAccessedMemberTypesFromBindingFlagsForNestedTypes (bindingFlags);
 
+		internal static bool HasMemberType (MemberTypes? memberTypes, MemberTypes? search) => memberTypes != null && (memberTypes & search) == search;
+
+		internal static DynamicallyAccessedMemberTypes GetDynamicallyAccessedMemberTypesFromMemberTypesForNestedTypes (MemberTypes? memberTypes) =>
+			(HasMemberType (memberTypes, MemberTypes.TypeInfo) ? DynamicallyAccessedMemberTypes.PublicNestedTypes | DynamicallyAccessedMemberTypes.NonPublicNestedTypes : DynamicallyAccessedMemberTypes.None) |
+			(HasMemberType (memberTypes, MemberTypes.NestedType) ? DynamicallyAccessedMemberTypes.PublicNestedTypes | DynamicallyAccessedMemberTypes.NonPublicNestedTypes : DynamicallyAccessedMemberTypes.None) |
```
hamarb123, line 1531: I'm slightly confused with the distinction between `TypeInfo` and `NestedType` - I think they both refer to nested types?
I assume `TypeInfo` would just be for when there's a nested type without `nested` attribute in IL? Otherwise, I have no idea what it is.
Can someone confirm what it means?

src/tools/illink/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -1381,6 +1461,29 @@ internal static bool BindingFlagsAreUnsupported (BindingFlags? bindingFlags)
...
+				MemberTypes.Event |
+				MemberTypes.Field |
+				MemberTypes.Method |
+				MemberTypes.Property |
+				MemberTypes.TypeInfo |
+				MemberTypes.NestedType;
+
+			// Member types that don't affect binding outside InvokeMember (that we don't analyze).
+			const MemberTypes IgnorableMemberTypes =
+				MemberTypes.Custom;
```
hamarb123, line 1481: I don't think we need to explicitly preserve attributes currently, do we?

src/tools/illink/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -394,13 +400,69 @@ GenericParameterValue genericParam
 					} else {
 						requiredMemberTypes = GetDynamicallyAccessedMemberTypesFromBindingFlagsForMembers (bindingFlags);
 					}
+					if (!MemberTypesAreUnsupported(memberTypes)) {
+						DynamicallyAccessedMemberTypes requiredMemberTypesMaskFromMemberTypes;
+						requiredMemberTypesMaskFromMemberTypes = GetDynamicallyAccessedMemberTypesFromMemberTypesForMembers (memberTypes);
+						requiredMemberTypes &= requiredMemberTypesMaskFromMemberTypes;
+					}
```
hamarb123, line 408: Should we exit early here if there's no flags in the `requiredMemberTypes`? This could happen, for example, with `MemberTypes.Custom`.

hamarb123:
Comment so this doesn't get closed - can someone please review this? Thanks!

sbomer (commented):
Thanks for the contribution!

src/tools/illink/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -367,18 +367,24 @@ GenericParameterValue genericParam
...
 					BindingFlags? bindingFlags;
+					MemberTypes? memberTypes;
 					if (calledMethod.HasMetadataParametersCount (1)) {
-						// Assume a default value for BindingFlags for methods that don't use BindingFlags as a parameter
-						bindingFlags = BindingFlags.Public | BindingFlags.Instance;
-					} else if (calledMethod.HasMetadataParametersCount (2) && calledMethod.HasParameterOfType ((ParameterIndex) 2, "System.Reflection.BindingFlags"))
+						// Assume a default value for MemberTypes for methods that don't use MemberTypes as a parameter
+						memberTypes = MemberTypes.All;
+						// Assume a default value for BindingFlags for methods that don't use BindingFlags as a parameter (Type.DefaultLookup)
+						bindingFlags = BindingFlags.Public | BindingFlags.Static | BindingFlags.Instance;
```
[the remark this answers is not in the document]
hamarb123: I will take a look through this next week some time (away currently).

src/tools/illink/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -1216,6 +1278,23 @@ private IEnumerable<MultiValue> ProcessGetMethodByName (TypeProxy type, string m
 				yield return NullValue.Instance;
 		}
 
+		private IEnumerable<MultiValue> ProcessGetMethodByNameWithPrefix (TypeProxy type, string methodName, bool nameIsPrefix, BindingFlags? bindingFlags)
```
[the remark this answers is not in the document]
vitek-karas: Do we think we need to recognize the prefix lookups intrinsically? It seems to bring complexity, but the benefit is questionable (How common is it, to use the prefix lookup functionality - I've never heard about it before this PR honestly). I think it would make sense to treat that case as "unsupported" and revert to the "keep all behavior" in that case.

There will probably always be cases which the trimming doesn't fully implement (it just doesn't make sense to invest resources into emulating every little detail of the runtime behavior, even if possible). As such it makes sense to consciously only implement the things which bring enough value. There's cost associated even with good code, we have to maintain it going forward, future us (and others) will re-read this code to understand and so on.

Same logic applies to the various `MemberType` values.

To put it simply - I don't believe there's value in implementing all of the functionality the runtime has. This is not about correctness, it's just about size. Intentionally leaving certain parts as "unrecognized" will not break applications, just make them larger.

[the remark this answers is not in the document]
hamarb123: @vitek-karas I use the `MemberType` values, and it was why I made the issue in the first place...

I can remove the prefix lookup code if we believe it overcomplicates things, as I don't think it provides a lot of value.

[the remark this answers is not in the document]
vitek-karas: @hamarb123 makes sense. I wasn't commenting on the whole change - specifically I was unsure about the prefixes. Removing that would make sense to me - and it should simplify the change (especially the shred code changes).

src/tools/illink/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -394,13 +400,72 @@ GenericParameterValue genericParam
...
+									// we may not need to search for constructors depending on the value of 'name', since they're always named '.ctor' (II.10.5.1)
+									DynamicallyAccessedMemberTypes requiredMemberTypes2 = requiredMemberTypes;
+									if (name != ".ctor") {
+										requiredMemberTypes2 &= ~(DynamicallyAccessedMemberTypes.PublicConstructors | DynamicallyAccessedMemberTypes.NonPublicConstructors);
+									}
+
+									AddReturnValue (MultiValueLattice.Top); // Initialize return value (so that it's not autofilled if there are no matching members)
+
+									// search for all the things we want by name
+									if ((requiredMemberTypes2 & (DynamicallyAccessedMemberTypes.PublicConstructors | DynamicallyAccessedMemberTypes.NonPublicConstructors)) != 0) {
```
vitek-karas, line 435: Why not something like `if (name == ".ctor" && ...)` and remove the check above about constructors - would probably simplify this a little bit.

src/tools/illink/src/ILLink.Shared/TrimAnalysis/HandleCallAction.cs
```diff
@@ -1381,6 +1461,29 @@ internal static bool BindingFlagsAreUnsupported (BindingFlags? bindingFlags)
...
+
+			// Member types we understand
+			const MemberTypes UnderstoodMemberTypes =
+				MemberTypes.Constructor |
+				MemberTypes.Event |
+				MemberTypes.Field |
+				MemberTypes.Method |
+				MemberTypes.Property |
+				MemberTypes.TypeInfo |
+				MemberTypes.NestedType;
```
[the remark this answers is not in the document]
hamarb123: I haven't implemented the changes to the tests yet, will close these comments when I do. I want to get the current tests to not fail first, then I'll change them to be what the new code should expect.

hamarb123:
> This pull request has been automatically marked `no-recent-activity` because it has not had any activity for 14 days. It will be closed if no further activity occurs within 14 more days. Any new comment (by anyone, not necessarily the author) will remove `no-recent-activity`.

I'm unable to look at this at the moment, or for another few weeks probably. Will re-open or open a new PR when I'm able to if/when I come back to it.

PR #94885, opened 2023-11-16
JIT: fix cross-block local assertion prop bug; add range enable

AndyAyersMS:
Fix condition under which we can share a pred's assertion out vector. Add the ability to disable cross-block local assertion prop via range.

Contributes to #93246.


AndyAyersMS:
FYI @dotnet/jit-contrib 
@jakobbotsch PTAL

jakobbotsch (approved):

AndyAyersMS:
This is a no-diff change, so failure in `Interop/DisabledRuntimeMarshalling` is almost certainly unrelated.

PR #94889, opened 2023-11-17
Type.GetMethod(name, genericParameterCount, bindingAttr, types) overload

### LINKED ISSUE #90995: [API Proposal]: Add overload of `Type.GetMethod` that takes a name, generic parameter count, binding flags and parameter types
### Background and motivation

In #42753, an overload of `Type.GetMethod` was added that takes a name, binding flags, and paramater types, to simplify usage when people don't care about the binder and paramater modifiers (which as I understand is there only for COM), which basically nobody passes in ever. However no equivalent overload was added that also takes genericParameterCount at the beginning. This is the overload I always use, even when the method isn't generic and I pass in 0, to be as specific as possible and make sure that my code still works even in case there is a generic overload added in the future. It seems that all the other overloads have an equivalent that takes genericParameterCount at the beginning, so this would just square off the feature in my opinion.

### API Proposal

```diff
 namespace System;

 public abstract partial class Type
 {
     public MethodInfo? GetMethod(string name, int genericParameterCount, Type[] types);
     public MethodInfo? GetMethod(string name, int genericParameterCount, Type[] types, ParameterModifier[]? modifiers);
+    public MethodInfo? GetMethod(string name, int genericParameterCount, BindingFlags bindingAttr, Type[] types);
     public MethodInfo? GetMethod(string name, int genericParameterCount, BindingFlags bindingAttr, Binder? binder, Type[] types, ParameterModifier[]? modifiers);
     public MethodInfo? GetMethod(string name, int genericParameterCount, BindingFlags bindingAttr, Binder? binder, CallingConventions callConvention, Type[] types, ParameterModifier[]? modifiers);
     public MethodInfo? GetMethod(string name);
     public MethodInfo? GetMethod(string name, Type[] types);
     public MethodInfo? GetMethod(string name, Type[] types, ParameterModifier[]? modifiers);
     public MethodInfo? GetMethod(string name, BindingFlags bindingAttr);
     public MethodInfo? GetMethod(string name, BindingFlags bindingAttr, Type[] types);
     public MethodInfo? GetMethod(string name, BindingFlags bindingAttr, Binder? binder, Type[] types, ParameterModifier[]? modifiers);
     public MethodInfo? GetMethod(string name, BindingFlags bindingAttr, Binder? binder, CallingConventions callConvention, Type[] types, ParameterModifier[]? modifiers);
 }
```

### API Usage

```csharp
using System.Reflection;

var method = typeof(C).GetMethod("M", 0, BindingFlags.Public | BindingFlags.Instance, Array.Empty<Type>());
// or "Type.EmptyTypes" instead of "Array.Empty<Type>()"

class C
{
    public void M();
}
```

### Alternative Designs
**Original issue alternative (not proposed)**: optionally, to square off the set of overloads completely, with complete parity between the ones that take genericParameterCount and the ones that don't (7 overloads for each):
```diff
 namespace System;

 public abstract partial class Type
 {
+    public MethodInfo? GetMethod(string name, int genericParameterCount);
     public MethodInfo? GetMethod(string name, int genericParameterCount, Type[] types);
     public MethodInfo? GetMethod(string name, int genericParameterCount, Type[] types, ParameterModifier[]? modifiers);
+    public MethodInfo? GetMethod(string name, int genericParameterCount, BindingFlags bindingAttr);
+    public MethodInfo? GetMethod(string name, int genericParameterCount, BindingFlags bindingAttr, Type[] types);
     public MethodInfo? GetMethod(string name, int genericParameterCount, BindingFlags bindingAttr, Binder? binder, Type[] types, ParameterModifier[]? modifiers);
     public MethodInfo? GetMethod(string name, int genericParameterCount, BindingFlags bindingAttr, Binder? binder, CallingConventions callConvention, Type[] types, ParameterModifier[]? modifiers);
     public MethodInfo? GetMethod(string name);
     public MethodInfo? GetMethod(string name, Type[] types);
     public MethodInfo? GetMethod(string name, Type[] types, ParameterModifier[]? modifiers);
     public MethodInfo? GetMethod(string name, BindingFlags bindingAttr);
     public MethodInfo? GetMethod(string name, BindingFlags bindingAttr, Type[] types);
     public MethodInfo? GetMethod(string name, BindingFlags bindingAttr, Binder? binder, Type[] types, ParameterModifier[]? modifiers);
     public MethodInfo? GetMethod(string name, BindingFlags bindingAttr, Binder? binder, CallingConventions callConvention, Type[] types, ParameterModifier[]? modifiers);
 }
```

### Risks

_No response_

I-SER-I:
close #90995

### DIFF
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/Runtime/General/ReflectionCoreCallbacksImplementation.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/Runtime/General/ReflectionCoreCallbacksImplementation.cs
@@ -324,7 +324,7 @@ namespace System.Reflection.Runtime.General
             Type? type = containingType.ToType();
             while (type != null)
             {
-                MethodInfo? methodInfo = type.GetMethod(method, 0, bindingFlags, null, parameterTypes, null);
+                MethodInfo? methodInfo = type.GetMethod(method, 0, bindingFlags, parameterTypes);
                 if (methodInfo != null && methodInfo.ReturnType.Equals(invokeMethod.ReturnType))
                     return (RuntimeMethodInfo)methodInfo; // This cast is safe since we already verified that containingType is runtime implemented.
 
--- a/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/Runtime/MethodInfos/CustomMethodMapper.cs
+++ b/src/coreclr/nativeaot/System.Private.CoreLib/src/System/Reflection/Runtime/MethodInfos/CustomMethodMapper.cs
@@ -59,7 +59,7 @@ namespace System.Reflection.Runtime.MethodInfos
             }
             else
             {
-                methodBase = declaringType.GetMethod(name, 0, bf, null, parameterTypes, null);
+                methodBase = declaringType.GetMethod(name, 0, bf, parameterTypes);
             }
 
             if (methodBase == null)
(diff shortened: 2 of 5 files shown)


buyaa-n (approved):
LGMT, thank you @I-SER-I !

Failures unrelated and known

PR #94896, opened 2023-11-17
JIT: Added DOTNET_JitRawHexCode/JitRawHexCodeFile

TIHan:
This adds a new environment variable, `DOTNET_JitRawHexCode/JitRawHexCodeFile`, in Debug/Checked builds that will enable writing the native code in hex to given file or stdout.

### DIFF
--- a/src/coreclr/jit/codegencommon.cpp
+++ b/src/coreclr/jit/codegencommon.cpp
@@ -2120,7 +2120,6 @@ void CodeGen::genEmitUnwindDebugGCandEH()
             FILE* dmpf = jitstdout();
 
             fprintf(dmpf, "Generated native code for %s:\n", compiler->info.compFullName);
-            fprintf(dmpf, "\n");
             hexDump(dmpf, addr, codeSize);
             fprintf(dmpf, "\n\n");
         }
--- a/src/coreclr/jit/codegencommon.cpp
+++ b/src/coreclr/jit/codegencommon.cpp
@@ -2098,21 +2098,31 @@ void CodeGen::genEmitUnwindDebugGCandEH()
 #endif // LATE_DISASM
 
 #ifdef DEBUG
-    if ((compiler->opts.disAsm || compiler->verbose) && compiler->opts.disAsmHexDumpFile)
+    if (JitConfig.JitRawHexCode().contains(compiler->info.compMethodHnd, compiler->info.compClassHnd,
+                                           &compiler->info.compMethodInfo->args))
     {
-        FILE*   hexDmpf;
-        errno_t ec = _wfopen_s(&hexDmpf, compiler->opts.disAsmHexDumpFile, W("at")); // NOTE: file append mode
-        if (ec == 0)
+        BYTE* addr = (BYTE*)*codePtr + compiler->GetEmitter()->writeableOffset;
+
+        const WCHAR* rawHexCodeFilePath = JitConfig.JitRawHexCodeFile();
+        if (rawHexCodeFilePath)
         {
-            assert(hexDmpf);
-
-            BYTE* addr = (BYTE*)*codePtr + compiler->GetEmitter()->writeableOffset;
-            for (unsigned int i = 0; i < codeSize; i++)
+            FILE*   hexDmpf;
+            errno_t ec = _wfopen_s(&hexDmpf, rawHexCodeFilePath, W("at")); // NOTE: file append mode
+            if (ec == 0)
             {
-                fprintf(hexDmpf, "%02X", *addr++);
+                assert(hexDmpf);
+                hexDump(hexDmpf, addr, codeSize);
+                fclose(hexDmpf);
             }
+        }
+        else
+        {
+            FILE* dmpf = jitstdout();
 
-            fclose(hexDmpf);
+            fprintf(dmpf, "Generated native code for %s:\n", compiler->info.compFullName);
+            fprintf(dmpf, "\n");
+            hexDump(dmpf, addr, codeSize);
+            fprintf(dmpf, "\n\n");
         }
     }
 #endif // DEBUG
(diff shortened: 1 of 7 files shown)


TIHan:
@dotnet/jit-contrib @kunalspathak

src/coreclr/jit/codegencommon.cpp
```diff
@@ -2097,6 +2097,26 @@ void CodeGen::genEmitUnwindDebugGCandEH()
...
+    if ((compiler->opts.disAsm || compiler->verbose) && compiler->opts.disAsmHexDumpFile)
+    {
+        FILE*   hexDmpf;
+        errno_t ec = _wfopen_s(&hexDmpf, compiler->opts.disAsmHexDumpFile, W("at")); // NOTE: file append mode
+        if (ec == 0)
+        {
+            assert(hexDmpf);
+
+            BYTE* addr = (BYTE*)*codePtr + compiler->GetEmitter()->writeableOffset;
+            for (unsigned int i = 0; i < codeSize; i++)
```
kunalspathak, line 2110: is it possible to use [fwrite](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fwrite?view=msvc-170) passing it `addr` and `codeSize` without the for loop?

kunalspathak (commented):
nit: I would go something without "disasm" (because it does not have mnemonics) and add "raw" in it . Something like `JitRawHexCodeFile` or similar.

src/coreclr/jit/compiler.h
```diff
@@ -9705,23 +9705,24 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
...
+        bool disAsmSpilled;             // Display native code when any register spilling occurs
+        bool disasmWithGC;              // Display GC info interleaved with disassembly.
+        bool disAddr;                   // Display process address next to each instruction in disassembly code
+        bool disAsm2;                   // Display native code after it is generated using external disassembler
+        bool dspOrder;                  // Display names of each of the methods that we ngen/jit
+        bool dspUnwind;                 // Display the unwind info output
+        bool compLongAddress;           // Force using large pseudo instructions for long address
+                                        // (IF_LARGEJMP/IF_LARGEADR/IF_LARGLDC)
+        bool dspGCtbls;                 // Display the GC tables
+        const WCHAR* disAsmHexDumpFile; // Writes the raw bytes in hex of native code to the specified file
```
jakobbotsch, line 9725: Can we just use the `JitConfig` variable directly from codegencommon.cpp?

src/coreclr/jit/codegencommon.cpp
```diff
@@ -2097,6 +2097,26 @@ void CodeGen::genEmitUnwindDebugGCandEH()
     getDisAssembler().disAsmCode((BYTE*)*codePtr, finalHotCodeSize, (BYTE*)coldCodePtr, finalColdCodeSize);
 #endif // LATE_DISASM
 
+#ifdef DEBUG
+    if ((compiler->opts.disAsm || compiler->verbose) && compiler->opts.disAsmHexDumpFile)
```
BruceForstall, line 2101: Why do you need to set disAsm or verbose? Couldn't this work fine with just setting the new variable?

kunalspathak: +1 on that.

BruceForstall: There could be:
```
CONFIG_METHODSET(JitRawHexCode,...
CONFIG_STRING(JitRawHexCodeFile,...
```
used as:
```
DOTNET_JitRawHexCode=foo bar baz
DOTNET_JitRawHexCodeFile=out.asm
```

TIHan: > although, one of the reason could be that we just want to output bytes for a particular method instead of all the methods.

That was the idea, but it would be fine to also change it to have the separate variables like @BruceForstall mentioned. I can do that.

src/coreclr/jit/codegencommon.cpp
TIHan: I can look at it, I imagine it would be preferrable.

src/coreclr/jit/compiler.h
TIHan: That's fair since it is just one place. I can change it.

src/coreclr/jit/codegencommon.cpp
TIHan: @BruceForstall @kunalspathak Wouldn't we need `verbose` if we want to capture the code when running the emitter tests?

BruceForstall: `genArm64EmitterUnitTests` currently only runs under `verbose`, but that's a separate issue, and can be changed (especially given this functionality)

src/coreclr/jit/codegencommon.cpp
TIHan: There doesn't seem to be a way to write out the bytes in hex form for `fwrite`.

TIHan:
@dotnet/jit-contrib this is ready again.

src/coreclr/jit/utils.cpp
```diff
@@ -1328,19 +1328,10 @@ void hexDump(FILE* dmpf, const char* name, BYTE* addr, size_t size)
 
     assert(addr);
 
-    fprintf(dmpf, "Hex dump of %s:\n", name);
-
     for (unsigned i = 0; i < size; i++)
     {
-        if ((i % 16) == 0)
```
kunalspathak, line 1335: I am assuming we don't want this anymore and we will never print the offsets?

kunalspathak (approved):
LGTM

src/coreclr/jit/utils.cpp
TIHan: Yea, we will just print the hex-code as is. The disassembly tool we will use can put in the offsets when decoding I believe.

PR #85912, opened 2023-05-08
Add Ubuntu Kinetic 22.10 and Mantic 23.10 RIDs

dviererbe:
Adds the Runtime Identifiers (RIDs) for Ubuntu 22.10 Kinetic Kudu and the recently announced Ubuntu 23.10 Mantic Minotaur.

Related to #84974, #84983, #84984

PR #84984, opened 2023-04-18
Add Ubuntu 22.10 kinetic kudu RIDs for .NET 7

dviererbe:
Related to https://github.com/dotnet/runtime/issues/84974

PR #84983, opened 2023-04-18
Add Ubuntu 22.10 kinetic kudu RIDs for .NET 6

dviererbe:
Related to https://github.com/dotnet/runtime/issues/84974

PR #94913, opened 2023-11-17
Add Ubuntu 24.04 LTS (Noble Numbat) Runtime Identifiers to .NET 6.0

dviererbe:
While building .NET 6 on Ubuntu 24.04 LTS (Noble Numbat) we encountered build failures,  because the runtime id was missing:

```
error NETSDK1083: The specified RuntimeIdentifier 'ubuntu.24.04-x64' is not recognized.
```

Adding the missing RIDs solved this problem.  

Last time (https://github.com/dotnet/runtime/pull/85912, https://github.com/dotnet/runtime/issues/84974, https://github.com/dotnet/runtime/pull/84983, https://github.com/dotnet/runtime/pull/84984), when I opened a PR to add the Ubuntu 22.10 (Kinetic Kudu) and 23.10 (Mantic Minotaur) RIDs, it was also required to add the patch to the main branch first. I skip this step, because the RID graph is frozen and noone is supposed to add RIDs there.

Note: It currently looks the way that we will not ship .NET 6 package with Ubuntu 24.04 LTS (Noble Numbat), because of the limited lifetime left once Ubuntu 24.04 LTS (Noble Numbat) is released, but this is not final yet. Therefore I don't know if it makes sense to actually include this patch. We just wanted to offer it.


dviererbe:
We just had a discussion and it is now final that we will not ship .NET 6 with Ubuntu 24.04 LTS (Noble Numbat).

carlossanlop:
> We just had a discussion and it is now final that we will not ship .NET 6 with Ubuntu 24.04 LTS (Noble Numbat).

So that means we can close this, right?

Issue #94786, opened 2023-11-15
crossgen2 fails on FIPS enabled systems (due to MD5 hashing)


tmds:
`crossgen2` fails on FIPS enabled systems due to it using an MD5 hash.

```
  Emitting R2R PE file: /home/tester/runtime/artifacts/bin/coreclr/linux.x64.Release/System.Private.CoreLib.dll
Error : error : 060800C8:digital envelope routines:EVP_DigestInit_ex:disabled for FIPS [/home/tester/runtime/src/coreclr/crossgen-corelib.proj]
Interop+Crypto+OpenSslCryptographicException : error : 060800C8:digital envelope routines:EVP_DigestInit_ex:disabled for FIPS [/home/tester/runtime/src/coreclr/crossgen-corelib.proj]
     at System.Security.Cryptography.HashProviderDispenser.OneShotHashProvider.HashData(String hashAlgorithmId, ReadOnlySpan`1 source, Span`1 destination)
     at System.Security.Cryptography.MD5.TryHashData(ReadOnlySpan`1 source, Span`1 destination, Int32& bytesWritten)
     at System.Security.Cryptography.MD5.HashData(ReadOnlySpan`1 source, Span`1 destination)
     at System.Security.Cryptography.MD5.HashData(ReadOnlySpan`1 source)
     at System.Security.Cryptography.MD5.HashData(Byte[] source)
     at ILCompiler.Diagnostics.PerfMapWriter.PerfMapV1SignatureHelper(IEnumerable`1 inputAssemblies, TargetDetails details) in /home/tester/runtime/src/coreclr/tools/aot/ILCompiler.Diagnostics/PerfMapWriter.cs:line 104
     at ILCompiler.DependencyAnalysis.ReadyToRunObjectWriter.EmitPortableExecutable() in /home/tester/runtime/src/coreclr/tools/aot/ILCompiler.ReadyToRun/CodeGen/ReadyToRunObjectWriter.cs:line 373
     at ILCompiler.ReadyToRunCodegenCompilation.Compile(String outputFile) in /home/tester/runtime/src/coreclr/tools/aot/ILCompiler.ReadyToRun/Compiler/ReadyToRunCodegenCompilation.cs:line 400
[...]

PR #94934, opened 2023-11-17
Permit MD5 regardless of FIPS configuration for Linux

vcsjones:
This changes the native shim to always permit the use of MD5, regardless of whether or not the system is configured to block the use MD5 for FIPS purposes. This is only for the MD5 primitive. HMAC-MD5 remains blocked.

We currently do not have any CI that exercises Linux in this configuration. This was manually verified in RHEL 8 and RHEL 9 for OpenSSL 1.1.1 and OpenSSL 3, respectively.

OpenSSL 1.1.1:

image

OpenSSL 3.0:

image

/cc @bartonjs @GrabYourPitchforks

### DIFF
--- a/src/native/libs/System.Security.Cryptography.Native/opensslshim.h
+++ b/src/native/libs/System.Security.Cryptography.Native/opensslshim.h
@@ -347,6 +347,7 @@ int EVP_DigestFinalXOF(EVP_MD_CTX *ctx, unsigned char *md, size_t len);
     REQUIRED_FUNCTION(EVP_MD_CTX_copy_ex) \
     RENAMED_FUNCTION(EVP_MD_CTX_free, EVP_MD_CTX_destroy) \
     RENAMED_FUNCTION(EVP_MD_CTX_new, EVP_MD_CTX_create) \
+    REQUIRED_FUNCTION(EVP_MD_CTX_set_flags) \
     LIGHTUP_FUNCTION(EVP_MD_fetch) \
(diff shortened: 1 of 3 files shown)


bartonjs (approved):

jeffhandley:
@tmds I saw you [referenced this PR](https://github.com/dotnet/runtime/issues/94786#issuecomment-1819165446) in a way that acknowledges this change in behavior. I am inferring that acknowledgement as a signoff from you on this PR--please let me know if that inference was incorrect.

tmds:
It was not an acknowledgement to sign-off from me.

I learned that this PR should fix the issue I had reported in https://github.com/dotnet/runtime/issues/94786, but I have not yet been able to verify that.

jeffhandley:
OK; thanks, @tmds! Do you have any concerns with this change? Permitting MD5 regardless of FIPS configuration yields consistent behavior for .NET and without this change, the behavior is in effect a regression. From what I could tell, this change should resolve #94786, which was a symptom of that regression. Nonetheless, I wanted to make sure you had the opportunity to review this change, as we're sending it for .NET 8 servicing.

tmds:
I kicked of a build to verify this change fixes https://github.com/dotnet/runtime/issues/94786. This build uses the source-build config (`DotNetBuildFromSource=true`) which links directly with the system OpenSSL.
Unless something unexpected happens, I'll have the result in less than an hour.

tmds:
Happy to confirm this fixes https://github.com/dotnet/runtime/issues/94786.

PR #94935, opened 2023-11-17
JIT: avoid local jtrue assertions involving floating types

AndyAyersMS:
Float relop equality does not imply bitwise equality. So skip making local jtrue assertions about floating types.

Contributes to #93246.

### DIFF
--- a/src/coreclr/jit/assertionprop.cpp
+++ b/src/coreclr/jit/assertionprop.cpp
@@ -2205,6 +2205,13 @@ AssertionInfo Compiler::optAssertionGenJtrue(GenTree* tree)
     GenTree* op1 = relop->AsOp()->gtOp1->gtCommaStoreVal();
     GenTree* op2 = relop->AsOp()->gtOp2->gtCommaStoreVal();
 
+    // Avoid creating local assertions for float types.
+    //
+    if (optLocalAssertionProp && varTypeIsFloating(op1))
+    {
+        return NO_ASSERTION_INDEX;
+    }
+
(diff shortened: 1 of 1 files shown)


jakobbotsch (approved):

src/coreclr/jit/assertionprop.cpp
```diff
@@ -2205,6 +2205,13 @@ AssertionInfo Compiler::optAssertionGenJtrue(GenTree* tree)
     GenTree* op1 = relop->AsOp()->gtOp1->gtCommaStoreVal();
     GenTree* op2 = relop->AsOp()->gtOp2->gtCommaStoreVal();
 
+    // Avoid creating local assertions for float types.
+    //
+    if (optLocalAssertionProp && varTypeIsFloating(op1))
```
tannergooding, line 2210: 👍. 

Just noting we could expand this in the future...

We have some well-defined APIs like `IsNaN`, `IsInfinity`, `IsFinite`, etc. These can all be used to drive other assertions.

Likewise if we know that op2 isn't NaN and isn't zero, then bitwise equality does hold (we take advantage of this in import to optimize `min` and `max` for example, when one input is a constant).

tannergooding (approved):

src/coreclr/jit/assertionprop.cpp
AndyAyersMS: Added a note to https://github.com/dotnet/runtime/issues/93246 so we don't forget about this.

Issue #94931, opened 2023-11-17
DisabledRuntimeMarshalling.PInvokeAssemblyMarshallingEnabled.DelegatesFromExternalAssembly.StructWithDefaultNonBlittableFields() failing


lewing:
## Build Information
Build: https://dev.azure.com/dnceng-public/cbb18261-c48f-4abb-8651-8cdcb5474649/_build/results?buildId=473177
Build error leg or test failing: Interop/DisabledRuntimeMarshalling/DisabledRuntimeMarshalling_Disabled_NativeAssemblyEnabled/DisabledRuntimeMarshalling_Disabled_NativeAssemblyEnabled.cmd
Pull request: https://github.com/dotnet/runtime/pull/94878

## Error Message

Fill the error message using [step by step known issues guidance](https://github.com/dotnet/arcade/blob/main/Documentation/Projects/Build%20Analysis/KnownIssues.md#how-to-fill-out-a-known-issue-error-section).

```json
{
  "ErrorMessage": "DisabledRuntimeMarshalling.PInvokeAssemblyMarshallingEnabled.DelegatesFromExternalAssembly.StructWithDefaultNonBlittableFields()",
  "ErrorPattern": "",
  "BuildRetry": false,
  "ExcludeConsoleLog": false
}
```

 ### Known issue validation
**Build: :mag_right:** https://dev.azure.com/dnceng-public/public/_build/results?buildId=473177
**Error message validated:** `DisabledRuntimeMarshalling.PInvokeAssemblyMarshallingEnabled.DelegatesFromExternalAssembly.StructWithDefaultNonBlittableFields()`
**Result validation: :white_check_mark:** Known issue matched with the provided build.
**Validation performed at:** 11/17/2023 6:22:15 PM UTC

### Report

|Build|Definition|Test|Pull Request|
|---|---|---|---|
|[475116](https://dev.azure.com/dnceng-public/public/_build/results?buildId=475116)|dotnet/runtime|[Interop/DisabledRuntimeMarshalling/DisabledRuntimeMarshalling_Disabled_NativeAssemblyEnabled/DisabledRuntimeMarshalling_Disabled_NativeAssemblyEnabled.cmd](https://dev.azure.com/dnceng-public/public/_build/results?buildId=475116&view=ms.vss-test-web.build-test-results-tab&runId=10855880&resultId=117052)||
[...]

PR #94938, opened 2023-11-17
Improve the hash function used in the `InstMethodHashTable`

### LINKED ISSUE #94936: Async methods become slower to compile in applications which have a very large number of assemblies
### Description

When there are _many_ different assemblies in the product which are used, and these assemblies all have async code in them, performance of `FindOrCreateAssociatedMethodDesc` becomes very slow. The repro is fairly simple (although not practical to include in this bug, as it is extremely large). Create 6000 small assemblies each with a number of async methods in it. Then call `System.Runtime.CompilerServices.RuntimeHelpers.PrepareMethod` on each async method. Observe in a profiler, that the time spent in the application during the series of calls to `PrepareMethod` that the runtime spends approximately 50% of the time in the JIT calling `FindOrCreateAssociatedMethodDesc`, and that the time in there is spent walking a linked list in the `InstMethodHashTable`.

### Configuration

All configurations, all OS.

### Regression?

This is not a regression.

### Data

The benchmark above takes approximately 56 seconds to run on my local machine.

### Analysis

An analysis of the `InstMethodHashTable` indicates that the hash function does not hash on the unique types used in a method instantiation. Instead it hashes based on the TypeDef token of those types. In the presence of 6000 small assemblies of similar shape, the hash tables ends up with extremely unbalanced bucket sizes, and a depth of 6000 entries in the linked list walk to find a single `MethodDesc`.

davidwrighton:
Instead of using the TypeDef token of the MethodTable, use the TypeHandle pointer instead. This is both faster to compute, and is more unique, as the TypeDef token could be shared across multiple types in different assemblies, and if the type is generic itself, it could be shared across multiple different instantiations of the type.

Improves the benchmark mentioned in isseu #94936 such that the cost is about 24 seconds, and the time in spent is spent almost entirely in the JIT, as should be expected.

Fixes #94936

### DIFF
--- a/src/coreclr/vm/instmethhash.cpp
+++ b/src/coreclr/vm/instmethhash.cpp
@@ -96,30 +96,19 @@ static DWORD Hash(TypeHandle declaringType, mdMethodDef token, Instantiation ins
 
     DWORD dwHash = 0x87654321;
 #define INST_HASH_ADD(_value) dwHash = ((dwHash << 5) + dwHash) ^ (_value)
+#ifdef TARGET_64BIT
+#define INST_HASH_ADDPOINTER(_value) INST_HASH_ADD((uint32_t)(uintptr_t)_value); INST_HASH_ADD((uint32_t)(((uintptr_t)_value) >> 32))
+#else
+#define INST_HASH_ADDPOINTER(_value) INST_HASH_ADD((uint32_t)(uintptr_t)_value);
+#endif
 
-    INST_HASH_ADD(declaringType.GetCl());
+    INST_HASH_ADDPOINTER(declaringType.AsPtr());
     INST_HASH_ADD(token);
 
     for (DWORD i = 0; i < inst.GetNumArgs(); i++)
     {
         TypeHandle thArg = inst[i];
-
-        if (thArg.GetMethodTable())
-        {
-            INST_HASH_ADD(thArg.GetCl());
-
-            Instantiation sArgInst = thArg.GetInstantiation();
-            for (DWORD j = 0; j < sArgInst.GetNumArgs(); j++)
-            {
-                TypeHandle thSubArg = sArgInst[j];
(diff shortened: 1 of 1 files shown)


src/coreclr/vm/instmethhash.cpp
```diff
@@ -96,30 +96,19 @@ static DWORD Hash(TypeHandle declaringType, mdMethodDef token, Instantiation ins
...
 #define INST_HASH_ADD(_value) dwHash = ((dwHash << 5) + dwHash) ^ (_value)
+#ifdef TARGET_64BIT
+#define INST_HASH_ADDPOINTER(_value) INST_HASH_ADD((uint32_t)(uintptr_t)_value); INST_HASH_ADD((uint32_t)(((uintptr_t)_value) >> 32))
+#else
+#define INST_HASH_ADDPOINTER(_value) INST_HASH_ADD((uint32_t)(uintptr_t)_value);
+#endif
 
-    INST_HASH_ADD(declaringType.GetCl());
+    INST_HASH_ADDPOINTER(declaringType.AsPtr());
     INST_HASH_ADD(token);
```
AaronRobinsonMSFT, line 106:
```suggestion
```

I think we can remove including the `token` now as well. If we leave this, I think we need to add a comment as to why the pointer hashing isn't enough.

AaronRobinsonMSFT (approved):

src/coreclr/vm/instmethhash.cpp
davidwrighton: The token is to identify twhich method on the type, so we still need that

AaronRobinsonMSFT: Oh right. That makes sense as the pointer is just for the type and not the member. Okay. Looks good to me.

davidwrighton:
Failing test is #94931

PR #94250, opened 2023-11-01
Fix JitOptRepeat

BruceForstall:
Many individual fixes have already been merged. This change adds two more fixes, updates the disabling of OptRepeat for one test, and enables OptRepeat in JitStress.

PR #94947, opened 2023-11-18
Fix LSRA handling of arm32 double registers and spilling

BruceForstall:
If LSRA uses `regsBusyUntilKill` to eliminate registers from consideration for spilling, for spilling arm32 double type registers, it needs to remove the even registers from the candidates. For example, if `regsBusyUntilKill` contains `f3`, candidates also needs to remove `f2`, since double register candidates only contains the even registers to represent to pair of registers for an arm32 double.

### DIFF
--- a/src/coreclr/jit/lsra.cpp
+++ b/src/coreclr/jit/lsra.cpp
@@ -7946,8 +7946,9 @@ regNumber LinearScan::getTempRegForResolution(BasicBlock*      fromBlock,
 #ifdef TARGET_ARM
     if (type == TYP_DOUBLE)
     {
-        // Exclude any doubles for which the odd half isn't in freeRegs.
-        freeRegs = freeRegs & ((freeRegs << 1) & RBM_ALLDOUBLE);
+        // Exclude any doubles for which the odd half isn't in freeRegs,
+        // and restrict down to just the even part of the even/odd pair.
+        freeRegs &= (freeRegs & RBM_ALLDOUBLE_HIGH) >> 1;
     }
 #endif
 
@@ -12462,13 +12463,10 @@ regMaskTP LinearScan::RegisterSelection::select(Interval*    currentInterval,
         // is also available. Thus, for every busyReg that is an odd floating-point register, we need to
         // remove from candidates the corresponding even floating-point register. For example, if busyRegs
         // contains `f3`, we need to remove `f2` from the candidates for a double register interval. The
-        // `~(busyRegs >> 1)` clause below creates a mask to do this. Note one oddity: if busyRegs contained
-        // f0, we would mask off some general-purpose register. If it contains any general-purpose registers,
-        // the mask will also contain general-purpose registers. That's ok because candidates for a double
-        // type starts off as only the even floating-point registers, so masking off anything else is a no-op.
+        // clause below creates a mask to do this.
         if (currentInterval->registerType == TYP_DOUBLE)
         {
-            candidates &= ~(busyRegs >> 1);
+            candidates &= ~((busyRegs & RBM_ALLDOUBLE_HIGH) >> 1);
         }
 #endif // TARGET_ARM
 
--- a/src/coreclr/jit/lsra.cpp
+++ b/src/coreclr/jit/lsra.cpp
@@ -12457,6 +12457,14 @@ regMaskTP LinearScan::RegisterSelection::select(Interval*    currentInterval,
         regMaskTP busyRegs = linearScan->regsBusyUntilKill | linearScan->regsInUseThisLocation;
         candidates &= ~busyRegs;
 
+#ifdef TARGET_ARM
+        // For TYP_DOUBLE on ARM, we can only use register for which the odd half is also available.
+        if (currentInterval->registerType == TYP_DOUBLE)
+        {
+            candidates &= ~(busyRegs >> 1);
+        }
+#endif // TARGET_ARM
+
 #ifdef DEBUG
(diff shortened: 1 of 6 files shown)


BruceForstall:
All the test failures are known or infra

BruceForstall:
fyi, I found this testing `JitOptRepeat`.

BruceForstall:
@kunalspathak PTAL
cc @dotnet/jit-contrib

src/coreclr/jit/lsra.cpp
```diff
@@ -12457,6 +12457,14 @@ regMaskTP LinearScan::RegisterSelection::select(Interval*    currentInterval,
         regMaskTP busyRegs = linearScan->regsBusyUntilKill | linearScan->regsInUseThisLocation;
         candidates &= ~busyRegs;
 
+#ifdef TARGET_ARM
+        // For TYP_DOUBLE on ARM, we can only use register for which the odd half is also available.
+        if (currentInterval->registerType == TYP_DOUBLE)
```
kunalspathak, line 12462: If `busyRegs` does not contain both registers of the pairs, a more appropriate fix would be to find place where we miss setting 1 of the register and fix it. When we set `regMask` in `regsBusyUntilKill` or `regsInUseThisLocation`, we mostly use `getRegMask` which takes into account the "double register" as well. 
https://github.com/dotnet/runtime/blob/740ecd0a8db3f6a7714f3e15bac0774f033720f0/src/coreclr/jit/lsra.h#L1740-L1746

kunalspathak: > `genRegMask` versus `getRegMask`

I think we should use `getRegMask(reg, var_type)` version and delete `genRegMask(reg, var_type)`. They are doing same thing.

> So, say f2 is a candidate, and f3 is a busy register. we need to remove f2 from the candidate set due to f3 being busy.

Correct, thats what I expect, but it should happen at place that added `f3` as busy register. Do you know which of `regsBusyUntilKill` or `regsInUseThisLocation` does it gets marked as busy? The change looks correct, but I want to make sure we do not accidently mark a register (probably in jitstressregs) not part of current even/odd pair as busy.

BruceForstall: The case was `regsBusyUntilKill`.

> but it should happen at place that added f3 as busy register

If `f3` is marked as busy because it is a float (single-precision) type, then you wouldn't want to mark `f2` busy at that point. It's only when we're looking for candidates for a double that we need to mask off both `f2` and `f3` if `f3` is already marked as busy.

kunalspathak: Do you mind sharing the repro or the JitDump file?

BruceForstall: Sent a link to JitDump offline.

You can also repro locally using:

```
py src\coreclr\scripts\superpmi.py replay -f coreclr_tests -arch x86 -target_arch arm -target_os linux -jitoption JitOptRepeat#* -jitoption JitOptRepeatCount#2 -jitoption JitStressRegs#3
```

And looking for:

```
Assertion failed '!isRegBusy(physRegRecord->regNum, current->registerType)'
```

(for me, MC#459399)

(I’ve fixed the other asserts in https://github.com/dotnet/runtime/pull/94250)

kunalspathak: I tried to repro and see how `f3` ends up in `regsBusyUntilKill`. I am ok with this fix, but can we add `assert((busyRegs & 0x555555550000) == RBM_NONE)` which will confirm at least that we never have `f2` in this mask and hence we don't accidently remove `f1` from the mask?

BruceForstall:
@kunalspathak I updated the comment

kunalspathak:
> @kunalspathak I updated the comment

can you also explain the implication of ` >> 1`  if we have `f2` in the mask as we spoke offline?

src/coreclr/jit/lsra.cpp
```diff
@@ -7946,8 +7946,9 @@ regNumber LinearScan::getTempRegForResolution(BasicBlock*      fromBlock,
 #ifdef TARGET_ARM
     if (type == TYP_DOUBLE)
     {
-        // Exclude any doubles for which the odd half isn't in freeRegs.
-        freeRegs = freeRegs & ((freeRegs << 1) & RBM_ALLDOUBLE);
+        // Exclude any doubles for which the odd half isn't in freeRegs,
+        // and restrict down to just the even part of the even/odd pair.
+        freeRegs &= (freeRegs & RBM_ALLDOUBLE_HIGH) >> 1;
```
kunalspathak, line 7951: For better understanding this concept, should we create a function or macro something like:

```c++
#define RBM_ALLDOUBLE_EVEN(mask) (mask & RBM_ALLDOUBLE_HIGH) >> 1
```

BruceForstall: hmm... that seems very confusing to me. `RBM_ALLDOUBLE` already is the even or "low" registers of the pairs.

This change fixes a bug I noticed just by inspection, and leads to some diffs. I'm still investigating, but I also found some pre-existing corruption in double register output in double register disasm, so I'm continuing down the rabbit hole...

BruceForstall:
I fixed a bug with disasm output of arm double registers d0-d15.

There are register allocation diffs due to the `getTempRegForResolution` fix because previously we would never consider d0 free (there may be other cases as well).

@kunalspathak PTAL

kunalspathak (approved):
LGTM

PR #94955, opened 2023-11-18
JIT: Process all EH successors for "in-block" defs in SSA

### LINKED ISSUE #94954: JIT: SSA does not add proper phi args for second pass EH successors
SSA only looks at the full set of EH successors as part of the "end-of-block" successor iteration. When it sees definitions inside the block, it only looks for enclosing handlers. For example:
```csharp
private static void Foo()
{
    int x = 123;
    try
    {
        try
        {
            throw new Exception();
        }
        finally
        {
            Console.WriteLine(x);
        }
    }
    catch (Exception) when (Bar(x = 50, Throw(x), x = 123))
    {
    }
}

private static bool Bar(int x, int y, int z)
{
    return true;
}

private static int Throw(int x)
{
    throw new Exception();
}

```
The `x = 50` here needs to induce a phi arg in the enclosed `finally` handler, but it doesn't. We effectively get a PHI that indicates the only possible value for `x` is `123` inside the enclosed finally, but this program enters the finally block with `x=50`.
It does not cause silent bad codegen here because there is a cycle in the flow-graph between the enclosed finally and the filter, so VN isn't able to deduce `x = 123` from the phis.

jakobbotsch:
When we saw a def we were only adding phi args to enclosing handlers, but this misses the special case of throws inside filters, where control can flow to enclosed handlers.

Fix #94954

### DIFF
--- a/src/coreclr/jit/block.h
+++ b/src/coreclr/jit/block.h
@@ -1416,6 +1416,9 @@ public:
     BasicBlockVisit VisitAllSuccs(Compiler* comp, TFunc func);
 
     template <typename TFunc>
+    BasicBlockVisit VisitEHSuccs(Compiler* comp, TFunc func);
+
+    template <typename TFunc>
     BasicBlockVisit VisitRegularSuccs(Compiler* comp, TFunc func);
 
     bool HasPotentialEHSuccs(Compiler* comp);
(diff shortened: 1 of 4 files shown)


jakobbotsch:

[No diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=475839&view=ms.vss-build-web.run-extensions-tab).

Fix here is just to expose the EH successor iteration that `VisitAllSuccs` already has (taking care to yield all EH successors for `BBJ_CALLFINALLY`) and then switch SSA to use it.

BruceForstall (approved):
Nice find!

LGTM

cincuranet:
Improvements:
- https://github.com/dotnet/perf-autofiling-issues/issues/25195

PR #94956, opened 2023-11-18
Fix comment typo in jithelp.asm

ericmutta:
The sentence:

> it is imperative that the addresses **of of** the values...

Should read:

> it is imperative that the addresses **of** the values

PS: Came across this will trying to figure out how the JIT transitions into native code after generating it from IL (I still haven't figured it all out so if there any pointers for this particular topic, I would appreciate some guidance 🙏).

### DIFF
--- a/src/coreclr/vm/amd64/JitHelpers_FastWriteBarriers.asm
+++ b/src/coreclr/vm/amd64/JitHelpers_FastWriteBarriers.asm
@@ -31,7 +31,7 @@ include asmconstants.inc
 ; for more info.
 ;
 ; READ THIS!!!!!!
-; it is imperative that the addresses of of the values that we overwrite
+; it is imperative that the addresses of the values that we overwrite
 ; (card table, ephemeral region ranges, etc) are naturally aligned since
 ; there are codepaths that will overwrite these values while the EE is running.
 ;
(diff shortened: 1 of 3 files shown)


jkotas:
The same typo is in multiple files:
```
src\coreclr\vm\amd64\JitHelpers_FastWriteBarriers.asm(34):; it is imperative that the addresses of of the values that we overwrite
src\coreclr\vm\i386\jithelp.asm(824):; it is imperative that the addresses of of the values that we overwrite
src\coreclr\vm\i386\jithelp.S(682)://  it is imperative that the addresses of of the values that we overwrite
```
Could you please fix all of them?

ericmutta:
@jkotas thanks for following up and spotting the extra typos. I have pushed more commits to fix them 👍

PS: if you can, please point me to the parts of the CLR where a method's IL is compiled to native code, then control is transferred to that native code.

jkotas:
> if you can, please point me to the parts of the CLR where a method's IL is compiled to native code, then control is transferred to that native code.

I think you are looking for PrestubWorker method in https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/prestub.cpp#L2341. This method is called by small piece of assembly code. This assembly code transfers the control to the native code address returned by PrestubWorker . This is Windows x64 version of that assembly code: https://github.com/dotnet/runtime/blob/main/src/coreclr/vm/amd64/ThePreStubAMD64.asm

jkotas (approved):
Thanks

ericmutta:
Many thanks @jkotas, that's exactly what I was looking for 👍

PR #94970, opened 2023-11-19
Set version in ZIP local header to ZIP64, when the file offset is >4GB

### LINKED ISSUE #94899: System.IO.Packaging.Open throws FileFormatException saying File contains corrupted data when the size of the package is more than 4GB and targeting .NET Framework
### Description

Our DacFx code uses System.IO.Packaging for packing the schema and when it is packaged with more than 4GB, it throw FileFormat Exception saying File contains corrupted data when it is used in a console app targeting .NET Framework but the same package loads fine in .NET Core

File contains corrupted data. 

Program Location:

```
at MS.Internal.IO.Zip.ZipIOLocalFileBlock.Validate(String fileName, ZipIOCentralDirectoryBlock centralDir, ZipIOCentralDirectoryFileHeader centralDirFileHeader)
at MS.Internal.IO.Zip.ZipIOLocalFileBlock.ParseRecord(BinaryReader reader, String fileName, Int64 position, ZipIOCentralDirectoryBlock centralDir, ZipIOCentralDirectoryFileHeader centralDirFileHeader)
at MS.Internal.IO.Zip.ZipIOLocalFileBlock.SeekableLoad(ZipIOBlockManager blockManager, String fileName)
at MS.Internal.IO.Zip.ZipIOBlockManager.LoadLocalFileBlock(String zipFileName)
at MS.Internal.IO.Zip.ZipArchive.GetFile(String zipFileName)
at MS.Internal.IO.Zip.ZipArchive.GetFiles()
at System.IO.Packaging.ZipPackage.ContentTypeHelper..ctor(ZipArchive zipArchive, IgnoredItemHelper ignoredItemHelper)
at System.IO.Packaging.ZipPackage..ctor(String path, FileMode mode, FileAccess access, FileShare share, Boolean streaming)
at System.IO.Packaging.Package.Open(String path, FileMode packageMode, FileAccess packageAccess, FileShare packageShare, Boolean streaming)
at Microsoft.Data.Tools.Schema.Sql.Build.SqlPackage.Artifact.InitializefromFile(FileInfo fileInfo, FileMode mode, FileAccess access)
```

### Reproduction Steps

1. Create a console app targeting **.NET core** and package a file of size greater than 4 or 6GB using System.IO.Packaging library
2. Create another console app targeting **.NET framework** and try to open the package created in step 1 using System.IO.Packaging library, it would throw File Format Exception.

### Expected behavior

It should open the package in the console app targeting .Net Framework

### Actual behavior

It throws File Format exception saying File contains corrupted data when it works fine in the console app targeting .NET core

### Regression?

_No response_

### Known Workarounds

Open the package in the console app targeting .NET Core

### Configuration

_No response_

### Other information

It has been reported here in https://github.com/dotnet/wpf/issues/2676 and because of this issue, we have an issue opened in DacFx https://github.com/microsoft/DacFx/issues/363

karakasa:
Fix #94899. Analysis in the comments. This PR fixes some compatibility issues of ZIP64 archives with certain decompressors (such as .NET Framework).

The spec is unclear whether the versionNeeded should be same in the CD and LH, and decompressors like 7-Zip & Windows don't complain about the discrepancy. Nevertheless I found 7-Zip set the LH to version ZIP64 under such circumstance.

The test is put in the outer loop because it will create a large 5GB file.

### DIFF
--- a/src/libraries/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.cs
+++ b/src/libraries/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.cs
@@ -964,7 +964,7 @@ namespace System.IO.Compression
             long finalPosition = _archive.ArchiveStream.Position;
             BinaryWriter writer = new BinaryWriter(_archive.ArchiveStream);
 
-            bool zip64Needed = SizesTooLarge()
+            bool zip64Needed = SizesTooLarge() || (_offsetOfLocalHeader > uint.MaxValue)
 #if DEBUG_FORCE_ZIP64
                 || _archive._forceZip64
 #endif
--- a/src/libraries/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.cs
+++ b/src/libraries/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.cs
@@ -490,7 +490,7 @@ namespace System.IO.Compression
             }
 
 
-            if (_offsetOfLocalHeader > uint.MaxValue
+            if (OffsetTooLarge()
 #if DEBUG_FORCE_ZIP64
                 || _archive._forceZip64
 #endif
@@ -799,6 +799,10 @@ namespace System.IO.Compression
 
         private bool SizesTooLarge() => _compressedSize > uint.MaxValue || _uncompressedSize > uint.MaxValue;
 
+        private bool OffsetTooLarge() => _offsetOfLocalHeader > uint.MaxValue;
+
+        private bool ShouldUseZIP64() => SizesTooLarge() || OffsetTooLarge();
+
         // return value is true if we allocated an extra field for 64 bit headers, un/compressed size
         private bool WriteLocalFileHeader(bool isEmptyFile)
         {
@@ -812,6 +816,9 @@ namespace System.IO.Compression
             Zip64ExtraField zip64ExtraField = default;
             bool zip64Used = false;
             uint compressedSizeTruncated, uncompressedSizeTruncated;
+
+            // save offset
+            _offsetOfLocalHeader = writer.BaseStream.Position;
 
             // if we already know that we have an empty file don't worry about anything, just do a straight shot of the header
             if (isEmptyFile)
@@ -840,7 +847,7 @@ namespace System.IO.Compression
                 {
                     // We are in seekable mode so we will not need to write a data descriptor
                     _generalPurposeBitFlag &= ~BitFlagValues.DataDescriptor;
-                    if (SizesTooLarge()
+                    if (ShouldUseZIP64()
 #if DEBUG_FORCE_ZIP64
                         || (_archive._forceZip64 && _archive.Mode == ZipArchiveMode.Update)
 #endif
@@ -864,9 +871,6 @@ namespace System.IO.Compression
                     }
                 }
             }
-
-            // save offset
-            _offsetOfLocalHeader = writer.BaseStream.Position;
 
             // calculate extra field. if zip64 stuff + original extraField aren't going to fit, dump the original extraField, because this is more important
             int bigExtraFieldLength = (zip64Used ? zip64ExtraField.TotalSize : 0)
@@ -964,7 +968,7 @@ namespace System.IO.Compression
             long finalPosition = _archive.ArchiveStream.Position;
             BinaryWriter writer = new BinaryWriter(_archive.ArchiveStream);
 
-            bool zip64Needed = SizesTooLarge() || (_offsetOfLocalHeader > uint.MaxValue)
+            bool zip64Needed = ShouldUseZIP64()
 #if DEBUG_FORCE_ZIP64
                 || _archive._forceZip64
 #endif
--- a/src/libraries/System.IO.Compression/tests/ZipArchive/zip_LargeFiles.cs
+++ b/src/libraries/System.IO.Compression/tests/ZipArchive/zip_LargeFiles.cs
@@ -47,7 +47,7 @@ namespace System.IO.Compression.Tests
         }
 
         [ConditionalFact(typeof(PlatformDetection), nameof(PlatformDetection.IsSpeedOptimized), nameof(PlatformDetection.Is64BitProcess))] // don't run it on slower runtimes
-        [OuterLoop]
+        [OuterLoop("It requires 5 GB of free disk space")]
         public static void CheckZIP64VersionIsSet_ForSmallFilesAfterBigFiles()
         {
             // issue #94899
--- a/src/libraries/System.IO.Compression/tests/ZipArchive/zip_LargeFiles.cs
+++ b/src/libraries/System.IO.Compression/tests/ZipArchive/zip_LargeFiles.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using Xunit;
 
 namespace System.IO.Compression.Tests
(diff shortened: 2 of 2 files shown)


src/libraries/System.IO.Compression/src/System/IO/Compression/ZipArchiveEntry.cs
```diff
@@ -964,7 +964,7 @@ private void WriteCrcAndSizesInLocalHeader(bool zip64HeaderUsed)
             long finalPosition = _archive.ArchiveStream.Position;
             BinaryWriter writer = new BinaryWriter(_archive.ArchiveStream);
 
-            bool zip64Needed = SizesTooLarge()
+            bool zip64Needed = SizesTooLarge() || (_offsetOfLocalHeader > uint.MaxValue)
```
stephentoub, line 967: Should the additional condition be part of SizesTooLarge? It's not clear to me from a cursory look what happens if some parts of the code think zip64 is needed and others don't.

bryanhunwardsen:
Is there a line of sight to a resolution here?

src/libraries/System.IO.Compression/tests/ZipArchive/zip_LargeFiles.cs
```diff
@@ -44,5 +45,81 @@ public static void UnzipOver4GBZipFile()
...
+                const string LargeFileName = "largefile";
+                const string SmallFileName = "smallfile";
+                const uint ZipLocalFileHeader_OffsetToVersionFromHeaderStart = 4;
+                const ushort Zip64Version = 45;
+
+                {
+                    // Create
+
+                    using var archive = new ZipArchive(fs, ZipArchiveMode.Create, true);
+                    ZipArchiveEntry file = archive.CreateEntry(LargeFileName, CompressionLevel.NoCompression);
```
danmoseley, line 73: is compression enabled an interesting variation?

danmoseley:
I think we thought we were waiting on @karakasa but I guess we're not. @carlossanlop does this change look reasonable to you - as you are best versed in the code/ the format?

ericstj:
@carlossanlop - can you please review this?

src/libraries/System.IO.Compression/tests/ZipArchive/zip_LargeFiles.cs
```diff
@@ -44,5 +45,81 @@ public static void UnzipOver4GBZipFile()
...
+                    {
+                        Assert.Fail("Cannot find the private field of _offsetOfLocalHeader in ZipArchiveEntry or the type is not long. Code may be changed after the test is written.");
+                    }
+
+                    foreach (ZipArchiveEntry entry in archive.Entries)
+                    {
+                        fs.Position = (long)offsetOfLHField.GetValue(entry) + ZipLocalFileHeader_OffsetToVersionFromHeaderStart;
+                        ushort versionNeeded = reader.ReadUInt16();
+
+                        Assert.True(versionNeeded >= Zip64Version, "ZIP64 version is not set for files with LH at >4GB offset.");
```
carlossanlop, line 113: Zip64Version is the last value in the enum: https://github.com/dotnet/runtime/blob/f0117c96ace4d475af63bce80d8afa31a740b836/src/libraries/System.IO.Compression/src/System/IO/Compression/ZipVersion.cs#L12

We should not be seeing values greater than that. If we do, this should fail because we either have a bug or we're unexpectedly getting an unsupported value.

So this should be:

```suggestion
                        Assert.Equal(Zip64Version, versionNeeded, "ZIP64 version is not set for files with LH at >4GB offset.");
```

_(I am assuming an optional string message can be passed to `Assert.Equal` as a 3rd argument, please confirm that for me.)_

carlossanlop (commented):
Thanks for your patience. Here's my review so far:

- I see that setting the value of `_offsetOfLocalHeader` a few lines earlier in `WriteLocalFileHeader` is safe as it is only used in the new invocation of `ShouldUseZIP64()`.
- I can see the new `ShouldUseZIP64()` method has the right scope for this specific issue.
- Left some suggestions for you to consider.

Please consider:

- Addressing @danmoseley 's suggestion of adding a test case that uses compression.
- Adding an extra test to `System.IO.Packaging`. We should test the original reported issue as close as possible. In the `src/libraries/System.IO.Packaging/tests/Tests.cs` file, there is a test called [VeryLargePart](https://github.com/dotnet/runtime/blob/f0117c96ace4d475af63bce80d8afa31a740b836/src/libraries/System.IO.Packaging/tests/Tests.cs#L3776), which can be used as inspiration for yours. Note that VeryLargePart is creating the zip file in-memory.
- Can you please test that the generated zip file can be read with external zip tools like 7-zip, windows explorer, winzip, unzip, etc? Testing reading the generated zip file with external programming languages would also be great.

Once merged, we can backport it to any requested active servicing branches (8.0 or 6.0).

danmoseley:
@karakasa thanks for your patience. Do you expect to be able to continue with this PR, and address the feedback?

karakasa:
> @karakasa thanks for your patience. Do you expect to be able to continue with this PR, and address the feedback?

My apologies I went through a surgery. I will look into the comments above in the following days. Or you may close this PR if someone else chooses to work on this. It's up to you. Many thanks.

src/libraries/System.IO.Compression/tests/ZipArchive/zip_LargeFiles.cs
karakasa: `Assert.Equal` doesn't have a 3 param overload. I've changed the line into:

    Assert.True(versionNeeded == Zip64Version, "Version is not ZIP64 for files with Local Header at >4GB offset.");

src/libraries/System.IO.Compression/tests/ZipArchive/zip_LargeFiles.cs
karakasa: > @karakasa would you mind adding a test that compresses? We just need to make sure the resulting compressed `largefile` should be larger than 4GB too.

Sure. But per my test, the compressed scenario is extremely time-consuming, even for `CompressionLevel.Fastest`.

karakasa:
@carlossanlop

> * Addressing @danmoseley 's suggestion of adding a test case that uses compression.

Sure.

> * Adding an extra test to `System.IO.Packaging`. We should test the original reported issue as close as possible. In the `src/libraries/System.IO.Packaging/tests/Tests.cs` file, there is a test called [VeryLargePart](https://github.com/dotnet/runtime/blob/f0117c96ace4d475af63bce80d8afa31a740b836/src/libraries/System.IO.Packaging/tests/Tests.cs#L3776), which can be used as inspiration for yours. Note that VeryLargePart is creating the zip file in-memory.

May I ask how can I achieve this? I believe it's essentially executing a single test, while the first part runs in the latest .NET and the second part runs in the .NET Framework.

The `VeryLargePart` test you referenced doesn't address such problem.



> * Can you please test that the generated zip file can be read with external zip tools like 7-zip, windows explorer, winzip, unzip, etc? Testing reading the generated zip file with external programming languages would also be great.

I have tested. Is replying this PR with images sufficient?

carlossanlop:
> I have tested. Is replying this PR with images sufficient?

Yes, absolutely. And thank you for doing that.

> May you elaborate how to achieve that?

My suggestion is that you open `src/libraries/System.IO.Packaging/tests/Tests.cs`, duplicate the `VeryLargePart`, test, rename it, and then adjust it slightly to ensure it touches the code you're fixing. Let me know if you have additional questions.

carlossanlop:
@karakasa please let us know if you'll be able to continue looking into this. Otherwise, I can take over.

karakasa:
> @karakasa please let us know if you'll be able to continue looking into this. Otherwise, I can take over.

Go ahead. Apologize I don’t have resources recently.

carlossanlop:
No worries, @karakasa. Thanks anyway for helping move this along.

PR #94982, opened 2023-11-19
[mono][wasm] Fix broken VM suspend while emitting events in Debug configuration

seclerp:
Currently, when building Mono Wasm runtime from sources using
```bash
./build.sh mono+libs -c debug -os browser
```
And using debugger agent, any suspend-related event causes crash:
```
blazor.webassembly.js:1 RuntimeError: null function or function signature mismatch
    at process_event (0434295e:0x19b1e2)
    at process_profiler_event (0434295e:0x19475f)
    at emit_type_load (0434295e:0x1cdc6a)
    at send_type_load (0434295e:0x1cd9c2)
    at jit_end (0434295e:0x194cb8)
    at jit_done (0434295e:0x1957b9)
    at mono_profiler_raise_jit_done (0434295e:0x2ffaee)
    at mono_interp_transform_method (0434295e:0x4684f)
    at do_transform_method (0434295e:0x21def)
    at mono_interp_exec_method (0434295e:0x218dd)
```

After a little bit of investigation I found that the error is related to `suspend_vm_func` function pointer which is not properly initialized in regular Wasm (Blazor) apps.

image_2023-11-19_15-27-11

Previously, `suspend_vm` has been used directly, so it worked correctly.

Please note that it's only reproducible in Debug builds. Seems like Release one just ignores it.

So I propose to use the same function pointers as before for Wasm apps.

### DIFF
--- a/src/mono/mono/component/debugger-agent.c
+++ b/src/mono/mono/component/debugger-agent.c
@@ -1636,6 +1636,10 @@ mono_init_debugger_agent_for_wasm (int log_level_parm, MonoProfilerHandle *prof)
 
 	agent_config.enabled = TRUE;
 
+	start_debugger_thread_func = start_debugger_thread;
+	suspend_vm_func = suspend_vm;
+	suspend_current_func = suspend_current;
+
 	mono_init_debugger_agent_common (prof);
 }
 


src/mono/mono/component/debugger-agent.c
```diff
@@ -1636,6 +1636,10 @@ mono_init_debugger_agent_for_wasm (int log_level_parm, MonoProfilerHandle *prof)
 
 	agent_config.enabled = TRUE;
 
+	start_debugger_thread_func = start_debugger_thread;
+	suspend_vm_func = suspend_vm;
```
lambdageek, line 1640: Not sure about this one: it's main side-effect looks like it will enable single-stepping in the current (only) thread. I see we have `mono_wasm_single_step_hit` as a single step trampoline for wasm, so that seems ok.  But I'm not sure if we expect to be in single-step mode on wasm after various events.  /cc @thaystg

lambdageek (commented):
For single-threaded wasm two of the three callbacks seem like they'll hang the runtime.  The third seems like it will work but maybe it puts the runtime into single-stepping mode when it is not expected?

I'm not sure about multi-threaded wasm - maybe @thaystg can clarify

lambdageek:
@seclerp About this PR and https://github.com/dotnet/runtime/pull/95002.  I hesitate to take them because it seems like you're doing something that is quite different from how our current WASM debugging works and it doesn't seem like we hit the RuntimeErrors that you're attempting to fix in our testing, it would be good if you could include some tests that exercise your changes.  (Perhaps as an additional testsuite for our debugger tests in https://github.com/dotnet/runtime/tree/main/src/mono/wasm/debugger)

seclerp:
My experiments are different from what debug proxy is currently offering. But I also think that these asserts and RuntimeErrors should be addressed, as they are a part of API that common Mono debugger agent offers to debugging clients.

Unfortunately, It's quite hard to prepare a public repro project (otherwise It will be attached immediately), but I will try to make miminal repro.

src/mono/mono/component/debugger-agent.c
```diff
@@ -1636,6 +1636,10 @@ mono_init_debugger_agent_for_wasm (int log_level_parm, MonoProfilerHandle *prof)
 
 	agent_config.enabled = TRUE;
 
+	start_debugger_thread_func = start_debugger_thread;
```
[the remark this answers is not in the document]
seclerp: The main problem was in `suspend_vm_func`, but I assumed that some other pointers will be required as well. But they are not mandatory in this case.

src/mono/mono/component/debugger-agent.c
seclerp: It is used here
https://github.com/dotnet/runtime/blob/34544b9416eb46c1397bc76dde257b4f5209d1a1/src/mono/mono/component/debugger-agent.c#L3768
I'm not 100% sure but I assume that it's because on the client side, I enable type-loading events. The common implementation is then offering a suspend policy different from NONE.

thaystg:
The way that we implemented browserdebugproxy, we should never receive an event that has `suspend_policy != SUSPEND_POLICY_NONE` in a wasm debugger scenario.

When we enable a breakpoint or a step or an execpetion, we set the suspend policy to none:
https://github.com/dotnet/runtime/blob/777f9d90edd39a35f791839fc3504ce162c6f30a/src/mono/wasm/debugger/BrowserDebugProxy/MonoSDBHelper.cs#L1532
https://github.com/dotnet/runtime/blob/777f9d90edd39a35f791839fc3504ce162c6f30a/src/mono/wasm/debugger/BrowserDebugProxy/MonoSDBHelper.cs#L1560C14-L1560C14
https://github.com/dotnet/runtime/blob/777f9d90edd39a35f791839fc3504ce162c6f30a/src/mono/wasm/debugger/BrowserDebugProxy/MonoSDBHelper.cs#L2278

seclerp:
@thaystg thanks! I have a couple of questions:
1. Is it true that instead of `VM_RESUME` the debug proxy resumes JS execution (`Debugger.resume`) and because of the single-thread environment it just unblocks execution?
1. Is it true that all event types (even ones not currently used by debug proxy) are working in such a way? 
1. How (possible) multi-threaded environment would work? For example, with `wasm-threads`.

thaystg:
> @thaystg thanks! I have a couple of questions:
> 
> 1. Is it true that instead of `VM_RESUME` the debug proxy resumes JS execution (`Debugger.resume`) and because of the single-thread environment it just unblocks execution?
- right
> 2. Is it true that all event types (even ones not currently used by debug proxy) are working in such a way?
- no, event types that are not being used be debug proxy, they are ignored so, it will never call process_event for this event types.
> 3. How (possible) multi-threaded environment would work? For example, with `wasm-threads`.
- in a similar way, the `debugger;` statement stops the execution of that thread and when we `Debugger.resume` the thread continues its execution.

seclerp:
Regarding 2 - they are ignored, but if we imagine they are required for some feature in the future, will theu also be enabled with `SUSPEND_POLICY=NONE`?

thaystg:
> Regarding 2 - they are ignored, but if we imagine they are required for some feature in the future, will theu also be enabled with `SUSPEND_POLICY=NONE`?

Correct!

seclerp:
Thank you! After a bit of testing and setting suspend policy to None everything is working fine without the fixes above. Will close this and https://github.com/dotnet/runtime/pull/95002.

PR #94986, opened 2023-11-20
Redirect stderr to stdout when invoking merged tests

BruceForstall:
Otherwise, only stdout gets logged to the test output .log file and stderr gets written to the console.

### DIFF
--- a/src/tests/Common/tests.targets
+++ b/src/tests/Common/tests.targets
@@ -80,7 +80,7 @@
 
   <Target Name="RunSingleMergedTest">
     <Exec Command="chmod +x $(TestWrapperScript)" WorkingDirectory="$(BaseOutputPathWithConfig)" EchoOff="true" Condition="'$(TargetOS)' != 'windows'" />
-    <Exec Command="$(TestWrapperScript) &gt;$(RedirectOutputToFile)" WorkingDirectory="$(BaseOutputPathWithConfig)" Timeout="$(__TestTimeout)" IgnoreExitCode="true">
+    <Exec Command="$(TestWrapperScript) &gt;$(RedirectOutputToFile) 2&gt;&amp;1" WorkingDirectory="$(BaseOutputPathWithConfig)" Timeout="$(__TestTimeout)" IgnoreExitCode="true">
       <Output TaskParameter="ExitCode" PropertyName="MergedTestExitCode" />
     </Exec>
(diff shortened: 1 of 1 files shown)


markples:
I don't understand how these scripts fit together.  In the line that you modified, I see that stdout is already redirected to a file, but if I look at a test leg, I see stdout output in console output.  For example, take the beginning of the merged group execution in the failing leg here - https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-94986-merge-680801a179d64e8986/Interop/1/console.a5635680.log?helixlogtype=result -

```
C:\h\w\9AC80876\w\9E55089D\e>call Interop\Interop\Interop.cmd -usewatcher 
BEGIN EXECUTION
"C:\h\w\9AC80876\p\watchdog.exe" 29 "C:\h\w\9AC80876\p\corerun.exe" -p "System.Reflection.Metadata.MetadataUpdater.IsSupported=false" -p "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization=true"  Interop.dll 
05:55:38.232 Running test: Interop/ICastable/Castable/Castable.dll
IsInstanceOfInterface has been called for type IRetThis
...
```

I don't see the redirection here, so it looks like something else is invoking the merged group's `.cmd` file.  In that case, we don't appear to redirect anything.  Do you know how this works and why one should redirect and other shouldn't?

BruceForstall:
I don't understand the difference. I'm guessing the code I modified is only invoked when running the tests locally using `run.py`, and not in the CI system (which as you know does things very differently from local runs).

When I run locally using `run.py`, now I first get a bunch of lines like:
```
C:\gh\runtime3\artifacts\tests\coreclr\windows.x64.Checked\JIT\jit64\jit64_2\jit64_2.cmd >C:\gh\runtime3\artifacts\tests\coreclr\windows.x64.Checked\JIT\jit64\jit64_2\jit64_2.log 2>&1
```

Note the trailing `2>&1`, where previously it wasn't there.

After all these comes the XUnitConsoleRunner output (for the unconverted tests?)

trylek:
IIRC this change only affects local test executions, not Helix executions in lab pipelines, so it seems reasonable to me.

trylek (approved):
LGTM, thank you!

PR #94942, opened 2023-11-18
P/Invokes should use C types not C++.

AaronRobinsonMSFT:
A recent C++ toolchain update uncovered an invalid P/Invoke signature using C++ types as opposed to C types. This caused the test to fail due to invalid assumptions. The test was updated to be valid.

Fixes #94931

PR #94987, opened 2023-11-20
Add ABI note about small return types

jkotas:


jkotas:
Follow up on https://github.com/dotnet/runtime/pull/94942#issuecomment-1817312110

docs/design/coreclr/botr/clr-abi.md
```diff
@@ -113,6 +113,10 @@ ARM64-only: When a method returns a structure that is larger than 16 bytes the c
 
 *Normal PInvoke* - The VM shares IL stubs based on signatures, but wants the right method to show up in call stack and exceptions, so the MethodDesc for the exact PInvoke is passed in the (x86) `EAX` / (AMD64) `R10` / (ARM, ARM64) `R12` (in the JIT: `REG_SECRET_STUB_PARAM`). Then in the IL stub, when the JIT gets `CORJIT_FLG_PUBLISH_SECRET_PARAM`, it must move the register into a compiler temp. The value is returned for the intrinsic `NI_System_StubHelpers_GetStubContext`.
 
+## x86 and AMD64 only: Small primitive returns
+
+Primitive value types smaller than 32-bits must be zero-extended or sign-extended to 32-bits. This is different from standard calling convention that leaves the state of unused bits in the value returned in `EAX`/`RAX` undefined.
```
jakobbotsch, line 118:
```suggestion
## Small primitive returns

Primitive value types smaller than 32-bits are normalized to 32-bits: signed small types are sign extended and unsigned small types are zero extended. This can be different from the standard calling conventions that may leave the state of unused bits in the return register undefined.
```

This is also different on arm64.

jkotas:
> This is also different on arm64.

If it is really different, this ifdef is wrong https://github.com/dotnet/runtime/blob/57085686bf7223e2305720383f4e00695321919a/src/coreclr/vm/fcall.h#L1335-L1348 and we are lucky that it has not showed up as an observable bug yet.

jakobbotsch:
> > This is also different on arm64.
> 
> If it is really different, this ifdef is wrong
> 
> https://github.com/dotnet/runtime/blob/57085686bf7223e2305720383f4e00695321919a/src/coreclr/vm/fcall.h#L1335-L1348
> 
> and we are lucky that it has not showed up as an observable bug yet.

For example: https://godbolt.org/z/59YMfWfh5

jakobbotsch (approved):

jakobbotsch (approved):

Issue #54110, opened 2021-06-13
ilverify fails with various file not found errors


AArnott:
After `dotnet tool update --global dotnet-ilverify --version 5.0.0`, any invocation of `ilverify` that includes a dll (I tried several) fails with:

```
Error: Internal.IL.VerifierException: Assembly or module not found: mscorlib
   at ILVerify.Verifier.SetSystemModuleName(AssemblyName name)
   at ILVerify.Program.Run()
   at ILVerify.Program.Run(Options options)
```

The latest prerelease also fails, but with a different message:
```
dotnet tool install --global dotnet-ilverify --version 6.0.0-preview.4.21253.7
```

```
Error:
  An assembly specified in the application dependencies manifest (ILVerify.deps.json) was not found:
    package: 'System.Collections.Immutable', version: '5.0.0'
    path: 'lib/netstandard2.0/System.Collections.Immutable.dll'
```

jkotas:
> Error: Internal.IL.VerifierException: Assembly or module not found: mscorlib

What is the command line that you are using to invoke ilverify?

ilverify requires explicitly all references, including mscorlib, to be explicitly specified on the command line. It looks like that mscorlib reference is not specified. https://github.com/dotnet/runtime/tree/main/src/coreclr/tools/ILVerify#how-to-use-ilverify

AArnott:
I only specified the one assembly on the command line. I had no idea I had to specify all its references too. Can that be added to the `-?` usage help? It didn't mention it. What's more, peverify.exe doesn't share that requirement, so it would be great to call it out.

lovettchris:
Yeah, I found this confusing too until I found the "-r" command line argument.  I found it to be quite difficult to find all the right folders, and it would be cool if ilverify provided an option that automatically does all this:

```powershell
function FindDotNet($dotnet) {
    $dotnet_path = $ENV:PATH.split([System.IO.Path]::PathSeparator) | ForEach-Object {
        $test = Join-Path -Path $_ -ChildPath $dotnet
        if ($IsWindows) {
            $test = $test + ".exe"
        }
        if (Test-Path -Path $test) {
            return $_
        }
        $candidate = [System.IO.Path]::Combine($_, "dotnet")
        if (Test-Path -Path $candidate) {
            $candidate = TraverseLink($candidate)
            if (Test-Path -Path $candidate -PathType Leaf) {
                $candidate = [System.IO.Path]::GetDirectoryname($candidate)
            }
            return $candidate
        }
[...]

PR #94988, opened 2023-11-20
Fix null reference exception in ilverify

jkotas:
Invoking ilverify with empty list of references lead to null reference exception.

Contributes to https://github.com/dotnet/runtime/issues/54110#issuecomment-1817961976

### DIFF
--- a/src/coreclr/tools/ILVerify/Program.cs
+++ b/src/coreclr/tools/ILVerify/Program.cs
@@ -34,7 +34,7 @@ namespace ILVerify
             _command = command;
 
             _inputFilePaths = Get(command.InputFilePath);
-            _referenceFilePaths = Get(command.Reference);
+            _referenceFilePaths = Get(command.Reference) ?? new Dictionary<string, string>();
             _verbose = Get(_command.Verbose);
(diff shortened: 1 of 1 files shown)


src/coreclr/tools/ILVerify/Program.cs
```diff
@@ -34,7 +34,7 @@ public Program(ILVerifyRootCommand command)
             _command = command;
 
             _inputFilePaths = Get(command.InputFilePath);
-            _referenceFilePaths = Get(command.Reference);
+            _referenceFilePaths = Get(command.Reference) ?? new Dictionary<string, string>();
```
am11, line 37: Ops, sorry about this. It was a breaking change in a later version of command-line-api. The recommended way in that breaking change was how we handled default case in crossgen2 and ilc: https://github.com/dotnet/runtime/commit/af6e7fe132ada839b61bfecdc355354e876323f8

jkotas: Could you please submit a PR with the fix you are proposing? (I am going to close this one.)

PR #94991, opened 2023-11-20
Replace LowLevelLock with Lock

MichalStrehovsky:
Context: https://github.com/dotnet/runtime/pull/94873#issuecomment-1818060075

Cc @dotnet/ilc-contrib


VSadov (approved):
LGTM

MichalStrehovsky:
> I do not think that this is a good change, and it should be reverted. This change can lead to reentrancy crashes if user code overrides `SynchronizationContext.Wait` and the override attempts to create `System.RuntimeType` instance (e.g. by calling `object.GetType()).
> 
> [#94873 (comment)](https://github.com/dotnet/runtime/pull/94873#issuecomment-1820951222) has more context.

Revert in #95071.

Issue #88733, opened 2023-07-12
Setting Size on MemoryCacheEntry after insertion causes faulty CurrentEstimatedSize


Lilk:
### Description

If holding onto a Microsoft.Extensions.Caching.Memory.CacheEntry, in order to update the size after it is known (If e.g. caching AsyncLazy values, and needing to update the size once the value has been materialized), the overall CurrentEstimatedSize is not updated.  

If the size is updated to a smaller value than the initally set value, this will lead to emptying of the cache and busy-loop of compaction upon reaching the size limit. 

Seen here is the result of a compaction with the parameter 0.5, causing half of the entries being dropped (lower graph), but seemingly no to little impact on the reported size. 
![image](https://github.com/dotnet/runtime/assets/710640/6f5a21c2-55b4-4f12-8e2c-1ef2545ab1da)

As the overcapacitycompaction scheduling is based on this metric, such a scenario will cause repeated scheduling of OvercapacityCompaction (for every insertion), and will cause a busy CPU loop, quickly dumping all of the entries in the cache, but never stopping the compaction cycle. See here graphs for patterns of entry count vs estimated size as well as cache misses.

![image](https://github.com/dotnet/runtime/assets/710640/8a81be82-32cc-42c3-824f-887f723e7a38)

And the CPU busy iterating the coherent state:

![image](https://github.com/dotnet/runtime/assets/710640/265ee8f5-0016-46e3-8672-0018141b3b1c)

I am at this point unsure of the symptoms if Size were to be update to a higher value, the reported CurrentEstimatedSize  would be less than the actual size, but what effects would it have to the MemoryCache, what effects will take place if the CurrentEstimatedSize  is less than zero?

### Reproduction Steps
```
// See https://aka.ms/new-console-template for more information
using Microsoft.Extensions.Caching.Memory;

Console.WriteLine("Hello, World!");
var _cache = new MemoryCache(new MemoryCacheOptions()
{
    TrackStatistics = true,
    ExpirationScanFrequency = TimeSpan.FromMinutes(1),
    CompactionPercentage = 0.1,
    SizeLimit = 10,

});

Task task = null;

_cache.GetOrCreate("test", (ICacheEntry entry) =>
{
    entry.Size = 4;
    task = Task.Delay(1000);
    task.ContinueWith((t) =>
    {
        try
        {
            entry.Size = 2;
        }
        catch (Exception e)
        {
            Console.WriteLine(e.Message);
        }
        return;
    });

    return new object();
});

[...]

PR #94992, opened 2023-11-20
extend MemoryCache with `ICollection<object> Keys` property

### LINKED ISSUE #36554: Will there be GetEnumerator() for MemoryCache ?
**LINK to the proposal**: https://github.com/dotnet/runtime/issues/36554#issuecomment-1818437941

Just wonder if there would be GetEnumerator() for MemoryCache, as supported in System.Runtime.Caching.Memorycache?

adamsitnik:
fixes #36554

### DIFF
--- a/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
+++ b/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
@@ -77,14 +77,13 @@ namespace Microsoft.Extensions.Caching.Memory
         public int Count => _coherentState.Count;
 
         /// <summary>
-        /// Gets a snapshot containing all the keys in the <see cref="MemoryCache"/>.
+        /// Gets an enumerable of the all the keys in the <see cref="MemoryCache"/>.
         /// </summary>
         public IEnumerable<object> Keys
         {
             get
             {
-                CoherentState coherentState = _coherentState; // Clear() can update the reference in the meantime
-                foreach (KeyValuePair<object, CacheEntry> pairs in coherentState._entries)
+                foreach (KeyValuePair<object, CacheEntry> pairs in _coherentState._entries)
                 {
                     yield return pairs.Key;
                 }
--- a/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
+++ b/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
@@ -79,7 +79,17 @@ namespace Microsoft.Extensions.Caching.Memory
         /// <summary>
         /// Gets a snapshot containing all the keys in the <see cref="MemoryCache"/>.
         /// </summary>
-        public ICollection<object> Keys => _coherentState._entries.Keys;
+        public IEnumerable<object> Keys
+        {
+            get
+            {
+                CoherentState coherentState = _coherentState; // Clear() can update the reference in the meantime
+                foreach (KeyValuePair<object, CacheEntry> pairs in coherentState._entries)
+                {
+                    yield return pairs.Key;
+                }
+            }
+        }
 
         /// <summary>
         /// Internal accessor for Size for testing only.
--- a/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
+++ b/src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
-using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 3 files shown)


src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
```diff
@@ -76,6 +76,21 @@ public MemoryCache(IOptions<MemoryCacheOptions> optionsAccessor, ILoggerFactory
...
 
+        /// <summary>
+        /// Gets a snapshot containing all the keys in the <see cref="MemoryCache"/>.
+        /// </summary>
+        public IEnumerable<object> Keys
+        {
+            get
+            {
+                CoherentState coherentState = _coherentState; // Clear() can update the reference in the meantime
+                foreach (KeyValuePair<object, CacheEntry> pairs in coherentState._entries)
```
stephentoub, line 87: Why is the coherentState local needed?

src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
```diff
@@ -76,6 +76,21 @@ public MemoryCache(IOptions<MemoryCacheOptions> optionsAccessor, ILoggerFactory
         /// </summary>
         public int Count => _coherentState.Count;
 
+        /// <summary>
+        /// Gets a snapshot containing all the keys in the <see cref="MemoryCache"/>.
```
stephentoub, line 80: It's not a snapshot.

src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
adamsitnik, line 87:
```suggestion
                foreach (KeyValuePair<object, CacheEntry> pairs in _coherentState._entries)
```

src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
adamsitnik, line 80:
```suggestion
        /// Gets an enumerable of the all the keys in the <see cref="MemoryCache"/>.
```

src/libraries/Microsoft.Extensions.Caching.Memory/tests/EnumerationTests.cs
```diff
@@ -0,0 +1,43 @@
...
+            for (int i = 0; i < KeyCount; i++)
+            {
+                cache.Set(i.ToString(), i);
+            }
+
+            foreach (object key in cache.Keys)
+            {
+                cache.Remove(key);
+            }
+            Assert.Equal(0, cache.Count);
```
Jozkee, line 40:
```suggestion
            int iterationCount = 0;
            foreach (object key in cache.Keys)
            {
                cache.Remove(key);
                iterationCount++;
            }
            Assert.Equal(0, cache.Count);
            Assert.Equal(KeyCount, iterationCount);
```

src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
```diff
@@ -76,6 +76,20 @@ public MemoryCache(IOptions<MemoryCacheOptions> optionsAccessor, ILoggerFactory
         /// </summary>
         public int Count => _coherentState.Count;
 
+        /// <summary>
+        /// Gets an enumerable of the all the keys in the <see cref="MemoryCache"/>.
+        /// </summary>
+        public IEnumerable<object> Keys
```
Jozkee, line 82: Did you consider returning `ICacheEntry` rather than just `Keys`? Is my understanding correct that some people want to make decisions based on `ICacheEntry.Value` https://github.com/dotnet/extensions/issues/4766#issuecomment-1845802796?

Jozkee: This can also be achieved by implementing `IEnumerable<KeyValuePair<object, ICacheEntry>>` as you mentioned in the proposal, but doing it in the property doesn't clutter the IDE with LINQ.

adamsitnik: I did (https://github.com/dotnet/runtime/issues/36554#issuecomment-1818437941):

> that would allow to mutate cache entries (values) after adding them to the cache, which is not supported and can lead into issues like #88733.

src/libraries/Microsoft.Extensions.Caching.Memory/tests/EnumerationTests.cs
Jozkee: Checks that the number of keys stays the same. Feel free to ignore.

src/libraries/Microsoft.Extensions.Caching.Memory/src/MemoryCache.cs
Jozkee: Good point, I didn't notice ICacheEntry had writable properties.

jozkee (approved):

PR #94993, opened 2023-11-20
Improve ConcurrentDictionary xml docs

adamsitnik:
After reading the discussion in https://github.com/dotnet/runtime/pull/93870#discussion_r1368745090 I've realized that `ConcurrentDictionary.Keys` and `Values` properties don't mention anything about them being snapshots. Moreover, some xml docs were referring to `this` as `Dictionary` rather than `ConcurrentDictionary`

### DIFF
--- a/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs
+++ b/src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs
@@ -1545,7 +1545,7 @@ namespace System.Collections.Concurrent
         /// <summary>
         /// Gets a snapshot containing all the keys in the <see cref="ConcurrentDictionary{TKey,TValue}"/>.
         /// </summary>
-        /// <remarks>It's a standalone copy of all the keys. It's not kept in sync with <see cref="ConcurrentDictionary{TKey,TValue}"/>.</remarks>
+        /// <remarks>The property returns a copy of all the keys. It's not kept in sync with <see cref="ConcurrentDictionary{TKey,TValue}"/>.</remarks>
         public ICollection<TKey> Keys => GetKeys();
 
         /// <inheritdoc cref="Keys"/>
@@ -1554,7 +1554,7 @@ namespace System.Collections.Concurrent
         /// <summary>
         /// Gets a snapshot containing all the values in the <see cref="ConcurrentDictionary{TKey,TValue}"/>.
         /// </summary>
-        /// <remarks>It's a standalone copy of all the values. It's not kept in sync with <see cref="ConcurrentDictionary{TKey,TValue}"/>.</remarks>
+        /// <remarks>The property returns a copy of all the values. It's not kept in sync with <see cref="ConcurrentDictionary{TKey,TValue}"/>.</remarks>
         public ICollection<TValue> Values => GetValues();
 
         /// <inheritdoc cref="Values"/>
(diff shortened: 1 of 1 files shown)


src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs
```diff
@@ -1660,11 +1645,11 @@ public bool IsEmpty
         /// elements.</exception>
         /// <exception cref="ArgumentException">
         /// <paramref name="key"/> is of a type that is not assignable to the key type <typeparamref
-        /// name="TKey"/> of the <see cref="Dictionary{TKey,TValue}"/>. -or-
+        /// name="TKey"/> of the <see cref="ConcurrentDictionary{TKey,TValue}"/>. -or-
```
eiriktsarpalis, line 1648: Shouldn't this just be `<inheritdoc />`?

eiriktsarpalis (approved):
LGTM, although I don't think we should try to document explicit interface implementations. Should this be ported to dotnet-api-docs as well?

src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs
```diff
@@ -1543,36 +1543,21 @@ public bool IsEmpty
         bool IDictionary<TKey, TValue>.Remove(TKey key) => TryRemove(key, out _);
 
         /// <summary>
-        /// Gets a collection containing the keys in the <see
-        /// cref="Dictionary{TKey,TValue}"/>.
+        /// Gets a snapshot containing all the keys in the <see cref="ConcurrentDictionary{TKey,TValue}"/>.
         /// </summary>
-        /// <value>An <see cref="ICollection{TKey}"/> containing the keys in the
-        /// <see cref="Dictionary{TKey,TValue}"/>.</value>
+        /// <remarks>It's a standalone copy of all the keys. It's not kept in sync with <see cref="ConcurrentDictionary{TKey,TValue}"/>.</remarks>
```
stephentoub, line 1548:
```suggestion
        /// <remarks>The property returns a copy of all the keys. It's not kept in sync with <see cref="ConcurrentDictionary{TKey,TValue}"/>.</remarks>
```

src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs
```diff
@@ -1543,36 +1543,21 @@ public bool IsEmpty
...
 
         /// <summary>
-        /// Gets a collection containing the values in the <see
-        /// cref="Dictionary{TKey,TValue}"/>.
+        /// Gets a snapshot containing all the values in the <see cref="ConcurrentDictionary{TKey,TValue}"/>.
         /// </summary>
-        /// <value>An <see cref="ICollection{TValue}"/> containing the values in
-        /// the
-        /// <see cref="Dictionary{TKey,TValue}"/>.</value>
+        /// <remarks>It's a standalone copy of all the values. It's not kept in sync with <see cref="ConcurrentDictionary{TKey,TValue}"/>.</remarks>
```
stephentoub, line 1557:
```suggestion
        /// <remarks>The property returns a copy of all the values. It's not kept in sync with <see cref="ConcurrentDictionary{TKey,TValue}"/>.</remarks>
```

stephentoub (approved):

src/libraries/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentDictionary.cs
adamsitnik: My understanding is that It could be if `IDictionary.Add` was providing xml docs. But it does not:

https://github.com/dotnet/runtime/blob/a9cdcd5716424c557d7a135d65b7ab88803de1dd/src/libraries/System.Private.CoreLib/src/System/Collections/IDictionary.cs#L30-L31

cc @carlossanlop

PR #94999, opened 2023-11-20
WIP: Testing iOS library mode with NativeAOT on CI

ivanpovazan:
Opening this PR just to test builds on Helix and to avoid polluting the original PR including the fix.


src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
```diff
@@ -134,13 +179,15 @@ The .NET Foundation licenses this file to you under the MIT license.
       <LinkerArg Include="-gz=zlib" Condition="'$(CompressSymbols)' != 'false'" />
       <LinkerArg Include="-fuse-ld=$(LinkerFlavor)" Condition="'$(LinkerFlavor)' != ''" />
       <LinkerArg Include="@(NativeLibrary)" />
-      <LinkerArg Include="--sysroot=$(SysRoot)" Condition="'$(SysRoot)' != ''" />
-      <LinkerArg Include="--target=$(TargetTriple)" Condition="'$(_IsApplePlatform)' != 'true' and '$(TargetTriple)' != ''" />
+      <LinkerArg Include="--sysroot=$(SysRoot)" Condition="'$(SysRoot)' != '' and '$(_iOSLibraryMode)' != 'true'" />
+      <LinkerArg Include="-isysroot $(SysRoot)" Condition="'$(SysRoot)' != '' and '$(_iOSLibraryMode)' == 'true'" />
```
jkotas, line 183: Should the sysroot conditions use `_IsApplePlatform` instead?

ivanpovazan: This effort is mainly done to have parity with Mono's library mode builds which only cover Apple mobile platforms. Is there a need to support macOS in the same fashion?

jkotas: We try to use logically correct conditions in situations like this, even when less logically correct condition produces same result for supported configurations.

Why do we need to differentiate between isysroot and sysroot here?

ivanpovazan: > Why do we need to differentiate between isysroot and sysroot here?

These switches seem to have different behaviours when cross-compiling.
- If we try to compile an iOS library with:
    ```
    --sysroot=/Applications/Xcode_15.0.0.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS17.0.sdk --target=arm64-apple-ios11.0-macho
    ```

    Clang produces the following warning:
    ```
    clang: warning: using sysroot for 'MacOSX' but targeting 'iPhone' [-Wincompatible-sysroot]
    ```

- If we use `-isysroot` (what Xcode is doing) with:
    ```
    -isysroot /Applications/Xcode_15.0.0.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS17.0.sdk --target=arm64-apple-ios11.0-macho
    ```
    Clang does not produce any warning.

Clang documentation does not provide much information:
- `sysroot` - no description - https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-sysroot
- `isysroot` - describes that the switch sets the system root directory - https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-isysroot-dir

Gcc documentation has a bit more information on this (so I suppose clang tries to mimic this behaviour with some internal caveats):
- `sysroot` - describes that the switch sets the logical root directory for includes and libraries, while when combined with `isysroot` then it sets the logical root directory only for libraries while `isysroot` sets the logical root directory for includes - https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html#index-sysroot
- `isysroot` - describes that the switch sets the logical root directory for includes (with the exception that on Darwin targets it applies to both includes and libraries) - https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html#index-isysroot

FWIW Xcode always seems to specify only `isysroot` when compiling a UI app for iOS or MacOS, or a console app for MacOS

---

> We try to use logically correct conditions in situations like this, even when less logically correct condition produces same result for supported configurations.

One solution would be to always use `isysroot` for all Apple platforms (due to above-mentioned differences) and extend the logic of finding the sdk path and target triple for MacOS for all cases (not just library mode) as well.

src/tasks/AppleAppBuilder/Xcode.cs
```diff
@@ -259,6 +259,16 @@ public void CreateXcodeProject(string projectName, string cmakeDirectoryPath)
...
         var predefinedExcludes = new List<string> { ".dll.o", ".dll.s", ".dwarf", ".m", ".h", ".a", ".bc", "libmonosgen-2.0.dylib", "libcoreclr.dylib", "icudt*" };
+        if (useNativeAOTRuntime)
+        {
+            predefinedExcludes.Add(".dll");
+            predefinedExcludes.Add(".pdb");
+            predefinedExcludes.Add(".json");
+            predefinedExcludes.Add(".txt");
+            predefinedExcludes.Add(".bin");
+            predefinedExcludes.Add(".dSYM");
+        }
```
ivanpovazan, line 270: Note to myself: these shouldn't be needed once we carefully construct the publish folder on Helix

ivanpovazan:
All relevant tests are passing.

src/coreclr/nativeaot/BuildIntegration/Microsoft.NETCore.Native.Unix.targets
jkotas: > One solution would be to always use isysroot for all Apple platforms (due to above-mentioned differences)

Yes.

> extend the logic of finding the sdk path and target triple for MacOS for all case

I do not think we need to set isysroot for MacOS. The default one works fine for MacOS. However, it should be possible for somebody to set isysroot e.g. if they have multiple xcode versions on their machine and want to use specific one.

ivanpovazan: Thank you for the confirmation.

Issue #18632, opened 2016-09-19
HttpWebRequest.GetRequestStreamAsync doesn't open the connection to the server - we always buffer the content


stephentoub:
The corefx implementation of HttpWebRequest.GetRequestStreamAsync effectively just returns an in-memory stream.  The connection to the server isn't made and the data isn't uploaded until Stream.Dispose is called on the that Stream.  This is problematic for several reasons, including that the actual asynchronous operation ends up happening in the synchronous Dispose call on the stream, and deadlocks can easily happen if code assumes that the server will respond once the supplied data has been written (but before Dispose is called).

liveans:
Fixed by https://github.com/dotnet/runtime/pull/95001

PR #95001, opened 2023-11-20
Implement HttpWebRequest AllowWriteStreamBuffering property

### LINKED ISSUE #60993: `HttpWebRequest.AllowWriteStreamBuffering` seems to lack an implementation
### Description

There is a properly on `HttpWebRequest` called `AllowWriteStreamBuffering`. However, setting it doesn't seem to have any effect; the request body is still buffered to a `MemoryStream` before the request is sent. This makes it rather difficult to do things like upload massive files via HTTP PUT or POST.

The following comment seems to indicate that the functionality of unbuffered write streams, as originally implemented in the .NET Framework, was torn out of the code as part of .NET Core development because .NET Core's `HttpWebRequest` lacked an `AllowWriteStreamBuffering`:

https://github.com/dotnet/runtime/blob/56135728983a635dad49441728d2c30e200eb575/src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs#L11-L15

...and then it was never reintroduced, even though the property was reintroduced in .NET Core 2.0 (according to the documentation).

In my testing, running against .NET 5.0, extremely large files cannot be transferred at all, because regardless of the value of `AllowWriteStreamBuffering`, trying to write them to the request body stream produces the following exception:
```
System.IO.IOException: Stream was too long.
```
This is an exception produced by the `MemoryStream` class.

### Reproduction Steps

```
			var request = (HttpWebRequest)WebRequest.Create("https://google.com/test"); // exact URL is irrelevant because request never gets that far

			request.AllowWriteStreamBuffering = false;
			request.Method = "PUT";

			using (var requestStream = request.GetRequestStream())
			{
				byte[] buffer = new byte[1048576];

				for (int megabyte = 0; megabyte < 2048; megabyte++)
					requestStream.Write(buffer, 0, buffer.Length);
			}
```

### Expected behavior

The above sample should actually connect to Google and try to send 2 GB of zeroes (don't do this :-) ).

### Actual behavior

No connection at all is made because `GetRequestStream` actually just creates a `MemoryStream` under the hood to buffer the full request body, even though `AllowWriteStreamBuffering` is `false`.

### Regression?

I am reasonably confident that this code works properly in the full .NET Framework right up to version 4.8.

### Known Workarounds

_No response_

### Configuration

I need to run more testing to confirm, but I believe the bug reproduces on all .NET Core versions, as well as .NET 5.0 and .NET 6 prereleases. The bug does not reproduce on any .NET Framework version, right up to 4.8.

### Other information

_No response_

liveans:
Fixes #60993 and #18632

This property is set (to `TRUE` or `FALSE`) in a 10.2% Nuget package that uses HttpWebRequest.


src/libraries/System.Net.Requests/src/System/Net/HttpClientContentStream.cs
```diff
@@ -0,0 +1,141 @@
...
+            if (_disposed)
+                ThrowDisposedException();
+
+            [StackTraceHidden]
+            static void ThrowDisposedException() => throw new ObjectDisposedException(nameof(HttpClientContentStream));
+        }
+
+        [DoesNotReturn]
+        private static void ThrowNotSupportedException() =>
+            throw new NotSupportedException();
```
rzikm, line 139: This method does not seem to be used

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1061,7 +1062,14 @@ private Task<Stream> InternalGetRequestStream()
                 throw new InvalidOperationException(SR.net_reqsubmitted);
             }
 
-            _requestStream = new RequestStream();
+            // Let's create stream buffer for transferring data from RequestStream to the StreamContent.
```
rzikm, line 1065: Nit

```suggestion
            // create stream buffer for transferring data from RequestStream to the StreamContent.
```

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1061,7 +1062,14 @@ private Task<Stream> InternalGetRequestStream()
...
+            StrongBox<int> box = new StrongBox<int>();
+            StreamBuffer streamBuffer = new StreamBuffer();
+            //s_cachedHttpClient = GetCachedOrCreateHttpClient(async, out bool disposeRequired);
+            _sendRequestTask = SendRequest(async, new StreamContent(new HttpClientContentStream(streamBuffer, box)));
+
+            // If any parameter changed let's change the RequestStream.
```
rzikm, line 1071:
```suggestion
            // If any parameter changed, change the RequestStream.
```

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1061,7 +1062,14 @@ private Task<Stream> InternalGetRequestStream()
                 throw new InvalidOperationException(SR.net_reqsubmitted);
             }
 
-            _requestStream = new RequestStream();
+            // Let's create stream buffer for transferring data from RequestStream to the StreamContent.
+            StrongBox<int> box = new StrongBox<int>();
+            StreamBuffer streamBuffer = new StreamBuffer();
+            //s_cachedHttpClient = GetCachedOrCreateHttpClient(async, out bool disposeRequired);
```
rzikm, line 1068: remove any commented out code.

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -45,17 +54,25 @@ public override bool CanWrite
             }
         }
 
+        // We're already sending data as StreamContent,
+        // so let's buffer data in memory and flush it in three cases:
```
rzikm, line 58:
```suggestion
        // so we buffer data in memory and flush it in three cases:
```

rzikm (commented):
Looks good so far.

ilmax:
I'm a bit puzzled by this. AFAIK `HttpClient` is the way to go, so I was expecting to the older `HttpWebRequest` to get deprecated and eventually removed from the framework altogether at some point, rather than implementing some more functionality.

liveans:
> AFAIK `HttpClient` is the way to go.

That's correct.

> so I was expecting to the older HttpWebRequest to get deprecated and eventually removed from the framework altogether at some point, rather than implementing some more functionality.

The main purpose of this PR is easing the migration from .NET Framework to .NET. We're implemeting some features to increase the compat.

stephentoub:
> We're implemeting some features to increase the compat.

Notably, this is not adding any new APIs. It's fixing bugs where existing APIs don't behave the way they're expected to. That's valuable even in obsolete functionality if it's highly used, which HWR still is.

stephentoub:
> Just out of curiosity, is there a plan to obsolete those classes in the medium/long term?

They already are, eg
https://github.com/dotnet/runtime/blob/623cf77a58f7a233b94dcc1c3ef8eb8d67e8d948/src/libraries/System.Net.Requests/src/System/Net/WebRequest.cs#L194

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,19 +2077,112 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+            await LoopbackServer.CreateClientAndServerAsync(
+                async (uri) =>
+                {
+                    const int size = 1048576 / 4;
+                    HttpWebRequest request = WebRequest.CreateHttp(uri);
+                    request.AllowWriteStreamBuffering = true;
+                    request.ContentType = "application/text";
+                    request.ContentLength = size;
+                    request.Method = "POST";
+                    request.Timeout = (size/16 * delaySec) + request.Timeout;
```
antonfirsov, line 2146: This could get flaky on slow test machines.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,19 +2077,112 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+                    using var stream = await request.GetRequestStreamAsync();
+                    for (int i = 0; i < size/16; i++)
+                    {
+                        await stream.WriteAsync(new ReadOnlyMemory<byte>(Encoding.UTF8.GetBytes(text)));
+                        if (delaySec > 0)
+                        {
+                            await Task.Delay(TimeSpan.FromSeconds(delaySec));
```
antonfirsov, line 2153: What is the behavior being tested with the delays?

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,19 +2077,112 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
                     request.Method = "POST";
                     request.ServicePoint.Expect100Continue = true;
                     request.ContinueTimeout = 30000;
-                    Stream requestStream = await request.GetRequestStreamAsync();
-                    requestStream.Write("aaaa\r\n\r\n"u8);
+                    request.ContentLength = 8;
+                    using (Stream requestStream = await request.GetRequestStreamAsync())
+                    {
+                        requestStream.Write("aaaa\r\n\r\n"u8);
+                    }
```
antonfirsov, line 2183: Are we intentionally disposing the stream before getting the response? Can't this result in a race?

antonfirsov (commented):
Some concerns around the management of the `StreamBuffer` lifecycle and around tests.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
liveans: Just making sure that, waiting for `StreamContent` on `HttpClient` will not cause any unexpected behavior.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,19 +2077,112 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+                    {
+                        await stream.WriteAsync(new ReadOnlyMemory<byte>(Encoding.UTF8.GetBytes(text)));
+                    }
+                },
+                async (server) =>
+                {
+                    await server.AcceptConnectionAsync(async connection =>
+                    {
+                        var data = await connection.ReadRequestDataAsync();
+                        Assert.Equal(text, Encoding.UTF8.GetString(data.Body[0..text.Length]));
```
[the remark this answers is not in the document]
liveans: Actually, I don't have any reason, I can change it to exact match.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,19 +2077,112 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
             });
         }
 
+        [Fact]
+        public async Task SendHttpPostRequest_ConnectionShouldStartWithRequestStream()
```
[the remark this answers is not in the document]
liveans: Normally, we were buffering `RequestStream` content all the time, but now we're not buffering anymore and we're expecting to connection should be started once we call `GetRequestStream`.

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -16,9 +17,17 @@ namespace System.Net
...
+        private readonly StreamBuffer _streamBuffer;
+        private readonly StrongBox<int> _refCount;
+        private readonly bool _isBuffered;
+        private bool _disposed;
 
-        public RequestStream()
+        public RequestStream(StrongBox<int> refCount, StreamBuffer streamBuffer, bool isBuffered)
         {
+            _refCount = refCount;
+            _buffer = new MemoryStream();
```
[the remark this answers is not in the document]
liveans: No, we don't need it, good catch!

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,98 +1134,121 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
 
-        private async Task<WebResponse> SendRequest(bool async)
+        private Task<HttpResponseMessage> SendRequest(bool async, HttpContent? content = null)
         {
             if (RequestSubmitted)
             {
                 throw new InvalidOperationException(SR.net_reqsubmitted);
             }
 
             var request = new HttpRequestMessage(HttpMethod.Parse(_originVerb), _requestUri);
```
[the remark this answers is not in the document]
liveans: Good catch, looks like we weren't disposing of it at all if I'm not missing anything. I'll dispose it after we get the response with `GetResponse`.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,98 +1134,123 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
+                    if (async)
+                    {
+                        await _requestStream!.FlushAsync().ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        _requestStream!.Flush();
+                    }
+                    _requestStream!.EndWriteOnStreamBuffer();
                 }
```
MihaZupan, line 1224: is it a given that the user will be done writing by this point?

That is, would something like this not have worked before?
```c#
var request = ...;
request.AllowWriteStreamBuffering = true;

using var requestStrem = await request.GetRequestStreamAsync();
Task responseTask = request.GetResponseAsync();

await requestStream.WriteAsync(foo);
await requestStream.FlushAsync();

await responseTask;
```

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2138,13 +2235,15 @@ public async Task SendHttpPostRequest_WithContinueTimeoutAndBody_Success(bool ex
         [InlineData(false)]
         public async Task SendHttpPostRequest_When100ContinueSet_ReceivedByServer(bool expect100Continue)
         {
+            TaskCompletionSource tcs = new TaskCompletionSource();
```
MihaZupan, line 2238: Why is this change needed?

src/libraries/System.Net.Requests/src/System/Net/HttpClientContentStream.cs
```diff
@@ -0,0 +1,132 @@
...
+        public override void SetLength(long value) => throw new NotSupportedException();
+
+        private void ThrowIfDisposed()
+        {
+            if (_disposed)
+                ThrowDisposedException();
+
+            [StackTraceHidden]
+            [DoesNotReturn]
+            static void ThrowDisposedException() => throw new ObjectDisposedException(nameof(HttpClientContentStream));
```
MihaZupan, line 129: Nit: This could be `ObjectDisposedException.ThrowIf(_disposed, this);`

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,98 +1134,123 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
 
-                _sendRequestTask = async ?
-                    client.SendAsync(request, _allowReadStreamBuffering ? HttpCompletionOption.ResponseContentRead : HttpCompletionOption.ResponseHeadersRead, _sendRequestCts!.Token) :
-                    Task.FromResult(client.Send(request, _allowReadStreamBuffering ? HttpCompletionOption.ResponseContentRead : HttpCompletionOption.ResponseHeadersRead, _sendRequestCts!.Token));
+            _sendRequestMessage.Version = ProtocolVersion;
+            HttpCompletionOption completionOption = _allowReadStreamBuffering ?
+                            HttpCompletionOption.ResponseContentRead : HttpCompletionOption.ResponseHeadersRead;
+            _sendRequestTask = async ?
+                _httpClient.SendAsync(_sendRequestMessage, completionOption, _sendRequestCts!.Token) :
+                Task.Run(() => _httpClient.Send(_sendRequestMessage, completionOption, _sendRequestCts!.Token));
```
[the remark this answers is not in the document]
liveans: Yes, this is basically removing the sync path of HttpWebRequest with this change, otherwise we will block the thread on `GetRequestStream` method call and request will never be completed.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,98 +1134,123 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
 
-                HttpResponseMessage responseMessage = await _sendRequestTask.ConfigureAwait(false);
+            return _sendRequestTask!;
+        }
 
-                HttpWebResponse response = new HttpWebResponse(responseMessage, _requestUri, _cookieContainer);
+        private async Task<WebResponse> HandleResponse(bool async)
+        {
+            _sendRequestTask ??= SendRequest(async);
```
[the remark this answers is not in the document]
liveans: This should be `??=`, because basically, we're starting a request on `RequestStream` if we call it, but if we call `GetResponse` right away, we're starting the request here.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
liveans: I believe once we call, `GetResponse` we're basically saying that `hey there is no more data I want to send`, so it should be safe to call it here.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
liveans: I believe this is not needed, I'll remove it.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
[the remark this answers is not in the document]
MihaZupan: This sort of defeats the point of having a non-async path at all though.
We should try to do the work synchronously on the calling thread. If that's not possible and we have to block every time, then we might as well just use real async.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,19 +2077,112 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+                    HttpWebRequest request = WebRequest.CreateHttp(uri);
+                    request.ContentType = "application/text";
+                    request.ContentLength = size;
+                    request.Method = "POST";
+                    request.AllowWriteStreamBuffering = false;
+                    using var stream = await request.GetRequestStreamAsync();
+                    for (int i = 0; i < size / 16; i++)
+                    {
+                        await stream.WriteAsync(new ReadOnlyMemory<byte>(Encoding.UTF8.GetBytes(text)));
+                    }
```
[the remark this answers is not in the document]
liveans: This test is trying to exceed `memory limit`, and since we're not buffering here, this should succeed and should not throw oom exception.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1066,7 +1070,12 @@ private Task<Stream> InternalGetRequestStream()
                 throw new InvalidOperationException(SR.net_reqsubmitted);
             }
 
-            _requestStream = new RequestStream();
+            // Create stream buffer for transferring data from RequestStream to the StreamContent.
+            StreamBuffer streamBuffer = new StreamBuffer();
+            _sendRequestTask = SendRequest(async, new StreamContent(new HttpClientContentStream(streamBuffer)));
+
+            // If any parameter changed, change the RequestStream.
+            _requestStream ??= new RequestStream(streamBuffer, AllowWriteStreamBuffering);
```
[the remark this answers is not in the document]
liveans: Looks like, this is not trivial, so I just changed it from `??=` to `=`.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,98 +1134,123 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
+                if (_requestStream is not null) // If we didn't flush and end the request stream, do it now, don't wait for dispose.
                 {
-                    throw new WebException(
-                        SR.Format(SR.net_servererror, (int)response.StatusCode, response.StatusDescription),
-                        null,
-                        WebExceptionStatus.ProtocolError,
-                        response);
+                    if (async)
+                    {
+                        await _requestStream!.FlushAsync().ConfigureAwait(false);
```
MihaZupan, line 1217:
```suggestion
                        await _requestStream.FlushAsync().ConfigureAwait(false);
```
Shouldn't be needed because of the null check above

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,98 +1134,123 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
+                    }
+                    _requestStream!.EndWriteOnStreamBuffer();
                 }
+            }
+            catch (ObjectDisposedException)
+            {
+                // If the request stream was disposed, don't do anything.
+            }
+
+            HttpResponseMessage responseMessage = await _sendRequestTask!.ConfigureAwait(false);
```
MihaZupan, line 1231:
```suggestion
            HttpResponseMessage responseMessage = await _sendRequestTask.ConfigureAwait(false);
```

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,98 +1134,123 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
+            {
+                throw new WebException(
+                    SR.Format(SR.net_servererror, (int)response.StatusCode, response.StatusDescription),
+                    null,
+                    WebExceptionStatus.ProtocolError,
+                    response);
             }
-            finally
+
+            _sendRequestMessage?.Dispose();
```
MihaZupan, line 1245:
```suggestion
            _sendRequestMessage.Dispose();
```

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,98 +1134,123 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
+
+            if (_disposeRequired)
             {
-                if (disposeRequired)
-                {
-                    client?.Dispose();
-                }
+                _httpClient?.Dispose();
```
MihaZupan, line 1249: This should likely stay in the finally block

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,98 +1134,123 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
+                _httpClient = null;
```
MihaZupan, line 1250: This is probably not needed? (why clear this and not `_sendRequestMessage` for example)

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,98 +1134,123 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
-                        null,
-                        WebExceptionStatus.ProtocolError,
-                        response);
+                    if (async)
+                    {
+                        await _requestStream!.FlushAsync().ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        _requestStream!.Flush();
```
MihaZupan, line 1221:
```suggestion
                        _requestStream.Flush();
```

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,98 +1134,123 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
+                    }
+                    _requestStream!.EndWriteOnStreamBuffer();
```
MihaZupan, line 1223:
```suggestion
                    _requestStream.EndWriteOnStreamBuffer();
```

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -15,10 +16,19 @@ namespace System.Net
...
 
-        public RequestStream()
+        public RequestStream(StreamBuffer streamBuffer, bool isBuffered)
         {
+            if (isBuffered)
+            {
+                _buffer = new MemoryStream();
+            }
+            _streamBuffer = streamBuffer;
+            _isBuffered = isBuffered;
```
MihaZupan, line 31: We already have a `StreamBuffer`. Can we skip the different handling here and just always write into it instead of the memory stream?
(should change max capacity if buffering is enabled)

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -45,17 +55,34 @@ public override bool CanWrite
...
+        // so we buffer data in memory and flush it in three cases:
+        // - When GetResponse called.
+        // - When RequestStream is getting disposed.
+        // - When user calls Flush.
         public override void Flush()
         {
-            // Nothing to do.
+            ThrowIfDisposed();
+
+            if (_isBuffered && _buffer!.Length > 0)
```
MihaZupan, line 67: If we always write into the stream buffer, the flushing logic also becomes unnecessary

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -95,40 +122,80 @@ public override void SetLength(long value)
...
+                _disposed = true;
+                _streamBuffer.EndWrite();
+            }
+
+            base.Dispose(disposing);
+        }
+
+        private void ThrowIfDisposed()
+        {
+            if (_disposed)
```
MihaZupan, line 194: Nit: `ObjectDisposedException.ThrowIf(_disposed, this);`

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,19 +2077,115 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+        }
+
+        [Fact]
+        public async Task SendHttpPostRequest_AllowWriteBufferingFalse_ConnectionShouldStartWithRequestStream()
+        {
+            const string text = "Hello World!!!!\n";
+            await LoopbackServer.CreateClientAndServerAsync(
+                async (uri) =>
+                {
+                    const int size = 1048576 * 2;
```
MihaZupan, line 2108:
```suggestion
                    int size = text.Length * 128 * 1024;
```

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,19 +2077,115 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
 
+        [Fact]
+        public async Task SendHttpPostRequest_ConnectionShouldStartWithRequestStream()
+        {
+            await LoopbackServer.CreateClientAndServerAsync(
+                async (uri) =>
+                {
+                    HttpWebRequest request = WebRequest.CreateHttp(uri);
+                    request.Method = "POST";
+                    var stream = await request.GetRequestStreamAsync();
```
MihaZupan, line 2088: This is leaving an unobserved exception.

Can we add something like
```c#
await Assert.ThrowsAnyAsync<Exception>(() => request.GetResponseAsync());
```

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,19 +2077,115 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+                        }
+                    });
+                }
+            );
+        }
+
+        [Theory]
+        [InlineData(0)]
+        [InlineData(1)]
+        public async Task SendHttpPost_WriteBufferingEnabled(int delayMs)
```
MihaZupan, line 2137: What is this test verifying?
Looks like we're writing & sleeping 16k times, but only checking that the first write went through?

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,19 +2077,115 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+                        Assert.Equal(text, Encoding.UTF8.GetString(data.Body[0..text.Length]));
+                    });
+                }
+            );
+        }
+        
         [Fact]
         public async Task SendHttpPostRequest_WithContinueTimeoutAndBody_BodyIsDelayed()
         {
+            TaskCompletionSource tcs = new TaskCompletionSource();
```
MihaZupan, line 2174: Is this needed?

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,19 +2077,115 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
             await LoopbackServer.CreateClientAndServerAsync(
                 async (uri) =>
                 {
                     HttpWebRequest request = WebRequest.CreateHttp(uri);
                     request.Method = "POST";
                     request.ServicePoint.Expect100Continue = true;
                     request.ContinueTimeout = 30000;
-                    Stream requestStream = await request.GetRequestStreamAsync();
-                    requestStream.Write("aaaa\r\n\r\n"u8);
+                    request.ContentLength = 8;
```
MihaZupan, line 2182: This test should work the same without the content length as well, right?

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1066,7 +1070,12 @@ private Task<Stream> InternalGetRequestStream()
                 throw new InvalidOperationException(SR.net_reqsubmitted);
             }
 
-            _requestStream = new RequestStream();
+            // Create stream buffer for transferring data from RequestStream to the StreamContent.
+            StreamBuffer streamBuffer = new StreamBuffer();
+            _sendRequestTask = SendRequest(async, new StreamContent(new HttpClientContentStream(streamBuffer)));
```
MihaZupan, line 1075: A subtle change here is that we're no longer sending the `Content-Length` header when buffering if we go through the `StreamContent`.
If we change this to not open the connection until GetResponse is called, we should set it manually on the content.

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -45,17 +55,34 @@ public override bool CanWrite
...
+        // - When user calls Flush.
         public override void Flush()
         {
-            // Nothing to do.
+            ThrowIfDisposed();
+
+            if (_isBuffered && _buffer!.Length > 0)
+            {
+                _streamBuffer.Write(_buffer!.GetBuffer().AsSpan(0, (int) _buffer!.Length));
+                _buffer!.SetLength(0);
```
MihaZupan, line 70:
```suggestion
                _streamBuffer.Write(_buffer.GetBuffer().AsSpan(0, (int) _buffer.Length));
                _buffer.SetLength(0);
```

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -45,17 +55,34 @@ public override bool CanWrite
...
-            // Nothing to do.
-            return cancellationToken.IsCancellationRequested ?
-                Task.FromCanceled(cancellationToken) :
-                Task.CompletedTask;
+            ThrowIfDisposed();
+
+            if (_isBuffered && _buffer!.Length > 0)
+            {
+                Task task = _streamBuffer.WriteAsync(_buffer!.GetBuffer().AsMemory(0, (int) _buffer!.Length), cancellationToken).AsTask();
+                _buffer!.SetLength(0);
```
MihaZupan, line 81:
```suggestion
                Task task = _streamBuffer.WriteAsync(_buffer.GetBuffer().AsMemory(0, (int) _buffer.Length), cancellationToken).AsTask();
                _buffer.SetLength(0);
```

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -95,40 +122,80 @@ public override void SetLength(long value)
...
+            ThrowIfDisposed();
             ValidateBufferArguments(buffer, offset, count);
-            _buffer.Write(buffer, offset, count);
+            if (_isBuffered)
+            {
+                _buffer!.Write(buffer, offset, count);
+            }
+            else
+            {
+                _streamBuffer!.Write(new(buffer, offset, count));
```
MihaZupan, line 133:
```suggestion
                _streamBuffer.Write(new(buffer, offset, count));
```

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -95,40 +122,80 @@ public override void SetLength(long value)
...
+            }
         }
 
         public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
+            ThrowIfDisposed();
             ValidateBufferArguments(buffer, offset, count);
-            return _buffer.WriteAsync(buffer, offset, count, cancellationToken);
+            return _isBuffered ? _buffer!.WriteAsync(buffer, offset, count, cancellationToken) :
+                _streamBuffer!.WriteAsync(new(buffer, offset, count), cancellationToken).AsTask();
```
MihaZupan, line 142:
```suggestion
                _streamBuffer.WriteAsync(new(buffer, offset, count), cancellationToken).AsTask();
```

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -45,17 +55,34 @@ public override bool CanWrite
...
 
         public override Task FlushAsync(CancellationToken cancellationToken)
         {
-            // Nothing to do.
-            return cancellationToken.IsCancellationRequested ?
-                Task.FromCanceled(cancellationToken) :
-                Task.CompletedTask;
+            ThrowIfDisposed();
+
+            if (_isBuffered && _buffer!.Length > 0)
```
MihaZupan, line 78: If you wanted to get rid of all these `!`, you could change `_isBuffered` to instead be something like
```c#
[MemberNotNullWhen(true, nameof(_buffer))]
private bool IsBuffered { get; set; }
```

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -95,40 +122,80 @@ public override void SetLength(long value)
...
+            return _isBuffered ? _buffer!.WriteAsync(buffer, offset, count, cancellationToken) :
+                _streamBuffer!.WriteAsync(new(buffer, offset, count), cancellationToken).AsTask();
         }
 
         public override ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken = default)
         {
-            return _buffer.WriteAsync(buffer, cancellationToken);
+            ThrowIfDisposed();
+            return _isBuffered ? _buffer!.WriteAsync(buffer, cancellationToken) :
+                _streamBuffer!.WriteAsync(buffer, cancellationToken);
```
MihaZupan, line 149:
```suggestion
                _streamBuffer.WriteAsync(buffer, cancellationToken);
```

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -95,40 +122,80 @@ public override void SetLength(long value)
...
         }
 
         public override IAsyncResult BeginWrite(byte[] buffer, int offset, int count, AsyncCallback? asyncCallback, object? asyncState)
         {
+            ThrowIfDisposed();
             ValidateBufferArguments(buffer, offset, count);
-            return _buffer.BeginWrite(buffer, offset, count, asyncCallback, asyncState);
+            return _isBuffered ? _buffer!.BeginWrite(buffer, offset, count, asyncCallback, asyncState) :
+                TaskToAsyncResult.Begin(_streamBuffer!.WriteAsync(new(buffer, offset, count)).AsTask(), asyncCallback, asyncState);
```
MihaZupan, line 157:
```suggestion
                TaskToAsyncResult.Begin(_streamBuffer.WriteAsync(new(buffer, offset, count)).AsTask(), asyncCallback, asyncState);
```

MihaZupan (commented):
A bunch of nits re: nullability.
The rest should hopefully be simple to address too.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
liveans: This is already running after try-catch, so it will run in any case.

MihaZupan: It's after the try/catch around flushing, not around awaiting the request task

liveans: Ah, that's right my bad. I'll move it into finally block again

src/libraries/System.Net.Requests/src/System/Net/HttpClientContentStream.cs
```diff
@@ -0,0 +1,127 @@
...
+        public override void Write(ReadOnlySpan<byte> buffer) => throw new NotSupportedException();
+
+        public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken) => throw new NotSupportedException();
+
+        public override ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken = default) => throw new NotSupportedException();
+
+        public override IAsyncResult BeginWrite(byte[] buffer, int offset, int count, AsyncCallback? callback, object? state) => throw new NotSupportedException();
+
+        public override void EndWrite(IAsyncResult asyncResult) => throw new NotSupportedException();
+
```
MihaZupan, line 116:
```suggestion
```

Nit: It shouldn't harm anything, but there's no need to override all of the write methods to throw, they'll delegate to the abstract `Write(byte[], int, int)`.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,78 +1138,108 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
+
+            HttpContent content;
+            if (_requestStream is not null)
+            {
+                content = new StreamContent(new HttpClientContentStream(_requestStream.GetBuffer()));
+                if (AllowWriteStreamBuffering)
                 {
-                    request.Headers.ExpectContinue = true;
+                    content.Headers.ContentLength = _requestStream.GetBuffer().ReadBytesAvailable;
                 }
```
MihaZupan, line 1230: This looks duplicated from what SendRequest will already do.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,78 +1138,108 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
+            }
+            else
+            {
+                content = new ByteArrayContent(Array.Empty<byte>());
```
MihaZupan, line 1234: Nit: This could remain as it was before the PR where we only lazily allocate it when we have to for headers. Unless I'm missing some detail?

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -45,17 +48,20 @@ public override bool CanWrite
             }
         }
 
+        // We're already sending data as StreamContent,
+        // so we buffer data in memory and flush it in three cases:
+        // - When GetResponse called.
+        // - When RequestStream is getting disposed.
+        // - When user calls Flush.
```
MihaZupan, line 55:
```suggestion
```

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -95,40 +101,63 @@ public override void SetLength(long value)
 
         public override void Write(byte[] buffer, int offset, int count)
         {
+            ThrowIfDisposed();
             ValidateBufferArguments(buffer, offset, count);
-            _buffer.Write(buffer, offset, count);
+                _streamBuffer.Write(new(buffer, offset, count));
```
MihaZupan, line 106:
```suggestion
            _streamBuffer.Write(new(buffer, offset, count));
```

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -95,40 +101,63 @@ public override void SetLength(long value)
...
-            _buffer.EndWrite(asyncResult);
+            ThrowIfDisposed();
+            TaskToAsyncResult.End(asyncResult);
         }
 
-        public ArraySegment<byte> GetBuffer()
+        protected override void Dispose(bool disposing)
         {
-            ArraySegment<byte> bytes;
+            ThrowIfDisposed();
```
MihaZupan, line 147: Double dispose is generally silently ignored instead of throwing

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,6 +2077,69 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+                    var stream = await request.GetRequestStreamAsync();
+                    await Assert.ThrowsAnyAsync<Exception>(() => request.GetResponseAsync());
+                },
+                async (server) => 
+                {
+                    Exception? exception = await Record.ExceptionAsync(() => server.AcceptConnectionAsync(_ =>
+                    {
+                        return Task.CompletedTask;
+                    }));
+                    Assert.Null(exception);
```
MihaZupan, line 2098:
```suggestion
                    await server.AcceptConnectionAsync(_ =>
                    {
                        return Task.CompletedTask;
                    });
```

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,78 +1138,108 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
-                foreach (string headerName in _webHeaderCollection)
-                {
-                    // The System.Net.Http APIs require HttpRequestMessage headers to be properly divided between the request headers
-                    // collection and the request content headers collection for all well-known header names.  And custom headers
-                    // are only allowed in the request headers collection and not in the request content headers collection.
-                    if (IsWellKnownContentHeader(headerName))
-                    {
-                        // Create empty content so that we can send the entity-body header.
-                        request.Content ??= new ByteArrayContent(Array.Empty<byte>());
+            _sendRequestMessage.Version = ProtocolVersion;
```
MihaZupan, line 1188:
```suggestion
```

You're already setting the version below

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,78 +1138,108 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
-                if (_servicePoint?.Expect100Continue == true)
+        private async Task<WebResponse> HandleResponse(bool async)
+        {
+            try
+            {
+                _requestStream?.EndWriteOnStreamBuffer();
+            }
+            catch (ObjectDisposedException)
+            {
+                // If the request stream was disposed, don't do anything.
```
MihaZupan, line 1220: `EndWriteOnStreamBuffer` doesn't look like it would ever throw.
We can drop the try/catch here

MihaZupan (approved):

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
liveans: requestStream can be disposed before we call, that's why I'm keeping it here.

MihaZupan: But even if it is disposed, `EndWriteOnStreamBuffer` won't throw

liveans: Because, I forgot to put it. Good catch, let me take a look.

MihaZupan: These are internal only helpers, so if we're going to ignore the exception anyway we might as well not throw it (especially given that the expectation here is that this stream will already be disposed in well behaved apps)

liveans: Yes, that's what I'm doing right now, I will push changes in a min, I needed to check `streamBuffer` implementation first, but disposing or ending to write 2 times is not harming us.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,78 +1138,95 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
-                HttpResponseMessage responseMessage = await _sendRequestTask.ConfigureAwait(false);
+        private async Task<WebResponse> HandleResponse(bool async)
+        {
+            // If user code used requestStream and didn't dispose it (end write on StreamBuffer)
+            // We're ending it here.
+            _requestStream?.EndWriteOnStreamBuffer();
 
-                HttpWebResponse response = new HttpWebResponse(responseMessage, _requestUri, _cookieContainer);
+            _sendRequestTask ??= _requestStream is not null ?
+                SendRequest(async, new StreamContent(new HttpClientContentStream(_requestStream.GetBuffer()))) :
```
ManickaP, line 1223: Why aren't we using `ByteArrayContent` anymore?

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1125,78 +1138,95 @@ public override Stream EndGetRequestStream(IAsyncResult asyncResult)
...
+            }
+            else
+            {
+                _sendRequestMessage.Headers.ConnectionClose = true;
+            }
 
-                request.Version = ProtocolVersion;
+            _sendRequestMessage.Version = ProtocolVersion;
+            HttpCompletionOption completionOption = _allowReadStreamBuffering ?
+                            HttpCompletionOption.ResponseContentRead : HttpCompletionOption.ResponseHeadersRead;
```
ManickaP, line 1206: NIT:
```suggestion
            HttpCompletionOption completionOption = _allowReadStreamBuffering ? HttpCompletionOption.ResponseContentRead : HttpCompletionOption.ResponseHeadersRead;
```

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,6 +2077,68 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+            const string text = "Hello World!!!!\n";
+            await LoopbackServer.CreateClientAndServerAsync(
+                async (uri) =>
+                {
+                    int size = text.Length * 128 * 1024;
+                    HttpWebRequest request = WebRequest.CreateHttp(uri);
+                    request.ContentType = "application/text";
+                    if (!buffering)
+                    {
+                        request.ContentLength = size;
```
ManickaP, line 2116: This shouldn't be necessary.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
liveans: Because we have an option not to buffer now and because of that we don't have the whole content upfront, ByteArrayContent is still useful if we're buffering but this is more suitable because of how `HttpWebRequest` works, IMO.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1066,7 +1070,16 @@ private Task<Stream> InternalGetRequestStream()
                 throw new InvalidOperationException(SR.net_reqsubmitted);
             }
 
-            _requestStream = new RequestStream();
+            // Create stream buffer for transferring data from RequestStream to the StreamContent.
+            StreamBuffer streamBuffer = new StreamBuffer(maxBufferSize: int.MaxValue);
```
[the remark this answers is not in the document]
liveans: Yes actually that would be better to limit, if we're not buffering.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1044,10 +1048,10 @@ public override WebResponse GetResponse()
 
         public override Stream GetRequestStream()
         {
-            return InternalGetRequestStream().Result;
+            return InternalGetRequestStream(async: false).Result;
         }
 
-        private Task<Stream> InternalGetRequestStream()
+        private Task<Stream> InternalGetRequestStream(bool async)
```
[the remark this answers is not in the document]
liveans: > It even gets ignored in SendRequest if we don't allow buffering, so the arg is moot here.

We're ignoring it because we're keeping the Sync path in case we're buffering, but we can't keep the Sync path if we're not buffering because we should open the connection upfront (`by starting request task`) and if we try to use Sync over thread we're going to lock the whole thread.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,6 +2077,68 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+                    }
+                    request.Method = "POST";
+                    request.AllowWriteStreamBuffering = buffering;
+                    using var stream = await request.GetRequestStreamAsync();
+                    for (int i = 0; i < size / text.Length; i++)
+                    {
+                        await stream.WriteAsync(new ReadOnlyMemory<byte>(Encoding.UTF8.GetBytes(text)));
+                    }
+                    await request.GetResponseAsync();
```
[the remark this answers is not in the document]
liveans: Yes, I can write a test for it with a raw `Socket`, instead of using the `HttpServer`.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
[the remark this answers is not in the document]
liveans: > Why does asynchronicity of getting the request stream affects the asynchronicity of invoking `HttpClient.Send(Async?)`? I should be able to do:
> 
> ```
> GetRequestStream() - synchronously retrieve the request content stream
> // write to stream asynchronously
> await TaskToApm(BeginGetResponse, EndGetResponse) - asynchronously invoke the request
> ```

That's a good point actually, and with the current workflow that should be possible. But as you already mentioned this argument might not be necessary here, because in case we're not buffering we're triggering the async path anyway. I will delete this parameter.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
liveans: Looks like I've deleted the other one but not this one :) will delete it

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
liveans: It should actually, if we're not buffering we should either set `Content-Length` or we should mark it as `SendChunked`.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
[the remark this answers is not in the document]
ManickaP: You don't have to write raw socket test. LoopbackServer has individual functions/helper to read headers and body, even to read a block of bytes:
https://github.com/dotnet/runtime/blob/bd5f6eb73f6dffe91386f75d3ed5531ddcb2aeb6/src/libraries/Common/tests/System/Net/Http/LoopbackServer.cs#L543-L543

liveans:
@ManickaP @MihaZupan Can you please check the last commit I've added an exception to match .NET Framework behavior here if we're not buffering.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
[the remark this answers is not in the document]
liveans: Added the test for it.

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1061,12 +1065,31 @@ private Task<Stream> InternalGetRequestStream()
...
+
             if (RequestSubmitted)
             {
                 throw new InvalidOperationException(SR.net_reqsubmitted);
             }
 
-            _requestStream = new RequestStream();
+            // Create stream buffer for transferring data from RequestStream to the StreamContent.
+            StreamBuffer streamBuffer = new StreamBuffer(
+                maxBufferSize: AllowWriteStreamBuffering is false ? int.MaxValue : StreamBuffer.DefaultMaxBufferSize);
```
ManickaP, line 1080:
```suggestion
            StreamBuffer streamBuffer = new StreamBuffer(maxBufferSize: AllowWriteStreamBuffering ? int.MaxValue : StreamBuffer.DefaultMaxBufferSize);
```
The condition seems to be inversed.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
ManickaP: No, that should do the web request (in this case `HttpContent`) on it's own.
If this wasn't necessary before, it cannot be required now, it will break users.

See this code path:
https://github.com/dotnet/runtime/blob/a3f74163d65fbbd81b9ed946e9d74f080be1dc70/src/libraries/System.Net.Http/src/System/Net/Http/Headers/HttpContentHeaders.cs#L34-L65

It calls `TryComputeLength` on the content:
https://github.com/dotnet/runtime/blob/a3f74163d65fbbd81b9ed946e9d74f080be1dc70/src/libraries/System.Net.Http/src/System/Net/Http/StreamContent.cs#L77-L89

So that's where the decision about ContentLength vs Chunked should be made. Because this all gets called from `SocketsHttpHandler` and it does the right thing:
https://github.com/dotnet/runtime/blob/a3f74163d65fbbd81b9ed946e9d74f080be1dc70/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/SocketsHttpHandler.cs#L631-L635

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,6 +2077,103 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+        [Fact]
+        public async Task SendHttpRequest_WhenNotBuffering_SendSuccess()
+        {
+            TaskCompletionSource tcs = new();
+            await LoopbackServer.CreateClientAndServerAsync(
+                async (uri) =>
+                {
+                    HttpWebRequest request = WebRequest.CreateHttp(uri);
+                    request.Method = "POST";
+                    request.ContentLength = 5;
```
ManickaP, line 2152: Why is content length used when non-buffering (=streaming) is on?

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,6 +2077,103 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+                    request.AllowWriteStreamBuffering = false;
+                    using (Stream requestStream = await request.GetRequestStreamAsync())
+                    {
+                        requestStream.Write("Hello"u8);
+                    }
+                    await tcs.Task;
```
ManickaP, line 2158: This isn't testing what I asked for. The await should be between different chunks for request content and unblocked by server reading at least some content data from the request content.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,6 +2077,103 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+                    await request.GetResponseAsync();
+                },
+                async (server) =>
+                {
+                    await server.AcceptConnectionAsync(async (client) =>
```
ManickaP, line 2163:
```suggestion
                    await server.AcceptConnectionAsync(async (connection) =>
```
This is still server side.

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
liveans: Because HttpWebRequest requires it, there is no way to know how many bytes will come through the stream if the user doesn't provide it upfront.

```c#
HttpWebRequest request = WebRequest.CreateHttp("https://httpbin.org/post");
request.AllowWriteStreamBuffering = false;
request.Method = "POST";
using (var requestStream = request.GetRequestStream())
{
    byte[] data = Encoding.UTF8.GetBytes("Hello, World!");
    requestStream.Write(data, 0, data.Length);
}
using (var response = request.GetResponse())
{
    using (var responseStream = response.GetResponseStream())
    {
        byte[] buffer = new byte[1024];
        int bytesRead;
        while ((bytesRead = responseStream.Read(buffer, 0, buffer.Length)) > 0)
        {
            Console.Write(Encoding.UTF8.GetString(buffer, 0, bytesRead));
        }
    }
}
```
This code throws that exception from here on .NET Framework.
image

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
liveans: It's not necessary anymore, if we don't provide `Content-Length` `HttpWebRequest` will use `Transfer-Encoding: chunked`.

src/libraries/System.Net.Requests/src/System/Net/RequestStreamContent.cs
```diff
@@ -0,0 +1,32 @@
...
+        protected override Task SerializeToStreamAsync(Stream stream, TransportContext? context)
+        {
+            return SerializeToStreamAsync(stream, context, default);
+        }
+        protected override async Task SerializeToStreamAsync(Stream stream, TransportContext? context, CancellationToken cancellationToken)
+        {
+            Debug.Assert(requestStream is not null);
+            Debug.Assert(stream is not null);
+
+            await stream.FlushAsync(cancellationToken).ConfigureAwait(false);
```
ManickaP, line 23: I don't understand why we need this here.

src/libraries/System.Net.Requests/src/System/Net/RequestStreamContent.cs
```diff
@@ -0,0 +1,32 @@
...
+            await requestStream.CopyToAsync(stream, requestStream.GetBuffer().ReadBytesAvailable, cancellationToken).ConfigureAwait(false);
```
ManickaP, line 24: This is copying data from `requestStream` to `stream` and one chunk of size `ReadBytesAvailable`. Is this what you really want?

ManickaP: Looking at framework code, looks like each write to `ConnectStream` (in our case `RequestStream`) is one chunk. This is not behaving like that. Or am I missing something?

MihaZupan: I think we should drop the buffer size argument here and let Stream use its default.

ManickaP: If we were to keep the chunk size corresponding to how big byte[] user wrote to the request stream. I think we need to implement the copying ourselves and not rely on `CopyToAsync`. Each write to the underlying stream will create one chunk, so we should propagate writes 1:1.

liveans:
I've ended up removing `StreamBuffer` and separating the Non-Buffering and Buffering streams. Currently, the buffering mode will use `ByteArrayContent` again, and for streaming mode, we have `RequestStreamContent` which connects the internal `HttpClient` stream to `RequestStream` directly, and basically `RequestStream` becomes a wrapper as Natalia suggested in our chat.

Is there any pushback on these changes?

cc @ManickaP @MihaZupan @CarnaViire

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
```diff
@@ -1066,9 +1070,24 @@ private Task<Stream> InternalGetRequestStream()
...
+            // Because we need to send the data as soon as possible when it's available from the RequestStream.
+            // Making this allows us to keep the sync send request path for buffering cases.
+            if (AllowWriteStreamBuffering is false)
+            {
+                // We're calling SendRequest with async, because we need to open the connection and send the request
+                // Otherwise, sync path will block the current thread until the request is sent.
+                TaskCompletionSource<Stream> getStreamTcs = new();
+                TaskCompletionSource completeTcs = new();
+                _sendRequestTask = SendRequest(async: true, new RequestStreamContent(getStreamTcs, completeTcs));
+                _requestStream = new RequestStream(getStreamTcs.Task.GetAwaiter().GetResult(), completeTcs);
```
liveans, line 1083: Looks like I can mark this method as async and await this `getStreamTcs.Task` instead of calling `GetAwaiter().GetResult()`

src/libraries/System.Net.Requests/src/System/Net/RequestStreamContent.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+        protected override bool TryComputeLength(out long length)
+        {
+            length = -1;
```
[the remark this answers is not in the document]
liveans: Since I've separated streams this is not needed anymore.

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -95,40 +94,67 @@ public override void SetLength(long value)
...
+            await FlushAsync(cancellationToken).ConfigureAwait(false);
         }
 
-        public override ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken = default)
+        public override async ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken = default)
         {
-            return _buffer.WriteAsync(buffer, cancellationToken);
+            ThrowIfDisposed();
+            await _internalStream.WriteAsync(buffer, cancellationToken).ConfigureAwait(false);
+            await FlushAsync(cancellationToken).ConfigureAwait(false);
```
MihaZupan, line 115: Why are we injecting flushes on writes?

liveans: To eliminate mini write buffer on HttpConnection. (Every write will be different chunk)

MihaZupan: Small buffers like this are fine though, no? The user is free to call `Flush` when they want to ensure data is written.
Manually flushing here only guarantees that each user `Write` call will translate into a `Write` call to the transport stream (e.g. Socket). The HTTP traffic sent will be the same.

liveans: It is, I didn't realize HTTP traffic will not change, once I looked at the code path now I got it, I can remove flushes

src/libraries/System.Net.Requests/src/System/Net/RequestStream.cs
```diff
@@ -95,40 +94,63 @@ public override void SetLength(long value)
...
+            _internalStream.Flush();
+            Complete();
+            base.Dispose(disposing);
+        }
 
-            bool success = _buffer.TryGetBuffer(out bytes);
-            Debug.Assert(success); // Buffer should always be visible since default MemoryStream constructor was used.
+        public override async ValueTask DisposeAsync()
+        {
+            await _internalStream.FlushAsync().ConfigureAwait(false);
```
ManickaP, line 146: Should you set `_disposed = true` here as well?

src/libraries/System.Net.Requests/src/System/Net/HttpWebRequest.cs
ManickaP: Please do.

src/libraries/System.Net.Requests/src/System/Net/RequestBufferingStream.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            ArraySegment<byte> bytes;
+
+            bool success = _buffer.TryGetBuffer(out bytes);
+            Debug.Assert(success); // Buffer should always be visible since default MemoryStream constructor was used.
+
+            return bytes;
+        }
+
+        // We need this to dispose the MemoryStream.
```
ManickaP, line 114: Why doesn't this class `Dispose` dispose the `MemoryStream`?

src/libraries/System.Net.Requests/tests/HttpWebRequestTest.cs
```diff
@@ -2077,6 +2077,123 @@ public async Task SendHttpGetRequest_WithCachePolicyBypassCache_DoNotAddHeaders(
...
+                    });
+                }
+            );
+        }
+
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public async Task SendHttpRequest_WhenNotBuffering_SendSuccess(bool isChunked)
+        {
```
ManickaP, line 2142: A lot of magic numbers around the byte array work. You could use something like:
```C#
byte[] firstBlock = "Hello"u8.ToArray();
byte[] secondBlock = "WorlddD"u8.ToArray();

await LoopbackServer.CreateClientAndServerAsync(
...
requestStream.Write("Hello"u8);
...
await connection.ReadBlockAsync(buffer, 0, firstBlock.Length);
```

ManickaP (commented):
This looks good! Last batch of mostly small comments. Thanks!

src/libraries/System.Net.Requests/src/System/Net/RequestBufferingStream.cs
liveans: Because user will be controlling this class and they can choose to dispose it before we send the request (Before calling `GetResponse`), in that case we'll try to use disposed `MemoryStream`

ManickaP: > they can choose to dispose it before we send the request (Before calling GetResponse)

Is this a valid use that is expected to work? Are you saying that this sequence of actions work on Framework?

liveans: Yes, it works on .NET Framework

```c#
HttpWebRequest request = (HttpWebRequest)WebRequest.Create("http://httpbin.org/post");
request.Method = "POST";
using (var requestStream = request.GetRequestStream())
{
    byte[] data = System.Text.Encoding.UTF8.GetBytes("Hello World");
    requestStream.Write(data, 0, data.Length);
}
using (var response = request.GetResponse())
{
    using (var responseStream = response.GetResponseStream())
    {
        using (var reader = new System.IO.StreamReader(responseStream))
        {
            string responseFromServer = reader.ReadToEnd();
            Console.WriteLine(responseFromServer);
        }
    }
}
Console.ReadLine();
```

ManickaP (approved):
LGTM, thanks!

PR #95004, opened 2023-11-20
Console.Unix: StdInReader: use stateful decoder.

### LINKED ISSUE #88343: Console.CursorTop causes invalid Console.ReadKey input
### Description

Calling `Console.CursorTop` when reading `Console.ReadKey` in a loop can cause `Console.ReadKey` to read an invalid character (�).

Simply removing the call to `Console.CursorTop` (with no other changes) solves the issue. There's no multithreading going on here, but it does require calling `Console.ReadKey` repeatedly in a loop.

It does not happen on Windows, but it does happen on Linux as well as WSL2 in Windows Terminal.

### Reproduction Steps

Reproduction steps:

1. Run the below program (it's a valid top-level program for .NET 7).
2. Put about 100 or 200 unicode characters on your clipboard (e.g. the character '❌' a bunch of times) and then paste it several times.
    - See the screen recording below as an example.

**Program showing the problem:**

```csharp
// the following is valid top-level statement program.

Console.WriteLine("Started. Press Ctrl-C to exit.");
Console.Write("> ");
ReadLine();

static void ReadLine()
{
    foreach (var key in ReadForever())
    {
        // comment out the following line to avoid the garbled characters.
        _ = Console.CursorTop;

        Console.Write(key);
    }
}

static IEnumerable<string> ReadForever()
{
    while (true)
    {
        var key = Console.ReadKey(true);

        if (!Console.KeyAvailable)
        {
            yield return key.KeyChar.ToString();
            continue;
        }

        // If the user pastes text, we see it as a bunch of key presses. We don't want to send
        // them all individually, as it will trigger unnecessary processing. Instead, batch
        // them up to send as single "pasted text" block.
        var keys = new List<ConsoleKeyInfo> { key };
        do
        {
            keys.Add(Console.ReadKey(true));
        } while (Console.KeyAvailable);

        yield return string.Join("", keys.Select(k => k.KeyChar));
    }
}
```

**Screen Recording of reproduction steps**

[!Screen Recording of reproduction steps](https://github.com/dotnet/runtime/assets/97195/272704e6-ef65-48c1-8bb0-fbf0ebaec9dc)

### Expected behavior

Should be able to call `Console.CursorTop` without affecting `Console.ReadKey`.

### Actual behavior

`Console.CursorTop` interferes with `Console.ReadKey`.

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

I have reports of users that are running Linux natively on their machines running into this issue. I have reproduced it with the following WSL2 configuration:

- WSL2 with Ubuntu: Linux 5.15.90.1-microsoft-standard-WSL2 x86_64
- .NET 7.0.305 (I also tried this with the latest .NET 8 pre-release)
- Windows Terminal 1.17.11461.0

### Other information

From debugging, it's an input issue, not an output issue. The char being read from Console.ReadKey has the unicode replacement character:

image

tmds:
Characters may be split across buffers that are handled separately so we need to use a decoder to track state.

Though it affects ReadLine, no issue has been reported for it because the default buffer of 1000 bytes is sufficient to fit typical terminal input.

The problem was reported against the Console Cursor API. The cursor position handling uses single byte read buffers. Any character that needs multiple bytes therefore gets split accross several buffers.
For the issue to occur, such characters needed to be on stdin while parsing the cursor position.

Fixes https://github.com/dotnet/runtime/issues/88343.

@adamsitnik @stephentoub @dotnet/area-system-console ptal.

cc @waf

### DIFF
--- a/src/libraries/System.Console/src/System/IO/StdInReader.cs
+++ b/src/libraries/System.Console/src/System/IO/StdInReader.cs
@@ -22,6 +22,7 @@ namespace System.IO
         private readonly StringBuilder _readLineSB; // SB that holds readLine output.  This is a field simply to enable reuse; it's only used in ReadLine.
         private readonly Stack<ConsoleKeyInfo> _tmpKeys = new Stack<ConsoleKeyInfo>(); // temporary working stack; should be empty outside of ReadLine
         private readonly Stack<ConsoleKeyInfo> _availableKeys = new Stack<ConsoleKeyInfo>(); // a queue of already processed key infos available for reading
+        private readonly Decoder _decoder;
         private readonly Encoding _encoding;
         private readonly Encoder _echoEncoder;
         private Encoder? _bufferReadEncoder;
@@ -41,6 +42,7 @@ namespace System.IO
             _endIndex = 0;
             _readLineSB = new StringBuilder();
             _echoEncoder = _encoding.GetEncoder();
+            _decoder = _encoding.GetDecoder();
         }
 
         /// <summary> Checks whether the unprocessed buffer is empty. </summary>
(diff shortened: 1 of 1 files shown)


adamsitnik (approved):
I love such simple fixes 🥇  Thank you for your help @tmds!

waf:
This is great, thanks for figuring this out!

PR #95015, opened 2023-11-20
[wasm] Set frozen cache emscripten variable conditionally

radekdoulik:
We should not set the FROZEN_CACHE variable when using a different cache than the default one from our nuget.

That should also help with other issues related to emscripten cache and act as a workaround.


lewing (approved):

radical:
How is a different cache path used? Should we instead be checking if the property has a non-default value (IOW, not one from the cache nuget)?

radical (approved):

PR #95016, opened 2023-11-20
Add Predicate register placeholder methods and Z registers

kunalspathak:
- Added placeholder methods for predicate registers. This is taken from https://github.com/dotnet/runtime/pull/94529
- Added "Z" register display. For now, I have added `EA_SCALABLE` to differentiate these registers. It can change in future.


src/coreclr/jit/emitarm64.cpp
```diff
@@ -10571,6 +10601,48 @@ void emitter::emitIns_Call(EmitCallType          callType,
...
+    assert((ureg >= 0) && (ureg <= 15));
+    return ureg;
+}
+
+/*****************************************************************************
+ *
+ *  Returns an encoding for the specified register used in the 'Pn' position
+ */
+
+/*static*/ emitter::code_t emitter::insEncodeReg_Pn(regNumber reg)
```
a74nh, line 10623: Seems a shame we can't reuse insEncodeReg_Zn etc - the function is very similar.

kunalspathak: The only differentiator is the following portion: Could use macro to share most of that code, but don't think it is worth it (for now).

```c++
    assert(emitter::isPredicateRegister(reg));
    emitter::code_t ureg = (emitter::code_t)reg - (emitter::code_t)REG_V0;
```

kunalspathak:
@dotnet/jit-contrib

src/coreclr/jit/instr.h
```diff
@@ -418,8 +418,10 @@ enum emitAttr : unsigned
                 EA_4BYTE         = 0x004,
                 EA_8BYTE         = 0x008,
                 EA_16BYTE        = 0x010,
-
-#if defined(TARGET_XARCH)
+#if defined(TARGET_ARM64)
+                EA_SCALABLE      = 0x020,
```
a74nh, line 422: Intention here is to use scalable alongside another size? For example, for an SVE vector of ints:
```csharp
    theEmitter->emitIns_R_R_R(INS_sve_and, EA_1BYTE|EA_SCALABLE, REG_V0, REG_V1, REG_V2, INS_OPTS_NONE);  // AND     <Zdn>.<T>, <Pg>/M, <Zdn>.<T>, <Zm>.<T>
```

src/coreclr/jit/emitarm64.cpp
BruceForstall: I think it's not worth sharing. The asserts will be different even if the code is similar. It'll probably either get inlined or comdat folded in a release build anyway.

BruceForstall (approved):

BruceForstall:
Were you going to add the predicate registers `REGALIAS` to this PR or another?

src/coreclr/jit/instr.h
kunalspathak: I think we need to add the following lane size enum entries in `insOpts` to represent number of lanes in scalable registers.
```
INS_OPTS_SCALABLE_B,
INS_OPTS_SCALABLE_H,
INS_OPTS_SCALABLE_S,
INS_OPTS_SCALABLE_D,
```

Then, for your example, it would be:

```c++
   theEmitter->emitIns_R_R_R(INS_sve_and, EA_SCALABLE, REG_V0, REG_V1, REG_V2, INS_OPTS_SCALABLE_B);  // AND     <Zdn>.<T>, <Pg>/M, <Zdn>.<T>, <Zm>.<T>
```

kunalspathak:
> Were you going to add the predicate registers `REGALIAS` to this PR or another?

Added. Can you pleae take another look?

BruceForstall (approved):
LGTM

src/coreclr/jit/instr.h
```diff
@@ -268,6 +268,11 @@ enum insOpts : unsigned
     INS_OPTS_1D,
     INS_OPTS_2D,
 
+    INS_OPTS_SCALABLE_B,
+    INS_OPTS_SCALABLE_H,
+    INS_OPTS_SCALABLE_S,
+    INS_OPTS_SCALABLE_D,
+
```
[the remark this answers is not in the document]
a74nh: Never mind, I've just seen `EA_SCALABLE`. Which gives:

```
theEmitter->emitIns_R_R_R(INS_sve_and, EA_SCALABLE, REG_Z0, REG_P1, REG_Z2, INS_OPTS_SCALABLE_S);
```

PR #95023, opened 2023-11-20
[perf] Run localized Blazor scenario

ilonatommy:
Change for running the benchmark from https://github.com/dotnet/performance/pull/3470.
Runs the app in two variants: without and with AOT.

Because we cannot pass the same argument multiple times (e.g. `pre.py publish --msbuild /p:_TrimmerDumpDependencies=true --msbuild /warnaserror:NU1602,NU1604` ignores all the repeated args but the last one - `warnaserror:NU1602,NU1604`), we are saving the args to MSBuild property and append to it when AOT needs to be switched on.

Should be merged after the changes in the perf repo.

### DIFF
--- a/eng/testing/performance/blazor_perf.proj
+++ b/eng/testing/performance/blazor_perf.proj
@@ -87,14 +87,14 @@
     </HelixWorkItem>
     <HelixWorkItem Include="SOD - Localized App - Publish$(RunConfigsString)">
       <PayloadDirectory>$(WorkItemDirectory)</PayloadDirectory>
-      <Command>cd $(BlazorLocalizedDirectory) &amp;&amp; $(PublishCommand) -f $(PerflabTargetFrameworks) &amp;&amp; $(Python) test.py sod --scenario-name &quot;%(Identity)&quot;</Command>
+      <Command>cd $(BlazorLocalizedDirectory) &amp;&amp; $(PublishCommand) -f $(PerflabTargetFrameworks) &amp;&amp; $(Python) test.py sod --scenario-name &quot;%(Identity)&quot; --dirs $(PizzaAppPubLocation)</Command>
       <PostCommands>$(Python) post.py --readonly-dotnet</PostCommands>
       <Timeout>1:00</Timeout>
     </HelixWorkItem>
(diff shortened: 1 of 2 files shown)


ilonatommy:
The failure on AOTed pizza `python3: can't open file '/home/helixbot/post.py': [Errno 2] No such file or directory` should not be connected.

radical:
These pass on `main`, so shouldn't be unrelated. Looking at:

```
+ export PYTHONPATH=/datadisks/disk1/work/9F1B089F/w/B59509DB/u/scripts:/datadisks/disk1/work/9F1B089F/w/B59509DB/u
+ chmod +x /datadisks/disk1/work/9F1B089F/w/B59509DB/u/SOD/SizeOnDisk
+ cd
+ export RestoreAdditionalProjectSources=/datadisks/disk1/work/9F1B089F/p/built-nugets
+ python3 pre.py publish --has-workload --readonly-dotnet --msbuild /p:_TrimmerDumpDependencies=true;/p:HybridGlobalization=False;/warnaserror:NU1602,NU1604 --msbuild-static AdditionalMonoLinkerOptions="$(AdditionalMonoLinkerOptions) --dump-dependencies" --binlog /datadisks/disk1/work/9F1B089F/w/B59509DB/uploads/blazor_publish.binlog -f net9.0
python3: can't open file '/home/helixbot/pre.py': [Errno 2] No such file or directory
```

.. notice the `cd` with no directory, that would reset the path to home directory - thus `/home/helixbot`, and thus causing the failure.

eng/testing/performance/blazor_perf.proj
```diff
@@ -36,7 +38,7 @@
     <BlazorPizzaDirectory>$(ScenarioDirectory)blazorpizza/</BlazorPizzaDirectory>
     <BlazorMinAOTDirectory>$(ScenarioDirectory)blazorminappaot/</BlazorMinAOTDirectory>
     <BlazorAOTDirectory>$(ScenarioDirectory)blazoraot/</BlazorAOTDirectory>
-    <BlazorPizzaAOTDirectory>$(ScenarioDirectory)blazorpizzaaot/</BlazorPizzaAOTDirectory>
```
radical, line 39: The failure is because of this getting removed.

radical (approved):

Issue #95025, opened 2023-11-20
Assertion failed 'genCountBits(availableSet) == 1'


BruceForstall:
arm64

https://dev.azure.com/dnceng-public/public/_build/results?buildId=475010&view=ms.vss-test-web.build-test-results-tab

```
DOTNET_TieredCompilation=0
DOTNET_JitStress=2
DOTNET_JitStressRegs=8
```

```
18:56:26.028 Running test: _AdvSimd_r::JIT.HardwareIntrinsics.Arm._AdvSimd.Program.StoreSelectedScalar_Vector64_Byte_7()
Beginning scenario: RunBasicScenario_UnsafeRead
Beginning scenario: RunBasicScenario_Load
Beginning scenario: RunReflectionScenario_UnsafeRead

Assert failure(PID 22 [0x00000016], Thread: 22 [0x0016]): Assertion failed 'genCountBits(availableSet) == 1' in 'System.Runtime.Intrinsics.Arm.AdvSimd:StoreSelectedScalar(ulong,System.Runtime.Intrinsics.Vector64`1[ubyte],ubyte)' during 'Generate code' (IL size 9; hash 0x2253e0a5; FullOpts)

    File: /__w/1/s/src/coreclr/jit/gentree.cpp Line: 26535
    Image: /root/helix/work/correlation/corerun

[createdump] Gathering state for process 22 corerun
[createdump] Crashing thread 0016 signal 6 (0006)
apply_reg_state: ip and cfa unchanged; stopping here (ip=0xff68144f7228)
apply_reg_state: ip and cfa unchanged; stopping here (ip=0xff68144f7228)
apply_reg_state: ip and cfa unchanged; stopping here (ip=0xff68144f7228)
[createdump] Writing crash report to file /home/helixbot/dotnetbuild/dumps/coredump.22.dmp.crashreport.json
[createdump] Crash report successfully written
[createdump] Writing minidump with heap to file /home/helixbot/dotnetbuild/dumps/coredump.22.dmp
[createdump] Written 241860608 bytes (59048 pages) to core file
[createdump] Target process is alive
[createdump] Dump successfully written in 412ms
waitpid() returned successfully (wstatus 00000000) WEXITSTATUS 0 WTERMSIG 0
App Exit Code: 22
Expected: 100
Actual: 22
END EXECUTION - FAILED
```

@kunalspathak 
cc @dotnet/jit-contrib

PR #95034, opened 2023-11-20
JIT: Keep fgBBcount up to date

jakobbotsch:
Ensure fgBBcount is always the exact count of basic blocks in the linked list of blocks. Assert this in the post phase checks.

The goal is to avoid renumbering blocks simply as a means of getting an up-to-date block count. Remove one place that does this (if-conversion). We should be able to remove another in morph, but I want to do it separately. Also, it is a prerequisite to a more general purpose and more liberally used dead block elimination pass that I want to look at implementing.

### DIFF
--- a/src/coreclr/jit/compiler.h
+++ b/src/coreclr/jit/compiler.h
@@ -4491,7 +4491,7 @@ public:
                                     // created.
     BasicBlockList* fgReturnBlocks; // list of BBJ_RETURN blocks
     unsigned        fgEdgeCount;    // # of control flow edges between the BBs
-    unsigned        fgBBcount;      // # of BBs in the method
+    unsigned        fgBBcount;      // # of BBs in the method (in the linked list that starts with fgFirstBB)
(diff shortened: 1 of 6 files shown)


jakobbotsch:

Failures are known (apart from the superpmi-replay ones, which are #95025).

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=476214&view=ms.vss-build-web.run-extensions-tab). The smaller ones seem to be because this renumbering that is being removed from if-conversion can have an impact on LSRA. I do not think that is something that if-conversion should be concerning itself with.

Larger ones are from switching from `rbp` to `rsp` based frames due to the use of `fgBBcount` in this code:
https://github.com/dotnet/runtime/blob/3b61ad9e3acb6302fd2fcef4947373475a0dcc29/src/coreclr/jit/regalloc.cpp#L215-L219

BruceForstall (approved):

PR #87087, opened 2023-06-03
Add managed entry points for raising Contention events

kouvel:
- Added the events similarly to how the PortableThreadPool events are currently set up
- The events are not raised anywhere from the managed side yet, it's expected that they will eventually be raised by a new Lock type and at least by NativeAOT's Lock

PR #95035, opened 2023-11-20
Fix managed ContentionStart data

verdie-g:
The managed ContentionStart event was recently added and I think it's only used for AOT so the impact was probably low.


verdie-g:
@kouvel small issue from https://github.com/dotnet/runtime/pull/87087

kouvel (approved):
Nice catch, thanks!

PR #87634, opened 2023-06-15
Limit number of tracked values in trim dataflow analysis

sbomer:
Fixes https://github.com/dotnet/runtime/issues/87596 by placing a limit on the number of values we track in dataflow analysis.

PR #95041, opened 2023-11-20
Fix ILLink/ILC hang when tracking too many hoisted local values

### LINKED ISSUE #94831: [.NET 8] <PublishTrimmed> no longers work but throws build process for an infinite loop
### Description

So build I built my ARM64 binary by
`dotnet publish -c Release -r linux-arm64 -p:PublishSingleFile=true -p:SelfContained=true -p:PublishReadyToRun=true MyProject.csproj`

Here are my project settings:
```
	<PropertyGroup>
		<OutputType>Exe</OutputType>
		<VersionPrefix>1.10.14</VersionPrefix>
		<RunAnalyzersDuringLiveAnalysis>False</RunAnalyzersDuringLiveAnalysis>
		<RunAnalyzersDuringBuild>False</RunAnalyzersDuringBuild>
		<EnableNETAnalyzers>False</EnableNETAnalyzers>
		<ProduceReferenceAssembly>False</ProduceReferenceAssembly>
		<PublishTrimmed>true</PublishTrimmed>
	</PropertyGroup>
```

Turns out `<PublishTrimmed>true</PublishTrimmed>` no longers works: the build which tools ~3 minutes before now did not terminate the 'optimization phase' after > 9 hours.

What's going wrong and how could I trim my build as before?

### Reproduction Steps

Please see above

### Expected behavior

Trimming should work with .NET8 as it worked with .NET 7

### Actual behavior

Trimming does not work with .NET8 as it worked with .NET 7

### Regression?

Yes

### Known Workarounds

None

### Configuration

_No response_

### Other information

_No response_

sbomer:
https://github.com/dotnet/runtime/pull/87634 introduced a limit on the number of values tracked in dataflow analysis, but this approach was incorrect because resetting the set of tracked values was effectively moving up the dataflow lattice, breaking the invariant and causing potential hangs.

The hang was observed in https://github.com/dotnet/runtime/issues/94831, where (as found by @vitek-karas) the hoisted local `state` field of an async method with many await points was getting a large number of tracked values, hitting the limit, being reset to "empty", but then growing again, causing the analysis not to converge.

The fix is to instead introduce a special value `ValueSet<TValue>.Unknown` that is used for this case. It acts like "bottom" in the lattice, so that it destroys any other inputs on meet ("bottom" meet anything else is "bottom").

In this testcase the `state` field isn't involved in dataflow warnings, so this actually allows the method to be analyzed correctly, producing the expected warning for the `Type` local that flows across all of the await points.

Fixes https://github.com/dotnet/runtime/issues/94831.

### DIFF
--- a/src/tools/illink/src/ILLink.Shared/DataFlow/ValueSet.cs
+++ b/src/tools/illink/src/ILLink.Shared/DataFlow/ValueSet.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text;
 
@@ -12,12 +13,14 @@ using System.Text;
 
 namespace ILLink.Shared.DataFlow
 {
-	public readonly struct ValueSet<TValue> : IEquatable<ValueSet<TValue>>, IEnumerable<TValue>, IDeepCopyValue<ValueSet<TValue>>
+	public readonly struct ValueSet<TValue> : IEquatable<ValueSet<TValue>>, IDeepCopyValue<ValueSet<TValue>>
 		where TValue : notnull
 	{
 		const int MaxValuesInSet = 256;
 
 		public static ValueSet<TValue> Empty;
+
+		public static ValueSet<TValue> Unknown = new ValueSet<TValue> (ValueSetSentinel.Unknown);
 
 		// Since we're going to do lot of type checks for this class a lot, it is much more efficient
 		// if the class is sealed (as then the runtime can do a simple method table pointer comparison)
@@ -119,11 +122,25 @@ namespace ILLink.Shared.DataFlow
 			}
 		}
 
+		private struct Enumerable : IEnumerable<TValue>
+		{
+			private readonly object? _values;
+
+			public Enumerable (object? values) => _values = values;
+
+			public Enumerator GetEnumerator () => new (_values);
+
+			IEnumerator<TValue> IEnumerable<TValue>.GetEnumerator () => GetEnumerator ();
+
+			IEnumerator IEnumerable.GetEnumerator () => GetEnumerator ();
+		}
+
 		// This stores the values. By far the most common case will be either no values, or a single value.
 		// Cases where there are multiple values stored are relatively very rare.
 		//   null - no values (empty set)
 		//   TValue - single value itself
 		//   EnumerableValues typed object - multiple values, stored in the hashset
+		//   ValueSetSentinel.Unknown - unknown value, or "any possible value"
 		private readonly object? _values;
 
 		public ValueSet (TValue value) => _values = value;
@@ -132,8 +149,16 @@ namespace ILLink.Shared.DataFlow
 
 		private ValueSet (EnumerableValues values) => _values = values;
 
+		private enum ValueSetSentinel
+		{
+			Unknown
+		}
+
+		private ValueSet (ValueSetSentinel sentinel) => _values = sentinel;
+
 		public static implicit operator ValueSet<TValue> (TValue value) => new (value);
 
+		// Note: returns false for Unknown
 		public bool HasMultipleValues => _values is EnumerableValues;
 
 		public override bool Equals (object? obj) => obj is ValueSet<TValue> other && Equals (other);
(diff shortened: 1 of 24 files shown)


src/tools/illink/src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -12,13 +13,15 @@
...
 {
-	public readonly struct ValueSet<TValue> : IEquatable<ValueSet<TValue>>, IEnumerable<TValue>, IDeepCopyValue<ValueSet<TValue>>
+	public readonly struct ValueSet<TValue> : IEquatable<ValueSet<TValue>>, IDeepCopyValue<ValueSet<TValue>>
 		where TValue : notnull
 	{
 		const int MaxValuesInSet = 256;
 
 		public static ValueSet<TValue> Empty;
 
+		public static ValueSet<TValue> Unknown = new ValueSet<TValue> (ValueSetSentinel.Unknown);
```
marek-safar, line 23: Why not make it readonly?

src/tools/illink/src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -173,17 +208,20 @@ public override int GetHashCode ()
 			return _values.GetHashCode ();
 		}
 
-		public Enumerator GetEnumerator () => new (_values);
-
-		IEnumerator<TValue> IEnumerable<TValue>.GetEnumerator () => GetEnumerator ();
-
-		IEnumerator IEnumerable.GetEnumerator () => GetEnumerator ();
+		public IEnumerable<TValue> GetKnownValues () => new Enumerable (_values is ValueSetSentinel.Unknown ? null : _values);
```
marek-safar, line 211: Why not use everywhere reference check against Unknown field?

src/tools/illink/src/ILLink.Shared/DataFlow/ValueSet.cs
```diff
@@ -119,11 +122,25 @@ public void Reset ()
 			}
 		}
 
+		private struct Enumerable : IEnumerable<TValue>
```
marek-safar, line 125: readonly?

src/tools/illink/src/ILLink.Shared/TrimAnalysis/ValueExtensions.cs
```diff
@@ -50,10 +52,20 @@ internal static string ValueToString (this SingleValue value, params object[] ar
...
+			if (values.Count () != 1)
 				return null;
 
-			return node.Single ();
+			return values.Single ();
+		}
+
+		// ValueSet<TValue> is not enumerable. This helper translates ValueSet<SingleValue>.Unknown
+		// into a ValueSet<SingleValue> whose sole element is UnknownValue.Instance.
+		internal static IEnumerable<SingleValue> AsEnumerable (this MultiValue multiValue)
```
vitek-karas, line 64: I must admit I'm not super excited about the need for this... but I also don't know about a better way - other than teaching ValueSet about the "error" value for all value sets (which is not ideal either).

src/tools/illink/src/ILLink.Shared/TrimAnalysis/ValueExtensions.cs
```diff
@@ -50,10 +52,20 @@ internal static string ValueToString (this SingleValue value, params object[] ar
...
+		{
+			return multiValue.IsUnknown ()
+				? ImmutableArray.Create (UnknownValue.Instance)
```
vitek-karas, line 67: This will allocate on every call - would be better to have this precached as a static field.

src/tools/illink/src/ILLink.Shared/TrimAnalysis/ValueExtensions.cs
vitek-karas: The other disadvantage of this approach is that each enumeration will now allocate, actually twice, once to box the enumerable and then to create the enumerator. If value set implement enumerable, it can rely on the compiler optimization to avoid these instead.

vitek-karas: One other approach might be this (not pretty, but might work):
Make MultiValue a wrapper (since struct can't inherit) - so declare MultiValue as a real type, and make it wrap the ValueSet. And them make it implement IEnumerable directly.
Not sure if there are places though where we heavily rely on the ability to operate on ValueSets directly. On the other hand the conversion could be implicit (since wrapping doesn't induce allocation), so it might be possible handle this nicely almost everywhere.

vitek-karas: Or (just idea):
* Change ValueSetLattice to a class (do we need this to be a struct, does it make much of a difference?)
  * There some weird things about the lattice as is now
  * It's a struct, but we use it through an interface -> boxing it anyway
  * It's a struct, so can't be derived, thus the implementation of all of its methods is effectively `static`.
  * For other lattices it's benefitial to be instance based (we need it in some cases), but for the ValueSetLattice this seems unnecessary, so having it provide the Top/Bottom as statics should work (alongside the interface implementations). Or maybe it would just provide the singleton value (it could be a singleton, we don't need to allocate more than one since they're all the same).
* Have ValueSet be a generic over TLattice
* Make the "Error" value a static on the lattice - or more precisely, make the ValueSetLattice define both Top and Bottom (only the ValueSetLattice for now, we would not need it on all lattices).

This would make the design overall slightly cleaner. For example now we rely on the implicit rule that the `default` value of a ValueSet is the `Top` value of the lattice.

src/tools/illink/src/ILLink.Shared/TrimAnalysis/ValueExtensions.cs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation and contributors. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
+using System.Collections.Immutable;
```
vitek-karas, line 5: Nit: Do we need the Immutable namespace?

vitek-karas (approved):
Thanks!

PR #95053, opened 2023-11-21
Ensure ExceptionStringIds are handled

### LINKED ISSUE #94107: NativeAOT compiler fails when it tries to emit a throw for BadImageFormatException
If the input for the AOT compiler is invalid in some sense, it should generate a throwing method body, but not fail the compilation.
In some cases this fails. An example is if we hit this code path:
https://github.com/dotnet/runtime/blob/77bd2ae57c67d9a7377305d4d1e169277c01daac/src/coreclr/tools/Common/TypeSystem/Ecma/EcmaModule.cs#L154

(happens when there's an unknown token type in the input)

This throws an internal exception in the compiler and attaches one argument to it.

Eventually this makes its way into the helper which generates the throwing method body where it will trigger an assert: https://github.com/dotnet/runtime/blob/77bd2ae57c67d9a7377305d4d1e169277c01daac/src/coreclr/tools/Common/TypeSystem/IL/Stubs/TypeSystemThrowingILEmitter.cs#L78

This is because the throw helper this will end up calling is:
https://github.com/dotnet/runtime/blob/77bd2ae57c67d9a7377305d4d1e169277c01daac/src/coreclr/nativeaot/System.Private.CoreLib/src/Internal/Runtime/CompilerHelpers/ThrowHelpers.cs#L76

That takes only a single parameter, the exception string ID, but it doesn't take additional argument. But our internal exception has additional argument.

Eventually this ends up producing invalid IL - calls the helper with more arguments then it declares. This will eventually fail in the JIT and ends up failing the entire compilation with a cryptic error:
```
EXEC : error : One or more errors occurred. (Invalid IL or CLR metadata) []
  System.AggregateException: One or more errors occurred. (Invalid IL or CLR metadata)
   ---> Internal.TypeSystem.TypeSystemException+InvalidProgramException: Invalid IL or CLR metadata
     at Internal.TypeSystem.ThrowHelper.ThrowInvalidProgramException() + 0x24
  ....
```

A repro is:

`dotnet new console`

`app.csproj`:
```xml
    <PropertyGroup>
        <OutputType>Exe</OutputType>
        <ImplicitUsings>enable</ImplicitUsings>
        <Nullable>enable</Nullable>
        <PublishAot>true</PublishAot>
        <!-- <PublishTrimmed>true</PublishTrimmed> -->
        <WarningsAsErrors>false</WarningsAsErrors>
        <NoWarn>NU1605;NU1603;NU1701</NoWarn>
        <RefAssemblyName>Microsoft.AspNetCore.Identity.EntityFrameworkCore</RefAssemblyName>
        <RefPackageVersion>3.1.10</RefPackageVersion>
        <RefPackageName></RefPackageName>
        <RefPackageName Condition="'$(PackageName)'==''">$(RefAssemblyName)</RefPackageName>
    </PropertyGroup>

    <ItemGroup>
        <TrimmerRootAssembly Include="$(RefAssemblyName)" />

        <PackageReference Include="$(RefPackageName)" Version="$(RefPackageVersion)" />
    </ItemGroup>

    <Target Name="ConfigureTrimming"
            BeforeTargets="PrepareForILLink">
        <ItemGroup>
        <ManagedAssemblyToLink Condition="'%(Filename)' == '$(RefAssemblyName)'">
            <TrimmerSingleWarn>false</TrimmerSingleWarn>
        </ManagedAssemblyToLink>
        </ItemGroup>
    </Target>
</Project>
```

`dotnet publish`

There's a side question about invalid IL in our library... but that's a separate problem (and probably not relevant as this is a really old version of the library)

MichalStrehovsky:
I went over all the ExceptionStringID in the compiler and fixed them up to ensure we have code that translates them to messages.

Fixes #94107.

Cc @dotnet/ilc-contrib

### DIFF
--- a/src/coreclr/tools/aot/ILCompiler.ReadyToRun/JitInterface/CorInfoImpl.ReadyToRun.cs
+++ b/src/coreclr/tools/aot/ILCompiler.ReadyToRun/JitInterface/CorInfoImpl.ReadyToRun.cs
@@ -1889,7 +1889,7 @@ namespace Internal.JitInterface
             // a static method would have never found an instance method.
             if (originalMethod.Signature.IsStatic && isCallVirt)
             {
-                ThrowHelper.ThrowInvalidProgramException(ExceptionStringID.InvalidProgramCallVirtStatic, originalMethod);
+                ThrowHelper.ThrowMissingMethodException("?");
             }
 
             exactType = type;
@@ -2095,7 +2095,7 @@ namespace Internal.JitInterface
                     // Compensate for always treating delegates as direct calls above
                     !(isLdftn && isVirtualBehaviorUnresolved))
                 {
-                    ThrowHelper.ThrowInvalidProgramException(ExceptionStringID.InvalidProgramCallAbstractMethod, targetMethod);
+                    ThrowHelper.ThrowInvalidProgramException(ExceptionStringID.InvalidProgramSpecific, targetMethod);
                 }
 
                 bool allowInstParam = (flags & CORINFO_CALLINFO_FLAGS.CORINFO_CALLINFO_ALLOWINSTPARAM) != 0;
--- a/src/coreclr/tools/Common/TypeSystem/Common/ExceptionStringID.cs
+++ b/src/coreclr/tools/Common/TypeSystem/Common/ExceptionStringID.cs
@@ -34,8 +34,6 @@ namespace Internal.TypeSystem
         InvalidProgramSpecific,
         InvalidProgramVararg,
         InvalidProgramCallVirtFinalize,
-        InvalidProgramCallAbstractMethod,
-        InvalidProgramCallVirtStatic,
         InvalidProgramNonStaticMethod,
         InvalidProgramGenericMethod,
         InvalidProgramNonBlittableTypes,
@@ -43,7 +41,6 @@ namespace Internal.TypeSystem
 
         // BadImageFormatException
         BadImageFormatGeneric,
-        BadImageFormatSpecific,
 
         // MarshalDirectiveException
         MarshalDirectiveGeneric,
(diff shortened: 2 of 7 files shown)


src/coreclr/tools/aot/ILCompiler.ReadyToRun/JitInterface/CorInfoImpl.ReadyToRun.cs
```diff
@@ -1889,7 +1889,7 @@ private static bool IsTypeSpecForTypicalInstantiation(TypeDesc t)
             // a static method would have never found an instance method.
             if (originalMethod.Signature.IsStatic && isCallVirt)
             {
-                ThrowHelper.ThrowInvalidProgramException(ExceptionStringID.InvalidProgramCallVirtStatic, originalMethod);
+                ThrowHelper.ThrowMissingMethodException("?");
```
MichalStrehovsky, line 1892: This is what CoreCLR VM throws from JitInterface.

src/coreclr/tools/aot/ILCompiler.ReadyToRun/JitInterface/CorInfoImpl.ReadyToRun.cs
```diff
@@ -2095,7 +2095,7 @@ private static bool IsTypeSpecForTypicalInstantiation(TypeDesc t)
                     // Compensate for always treating delegates as direct calls above
                     !(isLdftn && isVirtualBehaviorUnresolved))
                 {
-                    ThrowHelper.ThrowInvalidProgramException(ExceptionStringID.InvalidProgramCallAbstractMethod, targetMethod);
+                    ThrowHelper.ThrowInvalidProgramException(ExceptionStringID.InvalidProgramSpecific, targetMethod);
```
MichalStrehovsky, line 2098: Same - this is just Bad IL in the CoreCLR VM. We don't need to be more detailed. ILVerify is the validator.

vitek-karas (approved):
Did you try the repro project from the issue?
Looks good - thanks a lot!

MichalStrehovsky:
> Did you try the repro project from the issue?

Yep, it now emits a harmless message `ILC: Method '[Microsoft.AspNetCore.Identity.EntityFrameworkCore]Microsoft.Extensions.DependencyInjection.IdentityEntityFrameworkBuilderExtensions.AddStores(IServiceCollection,Type,Type,Type)' will always throw because: The format of a DLL or executable being loaded is invalid` and finishes compiling successfully.

PR #95063, opened 2023-11-21
Add eng/pipelines/common/variables.yml to coreclr pipelines

akoeplinger:
Most of the pipelines were missing the variables template. This caused e.g. CodeQL to run on them inadvertently.

/cc @BruceForstall


BruceForstall (approved):
LGTM

Do you need to test one or more before merge?

akoeplinger:
I don't think so, this is a very simple change and since the pipelines are independent testing one doesn't really tell us much.

I'll keep an eye on the pipelines :)

PR #95065, opened 2023-11-21
Introduce CallConvSwift and special register types to represent the Swift ABI calling convention

### LINKED ISSUE #64215: Introduce `CallConvSwift` to represent the Swift ABI calling convention
## Background and Motivation

The Swift programming language has a different ABI, runtime environment, and object model, making it challenging to call from the .NET without the runtime support. Ideally, we want to avoid generating extra wrappers in the objection tools and attempt to directly call all kinds of Swift functions.

For interop developers of managed APIs wrapping native Swift API layers, reconciling the Swift ABI with the existing P/Invoke options is close to intractable. The proposal would be to add built-in support for the Swift ABI via P/Invokes and Reverse P/Invokes through [`UnmanagedCallConvAttribute`](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.unmanagedcallconvattribute). This would directly help out binding efforts in the MAUI space.

## Proposed API

According to the calling convention, the `self` context has dedicated registers, and it is always passed through them since it's heavily used. Methods calling other methods on the same object can share the self context.

Here are cases when `self` context is passed via register:
 - Instance methods on class types: pointer to self
 - Class methods: pointer to type metadata (which may be subclass metadata)
 - Mutating method on value types: pointer to the value (i.e. value is passed indirectly)
 - Non-mutating methods on value types: self may fit in one or more registers, else passed indirectly
 - Thick closures, i.e. closures requiring a context: the closure context

Error handling is also handled through registers, so the caller needs to check for `error` and throw if necessary. Implementing P/Invoke thunks should simplify registers juggling by using predefined set of registers in these scenarios.

```diff
namespace System.Runtime.CompilerServices
{
+    public class CallConvSwift
+    {
+        public CallConvSwift() { }
+    }
}
```
Based on the [design doc](https://github.com/dotnet/designs/blob/main/proposed/swift-interop.md) we want to introduce types to represent each of the special registers. 
```diff
+ namespace System.Runtime.InteropServices.Swift
+ {
+     public readonly struct SwiftSelf
+     {
+         public SwiftSelf(IntPtr value)
+         {
+             Value = value;
+         }
+
+         public IntPtr Value { get; }
+     }
+
+     public readonly struct SwiftError
+     {
+         public SwiftError(IntPtr value)
+         {
+             Value = value;
+         }
+
+         public IntPtr Value { get; }
+     }
+ }

```

## Usage Examples

This API would be used in the P/Invoke and Reverse P/Invoke syntax as follows:

```csharp
using System.Runtime.InteropServices.Swift;

[UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvSwift) })]
[DllImport("SwiftLibrary", EntryPoint = "export")]
public unsafe static extern nint conditionallyThrowError(bool willThrow, SwiftError* error);

[UnmanagedCallConv(CallConvs = new Type[] { typeof(CallConvSwift) })]
[DllImport("SwiftLibrary", EntryPoint = "export")]
public static extern nint getMagicNumber(SwiftSelf self);
```

Using the `CallConv*` approach would also permit use in C# `unmanaged` function pointers. This would follow the Swift ABI during dispatch. Documentation for the Swift ABI can be found [here](https://github.com/apple/swift/tree/main/docs/ABI) and its design philosophy [here](https://github.com/apple/swift/blob/main/docs/ABIStabilityManifesto.md).

/cc @dotnet/jit-contrib @dotnet/interop-contrib @Redth

kotlarmilos:
## Description

Based on the [design doc](https://github.com/dotnet/designs/blob/main/proposed/swift-interop.md), this PR introduces types to represent each of the special registers.

The Swift programming language has a different ABI, runtime environment, and object model. According to the calling convention, the self context has dedicated registers, and it is always passed through them since it's heavily used. Methods calling other methods on the same object can share the self context. Errors are also handled through registers, so the caller needs to check for errors and throw if necessary. Implementing P/Invoke thunks should simplify registers juggling by using predefined set of registers in these scenarios. 

The Mono runtime currently does not compile native wrappers with the LLVM backend. However, support for LLVM is essential for handling scenarios involving `UnmanagedCallersOnly`, where Swift code calls into C#. The implementation of LLVM support will be addressed in a subsequent PRs.

We need approval for the https://github.com/dotnet/runtime/issues/64215 before merging this PR.

Fixes https://github.com/dotnet/runtime/issues/64215


src/libraries/System.Runtime/ref/System.Runtime.cs
```diff
@@ -13811,6 +13815,19 @@ public static unsafe class SpanMarshaller<T, TUnmanagedElement> where TUnmanaged
...
+{
+    public readonly struct SwiftSelf
+    {
+        public SwiftSelf(IntPtr value) { }
+        public IntPtr Value { get; }
+    }
+    public readonly struct SwiftError
+    {
+        public SwiftError(IntPtr value) { }
+        public IntPtr Value { get; }
```
AaronRobinsonMSFT, line 13828:
```suggestion
        public SwiftSelf(System.IntPtr value) { }
        public System.IntPtr Value { get; }
    }
    public readonly struct SwiftError
    {
        public SwiftError(System.IntPtr value) { }
        public System.IntPtr Value { get; }
```

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/CallingConventions.cs
```diff
@@ -15,6 +15,10 @@ public class CallConvStdcall
     {
         public CallConvStdcall() { }
     }
+    public class CallConvSwift
+    {
+        public CallConvSwift() { }
```
AaronRobinsonMSFT, line 20:
```suggestion
        /// <summary>
        /// Initializes a new instance of the <see cref="CallConvSwift" /> class.
        /// </summary>
        public CallConvSwift() { }
```

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/CallingConventions.cs
```diff
@@ -15,6 +15,10 @@ public class CallConvStdcall
     {
         public CallConvStdcall() { }
     }
+    public class CallConvSwift
```
AaronRobinsonMSFT, line 18:
```suggestion
    /// <summary>
    /// Indicates that a method should using the <see href="https://github.com/apple/swift/blob/main/docs/ABIStabilityManifesto.md#calling-convention">Swift</see>calling convention.
    /// </summary>
    public class CallConvSwift
```

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Swift/SwiftTypes.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// </remarks>
+    public readonly struct SwiftSelf
+    {
+        /// <summary>
+        /// Creates a new instance of the SwiftSelf struct with the specified pointer value.
+        /// </summary>
+        /// <param name="value">The pointer value representing the self context.</param>
+        public SwiftSelf(IntPtr value) {
+            Value = value;
+        }
```
AaronRobinsonMSFT, line 30:
```suggestion
        public SwiftSelf(IntPtr value)
        {
            Value = value;
        }
```

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Swift/SwiftTypes.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// </remarks>
+    public readonly struct SwiftError
+    {
+        /// <summary>
+        /// Creates a new instance of the SwiftError struct with the specified pointer value.
+        /// </summary>
+        /// <param name="value">The pointer value representing the error context.</param>
+        public SwiftError(IntPtr value) {
+            Value = value;
+        }
```
AaronRobinsonMSFT, line 61:
```suggestion
        public SwiftError(IntPtr value)
        {
            Value = value;
        }
```

jkoritzinsky (approved):

SamMonoRT (approved):

matouskozak (approved):

kotlarmilos:
@AaronRobinsonMSFT please sign off on this from the Interop side.

AaronRobinsonMSFT:
@kotlarmilos Note that unless @jkoritzinsky specifically asks for my sign-off, his review is sufficient for any interop related PRs. I do appreciate your dilligence in getting all the stakeholders involved though. Thanks!

AaronRobinsonMSFT (approved):

PR #95085, opened 2023-11-21
JIT: Add a "DFS blocks" phase

jakobbotsch:
Factor out SSA's general DFS that builds a post-order and make it its own pass. Use it to eliminate dead blocks as well and invoke it in a few different places within the JIT. The computed post-orders are not used yet, but I have plans to use it in local morph and in physical promotion. We should also be able to switch cross block local morph to use it instead of computing its own RPO. Finally, I want to see if this can replace some of the dense reachability computations that loop finding is doing... it's not totally clear to me yet.


jakobbotsch:
#95251 does the main chunk of this. I'll add the unreachable part once I get to a point where I want to start using the reverse post-order in more places.

PR #95090, opened 2023-11-21
[release/6.0-staging] Permit MD5 regardless of FIPS configuration for Linux

vcsjones:
Backport of #94934 to release/6.0-staging

/cc @bartonjs

## Customer Impact

Customers legitimately using the MD5 algorithm for non-cryptographic purposes get a `CryptographicException` on certain Linux configurations; including RHEL (with an opt-in) and Mariner (the new default?).

The main known purpose for continuing to use MD5 is to set the Content-MD5 header required on uploading to Azure Blob Storage.

## Testing

Verified by running the MD5 tests on a machine in this configuration.  CI does not currently have such a configuration.

## Risk

Low.

Customers not using MD5 are entirely unaffected.  Customers using MD5 on a system without a FIPS lockout are verified by CI.  Customers using MD5 on a system with a FIPS lockout are already broken, and have been manually verified as being unbroken.

The affected codepaths are initialization, they do not process arguments, so there are no argument-based edge cases.


jeffhandley (approved):

bartonjs (approved):

carlossanlop:

PR #95098, opened 2023-11-22
Maintain a per-thread cache of pinning GC handles

stephentoub:
(I'm not sure if this is worthwhile... opening it for a discussion...)

@Maoni0 and I were in a discussion today about a particular workload that was doing a lot of interop with {ReadOnly}Memory and as a result doing a lot {ReadOnly}Memory.Pin which in turn resulted in a lot of GCHandle.Alloc and in turn a lot of contention. Maoni inquired about the possibility of having a cache of pinning GCHandles to take the edge off.  I've implemented that here.

| Method                          | Job        | Toolchain         | Mean              | Ratio |
|-------------------------------- |----------- |------------------ |------------------:|------:|
| AllocFree                       | Job-YUHNJU | \main\corerun.exe |          36.32 ns |  1.00 |
| AllocFree                       | Job-LZRCOY | \pr\corerun.exe   |          26.45 ns |  0.70 |
|                                 |            |                   |                   |       |
| ParallelAllocFree               | Job-YUHNJU | \main\corerun.exe |  30,591,619.23 ns |  1.00 |
| ParallelAllocFree               | Job-LZRCOY | \pr\corerun.exe   |   5,154,509.97 ns |  0.16 |
|                                 |            |                   |                   |       |
| AsyncWithYieldAllocFree         | Job-YUHNJU | \main\corerun.exe | 132,349,354.76 ns |  1.00 |
| AsyncWithYieldAllocFree         | Job-LZRCOY | \pr\corerun.exe   | 129,156,684.62 ns |  0.98 |
|                                 |            |                   |                   |       |
| AllocFreeMany                   | Job-YUHNJU | \main\corerun.exe |     712,603.22 ns |  1.00 |
| AllocFreeMany                   | Job-LZRCOY | \pr\corerun.exe   |     676,165.12 ns |  0.95 |

```C#
using System.Buffers;

public partial class Tests
{
    private ReadOnlyMemory<int> _rom = new int[1];
    private MemoryHandle[] _manyHandles = new MemoryHandle[10_000];

    [Benchmark]
    public MemoryHandle AllocFree()
    {
        MemoryHandle handle = _rom.Pin();
        handle.Dispose();
        return handle;
    }

    [Benchmark]
    public void ParallelAllocFree()
    {
        Parallel.For(0, 1_000_000, i =>
        {
            _rom.Pin().Dispose();
        });
    }

    [Benchmark]
    public Task AsyncWithYieldAllocFree()
    {
        return Parallel.ForAsync(0, 1_000_000, async (i, ct) =>
        {
            MemoryHandle handle = _rom.Pin();
            await Task.Yield();
            handle.Dispose();
        });
    }

    [Benchmark]
    public void AllocFreeMany()
    {
        MemoryHandle[] handles = _manyHandles;
        ReadOnlyMemory<int> rom = _rom;
        
        for (int i = 0; i < handles.Length; i++)
        {
            handles[i] = rom.Pin();
        }

        for (int i = 0; i < handles.Length; i++)
        {
            handles[i].Dispose();
        }
    }
}
```

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/GCHandle.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/GCHandle.cs
@@ -49,7 +49,7 @@ namespace System.Runtime.InteropServices
         /// <summary>Per-thread cache of pinning handles.</summary>
         /// <remarks>Every handle in the cache is a GCHandleType.Pinned handle with a null target.</remarks>
         [ThreadStatic]
-        private static FixedCapacityIntPtrStack? s_pooledPinningHandles;
+        private static PerThreadHandleCache? s_pooledPinningHandles;
 
         /// <summary>Allocates a pinning handle whose target is the specified object.</summary>
         /// <param name="value">The object to which the allocated handle should reference.</param>
@@ -57,7 +57,7 @@ namespace System.Runtime.InteropServices
         private static IntPtr AllocPinnedWithPooling(object? value)
         {
             // If we have a cache and we can take a handle from it, do so.
-            if (s_pooledPinningHandles is FixedCapacityIntPtrStack handles && handles.TryPop(out IntPtr handle))
+            if (s_pooledPinningHandles is PerThreadHandleCache handles && handles.TryPop(out IntPtr handle))
             {
                 // We successfully got a handle from the cache.  Update it to point to the new target.
                 InternalSet(GetHandleValue(handle), value);
@@ -79,7 +79,7 @@ namespace System.Runtime.InteropServices
             Debug.Assert(IsPinned(handle));
 
             // Try to store the handle into the cache rather than actually freeing it.
-            FixedCapacityIntPtrStack handles = s_pooledPinningHandles ??= new();
+            PerThreadHandleCache handles = s_pooledPinningHandles ??= new();
             if (handles.TryPush(handle))
             {
                 // Clear the handle's target so that it doesn't keep the object pinned.
@@ -98,14 +98,23 @@ namespace System.Runtime.InteropServices
             }
         }
 
-        /// <summary>Provides a simple stack of IntPtrs with a fixed capacity.</summary>
+        /// <summary>Provides a simple stack for caching handles on a particular thread.</summary>
         /// <param name="capacity">The maximum number of items the stack can hold.</param>
-        private sealed class FixedCapacityIntPtrStack(int capacity = 8) // 8 == arbitrary limit that can be tuned over time
+        private sealed class PerThreadHandleCache(int capacity = 8) // 8 == arbitrary limit that can be tuned over time
         {
             /// <summary>Array of values.</summary>
             private readonly IntPtr[] _values = new IntPtr[capacity];
             /// <summary>Number of valid cached handles in <see cref="_values"/>.</summary>
             private int _count;
+
+            /// <summary>Release all handles cached on the thread when the thread goes away.</summary>
+            ~PerThreadHandleCache()
+            {
+                for (int i = 0; i < _count; i++)
+                {
+                    InternalFree(GetHandleValue(_values[i]));
+                }
+            }
 
             /// <summary>Tries to push a value onto the stack.</summary>
             /// <returns>true if the value was stored; otherwise, false.</returns>
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/GCHandle.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/GCHandle.cs
@@ -41,15 +41,108 @@ namespace System.Runtime.InteropServices
                 throw new ArgumentException(SR.ArgumentException_NotIsomorphic, nameof(value));
             }
 
-            IntPtr handle = InternalAlloc(value, type);
+            _handle = type == GCHandleType.Pinned ?
+                AllocPinnedWithPooling(value) :
+                InternalAlloc(value, type);
+        }
 
-            if (type == GCHandleType.Pinned)
+        /// <summary>Per-thread cache of pinning handles.</summary>
+        /// <remarks>Every handle in the cache is a GCHandleType.Pinned handle with a null target.</remarks>
+        [ThreadStatic]
+        private static FixedCapacityIntPtrStack? s_pooledPinningHandles;
+
+        /// <summary>Allocates a pinning handle whose target is the specified object.</summary>
+        /// <param name="value">The object to which the allocated handle should reference.</param>
+        /// <returns>The allocated handle.</returns>
+        private static IntPtr AllocPinnedWithPooling(object? value)
+        {
+            // If we have a cache and we can take a handle from it, do so.
+            if (s_pooledPinningHandles is FixedCapacityIntPtrStack handles && handles.TryPop(out IntPtr handle))
             {
-                // Record if the handle is pinned.
-                handle = (IntPtr)((nint)handle | 1);
+                // We successfully got a handle from the cache.  Update it to point to the new target.
+                InternalSet(GetHandleValue(handle), value);
+            }
+            else
+            {
+                // We could not get a handle from the cache.  Allocate a new one.
+                handle = MarkPinned(InternalAlloc(value, GCHandleType.Pinned));
             }
 
-            _handle = handle;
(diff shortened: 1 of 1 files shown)


src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/GCHandle.cs
```diff
@@ -41,15 +41,108 @@ private GCHandle(object? value, GCHandleType type)
...
+            {
+                // The cache was full: actually free the handle.
+                InternalFree(GetHandleValue(handle));
+            }
+        }
 
-            if (type == GCHandleType.Pinned)
+        /// <summary>Provides a simple stack of IntPtrs with a fixed capacity.</summary>
+        /// <param name="capacity">The maximum number of items the stack can hold.</param>
+        private sealed class FixedCapacityIntPtrStack(int capacity = 8) // 8 == arbitrary limit that can be tuned over time
```
jkotas, line 103: Does this need a finalizer to prevent the handles from leaking when threads come and go?

stephentoub: Yup.

davidfowl:
Out of curiosity was it a server workload ?

jkotas:
The handle table is trying to affinitize the handles to cores. This cache is likely going to reduce effectiveness of this affinitization in longer running processes - we have similar problem with array pool as well. We are slowly giving up on scaling well on NUMA machines with these optimizations. (I do see it as a showstopper for these optimizations - just highlighting what we are giving up.)

cc @mangod9

stephentoub:
> Out of curiosity was it a server workload ?

Yes. Their exact workload is likely served better by not using ReadOnlyMemory.Pin at all, and instead using fixed with a span from the ROM.  We also pointed them to AllocateArray(..., pinned: true) + CreateFromPinnedArray.  They're trying both of those out, but my expectation is it'll address their issues better than this.  This is just fallout from the general discussion.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/GCHandle.cs
```diff
@@ -41,15 +41,108 @@ private GCHandle(object? value, GCHandleType type)
                 throw new ArgumentException(SR.ArgumentException_NotIsomorphic, nameof(value));
             }
 
-            IntPtr handle = InternalAlloc(value, type);
+            _handle = type == GCHandleType.Pinned ?
+                AllocPinnedWithPooling(value) :
+                InternalAlloc(value, type);
```
jkotas, line 46: This will impact profiler APIs and event tracing. Profiler APIs are not going to see all handle allocations and free as before. E.g. if the diagnostic tools has feature to detect handle leaks, the stacktraces associated with the leaking handles are going to be bogus.

jkotas:
The handle table has similar per-core cache here: https://github.com/dotnet/runtime/blob/91f8ffa67c413bfa8c28712f62e62e93988a0224/src/coreclr/gc/handletablecache.cpp#L722-L730 . I am wondering whether there are ways how to make the current handle table cache faster/better.

stephentoub:
Thanks, @jkotas.  That's all helpful.  As I noted, I'm not sold on this, so I'd be fine closing it if we don't think it's the right tradeoffs; I just promised Maoni I'd throw it together to evaluate :)

jkotas:
Something would need to be done about the impact on diagnostics.

Otherwise, I do not have a strong opinion about this change.

jkotas:
I think the poor performance of ParallelAllocFree is caused by this global per counter: https://github.com/dotnet/runtime/blob/91f8ffa67c413bfa8c28712f62e62e93988a0224/src/coreclr/gc/handletable.cpp#L309 . We should replace it with a scalable counter, similar to what we have done in other places.

jkotas:
> The handle table is trying to affinitize the handles to cores. 

This is done for server GC only: https://github.com/dotnet/runtime/blob/91f8ffa67c413bfa8c28712f62e62e93988a0224/src/coreclr/gc/objecthandle.cpp#L536-L542 . I expect that results for ParallelAllocFree microbenchmark are going to be very different for workstation GC vs. server GC.

En3Tho:
I wonder if with DATAS there is a need to do this workstation/server distinction? Afaik DATAS makes gc start with one heap (like workstation)   and then scale up further if needed. So in this case it might work for some time as server gc running in sorta workstation mode.

mangod9:
> in a lot of GCHandle.Alloc and in turn a lot of contention

Is the contention during allocation or Dispose?

Maoni0:
@mangod9 the contention is during both. 

@stephentoub were you running with Server GC or Workstation GC? how many cores does your machine have? also, can you please elaborate on your opinion on these results? you said you were not sold, was that because you thought the improvements weren't enough or were you worried about other aspects of this change? I have not take a close look but I would like to.

stephentoub:
> were you running with Server GC or Workstation GC?

Server

> how many cores does your machine have?

12 logical

> you said you were not sold, was that because you thought the improvements weren't enough or were you worried about other aspects of this change?

A combination of:
- We could probably improve other aspects to decrease the gap, like the ones Jan called out
- There's already a multilevel cache in place on the native side and this is layering on another cache; seems like it'd be better just to improve that original one rather than add a second in a different location
- There are the issues Jan calls out around diagnostics and scale
- Some scenerios involving tons of handles at high contention may be mitigated by other mechanisms, like using fixed if the pinning is scoped, or using pre-pimned buffers rather than pinning/unpinning for every operation

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/GCHandle.cs
Maoni0: handle creation/free will be the same as before - you still get the callback and the stress log msg when a handle is allocated from the handle table and freed to the handle table. and you can already change the Target of a handle today so this doesn't introduce a new scenario. and there's an event that gets fired when this happens.

Maoni0:
> We could probably improve other aspects to decrease the gap, like the ones Jan called out
> There's already a multilevel cache in place on the native side and this is layering on another cache; seems like it'd be better just to improve that original one rather than add a second in a different location

yes, there could be changes done in the handle table but they are much risker than this. and this could be a change done in ROM which was what I was thinking, since it can be such a heavy usage of pinned handles. doing this in GCHandle is fine for showing the effect of course.

I don't think there's any issue with diagnostics since this doesn't introduce a new scenario. I responded to Jan's comment above.

src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/GCHandle.cs
jkotas: I am talking about these profiler callbacks https://github.com/dotnet/runtime/blob/cbc501ca196371572c38f8d12a66969864d99c08/src/coreclr/inc/corprof.idl#L2311-L2321 . I believe that these callbacks were specifically designed to allow profilers to capture callstacks for when the GC handle is created and freed and detect leaking handles.

The cache of GC handles breaks the diagnostic value of these callbacks. The profiler will only get the callback when the handle is allocated for the first time from the GC handle table. I won't get callbacks when the handle is reused and so it won't be able to capture the user code callstack that allocated the handle.

Maoni0: yeah, I knew that's what you meant - those are the only profiling API callbacks for handle creation/free. 

today you can create a bunch of handles and just change their targets, including setting the targets to null which is exactly what happens here. I don't see this as a blocking factor for doing perf improvement. if this is really needed, we could add another callback when target is changed.

Maoni0:
> I think the poor performance of ParallelAllocFree is caused by this global perf counter:
> 
> https://github.com/dotnet/runtime/blob/91f8ffa67c413bfa8c28712f62e62e93988a0224/src/coreclr/gc/handletable.cpp#L309
> 
> . We should replace it with a scalable counter, similar to what we have done in other places.

yep, I agree this should be fixed - I thought we had fixed this already as I recall this issue came up before. increasing this global counter, if any, has a negative value. since we don't even interlock, it just gives a (sometimes extremely) inaccurate value so it's already misleading, and it's costly.

I'll make a fix.

PR #95105, opened 2023-11-22
Arm64/SVE: Add insEncodeReg* methods

kunalspathak:
Add bunch of `insEncodeReg*` methods because not all instructions have registers position at same position. See details at 
https://github.com/dotnet/runtime/issues/94549#issuecomment-1821981987. I have updated the [emitOutputInstr_sve.cpp.txt](https://github.com/dotnet/runtime/files/13434525/emitOutputInstr_sve.cpp.txt) in #94549 to call these methods.

### DIFF
--- a/src/coreclr/jit/emitarm64.cpp
+++ b/src/coreclr/jit/emitarm64.cpp
@@ -3768,6 +3768,7 @@ emitter::code_t emitter::emitInsCodeSve(instruction ins, insFormat fmt)
     }
 
     assert(encoding_found);
+    ins = (instruction)((unsigned)ins - INS_sve_invalid);
 
     switch (index)
     {
--- a/src/coreclr/jit/emitarm64.cpp
+++ b/src/coreclr/jit/emitarm64.cpp
@@ -10633,25 +10633,129 @@ void emitter::emitIns_Call(EmitCallType          callType,
 
 /*****************************************************************************
  *
- *  Returns an encoding for the specified register used in the 'Pd' position
+ *  Return an encoding for the specified 'V' register used in '4' thru '0' position.
  */
 
-/*static*/ emitter::code_t emitter::insEncodeReg_Pd(regNumber reg)
+/*static*/ emitter::code_t emitter::insEncodeReg_V_4_to_0(regNumber reg)
 {
-    assert(emitter::isPredicateRegister(reg));
-    emitter::code_t ureg = (emitter::code_t)reg - (emitter::code_t)REG_P0;
-    assert((ureg >= 0) && (ureg <= 15));
-    return ureg;
+    assert(isVectorRegister(reg));
+    emitter::code_t ureg = (emitter::code_t)reg - (emitter::code_t)REG_V0;
+    assert((ureg >= 0) && (ureg <= 32));
+    return ureg << 0;
 }
 
 /*****************************************************************************
  *
- *  Returns an encoding for the specified register used in the 'Pn' position
+ *  Return an encoding for the specified 'V' register used in '9' thru '5' position.
  */
 
-/*static*/ emitter::code_t emitter::insEncodeReg_Pn(regNumber reg)
+/*static*/ emitter::code_t emitter::insEncodeReg_V_9_to_5(regNumber reg)
 {
-    assert(emitter::isPredicateRegister(reg));
+    assert(isVectorRegister(reg));
+    emitter::code_t ureg = (emitter::code_t)reg - (emitter::code_t)REG_V0;
+    assert((ureg >= 0) && (ureg <= 32));
+    return ureg << 5;
+}
+
+/*****************************************************************************
+ *
+ *  Return an encoding for the specified 'P' register used in '12' thru '10' position.
+ */
+
+/*static*/ emitter::code_t emitter::insEncodeReg_P_12_to_10(regNumber reg)
+{
+    assert(isPredicateRegister(reg));
+    emitter::code_t ureg = (emitter::code_t)reg - (emitter::code_t)REG_P0;
+    assert((ureg >= 0) && (ureg <= 15));
+    return ureg << 10;
+}
+
+/*****************************************************************************
+ *
+ *  Return an encoding for the specified 'V' register used in '21' thru '17' position.
+ */
+
+/*static*/ emitter::code_t emitter::insEncodeReg_V_21_to_17(regNumber reg)
+{
+    assert(isVectorRegister(reg));
+    emitter::code_t ureg = (emitter::code_t)reg - (emitter::code_t)REG_V0;
+    assert((ureg >= 0) && (ureg <= 32));
+    return ureg << 17;
+}
+
+/*****************************************************************************
+ *
+ *  Return an encoding for the specified 'R' register used in '21' thru '17' position.
+ */
+
+/*static*/ emitter::code_t emitter::insEncodeReg_R_21_to_17(regNumber reg)
+{
+    assert(isIntegerRegister(reg));
+    emitter::code_t ureg = (emitter::code_t)reg;
+    assert((ureg >= 0) && (ureg <= 32));
+    return ureg << 17;
+}
+
+/*****************************************************************************
+ *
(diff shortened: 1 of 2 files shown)


kunalspathak:
@a74nh @BruceForstall @TIHan @amanasifkhalid

src/coreclr/jit/emitarm64.cpp
```diff
@@ -10633,41 +10633,236 @@ void emitter::emitIns_Call(EmitCallType          callType,
...
+}
+
+/*****************************************************************************
+ *
+ *  Return an encoding for the specified 'V' register used in '9' thru '5' position.
+ */
+
+/*static*/ emitter::code_t emitter::insEncodeReg_V_9_to_5(regNumber reg)
+{
+    assert(isVectorRegister(reg));
```
a74nh, line 10654: I want to change this to `insEncodeReg_V_9_to_5(regNumber reg, int lowbit, int highbit)`
But your way is safer.

At the risk of making more hard to parse macros:
```
#define insEncodeFunc(highbit, lowbit) emitter::code_t emitter::insEncodeReg_V_#highbit_to_#lowbit(regNumber reg) /
{ /
    emitter::code_t ureg = (emitter::code_t)reg - (emitter::code_t)REG_V0; /
    assert((ureg >= 0) && (ureg <= 32)); /
    return ureg << 5; /
} /

insEncodeFunc(9, 5);
insEncodeFunc(12, 10);
etc
```

src/coreclr/jit/emitarm64.h
```diff
@@ -316,14 +316,59 @@ static code_t insEncodeReg_Vm(regNumber reg);
...
+static code_t insEncodeReg_P_7_to_5(regNumber reg);
+
+// Return an encoding for the specified register used in '3' thru '1' position.
+static code_t insEncodeReg_P_3_to_1(regNumber reg);
+
+// Return an encoding for the specified register used in '2' thru '0' position.
+static code_t insEncodeReg_P_2_to_0(regNumber reg);
+
+// Return an encoding for the specified register used in '19' thru '17' position.
+static code_t insEncodeReg_V_19_to_17(regNumber reg);
```
a74nh, line 365: Should all the `V` ones be `Z`?

src/coreclr/jit/emitarm64.h
```diff
@@ -316,14 +316,59 @@ static code_t insEncodeReg_Vm(regNumber reg);
...
+
+// Return an encoding for the specified register used in '20' thru '17' position.
+static code_t insEncodeReg_V_20_to_17(regNumber reg);
+
+// Return an encoding for the specified register used in '9' thru '6' position.
+static code_t insEncodeReg_V_9_to_6(regNumber reg);
```
a74nh, line 371: 9_to_5?
https://docsmirror.github.io/A64/2023-06/and_z_p_zz.html

a74nh: sorry, you already have 9_to_5.
Confused as to exactly what this one is, as it is only 4 bits.

src/coreclr/jit/emitarm64.h
kunalspathak: I haven't introduced any "Z" terminology except for displaying the disassembly so kept it as `insEncodeReg_V*`. Even we use `REG_V0`, etc. Perhaps I should add aliases for "Z" registers and use it as well as rename methods whereever necessary.

src/coreclr/jit/emitarm64.h
kunalspathak: If I search for `insEncodeReg_V_9_to_6` in `emitOutputInstr_sve.cpp`, I see this:

```c++
case IF_SVE_HG_2A:   // ................ ......nnnn.ddddd -- SVE2 FP8 downconverts
   code = emitInsCodeSve(ins, fmt);
   code |= insEncodeReg_V_4_to_0(id->idReg10()); // ddddd
   code |= insEncodeReg_V_9_to_6(id->idReg20()); // nnnn
   dst += emitOutput_Instr(dst, code);
   break;
case IF_SVE_GA_2A:   // ............iiii ......nnnn.ddddd -- SME2 multi-vec shift narrow
   code = emitInsCodeSve(ins, fmt);
   code |= insEncodeReg_V_4_to_0(id->idReg10()); // ddddd
   code |= insEncodeReg_V_9_to_6(id->idReg20()); // nnnn
   code |= insEncodeImm(); // iiii
   dst += emitOutput_Instr(dst, code);
   break;
```

Then, searching for `IF_SVE_HG_2A` and  `IF_SVE_GA_2A in `instrsarm64_sve.h`, I see this:

```c++
//    enum               name                     info                                              SVE_HG_2A                                    
INST1(bfcvtn,            "bfcvtn",                0,                       IF_SVE_HG_2A,            0x650A3800                                   )
    // BFCVTN  <Zd>.B, {<Zn1>.H-<Zn2>.H }                                                SVE_HG_2A           0110010100001010 001110nnnn0ddddd     650A 3800   

//    enum               name                     info                                              SVE_GA_2A                                    
INST1(sqrshrn,           "sqrshrn",               RSH,                     IF_SVE_GA_2A,            0x45B02800                                   )
    // SQRSHRN <Zd>.H, {<Zn1>.S-<Zn2>.S }, #<const>                                      SVE_GA_2A           010001011011iiii 001010nnnn0ddddd     45B0 2800  

```

image

image

Most likely the SVE2 instructions takes only 4 bits for Z registers. Also it seems that https://docsmirror.github.io/A64/2023-06/sveindex.html doesn't have SVE2 instructions. I didn't find an entry of "bfcvtn" in there.

src/coreclr/jit/emitarm64.cpp
kunalspathak: @BruceForstall - any preference? if we don't want to use macro, we can also convert in a function that takes `lowbit` as a parameter.

a74nh: For some of the predicate ones, I've had to change them to assert(isLowPredicateRegister(reg)); because they only have 4 bits, as opposed to 5 bits.

So, maybe better leaving as functions.

src/coreclr/jit/emitarm64.h
a74nh: My preference would be Z, and introduce `REG_Z0` etc to make the code easier to read.
Same with `REG_P0`.

src/coreclr/jit/emitarm64.h
a74nh: Ok, looks like this is similar to the instructions that only allow predicate registers 0 to 7.

src/coreclr/jit/emitarm64.h
kunalspathak: Done for https://github.com/dotnet/runtime/pull/95129

src/coreclr/jit/emitarm64.cpp
BruceForstall: I'd prefer functions, even if they are quite duplicative. Makes for easier debugging. Sounds like @a74nh already found reasons the asserts would be different.

src/coreclr/jit/emitarm64.h
```diff
@@ -316,14 +316,59 @@ static code_t insEncodeReg_Vm(regNumber reg);
 // Returns an encoding for the specified register used in the 'Va' position
 static code_t insEncodeReg_Va(regNumber reg);
 
-// Returns an encoding for the specified register used in the 'Pd' position
-static code_t insEncodeReg_Pd(regNumber reg);
+// Return an encoding for the specified register used in '4' thru '0' position.
```
BruceForstall, line 319: If there's a comment here, it seems like it should match the definition comment, e.g.:
```
Return an encoding for the specified 'V' register used in '4' thru '0' position.
```

(same below)

BruceForstall (approved):

src/coreclr/jit/emitarm64.h
kunalspathak: Done.

Issue #94652, opened 2023-11-13
Test failure: _ParallelCrashTester::ParallelCrashTester.ParallelCrashMainThread()


VincentBu:
**Failed in:** [runtime-coreclr gcstress0x3-gcstress0xc 20231112.1](https://dev.azure.com/dnceng-public/public/_build/results?buildId=467188&view=ms.vss-test-web.build-test-results-tab&runId=10617694&resultId=119696&paneView=debug)

**Failed tests:**
```
coreclr linux x64 Checked gcstress0xc @ Ubuntu.2204.Amd64.Open
    - _ParallelCrashTester::ParallelCrashTester.ParallelCrashMainThread()
```

**Error message:**
```
 Exit code = 162, expected 134
```

**Stack trace:**
```
    at ParallelCrashTester.RunParallelCrash(Int32 arg)
   at Program.<<Main>$>g__TestExecutor11|0_12(StreamWriter tempLogSw, StreamWriter statsCsvSw, <>c__DisplayClass0_0&)
```

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/v-bibu/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/v-bibu/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#467886" vlink="#96607D">

Queued | Pipeline | Pipeline_Configuration | OS | Arch | Test | Outcome | BuildReason
-- | -- | -- | -- | -- | -- | -- | --
2023-10-24T06:26:49.662Z | runtime-coreclr crossgen2 outerloop | Checked-no_tiered_compilation | osx.1200.arm64.open | arm64 | baseservices/exceptions/simple/ParallelCrashMainThread/ParallelCrashMainThread.sh | Failed | Schedule
[...]

PR #95117, opened 2023-11-22
JIT: Switch arm32 call-finally to be similar to other targets

### LINKED ISSUE #59453: RyuJIT: reconsider arm32 callfinally code pattern
The JIT doesn't allow retless callfinally on ARM. Perhaps we should allow them, and generate "normal" calls instead of the `mov lr, <continuation>; jmp <finally>` form we generate now. Then we might be able to get rid of all the `BBF_FINALLY_TARGET` code, but we would need to set `FEATURE_EH_CALLFINALLY_THUNKS=1`. It might generate some worse code, which shouldn't normally matter because we now have finally cloning. We'd have to make sure the VM stack walker is ok with it. It would make ARM32 work just like ARM64, which might be advantageous. And if we could get rid of BBF_FINALLY_TARGET, that would be nice. If ARM32 chips implement a return address stack keyed off bl instructions (I don't know if they do), it might improve performance.

category:design
theme:eh

### LINKED ISSUE #66578: Assertion failed 'interval->isSpilled' during 'LSRA allocate'
### Description

The following program hits a JIT assert on arm32 when optimizing.

### Reproduction Steps

```csharp
// Generated by Fuzzlyn v1.5 on 2022-03-13 15:19:20
// Run on Arm Linux
// Seed: 13161118502102171269
// Reduced from 313.8 KiB to 1.2 KiB in 00:16:29
// Hits JIT assert in Release:
// Assertion failed 'interval->isSpilled' in 'Program:M52(C0)' during 'LSRA allocate' (IL size 103; hash 0xb4fc4553; FullOpts)
// 
//     File: /__w/1/s/src/coreclr/jit/lsra.cpp Line: 5924
// 
public interface I0
{
}

public interface I1
{
}

public class C0 : I0, I1
{
    public sbyte F0;
    public bool F1;
    public uint F2;
    public ushort F5;
    public C0(sbyte f0, bool f1, uint f2, ushort f5)
    {
        F0 = f0;
        F1 = f1;
        F2 = f2;
        F5 = f5;
    }
}

public class Program
{
    public static I1[][] s_18;
    public static ushort[][] s_43;
    public static I1 s_64;
    public static I0 s_88;
    public static void Main()
    {
        var vr6 = new C0(0, false, 0, 0);
        M52(vr6);
    }

    public static void M52(C0 argThis)
    {
        I1 vr9 = s_18[0][0];
        if (argThis.F1)
        {
            return;
        }

        if (argThis.F1)
        {
            try
            {
                s_43 = new ushort[][]{new ushort[]{0}};
            }
            finally
            {
                for (int var8 = 0; var8 < 2; var8++)
                {
                    s_64 = argThis;
                }

                C0 vr10 = new C0(0, true, 0, 0);
                s_88 = new C0(-1, true, 0, 0);
            }

            I1 vr12 = s_18[0][0];
        }
    }
}
```

### Expected behavior

No assert hit.

### Actual behavior

Assert hit.

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

_No response_

### Other information

_No response_

category:correctness
theme:register-allocator

jakobbotsch:
This switches arm32 to generate call-finally in the same way as other targets, with a call to the finally funclet, instead of loading a different return address. Loading a different return address confuses the hardware's return address predictor, which has large negative perf impact.

Two micro benchmarks from my rpi, both run with `DOTNET_JitEnableFinallyCloning=0`, that show the cost of messing up return address prediction. The first runs a loop that calls a finally funclet on every iteration. The second sets up a deeper stack and calls a funclet, which means that the old scheme then mispredicts the return for all subsequent returns. The second benchmark is over 4 times faster with this change.
```csharp
    public static void Main()
    {
        for (int i = 0; i < 4; i++)
        {
            for (int j = 0; j < 100; j++)
            {
                Run(100);
                Call1();
            }

            Thread.Sleep(100);
        }

        Stopwatch timer = Stopwatch.StartNew();
        Run(100_000_000);
        timer.Stop();
        Console.WriteLine("Elapsed: {0}ms", timer.ElapsedMilliseconds);
        timer.Restart();
        for (int i = 0; i < 100_000_000; i++)
            Call1();
        timer.Stop();
        Console.WriteLine("Elapsed: {0}ms", timer.ElapsedMilliseconds);
    }

    public static long Run(int iters)
    {
        long sum = 0;
        for (int i = 0; i < iters; i++)
        {
            try
            {
                sum += i;
            }
            finally
            {
                sum += 2 * i;
            }
        }

        return sum;
    }

    public static int Call1() => Call2() + 1;
    public static int Call2() => Call3() + 2;
    public static int Call3() => Call4() + 3;
    public static int Call4() => Call5() + 4;
    public static int Call5() => Call6() + 5;
    public static int Call6() => Call7() + 6;
    public static int Call7() => Call8() + 7;
    public static int Call8() => Call9() + 8;
    public static int Call9() => Call10() + 9;
    public static int Call10() => Call11() + 10;
    public static int Call11() => Call12() + 11;
    public static int Call12() => Call13() + 12;
    public static int Call13() => Finally();

    public static int Finally()
    {
        int result = 10;
        try
        {
            result += 5;
        }
        finally
        {
            result += 10;
        }

        return result;
    }
```

Codegen diff: https://www.diffchecker.com/zKoHLm4v/

Output before:
```scala
Elapsed: 916ms
Elapsed: 18322ms
```

Output after:
```scala
Elapsed: 856ms
Elapsed: 4010ms
```

Fix #59453
Fix #66578


jakobbotsch:

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=478074&view=ms.vss-build-web.run-extensions-tab)

No arm32 failures in any of the test pipelines (except for #94652), so seems like not much else is needed here.

kunalspathak:
Nice cleanup. Changes LGTM

jakobbotsch:
I found #59453 that covers this and noticed the `FEATURE_EH_CALLFINALLY_THUNKS` it talks about, so I pushed a change to enable that. Will rerun testing once innerloop passes.

BruceForstall:
> I found #59453 that covers this and noticed the `FEATURE_EH_CALLFINALLY_THUNKS` it talks about, so I pushed a change to enable that. Will rerun testing once innerloop passes.

Can you update the diffchecker diffs linked at the top after this change? That was one thing I didn't see in the diff: I expected the "bl" to the finally to be in its own IG so the EH reporting could reference it.

In the diff that is there, there is a `nop` injected before the continuation (finally return target). I would think that would no longer be necessary.

BruceForstall:
[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=478961&view=ms.vss-build-web.run-extensions-tab)

Oddly large TP regression in some cases. Probably due to FEATURE_EH_CALLFINALLY_THUNKS handling?

Huge size improvements.

BruceForstall (approved):
The changes look great to me. Love to see all this ifdef code disappear, especially if it's a net perf and code size win.

BruceForstall:
cc @janvorli -- this PR changes how the JIT generates calls to finally clauses. fyi, in case you know of a reason the VM has a dependency on the current model (seems unlikely to me).

jakobbotsch:
> Oddly large TP regression in some cases. Probably due to FEATURE_EH_CALLFINALLY_THUNKS handling?

libraries.pmi (the collection with the large TP regression) has only 6 MinOpts contexts, so the results aren't very telling. The particular context that regresses is for this one:
https://github.com/dotnet/fsharp/blob/b4dc66c3d3cd4b45614d2b1b903890c9c846895a/src/FSharp.Core/Linq.fs#L429-L641

We switch to minopts because the function has over 60 KB of IL. The function has only 7 EH clauses but it has 2500 basic blocks. The detailed TP diff for this single context is:

```scala
Base: 185069812, Diff: 185844906, +0.4188%

552076 : NA        : 51.77% : +0.2983% : ?fgFindInsertPoint@Compiler@@QAEPAUBasicBlock@@I_NPAU2@1110@Z                            
131965 : +4.94%    : 12.37% : +0.0713% : ?Build@GcInfoEncoder@@QAEXXZ                                                             
98691  : NA        : 9.25%  : +0.0533% : ?isBBCallAlwaysPair@BasicBlock@@QBE_NXZ                                                  
44208  : NA        : 4.15%  : +0.0239% : ?emitAddLabel@emitter@@AAEPAXABQAI_K1@Z                                                  
29708  : +3636.23% : 2.79%  : +0.0161% : ?genReportEH@CodeGen@@AAEXXZ                                                             
13135  : +1.65%    : 1.23%  : +0.0071% : ?setBlockSequence@LinearScan@@AAEXXZ                                                     
12784  : +43.06%   : 1.20%  : +0.0069% : ?isBBCallAlwaysPairTail@BasicBlock@@QBE_NXZ                                              
11847  : +0.83%    : 1.11%  : +0.0064% : ?genCodeForBBlist@CodeGen@@IAEXXZ                                                        
5164   : +0.13%    : 0.48%  : +0.0028% : ?emitOutputInstr@emitter@@IAEIPAUinsGroup@@PAUinstrDesc@1@PAPAE@Z                        
2645   : +8.26%    : 0.25%  : +0.0014% : ?genMarkLabelsForCodegen@CodeGen@@AAEXXZ                                                 
2472   : +0.02%    : 0.23%  : +0.0013% : ?allocateRegisters@LinearScan@@QAEXXZ                                                    
2423   : +1.42%    : 0.23%  : +0.0013% : ?emitIns_J@emitter@@QAEXW4instruction@@PAUBasicBlock@@H@Z                                
1140   : +0.24%    : 0.11%  : +0.0006% : ?ClearD@?$BitSetOps@PAI$00PAVCompiler@@VBasicBlockBitSetTraits@@@@SGXPAVCompiler@@AAPAI@Z
-1923  : -0.40%    : 0.18%  : -0.0010% : ?emitEndCodeGen@emitter@@QAEIPAVCompiler@@_N11IPAI2PAPAX33@Z                             
-2316  : -0.72%    : 0.22%  : -0.0013% : ?fgRemoveDeadBlocks@Compiler@@IAE_NXZ                                                    
-5768  : -0.70%    : 0.54%  : -0.0031% : ?emitJumpDistBind@emitter@@AAEXXZ                                                        
-6405  : -1.46%    : 0.60%  : -0.0035% : ?WriteSlotStateVector@GcInfoEncoder@@AAEXAAVBitStreamWriter@@ABVBitArray@@@Z             
-8759  : -0.71%    : 0.82%  : -0.0047% : ?Write@BitStreamWriter@@QAEXII@Z                                                         
-10567 : -1.43%    : 0.99%  : -0.0057% : ?SizeofSlotStateVarLengthVector@GcInfoEncoder@@AAEXABVBitArray@@IIPAI11@Z                
-29120 : -100.00%  : 2.73%  : -0.0157% : ?emitStackPop@emitter@@QAEXPAE_NEI@Z                                                     
-31879 : -26.86%   : 2.99%  : -0.0172% : Compiler::fgRemoveUnreachableBlocks<`Compiler::fgRemoveDeadBlocks'::`2'::<lambda_1> >    
-44208 : -100.00%  : 4.15%  : -0.0239% : ?emitAddLabel@emitter@@AAEPAXABQAI_K1_N@Z                                                
```

jakobbotsch:
> Can you update the diffchecker diffs linked at the top after this change? That was one thing I didn't see in the diff: I expected the "bl" to the finally to be in its own IG so the EH reporting could reference it.
> 
> In the diff that is there, there is a `nop` injected before the continuation (finally return target). I would think that would no longer be necessary.

I updated the diff -- indeed, the `bl` is in its own IG now, but the `nop` is still there.

janvorli:
> cc @janvorli -- this PR changes how the JIT generates calls to finally clauses. fyi, in case you know of a reason the VM has a dependency on the current model (seems unlikely to me).

@BruceForstall if I understand it correctly, this only influences non-exceptionally invoked finallys, right? So the difference is that instead of `lr, <continuation>; jmp <finally>` JIT uses `call <finally>; jmp <continuation>`?
 I am not aware of the runtime depending on that in any way.

PR #95129, opened 2023-11-22
SVE: Add Z register aliases

kunalspathak:


kunalspathak:
@a74nh @dotnet/jit-contrib

a74nh:
LGTM.

BruceForstall:
Is this going to be more useful than confusing? I guess the Arm manual itself distinguishes V and Z registers, which seems odd, since V registers are also named Q, D, S, H, and B depending on size, and Z registers contain V registers.

It seems like this won't be helpful because you'll be able to use V instead of Z and vice-versa in the code, and there will be no checking.

kunalspathak:
> It seems like this won't be helpful because you'll be able to use V instead of Z and vice-versa in the code, and there will be no checking.

Sure, V and Z are same registers, but I think while reading code, it will be helpful to know if the instruction operates on scalable "Z" register or a fixed-width "V" register specially around instructions that operates on both Z and V registers. 

```diff
-theEmitter->emitIns_R_R_R(INS_sve_andqv, EA_8BYTE, REG_V0, REG_V0, REG_P0, INS_OPTS_SCALABLE_B);  // IF_SVE_AG_3A /* ANDQV   <Vd>.<T>, <Pg>, <Zn>.<Tb> */
+theEmitter->emitIns_R_R_R(INS_sve_andqv, EA_8BYTE, REG_V0, REG_Z0, REG_P0, INS_OPTS_SCALABLE_B);  // IF_SVE_AG_3A /* ANDQV   <Vd>.<T>, <Pg>, <Zn>.<Tb> */
```

To that matter, I don't think we use `REG_V*` explicitely anyway apart from unit test and few other places like unwind, where we can continue to use `REG_V*`.

BruceForstall:
> I think while reading code, it will be helpful to know if the instruction operates on scalable "Z" register or a fixed-width "V" register specially around instructions that operates on both Z and V registers.

The problem is, if there is no correctness checking, then it's not going to be too useful, IMO.

kunalspathak:
> To that matter, I don't think we use `REG_V*` explicitely anyway apart from unit test and few other places like unwind, where we can continue to use `REG_V*`.

Taking account of that and @BruceForstall POV, I will close this PR then.

PR #95133, opened 2023-11-22
Fix FormatException error message when parsing invalid UTF8 inputs

### LINKED ISSUE #95044: The input string 'System.ReadOnlySpan<Byte>[2]' was not in a correct format.
### Description

`System.Number.ThrowFormatException(ReadOnlySpan<>)` calls object.ToString on the parameter, which results in the thrown exception containing the type's display name instead of the value of the span. https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Number.Parsing.cs,1391
I think it should convert the value of the span to UTF-16 and use that in message formatting.

### Reproduction Steps

dotnet new console -f net8.0
echo _ = int.Parse(stackalloc byte[2]); > Program.cs
dotnet run

### Expected behavior

```
Unhandled exception. System.FormatException: The input string '
```

(to match `int.Parse("\0\0")` behavior, which I'm not a fan of)

### Actual behavior

```
Unhandled exception. System.FormatException: The input string 'System.ReadOnlySpan<Byte>[2]' was not in a correct format.
   at System.Number.ThrowFormatException[TChar](ReadOnlySpan`1 value)
   at System.Int32.Parse(ReadOnlySpan`1 utf8Text, NumberStyles style, IFormatProvider provider)
   at Program.<Main>$(String[] args) in path\Program.cs:line 1
```

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

net8.0

### Other information

_No response_

stephentoub:
Fixes https://github.com/dotnet/runtime/issues/95044


src/libraries/System.Runtime/tests/System.Runtime.Tests/System/HalfTests.cs
```diff
@@ -921,13 +921,26 @@ public static void Parse_Utf8Span_Invalid(string value, NumberStyles style, IFor
             if (value != null)
             {
                 ReadOnlySpan<byte> valueUtf8 = Encoding.UTF8.GetBytes(value);
-                Assert.Throws(exceptionType, () => float.Parse(Encoding.UTF8.GetBytes(value), style, provider));
+                Exception e = Assert.Throws(exceptionType, () => Half.Parse(Encoding.UTF8.GetBytes(value), style, provider));
```
stephentoub, line 924: This also fixed the existing call to test Half rather than float.

src/libraries/System.Runtime/tests/System.Runtime.Tests/System/IntPtrTests.cs
```diff
@@ -1021,13 +1021,26 @@ public static void Parse_Utf8Span_Invalid(string value, NumberStyles style, IFor
                     Assert.Equal(default, result);
                 }
 
-                Assert.Throws(exceptionType, () => int.Parse(Encoding.UTF8.GetBytes(value), style, provider));
+                Exception e = Assert.Throws(exceptionType, () => nint.Parse(Encoding.UTF8.GetBytes(value), style, provider));
```
stephentoub, line 1024: This also fixed the existing call to test nint rather than int.

src/libraries/System.Private.CoreLib/src/System/Number.Parsing.cs
```diff
@@ -1388,7 +1389,27 @@ internal enum ParsingStatus
...
+                // Decode the UTF8 value into a string we can include in the error message. We're here
+                // because we failed to parse, which also means the bytes might not be valid UTF8,
+                // so fallback to a message that doesn't include the value if the bytes are invalid.
+                // It's possible after we check the bytes for validity that they could be concurrently
+                // mutated, but if that's happening, all bets are off, anyway, and it simply impacts
+                // which exception is thrown.
+                ReadOnlySpan<byte> bytes = MemoryMarshal.Cast<TChar, byte>(value);
+                errorMessage = Utf8.IsValid(bytes) ?
+                    SR.Format(SR.Format_InvalidStringWithValue, Encoding.UTF8.GetString(bytes)) :
+                    SR.Format_InvalidString;
```
tannergooding, line 1405: I know this is an exception path and therefore likely cold, but is it worth just calling `Utf8.ToUtf16` instead and checking the `OperationStatus`, since its cheaper than using `Utf8.IsValid` and then trying to extract the string?

tannergooding (approved):

src/libraries/System.Private.CoreLib/src/System/Number.Parsing.cs
stephentoub: > but is it worth just calling Utf8.ToUtf16 instead and checking the OperationStatus

I'd need a sufficiently large destination buffer; didn't seem worth doing the whole stackalloc/ArrayPool dance to get one. Is there something else you had in mind?

stephentoub: > We're getting that allocation anyways with Encoding.UTF8.GetString

We shouldn't be. If the input is <= 32 bytes, it'll stackalloc, and if it's longer, it'll use GetCharCount to get the actual length, allocate the string of that exact length, and then decode directly into the string.

> The allocation is probably cheaper than the IsValid check and doesn't require the dance.

| Method     | Length | Mean     | Allocated |
|----------- |------- |---------:|----------:|
| GetString1 | 5      | 23.64 ns |      32 B |
| GetString2 | 5      | 22.61 ns |      72 B |
| GetString1 | 100    | 47.10 ns |     224 B |
| GetString2 | 100    | 59.77 ns |     456 B |

```C#
using System.Text;
using System.Text.Unicode;

[MemoryDiagnoser(false)]
public partial class Tests
{
    private byte[] _data;

    [Params(5, 100)]
    public int Length { get; set; }

    [GlobalSetup]
    public void Setup() => _data = Encoding.UTF8.GetBytes(new string('a', Length));

    [Benchmark]
    public string GetString1()
    {
        if (Utf8.IsValid(_data))
        {
            return Encoding.UTF8.GetString(_data);
        }
        return "";
    }

    [Benchmark]
    public string GetString2()
    {
        char[] dest = new char[Encoding.UTF8.GetMaxCharCount(_data.Length)];
        if (Utf8.ToUtf16(_data, dest, out _, out int charsWritten) == System.Buffers.OperationStatus.Done)
        {
            return dest.AsSpan(0, charsWritten).ToString();
        }
        return "";
    }
}
```

adamsitnik (dismissed):
The test failures seem to be related, please fix them.

PR #94239, opened 2023-10-31
JIT: Remove BBJ_NONE

amanasifkhalid:
Next step for #93020, per conversation on #93772. Replacing `BBJ_NONE` with `BBJ_ALWAYS` to the next block helps limit our use of implicit fall-through (though we still expect `BBJ_COND` to fall through when its false branch is taken; #93772 should eventually address this).

I've added a small peephole optimization to skip emitting unconditional branches to the next block during codegen.

PR #95139, opened 2023-11-22
[JIT] Add BasicBlock::bbFlags helper methods

amanasifkhalid:
Andy mentioned in #94239 that it would be nice to have a method handle verbose flag checks like `((block->bbFlags & BBF_SOME_FLAG) != 0)`, so this adds some helper methods for checking/modifying block flags.


amanasifkhalid:
cc @dotnet/jit-contrib. With these changes, we only have a few spots where we directly access `bbFlags` (usually either for assignment, or for calculating its bitwise AND with some flag without converting the result to a boolean). If we've made it this far, would it be worth just making `bbFlags` private, and adding `GetFlags()` and `SetFlags(...)` to cover those last few cases?

src/coreclr/jit/block.h
```diff
@@ -869,8 +894,7 @@ struct BasicBlock : private LIR::Range
     {
         if (this->bbWeight == BB_ZERO_WEIGHT)
         {
-            this->bbFlags &= ~BBF_RUN_RARELY;  // Clear any RarelyRun flag
-            this->bbFlags &= ~BBF_PROF_WEIGHT; // Clear any profile-derived flag
+            this->RemoveFlag((BBF_RUN_RARELY | BBF_PROF_WEIGHT));
```
BruceForstall, line 897: nit
```suggestion
            this->RemoveFlag(BBF_RUN_RARELY | BBF_PROF_WEIGHT);
```

src/coreclr/jit/codegenarmarch.cpp
```diff
@@ -5435,7 +5435,7 @@ void CodeGen::genFnEpilog(BasicBlock* block)
     }
 #endif // DEBUG
 
-    bool jmpEpilog = ((block->bbFlags & BBF_HAS_JMP) != 0);
+    bool jmpEpilog = (block->HasFlag(BBF_HAS_JMP));
```
BruceForstall, line 5438: nit
```suggestion
    bool jmpEpilog = block->HasFlag(BBF_HAS_JMP);
```

src/coreclr/jit/codegencommon.cpp
```diff
@@ -864,11 +864,11 @@ BasicBlock* CodeGen::genCreateTempLabel()
 #endif
 
     JITDUMP("Mark " FMT_BB " as label: codegen temp block\n", block->bbNum);
-    block->bbFlags |= BBF_HAS_LABEL;
+    block->SetFlag(BBF_HAS_LABEL);
 
     // Use coldness of current block, as this label will
     // be contained in it.
-    block->bbFlags |= (compiler->compCurBB->bbFlags & BBF_COLD);
+    block->SetFlag(compiler->compCurBB->bbFlags & BBF_COLD);
```
BruceForstall, line 871: You could add a `CopyFlags` function and have:
```suggestion
    block->CopyFlags(compiler->compCurBB, BBF_COLD);
```

src/coreclr/jit/block.h
```diff
@@ -714,6 +714,31 @@ struct BasicBlock : private LIR::Range
...
+    {
+        return CheckFlag(flag, flag);
+    }
+
+    bool HasFlag(const BasicBlockFlags flag) const
+    {
+        return !CheckFlag(flag, BBF_EMPTY);
+    }
+
+    void SetFlag(const BasicBlockFlags flag)
```
BruceForstall, line 732: Maybe it should be SetFlags/RemoveFlags?

src/coreclr/jit/block.h
```diff
@@ -714,6 +714,31 @@ struct BasicBlock : private LIR::Range
...
+    {
+        return ((bbFlags & flag) == target);
+    }
+
+    bool CheckFlag(const BasicBlockFlags flag) const
+    {
+        return CheckFlag(flag, flag);
+    }
+
+    bool HasFlag(const BasicBlockFlags flag) const
```
BruceForstall, line 727: Should this assert that `flag` has only one bit set? And add an explicit `HasAnyFlag` that allows for the rare cases where you're checking if any flag is set? (actually, we should probably just make callers use `HasFlag(one) || HasFlag(two)` in that case)

src/coreclr/jit/fgopt.cpp
```diff
@@ -2343,20 +2343,20 @@ void Compiler::fgCompactBlocks(BasicBlock* block, BasicBlock* bNext)
...
-        block->bbFlags &= ~BBF_INTERNAL; // Clear the BBF_INTERNAL flag
-        block->bbFlags |= BBF_IMPORTED;  // Set the BBF_IMPORTED flag
+        block->RemoveFlag(BBF_INTERNAL); // Clear the BBF_INTERNAL flag
+        block->SetFlag(BBF_IMPORTED);    // Set the BBF_IMPORTED flag
     }
 
     /* Update the flags for block with those found in bNext */
 
-    block->bbFlags |= (bNext->bbFlags & BBF_COMPACT_UPD);
+    block->SetFlag(bNext->bbFlags & BBF_COMPACT_UPD);
```
BruceForstall, line 2355: Another case for `CopyFlag`

src/coreclr/jit/flowgraph.cpp
```diff
@@ -3026,7 +3026,7 @@ void Compiler::fgInsertFuncletPrologBlock(BasicBlock* block)
     assert(nullptr == fgGetPredForBlock(block, newHead));
     fgAddRefPred(block, newHead);
 
-    assert((newHead->bbFlags & BBF_INTERNAL) == BBF_INTERNAL);
+    assert(newHead->CheckFlag(BBF_INTERNAL));
```
BruceForstall, line 3029: This can probably be `HasFlag`. It's equivalent if the flag being checked is a single bit. That's probably the case for most/all of the uses of `CheckFlag` with a single flag.

BruceForstall (commented):
I like this change. Left a few notes.

> If we've made it this far, would it be worth just making bbFlags private, and adding GetFlags() and SetFlags(...) to cover those last few cases?

With my suggested `CopyFlags` function, how many more cases are there? I think it would make sense to make it private. I would name the functions `GetFlagsRaw`/`SetFlagsRaw` to make it even more clear this is an "escape hatch" and not preferred usage.

src/coreclr/jit/codegencommon.cpp
amanasifkhalid: Sure thing. We do have a few cases where the second argument to `CopyFlags` would be multiple flags OR'd together, e.g. in `Compiler::fgSplitBlockBeforeTree`:
`block->SetFlag(originalFlags & (BBF_SPLIT_GAINED | BBF_IMPORTED | BBF_GC_SAFE_POINT | BBF_LOOP_PREHEADER | BBF_RETLESS_CALL));`
Would you be ok with this pattern?

BruceForstall: Something like:
```
block->CopyFlags(fromBlock, BBF_IMPORTED | BBF_GC_SAFE_POINT | BBF_LOOP_PREHEADER | BBF_RETLESS_CALL);
```
would be fine.

The more complex cases where we (1) grab the flags, (2) do some computations, then (3) set the flags, might be where the `GetFlagsRaw` would come in. So the example you reference might need to use that.

src/coreclr/jit/block.h
amanasifkhalid: I think that's a good idea. It doesn't come up often, but we do have a few `HasFlag(BBF_FLAG1 | BBF_FLAG2)` calls, so maybe we can use the templated approach we use for `BasicBlock::KindIs`? If not, we can just use more than one call to `HasFlag` in the caller.

BruceForstall: I think it's better to make the caller use `HasFlag(one) || HasFlag(two)`, or `HasFlag(one) && HasFlag(two)`.

Yes, we could use the template magic to have `HasAnyFlags(one,two,...)` and `HasAllFlags(one,two,...)` but does that happen very often? IMO, we should not have `HasFlag(one,two,...)` because it's not clear if it's "any" or "all".

amanasifkhalid: > Yes, we could use the template magic to have HasAnyFlags(one,two,...) and HasAllFlags(one,two,...) but does that happen very often?

I don't recall seeing it enough to justify introducing two more methods.

> I think it's better to make the caller use HasFlag(one) || HasFlag(two), or HasFlag(one) && HasFlag(two).

Now that you mention the `||` vs `&&` cases, I agree it's more readable to just have the caller make two calls.

amanasifkhalid: On second thought, there are some cases where multiple calls to `HasFlag` quickly becomes ugly, so I'll introduce `HasAnyFlags` to cover these scenarios (I haven't found a need for `HasAllFlags` anywhere).

amanasifkhalid:
> With my suggested CopyFlags function, how many more cases are there?

The cases that require `GetFlagsRaw` and/or `SetFlagsRaw` fall into the following categories:
* We want to cheaply assign `bbFlags` to some computed value, or to the flags of another block (we could introduce another version of `CopyFlags` to do the latter, but I think that might be confusing, since one version of `CopyFlags` would bitwise OR the flags with its arguments, and the other would just assign them). This occurs only a few (~3) times, for example when creating a duplicate block.
* We want to copy a block's flags to a local variable, do some computations on the local variable, and then pass the computed flags to `SetFlags/CopyFlags/etc`, or use them as a conditional. A one-line version of the former looks like `newBlock->CopyFlags(curr, (succ->GetFlagsRaw() & BBF_BACKWARD_JUMP));`
* We want to print a block's flags.

Overall, usage of the raw getters/setters is uncommon.

src/coreclr/jit/flowgraph.cpp
amanasifkhalid: Thanks for pointing this out. I was able to get rid of `CheckFlag` altogether.

src/coreclr/jit/block.h
BruceForstall: nit: in English, the preferred wording would be HasAnyFlag and HasAllFlags, IMO

jakobbotsch:
I don't have a strong opinion on a change like this, but I will say that I don't think the original pattern is verbose, and bitwise manipulation is such a primitive operation that I don't think this makes things clearer, only a bit inconsistent with other flags throughout the JIT.

If other people on the team prefer these wrapper methods, then I am ok with it.

amanasifkhalid:
> I don't think this makes things clearer, only a bit inconsistent with other flags throughout the JIT.

I agree this isn't a significant readability improvement, but this does fix the style inconsistencies with checking the flags (like use of implicit boolean, number of parentheses, etc).

I noticed from the last run that this change has unexpectedly large [TP diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=478651&view=ms.vss-build-web.run-extensions-tab). I suspect the templated definition of `HasAnyFlag` is to blame, and it would be cheaper to OR all the flags passed in first, and then check if `bbFlags` contains them with one AND instruction. 

Side note: I tried profiling the TP diff locally, but the recent changes to `jiteeversionguid.h` are causing the JIT to fail to initialize when SuperPMI tries to run a collection due to a version mismatch. Rebuilding the JIT/SuperPMI and redownloading the collections doesn't fix this. Any ideas?

jakobbotsch:
> Side note: I tried profiling the TP diff locally, but the recent changes to `jiteeversionguid.h` are causing the JIT to fail to initialize when SuperPMI tries to run a collection due to a version mismatch. Rebuilding the JIT/SuperPMI and redownloading the collections doesn't fix this. Any ideas?

Maybe your baseline is outdated? You may need a `git fetch upstream` for `superpmi tpdiff` to pick up a new enough baseline.

AndyAyersMS:
Sorry to bike shed but what was the rationale for `CheckFlag` instead of `HasFlag`?

src/coreclr/jit/block.h
```diff
@@ -697,8 +697,49 @@ struct BasicBlock : private LIR::Range
...
+    {
+        // Assert flag is not multiple BasicBlockFlags OR'd together
+        // by checking if it is a power of 2
+        // (CheckFlag expects to check only one flag at a time)
+        assert((flag & (flag - 1)) == 0);
+        return (bbFlags & flag);
+    }
+
+    BasicBlockFlags CheckFlags(const BasicBlockFlags flags) const
+    {
```
AndyAyersMS, line 714: Should you assert `flags` is not a power of two here?

amanasifkhalid:
> Maybe your baseline is outdated? You may need a git fetch upstream for superpmi tpdiff to pick up a new enough baseline.

Thank you for the suggestion, and sorry I never responded to this. Rebuilding everything did the trick.

> Sorry to bike shed but what was the rationale for CheckFlag instead of HasFlag?

No worries; I'm glad you asked. I traced the TP regressions back to `HasFlag` locally. Previously, `HasFlag` returned a boolean by evaluating `((bbFlags & flag) != 0)`. Perhaps due to some suboptimal codegen, I found it's faster for `HasFlag` to simply return `(bbFlags & flag)`, which is of type `BasicBlockFlags`. I don't see how these two approaches are different, since with the new approach, the returned `BasicBlockFlags` is still implicitly cast to a boolean in the caller, but this fixed the TP regressions locally. The name `HasFlag` implies to me that it returns a boolean, so I think this approach justifies changing the name too, but I'd appreciate suggestions here (or any insight into why this approach is faster).

src/coreclr/jit/block.h
amanasifkhalid: I think that makes sense.

src/coreclr/jit/compiler.h
```diff
@@ -3298,7 +3298,7 @@ class Compiler
     };
 
     GenTree* gtTryRemoveBoxUpstreamEffects(GenTree* tree, BoxRemovalOptions options = BR_REMOVE_AND_NARROW);
-    GenTree* gtOptimizeEnumHasFlag(GenTree* thisOp, GenTree* flagOp);
+    GenTree* gtOptimizeEnumCheckFlag(GenTree* thisOp, GenTree* flagOp);
```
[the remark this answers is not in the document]
amanasifkhalid: Thanks for catching that -- my `Ctrl+F` was unintentionally indiscriminate.

src/coreclr/jit/block.h
```diff
@@ -697,8 +697,49 @@ struct BasicBlock : private LIR::Range
...
+private:
     BasicBlockFlags bbFlags;
 
+public:
+    BasicBlockFlags CheckFlag(const BasicBlockFlags flag) const
+    {
+        // Assert flag is not multiple BasicBlockFlags OR'd together
+        // by checking if it is a power of 2
+        // (CheckFlag expects to check only one flag at a time)
+        assert((flag & (flag - 1)) == 0);
```
BruceForstall, line 709:
```suggestion
        assert(isPow2(flag));
```

BruceForstall (changes requested):
Why did you change it from `HasFlag` to `CheckFlag`? Because it no longer returns a bool? I think it reads poorly in the code as `CheckFlag`, e.g.,

```
if (block->CheckFlag(BBF_FUNCLET_BEG)) ...
```

"CheckFlag" obviously does something, but isn't obviously a predicate. It reads like it would be a void function. Whereas:

```
if (block->HasFlag(BBF_FUNCLET_BEG)) ...
```

is obviously a predicate.

I note that there are many places where the result of the current CheckFlag (a BasicBlockFlags) is assigned to a bool. Does the compiler actually insert an implicit `!= 0` and generate a 0/1 result, or does it generate a `!= 0 / == 0` result by just copying the value?

Did you remove:
```
bool HasFlag(const BasicBlockFlags flag) const
{
    return (bbFlags & flag) != 0;
}
```
because of TP concerns? Would it be possible to implement it as:
```
bool HasFlag(const BasicBlockFlags flag) const
{
    return (bool)(bbFlags & flag);
}
```
Would that be legal? Wouldn't that do the same as the implicit conversions you've already got?

amanasifkhalid:
> Why did you change it from HasFlag to CheckFlag? Because it no longer returns a bool? I think it reads poorly in the code as CheckFlag

Yes, my reasoning for changing its name was around its return value no longer being a bool. I did a quick pass, and the overwhelming majority of `CheckFlag` calls are used as predicates (such that changing the return value back to bool would not affect correctness), so I understand why we should name it `HasFlag`. I can revert the name change.

> Did you remove because of TP concerns? Would it be possible to implement it as...

Yes, I still see TP regressions locally with the `(bool)(bbFlags & flag)` approach. I'm guessing this is a codegen issue with MSVC, since we didn't see any [diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=478651&view=ms.vss-build-web.run-extensions-tab) on Linux with the initial boolean implementation, though [godbolt](https://godbolt.org/z/WYGrjoG5Y) isn't all that enlightening; with optimizations enabled, the two implementations look roughly equivalent (especially after `CheckFlag` gets inlined).

The updated approach of not casting to bool in `CheckFlag` seems to have fixed the [TP diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=483153&view=ms.vss-build-web.run-extensions-tab) on Windows x64, though x86 still has big regressions (though there are asmdiffs on Windows x86, too; I'll look into those).

BruceForstall:
It's hard to understand why there is any (but minor) TP impact to the bool-returning `((block->bbFlags & BBF_SOME_FLAG) != 0)` function, especially since that pattern was widely used anyway. And if it's the manifesting of a 0/1 bool that is expensive, then you would think the non-bool function would be a TP improvement, since all the cases currently using the `((block->bbFlags & BBF_SOME_FLAG) != 0)` form would "improve".

This all assumes `HasFlag` is inlined, which I would assume. But maybe that's a bad assumption?

amanasifkhalid:
> This all assumes HasFlag is inlined, which I would assume. But maybe that's a bad assumption?

I don't see why `HasFlag` wouldn't be inlined, considering its size and simplicity (though maybe I'm missing something). Adding the `inline` directive doesn't change any TP diffs locally (though that doesn't guarantee the compiler will inline it). I'll take a look at the disassembly...

amanasifkhalid:
> I'll take a look at the disassembly...

So the codegen for the `BasicBlockFlags` and `bool` implementations match godbolt; here they are, respectively:
```
mov         rax,qword ptr [rcx+30h]
and         rax,rdx
ret
```

```
test        qword ptr [rcx+30h],rdx
setne       al
ret
```

In 6 places in the JIT, neither version is inlined (though there are no instances where one is inlined, and the other isn't). These calls are in larger methods like `Compiler::impImportBlockCode` (which is almost 5000 lines long), so maybe MSVC is hitting its inlining budget? The `inline` directive doesn't change anything here. Similarly, `SetFlags` is called in 12 places. All the other helper methods are inlined everywhere.

We could use `__forceinline` to get around this. We have a few use cases of it, but I imagine this isn't ideal (though replacing these methods with macros probably wouldn't be any better). But regardless of how we address inlining, I'm still surprised I managed to make up most of the TP diff on Windows x64 by returning a `BasicBlockFlags` instead. I guess a `mov` and an `and` are that much cheaper?

BruceForstall:
> We could use __forceinline to get around this.

I think this is a good case for the use of `__forceinline`

> I guess a mov and an and are that much cheaper?

Note that TP diffs purely count instructions; they are not a true measure of run-time performance difference. An instruction count is the best, extremely low noise and reproducible, proxy for run-time perf differences that we have.

amanasifkhalid:
> I think this is a good case for the use of __forceinline

This seemed to fix the remaining [diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=484511&view=ms.vss-build-web.run-extensions-tab) on x86.

> Note that TP diffs purely count instructions; they are not a true measure of run-time performance difference.

I see, thanks for pointing that out. We're still seeing TP regressions with the inlined boolean implementation of the method. I can confirm that the boolean implementation results in a larger `clrjit.dll` disassembly. To get an idea of why, I looked at disassemblies of `Compiler::fgHasCycleWithoutGCSafePoint` (which has two calls to `HasFlag`, where the method is used as a predicate) with the two different implementations. Both calls to `HasFlag` are inlined, but the codegen is slightly different. For the `BasicBlockFlags` implementation, `HasFlag` is inlined into a `bt`, while for the boolean implementation, `HasFlag` is inlined into a `shr` and a `test`. So we are executing more instructions with the boolean approach.

If everyone's ok with the slightly less intuitive code, I'm going to revert to the `BasicBlockFlags` implementation for the better TP.

BruceForstall:
So the implementation with `__forceinline` and bool `return (bbFlags & flag) != 0` has a very slight (+0.02%) regression, but notably no regression on linux/x64 (built with clang)? I'm inclined to say we should just accept that and go for the intuitive implementation.

BruceForstall:
With the current variation that is running diffs, would you expect TP to _universally_ improve, since there are lots of places using `((block->bbFlags & flag) != 0)` that will now be doing (inlined) `(block->bbFlags & flag)` which you've shown has fewer instructions?

BruceForstall:
> I'm thinking of opening a followup PR where we `__forceinline` these methods to see if we can claw back some more TP (we could also chop up these large methods to fit them in MSVC's inlining budget, though that seems more daunting). What do we think?

Seems like a good experiment to try (the `__forceinline`). Splitting up large methods is generally "hard", but can be worthwhile. Probably best done while working in an area, not as a "go around breaking up large methods" task.

What really matters is the profile-feedback-based Release build. Note that TP diffs, while they use Release builds, disable profile feedback being used by the compiler, to ensure "apples to apples" comparison that doesn't bias the baseline (which presumably has more accurate profile data).

amanasifkhalid:
> With the current variation that is running diffs, would you expect TP to universally improve

Good question. This [previous run](https://dev.azure.com/dnceng-public/public/_build/results?buildId=484511&view=ms.vss-build-web.run-extensions-tab) is with `__forceinline` and returning `(bbFlags & flag)`. There are only some slight TP improvements on win-x86 when targeting Linux ARM. I'll take a look at the codegen with the original `((block->bbFlags & flag) != 0)` approach to see if there's any difference.

amanasifkhalid:
I checked the disassembly of `Compiler::fgHasCycleWithoutGCSafePoint` with the `HasFlag` calls replaced with `((block->bbFlags & flag) != 0)`, and the codegen is identical. However, there are some odd codegen diffs elsewhere that result in a larger assembly overall with the `HasFlag` changes. I've attached a snippet below; left is without this PR's changes, and right is with them (sorry for the excessive diff highlighting):

image

That snippet is from `CodeGen::genCodeForBBlist`; looking at the changes I made there, nothing stands out to me as responsible for this diff. So maybe just some odd MSVC behavior?

BruceForstall:
@amanasifkhalid I submitted a PR to update this PR, with fixes to a couple bugs I noticed while reviewing, as well as adding some additional FORCEINLINE cases (that unfortunately don't materially improve measured TP):

https://github.com/amanasifkhalid/runtime/pull/3

amanasifkhalid:
@BruceForstall thanks for investigating this further, and for the fixes! Hopefully we'll see an MSVC fix for this soon.

BruceForstall (approved):
LGTM if tests pass

BruceForstall:
[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=491251&view=ms.vss-build-web.run-extensions-tab)

No asm diffs (as expected). No TP diffs on linux. Small TP improvements to small regressions on Windows (-0.01% to 0.02%)

amanasifkhalid:
Failures are known NativeAOT issues.

Issue #95140, opened 2023-11-22
NativeAOT: Analysis doesn't produce warnings from generic parameter annotations from a string input


vitek-karas:
```csharp
class TypeWithAnnotation<[DAM(All)] T> {}

class TypeWithRUC
{
    [RequiresUnreferencedCode()]
    public void RUCMethod() {}
}

void Method<TUnknown>()
{
    typeof(TypeWithAnnotation<TypeWithRUC>); // IL2026 - correct
    Type.GetType("TypeWithAnnotation[[TypeWithRUC]]"); // No warning
}
```

The analysis can parser type names and correctly mark such types, but it doesn't process generic parameter annotations on them. Which leads to both analysis holes (as shown above), as well as marking holes (in the above sample the RUCMethod is not preserved).

Issue #95121, opened 2023-11-22
Trim analyzer: Doesn't produce warnings due to annotated generic parameter in various situations


vitek-karas:
```csharp
Method<string>();

Type Method<TUnknown>()
{
    // Should WARN - TUnknown does have All annotation
    return typeof(TRequires<TUnknown>);
}

class TRequires<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] T>
{
}
```

The able produces IL2091 in the trimmer, but doesn't produce it in the analyzer. The bug seems to be that we don't process generic parameter annotations in a `typeof`.

Also happens for a declaration of a local variable:
```csharp
void Method<TUnknown>()
{
    // Trimmer produces IL2091 here, analyzer doesn't
    TRequires<TUnknown> a;
}
```

Note that for local variables NativeAOT doesn't produce the warning either - see the comment here for why:
https://github.com/dotnet/runtime/blob/7f818485a669eef2271034cb1aee3b01bdeeaef6/src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterWarningLocation.cs#L12-L28

We should figure out if analyzer needs to produce the warning or not.

Another situation is a static field access:
```csharp
    TRequires<TUnknown>.Field;  // Should warn
```

Yet another situation - expression trees:
```csharp
Expression<Func<string>> a = () => TRequires<TUnknown>.Field
```

And then type checks (NativeAOT doesn't warn here, see the link above):
```csharp
if (value is TRequires<TUnknown>) {}

var a = value as TRequires<TUnknown>;
```

Catch and filter expressions (NativeAOT doesn't warn):
```csharp
try {}
catch (TRequires<TUnknown>) {}
catch (Exception ex) when (ex is TRequires<TUnknown>) {}
```

Nested generic for methods:
```csharp
GenericMethod<Generic1<Generic2<TRequires<TUnknown>>>();
```

Similarly nested generics don't cause DAM marking and thus diagnostics:
```csharp
GenericMethod<Generic1<Generic2<TRequires<TypeWIthRUCMethod>>>(); // Should produce IL2026
```

vitek-karas:
Good point - If we can get the same behavior as NativeAOT I think that would be great (it warns less). Also if we can clearly define the way to do that in the type system (NativeAOT implementation kind of fell out of the way the compiler works), we could apply it to trimmer.

[...]

Issue #95118, opened 2023-11-22
Trim analyzer: Implement type name parsing


vitek-karas:
For example:
```csharp
// Should produce warning because we should be able to parse the type name, resolve it
// and find the method with RUC and thus issue a warning due to potential access to that method.
Type.GetType("TypeWithRUCMethod").GetMethods();
```

This is also happening in places with annotated string values.

There's a TODO in the code about this: https://github.com/dotnet/runtime/blob/2096625a41b8284f4388a5a42137b675df4685da/src/tools/illink/src/ILLink.RoslynAnalyzer/TrimAnalysis/RequireDynamicallyAccessedMembersAction.cs#L26-L30

sbomer:
Looks like Roslyn bumped System.Reflection.Metadata to 9.0.0 so we can go ahead with this.

PR #95109, opened 2023-11-22
Ensure invalid type in Type.GetType doesn't crash

MichalStrehovsky:
This regressed when `ReflectedTypeNode` was introduced. Previously we'd just ask for an `EETypeNode` and that one has this check in the constructor.

Fixes #94830.

Cc @dotnet/ilc-contrib

PR #95143, opened 2023-11-22
Trimming: Add tests for handling of annotated string type names representing generic instantiations

vitek-karas:
This change:
* Adds new tests which validate behavior of generic parameter data flow when the type is specified as a string in either an annotated parameter or Type.GetType
  * This found new issues in both the analyzer and NativeAOT * https://github.com/dotnet/runtime/issues/95140 * https://github.com/dotnet/runtime/issues/95118
* Adds a test for invalid type name in the same situation - this is basically another test for https://github.com/dotnet/runtime/issues/94830 (And the related fix in https://github.com/dotnet/runtime/pull/95109)
* Fixes test names in the NativeAOT test infra for the shared tests
* Enable the generic parameter data flow warnings test full validation in the analyzer
  * This surfaced a lot of issues in the analyzer around generic parameter data flow - https://github.com/dotnet/runtime/issues/95121


sbomer (approved):
Thank you!

PR #95152, opened 2023-11-23
JIT: Assert bbJumpKind in BasicBlock::HasJump

amanasifkhalid:
Based on feedback in #94239, before reading a block's `bbJumpDest` in `BasicBlock::HasJump`, we should assert it has a valid jump kind. In other words, calling `HasJump` on a block that is never expected to have a jump destination should be considered invalid behavior.


amanasifkhalid:
cc @dotnet/jit-contrib, @BruceForstall PTAL

src/coreclr/jit/block.h
```diff
@@ -662,13 +662,13 @@ struct BasicBlock : private LIR::Range
 
     bool HasJump() const
     {
+        assert(KindIs(BBJ_ALWAYS, BBJ_CALLFINALLY, BBJ_COND, BBJ_EHCATCHRET, BBJ_EHFILTERRET, BBJ_LEAVE));
         return (bbJumpDest != nullptr);
     }
```
jakobbotsch, line 667: What is the point of `HasJump()`? It seems like a check for whether the block is in some half-constructed state, since I would expect the block kinds above always to have valid `bbJumpDest`. It does not really make sense for me to have this function with the assert that is added here.
An assert that would make more sense to me would be:
```
assert(KindIs(BBJ_ALWAYS, BBJ_CALLFINALLY, BBJ_COND, BBJ_EHCATCHRET, BBJ_EHFILTERRET, BBJ_LEAVE) == (bbJumpDest != nullptr))
```

jakobbotsch: Ah, I see that @BruceForstall's feedback in #94239 was about this being a union field.

I would like to understand when it is expected that these jump kinds have a null `bbJumpDest`. Maybe that happens only for a very short while when we're initializing the flow graph? In that case, consider renaming this function to `HasInitializedJumpDest()` or something of the sort, to make that clear. I think it is a bit confusing currently, as it makes it seem like an expected situation that some of these blocks do not have a valid `bbJumpDest`, which should not be possible.

amanasifkhalid: > I would like to understand when it is expected that these jump kinds have a null bbJumpDest. Maybe that happens only for a very short while when we're initializing the flow graph?

Yes, there are a few cases where we have to create a block that requires a jump target without knowing the target yet. For example, in `Compiler::fgMakeBasicBlocks`, we create blocks only knowing the offset of their jump targets, and set the actual target block later in `Compiler::fgLinkBasicBlocks`. The tougher cases are during importing, where we sometimes create a `BBJ_ALWAYS` or `BBJ_CALLFINALLY` without setting its jump target in a loop, and then create and set the jump target in a later iteration of the loop. There are a few other spots where we do this pattern, but I can't remember off the top of my head...

Your first suggested assert seems to best communicate what `HasJump` should do; for block kinds that should never read `bbJumpDest`, `bbJumpDest` should not be set, else `bbJumpDest` must be set (i.e. we shouldn't be in some half-constructed state) before performing any conditional logic based on whether the block has a jump or not.

jakobbotsch (approved):
LGTM now. Might want to wait for @BruceForstall too since this was his feedback.

BruceForstall (approved):
LGTM.

You now have six copies of `KindIs(BBJ_ALWAYS, BBJ_CALLFINALLY, BBJ_COND, BBJ_EHCATCHRET, BBJ_LEAVE)` I had suggested helper functions:

```
    HasJumpDest() => KindIs(BBJ_ALWAYS,<others>)  // bbJumpDest is valid
    HasJumpSwt() => KindIs(BBJ_SWITCH)            // bbJumpSwt is valid
    HasJumpEhf() => KindIs(BBJ_EHFINALLYRET)      // bbJumpEhf is valid
```

Also, it seems unfortunate that you're introducing two new BBJ_NONE cases that you'll have to remove with https://github.com/dotnet/runtime/pull/94239

amanasifkhalid:
> You now have six copies of KindIs(BBJ_ALWAYS, BBJ_CALLFINALLY, BBJ_COND, BBJ_EHCATCHRET, BBJ_LEAVE) I had suggested helper functions:

Ah sorry, I had interpreted those suggestions as replacements for `HasJump`. I can add `HasJumpDest` to this PR to replace all the long `KindIs` calls.

> Also, it seems unfortunate that you're introducing two new BBJ_NONE cases that you'll have to remove with #94239 

Yeah, I was planning on doing some cleanup work to fix all the new merge conflicts in #94239 from these follow-up PRs anyway, so I'm not too worried about having to fix those new `BBJ_NONE` instances.

amanasifkhalid:
@BruceForstall PTAL. I replaced all the `KindIs(BBJ_ALWAYS, ...)` with `HasJumpDest()`.

src/coreclr/jit/block.h
```diff
@@ -622,10 +622,16 @@ struct BasicBlock : private LIR::Range
         assert(KindIs(BBJ_ALWAYS, BBJ_COND, BBJ_LEAVE));
     }
 
+    bool HasJumpDest() const
+    {
+        // These block types should always have bbJumpDest set
+        return KindIs(BBJ_ALWAYS, BBJ_CALLFINALLY, BBJ_COND, BBJ_EHCATCHRET, BBJ_LEAVE);
```
jakobbotsch, line 628: This is missing `BBJ_EHFILTERRET`

BruceForstall (approved):

PR #95173, opened 2023-11-23
Update Array.cs

meokullu:
Removing additional space character.

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Array.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Array.cs
@@ -1045,7 +1045,7 @@ namespace System
         // index into the array.
         //
         // This method is to support the ICollection interface, and calls
-        // Array.Copy internally.  If you aren't using ICollection explicitly,
+        // Array.Copy internally. If you aren't using ICollection explicitly,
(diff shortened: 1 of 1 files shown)


src/libraries/System.Private.CoreLib/src/System/Array.cs
```diff
@@ -1045,7 +1045,7 @@ public static int BinarySearch<T>(T[] array, int index, int length, T value, ICo
         // index into the array.
         //
         // This method is to support the ICollection interface, and calls
-        // Array.Copy internally.  If you aren't using ICollection explicitly,
+        // Array.Copy internally. If you aren't using ICollection explicitly,
```
stephentoub, line 1048: Thanks, but it's very common to use two spaces after a period. It's very unlikely this is the only one in the codebase.

meokullu: Wow, it is the first time ever I heard this. I didn't know that was common thing. Out of that on all codes, I didn't see it.

adamsitnik:
Based on what @stephentoub wrote, I am closing the PR. Thank you for your contribution @meokullu

PR #95180, opened 2023-11-23
[browser] js string marshaling by value

pavelsavara:
New MSbuild flag `WasmJsInteropByValue` which is `false` by default on ST build and `true` on MT build.

Make string marshaling by value rather than by reference to `MonoString*`

The motivation is that 
- Native AOT is not Mono
- MT may need to marshal to unmanaged thread

This removes optimization we have for interned strings, when enabled.

I didn't fix it for legacy interop.

### DIFF
--- a/src/mono/wasm/build/WasmApp.targets
+++ b/src/mono/wasm/build/WasmApp.targets
@@ -73,6 +73,7 @@
                                               Defaults to false.
       - $(WasmAotProfilePath)               - Path to an AOT profile file.
       - $(WasmEnableLegacyJsInterop)        - Include support for legacy JS interop. Defaults to true.
+      - $(WasmEnableJsInteropByValue)       - Make JS interop to pass string by value instead of by reference. Defaults to false. Defaults to true for build with threads.
       - $(WasmEnableExceptionHandling)      - Enable support for the WASM post MVP Exception Handling runtime extension.
       - $(WasmEnableSIMD)                   - Enable support for the WASM post MVP SIMD runtime extension.
       - $(WasmEnableWebcil)                 - Enable conversion of assembly .dlls to Webcil wrapped in .wasm (default: true)
@@ -115,6 +116,8 @@
     <WasmEnableExceptionHandling Condition="'$(WasmEnableExceptionHandling)' == ''">true</WasmEnableExceptionHandling>
     <WasmEnableSIMD Condition="'$(WasmEnableSIMD)' == ''">$(WasmEnableExceptionHandling)</WasmEnableSIMD>
     <WasmEnableLegacyJsInterop Condition="'$(WasmEnableLegacyJsInterop)' == ''">true</WasmEnableLegacyJsInterop>
+    <WasmEnableJsInteropByValue Condition="'$(WasmEnableJsInteropByValue)' == '' and ( '$(WasmEnableThreads)' == 'true' or '$(MonoWasmBuildVariant)' == 'multithread' )">true</WasmEnableJsInteropByValue>
+    <WasmEnableJsInteropByValue Condition="'$(WasmEnableJsInteropByValue)' == ''">false</WasmEnableJsInteropByValue>
 
     <!--<WasmStripAOTAssemblies Condition="'$(AOTMode)' == 'LLVMOnlyInterp'">false</WasmStripAOTAssemblies>-->
     <!--<WasmStripAOTAssemblies Condition="'$(WasmStripAOTAssemblies)' == ''">$(RunAOTCompilation)</WasmStripAOTAssemblies>-->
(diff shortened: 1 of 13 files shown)


kg (approved):
I'm not sure I understand why we're doing this, but the code looks fine

src/mono/wasm/wasm.proj
```diff
@@ -27,6 +27,7 @@
     <WasmEnableSIMD Condition="'$(WasmEnableSIMD)' == ''">true</WasmEnableSIMD>
     <WasmEnableExceptionHandling Condition="'$(WasmEnableExceptionHandling)' == ''">true</WasmEnableExceptionHandling>
     <WasmEnableLegacyJsInterop Condition="'$(WasmEnableLegacyJsInterop)' == ''">true</WasmEnableLegacyJsInterop>
+    <WasmJsInteropByValue Condition="'$(TargetPlatformIdentifier)' == 'browser' and '$(WasmJsInteropByValue)' == '' and '$(MonoWasmThreads)' == 'true'">true</WasmJsInteropByValue>
```
maraf, line 30: nit: Just for consistency

```suggestion
    <WasmEnableJsInteropByValue Condition="'$(TargetPlatformIdentifier)' == 'browser' and '$(WasmEnableJsInteropByValue)' == '' and '$(MonoWasmThreads)' == 'true'">true</WasmEnableJsInteropByValue>
```

pavelsavara:
> I'm not sure I understand why we're doing this

@kg

- For NativeAOT sake - there is no Mono and MonoString* there.
- for deputy MT experiment, dealing with MonoString requires we enter `MONO_ENTER_GC_UNSAFE` on UI thread, which could be blocking operation.

maraf (approved):

src/mono/wasm/build/WasmApp.targets
```diff
@@ -115,6 +116,8 @@
     <WasmEnableExceptionHandling Condition="'$(WasmEnableExceptionHandling)' == ''">true</WasmEnableExceptionHandling>
     <WasmEnableSIMD Condition="'$(WasmEnableSIMD)' == ''">$(WasmEnableExceptionHandling)</WasmEnableSIMD>
     <WasmEnableLegacyJsInterop Condition="'$(WasmEnableLegacyJsInterop)' == ''">true</WasmEnableLegacyJsInterop>
+    <WasmEnableJsInteropByValue Condition="'$(WasmEnableJsInteropByValue)' == '' and ( '$(WasmEnableThreads)' == 'true' or '$(MonoWasmBuildVariant)' == 'multithread' )">true</WasmEnableJsInteropByValue>
```
radical, line 119: Isn't `WasmEnableThreads` user facing, and `MonoWasmBuildVariant` is meant more for the wasm/runtime build itself? I see that it is being used in a couple of other places, which we should fix too. Not necessary for this PR.

pavelsavara: This makes it easy to work in-tree with samples and unit tests when you just have `MonoWasmBuildVariant` set globally and sometimes native rebuild kicks it. Maybe it could be unified somehow, I'm not 100% sure about it.

radical: You can set `WasmEnableThreads` based on `MonoWasmBuildVariant` in `WasmApp.InTree.targets`.

pavelsavara: I would prefer if you could help me to improve it in different PR. But I would like to ask details in meantime

what are you supposed to use `WasmEnableThreads` or `MonoWasmBuildVariant` inside of
- wasm.proj
- WasmApp.Native.targets 
- C code like drive.c triggered by runtime build or application native rebuild
- inside .csproj of libraries.
- inside .csproj of libraries unit tests
- inside .csproj of in-tree samples
- in .csproj of WBT

What are you supposed to use `WasmEnableThreads` or `MonoWasmBuildVariant` on commandline ?
- when you build runtime and libraries with `build.sh`
- when you run unit tests. This could trigger re-build of C# libraries
- when you run samples in-tree.  This could trigger re-build of C# libraries

How does it propagate to the nested build/publish ?
Does build on helix changes anything about it ?
Does WBT runtime installation in artifacts changes anything about it ?
Is there really benefit of having 2 of them ?

radical: `MonoWasmBuildVariant` is really only for non-user projects, so wasm.proj would be an example.
All the other test cases (except WBT) use InTree targets, so adding overrides there would apply to all of them.
WBT strictly uses what the user gets, so no InTree targets.

> Is there really benefit of having 2 of them ?

Not really. I think we had that before for `multithread`, and `perftrace` builds but now we don't need a string property. It is clearer to have them separate but I wouldn't mind using only `WasmEnableThreads` either.
What I do want to avoid is using internal build properties in user facing targets.

cc @lambdageek

Issue #2389, opened 2020-01-30
[mono] Tests failed: System.Reflection.Emit.Tests


MaximLipnin:
This is a tracking issue for multiple failures in System.Reflection.Emit.Tests on Mono.
The following tests are disabled in CoreFX.issues.rsp file and will be moved into their respective xUnit tests in #2087.  

- System.Reflection.Emit.Tests.AssemblyTests.DefineDynamicModule (Mono behaves like coreclr but the test doesn't detect it)
- System.Reflection.Emit.Tests.MethodBuilderEquals.Equals
- System.Reflection.Emit.Tests.ConstructorBuilderToString.ToString_NullRequiredOptionalCustomModifiers
- System.Reflection.Emit.Tests.ConstructorBuilderToString.ToString_NoRequiredOptionalCustomModifiers
- System.Reflection.Emit.Tests.EnumBuilderMethodTests.DefineLiteral
- System.Reflection.Emit.Tests.EnumBuilderMethodTests.DefineLiteral_InvalidLiteralValue_ThrowsArgumentException
- System.Reflection.Emit.Tests.EnumBuilderMethodTests.SetCustomAttribute_ConstructorInfo_ByteArray
- System.Reflection.Emit.Tests.EnumBuilderMethodTests.SetCustomAttribute_CustomAttributeBuilder
- System.Reflection.Emit.Tests.EnumBuilderPropertyTests.Guid_TypeCreated
- System.Reflection.Emit.Tests.FieldBuilderSetConstant.SetConstant_InvalidType_ThrowsArgumentException
- System.Reflection.Emit.Tests.GenericTypeParameterBuilderSetInterfaceConstraints.SetInterfaceConstraints_OneCustomInterface
- System.Reflection.Emit.Tests.GenericTypeParameterBuilderSetInterfaceConstraints.SetInterfaceConstraints_MultipleCustomInterfaces
- System.Reflection.Emit.Tests.MethodBuilderDefineGenericParameters.DefineGenericParameters_SingleTypeParameter_SetImplementationFlagsCalled_ThrowsInvalidOperationException
- System.Reflection.Emit.Tests.MethodBuilderDefineGenericParameters.DefineGenericParameters_TwoTypeParameters_SetImplementationFlagsCalled_ThrowsInvalidOperationException
- System.Reflection.Emit.Tests.MethodBuilderDefineGenericParameters.DefineGenericParameters_SingleTypeParameter_AlreadyDefined_ThrowsInvalidOperationException
[...]

PR #78544, opened 2022-11-18
Make System.Reflection.Emit public types abstract

buyaa-n:
We plan to abstract out Reflection.Emit APIs for adding [`AssemblyBuilder.Save()` in .NET Core](https://github.com/dotnet/runtime/issues/15704). So that we can have two implementations, one that having old code to support all downlevel runtime, and a new implementation that replaces the current Reflection.Emit implementation with AssemblyBuilder.Save() support.

This PR is based on [@jkotas prototype](https://github.com/dotnet/runtime/compare/main...jkotas:runtime:reflection-emit), but only covers the effort of abstracting all public APIs in System.Reflection.Emit.

Fixes https://github.com/dotnet/runtime/issues/78542

PR #95183, opened 2023-11-23
[mono][System.Reflection.Emit] Fix explicitly setting the return type on a dynamic method

### LINKED ISSUE #94490: [mono][System.Reflection.Emit] VTable initialization failure for class implementing an interface
When using System.Reflection.Emit with Mono, using `MethodBuilder.SetReturnType` and `MethodBuilder.SetParameters` on methods that are supposed to implement interface methods does not work - the method is not considered to implement the interface method.  Using the long form of `TypeBuilder.DefineMethod` works.

```csharp
using System;
using System.Reflection;
using System.Reflection.Emit;

public interface IStr
{
    public string MStr(string x);
}

public class P {
    public static void Main()
    {
        var aname = "TestAssembly";
        var assemblyName = new AssemblyName(aname);
        var assemblyBuilder = AssemblyBuilder.DefineDynamicAssembly(assemblyName, AssemblyBuilderAccess.Run);
        var moduleBuilder = assemblyBuilder.DefineDynamicModule(aname);

        var typeBuilder = moduleBuilder.DefineType("TestType", TypeAttributes.Public, typeof(object), new Type[] { typeof(IStr) });

#if false
        //works
        var methodBuilder = typeBuilder.DefineMethod("MStr", MethodAttributes.Public | MethodAttributes.Virtual, typeof(string), new Type[] { typeof(string) });
#else
        // throws "Invalid VTable" from CreateInstance, below
        var methodBuilder = typeBuilder.DefineMethod("MStr", MethodAttributes.Public | MethodAttributes.Virtual);
        methodBuilder.SetReturnType(typeof(string));
        methodBuilder.SetParameters(new Type[] { typeof(string) });
#endif

        var ilg = methodBuilder.GetILGenerator();
        ilg.Emit(OpCodes.Ldarg_1);
        ilg.Emit(OpCodes.Ret);

        var type = typeBuilder.CreateType();
        var obj = Activator.CreateInstance(type);
        var m = type.GetMethod("MStr");
        var r = m.Invoke(obj, new object[] { "hello" });
        Console.WriteLine($"r = {r}");
    }
}
```

*Update* simpler repro and better diagnosis; original issue below

---

Needs some investigation still.  Some of the System.Reflection.Emit code generates a class with a bad vtable on net8.0.

The TypedClientBuilder is from aspnetcore:

https://github.com/dotnet/aspnetcore/blob/556f39af49b1a32f14089d3809557825932e4ca5/src/SignalR/server/Core/src/Internal/TypedClientBuilder.cs#L49

Repro:

Clone this repo https://github.com/lambdageek/repro-iclientproxy-vtable-fail

**Expected**

```console
$ dotnet run
SendCoreAsync - M1
```

**Actual**

```console
$ dotnet run
Unhandled Exception:
System.TypeLoadException: VTable setup of type Microsoft.AspNetCore.SignalR.TypedClientBuilder.I1Impl failed
   at System.RuntimeType.GetMethodsByName(String name, BindingFlags bindingAttr, MemberListType listType, RuntimeType reflectedType)
   at System.RuntimeType.GetMethodCandidates(String name, BindingFlags bindingAttr, CallingConventions callConv, Type[] types, Int32 genericParamCount, Boolean allowPrefixLookup)
   at System.RuntimeType.GetMethodImpl(String name, Int32 genericParamCount, BindingFlags bindingAttr, Binder binder, CallingConventions callConv, Type[] types, ParameterModifier[] modifiers)
   at System.RuntimeType.GetMethodImpl(String name, BindingFlags bindingAttr, Binder binder, CallingConventions callConvention, Type[] types, ParameterModifier[] modifiers)
   at System.Type.GetMethod(String name, BindingFlags bindingAttr)
   at Retrofit.TypedClientBuilder`1[[I1, promo, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]].GenerateClientBuilder() in /private/tmp/promo/TypedClientBuilder.cs:line 42
   at System.Lazy`1[[System.Func`2[[Retrofit.IClientProxy, promo, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null],[I1, promo, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]], System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1[[System.Func`2[[Retrofit.IClientProxy, promo, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null],[I1, promo, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]], System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1[[System.Func`2[[Retrofit.IClientProxy, promo, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null],[I1, promo, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]], System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].CreateValue()
   at System.Lazy`1[[System.Func`2[[Retrofit.IClientProxy, promo, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null],[I1, promo, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]], System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].get_Value()
   at Retrofit.TypedClientBuilder`1[[I1, promo, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]].Build(IClientProxy proxy) in /private/tmp/promo/TypedClientBuilder.cs:line 24
   at P.Main() in /private/tmp/promo/Program.cs:line 19
   at P.<Main>()
```

ivanpovazan:
## Description

This PR fixes the support of `SetReturnType` and `SetParameters` in the `MethodBuilder` API with Mono, by properly setting the return type of the dynamic method only if it is non null value.
 
This was manifested in two scenarios:
1. If the dynamic method return and parameter types are manually set in the following sequence: 
    ```cs
     methodBuilder.SetReturnType(interfaceMethodInfo.ReturnType);
     methodBuilder.SetParameters(paramTypes);
    ```
     The subsequent call to `SetParameters` (which internally calls `SetSignatureCore`) would overwrite the set return type to `null` causing the method not to be considered as an implementation of a interface method. This scenario is used in https://github.com/dotnet/aspnetcore/blob/4dc81f80abd45cb5c2e04d3b742bc6755c1153ee/src/SignalR/server/Core/src/Internal/TypedClientBuilder.cs#L136-L137 causing ASP.NET test failures when using Mono.

2. Setting explicit `null` as a return type via `SetReturnType` overwrites the default return type value which should be `typeof(void)`  

    This scenario was covered with `System.Reflection.Emit` libraries tests https://github.com/dotnet/runtime/blob/4fa760f59503b9bb24983ef4275e77ee2fd375c1/src/libraries/System.Reflection.Emit/tests/MethodBuilder/MethodBuilderSetReturnType.cs#L141-L143 but was disabled for Mono as reported in: https://github.com/dotnet/runtime/issues/2389

## Verification

This fix was verified by enabling the disabled unit test (and verifying locally that it passes) and manually verifying that the [repro-iclientproxy-vtable-fail](https://github.com/lambdageek/repro-iclientproxy-vtable-fail) as described in https://github.com/dotnet/runtime/issues/94490#issue-1982229368 does not crash.

---
Fixes https://github.com/dotnet/runtime/issues/94490

### DIFF
--- a/src/libraries/System.Reflection.Emit/tests/MethodBuilder/MethodBuilderSetReturnType.cs
+++ b/src/libraries/System.Reflection.Emit/tests/MethodBuilder/MethodBuilderSetReturnType.cs
@@ -132,7 +132,6 @@ namespace System.Reflection.Emit.Tests
         }
 
         [Fact]
-        [ActiveIssue("https://github.com/dotnet/runtime/issues/2389", TestRuntimes.Mono)]
         public void SetReturnType_NullReturnType_ReturnsVoid()
         {
             TypeBuilder type = Helpers.DynamicType(TypeAttributes.Abstract);
--- a/src/mono/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeMethodBuilder.Mono.cs
+++ b/src/mono/System.Private.CoreLib/src/System/Reflection/Emit/RuntimeMethodBuilder.Mono.cs
@@ -603,7 +603,9 @@ namespace System.Reflection.Emit
                 Array.Copy(parameterTypes, this.parameters, parameterTypes.Length);
             }
 
-            rtype = returnType;
+            if (returnType != null)
+                rtype = returnType;
+
             returnModReq = returnTypeRequiredCustomModifiers;
             returnModOpt = returnTypeOptionalCustomModifiers;
             paramModReq = parameterTypeRequiredCustomModifiers;


ivanpovazan:
Once this fix gets approved, we should consider backporting it to .net8 servicing release as it affects signalR users on RHEL.

lambdageek (approved):

buyaa-n (approved):
Thank you for the fix.

> Once this fix gets approved, we should consider backporting it to .net8 servicing release as it affects signalR users on RHEL.

💯, this looks regression introduced in 8.0 with https://github.com/dotnet/runtime/pull/78544

PR #95215, opened 2023-11-24
Bump dependencies of markdownlint action

vcsjones:
Node 16 is EOL, so bump to the latest LTS of node. While we are here, bump the versions of the actions.

### DIFF
--- a/.github/workflows/markdownlint.yml
+++ b/.github/workflows/markdownlint.yml
@@ -17,11 +17,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - name: Use Node.js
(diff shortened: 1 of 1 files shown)


ViktorHofer (approved):

.github/workflows/markdownlint.yml
```diff
@@ -17,11 +17,11 @@ jobs:
...
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - name: Use Node.js
-      uses: actions/setup-node@v1
+      uses: actions/setup-node@v4
       with:
-        node-version: 16.x
+        node-version: 20.x
```
ViktorHofer, line 24: Might make sense to use `lts` here: https://github.com/actions/setup-node#usage

PR #95221, opened 2023-11-25
[NativeAOT] Aging stacks

VSadov:
Do not scan stacks that cannot possibly have any roots interesting to the current GC.


src/coreclr/nativeaot/Runtime/gcrhenv.cpp
```diff
@@ -659,10 +659,48 @@ void GCToEEInterface::SyncBlockCacheWeakPtrScan(HANDLESCANPROC /*scanProc*/, uin
...
+
+        // the stack is too old to be interesing in this GC
+        if (generation > condemned)
+            continue;
+
+        // the stack is as young as it can be
+        if (generation == 0)
+            continue;
+
+        pThread->SetGeneration(0);
```
VSadov, line 676: Just setting generation to 0 as we do not want to dig through the entire stack and see what aged and what did not.

PR #95227, opened 2023-11-25
JIT: Disallow loop increments in the top block

### LINKED ISSUE #95199: JIT: Assertion failed 'i < BitSetTraits::GetSize(env)' during profile synthesis
If I run SPMI replay of `coreclr_tests` with `-jitoption JitSynthesizeCounts=1` I hit this assert on a few contexts:
```scala
*************** Inline @[000028] Starting PHASE Profile incorporation
Synthesizing profile data

New BlockSet epoch 1, # of blocks (including unused BB00): 4, bitset array size: 1 (short)
ISSUE: <ASSERT> #1 C:\dev\dotnet\runtime4\src\coreclr\jit\bitsetasshortlong.h (316) - Assertion failed 'i < BitSetTraits::GetSize(env)' in 'Program:<<Main>$>g__TestExecutor88|0_89(System.IO.StreamWriter,System.IO.StreamWriter,byref)' during 'Morph - Inlining' (IL size 271; hash 0x78af786b; FullOpts)
```

The inlinee inherits the EH table from the inliner (why?), so the RPO should probably skip handlers for inlinees. Or we should stop inheriting the EH table (I don't really see why we would do this).

### LINKED ISSUE #95226: JIT: Incorrect removal of bounds check in some loop structures
When `DOTNET_JitDoLoopInversion=0` the following program does not throw an exception, even though it should. We incorrectly remove the bounds check for `arr[i]`.
```csharp
public class Program
{
    public static void Main()
    {
        for (int i = 0; i < 4; i++)
        {
            for (int j = 0; j < 200; j++)
            {
                try
                {
                    Bar(new int[10], new C());
                }
                catch (IndexOutOfRangeException)
                {
                }
            }

            Thread.Sleep(500);
        }

        Bar(new int[1], new C());
    }

    private static void Bar(int[] arr, I iface)
    {
        if (arr == null)
        {
            return;
        }

        iface.Foo();

        int i = 0;
        goto Bottom;
Top:;
        i++;
Bottom:;
        Use(arr[i]);
        if (i < arr.Length)
            goto Top;
    }

    private static void Use(int x)
    {
    }
}

internal interface I
{
    void Foo();
}

internal class C : I
{
    public void Foo()
    {
    }
}
```

The problem arises due to several factors:
1. The following code is questionable: https://github.com/dotnet/runtime/blob/2f8ef582a8e2e7e7996267694cd7ac73f9e8f34b/src/coreclr/jit/optimizer.cpp#L1130-L1135
This means we proceed as if we successfully analyzed the induction structure of the loop. But in normal cases for an `LPFLG_ITER` loop we would expect the `i < arr.Length` to be a known fact in the loop body; that's not true here due to the `i++`. The code linked here allows loop shapes where the test and increment are in opposite order.
2. With default JIT settings the loop structure above gets broken by loop inversion + flow graph optimizations, hence why `DOTNET_JitDoLoopInversion=0` is required. I would expect the problem exists even with loop inversion enabled if you can trick flow graph optimizations to avoid doing some compactions here. The fact that our induction analysis is so "pattern matchey" makes it hard to get the wrong structure past loop inversion + flow graph opts in a way that it still matches.
3. We normally skip loop cloning entirely for any loop that was _not_ inverted: https://github.com/dotnet/runtime/blob/2f8ef582a8e2e7e7996267694cd7ac73f9e8f34b/src/coreclr/jit/loopcloning.cpp#L1893-L1900
However, we skip this check in methods with GDV.

Overall I think (1) is the actual bug; we should not make it seem like we managed to analyze the induction structure in this case.

(Edit: See https://github.com/dotnet/runtime/issues/95226#issuecomment-1826359517 for a test case that hits problem (1) with default JIT config)

jakobbotsch:
If we have a loop increment in the top block then the body won't be dominated by the loop test, so we should not consider the induction structure to have been analyzed successfully.

I suspect this code exists from a time where flow graph optimizations did not do compaction as well as today. If we did not do compaction after loop inversion then I think many loops would have structure such as the following, taken from one of the very few diffs of this change:

```
------------ BB50 [297..2AC), preds={BB51} succs={BB51}

***** BB50
STMT00061 ( 0x2A6[E-] ... ??? )
               [000230] DA---+-----                         ▌  STORE_LCL_VAR int    V14 loc13
               [000229] -----+-----                         └──▌  ADD       int
               [000227] -----+-----                            ├──▌  LCL_VAR   int    V14 loc13
               [000228] -----+-----                            └──▌  CNS_INT   int    1

------------ BB51 [2AC..2B4) -> BB50 (cond), preds={BB49,BB50} succs={BB52,BB50}

***** BB51
STMT00059 ( 0x2AC[E-] ... 0x2B2 )
               [000217] ---X-+-----                         ▌  JTRUE     void
               [000216] J--X-+-N---                         └──▌  LT        int
               [000213] -----+-----                            ├──▌  LCL_VAR   int    V14 loc13
               [000215] ---X-+-----                            └──▌  ARR_LENGTH int
               [000214] -----+-----                               └──▌  LCL_VAR   ref    V05 loc4

```
Here the top block is a BBJ_NONE with the increment, and the bottom block test succeeds it immediately. I do not think it's worth it to add additional checks for this case since compaction handles it normally.

Fix #95226

### DIFF
--- a/src/coreclr/jit/optimizer.cpp
+++ b/src/coreclr/jit/optimizer.cpp
@@ -1120,23 +1120,9 @@ bool Compiler::optExtractInitTestIncr(
         incrStmt = incrStmt->GetPrevStmt();
     }
 
-    if (incrStmt == nullptr || optIsLoopIncrTree(incrStmt->GetRootNode()) == BAD_VAR_NUM)
+    if (incrStmt == nullptr || (optIsLoopIncrTree(incrStmt->GetRootNode()) == BAD_VAR_NUM))
     {
-        if (top == nullptr || top->bbStmtList == nullptr || top->bbStmtList->GetPrevStmt() == nullptr)
-        {
-            return false;
-        }
-
-        // If the prev stmt to loop test is not incr, then check if we have loop test evaluated into a tmp.
-        Statement* toplastStmt = top->lastStmt();
-        if (optIsLoopIncrTree(toplastStmt->GetRootNode()) != BAD_VAR_NUM)
-        {
-            incrStmt = toplastStmt;
-        }
-        else
-        {
-            return false;
-        }
+        return false;
     }
 
     assert(testStmt != incrStmt);
--- a/src/tests/JIT/Regression/JitBlue/Runtime_95226/Runtime_95226.cs
+++ b/src/tests/JIT/Regression/JitBlue/Runtime_95226/Runtime_95226.cs
@@ -1,0 +1,41 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.CompilerServices;
+using Xunit;
+
+public class Runtime_95226
+{
+    [Fact]
+    public static void TestEntryPoint()
+    {
+        Assert.Throws<IndexOutOfRangeException>(() => Foo(new int[1]));
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static void Foo(int[] arr)
+    {
+        int i = 0;
+        goto Bottom;
+Top:;
+        i++;
+        while (true)
+        {
+            if (AlwaysTrue())
+                break;
+        }
+Bottom:;
+        Use(arr[i]);
+        if (i < arr.Length)
+            goto Top;
+    }
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    private static bool AlwaysTrue() => true;
+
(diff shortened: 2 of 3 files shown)


jakobbotsch:

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=480139&view=ms.vss-build-web.run-extensions-tab). Small number of cases where we now reject the induction structure of a loop and thus stop cloning.

BruceForstall (approved):

PR #94667, opened 2023-11-13
Disable optimization which sometimes results in incorrect case sensitivity in FrozenCollections

andrewjsaid:
See https://github.com/dotnet/runtime/pull/93986 for full context

Fix #93974

PR #95232, opened 2023-11-25
Reintroduce case sensitive comparison optimization for FrozenDictionary in some cases

andrewjsaid:
In https://github.com/dotnet/runtime/pull/94667 we fixed a bug where `FrozenDictionary<string, T>` was incorrectly using case sensitive comparison. The solution chosen minimized the diff as the PR would be ported to .NET 8. Upon reflection, alongside fixing the bug, it did de-optimize some cases where the optimization was in fact valid. The examples below give a quick summary.

```
// PR94667 correctly targeted this scenario. Case sensitivity matters here even if only hashing last character.
keys1 = ["abc1", "abc2", "abc3", "abc4", "abc5", "abc6"];

// PR94667 left this scenario unchanged. Case sensitivity doesn't matter but the hash is against the entire string.
keys2 = ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9"];

// PR94667 unnecessarily de-optimized this scenario. Case sensitivity doesn't matter even if only hashing last character.
keys3 = ["001", "002", "003", "004", "005", "006"];
```

This PR I am submitting re-introduces the optimization for scenarios like `keys3` above i.e. where:
1. The `KeyAnalyzer` has found a substring
2. The entirety of all keys are ASCII
3. There are no letters in any of the keys - not only checking the substring.

I am not submitting any benchmarks up-front as it is a re-introduction of an optimization in some cases and there are no benchmarks for case insensitive Frozen Dictionaries. Please let me know if there's any relevant benchmarks I should run.


src/libraries/System.Collections.Immutable/src/System/Collections/Frozen/String/KeyAnalyzer.cs
```diff
@@ -144,11 +144,21 @@ private static bool TryUseSubstring(ReadOnlySpan<string> uniqueStrings, bool ign
...
+                    if (canSwitchIgnoreCaseHashToCaseSensitive)
                     {
-                        canSwitchIgnoreCaseHashToCaseSensitive = false;
+                        // If we are testing the entire string then s == substring
+                        // and we have already checked IsAllAscii(s).
+                        if (isSubstring && !IsAllAscii(s.AsSpan()))
+                        {
+                            canSwitchIgnoreCaseHashToCaseSensitive = false;
+                        }
+                        else if (ContainsAnyLetters(s.AsSpan()))
```
stephentoub, line 158: Can we rename ContainsAnyLetters to ContainsAnyAsciiLetters?

src/libraries/System.Collections.Immutable/src/System/Collections/Frozen/String/KeyAnalyzer.cs
```diff
@@ -144,11 +144,21 @@ private static bool TryUseSubstring(ReadOnlySpan<string> uniqueStrings, bool ign
...
-                    if (canSwitchIgnoreCaseHashToCaseSensitive && ContainsAnyLetters(substring))
+                    // All substrings so far are still ASCII only.  Now we check whether
+                    // the entire string either is not ASCII or contains ASCII letters,
+                    // in which case we disable the switch to case sensitive.
+                    if (canSwitchIgnoreCaseHashToCaseSensitive)
                     {
-                        canSwitchIgnoreCaseHashToCaseSensitive = false;
+                        // If we are testing the entire string then s == substring
+                        // and we have already checked IsAllAscii(s).
+                        if (isSubstring && !IsAllAscii(s.AsSpan()))
```
stephentoub, line 154: The existing naming employed here is confusing in conjunction with this comment, since "is substring" being true means s _is not_ substring.

Let's rename "s" to "uniqueString" and "substring" to "uniqueStringSlice", or something like that.

andrewjsaid:
@stephentoub I have addressed your comments. Thanks

danmoseley:
Is it worth adding benchmarks, you mention they don't exist?

andrewjsaid:
@danmoseley I would recommend that at least a few benchmarks are added to dotnet/performance with case insensitive comparison, yes.

I can try to find the time to do so but I can not reliably make that commitment at this moment.

stephentoub (approved):
Thanks!

andrewjsaid:
Sorry to pester but as it's approved, could it also be merged, please?

PR #95233, opened 2023-11-25
Convert HELPER_METHOD_FRAME to QCalls (2/N)

jkotas:


vargaz:
The mono changes look ok.

src/coreclr/vm/olevariant.cpp
```diff
@@ -1154,13 +1154,6 @@ void VariantData::NewVariant(VariantData * const& dest, const CVTypes type, INT6
...
 }
 
-void SafeVariantClearHelper(_Inout_ VARIANT* pVar)
-{
-    WRAPPER_NO_CONTRACT;
-
-    VariantClear(pVar);
-}
-
 class OutOfMemoryException;
```
AaronRobinsonMSFT, line 1157: Is this forward declaration still needed?

src/coreclr/vm/olevariant.cpp
```diff
@@ -1176,52 +1169,16 @@ void SafeVariantClear(VARIANT* pVar)
...
-            // currently thinks that the exception is still on the fly.
-            GetThread()->GetCurrentStackGuard()->RestoreCurrentGuard();
-#endif
-
-            SCAN_EHMARKER_END_CATCH();
-        }
-        PAL_CPP_ENDTRY;
-
+        VariantClear(pVar);
         FillMemory(pVar, sizeof(VARIANT), 0x00);
```
AaronRobinsonMSFT, line 1173:
```suggestion
        FillMemory(pVar, sizeof(VARIANT), 0x00);
        V_VT(pVar) = VT_EMPTY;
```

This is sort of an odd thing. The `vt` field is being overwritten here. When `VariantClear` is called the instance is reset to `VT_EMPTY`. We should either explicit reset it to `VT_EMPTY` or comment that since `VT_EMPTY` is 0 this is ensuring that value remains.

AaronRobinsonMSFT (approved):

src/coreclr/vm/olevariant.cpp
jkotas: Yes, it is odd. It would be best to delete it, but it is legacy COM interop where users depend on every undocumented detail.

AaronRobinsonMSFT (approved):

jkoritzinsky (approved):

PR #95234, opened 2023-11-25
Improve branch optimizer around implied relops

EgorBo:
Should help with cases like:
```cs
void Test(int x)
{
    if (x > 100)
        if (x > 10) // always true
            Console.WriteLine();
}
```
```diff
; Method Program:Test(int):this (FullOpts)
       sub      rsp, 40
       cmp      edx, 100
       jle      SHORT G_M52364_IG04
-      cmp      edx, 10
-      jle      SHORT G_M52364_IG04
       call     [System.Console:WriteLine()]
G_M52364_IG04:
       nop      
       add      rsp, 40
       ret      
-; Total bytes of code: 26
+; Total bytes of code: 21
```

~~Ideally, it should be done in "redundant branch opt" phase, but it seems to be way more complicated so at least let's have it in assertprop for now: https://github.com/EgorBo/runtime-1/commit/cdbaac0689457118901d92227213acd39fb52271~~


src/coreclr/jit/assertionprop.cpp
```diff
@@ -3992,6 +4200,15 @@ GenTree* Compiler::optAssertionPropGlobal_RelOp(ASSERT_VALARG_TP assertions, Gen
         return optAssertionProp_Update(newTree, tree, stmt);
     }
 
+    if (op2->IsCnsIntOrI() && tree->OperIs(GT_EQ, GT_LT, GT_LE, GT_GE, GT_GT))
```
MichalPetryka, line 4203: Does `GT_NE` not need need to be handled here?

EgorBo: > Does `GT_NE` not need need to be handled here?

Heh, yeah, I was adding it separately - it's tricky because it's not representable via "{start, end}"

EgorBo:
PTAL @AndyAyersMS @dotnet/jit-contrib 

[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=485344&view=ms.vss-build-web.run-extensions-tab)

src/coreclr/jit/redundantbranchopts.cpp
```diff
@@ -338,6 +501,67 @@ void Compiler::optRelopImpliesRelop(RelopImplicationInfo* rii)
...
+                //   if (x <= 10) // never true
+                //       goto BB4
+                //
+                // BB3:
+                //   Console.WriteLine("Taken!");
+                //
+                // BB4:
+                //   return;
+
+                // Check whether the dominating compare being "false" implies the dominated compare is known
```
AndyAyersMS, line 550: The reversal/negation here is a bit tricky, but I think I follow it.

Seems like there's a complementary case where you don't negate the upper relop and then may instead have `canInferFromTrue` abilities?

AndyAyersMS (approved):

danmoseley:
> (not too big but I have some follow-up ideas how to extend them

How many of those diffs represent places code itself should be cleaned up? Like your example.

src/coreclr/jit/redundantbranchopts.cpp
EgorBo: Yeah, basically there are two cases:

`canInferFromFalse`:

```cs
if (x > 100)
{
    if (x > 10)
    {
        ...
    }
}
```

`canInferFromTrue`:

```cs
if (x > 100)
{
    ...
}
if (x > 10)
{
    ...
}
```

I'll check separately the 2nd case (last I tried it added very little diffs)

EgorBo:
Since the last review I found a small bug - I replaced `ssize_t` with `target_ssize_t` (Linux arm failed on CI and it was the only target where we generate R2R'd code for corelib using 64bit jit for it via crosscompilation).

Issue #95032, opened 2023-11-20
Problems running coreclr tests with run.cmd/py


BruceForstall:
There are a number of problems when running coreclr tests using the run.cmd/py scripts:

### Repro files
run.py creates "repro" files to aid in reproducing test failures, but encapsulating all the environment variable and path settings needed. There are various problems:

It outputs:
```
Creating repro files at: C:\gh\runtime3\artifacts\repro\windows.x64.Checked
Repro files written.
```
But actually doesn't write the repo files there. Instead, it writes files like:
```
C:\gh\runtime3\artifacts\tests\coreclr\windows.x64.Checked\Interop\DisabledRuntimeMarshalling\DisabledRuntimeMarshalling_Disabled_NativeAssemblyEnabled\DisabledRuntimeMarshalling_Disabled_NativeAssemblyEnabled.cmd_windows_x64_Checked.cmd
```
So the repro files are written to the test directory itself, and don't have the word "repro" in them, or any obvious way to know they are repro files, and also it is hard to find them. This also confuses run.py when running tests again, because it thinks these are test wrappers that themselves need repro files.

To fix:
- [x] output the actual filename/path of each repro file created
- [x] move the repro file to the repro directory
- [x] remove the unnecessary `_windows_x64_checked.cmd` part (that's in the repro directory path already)
- [x] add prefix `repo_` to the filename to be clear
- [x] related, add "setlocal" to repro script to avoid polluting the environment when invoking the repro file

### Results directory

Currently, running the tests puts results in a `artifacts\log` directory, and (except for the problems listed above), results in a `artifacts\repro\windows.x64.Checked` directory (for example).

- [ ] I propose that all files output from a test run get put in a `results` directory (kind of like the `log` directory but better named). The files in this directory would not be changed or overwritten: if tests are run again, a new `results.1`, `results.2`, etc. directory name would be created and used.

Merged test results are spread around the tree, in directories like C:\gh\runtime3\artifacts\tests\coreclr\windows.x64.Checked\JIT\Regression\Regression_4

This includes files like:

- Regression_4.log
- Regression_4.tempLog.xml
- Regression_4.testResults.xml
- Regression_4.testStats.csv

- [ ] I propose these also go in the `results` directory, to avoid trashing them on test re-runs. E.g., C:\gh\runtime3\artifacts\results\JIT\Regression\Regression_4.

### Too verbose output

[...]

Issue #93452, opened 2023-10-13
Add display name filtering in merged tests


jakobbotsch:
The merged test wrapper currently only supports filtering by fully qualified name, which is very frustrating for two reasons:
1. Running a merged test wrapper does not print the fully qualified name of the tests it is running anywhere, so you have to know what string it is expecting and go look it up for your test
2. Multiple tests can have the same fully qualified name, making it impossible to actually filter down to what you want (e.g. tests with the fully qualified name `OVFTest.TestEntryPoint` are present 9 times in `jit64_5` which is what actually sparked this issue)

Can we just match the input string on both the fully qualified name and the display name?

cc @jkoritzinsky @markples

jkoritzinsky:
I'm okay adding DisplayName support, but I want to match the dotnet test/xunit syntax as much as possible.

markples:
Just to be clear, do you mean like in https://learn.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests?pivots=xunit? Therefore, the minimal support here would be `DisplayName~<name>`.

The display name that we are using is specific to the dotnet repo.  If I recall correctly, it was to minimize test name churn when converting the tests to merged groups (or more specifically to xunit-style at all).  Do you plan to remove this at some point to get rid of the special case?  I guess perhaps that should be after figuring out how to merge the building of tests too, at which point these tests would fail to build and need to be updated with unique names.

jkoritzinsky:
> Just to be clear, do you mean like in https://learn.microsoft.com/en-us/dotnet/core/testing/selective-unit-tests?pivots=xunit? Therefore, the minimal support here would be `DisplayName~<name>`.

Yep that's what I'm thinking of. The TestFilter type already has support for it, but I never implemented the parsing.

> 
> The display name that we are using is specific to the dotnet repo.  If I recall correctly, it was to minimize test name churn when converting the tests to merged groups (or more specifically to xunit-style at all).  Do you plan to remove this at some point to get rid of the special case?  I guess perhaps that should be after figuring out how to merge the building of tests too, at which point these tests would fail to build and need to be updated with unique names.

Yes, I hope eventually we can get rid of issues.targets, the compat DisplayName generation, and merge the test builds for any non-out-of-proc tests. Once we do that work, there's even further simplifications we can do in our test infra to make it significantly easier to maintain.

[...]

Issue #90682, opened 2023-08-16
[Quality Week Design Proposal] Clean up structure of artifacts/log


trylek:
**ETA**: 1 work-week for design and initial PR switching over parts of CoreCLR build and test execution to the new model

Today, ```artifacts/log``` is a mess combining build and execution logs without clear indication of what and when was running; among others this means that running a limited test run erroneously ingests the results from a potential prior broader run, sometimes even leading to duplication errors in the Python script.

This proposal is to create (possibly gradually as many scripts will need modifying) a cleaner log structure that lets developers reason about their individual builds and runs. My initial thinking is to modify log paths according to the following examples (feel free to suggest any modifications to this scheme):

* ```artifacts/log/windows.x64.Release/230302-1907.build/(build logs from 19:07 on 3/2 2023)```
* ```artifacts/log/Linux.arm64.Debug/230302-1909.run/(run logs from 19:09 on 3/2 2023)```

The nice bit is that you can use filename vs. extension sorting in the file explorer to organize the logs by builds vs. runs vs. a combo of everything ordered by timestamps. 

We can consider having an extra level corresponding to runtime flavor i.e. ```artifacts/log/coreclr``` vs. ```artifacts/log/mono```.

We might want to have something like "latest build / latest run", either as a copy of the respective log directory or as a text file containing the name of the "latest" folder.

/cc @dotnet/runtime-infrastructure @janvorli @jkotas @agocke @jkoritzinsky @vihofer @hoyosjs @stephentoub

jkoritzinsky:
Would it be interesting to do this for the Arcade-managed binlogs as well, or just for our one-off text-based logs?

trylek:
Thanks Jeremy, of course it would be great to clean up all logs. @SamMonoRT has also just reminded me that there's an equivalent mess in the pipeline logs i.o.w. in naming of the artifacts in yml files, that would be great to clean up too but it remains to be seen how much work can fit in the quality week timeframe. I think the main thing is to agree on a structure that annoys the smallest number of people, the actual work can be incremental and take a longer period of time as schedule allows considering this is just a "nice-to-have" without any blocking implications.

jkotas:
Are there build environments out there that try to sort build and test logs into time-stamped directories?

Would this mean that developers iterating on their changes and building repeatedly are going to end up with many sets of logs?

trylek:
[...]

PR #95237, opened 2023-11-26
Update coreclr test run scripts

BruceForstall:
1. Add a run.py `-logs_dir` argument (run.cmd: `logsdir`; run.sh: `-logsDir`) to allow specifying a particular logs directory name. By default, use "logs", as before. This can be used to help preserve per-test-run output (note, however, that currently many per-test-run assets litter the tests artifacts directories). This can be useful if you have multiple test run results logs directory and you want to use `--analyze_results_only` (run.cmd: `printlastresultsonly`) to analyze a run that is not the default logs directory.
2. Remove the unnecessary run.py `--skip_test_run` argument. It was added to use the logic to precompile the libraries, but that logic is now gone.
3. Move generated repro files for failed tests to the "log\repro" directory, and name them with a `repro_` filename prefix.
4. Fix repro file generation. Currently, it gives up on creating repro files for merged tests (it doesn't know how to generate a repro file). Repro files now work for `BuildAsStandalone=true` test builds.
5. Print out the path of all repro files that are created.
6. Add a `setlocal` to the generated Windows repro files.
7. Misc. minor improvements.

Contributes to #95032

### DIFF
--- a/src/tests/run.cmd
+++ b/src/tests/run.cmd
@@ -230,7 +230,7 @@ echo printlastresultsonly      - Print the last test results without running tes
 echo runincontext              - Run each tests in an unloadable AssemblyLoadContext
 echo timeout ^<n^>               - Sets the per-test timeout in milliseconds ^(default is 10 minutes = 10 * 60 * 1000 = 600000^).
 echo                             Note: some options override this ^(gcstresslevel, longgc, gcsimulator^).
-echo logsdir ^<dir^>             - Specify the logs directory ^(default: a new unique directory in the series: log, log.1, log.2, etc.^)
+echo logsdir ^<dir^>             - Specify the logs directory ^(default: artifacts/log^)
 echo msbuildargs ^<args...^>     - Pass all subsequent args directly to msbuild invocations.
 echo ^<CORE_ROOT^>               - Path to the runtime to test ^(if specified^).
 echo.
--- a/src/tests/run.cmd
+++ b/src/tests/run.cmd
@@ -18,8 +18,6 @@ for %%i in ("%__RepoRootDir%") do set "__RepoRootDir=%%~fi"
 if %__ProjectDir:~-1%==\ set "__ProjectDir=%__ProjectDir:~0,-1%"
 set "__ProjectFilesDir=%__ProjectDir%"
 set "__RootBinDir=%__RepoRootDir%\artifacts"
-set "__LogsDir=%__RootBinDir%\log"
-set "__MsbuildDebugLogsDir=%__LogsDir%\MsbuildDebugLogs"
 set __ToolsDir=%__ProjectDir%\..\Tools
 set "DotNetCli=%__RepoRootDir%\dotnet.cmd"
 
@@ -30,6 +28,7 @@ set __LongGCTests=
 set __GCSimulatorTests=
 set __IlasmRoundTrip=
 set __PrintLastResultsOnly=
+set LogsDirArg=
 set RunInUnloadableContext=
 set TieringTest=
 
@@ -60,6 +59,7 @@ if /i "%1" == "jitminopts"                              (set DOTNET_JITMinOpts=1
 if /i "%1" == "jitforcerelocs"                          (set DOTNET_ForceRelocs=1&shift&goto Arg_Loop)
 
 if /i "%1" == "printlastresultsonly"                    (set __PrintLastResultsOnly=1&shift&goto Arg_Loop)
+if /i "%1" == "logsdir"                                 (set LogsDirArg=%2&shift&shift&goto Arg_Loop)
 if /i "%1" == "runcrossgen2tests"                       (set RunCrossGen2=1&shift&goto Arg_Loop)
 REM This test feature is currently intentionally undocumented
 if /i "%1" == "runlargeversionbubblecrossgen2tests"     (set RunCrossGen2=1&set CrossgenLargeVersionBubble=1&shift&goto Arg_Loop)
@@ -108,6 +108,10 @@ if not defined XunitTestReportDirBase set  XunitTestReportDirBase=%XunitTestBinB
 REM Set up arguments to call run.py
 
 set __RuntestPyArgs=-arch %__BuildArch% -build_type %__BuildType%
+
+if defined LogsDirArg (
+    set __RuntestPyArgs=%__RuntestPyArgs% -logs_dir %LogsDirArg%
+)
 
 if defined DoLink (
     set __RuntestPyArgs=%__RuntestPyArgs% --il_link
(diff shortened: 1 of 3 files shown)


BruceForstall:
@trylek @jkoritzinsky @markples I'm trying to fix run.py, especially the ability to automatically create repro scripts for failing tests. It works now for BuildAsStandalone=true, but not for merged tests.

In a merged test, we end up with an xml file like JIT.SIMD.JIT.SIMD.testRun.xml in the log directory (it looks like the testRun.xml file is copied to the log directory, replacing path separators with periods. Aside: couldn't we copy it preserving the same relative path and not munge the filename?), with contents like:
```
<assemblies>

<assembly
    name="JIT.SIMD"
    test-framework="XUnitWrapperGenerator-generated-runner"
    run-date="2023-11-26"
    run-time="22:40:26"
    time="14.6475557"
    total="116"
    passed="116"
    failed="0"
    skipped="0"
    errors="0">

<collection
    name="Collection"
    time="14.6475557"
    total="116"
    passed="116"
    failed="0"
    skipped="0"
    errors="0"
>
<test name="JIT\SIMD\AbsGeneric_r\AbsGeneric_r.dll" type="VectorMathTests.Program" method="TestEntryPoint" time="0.123840" result="Pass"></test>

<test name="JIT\SIMD\AbsGeneric_ro\AbsGeneric_ro.dll" type="VectorMathTests.Program" method="TestEntryPoint" time="0.002934" result="Pass"></test>

<test name="JIT\SIMD\AbsSqrt_r\AbsSqrt_r.dll" type="VectorMathTests.Program" method="TestEntryPoint" time="0.004617" result="Pass"></test>

<test name="JIT\SIMD\AbsSqrt_ro\AbsSqrt_ro.dll" type="VectorMathTests.Program" method="TestEntryPoint" time="0.000910" result="Pass"></test>

<test name="JIT\SIMD\AddingSequence_r\AddingSequence_r.dll" type="VectorMathTests.Program" method="TestEntryPoint" time="0.003844" result="Pass"></test>

<test name="JIT\SIMD\AddingSequence_ro\AddingSequence_ro.dll" type="VectorMathTests.Program" method="TestEntryPoint" time="0.002271" result="Pass"></test>

<test name="_BitwiseOperations_r::VectorMathTests.Program.TestDouble()" type="VectorMathTests.Program" method="TestDouble" time="0.005970" result="Pass"></test>

<test name="_BitwiseOperations_r::VectorMathTests.Program.TestBool()" type="VectorMathTests.Program" method="TestBool" time="0.008456" result="Pass"></test>

<test name="_BitwiseOperations_ro::VectorMathTests.Program.TestDouble()" type="VectorMathTests.Program" method="TestDouble" time="0.003036" result="Pass"></test>

<test name="_BitwiseOperations_ro::VectorMathTests.Program.TestBool()" type="VectorMathTests.Program" method="TestBool" time="0.002500" result="Pass"></test>
...
```

How do we convert this information into a script to run the test, for each test that has `result="Fail"`? It looks like the `<assembly>` "name" is the name of the wrapper assembly (minus the ".dll" extension). So assuming these are all unique, we can just search the tree for it. Then invoke `corerun JIT.SIMD.dll XXX` where `XXX` is the `<test name="XXX"...`? I tried that and it seems to work, but then again, `corerun JIT.SIMD.dll foobar` doesn't fail, and in fact, returns a 100 return code and produces no output or error message, so I'm not confident.

janvorli:
Adding the index to the log directory means that I cannot use the same shell path to search in the generated logs. I always rely on shell history to run stuff from the command line, so I'd have to modify such a search command unless I deleted the log folder before every run. It seems that we could change the naming schema as follows. If there was a log folder already, then rename it to log.{datetime} where the datetime would be the file creation or modification time. Then just create the log folder without any suffix and use it for the new run.

trylek:
@BruceForstall - if you're suggesting basically creating a subtree with the xml results under ```artifacts\log```, that should work even though we'd need to modify the run.py script to collect the xml files across the entire subtree, probably not super hard. Alternatively we could probably somehow encode the relative path to the merged wrapper in some new attribute of the ```assembly``` tag. It might also be possible to generate the repro scripts immediately when building the tests, emitting a single-line file should take a negligible amount of time in comparison with all the msbuild and other machinery that happens under the bonnet. I'm curious whether @jkoritzinsky or @markples might have additional ideas.

For @janvorli's comment, I like the idea of renaming the log to include a timestamp when rerunning the tests; having said that, I suggested something similar in

https://github.com/dotnet/runtime/issues/90682

and several people including JanK expressed concerns towards such a scheme so these would need reaching an agreement on if we decide to proceed in this direction. Other than that I'm not sure what exactly you meant by ```index```, if I understand Bruce's suggestion correctly he basically proposed putting the xml results into subfolders so that their paths relative to ```artifacts\log``` would let him easily identify the relative path of the corresponding merged wrapper under ```artifacts\tests\coreclr\os.arch.config```; please clarify or correct me in case that was just my misunderstanding.

trylek:
@BruceForstall - IIRC it should suffice to run the merged wrapper passing in the test in question as the first command-line argument, @jkoritzinsky should confirm. According to my recollection there should be a minimalistic set of ```if``` statements generated into the merged wrapper to support narrowing its execution to a single test, when Jeremy wrote this logic he was aware of the general ask for minimum overhead originally raised by the JIT team. If that doesn't work or doesn't meet expectations, it should be relatively simple to update the merged wrapper generator to take this into consideration.

I also think you're right that the merged wrapper always returns 100 even when some tests failed, otherwise Helix nukes the work item without uploading any results AFAIK, that will likely need fixing for this execution mode if we want the merged wrapper to return the actual test exit code in this case - in a way, it seems to me that passing in the test name should internally turn the merged wrapper functionality into that of the "simple wrapper" used for out-of-process tests. According to my knowledge Jeremy is back in office now, once we decide on what needs to be done I'll be happy to lend a hand to get this fixed in the next couple of days.

markples:
See #93452 about specifying tests.

Also, just in case, there was a bug in the `.cmd` scripts where passing a test name to the script didn't work.  This was fixed in July, but you (if like me) might have a memory of not being able to get this to work at all.

I believe that having "passing the test name" change the exit behavior may be difficult to reconcile because the argument is a general filter argument.  In theory this could end up being used for various tasks (e.g., use filters instead of compiling directories into multiple merged groups?).  But I assume that the merged test script is called from some other script that could behave differently (or we could add a layer if the "next layer up" is Helix and difficult to change)?

BruceForstall:
@trylek @janvorli I removed the code that created new 'log' directories, so now it uses the same "artifacts/log" directory as before. Mostly, the change now is to fix repro file generation, plus clean up some code (add comments, e.g.). I left the "-logs_dir" arguments since that could potentially still be useful either now or in the future.

jakobbotsch (approved):

PR #95247, opened 2023-11-26
[DataTable] Optimize [NameNode/ConstNode] [=|!=|>|<|>=|<=] [NameNode/ConstNode] filtering

TrayanZapryanov:
This PR optimizes several of most common scenarios when filtering database table(according my knowledge).
Affected method [DataTable.Select(string? filterExpression)](https://github.com/dotnet/runtime/blob/08bef8ee129b9f6d4e7bdc3229b41856b96fd722/src/libraries/System.Data.Common/src/System/Data/DataTable.cs#L4224).
I've added shortcut based on [DataColumn.CompareValueTo](https://github.com/dotnet/runtime/blob/08bef8ee129b9f6d4e7bdc3229b41856b96fd722/src/libraries/System.Data.Common/src/System/Data/DataColumn.cs#L1475) in NameNode and use it in BinaryNode.
Cases:

- [Column] [Operation] [Constant]
- [Constant] [Operation] [Column] 

Operation:

- EqualTo 
- GreaterThen 
- LessThen 
- GreaterOrEqual
- LessOrEqual 
- NotEqual 

This change minimize boxing of stored values in the column.

Benchmark:
I am using table with 1000 rows.
```
[BenchmarkCategory(Categories.Libraries, Categories.Data)]
public class DataTableSelectTests
{
	private DataTable _table;
	private string _ageColumnCondition;

	[GlobalSetup]
	public void Setup()
	{
		_table = new DataTable("test");
		var c = new DataColumn("name");
		_table.Columns.Add(c);
		c = new DataColumn("age");
		c.DataType = typeof(int);
		_table.Columns.Add(c);
		c = new DataColumn("id");
		_table.Columns.Add(c);

		for (var i = 0; i < 1000; i++)
		{
			var row = _table.NewRow();
			row[0] = "human" + i;
			row[1] = i;
			row[2] = i;
			_table.Rows.Add(row);
		}

		_ageColumnCondition = string.Join("OR",
			Enumerable.Range(0, ConditionsCount)
				.Select(cnt => $"(age = {cnt})"));
	}

	[Params(1, 5, 10, 15)]
	public int ConditionsCount = 5;

	[Benchmark]
	public int SelectByAge() => _table.Select(_ageColumnCondition).Length;
}
```

Results:
Before
BenchmarkDotNet v0.13.10-nightly.20231019.90, Windows 11 (10.0.22621.2715/22H2/2022Update/SunValley2)
[...]

PR #95248, opened 2023-11-26
[DataTable] Cache boolean result from BinaryNode.EvalBinary

TrayanZapryanov:
This PR will minimize boxing of bools in [BinaryNode.EvalBinary](https://github.com/dotnet/runtime/blob/08bef8ee129b9f6d4e7bdc3229b41856b96fd722/src/libraries/System.Data.Common/src/System/Data/Filter/BinaryNode.cs#L283) method by caching static object instances for true/false.

I was really amazed that this is not done by compiler and will suggest to built in if possible.
Meanwhile this removed tons of allocations when using [DataTable.Select(string? filterExpression)](https://github.com/dotnet/runtime/blob/08bef8ee129b9f6d4e7bdc3229b41856b96fd722/src/libraries/System.Data.Common/src/System/Data/DataTable.cs#L4224).

Benchmark is the same as in [[DataTable] Optimize [NameNode/ConstNode] [=|!=|>|<|>=|<=] [NameNode/ConstNode] filtering](https://github.com/dotnet/runtime/pull/95247)

Results:
Before
BenchmarkDotNet v0.13.10-nightly.20231019.90, Windows 11 (10.0.22621.2715/22H2/2022Update/SunValley2)
11th Gen Intel Core i9-11900K 3.50GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK 8.0.100
  [Host]     : .NET 8.0.0 (8.0.23.53103), X64 RyuJIT AVX2
  Job-RYTSRU : .NET 8.0.0 (8.0.23.53103), X64 RyuJIT AVX2

PowerPlanMode=00000000-0000-0000-0000-000000000000  Arguments=/p:EnableUnsafeBinaryFormatterSerialization=true  IterationTime=250.0000 ms
MaxIterationCount=20  MinIterationCount=15  WarmupCount=1

| Method      | ConditionsCount | Mean       | Error    | StdDev   | Median     | Min        | Max        | Gen0     | Gen1    | Allocated  |
|------------ |---------------- |-----------:|---------:|---------:|-----------:|-----------:|-----------:|---------:|--------:|-----------:|
| SelectByAge | 1               |   250.1 us |  7.51 us |  8.65 us |   246.9 us |   240.4 us |   269.3 us |  10.5769 |  0.9615 |   86.97 KB |
| SelectByAge | 5               |   560.4 us | 10.32 us |  9.65 us |   561.5 us |   546.5 us |   578.2 us |  45.2586 |  4.3103 |  369.74 KB |
| SelectByAge | 10              |   965.6 us | 19.05 us | 20.38 us |   961.3 us |   943.1 us |   999.6 us |  85.9375 |  7.8125 |  722.18 KB |
| SelectByAge | 15              | 1,361.4 us | 26.62 us | 27.34 us | 1,356.9 us | 1,323.4 us | 1,411.5 us | 130.2083 | 15.6250 | 1073.35 KB |

After:
| Method      | ConditionsCount | Mean       | Error    | StdDev   | Median     | Min        | Max        | Gen0    | Gen1   | Allocated |
|------------ |---------------- |-----------:|---------:|---------:|-----------:|-----------:|-----------:|--------:|-------:|----------:|
| SelectByAge | 1               |   226.8 us |  5.60 us |  6.23 us |   222.8 us |   221.1 us |   239.3 us |  7.4280 |      - |  63.53 KB |
| SelectByAge | 5               |   512.4 us | 13.87 us | 15.97 us |   507.8 us |   492.4 us |   545.6 us | 18.7891 | 2.0877 | 159.03 KB |
| SelectByAge | 10              |   831.0 us | 16.23 us | 17.37 us |   826.3 us |   810.5 us |   870.9 us | 31.0345 | 3.4483 | 277.92 KB |
| SelectByAge | 15              | 1,213.6 us | 26.43 us | 29.38 us | 1,201.5 us | 1,165.1 us | 1,261.4 us | 47.6190 | 4.7619 | 396.13 KB |

I really hope if this kind of caching can be integrated in compiler.

### DIFF
--- a/src/libraries/System.Data.Common/src/System/Data/Filter/BinaryNode.cs
+++ b/src/libraries/System.Data.Common/src/System/Data/Filter/BinaryNode.cs
@@ -12,6 +12,8 @@ namespace System.Data
 {
     internal class BinaryNode : ExpressionNode
     {
+        private static readonly object s_true = true;
+        private static readonly object s_false = false;
         internal int _op;
 
         internal ExpressionNode _left;
@@ -862,37 +864,37 @@ namespace System.Data
                         if ((vLeft == DBNull.Value) || (left.IsSqlColumn && DataStorage.IsObjectSqlNull(vLeft)) ||
                              (vRight == DBNull.Value) || (right.IsSqlColumn && DataStorage.IsObjectSqlNull(vRight)))
                             return DBNull.Value;
-                        return (0 == BinaryCompare(vLeft, vRight, resultType, Operators.EqualTo));
+                        return (0 == BinaryCompare(vLeft, vRight, resultType, Operators.EqualTo)) ? s_true : s_false;
 
                     case Operators.GreaterThen:
                         if ((vLeft == DBNull.Value) || (left.IsSqlColumn && DataStorage.IsObjectSqlNull(vLeft)) ||
                              (vRight == DBNull.Value) || (right.IsSqlColumn && DataStorage.IsObjectSqlNull(vRight)))
                             return DBNull.Value;
-                        return (0 < BinaryCompare(vLeft, vRight, resultType, op));
+                        return (0 < BinaryCompare(vLeft, vRight, resultType, op)) ? s_true : s_false;
 
                     case Operators.LessThen:
                         if ((vLeft == DBNull.Value) || (left.IsSqlColumn && DataStorage.IsObjectSqlNull(vLeft)) ||
                              (vRight == DBNull.Value) || (right.IsSqlColumn && DataStorage.IsObjectSqlNull(vRight)))
                             return DBNull.Value;
-                        return (0 > BinaryCompare(vLeft, vRight, resultType, op));
+                        return (0 > BinaryCompare(vLeft, vRight, resultType, op)) ? s_true : s_false;
 
                     case Operators.GreaterOrEqual:
                         if ((vLeft == DBNull.Value) || (left.IsSqlColumn && DataStorage.IsObjectSqlNull(vLeft)) ||
                              (vRight == DBNull.Value) || (right.IsSqlColumn && DataStorage.IsObjectSqlNull(vRight)))
                             return DBNull.Value;
-                        return (0 <= BinaryCompare(vLeft, vRight, resultType, op));
+                        return (0 <= BinaryCompare(vLeft, vRight, resultType, op)) ? s_true : s_false;
 
                     case Operators.LessOrEqual:
                         if (((vLeft == DBNull.Value) || (left.IsSqlColumn && DataStorage.IsObjectSqlNull(vLeft))) ||
                              ((vRight == DBNull.Value) || (right.IsSqlColumn && DataStorage.IsObjectSqlNull(vRight))))
                             return DBNull.Value;
-                        return (0 >= BinaryCompare(vLeft, vRight, resultType, op));
+                        return (0 >= BinaryCompare(vLeft, vRight, resultType, op)) ? s_true : s_false;
 
                     case Operators.NotEqual:
                         if (((vLeft == DBNull.Value) || (left.IsSqlColumn && DataStorage.IsObjectSqlNull(vLeft))) ||
                              ((vRight == DBNull.Value) || (right.IsSqlColumn && DataStorage.IsObjectSqlNull(vRight))))
                             return DBNull.Value;
-                        return (0 != BinaryCompare(vLeft, vRight, resultType, op));
+                        return (0 != BinaryCompare(vLeft, vRight, resultType, op)) ? s_true : s_false;
 
                     case Operators.Is:
                         vLeft = BinaryNode.Eval(left, row, version, recordNos);
@@ -932,7 +934,7 @@ namespace System.Data
                         {
                             if ((bool)vLeft == false)
                             {
-                                value = false;
+                                value = s_false;
                                 break;
                             }
                         }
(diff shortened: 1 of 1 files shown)


TrayanZapryanov:
Spoiler: once both PRs are in we should receive following results

| Method      | ConditionsCount | Mean     | Error    | StdDev   | Median   | Min      | Max      | Gen0   | Allocated |
|------------ |---------------- |---------:|---------:|---------:|---------:|---------:|---------:|-------:|----------:|
| SelectByAge | 1               | 200.8 us |  3.89 us |  4.48 us | 199.3 us | 195.4 us | 209.8 us | 4.7506 |  40.12 KB |
| SelectByAge | 5               | 327.1 us |  9.51 us | 10.95 us | 321.6 us | 315.8 us | 353.6 us | 3.8710 |   42.1 KB |
| SelectByAge | 10              | 495.3 us |  8.05 us |  7.13 us | 495.0 us | 484.2 us | 509.1 us | 3.8911 |  44.63 KB |
| SelectByAge | 15              | 723.7 us | 14.60 us | 16.81 us | 716.6 us | 700.1 us | 751.3 us | 2.8818 |  47.05 KB |

roji:
@TrayanZapryanov it will likely be a while before I have any time to look at this PR. DataTable in general is de-prioritized (and to a certain extent no longer recommended); it's also not particularly suited for high-performance scenarios. So while I'm sure it's possible to find opportunities to (micro-)optimize it, I'm not sure it's worth the time and the churn.

Having said that, the above does show some non-trivial gains and if you feel it's really worth it, I'd be happy to take a look (but again, probably not anytime soon).

TrayanZapryanov:
@roji Thank you very much for the response. Unfortunately we are using this functionality in our software and everyday millions of operations goes through it. Both PRs looks quite simple and will reduce a lot overhead and make planet more green :). I will wait until you have some time to check, don't worry.

TrayanZapryanov:
@stephentoub I am sorry to bother you, but if I remember correctly there was a cache of bools in Task. FromResult before. And it was removed soon(this year). Unfortunately I have no more details, but see some similarities with this PR. Is it possible somewhere in .Net ecosystem to be done such static caching for boxing of bools to object? Then my PR will not be needed and will solve the problem in general.

stephentoub:
> but if I remember correctly there was a cache of bools in Task. FromResult before. And it was removed soon(this year).

It was not removed. Task.FromResult(bool) will return a cached task. That was added in NET 6 (previously it was used only for async methods).

MichalPetryka:
> static caching for boxing

This would be a huge breaking change and would require an ECMA modification.

TrayanZapryanov:
Sure, but is there any real use case where you want to distinguish two boxed true/false values?

EgorBo:
> Sure, but is there any real use case where you want to distinguish two boxed true/false values?

You can find some discussion here https://github.com/dotnet/runtime/issues/7079

TrayanZapryanov:
@EgorBo thank you for pointing me this interesting discussion. Looks like it will not be changed in the .NET, but do you see any value in current PR to cache bools? In this place boxing scales by number of rows and number of conditions, so easily can reach 50k objects per operation. Another question is it valuable to have analyser for boxing bools(maybe also several other values like 0, DateTime.Min/max.....) which proposes static fields as optimization?

EgorBo:
> but do you see any value in current PR to cache bools?

The optimization itself is not uncommon, I've seen it in a few internal and public projects, but, presumably, it should be applied only where it indeed saves a lot of work for GC in a realistic scenario, so I'd defer the decision to @roji who definitely knows System.Data better)

roji:
I definitely see this making sense in scenarios where the cached instances are used internally, and the usages are under control of the application/library (as is the case in e.g. Roslyn). However, this is a BCL API which would be handing out these cached instances to arbitrary consuming user code (please correct me if that's not the case), so at least in theory there's the potential of a breaking change here.

And once again, DataTable/DataSet are definitely not considered high-performance APIs, and are not longer being actively evolved, so I'm not sure this is the right fit here.

TrayanZapryanov:
@roji I understand your concerns about possible breaking change. In my eyes this is totally internal API and result of it is used only in DataExpression(also internal one), but would be good if you check again. You mention already that it is not advisable to use DataTable API, but what is alternative? I know that a lot of desktop software is having grids, that are bound to DataTables and use for free filtering and indexes of it. Sometimes they store 10 to 100000 rows in it and if you apply a simple select over 1 or 2 columns easily can result to 1 million boxing allocations. Decision is your and I will agree with it, just cannot accept that this API is somehow obsolete and should not be touched or modernized.

roji:
As a general rule, loading a huge (unbounded?) number of rows and binding them to a UI component is in itself problematic practice. In scenarios where arbitrary large data is displayed via a UI, some sort of pagination/progressive downloading needs to be implemented, limiting the number of rows fetched from the database (and transferred to the client) at for every page.

When things are done this way, then the extra allocation overhead of the DataTable bools really should become negligible.  And conversely, in a real-world scenario, downloading and binding a huge number of rows should already be so problematic for performance, that the extra bool allocations also shouldn't matter much.

I admit I'm talking theoretically at this point, since I haven't had time to dive into the PR and the DataTable parts it touches. If you still feel optimizing this is worthwhile, I'll try take a look - but it will be a while before I'm able to do so.

SamMonoRT:
Sorry to mention, but DataTable/DataSet is not being actively worked on and we will not be taking changes unless absolutely necessary for maintenance/build purposes only.  Closing this PR.
cc @roji

PR #95249, opened 2023-11-26
JIT: Remove questionable transformation

### LINKED ISSUE #95318: JIT: Assertion failed '!"Unexpected small type in IR"'
```csharp
// Generated by Fuzzlyn v1.6 on 2023-11-26 16:15:10
// Run on X64 Windows
// Seed: 3845092599037652036
// Reduced from 72.7 KiB to 0.2 KiB in 00:01:18
// Hits JIT assert in Release:
// Assertion failed '!"Unexpected small type in IR"' in 'Program:Main(Fuzzlyn.ExecutionServer.IRuntime)' during 'Post-Morph' (IL size 19; hash 0xade6b36b; FullOpts)
// 
//     File: D:\a\_work\1\s\src\coreclr\jit\fgdiagnostic.cpp Line: 3293
// 

public class Program
{
    public static void Main()
    {
        short[] vr5 = new short[]{0};
        vr5[0] = (short)((long)0 | vr5[0]);
    }
}
```

EgorBo:


src/coreclr/jit/optimizer.cpp
```diff
@@ -6029,34 +6029,14 @@ bool Compiler::optNarrowTree(GenTree* tree, var_types srct, var_types dstt, Valu
...
-                            tree->gtFlags &= ~GTF_UNSIGNED;
-                            tree->AsOp()->gtOp2 = nullptr;
-                            tree->gtVNPair      = op1->gtVNPair; // Set to op1's ValueNumber
-                        }
-                        else
-                        {
-                            // oprSize is smaller or there is a signedness mismatch for small types
+                        // oprSize is smaller or there is a signedness mismatch for small types
+                        // Change the CastToType in the GT_CAST node
+                        tree->CastToType() = dstt;
```
jakobbotsch, line 6034: What IR does this create for the example case we were looking at?

EgorBo: It used to be:
```
    [000002] ----G+-----                         *  CAST      long <- int
    [000001] n---G+-----                         \--*  IND       byte  
    [000000] H----+-----                            \--*  CNS_INT(h) long   0x7ffa2213f21a static Fseq[s_13]

to

    [000002] ----G+-----                         *  NOP       byte  
    [000001] n---G+-----                         \--*  IND       byte  
    [000000] H----+-----                            \--*  CNS_INT(h) long   0x7ffa2213f21a static Fseq[s_13]
```

Now it's:
```
    [000002] ----G+-----                         *  CAST      long <- int
    [000001] n---G+-----                         \--*  IND       byte  
    [000000] H----+-----                            \--*  CNS_INT(h) long   0x7ffa22c7f21a static Fseq[s_13]

to

    [000002] ----G+-----                         *  CAST      int <- byte <- int
    [000001] n---G+-----                         \--*  IND       byte  
    [000000] H----+-----                            \--*  CNS_INT(h) long   0x7ffa22c7f21a static Fseq[s_13]

```

jakobbotsch: That's still an illegal transformation in this case since the value was an argument. The new cast does not guarantee widening into the upper 32 bits like the original cast does.

SingleAccretion:
If you are deleting this quirky `NOP` variant, you can also delete code here:

https://github.com/dotnet/runtime/blob/721c445e494a1d85d2d12e7a99ac2cd94bb898fa/src/coreclr/jit/valuenum.cpp#L11371-L11375

and `case GT_NOP` (as well as adjust the comment) here:

https://github.com/dotnet/runtime/blob/721c445e494a1d85d2d12e7a99ac2cd94bb898fa/src/coreclr/jit/rationalize.cpp#L252-L264

src/coreclr/jit/optimizer.cpp
EgorBo: @jakobbotsch are you suggesting to remove the whole transformation?

jakobbotsch: It seems there are two separate questions:
1. Who is invoking `optNarrowTree` in this way that changes semantics of the program? The transformation is questionable to me but it's because the end result is a change change in behavior -- and this PR still has that problem.
2. Can we get rid of the weird passthrough `GT_NOP`? Yes, and this PR probably accomplishes that, but it seems a better way would be to refactor `optNarrowTree` to take a `GenTree**` so that it can just replace the use of the cast with the operand in this situation.

I'm more worried about (1) than (2), though it would also be nice to clean up (2).

jakobbotsch (approved):
There's probably some more stuff to fix around `optNarrowTree`, but this looks good to me to unblock the Fuzzlyn runs.
 
`gtEffectiveVal` can be cleaned up too now if passthrough NOPs no longer exist... but doesn't necessarily have to be in this PR.

PR #95254, opened 2023-11-27
add README.md for System.Console

adamsitnik:


src/libraries/System.Console/README.md
```diff
@@ -0,0 +1,13 @@
+# System.Console
+Contains the source and tests of assembly System.Console which includes a subset of members of the System.Console namespace that allow you to interact with `Console`.
```
Jozkee, line 2: There's no System.Console namespace in the library or am I missing something?
```suggestion
Contains the source and tests of assembly System.Console which includes a subset of members of the System namespace that allow you to interact with the standard input and output of console applications.
```

jozkee (approved):
Otherwise, LGTM. Thanks!

PR #95260, opened 2023-11-27
[iOS][non-icu] HybridGlobalization implement GetSortKey on hybrid mode

mkhamoyan:
Implements https://github.com/dotnet/runtime/issues/94216

cc @SamMonoRT


src/native/libs/System.Globalization.Native/pal_collation.m
```diff
@@ -295,4 +295,49 @@ int32_t GlobalizationNative_EndsWithNative(const uint16_t* localeName, int32_t l
...
+        const char *utf8Bytes = [transformedString UTF8String];
+        NSData *dataToUse = nil;
+        NSUInteger utf8Length = 0;
+        if (utf8Bytes != NULL) {
+            utf8Length = [transformedString lengthOfBytesUsingEncoding:NSUTF8StringEncoding];
+            dataToUse = [NSData dataWithBytes:utf8Bytes length:utf8Length];
+        } else {
+            // Convert the string to UTF-16 representation
+            dataToUse = [transformedString dataUsingEncoding:NSUTF16StringEncoding];
+            utf8Length = ([dataToUse length] / sizeof(uint16_t)) * 2;
```
akoeplinger, line 328: I'm not sure how the else can be happening and assuming UTF16 doesn't seem safe to me.
Also UTF8 is a variable length encoding so could be 1-4bytes, calculating the length based on `sizeof(uint16_t)) * 2` doesn't seem safe to me.

akoeplinger (approved):
One question, looks good otherwise

src/libraries/System.Private.CoreLib/src/System/Globalization/CompareInfo.Icu.cs
```diff
@@ -700,14 +700,36 @@ private unsafe SortKey IcuCreateSortKey(string source, CompareOptions options)
...
                 keyData = new byte[sortKeyLength];
 
                 fixed (byte* pSortKey = keyData)
                 {
-                    if (Interop.Globalization.GetSortKey(_sortHandle, pSource, source.Length, pSortKey, sortKeyLength, options) != sortKeyLength)
+#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS
+                    if (GlobalizationMode.Hybrid)
                     {
-                        throw new ArgumentException(SR.Arg_ExternalException);
+                        if (Interop.Globalization.GetSortKey(_sortHandle, pSource, source.Length, null, 0, options) != sortKeyLength)
```
mdh1418, line 721: Should we be passing in `pSortKey` as `sortKey` here, similar to the else case? And `sortKeyLength` as well? 
Should we be using the `GetSortKeyNative` here?

src/native/libs/System.Globalization.Native/pal_collation.m
```diff
@@ -295,4 +295,49 @@ int32_t GlobalizationNative_EndsWithNative(const uint16_t* localeName, int32_t l
...
+                                             uint8_t* sortKey, int32_t cbSortKeyLength, int32_t options)
+{
+    @autoreleasepool {
+        if (cwStrLength == 0)
+        {
+            if (sortKey == NULL)
+                sortKey = malloc(1);
+            sortKey[0] = '\0';
+            return 1;
+        }
```
mdh1418, line 308: I'm not sure I'm following what this function is supposed to do.

Why do we need to allocate sortKey and return `1` when `cwStrLength` is 0?
Callers of this function passing in `null` for `sortKey` wouldn't necessarily free the memory we are allocating here woud they?

src/native/libs/System.Globalization.Native/pal_collation.m
mkhamoyan: `else` block was added for `System.Collections.Tests` failure. There are cases when `[transformedString UTF8String]` is null for passed string (like this line https://github.com/dotnet/runtime/blob/main/src/libraries/System.Collections/tests/Generic/Dictionary/HashCollisionScenarios/OutOfBoundsRegression.cs#L177)

src/libraries/Common/tests/Tests/System/StringTests.cs
```diff
@@ -3194,7 +3197,7 @@ public static void IndexOf_TurkishI_EnglishUSCulture()
             }
         }
 
-        [ConditionalFact(typeof(PlatformDetection), nameof(PlatformDetection.IsNotInvariantGlobalization))]
+        [ConditionalFact(typeof(PlatformDetection), nameof(PlatformDetection.IsIcuGlobalizationAndNotHybrid))]
```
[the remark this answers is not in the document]
mkhamoyan: As `IsIcuGlobalizationAndNotHybrid => IsIcuGlobalization && IsNotHybridGlobalization` .
Either `IsInvariantGlobalization` will be true, either`IsIcuGlobalization` will be true, both same time can't. With that logic test will not run when `Invariant` or `Hybrid` mode.

src/libraries/Common/tests/Tests/System/StringTests.cs
```diff
@@ -5955,7 +5960,7 @@ public static void ToUpper_TurkishI_EnglishUSCulture(string s, string expected)
                 new KeyValuePair<char, char>('\u0130', '\u0130'),
                 new KeyValuePair<char, char>('\u0131', '\u0131'));
 
-        [ConditionalTheory(typeof(PlatformDetection), nameof(PlatformDetection.IsNotInvariantGlobalization))]
+        [ConditionalTheory(typeof(PlatformDetection), nameof(PlatformDetection.IsIcuGlobalizationAndNotHybrid))]
```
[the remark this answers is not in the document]
mkhamoyan: As `IsIcuGlobalizationAndNotHybrid => IsIcuGlobalization && IsNotHybridGlobalization` .
Either `IsInvariantGlobalization` will be true, either`IsIcuGlobalization` will be true, both same time can't. With that logic test will not run when `Invariant` or `Hybrid` mode.

src/libraries/System.Private.CoreLib/src/System/Globalization/CompareInfo.Icu.cs
mkhamoyan: Good catch, here also should be called `GetSortKeyNative`

src/native/libs/System.Globalization.Native/pal_collation.m
mkhamoyan: When `cwStrLength` is 0 , we want to return for empty string valid answer (terminated empty string), otherwise https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Globalization/CompareInfo.Icu.cs#L831 will consider that there was a failure.

src/libraries/Common/tests/Tests/System/StringTests.cs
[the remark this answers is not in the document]
mdh1418: It still feels like this change will skip the test for some situations. For example, the test was running for `IsHybridGlobalizationOnBrowser` before, but now with `IsNotHybridGlobalization`, it will be skipped. `IsNlsGlobalization` will also now be skipped as it is `IsNotInvariantGlobalization && !IsIcuGlobalization`.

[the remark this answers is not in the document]
mkhamoyan: yes that's true, updated it.

src/native/libs/System.Globalization.Native/pal_collation.m
akoeplinger: hmm that might just happen for strings with invalid surrogate pairs: https://stackoverflow.com/questions/47130742/whats-an-example-of-an-nsstring-for-which-utf8string-returns-null

I wonder if we need to convert to utf8 at all, maybe we could just return the `transformedString` as utf16 to managed?

mkhamoyan: I will try returning `transformedString` as utf16 to managed to see how tests will pass.

src/libraries/System.Private.CoreLib/src/System/Globalization/CompareInfo.Icu.cs
```diff
@@ -823,7 +872,16 @@ private unsafe int IcuGetHashCodeOfString(ReadOnlySpan<char> source, CompareOpti
...
                     fixed (byte* pSortKey = &MemoryMarshal.GetReference(sortKey))
                     {
-                        sortKeyLength = Interop.Globalization.GetSortKey(_sortHandle, pSource, source.Length, pSortKey, sortKey.Length, options);
+#if TARGET_MACCATALYST || TARGET_IOS || TARGET_TVOS
+                        if (GlobalizationMode.Hybrid)
+                        {
+                            sortKeyLength = Interop.Globalization.GetSortKeyNative(m_name, m_name.Length, pSource, source.Length, pSortKey, sortKey.Length, options);
+                        }
+                        else
+    #endif
```
mdh1418, line 881:
```suggestion
#endif
```

src/libraries/System.Runtime/tests/System.Globalization.Tests/CompareInfo/CompareInfoTests.cs
mdh1418, line 1: Were we running this test for HybridGlobalizationOnOSX before this PR? If so, how come adding the implementation for SortKey changes the behavior for some of the comparison tests? Specifically the ones like `PlatformDetection.IsHybridGlobalizationOnOSX ? -1 : 1` suggests that before SortKey was added, we were passing the test, so it feels odd that we would cause a regression with SortKey.

mdh1418 (commented):
And as discussed offline, I think it would be good to note somewhere that this implementation does not construct SortKeys like `ucol_getSortKey` does, and might not adhere to the specifications [specifications of SortKey](https://unicode-org.github.io/icu/userguide/collation/api.html#getsortkey) such as SortKeys from different collators not being comparable and merging sortkeys.

src/native/libs/System.Globalization.Native/pal_collation.m
akoeplinger: If managed is only expecting a byte array or span then I think we don't need to do any conversion to utf8/utf16 at all, we can use something like this:

```objc
NSUInteger transformedStringBytes = [transformedString lengthOfBytesUsingEncoding: NSUTF16StringEncoding];
sortKey = malloc(transformedStringBytes);
NSRange range = NSMakeRange(0, [transformedString length]);
BOOL result = [transformedString getBytes:sortKey maxLength:transformedStringBytes usedLength:NULL encoding:NSUTF16StringEncoding options:0 range:range remainingRange:NULL];

// check result etc
```

mkhamoyan: Thanks, refactored the function.

mkhamoyan:
Failures are not related.

PR #95262, opened 2023-11-27
[browser] Document `DebuggerSupport` for non-debug configurations

maraf:
Describe the need to set `DebuggerSupport` to `true` for debugging configurations other than `Debug`

### DIFF
--- a/src/mono/wasm/features.md
+++ b/src/mono/wasm/features.md
@@ -368,6 +368,14 @@ You can add following elements in your .csproj
 </PropertyGroup>
 ```
 
+### Debugging non-debug configuration
+If you want to debug the application in other than Debug configuration, you need to set MSBuild property `DebuggerSupport` to `true` (globally or with condition for that particular configuration).
+```xml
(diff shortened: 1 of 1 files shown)


src/mono/wasm/features.md
```diff
@@ -368,6 +368,14 @@ You can add following elements in your .csproj
 </PropertyGroup>
 ```
 
+### Debugging non-debug configuration
+If you want to debug the application in other than Debug configuration, you need to set MSBuild property `DebuggerSupport` to `true` (globally or with condition for that particular configuration).
+```xml
+<PropertyGroup Condition="'$(Configuration)' == '[DESIRED CONFIGURATION NAME]'">
+  <DebuggerSupport>true</DebuggerSupport>
+</PropertyGroup>
+```
```
pavelsavara, line 377: please describe the costs

ilonatommy: Like
```suggestion
```
Debugging in `Release` mode can affect the behavior of your code due to optimizations, making it harder to trace the execution flow, observe variable values, or set breakpoints.
```
?

maraf:
After offline discussion, we are going in other direction - reverting to the previous behavior

PR #95266, opened 2023-11-27
[mono][debugger] If debugger is disconnected but vm is not exited restart the debugger thread

### LINKED ISSUE #95256: mono debug component not able to re-connect after disconnect
### Description

Recently some rider user in our company reported to us that rider's mono remote debug tool is not able to re-connect to runtime debug component. 

When looking into the component log, we found that rider will send a CMD_VM + CMD_VM_DISPOSE command set when disconnecting, while our custom visual studio debugger not (which is working fine).

According to pull request [#67272](https://github.com/dotnet/runtime/pull/67272/files) [debugger-agent.c](https://github.com/thaystg/runtime/blob/09a60af4557c79675708de9d20e5f3510df06f3c/src/mono/mono/component/debugger-agent.c) line 10463, if the debugger send this command set, the process function will reverse the bool check result, making the runtime skip the further debug thread restart code.

Compared to the previous version of this file, it seems the reverse is unnecessary, will you please check it? @thaystg 

### Reproduction Steps

1. Enable mono debug component log with level equals or greater than 1, then start the mono runtime with debug enabled.
2. using rider (rider EAP version is equal)'s mono remote debug tool to connect to the mono runtime's debug component
3. stop the rider debug tool, while keep the mono runtime running.
4. re-start the rider debug tool

### Expected behavior

Expect rider's debug tool re-connect to mono debug component successfully.

### Actual behavior

rider's debug tool is unable to re-connect, debug functions such as breakpoint event are not working as first launch.

### Regression?

_No response_

### Known Workarounds

If we remove the bool reverse we mentioned above and use the modified runtime, rider's debug tool re-connected as expect.

### Configuration

_No response_

### Other information

_No response_

thaystg:
Fixing wrong check.

Fixes #95256

### DIFF
--- a/src/mono/mono/component/debugger-agent.c
+++ b/src/mono/mono/component/debugger-agent.c
@@ -11090,7 +11090,7 @@ bool mono_debugger_agent_receive_and_process_command (void)
 		if (command_set == CMD_SET_VM && (command == CMD_VM_DISPOSE || command == CMD_VM_EXIT))
 			break;
 	}
-	return !(command_set == CMD_SET_VM && command == CMD_VM_DISPOSE);
+	return (command_set == CMD_SET_VM && command == CMD_VM_DISPOSE);
 }
 
 static gboolean


lambdageek:
By the way @thaystg  there's a call to `mono_debugger_agent_receive_and_process_command` on WASI that doesn't check the return value.  Probably should make a separate issue for it, but there's probably an issue there, too.

https://github.com/dotnet/runtime/blob/f38242bd94e4ba67c8b985ffe4d52bd04cc2a36c/src/mono/mono/component/debugger-agent.c#L3790-L3798

lambdageek (approved):

lambdageek:
We should backport

PR #95278, opened 2023-11-27
Fix _AssemblyInTargetingPack value during servicing

ViktorHofer:
The _AssemblyInTargetingPack property in packaging.targets depends on `IsNETCoreAppSrc` or `IsNetCoreAppRef` which aren't defined until src/libraries/Directory.Build.targets is imported.

As packaging.targets is imported first, there's a property sequencing issue.

The fix is to move the `_AssemblyInTargetingPack` logic out of packaging.targets as that's code that is specific to targeting pack libraries which reside under src/libraries.

**This fixes the issue that appeared in the [release/8.0 branch](https://github.com/dotnet/runtime/pull/94882) in main.**

### DIFF
--- a/src/libraries/Directory.Build.targets
+++ b/src/libraries/Directory.Build.targets
@@ -72,6 +72,7 @@
 
   <!-- The assembly version gets updated during servicing when the assembly isn't part of a targeting pack. -->
   <PropertyGroup Condition="'$(PreReleaseVersionLabel)' == 'servicing' and
+                            '$(IsPackable)' == 'true'
                             '$(PackageUseIncrementalServicingVersion)' == 'true'">
     <_IsWindowsDesktopApp Condition="$(WindowsDesktopCoreAppLibrary.Contains('$(AssemblyName);'))">true</_IsWindowsDesktopApp>
     <_IsAspNetCoreApp Condition="$(AspNetCoreAppLibrary.Contains('$(AssemblyName);'))">true</_IsAspNetCoreApp>
--- a/src/libraries/Directory.Build.targets
+++ b/src/libraries/Directory.Build.targets
@@ -72,7 +72,7 @@
 
   <!-- The assembly version gets updated during servicing when the assembly isn't part of a targeting pack. -->
   <PropertyGroup Condition="'$(PreReleaseVersionLabel)' == 'servicing' and
-                            '$(IsPackable)' == 'true'
+                            '$(IsPackable)' == 'true' and
                             '$(PackageUseIncrementalServicingVersion)' == 'true'">
     <_IsWindowsDesktopApp Condition="$(WindowsDesktopCoreAppLibrary.Contains('$(AssemblyName);'))">true</_IsWindowsDesktopApp>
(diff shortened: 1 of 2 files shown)


src/libraries/Directory.Build.targets
```diff
@@ -70,6 +70,15 @@
                                    '$(GeneratePlatformNotSupportedAssemblyMessage)' == ''">true</ILLinkTrimAssembly>
   </PropertyGroup>
 
+  <!-- The assembly version gets updated during servicing when the assembly isn't part of a targeting pack. -->
+  <PropertyGroup Condition="'$(PreReleaseVersionLabel)' == 'servicing' and
+                            '$(PackageUseIncrementalServicingVersion)' == 'true'">
+    <_IsWindowsDesktopApp Condition="$(WindowsDesktopCoreAppLibrary.Contains('$(AssemblyName);'))">true</_IsWindowsDesktopApp>
+    <_IsAspNetCoreApp Condition="$(AspNetCoreAppLibrary.Contains('$(AssemblyName);'))">true</_IsAspNetCoreApp>
+    <_AssemblyInTargetingPack Condition="('$(IsNETCoreAppSrc)' == 'true' or '$(IsNetCoreAppRef)' == 'true' or '$(_IsAspNetCoreApp)' == 'true' or '$(_IsWindowsDesktopApp)' == 'true') and '$(TargetFrameworkIdentifier)' != '.NETFramework'">true</_AssemblyInTargetingPack>
+    <AssemblyVersion Condition="'$(_AssemblyInTargetingPack)' != 'true'">$(MajorVersion).$(MinorVersion).0.$(ServicingVersion)</AssemblyVersion>
```
ericstj, line 79: Will this start applying the change to `ref` projects as well - and have we validated it does so correctly?  I see we previously conditioned inclusion of `packaging.targets` on `IsPackable` which was only set in `src` projects and now this will apply to both `src` and `ref`.  I haven't spotted a problem with that right now, and changing the version in both is the right thing.  Just double check that things are working correctly with `ref` projects.

ViktorHofer: Good point, thanks for catching this. This logic also relies on the `ServicingVersion` property which is defaultedin packaging.targets and only changed in source projects, so only for packable projects. I think it's best to keep the conditions on this that we had previously. Any objection to that?

carlossanlop: Out of curiosity - What kinds of problems will you look for in the ref projects, @ViktorHofer? What things could break with this change?

carlossanlop (approved):
Assuming no problem found with ref projects, lgtm.

src/libraries/Directory.Build.targets
ViktorHofer: With the previous state of this PR, reference source projects would have had the following assembly version: `9.0.0.` during servicing. So the last servicing digit would have been missing because `ServicingVersion` wasn't defined.

ericstj:
LGTM With the `IsPackable` condition this makes this change more of just refactoring and I'd expect it to be a noop for ref projects.

Issue #95273, opened 2023-11-27
[wasi] Single file build fails with AOT


ilonatommy:
This will show up after adding new AOT tests on wasi in https://github.com/dotnet/runtime/pull/95281.

```
        Stack Trace:
          C:\Users\source\repos\runtime-fork\src\mono\wasi\Wasi.Build.Tests\BuildTestBase.cs(445,0): at Wasm.Build.Tests.BuildTestBase.AssertBuild(String args, String label, Boolean expectSuccess, IDictionary`2 envVars, Nullable`1 timeoutMs)
          C:\Users\source\repos\runtime-fork\src\mono\wasi\Wasi.Build.Tests\BuildTestBase.cs(255,0): at Wasm.Build.Tests.BuildTestBase.BuildProject(BuildArgs buildArgs, String id, BuildProjectOptions options)
          C:\Users\source\repos\runtime-fork\src\mono\wasi\Wasi.Build.Tests\BuildThenPublish.cs(50,0): at Wasi.Build.Tests.BuildThenPublishTests.ConsoleBuildThenPublish(String config, Boolean aot, Boolean singleFileBundle)
             at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
             at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)
        Output:
          [] Executing (Captured Output) - C:\Users\source\repos\runtime-fork\artifacts\bin\dotnet-latest\dotnet.exe new wasiconsole  -  in pwd C:\Users\source\repos\runtime-fork\artifacts\bin\Wasi.Build.Tests\Release\net8.0\wasi-wasm\wbt\Debug_dslxjqc1_otc
          [] The template "Wasi Console App" was created successfully.
          [] 
          -------- Building ---------
          Binlog path: C:\Users\source\repos\runtime-fork\artifacts\bin\Wasi.Build.Tests\Release\net8.0\wasi-wasm\xharness-output\logs\Debug_dslxjqc1_otc\Debug_dslxjqc1_otc.binlog
          Building Debug_dslxjqc1_otc in C:\Users\source\repos\runtime-fork\artifacts\bin\Wasi.Build.Tests\Release\net8.0\wasi-wasm\wbt\Debug_dslxjqc1_otc
[...]

Issue #95136, opened 2023-11-22
Stand up AOT tests in ci


lewing:

radical:
We already have Wasi.Build.Tests, and that has *one* AOT test :D More should be added.
For example:
- [ ] aot with single file bundle
- [ ] aot without single file bundle
- [ ] running an AOT'ed app (once that is working)

ilonatommy:
What does "runtime tests" mean?

radical:
> What does "runtime tests" mean?

These are in `src/tests`. Note that we don't run wasm runtime tests with AOT either, yet.

PR #95281, opened 2023-11-27
[wasi] Add AOT option to existing Wasi.Build.Tests

ilonatommy:
Contributes to https://github.com/dotnet/runtime/issues/95136.

AOT + single file bundle fails - https://github.com/dotnet/runtime/issues/95273
When we just publish and run, AOT works fine in both single file bundle and without it.


src/mono/wasi/Wasi.Build.Tests/PublishThenRun.cs
```diff
@@ -0,0 +1,107 @@
...
+
+    private void ConsolePublishThenRunForDifferentBundlings(string config, bool relinking, bool invariantTimezone, bool aot, bool singleFileBundle)
+    {
+        string id = $"{config}_{GetRandomId()}";
+        string projectFile = CreateWasmTemplateProject(id, "wasiconsole");
+        string projectName = Path.GetFileNameWithoutExtension(projectFile);
+        File.Copy(Path.Combine(BuildEnvironment.TestAssetsPath, "SimpleMainWithArgs.cs"), Path.Combine(_projectDir!, "Program.cs"), true);
+
+        string extraProperties = @$"<WasmSingleFileBundle>{singleFileBundle}</WasmSingleFileBundle>
+            <InvariantTimezone>{invariantTimezone}</InvariantTimezone>";
```
radical, line 64: This should be tested in a separate test.

src/mono/wasi/Wasi.Build.Tests/BuildThenRun.cs
```diff
@@ -0,0 +1,109 @@
...
+using Xunit;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+using Wasm.Build.Tests;
+
+#nullable enable
+
+namespace Wasi.Build.Tests;
+
+public class BuildThenRunTests : BuildTestBase
```
radical, line 16: `BuildThenRun`, and `PublishThenRun` will be the most common scenarios anyway, and that does not help differentiate the tests in a meaningful way.
`BuildThenPublish` is different because it is specifically about build behavior when doing a `build` followed by a `publish` because the latter has overlap with the former.

The basic tests can still be build/publish for single-file, and non-single-file bundle cases. The TZ one can be in a new file.

radical (dismissed):

src/mono/wasi/Wasi.Build.Tests/BuildThenRun.cs
```diff
@@ -13,113 +13,90 @@
...
-    [ConditionalTheory(typeof(BuildTestBase), nameof(IsUsingWorkloads))]
-    [MemberData(nameof(TestDataForConsolePublishAndRun))]
-    public void ConsolePublishAndRunForSingleFileBundle(string config, bool relinking, bool invariantTimezone)
+    [Theory]
+    [InlineData("Debug", /*aot*/ false, /*singleFileBundle*/ false)]
+    [InlineData("Debug", /*aot*/ true, /*singleFileBundle*/ false)]
+    [InlineData("Release", /*aot*/ false, /*singleFileBundle*/ false)]
+    [InlineData("Release", /*aot*/ true, /*singleFileBundle*/ false)]
+    [InlineData("Debug", /*aot*/ false, /*singleFileBundle*/ true)]
+    // [InlineData("Debug", /*aot*/ true, /*singleFileBundle*/ true)] // https://github.com/dotnet/runtime/issues/95273
```
radical, line 70: I merged a fix for this today.

src/mono/wasi/Wasi.Build.Tests/BuildThenRun.cs
```diff
@@ -13,113 +13,90 @@
...
 
-        var buildArgs = new BuildArgs(projectName, config, /*aot*/false, id, null);
-        buildArgs = ExpandBuildArgs(buildArgs);
-
-        bool expectRelinking = config == "Release" || relinking;
         BuildProject(buildArgs,
                     id: id,
                     new BuildProjectOptions(
-                        DotnetWasmFromRuntimePack: !expectRelinking,
+                        DotnetWasmFromRuntimePack: true,
```
radical, line 94: This does not look correct. This should be `false` for relinking, and aot. How is it not failing at all?

radical:
This is difficult to review because:
1. it spreads tests over various new files
2. the tests themselves have new changes like in the code, adding new cases
3. and there are new tests

Could you please separate the two into separate PRs?

I think maybe even avoid moving tests around to new files just for avoiding `too many tests in one file` issue. I have a different solution in mind for that. I can discuss that with you, or probably open an issue for it. But this part is optional for the changes here.

radical:
Another test to add - project file has `RunAOTcompilation=true`, but the test only runs `build`, and not `publish`.

src/mono/wasi/Wasi.Build.Tests/BuildThenRun.cs
ilonatommy: Interesting. I merged with main, including your fixing PR, then I rebuilt and tested: fails with the same error.

src/mono/wasi/Wasi.Build.Tests/BuildThenRun.cs
ilonatommy: It does not fail, no matter what value we pass to `DotnetWasmFromRuntimePack`. We use `DotnetWasmFromRuntimePack` only in interpreted version, see:
https://github.com/dotnet/runtime/blob/f66c1c15409ff505ba55f8fdf86bc5c041bf47a3/src/mono/wasm/Wasm.Build.Tests/TestMainJsProjectProvider.cs#L89-L93
Never again we check for its value.

ilonatommy:
> RunAOTcompilation

I am not sure I understand the need to have a test with only build inside. Isn't it covered by e.g. `ConsoleBuildThenRunThenPublish` when we perform the build alone and run? We also publish in the end but if `ConsoleBuildThenRunThenPublish` fails on the build step (pure build, not build from publish) then we will know that build fails, seeing the line from the stack trace.

eng/pipelines/common/evaluate-default-paths.yml
```diff
@@ -228,6 +228,10 @@ jobs:
       - src/mono/wasm/templates/*
       - src/mono/wasm/testassets/*
       - src/mono/wasm/Wasm.Build.Tests/*
+      - src/mono/wasi/build/*
+      - src/mono/wasi/runtime/*
+      - src/mono/wasi/testassets/*
+      - src/mono/wasi/Wasi.Build.Tests/*
       - ${{ parameters._const_paths._wasm_chrome }}
```
radical, line 235: Good catch!!
Add this to `wasm_specific_except_wbt_dbg` also. We will split up the wasm/wasi bits in a future PR, but for now they get triggered together.

src/mono/wasi/Wasi.Build.Tests/BuildThenRun.cs
radical: Hm this should get cleaned up to be same as wasm in an upcoming PR to share more code between wasm-bt and wasi-bt.

ilonatommy:
> I think maybe even avoid moving tests around to new files just for avoiding too many tests in one file issue. I have a different solution in mind for that.

I tested locally and all of them pass. Is this solution small enough to apply in this PR? If not, could you please review in this state and after I apply all the comments, I will move at least `InvariantTimezones/Globalization` tests to a separate file.

```
[EXECUTION TIMED OUT]
Exit Code:-3Executor timed out after 1800 seconds and was killed
['Workloads-Wasi.Build.Tests' END OF WORK ITEM LOG: Command timed out, and was killed]
```

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
```diff
@@ -21,24 +21,126 @@ public WasiTemplateTests(ITestOutputHelper output, SharedBuildPerTestClassFixtur
...
-    [InlineData("Debug", /*aot*/ true)]
-    [InlineData("Release", /*aot*/ false)]
-    [InlineData("Release", /*aot*/ true)]
-    public void ConsoleBuildThenPublish(string config, bool aot)
+    [InlineData("Debug", /*aot*/ false, /*singleFileBundle*/ false)]
+    [InlineData("Debug", /*aot*/ true, /*singleFileBundle*/ false)]
+    [InlineData("Release", /*aot*/ false, /*singleFileBundle*/ false)]
+    [InlineData("Release", /*aot*/ true, /*singleFileBundle*/ false)]
+    [InlineData("Debug", /*aot*/ false, /*singleFileBundle*/ true)]
+    // [InlineData("Debug", /*aot*/ true, /*singleFileBundle*/ true)] // https://github.com/dotnet/runtime/issues/95273
```
radical, line 29: Is this still failing? Could you please re-open the issue?

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
```diff
@@ -21,24 +21,126 @@ public WasiTemplateTests(ITestOutputHelper output, SharedBuildPerTestClassFixtur
...
+        string extraProperties = "";
         if (aot)
-        {
-            AddItemsPropertiesToProject(projectFile, "<RunAOTCompilation>true</RunAOTCompilation><_WasmDevel>false</_WasmDevel>");
-        }
+            extraProperties = "<RunAOTCompilation>true</RunAOTCompilation><_WasmDevel>false</_WasmDevel>";
+        if (singleFileBundle)
+            extraProperties += "<WasmSingleFileBundle>true</WasmSingleFileBundle>";
+        if (!string.IsNullOrEmpty(extraProperties))
+        AddItemsPropertiesToProject(projectFile, extraProperties);
```
radical, line 48:
```suggestion
            AddItemsPropertiesToProject(projectFile, extraProperties);
```

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
```diff
@@ -21,24 +21,126 @@ public WasiTemplateTests(ITestOutputHelper output, SharedBuildPerTestClassFixtur
...
+                    id: id,
+                    new BuildProjectOptions(
+                        DotnetWasmFromRuntimePack: true,
+                        CreateProject: false,
+                        Publish: false,
+                        TargetFramework: BuildTestBase.DefaultTargetFramework));
+
+        // ActiveIssue: https://github.com/dotnet/runtime/issues/82515
+        int expectedExitCode = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? 1 : 42;
+    }
```
radical, line 60: Run the app too.

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
```diff
@@ -21,24 +21,126 @@ public WasiTemplateTests(ITestOutputHelper output, SharedBuildPerTestClassFixtur
...
+
+    [Theory]
+    [InlineData("Debug", /*aot*/ false, /*singleFileBundle*/ false)]
+    [InlineData("Debug", /*aot*/ true, /*singleFileBundle*/ false)]
+    [InlineData("Release", /*aot*/ false, /*singleFileBundle*/ false)]
+    [InlineData("Release", /*aot*/ true, /*singleFileBundle*/ false)]
+    [InlineData("Debug", /*aot*/ false, /*singleFileBundle*/ true)]
+    // [InlineData("Debug", /*aot*/ true, /*singleFileBundle*/ true)] // https://github.com/dotnet/runtime/issues/95273
+    [InlineData("Release", /*aot*/ false, /*singleFileBundle*/ true)]
+    // [InlineData("Release", /*aot*/ true, /*singleFileBundle*/ true)] // https://github.com/dotnet/runtime/issues/95273
```
radical, line 70: Can we make this into a static property/method since it is shared by multiple tests?

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
```diff
@@ -21,24 +21,126 @@ public WasiTemplateTests(ITestOutputHelper output, SharedBuildPerTestClassFixtur
...
+                    id: id,
+                    new BuildProjectOptions(
+                        DotnetWasmFromRuntimePack: true,
+                        CreateProject: false,
+                        Publish: false,
+                        TargetFramework: BuildTestBase.DefaultTargetFramework));
+
+        // ActiveIssue: https://github.com/dotnet/runtime/issues/82515
+        int expectedExitCode = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? 1 : 42;
+
```
radical, line 99: Run the app.

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
```diff
@@ -21,24 +21,126 @@ public WasiTemplateTests(ITestOutputHelper output, SharedBuildPerTestClassFixtur
...
+        bool expectRelinking = config == "Release";
+        BuildProject(buildArgs,
+                    id: id,
+                    new BuildProjectOptions(
+                        DotnetWasmFromRuntimePack: !expectRelinking,
+                        CreateProject: false,
+                        Publish: true,
+                        TargetFramework: BuildTestBase.DefaultTargetFramework,
+                        UseCache: false));
```
radical, line 115: Run the app. But for this particular one, it can be in a follow up PR.

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
```diff
@@ -21,24 +21,126 @@ public WasiTemplateTests(ITestOutputHelper output, SharedBuildPerTestClassFixtur
...
-    public void ConsoleBuildThenPublish(string config, bool aot)
+    [InlineData("Debug", /*aot*/ false, /*singleFileBundle*/ false)]
+    [InlineData("Debug", /*aot*/ true, /*singleFileBundle*/ false)]
+    [InlineData("Release", /*aot*/ false, /*singleFileBundle*/ false)]
+    [InlineData("Release", /*aot*/ true, /*singleFileBundle*/ false)]
+    [InlineData("Debug", /*aot*/ false, /*singleFileBundle*/ true)]
+    // [InlineData("Debug", /*aot*/ true, /*singleFileBundle*/ true)] // https://github.com/dotnet/runtime/issues/95273
+    [InlineData("Release", /*aot*/ false, /*singleFileBundle*/ true)]
+    // [InlineData("Release", /*aot*/ true, /*singleFileBundle*/ true)] // https://github.com/dotnet/runtime/issues/95273
+    public void ConsoleBuild(string config, bool aot, bool singleFileBundle)
```
radical, line 32: This seems redundant since `ConsoleBuildThenPublish` is already being added below.

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
```diff
@@ -21,24 +21,126 @@ public WasiTemplateTests(ITestOutputHelper output, SharedBuildPerTestClassFixtur
...
+    [Theory]
+    [InlineData("Debug", /*aot*/ false, /*singleFileBundle*/ false)]
+    [InlineData("Debug", /*aot*/ true, /*singleFileBundle*/ false)]
+    [InlineData("Release", /*aot*/ false, /*singleFileBundle*/ false)]
+    [InlineData("Release", /*aot*/ true, /*singleFileBundle*/ false)]
+    [InlineData("Debug", /*aot*/ false, /*singleFileBundle*/ true)]
+    // [InlineData("Debug", /*aot*/ true, /*singleFileBundle*/ true)] // https://github.com/dotnet/runtime/issues/95273
+    [InlineData("Release", /*aot*/ false, /*singleFileBundle*/ true)]
+    // [InlineData("Release", /*aot*/ true, /*singleFileBundle*/ true)] // https://github.com/dotnet/runtime/issues/95273
+    public void ConsoleBuildThenRunThenPublish(string config, bool aot, bool singleFileBundle)
```
radical, line 127: This is also same as `ConsoleBuildThenPublish`. Run the app in that test itself.

radical (changes requested):
This is almost done. Thanks for all the effort.

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
ilonatommy: > Another test to add - project file has RunAOTcompilation=true, but the test only runs build, and not publish.

That's this test. Only build, no publish. Now I'm confused

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
ilonatommy: No, it has `Run` while the other did not try to run. If we don't care about having tests of each part separately, then I will leave only `ConsoleBuildThenRunThenPublish` and remove the others.

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
radical: Then you can remove the aot parameter

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
ilonatommy: Removing the whole test, leaving `ConsoleBuildThenRunThenPublish` only.

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
ilonatommy: Failing, reopened

src/mono/wasi/Wasi.Build.Tests/WasiTemplateTests.cs
ilonatommy: Changing to `IEnumerable.Multiply` format

radical (approved):
LGTM! Thank you for the cleanup, and the new tests.

radical:
FYI: I'm working on a build-related PR which will add more tests, and another one that updates wasi.build.tests to use the recent Provider infrastructure. I will rebase my changes to include this one, but till those PRs land I would suggest avoiding the wasi.build.tests .

ilonatommy:
I don't understand why the timeouts still occur. Locally:
Win: `InvariantTests`: 279.047s, `WasiTemplateTests`: 684.603s, 
Codespaces (slowest possible): `InvariantTests`: 347.452s., `WasiTemplateTests`: 884.240s. 
Timeout message: exceeded 1800s.

radical:
The problem seems to be that all the wasi.build.tests are running for each helix item. For `wasm.build.tests` we run one test class per helix item. See `_AddWorkItemsForBuildWasmApps` in https://github.com/dotnet/runtime/blob/2f5cf1c1dfce00a563679e072c50340068ee36d8/src/libraries/sendtohelix-wasm.targets#L287-L323 . The same target in `src/libraries/sendtohelix-wasi.targets` needs the same test splitting based on test class names. The target should be shareable with a few changes.

eng/testing/scenarios/BuildWasiAppsJobsList.txt
```diff
@@ -0,0 +1,2 @@
+Wasi.Build.Tests.Blazor.InvariantTests
+Wasi.Build.Tests.Blazor.WasiTemplateTests
```
radical, line 2:
```suggestion
Wasi.Build.Tests.InvariantTests
Wasi.Build.Tests.WasiTemplateTests
```

src/mono/wasi/Wasi.Build.Tests/BuildTestBase.cs
```diff
@@ -692,8 +692,38 @@ public static int Main()
                     return 42;
                 }
             }";
-    }
 
+
```
radical, line 696:
```suggestion
```

src/mono/wasi/Wasi.Build.Tests/BuildTestBase.cs
```diff
@@ -692,8 +692,38 @@ public static int Main()
...
+            new IEnumerable<object?>[]
+            {
+                new object?[] { "Debug" },
+                new object?[] { "Release" }
+            }
+            .AsEnumerable()
+            .MultiplyWithSingleArgs(true, false) /*propertyValue*/
+            .MultiplyWithSingleArgs(true, false) /*aot*/
+            .UnwrapItemsAsArrays();
+
```
radical, line 707:
```suggestion

```

radical (approved):

PR #95286, opened 2023-11-27
bump-chrome-version.yml: Use `.dotnet/dotnet` if available, else use the system installed one

radical:
…e system installed one.

Fixes:
`/home/runner/work/_temp/a5c062c7-589b-41d2-8f29-8769ac5e9ffd.sh: line 1: .dotnet/dotnet: No such file or directory`

`.dotnet/dotnet` is installed only if version in `global.json` is not already installed system-wide.

### DIFF
--- a/.github/workflows/bump-chrome-version.yml
+++ b/.github/workflows/bump-chrome-version.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Run UpdateToLatestVersion
         run: >-
           make -C src/mono/wasm build-tasks &&
-          .dotnet/dotnet build eng/testing/bump-chrome-version.proj -p:Configuration=Release &&
+          PATH=$PWD/.dotnet:$PATH dotnet build eng/testing/bump-chrome-version.proj -p:Configuration=Release &&
           git add eng/testing/ChromeVersions.props &&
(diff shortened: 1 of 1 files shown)


lewing (approved):

radical:
Test run: https://github.com/dotnet/runtime/actions/runs/7009519217/job/19068145497

.github/workflows/bump-chrome-version.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Run UpdateToLatestVersion
         run: >-
           make -C src/mono/wasm build-tasks &&
-          .dotnet/dotnet build eng/testing/bump-chrome-version.proj -p:Configuration=Release &&
+          PATH=$PWD/.dotnet:$PATH dotnet build eng/testing/bump-chrome-version.proj -p:Configuration=Release &&
```
akoeplinger, line 29: @radical FYI you can use `dotnet.sh` instead which will do the same thing: use the system dotnet if possible, otherwise use the local dotnet

radical: I'll keep this in mind for future!

PR #95291, opened 2023-11-27
Add memory randomization runs

DrewScoggins:
This is adding it to run against all tests, but only in one single configuration. We are doing this to collect data that we can then use to test the visualization system.

### DIFF
--- a/eng/testing/performance/performance-setup.ps1
+++ b/eng/testing/performance/performance-setup.ps1
@@ -139,10 +139,6 @@ if ($NoR2R) {
     $SetupArguments = "$SetupArguments --no-r2r"
 }
 
-if ($MemoryRandomization) {
-    $SetupArguments = "$SetupArguments --memory-randomization"
-}
-
 if ($ExperimentName) {
     $SetupArguments = "$SetupArguments --experiment-name '$ExperimentName'"
 }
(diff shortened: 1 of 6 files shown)


eng/testing/performance/performance-setup.ps1
```diff
@@ -136,6 +139,10 @@ if ($NoR2R) {
     $SetupArguments = "$SetupArguments --no-r2r"
 }
 
+if ($MemoryRandomization) {
+    $SetupArguments = "$SetupArguments --memory-randomization"
+}
+
```
cincuranet, line 145: This is not needed.

eng/testing/performance/performance-setup.sh
```diff
@@ -385,8 +385,16 @@ if [[ "$nor2r" == "true" ]]; then
     configurations="$configurations R2RType=nor2r"
 fi
 
+if [[ "$memoryrandomization" == "true" ]]; then
+    configurations="$configurations MemoeryRandomization=true"
+    extra_benchmark_dotnet_arguments="$extra_benchmark_dotnet_arguments --memory-randomization true"
+fi
+
```
cincuranet, line 392: This is not needed.

cincuranet (changes requested):
I guess the changes in YAML files will be fixed, it was only for testing, right?

LoopedBard3 (approved):
Looks good, can't wait to see the results of the experiment. 👍

cincuranet (approved):

PR #95296, opened 2023-11-27
Alternative fix for RuntimeHelpers.GetSubArray AOT compatibility warning 

### LINKED ISSUE #69053: Slice pattern with type pattern generates redundant type test (or fails to perform the test intended)
**Version Used**: C# 11.0, .NET SDK 7.0.302.

**Steps to Reproduce**:

Compile this code:

```cs
object[] list = new string[]{ "alpha", "papa", "omega" };
if (list is ["alpha", .. string[] theRest, "omega"])
{
    Console.WriteLine($"Match! {theRest.Length}");
}
```

Inspect the generated IL. The part that evaluates the slice pattern includes this code:

```cil
call   !!0[] [System.Runtime]System.Runtime.CompilerServices.RuntimeHelpers::GetSubArray<object>(!!0[],
            valuetype [System.Runtime]System.Range)
isinst  string[]
```

I believe that `isinst` instruction is aiming to deal with the fact that although the static type of `list` is `object[]`, the type compatibility rules for arrays mean it might actually be a `string[]` (and in this example, it is). The thinking then appears to be that although that `string[] theRest` doesn't match the static type of `list`, it might be a runtime match.

So this calls `GetSubArray`, and then asks whether the array that came back is really a `string[]`.

Unfortunately, this serves no purpose. `GetSubArray<object>` always returns an `object[]` even when, as in this case, the input was actually a `string[]`. (To put it another way, the compiler seems to assume that `GetSubArray` will preserve the array element type. But that's wrong. In fact `GetSubArray` returns an array where the element type is determined by the type argument.)

So that `isinst string[]` will always fail.

So at best, that test is redundant. At worst, this is a bug: if the intention here is that the pattern should succeed when the runtime type of `list` is `string[]`, this code fails to achieve that. That pattern only matches the array in this example if we change `list`'s declared type to `string[]`. (Since in general, slice patterns reject impossible matches at compile time, it seems more likely that this is indeed a bug, and that this pattern was meant to match in this example.)

**Expected Behavior**: well this depends on what the compiler is trying to achieve. If the actual behaviour of this code (the pattern does not match) is correct, then the compiler should not emit that `isinst string[]` test, because it will always fail. (And perhaps the compiler should be reporting an error, because it does that in most cases where a slice pattern is an impossible match.) But if that pattern is in fact meant to match in this scenario, then the compiler is emitting code that fails to achieve this.

**Actual Behavior**: the compiler emits a type check that always fails.

jkotas:
C# LDM would like to keep the original "virtual" RuntimeHelpers.GetSubArray behavior.

Fixes https://github.com/dotnet/roslyn/issues/69053

### DIFF
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
@@ -29,10 +29,9 @@ namespace System.Runtime.CompilerServices
 
             T[] dest;
 
-            if (typeof(T).IsValueType || typeof(T[]) == array.GetType())
+            if (typeof(T[]) == array.GetType())
             {
-                // We know the type of the array to be exactly T[] or an array variance
-                // compatible value type substitution like int[] <-> uint[].
+                // We know the type of the array to be exactly T[].
 
                 if (length == 0)
                 {
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
@@ -47,7 +47,7 @@ namespace System.Runtime.CompilerServices
                 // an array of the exact same backing type. The cast to T[] will
                 // never fail.
 
-                dest = Unsafe.As<T[]>(Array.CreateInstance(array.GetType().GetElementType()!, length));
+                dest = Unsafe.As<T[]>(Array.CreateInstanceFromArrayType(array.GetType(), length));
             }
 
             // In either case, the newly-allocated array is the exact same type as the
--- a/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
+++ b/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
@@ -27,18 +27,33 @@ namespace System.Runtime.CompilerServices
 
             (int offset, int length) = range.GetOffsetAndLength(array.Length);
 
-            if (length == 0)
+            T[] dest;
+
+            if (typeof(T).IsValueType || typeof(T[]) == array.GetType())
             {
-                return Array.Empty<T>();
+                // We know the type of the array to be exactly T[] or an array variance
+                // compatible value type substitution like int[] <-> uint[].
+
+                if (length == 0)
+                {
+                    return Array.Empty<T>();
+                }
+
+                dest = new T[length];
+            }
+            else
+            {
+                // The array is actually a U[] where U:T. We'll make sure to create
+                // an array of the exact same backing type. The cast to T[] will
+                // never fail.
+
+                dest = Unsafe.As<T[]>(Array.CreateInstance(array.GetType().GetElementType()!, length));
             }
 
-            T[] dest = new T[length];
-
-            // Due to array variance, it's possible that the incoming array is
-            // actually of type U[], where U:T; or that an int[] <-> uint[] or
-            // similar cast has occurred. In any case, since it's always legal
-            // to reinterpret U as T in this scenario (but not necessarily the
-            // other way around), we can use Buffer.Memmove here.
+            // In either case, the newly-allocated array is the exact same type as the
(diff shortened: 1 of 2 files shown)


jkotas:
@jcouv Here is revert of https://github.com/dotnet/roslyn/issues/69053#issuecomment-1828511548 . 

I think we will need to file a breaking notice for this change. This can both:
- Change the behavior of complex pattern matching expression
- Break code that expects the array slicing to return the "non-virtual" type. For example:
```
test(new string[10]);

static object test(object[] a)
{
    var b = a[1..2];
    b[0] = new object();
    return b;
}
```
I will need your help with providing the justification.

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
```diff
@@ -27,18 +27,33 @@ public static T[] GetSubArray<T>(T[] array, Range range)
...
-            T[] dest = new T[length];
+                dest = Unsafe.As<T[]>(Array.CreateInstanceFromArrayType(array.GetType(), length));
+            }
 
-            // Due to array variance, it's possible that the incoming array is
-            // actually of type U[], where U:T; or that an int[] <-> uint[] or
-            // similar cast has occurred. In any case, since it's always legal
-            // to reinterpret U as T in this scenario (but not necessarily the
-            // other way around), we can use Buffer.Memmove here.
+            // In either case, the newly-allocated array is the exact same type as the
```
stephentoub, line 53: > is the exact same type

This isn't the case for the int[] <-> uint[] case, right?  That's not a problem, but maybe something to clarify in the comment?

src/libraries/System.Runtime/tests/System.Runtime.Tests/System/Runtime/CompilerServices/RuntimeHelpersTests.cs
```diff
@@ -361,6 +361,14 @@ public static void ArrayRangeHelperTest()
             Assert.Throws<ArgumentOutOfRangeException>(() => { int [] array = RuntimeHelpers.GetSubArray(a, range); });
         }
 
+        [Fact]
+        public static void ArrayGetSubArrayCoVarianceTest()
```
stephentoub, line 365: For completeness, do we have / want a test for the int[]/uint[] case?

stephentoub (approved):

stephentoub:
> I think we will need to file a breaking notice for this change.

This was introduced in .NET 7, right?  Should we consider backporting it to .NET 8 asap, such that effectively it would only impact folks that were on .NET 7 or early .NET 8 adopters?

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
jkotas: Good point. Does C# LDM want the exact type preserved even for compatible valuetype substitutions?

This prints `uint[]` today:
```
int[] a = new int[10];
uint[] b = (uint[])(object)a;
Console.WriteLine(b[1..2].GetType());
```
Should it print `int[]` instead?

In other words, should the actual type of array slice always match the type of the source array?

jkotas:
> This was introduced in .NET 7, right?

Right.

> Should we consider backporting it to .NET 8 asap, such that effectively it would only impact folks that were on .NET 7 or early .NET 8 adopters?

It is a corner case and fixing it is a potential breaking change. We have lived with the current behavior for 1.5 years. I do not think that backporting this change meets the servicing bar.

src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RuntimeHelpers.cs
jcouv: Yes, I think this should print `int[]` to mimic a virtual call. Tagging @MadsTorgersen to confirm

MadsTorgersen: Ideally, I guess, but this seems like such a corner case that I doubt anyone would notice in practice, let alone be hurt by it. In the C# language case, since a runtime type check will succeed between `int[]` and `uint[]`, the motivating scenario will be back to working again either way.

PR #95300, opened 2023-11-27
Fix OuterLoop attribute handling

jkoritzinsky:
These codepaths would never trigger because the attribute is `OuterLoopAttribute`, not `OuterloopAttribute`. Fix the handling to actually skip the tests in Pri0 builds.


trylek (approved):
LGTM, thanks for fixing!

PR #95302, opened 2023-11-27
[release/8.0-staging] Fix ILLink/ILC hang when tracking too many hoisted local values

sbomer:
Backport of #95041 to release/8.0-staging

## Customer Impact
Fixes a customer-reported hang when using `PublishTrimmed` in an app that happened to have a large method with a lot of `await`s. The same issue is also present with `PublishAot`.

## Testing

Validated fix with a repro testcase.

## Risk

Low. The fix can lead to new warnings in some cases (as seen in the analyzer testcase), but these represent potential correctness issues, can be easily silenced if necessary, and should be very rare (only affect very large methods with particular patterns that cause tracking of a large number of dataflow values where we give up on the tracking).


vitek-karas (approved):
Thanks!

jeffschwMSFT (approved):
approved. we will take for consideration in 8.0.x

PR #95303, opened 2023-11-28
Add PropertyBuilder implementation

buyaa-n:
Add PropertyBuilder implementation for new persist able reflection emit implementation.

### DIFF
--- a/src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/FieldBuilderImpl.cs
+++ b/src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/FieldBuilderImpl.cs
@@ -36,7 +36,6 @@ namespace System.Reflection.Emit
         {
             _typeBuilder.ThrowIfCreated();
             ValidateDefaultValueType(defaultValue, _fieldType);
-
             _defaultValue = defaultValue;
         }
 
--- a/src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/FieldBuilderImpl.cs
+++ b/src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/FieldBuilderImpl.cs
@@ -34,56 +34,70 @@ namespace System.Reflection.Emit
 
         protected override void SetConstantCore(object? defaultValue)
         {
+            _typeBuilder.ThrowIfCreated();
+            ValidateDefaultValueType(defaultValue, _fieldType);
+
+            _defaultValue = defaultValue;
+        }
+
+        internal static void ValidateDefaultValueType(object? defaultValue, Type destinationType)
+        {
             if (defaultValue == null)
             {
                 // nullable value types can hold null value.
-                if (_fieldType.IsValueType && !(_fieldType.IsGenericType && _fieldType.GetGenericTypeDefinition() == typeof(Nullable<>)))
+                if (destinationType.IsValueType && !(destinationType.IsGenericType && destinationType.GetGenericTypeDefinition() == typeof(Nullable<>)))
+                {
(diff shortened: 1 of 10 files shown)


src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/FieldBuilderImpl.cs
```diff
@@ -33,57 +33,71 @@ internal FieldBuilderImpl(TypeBuilderImpl typeBuilder, string fieldName, Type ty
         }
 
         protected override void SetConstantCore(object? defaultValue)
+        {
+            _typeBuilder.ThrowIfCreated();
+            ValidateDefaultValueType(defaultValue, _fieldType);
+
```
steveharter, line 39: nit: remove extra line?

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ModuleBuilderImpl.cs
```diff
@@ -188,6 +190,42 @@ internal void AppendMetadata(MethodBodyStreamEncoder methodBodyEncoder)
             }
         }
 
+        private void WriteProperties(TypeBuilderImpl typeBuilder)
+        {
+            if (typeBuilder._propertyDefinitions.Count > 0)
```
steveharter, line 195: nit: just return here if `Count == 0`? (to avoid the code block below)

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ModuleBuilderImpl.cs
```diff
@@ -188,6 +190,42 @@ internal void AppendMetadata(MethodBodyStreamEncoder methodBodyEncoder)
...
+                    }
+
+                    if (property.SetMethod is MethodBuilderImpl sMb)
+                    {
+                        AddMethodSemantics(propertyHandle, MethodSemanticsAttributes.Setter, sMb._handle);
+                    }
+
+                    if (property._otherMethods != null)
+                    {
+                        foreach (MethodBuilderImpl method in property._otherMethods)
```
steveharter, line 215: Do we have tests for other property methods? IL-only tests?

steveharter (approved):
LGTM; some misc nits\questions.

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ModuleBuilderImpl.cs
buyaa-n: I have only this: https://github.com/dotnet/runtime/blob/dece90dd90b5e9197e9089ddb3b21d07f0c60422/src/libraries/System.Reflection.Emit/tests/PersistableAssemblyBuilder/AssemblySavePropertyBuilderTests.cs#L56-L58

In the generated assembly it was added just like a normal method, was not sure what else to verify

buyaa-n:
The failures unrelated and known

src/libraries/System.Reflection.Emit/src/System/Reflection/Emit/ModuleBuilderImpl.cs
steveharter: I assume this was for ".other" methods which are different than ".get" and ".set" (ECMA allows methods other than get\set)

buyaa-n: Yes, it is the "other" method

Issue #95213, opened 2023-11-24
TimeZoneInfo.StandardName empty when run under ubuntu-latest NET 8, while value present with NET 6


lahma:
### Description

I'm having some trouble wrapping my head around this. We have a test case in Jint that work under NET 6 and full framework, but now when trying to target NET 8 the `TimeZoneInfo.StandardName` is empty. 

I created a simple assert that `Pacific Standard Time` has a name here:

https://github.com/lahma/timezone-repro/blob/main/UnitTest1.cs

Fails on NET 8 but not on NET 6, test project multi-targets both:

https://github.com/lahma/timezone-repro/actions/runs/6983316148/job/19004227812

I would understand that this could be any tzdata/ICU problem, but the same GitHub Actions runner is running both .NET 6 and 8 and thus I would expect that those databases are valid.

/cc @sebastienros 

### Reproduction Steps

https://github.com/lahma/timezone-repro

### Expected behavior

Would print the correct time zone on both NET 6 and NET 8.

### Actual behavior

NET 8 has empty `StandardName`

Clockwork-Muse:
As remarked upon [in the documentation](https://learn.microsoft.com/en-us/dotnet/api/system.timezoneinfo.findsystemtimezonebyid?view=net-8.0#remarks), Linux uses ICU ids.  `Pacific Standard Time` isn't an ICU id - the equivalent is `America/Los_Angeles`

I'm not sure why it worked on the .Net 6 version, if all other aspects were the same.
[Note that there are methods to convert to/from ICU ids](https://learn.microsoft.com/en-us/dotnet/api/system.timezoneinfo.tryconvertwindowsidtoianaid?view=net-8.0) - you should likely switch to using ICU ids, and if you're running on Windows perform the conversion if required.

lahma:
I do understand that there are different ids and some more preferable, what is puzzling here is why the behavior change and I did try to go trough all the breaking changes documentation for NET 8 and couldn't find anything related to this. Successfully resolving with an (non-preferred) id works but somehow it isn't identifiable with a name anymore.

lahma:
There was also fallbacks in place to support different naming schemes (net462, net6.0), it's a bit hard to get them working targeting both full and modern runtime.

https://github.com/sebastienros/jint/blob/473ca00c598cadaabc3259cabe271bc9e1d934f8/Jint.Tests/Runtime/EngineTests.cs#L30-L39

Clockwork-Muse:
[...]

PR #95304, opened 2023-11-28
Correct time zone name creation when generated lazily.

tarekgh:
https://github.com/dotnet/runtime/issues/95213

This addresses the regression introduced in .NET 8.0 https://github.com/dotnet/runtime/pull/88368, where creating a `TimeZoneInfo` object using a Windows Id on Linux/Mac resulted in empty strings for Standard, Daylight, and display names.


mdh1418 (approved):
Looks good to me, Thanks for the catch!

PR #95306, opened 2023-11-28
Convert HELPER_METHOD_FRAME to QCalls (3/N)

jkotas:


src/coreclr/jit/importercalls.cpp
```diff
@@ -5407,9 +5407,10 @@ void Compiler::impCheckForPInvokeCall(
     }
     optNativeCallCount++;
 
-    if (methHnd == nullptr && (opts.jitFlags->IsSet(JitFlags::JIT_FLAG_IL_STUB) || IsTargetAbi(CORINFO_NATIVEAOT_ABI)))
+    if (methHnd == nullptr && (IsTargetAbi(CORINFO_NATIVEAOT_ABI) ||
+                               (opts.jitFlags->IsSet(JitFlags::JIT_FLAG_IL_STUB) && !compIsForInlining())))
```
jkotas, line 5411: Switching InterfaceMarshaler.ClearNative to managed implementation exposed a bug here. The problem was that the unmanaged calli from Marshal.Release got inlined into the finally block. The exception handling system is not able to deal with pinvoke transitions inlined into the finally blocks.

The fix is to give the free pass for CALLI inlining in IL stubs to the main IL stub body. The inlined helpers should get proper legality check.

@jakobbotsch Could you please review this change?

jakobbotsch (approved):

src/coreclr/jit/importercalls.cpp
jakobbotsch: LGTM. I assume we haven't seen this before because we have very limited inline candidates into IL stubs.

jkoritzinsky (approved):

PR #95308, opened 2023-11-28
Make GetKeyedServices(KeyedService.AnyKey) return all services

stephentoub:
I expected GetKeyedServices(KeyedService.AnyKey) to return all services, since AnyKey is documented as "Gets a key that matches any key." This changes the behavior of KeysMatch accordingly.

Feel free to say "this isn't the right answer" if there's a better answer.

### DIFF
--- a/src/libraries/Microsoft.Extensions.DependencyInjection.Specification.Tests/src/KeyedDependencyInjectionSpecificationTests.cs
+++ b/src/libraries/Microsoft.Extensions.DependencyInjection.Specification.Tests/src/KeyedDependencyInjectionSpecificationTests.cs
@@ -101,6 +101,30 @@ namespace Microsoft.Extensions.DependencyInjection.Specification
         }
 
         [Fact]
+        public void ResolveKeyedServicesAnyKey()
+        {
+            var service1 = new Service();
+            var service2 = new Service();
+            var service3 = new Service();
+            var service4 = new Service();
+            var service5 = new Service();
+            var service6 = new Service();
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddKeyedSingleton<IService>("first-service", service1);
+            serviceCollection.AddKeyedSingleton<IService>("service", service2);
+            serviceCollection.AddKeyedSingleton<IService>("service", service3);
+            serviceCollection.AddKeyedSingleton<IService>("service", service4);
+            serviceCollection.AddKeyedSingleton<IService>(null, service5);
+            serviceCollection.AddSingleton<IService>(service6);
+
+            var provider = CreateServiceProvider(serviceCollection);
+
+            var allServices = provider.GetKeyedServices<IService>(KeyedService.AnyKey).ToList();
+            Assert.Equal(6, allServices.Count);
+            Assert.Equal(new[] { service1, service2, service3, service4, service5, service6 }, allServices);
+        }
+
+        [Fact]
         public void ResolveKeyedGenericServices()
         {
             var service1 = new FakeService();
--- a/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceLookup/CallSiteFactory.cs
+++ b/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceLookup/CallSiteFactory.cs
@@ -686,17 +686,27 @@ namespace Microsoft.Extensions.DependencyInjection.ServiceLookup
         }
 
         /// <summary>
-        /// Returns true if both keys are null or equals, or if key1 is KeyedService.AnyKey and key2 is not null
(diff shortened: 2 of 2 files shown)


skyoxZ:
Is there any docs/comments that let users learn that a magic value `KeyedService.AnyKey` can be used on relative Add/Get methods?

benjaminpetit:
We need first to decide what to do when there is a `KeyedService.AnyKey` service registration. Currently with your PR, it does this:

```csharp
var service1 = new Service();
var service2 = new Service();
var service3 = new Service();
var service4 = new Service();
var service5 = new Service();
var service6 = new Service();
var anyService = new Service();
var serviceCollection = new ServiceCollection();
serviceCollection.AddKeyedSingleton<IService>(KeyedService.AnyKey, anyService);
serviceCollection.AddKeyedSingleton<IService>("first-service", service1);
serviceCollection.AddKeyedSingleton<IService>("service", service2);
serviceCollection.AddKeyedSingleton<IService>("service", service3);
serviceCollection.AddKeyedSingleton<IService>("service", service4);
serviceCollection.AddKeyedSingleton<IService>(null, service5);
serviceCollection.AddSingleton<IService>(service6);

var provider = CreateServiceProvider(serviceCollection);

var allServices = provider.GetKeyedServices<IService>(KeyedService.AnyKey).ToList(); // returns only anyService
```

I don't think that's what we want?

stephentoub:
> We need first to decide what to do when there is a KeyedService.AnyKey service registration

I still don't fully understand when I would want one of those and so don't have a good answer for what it should mean when there is one.

benjaminpetit:
Pinging @davidfowl 

For example, you could do something like this:

``` csharp
serviceCollection.AddKeyedTransient<ILogger>(KeyedService.AnyKey, (sp, key) =>
{
    var factory = sp.GetRequiredService<ILoggerFactory>();
    return factory.CreateLogger(key.ToString());
});
```

And then in your class:

```csharp
public class Service()
{
    public Service([FromKeyedServices("SomeCategory") ILogger logger]) { }
}
```

davidfowl:
> then calling GetKeyedService<IService>("a") will return anyService, I also don't think it's correct.

That is fine if AnyKey means *any key* including the anykey itself no?

stephentoub:
> plus, I don't think GetKeyedServices<IService>(KeyedService.AnyKey) should return non-keyed services too.

Why? It returns them today if you pass null as the key. null is valid in the domain of keys, so I don't know why it would be considered special here.

Taking a step back, @benjaminpetit, we need a backportable solution to enumerating all of the instances. I don't partivularly care whether it's `GetServices<T>()` or `GetKeyedServices<T>(AnyKey)`, and I don't have strong views on what the behavior should be when AnyKey is used during registration. Can you help drive to a solution? If this PR isn't the answer, happy to close it, or if there's a small change you'd like me to make, happy so that, too. We just need a solution.

benjaminpetit:
> That is fine if AnyKey means any key including the anykey itself no?

Maybe? After some thoughts, ss long as we document it, I think both solutions are fine.

> Why? It returns them today if you pass null as the key, e.g. this prints 3:

Because `collection.AddKeyedSingleton<IService, Service>(null)`  and ` collection.AddSingleton<IService, Service>()`  are considered equivalent, same thing for `collection.GetKeyedService<IMyService>(null)`  and `collection.GetService<IMyService>()`.

A service with a `null` key is equivalent as a non-keyed service.

IMO the correct behavior would be: `GetKeyedServices<T>(KeyedService.AnyKey)` shoud return all the **explicitely** registered services, not the one that were "created" with `AnyKey`:

```csharp
serviceCollection.AddKeyedSingleton<IMyService, MyService>(KeyedService.AnyKey);
serviceCollection.AddKeyedSingleton<IMyService, MyOtherService>("another_key");
 
var serviceProvider = serviceCollection.BuildServiceProvider();

// Will return two services, one with key "KeyedService.AnyKey", the other with "another_key"?
var services1 = serviceProvider.GetKeyedServices<IMyService>(KeyedService.AnyKey);
 
// Returns "MyService" -- thanks to the KeyedService.AnyKey registration
_ = serviceProvider.GetKeyedService<IMyService>("something_else");
 
// Returns two services, same as service1
var services2 = serviceProvider.GetKeyedServices<IMyService>(KeyedService.AnyKey);
 
// What should it return? To be consistent with what was returned in services2, nothing?
var services3 = serviceProvider.GetKeyedServices<IMyService>("plop");
```

What we choose to do for `services2` and `services3` will really impact this PR.

And maybe we should add another method like this (probaby for 9?):

```csharp
IEnumerable<(object key, T service)> GetKeyedServices<T>();
```

I imagine that some people would be interested to get the key too.

stephentoub:
Closing in favor of https://github.com/dotnet/runtime/pull/95582

Issue #92529, opened 2023-09-23
[API Proposal]: Add source-generator for enum type


HakamFostok:
### Background and motivation

Andrew Lock (Microsoft MVP) created [NetEscapades.EnumGenerators](https://github.com/andrewlock/NetEscapades.EnumGenerators) library
Basically, it is a source generator that emits static methods for a specific enum that is equivalent to the methods provided by the [`Enum`](https://learn.microsoft.com/en-us/dotnet/api/system.enum) static class.
But the catch here is those source-generated methods do not use reflection so their performance is much faster.
here is a benchmark from [Andrew Lock's article](https://andrewlock.net/netescapades-enumgenerators-a-source-generator-for-enum-performance/)

1. Benchmark for standard `ToString` method vs source-generated `ToString` method
![image](https://github.com/dotnet/runtime/assets/12747172/60b0a359-6653-455f-a1a9-2fd1feb2c3b1)

2. Benchmark for `Enum.IsDefined` method which uses reflection vs source-generated `IsDefined` method
![image](https://github.com/dotnet/runtime/assets/12747172/453a87bb-153e-46a4-ab01-bbb11ea8ceaa)

Those benchmarks could excite only @stephentoub 

I know showing only the benchmark is not enough and the code is needed, but all the details are in the article mentioned above.

I hope, one day, those source-generators can find their ways to be first-class citizens in .NET.
Thank you very much for all your efforts to make .NET the best .NET possible.

### API Proposal

```csharp
namespace System;

[AttributeUsage(AttributeTargets.Enum)]
public sealed class EnumGenerator : Attribute
{
}
```

### API Usage

```csharp
[EnumGenerator]
public enum SomeEnum 
{
   value1,
   value2
}
```

EgorBo:
This idea was discussed here: https://github.com/dotnet/runtime/issues/57748#issuecomment-971564143

HakamFostok:
Hi @EgorBo, I checked the issue you mentioned, actually, it is one part of this suggestion, it is talking only about `ToString`, the library I mentioned is about all the static methods that are defined in the `Enum` class.
for example `IsDefined`, `GetValues`, `GetNames` ext. 
so the scope is a little bit broader here, but I see there, maybe there is a trend to keep .NET as base, and encouragement to do those things as Nuget packages (which, I understand, is fine and for the general good)
I just wanted to suggest that, thank you very much

Tornhoof:
[...]

Issue #57748, opened 2021-08-19
Enum.ToString is slower than explicit implementation


deeprobin:
### Description

Since [Enum.ToString](https://github.com/dotnet/runtime/blob/895c99cc502913fd7393b95af7cae5b9e6fbdea3/src/libraries/System.Private.CoreLib/src/System/Enum.cs#L1136-L1146) usually accesses [Enum.GetEnumName](https://github.com/dotnet/runtime/blob/895c99cc502913fd7393b95af7cae5b9e6fbdea3/src/libraries/System.Private.CoreLib/src/System/Enum.cs#L124-L133), which accesses it in the runtime.

https://github.com/dotnet/runtime/blob/895c99cc502913fd7393b95af7cae5b9e6fbdea3/src/libraries/System.Private.CoreLib/src/System/Enum.cs#L1136-L1146

https://github.com/dotnet/runtime/blob/895c99cc502913fd7393b95af7cae5b9e6fbdea3/src/libraries/System.Private.CoreLib/src/System/Enum.cs#L135-L147

https://github.com/dotnet/runtime/blob/895c99cc502913fd7393b95af7cae5b9e6fbdea3/src/libraries/System.Private.CoreLib/src/System/Enum.cs#L124-L133

However, since the name is already given at compile time I think this should be treated as a constant and assembled at compile time.

### Configuration

```
BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19043.1165 (21H1/May2021Update)
AMD Ryzen 5 3600, 1 CPU, 12 logical and 6 physical cores
.NET SDK=5.0.102
  [Host]     : .NET 5.0.2 (5.0.220.61120), X64 RyuJIT
  DefaultJob : .NET 5.0.2 (5.0.220.61120), X64 RyuJIT
```

#### Benchmarks

|                         Method |      Mean |     Error |    StdDev | Code Size |  Gen 0 | Allocated |
|------------------------------- |----------:|----------:|----------:|----------:|-------:|----------:|
|                RegularToString | 35.004 ns | 0.5859 ns | 0.5481 ns |     104 B | 0.0029 |      24 B |
|                    EnumGetName | 72.600 ns | 0.8403 ns | 0.7860 ns |      68 B | 0.0029 |      24 B |
| ExplicitToStringImplementation |  1.764 ns | 0.0613 ns | 0.0573 ns |     209 B |      - |         - |

##### References
[...]

PR #95311, opened 2023-11-28
Update Microsoft.Extensions.Primitives/ThrowHelper.cs to use fast enum lookup

Meir017:
similar to the logic in https://github.com/dotnet/runtime/blob/main/src/libraries/System.Linq/src/System/Linq/ThrowHelper.cs


EgorBo:
Is this for some NativeAOT workload? Should we discuss a SG-based ToString() instead? (see https://github.com/dotnet/runtime/issues/92529 and https://github.com/dotnet/runtime/issues/57748#issuecomment-971564143)

tarekgh:
I am closing this PR till we finalize the discussion as @EgorBo pointed out.

PR #95317, opened 2023-11-28
[RISC-V] Replace rsGetRsvdReg with ordinary temp registers

tomeksowi:
Replace rsGetRsvdReg() calls with temporary registers assigned in LSRA in some places.

Note: getting rid completely of rsGetRsvdReg() is possible but difficult due to many uses in the emitter and the conditions for assigning temps would be intricate and hard to distill in LSRA so it's deferred for later.

Part of #84834
cc @wscho77 @HJLeee @clamp03 @JongHeonChoi @t-mustafin @gbalykov @viewizard @ashaurtaev @sirntar @yurai007 @bajtazar @brucehoult

### DIFF
--- a/src/coreclr/jit/codegen.h
+++ b/src/coreclr/jit/codegen.h
@@ -807,12 +807,7 @@ protected:
                       regNumber targetReg,
                       regNumber baseReg,
                       regNumber indexReg,
-                      int       scale
-#ifdef TARGET_RISCV64
-                      ,
-                      regNumber scaleTempReg
-#endif
-                      );
+                      int scale RISCV64_ARG(regNumber scaleTempReg));
 #endif // TARGET_ARMARCH || TARGET_LOONGARCH64 || TARGET_RISCV64
 
 #if defined(TARGET_ARMARCH)
--- a/src/coreclr/jit/codegen.h
+++ b/src/coreclr/jit/codegen.h
@@ -803,11 +803,16 @@ protected:
     void genSetRegToCond(regNumber dstReg, GenTree* tree);
 
 #if defined(TARGET_ARMARCH) || defined(TARGET_LOONGARCH64) || defined(TARGET_RISCV64)
-    void genScaledAdd(emitAttr attr, regNumber targetReg, regNumber baseReg, regNumber indexReg, int scale
-    #ifdef TARGET_RISCV64
-        , regNumber scaleTempReg
-    #endif
-    );
+    void genScaledAdd(emitAttr  attr,
+                      regNumber targetReg,
+                      regNumber baseReg,
+                      regNumber indexReg,
+                      int       scale
+#ifdef TARGET_RISCV64
+                      ,
+                      regNumber scaleTempReg
+#endif
+                      );
 #endif // TARGET_ARMARCH || TARGET_LOONGARCH64 || TARGET_RISCV64
 
 #if defined(TARGET_ARMARCH)
@@ -824,6 +829,7 @@ protected:
     // Generate the instruction to move a value between register files
     void genBitCast(var_types targetType, regNumber targetReg, var_types srcType, regNumber srcReg);
 
+public:
     struct GenIntCastDesc
     {
         enum CheckKind
@@ -901,6 +907,7 @@ protected:
         }
     };
 
+protected:
     void genIntCastOverflowCheck(GenTreeCast* cast, const GenIntCastDesc& desc, regNumber reg);
     void genIntToIntCast(GenTreeCast* cast);
     void genFloatToFloatCast(GenTree* treeNode);
--- a/src/coreclr/jit/codegen.h
+++ b/src/coreclr/jit/codegen.h
@@ -803,7 +803,11 @@ protected:
     void genSetRegToCond(regNumber dstReg, GenTree* tree);
 
 #if defined(TARGET_ARMARCH) || defined(TARGET_LOONGARCH64) || defined(TARGET_RISCV64)
-    void genScaledAdd(emitAttr attr, regNumber targetReg, regNumber baseReg, regNumber indexReg, int scale);
+    void genScaledAdd(emitAttr attr, regNumber targetReg, regNumber baseReg, regNumber indexReg, int scale
+    #ifdef TARGET_RISCV64
+        , regNumber scaleTempReg
+    #endif
+    );
 #endif // TARGET_ARMARCH || TARGET_LOONGARCH64 || TARGET_RISCV64
 
 #if defined(TARGET_ARMARCH)
--- a/src/coreclr/jit/codegenriscv64.cpp
+++ b/src/coreclr/jit/codegenriscv64.cpp
@@ -3463,13 +3463,14 @@ void CodeGen::genCodeForCompare(GenTreeOp* tree)
             }
             else if (tree->OperIs(GT_EQ))
             {
-                skipLabel = genCreateTempLabel();
+                regNumber tempReg = tree->GetSingleTempReg();
+                skipLabel         = genCreateTempLabel();
                 emit->emitIns_R_R(cmpSize == EA_4BYTE ? INS_fclass_s : INS_fclass_d, cmpSize, targetReg, regOp1);
-                emit->emitIns_R_R(cmpSize == EA_4BYTE ? INS_fclass_s : INS_fclass_d, cmpSize, rsGetRsvdReg(), regOp2);
-                emit->emitIns_R_R_R(INS_or, EA_8BYTE, rsGetRsvdReg(), targetReg, rsGetRsvdReg());
-                emit->emitIns_R_R_I(INS_andi, EA_8BYTE, rsGetRsvdReg(), rsGetRsvdReg(), 0x300);
+                emit->emitIns_R_R(cmpSize == EA_4BYTE ? INS_fclass_s : INS_fclass_d, cmpSize, tempReg, regOp2);
+                emit->emitIns_R_R_R(INS_or, EA_8BYTE, tempReg, targetReg, tempReg);
+                emit->emitIns_R_R_I(INS_andi, EA_8BYTE, tempReg, tempReg, 0x300);
                 emit->emitIns_R_R_I(INS_addi, EA_8BYTE, targetReg, REG_R0, 1);
-                emit->emitIns_J(INS_bnez, skipLabel, rsGetRsvdReg());
+                emit->emitIns_J(INS_bnez, skipLabel, tempReg);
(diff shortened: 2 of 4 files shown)


src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -1762,7 +1759,9 @@ void CodeGen::genLclHeap(GenTree* tree)
         unsigned outgoingArgSpaceAligned = roundUp(compiler->lvaOutgoingArgSpaceSize, STACK_ALIGN);
         // assert((compiler->lvaOutgoingArgSpaceSize % STACK_ALIGN) == 0); // This must be true for the stack to remain
         //                                                                // aligned
-        genInstrWithConstant(INS_addi, EA_PTRSIZE, REG_SPBASE, REG_SPBASE, outgoingArgSpaceAligned, rsGetRsvdReg());
+        if (tempReg == REG_NA)
```
yurai007, line 1763: It looks that at this point tempReg is always REG_NA.

src/coreclr/jit/lsrariscv64.cpp
```diff
@@ -286,10 +286,65 @@ int LinearScan::BuildNode(GenTree* tree)
...
+                {
+                    needTemp = true; // divisorReg
+                }
+                else if (tree->OperIsCommutative())
+                {
+                    if (tree->gtGetOp1()->isContainedIntOrIImmed())
+                        needTemp = true; // reg1
+                }
+
+                if (tree->gtOper == GT_DIV || tree->gtOper == GT_MOD)
```
yurai007, line 312: Suppose that at this point we know we need temp. Does it still make sense to enter this branch?

src/coreclr/jit/lsrariscv64.cpp
```diff
@@ -286,10 +286,65 @@ int LinearScan::BuildNode(GenTree* tree)
...
+                {
+                    bool checkDividend = true;
+                    // Do we have an immediate for the 'divisorOp'?
+                    if (divisorOp->IsCnsIntOrI())
+                    {
+                        ssize_t intConstValue = divisorOp->AsIntCon()->gtIconVal;
+                        if (intConstValue != -1)
```
yurai007, line 319: nit: I have troubles with understanding this part. Maybe adding one extra comment explaining meaning of -1 and impact on next check could help a bit.

src/coreclr/jit/emitriscv64.cpp
```diff
@@ -4316,25 +4314,24 @@ regNumber emitter::emitInsTernary(instruction ins, emitAttr attr, GenTree* dst,
...
-                            codeGen->genJumpToThrowHlpBlk_la(SCK_OVERFLOW, INS_bne, codeGen->rsGetRsvdReg());
+                            codeGen->genJumpToThrowHlpBlk_la(SCK_OVERFLOW, INS_bne, tempReg);
                         }
                         else
                         {
-                            regNumber tmpReg = dst->GetSingleTempReg();
-                            assert(tmpReg != dst->GetRegNum());
-                            assert(tmpReg != src1->GetRegNum());
-                            assert(tmpReg != src2->GetRegNum());
+                            regNumber tempReg2 = dst->ExtractTempReg();
```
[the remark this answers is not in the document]
tomeksowi: It is but tempReg is to replace rsGetRsvdReg and tempReg2 is another temp that was assigned here before (I just renamed it from tmpReg to avoid confusion).

src/coreclr/jit/lsrariscv64.cpp
tomeksowi: > Suppose that at this point we know we need temp

Do we? needTemp is true only when one of the ifs above is entered.

src/coreclr/jit/emitriscv64.cpp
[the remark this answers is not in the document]
yurai007: Sure, but isn't tempReg2 always equal to tempReg? If so we can simply use tempReg and remove tempReg2 completely. I don't think dst is mutated somewhere in-between but maybe I'm missing something.

src/coreclr/jit/lsrariscv64.cpp
yurai007: What I'm trying to say is: I guess that if needTemp is true we don't to enter this block at all.

src/coreclr/jit/lsrariscv64.cpp
tomeksowi: To know whether we need a temp it's enough to mimic the logic from CodeGen::genCodeForDivMod (which also doesn't explain the meaning of -1:) There is some explanation in the condition under which ArithmeticException flag is set, which is GenTree::CanDivOrModPossiblyOverflow.

But basic idea IMO is clear: integer division can overflow only for div by -1, any other other divisor results in a smaller number. And it does overflow if dividend is MinValue because -MinValue > MaxValue.

src/coreclr/jit/lsrariscv64.cpp
tomeksowi: ok, now I get it:) yes, you're right

src/coreclr/jit/emitriscv64.cpp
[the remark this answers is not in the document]
tomeksowi: Both are used in genJumpToThrowHlpBlk_la below so probably not.

src/coreclr/jit/codegenriscv64.cpp
tomeksowi: I think it's ok without an assertion, var declaration isn't that far away.

Bajtazar (approved):

src/coreclr/jit/codegenriscv64.cpp
```diff
@@ -1815,8 +1815,10 @@ void CodeGen::genLclHeap(GenTree* tree)
             }
             else
             {
-                emit->emitLoadImmediate(EA_PTRSIZE, rsGetRsvdReg(), amount);
-                emit->emitIns_R_R_R(INS_sub, EA_PTRSIZE, REG_SPBASE, REG_SPBASE, rsGetRsvdReg());
+                if (tempReg == REG_NA)
+                    tempReg = tree->ExtractTempReg();
```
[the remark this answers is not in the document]
tannergooding: > It doesn't really matter. It's allowed

The inverse is also allowed and it does ultimately matter. The end result comes down to area owner preference and while some people on the team are fine with it, other people on the team request that it not be done in areas they are regularly maintaining.

At the same time, something being allowed doesn't necessarily mean it is a good idea. There is always nuance and cases where it may be objectively better or worse than the alternative.

> and simple code takes less space.

This is absolutely not a definitive truth. Simple code *can* take less space. Equally, you can write complex or hard to read code that also takes less space.

It is frequently the case that adding whitespace, adding braces, adding comments, inserting additional named variables, and more can all make code simpler and easier to read.

Skipping out on brackets have been the root cause of several bugs, CVEs (security issues), and other problems across a multitude of code bases (including large enterprise ones). This can become particularly prevalent as a part of using macros, after doing a refactoring, etc. So while you might be able to save 2 lines by skipping braces, it is always safer to explicitly define them.

yurai007 (approved):

src/coreclr/jit/codegenriscv64.cpp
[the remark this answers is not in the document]
tomeksowi: I was under impression [CLR JIT Coding Conventions](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/clr-jit-coding-conventions.md#12.1) specifically enumerating cases for which braceless ifs are allowed leaves that to the contributor's preference.

But I can of course change it if area owners want braces here anyway.

tomeksowi:
@jakobbotsch PR ready for review

clamp03 (approved):

sirntar (approved):

src/coreclr/jit/codegen.h
```diff
@@ -803,7 +803,16 @@ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
...
 #if defined(TARGET_ARMARCH) || defined(TARGET_LOONGARCH64) || defined(TARGET_RISCV64)
-    void genScaledAdd(emitAttr attr, regNumber targetReg, regNumber baseReg, regNumber indexReg, int scale);
+    void genScaledAdd(emitAttr  attr,
+                      regNumber targetReg,
+                      regNumber baseReg,
+                      regNumber indexReg,
+                      int       scale
+#ifdef TARGET_RISCV64
+                      ,
+                      regNumber scaleTempReg
```
jakobbotsch, line 813: Please introduce something similar to `UNIX_LOONGARCH64_ONLY_ARG`.

src/coreclr/jit/lsrariscv64.cpp
```diff
@@ -242,6 +243,8 @@ int LinearScan::BuildNode(GenTree* tree)
             {
                 // Need a register different from target reg to check for overflow.
                 buildInternalIntRegisterDefForNode(tree);
+                if (!(tree->gtFlags & GTF_UNSIGNED))
```
jakobbotsch, line 246: Nit: new code should check explicitly against zero.

src/coreclr/jit/lsrariscv64.cpp
```diff
@@ -1271,6 +1385,12 @@ int LinearScan::BuildBlockStore(GenTreeBlk* blkNode)
 //
 int LinearScan::BuildCast(GenTreeCast* cast)
 {
+    enum CodeGen::GenIntCastDesc::CheckKind kind = CodeGen::GenIntCastDesc(cast).CheckKind();
+    if (kind != CodeGen::GenIntCastDesc::CHECK_NONE && kind != CodeGen::GenIntCastDesc::CHECK_POSITIVE)
```
jakobbotsch, line 1389:
```suggestion
    if ((kind != CodeGen::GenIntCastDesc::CHECK_NONE) && (kind != CodeGen::GenIntCastDesc::CHECK_POSITIVE))
```

src/coreclr/jit/lsrariscv64.cpp
```diff
@@ -549,6 +655,14 @@ int LinearScan::BuildNode(GenTree* tree)
             }
             assert(dstCount == 1);
 
+            if (base && index)
+            {
+                DWORD scale;
+                BitScanForward(&scale, lea->gtScale);
+                if (scale)
```
jakobbotsch, line 662: Ditto here, and other places.

src/coreclr/jit/codegen.h
tomeksowi: There's already RISCV64_ARG in stdmacros.h, I think I can reuse.

jakobbotsch (approved):

PR #95324, opened 2023-11-28
[browser] Added an optional HybridGlobalization run of libraries

### LINKED ISSUE #95282: [wasm] Create an optional job for HybridGlobalization on CI
Currently we are testing only `System.Globalization` and `System.Globalization.Calendars` with the flag `HybridGlobalization=true`. We would like to leave these tests in the default CI lane and add a opt-in job for running all the libs in the browser with `HybridGlobalization=true` (both AOT and non-AOT).

ilonatommy:
Fixes https://github.com/dotnet/runtime/issues/95282.

Runs on:
~~- rolling build~~ add it in a few weeks in a follow-up PR, when we decide that it is stable
- on request as a part of: `/azp run runtime-wasm-libtests`

We have a few not supported operations, e.g. `GetHashCode` or `Compare` with certain sets of `StringComparison` options. Tests that use them were blocked. Some tests need further investigation, connected issues:
https://github.com/dotnet/runtime/issues/95503
https://github.com/dotnet/runtime/issues/95471
https://github.com/dotnet/runtime/issues/95473
https://github.com/dotnet/runtime/issues/95594
https://github.com/dotnet/runtime/issues/95623


ilonatommy:
> No pipelines are associated with this pull request.

@radical, am I missing something to be able to trigger the template's job?

radical:
> > No pipelines are associated with this pull request.
> 
> @radical, am I missing something to be able to trigger the template's job?

You will need to ask FR to add a corresponding pipeline. Maybe add this to `runtime-wasm-libtests`?

eng/pipelines/extra-platforms/runtime-extra-platforms-wasm.yml
```diff
@@ -179,6 +194,22 @@ jobs:
...
+    parameters:
+      platforms:
+        - browser_wasm
+        - browser_wasm_win
+      nameSuffix: _HybridGlobalization
+      extraHelixArgs: /p:HybridGlobalization=true
+      isExtraPlatformsBuild: ${{ parameters.isExtraPlatformsBuild }}
+      isWasmOnlyBuild: ${{ parameters.isWasmOnlyBuild }}
+      scenarios:
+        - normal
```
radical, line 209: Is running on v8 useful?

eng/pipelines/extra-platforms/runtime-extra-platforms-wasm.yml
```diff
@@ -54,6 +54,21 @@ jobs:
       shouldContinueOnError: true
       alwaysRun: true
 
+  #ToDo: add AOT version when this is confirmed to work
+  #ToDo: confirm if adding it here is fine
+  - template: /eng/pipelines/common/templates/wasm-library-tests.yml
```
radical, line 59: Is this expected to be stable right now?

radical: If not, then we should add only to `runtime-wasm*` for now.

eng/pipelines/extra-platforms/runtime-extra-platforms-wasm.yml
```diff
@@ -179,6 +194,22 @@ jobs:
...
+        - browser_wasm
+        - browser_wasm_win
+      nameSuffix: _HybridGlobalization
+      extraHelixArgs: /p:HybridGlobalization=true
+      isExtraPlatformsBuild: ${{ parameters.isExtraPlatformsBuild }}
+      isWasmOnlyBuild: ${{ parameters.isWasmOnlyBuild }}
+      scenarios:
+        - normal
+        - WasmTestOnBrowser
+        - WasmTestOnNodeJS
```
radical, line 211: We avoid running jobs in `runtime-wasm` if their duplicate for `runtime` would be run. In this case, since the new job will not be running on PRs by default, use `alwaysRun: true` so it is always run on `runtime-wasm`.

eng/pipelines/extra-platforms/runtime-extra-platforms-wasm.yml
radical: > The `runtime-wasm-libtests` are run as part of rolling build?

`runtime-wasm-libtests` is a separate manual-run-only pipeline. Some of the jobs are part of `runtime-extra-platforms` though, and those run for rolling builds.

eng/pipelines/extra-platforms/runtime-extra-platforms-wasm.yml
ilonatommy: It is not a high priority but we expect the tests to pass on each host, on V8 as well, e.g. see in https://github.com/dotnet/runtime/blob/65d38607acb332f2ccd1fbf7aed2e37f659a4d56/src/libraries/System.Runtime/tests/System.Globalization.Tests/CompareInfo/CompareInfoTests.IsPrefix.cs#L82
we detect the differences in behavior between hosts and changed the expected results accordingly.

eng/pipelines/extra-platforms/runtime-extra-platforms-wasm.yml
ilonatommy: "Is this expected to be stable right now?" I only know that for iOS it's not. For browser it might be, we'll see in this PR.

ilonatommy:
Locally, running with node v20.8.0 `WasmTestOnNodeJS-System.Globalization.Hybrid.WASM.Tests` passes. Is it possible that Helix uses different node versions on Windows than on Linux? Window's logs are truncated - there's no info about version there.

ilonatommy:
`HybridGlobalization` does not get propagated to helix, binlog of non-hybrid libs does not contain this property.

ilonatommy:
Needs a re-run when https://github.com/dotnet/dotnet-buildtools-prereqs-docker/pull/937 gets merged.

src/libraries/Common/tests/Tests/System/StringTests.cs
```diff
@@ -4924,6 +4932,7 @@ public static IEnumerable<object[]> StartsWith_StringComparison_TestData()
 
         [Theory]
         [MemberData(nameof(StartsWith_StringComparison_TestData))]
+		[ActiveIssue("https://github.com/dotnet/runtime/issues/95473", typeof(PlatformDetection), nameof(PlatformDetection.IsHybridGlobalizationOnBrowser))]
```
radical, line 4935:
```suggestion
        [ActiveIssue("https://github.com/dotnet/runtime/issues/95473", typeof(PlatformDetection), nameof(PlatformDetection.IsHybridGlobalizationOnBrowser))]
```

src/libraries/tests.proj
```diff
@@ -84,6 +84,12 @@
     <ProjectExclusions Include="$(MSBuildThisFileDirectory)System.Runtime\tests\System.Globalization.Tests\Hybrid\System.Globalization.Hybrid.WASM.Tests.csproj" />
   </ItemGroup>
 
+  <!-- invariant project cannot be tested in hybrid mode; projects using functions that throw PNSE too often to block separate tests (HashCode, CompareOptions) -->
```
radical, line 87: `projects using functions that throw PNSE too often to block separate tests (HashCode, CompareOptions) `

- can you explain this? And is there an issue for it?

radical (approved):
The CI related changes LGTM 👍

src/libraries/tests.proj
ilonatommy: It's not an issue, it's by design, here is the [doc](https://github.com/dotnet/runtime/blob/69bbdb8869b781f1d86434a340a6ba4b391338f0/docs/design/features/globalization-hybrid-mode.md)

ilonatommy: As a follow-up job, I should add the APIs from tests that I blocked in this PR to the doc (into "Affected public APIs" sections).

PR #95127, opened 2023-11-22
Add Arm64 encodings for IF_SVE_AA_3A to IF_SVE_HL_3A

a74nh:
Add the first set of 17 SVE groups.

Contributes towards https://github.com/dotnet/runtime/issues/94549
<strike>
Based on top of https://github.com/dotnet/runtime/pull/95105
Code starts from https://github.com/dotnet/runtime/pull/95127/commits/55731b41720fc6a8c1f127e7e4746e58cb717632
</strike>

PR #95337, opened 2023-11-28
Add Arm64 encodings for IF_SVE_AF_3A to IF_SVE_AQ_3A

a74nh:
Next set of encodings to contribute towards https://github.com/dotnet/runtime/issues/94549


a74nh:
Some of the encodings aren't yet supported by capstone. These are all quadword instructions are are not available in N2. These are untested, so the codegen code has been commented out. I've left in the code emit code for these as it should be mostly correct - but it is the job of whoever enables the codegen code to fully test it.

a74nh:
@kunalspathak @BruceForstall

kunalspathak:
> Next set of encodings to contribute towards #94549

I don't see these 2 encodings...are you planning to add them later?

```
SVE_AB_3B
SVE_HL_3B
```

src/coreclr/jit/emitarm64.cpp
```diff
@@ -18525,6 +18812,14 @@ emitter::insExecutionCharacteristics emitter::getInsExecutionCharacteristics(ins
             }
             break;
 
+        // Not available in Arm Neoverse N2 Software Optimization Guide.
+        case IF_SVE_AG_3A: // ........xx...... ...gggnnnnnddddd -- SVE bitwise logical reduction (quadwords)
+        case IF_SVE_AJ_3A: // ........xx...... ...gggnnnnnddddd -- SVE integer add reduction (quadwords)
+        case IF_SVE_AL_3A: // ........xx...... ...gggnnnnnddddd -- SVE integer min/max reduction (quadwords)
+            result.insLatency    = PERFSCORE_LATENCY_20C;    // Invalid
```
kunalspathak, line 18819: why is this "// Invalid"?

src/coreclr/jit/emitarm64.cpp
```diff
@@ -13941,8 +14132,16 @@ size_t emitter::emitOutputInstr(insGroup* ig, instrDesc* id, BYTE** dp)
         case IF_SVE_AC_3A: // ........xx...... ...gggmmmmmddddd -- SVE integer divide vectors (predicated)
         case IF_SVE_AD_3A: // ........xx...... ...gggmmmmmddddd -- SVE integer min/max/difference (predicated)
         case IF_SVE_AE_3A: // ........xx...... ...gggmmmmmddddd -- SVE integer multiply vectors (predicated)
+        case IF_SVE_AF_3A: // ........xx...... ...gggnnnnnddddd -- SVE bitwise logical reduction (predicated)
```
kunalspathak, line 14135: I am actually thinking that instead of adding all the SVE instructions and format in existing switch case, we should have them handled under `default` to minimize the TP impact. Something like:

```c++
switch (fmt)
{
  case non_sve_fmt1:
     break;
  case non_sve_fmt2:
     break;
  ...
 default:
    emitOutputSveInstr(...);
    break;
}

emitOutputSveInstr()
{
  switch(..)
 {
   ...
   default:
     assert(!"unexpected");
     break;
 }
}
```

From #95127, here is the impact:

image

kunalspathak (changes requested):
changes LGTM. added some suggestion on reducing the TP impact.

src/coreclr/jit/emitarm64.cpp
```diff
@@ -18432,13 +18667,65 @@ emitter::insExecutionCharacteristics emitter::getInsExecutionCharacteristics(ins
...
+            result.insThroughput = PERFSCORE_THROUGHPUT_2X;
+            break;
+
+        case IF_SVE_AQ_3A:
+            switch (ins)
+            {
+                // Arithmetic, basic
+                case INS_sve_abs:
+                case INS_sve_neg:
+                    result.insLatency    = PERFSCORE_LATENCY_2C;
```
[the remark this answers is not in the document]
a74nh: this is coming from your table

src/coreclr/jit/emitarm64.cpp
a74nh: These aren't in the manual. So, I wanted something to say it was just a randomly chosen high value. Maybe `// Placeholder` ?

src/coreclr/jit/instr.h
```diff
@@ -282,6 +282,11 @@ enum insOpts : unsigned
     INS_OPTS_SCALABLE_S_TO_SIMD,
     INS_OPTS_SCALABLE_D_TO_SIMD,
 
+    INS_OPTS_SCALABLE_B_TO_SIMD_VECTOR,
```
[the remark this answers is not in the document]
a74nh: `INS_OPTS_SCALABLE_B_TO_SIMD` goes to a scalar in an FP register (eg S1). I think it needs renaming to `INS_OPTS_SCALABLE_B_TO_FP_SCALAR`.

`INS_OPTS_SCALABLE_B_TO_SIMD_VECTOR` goes to an NEON vector register (eg v1.4s). I was using `SIMD` to mean `fixed`. Maybe this one is ok as it is?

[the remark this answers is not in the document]
kunalspathak: yes, i think renaming `INS_OPTS_SCALABLE_B_TO_SIMD` -> `INS_OPTS_SCALABLE_B_TO_FP_SCALAR`  will work

src/coreclr/jit/emitarm64.cpp
kunalspathak: probably use `// TODO-SVE` that we use elsewhere too

src/coreclr/jit/emitarm64.cpp
a74nh: I'm happy with that.

Addition thoughts:
- Add an assert in the default case: `assert(fmt >= first_sve_fmt && fmt <= last_sve_fmt)`.
- Do this for all the functions with SVE checks. Eg `Emit_R_R_R()`, `emitOutputInstr()` `getInsExecutionCharacteristics()` etc etc.
- Move all the new SVE functions to a new file emitarm64sve.cpp

Were you planning on doing that?

kunalspathak: > I'm happy with that.
> 
> Addition thoughts:
> 
> * Add an assert in the default case: `assert(fmt >= first_sve_fmt && fmt <= last_sve_fmt)`.
> * Do this for all the functions with SVE checks. Eg `Emit_R_R_R()`, `emitOutputInstr()` `getInsExecutionCharacteristics()` etc etc.
> * Move all the new SVE functions to a new file emitarm64sve.cpp

They all sound good to me.

> Were you planning on doing that?

I was not (yet). It just that I checked the TP impact that made me realize that we should do it. Let me know if you want to do it as part of this PR or I am happy to do it but this PR will have to wait for it to finish. Let me know what works for you.

src/coreclr/jit/instr.h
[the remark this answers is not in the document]
a74nh: Looking at this, `INS_OPTS_SCALABLE_B_TO_SIMD` -> `INS_OPTS_SCALABLE_B_TO_SIMD_SCALAR` works better, as doesn't indicate if the value in the simd scalar is an int or float.

src/coreclr/jit/emitarm64.cpp
a74nh: > I was not (yet). It just that I checked the TP impact that made me realize that we should do it. Let me know if you want to do it as part of this PR or I am happy to do it but this PR will have to wait for it to finish. Let me know what works for you.

Happy to hold this PR back. In the meantime, I'll probably add a few more into it. Either way, I didn't want to start something new myself today.

src/coreclr/jit/emitarm64.cpp
```diff
@@ -1024,6 +1026,75 @@ void emitter::emitInsSanityCheck(instrDesc* id)
...
+        case IF_SVE_AI_3A: // ........xx...... ...gggnnnnnddddd -- SVE integer add reduction (predicated)
+            elemsize = id->idOpSize();
+            assert(insOptsScalableWideningToSimdScalar(id->idInsOpt())); // xx
+            assert(isVectorRegister(id->idReg1()));                      // ddddd
+            assert(isLowPredicateRegister(id->idReg2()));                // ggg
+            assert(isVectorRegister(id->idReg3()));                      // mmmmm
+            assert(isValidVectorElemsizeWidening(elemsize));
+            break;
+
+        // Scable, possibly FP.
```
BruceForstall, line 1051:
```suggestion
        // Scalable, possibly FP.
```

src/coreclr/jit/emitarm64.cpp
```diff
@@ -1024,6 +1026,75 @@ void emitter::emitInsSanityCheck(instrDesc* id)
...
+        case IF_SVE_AP_3A: // ........xx...... ...gggnnnnnddddd -- SVE bitwise unary operations (predicated)
+            switch (id->idIns())
+            {
+                case INS_sve_fabs:
+                case INS_sve_fneg:
+                    assert(insOptsScalableFloat(id->idInsOpt())); // xx
+                    break;
+                default:
+                    assert(insOptsScalableSimple(id->idInsOpt())); // xx
```
BruceForstall, line 1060: nit: add a `break;`

src/coreclr/jit/emitarm64.cpp
```diff
@@ -13941,8 +14132,16 @@ size_t emitter::emitOutputInstr(insGroup* ig, instrDesc* id, BYTE** dp)
         case IF_SVE_AC_3A: // ........xx...... ...gggmmmmmddddd -- SVE integer divide vectors (predicated)
         case IF_SVE_AD_3A: // ........xx...... ...gggmmmmmddddd -- SVE integer min/max/difference (predicated)
         case IF_SVE_AE_3A: // ........xx...... ...gggmmmmmddddd -- SVE integer multiply vectors (predicated)
+        case IF_SVE_AF_3A: // ........xx...... ...gggnnnnnddddd -- SVE bitwise logical reduction (predicated)
+        case IF_SVE_AG_3A: // ........xx...... ...gggnnnnnddddd -- SVE bitwise logical reduction (quadwords)
+        case IF_SVE_AI_3A: // ........xx...... ...gggnnnnnddddd -- SVE integer add reduction (predicated)
+        case IF_SVE_AJ_3A: // ........xx...... ...gggnnnnnddddd -- SVE integer add reduction (quadwords)
+        case IF_SVE_AK_3A: // ........xx...... ...gggnnnnnddddd -- SVE integer min/max reduction (predicated)
+        case IF_SVE_AL_3A: // ........xx...... ...gggnnnnnddddd -- SVE integer min/max reduction (quadwords)
```
BruceForstall, line 14140: Should we mix the 'n' and 'm' patterns in the same "case"? I suppose it's assuming that 'idReg3()' is either 'm' or 'n', depending on the format.

src/coreclr/jit/emitarm64.cpp
```diff
@@ -16218,6 +16421,38 @@ void emitter::emitDispInsHelp(
...
 
+        // <V><d>, <Pg>, <Zn>.<T>
+        case IF_SVE_AF_3A: // ........xx...... ...gggnnnnnddddd -- SVE bitwise logical reduction (predicated)
+        case IF_SVE_AK_3A: // ........xx...... ...gggnnnnnddddd -- SVE integer min/max reduction (predicated)
+            emitDispReg(id->idReg1(), size, true);                    // ddddd
+            emitDispPredicateReg(id->idReg2(), PREDICATE_NONE, true); // ggg
+            emitDispSveReg(id->idReg3(), id->idInsOpt(), false);      // mmmmm
+            break;
+
+        // <Vd>.<T>, <Pg>, <Zn>.<Tb>
```
BruceForstall, line 16432: This comment is very helpful. Maybe we should add them to all case "sets"?

BruceForstall (commented):
A couple nits/minor comments

src/coreclr/jit/emitarm64.cpp
kunalspathak: latest runs from this PR makes me wonder how much to trust them:

image

but I guess moving the sve code in `emitarm64sve.cpp` will be general goodness.

a74nh: If moving sve out _does_ improve throughput, then you might want to consider doing the same for NEON instructions? This might make NEON throughput slower, but would improve throughput for general instructions.

NEON is only used for longer running, frequently used methods, so the compiler throughput is less important. Whereas short running, single use methods will all be general instructions, so the compiler throughput matters more for those.

(it also shrinks file size, which makes me happy)

src/coreclr/jit/emitarm64.cpp
a74nh: Yes, idReg3() is the same in both.

I wasn't sure whether to keep them the same either - but, because it's the same code I went with the one block.

src/coreclr/jit/emitarm64.cpp
a74nh: I've already done it for all the other SVE instructions done so far. I'm not planning on going back over NEON etc. :)

src/coreclr/jit/codegenarm64.cpp
```diff
@@ -5437,6 +5437,7 @@ void CodeGen::genProfilingLeaveCallback(unsigned helper)
 // #define ALL_ARM64_EMITTER_UNIT_TESTS_GENERAL
 // #define ALL_ARM64_EMITTER_UNIT_TESTS_ADVSIMD
 // #define ALL_ARM64_EMITTER_UNIT_TESTS_SVE
+// #define ALL_ARM64_EMITTER_UNIT_TESTS_SVE_UNSUPPORTED
```
kunalspathak, line 5440: not sure if it is worth adding a comment in next PR why they are not supported. "The encodings of the instructions cannot be verified currently, so are added under this flag"

src/coreclr/jit/emitarm64.cpp
```diff
@@ -8184,6 +8257,68 @@ void emitter::emitIns_R_R_R(
...
+            assert(isLowPredicateRegister(reg2));
+            assert(isVectorRegister(reg3));
+            assert(insOptsScalableToSimdScalar(opt));
+            fmt = IF_SVE_AF_3A;
+            break;
+
+        case INS_sve_andqv:
+        case INS_sve_eorqv:
+        case INS_sve_orqv:
+            unreached(); // TODO-SVE: Not yet supported.
```
kunalspathak, line 8273: likewise here.

kunalspathak: can any of these instructions block us from the basic scenario that we are thinking of enabling?

kunalspathak (approved):
LGTM

PR #95341, opened 2023-11-28
Make source generated JsonTypeInfo properties nullable-oblivious

### LINKED ISSUE #95083: Passing generic class with nullable reference type in JsonSerializableAttribute raises CS8620 when used
### Description

Take this class
```cs
[JsonSerializable(typeof(IDictionary<string, object?>))]
public partial class JsonContextTest : JsonSerializerContext { }
```
When used like this:
```cs
IDictionary<string, object?> d = new Dictionary<string, object?>()
{
    ["A"] = "b",
    ["D"] = 1,
    ["C"] = null
};
var b = JsonContent.Create(d, JsonContextTest.Default.IDictionaryStringObject);
```
It raises CS8620 `Argument of type 'System.Collections.Generic.IDictionary<string,object?>' cannot be used for parameter 'inputValue' of type 'System.Collections.Generic.IDictionary<string,object>?' in 'System.Net.Http.Json.JsonContent.Create<T>' because of differences in the nullability of reference types`
Generated code for JsonContextTest:
```cs
public global::System.Text.Json.Serialization.Metadata.JsonTypeInfo<global::System.Collections.Generic.IDictionary<string, object>> IDictionaryStringObject
``` 
Notice the absence of `object?`
It does not raise any runtime errors however when using
My colleague is treating all warnings as errors so is there any way to avoid raising CS8620 besides `NoWarn`?

### Reproduction Steps

Create class
```cs
[JsonSerializable(typeof(IDictionary<string, object?>))]
public partial class JsonContextTest : JsonSerializerContext { }
```
Use it
```cs
IDictionary<string, object?> d = new Dictionary<string, object?>()
{
    ["A"] = "b",
    ["D"] = 1,
    ["C"] = null
};
var b = JsonContent.Create(d, JsonContextTest.Default.IDictionaryStringObject);
```

### Expected behavior

No warning raised

### Actual behavior

CS8620 is raised

### Regression?

_No response_

### Known Workarounds

_No response_

### Configuration

.NET 8

### Other information

_No response_

eiriktsarpalis:
Fix #95083.

### DIFF
--- a/src/libraries/System.Text.Json/gen/JsonSourceGenerator.Emitter.cs
+++ b/src/libraries/System.Text.Json/gen/JsonSourceGenerator.Emitter.cs
@@ -3,12 +3,10 @@
 
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Globalization;
 using System.Linq;
 using System.Reflection;
 using System.Text.Json.Serialization;
 using Microsoft.CodeAnalysis;
-using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.Text;
 using SourceGenerators;
 
@@ -1009,7 +1007,9 @@ namespace System.Text.Json.SourceGeneration
                     /// <summary>
                     /// Defines the source generated JSON serialization contract metadata for a given type.
                     /// </summary>
+                    #nullable disable annotations // Marking the property type as nullable-oblivious.
                     public {{typeInfoFQN}} {{typeInfoPropertyName}}
+                    #nullable enable annotations
                     {
                         get => _{{typeInfoPropertyName}} ??= ({{typeInfoFQN}}){{OptionsInstanceVariableName}}.GetTypeInfo(typeof({{typeFQN}}));
                     }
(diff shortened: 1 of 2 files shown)


stephentoub (approved):
Since there's no ability or need to create different contexts for different annotations of the same types, this looks like the right answer.

src/libraries/System.Text.Json/gen/JsonSourceGenerator.Emitter.cs
```diff
@@ -1009,7 +1007,9 @@ private static void GenerateTypeInfoFactoryHeader(SourceWriter writer, TypeGener
                     /// <summary>
                     /// Defines the source generated JSON serialization contract metadata for a given type.
                     /// </summary>
+                    #nullable disable annotations // Marking the property type as nullable-oblivious.
                     public {{typeInfoFQN}} {{typeInfoPropertyName}}
+                    #nullable enable annotations
```
stephentoub, line 1012: Are there other places where the generic type arguments used end up surfacing and that would also benefit from being made oblivious?

eiriktsarpalis: That's the only public API surfacing generic metadata types. Everything else is using the non-generic `Type => JsonTypeInfo` resolver methods.

PR #95352, opened 2023-11-28
Add regression testing for the STJv8 source generator.

eiriktsarpalis:
Adds .NET 8 source generated sources to complement the existing .NET 6 and .NET 7 regression suites.


tarekgh (approved):

jozkee (approved):
Given that Net70RegressionTests.cs is identical to Net80RegressionTests.cs, is this only a snapshot of some generated types on 8.0? Will we continue adding them each release indiscriminately?

eiriktsarpalis:
> Given that Net70RegressionTests.cs is identical to Net80RegressionTests.cs, is this only a snapshot of some generated types on 8.0? 

Correct, this is smoke testing functionality of the source generated baselines against the current runtime bits, so what tests are being run doesn't change much.

> Will we continue adding them each release indiscriminately?

It depends on whether a particular release makes substantial changes to how it generates code. We have been updating the SG on every release so it made sense to include new regression tests every time.

PR #95353, opened 2023-11-28
Clean up GT_NOP

EgorBo:
This PR addresses SingleAccretion and Jakob's feedback in https://github.com/dotnet/runtime/pull/95249 (last comment)

So we now validate that `GT_NOP` has no child nodes and is always of `TYP_VOID` type

### DIFF
--- a/src/coreclr/jit/fgdiagnostic.cpp
+++ b/src/coreclr/jit/fgdiagnostic.cpp
@@ -2576,8 +2576,8 @@ Compiler::fgWalkResult Compiler::fgStress64RsltMulCB(GenTree** pTree, fgWalkData
     *pTree              = pComp->gtNewCastNode(TYP_INT, tree, false, TYP_INT);
 
     // To ensure optNarrowTree() doesn't fold back to the original tree.
-    tree->gtGetOp1()->gtFlags |= GTF_CAST_STRESS_KEEP;
-    tree->gtGetOp2()->gtFlags |= GTF_CAST_STRESS_KEEP;
+    tree->gtGetOp1()->gtDebugFlags |= GTF_DEBUG_CAST_DONT_FOLD;
+    tree->gtGetOp2()->gtDebugFlags |= GTF_DEBUG_CAST_DONT_FOLD;
 
     JITDUMP("STRESS_64RSLT_MUL after:\n")
     DISPTREE(*pTree)
--- a/src/coreclr/jit/fgdiagnostic.cpp
+++ b/src/coreclr/jit/fgdiagnostic.cpp
@@ -2570,12 +2570,14 @@ Compiler::fgWalkResult Compiler::fgStress64RsltMulCB(GenTree** pTree, fgWalkData
     JITDUMP("STRESS_64RSLT_MUL before:\n")
     DISPTREE(tree)
 
-    // To ensure optNarrowTree() doesn't fold back to the original tree.
-    tree->AsOp()->gtOp1 = pComp->gtNewCastNode(TYP_LONG, pComp->gtAddFakeSideEffect(tree->gtGetOp1()), false, TYP_LONG);
-    tree->AsOp()->gtOp1 = pComp->gtNewCastNode(TYP_LONG, pComp->gtAddFakeSideEffect(tree->gtGetOp1()), false, TYP_LONG);
-    tree->AsOp()->gtOp2 = pComp->gtNewCastNode(TYP_LONG, pComp->gtAddFakeSideEffect(tree->gtGetOp2()), false, TYP_LONG);
+    tree->AsOp()->gtOp1 = pComp->gtNewCastNode(TYP_LONG, tree->gtGetOp1(), false, TYP_LONG);
+    tree->AsOp()->gtOp2 = pComp->gtNewCastNode(TYP_LONG, tree->gtGetOp2(), false, TYP_LONG);
     tree->gtType        = TYP_LONG;
     *pTree              = pComp->gtNewCastNode(TYP_INT, tree, false, TYP_INT);
+
+    // To ensure optNarrowTree() doesn't fold back to the original tree.
+    tree->gtGetOp1()->gtFlags |= GTF_CAST_STRESS_KEEP;
+    tree->gtGetOp2()->gtFlags |= GTF_CAST_STRESS_KEEP;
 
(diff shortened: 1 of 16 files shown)


jakobbotsch:
You can also remove the `GTK_UNOP` from gtlist.h.

src/coreclr/jit/morph.cpp
```diff
@@ -6384,11 +6384,6 @@ void Compiler::fgValidateIRForTailCall(GenTreeCall* call)
                 return WALK_ABORT;
             }
 
-            // GT_NOP might appear due to stores that end up as
-            // self-stores, which get morphed to GT_NOP.
-            if (tree->OperIs(GT_NOP))
-            {
-            }
```
jakobbotsch, line 6391: I don't think this one can be removed. Block morphing will still morph these to `GT_NOP`.

EgorBo: but um.. isn't a no-op code? it checks for GT_NOP and does nothing?

EgorBo: ah, I see! didn't notice `else`

jakobbotsch: Feel free to add a `// fallthrough` or something to make it clear 🙂

src/coreclr/jit/gentree.cpp
```diff
@@ -10142,7 +10143,6 @@ GenTreeUseEdgeIterator::GenTreeUseEdgeIterator(GenTree* node)
             return;
 
         // Unary operators with an optional operand
-        case GT_NOP:
```
SingleAccretion, line 10145: `NullOp1Legal` should be updated too.

src/coreclr/jit/fgdiagnostic.cpp
```diff
@@ -2572,7 +2572,6 @@ Compiler::fgWalkResult Compiler::fgStress64RsltMulCB(GenTree** pTree, fgWalkData
 
     // To ensure optNarrowTree() doesn't fold back to the original tree.
     tree->AsOp()->gtOp1 = pComp->gtNewCastNode(TYP_LONG, tree->AsOp()->gtOp1, false, TYP_LONG);
```
SingleAccretion, line 2574: `optNarrowTree` might "fold that back" now. You could use a `COMMA` with a volatile store to an unused local to prevent it from doing so, perhaps.

src/coreclr/jit/gtlist.h
```diff
@@ -53,7 +53,7 @@ GTNODE(CNS_VEC          , GenTreeVecCon      ,0,0,GTK_LEAF)
 //-----------------------------------------------------------------------------
 
 GTNODE(NOT              , GenTreeOp          ,0,0,GTK_UNOP)
-GTNODE(NOP              , GenTree            ,0,0,GTK_UNOP|DBK_NOCONTAIN)
+GTNODE(NOP              , GenTree            ,0,0,GTK_LEAF|DBK_NOCONTAIN)
```
SingleAccretion, line 56:
```suggestion
GTNODE(NOP              , GenTree            ,0,1,GTK_LEAF|DBK_NOCONTAIN)
```
`NOP` is now illegal as a VN func (not that it ever was actually legal).

src/coreclr/jit/valuenum.cpp
```diff
@@ -11165,6 +11165,7 @@ void Compiler::fgValueNumberTree(GenTree* tree)
 
             // These do not represent values.
             case GT_NO_OP:
+            case GT_NOP:
```
SingleAccretion, line 11168: There is code above:

https://github.com/dotnet/runtime/blob/09eebddbbb1e75a6ab5ecd9dc85520f0ad9b93cf/src/coreclr/jit/valuenum.cpp#L11391-L11394

which should be deleted. I don't think the "actually nullary" case there is reachable now.

src/coreclr/jit/morph.cpp
```diff
@@ -9042,6 +9042,11 @@ GenTree* Compiler::fgMorphSmpOp(GenTree* tree, MorphAddrContext* mac, bool* optA
         return tree;
     }
 
+    if (tree->IsNothingNode())
```
SingleAccretion, line 9045: Is this actually reachable? Presumably, `NOP`s should go through `fgMorphLeaf` now.

EgorBo: Yes, `gtFoldExpr` called above may return a NOP (to be precise, `gtFoldExprConst` does that for:
```
*  BOUNDS_CHECK_Rng void   <l:$285, c:$284>
+--*  CNS_INT   int    0 $40
\--*  CNS_INT   int    1 $41
```

src/coreclr/jit/fgdiagnostic.cpp
jakobbotsch: Alternatively a debug-only flag might be a bit simpler, since this is just for stress-only code anyway.

src/coreclr/jit/gentree.h
```diff
@@ -499,6 +499,8 @@ enum GenTreeFlags : unsigned int
     GTF_ADDRMODE_NO_CSE         = 0x80000000, // GT_ADD/GT_MUL/GT_LSH -- Do not CSE this node only, forms complex
                                               //                         addressing mode
 
+    GTF_CAST_STRESS_KEEP        = 0x00400000, // GT_CAST    -- A recommendation to keep this cast (used by stress mode)
```
jakobbotsch, line 502: It should be in `GenTreeDebugFlags`.

EgorBo: Suprisingly, CI + outerloops is fully green from without re-runs 😮

EgorBo:
@jakobbotsch @dotnet/jit-contrib PTAL, this is ready for review

jakobbotsch (approved):
LGTM, nice cleanup. Why is there a diff?

EgorBo:
> LGTM, nice cleanup. Why is there a diff?

It seems to only 1 method slightly regressed, but that method is quite huge (5k bytes of codegen).
I was able to obtain JitDump for before/after and the only difference I see is in the VN phase, so the change happend because we no longer number NOP, and it changed VN for the parents. So I presume it slightly changed CSE behaviour due to different (or same?) numbers. But since it's just 1 method I guess it's not worth investigating much.

image

PR #95362, opened 2023-11-28
Free the tls memory on thread termination

kunalspathak:
The TLS memory that we used to store the TLS data per thread was never freed on thread termination. This results in memory leak.


src/coreclr/vm/threads.cpp
```diff
@@ -7683,6 +7683,10 @@ Frame * Thread::NotifyFrameChainOfExceptionUnwind(Frame* pStartFrame, LPVOID pvL
     return pFrame;
 }
 
+#ifndef _MSC_VER
```
AaronRobinsonMSFT, line 7686: Why does the non-MSVC need a redeclaration here? Is this because of `__declspec(selectany)` mechanism? If so, it might be appropriate to remove the duplicate definition/declaration pattern and instead follow a simple compiler agnostic `extern` declaration and define the instance in `jitinterface.cpp`.

kunalspathak: fixed.

kunalspathak:
@jkotas

AaronRobinsonMSFT (approved):

AndyAyersMS (approved):
Changes look ok to me. I think @AaronRobinsonMSFT is also taking a look.

mangod9 (approved):

src/coreclr/vm/threads.cpp
```diff
@@ -7702,6 +7708,15 @@ void Thread::DeleteThreadStaticData()
     CONTRACTL_END;
 
     m_ThreadLocalBlock.FreeTable();
+    
+    delete[] t_ThreadStatics.NonGCThreadStaticBlocks;
```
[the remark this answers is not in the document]
kunalspathak: correct.

AndyAyersMS:
@kunalspathak no symbolization on the stack traces so I can only guess what's behind the failures -- maybe the cleanup needs to be conditional on the TLS not being nullptr? Linux may be more tolerant of calling delete on nulls.

kunalspathak:
> maybe the cleanup needs to be conditional on the TLS not being nullptr

Here is what I found for 1 repro, but it doesn't always stay like that. Notice the `JIT_GetSharedGCThreadStaticBaseOptimized ` at the end, so my first guess was it has to do with `CONSISTENCY_CHECK(staticBlock != NULL);`, but adding printf doesn't help as it is not consistent.

```
CORECLR! CHECK::Trigger + 0x1E6 (0x00007ffe`907a7356)
CORECLR! CLRVectoredExceptionHandlerPhase3 + 0x4B5 (0x00007ffe`900d7135)
CORECLR! CLRVectoredExceptionHandlerPhase2 + 0x91 (0x00007ffe`900d6a11)
CORECLR! CLRVectoredExceptionHandler + 0xDFD (0x00007ffe`900d696d)
CORECLR! CLRVectoredExceptionHandlerShim + 0x1B4 (0x00007ffe`900d73b4)
NTDLL! RtlGetLengthWithoutLastFullDosOrNtPathElement + 0x3DA (0x00007fff`c6e67a8a)
NTDLL! RtlFindCharInUnicodeString + 0x872 (0x00007fff`c6e0e242)
NTDLL! KiUserExceptionDispatcher + 0x2E (0x00007fff`c6e9340e)
CORECLR! JIT_GetSharedGCThreadStaticBaseOptimized + 0x4CE (0x00007ffe`903f8e9e)
<no module>! <no symbol> + 0x0 (0x00007ffe`375eb155)
    File: D:\git\runtime2\src\coreclr\vm\excep.cpp Line: 7177
    Image: E:\helix\95362_2\correlation-payload\test-runtime-net9.0-windows-Release-x64\dotnet.exe
```

Under windbg, I always hit this kind of AV, which I think is coming because `staticBlock` is nullptr? The AV in windbg is consistent but I don't understand why I don't see if windbg is not attached:

```asm
0007ffe`4a379c9e c5fe7f65e0      vmovdqu ymmword ptr [rbp-20h],ymm4
00007ffe`4a379ca3 48894d10        mov     qword ptr [rbp+10h],rcx
00007ffe`4a379ca7 488b4d10        mov     rcx,qword ptr [rbp+10h]
00007ffe`4a379cab 488b4908        mov     rcx,qword ptr [rcx+8]
00007ffe`4a379caf 48894df8        mov     qword ptr [rbp-8],rcx
00007ffe`4a379cb3 488b4df8        mov     rcx,qword ptr [rbp-8]
00007ffe`4a379cb7 48baf8597e4afe7f0000 mov rdx,7FFE4A7E59F8h
00007ffe`4a379cc1 e89a14435c      call    coreclr!JIT_ClassProfile32 (00007ffe`a67ab160)
00007ffe`4a379cc6 488b4df8        mov     rcx,qword ptr [rbp-8]
00007ffe`4a379cca 48894de8        mov     qword ptr [rbp-18h],rcx
00007ffe`4a379cce 488b4de8        mov     rcx,qword ptr [rbp-18h]
00007ffe`4a379cd2 488b45e8        mov     rax,qword ptr [rbp-18h]
00007ffe`4a379cd6 488b00          mov     rax,qword ptr [rax] ds:00000000`00000000=???????????????? ; <----
00007ffe`4a379cd9 488b4060        mov     rax,qword ptr [rax+60h]
00007ffe`4a379cdd ff5018          call    qword ptr [rax+18h]
00007ffe`4a379ce0 85c0            test    eax,eax

```

kunalspathak:
I just pushed the conditional checking as well as reset the other two variables.

kunalspathak:
> I just pushed the conditional checking as well as reset the other two variables.

I verified locally and MSVC doesn't complain if using `delete[]` on `nullptr`.

jkotas:
Is the `DeleteThreadStaticData` method guaranteed to be called on the terminating thread? Can it be called on some other thread?

kunalspathak:
> Is the `DeleteThreadStaticData` method guaranteed to be called on the terminating thread? Can it be called on some other thread?

Looking at the comments about `OnThreadTerminate`, it seems to be the case:

https://github.com/dotnet/runtime/blob/8cb3c6187f9e1338f22f9e37c4dfe615123eaa59/src/coreclr/vm/threads.cpp#L2353-L2359)

One of the caller `DestroyThread` of `OnThreadTerminate` has this assert:

https://github.com/dotnet/runtime/blob/8cb3c6187f9e1338f22f9e37c4dfe615123eaa59/src/coreclr/vm/threads.cpp#L894

but the fact that inside `OnThreadTerminate()` has this code just few lines after `DeleteThreadStaticData()` makes me wonder if the terminating thread is the one that calls `DeleteThreadStaticData`: 

https://github.com/dotnet/runtime/blob/8cb3c6187f9e1338f22f9e37c4dfe615123eaa59/src/coreclr/vm/threads.cpp#L2962-L2964

I don't want to disturb the CI, but once it is done, I can try adding this check inside `DeleteThreadStaticData`

JulieLeeMSFT:
@mangod9 for his input:
> but the fact that inside `OnThreadTerminate()` has this code just few lines after `DeleteThreadStaticData()` makes me wonder if the terminating thread is the one that calls `DeleteThreadStaticData`:

kunalspathak:
> but the fact that inside `OnThreadTerminate()` has this code just few lines after `DeleteThreadStaticData()` makes me wonder if the terminating thread is the one that calls `DeleteThreadStaticData`:

alright, I added an assert locally and I hit that assert.

```diff
     m_ThreadLocalBlock.FreeTable();

+    Thread *pCurrentThread = GetThreadNULLOk();
+    DWORD CurrentThreadID = pCurrentThread?pCurrentThread->GetThreadId():0;
+    DWORD ThisThreadID = GetThreadId();
+
+    _ASSERTE(CurrentThreadID == ThisThreadID);
```

```
Assert failure(PID 84504 [0x00014a18], Thread: 99544 [0x184d8]): CurrentThreadID == ThisThreadID

CORECLR! Thread::DeleteThreadStaticData + 0x13F (0x00007ffe`ad4f7b6f)
CORECLR! Thread::OnThreadTerminate + 0x22A (0x00007ffe`ad500f6a)
CORECLR! Thread::CleanupDetachedThreads + 0x219 (0x00007ffe`ad4f5429)
CORECLR! Thread::DoExtraWorkForFinalizer + 0x1A7 (0x00007ffe`ad4f9227)
CORECLR! FinalizerThread::FinalizerThreadWorker + 0x8C2 (0x00007ffe`ad610952)
CORECLR! ManagedThreadBase_DispatchInner + 0xB4 (0x00007ffe`ad4ffd84)
CORECLR! ManagedThreadBase_DispatchMiddle + 0xB6 (0x00007ffe`ad4ffe86)
CORECLR! ``ManagedThreadBase_DispatchOuter'::`11'::__Body::Run'::`5'::__Body::Run + 0x7B (0x00007ffe`ad50265b)
CORECLR! `ManagedThreadBase_DispatchOuter'::`11'::__Body::Run + 0x93 (0x00007ffe`ad502953)
CORECLR! ManagedThreadBase_DispatchOuter + 0xDD (0x00007ffe`ad4fff9d)
    File: D:\git\runtime2\src\coreclr\vm\threads.cpp Line: 7716
    Image: E:\helix\95362_2\correlation-payload\test-runtime-net9.0-windows-Release-x64\dotnet.exe
```

So given that I can guard the deletion only if we are on current thread, where can we delete the memory for threads whose `DeleteThreadStaticData` gets called by different thread?

kunalspathak:
> Is there code that detached threads execute before termination?

Seems `DetachThread` could be the right method.

https://github.com/dotnet/runtime/blob/8cb3c6187f9e1338f22f9e37c4dfe615123eaa59/src/coreclr/vm/threads.cpp#L983

JulieLeeMSFT:
@kouvel, please see Kunal's question. We need to merge this PR tonight.

jkotas (approved):

src/coreclr/vm/ceemain.cpp
```diff
@@ -1767,6 +1768,35 @@ void EnsureTlsDestructionMonitor()
...
+    if (t_ThreadStatics.NonGCThreadStaticBlocks != nullptr)
+    {
+        delete[] t_ThreadStatics.NonGCThreadStaticBlocks;
+        t_ThreadStatics.NonGCThreadStaticBlocks = nullptr;
+    }
+    if (t_ThreadStatics.GCThreadStaticBlocks != nullptr)
+    {
+        delete[] t_ThreadStatics.GCThreadStaticBlocks;
+        t_ThreadStatics.GCThreadStaticBlocks = nullptr;
+    }
```
AaronRobinsonMSFT, line 1797: Nit: None of these null checks are needed.

AaronRobinsonMSFT (approved):

PR #95368, opened 2023-11-28
[wasm] CI: trigger only WBT on template changes

radical:
.


radical:
The problem can be seen in https://dev.azure.com/dnceng-public/public/_build/results?buildId=482842&view=logs&jobId=d4e38924-13a0-58bd-9074-6a4810543e7c .

lewing (approved):

PR #95376, opened 2023-11-29
Avoid redundant GetVirtualMemoryLimit syscall

jkotas:


src/coreclr/gc/gc.cpp
```diff
@@ -48097,7 +48097,8 @@ HRESULT GCHeap::Initialize()
             // If no hard_limit is configured the reservation size is min of 1/2 GetVirtualMemoryLimit() or max of 256Gb or 2x physical limit.
             gc_heap::regions_range = max((size_t)256 * 1024 * 1024 * 1024, (size_t)(2 * gc_heap::total_physical_mem));
         }
-        gc_heap::regions_range = min(gc_heap::regions_range, GCToOSInterface::GetVirtualMemoryLimit()/2);
+        size_t virtualMemoryLimit = GCToOSInterface::GetVirtualMemoryLimit();
+        gc_heap::regions_range = min(gc_heap::regions_range, virtualMemoryLimit/2);
```
jkotas, line 48101: min/max macros from stdlib.h are perf traps. They will evaluate their arguments twice.

`#define min(a,b) (((a) < (b)) ? (a) : (b))`

stephentoub (approved):

src/coreclr/gc/gc.cpp
```diff
@@ -48097,7 +48097,8 @@ HRESULT GCHeap::Initialize()
             // If no hard_limit is configured the reservation size is min of 1/2 GetVirtualMemoryLimit() or max of 256Gb or 2x physical limit.
             gc_heap::regions_range = max((size_t)256 * 1024 * 1024 * 1024, (size_t)(2 * gc_heap::total_physical_mem));
         }
-        gc_heap::regions_range = min(gc_heap::regions_range, GCToOSInterface::GetVirtualMemoryLimit()/2);
+        size_t virtualMemoryLimit = GCToOSInterface::GetVirtualMemoryLimit();
```
Maoni0, line 48100:
```suggestion
        size_t virtual_mem_limit = GCToOSInterface::GetVirtualMemoryLimit();
```
nit - formatting

Maoni0 (approved):

janvorli (approved):
LGTM, thank you!

src/coreclr/gc/gc.cpp
teo-tsirpanis: Couldn't it be using `std::min`?

jkotas: Maybe, I am not sure whether it would work here.

In general, we are not able to use C++ standard libraries (yet) because of collisions with the CoreCLR PAL.

PR #95379, opened 2023-11-29
JIT: removed unneeded throw helper blocks

### LINKED ISSUE #93948: JIT: clean up unused throw helpers
https://github.com/dotnet/runtime/pull/93371 deferred creation of shared throw helper blocks, but introduced more cases where throw helper blocks end up not being needed.

There's a divergence between the point that the JIT "demands" the creation of a helper (via `fgAddCodeRef`) and requests the helper block, and in between the JIT may optimize away the code that needed the helper.

So the fix here is either to defer or reaffirm the demands, so that they better reflect actual needs, or else allow unused helper blocks to be removed later on.

AndyAyersMS:
Various optimizations can happen between the time the throw helper blocks are first requested (morph) and finally used (codegen). Prune away ones that aren't needed during the stack level setter.

Fixes #93948.

### DIFF
--- a/src/coreclr/jit/stacklevelsetter.cpp
+++ b/src/coreclr/jit/stacklevelsetter.cpp
@@ -14,9 +14,9 @@ StackLevelSetter::StackLevelSetter(Compiler* compiler)
     , maxStackLevel(0)
     , memAllocator(compiler->getAllocator(CMK_CallArgs))
     , putArgNumSlots(memAllocator)
+    , throwHelperBlocksUsed(comp->fgUseThrowHelperBlocks() && comp->compUsesThrowHelper)
 #if !FEATURE_FIXED_OUT_ARGS
     , framePointerRequired(compiler->codeGen->isFramePointerRequired())
-    , throwHelperBlocksUsed(comp->fgUseThrowHelperBlocks() && comp->compUsesThrowHelper)
 #endif // !FEATURE_FIXED_OUT_ARGS
 {
     // The constructor reads this value to skip iterations that could set it if it is already set.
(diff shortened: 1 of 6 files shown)


AndyAyersMS:
[Diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=483284&view=ms.vss-build-web.run-extensions-tab)

For code size this is pure goodness, but the TP impact is a bit high. Some possible mitigations:
* there's not much benefit on min opts so we can do some of this more selectively (only if required)
* the caching done by `fgFindExcptnTarget` is likely not that effective, seems like we should just keep a full hash table. Want to verify this is the big extra cost first.

AndyAyersMS:
Also what I'm doing likely won't work for `SCK_FAIL_FAST` but we don't seem to be using it? It is easy enough to just always keep these if they are pre-generated.

src/coreclr/jit/stacklevelsetter.cpp
```diff
@@ -89,16 +105,10 @@ void StackLevelSetter::ProcessBlock(BasicBlock* block)
             SubStackLevel(numSlots);
         }
 
-#if !FEATURE_FIXED_OUT_ARGS
-        // Set throw blocks incoming stack depth for x86.
-        if (throwHelperBlocksUsed && !framePointerRequired)
+        if (throwHelperBlocksUsed && node->OperMayThrow(comp))
```
jakobbotsch, line 108: Since `GTF_CALL` and `GTF_EXCEPT` are up-to-date and overapproximations here you can optimize TP with this trick:
https://github.com/dotnet/runtime/blob/3805c174d0a72dadfdbef98011b10b32df9e93f3/src/coreclr/jit/promotion.cpp#L2340-L2349

AndyAyersMS:
Should be ready for final review.

FYI @dotnet/jit-contrib

BruceForstall (approved):
Minor notes

AndyAyersMS:
SPMI failures are known bad contexts for OSX. Ignoring.

Issue #95390, opened 2023-11-29
SignedXml.CheckSignature failing for valid Enveloped signature with #xpointer(/) reference


m0sa:
### Description

`SignedXml.CheckSignature` is returning `false` for valid a enveloped signature document that is using a `<ds:References URI="#xpointer(/)">`.

This should work, given that `xpointer(/)` is explicitly [mentioned in the documentation](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.xml.signedxml?view=dotnet-plat-ext-7.0#choosing-the-transform-algorithm), and `SignedXml` [code](https://github.com/dotnet/runtime/blob/f66c1c15409ff505ba55f8fdf86bc5c041bf47a3/src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/SignedXml.cs#L836)

### Reproduction Steps

```csharp
var xmlDoc = new XmlDocument();
[...]

PR #95404, opened 2023-11-29
Fixing SignedXml.CheckSignature for enveloped signature with `#xpointer(/)` Reference

m0sa:
Adds tests for #95390 and implements a fix.


src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Reference.cs
```diff
@@ -266,18 +266,35 @@ public void LoadXml(XmlElement value)
...
+                            // This needs to match the way CalculateSignature resolves URI references.
+                            XmlNode? referenceTarget = null;
+                            if (_uri == null)
+                            {
+                                referenceTarget = transformElement.OwnerDocument;
+                            }
+                            else if (_uri.Length == 0)
+                            {
+                                referenceTarget = transformElement.OwnerDocument;
+                            }
```
vcsjones, line 285: These blocks have the same body. Can we combine them?

```suggestion
                            if (_uri == null || _uri.Length == 0)
                            {
                                referenceTarget = transformElement.OwnerDocument;
                            }
```

src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Reference.cs
```diff
@@ -266,18 +266,35 @@ public void LoadXml(XmlElement value)
...
+                            else if (_uri[0] == '#')
+                            {
+                                var idref = Utils.ExtractIdFromLocalUri(_uri);
```
vcsjones, line 288: See point 10 in the [coding style](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md) but we should not use `var` here.

```suggestion
                                string idref = Utils.ExtractIdFromLocalUri(_uri);
```

src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Reference.cs
```diff
@@ -266,18 +266,35 @@ public void LoadXml(XmlElement value)
...
+                                if (idref == "xpointer(/)")
+                                {
+                                    referenceTarget = transformElement.OwnerDocument;
+                                }
+                                else
+                                {
+                                    referenceTarget = SignedXml!.GetIdElement(transformElement.OwnerDocument, idref);
```
vcsjones, line 295: Previously, the result from `Utils.GetIdFromLocalUri` was going in as the second parameter to `GetIdElement`. Now it is `Utils.ExtractIdFromLocalUri`.

Was this intentional?

src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Reference.cs
```diff
@@ -266,18 +266,35 @@ public void LoadXml(XmlElement value)
...
+                                referenceTarget = transformElement.OwnerDocument;
+                            }
+                            else if (_uri.Length == 0)
+                            {
+                                referenceTarget = transformElement.OwnerDocument;
+                            }
+                            else if (_uri[0] == '#')
+                            {
+                                var idref = Utils.ExtractIdFromLocalUri(_uri);
+                                if (idref == "xpointer(/)")
```
vcsjones, line 289: This fixes the very specific case of `#xpointer(/)` but not something like `#xpointer(id("foo"))`. Did that work before? Or only the root document?

src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Reference.cs
m0sa: Yes, this was intentional. `GetIdFromLocalUri` and `ExtracIdFromLocalUri` [identical](https://github.com/m0sa/dotnet-runtime-95390/blob/d7542caffebeb6a7032c5e2255b605660d61e146/src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Utils.cs#L321-L358), except for the `out bool discardComments` parameter and it's implementation details. The resulting `out` value was discarded anyway.

src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Reference.cs
m0sa: Yes, that worked correctly before, since `Get/ExtractIdFromLocalUri` would return the `foo` from `xpointer(id("foo"))`, which SignedXml.GetIdElement knows how to handle. `SignedXml.GetIdElement` doesn't know how to handle  `xpointer(/)` and would return `null` in that case, which I think was the root cause for the bug.

m0sa: The `xpointer(id("xyz"))` cases are not tested at all, though. They should behave the same as URI="#id", so it would probably make more sense to tweak the existing tests with such references, to also have the `xpointer(id("id"))` counterparts.

src/libraries/System.Security.Cryptography.Xml/tests/SignedXmlTest.cs
```diff
@@ -1993,5 +1994,24 @@ public void CheckSignatureMultipleEnvelopedSignatures(string signatureParent/*,
 
             Assert.False(subject.CheckSignature());
         }
+
+        private static Dictionary<string, string> EnvelopedSignatureWithRootXpointerReference = new ()
```
vcsjones, line 1998: This dictionary is could be `readonly` and doesn't follow the static naming convention. It should be `s_envelopedSignatureWithRootXpointerReference`.

That said, maybe this would look better as a theory using `MemberData` instead of a dictionary as a lookup. What do you think?

src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Reference.cs
```diff
@@ -266,18 +266,31 @@ public void LoadXml(XmlElement value)
...
                             XmlNode? signatureTag = transformElement.SelectSingleNode("ancestor::ds:Signature[1]", nsm);
 
                             // Resolve the reference to get starting point for position calculation.
-                            XmlNode? referenceTarget =
-                                _uri.Length == 0
-                                ? transformElement.OwnerDocument
-                                : SignedXml!.GetIdElement(transformElement.OwnerDocument, Utils.GetIdFromLocalUri(_uri, out bool _));
+                            // This needs to match the way CalculateSignature resolves URI references.
+                            XmlNode? referenceTarget = null;
+                            if (_uri == null || _uri.Length == 0)
```
bartonjs, line 278: Previously, a null uri or a uri that doesn't start with `'#'` would skip all of this code.  

With this change, URIs that don't start with '#' will see `referenceTarget` as null, so there's no change there.. but now a null value for `_uri` matches OwnerDocument instead of nothing.  I can't think off of the top of my head what impact that will have, so can you explain why it was changed?

If the answer was "because otherwise [0] is hard", then the null check should just be put back into line 269:

```C#
if (transform is XmlDsigEnvelopedSignatureTransform && _uri != null)
```

m0sa: I guess what I got wrong is the comment... This should match how Reference.CalculateHashValue works. That case explicitly mentions URI null meaning the whole doc:

https://github.com/dotnet/runtime/blob/release/8.0/src/libraries/System.Security.Cryptography.Xml/src/System/Security/Cryptography/Xml/Reference.cs#L375

m0sa: Ok, so after a bit of poking around I was able to craft a document with a valid enveloped signature that has a `<Reference>` element w/o any attributes (effectively causing it to have `Uri = null`):

```
<?xml version="1.0" encoding="UTF-8"?><hello><world>Hi</world><Signature xmlns="http://www.w3.org/2000/09/xmldsig#"><SignedInfo><CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#WithComments" /><SignatureMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-sha256" /><Reference><Transforms><Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#WithComments" /></Transforms><DigestMethod Algorithm="http://www.w3.org/2001/04/xmlenc#sha256" /><DigestValue>SVaCE5w9iLXTVYTKP1t/yjjmPXvWovMYpgljGgpgz2Y=</DigestValue></Reference></SignedInfo><SignatureValue>UHeu+oXIteU77QrnOUA10sZrJYHjX8BC4pF+mL9ha64XNmPL05qGbvXyNxZlERBaSxCqsLmHTmgEEIew/BtMn/6KvI4A2prs25/ZDRjMZ1MhXQ5tHJUM+0hwnyGEklAgFU2dYtbR8Gzj541/yWcaAyo71KXmkD6SeEDHCK5qwNM=</SignatureValue><KeyInfo><KeyValue><RSAKeyValue><Modulus>pIed5HbnnEgAyc84LYtvx8CDWUotUv5+skgVTay5KQzgbRQZRJFv3x3Uioaoyfcq0gG0MtO0ZATjKUJcGWTvJoqhvR6YjiM8iR5nnksXUYjCd+zqx7UuA3QXdShXt1JLOtEOv/WNnBM6IjpKnaRDBT1+bZwOst/UGrtXdQxCqXk=</Modulus><Exponent>AQAB</Exponent></RSAKeyValue></KeyValue></KeyInfo></Signature></hello>
```

Unfortunately, though, none of the built-in signature verification methods work out of the box on this document, neither in dotnet 6 nor 8. By default `XmlDsigEnvelopedSignatureTransform.GetOutput()` throws, e.g.:

```
An XmlDocument context is required for enveloped transforms.
   at System.Security.Cryptography.Xml.XmlDsigEnvelopedSignatureTransform.GetOutput()
   at System.Security.Cryptography.Xml.TransformChain.TransformToOctetStream(Object inputObject, Type inputType, XmlResolver resolver, String baseUri)
   at System.Security.Cryptography.Xml.Reference.CalculateHashValue(XmlDocument document, CanonicalXmlNodeList refList)
   at System.Security.Cryptography.Xml.SignedXml.CheckDigestedReferences()
   at System.Security.Cryptography.Xml.SignedXml.CheckSignature(AsymmetricAlgorithm key)
   at System.Security.Cryptography.Xml.SignedXml.CheckSignatureReturningKey(AsymmetricAlgorithm& signingKey)
   at System.Security.Cryptography.Xml.SignedXml.CheckSignature()
```

Other framework / online tools (e.g. paste the above XML into https://tools.chilkat.io/xmlDsigVerify.cshtml), successfully  validate that signature out of the box.

I only managed to validate the signature by explicitly loading the `XmlDocument` into the `Transform`, manually via `LoadInput`, before calling `CheckSignature`. In that case, `CheckSignature` returns `true` in dotnet 6, but it returns `false` in dotnet 8, basically exhibiting the same behavior as outlined in #95390, but for the `Uri == null` case:

```
    XmlDocument xmlDoc = new();
	xmlDoc.LoadXml(xmlStr.Dump());

	SignedXml signedXml = new(xmlDoc);
	var signature = xmlDoc.GetElementsByTagName("Signature", SignedXml.XmlDsigNamespaceUrl)[0] as XmlElement;
	signedXml.LoadXml(signature);
	
	// without this, CheckSignature throws
	((Reference)signedXml.SignedInfo.References[0]).TransformChain[0].LoadInput(xmlDoc);
	signedXml.CheckSignature().Dump();
```

m0sa: Code now has tests to show this is indeed the desired behavior in cases where there's a `<Reference>` element without attributes.

bartonjs (approved):
I'm slightly displeased at how the null URI got included in this change sort of just-because (i.e. it wasn't part of the original problem description); but there's not really reason to break it apart at this point.  (Except that the PR / final commit needs a better title.)

`_uri` values that are non-empty and don't start with `#` will now take a different path through the code; but it ends up having the same effect.

Issue #86829, opened 2023-05-27
Unify tryrun scripts


MichalStrehovsky:
https://github.com/dotnet/runtime/pull/86781#discussion_r1207000325

We have tryrun.cmake that is mostly used for `-cross` builds.

Then we have similar things in [runtime/src/native/libs/configure.cmake](https://github.com/dotnet/runtime/blob/c5256458235e00c4bc4634c3cc389fefdd4c10d8/src/native/libs/configure.cmake#L572-L605) and https://github.com/dotnet/runtime/blob/main/eng/native/tryrun_ios_tvos.cmake.

We should reconcile this.

Related: #56622

janvorli:
@MichalStrehovsky the `tryrun.cmake` contains values for cross build that the configure.cmake would detect if the build was executed on the target OS. It is used only for checks that execute code, so they cannot be evaluated on the build hosting OS. This is a cmake concept - the `tryrun.cmake` is meant to be separate, as it is loaded by cmake as an initialization of its cache.
It is not clear to me how we could reasonably unify these.

MichalStrehovsky:
> @MichalStrehovsky the `tryrun.cmake` contains values for cross build that the configure.cmake would detect if the build was executed on the target OS. It is used only for checks that execute code, so they cannot be evaluated on the build hosting OS.

Yes, it's the same thing that the quoted part of configure.cmake is also doing. It's not clear why we spread these around the codebase

janvorli:
Oh, I am surprised to see it in the configure.cmake in libraries. The ones in coreclr don't have that, they use the tryrun.cmake.

akoeplinger:
@ViktorHofer it can stay in 8.0 for now

ilonatommy:
@akoeplinger, this looks like a fast cleanup issue, will you have capacity for it?

akoeplinger:
Hopefully 😄

PR #80295, opened 2023-01-06
Honor virtual memory limit

janvorli:
The coreclr runtime currently doesn't honor virtual memory size limit specified by the rlimit APIs (or the `ulimit -v` shell command).

There were couple of cases in the past where people have hit this problem when trying to run .NET on systems where the administrators have applied virtual memory limit for all users as a simple mean to limit the physical memory usage.

This change enables honoring that limit in the GC and PAL preallocator of the initial executable memory, as those two are the factors preventing coreclr execution on such systems.

PR #95405, opened 2023-11-29
Fix iOS/tvOS NativeAOT builds

akoeplinger:
https://github.com/dotnet/runtime/pull/80295 added a new entry to tryrun.cmake but iOS/tvOS use a separate tryrun_ios_tvos.cmake which was missing the new entry. Removing the separate file is tracked in https://github.com/dotnet/runtime/issues/86829


steveisok (approved):

akoeplinger:
Failures are unrelated.

PR #94353, opened 2023-11-03
Replace `pushd/popd` Calls With `-S/-B` CMake Flags in the Build System

ivdiazsa:
Completes #93702. With our build system requiring at least CMake version 3.20, we can now use its `-S` and `-B` flags to choose source and build paths, rather than changing directories with `pushd` and `popd`. This PR adds this enhancement to the build system.

PR #86530, opened 2023-05-19
Up Minimum CMake Version to 3.20

ivdiazsa:
The PR for #74027. All the necessary changes to increase the version of CMake will go here.

PR #95408, opened 2023-11-29
Fix propagating cmake exit code from gen-buildsys.sh

akoeplinger:
While looking at https://github.com/dotnet/runtime/pull/95405 I noticed that we didn't fail the build when the cmake configure step failed but tried to run the build regardless.

The reason is because in ece10adbddc8d7b18711f1ad753f75a6e2234caa another command `popd` was added after the cmake invocation which overwrote the exit code so build-commons.sh didn't know cmake failed.

We can use the `-S` and `-B` flags to set the directories instead again now that we use a new enough minimum CMake version.

### DIFF
--- a/eng/native/gen-buildsys.sh
+++ b/eng/native/gen-buildsys.sh
@@ -116,4 +116,6 @@ $cmake_command \
   $__UnprocessedCMakeArgs \
   "$1"
 
+cmake_exit_code=$?
 popd
+exit $cmake_exit_code
--- a/eng/native/gen-buildsys.sh
+++ b/eng/native/gen-buildsys.sh
@@ -104,9 +104,6 @@ if [[ "$host_arch" == "wasm" ]]; then
     fi
 fi
 
-# We have to be able to build with CMake 3.6.2, so we can't use the -S or -B options
-pushd "$2"
-
 $cmake_command \
   --no-warn-unused-cli \
   -G "$generator" \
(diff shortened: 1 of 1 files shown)


jkoritzinsky:
Alternatively, can we change the script to use the -S and -B flags so we don't need the pushd/pops in the first place?

akoeplinger:
@jkoritzinsky I didn't do that because of https://github.com/dotnet/runtime/blob/b95bccb1e6117acb7d64954117672493e35fe6af/eng/native/gen-buildsys.sh#L107-L108 but we're using a newer minimum CMake version now which supports those flags right?

akoeplinger:
Yup it's 3.20 now: https://github.com/dotnet/runtime/pull/86530
I'll use the flags instead.

eng/native/gen-buildsys.sh
```diff
@@ -104,16 +104,14 @@ if [[ "$host_arch" == "wasm" ]]; then
...
   "-DCMAKE_BUILD_TYPE=$buildtype" \
   "-DCMAKE_INSTALL_PREFIX=$__CMakeBinDir" \
   $cmake_extra_defines \
   $__UnprocessedCMakeArgs \
-  "$1"
+  -S "$1" \
+  -B "$3"
 
-popd
+exit $?
```
akoeplinger, line 117: still keeping this so it's clearer for the next person touching this line that the exit code is important

am11:
Duplicate of https://github.com/dotnet/runtime/pull/94353?

akoeplinger:
Thanks, closing this one then.

PR #95411, opened 2023-11-29
[browser] eager allocation of Promise/Task result for async JSImport/JSExport

pavelsavara:
# Motivation

"Because the caller already knows they need to create a proxy-promise, which would be later bound to result of the called function."

This makes it possible to send marshaling "stack frame" of async `JSImport`/`JSExport` methods as **one-way message**. 
Because `Promise`/`Task` result is created by sender thread ahead of time and sending thread could `await` it even before receiving thread saw the message. 

This PR doesn't contain the MT changes which will use this.

# Summary

PR https://github.com/dotnet/runtime/pull/93648 introduced virtual GC and JS handles.
They are good because we could allocate the handle to the future object of the other side, without crossing to that other side to create the real object. That matters in MT scenarios, in which we also need to cross to the correct thread.

This PR optimizes it further by allocating the real `PromiseHolder` or `TaskHolder` and normal non-virtual handle for it. 
We do it for the return value of async `JSImport` or `JSExport`.

The `Task`/`Promise` passed as parameter keep using the virtual handles.
"Because we only need to pass _virtual_ identity of the proxy-promise which will be created on the other side".

We still support case when target method returns synchronously resolved `Task`/`Promise` or `null`. Only when caller invokes the `JSImport`/`JSExport` synchronously. Which is currently always. 

`null` value and synchronous resolve/reject would not be possible return value of real async cross-thread MT dispatch. There are currently WS client scenarios which are using synchronous `null` as signal. That would need to be fixed on another PR.

# Changes

- pre-allocate `PromiseHolder` for async `JSImport` in `InvokeJSImportImpl`
- rename `CreateJSOwnedHolder` to `GetPromiseHolder`
- renames back from `gcvHandle` to `gcHandle` where both could be used
- rename `InvokeJSImpl` to `InvokeJSFunction`
- rename `mono_wasm_invoke_bound_function` to `mono_wasm_invoke_js_function`
- use `OwnerTID` in `JSObject` instead of `ManagedThreadId`
- make `JSObject.Dispose` asynchronously `Post`ed to thread of affinity.
- added async methods to the advanced sample


src/libraries/System.Runtime.InteropServices.JavaScript/System.Runtime.InteropServices.JavaScript.sln
```diff
@@ -1,4 +1,8 @@
-﻿Microsoft Visual Studio Solution File, Format Version 12.00
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.8.34004.107
```
kg, line 4: Are all these sln changes intentional?

src/libraries/System.Runtime.InteropServices.JavaScript/src/System/Runtime/InteropServices/JavaScript/JSHostImplementation.Types.cs
```diff
@@ -12,20 +12,32 @@ internal static partial class JSHostImplementation
 
         public sealed class PromiseHolder
         {
-            public nint GCVHandle;
+            public nint GCHandle;
```
kg, line 15: why is this nint instead of a GCHandle?

src/libraries/System.Runtime.InteropServices.JavaScript/src/System/Runtime/InteropServices/JavaScript/JSHostImplementation.Types.cs
```diff
@@ -12,20 +12,32 @@ internal static partial class JSHostImplementation
...
             public ToManagedCallback? Callback;
 #if FEATURE_WASM_THREADS
             // the JavaScript object could only exist on the single web worker and can't migrate to other workers
-            internal int OwnerThreadId;
-            internal SynchronizationContext? SynchronizationContext;
+            internal JSSynchronizationContext SynchronizationContext;
+#endif
+
+#if FEATURE_WASM_THREADS
+            public PromiseHolder(JSSynchronizationContext targetContext)
```
kg, line 23: I feel like having a different signature for this per-ifdef is going to cause an explosion of ifdef'd code everywhere, when we could just have CurrentJSSynchronizationContext exist and return null

src/mono/wasm/runtime/invoke-cs.ts
```diff
@@ -242,14 +319,22 @@ function bind_fn(closure: BindingClosure) {
                     marshaler(args, js_arg);
                 }
             }
+            let js_result = undefined;
+            if (is_async && res_converter) {
+                // pre-allocate the promise
+                js_result = res_converter(args);
```
kg, line 325: What does it mean if it's async but there's no res_converter? shouldn't there still be a pre-allocated promise? the is_async below doesn't check for res_converter

src/mono/wasm/runtime/marshal-to-cs.ts
```diff
@@ -315,13 +315,16 @@ function _marshal_task_to_cs(arg: JSMarshalerArgument, value: Promise<any>, _?:
     }
     mono_check(isThenable(value), "Value is not a Promise");
 
-    const gc_handle = alloc_gcv_handle();
-    set_gc_handle(arg, gc_handle);
-    set_arg_type(arg, MarshalerType.Task);
+    const handleIsPreallocated = get_arg_type(arg) == MarshalerType.Task;
```
kg, line 318: Why is this `MarshalerType.Task` and not `MarshalerType.TaskPreCreated`?

src/mono/wasm/runtime/marshal-to-js.ts
```diff
@@ -216,11 +217,58 @@ function _marshal_delegate_to_js(arg: JSMarshalerArgument, _?: MarshalerType, re
...
 export function marshal_task_to_js(arg: JSMarshalerArgument, _?: MarshalerType, res_converter?: MarshalerToJs): Promise<any> | null {
+    // this path is used only when Task is passed as argument to JSImport and virtual JSHandle would be used
+
+    // if there is synchronous result, return it
+    const promise = sync_marshal_task_to_js(arg, res_converter);
+    if (promise !== false) {
+        return promise;
+    }
+
+    const jsv_handle = get_arg_js_handle(arg);
```
kg, line 233: i don't like that `get_arg_js_handle` returns a jsv_handle, yet `mono_wasm_get_js_handle` returns a js_handle. it's magical and confusing to the reader that there's no type system or naming convention hints here to make it evident whether a given piece of code is correct. i don't expect you to fix that in this PR though

src/mono/wasm/runtime/marshal-to-js.ts
```diff
@@ -364,6 +407,7 @@ function _marshal_js_object_to_js(arg: JSMarshalerArgument): any {
     }
     const js_handle = get_arg_js_handle(arg);
     const js_obj = mono_wasm_get_jsobj_from_js_handle(js_handle);
+    mono_assert(js_obj, "Expected JS object instance");
```
kg, line 410: A better message for this might be good if we expect it to ever be hit in production

kg (approved):
Other than my comments I don't see anything blocking this, but I don't feel confident at all in my understanding of jsv/gcv vs js/gc handles or the correctness of the code using them

src/mono/wasm/runtime/invoke-cs.ts
pavelsavara: I was just lazy. I will uses `!` to convince type system at compile time.

src/mono/wasm/runtime/marshal-to-cs.ts
pavelsavara: Good point, it will clarify things further.

src/libraries/System.Runtime.InteropServices.JavaScript/System.Runtime.InteropServices.JavaScript.sln
pavelsavara: VS wants to do them any time I open the solution. So I'm thinking I should not fight it anymore.
I will drop the versions, that's not necessary.

src/libraries/System.Runtime.InteropServices.JavaScript/src/System/Runtime/InteropServices/JavaScript/JSHostImplementation.Types.cs
pavelsavara: Because sometimes it's a GCVHandle (virtual one). I will add comment.

src/mono/wasm/runtime/marshal-to-js.ts
pavelsavara: Hmm, re-reading your comment, you are right. `JSVHandle` is special type of `JSHandle` on logical level. (recognized by negative values). We currently only use it pro promises/tasks.

pavelsavara: The usage of `TaskPreCreated` allowed me to make usage bit clearer. But still they are (and should be) interchangeable.

pavelsavara:
There is unrelated `System.Collections.Concurrent.Tests` in WASM MT.

PR #94485, opened 2023-11-07
Report StackOverflowException on NativeAOT on Linux

jtschuster:
Moved https://github.com/dotnet/runtime/pull/93901 here. Pulling main messed up something, so just going to make a new PR.

Uses similar code as CoreCLR to allocate an alternate stack for handling SIGSEGV, and properly reports a StackOverflowException.

Fixes https://github.com/dotnet/runtime/issues/82334

Updates the baseservices/exceptions/stackoverflow test to be one project and run on NativeAOT.

Issue #95414, opened 2023-11-29
[outerloop] [NativeAOT] Segmentation fault in System.Runtime.InteropServices.Tests


carlossanlop:
### Error Blob

```json
{
  "ErrorMessage": "Segmentation fault      (core dumped) ./System.Runtime.InteropServices.Tests -notrait category=IgnoreForCI -notrait category=OuterLoop",
  "BuildRetry": false,
  "ErrorPattern": "",
  "ExcludeConsoleLog": true
}
```

### Reproduction Steps

@dotnet/ilc-contrib @dotnet/interop-contrib PTAL

- Affected PR: https://github.com/dotnet/runtime/pull/94868 via `/azp run runtime-nativeaot-outerloop`
- Build: https://dev.azure.com/dnceng-public/public/_build/results?buildId=483727
- Job: https://dev.azure.com/dnceng-public/public/_build/results?buildId=483727&view=logs&j=20e6f673-0b0c-5c85-377c-5d941d4c8ee3&t=5801e1cc-59c5-5e47-8484-142b970afd58&l=55
- Queue: `linux_musl-x64 Release NativeAOT_Libs`
- Log file: https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-94868-merge-67c5e1f44298409881/System.Runtime.InteropServices.Tests/1/console.cb42cc43.log?helixlogtype=result
- Output: 
```
===========================================================================================================
/root/helix/work/workitem/e /root/helix/work/workitem/e
DOTNET_DbgEnableMiniDump is set and the createdump binary does not exist: ./createdump
Running assembly:System.Runtime.InteropServices.Tests, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51
[SKIP] System.Runtime.InteropServices.Tests.GetHINSTANCETests.GetHINSTANCE_ModuleBuilder_ReturnsSameInstance
[SKIP] System.Runtime.InteropServices.Tests.Int16Tests.WriteInt16_NotReadable_ThrowsArgumentException
[SKIP] System.Runtime.InteropServices.Tests.Int16Tests.ReadInt16_NotReadable_ThrowsArgumentException
[SKIP] System.Runtime.InteropServices.Tests.CreateWrapperOfTypeTests.CreateWrapperOfType_NullType_ThrowsArgumentNullException
[...]

Issue #94856, opened 2023-11-16
SIGSEGV in System.Numerics.Vectors


MichalStrehovsky:
Saw this here: https://helixre107v0xdcypoyl9e7f.blob.core.windows.net/dotnet-runtime-refs-pull-94756-merge-dc464ce7a6094aecb0/System.Numerics.Vectors.Tests/1/console.d55d86b1.log?helixlogtype=result

```
/root/helix/work/workitem/e /root/helix/work/workitem/e
Running assembly:System.Numerics.Vectors.Tests, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51
./RunTests.sh: line 204:    23 Segmentation fault      (core dumped) ./System.Numerics.Vectors.Tests -notrait category=IgnoreForCI -notrait category=OuterLoop -notrait category=failing -xml testResults.xml $RSP_FILE
/root/helix/work/workitem/e
----- end Thu 16 Nov 2023 03:38:07 AM UTC ----- exit code 139 ----------------------------------------------------------
exit code 139 means SIGSEGV Illegal memory access. Deref invalid pointer, overrunning buffer, stack overflow etc. Core dumped.
```

I pulled down the dump, the stack looks like this:

```
00 00007f00`3e9d9378 00007f40`d401e140     System_Numerics_Vectors!SIGSEGVHandler
01 00007f00`3e9d9380 00005616`bfd52584     libpthread_2_31!funlockfile+0x50
02 00007f00`36ffb520 00005616`bfbff1a9     System_Numerics_Vectors!S_P_CoreLib_System_Numerics_Vector_1<Int8>::CopyTo+0x4 [/_/src/libraries/System.Private.CoreLib/src/System/Numerics/Vector_1.cs @ 681] 
03 00007f00`36ffb530 00005616`bff93f79     System_Numerics_Vectors!xunit_assert_Xunit_Assert::RecordException+0x29 [/_/src/Microsoft.DotNet.XUnitAssert/src/Record.cs @ 37] 
04 00007f00`36ffb550 00005616`bff02f34     System_Numerics_Vectors!xunit_assert_Xunit_Assert::Throws+0x29 [/_/src/Microsoft.DotNet.XUnitAssert/src/ExceptionAsserts.cs @ 100] 
[...]

PR #95415, opened 2023-11-29
Revert "Report StackOverflowException on NativeAOT on Linux"

jtschuster:
Reverts dotnet/runtime#94485

We're hitting seg faults instead of null reference exceptions in tests after this change. Reverting to make sure this caused the issue, and then investigate.


jkotas (approved):

jkotas:
Related issues: #95414 #95414 #94856

PR #95432, opened 2023-11-30
Migrate to the 'locker' GitHub action for locking closed/stale issues/PRs

jeffhandley:
This migrates away from the FabricBot implementation of a scheduled search over closed issues and adopts the [Locker](https://github.com/microsoft/vscode-github-triage-actions/?tab=readme-ov-file#locker) GitHub action authored by the VS Code team.

The updated fabricbot.json was generated via https://github.com/dotnet/fabricbot-config/pull/83. The workflow file was implemented and tested at https://github.com/dotnet/fabricbot-config/blob/main/.github/workflows/locker.yml.

Since this workflow uses the actions/checkout action, we need to ensure the following GitHub Actions settings are enabled:

1. Allow enterprise, and select non-enterprise, actions and reusable workflows
2. Allow actions created by GitHub


.github/workflows/locker.yml
```diff
@@ -0,0 +1,36 @@
+name: Locker - Lock stale issues and PRs
+on:
+  schedule:
+    - cron: '0 9 * * *' # Once per day, early morning PT
```
jeffhandley, line 4:
```suggestion
    - cron: '37 8 * * *' # Once per day, early morning PT
```

@terrajobst pointed out that GitHub suggests not using `:00` on cron jobs because it's so commonly used. Picking an arbitrary time.

bartonjs (approved):

PR #95443, opened 2023-11-30
[mono][jit] Fix the bblock check in the box+callvirt optimization.

### LINKED ISSUE #95347: [mono] box+callvirt optimization incorrect when instructions are in different basic blocks
The `box+callvirt` optimization from https://github.com/dotnet/runtime/pull/73456 is firing incorrectly when the `box` and the `callvirt` are consecutive IL instructions but in different basic blocks and there is another basic block that jumps to the `callvirt`.

See https://github.com/dotnet/runtime/issues/95347#issuecomment-1832878361 and the next 3 comments for repro and diagnosis

Original issue description below:

----

### Description

Runtime behaviour changes in net8.0.
In net8.0 when running the code under/linked repo in Release mode on Mono (Android, ios, wasm) vs Debug it behaves different.
It also behaves different to release mode of a windows console app. 

In net7.0 - both of them worked the same, in both debug and release mode

`(nullableObject ?? false).ToString()`

### Reproduction Steps

1. Checkout code - https://github.com/iXyles/net8-regression-bug (or copy the code part from it)
2. Run console app in Release mode Should print 'False': False
3. Run Andrioid app in Release mode Should print same, but instead it prints 'False': true

Optional:
4. Uninstall app from device
5. Run app in Debug mode - now it shows 'False': False (as expected)

### Expected behavior

Should work the same as in previous versions.
Debug & Release mode should behave the same with the given code.
Output should be the same in Android, iOS & Windows console app

### Actual behavior

Returns different values in Debug vs Release mode & works different on different platforms and previous net versions
Output value on Android & iOS: `true`
Output value on Windows console app: `FALSE`

### Regression?

This works with previous net7.0 version (latest)

### Known Workarounds

`nullableObject != null ? nullableObject.ToString() : false.ToString()` (in this sample)
or
Downgrade back to net7.0

### Configuration

Android - Unsure
iOS - ARM64

### Other information

_No response_

vargaz:
Fixes https://github.com/dotnet/runtime/issues/95347.

### DIFF
--- a/src/mono/mono/mini/method-to-ir.c
+++ b/src/mono/mono/mini/method-to-ir.c
@@ -9792,12 +9792,11 @@ calli_end:
 			//   callvirt instance ISomeIface::Method()
 			// TO
 			//   call SomeStruct::Method()
-			guchar* callvirt_ip;
 			guint32 callvirt_proc_token;
 			if (!((cfg->compile_aot || cfg->compile_llvm) && !mono_class_is_def(klass)) && // we cannot devirtualize in AOT when using generics
 				next_ip < end &&
-				(callvirt_ip = il_read_callvirt (next_ip, end, &callvirt_proc_token)) && 
-				ip_in_bb (cfg, cfg->cbb, callvirt_ip) ) {
+				il_read_callvirt (next_ip, end, &callvirt_proc_token) &&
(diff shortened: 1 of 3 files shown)


vargaz:
Still needs a test.

lambdageek (approved):

lambdageek:
Please backport to net8

PR #95456, opened 2023-11-30
[LoongArch64] add LoongArch64's SIMD128/256 instructions within emitter.

shushanhf:
This is the first PR to adding LoongArch64's SIMD128/256 within the JIT.

Add LoongArch64's SIMD128/256 instructions within emitter.


shushanhf:
Hi, @jakobbotsch 
Could you please review this PR？
Thanks

jakobbotsch:
@shushanhf Thanks, I'll try to take a look soon, but it's a large PR and I need to set aside some time to do it. Hopefully I should be able to do that by the end of this week or at the beginning of the next one.

shushanhf:
> @shushanhf Thanks, I'll try to take a look soon, but it's a large PR and I need to set aside some time to do it. Hopefully I should be able to do that by the end of this week or at the beginning of the next one.

OK, Thanks!

JulieLeeMSFT:
CC @BruceForstall.

src/coreclr/jit/instrsloongarch64.h
```diff
@@ -484,6 +484,1520 @@ INST(fld_d,         "fld.d",          LD,   0x2b800000)
...
+INST(vslt_w,            "vslt.w",           0,  0x70070000)
+INST(vslt_d,            "vslt.d",           0,  0x70078000)
+INST(vslt_bu,           "vslt.bu",          0,  0x70080000)
+INST(vslt_hu,           "vslt.hu",          0,  0x70088000)
+INST(vslt_wu,           "vslt.wu",          0,  0x70090000)
+INST(vslt_du,           "vslt.du",          0,  0x70098000)
+INST(vadd_b,            "vadd.b",           0,  0x700a0000)
+INST(vadd_h,            "vadd.h",           0,  0x700a8000)
+INST(vadd_w,            "vadd.w",           0,  0x700b0000)
+INST(vadd_d,            "vadd.d",           0,  0x700b8000)
```
BruceForstall, line 581: I presume each of these instructions is a variant of `vadd` with a different SIMD lane size. Why did you decide to implement them all as separate instructions instead of a single INS_vadd instruction with a lane size "option", like other platforms do? One result of the way it is done here is a lot of code like:
```
(INS_vreplgr2vr_b <= ins) && (ins <= INS_vreplgr2vr_d)
```
instead of a simple:
```
(ins == INS_vreplgr2vr)
```

And, every single SIMD instruction needs to be listed 4 times.

BruceForstall: I see several concerns with having so many ranges in the code:
1. It's difficult when reading the code to understand the set of instructions in any given range.
2. It makes it impossible to rearrange the instruction list.

It some cases (many cases?) maybe it would be useful to have named small helper functions that make it more obvious?

src/coreclr/jit/emitloongarch64.cpp
```diff
@@ -5029,88 +5433,87 @@ void emitter::emitDisInsName(code_t code, const BYTE* addr, instrDesc* id)
...
-            const char* fd = RegNames[(code & 0x1f) + 32];
-            const char* fj = RegNames[((code >> 5) & 0x1f) + 32];
-            const char* fk = RegNames[((code >> 10) & 0x1f) + 32];
-            const char* fa = RegNames[((code >> 15) & 0x1f) + 32];
             printf("fnmsub.d     %s, %s, %s, %s\n", fd, fj, fk, fa);
             return;
-        }
+
+#ifdef FEATURE_SIMD
+        case 0x91:
```
BruceForstall, line 5468: Aren't there `#define` values / names for these numbers?

I'm surprised the disassembly output has all these hard-coded text names instead of having the text come from the instruction table. If they came from a table, then all this code would reduce to a couple lines, since they have the same set of arguments (for `v` and `xv` cases).

BruceForstall: I guess this comment applies to this entire disassembler, which seems insufficiently data driven. So it's nothing new.

There are an extraordinary number of "magic numbers" in the code.

BruceForstall: It seems like this whole thing could be written much better, more data driven, and not require an additional 10,000 lines of code.

src/coreclr/jit/instrsloongarch64.h
MichalPetryka: > I presume each of these instructions is a variant of `vadd` with a different SIMD lane size.

It's `b` for byte, `h` for halfword, `w` for word and `d` for doubleword AFAIR.

shushanhf: > > I presume each of these instructions is a variant of `vadd` with a different SIMD lane size.
> 
> It's `b` for byte, `h` for halfword, `w` for word and `d` for doubleword AFAIR.

Yes, you are right.

shushanhf: > I presume each of these instructions is a variant of `vadd` with a different SIMD lane size. Why did you decide to implement them all as separate instructions instead of a single INS_vadd instruction with a lane size "option", like other platforms do? One result of the way it is done here is a lot of code like:
> ```
> (INS_vreplgr2vr_b <= ins) && (ins <= INS_vreplgr2vr_d)
> ```

Hi, @BruceForstall 
First of all, thanks for your reviewing.
Even we use a single SIMD instruction `vadd` to express the four `vadd_b/h/w/d`, here the `(INS_vreplgr2vr_b <= ins) && (ins <= INS_vreplgr2vr_d)`  is the same, because we encode the simlar type instructions together not only the `vadd_b/h/w/d`, that is here using `vadd` or `vadd_b/h/w/d` is the same.

Besides, not all SIMD instructions are `b/h/w/d`, using the `b/h/w/d` separately is more exact and there is no need to add size "option" to avoid extra encoding or decoding.

> instead of a simple:
> 
> ```
> (ins == INS_vreplgr2vr)
> ```
> 
> And, every single SIMD instruction needs to be listed 4 times.

src/coreclr/jit/emitloongarch64.cpp
shushanhf: Thanks.
As the ISA's encoding is very stable and only used for debugging, only pushed one time, here using the explicit ‘‘magic numbers’’.
If needed, we can amend these further.

shushanhf: @BruceForstall 
I will reduce the lines for this disasm function by pushing a patch later.

shushanhf: Hi, @BruceForstall 

The previous LoongArch64's implementation of the `emitDisInsName()` is so large as we disassemble from the memory and decoding one by one fields.

For ARM64 and X86, issuing the instruction from the `instrDesc` to the memory, while the disassembly is also from the `instrDesc` other than from the memory.
BTW, I think disassembling from the memory is better.

I have redesigned the `emitDisInsName()`.
Thanks very much for your reviewing and advices!!!

BruceForstall (approved):
LGTM. Thanks for the changes. I hope the updated instruction dumper proves useful.

Issue #42820, opened 2020-09-28
Standardize pattern for exposing advanced configuration for compression streams


terrajobst:
There is [a desire](https://github.com/dotnet/runtime/issues/2236#issuecomment-699927812) to expose compression-algorithm specific configuration when constructing compression streams. The same options can be used for `ZLibStream`, `DeflateStream`, `GZipStream`. Named those options with `Zlib` prefix same as the native library name they use another prefix we considered is `Deflate` - the algorithm name, the prefix needed because we have existing `CompressionLevel` enum that used for all compression streams including `BrotliStream` and it cannot be used for fine tuning the compression level for specific stream
https://github.com/dotnet/runtime/blob/d710379da9b2699bba0a8f26d65d3a6e8549f26a/src/libraries/System.IO.Compression/ref/System.IO.Compression.cs#L9-L15 

### API Proposal
```C#
namespace System.IO.Compression
{
    public enum ZlibCompressionLevel
    {
        DefaultCompression = -1,
        NoCompression = 0,
        Level1 = 1,
        BestSpeed = Level1,
        Level2 = 2,
        Level3 = 3,
        Level4 = 4,
        Level5 = 5,
        Level6 = 6,
        Level7 = 7,
        Level8 = 8,
        Level9 = 9,
        BestCompression = Level9,
    }
    public enum ZlibCompressionStrategy
    {
        DefaultStrategy = 0,
        Filtered = 1,
        HuffmanOnly = 2,
        Rle = 3,
        Fixed = 4
    }
    public enum ZlibFlushMode
    {
        NoFlush = 0,
        PartialFlush = 1,
        SyncFlush = 2,
        FullFlush = 3,
        Finish = 4,
    }
    public sealed class ZLibStream : Stream
    {
        // CompressionMode.Compress
        public ZLibStream(Stream stream, ZlibCompressionLevel compressionLevel = ZlibCompressionLevel.DefaultCompression,
            ZlibCompressionStrategy strategy = ZlibCompressionStrategy.DefaultStrategy, bool leaveOpen = false);
        public ZlibFlushMode FlushMode { get; set; }
    }
    public partial class DeflateStream : Stream
    {
        public DeflateStream(Stream stream, ZlibCompressionLevel compressionLevel = ZlibCompressionLevel.DefaultCompression,
            ZlibCompressionStrategy strategy = ZlibCompressionStrategy.DefaultStrategy, bool leaveOpen = false);
[...]

PR #95460, opened 2023-11-30
Changing compression level validation to allow any valid value

davemanton:
Compression level was validating to restrict to values represented in the compression level enum

However there are other values which are appropriate and users may want to use

This PR enables the usage of any compression level lieing between the defined minimum and maximum and throws the existing exception if the value set sits outside the valid range

### DIFF
--- a/src/libraries/System.IO.Compression.Brotli/src/System/IO/Compression/BrotliUtils.cs
+++ b/src/libraries/System.IO.Compression.Brotli/src/System/IO/Compression/BrotliUtils.cs
@@ -20,7 +20,7 @@ namespace System.IO.Compression
                 CompressionLevel.Fastest => 1,
                 CompressionLevel.Optimal => Quality_Default,
                 CompressionLevel.SmallestSize => Quality_Max,
-                _ when (int)compressionLevel is > Quality_Min and <= Quality_Max => (int)compressionLevel,
+                _ when (int)compressionLevel is >= Quality_Min and <= Quality_Max => (int)compressionLevel,
                 _ => throw new ArgumentException(SR.ArgumentOutOfRange_Enum, nameof(compressionLevel))
             };
     }
--- a/src/libraries/System.IO.Compression.Brotli/src/System/IO/Compression/BrotliUtils.cs
+++ b/src/libraries/System.IO.Compression.Brotli/src/System/IO/Compression/BrotliUtils.cs
@@ -20,6 +20,7 @@ namespace System.IO.Compression
                 CompressionLevel.Fastest => 1,
                 CompressionLevel.Optimal => Quality_Default,
                 CompressionLevel.SmallestSize => Quality_Max,
+                _ when (int)compressionLevel is > Quality_Min and <= Quality_Max => (int)compressionLevel,
                 _ => throw new ArgumentException(SR.ArgumentOutOfRange_Enum, nameof(compressionLevel))
             };
     }
--- a/src/libraries/System.IO.Compression.Brotli/tests/CompressionStreamUnitTests.Brotli.cs
+++ b/src/libraries/System.IO.Compression.Brotli/tests/CompressionStreamUnitTests.Brotli.cs
@@ -43,7 +43,7 @@ namespace System.IO.Compression
         [InlineData((CompressionLevel)5)]
         public void Ctor_ArgumentValidation_ValidCompressionLevel(CompressionLevel compressionLevel)
         {
-            Assert.DoesNotThrow(() => new BrotliStream(new MemoryStream(), compressionLevel));
+            _ = new BrotliStream(new MemoryStream(), compressionLevel);
         }
 
         [Fact]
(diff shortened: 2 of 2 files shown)


davemanton:
@dotnet-policy-service agree

stephentoub:
Thank you, but this validation is by design:
https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/7.0/brotlistream-ctor
There are open issues I hope will be addressed in .NET 9 to add more knobs here, but in the meantime you can use BrotliEncoder directly.

davemanton:
@stephentoub

Thank you for providing the documentation.  Respectfully, I believe this PR is inline with the expectations of the design

The documentation suggests that allowing an arbitrary level to the underlying library could result in inconsistent and unexpected behaviour, this PR is designed to avoid that and allow values within a defined valid range

For us this is a blocker to progressing to dotnet 8 because we've chosen a valid compression level - 5 which for us provides the correct trade off between speed and payload size.  A larger payload would affect our mobile users experience and payload is something we pay close attention to

We are a very large e-commerce retailer in the UK but for us writing our own custom compression middleware isn't something I want our devs to get involved with when they don't have knowledge or experience of that area.  The existing middleware is great, I don't really understand why we can't pass it a valid (not arbitrary) value

As a listed company we try to avoid versions of dotnet which aren't LTS, so you're effectively asking us to wait 2 years to get our payload size back to where it should be, or write an encoder middleware when there's a perfectly good one available and we're not really qualified to do it

stephentoub:
Thanks, I understand the concerns. Ideally the bug fix around validation would have happened at the same time as adding additional ctors, but alas, they were split from each other, one happened, one didn't. cc: @jeffhandley 

If level 4 instead of 5 would meet your needs, you can choose Optimal. Or you can write a custom Stream for when you need this specific encoding level.

The problem with allowing arbitrary values through isn't about the min/max but about the numbers not always meaning what the caller thinks they might mean, eg if you pass 0, you don't get level 0, if you pass 3 you don't get 3, etc. This ctor works in terms of a specific mapping; that at one point it accidentally interpreted the input differently was a bug and one we don't want to reintroduce.

I'm not arguing there isn't a need for additional support here, I agree there is, just that this ctor is not the answer.

davemanton:
I remember reading the issue at the time when we chose our level and we ran foul of optimal not meaning optimal when the Quality_Default was set at maximum compression - https://github.com/dotnet/runtime/issues/26097

The changes that have been made since then much better describe optimal and meet the requirements of the enum describing min/max and fastest through applicable values

Please could you clarify why an integer value doesn't give you the level you're asking for on a sliding scale?  I appreciate this is a lack of knowledge on my part but I'm unclear why that would be the case

stephentoub:
> Please could you clarify why an integer value doesn't give you the level you're asking for on a sliding scale? I appreciate this is a lack of knowledge on my part but I'm unclear why that would be the case

If it actually took an int, it'd be fine, and that's what a future ctor would accept. But it doesn't, it accepts a CompressionLevel, which the implementation then interprets and maps to an integer value. There's no way for the implementation to distinguish, for example, between `CompressionLevel.SmallestSize` and `(CompressionLevel)3`; at the IL level, those are identical. And as such, if you specify 3, you actually get 11, because that's what SmallestSize maps to.

davemanton:
Okay, understood because of the CompressionLevel enum having those values predefined and therefore it getting mapped there even when you pass through an integer

I think to be fair to most users who set an integer value themselves they're probably going to experiment and realise there's a problem.  Removing the ability to set an arbitrary value to stop 0-3 from used incorrectly when there's the majority of the scale left is a bit much

I understand that Microsoft doesn't necessarily want the issues raising that it doesn't do what's expected with an integer level, but not using the enum is not using it as described so a user shouldn't expect it to be an exact correlation and the enum is easy enough to view to understand why

Blocking this functionality effectively for 2 years is really bad for users like us.  If you count from dotnet 7 -> the next LTS at dotnet 10, it's 3 years.  Surely we're better off allowing users to put in a valid value provided it falls within the correct range rather than stopping them having the compression level they want/need for so long

davemanton:
@stephentoub 

I'd like to offer an alternative through this PR which enables an integer specific constructor https://github.com/dotnet/runtime/pull/95467

This way I can use this to create a compression provider which more closely matches the original brotli provider without needing a lot of knowledge of compression - https://github.com/dotnet/aspnetcore/blob/474d2944dba601440987b328d7044155023c3c14/src/Middleware/ResponseCompression/src/BrotliCompressionProvider.cs

stephentoub:
> an integer specific constructor #95467
> 
> This way I can use this to create a compression provider which more closely matches the original brotli provider without needing a lot of knowledge of compression - https://github.com/dotnet/aspnetcore/blob/474d2944dba601440987b328d7044155023c3c14/src/Middleware/ResponseCompression/src/BrotliCompressionProvider.cs

Thanks. That is the planned approach:
https://github.com/dotnet/runtime/issues/42820
The actual shape of the APIs still needs to be agreed upon before we can accept a PR for it.

davemanton:
No problem, thanks for your help today

PR #95469, opened 2023-11-30
[Mono] Stop falling back to interpreter for runtime invoke

fanyang-mono:
This PR stops most of the runtime invoke calls falling back to interpreter when `MONO_AOT_MODE_LLVMONLY_INTERP` mode is set. `MONO_AOT_MODE_LLVMONLY_INTERP` mode is set only when one of these two switches is passed to Mono AOT:  

- --llvmonly
- --llvmonly-interp

WASM AOT is the only major target uses it. It was setup here:
https://github.com/dotnet/runtime/blob/7bbfcc40061d4a542db168395ca3f6e2687cc09a/src/mono/wasm/build/WasmApp.Native.targets#L665

This PR should resolve some of, if not all of, the performance regression (https://github.com/dotnet/perf-autofiling-issues/issues/24928) introduced by enabling `WasmStripILAfterAOT` by default.

This PR also contains a fix to the problem with Mono LLVM AOT when a runtime invoke calling a method with delegate parameters. The delegate parameter wasn't handled properly before.

The following C# program is a demonstration of the problem
```c#
public class Test
{
    unsafe public static void Main(string[] args)
    {
        void* fn = FunctionPointerMethods.GetFunctionPointer();
        MethodInfo m = typeof(FunctionPointerMethods).GetMethod(nameof(FunctionPointerMethods.CallFcnPtr_FP));
        m.Invoke(null, new object[] { (IntPtr)fn, 42 });  <--- Error out
    }
}

public static class FunctionPointerMethods
{
    public static bool CallMe(int i)
    {
        return i == 42;
    }

    public static unsafe bool CallFcnPtr_FP(delegate*<int, bool> fn, int value)
    {
        return fn(value);
    }

    public static unsafe delegate*<int, bool> GetFunctionPointer() => &CallMe;
}
``` 

cc: @lewing


fanyang-mono:
All CI failures are known and unrelated to this PR.

src/mono/mono/mini/mini-runtime.c
```diff
@@ -3401,7 +3401,7 @@ mono_llvmonly_runtime_invoke (MonoMethod *method, RuntimeInvokeInfo *info, void
 	for (i = 0; i < sig->param_count; ++i) {
 		MonoType *t = sig->params [i];
 
-		if (!m_type_is_byref (t) && (MONO_TYPE_IS_REFERENCE (t) || t->type == MONO_TYPE_PTR)) {
+		if (!m_type_is_byref (t) && (MONO_TYPE_IS_REFERENCE (t) || t->type == MONO_TYPE_PTR || t->type == MONO_TYPE_FNPTR)) {
 			param_refs [i] = params [i];
```
SamMonoRT, line 3405: does this also affect cases in iOS where interpreter fall back happens? @BrzVlad @vargaz

fanyang-mono: I tested locally. Full aot didn't go through this codepath.

vargaz (approved):

PR #95470, opened 2023-11-30
[RISC-V] Add X0 field init at capture context.

viewizard:
Related to issue in managed stepping (related PR: https://github.com/dotnet/runtime/pull/94853), where during stepping we use stored X0 that not zero, that ruin NativeWalker and SW single stepper logic (for example, at conditional branch, where X0 could be used as one of register).

CC @clamp03 @wscho77 @HJLeee @JongHeonChoi @t-mustafin @gbalykov @brucehoult

### DIFF
--- a/src/coreclr/pal/src/arch/riscv64/context2.S
+++ b/src/coreclr/pal/src/arch/riscv64/context2.S
@@ -160,6 +160,7 @@ LOCAL_LABEL(Done_CONTEXT_CONTROL):
     ld  t1, 8(sp)
     ld  t3, 16(sp)
 
+    sd  x0, (CONTEXT_X0)(a0)
     sd  tp, (CONTEXT_Tp)(a0)
     sd  gp, (CONTEXT_Gp)(a0)
     sd  a0, (CONTEXT_A0)(a0)


gbalykov (approved):

src/coreclr/pal/src/arch/riscv64/context2.S
```diff
@@ -156,6 +156,9 @@ LOCAL_LABEL(Done_CONTEXT_CONTROL):
     and  t3, t1, t0
     bne  t3, t0, LOCAL_LABEL(Done_CONTEXT_INTEGER)
 
+    li  t0, 0
+    sd  t0, (CONTEXT_X0)(a0)
+
```
clamp03, line 161: My suggestion is 
```
     ld  t1, 8(sp)
     ld  t3, 16(sp)
 
+    sd  x0, (CONTEXT_X0)(a0)
     sd  tp, (CONTEXT_Tp)(a0)
     sd  gp, (CONTEXT_Gp)(a0)
     sd  a0, (CONTEXT_A0)(a0)
```
If I misunderstand your code, please let me know.
Thank you.

viewizard: Same logic but much better (for some reason I thought `sd` don't work with `x0` directly and don't even check it -_-). Thanks!

clamp03 (approved):

jkotas:
Would it be better to drop the zero register from the CONTEXT?

brucehoult:
> Would it be better to drop the zero register from the CONTEXT?

There's certainly no point in saving and attempting to restore it, since it is immutably 0.

gbalykov:
> Would it be better to drop the zero register from the CONTEXT?

@jkotas removing x0 from context will break logic relying on registers indexing, where integer number of register is used (e.g. in `RiscV64SingleStepper::GetReg`, etc.). This change is rather simple and fixes icordebug api stepping, and we've tested that stepping works with this change in netcoredbg. Removal of x0 from context will be more complex, so how about adding `TODO-RISCV64` for further improvement, and keeping this change as is to fix icordebug stepping for now?

jkotas:
How many places do index from X0? It should be a very few and they should be easy to find by just building with X0 removed.

gbalykov:
> How many places do index from X0? It should be a very few and they should be easy to find by just building with X0 removed.

Removal will complicate places like `(&pCtx->X0)[reg]`. We'll have to add conditions there or use some numbers mapping, so not sure it's worth it to complicate things in other places.

jkotas:
The zero register needs special casing in number of places. I think it is likely that there are more places where this special casing is missing. If the zero register is omitted from CONTEXT, these places are going to be more apparent.

I agree that this is small enough change that it can be merged.

jkotas (approved):

gbalykov:
Thank you!

Issue #81358, opened 2023-01-30
NativeAOT Data flow annotations are not applied to DI created generic types


vitek-karas:
This only occurs in NativeAOT if we stop doing generic parameter data flow validation
from generic dictionaries and instead switch to the approach taken in https://github.com/dotnet/runtime/pull/80956.

## Problem description

If there's a service with a generic parameter which has data flow annotations (`DynamicallyAccessedMembers`)
and such a service is injected into the application due to some other service having a `.ctor` parameter
of its type with specific instantiation, the annotation is not applied to the instantiation.

This can lead to failures due to missing members or similar problems. This most readily reproes itself
in many of the `Microsoft.Extensions.*` tests. For example the `Microsoft.Extensions.Logging` tests
hit this problem because they have a logger factory type which takes `IOptions<LoggerOptions>` parameter
in its constructor. `IOptions<T>` has a data flow annotation on the `T`.

## Repro

Sample code which reproes the problem:
```C#
using System;
using System.Diagnostics.CodeAnalysis;
using System.Reflection;

namespace DIBreak
{
    public class Program
    {
        public static void Main()
        {
            Services services = new();
            services.RegisterService(typeof(INameProvider<>), typeof(NameProviderService<>));
            services.RegisterService(typeof(IDataObjectPrinter), typeof(DataObjectPrinterService));

            var printer = services.GetService<IDataObjectPrinter>();
            printer.Print(new DataObject());
        }
    }

    public class DataObject {
        public string Name => "Hello World";
    }

    // Simplistic implementation of DI which is comparable in behavior to our DI
    class Services {
        private Dictionary<Type, Type> _services = new Dictionary<Type, Type>();

        public void RegisterService(Type interfaceType, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type implementationType)
        {
            // TODO: Validate that implementationType implements interfaceType

            _services.Add(interfaceType, implementationType);
        }

        public T GetService<T>() {
            return (T)GetService(typeof(T));
        }

        public object GetService(Type interfaceType) {
            Type typeDef = interfaceType.IsGenericType ? interfaceType.GetGenericTypeDefinition() : interfaceType;
            Type implementationType = GetImplementationType(typeDef);

            if (implementationType.IsGenericTypeDefinition) {
                for (int i = 0; i < implementationType.GetGenericArguments().Length; i++) {
                    Type genericArgument = implementationType.GetGenericArguments()[i];
[...]

PR #95482, opened 2023-11-30
[ILLink analyzer] Move generic parameter analysis to dataflow

### LINKED ISSUE #95121: Trim analyzer: Doesn't produce warnings due to annotated generic parameter in various situations

sbomer:
And align behavior more closely with NativeAot. I tried to roughly follow the implementation approach used by NativeAot in GenericArgumentDataFlow.cs to make it easier to share the logic (without actually sharing the code for now).

Fixes https://github.com/dotnet/runtime/issues/95121.

Some things to note:
- Gets rid of some analyzer warnings in cases where NativeAot doesn't warn either
- Doesn't warn for generics in `typeof`
- Doesn't warn for generics in signatures of reflectable members (ILC does this to work around an incorrect suppression in DI, as discussed in https://github.com/dotnet/runtime/issues/81358).
- Generics in base types or interface types are still analyzed outside of dataflow


src/tools/illink/test/Mono.Linker.Tests.Cases/DataFlow/GenericParameterDataFlow.cs
```diff
@@ -406,7 +406,7 @@ class TypeGenericRequirementsOnMembers<[DynamicallyAccessedMembers (DynamicallyA
 		{
 			public TypeRequiresPublicFields<TOuter> PublicFieldsField;
 
-			[ExpectedWarning ("IL2091", nameof (TypeRequiresPublicMethods<TOuter>), ProducedBy = Tool.Trimmer | Tool.Analyzer)] // NativeAOT_StorageSpaceType
+			[ExpectedWarning ("IL2091", nameof (TypeRequiresPublicMethods<TOuter>), ProducedBy = Tool.Trimmer)] // NativeAOT_StorageSpaceType
```
vitek-karas, line 409: Nit: Maybe update the comment about NativeAOT_StorageSpaceType thing at the top of the file to also mention that we now expect the same behavior from the analyzer.

vitek-karas (approved):
Nice!

ShreyasJejurkar:
This is great. I just had one question can't we share the trimming logic between the trimmer, NativeAOT, and the rest of the app models where we do trimming? This is hard to do or is it technically impossible to do so?

vitek-karas:
The additional complication is that trimmer (illink) and NativeAOT see IL, while the analyzer sees the Roslyn's operations/CFG. We could probably share more if(when) we implement CFG reconstruction from IL in trimmer and NativeAOT. But currently the two approaches have to do some things differently.

That said, we already share quite a lot of the core data flow logic, and almost all of the tests.

Issue #53849, opened 2021-06-08
VC++ compiler optimization switches


BruceForstall:
`eng\native\configureoptimization.cmake` contains:
```
if(CLR_CMAKE_HOST_WIN32)                                                                        
    add_compile_options($<$<AND:$<COMPILE_LANGUAGE:C,CXX>,$<CONFIG:Debug>>:/Od>)                
    add_compile_options($<$<AND:$<COMPILE_LANGUAGE:C,CXX>,$<CONFIG:Checked>>:/O1>)              
    add_compile_options($<$<AND:$<COMPILE_LANGUAGE:C,CXX>,$<CONFIG:Release>>:/Ox>)              
    add_compile_options($<$<AND:$<COMPILE_LANGUAGE:C,CXX>,$<CONFIG:RelWithDebInfo>>:/O2>)       
```

I have some questions:
1. Why don't we use `/O2` for all non-Debug builds?
2. Is `RelWithDebInfo` actually used? For what?

@dotnet/runtime-infrastructure

jkoritzinsky:
cc: @jkotas @janvorli

jkotas:
It has been like that since we have open sourced: https://github.com/dotnet/coreclr/commits/release/2.1/src/pal/tools/windows-compiler-override.txt . I do not think we have history for this file prior to open sourcing.

The config for release is configuration is the only sensitive one, changing that would require performance testing. Feel free to clean it up and unify the rest. 

You may want to check impact of the checked build optimization settings on build performance. It is possible that `/O1` was choosen as a trade-off between build and runtime performance.

You can also change https://github.com/dotnet/runtime/blob/main/src/libraries/Native/Windows/CMakeLists.txt to include the central optimization settings file while you are on it.

BruceForstall:
cc @AndyAyersMS

BruceForstall:
Well, it's messier than this. If I force a compiler error in a win-x64 release build, I see a command line with these optimization switches (this is a subset of the compiler switches):
```
/O2
/Ob2
/Ox
/Oi
/Oy-
/Gy
```
[...]

PR #95484, opened 2023-11-30
JIT: Force inline small BasicBlock methods

amanasifkhalid:
Several small methods aren't inlined in very large methods (such as `Compiler::impImportBlockCode`) by MSVC, probably due to its inlining budget. Force-inlining some of these methods, starting with a few `BasicBlock` methods, may slightly improve TP (at least this was the case locally).

### DIFF
--- a/src/coreclr/jit/block.cpp
+++ b/src/coreclr/jit/block.cpp
@@ -1631,59 +1631,6 @@ BasicBlock* BasicBlock::New(Compiler* compiler, BBjumpKinds jumpKind, unsigned j
 }
 
 //------------------------------------------------------------------------
-// isBBCallAlwaysPair: Determine if this is the first block of a BBJ_CALLFINALLY/BBJ_ALWAYS pair
-//
-// Return Value:
-//    True iff "this" is the first block of a BBJ_CALLFINALLY/BBJ_ALWAYS pair
-//    -- a block corresponding to an exit from the try of a try/finally.
-//
-// Notes:
-//    In the flow graph, this becomes a block that calls the finally, and a second, immediately
-//    following empty block (in the bbNext chain) to which the finally will return, and which
-//    branches unconditionally to the next block to be executed outside the try/finally.
-//    Note that code is often generated differently than this description. For example, on ARM,
-//    the target of the BBJ_ALWAYS is loaded in LR (the return register), and a direct jump is
-//    made to the 'finally'. The effect is that the 'finally' returns directly to the target of
-//    the BBJ_ALWAYS. A "retless" BBJ_CALLFINALLY is one that has no corresponding BBJ_ALWAYS.
-//    This can happen if the finally is known to not return (e.g., it contains a 'throw'). In
-//    that case, the BBJ_CALLFINALLY flags has BBF_RETLESS_CALL set. Note that ARM never has
-//    "retless" BBJ_CALLFINALLY blocks due to a requirement to use the BBJ_ALWAYS for
-//    generating code.
-//
-bool BasicBlock::isBBCallAlwaysPair() const
-{
-    if (this->KindIs(BBJ_CALLFINALLY) && !(this->bbFlags & BBF_RETLESS_CALL))
-    {
-        // Some asserts that the next block is a BBJ_ALWAYS of the proper form.
-        assert(!this->IsLast());
-        assert(this->Next()->KindIs(BBJ_ALWAYS));
-        assert(this->Next()->bbFlags & BBF_KEEP_BBJ_ALWAYS);
-        assert(this->Next()->isEmpty());
-
-        return true;
-    }
-    else
-    {
-        return false;
-    }
-}
-
-//------------------------------------------------------------------------
-// isBBCallAlwaysPairTail: Determine if this is the last block of a BBJ_CALLFINALLY/BBJ_ALWAYS pair
-//
-// Return Value:
-//    True iff "this" is the last block of a BBJ_CALLFINALLY/BBJ_ALWAYS pair
-//    -- a block corresponding to an exit from the try of a try/finally.
-//
-// Notes:
-//    See notes on isBBCallAlwaysPair(), above.
-//
-bool BasicBlock::isBBCallAlwaysPairTail() const
-{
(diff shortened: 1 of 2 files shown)


amanasifkhalid:
cc @dotnet/jit-contrib. My intent was to limit this to roughly one-line helper methods, but I did see some non-inlined calls to `isBBCallAlwaysPair` and `isBBCallAlwaysPairTail`, so I decided to include them. [TP diffs](https://dev.azure.com/dnceng-public/public/_build/results?buildId=485388&view=ms.vss-build-web.run-extensions-tab) are surprisingly biggest on Linux x64, so I guess Clang was running into similar inlining issues?

If we think the TP diffs are worth it, I'll add some comments explaining the decision (and maybe a TODO to remove these eventually?).

src/coreclr/jit/block.cpp
```diff
@@ -1630,59 +1630,6 @@ BasicBlock* BasicBlock::New(Compiler* compiler, BBjumpKinds jumpKind, unsigned j
...
-//    Note that code is often generated differently than this description. For example, on ARM,
-//    the target of the BBJ_ALWAYS is loaded in LR (the return register), and a direct jump is
-//    made to the 'finally'. The effect is that the 'finally' returns directly to the target of
-//    the BBJ_ALWAYS. A "retless" BBJ_CALLFINALLY is one that has no corresponding BBJ_ALWAYS.
-//    This can happen if the finally is known to not return (e.g., it contains a 'throw'). In
-//    that case, the BBJ_CALLFINALLY flags has BBF_RETLESS_CALL set. Note that ARM never has
-//    "retless" BBJ_CALLFINALLY blocks due to a requirement to use the BBJ_ALWAYS for
-//    generating code.
-//
-bool BasicBlock::isBBCallAlwaysPair() const
```
SingleAccretion, line 1653: Moving methods from a `.cpp` file to a header is a takeback w.r.t. maintainability (the coding convention encourages placing methods in implementation files). Does it make a good enough difference to pay for that?

SingleAccretion (commented):
If you could collect a per-method TP diff, it would be interesting to see where this makes a difference.

src/coreclr/jit/block.cpp
amanasifkhalid: Locally on Windows x64, I observed an additional 0.01% or 0.02% TP improvement; technically a 25-33% improvement over not force-inlining these methods, but only because the TP improvements were small on Windows x64 to begin with. As for maintainability, I agree I'd rather not move code previously in implementation files into header files, but I'm curious where we draw the line for ["very small inline member function implementations"](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/clr-jit-coding-conventions.md#9-function-structure) in the coding convention -- just one-line methods? Small methods with no function calls to avoid size increases from inlining?

amanasifkhalid:
> If we want to do that we should wait until we have native PGO enabled again and then do one-off TP collection in a PGO build with and without the change.

Good idea, I'll wait for #91297 to be merged in. I think this should happen soon, since #95510 was recently opened.

> FWIW, there's some MSVC switches like /d2inlinelogfull:FunctionName and /d2inlinestats that can be used to give more information about inlining decisions.

Thanks for mentioning that. I suspect our exceptionally large callers are exhausting the inlining budget pretty quickly, leaving out the easy inlining candidates. I'll take another look once we have PGO data to see if this is an issue.

amanasifkhalid:
By the way, we have some existing uses of `FORCEINLINE` (and `__forceinline`) in the codebase that might be worth revisiting with the updated PGO data. If the new PGO data negates the need for `FORCEINLINE`, then maybe I can turn this into a cleanup PR.

amanasifkhalid:
> However, most users won't be trying to debug a Release build that we've shipped to production and so we really want it to run as quickly as possible there.

I agree with this sentiment, but could switching to /O2 impact how we support users facing issues in production? In other words, are there situations where we only have a Release build we can debug? Perhaps such cases are rare enough (and already tricky to debug) that they're worth the performance boost.

jkotas:
`/O2` vs `/Ox` is not about debuggability. It is about performance. Our historical experience is that enabling max speed optimizations makes the runtime slower. If you can prove that it is not the case, feel to change it.

amanasifkhalid:
> Our historical experience is that enabling max speed optimizations makes the runtime slower.

Do you know when we last did a comparison? If we haven't tried the new inlining heuristics, I'd be interested in trying them.

tannergooding:
> Our historical experience is that enabling max speed optimizations makes the runtime slower.

That's surprising. I wouldn't expect that given as it looks like we're only really missing `/GF` (Eliminate Duplicate Strings); given that we seem to be manually enabling `/Gy` (Enable Function-Level Linking).

I think it'd be worth testing again and then testing with `/Ob3` and potentially `/GL`, which are relatively newer options.

jkotas:
> Do you know when we last did a comparison?

It has been a while.

> /GL

/GL should be enabled by [CMAKE_INTERPROCEDURAL_OPTIMIZATION](https://github.com/dotnet/runtime/blob/1d8e957ada28a084128ac8323a3abb0b1fd04858/eng/native/configurecompiler.cmake#L864).

EgorBo:
I tested O2 here https://github.com/dotnet/runtime/pull/93336 and didn't detect any TP diffs (for jit)

BruceForstall:
We absolutely should change MSVC compilation switches for Release to use -O2 instead of -Ox. https://github.com/dotnet/runtime/issues/53849. Checked and Release should have the same compiler optimization switches. There are bugs for doing that for x86. I might have done it for non-x86 but testing Release is "hard" -- our CI systems don't do it, and you'd want to do some perf testing on it also.

BruceForstall:
> I disagree. /GL makes sense for Release, but it does not make sense for Checked.

Why do you think this? The goal for Checked is to run as fast as possible but still have DEBUG checking (asserts, etc.) enabled.

Is it because you want Checked builds to compile faster (and maybe be incremental?) for faster dev inner-loop?

One optimization I admit would not make sense in Checked is native PGO, since we'll presumably never collect Checked PGO data.

PR #95485, opened 2023-11-30
Make sure LOHThreshold less than region_size

### LINKED ISSUE #95219: Crash when allocating 4+ MB with high LOH threshold
### Description

I realized that there is no higher limit on `DOTNET_GCLOHThreshold` so I wondered how the GC would handle it if I tried to allocate 4+ MB with a LOH threshold that is bigger than that.

The application hangs with the memory usage continuously increasing. I didn't debug in detail, but it looks like the GC enters in an infinite loop where it tries to allocate a new 4MB basic region, only to find that the allocation doesn't fit in any available region.

### Reproduction Steps

Set `DOTNET_GCLOHThreshold=6400000` then run the code:

```csharp
var alloc = new byte[1024 * 1024 * 4];
```

### Expected behavior

Either of:
 - The array is successfully allocated in gen 0
 - The LOH threshold is capped at ~3 MB and the array is successfully allocated in LOH
 - The CLR refuses to start

### Actual behavior

The GC is stuck in a seemingly infinite loop, creating new regions.

### Regression?

The issue only happens with regions enabled.

### Known Workarounds

Reduce the LOH threshold or disable regions.

### Configuration

_No response_

### Other information

_No response_

cshung:
Fixes https://github.com/dotnet/runtime/issues/95219.

Here are the testing result on CoreCLR after all code review comments are incorporated, the table headers are the byte array sizes (which we need to add 24 to get the real object size), all tests are performed on Windows X64.

|     |         | Config   | GC.GetConfigurationVariable | 85000 - 25 | 85000 - 24 | 4194216 - 25 | 4194216 - 24 |
| --- | ------- | -------- | --------------------------- | ---------- | ---------- |------------- |------------- |
| wks | regions | N/A      | 85000                       | SOH        | LOH        | LOH          | LOH          |
| wks | regions | 1M       | 1048576                     | SOH        | SOH        | LOH          | LOH          |
| wks | regions | 4M       | 4194216                     | SOH        | SOH        | SOH          | LOH          |
| svr | regions | N/A      | 85000                       | SOH        | LOH        | LOH          | LOH          |
| svr | regions | 1M       | 1048576                     | SOH        | SOH        | LOH          | LOH          |
| svr | regions | 4M       | 4194216                     | SOH        | SOH        | SOH          | LOH          |
| wks | segment | N/A      | 85000                       | SOH        | LOH        | LOH          | LOH          |
| wks | segment | 1M       | 1048576                     | SOH        | SOH        | LOH          | LOH          |
| wks | segment | 4M       | 4194216                     | SOH        | SOH        | SOH          | LOH          |
| svr | segment | N/A      | 85000                       | SOH        | LOH        | LOH          | LOH          |
| svr | segment | 1M       | 1048576                     | SOH        | SOH        | LOH          | LOH          |
| svr | segment | 4M       | 4194304                     | SOH        | SOH        | SOH          | SOH          |

Additional tests on segments to see how far we can go there.

|     |         | Config   | GC.GetConfigurationVariable | 500M - 25  |
| --- | ------- | -------- | --------------------------- | ---------- |
| wks | segment | 1G       | 1073741824                  | OOM        |
| svr | segment | 1G       | 1073741824                  | SOH        |

### DIFF
--- a/src/coreclr/gc/gc.cpp
+++ b/src/coreclr/gc/gc.cpp
@@ -48147,6 +48147,9 @@ HRESULT GCHeap::Initialize()
 
     GCConfig::SetHeapCount(static_cast<int64_t>(nhp));
 
+    loh_size_threshold = (size_t)GCConfig::GetLOHThreshold();
+    loh_size_threshold = max (loh_size_threshold, LARGE_OBJECT_SIZE);
+
 #ifdef USE_REGIONS
     gc_heap::enable_special_regions_p = (bool)GCConfig::GetGCEnableSpecialRegions();
     size_t gc_region_size = (size_t)GCConfig::GetGCRegionSize();
@@ -48216,8 +48219,6 @@ HRESULT GCHeap::Initialize()
 #error "FEATURE_STRUCTALIGN is not supported for USE_REGIONS"
 #endif //FEATURE_STRUCTALIGN
 
-    loh_size_threshold = (size_t)GCConfig::GetLOHThreshold();
-    loh_size_threshold = max (loh_size_threshold, LARGE_OBJECT_SIZE);
     loh_size_threshold = min (loh_size_threshold, effective_max_small_object_size);
     GCConfig::SetLOHThreshold(loh_size_threshold);
 
--- a/src/coreclr/gc/gc.cpp
+++ b/src/coreclr/gc/gc.cpp
@@ -48189,21 +48189,16 @@ HRESULT GCHeap::Initialize()
 
     /*
      * Allocation requests less than loh_size_threshold will be allocated on the small object heap.
-     * For regions, we constrainted that objects must stay on a single region, so it is impossible to allocate
-     * an object larger than the available space on a single region.
      *
-     * Note that that available space on the region is not the region size, but a bit smaller.
-     * The adjustment formula is based on these three implementation details:
+     * An object cannot span more than one region and regions in small object heap are of the same size - gc_region_size. 
+     * However, the space available for actual allocations is reduced by the following implementation details -
      *
-     * 1.) heap_segment_mem is set to the new pages + sizeof(aligned_plug_and_gap) in make_heap_segment,
-     *     effectively making the bytes before that unusable for objects.
-     * 2.) a_fit_segment_end_p set pad to Align(min_obj_size, align_const), effectively making some end bytes
-     *     unusable
-     * 3.) a_size_fit_p requires the available space to be larger than the allocated size + Align(min_obj_size, align_const)
-     *     again, effectively making some end bytes unusable.
+     * 1.) heap_segment_mem is set to the new pages + sizeof(aligned_plug_and_gap) in make_heap_segment.
+     * 2.) a_fit_segment_end_p set pad to Align(min_obj_size, align_const).
+     * 3.) a_size_fit_p requires the available space to be >= the allocated size + Align(min_obj_size, align_const)
      *
-     * It is guaranteed that allocation request with exactly this amount or less will either succeed or fail (because we either
-     * cannot get new region or cannot commit memory), but never run into an infinite loop.
+     * It is guaranteed that an allocation request with this amount or less will succeed unless
+     * we cannot commit memory for it.
      */
     int align_const = get_alignment_constant (TRUE);
     size_t effective_max_small_object_size = gc_region_size - sizeof(aligned_plug_and_gap) - Align(min_obj_size, align_const) * 2;
@@ -48224,6 +48219,7 @@ HRESULT GCHeap::Initialize()
 
     gc_heap::min_segment_size_shr = index_of_highest_set_bit (gc_region_size);
 #else
+    // TODO, andrewau, we cannot let user to set an arbitrarily high loh_size_threshold value for segments as well
     gc_heap::min_segment_size_shr = index_of_highest_set_bit (gc_heap::min_segment_size);
 #endif //USE_REGIONS
 
--- a/src/coreclr/gc/gc.cpp
+++ b/src/coreclr/gc/gc.cpp
@@ -51872,9 +51872,9 @@ void GCHeap::DiagScanDependentHandles (handle_scan_fn fn, int gen_number, ScanCo
     GCScan::GcScanDependentHandlesForProfilerAndETW (gen_number, context, fn);
 }
 
-uint64_t GCHeap::GetLOHThreshold()
+size_t GCHeap::GetLOHThreshold()
 {
-    return (uint64_t)loh_size_threshold;
+    return loh_size_threshold;
 }
 
 void GCHeap::DiagGetGCSettings(EtwGCSettingsInfo* etw_settings)
--- a/src/coreclr/gc/gc.cpp
+++ b/src/coreclr/gc/gc.cpp
@@ -14389,9 +14389,6 @@ gc_heap::init_semi_shared()
     loh_compaction_mode = loh_compaction_default;
 #endif //FEATURE_LOH_COMPACTION
 
-    loh_size_threshold = (size_t)GCConfig::GetLOHThreshold();
-    assert (loh_size_threshold >= LARGE_OBJECT_SIZE);
-
 #ifdef BGC_SERVO_TUNING
     memset (bgc_tuning::gen_calc, 0, sizeof (bgc_tuning::gen_calc));
     memset (bgc_tuning::gen_stats, 0, sizeof (bgc_tuning::gen_stats));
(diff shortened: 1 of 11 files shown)


src/coreclr/vm/ceemain.cpp
```diff
@@ -871,6 +871,9 @@ void EEStartupHelper()
         // requires write barriers to have been set up on x86, which happens as part
         // of InitJITHelpers1.
         hr = g_pGCHeap->Initialize();
+
+        g_pConfig->SetGCLOHThreshold((DWORD)GCHeapUtilities::GetGCHeap()->GetLOHThreshold());
```
jkotas, line 875: It looks odd to cache the value on `g_pConfig`. It would look better to cache it in a static variable in gcheaputilities.* or gchelpers.*.

jkotas: This change should be just about LOHThreshold. My point is that LOHThreshold is not read from the config by EEConfig after this change, so there is no point in keeping the cached value for it on EEConfig. After this change, LOHThreshold is yet another value computed by the GC that the VM decided to cache.

The other flags in your list are read from the config by EEConfig so I do not see a problem with keeping those on EEConfig.

BTW: None of the few places where LOHThreshold is used on the VM side look perf sensitive. I think it would ok to just call the GC API in these places, without caching the value ahead of time. It is fine to cache it if you like it better - I do not have a problem with it.

src/coreclr/vm/gchelpers.cpp
```diff
@@ -706,7 +706,7 @@ OBJECTREF AllocateArrayEx(MethodTable *pArrayMT, INT32 *pArgs, DWORD dwNumArgs,
     }
 #endif
 
-    if (totalSize >= g_pConfig->GetGCLOHThreshold())
+    if (totalSize >= (DWORD)GCHeapUtilities::GetGCHeap()->GetLOHThreshold())
```
jkotas, line 709:
```suggestion
    if (totalSize >= LARGE_OBJECT_SIZE && totalSize >= (DWORD)GCHeapUtilities::GetGCHeap()->GetLOHThreshold())
```

src/coreclr/vm/gchelpers.cpp
```diff
@@ -913,7 +913,7 @@ STRINGREF AllocateString( DWORD cchStringLength )
     SetTypeHandleOnThreadForAlloc(TypeHandle(g_pStringClass));
 
     GC_ALLOC_FLAGS flags = GC_ALLOC_NO_FLAGS;
-    if (totalSize >= g_pConfig->GetGCLOHThreshold())
+    if (totalSize >= (DWORD)GCHeapUtilities::GetGCHeap()->GetLOHThreshold())
```
jkotas, line 916:
```suggestion
    if (totalSize >= LARGE_OBJECT_SIZE && totalSize >= (DWORD)GCHeapUtilities::GetGCHeap()->GetLOHThreshold())
```

src/coreclr/vm/gchelpers.cpp
```diff
@@ -1062,7 +1062,7 @@ OBJECTREF AllocateObject(MethodTable *pMT
             flags |= GC_ALLOC_FINALIZE;
 
         DWORD totalSize = pMT->GetBaseSize();
-        if (totalSize >= g_pConfig->GetGCLOHThreshold())
+        if (totalSize >= (DWORD)GCHeapUtilities::GetGCHeap()->GetLOHThreshold())
```
jkotas, line 1065:
```suggestion
        if (totalSize >= LARGE_OBJECT_SIZE && totalSize >= (DWORD)GCHeapUtilities::GetGCHeap()->GetLOHThreshold())
```

jkotas (approved):
It may be worth it to have a `LARGE_OBJECT_SIZE` fast path to avoid the virtual call for small sizes.

src/coreclr/gc/gcinterface.h
```diff
@@ -1013,6 +1013,8 @@ class IGCHeap {
     virtual FinalizerWorkItem* GetExtraWorkForFinalization() PURE_VIRTUAL
 
     virtual uint64_t GetGenerationBudget(int generation) PURE_VIRTUAL
+
+    virtual uint64_t GetLOHThreshold() PURE_VIRTUAL
```
jkotas, line 1017: Would `size_t` make more sense as the return type? This cannot be more than `size_t`. It would save the casts in all places that it is called.

I see that you have matched `GetGenerationBudget`, but the return type of `GetGenerationBudget` looks like an oversight. There are many address-space sized values in gcinterface.h that use `size_t`.

jkotas (approved):
Thanks

src/coreclr/gc/gc.cpp
```diff
@@ -48186,6 +48186,42 @@ HRESULT GCHeap::Initialize()
     {
         return E_OUTOFMEMORY;
     }
+
+    /*
+     * Allocation requests less than loh_size_threshold will be allocated on the small object heap.
+     * For regions, we constrainted that objects must stay on a single region, so it is impossible to allocate
```
Maoni0, line 48192: that's the same for segments too. it's just segments are a lot larger. as we talked about, we should decide if we want to have a different size for segments. I'm ok with having the same value for both so users don't need to concern themselves with whether they are using segments or regions.

src/coreclr/gc/gc.cpp
```diff
@@ -48186,6 +48186,42 @@ HRESULT GCHeap::Initialize()
...
+     * an object larger than the available space on a single region.
+     *
+     * Note that that available space on the region is not the region size, but a bit smaller.
+     * The adjustment formula is based on these three implementation details:
```
Maoni0, line 48196:
```suggestion
     * An object cannot span more than one region and regions in small object heap are of the same size - gc_region_size. 
     *  However, the space available for actual allocations is reduced by the following implementation details -

File: src/coreclr/gc/gc.cpp [hunk 8]
```diff
@@ -48186,6 +48186,42 @@ HRESULT GCHeap::Initialize()
...
+     *
+     * 1.) heap_segment_mem is set to the new pages + sizeof(aligned_plug_and_gap) in make_heap_segment,
+     *     effectively making the bytes before that unusable for objects.
+     * 2.) a_fit_segment_end_p set pad to Align(min_obj_size, align_const), effectively making some end bytes
+     *     unusable
+     * 3.) a_size_fit_p requires the available space to be larger than the allocated size + Align(min_obj_size, align_const)
+     *     again, effectively making some end bytes unusable.
```
Maoni0, line 48203: nit - I think it's fine to delete all the lines that say "effectively making some bytes unusable" since you are talking about how the available size is smaller.

src/coreclr/gc/gc.cpp
```diff
@@ -48186,6 +48186,42 @@ HRESULT GCHeap::Initialize()
...
+     *
+     * It is guaranteed that allocation request with exactly this amount or less will either succeed or fail (because we either
+     * cannot get new region or cannot commit memory), but never run into an infinite loop.
```
Maoni0, line 48206:
```suggestion
     * It is guaranteed that an allocation request with this amount or less will succeed unless
     * we cannot commit memory for it.
```
if you can't a new region you wouldn't even be doing any check to see if the size can fit so we don't need to say "if we cannot get new region". succeeding also implies we cannot run into an infinite loop

src/coreclr/gc/gc.cpp
```diff
@@ -48186,6 +48186,42 @@ HRESULT GCHeap::Initialize()
...
+     * an object larger than the available space on a single region.
+     *
+     * Note that that available space on the region is not the region size, but a bit smaller.
+     * The adjustment formula is based on these three implementation details:
+     *
+     * 1.) heap_segment_mem is set to the new pages + sizeof(aligned_plug_and_gap) in make_heap_segment,
+     *     effectively making the bytes before that unusable for objects.
+     * 2.) a_fit_segment_end_p set pad to Align(min_obj_size, align_const), effectively making some end bytes
+     *     unusable
+     * 3.) a_size_fit_p requires the available space to be larger than the allocated size + Align(min_obj_size, align_const)
```
Maoni0, line 48202:
```suggestion
     * 3.) a_size_fit_p requires the available space to be >= the allocated size + Align(min_obj_size, align_const)
```

src/coreclr/gc/gc.cpp
Maoni0: oh and please remember to update the documentation for this config. thanks!

src/coreclr/vm/gchelpers.cpp
```diff
@@ -410,7 +410,7 @@ OBJECTREF AllocateSzArray(MethodTable* pArrayMT, INT32 cElements, GC_ALLOC_FLAGS
     }
 #endif
 
-    if (totalSize >= g_pConfig->GetGCLOHThreshold())
+    if (totalSize >= LARGE_OBJECT_SIZE && totalSize >= GCHeapUtilities::GetGCHeap()->GetLOHThreshold())
```
jkotas, line 413: Could you please make the same change for NAOT? 

https://github.com/dotnet/runtime/blob/d636522f4ab9615d4358e8bee037ba38ed9e35d1/src/coreclr/nativeaot/Runtime/gcrhenv.cpp#L172

src/coreclr/gc/gc.cpp
cshung: Here (https://github.com/dotnet/docs/pull/38763) is the documentation update PR.

Maoni0 (approved):

PR #95489, opened 2023-12-01
Delete area-owners.json

jeffhandley:
We were using this auxiliary file to add two types of additional metadata for area ownership, with that metadata being incorporated into the https://issuesof.net queries.

1. Definition of "area pods" for the Libraries teams
2. Expanding the GitHub `area-` teams out to include their members

We don't need the issuesof.net queries for area pods anymore, and we can work around the GitHub team member expansion a couple different ways, potentially even incorporating that behavior directly into issuesof.net.


terrajobst:
FYI, issuesof.net now expands GitHub teams during indexing, which is run once a day. Also, I remove the code that loads the JSON file, it's now taken by parsing `area-owners.md` plus expanding the GitHub teams.

terrajobst (approved):

